PR #9931, opened 2024-02-06
Unblock EXTERNAL/EXTERNAL_VPC Cloud KMS key creation.

tdbhacks:
The `external_protection_level_options` field had been nested within the attestation field by accident. See https://github.com/hashicorp/terraform-provider-google/issues/15004 for more information about this change..

NOTE: this is the initial deprecation notice, and we are just adding a duplicate set of fields outside of `attestation`. When the old field is eventually removed, it should not break any CUJs because the current state of things does not allow for successful key creation (malformed requests would be rejected by the server). The old, deprecated field will be removed in the next release.

```release-note:bug
kms: added top-level `external_protection_level_options` field in `google_kms_crypto_key_version` resource
```
```release-note:deprecation
kms: `attestation.external_protection_level_options` has been deprecated in favor of `external_protection_level_options` in `google_kms_crypto_key_version` 
```
```release-note:enhancement
kms: added `crypto_key_backend` field to `google_kms_crypto_key` resource
```

### DIFF
--- a/mmv1/products/kms/CryptoKeyVersion.yaml
+++ b/mmv1/products/kms/CryptoKeyVersion.yaml
@@ -39,6 +39,7 @@ examples:
 custom_code: !ruby/object:Provider::Terraform::CustomCode
   custom_delete: templates/terraform/custom_delete/kms_crypto_key_version.erb
   custom_import: templates/terraform/custom_import/kms_crypto_key_version.go.erb
+  pre_update: templates/terraform/pre_update/kms_crypto_key_version.go.erb
 parameters:
   - !ruby/object:Api::Type::String
     name: 'cryptoKey'
@@ -123,6 +124,9 @@ properties:
         name: 'externalProtectionLevelOptions'
         description: |
           ExternalProtectionLevelOptions stores a group of additional fields for configuring a CryptoKeyVersion that are specific to the EXTERNAL protection level and EXTERNAL_VPC protection levels.
+        deprecation_message: >-
+          `externalProtectionLevelOptions` is being un-nested from the `attestation` field.
+          Please use the top level `externalProtectionLevelOptions` field instead.
         properties:
           - !ruby/object:Api::Type::String
             name: 'externalKeyUri'
@@ -132,3 +136,16 @@ properties:
             name: 'ekmConnectionKeyPath'
             description: |
               The path to the external key material on the EKM when using EkmConnection e.g., "v0/my/key". Set this field instead of externalKeyUri when using an EkmConnection.
+  - !ruby/object:Api::Type::NestedObject
+    name: 'externalProtectionLevelOptions'
+    description: |
+      ExternalProtectionLevelOptions stores a group of additional fields for configuring a CryptoKeyVersion that are specific to the EXTERNAL protection level and EXTERNAL_VPC protection levels.
+    properties:
+      - !ruby/object:Api::Type::String
+        name: 'externalKeyUri'
+        description: |
+          The URI for an external resource that this CryptoKeyVersion represents.
+      - !ruby/object:Api::Type::String
+        name: 'ekmConnectionKeyPath'
+        description: |
+          The path to the external key material on the EKM when using EkmConnection e.g., "v0/my/key". Set this field instead of externalKeyUri when using an EkmConnection.
--- a/mmv1/third_party/terraform/services/kms/resource_kms_crypto_key_test.go
+++ b/mmv1/third_party/terraform/services/kms/resource_kms_crypto_key_test.go
@@ -576,6 +576,12 @@ func TestAccKmsCryptoKeyVersion_externalProtectionLevelOptions(t *testing.T) {
 }
 
 func TestAccKmsCryptoKeyVersion_externalProtectionLevelOptionsVpc(t *testing.T) {
+	// This test relies on manual steps to set up the EkmConnection used for the
+	// CryptoKeyVersion creation, which means we can't spin up a temporary project.
+	// We also can't use bootstrapped keys because that would defeat the purpose of
+	// this key creation test, so we skip this test for VCR to avoid KMS resource
+	// accumulation in the TF test project (since KMS resources can't be deleted).
+	acctest.SkipIfVcr(t)
 	t.Parallel()
 
 	projectId := envvar.GetTestProjectFromEnv()
--- a/mmv1/third_party/terraform/services/kms/resource_kms_crypto_key_test.go
+++ b/mmv1/third_party/terraform/services/kms/resource_kms_crypto_key_test.go
@@ -538,6 +538,79 @@ func TestAccKmsCryptoKeyVersion_patch(t *testing.T) {
 	})
 }
 
+func TestAccKmsCryptoKeyVersion_externalProtectionLevelOptions(t *testing.T) {
+	t.Parallel()
+
+	projectId := fmt.Sprintf("tf-test-%d", acctest.RandInt(t))
+	projectOrg := envvar.GetTestOrgFromEnv(t)
+	projectBillingAccount := envvar.GetTestBillingAccountFromEnv(t)
+	keyRingName := fmt.Sprintf("tf-test-%s", acctest.RandString(t, 10))
+	cryptoKeyName := fmt.Sprintf("tf-test-%s", acctest.RandString(t, 10))
+	keyUri := "data.google_secret_manager_secret_version.key_uri.secret_data"
+	updatedKeyUri := "data.google_secret_manager_secret_version.key_uri_updated.secret_data"
+
+	acctest.VcrTest(t, resource.TestCase{
+		PreCheck:                 func() { acctest.AccTestPreCheck(t) },
+		ProtoV5ProviderFactories: acctest.ProtoV5ProviderFactories(t),
+		Steps: []resource.TestStep{
+			{
+				Config: testGoogleKmsCryptoKeyVersion_externalProtectionLevelOptions(projectId, projectOrg, projectBillingAccount, keyRingName, cryptoKeyName, keyUri),
+			},
+			{
+				ResourceName:            "google_kms_crypto_key_version.crypto_key_version",
+				ImportState:             true,
+				ImportStateVerify:       true,
+				ImportStateVerifyIgnore: []string{"labels", "terraform_labels"},
+			},
+			{
+				Config: testGoogleKmsCryptoKeyVersion_externalProtectionLevelOptions(projectId, projectOrg, projectBillingAccount, keyRingName, cryptoKeyName, updatedKeyUri),
+			},
+			{
+				ResourceName:            "google_kms_crypto_key_version.crypto_key_version",
+				ImportState:             true,
+				ImportStateVerify:       true,
+				ImportStateVerifyIgnore: []string{"labels", "terraform_labels"},
+			},
+		},
+	})
+}
+
+func TestAccKmsCryptoKeyVersion_externalProtectionLevelOptionsVpc(t *testing.T) {
+	t.Parallel()
+
+	projectId := envvar.GetTestProjectFromEnv()
+	keyRingName := fmt.Sprintf("tf-test-%s", acctest.RandString(t, 10))
+	cryptoKeyName := fmt.Sprintf("tf-test-%s", acctest.RandString(t, 10))
+	ekmConnectionName := fmt.Sprintf("tf-test-%s", acctest.RandString(t, 10))
+	keyPath := "data.google_secret_manager_secret_version.key_path.secret_data"
+	updatedKeyPath := "data.google_secret_manager_secret_version.key_path_updated.secret_data"
+
+	acctest.VcrTest(t, resource.TestCase{
+		PreCheck:                 func() { acctest.AccTestPreCheck(t) },
+		ProtoV5ProviderFactories: acctest.ProtoV5ProviderFactories(t),
+		Steps: []resource.TestStep{
+			{
+				Config: testGoogleKmsCryptoKeyVersion_externalProtectionLevelOptionsVpc(projectId, keyRingName, cryptoKeyName, ekmConnectionName, keyPath),
+			},
+			{
+				ResourceName:            "google_kms_crypto_key_version.crypto_key_version",
+				ImportState:             true,
+				ImportStateVerify:       true,
+				ImportStateVerifyIgnore: []string{"labels", "terraform_labels"},
+			},
+			{
+				Config: testGoogleKmsCryptoKeyVersion_externalProtectionLevelOptionsVpc(projectId, keyRingName, cryptoKeyName, ekmConnectionName, updatedKeyPath),
+			},
+			{
+				ResourceName:            "google_kms_crypto_key_version.crypto_key_version",
+				ImportState:             true,
+				ImportStateVerify:       true,
+				ImportStateVerifyIgnore: []string{"labels", "terraform_labels"},
+			},
+		},
+	})
+}
+
 // This test runs in its own project, otherwise the test project would start to get filled
 // with undeletable resources
 func testGoogleKmsCryptoKey_basic(projectId, projectOrg, projectBillingAccount, keyRingName, cryptoKeyName string) string {
@@ -952,4 +1025,148 @@ resource "google_kms_crypto_key_version" "crypto_key_version" {
 	state = "%s"
 }
 `, projectId, projectId, projectOrg, projectBillingAccount, keyRingName, cryptoKeyName, preventDestroy, state)
+}
+
+func testGoogleKmsCryptoKeyVersion_externalProtectionLevelOptions(projectId, projectOrg, projectBillingAccount, keyRingName, cryptoKeyName, keyUri string) string {
+	return fmt.Sprintf(`
+resource "google_project" "acceptance" {
+	name            = "%s"
+	project_id      = "%s"
+	org_id          = "%s"
+	billing_account = "%s"
+}
+
+resource "google_project_service" "acceptance" {
+	project = google_project.acceptance.project_id
+	service = "cloudkms.googleapis.com"
+}
+
+resource "google_kms_key_ring" "key_ring" {
+	project  = google_project_service.acceptance.project
+	name     = "%s"
+	location = "us-central1"
+}
+
+resource "google_kms_crypto_key" "crypto_key" {
+	name     = "%s"
+	key_ring = google_kms_key_ring.key_ring.id
+
+	version_template {
+		algorithm = "EXTERNAL_SYMMETRIC_ENCRYPTION"
+		protection_level = "EXTERNAL"
+	}
+
+	labels = {
+		key = "value"
+	}
+	skip_initial_version_creation = true
+}
+
+data "google_secret_manager_secret_version" "key_uri" {
+  secret = "external-full-key-uri"
+  project = "315636579862"
+}
+data "google_secret_manager_secret_version" "key_uri_updated" {
+  secret = "external-full-key-uri-update-test"
+  project = "315636579862"
+}
+
+resource "google_kms_crypto_key_version" "crypto_key_version" {
+	crypto_key = google_kms_crypto_key.crypto_key.id
+	external_protection_level_options {
+		external_key_uri = %s
+	}
+}
+`, projectId, projectId, projectOrg, projectBillingAccount, keyRingName, cryptoKeyName, keyUri)
+}
+
+// EkmConnection setup and creation is based off of resource_kms_ekm_connection_test.go
+func testGoogleKmsCryptoKeyVersion_externalProtectionLevelOptionsVpc(projectId, keyRingName, cryptoKeyName, ekmConnectionName, keyPath string) string {
+	return fmt.Sprintf(`
+data "google_project" "vpc-project" {
+  project_id = "cloud-ekm-refekm-playground"
+}
+data "google_project" "project" {
+  project_id = "%s"
+}
+
+data "google_secret_manager_secret_version" "raw_der" {
+  secret = "playground-cert"
+  project = "315636579862"
+}
+data "google_secret_manager_secret_version" "hostname" {
+  secret = "external-uri"
+  project = "315636579862"
+}
+data "google_secret_manager_secret_version" "servicedirectoryservice" {
+  secret = "external-servicedirectoryservice"
+  project = "315636579862"
+}
+
+resource "google_project_iam_member" "add_sdviewer" {
+  project = data.google_project.vpc-project.number
+  role    = "roles/servicedirectory.viewer"
+  member  = "serviceAccount:service-${data.google_project.project.number}@gcp-sa-ekms.iam.gserviceaccount.com"
+}
+resource "google_project_iam_member" "add_pscAuthorizedService" {
+  project = data.google_project.vpc-project.number
+  role    = "roles/servicedirectory.pscAuthorizedService"
+  member  = "serviceAccount:service-${data.google_project.project.number}@gcp-sa-ekms.iam.gserviceaccount.com"
+}
+
+resource "google_kms_ekm_connection" "example-ekmconnection" {
+  name            	= "%s"
+  location		= "us-central1"
+  key_management_mode 	= "MANUAL"
+  service_resolvers  	{
+      service_directory_service  = data.google_secret_manager_secret_version.servicedirectoryservice.secret_data
+      hostname 			 = data.google_secret_manager_secret_version.hostname.secret_data
+      server_certificates        {
+      		raw_der	= data.google_secret_manager_secret_version.raw_der.secret_data
+      }
+  }
+  depends_on = [
+        google_project_iam_member.add_pscAuthorizedService,
+        google_project_iam_member.add_sdviewer
+  ]
+}
+
+resource "google_kms_key_ring" "key_ring" {
+	project  = data.google_project.project.project_id
+	name     = "%s"
+	location = "us-central1"
+}
+
+resource "google_kms_crypto_key" "crypto_key" {
+	name     = "%s"
+	key_ring = google_kms_key_ring.key_ring.id
+
+	version_template {
+		algorithm = "EXTERNAL_SYMMETRIC_ENCRYPTION"
+		protection_level = "EXTERNAL_VPC"
+	}
+
+	labels = {
+		key = "value"
+	}
+	crypto_key_backend = google_kms_ekm_connection.example-ekmconnection.id
+	skip_initial_version_creation = true
+}
+
+data "google_secret_manager_secret_version" "key_path" {
+  secret = "external-keypath"
+  project = "315636579862"
+}
+data "google_secret_manager_secret_version" "key_path_updated" {
+  secret = "external-keypath-update-test"
+  project = "315636579862"
+}
+
+resource "google_kms_crypto_key_version" "crypto_key_version" {
+	crypto_key = google_kms_crypto_key.crypto_key.id
+	external_protection_level_options {
+		ekm_connection_key_path = %s
+	}
+}
+`, projectId, ekmConnectionName, keyRingName, cryptoKeyName, keyPath)
 }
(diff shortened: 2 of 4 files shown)


mmv1/third_party/terraform/services/kms/resource_kms_crypto_key_test.go
```diff
@@ -886,3 +912,40 @@ resource "google_kms_crypto_key_version" "crypto_key_version" {
...
+resource "google_kms_key_ring" "key_ring" {
+	project  = google_project_service.acceptance.project
+	name     = "%s"
+	location = "us-central1"
+}
+
+resource "google_kms_crypto_key" "crypto_key" {
+	name     = "%s"
+	key_ring = google_kms_key_ring.key_ring.id
+	labels = {
```
rishamchokshi, line 939: You would need to set the protection type here to external

tdbhacks: yeah, that's what happens when you forget to `git commit --amend` your pending changes :) I'm still debugging the test to find the right incantation to make it work, sorry for the noise

tdbhacks:
Note to reviewer: following internal discussions, this PR will need to be reworked a bit. I will let you know when it's ready for review.

tdbhacks:
Reworked this PR, looking for `/gcbrun` to see if the test works now, as well as an initial review!

tdbhacks:
Updated to undo remaining breaking changes, which will be done in a follow up PR. Looking for /gcbrun again

tdbhacks:
@trodge mind /gcbrun -ing again? Thanks!

c2thorn:
@trodge You can assign me to this if you want, it's similar to another PR that I reviewed.

c2thorn (changes requested):
Please add a release note following https://googlecloudplatform.github.io/magic-modules/contribute/release-notes/

mmv1/products/kms/CryptoKeyVersion.yaml
```diff
@@ -129,3 +132,16 @@ properties:
...
+    name: 'externalProtectionLevelOptions'
+    description: |
+      ExternalProtectionLevelOptions stores a group of additional fields for configuring a CryptoKeyVersion that are specific to the EXTERNAL protection level and EXTERNAL_VPC protection levels.
+    properties:
+      - !ruby/object:Api::Type::String
+        name: 'externalKeyUri'
+        description: |
+          The URI for an external resource that this CryptoKeyVersion represents.
+      - !ruby/object:Api::Type::String
+        name: 'ekmConnectionKeyPath'
```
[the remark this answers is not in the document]
tdbhacks: Added now that EkmConnection support has been merged (hopefully it works, let's see)

mmv1/products/kms/CryptoKeyVersion.yaml
```diff
@@ -129,3 +132,16 @@ properties:
             name: 'ekmConnectionKeyPath'
             description: |
               The path to the external key material on the EKM when using EkmConnection e.g., "v0/my/key". Set this field instead of externalKeyUri when using an EkmConnection.
+  - !ruby/object:Api::Type::NestedObject
+    name: 'externalProtectionLevelOptions'
```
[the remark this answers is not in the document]
tdbhacks: Done, I haven't tested manually because it's kind of tricky given the extensive setup required, let's see if /gcbrun fails

tdbhacks:
@c2thorn Added release notes, tests for the field that is now supported (it required EkmConnection support, which has since been merged), as well as update tests.

@rishamchokshi can you also take a look? I've updated the stored secret used for the key uri because I think the key path was wrong. Thanks!

c2thorn:
tweaked the release notes

tdbhacks:
Ugh, lint error, will fix soon!

tdbhacks:
Squashed commits, fixed lint, and updated secrets used in tests after offline chat with Risham. Looking for /gcbrun as usual :)

tdbhacks:
I was reviewing a different PR when I noticed that Terraform currently doesn't include the `cryptoKeyBackend` field we use for `EXTERNAL_VPC` key creation: https://cloud.google.com/kms/docs/reference/rest/v1/projects.locations.keyRings.cryptoKeys#CryptoKey 🫠

I'll have to add it to this PR if we want comprehensive tests / full key creation support. Thanks for the /gcbrun though, this is turning out to be much trickier than I initially expected.

tdbhacks:
Added cryptoKeyBackend field as mentioned in my previous comment. @c2thorn I also added `update_mask_fields` to the yaml to list both fields in external_protection_level_options as updatable, does that LGTY? I think that's what the latest /gcbrun error is about (first test, the second test failed because of the missing field)

tdbhacks:
re: non-vpc test failure: @c2thorn I think I'm running into the opposite error now, where it's using both fields in the update mask, but we have server-side checks where only one of the fields can be specified in the update request, depending on the type of key (EXTERNAL vs EXTERNAL_VPC).

What's the right incantation to only specify one of the two fields in the update mask? My acceptance test config only sets `external_key_uri` because we're updating an EXTERNAL key, but it looks like it's still passing the other field in the update mask.

Dropping `update_mask_fields` causes the following error
```
Error 400: Cannot update the entire CryptoKeyVersion.external_protection_level_options submessage. Please include individual fields of CryptoKeyVersion.external_protection_level_options in the field mask.
```
and adding `update_mask_fields` causes the latest error
```
Error: Error updating CryptoKeyVersion "projects/tf-test-9006404295424275213/locations/us-central1/keyRings/tf-test-gpot1xa3yc/cryptoKeys/tf-test-4htxzq7p28/cryptoKeyVersions/1": googleapi: Error 400: Updating ekm_connection_key_path is only supported when protection level is EXTERNAL_VPC.
```

re: vpc error: @rishamchokshi I believe the only edit I made to the EkmConnection setup used in your acceptance test is [this line](https://github.com/GoogleCloudPlatform/magic-modules/blob/4c432c6d4d96c2537b7818afdbef4387d8894733/mmv1/third_party/terraform/services/kms/resource_kms_ekm_connection_test.go#L97). In my test, I'm explicitly using the auto-created project, while your acceptance test doesn't set the project ID explicitly. I assume that's the new way the acceptance tests work, while mine is pre-existing and so sadly I have to work with what I have. The project number of "google_project.project" is then used as part of the service account we are granting permissions to, which is currently returning [a NOT_FOUND error](https://ff80fc688db646e85c0974dd849727d93396b627ed0f9913da4a80e-apidata.googleusercontent.com/download/storage/v1/b/ci-vcr-logs/o/beta%2Frefs%2Fheads%2Fauto-pr-9931%2Fartifacts%2Fb6deb36d-d845-4881-a730-2b864f0ef681%2Fbuild-log%2Frecording_build%2FTestAccKmsCryptoKeyVersion_externalProtectionLevelOptionsVpc_recording_test.log?jk=ATxpoHf6HYxo0SJUrMM4Qxtcb75T8dsutaiKfoK7ZO8en2RHt9IB1xFhq1UsisuuMpqtt1qyDVfztCea5a6xiLckHhSwqt9W74xw9U2SW3DSqBkjukPe6INZ95aYCmsLlW3MD3rMU8xlH-HVBqFfQVWeGHFbYbdN-Kswk_922OU5_0Wz2yAPVrR8KnemPoo2dN-F_k62it0crf_5aiqu9Q2eeQcZ6LI4qZJpW6cPWMrN_7oDuPdTCGE_eT1PTjTy8avpKWvsP-lXHI6xsR09g5flWA1pipN564AHOwoGO4bs1XzMi0f9z3tDlgA7LlNc8Wm6jAoKQoH1sUNY2JxP_RpNhsMQ3vsS3z6QnOe8kxoe3pXZzc1O5eaAYoPKSSZ5H-GPUWtZVh1fFTBqWjI9-XuMQSrpEmmN-2BoqTvCqKTNOAYRSBhVKn-RDYxBlU7eaprhErCbsCMBFpKcyOCvpmu6gTdDlEdN81FnNQTELZSyOUitDQPxT1todPDT8C54Tdv1pgHvFl0cTKKBRKMCzY_t5u6TZP3PFI2PEgHFLu-SxgX8MOgGn-568bVYrbqPEvyimrBo6w35FIv9IcjgMsBHvEgpoRPo0I7VvBNAXxW26cicmXT4zguJoUBEsKXTQs-UR4hiLQCb7Xml45wfYDoWLJUINLJ4o9Nt79G0Bp0QdI1BtTlEnKZHTzdWbeKOVdb2R7Ygau0zFBxSE-mpx6qaDIDvpAX_LzBolGdjtYU2yza1a554ZUXrA2YxpjkNmzc1ZpPJRh5PKAuGyWhEyyb-U9BOgVGVepkIZxlGWg37i-SX-RIE7bhu9s60aFiGHAnkIeHGvvaZRytTDl8VO_tsmr3owAOxPVbQjUOaGy4A6UlR0hB6EF-lg4bU6eYcZOaHplVKuSgFlsfS3jU5dgqz6QtHJm4NdiFWbnrMS7Z6emOIy1oHd3tUXs4T5LI6GX5kHYltbqyC8kYSPDHhQNpQ1pnwH2Ys2DHBCpxRobg5IvtnJ-9qGR-_PH8F5t9Zs4rwM8-m4YStE1NSYnBkHci4Xo2TH23YshCQxUL9xFPqYXYaijEOz71z14CXcInxR18Tt26TZurh3I7fxGbWSBu6iLvQY8SxkQkBhk3E_sq8PwHTmB45C3ioKhRQAsZ3gKk6TzqRl2gxPX_MDxrNdoGCNPiSvyn_-dq0dpfCY2ph1arm2BHyl8Xy56JWweRos0c5U5Yome8I9IoRQdQLLgq6carb0j5EBD13bepIOG4eM4zuAgZtA7ipUzyPZzH_M6RGZmBFBfXAHfkhZTgRo2pEzqZC5tlRLnh_Gz4vSTOL6KBbSuKFo02rsHVOpeAe_oxwbkglGSdZOAKPm8tqcNnHsBQFMIXZDJ_wEgMbTUrMWCAiiru1J0VabbVikBU15EsDmJEjin3uMT4GXyYYOismtMC0&isca=1) in my test. Any tips? I'm also wondering which project your "example-ekmconnection" EkmConnection is created in, it's not specified in your acceptance test so I assume it's also being inferred somehow?

Sorry, long and very verbose comment, but this stuff it tricky to explain and debug :)

c2thorn:
> What's the right incantation to only specify one of the two fields in the update mask? My acceptance test config only sets `external_key_uri` because we're updating an EXTERNAL key, but it looks like it's still passing the other field in the update mask.

We don't have more robust controls for the update mask in YAML because API's don't usually have these requirements. To apply new logic, you'll have to write go code here: https://github.com/GoogleCloudPlatform/magic-modules/tree/main/mmv1/templates/terraform/pre_update

and reference that file in the `custom_code` block like here: https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/products/alloydb/Cluster.yaml#L107

This code snippet you write will be injected right before it actually sends the API request. Here you can conditionally set the fields in the update mask object.

I think you'll want to remove `update_mask` in the YAML, so MMv1 won't generate its own logic. You canstart off writing your code by copying the current generated update mask logic: [modular-magician/terraform-provider-google-beta@`auto-pr-9931-old..auto-pr-9931`#diff-0bba3747d0](https://github.com/modular-magician/terraform-provider-google-beta/compare/auto-pr-9931-old..auto-pr-9931#diff-0bba3747d060391f13c15c3f3ef792a366fcd65302b07c0b01263ce28b97785bR334-R347) and editing from there.

c2thorn:
I don't know if that last link works well, but I think you should start off by copying this code block:
```
	if d.HasChange("state") {
		updateMask = append(updateMask, "state")
	}

	if d.HasChange("external_protection_level_options") {
		updateMask = append(updateMask, "externalProtectionLevelOptions.externalKeyUri",
			"externalProtectionLevelOptions.ekmConnectionKeyPath")
	}
	// updateMask is a URL parameter but not present in the schema, so ReplaceVars
	// won't set it
	url, err = transport_tpg.AddQueryParams(url, map[string]string{"updateMask": strings.Join(updateMask, ",")})
	if err != nil {
		return err
	}
```

tdbhacks:
Aah, gotcha, thank you for the tips and samples! I've added the conditional logic, let me know if it looks good to you! I've also updated the yaml accordingly

mmv1/products/kms/CryptoKeyVersion.yaml
```diff
@@ -39,6 +39,7 @@ examples:
 custom_code: !ruby/object:Provider::Terraform::CustomCode
   custom_delete: templates/terraform/custom_delete/kms_crypto_key_version.erb
   custom_import: templates/terraform/custom_import/kms_crypto_key_version.go.erb
+  pre_update: templates/terraform/pre_update/kms_crypto_key_version.go.erb
```
c2thorn, line 42: I _think_ `update_mask: true` will need to be removed as well, but I will run the CI anyway to see

c2thorn:
this is the patch request from the debug logs (id's scrubbed):
```
PATCH /v1/projects/<id>/locations/us-central1/keyRings/<id>/cryptoKeys/<id>/cryptoKeyVersions/1?alt=json&updateMask=externalProtectionLevelOptions.externalKeyUri HTTP/1.1
Host: cloudkms.googleapis.com

{
 "externalProtectionLevelOptions": {
  "externalKeyUri": "v0/longlived/footest"
 },
 "state": "ENABLED"
}
```
anything look wrong here?

tdbhacks:
So I think the custom code worked (woo-hoo!), because only externalKeyUri is being set which is correct. The issue is that the value there is not valid, and that's on me because I thought Update would not have the same level of server-side validation as Create.

I checked and we do have some significant validation in the backend to make sure that the new externalKeyUri is okay, and that's tricky to satisfy for me without knowing what's available in the test EKM we are using for these tests. @rishamchokshi is there another key uri that we can add as a new secret to load in the update test? I'm looking for something that would pass the backend update checks.

Let's get this fixed first, and we can then check back on the VPC issue I mentioned in my earlier comment

mmv1/third_party/terraform/services/kms/resource_kms_crypto_key_test.go
```diff
@@ -953,3 +1028,159 @@ resource "google_kms_crypto_key_version" "crypto_key_version" {
...
+data "google_secret_manager_secret_version" "hostname" {
+  secret = "external-uri"
+  project = "315636579862"
+}
+data "google_secret_manager_secret_version" "servicedirectoryservice" {
+  secret = "external-servicedirectoryservice"
+  project = "315636579862"
+}
+
+resource "google_project_iam_member" "add_sdviewer" {
```
rishamchokshi, line 1126: lets get rid of both of these iam policy resource addition. Looks like this is failing for the tests

tdbhacks:
Hopefully ready for another /gcbrun 🤞

tdbhacks:
Woo-hoo, finally seeing some green! 🎉 will take a look at the VPC error shortly, at first glance it sounds like I forgot to update something in the test

tdbhacks:
Yup, I had forgot to drop the `depends_on` below

tdbhacks:
Okay, the newer error is more interesting: looking at our existing yaml definitions for KMS resources, `name` refers to what we generally define as the `id` of the resource, whereas "resource name" generally means the full resource name (`projects/someproject/locations/....`). @rishamchokshi Do we have a way to get the full resource name from the EkmConnection resource?

In the meantime, I've updated the test to just build it by concatenating the known pieces.

c2thorn:
> Okay, the newer error is more interesting: looking at our existing yaml definitions for KMS resources, `name` refers to what we generally define as the `id` of the resource, whereas "resource name" generally means the full resource name (`projects/someproject/locations/....`). @rishamchokshi Do we have a way to get the full resource name from the EkmConnection resource?
> 
> In the meantime, I've updated the test to just build it by concatenating the known pieces.

I believe `google_kms_ekm_connection.example-ekmconnection.id` will return the long form name

tdbhacks:
> > Okay, the newer error is more interesting: looking at our existing yaml definitions for KMS resources, `name` refers to what we generally define as the `id` of the resource, whereas "resource name" generally means the full resource name (`projects/someproject/locations/....`). @rishamchokshi Do we have a way to get the full resource name from the EkmConnection resource?
> > In the meantime, I've updated the test to just build it by concatenating the known pieces.
> 
> I believe `google_kms_ekm_connection.example-ekmconnection.id` will return the long form name

Got it! I assume that's based on the `id_format`: https://github.com/GoogleCloudPlatform/magic-modules/blob/2089d74fe1ddf00362c76d82dbcebe95fafc9142/mmv1/products/kms/EkmConnection.yaml#L31

Re: the latest error, I double-checked the resource name I'm currently using and I think it's well-formatted (though maybe I need a second pair of eyes to check for any typos). If it is correct, I can still try to use the id field, but I'm wondering if there's something else I need to do IAM permissions-wise. Which user are the acceptance tests being run as? Do I need to grant additional permissions to a specific service account or something along those lines? Thanks!

c2thorn:
> > > Okay, the newer error is more interesting: looking at our existing yaml definitions for KMS resources, `name` refers to what we generally define as the `id` of the resource, whereas "resource name" generally means the full resource name (`projects/someproject/locations/....`). @rishamchokshi Do we have a way to get the full resource name from the EkmConnection resource?
> > > In the meantime, I've updated the test to just build it by concatenating the known pieces.
> > 
> > 
> > I believe `google_kms_ekm_connection.example-ekmconnection.id` will return the long form name
> 
> Got it! I assume that's based on the `id_format`:
> 
> https://github.com/GoogleCloudPlatform/magic-modules/blob/2089d74fe1ddf00362c76d82dbcebe95fafc9142/mmv1/products/kms/EkmConnection.yaml#L31
> 
> Re: the latest error, I double-checked the resource name I'm currently using and I think it's well-formatted (though maybe I need a second pair of eyes to check for any typos). If it is correct, I can still try to use the id field, but I'm wondering if there's something else I need to do IAM permissions-wise. Which user are the acceptance tests being run as? Do I need to grant additional permissions to a specific service account or something along those lines? Thanks!

Responded to you internally.

tdbhacks:
Updated to use `google_kms_ekm_connection.example-ekmconnection.id` as suggested, just in case that's playing a part

tdbhacks:
Ah-ha! So the permission error is gone, but the issue is that the EkmConnection is somehow being created in a different project? I sort of mentioned this a couple days ago: the EkmConnection acceptance test I based my test off of doesn't specify any `project` in the EkmConnection resource creation (as opposed to other sample tests [like this one](https://github.com/GoogleCloudPlatform/magic-modules/blob/89267ec3216d345fd385cab8ed30b7af41a814a3/mmv1/third_party/terraform/services/kms/resource_kms_key_ring_test.go#L73)). @c2thorn is the `project` field always available to be specified? I don't see it in the yaml definition of KeyRing but it's being used in the test I linked above, so I assume it's autogenerated like the `id` field. I'm also curious, what does it default to if not specified?

c2thorn:
> Ah-ha! So the permission error is gone, but the issue is that the EkmConnection is somehow being created in a different project? I sort of mentioned this a couple days ago: the EkmConnection acceptance test I based my test off of doesn't specify any `project` in the EkmConnection resource creation (as opposed to other sample tests [like this one](https://github.com/GoogleCloudPlatform/magic-modules/blob/89267ec3216d345fd385cab8ed30b7af41a814a3/mmv1/third_party/terraform/services/kms/resource_kms_key_ring_test.go#L73)). @c2thorn is the `project` field always available to be specified? I don't see it in the yaml definition of KeyRing but it's being used in the test I linked above, so I assume it's autogenerated like the `id` field. I'm also curious, what does it default to if not specified?

Yes the `project` field is autogenerated by MMv1: https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/kms_ekm_connection#project
`project` will default to the provider block or `GOOGLE_PROJECT` environment variable: https://registry.terraform.io/providers/hashicorp/google/latest/docs/guides/provider_reference#project

In our test environment we have `GOOGLE_PROJECT` set.

tdbhacks:
Thanks for the additional context!

I believe we have two choices here:
1. drop the temporary project creation in my test and rely on the project specified in `GOOGLE_PROJECT` and set in the Terraform test environment. I believe this should work in VCR and will fail with the known test error in nightly and GA builds. However, I think we are implicitly running into something that we wanted to avoid, which is that we are constantly creating new KMS resources that can't be deleted (I think this is already happening for EkmConnections if we check the test project used by Terraform). The other KMS tests create a temporary project to bypass this problem by destroying the whole project.
2. create the EkmConnection (and CryptoKeyVersion) in the temporary project created in the test. I'm not sure this is even viable, if creating EkmConnections requires manual setup steps (because the project would change for each test run).

If the Terraform team is okay with option 1 and the relative accumulation of KMS resources, I don't have a problem with that approach. If the resource accumulation is a problem, we should make both this test and the EkmConnection test merged in #10094 manual. Thoughts?

c2thorn:
> 1. drop the temporary project creation in my test and rely on the project specified in `GOOGLE_PROJECT` and set in the Terraform test environment. I believe this should work in VCR and will fail with the known test error in nightly and GA builds. However, I think we are implicitly running into something that we wanted to avoid, which is that we are constantly creating new KMS resources that can't be deleted (I think this is already happening for EkmConnections if we check the test project used by Terraform). The other KMS tests create a temporary project to bypass this problem by destroying the whole project.

Thanks for the refresher that KMS resources cannot always be deleted. This reminded me that we've run into this problem before and have created bootstrap functions to share KMS resources between tests: https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/third_party/terraform/acctest/bootstrap_test_utils.go#L44

You can see it being used in tests here: https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/third_party/terraform/services/datalossprevention/resource_data_loss_prevention_deidentify_template_test.go#L17

Option 1 should be fine if we utilize these

tdbhacks:
> > 1. drop the temporary project creation in my test and rely on the project specified in `GOOGLE_PROJECT` and set in the Terraform test environment. I believe this should work in VCR and will fail with the known test error in nightly and GA builds. However, I think we are implicitly running into something that we wanted to avoid, which is that we are constantly creating new KMS resources that can't be deleted (I think this is already happening for EkmConnections if we check the test project used by Terraform). The other KMS tests create a temporary project to bypass this problem by destroying the whole project.
> 
> Thanks for the refresher that KMS resources cannot always be deleted. This reminded me that we've run into this problem before and have created bootstrap functions to share KMS resources between tests: https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/third_party/terraform/acctest/bootstrap_test_utils.go#L44
> 
> You can see it being used in tests here: https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/third_party/terraform/services/datalossprevention/resource_data_loss_prevention_deidentify_template_test.go#L17
> 
> Option 1 should be fine if we utilize these

I think that works for tests involving other products using a KMS key (eg. the datalossprevention test you linked to), but it wouldn't work for tests involving the CryptoKey/CryptoKeyVersion creation itself. If we were to bootstrap a KMS key and reuse it, that would defeat the purpose of the key creation test because it would only test key creation once. Every future test run would essentially just use GetCryptoKeyVersion behind the scenes.

Note that I'm not sure I have a very good solution here, I think EkmConnection-related tests are especially tricky because of the required initial setup. @rishamchokshi WDYT? I'm starting to lean towards going for option 1 to verify everything works, and then marking the EkmConnection-related tests as manual to avoid resource accumulation. Maybe we could only run them as part of the lower frequency nightly/GA builds?

rishamchokshi:
> > > 1. drop the temporary project creation in my test and rely on the project specified in `GOOGLE_PROJECT` and set in the Terraform test environment. I believe this should work in VCR and will fail with the known test error in nightly and GA builds. However, I think we are implicitly running into something that we wanted to avoid, which is that we are constantly creating new KMS resources that can't be deleted (I think this is already happening for EkmConnections if we check the test project used by Terraform). The other KMS tests create a temporary project to bypass this problem by destroying the whole project.
> > 
> > 
> > Thanks for the refresher that KMS resources cannot always be deleted. This reminded me that we've run into this problem before and have created bootstrap functions to share KMS resources between tests: https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/third_party/terraform/acctest/bootstrap_test_utils.go#L44
> > You can see it being used in tests here: https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/third_party/terraform/services/datalossprevention/resource_data_loss_prevention_deidentify_template_test.go#L17
> > Option 1 should be fine if we utilize these
> 
> I think that works for tests involving other products using a KMS key (eg. the datalossprevention test you linked to), but it wouldn't work for tests involving the CryptoKey/CryptoKeyVersion creation itself. If we were to bootstrap a KMS key and reuse it, that would defeat the purpose of the key creation test because it would only test key creation once. Every future test run would essentially just use GetCryptoKeyVersion behind the scenes.
> 
> Note that I'm not sure I have a very good solution here, I think EkmConnection-related tests are especially tricky because of the required initial setup. @rishamchokshi WDYT? I'm starting to lean towards going for option 1 to verify everything works, and then marking the EkmConnection-related tests as manual to avoid resource accumulation. Maybe we could only run them as part of the lower frequency nightly/GA builds?

I agree, lets first verify it works completely end to end and go towards a similar path as ekmconnection tests since it depends on creating ekmconnection resource

mmv1/third_party/terraform/services/kms/resource_kms_crypto_key_test.go
```diff
@@ -953,3 +1026,152 @@ resource "google_kms_crypto_key_version" "crypto_key_version" {
...
+	}
+}
+`, projectId, projectId, projectOrg, projectBillingAccount, keyRingName, cryptoKeyName, keyUri)
+}
+
+// EkmConnection setup and creation is based off of resource_kms_ekm_connection_test.go
+func testGoogleKmsCryptoKeyVersion_externalProtectionLevelOptionsVpc(projectId, keyRingName, cryptoKeyName, ekmConnectionName, keyPath string) string {
+	return fmt.Sprintf(`
+resource "google_project_service" "acceptance" {
+	project = data.google_project.project.project_id
```
c2thorn, line 1087: @tdbhacks Running this test disabled our `cloudkms.googleapis.com` on our test project when it was destroyed at the end of the test. This API should always be enabled on our project, so shouldn't be needed.

c2thorn (changes requested):

mmv1/third_party/terraform/services/kms/resource_kms_crypto_key_test.go
tdbhacks: Ack, sorry about that! Removed, this was leftover from when we were still creating a new test project for each run, which required the API to be enabled.

mmv1/third_party/terraform/services/kms/resource_kms_crypto_key_test.go
```diff
@@ -953,3 +1026,147 @@ resource "google_kms_crypto_key_version" "crypto_key_version" {
...
+func testGoogleKmsCryptoKeyVersion_externalProtectionLevelOptions(projectId, projectOrg, projectBillingAccount, keyRingName, cryptoKeyName, keyUri string) string {
+	return fmt.Sprintf(`
+resource "google_project" "acceptance" {
+	name            = "%s"
+	project_id      = "%s"
+	org_id          = "%s"
+	billing_account = "%s"
+}
+
+resource "google_project_service" "acceptance" {
```
c2thorn, line 1039: same thing here, needs to be removed

c2thorn: I was incorrect, disregard.

tdbhacks:
I just squashed all commits, now waiting for a successful VCR run before I add the tag to only run the VPC test in nightly/GA builds

tdbhacks:
Alright, looks like the test was skipped correctly! @c2thorn ready for the final review, I've updated the PR title and description but let me know if you'd like me to tweak something. Thanks!

c2thorn (approved):
LGTM, confirmed test logs stored on internal ticket for other EKM tests.
I'll check the nightly tomorrow to see how it goes with the permissions.

PR #9934, opened 2024-02-07
add deprecation message to `google_notebook_instance`

BBBmau:
Adds a deprecation message to `google_notebook_instance`

Fixes https://github.com/hashicorp/terraform-provider-google/issues/17187

```release-note:none

```

### DIFF
--- a/mmv1/products/notebooks/Instance.yaml
+++ b/mmv1/products/notebooks/Instance.yaml
@@ -15,7 +15,7 @@
 name: 'Instance'
 deprecation_message: >-
   `google_notebook_instance` is deprecated and will be removed in a future major release.
-  Use `google_workbench_instance` instead. 
+  Use `google_workbench_instance` instead.
 description: |
   A Cloud AI Platform Notebook instance.
 
--- a/mmv1/products/notebooks/Instance.yaml
+++ b/mmv1/products/notebooks/Instance.yaml
@@ -13,7 +13,7 @@
 
 --- !ruby/object:Api::Resource
 name: 'Instance'
-# This resource should not be removed until the 2025 major release or later. 
+# This resource should not be removed until the 2025 major release or later.
 # Check instance availability first before fully removing.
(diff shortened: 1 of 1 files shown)


bcreddy-gcp:
When will the resource be removed from Terraform? From notebooks end, we plan to support the resource for the next year, after which it will still be available but not be supported. We'd prefer the resource being in Terraform for as long instances are available.

BBBmau:
@bcreddy-gcp no plans to remove it anytime soon, it would be good practice to mark it as deprecated so that users can shift to using `google_workbench_instance` instead based off of your comment left from another issue.

https://github.com/hashicorp/terraform-provider-google/issues/12515#issuecomment-1927822049

c2thorn:
> When will the resource be removed from Terraform? From notebooks end, we plan to support the resource for the next year, after which it will still be available but not be supported. We'd prefer the resource being in Terraform for as long instances are available.

Can we capture this in a comment right above the deprecation message? @BBBmau 
There is a non-zero chance that as we plan the next major release, someone goes through deprecated resources and marks them for removal. I want to make sure this doesn't get accidentally marked until appropriate

mmv1/products/notebooks/Instance.yaml
```diff
@@ -13,6 +13,9 @@
 
 --- !ruby/object:Api::Resource
 name: 'Instance'
+deprecation_message: >-
```
c2thorn, line 16:
```suggestion
# This resource should not be removed until the 2025 major release or later. 
# Check instance availability first before fully removing.
deprecation_message: >-
```

BBBmau: Thanks for the comment suggestion!

c2thorn: sorry about the whitespace -_-

c2thorn (approved):

PR #9936, opened 2024-02-07
Add custom domain configuration for Looker

nkhasuntsev:
Fixes: https://github.com/hashicorp/terraform-provider-google/issues/17193

```release-note:enhancement
looker: added `custom_domain` field to `google_looker_instance ` resource
```

### DIFF
--- a/mmv1/products/looker/Instance.yaml
+++ b/mmv1/products/looker/Instance.yaml
@@ -434,24 +434,9 @@ properties:
         name: 'domain'
         description: |
           Domain name
-      - !ruby/object:Api::Type::Enum
+      - !ruby/object:Api::Type::String
         name: 'state'
-        output: true
         description: |
-          Status of the custom domain:
-          - CUSTOM_DOMAIN_STATE_UNSPECIFIED: Unspecified state
-          - UNVERIFIED: DNS record is not created
-          - VERIFIED: DNS record is created
-          - MODIFYING: Calling SLM to update
-          - AVAILABLE: ManagedCertificate is ready
-          - UNAVAILABLE: ManagedCertificate is not ready
-          - UNKNOWN: Status is not known
-        values:
-          - :CUSTOM_DOMAIN_STATE_UNSPECIFIED
-          - :UNVERIFIED
-          - :VERIFIED
-          - :MODIFYING
-          - :AVAILABLE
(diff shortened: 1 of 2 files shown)


bschaatsbergen (approved):
LGTM! 🚀

nkhasuntsev:
It works as expected, but since I don't own the domain, it remains in the unverified state, but it's added to the instance. Result of the test in the console:

Screenshot 2024-02-08 at 10 10 02

mmv1/products/looker/Instance.yaml
```diff
@@ -416,3 +424,34 @@ properties:
...
+      Custom domain settings for a Looker instance.
+    properties:
+      - !ruby/object:Api::Type::String
+        name: 'domain'
+        description: |
+          Domain name
+      - !ruby/object:Api::Type::Enum
+        name: 'state'
+        output: true
+        description: |
```
roaks3, line 440: For output-only enums, we generally use a string instead, because validation is not needed. You can also simplify the description to just `Status of the custom domain`, which is nice in case the options change in the future.

roaks3 (commented):
Overall LGTM, just one minor comment.

Regarding the tests, we'll see what the automated tests return. We may need to use a working domain and/or set up infrastructure for this test specifically.

bschaatsbergen:
> Overall LGTM, just one minor comment.
> 
> 
> 
> Regarding the tests, we'll see what the automated tests return. We may need to use a working domain and/or set up infrastructure for this test specifically.

For my information, do we actually have a domain available for resource creation in the project they get ran in?

roaks3:
@bschaatsbergen it would depend on the use case, but I believe we primarily use `hashicorptest.com` for cases where we need to have ownership of the domain in our tests: https://github.com/GoogleCloudPlatform/magic-modules/blob/main/.ci/infra/terraform/README.md?plain=1#L16.

mmv1/products/looker/Instance.yaml
nkhasuntsev: All right, yeah makes sense. I can definitely change it. I kept it the same as the other output enums in the Looker resource, like kmsKeyState. But simplifying it sounds like a good idea.

roaks3 (approved):

PR #9944, opened 2024-02-07
Fix test failures TestAccGKEHubFeatureMembership_gkehubFeaturePolicyController

tonyzhc:
Fix test failures where Policy Controller FeatureMembership resource cannot be deleted if Anthos API is disabled.

Fix issue: https://github.com/hashicorp/terraform-provider-google/issues/10224

```release-note:none
```


trodge (approved):

melinath:
Drive-by comment - we should also probably add `disable_on_destroy = false` to all the other APIs enabled by this test - or else rely on setting them on the test projects & not manage the enablement with resources at all.

PR #9946, opened 2024-02-07
Allow vcenter_network to be set

naitianliu-google:
https://github.com/hashicorp/terraform-provider-google/issues/16821

```release-note:enhancement
pkeonprem: allowed `vcenter_network` to be set in `google_gkeonprem_vmware_cluster`, while the field does not support in-place update
```

### DIFF
--- a/mmv1/products/gkeonprem/VmwareCluster.yaml
+++ b/mmv1/products/gkeonprem/VmwareCluster.yaml
@@ -253,6 +253,7 @@ properties:
           vcenter_network specifies vCenter network name. Inherited from the
           admin cluster.
         immutable: true
+        default_from_api: true
       - !ruby/object:Api::Type::NestedObject
         name: 'hostConfig'
         description:
--- a/mmv1/products/gkeonprem/VmwareCluster.yaml
+++ b/mmv1/products/gkeonprem/VmwareCluster.yaml
@@ -252,7 +252,7 @@ properties:
         description:
           vcenter_network specifies vCenter network name. Inherited from the
           admin cluster.
-        output: true
+        immutable: true
       - !ruby/object:Api::Type::NestedObject
         name: 'hostConfig'
         description:
--- a/mmv1/templates/terraform/examples/gkeonprem_vmware_cluster_f5lb.tf.erb
+++ b/mmv1/templates/terraform/examples/gkeonprem_vmware_cluster_f5lb.tf.erb
@@ -21,6 +21,7 @@ resource "google_gkeonprem_vmware_cluster" "<%= ctx[:primary_resource_id] %>" {
         gateway="test-gateway"
       }
     }
+    vcenter_network = "test-vcenter-network"
   }
   control_plane_node {
      cpus = 4


naitianliu-google:
@shuyama1 Please help take a look when you get a chance. Thanks!

mmv1/templates/terraform/examples/gkeonprem_vmware_cluster_f5lb.tf.erb
```diff
@@ -21,6 +21,7 @@ resource "google_gkeonprem_vmware_cluster" "<%= ctx[:primary_resource_id] %>" {
         gateway="test-gateway"
       }
     }
+    vcenter_network = "test-vcenter-network"
```
shuyama1, line 24: Is this field actually set the network or is this just passing the value? as I don't think network `test-vcenter-network` exists in our project

naitianliu-google:
> Just want to confirm, is this a field that can be specified by user? Looks like the API doc still shows it as an output-only field https://cloud.google.com/anthos/clusters/docs/on-prem-api/reference/rest/v1/projects.locations.vmwareClusters#vmwarenetworkconfig

Thanks for checking! Yes, we need to update the public doc for on-prem API as well. We are drafting the changes in parallel. Here is the CRD docs for the field which shows the field can be specified but immutable. https://cloud.google.com/anthos/clusters/docs/on-prem/latest/how-to/user-cluster-configuration-file#network-vcenter-networkname-field

mmv1/templates/terraform/examples/gkeonprem_vmware_cluster_f5lb.tf.erb
naitianliu-google: Thanks for checking. This is just passing the value, the test would not actually set the network. @shuyama1

naitianliu-google:
@shuyama1 could you please help take another look at this PR when you get a chance? Thanks!

shuyama1:
@naitianliu-google Sorry, I didn't realize it's waiting on my review since it's not showing up on my review list. Taking a review now. Thanks for the ping.

naitianliu-google:
> @naitianliu-google Sorry, I didn't realize it's waiting on my review since it's not showing up on my review list. Taking a review now. Thanks for the ping.

@shuyama1 Thanks for taking time to review this PR. Please let me know if anything else needs to be changed.

shuyama1 (changes requested):

mmv1/products/gkeonprem/VmwareCluster.yaml
```diff
@@ -252,7 +252,7 @@ properties:
         description:
           vcenter_network specifies vCenter network name. Inherited from the
           admin cluster.
-        output: true
+        immutable: true
```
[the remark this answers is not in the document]
naitianliu-google: Thanks for the guidance! Updated accordingly. @shuyama1

shuyama1 (approved):

PR #9948, opened 2024-02-08
[#16607] Fix: support iamMember (i.e. for WIF ids) in big query datasets

LucaPrete:
Fix: allow users to set permissions for `principal`/`principalSets` (`iamMember`) in BigQuery datasets.

Fixes [github.com/hashicorp/terraform-provider-google/issues/16607](https://github.com/hashicorp/terraform-provider-google/issues/16607)

```release-note:bug
bigquery: allowed users to set permissions for `principal`/`principalSets` (`iamMember`) in `google_bigquery_dataset_iam_member`.
```

### DIFF
--- a/mmv1/third_party/terraform/services/bigquery/iam_bigquery_dataset.go
+++ b/mmv1/third_party/terraform/services/bigquery/iam_bigquery_dataset.go
@@ -228,8 +228,6 @@ func iamMemberToAccess(member string) (string, string, error) {
 			return "userByEmail", pieces[1], nil
 		case "serviceAccount":
 			return "userByEmail", pieces[1], nil
-		default:
-			return "", "", fmt.Errorf("Failed to parse BigQuery Dataset IAM member type: %s", member)
 		}
 	}
 	if member == "projectOwners" || member == "projectReaders" || member == "projectWriters" || member == "allAuthenticatedUsers" {
--- a/mmv1/third_party/terraform/services/bigquery/iam_bigquery_dataset.go
+++ b/mmv1/third_party/terraform/services/bigquery/iam_bigquery_dataset.go
@@ -222,6 +222,8 @@ func iamMemberToAccess(member string) (string, string, error) {
 			return "groupByEmail", pieces[1], nil
 		case "domain":
 			return "domain", pieces[1], nil
+		case "iamMember":
+			return "iamMember", pieces[1], nil
 		case "user":
 			return "userByEmail", pieces[1], nil
 		case "serviceAccount":
@@ -234,7 +236,7 @@ func iamMemberToAccess(member string) (string, string, error) {
 		// These are special BigQuery Dataset permissions
 		return "specialGroup", member, nil
 	}
-	return "iamMember", member, nil
+	return "", "", fmt.Errorf("Failed to parse BigQuery Dataset IAM member type: %s", member)
 }
 
 func accessToIamMember(access map[string]interface{}) (string, error) {
--- a/mmv1/third_party/terraform/services/bigquery/iam_bigquery_dataset.go
+++ b/mmv1/third_party/terraform/services/bigquery/iam_bigquery_dataset.go
@@ -214,7 +214,6 @@ func iamMemberToAccess(member string) (string, string, error) {
 	if strings.HasPrefix(member, "deleted:") {
 		return "", "", fmt.Errorf("BigQuery Dataset IAM member is deleted: %s", member)
 	}
-
 	pieces := strings.SplitN(member, ":", 2)
 	if len(pieces) > 1 {
 		switch pieces[0] {
@@ -249,7 +248,7 @@ func accessToIamMember(access map[string]interface{}) (string, error) {
 		return member.(string), nil
 	}
 	if member, ok := access["iamMember"]; ok {
-		return member.(string), nil
+		return fmt.Sprintf("iamMember:%s", member.(string)), nil
 	}
 	if _, ok := access["view"]; ok {
 		// view does not map to an IAM member, use access instead
--- a/mmv1/third_party/terraform/services/bigquery/resource_bigquery_dataset_iam_member_test.go
+++ b/mmv1/third_party/terraform/services/bigquery/resource_bigquery_dataset_iam_member_test.go
@@ -9,7 +9,7 @@ import (
 	"github.com/hashicorp/terraform-provider-google/google/envvar"
 )
 
-func TestAccBigqueryDatasetIamMember_basic(t *testing.T) {
+func TestAccBigqueryDatasetIamMember_serviceAccount(t *testing.T) {
 	t.Parallel()
 
 	datasetID := fmt.Sprintf("tf_test_%s", acctest.RandString(t, 10))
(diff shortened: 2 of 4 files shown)


LucaPrete:
Please wait to merge this. Adding tests and fixing a few things.

mmv1/third_party/terraform/services/bigquery/iam_bigquery_dataset.go
```diff
@@ -222,19 +222,19 @@ func iamMemberToAccess(member string) (string, string, error) {
 			return "groupByEmail", pieces[1], nil
 		case "domain":
 			return "domain", pieces[1], nil
+		case "iamMember":
+			return "iamMember", pieces[1], nil
```
tgmof, line 226: If I may add a small remark:
I think this switch/case mapping was heavily inspired by the V1 prefixes used by IAM:
https://cloud.google.com/iam/docs/principal-identifiers
(I only found these "deleted" information in this doc for instance).
If you're not planning to move to V2 API anytime soon, might make sense to follow the same logic and use principal/principalSet.
Moreover pieces[1] won't work for my use case here since I need the API to be fed with the entire string (not the cut string), so the line 237 was actually correct.
I mean that the API is expecting "iamMember":"principal://iam.googleapis.com/locations/global/workforcePools/POOL_ID/subject/SUBJECT_ATTRIBUTE_VALUE" 
but NOT
"iamMember":"//iam.googleapis.com/locations/global/workforcePools/POOL_ID/subject/SUBJECT_ATTRIBUTE_VALUE"

LucaPrete:
@rileykarson @ScottSuarez this is ready for review on my side but apparently the test here is failing.

I cannot reproduce the error on `TestAccBigqueryDatasetIamMember_iamMember`, while the other error seems completely unrelated with what I did.

Can you please help me to figure out what's going on?

Thank you!

mmv1/third_party/terraform/services/bigquery/iam_bigquery_dataset.go
LucaPrete: @tgmof thanks for the feedback.
I'm not understanding. Can you please show me a use case where this wouldn't work so I can better understand?
Thanks

tgmof: Sure!
I am trying to give rights to a list of Federated Identity users that all have this format:
principal://iam.googleapis.com/locations/global/workforcePools/POOL_ID/subject/SUBJECT_ATTRIBUTE_VALUE

It works via GCP Console UI (where I need to put the full string above) but not via Terraform. It also works via this API:
https://cloud.google.com/bigquery/docs/reference/rest/v2/datasets/patch
image

I understand that, for instance, the API expects "groupByEmail": "some-gcp-group-email" without the leading "group:" but for Federated Identity iamMember it's different, the API needs to full string:
"iamMember": "principal://iam.googleap..." and the principal: should not be cut off into pieces[0] (ignored) and pieces[1] (passed to the API) like it is in your PR proposal:
https://cloud.google.com/iam/docs/principal-identifiers
https://cloud.google.com/bigquery/docs/reference/rest/v2/datasets/patch

LucaPrete: pieces[1] will return the full principal (principal://.....)
Please, notice that we splitN 2. So we're splitting the string at most in two parts. Meaning that in case of "iamMember:principal://...." it will return `pieces[0]` = `iamMember` and `pieces[1]` = `principal://iam....`

tgmof: Got it now, thank you for having taken the time to answer! I wrongly assumed that it needed to be passed without iamMember: to Terraform (because the terraform argument itself it called member, but anyway wrong assumption).

Anyway, I guess your commit will also re-generate the doc here https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/bigquery_dataset_iam#google_bigquery_dataset_iam_member so I won't have to make any assumption after you merge your PR.
Thanks again!

mmv1/third_party/terraform/services/bigquery/iam_bigquery_dataset.go
```diff
@@ -222,19 +222,19 @@ func iamMemberToAccess(member string) (string, string, error) {
...
 		case "user":
 			return "userByEmail", pieces[1], nil
 		case "serviceAccount":
 			return "userByEmail", pieces[1], nil
-		default:
-			return "", "", fmt.Errorf("Failed to parse BigQuery Dataset IAM member type: %s", member)
 		}
 	}
 	if member == "projectOwners" || member == "projectReaders" || member == "projectWriters" || member == "allAuthenticatedUsers" {
```
tgmof, line 233: Opportunistic refactoring/fix opportunity:
1- projectWriters, projectReaders, projectOwners are not documented as possible output https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/bigquery_dataset_iam#google_bigquery_dataset_iam_member 

2- allUsers is documented but doesn't look to be a case handled by the API (?). https://cloud.google.com/bigquery/docs/reference/rest/v2/datasets#Dataset

mmv1/third_party/terraform/services/bigquery/iam_bigquery_dataset.go
LucaPrete: Thanks for double checking @tgmof !
Yes, the PR will also generate the new doc.

c2thorn:
This may be due to an existing bug with `google_bigquery_dataset_iam_member`: https://github.com/hashicorp/terraform-provider-google/issues/17172

I'll try running again to see if it is consistent

LucaPrete:
Hello @c2thorn do we know why this is still failing? We've a p1 open and we'd like to go through the review/merge as soon as we can. Please, let me know if there's anything I can do from my side.

c2thorn:
@LucaPrete I do not know why it is failing, I believe it is an existing bug as seen in https://github.com/hashicorp/terraform-provider-google/issues/17172

Nothing from the test config for your addition points to a bug that I can see, but I'm not going to be able to merge this yet until we solve it. I won't have time to do a deep debugging dive until late tomorrow.

I can run the test in a different environment to see if it environment-specific if you aren't able to recreate it.

c2thorn:
Test failed again. 

@LucaPrete do you have a debug log of a successful run?

mmv1/third_party/terraform/services/bigquery/resource_bigquery_dataset_iam_member_test.go
```diff
@@ -62,3 +90,32 @@ resource "google_service_account" "bqviewer" {
 }
 `, datasetID, saID)
 }
+
+func testAccBigqueryDatasetIamMember_iamMember(datasetID, wifIDs string) string {
+	return fmt.Sprintf(`
+resource "google_bigquery_dataset_iam_member" "access" {
+  dataset_id = google_bigquery_dataset.dataset.dataset_id
+  role       = "roles/viewer"
+  member     = "iamMember:principal://iam.googleapis.com/${google_iam_workload_identity_pool.wif_pool.name}/subject/test"
```
c2thorn, line 99: API is returning
```
    {
      "role": "roles/viewer",
      "iamMember": "principal://iam.googleapis.com/projects/<project-num>/locations/global/workloadIdentityPools/<pool-id>/subject/test"
    }
```

Yet Terraform is looking for the exact string `iamMember:principal://iam.googleapis.com/projects/<project-num>/locations/global/workloadIdentityPools/<pool-id>/subject/test`

```
2024/02/09 19:12:01 [DEBUG]: Looking for member "iamMember:principal://iam.googleapis.com/projects/<project-num>/locations/global/workloadIdentityPools/<pool-id>/subject/test" in found binding
2024/02/09 19:12:01 [DEBUG]: Member "iamMember:principal://iam.googleapis.com/projects/<project-num>/locations/global/workloadIdentityPools/<pool-id>/subject/test" for binding for role "roles/viewer" with condition tpgiamresource.conditionKey{Description:"", Expression:"", Title:""} does not exist in policy of Bigquery Dataset ci-test-project-188019/<dataset-id>, removing from state.
2024-02-09T19:12:01.373Z [ERROR] vertex "google_bigquery_dataset_iam_member.access" error: Provider produced inconsistent result after apply
```

Is this just a matter of changing the test string?

mmv1/third_party/terraform/services/bigquery/iam_bigquery_dataset.go
tgmof: Feel free to ignore this comment if not relevant / I lack of context but I just ping you in case you missed this comment @LucaPrete .

LucaPrete:
@c2thorn it should be all fixed now!

c2thorn (approved):

PR #9957, opened 2024-02-09
Address dependency issues in TestAccFirestoreField_* tests

SarahFrench:
Closes https://github.com/hashicorp/terraform-provider-google/issues/17392

Adding an additional delay in TestAccFirestoreField_* tests to address an 100% error occurring since late Jan:

Screenshot 2024-02-09 at 16 43 41

```
------- Stdout: -------
=== RUN   TestAccFirestoreField_firestoreFieldUpdateAddIndexExample
=== PAUSE TestAccFirestoreField_firestoreFieldUpdateAddIndexExample
=== CONT  TestAccFirestoreField_firestoreFieldUpdateAddIndexExample
    testing_new.go:84: Error running post-test destroy, there may be dangling resources: googleapi: Error 403: Cloud Firestore API has not been used in project tf-test3abifux6ev before or it is disabled. Enable it by visiting https://console.developers.google.com/apis/api/firestore.googleapis.com/overview?project=tf-test3abifux6ev then retry. If you enabled this API recently, wait a few minutes for the action to propagate to our systems and retry.
        Details:
        [
          {
            "@type": "type.googleapis.com/google.rpc.Help",
            "links": [
              {
                "description": "Google developers console API activation",
                "url": "https://console.developers.google.com/apis/api/firestore.googleapis.com/overview?project=tf-test3abifux6ev
              }
            ]
          },
          {
            "@type": "type.googleapis.com/google.rpc.ErrorInfo",
            "domain": "googleapis.com",
            "metadata": {
              "consumer": "projects/tf-test3abifux6ev",
              "service": "firestore.googleapis.com"
            },
            "reason": "SERVICE_DISABLED"
          }
        ]
--- FAIL: TestAccFirestoreField_firestoreFieldUpdateAddIndexExample (1163.75s)
FAIL
```

```release-note:none

```

### DIFF
--- a/mmv1/templates/terraform/custom_delete/firestore_field_delete.go.erb
+++ b/mmv1/templates/terraform/custom_delete/firestore_field_delete.go.erb
@@ -52,7 +52,8 @@ if err != nil {
 	return err
 }
 
+log.Printf("[DEBUG] Finished deleting Field %q", d.Id())
+
 d.SetId("") // Remove the resource from state
 
-log.Printf("[DEBUG] Finished deleting Field %q", d.Id())
 return nil
--- a/mmv1/templates/terraform/custom_delete/firestore_field_delete.go.erb
+++ b/mmv1/templates/terraform/custom_delete/firestore_field_delete.go.erb
@@ -52,5 +52,7 @@ if err != nil {
 	return err
 }
 
+d.SetId("") // Remove the resource from state
+
 log.Printf("[DEBUG] Finished deleting Field %q", d.Id())
 return nil
--- a/mmv1/templates/terraform/custom_delete/firestore_field_delete.go.erb
+++ b/mmv1/templates/terraform/custom_delete/firestore_field_delete.go.erb
@@ -53,7 +53,4 @@ if err != nil {
 }
 
 log.Printf("[DEBUG] Finished deleting Field %q", d.Id())
-
-d.SetId("") // Remove the resource from state
-
 return nil
--- a/mmv1/templates/terraform/custom_check_destroy/firestore_field.go.erb
+++ b/mmv1/templates/terraform/custom_check_destroy/firestore_field.go.erb
@@ -16,7 +16,16 @@ res, err := transport_tpg.SendRequest(transport_tpg.SendRequestOptions{
 	UserAgent: config.UserAgent,
 })
 if err != nil {
-  return err
+	e := err.(*googleapi.Error)
+	if e.Code == 403 && strings.Contains(e.Message, "Cloud Firestore API has not been used in project") {
+		// The acceptance test has provisioned the resources under test in a new project, and the destory check is seeing the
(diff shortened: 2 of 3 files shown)


SarahFrench:
[Running a build against this PR in TeamCity here](https://hashicorp.teamcity.com/buildConfiguration/TerraformProviders_GoogleCloud_GOOGLE_MMUPSTREAMTESTS_GOOGLE_PACKAGE_FIRESTORE/98992?mode=builds&buildTab=overview&hideTestsFromDependencies=false&hideProblemsFromDependencies=false&expandBuildChangesSection=true)

melinath (changes requested):
Looks like the teamcity tests are still failing

SarahFrench:
Ah, I just realised that this permission error is coming at the end of the test during cleanup. I had assumed the problem was enabling the service at the start of the test.

SarahFrench:
I'm re-running this test in TeamCity after each change but the test cleanup keeps deleting the project first. To my understanding, this shouldn't be happening as the project resource has dependant resources (via implicit and explicit dependencies). The destroy order should be the reverse of the create order, so I think I'm overlooking something

SarahFrench:
Order of events in the test is:
- 'Deletion' of firestore field resource - it's not deletable but is altered via PATCH request
- Deletion of DB is skipped: "Firestore database "(default)" deletion_policy is not set to 'DELETE', skipping deletion"
- Firestore service is disabled
- Project is deleted
- Then, there's a random GET request for the firestore field that triggers the 403

The GET request makes sense if the field cannot be deleted and the data base also isn't deleted. Terraform presumably wants to refresh state.

SarahFrench:
Ah the problem appears to be that the `CheckDestroy` for the test isn't well equipped for when the test creates a new separate project

SarahFrench:
This build in TeamCity runs TestAccFirestoreField_firestoreFieldUpdateAddIndexExample against this PR: https://hashicorp.teamcity.com/buildConfiguration/TerraformProviders_GoogleCloud_GOOGLE_MMUPSTREAMTESTS_GOOGLE_PACKAGE_FIRESTORE/108307?mode=builds&buildTab=overview&hideTestsFromDependencies=false&hideProblemsFromDependencies=false

SarahFrench:
Test passed!

mmv1/templates/terraform/custom_delete/firestore_field_delete.go.erb
```diff
@@ -53,4 +53,7 @@ if err != nil {
 }
 
 log.Printf("[DEBUG] Finished deleting Field %q", d.Id())
+
+d.SetId("") // Remove the resource from state
```
SarahFrench, line 57: I'll remove this as, while it might be a legitimate thing to add, this PR isn't the correct place

melinath (approved):
LGTM - I think that this needs to be marked as fixing a few different test failure tickets before merge though

PR #9963, opened 2024-02-10
fix forceNew on resource_manager_tags

edwardmedia:
Fixes https://github.com/hashicorp/terraform-provider-google/issues/17239

```release-note:bug
compute: fixed doing nothing when `resource_manager_tags` is changed on `google_compute_instance_template` and `google_compute_region_instance_template`
```


edwardmedia:
There are two resource_manager_tags in the google_compute_instance_template and google_compute_region_instance_template

While forceNew is correctly set on {root}.disk.resource_manager_tags. It is wrong on {root}.resource_manager_tags

trodge (approved):

PR #9964, opened 2024-02-12
All instances config GA

askubis:
This promotes All Instances Config functionalities to GA . Modifies Compute (Region)InstanceGroupManagers.

Fixes https://github.com/hashicorp/terraform-provider-google/issues/11653

 Searched through the [issue tracker](https://github.com/hashicorp/terraform-provider-google/issues) for an open issue that this either resolves or contributes to, commented on it to claim it, and written "fixes {url}" or "part of {url}" in this PR description. If there were no relevant open issues, I opened one and commented that I would like to work on it (not necessary for very small changes).
 [Generated Terraform](https://github.com/GoogleCloudPlatform/magic-modules#generating-downstream-tools), and ran make test and make lint to ensure it passes unit and linter tests.
 Ensured that all new fields I added that can be set by a user appear in at least one [example](https://github.com/GoogleCloudPlatform/magic-modules/tree/main/templates/terraform/examples) (for generated resources) or [third_party test](https://github.com/GoogleCloudPlatform/magic-modules/tree/main/third_party/terraform/tests) (for handwritten resources or update tests).
 [Ran](https://github.com/hashicorp/terraform-provider-google/blob/main/.github/CONTRIBUTING.md#tests) relevant acceptance tests (If the acceptance tests do not yet pass or you are unable to run them, please let your reviewer know).
 Read the [Release Notes Guide](https://github.com/GoogleCloudPlatform/magic-modules/blob/main/.ci/RELEASE_NOTES_GUIDE.md) before writing my release note below.

```release-note:enhancement
compute: promoted `all_instances_config` field in resources `google_compute_instance_group_manager` and `google_compute_region_instance_group_manager` to GA
```


mmv1/third_party/terraform/go.mod.erb
```diff
@@ -26,9 +26,9 @@ require (
 	github.com/sirupsen/logrus v1.8.1
 	golang.org/x/net v0.20.0
 	golang.org/x/oauth2 v0.16.0
-	google.golang.org/api v0.156.0
```
roaks3, line 29: Do you need these version bumps? It seems like they should not be needed considering this is already released under the beta version. If we don't need them, they should probably be removed from the PR.

roaks3 (approved):
LGTM, although I think the dependency bumps should be removed if possible.

Also, it would be good to run the tests in the GA version locally if you can (our CI checks only run tests for the beta version, which doesn't change for a promotion).

mmv1/third_party/terraform/go.mod.erb
askubis: I definitely need them. I had the code ready a long time ago, but the tests weren't passing, as it didn't build.

wrt. Tests - I did run them, they pass with no issues

roaks3: Ok, so just to clarify, you're saying this behavior is not working right now in beta, and the dependency bumps will fix that for beta (and now GA as well)?

roaks3:
Thanks for running tests locally. I'm going to run the checks as well now, but don't expect them to flag anything.

roaks3:
@askubis this should be good to go, but it looks like the libraries were updated in another PR to a more recent version. So to merge, we'll need to resolve those conflicts (sorry this took a bit longer to approve and allowed the merge conflict to take place).

askubis:
Merge/conflict resolving done, thanks!

PR #7661, opened 2023-04-07
Add regional support for certificatemanager resources

Hamzawy63:
- Add a url-only location property that is optional and defaults to "global"
- Add a new example that creates a regional self-managed certificate 

Terraform Issue: https://github.com/hashicorp/terraform-provider-google/issues/13090

If this PR is for Terraform, I acknowledge that I have:

- [X] Searched through the [issue tracker](https://github.com/hashicorp/terraform-provider-google/issues) for an open issue that this either resolves or contributes to, commented on it to claim it, and written "fixes {url}" or "part of {url}" in this PR description. If there were no relevant open issues, I opened one and commented that I would like to work on it (not necessary for very small changes).
- [X] Ensured that all new fields I added that can be set by a user appear in at least one [example](https://github.com/GoogleCloudPlatform/magic-modules/tree/main/mmv1/templates/terraform/examples) (for generated resources) or [third_party test](https://github.com/GoogleCloudPlatform/magic-modules/tree/main/mmv1/third_party/terraform/tests) (for handwritten resources or update tests).
- [X] [Generated Terraform providers](https://github.com/GoogleCloudPlatform/magic-modules/blob/main/docs/content/docs/getting-started/generate-providers.md), and ran [`make test` and `make lint`](https://googlecloudplatform.github.io/magic-modules/docs/getting-started/run-provider-tests/#run-unit-tests) in the generated providers to ensure it passes unit and linter tests.
- [X] [Ran](https://github.com/GoogleCloudPlatform/magic-modules/blob/main/docs/content/docs/getting-started/run-provider-tests.md) relevant acceptance tests using my own Google Cloud project and credentials (If the acceptance tests do not yet pass or you are unable to run them, please let your reviewer know).
- [X] Read the [Release Notes Guide](https://github.com/GoogleCloudPlatform/magic-modules/blob/main/.ci/RELEASE_NOTES_GUIDE.md) before writing my release note below.

**Release Note Template for Downstream PRs (will be copied)**

```release-note:enhancement
certificatemanager: added `location` field to `certificatemanager` resources
```

PR #9968, opened 2024-02-12
Add location field in DNS authorization resource.

Hamzawy63:
Add a new field `location` in `DnsAuthorization` resource in certificate managers.  The field will default to global in case it's unset. 

This is pretty much similar to what happened in https://github.com/GoogleCloudPlatform/magic-modules/pull/7825, I used this PR as a reference to make a state_migration to avoid breaking changes. 

The resource does not yet support non global instances, but once it does, we will add a new example with a different location. 

```release-note:enhancement
certificatemanager: added `location` field to `google_certificate_manager_dns_authorization` resource

```

### DIFF
--- a/mmv1/templates/terraform/state_migrations/certificate_manager_dns_authorization.go.erb
+++ b/mmv1/templates/terraform/state_migrations/certificate_manager_dns_authorization.go.erb
@@ -1,0 +1,91 @@
+func ResourceCertificateManagerDnsAuthorizationUpgradeV0(_ context.Context, rawState map[string]interface{}, meta interface{}) (map[string]interface{}, error) {
+	log.Printf("[DEBUG] Attributes before migration: %#v", rawState)
+	// Version 0 didn't support location. Default it to global.
+	rawState["location"] = "global"
+	log.Printf("[DEBUG] Attributes after migration: %#v", rawState)
+	return rawState, nil
+}
+
+func resourceCertificateManagerDnsAuthorizationResourceV0() *schema.Resource {
+    return &schema.Resource{
+        Schema: map[string]*schema.Schema{
+			"domain": {
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+				Description: `A domain which is being authorized. A DnsAuthorization resource covers a
+single domain and its wildcard, e.g. authorization for "example.com" can
+be used to issue certificates for "example.com" and "*.example.com".`,
+			},
+			"name": {
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+				Description: `Name of the resource; provided by the client when the resource is created.
+The name must be 1-64 characters long, and match the regular expression [a-zA-Z][a-zA-Z0-9_-]* which means the first character must be a letter,
+and all following characters must be a dash, underscore, letter or digit.`,
+			},
+			"description": {
+				Type:        schema.TypeString,
+				Optional:    true,
+				Description: `A human-readable description of the resource.`,
+			},
+			"labels": {
+				Type:     schema.TypeMap,
+				Optional: true,
+				Description: `Set of label tags associated with the DNS Authorization resource.
+
+**Note**: This field is non-authoritative, and will only manage the labels present in your configuration.
+Please refer to the field 'effective_labels' for all of the labels present on the resource.`,
+				Elem: &schema.Schema{Type: schema.TypeString},
+			},
+			"dns_resource_record": {
+				Type:     schema.TypeList,
+				Computed: true,
+				Description: `The structure describing the DNS Resource Record that needs to be added
+to DNS configuration for the authorization to be usable by
+certificate.`,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"data": {
+							Type:        schema.TypeString,
+							Computed:    true,
+							Description: `Data of the DNS Resource Record.`,
+						},
+						"name": {
+							Type:     schema.TypeString,
+							Computed: true,
+							Description: `Fully qualified name of the DNS Resource Record.
+E.g. '_acme-challenge.example.com'.`,
+						},
+						"type": {
+							Type:        schema.TypeString,
+							Computed:    true,
+							Description: `Type of the DNS Resource Record.`,
+						},
+					},
+				},
+			},
+			"effective_labels": {
+				Type:        schema.TypeMap,
+				Computed:    true,
+				Description: `All of labels (key/value pairs) present on the resource in GCP, including the labels configured through Terraform, other clients and services.`,
+				Elem:        &schema.Schema{Type: schema.TypeString},
+			},
+			"terraform_labels": {
+				Type:     schema.TypeMap,
+				Computed: true,
+				Description: `The combination of labels configured directly on the resource
+ and default labels configured on the provider.`,
+				Elem: &schema.Schema{Type: schema.TypeString},
+			},
+			"project": {
+				Type:     schema.TypeString,
+				Optional: true,
+				Computed: true,
+				ForceNew: true,
+			},
+		},    
+    }
+
+}
--- a/mmv1/third_party/terraform/services/certificatemanager/resource_certificate_manager_dns_authorization_upgrade_test.go
+++ b/mmv1/third_party/terraform/services/certificatemanager/resource_certificate_manager_dns_authorization_upgrade_test.go
@@ -1,0 +1,77 @@
+package certificatemanager_test
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"github.com/hashicorp/terraform-provider-google/google/acctest"
+)
+
+// Tests schema version migration by creating a dns authorization with an old version of the provider (5.15.0)
+// and then updating it with the current version the provider.
+func TestAccCertificateManagerDnsAuthorization_migration(t *testing.T) {
+	acctest.SkipIfVcr(t)
+	t.Parallel()
+	name := fmt.Sprintf("tf-test-%d", acctest.RandInt(t))
+
+	oldVersion := map[string]resource.ExternalProvider{
+		"google": {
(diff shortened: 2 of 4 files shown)


mmv1/templates/terraform/state_migrations/certificate_manager_dns_authorization.go.erb
```diff
@@ -0,0 +1,91 @@
+func ResourceCertificateManagerDnsAuthorizationUpgradeV0(_ context.Context, rawState map[string]interface{}, meta interface{}) (map[string]interface{}, error) {
+	log.Printf("[DEBUG] Attributes before migration: %#v", rawState)
```
Hamzawy63, line 2: I noticed that this log is not printed in while running the migration test, when it will be used? 

Also what can I do to mimic a customer that has a configuration in an old version and then upgrades it? I didn't know how to define the resource to be in an older version. 

I had an old journey with a similar change and you suggested some solutions for testing: https://github.com/GoogleCloudPlatform/magic-modules/pull/7661#issuecomment-1503888510, I would appreciate  if you could elaborate more on both of the mentioned solutions.

mmv1/third_party/terraform/services/certificatemanager/resource_certificate_manager_dns_authorization_upgrade_test.go
```diff
@@ -0,0 +1,77 @@
...
+		CheckDestroy: testAccCheckCertificateManagerDnsAuthorizationDestroyProducer(t),
+		Steps: []resource.TestStep{
+			{
+				Config:            dnsAuthorizationResourceConfig(name),
+				ExternalProviders: oldVersion,
+			},
+			{
+				ResourceName:            "google_certificate_manager_dns_authorization.default",
+				ImportState:             true,
+				ImportStateVerifyIgnore: []string{"location"},
```
Hamzawy63, line 40: How to decide if a field should exist in `ImportStateVerifyIgnore` or not?

mmv1/templates/terraform/state_migrations/certificate_manager_dns_authorization.go.erb
SarahFrench: >I noticed that this log is not printed in while running the migration test, when it will be used?

I ran your `TestAccCertificateManagerDnsAuthorization_migration` test locally with the environment variable TF_LOG set to "DEBUG" to make debug logs visible and I also didn't see the log. I believe this shows that the test isn't using the local code from `mynewprovider` - see other comment for suggested changes!

>Also what can I do to mimic a customer that has a configuration in an old version and then upgrades it? I didn't know how to define the resource to be in an older version.

You can achieve this with the `TestAccCertificateManagerDnsAuthorization_migration` test (following the fix in my other comment). In that test what's happening is:

1. the resource is being provisioned with the old version of the provider + the first config
2. test verifies ability to import the resource using the old version of the provider
3. the resource is being updated using the new version of the provider + the second config with updated description
4. test verifies ability to import the resource using the new version of the provider

You could add further test steps if they were needed. For example you could insert a step between 2 and 3 of the steps above where you use the first config with the new provider and check that the plan is empty (the snippet below has `ExpectNonEmptyPlan: false,` implied). This would mimic a user just upgrading their provider version and not changing their config:

```
			{
				Config:            dnsAuthorizationResourceConfig(name),
				ProviderFactories: newVersion,
				PlanOnly:          true,
			},
```

mmv1/third_party/terraform/services/certificatemanager/resource_certificate_manager_dns_authorization_upgrade_test.go
```diff
@@ -0,0 +1,77 @@
...
+	return fmt.Sprintf(`
+	provider "mynewprovider" {}
+	
+	resource "google_certificate_manager_dns_authorization" "default" {
+		name        = "%s"
+		description = "The migrated default dns"
+		domain      = "domain.hashicorptest.com"
+	  }
+	`, name)
+}
```
SarahFrench, line 77: To enable you to see the debug log mentioned in another comment, you either need to change this code to:

```diff
resource "google_certificate_manager_dns_authorization" "default" {
+ 		provider = mynewprovider
		name        = "%s"
		description = "The migrated default dns"
		domain      = "domain.hashicorptest.com"
	  }
```

or remove the `provider "mynewprovider" {}` line and change `newVersion ` to have a `"google` key instead of `"mynewprovider"`

SarahFrench: To follow this up, without the explicit `provider` argument Terraform will use the `google_` prefix of the resource to decide which provider is needed. I'm not sure exactly what is happening, but it may be that the "google" provider from `oldVersion` is still accessible and being used 😱 . If we change `newVersion` to have a "google" key we would be ensuring the expected code is used. Adding `provider = mynewprovider` would solve the problem in a different way because it stops Terraform looking for "google" and instead looks for "mynewprovider".

Hamzawy63: Thanks for the thorough explanation and for testing this locally! I really appreciate the efforts!

I have tried to do as mentioned, but I still can not see the print statement in the debug logs (I made sure that TF_LOG=DEBUG). 
I first tried to remove mynewprovoider at from the example, and replace "mynewprovider" key in newVersion with "google".  The second thing I tried was to embed `provider = mynewprovider` inside the updated example as you suggested. 

Did I miss something?

Hamzawy63: @SarahFrench 
I also tried to run an existing upgrade test:`make testacc TEST=./google/services/cloudbuild TESTARGS='-run=TestAccCloudBuildTrigger_migration'
` and it's not printing the debug logs as well (https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/templates/terraform/state_migrations/cloud_build_trigger.go.erb#L2)

SarahFrench: Do you have TF_LOG=DEBUG set to make those logs visible? If you run `TF_LOG=DEBUG make testacc TEST=./google/services/cloudbuild TESTARGS='-run=TestAccCloudBuildTrigger_migration'` you should be able to see the 'Attributes before migration..." "Attributes after migration..." log entries. I can see that debug log when I run it locally, and it works because that test uses "mynewprovider" and [has that explicitly set as the provider to be used for the resource](https://github.com/hashicorp/terraform-provider-google/blob/0b64b416469014808ba1e27b6de9e0a9af473d47/google/services/cloudbuild/resource_cloudbuild_trigger_upgrade_test.go#L81) in the relevant test step.

Hamzawy63: @SarahFrench 
Yes, I executed `TF_LOG=DEBUG make testacc TEST=./google/services/cloudbuild TESTARGS='-run=TestAccCloudBuildTrigger_migration'` and can not see the log 'Attributes before migration...'

This is the logs I receive: https://paste.googleplex.com/5093583482257408

SarahFrench: I'm afraid I'm not a Googler so I don't have access to that resource, and I don't have a good alternative for privately sharing logs that might contain sensitive info.

I ran the certificatemanager package's acceptance tests in our overnight test environment using [this PR's proposed code changes in the GA provider repo](https://github.com/modular-magician/terraform-provider-google/compare/auto-pr-9968-old..auto-pr-9968) and I can see this in the debug logs for `TestAccCertificateManagerDnsAuthorization_migration`:

```
2024/02/21 14:36:37 [DEBUG] Attributes before migration: map[string]interface {}{"description":"The default dns", "dns_resource_record":[]interface {}{map[string]interface {}{"data":"3ee21668-b128-4d5d-9c72-34e59dd32c04.13.authorize.certificatemanager.goog.", "name":"_acme-challenge.domain.hashicorptest.com.", "type":"CNAME"}}, "domain":"domain.hashicorptest.com", "effective_labels":map[string]interface {}{}, "id":"projects/ci-test-project-188019/locations/global/dnsAuthorizations/tf-test-8492830205563058242", "labels":map[string]interface {}{}, "name":"tf-test-8492830205563058242", "project":"ci-test-project-188019", "terraform_labels":map[string]interface {}{}, "timeouts":interface {}(nil)}
2024/02/21 14:36:37 [DEBUG] Attributes after migration: map[string]interface {}{"description":"The default dns", "dns_resource_record":[]interface {}{map[string]interface {}{"data":"3ee21668-b128-4d5d-9c72-34e59dd32c04.13.authorize.certificatemanager.goog.", "name":"_acme-challenge.domain.hashicorptest.com.", "type":"CNAME"}}, "domain":"domain.hashicorptest.com", "effective_labels":map[string]interface {}{}, "id":"projects/ci-test-project-188019/locations/global/dnsAuthorizations/tf-test-8492830205563058242", "labels":map[string]interface {}{}, "location":"global", "name":"tf-test-8492830205563058242", "project":"ci-test-project-188019", "terraform_labels":map[string]interface {}{}, "timeouts":interface {}(nil)}
```
---

If you have TF_LOG=DEBUG set to see detailed logs, TF_ACC=1 set to enable acceptance tests running (set for you via `make testacc`), and you're using the correct code that includes the new TestAccCertificateManagerDnsAuthorization_migration test I'm not sure what the problem would be. As you seem to be a Googler, it might be worth attending the Google provider team's internal office hours to ask for help.

Hamzawy63: Thanks again for helping testing this! 

From the logs you shared, the `location` field seems to exist after the migration. Should anything else be done from my side?

SarahFrench: Nothing more - thanks for your patience!

SarahFrench (approved):
LGTM and the migration test is demonstrating that state is migrated successfully

PR #9969, opened 2024-02-12
Expose version_id on google_cloudfunctions_function

drcapulet:
This was already in the YAML but the lookup has custom code so it wasn't exposed - updated the type to be an integer since that's what the Go SDK returns, but happy to keep it a string.

```release-note:enhancement
Expose version_id on google_cloudfunctions_function
```


mmv1/third_party/terraform/services/cloudfunctions/resource_cloudfunctions_function.go
```diff
@@ -490,6 +490,11 @@ func ResourceCloudFunctionsFunction() *schema.Resource {
 				Computed:    true,
 				Description: `Describes the current stage of a deployment.`,
 			},
+			"version_id": {
+				Type:        schema.TypeInt,
```
melinath, line 494: It looks like this is a `string (int64)` in the API, so it should be represented as a string in Terraform for consistency. TypeInt can't reliably represent int64 values.

melinath (changes requested):

mmv1/third_party/terraform/services/cloudfunctions/resource_cloudfunctions_function.go
drcapulet: Sure, updated!

melinath (approved):

PR #9945, opened 2024-02-07
Add desired state to workbench instances.

bcreddy-gcp:
**Release Note Template for Downstream PRs (will be copied)**

```release-note:enhancement
workbench: added `desired_state` field to `google_workbench_instance` resource
```

PR #9971, opened 2024-02-12
Make notebooks wait for active state and enable stopping/starting ins…

bcreddy-gcp:
…tances using the desired_state field

```release-note:enhancement

```

### DIFF
--- a/mmv1/templates/terraform/constants/notebooks_instance.go
+++ b/mmv1/templates/terraform/constants/notebooks_instance.go
@@ -33,3 +33,56 @@ func NotebooksInstanceKmsDiffSuppress(_, old, new string, _ *schema.ResourceData
 	}
 	return false
 }
+
+<% unless compiler == "terraformgoogleconversion-codegen" -%>
+// waitForNotebooksInstanceActive waits for an Notebook instance to become "ACTIVE"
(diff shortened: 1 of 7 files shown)


bcreddy-gcp:
Similar to https://github.com/GoogleCloudPlatform/magic-modules/pull/9945

mmv1/templates/terraform/constants/notebooks_instance.go
```diff
@@ -33,3 +33,56 @@ func NotebooksInstanceKmsDiffSuppress(_, old, new string, _ *schema.ResourceData
 	}
 	return false
 }
+
+<% unless compiler == "terraformgoogleconversion-codegen" -%>
```
hao-nan-li, line 37: Any particular reason to wrap this function with <% unless compiler == "terraformgoogleconversion-codegen" -%>?

hao-nan-li:
Could you add a release note? Thanks

bcreddy-gcp:
Added the release note

mmv1/templates/terraform/constants/notebooks_instance.go
bcreddy-gcp: I am actually not familiar with the reason this is required. If this is not present, I had the conversion test fail for workbench PR, so I added it here as well.

hao-nan-li (approved):

PR #9973, opened 2024-02-13
Add support for custom containers for workbench instances

bcreddy-gcp:

```release-note:enhancement
workbench: added `container_image` field to `google_workbench_instance` resource
```

### DIFF
--- a/mmv1/products/workbench/Instance.yaml
+++ b/mmv1/products/workbench/Instance.yaml
@@ -44,6 +44,15 @@ examples:
     vars:
       instance_name: 'workbench-instance'
   - !ruby/object:Provider::Terraform::Examples
+    name: 'workbench_instance_basic_container'
+    primary_resource_id: 'instance'
+    primary_resource_name: "fmt.Sprintf(\"tf-test-workbench-instance%s\",
+      context[\"\
+      random_suffix\"])"
+    region_override: 'us-west1-a'
+    vars:
+      instance_name: 'workbench-instance'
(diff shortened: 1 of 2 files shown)


rileykarson:
> Error: Error creating Instance: Post "https://notebooks.googleapis.com/v2/projects/ci-test-project-188019/locations/us-west1-a/instances?alt=json&instanceId=tf-test-workbench-instancei6uzd1hvcq": net/http: request canceled (Client.Timeout exceeded while awaiting headers)

That error generally indicates an issue with the API- it's a synchronous request timeout where an initial response took longer than 30s to get back.

bcreddy-gcp:
Could you retest this? I believe we had an outage in us-west1-a.

rileykarson:
Sure!

/gcbrun

melinath:
Drive-by /gcbrun

mmv1/products/workbench/Instance.yaml
```diff
@@ -176,6 +185,9 @@ properties:
         name: vmImage
         custom_flatten: templates/terraform/custom_flatten/workbench_instance_vm_image_flatten.go.erb
         default_from_api: true
+        exactly_one_of:
```
rileykarson, line 188: For nested fields this needs to be the full terraform path i.e. `gce_setup.0.field`. I don't think we want to use exactly_one_of though, since that will make the field(s) required and that's backwards incompatible. Instead, we want `conflicts` which will stop them being specified at the same time.

rileykarson (approved):

PR #9978, opened 2024-02-13
App Check DeviceCheck provider

rainshen49:
Add support for Apple's [DeviceCheck](https://firebase.google.com/docs/reference/appcheck/rest/v1/projects.apps.deviceCheckConfig) provider for App Check. 

Part of https://github.com/hashicorp/terraform-provider-google/issues/17095

```release-note:new-resource
`google_firebase_app_check_device_check_config`
```


mmv1/products/firebaseappcheck/DeviceCheckConfig.yaml
```diff
@@ -0,0 +1,113 @@
...
+    required: true
+    immutable: true
+    url_param_only: true
+properties:
+  - !ruby/object:Api::Type::String
+    name: name
+    description: |
+      The relative resource name of the DeviceCheck configuration object
+    output: true
+    pattern: projects/{{project}}/apps/{{app_id}}/deviceCheckConfig
```
melinath, line 87: Even though there are a number of places this is currently set, I don't think this particular property actually does anything. It seems to be legacy. Created https://github.com/hashicorp/terraform-provider-google/issues/17269 to track removing support.

```suggestion
```

mmv1/products/firebaseappcheck/DeviceCheckConfig.yaml
```diff
@@ -0,0 +1,113 @@
...
+  - !ruby/object:Api::Type::String
+    name: tokenTtl
+    description: |
+      Specifies the duration for which App Check tokens exchanged from DeviceCheck artifacts will be valid.
+      If unset, a default value of 1 hour is assumed. Must be between 30 minutes and 7 days, inclusive.
```
melinath, line 92: If this default is expected to remain the same long-term, `default_value` may provide a better UX - see https://googlecloudplatform.github.io/magic-modules/develop/permadiff/#default for details / caveats.

mmv1/products/firebaseappcheck/DeviceCheckConfig.yaml
```diff
@@ -0,0 +1,113 @@
...
+
+      A duration in seconds with up to nine fractional digits, ending with 's'. Example: "3.5s".
+    default_from_api: true
+  - !ruby/object:Api::Type::String
+    name: keyId
+    description: |
+      The key identifier of a private key enabled with DeviceCheck, created in your Apple Developer account.
+    required: true
+  - !ruby/object:Api::Type::String
+    name: privateKey
```
melinath, line 102: this field is listed as input-only in the [API docs](https://firebase.google.com/docs/reference/appcheck/rest/v1/projects.apps.deviceCheckConfig); I would assume that means that it is immutable? If so please add `immutable: true` and make sure the update test doesn't modify this field.

mmv1/templates/terraform/examples/firebase_app_check_device_check_config_minimal.tf.erb
```diff
@@ -0,0 +1,29 @@
+resource "google_firebase_apple_app" "default" {
```
melinath, line 1: Since the only difference is the TTL field, there's not much value (for tests or for docs) in having the minimal config in addition to the full config. You could leave this out if you want.

mmv1/third_party/terraform/services/firebaseappcheck/test-fixtures/private-key.p8
```diff
@@ -0,0 +1,3 @@
+-----BEGIN PRIVATE KEY-----
+NotAValidKey
```
melinath, line 2: Just want to flag that we've run into problems in the past with APIs adding validation for things like private keys, which then causes test failures that have to be addressed - if that's something that might happen here it could make sense to provide a valid private key that is generated specifically for these tests and not used elsewhere.

melinath (changes requested):
a few changes to request

mmv1/products/firebaseappcheck/DeviceCheckConfig.yaml
rainshen49: Done.

mmv1/products/firebaseappcheck/DeviceCheckConfig.yaml
rainshen49: Even though it's input only, it's not immutable. It's a supported use case to update the privateKey. The input only attribute is in the sense that the value won't be returned in the response due to being sensitive. A more thorough explanation can be found in https://google.aip.dev/147

mmv1/templates/terraform/examples/firebase_app_check_device_check_config_minimal.tf.erb
rainshen49: Done.

mmv1/products/firebaseappcheck/DeviceCheckConfig.yaml
melinath: gotcha, thanks for confirming!

mmv1/products/firebaseappcheck/DeviceCheckConfig.yaml
rainshen49: Requesting for a valid key at the moment. I'll update the thread when ready.

mmv1/third_party/terraform/services/firebaseappcheck/test-fixtures/private-key.p8
rainshen49: Requesting for a valid key at the moment. I'll update the thread when ready.

mmv1/products/firebaseappcheck/DeviceCheckConfig.yaml
rainshen49: The default ttl value is expected to change. Since App Check is fairly new, 1 hour is more of an educated guess. As more information becomes available, we may change the default ttl to a more reasonable value.

melinath (changes requested):
removing from my queue pending the private key changes

rainshen49:
Please don't approve/merge yet. Let me fix the same issue as https://github.com/GoogleCloudPlatform/magic-modules/pull/10120

melinath (approved):

PR #9979, opened 2024-02-13
Add ingestion settings to `google_pubsub_topic`

fayssalmartanigcp:
Add ingestion settings to `google_pubsub_topic`.

These are new fields on the Topic proto message published here: https://github.com/googleapis/googleapis/blob/master/google/pubsub/v1/pubsub.proto#L183

```release-note:enhancement
Add ingestion settings to `google_pubsub_topic`.
```


hao-nan-li (commented):
Could you sign the CLA as commented in this PR?
Also could you add a test to cover the added field?

fayssalmartanigcp:
> Could you sign the CLA as commented in this PR? Also could you add a test to cover the added field?

Regarding the test, it unfortunately requires working AWS resources including a Kinesis stream, consumer and role. I have tested them locally and it works, but we won't be able to have automated tests that depend on AWS resources.

PR #9985, opened 2024-02-14
Add ingestion settings to `google_pubsub_topic`

fayssalmartanigcp:
Add ingestion settings to google_pubsub_topic.

These are new fields on the Topic proto message published here: https://github.com/googleapis/googleapis/blob/master/google/pubsub/v1/pubsub.proto#L183

```release-note:enhancement
pubsub: added `ingestion_data_source_settings` field to `google_pubsub_topic` resource
```


roaks3:
Sorry for the delay on this review. On first glance while we wait for tests:
- I don't see these fields in the API documentation. Are they just not documented yet? If they are not released, this may need to be implemented in our internal provider go/terraform-eap
- The other thing that stands out is these fields are not being tested anywhere. The simplest option is to add them to an existing example. If they are mutable values after the resource is created (you have them configured that way currently), you may instead need to use them in an "update test", to demonstrate that the values can be changed.

fayssalmartanigcp:
> Sorry for the delay on this review. On first glance while we wait for tests:
> 
> * I don't see these fields in the API documentation. Are they just not documented yet? If they are not released, this may need to be implemented in our internal provider go/terraform-eap

The fields are released as part of the public proto here https://github.com/googleapis/googleapis/blob/master/google/pubsub/v1/pubsub.proto#L183-L245. Is this what you mean by API documentation or something else?

> * The other thing that stands out is these fields are not being tested anywhere. The simplest option is to add them to an existing example. If they are mutable values after the resource is created (you have them configured that way currently), you may instead need to use them in an "update test", to demonstrate that the values can be changed.

The issue with the fields is that they require real AWS Kinesis/Role resources that need to be set up. I'm not sure if this is a supported capability in the test framework here. I have tested them manually though with private AWS Kinesis/Role and confirmed they work as expected.

roaks3:
Ah ok, I was referring to this API documentation, which doesn't seem to include these fields: https://cloud.google.com/pubsub/docs/reference/rest/v1/projects.topics

For testing, you can see some existing examples in the code. The simplest option is if fake values an be used without causing errors. If a live AWS account is truly needed to use the resource, you should look into go/terraform-team-testing#managing-external-dependencies to see if your team could manage that account and have tests reference it. Otherwise, it is possible we need to make an exception here, but even in that case, the acceptance tests should be written in this PR similar to all other tests so that they can be run in a repeatable way (and we would skip them during automated tests).

fayssalmartanigcp:
> Ah ok, I was referring to this API documentation, which doesn't seem to include these fields: https://cloud.google.com/pubsub/docs/reference/rest/v1/projects.topics

The documentation will be updated soon prior to the launch of this feature.
 
> For testing, you can see some existing examples in the code...

Added Create/Update tests as discussed offline.

mmv1/products/pubsub/Topic.yaml
```diff
@@ -149,3 +154,40 @@ properties:
...
+            required: true
+          - !ruby/object:Api::Type::String
+            name: 'consumerArn'
+            description: |
+              The Kinesis consumer ARN to used for ingestion in
+              Enhanced Fan-Out mode. The consumer must be already
+              created and ready to be used.
+            required: true
+          - !ruby/object:Api::Type::String
+            name: 'awsRoleArn'
```
roaks3, line 180: I noticed you aren't updating this value in the tests. Is it immutable? If so, we should add `immutable: true` to make sure that if a user does want to change this part of their config, it will correctly create a new resource with the updated value.

roaks3 (approved):
LGTM, one minor comment, and waiting on tests to merge

mmv1/products/pubsub/Topic.yaml
fayssalmartanigcp: Fixed the test to update the value. All these ingestion fields are mutable.

roaks3 (approved):

PR #9987, opened 2024-02-14
Fix issues with vcr status and logs

trodge:
Fixes an issue with replaying after recording logs getting improperly uploaded and test state being improperly reported.

```release-note:none

```


c2thorn (approved):

shuyama1:
I could be wrong, but it seems we'll mistaken mark the VCR status as `success` if there're some tests failed in recording but passed ones run with no issues in re-replaying

PR #9988, opened 2024-02-14
Split github tokens

trodge:

```release-note:none

```


.ci/magician/cmd/generate_comment.go
```diff
@@ -76,7 +77,7 @@ var generateCommentCmd = &cobra.Command{
 			fmt.Println("Error creating a runner: ", err)
 			os.Exit(1)
 		}
-		ctlr := source.NewController(filepath.Join("workspace", "go"), "modular-magician", env["GITHUB_TOKEN"], rnr)
+		ctlr := source.NewController(filepath.Join("workspace", "go"), "modular-magician", env["GITHUB_TOKEN_DOWNSTREAMS"], rnr)
```
melinath, line 80: It looks like this probably needs both tokens passed in - which means refactoring the controller to support both tokens (and support only getting one or the other), or refactoring the calling code to make multiple controllers and decide which to use based on the task. I'd lean towards the second option probably.

melinath (changes requested):

.ci/gcb-push-downstream.yml
```diff
@@ -132,22 +129,21 @@ steps:
...
             else
-              git push https://modular-magician:$$GITHUB_TOKEN@github.com/GoogleCloudPlatform/magic-modules $COMMIT_SHA:tgc-sync-$BRANCH_NAME
+              git push https://modular-magician:$$GITHUB_TOKEN_MAGIC_MODULES@github.com/GoogleCloudPlatform/magic-modules $COMMIT_SHA:tgc-sync-$BRANCH_NAME
             fi
 
     # TF-OICS
     - name: 'gcr.io/graphite-docker-images/bash-plus'
       entrypoint: '/workspace/.ci/scripts/bash-plus/downstream-waiter/wait_for_commit.sh'
       id: tf-oics-sync
-      secretEnv: ["GITHUB_TOKEN"]
```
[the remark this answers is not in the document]
trodge: This token doesn't seem to actually be used by this script anymore.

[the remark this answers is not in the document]
melinath: indeed! great!

.ci/magician/cmd/generate_comment.go
trodge: I don't see where the controller needs the magic modules token.

melinath: you're right, it doesn't. I was misunderstanding what the ctlr was used for.

.ci/gcb-push-downstream.yml
```diff
@@ -228,7 +224,7 @@ logsBucket: 'gs://cloudbuild-downstream-builder-logs'
 availableSecrets:
   secretManager:
     - versionName: projects/673497134629/secrets/github-classic--repo-workflow/versions/latest
-      env: GITHUB_TOKEN
+      env: GITHUB_TOKEN_CLASSIC
```
melinath, line 227: `vcr-merge` and `magician-check-vcr-cassettes` in this file still reference GITHUB_TOKEN & need to be updated to use GITHUB_TOKEN_CLASSIC instead. The scripts (vcr_merge.sh and check_cassettes respectively) also need to be updated accordingly.

.ci/magician/github/init.go
```diff
@@ -16,21 +16,18 @@
...
+	classicToken string
+	token        string
 }
 
 func NewClient() *Client {
-	githubToken, ok := os.LookupEnv("GITHUB_TOKEN")
-	if !ok {
-		fmt.Println("Did not provide GITHUB_TOKEN environment variable")
-		os.Exit(1)
+	return &Client{
```
melinath, line 29: Rather than giving the client multiple tokens, we should pass in the token as an argument to `NewClient` - if the caller needs to make calls with different tokens it should create two clients. (It'll also be better to have the GitHub client not have to know about what the env vars are called.)

Having one token per client is a pretty common pattern when making clients like this. Making this change will also make it easier for us to transition to go-github later: https://github.com/google/go-github?tab=readme-ov-file#authentication

.ci/magician/github/set.go
```diff
@@ -119,7 +119,7 @@ func (gh *Client) CreateWorkflowDispatchEvent(workflowFileName string, inputs ma
 
 func (gh *Client) MergePullRequest(owner, repo, prNumber string) error {
 	url := fmt.Sprintf("https://api.github.com/repos/%s/%s/pulls/%s/merge", owner, repo, prNumber)
-	err := utils.RequestCall(url, "PUT", gh.token, nil, map[string]any{
+	err := utils.RequestCall(url, "PUT", gh.classicToken, nil, map[string]any{
```
melinath, line 122:
```suggestion
	err := utils.RequestCall(url, "PUT", gh.token, nil, map[string]any{
```

.ci/magician/vcr/tester.go
```diff
@@ -199,7 +199,7 @@ func (vt *Tester) Run(mode Mode, version provider.Version, testDirs []string) (*
 	}
 	var printedEnv string
 	for ev, val := range env {
-		if ev == "SA_KEY" || ev == "GITHUB_TOKEN" {
+		if ev == "SA_KEY" || ev == "GITHUB_TOKEN_DOWNSTREAMS" || ev == "GITHUB_TOKEN_MAGIC_MODULES" {
```
melinath, line 202: What would you think about this? Possibly more accident-proof.

```suggestion
		if ev == "SA_KEY" || strings.HasPrefix(ev, "GITHUB_TOKEN") {
```

tools/diff-processor/labels/get_issue.go
```diff
@@ -20,7 +20,7 @@ func GetIssue(repository string, id uint64) (labeler.Issue, error) {
 		return issue, fmt.Errorf("Error creating request: %w", err)
 	}
 	req.Header.Add("Accept", "application/vnd.github+json")
-	req.Header.Add("Authorization", "Bearer "+os.Getenv("GITHUB_TOKEN"))
+	req.Header.Add("Authorization", "Bearer "+os.Getenv("GITHUB_TOKEN_MAGIC_MODULES"))
```
melinath, line 23: Ideally this would also be passed in - but it's nested deeply enough (and not using our github client anyway) that this seems like a reasonable compromise 👍

.ci/magician/cmd/check_cassettes.go
```diff
@@ -62,7 +62,7 @@ var checkCassettesCmd = &cobra.Command{
 			os.Exit(1)
 		}
 
-		ctlr := source.NewController(env["GOPATH"], "modular-magician", env["GITHUB_TOKEN"], rnr)
+		ctlr := source.NewController(env["GOPATH"], "modular-magician", env["GITHUB_TOKEN_DOWNSTREAMS"], rnr)
```
melinath, line 65: Here & above

```suggestion
		ctlr := source.NewController(env["GOPATH"], "modular-magician", env["GITHUB_TOKEN_CLASSIC"], rnr)
```

.ci/gcb-vcr-nightly.yml
```diff
@@ -15,8 +15,6 @@ options:
 logsBucket: 'gs://cloudbuild-vcr-nightly-logs'
 availableSecrets:
   secretManager:
-    - versionName: projects/673497134629/secrets/github-magician-token/versions/latest
```
melinath, line 18: 👍 removed this permission.

melinath (changes requested):
looks reasonable - a couple things to change

.ci/magician/cmd/check_cassettes.go
trodge: I think this step can use the fine-grained token since it's only cloning the downstreams.

melinath: sure, you could pass in GITHUB_TOKEN_DOWNSTREAMS instead if the classic token isn't needed - but what's being passed in needs to match what's being referenced.

.ci/scripts/go-plus/tgc-tester-integration/test_tgc_integration.sh
```diff
@@ -12,7 +12,7 @@ github_username=modular-magician
 
 
 new_branch="auto-pr-$pr_number"
-git_remote=https://$github_username:$GITHUB_TOKEN@github.com/$github_username/$gh_repo
+git_remote=https://$github_username:$GITHUB_TOKEN_CLASSIC@github.com/$github_username/$gh_repo
```
melinath, line 15:
```suggestion
git_remote=https://github.com/$github_username/$gh_repo
```

Shouldn't need a token here since we're not pushing code.

.ci/scripts/go-plus/tgc-tester-integration/test_tgc_integration.sh
```diff
@@ -38,7 +38,7 @@ post_body=$( jq -n \
 
 curl \
   -X POST \
-  -u "$github_username:$GITHUB_TOKEN" \
+  -u "$github_username:$GITHUB_TOKEN_CLASSIC" \
```
melinath, line 41: This needs to match what's being passed in.

```suggestion
  -u "$github_username:$GITHUB_TOKEN_MAGIC_MODULES" \
```

.ci/scripts/go-plus/tgc-tester-integration/test_tgc_integration.sh
```diff
@@ -68,7 +68,7 @@ post_body=$( jq -n \
...
```
melinath, line 71:
```suggestion
  -u "$github_username:$GITHUB_TOKEN_MAGIC_MODULES" \
```

melinath (changes requested):

mmv1/third_party/terraform/acctest/bootstrap_iam_test_utils.go
```diff
@@ -10,6 +10,8 @@ import (
 	cloudresourcemanager "google.golang.org/api/cloudresourcemanager/v1"
 )
 
+// Test change
```
melinath, line 13: revert before merging

melinath (approved):
The last thing remaining had been the TGC integration tests and that's updated now, so LGTM!

PR #9989, opened 2024-02-15
fixes: permadiff issue if event trigger region is not specified

kumadee:
As per https://cloud.google.com/functions/docs/reference/rest/v2/projects.locations.functions#eventtrigger, the trigger_region if not provided, defaults to the same region as the function.

fixes https://github.com/hashicorp/terraform-provider-google/issues/17161

```release-note:bug
cloudfunctions2: fixed an issue where not specifying `event_config.trigger_region` in `google_cloudfunctions2_function` resulted in a permanent diff. The field now pulls a default value from the API when unset.
```

### DIFF
--- a/mmv1/products/cloudfunctions2/Function.yaml
+++ b/mmv1/products/cloudfunctions2/Function.yaml
@@ -594,6 +594,7 @@ properties:
           events originating in this region. It can be the same
           region as the function, a different region or multi-region, or the global
           region. If not provided, defaults to the same region as the function.
+        default_from_api: true
       - !ruby/object:Api::Type::String
         name: 'eventType'
         description: 'Required. The type of event to observe.'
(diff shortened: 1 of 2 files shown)


SarahFrench (commented):
Hi @kumadee , thanks for your PR! The change looks fine but it'd be good to update a test to verify the new behaviour of defaulting from the API.

Could you please remove [this line](https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/templates/terraform/examples/cloudfunctions2_basic_gcs.tf.erb#L94) from that example file? That file is used to generate acceptance tests and also used to add that example to the docs for the google_cloudfunctions2_function resource. There are lots of other acceptance tests that explicitly set `trigger_region` but none currently don't set that field, hence asking you to delete that line.

kumadee:
Hi @SarahFrench, Thanks for the hint. I was also wondering which test to update for this change.

SarahFrench (approved):
I confirmed locally that the changes to that acceptance test (via editing mmv1/templates/terraform/examples/cloudfunctions2_basic_gcs.tf.erb) cause the test to fail with the current version of the provider and the test passes after `default_from_api: true` is added. LGTM!

PR #9991, opened 2024-02-15
App Check reCAPTCHA V3 and reCAPTCHA Enterprise providers

rainshen49:
Part of https://github.com/hashicorp/terraform-provider-google/issues/17095

Add support for [reCAPTCHA V3](https://firebase.google.com/docs/reference/appcheck/rest/v1/projects.apps.recaptchaV3Config) and [reCAPTCHA Enterprise](https://firebase.google.com/docs/reference/appcheck/rest/v1/projects.apps.recaptchaEnterpriseConfig) providers

```release-note:new-resource
`google_firebase_app_check_recaptcha_enterprise_config`
`google_firebase_app_check_recaptcha_v3_config `
```


mmv1/products/firebaseappcheck/RecaptchaEnterpriseConfig.yaml
```diff
@@ -0,0 +1,81 @@
...
+---
+!ruby/object:Api::Resource
+name: "RecaptchaEnterpriseConfig"
+base_url: projects/{{project}}/apps/{{app_id}}/recaptchaEnterpriseConfig
+self_link: projects/{{project}}/apps/{{app_id}}/recaptchaEnterpriseConfig
+create_url: projects/{{project}}/apps/{{app_id}}/recaptchaEnterpriseConfig?updateMask=tokenTtl,siteKey
+create_verb: :PATCH
+update_verb: :PATCH
+update_mask: true
+skip_delete: true
```
hao-nan-li, line 23: Any particular reason to set this field to true?

mmv1/products/firebaseappcheck/RecaptchaEnterpriseConfig.yaml
```diff
@@ -0,0 +1,81 @@
...
+      The relative resource name of the reCAPTCHA Enterprise configuration object
+    output: true
+  - !ruby/object:Api::Type::String
+    name: tokenTtl
+    description: |
+      Specifies the duration for which App Check tokens exchanged from reCAPTCHA Enterprise artifacts will be valid.
+      If unset, a default value of 1 hour is assumed. Must be between 30 minutes and 7 days, inclusive.
+
+      A duration in seconds with up to nine fractional digits, ending with 's'. Example: "3.5s".
+    default_from_api: true
```
hao-nan-li, line 74: I see this field as part of the creation URL, why is it also default_from_api?

mmv1/products/firebaseappcheck/RecaptchaV3Config.yaml
```diff
@@ -0,0 +1,87 @@
...
+---
+!ruby/object:Api::Resource
+name: "RecaptchaV3Config"
+base_url: projects/{{project}}/apps/{{app_id}}/recaptchaV3Config
+self_link: projects/{{project}}/apps/{{app_id}}/recaptchaV3Config
+create_url: projects/{{project}}/apps/{{app_id}}/recaptchaV3Config?updateMask=tokenTtl,siteSecret
+create_verb: :PATCH
+update_verb: :PATCH
+update_mask: true
+skip_delete: true
```
hao-nan-li, line 23: Same questions as the above resource.

mmv1/products/firebaseappcheck/RecaptchaEnterpriseConfig.yaml
rainshen49: The underlying API doesn't have a `DELETE` method. https://firebase.google.com/docs/reference/appcheck/rest/v1/projects.apps.recaptchaV3Config

I've considered an empty patch, but after discussing with the App Check experts, we've decided to make deletion a no-op. This is because an empty patch doesn't do what the user think it does. It doesn't make App Check "forget about" the app, but rather makes the config invalid (blank) which will break their app if still in use.

mmv1/products/firebaseappcheck/RecaptchaEnterpriseConfig.yaml
rainshen49: The `updateMask` parameter specifies which field to update. The value of the `tokenTtl` itself is passed in the request body. See https://google.aip.dev/161 for more info about field masks in general. In this case, if the request sends an empty value `tokenTtl`, the backend will assign a default that is currently `3600s`, but may change in the future, thus the `default_from_api`.

mmv1/products/firebaseappcheck/RecaptchaEnterpriseConfig.yaml
hao-nan-li: Just want to make sure, if the API doesn't have a delete method, how shall this resource be deleted at the end of the test?

hao-nan-li: Sounds good! Good to know that this resource won't cause any leak.

hao-nan-li (approved):

PR #9999, opened 2024-02-16
TeamCity: Add validating GHAs that check Kotlin files listing services

BBBmau:
Closes [#17168](https://github.com/hashicorp/terraform-provider-google/issues/17168)

```release-note:none

```


BBBmau:
Some notes:

it seemed to make sense to keep the `diff_check.go` file within the `.teamcity` folder however I understand that it could be moved to a more appropriate place within the `.teamcity` directory.

With the time it takes for the provider to be built and since the provider has to be built regardless for unit-tests, it would be good to have it use the built provider from the unit-tests for it to be used in the teamcity-diff-check. Although, since we don't expect new services to be added we could just ignore this also.

mmv1/third_party/terraform/.teamcity/components/inputs/diff_check.go
```diff
@@ -0,0 +1,91 @@
...
+		return
+	}
+
+	pattern := regexp.MustCompile(`github\.com\/hashicorp\/terraform-provider-(google|google-beta)\/(google|google-beta)\/services\/(?P<service>\w+)`)
+
+	// Template to convert "key: value" to "key=value" by
+	// referencing the values captured by the regex pattern.
+	template := []byte("$service\n")
+
+	googleServices := []byte{}
```
SarahFrench, line 39: Instead of collecting this information as a long slice of bytes, I'd recommend collecting a list of the service names as human readable strings. This would make the tool easier to debug during development and also help with making actionable error messages when a service is identified as missing.

The loop over submatches on lines 42-46 could contain something like:

```
dst := []byte{}
service := pattern.Expand(dst, template, stdout, submatches)
googleServices = append(googleServices, string(service))
```

---

Following that, another idea to think about might be storing the found services in a map where the key is the service name, instead of using a list. This would make it much easier to identify if a service is present or not:

```
// assuming data gets stored as map[string]bool (value in key-value pair not super important)

// from processing services_ga.kt
servicesInTeamCity := map[string]bool {
  "serviceA" : true,
  // serviceB isn't a key because it's missing from the services_ga.kt file used to make the map
  "serviceC" : true,
  ...
}

// from processing stdout of go list command
servicesInGaProvider := map[string]bool {
  "serviceA" : true,
  "serviceB" : true,
  "serviceC" : true,
  ...
}

for service, _ := range servicesInGaProvider {
  val, ok := servicesInTeamCity[service]; !ok {
    fmt.Printf("service %s is in the GA provider but not in services_gs.kt!", service)
  }
}
```

mmv1/third_party/terraform/.teamcity/components/inputs/diff_check.go
```diff
@@ -0,0 +1,91 @@
...
+	if err != nil {
+		fmt.Println(err.Error())
+		return
+	}
+
+	pattern := regexp.MustCompile(`github\.com\/hashicorp\/terraform-provider-(google|google-beta)\/(google|google-beta)\/services\/(?P<service>\w+)`)
+
+	// Template to convert "key: value" to "key=value" by
+	// referencing the values captured by the regex pattern.
+	template := []byte("$service\n")
```
SarahFrench, line 37: Is the `\n` needed here?

mmv1/third_party/terraform/.teamcity/components/inputs/diff_check.go
```diff
@@ -0,0 +1,91 @@
...
+	"fmt"
+	"io"
+	"os"
+	"os/exec"
+	"regexp"
+)
+
+func main() {
+	repo := os.Args
+	services := fmt.Sprintf("../../../%v/services/...", repo[1])
```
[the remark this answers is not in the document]
SarahFrench: From taking a look at this code I think flags to receive the path to the service_x.kt Kotlin file and the path to the services folder in the unzipped artifact containing the TPG/TPGB codebase would be a good idea!

mmv1/third_party/terraform/.teamcity/components/inputs/diff_check.go
```diff
@@ -0,0 +1,91 @@
...
+	fmt.Println(services)
+	cmd := exec.Command("go", "list", services)
+	stdout, err := cmd.Output()
+	if err != nil {
+		fmt.Println(err.Error())
+		return
+	}
+
+	pattern := regexp.MustCompile(`github\.com\/hashicorp\/terraform-provider-(google|google-beta)\/(google|google-beta)\/services\/(?P<service>\w+)`)
```
SarahFrench, line 33: If "google" / "google-beta" is obtained as an argument, or can be inferred from other arguments, that could be used to produce a slightly simpler regex.

SarahFrench (commented):
Thanks for sharing this @BBBmau! It's a great start and I like how you've navigated pulling information out of the Kotlin map!

You asked about where to store this work/the diff_check.go file. I'd suggest moving the file into it's own folder in [the tools folder of magic modules](https://github.com/GoogleCloudPlatform/magic-modules/tree/main/tools). This would mean that the GHA would need to both checkout the MM repo (to get access to the tool) and also open the TPG/TPGB codebase via the `Download built artifacts` step (to be able to run go list). This would also help you navigate [the problem in this test GHA run of yours](https://github.com/BBBmau/magic-modules/actions/runs/7925793270/job/21639590507), which fails because the google-beta provider doesn't include the .teamcity folder.

mmv1/third_party/terraform/.teamcity/components/inputs/diff_check.go
BBBmau: Removed since I found that it's really not necessary for template

tools/teamcity-diff-check/main.go
```diff
@@ -0,0 +1,107 @@
...
+
+	test.Dir = "../../provider/.teamcity"
+	root, _ = test.Output()
+	fmt.Println(string(root))
+
+	test.Dir = "../../provider/.teamcity/components"
+	root, _ = test.Output()
+	fmt.Println(string(root))
+
+	f, err := os.Open(fmt.Sprintf("../../provider/.teamcity/components/inputs/%s", *serviceFile))
```
SarahFrench, line 67: The confusion about the GHA working with GA but not Beta is because this GHA should validate the service files present in the magic-modules repo, not files copied into the downstream provider repos. I think this change is needed, but you'll need to test it:

```suggestion
	f, err := os.Open(fmt.Sprintf("../mmv1/third_party/terraform/.teamcity/components/inputs/%s", *serviceFile))
```

Only the GA version of the provider contains a .teamcity folder, which is why the GHA currently fails when it's attempting to check the services_beta.kt file in the Beta provider code base. Only the GA provider having a .teamcity folder is intentional.

The GHAs need to download the GA and Beta artefacts to perform the `go list` step in your script and determine which service packages are present, but otherwise you shouldn't need to interact with the downstream providers' files.

tools/teamcity-diff-check/main.go
```diff
@@ -0,0 +1,107 @@
...
+	// For each match of the regex in the content.
+	for _, submatches := range pattern.FindAllSubmatchIndex(bs, -1) {
+		service := pattern.Expand(dst, template, bs, submatches)
+		teamcityServices = append(teamcityServices, string(service))
+	}
+
+	if !reflect.DeepEqual(googleServices, teamcityServices) {
+		fmt.Fprintf(os.Stderr, "error: diff in %s\n", *serviceFile)
+		os.Exit(1)
+	}
```
SarahFrench, line 104: Could you please implement a way of telling users which service(s) is missing? Currently, if this GHA identifies there's a diff the user will need to compare a list of ~135 folders to the map of services in the matching Kotlin file to find out what fix is required

tools/teamcity-diff-check/main.go
```diff
@@ -0,0 +1,111 @@
...
+
+	var diff []string
+	for _, s := range tS {
+		if _, found := g[s]; !found {
+			diff = append(diff, s)
+		}
+	}
+
+	return diff
+}
```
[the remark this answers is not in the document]
BBBmau: This is true, the correct way where `services in the provider` is the ground truth would be to compare the services found in teamcity with that of google services (which would be a complete flip in how it's shown in the code).

tools/teamcity-diff-check/main.go
```diff
@@ -0,0 +1,111 @@
...
+
+	servicesPath := fmt.Sprintf("../../provider/%s/services/", *provider)
+	cmd := exec.Command("go", "list", "./...")
+	cmd.Dir = servicesPath
+	stdout, err := cmd.Output()
+	// if err != nil {
+	// 	fmt.Println(string(stdout))
+	// 	fmt.Println(err.Error())
+	// 	return
+	// }
```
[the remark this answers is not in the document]
BBBmau: https://github.com/BBBmau/magic-modules/actions/runs/7992740233/job/21827124224

.github/workflows/teamcity-services-diff-check.yml
```diff
@@ -68,4 +77,6 @@ jobs:
         run: |
           cd tools/teamcity-diff-check
           go run main.go
```
SarahFrench, line 85: Looks like an accidental duplication here:

```suggestion
```

.github/workflows/teamcity-services-diff-check.yml
```diff
@@ -10,6 +10,15 @@ on:
...
       - 'mmv1/third_party/terraform/.teamcity/components/inputs/services_beta.kt'
 jobs:
+  check-pr:
+    runs-on: ubuntu-22.04
+    steps:
+      - name: Check for New Services
+        run: |
+          newServices=$(($(git diff --name-only --diff-filter=A origin/main HEAD | grep "mmv1/products" | wc -l)))
+          if [ "$newServices" != "0" ];then
+          exit 1
```
SarahFrench: Lets pair on this section, but here's a list of things that we should discuss when we pair:

- What is the if condition checking here? Is it correct?
- Adding feedback to the user about the significance of this step, e.g. via an `echo` command
- Ways to make the newServices count more accurate using patterns/regex with `grep`
   -  On Mac, `grep -e "mmv1/products/.*/product.yaml"` would help identify when new services are added
- Use of [outputs](https://docs.github.com/en/actions/using-jobs/defining-outputs-for-jobs) and [conditional job execution](https://docs.github.com/en/actions/using-jobs/using-conditions-to-control-job-execution) to control when the GHA runs (replacing `exit 1`)

.github/workflows/teamcity-services-diff-check.yml
```diff
@@ -0,0 +1,88 @@
...
+      with:
+        repo: 'terraform-provider-google-beta'
+      
+  teamcity-services-diff-check:
+    needs: [terraform-provider-google, terraform-provider-google-beta]
+    runs-on: ubuntu-22.04
+    steps:
+      - uses: actions/checkout@v3
+      
+      - name: Download built artifacts
```
SarahFrench, line 44:
```suggestion
      - name: Download built artifacts - GA provider
```

.github/workflows/teamcity-services-diff-check.yml
```diff
@@ -0,0 +1,88 @@
...
+        with:
+          name: artifact-terraform-provider-google
+          path: artifacts
+      
+      - name: Unzip the artifacts and delete the zip
+        run: |
+          unzip -o artifacts/output.zip -d ./provider
+          rm artifacts/output.zip
+
+      - name: Download built artifacts
```
SarahFrench, line 55:
```suggestion
      - name: Download built artifacts - Beta provider
```

SarahFrench (commented):
Thanks for taking on my feedback! I think we should pair on making this GHA only run fully if new services are added, but in my review comments I left a bullet point list of things we should discuss while pairing. I'll put something in our calendars

.github/workflows/teamcity-services-diff-check.yml
```diff
@@ -0,0 +1,93 @@
...
+permissions: read-all
+
+on:
+  workflow_dispatch:
+  pull_request:
+    paths:
+      - '.github/workflows/teamcity-services-diff-check.yml'
+      - 'mmv1/third_party/terraform/.teamcity/components/inputs/services_ga.kt'
+      - 'mmv1/third_party/terraform/.teamcity/components/inputs/services_beta.kt'
+      - 'mmv1/products/**'
```
SarahFrench, line 11: I'm glad we established that the use of paths here + the `git diff` step will help us identify a new service being added pretty accurately. Though after thinking that through I realised that another edge case is probably when a service is promoted to GA. In that situation services_ga.kt will be mis-matched with the GA provider but this GHA won't have an opportunity to be triggered automatically.

I think a solution to this is adding a [schedule](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#schedule) trigger and making it run once a week. [This is a useful website](https://crontab.guru/) to check the cron expression. I'd recommend running this on early Tuesday mornings (i.e. sometime ~4am UTC) because that's the last time we could catch a newly-added service missing from TeamCity before the the tests run on Wednesday morning and are used for the release cut.

.github/workflows/teamcity-services-diff-check.yml
```diff
@@ -0,0 +1,93 @@
...
+  check-pr:
+    runs-on: ubuntu-22.04
+    outputs:
+      services: ${{steps.services.outputs.services}}
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+      - id: services
+        run: |
```
SarahFrench, line 22: Could you add a `name` property here to help indicate what the step is doing, e.g. "Check for new services"

.github/workflows/teamcity-services-diff-check.yml
SarahFrench: Given the PR-specific stuff below using `git diff`, it might be that we can't add a schedule trigger. Instead we might need to pull out some of the stuff below into reusable workflows, and then create a PR-triggered and schedule-triggered version of the workflow that both use the TeamCity-checking steps.

SarahFrench (commented):
Thanks for bearing with me - here are some more thoughts on when to trigger the GHA

.github/workflows/teamcity-services-diff-check.yml
SarahFrench: Addressed by adding explicit flags for both GA and Beta

.github/workflows/teamcity-services-diff-check.yml
SarahFrench: See also: https://github.com/hashicorp/terraform-provider-google/issues/15633#issuecomment-1961697734

tools/teamcity-diff-check/main.go
```diff
@@ -0,0 +1,92 @@
...
+
+	googleServices := []string{}
+	scanner := bufio.NewScanner(file)
+	for scanner.Scan() {
+		googleServices = append(googleServices, scanner.Text())
+	}
+
+	////////////////////////////////////////////////////////////////////////////////
+
+	f, err := os.Open(fmt.Sprintf("../../mmv1/third_party/terraform/.teamcity/components/inputs/%s", *serviceFile+".kt"))
```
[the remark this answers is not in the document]
BBBmau: This would be good to follow, this would also require another flag to include the use of the `service_file.txt` which is where all services are stored from the `git diff` command. This is due to the `service_file` flag also being used for the `.txt` file also

They could be separated to be two different flags: `-servicekt_path` and `-service_txt`

For now we can save this for another time 👍🏼

.github/workflows/teamcity-services-diff-check.yml
BBBmau: Latest commit adds an almost identical workflow that runs on PRs, only difference is the removal of `check-pr` job and the addition of `schedule` where it'll run every tuesday morning (4am UTC)

SarahFrench: Looks good! A future improvement could be to use reusable workflows to reduce repetition, but lets leave that for now and get this merged 🚢

SarahFrench (approved):
LGTM - once this is merged lets open some pretend PRs to confirm that it's behaving as expected, just to be certain

PR #10003, opened 2024-02-16
Fix path in TeamCity GHA

SarahFrench:
Fixes an issue with this workflow that was just added.

I realised that when I tested this workflow previously the action was triggered by editing the workflow file, and I hasn't adequately tested editing files in .teamcity 🤦 

```release-note:none

```


shuyama1 (approved):

SarahFrench:
Thanks!

PR #10004, opened 2024-02-16
Adding Data Source Forwarding Rules

jesse-l:
Adding a new data source that will allow users to list forwarding rules that they have inside a specific region of a Google Cloud Project. This PR is for [#17294](https://github.com/hashicorp/terraform-provider-google/issues/17294)

```release-note:new-datasource
google_compute_forwarding_rules
```


mmv1/third_party/terraform/services/compute/data_source_google_compute_forwarding_rules.go
```diff
@@ -0,0 +1,101 @@
...
+
+			"region": {
+				Type:     schema.TypeString,
+				Optional: true,
+			},
+
+			"rules": {
+				Type:     schema.TypeList,
+				Computed: true,
+				Elem:     &schema.Schema{Type: schema.TypeMap},
```
c2thorn, line 30: why not reuse the resource schema of the existing forwarding rule schema? similar to https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/third_party/terraform/services/sql/data_source_sql_databases.go#L31

I'm not sure a string map will be the best experience for users.

mmv1/third_party/terraform/services/compute/data_source_google_compute_forwarding_rules_test.go
```diff
@@ -0,0 +1,53 @@
...
+	poolName := fmt.Sprintf("tf-%s", acctest.RandString(t, 10))
+	ruleName := fmt.Sprintf("tf-%s", acctest.RandString(t, 10))
+
+	acctest.VcrTest(t, resource.TestCase{
+		PreCheck:                 func() { acctest.AccTestPreCheck(t) },
+		ProtoV5ProviderFactories: acctest.ProtoV5ProviderFactories(t),
+		Steps: []resource.TestStep{
+			{
+				Config: testAccDataSourceGoogleForwardingRuleConfig(poolName, ruleName),
+				Check:  acctest.CheckDataSourceStateMatchesResourceStateWithIgnores("data.google_compute_forwarding_rule.my_forwarding_rule", "google_compute_forwarding_rule.foobar-fr", map[string]struct{}{"port_range": {}, "target": {}}),
```
c2thorn, line 23: what's the reasoning for ignoring fields here?

c2thorn (changes requested):

mmv1/third_party/terraform/services/compute/data_source_google_compute_forwarding_rules.go
jesse-l: I have updated the return type to be the Forwarding Rules schema now.

mmv1/third_party/terraform/services/compute/data_source_google_compute_forwarding_rules_test.go
c2thorn: @jesse-l why should we ignore here? Are we expecting it to not match on purpose?

c2thorn (approved):

PR #8941, opened 2023-09-13
Cross-regional Internal HTTPS Load Balancer Certificate Manager validation

DanielRieske:
The resource `google_compute_target_https_proxy` can reference sslCertificates from two different sources.

Primarly you reference compute SSL certifications but in certain edge cases the API also accepts a reference to a certificate in a certificate manager.

In it's current implementation we only accept a resource link from `sslCertificate` which should be removed to allow references to a certificate in a certificate manager.

See: https://github.com/hashicorp/terraform-provider-google/issues/15805#issuecomment-1716765558

Closes: https://github.com/hashicorp/terraform-provider-google/issues/15805

**Release Note Template for Downstream PRs (will be copied)**

```release-note:enhancement
compute: removed validation on `sslCertificates` in `google_compute_https_target_proxy`
```

PR #10011, opened 2024-02-19
Add ceritificateManagerCertificates field to ComputeRegionTargetHttpsProxy resource

Hamzawy63:
Add a new field certificateManagerCertificates field to ComputeRegionTargetHttpsProxy resource. 

This is exactly the same to what was done for ComputeTargetHttpsProxy https://github.com/GoogleCloudPlatform/magic-modules/pull/9144

**Context** :
The resource ComputeRegionTargetHttpsProxy has a field called sslCertificates, this field used to reference only ssl certificates. Recently, certificates of type CertificateManagerCertificates has been allowed. However, either all the items of the sslCertificates array will be sslCertificates or certificate manager certificates.

Furthermore, the field in TF couldn't accept certificate manager certificates because of a custom_expand function that only validates compute certificates (sslcertificates). It was agreed to handle this situation by defining a new field that shall be used when the customer wants to reference certificate manager certificates instead of sslCertificates. 

[More discussion about this here](https://github.com/GoogleCloudPlatform/magic-modules/pull/8941#issuecomment-1743191092)

What has been done in this PR: 
1. Added a new field `CertificateManagerCerticates` in the resource
2. Removed`required=true` from sslCertificates field
3. stated that SslCertificates and CertificateManagerCertificates are conflicting fields 
4. Used the encoder to rename CertificateManagerCertificates to sslCertificates before sending it to the API
5. Decoder is used to rename back `sslCertificates` to `CertificateManagerCertificates` in TF in case the user used certificate manager certs instead of sslCerts (note that we can't have both sslCertificates and CertificateManagerCertificates defined together) 
6. The API expects the format of the certificate to be `//certificatemanager.googleapis.com/projects/{{project}}/..`. For a better and more convenient UX, an expander is defined so that the customer can use the `id` of the resource (i.e. `projects/{{project}}/...`) and the expander will take care of the conversion 
7. If the customer used the id, the API still responds back with `//certificatemanager.googleapis.com/projects/{{project}}/..`, that's why a diff_suppress_func is used.
8. Added an example that demonstrates using CertificateManagerCertificates in ComputeRegionTargetHttpsProxy

Fixes: https://github.com/hashicorp/terraform-provider-google/issues/16998
b/320650971

```release-note:enhancement
compute: added `certificate_manager_certificates` field to `google_compute_region_target_https_proxy` resource
```

### DIFF
--- a/mmv1/products/compute/RegionTargetHttpsProxy.yaml
+++ b/mmv1/products/compute/RegionTargetHttpsProxy.yaml
@@ -41,6 +41,9 @@ async: !ruby/object:Api::OpAsync
   error: !ruby/object:Api::OpAsync::Error
     path: 'error/errors'
     message: 'message'
+custom_code: !ruby/object:Provider::Terraform::CustomCode
+  encoder: templates/terraform/encoders/compute_region_target_https_proxy.go.erb
+  decoder: templates/terraform/decoders/compute_region_target_https_proxy.go.erb
 examples:
   - !ruby/object:Provider::Terraform::Examples
     name: 'region_target_https_proxy_basic'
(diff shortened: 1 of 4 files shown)


Hamzawy63:
@c2thorn Gentle reminder.

Thanks in advance!

c2thorn (approved):
LGTM due to ComputeTargetHttpsProxy precedent

PR #10013, opened 2024-02-20
Provider functions 1

SarahFrench:

```release-note:none

```


mmv1/provider/terraform/common~compile.yaml
```diff
@@ -102,6 +102,12 @@
 -%>
 '<%= dir -%>/<%= fname.delete_suffix(".erb") -%>': 'third_party/terraform/framework_models/<%= fname -%>'
 <% end -%>
+<%
+  Dir["third_party/terraform/functions/*.go.erb"].each do |file_path|
```
rileykarson, line 106: I think we'll want a rule in https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/provider/terraform/common~copy.yaml as well, to avoid a surprise if we add a copied file in the future

The directory structure under `third_party/terraform` is _so close_ to 1:1 with the provider nowadays, which would remove the need for these mapping rules at all...

rileykarson (approved):

SarahFrench:
Test failure is unrelated, see : https://github.com/hashicorp/terraform-provider-google/issues/17319

PR #10015, opened 2024-02-20
Removed GH token from VCR nightly job

shuyama1:
After a quick investigation, I don't think it's needed for VCR nightly job for now.

```release-note:none

```


melinath (approved):
LGTM

PR #10016, opened 2024-02-20
Add check for Environment proto field before accessing in Dataflow provider

jrmccluskey:
Adds an extra check to the Dataflow provider when accessing job proto fields; an edge case existed where the `Environment` proto field is nil, causing a nil-pointer exception when accessing sub-fields rather than an error that can be returned up the call stack. Checking for existence and bubbling an error up should eliminate that issue.

Validated the code change is propagated to the generated repos following https://googlecloudplatform.github.io/magic-modules/get-started/generate-providers/, also passed linting + unit tests + dataflow acceptance tests

Fixes https://github.com/hashicorp/terraform-provider-google/issues/17046

```release-note:bug
dataflow: fixes potential nil-pointer error if a job's Environment field is nil when reading job information
```

### DIFF
--- a/mmv1/third_party/terraform/services/dataflow/resource_dataflow_job.go.erb
+++ b/mmv1/third_party/terraform/services/dataflow/resource_dataflow_job.go.erb
@@ -371,6 +371,9 @@ func resourceDataflowJobRead(d *schema.ResourceData, meta interface{}) error {
 	if err := d.Set("effective_labels", job.Labels); err != nil {
 		return fmt.Errorf("Error setting effective_labels: %s", err)
 	}
+	if job.Environment == nil {
+		return fmt.Errorf("Error accessing Environment proto: proto is nil")
+	}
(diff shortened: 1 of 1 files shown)


c2thorn:
agree that this nil check is needed.
But does this actually fix https://github.com/hashicorp/terraform-provider-google/issues/17046 or just bubble up an error vs a crash?

c2thorn:
@jrmccluskey

jrmccluskey:
> agree that this nil check is needed. But does this actually fix [hashicorp/terraform-provider-google#17046](https://github.com/hashicorp/terraform-provider-google/issues/17046) or just bubble up an error vs a crash?

As best I can tell, the true fix is on the service side making sure that proto field always exists. At the very least this avoids a panic -> crash in that situation and gives a clearer error message

c2thorn (approved):

PR #10019, opened 2024-02-20
 Added `CloudQuotas` service and Create `QuotaPreference` Resource

lola98:
Add new resource `google_cloud_quotas_quota_preference` to manage a project's quota preference. 

```release-note:new-resource
`google_cloud_quotas_quota_preference`
```


mmv1/products/cloudquotas/QuotaPreference.yaml
```diff
@@ -0,0 +1,164 @@
...
+skip_delete: true
+autogen_async: false
+custom_code: !ruby/object:Provider::Terraform::CustomCode
+  encoder: templates/terraform/encoders/quota_preference.go.erb
+
+examples:
+  - !ruby/object:Provider::Terraform::Examples
+    name: "cloudquotas_quota_preference_basic"
+    primary_resource_id: "preference"
+    skip_test: true
```
roaks3, line 37: Am I understanding correctly the this test is skipped because it is covered by the handwritten tests?

mmv1/products/cloudquotas/QuotaPreference.yaml
```diff
@@ -0,0 +1,164 @@
...
+  - !ruby/object:Api::Type::String
+    name: 'parent'
+    immutable: true
+    url_param_only: true
+    description: The parent of the quota preference. Allowed parents are "projects/[project-id / number]" or "folders/[folder-id / number]" or "organizations/[org-id / number]".
+  - !ruby/object:Api::Type::String
+    name: 'quota_preference_id'
+    immutable: true
+    url_param_only: true
+    default_value: ""
```
roaks3, line 53: Do you know if this default is needed?

mmv1/products/cloudquotas/QuotaPreference.yaml
```diff
@@ -0,0 +1,164 @@
...
+    default_value: :QUOTA_SAFETY_CHECK_UNSPECIFIED
+    description: The list of quota safety checks to be ignored.
+    values:
+      - :QUOTA_SAFETY_CHECK_UNSPECIFIED
+      - :QUOTA_DECREASE_BELOW_USAGE
+      - :QUOTA_DECREASE_PERCENTAGE_TOO_HIGH
+
+properties:
+  - !ruby/object:Api::Type::String
+    name: 'name'
```
roaks3, line 77: If I'm understanding correctly, using `name` takes precedence over `parent` and `quota_preference_id`, even if they are specified. Is that correct? If so, we may want to consider using `conflicts_with` or similar to prevent unexpected behavior.

mmv1/products/cloudquotas/QuotaPreference.yaml
```diff
@@ -0,0 +1,164 @@
...
+        description: |
+          The trace id that the Google Cloud uses to provision the requested quota. This trace id may be used by the client to contact Cloud support to track the state of a quota preference request. The trace id is only produced for increase requests and is unique for each request. The quota decrease requests do not have a trace id.
+      - !ruby/object:Api::Type::KeyValuePairs
+        name: 'annotations'
+        description: |-
+          The annotations map for clients to store small amounts of arbitrary data. Do not put PII or other sensitive information here. See https://google.aip.dev/128#annotations.
+
+          An object containing a list of "key: value" pairs. Example: { "name": "wrench", "mass": "1.3kg", "count": "3" }.
+      - !ruby/object:Api::Type::Enum
+        name: 'requestOrigin'
```
roaks3, line 120: Since this is output-only, it is usually simpler to define it as a string. Then we don't need to manage a client-side set of options.

mmv1/products/cloudquotas/QuotaPreference.yaml
```diff
@@ -0,0 +1,164 @@
...
+          - :CLOUD_CONSOLE`
+          - :AUTO_ADJUSTER
+  - !ruby/object:Api::Type::KeyValuePairs
+    name: 'dimensions'
+    description: |-
+      The dimensions that this quota preference applies to. The key of the map entry is the name of a dimension, such as "region", "zone", "network_id", and the value of the map entry is the dimension value. If a dimension is missing from the map of dimensions, the quota preference applies to all the dimension values except for those that have other quota preferences configured for the specific value.
+
+      NOTE: QuotaPreferences can only be applied across all values of "user" and "resource" dimension. Do not set values for "user" or "resource" in the dimension map.
+
+      Example: {"provider", "Foo Inc"} where "provider" is a service specific dimension.
```
roaks3, line 134: Is there a typo here? Should it be `{"provider": "Foo Inc"}`?

mmv1/products/cloudquotas/QuotaPreference.yaml
```diff
@@ -0,0 +1,164 @@
...
+  - !ruby/object:Api::Type::String
+    name: 'etag'
```
roaks3, line 136: Normally `etag` is not needed in Terraform. Do you know if it has a practical use case for this resource?

mmv1/templates/terraform/examples/cloudquotas_quota_preference_basic.tf.erb
```diff
@@ -0,0 +1,15 @@
+# [START cloudquotas_quota_preference_basic]
```
roaks3, line 1: These tags are no longer needed, so they can be removed

mmv1/templates/terraform/examples/cloudquotas_quota_preference_basic.tf.erb
```diff
@@ -0,0 +1,15 @@
...
+  name          = "projects/<%= ctx[:test_env_vars]['project'] %>/locations/global/quotaPreferences/<%= ctx[:vars]['quota_preference_id'] %>"
+  quota_config  {
+    preferred_value = 210
+  }
+  dimensions = {
+    region = "us-central2"
+  }
+  service       = "compute.googleapis.com"
+  quota_id      = "CPUS-per-project-region"
+  contact_email = "liulola@google.com"
```
roaks3, line 12: These examples are made public on our documentation website, so I would consider using a fake email address here.

mmv1/third_party/terraform/services/cloudquotas/resource_cloud_quotas_quota_preference_test.go
```diff
@@ -0,0 +1,82 @@
...
+}
+
+func testAccCloudQuotasQuotaPreference_cloudquotasQuotaPreferenceBasicExample_basic(context map[string]interface{}) string {
+	return acctest.Nprintf(`
+resource "google_cloud_quotas_quota_preference" "preference" {
+  name          = "projects/%{project}/locations/global/quotaPreferences/tf-test-compute_googleapis_com-CPUS-per-project_us-central2%{random_suffix}"
+  quota_config  {
+    preferred_value = 210
+  }
+  dimensions = {
```
roaks3, line 53: Ideally, one of these test configs should be minimal, with the least amount of fields provided. That helps to verify that optional fields are in fact optional.

Then, we should be updating all fields that can be mutated (for example, the schema defines `service` as being optional).

roaks3 (changes requested):
Overall the resource looks good, I just had some comments on a few details, and the test should be updated to be more complete.

mmv1/products/cloudquotas/QuotaPreference.yaml
lola98: I see, switched to string.

mmv1/products/cloudquotas/QuotaPreference.yaml
lola98: This test is skipped because the resource can only be created once and cannot be deleted (our API has no deletion method). Recreating the same resource would cause API throwing an exception. The handwritten test is also failing due to the same reason, I'm wondering is there any workaround solutions?

mmv1/products/cloudquotas/QuotaPreference.yaml
lola98: Done, switched to string.

mmv1/products/cloudquotas/QuotaPreference.yaml
lola98: Yes I was thinking `parent` and `quota_preference_id` fields will be fetched from `name` if none are specified. Also name is a required field for UPDATE, so the user do not need to input redundant information. I've added `conflicts_with` for `name`.

mmv1/templates/terraform/examples/cloudquotas_quota_preference_basic.tf.erb
lola98: Got it, switched to a fake email.

mmv1/templates/terraform/examples/cloudquotas_quota_preference_basic.tf.erb
lola98: Done.

mmv1/products/cloudquotas/QuotaPreference.yaml
lola98: Yes fixed.

mmv1/products/cloudquotas/QuotaPreference.yaml
lola98: Seems like not needed, API will generate a UUID if given an empty id. Removed.

mmv1/products/cloudquotas/QuotaPreference.yaml
lola98: Quota preference may not be approved immediately. The users can use `etag` as an option to get the latest version if they want the check the status of a preference. I'd say to keep this field as it aligns with our API design.

mmv1/products/cloudquotas/QuotaPreference.yaml
roaks3: You may need to create a new project as part of the test, and then apply the quota to that project, in order to run this test nightly. But within a single run, I think it is fair to keep this example skipped (to reduce run time of tests, and the number of additional projects created).

There are many examples in the codebase you can see where a `google_project` is being created within the test config, but let me know if you need help.

mmv1/products/cloudquotas/QuotaPreference.yaml
roaks3: Ok, the "conflicts with" looks good. It may still be worth reducing the duplication, and I'm remembering that we usually use a pattern here where `name` doubles as the name and the id. See how Cloud Run V2 Service is set up with https://cloud.google.com/run/docs/reference/rest/v2/projects.locations.services/create and https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/products/cloudrunv2/Service.yaml#L135-L145. The user would specify the `parent` and optionally the `name`, and after creation, the `name` would contain the full name value including the parent.

```
resource "google_cloud_quotas_quota_preference" "my-preference" {
  parent = "projects/123/locations/global"
  name = "compute_googleapis_com-CPUS-per-project-us-central1"
  ...
```

After apply, the `name` in TF state is `"projects/123/locations/global/quotaPreferences/compute_googleapis_com-CPUS-per-project-us-central1"`.

mmv1/third_party/terraform/services/cloudquotas/resource_cloud_quotas_quota_preference_test.go
lola98: Done, I added more tests to cover scenarios where mutable fields are modified. . 
I have a question regarding the design:  Our API accepts the `allow-missing` flag for `update`. When enabled, the `update` method is equivalent to `create`. If we take out the flag, certain fields (like `service`, `quotaId` and `dimensions`) become immutable. When the user wants to update those fields, they will have to create a new preference. I'd like your suggestion whether we want to support such feature in TF?

mmv1/products/cloudquotas/QuotaPreference.yaml
lola98: Implementation updated to match the example's approach.

mmv1/products/cloudquotas/QuotaPreference.yaml
lola98: Added project creation as part of the test. Could you help run `/gcbrun`? It doesn't seem to respond when I trigger it.

roaks3: Ah yep, `/gcbrun` can only be trigger by our team at the moment.

mmv1/third_party/terraform/services/cloudquotas/resource_cloud_quotas_quota_preference_test.go
```diff
@@ -0,0 +1,292 @@
...
+			name       		= "tf-test%{random_suffix}"
+			org_id			= "%{org_id}"
+			billing_account	= "%{billing_account}"
+		}
+
+		resource "google_project_iam_binding" "project_iam" {
+			project 	= google_project.new_project.project_id
+			role    	= "roles/cloudquotas.admin"
+			members 	= [
+				"user:testinguser2@google.com"
```
roaks3, line 84: I think you will need to use the configured service account (which will be our test runner in CI environments). We just came across a similar case, so if you have issues setting up this permission, we can consider alternative approaches.

mmv1/third_party/terraform/services/cloudquotas/resource_cloud_quotas_quota_preference_test.go
lola98: Using `allow-missing` could modify the `parent``and `name` fields, which are used to generate the resource ID. This led to provider import errors, so I removed the allow-missing flag.

mmv1/third_party/terraform/services/cloudquotas/resource_cloud_quotas_quota_preference_test.go
lola98: Previously our backend is checking permission for the `contact_email` field, and the block grants essential permissions to the email account. We've removed our IAM checking mechanism and the change is now live in prod, so I've removed the block.

Looking into the test report, it seems the API is complaining about decreasing the quota by too much. I've added a parameter to force the change.

mmv1/third_party/terraform/services/cloudquotas/resource_cloud_quotas_quota_preference_test.go
roaks3: Yea, I think that was the correct outcome here in terms of ergonomics as well. Terraform expects more explicit CRUD actions, so an upsert typically does not fit as well.

lola98:
Hi @roaks3, could you help run /gcbrun? I've addressed comments from the `missing test report` and all other tests passed. Thanks!

mmv1/templates/terraform/constants/cloudquotas_quota_preference.go.erb
```diff
@@ -0,0 +1,3 @@
+func QuotaConfigAnnotationsDiffSuppress(k, old, new string, d *schema.ResourceData) bool {
```
roaks3, line 1: I would guess that we don't want to always suppress diffing on this field. Was this added to support server-controlled annotations?

mmv1/third_party/terraform/services/cloudquotas/resource_cloud_quotas_quota_preference_test.go
```diff
@@ -0,0 +1,200 @@
+// Copyright (c) HashiCorp, Inc.
```
roaks3, line 1: Nit: license added twice. You actually shouldn't need it at all in this repository.

mmv1/third_party/terraform/services/cloudquotas/resource_cloud_quotas_quota_preference_test.go
```diff
@@ -0,0 +1,200 @@
...
+		resource "google_project_service" "cloudquotas" {
+			project  	= google_project.new_project.project_id
+			service 	= "cloudquotas.googleapis.com"
+			depends_on	= [google_project.new_project]
+		}
+
+		resource "google_project_service" "compute" {
+			project  	= google_project.new_project.project_id
+			service 	= "compute.googleapis.com"
+			depends_on	= [google_project_service.cloudquotas]
```
roaks3, line 82: Nit: you might be able to speed up the test by depending on all 3 services in the sleep, and having them enabled in parallel

mmv1/third_party/terraform/services/cloudquotas/resource_cloud_quotas_quota_preference_test.go
```diff
@@ -0,0 +1,200 @@
...
+		
+		resource "time_sleep" "wait_180_seconds" {
+			create_duration	= "180s"
+			depends_on		= [
+				google_project_service.billing, 
+			]
+		}
+
+		resource "google_cloud_quotas_quota_preference" "my_preference"{
+			parent					= "projects/${google_project_service.billing.project}"
```
roaks3, line 99: You have an implicit dependency on the billing service being enabled, but you don't block on the sleep here. I suspect you need a depends_on for the `time_sleep.wait_180_seconds`, and this line could then use `google_project.new_project.project_id` instead.

roaks3 (commented):
LGTM, just need to make sure annotations are working properly, and that the tests will be reliable with the sleeps.

mmv1/third_party/terraform/services/cloudquotas/resource_cloud_quotas_quota_preference_test.go
```diff
@@ -0,0 +1,200 @@
...
+				ImportStateVerifyIgnore: []string{"parent", "quota_preference_id", "validate_only", "ignore_safety_checks", "contact_email"},
+			},
+			{
+				Config: testAccCloudQuotasQuotaPreference_cloudquotasQuotaPreferenceBasicExample_increaseQuota(context),
+			},
+			{
+				ResourceName:            "google_cloud_quotas_quota_preference.my_preference",
+				ImportState:             true,
+				ImportStateVerify:       true,
+				ExpectNonEmptyPlan:      true,
```
roaks3, line 48: Oops, missed this one: I believe we should remove this line, but there is probably something going on that prompted adding it here (and my guess is that it's the same reason tests just failed). Perhaps the preferred value is not returned from the API or something like that?

lola98:
Hmm It seems the unused import is causing a build failure. I added this for a new data source, and no resources exist under the service yet.  I've remove the import block.
build failure

roaks3 (changes requested):
Looks like tests passed, but I see 2 remaining items on this one:
- Handling of annotations (and potentially removing the diff suppress)
- We probably still want a 3rd test step using `validate_only`, unless you found that to be infeasible

lola98:
`annotations` is a input-only field and our API will not store / modify the data, and will pass it to the backend directly.
I tried to include "quota_config.0.annotations" in `ImportStateVerifyIgnore` but still receives an import error since this is an input-only field. I wasn't able to use `Type::KeyValueAnnotations` since the field is not at the top level. 

Screenshot 2024-03-18 at 12 21 15 PM

lola98:
According to https://yaqs.corp.google.com/eng/q/3735742487973593088, it seems terraform does not support `validate_only` field, so I removed it.

lola98:
> `annotations` is a input-only field and our API will not store / modify the data, and will pass it to the backend directly. I tried to include "quota_config.0.annotations" in `ImportStateVerifyIgnore` but still receives an import error since this is an input-only field. I wasn't able to use `Type::KeyValueAnnotations` since the field is not at the top level.
> 
> I got the following error: Screenshot 2024-03-18 at 12 21 15 PM
> 
> Do you know how to resolve this?

Instead of suppressing diff, I switched to implementing the ignore_read feature suggested in the [doc](https://googlecloudplatform.github.io/magic-modules/develop/permadiff/#ignore_read).

mmv1/third_party/terraform/provider/provider_mmv1_resources.go.erb
```diff
@@ -15,9 +15,6 @@ import (
 	"github.com/hashicorp/terraform-provider-google/google/services/dataflow"
 	"github.com/hashicorp/terraform-provider-google/google/services/servicenetworking"
 	"github.com/hashicorp/terraform-provider-google/google/tpgiamresource"
-	<% if version == 'ga' -%> // https://github.com/hashicorp/terraform-provider-google/issues/15633 for details
```
roaks3, line 18: With the other PRs merged, do we still need to remove this, or do we possibly need a rebase?

roaks3 (approved):
LGTM pending tests, and one note to double check that we want to remove the import

mmv1/third_party/terraform/services/cloudquotas/resource_cloud_quotas_quota_preference_test.go
```diff
@@ -0,0 +1,195 @@
...
+			project_id 		= "tf-test%{random_suffix}"
+			name       		= "tf-test%{random_suffix}"
+			org_id			= "%{org_id}"
+			billing_account	= "%{billing_account}"
+		}
+
+		resource "google_project_service" "cloudquotas" {
+			project  	= google_project.new_project.project_id
+			service 	= "cloudquotas.googleapis.com"
+			depends_on	= [google_project.new_project]
```
roaks3, line 71: Nit: for the `google_project_service`, it already uses `google_project.new_project.project_id`, which sets up an implicit dependency. So you don't need the `depends_on` for these. Same for the other 2 templates.

mmv1/third_party/terraform/services/cloudquotas/resource_cloud_quotas_quota_preference_test.go
```diff
@@ -0,0 +1,195 @@
...
+		}
+
+		resource "google_project_service" "compute" {
+			project  	= google_project.new_project.project_id
+			service 	= "compute.googleapis.com"
+			depends_on	= [google_project.new_project]
```
roaks3, line 77: I was not aware of this, but when multiple `google_project_service` resources are used, they are actually batched into a single request, and that appears to cause issues with our VCR recording. I think you will need to chain these as an alternative unfortunately (see https://github.com/GoogleCloudPlatform/magic-modules/pull/7282 as an example).

roaks3 (approved):
This looks ready to merge, but did you see my comment above regarding the import? https://github.com/GoogleCloudPlatform/magic-modules/pull/10019/files#r1532158791

lola98:
> This looks ready to merge, but did you see my comment above regarding the import? https://github.com/GoogleCloudPlatform/magic-modules/pull/10019/files#r1532158791

Yes IIUC, after we add this resource, the service will be automatically imported. So manual import is no longer required. The discussion was here: https://github.com/GoogleCloudPlatform/magic-modules/pull/10071#discussion_r1518293147.

PR #10025, opened 2024-02-21
Ensured that beta runs in TeamCity use only beta paths

melinath:
Made sure that all packages used in beta configurations use `google-beta` in their path instead of `google`. I also added tests that I think would catch this issue in the future for sweepers, since that's the problem at hand - I'm also not as confident about how to add tests for the other packages.

```release-note:none

```


SarahFrench (approved):
These changes look good - I've left some comments but they're all effectively +1's to the changes you've made or notes about how I'll pull your changes to PRs I have in-flight!

SarahFrench:
Merging so I can ensure the changes work ok in TeamCity and will be in place for tonight's tests

PR #10033, opened 2024-02-22
make go-converted YAML compatible with go compiler

c2thorn:
combines https://github.com/GoogleCloudPlatform/magic-modules/pull/9869 and https://github.com/GoogleCloudPlatform/magic-modules/pull/9853

with fixes for both + formatting.

End result is 
- compatible+linted go YAML output for datafusion+pubsub
- conversion from these YAML to full(?) resource DTO in go

My specific changes are in [`b9cc170` (#10033)](https://github.com/GoogleCloudPlatform/magic-modules/pull/10033/commits/b9cc1707f043275339f67616b4b3240da9359e79)

```release-note:none

```


NickElliot (commented):
Thanks for cleaning up the janky "layer" logic -- I was completely at a loss for getting the indentation to format correctly for the nested fields and that was the best I could come up with despite it not seeming like an issue for actual .go resource generation.

Should there be a comment on the example go_*.yaml files clarifying that these exist for/in a state of in-progress testing and should not be modified by contributors, if they're going to be merged to the main branch in current state?

c2thorn:
> Should there be a comment on the example go_*.yaml files clarifying that these exist for/in a state of in-progress testing and should not be modified by contributors, if they're going to be merged to the main branch in current state?

I think this makes a lot of sense! I'll add this in.

Also, I've commented out YAML generation to the providers. I am somewhat concerned about a potential side effect from having the YAML generate (CI issue or something, it already isn't detected by the diff generator) so I'd rather just deal with uncommenting that than running that risk. Does that seem reasonable to you?

NickElliot (approved):
all seems good to me now then!

PR #10037, opened 2024-02-23
Bump nodejs10 to nodejs18 in GAE tests and docs

SarahFrench:
This addresses a bunch of TestAccAppEngine* test failures due to :

```
        Error: Error creating StandardAppVersion: googleapi: Error 400: Error(s) encountered validating runtime. Runtime nodejs10 is end of support and no longer allowed. Please use the latest Node.js runtime for App Engine Standard.
```

```release-note:none

```

### DIFF
--- a/mmv1/templates/terraform/examples/app_engine_application_url_dispatch_rules_basic.tf.erb
+++ b/mmv1/templates/terraform/examples/app_engine_application_url_dispatch_rules_basic.tf.erb
@@ -15,7 +15,7 @@ resource "google_app_engine_application_url_dispatch_rules" "<%= ctx[:primary_re
 resource "google_app_engine_standard_app_version" "admin_v3" {
   version_id = "v3"
   service    = "admin"
-  runtime    = "nodejs10"
+  runtime    = "nodejs18"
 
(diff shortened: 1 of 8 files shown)


melinath:
It looks like there's maybe a delay between the version resource being created and it being available via GET? Maybe that resource should use `resource_inside_response: true` instead of making a follow-up GET request?

melinath (approved):
I'm tentatively approving this since it resolves a bunch of errors - the remainder are likely not caused by the changes being made here. But if the resource_inside_response fix worked & you wanted to add it go for it!

shuyama1:
@SarahFrench @melinath Sorry, I didn't see this PR and made a similar PR https://github.com/GoogleCloudPlatform/magic-modules/pull/10047 (bump the version to the latest - nodejs20 ). I have a clean test pass, so we can go ahead and merge that one if it's easier. I think what caused the tests failed here is that same tests triggered twice in recording and ended up interrupting each other. I've already flagged @trodge about this VCR issue, which seems coming from how the tests are extracted from the replaying build log.

melinath:
SGTM

PR #10039, opened 2024-02-23
Add support for string --> object map for DCL resources

slevenick:
Fixes: https://github.com/hashicorp/terraform-provider-google/issues/16777

Tests for gkehub pass in CI

This adds an override for these maps that specifies the key name and must be present for these fields to work.

```release-note:enhancement
gkehub: added support for `policycontroller.policy_controller_hub_config.policy_content.bundles` and 
`policycontroller.policy_controller_hub_config.deployment_configs` fields to `google_gke_hub_feature_membership`
```

### DIFF
--- a/tpgtools/property.go
+++ b/tpgtools/property.go
@@ -107,6 +107,10 @@ type Property struct {
 	// Sub-properties of nested objects or arrays with nested objects
 	Properties []Property
 
+	// If this is a complex map type, this string represents the name of the
+	// field that the key to the map can be set with
+	ComplexMapKeyName string
+
 	// Reference to the parent resource.
 	// note: "Properties" will not be available.
 	resource *Resource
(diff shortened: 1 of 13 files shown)


slevenick:
Tests pass in CI except for `TestAccGKEHubFeatureMembership_gkehubFeatureAcmUpdate` which is unrelated

tpgtools/property.go
```diff
@@ -651,6 +673,38 @@ func createPropertiesFromSchema(schema *openapi.Schema, typeFetcher *TypeFetcher
...
+			}
+			keyProp := Property{
+				title:    cm.KeyName,
+				Type:     Type{&openapi.Schema{Type: "string"}},
+				resource: resource,
+				parent:   &p,
+				Required: true,
+				Description: "The name for the key in the map for which this object is mapped to in the API",
+			}
+			props = append(props, keyProp)
```
rileykarson, line 700: nit: Would we be able to prepend instead? That'll make it appear first rather than last. We don't seem to reorder these fields by required->optional->output, unless that transformation is docs only.

rileykarson (commented):
Docs for these resources are weird (I think they're handwritten?), so you'll need to manually regenerate those!

slevenick:
> Docs for these resources are weird (I think they're handwritten?), so you'll need to manually regenerate those!

Uhg, forgot about handwritten docs!

tpgtools/property.go
slevenick: Done

slevenick:
I'm unsure how there are failing tests, they are in firebase and those files are unchanged in downstreams

rileykarson (approved):
LGTM

Yeah those are clearly not caused by this change

PR #10046, opened 2024-02-25
Retry delete network step while creating a google project.

ajaybgunjal1:
compute.googleapis.com API must be enabled to delete the network related to the newly created project. 

We first enable the compute API. But this takes some time to propagate to all the systems. Hence, delete network API sometimes returns `403- compute api is not enabled` error.

To mitigate this issue, we will wait for 10s and retry the delete network step when there is 403 error.

```release-note:bug
resourcemanager: added a retry to deleting the default network when `auto_create_network` is false in `google_project`
```

### DIFF
--- a/mmv1/third_party/terraform/services/resourcemanager/resource_google_project.go
+++ b/mmv1/third_party/terraform/services/resourcemanager/resource_google_project.go
@@ -227,7 +227,13 @@ func resourceGoogleProjectCreate(d *schema.ResourceData, meta interface{}) error
 			return errwrap.Wrapf("Error enabling the Compute Engine API required to delete the default network: {{err}} ", err)
 		}
 
-		if err = forceDeleteComputeNetwork(d, config, project.ProjectId, "default"); err != nil {
+		err = forceDeleteComputeNetwork(d, config, project.ProjectId, "default")
+		// Retry if API is not yet enabled.
+		if err != nil && transport_tpg.IsGoogleApiErrorWithCode(err, 403) {
+			time.Sleep(10 * time.Second)
+			err = forceDeleteComputeNetwork(d, config, project.ProjectId, "default")
+		}
+		if err != nil {
 			if transport_tpg.IsGoogleApiErrorWithCode(err, 404) {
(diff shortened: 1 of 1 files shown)


c2thorn:
can you write a release note following https://googlecloudplatform.github.io/magic-modules/contribute/release-notes/ @ajaybgunjal1

mmv1/third_party/terraform/services/resourcemanager/resource_google_project.go
```diff
@@ -227,7 +227,13 @@ func resourceGoogleProjectCreate(d *schema.ResourceData, meta interface{}) error
 			return errwrap.Wrapf("Error enabling the Compute Engine API required to delete the default network: {{err}} ", err)
 		}
 
-		if err = forceDeleteComputeNetwork(d, config, project.ProjectId, "default"); err != nil {
+		err = forceDeleteComputeNetwork(d, config, project.ProjectId, "default")
+		// Retry if API is not yet enabled.
+		if err != nil && transport_tpg.IsGoogleApiErrorWithCode(err, 403) {
+			time.Sleep(10 * time.Second)
```
c2thorn, line 233: Noting for future reviewers: Normally we would want to send the request in a retry loop using our [standard retry predicates](https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/third_party/terraform/transport/error_retry_predicates.go#L23)

But forceDeleteComputeNetwork actually involves multiple steps, so a single sleep retry should be fine in this case.

c2thorn (approved):

ajaybgunjal1:
Thank you for the approval!

mmv1/third_party/terraform/services/resourcemanager/resource_google_project.go
melinath: Why are we adding the retry here instead of adding retries inside forceDeleteComputeNetwork?

c2thorn: I thought about that, but the 403's we're trying to retry on here occur right after project creation. If anything wanted to reuse the function (this is the only use so far), I thought it would be better to keep this case contained to `google_project`. Open to other ideas though.

PR #10050, opened 2024-02-26
Instance lifecycle policy default action on failure implementation

askubis:
This Creates (R)IGM.InstanceLifecyclePolicy.DefaultActionOnFailure functionalities in Beta and GA . Modifies Compute (Region)InstanceGroupManagers.

fixes #17409

Searched through the [issue tracker](https://github.com/hashicorp/terraform-provider-google/issues) for an open issue that this either resolves or contributes to, commented on it to claim it, and written "fixes {url}" or "part of {url}" in this PR description. If there were no relevant open issues, I opened one and commented that I would like to work on it (not necessary for very small changes).
[Generated Terraform](https://github.com/GoogleCloudPlatform/magic-modules#generating-downstream-tools), and ran make test and make lint to ensure it passes unit and linter tests.
Ensured that all new fields I added that can be set by a user appear in at least one [example](https://github.com/GoogleCloudPlatform/magic-modules/tree/main/templates/terraform/examples) (for generated resources) or [third_party test](https://github.com/GoogleCloudPlatform/magic-modules/tree/main/third_party/terraform/tests) (for handwritten resources or update tests).
[Ran](https://github.com/hashicorp/terraform-provider-google/blob/main/.github/CONTRIBUTING.md#tests) relevant acceptance tests (If the acceptance tests do not yet pass or you are unable to run them, please let your reviewer know).
Read the [Release Notes Guide](https://github.com/GoogleCloudPlatform/magic-modules/blob/main/.ci/RELEASE_NOTES_GUIDE.md) before writing my release note below.

compute: added `instance_lifecycle_policy.default_action_on_failure` field in resources `google_compute_instance_group_manager` and `google_compute_region_instance_group_manager` in Beta and GA


askubis:
Fixed description

SarahFrench (commented):
Thanks for your PR, everything currently looks good!

One question I have is about what would happen if a user provisioned an instance group manager resource (regional or not) with default_action_on_failure set to the non-default value `DO_NOTHING` and then subsequently removed that field from their config. Should they expect Terraform to change the field to the default value when the config isn't explicitly setting the value to `DO_NOTHING`?

Could you please add some lines like this to the 'update' tests you have edit to assert the expected value of the field in each test step after the config changes are applied?:

```diff
{
	Config: testAccInstanceGroupManager_update(template1, target1, description, igm),
+	Check: resource.ComposeTestCheckFunc(
+		resource.TestCheckResourceAttr("google_compute_instance_group_manager.igm-update", "instance_lifecycle_policy.0.default_action_on_failure", "DO_NOTHING"),
+	),
},
```

For context, if a resource has a boolean field to enable a feature on a resource users tend to expect that deleting `enabled = true` from their Terraform configuration to have the same effect as explicitly changing it to `enabled = false`, assuming that the API's default behaviour is to have that feature NOT enabled. I'm wondering if it makes sense to update these resources to revert to the default value `"REPAIR"` if the field is deleted too, or if it doesn't make sense to do so. What do you think?

askubis:
Thanks for the review Sarah!
I've added the assertions you've mentioned.

Re: "I'm wondering if it makes sense to update these resources to revert to the default value "REPAIR" if the field is deleted too, or if it doesn't make sense to do so. What do you think?"
I'll keep it in mind, but I've tested some fields both on the instance_group_manager and instance resources, and it seems we do not do it (removing a field like this results in no delta seen by terraform, so it does not update the resources at all).
For now lets keep it consistent, and I'll think about it wrt. coming major release, OK?

Generally speaking there are two contradicting arguments I see in this case:
1. Since terraform is declarative, the same configuration should always produce the same resources, so we should always send all the default values to update possibly changed fields.
2. Since one did not specify a field in the configuration they don't care about mainpulating the value (and sometimes it is on purpose, e.g. if we would always specify target_size that would be conflicting with autoscaler that changes the size of the igm by using the same field(target_size))

SarahFrench:
> I've tested some fields both on the instance_group_manager and instance resources, and it seems we do not do it (removing a field like this results in no delta seen by terraform, so it does not update the resources at all).

If un-setting/removing the field doesn't result in a plan that suggests Terraform is going to do anything to the field then I think it's ok to leave the resource as it is currently. [When this situation has arisen in the past with boolean fields](https://github.com/hashicorp/terraform-provider-google/issues/17178) Terraform has indicated in a plan that the field would be set to `null`, in which case it's reasonable for users to assume that the field will indeed be unset and resume to using any default values.

Thanks for taking a look into this!

SarahFrench (approved):
Approving, will merge once the checks complete (and assuming they pass)

mmv1/third_party/terraform/services/compute/resource_compute_region_instance_group_manager_test.go.erb
```diff
@@ -83,6 +83,9 @@ func TestAccRegionInstanceGroupManager_update(t *testing.T) {
 		Steps: []resource.TestStep{
 			{
 				Config: testAccRegionInstanceGroupManager_update(template1, target1, igm),
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr("google_compute_instance_group_manager.igm-update", "instance_lifecycle_policy.0.default_action_on_failure", "DO_NOTHING"),
```
SarahFrench, line 87:
```suggestion
					resource.TestCheckResourceAttr("google_compute_region_instance_group_manager.igm-update", "instance_lifecycle_policy.0.default_action_on_failure", "DO_NOTHING"),
```

mmv1/third_party/terraform/services/compute/resource_compute_region_instance_group_manager_test.go.erb
```diff
@@ -92,6 +95,9 @@ func TestAccRegionInstanceGroupManager_update(t *testing.T) {
 			},
 			{
 				Config: testAccRegionInstanceGroupManager_update2(template1, target1, target2, template2, igm),
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr("google_compute_instance_group_manager.igm-update", "instance_lifecycle_policy.0.default_action_on_failure", "REPAIR"),
```
SarahFrench, line 99:
```suggestion
					resource.TestCheckResourceAttr("google_compute_region_instance_group_manager.igm-update", "instance_lifecycle_policy.0.default_action_on_failure", "REPAIR"),
```

mmv1/third_party/terraform/services/compute/resource_compute_region_instance_group_manager_test.go.erb
```diff
@@ -101,6 +107,9 @@ func TestAccRegionInstanceGroupManager_update(t *testing.T) {
 			},
 			{
 				Config: testAccRegionInstanceGroupManager_update3(template1, target1, target2, template2, igm),
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr("google_compute_instance_group_manager.igm-update", "instance_lifecycle_policy.0.default_action_on_failure", "REPAIR"),
```
SarahFrench, line 111:
```suggestion
					resource.TestCheckResourceAttr("google_compute_region_instance_group_manager.igm-update", "instance_lifecycle_policy.0.default_action_on_failure", "REPAIR"),
```

askubis:
Thanks, I did not regenerate the code after modifying the regional_... tests.
Thanks for fixing that!

SarahFrench:
No problem!

/gcbrun

PR #10051, opened 2024-02-26
Add IAM resources for Cloud Deploy Custom Target Type

mattsanta:
Part of https://github.com/hashicorp/terraform-provider-google/issues/13042

```release-note:new-resource
google_clouddeploy_custom_target_type_iam_*
```


SarahFrench (approved):
LGTM - thanks for navigating that error about import IDs. I'm going to overlook the [warning about fields not being tested](https://github.com/GoogleCloudPlatform/magic-modules/pull/10051#issuecomment-1968963078) because the IAM resource tests are generated and other similar IAM resources' tests don't include these fields.

PR #10056, opened 2024-02-26
Make kmsServiceAccountId editable for Logging Settings

zachberger:
Google Cloud Logging now supports setting the `kmsServiceAccountId` setting at the Project, Folder, and Organization level. This PR updates the `google_logging_folder_settings` and  `google_logging_organization_settings` resource to allow setting this field. A future PR will be opened to create a new `google_logging_project_settings` resource, there is only a datasource right now as until recently there were no modifiable fields  on `google_logging_project_settings`

```release-note:enhancement
`google_logging_folder_settings` and  `google_logging_organization_settings` resources to allow setting `kms_service_account_id`
```

### DIFF
--- a/mmv1/products/logging/FolderSettings.yaml
+++ b/mmv1/products/logging/FolderSettings.yaml
@@ -23,8 +23,8 @@ base_url: 'folders/{{folder}}/settings'
 self_link: 'folders/{{folder}}/settings'
 import_format: ['folders/{{folder}}/settings']
 # Hardcode the updateMask since d.HasChanged does not work on create.
-create_url: 'folders/{{folder}}/settings?updateMask=disableDefaultSink,storageLocation,kmsKeyName'
-update_url: 'folders/{{folder}}/settings?updateMask=disableDefaultSink,storageLocation,kmsKeyName'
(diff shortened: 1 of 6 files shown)


zachberger:
@ScottSuarez your advice would be appreciated here. The property kms_service_account_id can be updated once and only once, and only when its a migration from a legacy Service Account (prefixed with cmek-) to a different Service Account (prefixed with service-). It is not possible to create organizations, folders, or projects configured with that legacy service account, so I'm not sure how to reliably and repeatedly test that within terraform. More details on why this migration is required can be found here [Migrate CMEK service accounts](https://cloud.google.com/logging/docs/routing/troubleshoot-cmek-orgs#migrate-cmek-sa)

ScottSuarez:
> @ScottSuarez your advice would be appreciated here. The property kms_service_account_id can be updated once and only once, and only when its a migration from a legacy Service Account (prefixed with cmek-) to a different Service Account (prefixed with service-). It is not possible to create organizations, folders, or projects configured with that legacy service account, so I'm not sure how to reliably and repeatedly test that within terraform. More details on why this migration is required can be found here [Migrate CMEK service accounts](https://cloud.google.com/logging/docs/routing/troubleshoot-cmek-orgs#migrate-cmek-sa)

I don't believe we should support this at the terraform level if it can done only once and only applies to legacy service accounts. I think that users should set this out of band of terraform in that case.

PR #10058, opened 2024-02-27
Update the GCF resource to reflect transition from Container Registry to Artifact Registry

Ectelion:
The general transition from Container Registry is outlined in [this](https://cloud.google.com/artifact-registry/docs/transition/transition-from-gcr) public doc. The GCF-specific API transition has already completed. The field `docker_registry` was already marked as optional and computed -- no need for any explicit changes to the defaulting logic here, we only update the docs and comments to reflect the new auto-defaulting.

```release-note:note
cloudfunctions: Updated docs for Cloud Function's `docker_registry` field to reflect the change of default from Container Registry to Artifact Registry.
```


rileykarson (approved):

rileykarson:
Running tests to make sure there are no surprises, but lgtm assuming there's none

Ectelion:
@rileykarson thanks!

rileykarson (approved):

PR #10060, opened 2024-02-27
Add `region_from_id` and `zone_from_id` provider-defined functions

SarahFrench:
This PR adds support for 2 new provider-defined functions:
- region_from_id
- zone_from_id

There are unit and acceptance tests for both of these, and documentation

## Note about testing

Currently, testing this PR requires downloading an [alpha release of TF 1.8.0](https://github.com/hashicorp/terraform/releases/tag/v1.8.0-alpha20240216) and running the acceptance test locally. The unit tests should work fine because they don't use a TF binary like acc tests do, and the dependencies on this branch include provider function support.

------

```release-note:enhancement
provider: added provider-defined function `region_from_id` for retrieving the region from a resource's self link or id
```

```release-note:enhancement
provider: added provider-defined function `zone_from_id` for retrieving the zone from a resource's self link or id
```

### DIFF
--- a/mmv1/third_party/terraform/functions/zone_from_id.go
+++ b/mmv1/third_party/terraform/functions/zone_from_id.go
@@ -1,0 +1,59 @@
+package functions
+
+import (
+	"context"
+	"regexp"
+
+	"github.com/hashicorp/terraform-plugin-framework/function"
+)
+
+var _ function.Function = ZoneFromIdFunction{}
+
+func NewZoneFromIdFunction() function.Function {
+	return &ZoneFromIdFunction{}
+}
+
+type ZoneFromIdFunction struct{}
+
+func (f ZoneFromIdFunction) Metadata(ctx context.Context, req function.MetadataRequest, resp *function.MetadataResponse) {
+	resp.Name = "zone_from_id"
+}
+
+func (f ZoneFromIdFunction) Definition(ctx context.Context, req function.DefinitionRequest, resp *function.DefinitionResponse) {
+	resp.Definition = function.Definition{
+		Summary:     "Returns the zone name within the resource id or self link provided as an argument.",
+		Description: "Takes a single string argument, which should be an id or self link of a resource. This function will either return the zone name from the input string or raise an error due to no zone being present in the string. The function uses the presence of \"zones/{{zone}}/\" in the input string to identify the zone name, e.g. when the function is passed the id \"projects/my-project/zones/us-central1-c/instances/my-instance\" as an argument it will return \"us-central1-c\".",
+		Parameters: []function.Parameter{
+			function.StringParameter{
+				Name:        "id",
+				Description: "An id of a resouce, or a self link. For example, both \"projects/my-project/zones/us-central1-c/instances/my-instance\" and \"https://www.googleapis.com/compute/v1/projects/my-project/zones/us-central1-c/instances/my-instance\" are valid inputs",
+			},
+		},
+		Return: function.StringReturn{},
+	}
+}
+
+func (f ZoneFromIdFunction) Run(ctx context.Context, req function.RunRequest, resp *function.RunResponse) {
+	// Load arguments from function call
+	var arg0 string
+	resp.Diagnostics.Append(req.Arguments.GetArgument(ctx, 0, &arg0)...)
+
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	// Prepare how we'll identify zone name from input string
+	regex := regexp.MustCompile("zones/(?P<ZoneName>[^/]+)/") // Should match the pattern below
+	template := "$ZoneName"                                   // Should match the submatch identifier in the regex
+	pattern := "zones/{zone}/"                                // Human-readable pseudo-regex pattern used in errors and warnings
+
+	// Validate input
+	ValidateElementFromIdArguments(arg0, regex, pattern, resp)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	// Get and return element from input string
+	zone := GetElementFromId(arg0, regex, template)
+	resp.Diagnostics.Append(resp.Result.Set(ctx, zone)...)
+}
--- a/mmv1/third_party/terraform/functions/zone_from_id_internal_test.go
+++ b/mmv1/third_party/terraform/functions/zone_from_id_internal_test.go
@@ -1,0 +1,114 @@
+package functions
+
+import (
+	"context"
+	"fmt"
+	"testing"
+
+	"github.com/google/go-cmp/cmp"
+	"github.com/hashicorp/terraform-plugin-framework/attr"
+	"github.com/hashicorp/terraform-plugin-framework/diag"
+	"github.com/hashicorp/terraform-plugin-framework/function"
+	"github.com/hashicorp/terraform-plugin-framework/types"
+	"github.com/hashicorp/terraform-plugin-framework/types/basetypes"
+	"github.com/hashicorp/terraform-provider-google/google/envvar"
+)
+
+func TestFunctionRun_zone_from_id(t *testing.T) {
+	t.Parallel()
+
+	zone := envvar.GetTestZoneFromEnv()
+
+	// Happy path inputs
+	validId := fmt.Sprintf("projects/my-project/zones/%s/networkEndpointGroups/my-neg", zone)
+	validSelfLink := fmt.Sprintf("https://www.googleapis.com/compute/v1/projects/my-project/zones/%s/networkEndpointGroups/my-neg", zone)
+	validOpStyleResourceName := fmt.Sprintf("//compute.googleapis.com/projects/my-project/zones/%s/instances/my-instance", zone)
+
+	// Unhappy path inputs
+	repetitiveInput := fmt.Sprintf("https://www.googleapis.com/compute/v1/projects/my-project/zones/%s/zones/not-this-one/networkEndpointGroups/my-neg", zone)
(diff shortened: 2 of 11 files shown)


SarahFrench:
Update: I'm in the process of adding the ability to test this via TeamCity here: https://github.com/GoogleCloudPlatform/magic-modules/pull/10088

mmv1/third_party/terraform/functions/zone_from_id_internal_test.go
```diff
@@ -0,0 +1,114 @@
...
+			request: function.RunRequest{
+				Arguments: function.NewArgumentsData([]attr.Value{types.StringValue(repetitiveInput)}),
+			},
+			expected: function.RunResponse{
+				Result: function.NewResultData(types.StringValue(zone)),
+				Diagnostics: diag.Diagnostics{
+					diag.NewArgumentWarningDiagnostic(
+						0,
+						ambiguousMatchesWarningSummary,
+						fmt.Sprintf("The input string \"%s\" contains more than one match for the pattern \"regions/{region}/\". Terraform will use the first found match.", repetitiveInput),
```
SarahFrench, line 69:
```suggestion
						fmt.Sprintf("The input string \"%s\" contains more than one match for the pattern \"zones/{zone}/\". Terraform will use the first found match.", repetitiveInput),
```

mmv1/third_party/terraform/functions/zone_from_id_internal_test.go
```diff
@@ -0,0 +1,114 @@
...
+			request: function.RunRequest{
+				Arguments: function.NewArgumentsData([]attr.Value{types.StringValue(invalidInput)}),
+			},
+			expected: function.RunResponse{
+				Result: function.NewResultData(types.StringNull()),
+				Diagnostics: diag.Diagnostics{
+					diag.NewArgumentErrorDiagnostic(
+						0,
+						noMatchesErrorSummary,
+						fmt.Sprintf("The input string \"%s\" doesn't contain the expected pattern \"regions/{region}/\".", invalidInput),
```
SarahFrench, line 84:
```suggestion
						fmt.Sprintf("The input string \"%s\" doesn't contain the expected pattern \"zones/{zone}/\".", invalidInput),
```

mmv1/third_party/terraform/functions/zone_from_id_internal_test.go
```diff
@@ -0,0 +1,114 @@
...
+	"github.com/hashicorp/terraform-plugin-framework/function"
+	"github.com/hashicorp/terraform-plugin-framework/types"
+	"github.com/hashicorp/terraform-plugin-framework/types/basetypes"
+	"github.com/hashicorp/terraform-provider-google/google/envvar"
+)
+
+func TestFunctionRun_zone_from_id(t *testing.T) {
+	t.Parallel()
+
+	zone := envvar.GetTestZoneFromEnv()
```
SarahFrench, line 20: It looks like `envvar.GetTestZoneFromEnv()` is returning an empty string and resulting in a unit test failure - I'll refactor this non-acceptance test to not depend on the environment the test is running in.

SarahFrench: [3ec0caa](https://github.com/GoogleCloudPlatform/magic-modules/pull/10060/commits/3ec0caa38681c8810b02c1d66ce0b6da16a28291)

SarahFrench:
Using the diff from the latest report I've triggered tests in TeamCity here : https://hashicorp.teamcity.com/buildConfiguration/TerraformProviders_GoogleCloud_FEATUREBRANCHPROVIDERFUNCTIONS_MM_GA_GOOGLE_PACKAGE_FUNCTIONS/115271?mode=builds&buildTab=overview&hideTestsFromDependencies=false&hideProblemsFromDependencies=false

c2thorn:
let me know when this is out of draft and ready for review

SarahFrench:
@c2thorn - sorry for the notification spam this PR must have given you. This is now ready for review :) - the unrelated test failures are now resolved by syncing in commits from main.

mmv1/third_party/terraform/functions/zone_from_id_test.go
```diff
@@ -0,0 +1,116 @@
...
+  }
+}
+
+data "google_compute_network" "default" {
+  name = "default"
+}
+
+resource "google_compute_subnetwork" "default" {
+  name          = "%{resource_name}"
+  ip_cidr_range = "10.2.0.0/16"
```
SarahFrench, line 67: CIDRs will cause an issue, so this test should be refactored with a [persistent disk](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/compute_disk) as a better zonal resource example versus `google_compute_network_endpoint_group` below.

SarahFrench: Same change should be made in `testProviderFunction_get_zone_from_resource_self_link`

@c2thorn I'm holding off pushing a new commit with this to avoid blocking review, but this is part of my self review

mmv1/third_party/terraform/website/docs/functions/region_from_id.html.markdown
```diff
@@ -0,0 +1,61 @@
...
+  required_providers {
+    google = {
+      source = "hashicorp/google"
+    }
+  }
+}
+
+# Value is "us-central1"
+output "function_output" {
+  value = provider::google::region_from_id("https://www.googleapis.com/compute/v1/projects/my-project/regions/us-central1/subnetworks/my-subnetwork")
```
c2thorn, line 28: Would it be more useful to users to see this used with a resource reference instead/in addition to? Initially, this seems like the more common use case?

mmv1/third_party/terraform/functions/zone_from_id.go
```diff
@@ -0,0 +1,59 @@
...
+		return
+	}
+
+	// Prepare how we'll identify zone name from input string
+	regex := regexp.MustCompile("zones/(?P<ZoneName>[^/]+)/") // Should match the pattern below
+	template := "$ZoneName"                                   // Should match the submatch identifier in the regex
+	pattern := "zones/{zone}/"                                // Human-readable pseudo-regex pattern used in errors and warnings
+
+	// Validate input
+	ValidateElementFromIdArguments(arg0, regex, pattern, resp)
```
c2thorn, line 51: Where do warnings come into play? Attached to the `resp`?

SarahFrench: Yeah that's correct, the `resp` / `function.RunResponse` sends the warning back to Terraform core alongside the result from the function call.

That's currently correct for terraform-plugin-framework v1.5.0, but v1.6.0 has a breaking change of replacing error handling with diagnostics (errors and warnings) to just errors. My plan is to do that once all provider-defined functions complete on the FEATURE-BRANCH-provider-functions branch.

Just wanted to give you a heads up as any review about the warning behaviours won't be as relevant once that change takes place

mmv1/third_party/terraform/website/docs/functions/region_from_id.html.markdown
SarahFrench: Yeah that's true. I think any example is going to look a bit odd and a bit contrived (why provide a region when provisioning a regional resource and then derive that value you already have?) but it'd be more useful to a user.

I'll push a commit that [updates resources used in acc tests](https://github.com/GoogleCloudPlatform/magic-modules/pull/10060#discussion_r1520323575) and I'll use the acc test config to update the docs too. Thanks!

mmv1/third_party/terraform/functions/zone_from_id.go
c2thorn: Okay, sounds good. As long as we aren't returning an error in the duplicated zone/region scenario, which looking at the PR you are not 👍

mmv1/third_party/terraform/website/docs/functions/region_from_id.html.markdown
c2thorn: Go for it! This looks good from my initial review.

mmv1/third_party/terraform/functions/zone_from_id_test.go
SarahFrench: Addressed in [8a81243](https://github.com/GoogleCloudPlatform/magic-modules/pull/10060/commits/8a81243db5a8cb09b2962b400c5c3ab8dc692cbd) and [1f17ec0](https://github.com/GoogleCloudPlatform/magic-modules/pull/10060/commits/1f17ec0863fd262c251e5152b30e5b6036758b70)

mmv1/third_party/terraform/website/docs/functions/zone_from_id.html.markdown
```diff
@@ -0,0 +1,81 @@
...
+    google-beta = {
+      source = "hashicorp/google-beta"
+    }
+  }
+}
+
+resource "google_compute_disk" "default" {
+  name  = "my-disk"
+  zone  = "us-central1-c"
+}
```
SarahFrench, line 56:
```suggestion
resource "google_compute_disk" "default" {
  # provider argument omitted - provisioning by google or google-beta doesn't impact this example
  name  = "my-disk"
  zone  = "us-central1-c"
}
```

c2thorn (approved):
looks solid to me

SarahFrench:
Thanks! Merging now, & I'll do the [mentioned refactoring](https://github.com/GoogleCloudPlatform/magic-modules/pull/10109) once all functions are on the feature branch

PR #10061, opened 2024-02-27
Add `location_from_id` provider-defined function

SarahFrench:
This PR adds support for a new provider-defined function:
- location_from_id

There are unit and acceptance tests for both of these, and documentation

## Note about testing

Currently, testing this PR requires downloading an [alpha release of TF 1.8.0](https://github.com/hashicorp/terraform/releases/tag/v1.8.0-alpha20240216) and running the acceptance test locally. The unit tests should work fine because they don't use a TF binary like acc tests do, and the dependencies on this branch include provider function support.

I'm working on adding automated testing for the feature branch and provider-defined function PR here: https://github.com/hashicorp/terraform-provider-google/issues/17451

------

```release-note:enhancement
provider: added provider-defined function `location_from_id` for retrieving the location from a resource's self link or id
```

### DIFF
--- a/mmv1/third_party/terraform/functions/location_from_id.go
+++ b/mmv1/third_party/terraform/functions/location_from_id.go
@@ -1,0 +1,59 @@
+package functions
+
+import (
+	"context"
+	"regexp"
+
+	"github.com/hashicorp/terraform-plugin-framework/function"
+)
+
+var _ function.Function = LocationFromIdFunction{}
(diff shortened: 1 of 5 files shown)


BBBmau (approved):
ran test and testacc. All passes and unit tests look good 👍🏼

SarahFrench:
Thanks Mau!

PR #10068, opened 2024-02-27
Prevent duplicate entries in vcr results

trodge:

```release-note:none

```


.ci/magician/vcr/tester.go
```diff
@@ -223,12 +223,13 @@ func (vt *Tester) Run(mode Mode, version provider.Version, testDirs []string) (*
 	logFileName := filepath.Join(vt.baseDir, "testlogs", fmt.Sprintf("%s_test.log", mode.Lower()))
```
shuyama1, line 223: It seems that we're facing one issue that the build log of the replaying after recording will replace the initial replaying build log as the file is named based on the test mode and they're both named "replaying_test.log"

shuyama1 (approved):
Thanks for the quick fix!

Spotted another issue and left comment inline. It shouldn't block this PR tho

PR #10069, opened 2024-02-27
Handwrite sweepers for networkConnectivity hubs and spokes

hao-nan-li:
fixes [17420](https://github.com/hashicorp/terraform-provider-google/issues/17420)
Tested locally and the new sweeper cleans up the resource

```release-note:none

```

### DIFF
--- a/mmv1/third_party/terraform/services/networkconnectivity/resource_network_connectivity_hub_sweeper.go.erb
+++ b/mmv1/third_party/terraform/services/networkconnectivity/resource_network_connectivity_hub_sweeper.go.erb
@@ -1,0 +1,127 @@
+<% autogen_exception -%>
+package networkconnectivity
+
+<% unless version == "ga" -%>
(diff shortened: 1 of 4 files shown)


melinath (approved):

melinath:
LGTM but why is this change needed exactly?

hao-nan-li:
> LGTM but why is this change needed exactly? Like, what was the core change needed?

These resources are not swept properly, and compute networks could not be deleted because they are used by networkConnectivity spokes.

melinath:
@hao-nan-li right, but why were they not getting swept properly? What did you change to make them get swept properly?

hao-nan-li:
> @hao-nan-li right, but why were they not getting swept properly? What did you change to make them get swept properly?

These resources were generated from DCL, and the generated sweeper was not sending the delete request. I've overridden the sweepers.

PR #10074, opened 2024-02-28
Create quota infos datasource

lola98:

```release-note:new-datasource
`google_cloud_quotas_quota_infos`
```


mmv1/third_party/terraform/services/cloudquotas/data_source_google_cloud_quotas_quota_infos.go
```diff
@@ -0,0 +1,217 @@
+// Copyright (c) HashiCorp, Inc.
```
roaks3, line 1: Nit: shouldn't need these lines in this repo

mmv1/third_party/terraform/services/cloudquotas/data_source_google_cloud_quotas_quota_infos.go
```diff
@@ -0,0 +1,217 @@
...
+		Schema: map[string]*schema.Schema{
+			"parent": {
+				Type:     schema.TypeString,
+				Required: true,
+			},
+			"service": {
+				Type:     schema.TypeString,
+				Required: true,
+			},
+			"page_token": {
```
roaks3, line 26: I think we probably want to hide the paging parameters, but I'm not sure how we would want to handle this data source in particular. Since Terraform is declarative, it would make usage weird to somehow get a `next_page_token` from an initial page, and use that as the page_token for another page, until the user presumably has what they need.

If the size is sufficiently small when specifying a `service`, you may be able to just remove these parameters, or you may want to add logic to fetch all pages and return the full result.

mmv1/third_party/terraform/services/cloudquotas/data_source_google_cloud_quotas_quota_infos.go
```diff
@@ -0,0 +1,217 @@
...
+				Computed: true,
+			},
+			"quota_infos": {
+				Type:     schema.TypeList,
+				Computed: true,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"service": {
+							Type:     schema.TypeString,
+							Required: true,
```
roaks3, line 46: I think these should all be `Computed: true`, and none should be `Required: true`

roaks3 (commented):
LGTM, mostly just wondering how we should handle the paging parameters

mmv1/third_party/terraform/services/cloudquotas/data_source_google_cloud_quotas_quota_infos.go
lola98: The response can be long since the command lists all quota under a service, and all dimensions within each quota. If paging parameters is not usually used in terraform, I guess I need to concat all pages and return the full list?

roaks3: I think it would be worth considering how you expect this data source to be used. My assumption is that users will not actually be utilizing hundreds of quota infos in their setup, but perhaps there is a reason? I would more so expect users to need a small subset of infos, probably with some sort of filtering, which you could offer as part of the data source implementation (perhaps dimensions like region?). If you supply sufficient filtering options, you could probably leave the limit at 100 entries and ignore paging.

If this is only being added as a generic data source, then I would probably concat all pages like you mentioned, but I am curious about the scale. Is it >1000 entries for any services? Is it common for services to have a large number of entries?

lola98: This is a generic data source, and our backend doesn't implement filtering. To be consistent with our API, I'd vote for concatenating all the results. I just checked that large services like `compute.googleapis.com` return approximately 100 entries, so concatenation is doable.

mmv1/third_party/terraform/services/cloudquotas/data_source_google_cloud_quotas_quota_infos.go
lola98: Removed.

mmv1/third_party/terraform/services/cloudquotas/data_source_google_cloud_quotas_quota_infos.go
lola98: Oh `compute.googleapis.com` returns approximately 300 entries, but most services should have less than 50 entries or so. Anyways I've remove the pagination parameters and concatenate the results.

mmv1/third_party/terraform/services/cloudquotas/data_source_google_cloud_quotas_quota_infos.go
lola98: Updated to `Required: true`.

mmv1/third_party/terraform/services/cloudquotas/data_source_google_cloud_quotas_quota_infos_test.go
```diff
@@ -30,7 +26,7 @@ func TestAccDataSourceGoogleQuotaInfos_basic(t *testing.T) {
 			{
 				Config: testAccDataSourceGoogleQuotaInfos_basic(context),
 				Check: resource.ComposeTestCheckFunc(
-					resource.TestCheckResourceAttr(resourceName, "quota_infos.#", "9"),
+					resource.TestCheckResourceAttr(resourceName, "quota_infos.#", "319"),
```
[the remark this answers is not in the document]
lola98: Yes removed.

roaks3 (approved):

roaks3 (approved):

PR #10089, opened 2024-02-29
fix forceNew on master_ipv4_cidr_block and private_endpoint_subnetwork

yiyinglovecoding:
Remove forceNew on these two fields as
- cx no longer need to provide `master_ipv4_cidr_block` for creating PSC clusters
- `private_endpoint_subnetwork` field is no more used

```release-note:bug
container: changed `master_ipv4_cidr_block`/`private_endpoint_subnetwork` to `ForceNew: false` on `Cluster`
```

### DIFF
--- a/mmv1/third_party/terraform/services/container/resource_container_cluster.go.erb
+++ b/mmv1/third_party/terraform/services/container/resource_container_cluster.go.erb
@@ -3079,46 +3079,6 @@ func resourceContainerClusterUpdate(d *schema.ResourceData, meta interface{}) er
 
 		log.Printf("[INFO] GKE cluster %s's enable private endpoint has been updated to %v", d.Id(), enabled)
 	}
-	
-	if d.HasChange("private_cluster_config.0.private_endpoint_subnetwork") {
-		subnetwork := d.Get("private_cluster_config.0.private_endpoint_subnetwork").(string)
-		req := &container.UpdateClusterRequest{
-			Update: &container.ClusterUpdate{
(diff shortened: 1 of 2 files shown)


yiyinglovecoding:
@c2thorn Hi kindly bumping as PR is ready for review and need kicking off cloudbuild run

rileykarson:
This should be reverted, I think, as is proposed in #10096. There was a change that impacted Terraform related to these fields, but the mitigation would be to mark fields as default_from_api/ignore_read as appropriate rather than changing the immutability of the field. If the mutability of the field **was** changed, we should capture that in update tests.

Some thinking on ForceNew (Immutable) is in https://googlecloudplatform.github.io/magic-modules/best-practices/#forcenew, while removing ForceNew can reduce the impact of a bad diff the cause of the diff is the bug we need to resolve, and marking an immutable field as non-immutable is a new bug. The initial issue generally isn't resolved either- it's just an unresolvable permadiff afterwards.

c2thorn:
> This should be reverted, I think, as is proposed in #10096. There was a change that impacted Terraform related to these fields, but the mitigation would be to mark fields as default_from_api/ignore_read as appropriate rather than changing the immutability of the field. If the mutability of the field **was** changed, we should capture that in update tests.

agreed. it should be reverted. I should have at least asked for testing on the updatability.

c2thorn:
Reverted. I'll remove the changelog entries downstreams while I'm at it.

PR #10090, opened 2024-02-29
Adding nodeType field in the terraform

shailymittal-github:
Adding nodeType and preciseSizeGb in the Cluster resource in redis.

 * Searched through the [issue tracker](https://github.com/hashicorp/terraform-provider-google/issues) for an open issue that this either resolves or contributes to, commented on it to claim it, and written "fixes {url}" or "part of {url}" in this PR description. If there were no relevant open issues, I opened one and commented that I would like to work on it (not necessary for very small changes).
 * Ensured that all new fields I added that can be set by a user appear in at least one [example](https://github.com/GoogleCloudPlatform/magic-modules/tree/main/mmv1/templates/terraform/examples) (for generated resources) or [third_party test](https://github.com/GoogleCloudPlatform/magic-modules/tree/main/mmv1/third_party/terraform/tests) (for handwritten resources or update tests). \
 * [Generated Terraform providers](https://github.com/GoogleCloudPlatform/magic-modules/blob/main/docs/content/docs/getting-started/generate-providers.md), and ran [make test and make lint](https://googlecloudplatform.github.io/magic-modules/docs/getting-started/run-provider-tests/#run-unit-tests) in the generated providers to ensure it passes unit and linter tests. \
 * [Ran](https://github.com/GoogleCloudPlatform/magic-modules/blob/main/docs/content/develop/run-tests.md) relevant acceptance tests using my own Google Cloud project and credentials (If the acceptance tests do not yet pass or you are unable to run them, please let your reviewer know).
 * Read [Write release notes](https://googlecloudplatform.github.io/magic-modules/contribute/release-notes/) before writing my release note below.
`google_redis_cluster`

```release-note:enhancement
redis: added `node_type` and `precise_size_gb` fields to `google_redis_cluster`
```


shailymittal-github:
Validated the acceptance test in autopush https://screenshot.googleplex.com/82MqY4DekCpVeLX

c2thorn:
/gcbrun

sorry, not sure why build isn't running.

mmv1/third_party/terraform/services/redis/resource_redis_cluster_test.go.erb
```diff
@@ -10,6 +10,34 @@ import (
...
+        name := fmt.Sprintf("tf-test-%d", acctest.RandInt(t))
+
+        acctest.VcrTest(t, resource.TestCase{
+                PreCheck:                 func() { acctest.AccTestPreCheck(t) },
+                ProtoV5ProviderFactories: acctest.ProtoV5ProviderBetaFactories(t),
+                CheckDestroy:             testAccCheckRedisClusterDestroyProducer(t),
+                Steps: []resource.TestStep{
+                        {
+                                // create cluster with replica count 1
+                                Config: createOrUpdateRedisCluster(name /* replicaCount = */, 1 /* shardCount = */, 3, true, /*nodeType = */ "REDIS_STANDARD_SMALL"),
```
c2thorn, line 25: the test is throwing a prevent_destroy error here. maybe setting it to false here would fix? Does it need to be `true`?
```suggestion
                                Config: createOrUpdateRedisCluster(name /* replicaCount = */, 1 /* shardCount = */, 3, false, /*nodeType = */ "REDIS_STANDARD_SMALL"),
```

mmv1/products/redis/Cluster.yaml
```diff
@@ -110,6 +110,20 @@ properties:
...
+    description: |
+      The nodeType for the Redis cluster.
+      If not provided, REDIS_HIGHMEM_MEDIUM will be used as default
+    values:
+      - :NODE_TYPE_UNSPECIFIED
+      - :REDIS_SHARED_CORE_NANO
+      - :REDIS_HIGHMEM_MEDIUM
+      - :REDIS_HIGHMEM_XLARGE
+      - :REDIS_STANDARD_SMALL
+    default_value: :REDIS_HIGHMEM_MEDIUM
```
c2thorn, line 124: we cannot add a new field with a default value within an existing block, as that would be a breaking change. We should set to `default_from_api: true`, which would allow Terraform to default to the API's value. It seems like this would fix some of the tests as well.

c2thorn (changes requested):

mmv1/products/redis/Cluster.yaml
shailymittal-github: makes sense, updated

mmv1/third_party/terraform/services/redis/resource_redis_cluster_test.go.erb
shailymittal-github: This is similar to the current tests now. We are keeping prevent destroy enabled while creating/updating the cluster and disable it in the delete call

shailymittal-github: This is to ensure that when terraform find this extra field in the existing cluster during update, it doesn't accidently delete the resources

c2thorn (approved):

PR #10097, opened 2024-03-01
Remove provider-related caches to address cache issues

SarahFrench:
We keep exceeding the 10GB cache limit for the repo, resulting in workflows stalling for 30min on caching steps and then timing out.

Recent examples:
- https://github.com/GoogleCloudPlatform/magic-modules/actions/runs/8110750120/job/22168743566
- https://github.com/GoogleCloudPlatform/magic-modules/actions/runs/8111707791
- https://github.com/GoogleCloudPlatform/magic-modules/actions/runs/8114921062/job/22181809313

This PR removes caching for the providers, which are the larger items stored in the cache. I've left in caching for things like bundler gems, Maven dependencies, and the cache for tgc.

```release-note:none

```

### DIFF
--- a/.github/workflows/test-tpg.yml
+++ b/.github/workflows/test-tpg.yml
@@ -41,16 +41,6 @@ jobs:
         repository: ${{ github.event.inputs.owner }}/${{ github.event.inputs.repo }}
         ref: ${{ github.event.inputs.branch }}
         fetch-depth: 2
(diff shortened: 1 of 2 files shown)


melinath (approved):
LGTM - it looks like we could make this only restore cache in PRs and do a full restore / save on main, but we can figure that out in a separate PR.

PR #10101, opened 2024-03-03
Suppress private_endpoint_subnetwork when master_ipv4_cidr_block is set

Francis-Liu:
For private flexible clusters, user are still allowed to set `master_ipv4_cidr_block`. In this situation, `private_endpoint_subnetwork` will return a non-empty subnet name, which doesn't match the empty subnet in Terraform config (user can set either cidr block or subnet, or neither, but not both, which is error-checked by API). We should ignore the subnet diff in this case. This PR is intended to fix the same issue #10089 attempted to fix. #10089 was reverted by #10096.

```release-note:bug
container: fixed `google_container_cluster` permadiff when `master_ipv4_cidr_block` is set for a private flexible cluster
```

### DIFF
--- a/mmv1/third_party/terraform/services/container/resource_container_cluster.go.erb
+++ b/mmv1/third_party/terraform/services/container/resource_container_cluster.go.erb
@@ -1684,7 +1684,7 @@ func ResourceContainerCluster() *schema.Resource {
 							Optional:     true,
 							ForceNew:     true,
 							AtLeastOneOf: privateClusterConfigKeys,
-							DiffSuppressFunc: tpgresource.CompareSelfLinkOrResourceName,
+							DiffSuppressFunc: containerClusterPrivateClusterConfigSuppress,
 							Description: `Subnetwork in cluster's network where master's endpoint will be provisioned.`,
 						},
 						"public_endpoint": {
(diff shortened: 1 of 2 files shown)


Francis-Liu:
@SarahFrench a kindly reminder

SarahFrench:
Thanks! This had fallen off my radar

SarahFrench (commented):
Hi @Francis-Liu - the code changes make sense to me, but I'm wondering about what testing can be done to verify this change.

Am I right that prior to this PR's changes a user would provision a private flexible cluster with `master_ipv4_cidr_block` set and on the next apply Terraform would suggest to replace the whole cluster due to the detected diff?

Would you be comfortable adding a new acceptance test or updating an existing one so that it fails prior to this change (due to a non-empty plan) and passes afterwards? [By default the test framework will not expect a non empty plan and will fail](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-testing/helper/resource#TestStep:~:text=//%20ExpectNonEmptyPlan%20can%20be%20set%20to%20true%20for%20specific%20types%20of%20tests%20that%20are%0A%09//%20looking%20to%20verify%20that%20a%20diff%20occurs%0A%09ExpectNonEmptyPlan%20bool) after provisioning something, which suggests we might not have a test that provisions a cluster configured in the way that triggers this problem.

Francis-Liu:
> Am I right that prior to this PR's changes a user would provision a private flexible cluster with master_ipv4_cidr_block set and on the next apply Terraform would suggest to replace the whole cluster due to the detected diff?

Yes. Precisely, the diff is in `private_endpoint_subnetwork`. What's interesting about this PR is, it's preparing for our next API change. Currently, there is no diff in `private_endpoint_subnetwork` with our API, with or without this PR, TF wouldn't suggest to replace the whole cluster. Once this PR is released, we are about to make the change such that the returned `private_endpoint_subnetwork` will be non empty when `master_ipv4_cidr_block` is set.

Francis-Liu:
Let me try to add a acceptance test case as you suggested, to capture potential failures after we make the further API change.

SarahFrench:
Thanks, that'd be useful to have!

I asked in the provider team for advice about how to time this release with the API update.

Next week I'm at a conference so won't be able to review things properly until March 25th onwards. If this PR needs to be reviewed and merged earlier, to enable release on a specific date etc, please ping @/rileykarson. He's got context on this.

SarahFrench:
Following on from above, handing off to our on-call for the next week: @roaks3. Ryan, if this PR doesn't need anything before the 25th please feel free to reassign back to me

Francis-Liu:
@roaks3 I have added the acceptance test. Could you please approve & merge this PR so it goes to release `5.22.0`

roaks3 (approved):

roaks3 (approved):

PR #10106, opened 2024-03-04
Add resource `name_from_id` provider-defined function

SarahFrench:
This is the final PR for adding provider defined functions that enable pulling these values out of identifiers:
- project
- region
- zone
- location
- **(short-form) name**

I've taken into account feedback from other PRs for this PR, e.g. updating docs and acc test resources used.

This function is special as it pulls out the end of an input string starting at the last `/`. There isn't an easy consistent regex to use, versus `/projects/{project}`

```release-note:enhancement
provider: added provider-defined function `name_from_id ` for retrieving the short-form name from a resource's self link or id
```

### DIFF
--- a/mmv1/third_party/terraform/functions/name_from_id.go
+++ b/mmv1/third_party/terraform/functions/name_from_id.go
@@ -1,0 +1,59 @@
+package functions
+
+import (
+	"context"
+	"regexp"
+
+	"github.com/hashicorp/terraform-plugin-framework/function"
+)
+
+var _ function.Function = NameFromIdFunction{}
+
+func NewNameFromIdFunction() function.Function {
+	return &NameFromIdFunction{}
+}
+
+type NameFromIdFunction struct{}
+
+func (f NameFromIdFunction) Metadata(ctx context.Context, req function.MetadataRequest, resp *function.MetadataResponse) {
+	resp.Name = "name_from_id"
+}
+
+func (f NameFromIdFunction) Definition(ctx context.Context, req function.DefinitionRequest, resp *function.DefinitionResponse) {
+	resp.Definition = function.Definition{
+		Summary:     "Returns the short-form name of a resource within a provided resource's id, resource URI, self link, or full resource name.",
+		Description: "Takes a single string argument, which should be a resource's id, resource URI, self link, or full resource name. This function will return the short-form name of a resource from the input string, or raise an error due to a problem with the input string. The function returns the final element in the input string as the resource's name, e.g. when the function is passed the id \"projects/my-project/zones/us-central1-c/instances/my-instance\" as an argument it will return \"my-instance\".",
+		Parameters: []function.Parameter{
+			function.StringParameter{
+				Name:        "id",
+				Description: "A string of a resource's id, resource URI, self link, or full resource name. For example, \"projects/my-project/zones/us-central1-c/instances/my-instance\", \"https://www.googleapis.com/compute/v1/projects/my-project/zones/us-central1-c/instances/my-instance\" and \"//gkehub.googleapis.com/projects/my-project/locations/us-central1/memberships/my-membership\" are valid values",
+			},
+		},
+		Return: function.StringReturn{},
+	}
+}
+
+func (f NameFromIdFunction) Run(ctx context.Context, req function.RunRequest, resp *function.RunResponse) {
+	// Load arguments from function call
+	var arg0 string
+	resp.Diagnostics.Append(req.Arguments.GetArgument(ctx, 0, &arg0)...)
+
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	// Prepare how we'll identify resource name from input string
+	regex := regexp.MustCompile("/(?P<ResourceName>[^/]+)$") // Should match the pattern below
+	template := "$ResourceName"                              // Should match the submatch identifier in the regex
+	pattern := "resourceType/{name}$"                        // Human-readable pseudo-regex pattern used in errors and warnings
+
+	// Validate input
+	ValidateElementFromIdArguments(arg0, regex, pattern, resp)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	// Get and return element from input string
+	name := GetElementFromId(arg0, regex, template)
+	resp.Diagnostics.Append(resp.Result.Set(ctx, name)...)
+}
--- a/mmv1/third_party/terraform/functions/name_from_id_internal_test.go
+++ b/mmv1/third_party/terraform/functions/name_from_id_internal_test.go
@@ -1,0 +1,97 @@
+package functions
+
+import (
+	"context"
+	"fmt"
+	"testing"
+
+	"github.com/google/go-cmp/cmp"
+	"github.com/hashicorp/terraform-plugin-framework/attr"
+	"github.com/hashicorp/terraform-plugin-framework/diag"
+	"github.com/hashicorp/terraform-plugin-framework/function"
+	"github.com/hashicorp/terraform-plugin-framework/types"
+	"github.com/hashicorp/terraform-plugin-framework/types/basetypes"
+)
+
+func TestFunctionRun_name_from_id(t *testing.T) {
+	t.Parallel()
+
+	name := "foobar"
+
+	// Happy path inputs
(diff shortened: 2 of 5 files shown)


SarahFrench:
I can't see checks for running tests etc... /gcbrun

SarahFrench:
Rebased to include the latest commits from main that were synced into the feature branch. This should address VCR failures above.

mmv1/third_party/terraform/functions/name_from_id_test.go
```diff
@@ -0,0 +1,88 @@
...
+				// Can get the project from a resource's id in one step
+				// Uses google_pubsub_topic resource's id attribute with format projects/{{project}}/topics/{{name}}
+				Config: testProviderFunction_get_project_from_resource_id(context),
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestMatchOutput(context["output_name"].(string), nameRegex),
+				),
+			},
+			{
+				// Can get the project from a resource's self_link in one step
+				// Uses google_compute_subnetwork resource's self_link attribute
```
SarahFrench, line 38:
```suggestion
				// Uses google_compute_disk resource's self_link attribute
```

mmv1/third_party/terraform/functions/name_from_id_internal_test.go
```diff
@@ -0,0 +1,97 @@
...
+		},
+		"it returns the expected output value when given a valid OP style resource name input": {
+			request: function.RunRequest{
+				Arguments: function.NewArgumentsData([]attr.Value{types.StringValue(validOpStyleResourceName)}),
+			},
+			expected: function.RunResponse{
+				Result: function.NewResultData(types.StringValue(name)),
+			},
+		},
+		"it returns an error when given input with no submatches": {
```
SarahFrench, line 57: There isn't a test for multiple submatches because you cannot have multiple ends of a string (`$` is used in the regex)

SarahFrench:
Redistributing review as I've used a lot of Shuya's time recently.

@c2thorn could you please review this PR? 🙏 

It's very similar to the region/zone one you previously reviewed, but now we're pulling out the short name of a resource from the long id/URL. The regex for this is more permissive though, so there might be edge cases to test for/add extra validation for.

mmv1/third_party/terraform/functions/name_from_id.go
```diff
@@ -0,0 +1,59 @@
+package functions
+
+import (
+	"context"
+	"regexp"
+
+	"github.com/hashicorp/terraform-plugin-framework/function"
+)
+
+var _ function.Function = NameFromIdFunction{}
```
c2thorn, line 10: Side question: in your experience does "name" usually refer to the "short name"? I think we've seen both long/short name when referencing "name" internally, so sometimes there is confusion. I don't have a lot of references externally for what folks usually mean by "name".

mmv1/third_party/terraform/functions/name_from_id_internal_test.go
c2thorn: Makes sense!

mmv1/third_party/terraform/functions/name_from_id_test.go
```diff
@@ -0,0 +1,88 @@
...
+    }
+  }
+}
+
+resource "google_pubsub_topic" "default" {
+  name = "%{resource_name}"
+}
+
+output "%{output_name}" {
+  value = provider::google::%{function_name}(google_pubsub_topic.default.id)
```
c2thorn, line 64: Provider functions can't imply the provider like resources do? The provider block will always be needed?

c2thorn (approved):
Looks good overall, just some minor questions

mmv1/third_party/terraform/functions/name_from_id_test.go
SarahFrench: That's right, the `provider::google::` syntax cannot be skipped and [it's not recommended to include provider names in function names](https://developer.hashicorp.com/terraform/plugin/best-practices/naming#function-names:~:text=Unlike%20resource%20and%20data%20source%20naming%2C%20function%20names%20should%20NOT%20include%20the%20provider%20name%20since%20the%20configuration%20language%20syntax%20for%20calling%20the%20function%20must%20already%20specify%20the%20provider%20name%20separately).

mmv1/third_party/terraform/functions/name_from_id.go
SarahFrench: This has been confusing to me too.

Here are things that make me think of "name" as referring to the short name of a resource:

1) A guide in the Google provider's docs says ([link](https://registry.terraform.io/providers/hashicorp/google/latest/docs/guides/getting_started#linking-gcp-resources:~:text=Like%20this%20VM%20instance%2C%20nearly%20every%20GCP%20resource%20will%20have%20a%20name%20field.%20They%20are%20used%20as%20a%20short%20way%20to%20identify%20resources%2C%20and%20a%20resource%27s%20display%20name%20in%20the%20Cloud%20Console%20will%20be%20the%20one%20defined%20in%20the%20name%20field.)):

> Like this VM instance, nearly every GCP resource will have a `name` field. They are used as a short way to identify resources, and a resource's display name in the Cloud Console will be the one defined in the `name` field.

Equating display name to `name` makes me think it's definitely referring to the short-form name.

2) Our documentation for the provider shows [examples with `name` (or equivalent) arguments being set like `name = "my-instance"`](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/compute_instance#example-usage). If an API returns a longer value after the resource is created there are often DiffSuppress functions like [`CompareSelfLinkOrResourceName`](https://github.com/hashicorp/terraform-provider-google/blob/099222630bbeac3ede04c7bf05217c3be2f32917/google/tpgresource/self_link_helpers.go#L40-L57) that ensures an end user only sees the name field as being the short-form version.

3) Our codebase includes code that refers to the last bit of a string that contains `/` characters as being the name. As well as [`CompareSelfLinkOrResourceName`](https://github.com/hashicorp/terraform-provider-google/blob/099222630bbeac3ede04c7bf05217c3be2f32917/google/tpgresource/self_link_helpers.go#L40-L57) above there are other helpers like [the `GetResourceNameFromSelfLink` helper function](https://github.com/hashicorp/terraform-provider-google/blob/099222630bbeac3ede04c7bf05217c3be2f32917/google/tpgresource/self_link_helpers.go#L85-L88).



In the past my understanding was that:
- self link = `https://cloudkms.googleapis.com/v1/projects/my-project/locations/us-central1/keyRings/my-key-ring/cryptoKeys/my-crypto-key`
    - [provider docs describe this in this guide](https://registry.terraform.io/providers/hashicorp/google/latest/docs/guides/getting_started#linking-gcp-resources)
- id = `projects/my-project/locations/us-central1/keyRings/my-key-ring/cryptoKeys/my-crypto-key`
    -  `id` attributes for resources are documented to have this format, [e.g. for BigQuery datasets](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/bigquery_dataset#id)
- name = `my-crypto-key`
     - [provider docs describe this in this guide](https://registry.terraform.io/providers/hashicorp/google/latest/docs/guides/getting_started#linking-gcp-resources)


_But_ then Riley referred me to this documentation about[ "full resource names" ](https://google.aip.dev/122#full-resource-names), i.e. `//cloudkms.googleapis.com/v1/projects/my-project/locations/us-central1/keyRings/my-key-ring/cryptoKeys/my-crypto-key`, which adds confusion. Compared to the provider documentation etc I don't know how much impact the AIP docs change what end users think about the concept of a name.

SarahFrench: Sorry for the essay! 😅 If you think that the function name could be named in a more accurate way I'm happy to change it now, before it's released. Perhaps `short_name_from_id`? What do you think?

c2thorn: Thanks for compiling that, it definitely looks like we use `name` in the short form more often than not externally. So I think it is fine to leave it as is!

SarahFrench: Perfect, thanks 😁 If any strong reason to rename the function comes up it'll be relatively easy to create an identical function with a different name and then deprecate the older version in future

PR #10112, opened 2024-03-04
Delete .github/workflows/build-downstream.yml

melinath:
This was part of our initial testing for downstream builds in GHA but is no longer used.

https://github.com/GoogleCloudPlatform/magic-modules/actions/runs/6726253692

```release-note:none

```


melinath:
nvm, this is used as a sub-workflow of `downstreams`

PR #10114, opened 2024-03-04
Use old github token if new tokens are not available

trodge:

```release-note:none

```


melinath (changes requested):
The current fallback system ends up being a little confusing since it's inconsistent / changes behavior somewhat. I'd recommend two new functions:

- `GetTokenOrFallback` (corresponding to os.Getenv - drop-in replacement for `env["GITHUB_TOKEN_VAR"]`)
- `LookupTokenOrFallback` (drop-in replacement for `os.LookupEnv`)

Additionally, `test_tgc_integration.sh` needs to be updated.

.ci/magician/cmd/check_cassettes.go
```diff
@@ -56,6 +55,8 @@ var checkCassettesCmd = &cobra.Command{
 			env[ev] = val
 		}
 
+		env["GITHUB_TOKEN_DOWNSTREAMS"] = githubTokenOrFallback("GITHUB_TOKEN_DOWNSTREAMS")
```
melinath, line 58: Could you use githubTokenOrFallback("GITHUB_TOKEN_DOWNSTREAMS") instead of accessing env later in this function?

.ci/magician/cmd/generate_comment.go
```diff
@@ -71,6 +69,14 @@ var generateCommentCmd = &cobra.Command{
 			env[ev] = val
 		}
 
+		for _, tokenName := range []string{"GITHUB_TOKEN_DOWNSTREAMS", "GITHUB_TOKEN_MAGIC_MODULES"} {
+			val, ok := lookupGithubTokenOrFallback(tokenName)
+			if !ok {
+				fmt.Printf("Did not provide %s or GITHUB_TOKEN environment variable\n", tokenName)
+				os.Exit(1)
+			}
+			env[tokenName] = val
```
melinath, line 78: As written this will fail if GITHUB_TOKEN_DOWNSTREAMS and GITHUB_TOKEN are both unset (even if 
GITHUB_TOKEN_MAGIC_MODULES is set)

```suggestion
			if ok {
				env[tokenName] = val
				break
			}
```

.ci/magician/cmd/test_terraform_vcr.go
```diff
@@ -55,6 +53,15 @@ var testTerraformVCRCmd = &cobra.Command{
 			env[ev] = val
 		}
 
+		for _, tokenName := range []string{"GITHUB_TOKEN_DOWNSTREAMS", "GITHUB_TOKEN_MAGIC_MODULES"} {
+			val, ok := lookupGithubTokenOrFallback(tokenName)
+			if !ok {
+				fmt.Printf("Did not provide %s or GITHUB_TOKEN environment variable\n", tokenName)
+				os.Exit(1)
+			}
+			env[tokenName] = val
```
melinath, line 62:
```suggestion
			if ok {
				env[tokenName] = val
				break
			}
```

melinath (changes requested):
Using lookup everywhere is fine but if you want to preserve the behavior of validating the inputs you'll need to do a separate check after the loop

.ci/magician/cmd/generate_comment.go
```diff
@@ -71,6 +69,13 @@ var generateCommentCmd = &cobra.Command{
 			env[ev] = val
 		}
 
+		for _, tokenName := range []string{"GITHUB_TOKEN_DOWNSTREAMS", "GITHUB_TOKEN_MAGIC_MODULES"} {
+			val, ok := lookupGithubTokenOrFallback(tokenName)
+			if ok {
+				env[tokenName] = val
+				break
```
melinath, line 76: I was wrong - this shouldn't have a break - both need to get set if present.

```suggestion
```

melinath: actually, I may have been wrong altogether and the previous logic was correct, since we _do_ want to require both of these to be set.

.ci/magician/cmd/generate_downstream.go
```diff
@@ -25,6 +25,7 @@ var gdEnvironmentVariables = [...]string{
 var gdTokenEnvironmentVariables = [...]string{
 	"GITHUB_TOKEN_CLASSIC",
 	"GITHUB_TOKEN_DOWNSTREAMS",
+	"GITHUB_TOKEN",
```
melinath, line 28: Is the logic around this correct? Right now the command will error if _any_ of these are not set, so it won't function as a fallback properly. Should we use the lookup with fallback when looping through this variable instead?

.ci/magician/cmd/test_terraform_vcr.go
```diff
@@ -55,6 +53,14 @@ var testTerraformVCRCmd = &cobra.Command{
 			env[ev] = val
 		}
 
+		for _, tokenName := range []string{"GITHUB_TOKEN_DOWNSTREAMS", "GITHUB_TOKEN_MAGIC_MODULES"} {
+			val, ok := lookupGithubTokenOrFallback(tokenName)
+			if ok {
+				env[tokenName] = val
```
melinath, line 59: same as above - the previous logic was correct, since these are both required (or need to have a fallback)

melinath (changes requested):

.ci/magician/cmd/generate_downstream.go
melinath: the logic is already correct

melinath (approved):
LGTM

PR #10118, opened 2024-03-05
Change `advertised_ip_ranges` fields in `google_compute_router_peer` and `google_compute_router` resources to Sets

Maarc-D:
# Description

Linked to https://github.com/hashicorp/terraform-provider-google/issues/9353

```release-note:bug
compute: changed the `advertised_ip_ranges` fields in both `google_compute_router_peer` and `google_compute_router` resources to be sets. This will reduce diffs due to re-ordering.
```

# ToDo

Need to test it to be sure their is no impacts with https://registry.terraform.io/providers/public-cloud-wl/google/5.17.4 from https://github.com/public-cloud-wl/terraform-provider-google/commit/1a4d7e6ecb001cf279c502512133edda167f6f4a

### DIFF
--- a/mmv1/products/compute/Router.yaml
+++ b/mmv1/products/compute/Router.yaml
@@ -150,7 +150,7 @@ properties:
           This enum field has the one valid value: ALL_SUBNETS
         item_type: Api::Type::String  # TODO(#324): enum?
         send_empty_value: true
-      - !ruby/object:Api::Type::Array
+      - !ruby/object:Api::Type::Set
         name: advertisedIpRanges
         description: |
           User-specified list of individual IP ranges to advertise in
--- a/mmv1/third_party/terraform/services/compute/resource_compute_router_peer.go.erb
+++ b/mmv1/third_party/terraform/services/compute/resource_compute_router_peer.go.erb
@@ -113,7 +113,7 @@ Leave this field blank to advertise no custom groups.`,
 				},
 			},
 			"advertised_ip_ranges": {
-				Type:     schema.TypeList,
+				Type:     schema.TypeSet,
 				Optional: true,
 				Description: `User-specified list of individual IP ranges to advertise in
 custom mode. This field can only be populated if advertiseMode
@@ -1095,8 +1095,9 @@ func expandNestedComputeRouterBgpPeerAdvertisedGroups(v interface{}, d tpgresour
 }
 
 func expandNestedComputeRouterBgpPeerAdvertisedIpRanges(v interface{}, d tpgresource.TerraformResourceData, config *transport_tpg.Config) (interface{}, error) {
+	v = v.(*Schema.Set).List()
 	l := v.([]interface{})
-	req := make([]interface{}, 0, len(l))
+	req := make(map[string]interface{}, 0, len(l))
 	for _, raw := range l {
 		if raw == nil {
 			continue


SarahFrench:
Just a heads up - this would be a breaking change for the field ([see our guidance on this here](https://googlecloudplatform.github.io/magic-modules/develop/breaking-changes/breaking-changes/#field-level-breaking-changes)) and would need to be included in an upcoming major release. We [make exceptions if the 'breaking change' fixes something that is broken for all users](https://googlecloudplatform.github.io/magic-modules/develop/breaking-changes/make-a-breaking-change/#in-minor-releases) but I don't think that is the case here.

Maarc-D:
@SarahFrench In fact I would need to update the label because I tested It is not breaking change if the refresh is on nor change detected if there is not dans if refresh is set to false it wil say that it wat to reoder everythin so 1change but on apply everything is fine, tested and approved on my side, I will use my fork in the meaning time this one can be merge

Maarc-D:
> @Maarc-D sorry for the delay, this PR slipped off my radar!
> 
> We'd need to double check what users will experience when upgrading from a previous version of the provider to a version including this change. Even if there's a new diff that's resolved with a single apply, navigating that diff might not work well for people with more complex configurations.
> 
> I'll trigger tests to run: /gcbrun

no problem I putted back as draft because we experienced few cases where this one does not resolve completely the idempotency issue :/ sometime still prompt more changes that it should. No working impact but it is on this cases still hard to make review before apply.

LucaPrete:
Hello @SarahFrench this seems a huge pain for many of our large Google customers. Is there anything we can do to speed up the development/review? Happy to help!

BTW should we start to "undraft" this? :)

Thanks!

SarahFrench (changes requested):
Changing an Array to a Set is a breaking change and will need to be part of a major release. The 6.0.0 major release is being planned currently and this PRs changes could be included then. Should we choose this option?

Alternatively, you can return the field to being an Array/list and then implement a custom flattener function for the field that reorders the API response to match the users configuration and minimise unnecessary diffs. There is new guidance about [how to do this in our contribution guide](https://googlecloudplatform.github.io/magic-modules/develop/permadiff/#list-order) and this PR is an example of the new helper functions being used: https://github.com/GoogleCloudPlatform/magic-modules/pull/10429 . If you choose this option please mark the PR as ready for review once you want feedback!

mmv1/products/compute/Router.yaml
```diff
@@ -150,7 +150,7 @@ properties:
           This enum field has the one valid value: ALL_SUBNETS
         item_type: Api::Type::String  # TODO(#324): enum?
         send_empty_value: true
-      - !ruby/object:Api::Type::Array
+      - !ruby/object:Api::Type::Set
         name: advertisedIpRanges
```
SarahFrench, line 154: This is currently causing a build error - to specify a field as a Set in these yaml files you will need to revert it back to  `!ruby/object:Api::Type::Array` and then add the property `is_set: true` to that field. However if you do that please refer to my previous review where making the change to a Set can only be in a major version.

SarahFrench:
Hi @LucaPrete - My review comments above describes the two routes forward for this PR/the [original issue](https://github.com/hashicorp/terraform-provider-google/issues/9353). @Maarc-D hasn't had enough time to respond to the review yet, but if this PR was to become stale then someone could open another PR informed from the discussion here.

LucaPrete:
Thank you @SarahFrench ! It makes absolutely sense to me.

I'd proceed with both paths in parallel:

* I opened a [new PR](https://github.com/GoogleCloudPlatform/magic-modules/pull/10572) that uses the flatten function. Waiting for the tests to complete. I'll add you in cc.

* @Maarc-D can you please address the comments of @SarahFrench. We can then think to merge this once we cut a new major? @SarahFrench do we have timelines for it (assuming the PR will be complete)?

SarahFrench:
@Maarc-D The 6.0.0 major release is coming up later this year. Do you want to proceed with this PR for that release?

Maarc-D:
> Thank you @SarahFrench ! It makes absolutely sense to me.
> 
> I'd proceed with both paths in parallel:
> 
> * I opened a [new PR](https://github.com/GoogleCloudPlatform/magic-modules/pull/10572) that uses the flatten function. Waiting for the tests to complete. I'll add you in cc.
> * @Maarc-D can you please address the comments of @SarahFrench. We can then think to merge this once we cut a new major? @SarahFrench do we have timelines for it (assuming the PR will be complete)?

I did not get time to deep dive on it with my day to day job stuffs increasing, I seen you've creating your own PR, great, hope it will fix.

But this problem is recurrent on multiple resources I have the same mess with `consumer_accept_lists` on `google_compute_service_attachment`

SarahFrench:
Hi @Maarc-D - please open new GH issue(s) for any problems you're experiencing so that they can go through the triage process. Thanks!

PR #10122, opened 2024-03-05
add billing project id support to firewall endpoint resource

maxi-cit:
Adds support for optional 'description' field and required 'billingProjectId' field to 'google_network_security_firewall_endpoint'

Part of: https://github.com/hashicorp/terraform-provider-google/issues/15779
Fixes: https://github.com/hashicorp/terraform-provider-google/issues/17396

```release-note:breaking-change
networksecurity: added required field `billing_project_id` to `google_network_security_firewall_endpoint` resource. Any configuration without `billing_project_id` specified will cause resource creation fail (beta)
```


mmv1/products/networksecurity/FirewallEndpoint.yaml
```diff
@@ -110,3 +114,8 @@ properties:
     name: 'state'
     description: The current state of the endpoint.
     output: true
+  - !ruby/object:Api::Type::String
+    name: 'billingProjectId'
```
shuyama1, line 118: Will not providing the billing project make the resource creation failed in all cases?
Adding a required field is a breaking change that we'd wait until the next major release, unless without the field specified, all current configurations will fail. See details at https://googlecloudplatform.github.io/magic-modules/develop/breaking-changes/breaking-changes/#field-level-breaking-changes and https://googlecloudplatform.github.io/magic-modules/develop/breaking-changes/make-a-breaking-change/#in-minor-releases

mmv1/products/networksecurity/FirewallEndpoint.yaml
```diff
@@ -84,6 +84,10 @@ properties:
     description: |
       Server-defined URL of this resource.
     output: true
+  - !ruby/object:Api::Type::String
+    name: 'description'
```
shuyama1, line 88: would you mind adding this field in a test?

shuyama1 (changes requested):

mmv1/products/networksecurity/FirewallEndpoint.yaml
maxi-cit: Hey @shuyama1, thanks for the quick feedback. This breaks all tests cases according to [this](https://github.com/hashicorp/terraform-provider-google/issues/17396).
Unfortunately, this field was introduced as a required field after the enhacement PR was merged.

shuyama1: Would you mind updating the test to include this field?

maxi-cit: Hello @shuyama1 I added the field to the basic test so it fixes it.

mmv1/products/networksecurity/FirewallEndpoint.yaml
maxi-cit: I removed this field from this PR because API is not saving this value and causes tests to fail

maxi-cit:
Hello @shuyama1 is there anything else blocking this merge?

shuyama1:
@maxi-cit  Thanks for the ping. The test skipped in VCR. I triggered a independent test run on [TC](https://hashicorp.teamcity.com/buildConfiguration/TerraformProviders_GoogleCloud_GOOGLE_BETA_MMUPSTREAMTESTS_GOOGLEBETA_PACKAGE_NETWORKSECURITY/120767?buildTab=overview&focusLine=NaN&hideTestsFromDependencies=false&hideProblemsFromDependencies=false&expandBuildChangesSection=true&expandBuildDeploymentsSection=false) and will update the result here when test's done (it may take ~20-30 mins).

Overall implementation LGTM. I do have one question regarding the test: Is there any particular reason that we need to use ADC  in this test?

maxi-cit:
Hey @shuyama1, thanks for the update. I used ADC just for a matter of convinience so I could iterate on the test faster

shuyama1:
> Hey @shuyama1, thanks for the update. I used ADC just for a matter of convinience so I could iterate on the test faster

Oh, got it! Then can we remove the provider override and what's modified in the CheckDestroy function? It'd be easier to read for future reference. I've tested it locally and they should be working fine with just `billing_project_id` specified in the resource in our testing environment. Thank you and apologize for the extra work.

maxi-cit:
Hey @shuyama1 I removed the ADC logic could you run the tests, please?

shuyama1 (approved):
LGTM! Triggered skipped tests on [TC](https://hashicorp.teamcity.com/buildConfiguration/TerraformProviders_GoogleCloud_GOOGLE_BETA_MMUPSTREAMTESTS_GOOGLEBETA_PACKAGE_NETWORKSECURITY/120895?buildTab=tests&expandedTest=build%3A%28id%3A120895%29%2Cid%3A2000000000) and tests passed!

Thank you!

PR #10124, opened 2024-03-05
Clarified permissions for a few workflows

melinath:
Brings things a bit more in line with go/github-actions recommendations. I split the repository-documentation action into two because they're fairly simple and I didn't see an easy way to represent the different permissions otherwise.

```release-note:none

```

### DIFF
--- a/.github/workflows/repository-documentation-deploy.yml
+++ b/.github/workflows/repository-documentation-deploy.yml
@@ -5,10 +5,7 @@ permissions: read-all
 on:
   push:
     branches:
-      - main  # Set a branch to deploy
-  pull_request:
-    paths:
-      - 'docs/**'
+      - main
 
 jobs:
   deploy:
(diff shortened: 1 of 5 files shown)


.github/workflows/test-tgc.yml
```diff
@@ -1,9 +1,6 @@
 name: TGC Build and Unit Test
 
-permissions:
-  actions: read
-  contents: read
-  statuses: write
+permissions: read-all
```
ScottSuarez, line 3: If any permission is specified the remaining are set to none
https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs

Just context on why it was written in this way.

ScottSuarez (commented):
If we are separating the deploy and test of the documentation should the targets not be set accordingly.

Notably we only want to deploy in push and test in pull_request.

.github/workflows/repository-documentation-deploy.yml
```diff
@@ -1,4 +1,6 @@
-name: repository-documentation
+name: repository-documentation-deploy
+
+permissions: read-all
 
 on:
   push:
```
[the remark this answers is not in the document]
melinath: fixed in aef3e500b - thanks!

.github/workflows/test-tgc.yml
melinath: Acknowledged - thanks! I think I'd rather switch to read-all for consistency with go/github-docs/security-recommendations and other workflows. Although this pattern is not explicitly discouraged - I don't think it adds any benefit since this is a public repo.

ScottSuarez (approved):

PR #10128, opened 2024-03-05
Post statuses earlier

melinath:
This should help avoid issues like https://github.com/GoogleCloudPlatform/magic-modules/actions/runs/8144704771 where the PR was never updated due to issues during cloning. Since the status update doesn't use repository code, it should be safe to update the status before checking out the repo.

```release-note:none

```


ScottSuarez (approved):

melinath:
🤦  well, the testing on the PR didn't do anything because the workflows run based on `main` so I'll just test them now that it's merged

PR #10130, opened 2024-03-06
Clarify usage of min/max throughput and instances

jashansudan:
Adding more descriptive comments for min/max throughput and instances to make behaviour more clear.
This is to address issues: https://github.com/hashicorp/terraform-provider-google/issues/10244 and https://github.com/hashicorp/terraform-provider-google/issues/15278.

```release-note:Update documentation for vpcaccess min/max instances and throughput

```

### DIFF
--- a/mmv1/products/vpcaccess/Connector.yaml
+++ b/mmv1/products/vpcaccess/Connector.yaml
@@ -106,19 +106,23 @@ properties:
   - !ruby/object:Api::Type::Integer
     name: minThroughput
     description: |
-      Minimum throughput of the connector in Mbps. Default and min is 200.
+      Minimum throughput of the connector in Mbps. Default and min is 200. Refers to the expected throughput when using an e2-micro machine type. 
+      Value must be a multiple of 100 from 200 through 900. Must be lower than the value specified by maxThroughput. If both minThroughput and
+      minInstances are provided, minInstances takes precedence over minThroughput. The use of minThroughput is discouraged in favor of minInstances.
     default_value: 200
     validation: !ruby/object:Provider::Terraform::Validation
       function: 'validation.IntBetween(200, 1000)'
   - !ruby/object:Api::Type::Integer
     name: minInstances
     description: |
-      Minimum value of instances in autoscaling group underlying the connector.
+      Minimum value of instances in autoscaling group underlying the connector. Value must be between 2 and 9, inclusive. Must be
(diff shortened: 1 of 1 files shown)


mmv1/products/vpcaccess/Connector.yaml
```diff
@@ -106,27 +106,34 @@ properties:
   - !ruby/object:Api::Type::Integer
     name: minThroughput
     description: |
-      Minimum throughput of the connector in Mbps. Default and min is 200.
+      Minimum throughput of the connector in Mbps. Default and min is 200. Refers to the expected throughput when using an e2-micro machine type.
+      Value must be a multiple of 100 from 200 through 900. Must be lower than the value specified by maxThroughput. If both minThroughput and
```
roaks3, line 110: Nit: From Terraform's perspective, these fields are snake cased (ie. max_throughput)

roaks3 (approved):
Looks good, but looking over the mentioned issues, I'm wondering why we are avoiding some mechanism to prevent the use of conflicting fields. Is there a reason that we don't want to go in that direction?

roaks3:
Oh I see, so the client-side defaults might be preventing us from validating that only one field is allowed. I don't recall if there is a way for us to tell if the user has specified a `*throughput` value since it has a default, but if we can, that might allow us to prevent the case where both fields are explicitly specified.

The long-term solution here might actually be to remove the default value though, and just set these up as conflicting fields. But, that would be a breaking change we need to make during a major release.

Unless you can quickly find a solution with the default present, I would suggest we merge this change as-is since it offers more clarity, and leave the issues open, potentially adding the `breaking-change` label to them if removing the default is the ideal solution.

jashansudan:
> Oh I see, so the client-side defaults might be preventing us from validating that only one field is allowed. I don't recall if there is a way for us to tell if the user has specified a `*throughput` value since it has a default, but if we can, that might allow us to prevent the case where both fields are explicitly specified.
> 
> The long-term solution here might actually be to remove the default value though, and just set these up as conflicting fields. But, that would be a breaking change we need to make during a major release.
> 
> Unless you can quickly find a solution with the default present, I would suggest we merge this change as-is since it offers more clarity, and leave the issues open, potentially adding the `breaking-change` label to them if removing the default is the ideal solution.

Fair enough, I'll leave the issues open and submit this.

roaks3 (approved):

PR #10134, opened 2024-03-06
Refactor TeamCity Services Diff Check to run as one job

BBBmau:
To reduce the amount of checks that appear in a PR. we reduced the amount of jobs to just require one job.

Doing this required to create a `build-downstream` action.yml since workflows are used in jobs and not steps.

We may want to consider using the action over the workflow to prevent having two `build-downstreams` although this could be something we discuss further.

`build-downstream/action` was used for the TeamCity PR Diff Check and TeamCity Weekly Diff Check 

```release-note:NONE

```


BBBmau:
We'd want to refactor all places where `downstream.yml` workflow is used and replace with the action.yml that's been created in this PR. Only other place is in `downstream.yml` https://github.com/GoogleCloudPlatform/magic-modules/blob/ab2122202ab62d620cbbbc8928078b0782146994/.github/workflows/downstreams.yml#L16-L31

PR #10137, opened 2024-03-07
Fix two lifecycle rules with different no_age value always generates change issue

kautikdk:
Fixes [hashicorp/terraform-provider-google/issues/17314](https://github.com/hashicorp/terraform-provider-google/issues/17314)

```release-note:bug
storage: fixed two or more lifecycle rules with different values of `no_age` field always generates change in `google_storage_bucket` resource.
```

### DIFF
--- a/mmv1/third_party/terraform/services/storage/resource_storage_bucket.go.erb
+++ b/mmv1/third_party/terraform/services/storage/resource_storage_bucket.go.erb
@@ -1237,10 +1237,10 @@ func flattenBucketLifecycle(d *schema.ResourceData, lifecycle *storage.BucketLif
 
 	rules := make([]map[string]interface{}, 0, len(lifecycle.Rule))
 
-	for _, rule := range lifecycle.Rule {
+	for index, rule := range lifecycle.Rule {
 		rules = append(rules, map[string]interface{}{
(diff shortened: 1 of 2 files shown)


hao-nan-li (commented):
Could you briefly explain why this change is necessary?

kautikdk:
Hi @hao-nan-li, As one google cloud storage bucket can have multiple lifecycle rules, `no_age` can be used in all of them. So current code doesn't check `no_age` for all of them rather it reads value from only one rule which is first one. So because of this bug whenever we assign different `no_age` values in different rules, terraform plan/apply always shows change as flattener only take first lifecycle rule's `no_age` value for all lifecycle rules.

hao-nan-li:
> Hi @hao-nan-li, As one google cloud storage bucket can have multiple lifecycle rules, `no_age` can be used in all of them. So current code doesn't check `no_age` for all of them rather it reads value from only one rule which is first one. So because of this bug whenever we assign different `no_age` values in different rules, terraform plan/apply always shows change as flattener only take first lifecycle rule's `no_age` value for all lifecycle rules.

Thanks for the explanation, LGTM.

hao-nan-li (approved):

PR #10139, opened 2024-03-07
New resource: Managed Instance Group Resize Request

mike-callahan:
Fixes https://github.com/hashicorp/terraform-provider-google/issues/17516

```release-note:new-resource
`google_compute_resize_request`
```


mmv1/products/compute/ResizeRequest.yaml
```diff
@@ -0,0 +1,168 @@
...
+    url_param_only: true
+  - !ruby/object:Api::Type::ResourceRef
+    name: "resizeRequest"
+    resource: "ResizeRequest"
+    imports: "name"
+    description: |
+      The name of an existing resize request. The name should conform to RFC1035 or be a resource ID.
+    url_param_only: true
+
+# Resize request parameters sent in the body of the CRUD action
```
askubis, line 112: seeing there is creation timestamp I would consider phrasing this differently (cause we don't send the output params) - unless this is the same for other similar requests.
"Resize request parameters"?

mmv1/products/compute/ResizeRequest.yaml
```diff
@@ -0,0 +1,168 @@
...
+  - !ruby/object:Api::Type::String
+    name: "description"
+    description: |
+      An optional description of this resize-request.
+  - !ruby/object:Api::Type::String
+    name: "zone"
+    description: |
+      The URL of a zone where the resize request is located. Populated only for zonal resize requests.
+    output: true
+  - !ruby/object:Api::Type::Integer
```
askubis, line 133: can we remove the deprecated parameter? I would rather not introduce this, so that we don't have to delete it as a part of breaking change in the future...

mmv1/templates/terraform/custom_delete/mig_resize_request_delete.go.erb
askubis, line 1: this looks like a sync artifact (the compute_mig_resize_request_delete referenced in mmv1/products/compute/ResizeRequest.yaml has same contents)
Please delete one of these files

mmv1/templates/terraform/custom_delete/mig_resize_request_delete.go.erb
```diff
@@ -0,0 +1,60 @@
...
+	return err
+}
+
+// Get delete url
+deleteUrl, err = tpgresource.ReplaceVars(d, config, "{{ComputeBasePath}}projects/{{project}}/zones/{{zone}}/instanceGroupManagers/{{instanceGroupManager}}/resizeRequests/{{resizeRequest}}")
+if err != nil {
+	return err
+}
+
+// Get current state (accepted or not) and delete
```
askubis, line 19: we may need a loop around this, as this may fail with a retryable error

mmv1/templates/terraform/custom_delete/mig_resize_request_delete.go.erb
```diff
@@ -0,0 +1,60 @@
...
+
+	if err != nil {
+		return transport_tpg.HandleNotFoundError(err, d, "ResizeRequest")
+	}
+
+	err = ComputeOperationWaitTime(
+		config, res, project, "Cancelling the resize request", userAgent,
+		d.Timeout(schema.TimeoutDelete))
+	
+	if err != nil {
```
askubis, line 40: and instead of returning here we should  read the resource again to check if the state has changed

askubis (commented):
These comments are about a very rare cornercase when while it's provisioning we can go back to ACCEPTED state

mike-callahan:
bug per askubis:
2024-07-09T11:42:30.388Z [DEBUG] provider.terraform-provider-google-beta: DELETE /compute/beta/projects/estest0/zones/us-central1-a/instanceGroupManagers/a3-dws/resizeRequests/?alt=json HTTP/1.1

mike-callahan:
> bug per askubis: ---[ REQUEST ]--------------------------------------- 2024-07-09T11:42:30.388Z [DEBUG] provider.terraform-provider-google-beta: DELETE /compute/beta/projects/estest0/zones/us-central1-a/instanceGroupManagers/a3-dws/resizeRequests/?alt=json HTTP/1.1

mike-callahan:
if d.Get("state") == "ACCEPTED" 
does not return true

askubis (approved):

mmv1/templates/terraform/examples/compute_mig_resize_request.tf.erb
```diff
@@ -0,0 +1,69 @@
...
+  instance_lifecycle_policy {
+    default_action_on_failure = "DO_NOTHING"
+  }
+
+  wait_for_instances = false
+
+}
+
+resource "google_compute_resize_request" "<%= ctx[:primary_resource_id] %>" {
+  name                   = "<%= ctx[:vars]['resize_request_name'] %>"
```
BBBmau, line 60: Thanks for the initial review @askubis!

Looks like the test isn't passing due to name being empty. In the yaml their isn't an existing variable for `resize_request_name` which is the reason for the error below.

```
Error: Error creating ResizeRequest: googleapi: Error 400: Invalid value for field 'resource.name': ''. Must be a match of regex '(?:[a-z](?:[-a-z0-9]{0,61}[a-z0-9])?)', invalid
        
          with google_compute_resize_request.a3_resize_request,
          on terraform_plugin_test.tf line 60, in resource "google_compute_resize_request" "a3_resize_request":
          60: resource "google_compute_resize_request" "a3_resize_request" {
```

Other failing test isn't clear as this PR doesn't touch it and may just be a flaky test.

mike-callahan: @BBBmau Resolved the resize request name issue. The other test is not owned by our resource so I am not sure what that is.

mike-callahan:
@BBBmau All tests now have passed. Would you be able to take a look?

mmv1/products/compute/ResizeRequest.yaml
```diff
@@ -0,0 +1,378 @@
...
+                        name: "locale"
+                        description: |
+                          The locale used following the specification defined at https://www.rfc-editor.org/rfc/bcp/bcp47.txt. Examples are: "en-US", "fr-CH", "es-MX"
+                      - !ruby/object:Api::Type::String
+                        name: "message"
+                        description: |
+                          The localized error message in the above locale.
+      # Status.lastAttempt
+      - !ruby/object:Api::Type::NestedObject
+        name: "lastAttempt"
```
BBBmau, line 262: seems like their are missing values for tests but a lot of them have `output only` in the description.

Be sure to mark the attributes that are meant to be output only with `output: true` as well as include any attributes that aren't output into tests. Whether that's the test that's already been made or a new test that covers the rest of the missing attributes.

mike-callahan: Thanks. I was under the impression that putting _output: true_ at the status block would inherit all the way down to its childrens objects. Currently all non-output values have tests.

mmv1/products/compute/ResizeRequest.yaml
```diff
@@ -0,0 +1,427 @@
...
+### Delete Method ###
+# Custom delete method to handle resize request cancellations vs. deletions.
+# If a resize request is in the ACCEPTED state, it must be canceled before it can be
+# deleted. If a resize request is NOT in the ACCEPTED state, it can be directly deleted.
+custom_code: !ruby/object:Provider::Terraform::CustomCode
+  custom_delete: templates/terraform/custom_delete/compute_mig_resize_request_delete.go.erb
+
+# If true, code for handling long-running operations is generated along with
+# the resource. If false, that code is not generated.
+autogen_async: false
```
BBBmau, line 57: This can be removed since by default `autogen_async` is set to false. Everything else is good. Once this is addressed and tests pass once more I can approve and merge this. 👍🏼

BBBmau (approved):
Looks great. Thanks for working on this!

PR #10142, opened 2024-03-07
Simplified / clarified approval flow for presubmits

melinath:
Previously, the flow looked like:

Trusted contributor:

1. Commit added to PR triggers membership-checker and community-checker
2. membership-checker triggers presubmit-generate-diffs and approves community-checker
3. community-checker is a no-op for trusted contributors

Untrusted contributor:
1. Commit added to PR triggers membership-checker and community-checker
2. [on /gcbrun] membership-checker approves community-checker
3. community-checker triggers presubmit-generate-diffs

This PR simplifies the flow to always be:

1. Commit added to PR triggers membership-checker and community-checker
2. [automatically or on /gcbrun] membership-checker approves community-checker
3. community-checker triggers presubmit-generate-diffs

```release-note:none

```

### DIFF
--- a/.ci/gcb-contributor-membership-checker.yml
+++ b/.ci/gcb-contributor-membership-checker.yml
@@ -68,6 +68,7 @@ steps:
       - "membership-checker"
       - $_PR_NUMBER
       - $COMMIT_SHA
+      # The following are unused and should be safe to remove at least 2 weeks after this comment was added.
       - $BRANCH_NAME
       - $_HEAD_REPO_URL
       - $_HEAD_BRANCH
--- a/.ci/gcb-contributor-membership-checker.yml
+++ b/.ci/gcb-contributor-membership-checker.yml
@@ -68,11 +68,6 @@ steps:
       - "membership-checker"
       - $_PR_NUMBER
       - $COMMIT_SHA
-      # The following are unused and should be safe to remove at least 2 weeks after this comment was added.
-      - $BRANCH_NAME
-      - $_HEAD_REPO_URL
-      - $_HEAD_BRANCH
-      - $_BASE_BRANCH
 
 availableSecrets:
   secretManager:
--- a/mmv1/third_party/terraform/services/compute/metadata.go.erb
+++ b/mmv1/third_party/terraform/services/compute/metadata.go.erb
@@ -15,7 +15,6 @@ import (
 	transport_tpg "github.com/hashicorp/terraform-provider-google/google/transport"
 )
 
-// force generation
 // Since the google compute API uses optimistic locking, there is a chance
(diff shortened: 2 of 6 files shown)


.ci/gcb-contributor-membership-checker.yml
```diff
@@ -68,6 +68,7 @@ steps:
       - "membership-checker"
       - $_PR_NUMBER
       - $COMMIT_SHA
+      # The following are unused and should be safe to remove at least 2 weeks after this comment was added.
       - $BRANCH_NAME
       - $_HEAD_REPO_URL
       - $_HEAD_BRANCH
```
shuyama1, line 74: Should we remove these now since it should be safe as we have `Args: cobra.RangeArgs(2, 6)` in the checker and change  it  to `Args: cobra.ExactArgs(2)` a few weeks after we make this change? The soak time is needed after we remove the unused args in the configuration file and before we change the check to accept exact two arguments, right?

melinath: I think we need the soak time now because people with open PRs will _immediately_ get the changes to the cloudbuild.yml (but will still have the old magician code that requires all 6 args.) In two weeks it should be safe to remove these from cloudbuild since the new magician code allows either 2 or 6 - and at the same time we can update the magician code to only allow 2, which should be fine since people getting that version of the magician code will also have the cloudbuild.yml that only supplies 2

shuyama1: ah, I'm thinking it's the opposite - open PRs should get old cloudbuild.yml for both membership and community checker but updated magician code since we check in what's in the main for both .ci/ and tools/ folder at https://github.com/GoogleCloudPlatform/magic-modules/blob/main/.ci/gcb-contributor-membership-checker.yml#L33. But I could be wrong

melinath: hmm. No, I think you're right - cloudbuild.yml must come from the PR branch. With the token split, the issue was that the cloudbuild.yml was still supplying the old token, but the code expected the new token already. Thanks for catching this!

melinath: fixed in f53f493fa

mmv1/third_party/terraform/services/compute/metadata.go.erb
```diff
@@ -15,6 +15,7 @@ import (
 	transport_tpg "github.com/hashicorp/terraform-provider-google/google/transport"
 )
 
+// force generation
```
shuyama1, line 18: Do forget to remove this change before merge

shuyama1 (approved):
LGTM, assuming unit tests passed locally, as I'm not sure if we're running them in our current presubmit checks

PR #10143, opened 2024-03-07
Move pre_delete code closer to the request

rileykarson:
`pre_create` runs right before the request, do the same with `pre_delete`. The update actions are both in different places as well (since update is structured as two bodies, standard and custom update) but I'm not fixing that here. For any future readers, consider that a reasonable precedent though :)

This change effectively just moves the function to after determining the final `billing_project` for all current resources.

Several resources send requests in pre_deletes that should use the billing project but use the resource project- those are correctable now. Filed https://github.com/hashicorp/terraform-provider-google/issues/17528. (Well, they were correctable before but would have had a slight bug like the ones below)

Interesting cases, all more correct than before IMO as we now always use the same billingProject as the final delete:

* google-beta/services/firebasedatabase/resource_firebase_database_instance.go
* google-beta/services/netapp/resource_netapp_volume_replication.go
* google-beta/services/privateca/resource_privateca_certificate_authority.go
* google-beta/services/spanner/resource_spanner_instance.go

```release-note:none
```

### DIFF
--- a/mmv1/templates/terraform/resource.erb
+++ b/mmv1/templates/terraform/resource.erb
@@ -1032,7 +1032,6 @@ func resource<%= object.resource_name -%>Delete(d *schema.ResourceData, meta int
 
 <%# If the deletion of the object requires sending a request body, the custom code will set 'obj' -%>
     var obj map[string]interface{}
-<%= lines(compile(pwd + '/' + object.custom_code.pre_delete)) if object.custom_code.pre_delete -%>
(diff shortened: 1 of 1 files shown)


NickElliot (approved):
The diffs confused me for a moment due to it displaying the change as shuffling the `getBillingProject` as opposed to the code of the pre_creates but realized the real shuffle a moment later. LGTM!

PR #10147, opened 2024-03-08
Access Context Manager - make ingress and egress rules immutable

Charlesleonius:
Make Service Perimeter Ingress and Egress policy resources immutable so that updates will be applied by replacing the resources.

Fixes:
[14895](https://github.com/hashicorp/terraform-provider-google/issues/14895) - b/287060819

```release-note:bug
accesscontextmanager: fixed an issue with `access_context_manager_service_perimeter_ingress_policy` and `access_context_manager_service_perimeter_egress_policy` where updates could not be applied after initial creation. Any updates applied to these resources will now involve their recreation. To ensure that new policies are added before old ones are removed, add a `lifecycle` block with `create_before_destroy = true` to your resource configuration alongside other updates.
```


NickElliot (commented):
Are the resources actually immutable? Per the API reference I'm seeing a PATCH command. If a resource recreation needs to be prompted when changing *specific* fields, thats accomplishable via creating a customize_diff command for those fields. But as is if the resource can be changed in-place at all, this is a breaking change and one that does not reflect resource behavior and can not be merged.

Charlesleonius:
> Are the resources actually immutable? Per the API reference I'm seeing a PATCH command. If a resource recreation needs to be prompted when changing _specific_ fields, thats accomplishable via making those fields forcenew. But as is if the resource can be changed in-place at all, this is a breaking change and one that does not reflect resource behavior and can not be merged.

This resource is not technically immutable but because of the way the fine-grained resource logic works currently, it has to be. If its not immutable, the provider will try to compare the updated resource with the existing resource returned from the API. Since there is no unique identifier, it compared field by field which doesn't match when a change has been made. When its immutable it compares the existing state and replaces it with the new one.

NickElliot (approved):
In that case LGTM and aligns with expected behavior for service perimeter sub resources.

PR #10152, opened 2024-03-08
Add apphub.admin role for the organization in the test environment

krishnangopal1810:

```release-note:REPLACEME

```


.ci/infra/terraform/main.tf
```diff
@@ -135,6 +135,12 @@ resource "google_organization_iam_member" "sa_storage_admin" {
   member = google_service_account.sa.member
 }
 
+resource "google_organization_iam_member" "apphub_admin" {
+  billing_account_id = data.google_organization.org.org_id
```
roaks3, line 139: Should be `org_id`

roaks3 (changes requested):
LGTM except small typo.

I've already gone ahead and granted this permission for all 3 test runners in our test org

roaks3 (approved):
Approved (and changes have been made to the test env), but I'm going to keep this unmerged until everyone is fully settled on this approach. It's possible we want to revisit this later and use project-level permissions in some way.

roaks3:
@krishnangopal1810 circling back on this: do you see any reason to change the strategy here in the short-term. I believe we should probably go ahead and merge this unless we expect it to change soon.

krishnangopal1810:
> @krishnangopal1810 circling back on this: do you see any reason to change the strategy here in the short-term. I believe we should probably go ahead and merge this unless we expect it to change soon.

I don't see a reason to change the strategy here. I agree we should merge this change.

roaks3 (approved):

PR #10154, opened 2024-03-09
Add support for Application Data Source

krishnangopal1810:
FIXES https://github.com/hashicorp/terraform-provider-google/issues/17538

```release-note:new-datasource
`google_apphub_application`
```


mmv1/third_party/terraform/services/apphub/data_source_apphub_application_test.go
```diff
@@ -0,0 +1,67 @@
...
+		"random_suffix": acctest.RandString(t, 10),
+	}
+
+	acctest.VcrTest(t, resource.TestCase{
+		PreCheck:                 func() { acctest.AccTestPreCheck(t) },
+		ProtoV5ProviderFactories: acctest.ProtoV5ProviderFactories(t),
+		CheckDestroy:             testAccCheckApphubApplicationDestroyProducer(t),
+		Steps: []resource.TestStep{
+			{
+				Config: testDataSourceApphubApplication_basic(context),
```
c2thorn, line 23: Can we add a check for some of the datasource's fields here? That way we are actually verifying some of the data? something like https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/third_party/terraform/services/pubsub/data_source_pubsub_topic_test.go#L24

c2thorn (changes requested):

c2thorn:
@krishnangopal1810 looks like there's a merge conflict

mmv1/third_party/terraform/services/apphub/data_source_apphub_application_test.go
krishnangopal1810: Done

mmv1/third_party/terraform/website/docs/d/apphub_application.html.markdown
```diff
@@ -0,0 +1,164 @@
...
+
+The following arguments are supported:
+
+* `project` - The host project of the application.
+* `application_id` - (Required) The user-defined identifier of the application.
+* `location` - (Required) The location of the application.
+
+## Attributes Reference
+
+In addition to the arguments listed above, the following computed attributes are exported:
```
c2thorn, line 33: Instead of listing out each attribute from the original resource, can we just refer to the resource's documentation like this: https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/third_party/terraform/website/docs/d/sourcerepo_repository.html.markdown?plain=1#L31

c2thorn (changes requested):

mmv1/third_party/terraform/website/docs/d/apphub_application.html.markdown
c2thorn: this will require fewer updates in the future and protect against drift.

krishnangopal1810: Done

c2thorn (approved):

mmv1/third_party/terraform/services/apphub/data_source_apphub_application_test.go
```diff
@@ -0,0 +1,70 @@
+package apphub_test
+
+import (
+	"testing"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
+	"github.com/hashicorp/terraform-provider-google/google/acctest"
+)
+
+func TestDataSourceApphubApplication_basic(t *testing.T) {
```
shuyama1, line 10: I don't think this test is ever triggered during presubmit or nightly. This because we are only running acceptance tests with `TestAcc` prefix. Probably need to rename the test to something like `TestAccDataSourceApphubApplication_basic`

PR #10107, opened 2024-03-04
Data source discovered workload

praseedhaPK:
Add data source for apphub discovered workload to automate querying and facilitate its use while registering workloads.
FIXES https://github.com/hashicorp/terraform-provider-google/issues/17474

```
google_apphub_discovered_workload
```

PR #10155, opened 2024-03-11
Add new resource Workload for Apphub

eshagoel06:
Add resource for registering identified Workload to the Application
FIXES https://github.com/hashicorp/terraform-provider-google/issues/17539

Based on other PR:
1. https://github.com/GoogleCloudPlatform/magic-modules/pull/10107

```release-note:new-resource
`google_apphub_workload`
```


mmv1/products/apphub/Workload.yaml
```diff
@@ -0,0 +1,241 @@
...
+  It registers identified workload to the Application.'
+autogen_async: true
+custom_code: !ruby/object:Provider::Terraform::CustomCode
+  constants: 'templates/terraform/constants/apphub_discovered_resource.go.erb'
+examples:
+  - !ruby/object:Provider::Terraform::Examples
+    name: "workload_basic"
+    pull_external: true
+    primary_resource_id: "example"
+    config_path: "templates/terraform/examples/apphub_workload_basic.tf.erb"
```
roaks3, line 32: Nit: if you use `name: "apphub_workload_basic"` you should be able to remove `config_path`

mmv1/products/apphub/Workload.yaml
```diff
@@ -0,0 +1,241 @@
...
+      org_id: :ORG_ID
+      billing_account: :BILLING_ACCT
+properties:
+  - !ruby/object:Api::Type::String
+    name: name
+    output: true
+    description: "Identifier. The resource name of the Workload. Format:\n\"projects/{host-project-id}/locations/{location}/applications/{application-id}/workloads/{workload-id}\" "
+  - !ruby/object:Api::Type::String
+    name: displayName
+    description: 'Optional. User-defined name for the Workload. '
```
roaks3, line 73: You don't need to include `Optional` or `Required` in any of these descriptions; the generator will add that for you.

mmv1/products/apphub/Workload.yaml
```diff
@@ -0,0 +1,241 @@
...
+    name: description
+    description: 'Optional. User-defined description of a Workload. '
+  - !ruby/object:Api::Type::NestedObject
+    properties:
+      - !ruby/object:Api::Type::String
+        name: uri
+        description: 'Output only. The underlying compute resource uri. '
+        output: true
+    output: true
+    name: workloadReference
```
roaks3, line 84: Nit: we usually put the `name` and other field properties at the top, before listing out nested `properties`. It makes it a little easier to read.

mmv1/products/apphub/Workload.yaml
```diff
@@ -0,0 +1,241 @@
...
+    description: 'Reference of an underlying compute resource represented by the Workload. '
+  - !ruby/object:Api::Type::NestedObject
+    properties:
+      - !ruby/object:Api::Type::String
+        name: gcpProject
+        description: "Output only. The service project identifier that the underlying
+          cloud resource\nresides in. Empty for non cloud resources. "
```
roaks3, line 91: Nit: we probably don't need the `\n` in these descriptions

mmv1/products/apphub/Workload.yaml
```diff
@@ -0,0 +1,241 @@
...
+              - :DEVELOPMENT
+        name: environment
+        description: 'Environment of the Application, Service, or Workload '
+      - !ruby/object:Api::Type::Array
+        name: developerOwners
+        item_type: !ruby/object:Api::Type::NestedObject
+          properties:
+            - !ruby/object:Api::Type::String
+              name: displayName
+              description: 'Optional. Contact''s name. '
```
roaks3, line 147: I'm curious how this renders with the apostophes, but it may need to be cleaned up with a `\`? Same below

mmv1/products/apphub/Workload.yaml
```diff
@@ -0,0 +1,241 @@
...
+  - !ruby/object:Api::Type::String
+    name: updateTime
+    description: 'Output only. Update time. '
+    output: true
+  - !ruby/object:Api::Type::String
+    name: uid
+    description: "Output only. A universally unique identifier (UUID) for the `Workload`
+      in the UUID4\nformat. "
+    output: true
+  - !ruby/object:Api::Type::Enum
```
roaks3, line 193: For output-only enums, we usually define them as strings, so that we don't need to keep a client-side list of possible values.

mmv1/products/apphub/Workload.yaml
```diff
@@ -0,0 +1,241 @@
...
+    name: state
+    description: "Output only. Workload state. \n Possible values:\n STATE_UNSPECIFIED\nCREATING\nACTIVE\nDELETING\nDETACHED"
+    output: true
+    values:
+      - :STATE_UNSPECIFIED
+      - :CREATING
+      - :ACTIVE
+      - :DELETING
+      - :DETACHED
+parameters:
```
roaks3, line 203: Nit: `parameters` and the other fields below are normally listed before `properties`

mmv1/products/apphub/Workload.yaml
```diff
@@ -0,0 +1,241 @@
...
+  - !ruby/object:Api::Type::String
+    name: location
+    description: 'Part of `parent`. See documentation of `projectsId`. '
```
roaks3, line 206: I would probably change the description here and for `applicationsId`: the user has no knowledge of a "parent" or a fields called "projectsId"

mmv1/products/apphub/Workload.yaml
```diff
@@ -0,0 +1,241 @@
...
+    url_param_only: true
+    required: true
+    immutable: true
+  - !ruby/object:Api::Type::String
+    name: applicationsId
```
roaks3, line 211: I'm not sure if this is intended or a typo. Would we want `applicationId` instead?

roaks3 (commented):
I did a first pass on the yaml and left some comments. Looks like tests pass, so should just be some small tweaks.

roaks3:
Looks like there are conflict after some of the other work was merged.

Question on the changes to `data_source_apphub_discovered_workload.go`: these would be breaking changes, unless the data source has not been released yet. Do you know if that has happened?

roaks3:
Oh nevermind, I see that this is based on those changes.

mmv1/products/apphub/Workload.yaml
eshagoel06: We've verified this. It renders as expected.

mmv1/products/apphub/Workload.yaml
```diff
@@ -0,0 +1,230 @@
...
+        name: zone
+        description: "Output only. The location that the underlying compute resource resides
+          in if it is zonal (e.g us-west1-a). "
+        output: true
+    output: true
+    name: workloadProperties
+    description: 'Properties of an underlying compute resource represented by the Workload. '
+  - !ruby/object:Api::Type::String
+    name: discoveredWorkload
+    diff_suppress_func: 'DiscoveredResourceProjectNumberDiffSuppress'
```
roaks3, line 145: I believe this can just use `diff_suppress_func: tpgresource.ProjectNumberDiffSuppress`. Did you try that? Then you could remove the custom code file.

eshagoel06: We checked that out. But our format doesn't match with that function. Please refer to https://github.com/GoogleCloudPlatform/magic-modules/pull/10132#discussion_r1517505047

mmv1/third_party/terraform/services/apphub/resource_apphub_workload_test.go
```diff
@@ -13,7 +13,6 @@ func TestAccApphubWorkload_apphubWorkloadUpdate(t *testing.T) {
 	t.Parallel()
 
 	context := map[string]interface{}{
-		"org_id":          envvar.GetTestOrgFromEnv(t),
```
roaks3: I'm going to do the gcbrun, but you may still need this line

krishnangopal1810: I unintentionally removed it. Adding back

roaks3 (approved):

PR #10158, opened 2024-03-11
Refactor TeamCity Services Diff Check to run as one job

BBBmau:
To reduce the amount of checks that appear in a PR. we reduced the amount of jobs to just require one job.

Doing this required to create a `build-downstream` action.yml since workflows are used in jobs and not steps.

We may want to consider using the action over the workflow to prevent having two `build-downstreams` although this could be something we discuss further.

`build-downstream/action` was used for the TeamCity PR Diff Check and TeamCity Weekly Diff Check 

We'd want to refactor all places where `downstream.yml` workflow is used and replace with the action.yml that's been created in this PR. Only other place is in `downstream.yml`, doing this would shorten it to just be one job instead of four. This can be done in a second PR. https://github.com/GoogleCloudPlatform/magic-modules/blob/ab2122202ab62d620cbbbc8928078b0782146994/.github/workflows/downstreams.yml#L16-L31

```release-note:NONE

```

### DIFF
--- a/.github/actions/build-downstream/action.yml
+++ b/.github/actions/build-downstream/action.yml
@@ -1,0 +1,104 @@
+name: build
+description: Generates Provider
+
+
+runs:
+    using: "composite"
+    steps:
+      - name: Checkout Repository
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - name: Set up Ruby
+        uses: ruby/setup-ruby@036ef458ddccddb148a2b9fb67e95a22fdbf728b # v1.160.0
+        with:
+          ruby-version: '3.1'
+
+      - name: Cache Bundler gems
+        uses: actions/cache@v3
+        with:
+          path: mmv1/vendor/bundle
+          key: ${{ runner.os }}-gems-${{ hashFiles('mmv1/**/Gemfile.lock') }}
+          restore-keys: |
+            ${{ runner.os }}-gems-
+
+      - name: Install Ruby dependencies
+        shell: bash
+        run: |
+          bundle config path mmv1/vendor/bundle
+          bundle install
+        working-directory: mmv1
+
+      - name: Set up Go
+        uses: actions/setup-go@v4
+        with:
+          go-version: '^1.20'
+
+      # Cache Go modules
+      - name: Cache Go modules
+        uses: actions/cache@v3
+        with:
+          path: ~/go/pkg/mod
+          key: ${{ runner.os }}-go-${{ hashFiles('**/go.sum') }}
+          restore-keys: |
+            ${{ runner.os }}-go-
+
+      - run: go install golang.org/x/tools/cmd/goimports@latest
+        shell: bash
+      - name: Build ${{ inputs.repo  }}
+        shell: bash
+        env:
+          BASE_BRANCH: ${{ github.event.pull_request.base.ref || github.ref_name }}
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          set -e
+          set -x
+          # Set GOPATH to a directory the runner user has access to
+          export GOPATH=~/go
+
+          function clone_repo() {
+            export OUTPUT_PATH=$GOPATH/src/github.com/$UPSTREAM_OWNER/$GH_REPO
+            GITHUB_PATH=https://x-access-token:$GITHUB_TOKEN@github.com/$UPSTREAM_OWNER/$GH_REPO
+            mkdir -p "$(dirname $OUTPUT_PATH)"
+            git clone $GITHUB_PATH $OUTPUT_PATH --branch $BASE_BRANCH
+          }
+
+          GH_REPO="${{ inputs.repo  }}"
+          if [ "$GH_REPO" == "docs-examples" ] && [ "$BASE_BRANCH" == "main" ]; then
+              BASE_BRANCH="master"
+          fi
+
+          GITHUB_PATH=https://x-access-token:$GITHUB_TOKEN@github.com/$UPSTREAM_OWNER/$GH_REPO
+
+          if [[ "$GH_REPO" == terraform-provider-google* ]]; then
+            UPSTREAM_OWNER=hashicorp
+            clone_repo
+          if [ "$GH_REPO" == "terraform-provider-google" ]; then
+              export VERSION=ga
+            else
+              export VERSION=beta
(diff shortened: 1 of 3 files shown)


.github/actions/build-downstream/action.yml
```diff
@@ -0,0 +1,110 @@
...
+            exit 1
+          fi
+
+          (current_dir=$(pwd) && cd $OUTPUT_PATH && zip -r "$current_dir/output.zip" .)
+
+      - name: Upload built artifacts
+        uses: actions/upload-artifact@3cea5372237819ed00197afe530f5a7ea3e805c8 # v3.1.0
+        with:
+          name: artifact-${{ inputs.repo  }}
+          path: output.zip
```
SarahFrench, line 110: Now that this is an action and can be performed as a step within a job (vs before where it was a job within a workflow like .github/workflows/teamcity-services-diff-check.yml) do we need to upload the built provider as an artifact? I believe artifacts are only required to [share data between jobs in a workflow](https://docs.github.com/en/rest/actions/artifacts?apiVersion=2022-11-28#:~:text=Artifacts%20enable%20you%20to%20share%20data%20between%20jobs%20in%20a%20workflow) (or if you want to persist data to reference in the future).

.github/workflows/teamcity-services-diff-check-weekly.yml
```diff
@@ -11,57 +11,60 @@ on:
...
+        - name: Download built artifacts - Beta provider
+          uses: actions/download-artifact@v2
+          with:
+            name: artifact-terraform-provider-google-beta
+            path: artifacts
+    
+        - name: Unzip the artifacts and delete the zip
+          run: |
+              unzip -o artifacts/output.zip -d ./provider
+              rm artifacts/output.zip
```
SarahFrench, line 59: I don't think you'd need to download artifacts (see other comment) as I believe the steps of the `teamcity-services-diff-check` job would all run on the same runner? Wouldn't we be using the same machine to build the providers, upload them, download them, and then reference them in later steps?

SarahFrench (commented):
Left some comments wondering if there are some unnecessary steps we can remove now that the `teamcity-services-diff-check` jobs are a single set of steps, and aren't dependent on other jobs in a workflow anymore

.github/actions/build-downstream/action.yml
BBBmau: Yep you're right. Since for the diff check we keep it in just one job there's no need to use artifact, it'll stay in the same runner like you said. Good catch!

BBBmau:
> Left some comments wondering if there are some unnecessary steps we can remove now that the `teamcity-services-diff-check` jobs are a single set of steps, and aren't dependent on other jobs in a workflow anymore

I'll apply the changes later today and request a review once ready.

BBBmau:
Did a second look at 'downstream.yaml', the actions.yml could still be used there but this would require the unit-tests to be all in one job instead of just multiple jobs.

Thoughts on this? It looks like there's improvements that can be done since we don't generate the providers in parallel in 'downstreams.yml'

SarahFrench:
> Thoughts on this? It looks like there's improvements that can be done since we don't generate the providers in parallel in 'downstreams.yml'

The TeamCity-related GHA is, I think, the only time that we generate both in parallel an that is needed. I'd recommend not refactoring anything not related to the TeamCity-related GHA in this PR and treating that as a separate piece of work.

BBBmau:
TeamCity Weekly Test Passing: https://github.com/BBBmau/magic-modules/actions/runs/8241230001/job/22538062627

Team City PR Diff Check Failing (Due to New Service): https://github.com/BBBmau/magic-modules/actions/runs/8241306698/job/22538297815?pr=19

BBBmau:
Removing the use of artifacts resulted in needing to grab the `OUTPUT_PATH` of where the provider was being generated. This done with `- run: echo "GOOGLE_REPO_PATH=${{ env.OUTPUT_PATH}}" >> $GITHUB_ENV` where `env.OUTPUT_PATH` is grabbed from within the `action.yml` file.

along with artifacts being removed I also removed where `setup-go` is called in both diff check workflows since the action.yml already does the setup for go. The extra setup in the workflows is redundant.

.github/workflows/teamcity-services-diff-check-weekly.yml
```diff
@@ -11,57 +11,32 @@ on:
...
+          with:
+            repo: 'terraform-provider-google'
+            token: '$GITHUB_TOKEN'
+        - run: echo "GOOGLE_REPO_PATH=${{ env.OUTPUT_PATH}}" >> $GITHUB_ENV
+        - name: TeamCity Google Beta Provider Generate
+          uses: ./.github/actions/build-downstream
+          with:
+            repo: 'terraform-provider-google-beta'
+            token: '$GITHUB_TOKEN'
+        - run: echo "GOOGLE_BETA_REPO_PATH=${{ env.OUTPUT_PATH}}" >> $GITHUB_ENV
```
SarahFrench, line 32: Could you please add a name to the two echo command steps? Also, it'd be helpful if the name (or a comment) explained how these steps are capturing the value of `OUTPUT_PATH` after `./.github/actions/build-downstream` changes that value.

Currently the names are just the script:
Screenshot 2024-03-12 at 18 09 25

.github/actions/build-downstream/action.yml
```diff
@@ -0,0 +1,105 @@
+name: build
+description: Generates Google/Google-Beta Provider
```
SarahFrench, line 2: It'd be worth rewording this to be a bit clearly. Currently what this generates depends on the input `repo`, and the description implies it could only create TPG or TPGB but you could generate `terraform-google-conversion`.

It could be something like "Generates a given downstream repository from Magic Modules, controlled by the actions inputs"

.github/workflows/teamcity-services-diff-check-weekly.yml
SarahFrench: Same for .github/workflows/teamcity-services-diff-check.yml

SarahFrench (commented):
This looks good! Comments are only around clarity and user-experience of this code.

.github/actions/build-downstream/action.yml
BBBmau: Good point, fixed and should be good for merging!

SarahFrench (approved):
Awesome, thanks!

PR #10164, opened 2024-03-11
Fix generate comment diff processor build failure

melinath:
This PR makes the generate_comments script more resilient against errors at various steps, and reports any errors encountered (in a general sense) as part of the diff comment. It also switches to rendering the diff comment as a go template and moves the comment post to the end of the task (so that it can include errors encountered when trying to run the missing test detector.)

```release-note:none

```


melinath:
rerunning now that GH is back online /gcbrun

trodge (approved):

PR #10166, opened 2024-03-11
Update `location_from_id` acceptance test to not use GOOGLE_REGION ENV

SarahFrench:
This PR makes `TestAccProviderFunction_location_from_id` not use `GOOGLE_REGION` to get a region value.

I won't update `TestAccProviderFunction_project_from_id` in a similar way because the acceptance test cannot run without GOOGLE_PROJECT set.

```release-note:none

```


BBBmau (approved):

SarahFrench:
Thanks Mau!

PR #10168, opened 2024-03-12
Run Firestore index tests in fresh databases.

pcostell:
This removes the index dependency on creating a document, as creating indexes on a non-existent collection is a valid operation in Firestore.

This removes the dependency on a fixed (default) database, which should deflake tests.

Fixes https://github.com/hashicorp/terraform-provider-google/issues/17393.

Verified with a local execution of `make testacc TEST=./google/services/f
irestore TESTARGS='-run=TestAccFirestoreIndex_'`

```release-note:none
```

### DIFF
--- a/mmv1/templates/terraform/examples/firestore_index_basic.tf.erb
+++ b/mmv1/templates/terraform/examples/firestore_index_basic.tf.erb
@@ -1,45 +1,17 @@
-resource "google_project" "project" {
-  project_id = "<%= ctx[:vars]['project_id'] %>"
-  name       = "<%= ctx[:vars]['project_id'] %>"
-  org_id     = "<%= ctx[:test_env_vars]['org_id'] %>"
-}
-
-resource "time_sleep" "wait_60_seconds" {
-  depends_on = [google_project.project]
-
-  create_duration = "60s"
-}
-
-resource "google_project_service" "firestore" {
-  project = google_project.project.project_id
-  service = "firestore.googleapis.com"
-
-  # Needed for CI tests for permissions to propagate, should not be needed for actual usage
-  depends_on = [time_sleep.wait_60_seconds]
-}
-
 resource "google_firestore_database" "database" {
-  project     = google_project.project.project_id
-  name        = "(default)"
+  project     = "<%= ctx[:test_env_vars]['project_id'] %>"
+  name        = "<%= ctx[:vars]['database_id'] %>"
   location_id = "nam5"
   type        = "FIRESTORE_NATIVE"
 
-  depends_on = [google_project_service.firestore]
-}
-
-# Creating a document also creates its collection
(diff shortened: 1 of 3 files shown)


pcostell:
Including @rwhogg as a Firestore expert reviewer.

slevenick (approved):
Added @rwhogg as a reviewer, I'm fine with this change as-is, but want to confirm that!

mmv1/templates/terraform/examples/firestore_index_basic.tf.erb
```diff
@@ -1,45 +1,17 @@
...
-  service = "firestore.googleapis.com"
-
-  # Needed for CI tests for permissions to propagate, should not be needed for actual usage
-  depends_on = [time_sleep.wait_60_seconds]
-}
-
 resource "google_firestore_database" "database" {
-  project     = google_project.project.project_id
-  name        = "(default)"
+  project     = "<%= ctx[:test_env_vars]['project_id'] %>"
```
rwhogg, line 2: I'm a little surprised this works given that you removed the `test_env_vars` section from the YAML

mmv1/templates/terraform/examples/firestore_index_basic.tf.erb
```diff
@@ -1,45 +1,17 @@
...
-}
-
-# Creating a document also creates its collection
-resource "google_firestore_document" "document" {
-  project     = google_project.project.project_id
-  database    = google_firestore_database.database.name
-  collection  = "somenewcollection"
-  document_id = "<%= ctx[:vars]['document_id'] %>"
-  fields      = "{\"something\":{\"mapValue\":{\"fields\":{\"akey\":{\"stringValue\":\"avalue\"}}}}}"
+  delete_protection_state = "<%= ctx[:vars]['delete_protection_state'] %>"
```
rwhogg, line 7: This field wasn't added to the `vars` in the YAML

pcostell: Realize we just want this to be disabled anyways, making it static.

rwhogg (approved):

rwhogg (changes requested):
Sorry, approved the wrong PR. Haven't had time to look at the changes yet

mmv1/templates/terraform/examples/firestore_index_basic.tf.erb
pcostell: I'm guessing the keys may just always be present?

The only combo I tried that didn't work was using [:vars]['project_id'], which failed because it used a test project that I don't have access to.

I cleaned this up so the yaml and erb files _should_ align. (but please double check me)

rwhogg (approved):
Thanks a bunch!

slevenick (approved):
Thanks!

PR #10170, opened 2024-03-12
Allow in-place column drop for bigquery table

obada-ab:
Partial fix for https://github.com/hashicorp/terraform-provider-google/issues/14769

Column deletion reference: https://cloud.google.com/bigquery/docs/managing-table-schemas#delete_a_column

```release-note:enhancement
bigquery: added in-place column drop support for `bigquery_table`
```


hao-nan-li (commented):
Could you briefly explain why this change is necessary?

mmv1/third_party/terraform/services/bigquery/resource_bigquery_table_internal_test.go
```diff
@@ -550,14 +570,18 @@ var testUnitBigQueryDataTableIsChangableTestCases = []testUnitBigQueryDataTableJ
 
 func TestUnitBigQueryDataTable_schemaIsChangable(t *testing.T) {
```
wj-chen, line 571: Do I understand correctly that this test is testing two different things - isChangeable when the schema is at the top level and when it's nested? Can we split these two cases?

Also, not related to your change but "Changable" should be "Changeable".

wj-chen (changes requested):

obada-ab:
> Could you briefly explain why this change is necessary?

@hao-nan-li many customers are having a hard time managing BigQuery tables when applying some types of changes to the `schema` field that can happen in-place, including column drops which are addressed in this PR. The current behavior of the BigQuery table resource is to force-new when a column is dropped, which causes all existing table data to be removed.

More info: https://github.com/hashicorp/terraform-provider-google/issues/14769 b/298496481

mmv1/third_party/terraform/services/bigquery/resource_bigquery_table.go
```diff
@@ -289,16 +286,21 @@ func resourceBigQueryTableSchemaIsChangeable(old, new interface{}) (bool, error)
...
 			}
 			if isChangable, err :=
-				resourceBigQueryTableSchemaIsChangeable(mapOld[key], mapNew[key]); err != nil || !isChangable {
+				resourceBigQueryTableSchemaIsChangeable(mapOld[key], mapNew[key], false); err != nil || !isChangable {
 				return false, err
 			}
 		}
-		return true, nil
+		// only pure column drops allowed
+		return (droppedColumns == 0) || (len(arrayOld) == len(arrayNew)+droppedColumns), nil
```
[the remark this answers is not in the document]
obada-ab: I think "changeableColumns" would be more precise since changing a column's mode would be allowed alongside column dropped, let me know if you think we should change this to only column drops without other in-place changes (besides adding columns)

> and just check old map size against the sum here instead?

hmm not sure if I get it, if the condition is something like
```
(droppedColumns == 0) || (len(arrayOld) == changeableColumns+droppedColumns)
```

then it wouldn't work correctly, for example when changing `{col1, col2, col3}` to `{col1, col4, col5}` the condition would result in `true` (`changeableColumns=1`, `droppedColumns=2`)

I think we can do something like:

```
newColumns := len(arrayNew) - changeableColumns
return (droppedColumns == 0) || (newColumns == 0), nil
```

what do you think?

mmv1/third_party/terraform/services/bigquery/resource_bigquery_table.go
```diff
@@ -1725,6 +1727,40 @@ func resourceBigQueryTableUpdate(d *schema.ResourceData, meta interface{}) error
...
+		if !newTableFields[field.Name] {
+			droppedColumns = append(droppedColumns, field.Name)
+		}
+	}
+
+	if len(droppedColumns) > 0 {
+		droppedColumnsString := strings.Join(droppedColumns, " ")
+		log.Printf("[INFO] Dropping columns in-place: %s", droppedColumnsString)
+
+		dropColumnsDDL := fmt.Sprintf("ALTER TABLE `%s.%s.%s` DROP COLUMN `%s`", d.Get("project").(string), d.Get("dataset_id").(string), d.Get("table_id").(string), droppedColumnsString)
```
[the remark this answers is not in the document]
obada-ab: Done

mmv1/third_party/terraform/services/bigquery/resource_bigquery_table.go
```diff
@@ -1725,6 +1727,40 @@ func resourceBigQueryTableUpdate(d *schema.ResourceData, meta interface{}) error
 	datasetID := d.Get("dataset_id").(string)
 	tableID := d.Get("table_id").(string)
 
+	oldTable, err := config.NewBigQueryClient(userAgent).Tables.Get(project, datasetID, tableID).Do()
```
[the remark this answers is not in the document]
obada-ab: Done

mmv1/third_party/terraform/services/bigquery/resource_bigquery_table_internal_test.go
```diff
@@ -517,12 +517,32 @@ var testUnitBigQueryDataTableIsChangableTestCases = []testUnitBigQueryDataTableJ
 		jsonNew:    "[{\"name\": \"value2\", \"type\" : \"BOOLEAN\", \"mode\" : \"NULLABLE\", \"description\" : \"newVal\" },  {\"name\": \"value1\", \"type\" : \"INTEGER\", \"mode\" : \"NULLABLE\", \"description\" : \"someVal\" }]",
 		changeable: true,
 	},
+	// This is changeable but it is currently treated by terraform as
+	// the renamed column being dropped and a new column added
```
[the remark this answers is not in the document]
obada-ab: Done

mmv1/third_party/terraform/services/bigquery/resource_bigquery_table_internal_test.go
obada-ab: Done

mmv1/third_party/terraform/services/bigquery/resource_bigquery_table.go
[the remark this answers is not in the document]
wj-chen: Let's still include other in-place changes. Naming is hard though - the columns would have the same name but maybe different other attributes. "SameNameColumns"?

Your updated logic looks good.

mmv1/third_party/terraform/services/bigquery/resource_bigquery_table.go
```diff
@@ -1703,6 +1705,12 @@ func resourceBigQueryTableRead(d *schema.ResourceData, meta interface{}) error {
 	return nil
 }
 
+type TableInfo struct {
```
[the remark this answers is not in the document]
obada-ab: Changed it to `TableReference`

mmv1/third_party/terraform/services/bigquery/resource_bigquery_table_internal_test.go
```diff
@@ -548,16 +566,26 @@ var testUnitBigQueryDataTableIsChangableTestCases = []testUnitBigQueryDataTableJ
...
-			testcase.name + "Nested",
-			fmt.Sprintf("[{\"name\": \"someValue\", \"type\" : \"INTEGER\", \"fields\" : %s }]", testcase.jsonOld),
-			fmt.Sprintf("[{\"name\": \"someValue\", \"type\" : \"INT64\", \"fields\" : %s }]", testcase.jsonNew),
-			testcase.changeable,
+	}
+}
+
+func TestUnitBigQueryDataTable_schemaIsChangeableNested(t *testing.T) {
+	t.Parallel()
+	// Only top level column drops are changeable
```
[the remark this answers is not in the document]
obada-ab: Good point, changed the logic to test custom outcomes for nested cases

obada-ab:
I tested the configuration provided in https://github.com/hashicorp/terraform-provider-google/issues/12465#issuecomment-1237308919, and when performing `terraform apply` the second time the provider attempts to remove some columns and fails with an error:

```
Terraform used the selected providers to generate the following execution plan. Resource actions are
indicated with the following symbols:
  ~ update in-place

  # google_bigquery_table.my_table will be updated in-place
  ~ resource "google_bigquery_table" "my_table" {
        id                       = "projects/obada-test-383610/datasets/my_source/tables/my_table"
      ~ schema                   = jsonencode(
          ~ [
                # (2 unchanged elements hidden)
                {
                    mode = "NULLABLE"
                    name = "weight"
                    type = "BIGNUMERIC"
                },
              - {
                  - mode = "NULLABLE"
                  - name = "environment"
                  - type = "STRING"
                },
              - {
                  - mode = "NULLABLE"
                  - name = "frequency"
                  - type = "STRING"
                },
            ]
        )
        # (18 unchanged attributes hidden)

        # (1 unchanged block hidden)
    }

  Enter a value: yes  

google_bigquery_table.my_table: Modifying... [id=projects/obada-test-383610/datasets/my_source/tables/my_table]
╷
│ Error: googleapi: Error 400: ALTER TABLE only supports altering schema for tables while my_table is a EXTERNAL TABLE
```

It looks like `ALTER TABLE` won't work at all for tables with external data configurations, looking into workarounds to avoid the error.

obada-ab:
The new behavior is problematic for external tables, I tested the configuration in https://github.com/hashicorp/terraform-provider-google/issues/12465#issuecomment-1237308919, and I got an error when performing `terraform apply` the second time:

```
ALTER TABLE only supports altering schema for tables while my_table is a EXTERNAL TABLE
```

Made changes so that the new behavior doesn't apply for external tables, tested the external table configuration again and now the provider forces replacement.

mmv1/third_party/terraform/services/bigquery/resource_bigquery_table_internal_test.go
```diff
@@ -576,16 +576,18 @@ func TestUnitBigQueryDataTable_schemaIsChangeable(t *testing.T) {
...
 	for _, testcase := range testUnitBigQueryDataTableIsChangeableTestCases {
-		if !skipNested[testcase.name] {
-			testcaseNested := &testUnitBigQueryDataTableJSONChangeableTestCase{
-				testcase.name + "Nested",
-				fmt.Sprintf("[{\"name\": \"someValue\", \"type\" : \"INTEGER\", \"fields\" : %s }]", testcase.jsonOld),
-				fmt.Sprintf("[{\"name\": \"someValue\", \"type\" : \"INT64\", \"fields\" : %s }]", testcase.jsonNew),
-				testcase.changeable,
-			}
-			testcaseNested.check(t)
+		changeable, ok := customNestedValues[testcase.name]
```
wj-chen, line 594: nit: This works but I think the reverse would be easier to understand - `changeable` is default to `testcase.changeable`, except when it's one of the two, then it's false.

mmv1/third_party/terraform/services/bigquery/resource_bigquery_table.go
```diff
@@ -298,9 +299,13 @@ func resourceBigQueryTableSchemaIsChangeable(old, new interface{}, topLevel bool
 				resourceBigQueryTableSchemaIsChangeable(mapOld[key], mapNew[key], false); err != nil || !isChangable {
 				return false, err
 			}
```
wj-chen, line 302: I thought we'd be updating this as we look through `mapOld` and comparing key to `mapNew` - could you explain how this is working as intended?

wj-chen:
> The new behavior is problematic for external tables, I tested the configuration in [hashicorp/terraform-provider-google#12465 (comment)](https://github.com/hashicorp/terraform-provider-google/issues/12465#issuecomment-1237308919), and I got an error when performing `terraform apply` the second time:
> 
> ```
> ALTER TABLE only supports altering schema for tables while my_table is a EXTERNAL TABLE
> ```
> 
> Made changes so that the new behavior doesn't apply for external tables, tested the external table configuration again and now the provider forces replacement.

Thank you for double checking and making the corresponding change!

mmv1/third_party/terraform/services/bigquery/resource_bigquery_table_internal_test.go
obada-ab: Done, changed it to a more readable form

mmv1/third_party/terraform/services/bigquery/resource_bigquery_table.go
obada-ab: not sure I understand the question correctly, we are going through the keys in `mapOld` and comparing to `mapNew`, but we're doing it recursively because of the nature of the schema (nested interfaces). We immediately return false (not changeable) if we encounter an unchangeable case, for example a column that had it's type changed.

wj-chen: I was referring to this block:
```
if topLevel {
  sameNameColumns += 1
}
```

At a glance, in a single given level, top or nested, `sameNameColumns` will be 0 or 1. Additional comments on the variables and the logic here would help.

obada-ab: `sameNameColumns ` is always 0 for any nested level, and could be 0 or more at the top level. I refactored this part of the code, it should still be the same logic but I think It's more readable now and less prone to accidental logic-breaking changes.

wj-chen (approved):

obada-ab:
@hao-nan-li can you take a look

hao-nan-li (approved):

PR #10179, opened 2024-03-12
fix tags cause notebooks instances to recreate

bcreddy-gcp:
Fixes https://github.com/hashicorp/terraform-provider-google/issues/17075

```release-note:bug
notebooks: fixed an issue where default tags would cause a diff recreating `google_notebooks_instance` resources
```


rileykarson (approved):
LGTM assuming no surprises from tests. Will merge post-results if so.

We may want to consider a rework of the field in `6.0.0`, similar to the one done for labels/annotations in https://registry.terraform.io/providers/hashicorp/google/latest/docs/guides/version_5_upgrade#provider (more like annotations than labels), or using a CustomizeDiff to propose the new values during plan.

bcreddy-gcp:
Noted. This is a deprecated resource, so not very keen on reworking the fields. I'll keep this in mind for other supported resources.

PR #10182, opened 2024-03-13
Fix the bug for computed labels and annotations

zli82016:
fixes https://github.com/hashicorp/terraform-provider-google/issues/16217

Fix the bug for root computed labels and annotations.

The bug for nested labels field (`metadata.labels` and `metadata.annotations`) will be fixed later. Two resources (`google_cloud_run_service` and `google_cloud_run_domain_mapping`) are affected by this case.

```release-note:none

```


rileykarson (approved):
LGTM- I tested locally as well, and saw the bad behaviour on an old version (`5.7.0`) & the fixed behaviour after building a provider w/ your change.

PR #10187, opened 2024-03-13
Remove misleading examples from docs for `google_dataform_repository`, add new example

SarahFrench:
Closes https://github.com/hashicorp/terraform-provider-google/issues/17335

The examples removed in this PR show use of `google_dataform_repository` with `google_sourcerepo_repository`. Acceptance tests generated from that example have consistently passed. However when those resources are viewed in the Google Cloud Console there is feedback about errors and the two components cannot function together.

This PR removed those misleading examples from the documentation but retain them as acceptance tests. Removing those examples from the docs would leave the documentation with no examples, so I've added a replacement where the reference `google_sourcerepo_repository.git_repository.url` is replaced with the obvious stand-in value `"https://github.com/OWNER/REPOSITORY.git""`.

```release-note:none

```


mmv1/products/dataform/Repository.yaml
```diff
@@ -33,7 +33,10 @@ import_format:
   ['projects/{{project}}/locations/{{region}}/repositories/{{name}}']
 examples:
   - !ruby/object:Provider::Terraform::Examples
-    name: 'dataform_repository'
+    name: 'dataform_respository'
```
SarahFrench, line 36: I have created a new file with this name but different contexts. The original `dataform_respository` was renamed to  `dataform_repository_with_cloudsource_repo`, and `dataform_repository_ssh` as renamed to `dataform_repository_with_cloudsource_repo_and_ssh`

roaks3 (approved):
LGTM, fix here makes sense.

I do have a related question, that we may want to add to the end of the issue: do we even need these as acceptance tests if they ultimately don't work?

SarahFrench:
>I do have a related question, that we may want to add to the end of the issue: do we even need these as acceptance tests if they ultimately don't work?

That's a good question... I think the only acceptance tests for this resource are the original tests that use `google_sourcerepo_repository`. The problem wasn't flagged at the time because the API accepts source repos as input but cannot function when being used versus simply being provisioned.

I was imagining that we'd keep the acc tests for the resource and re-add the examples to the docs when the two services are made compatible with each other

SarahFrench:
I'm going to go ahead with merge and happy to open a subsequent issue based on the Q about removing acc tests

PR #10189, opened 2024-03-13
Add SarahFrench to on vacation list

SarahFrench:
To be merged on Friday.
I'll be on reduced availability and then OOO from 18th March onwards, and then back on the 25th March.

```release-note:none

```


BBBmau (approved):

SarahFrench:
Thanks @BBBmau!

Now you're on the review rotation its worth you knowing - my adding your username to this list you can avoid being assigned PRs when you're on vacation. It's nice to not return to a full inbox of notifications!

PR #10190, opened 2024-03-13
Change load_balancer_type to optional in dns_record_set resource

nbodepudi:
`load_balancer_type` is becoming optional to align with upcoming API changes. This change will introduce support for internal regional network proxy load balancers, as well as any other GCP load balancer that can be health checked.

**b/290267743**

```release-note:enhancement
dns: changed `load_balancer_type` field from required to optional in `google_dns_record_set`
```


slevenick (changes requested):
Looks like the test is failing. Is not having a load balancer type actually valid?

nbodepudi:
Hey, API changes are not in prod yet so the test is failing. Will make this PR a draft and reopen once they have landed.

slevenick (changes requested):
Looks like the API sends back the value as "none" instead of what we expect which would be not sending the value back at all.

There's a couple ways to fix that, maybe we should copy the API's pattern and allow sending the value of "none" specifically?

nbodepudi:
Hi @slevenick, For some reason acceptance tests for GA provider are not running in my machine  and it is throwing a weird error. I added a diff suppress func to fix this. I ran acceptance tests in beta, they were successful.  Please take a look when you have time.

mmv1/third_party/terraform/services/dns/resource_dns_record_set.go
```diff
@@ -16,6 +16,16 @@ import (
 	"google.golang.org/api/dns/v1"
 )
 
+func lbTypeNoneDiffSuppress(k, old, new string, d *schema.ResourceData) bool {
+	// Check if the key matches the relevant attribute
+	if k != "routing_policy.0.primary_backup.0.primary.0.internal_load_balancers.0.load_balancer_type" {
```
slevenick, line 21: So, `internal_load_balancers` is a list, which means this diff suppress func will only function on the first element of that list. Can you add a test that specifies 2 `internal_load_balancers` that would need to use this diff suppress func?

slevenick (changes requested):

slevenick (approved):

PR #10191, opened 2024-03-13
Improve the error messages when mmv1 compilation fails

roaks3:
fixes https://github.com/hashicorp/terraform-provider-google/issues/16650

When errors are raised during compilation, we don't always get clear signal regarding where the failure happened (due in part to multiple threads being used). This comes up fairly regularly when we see EAP build failures, and it is often difficult to track down where a fix is needed.

For example, during one recent EAP failure, we saw this message:
```
compiler.rb:206:in `block (2 levels) in <top (required)>': undefined local variable or method `file_path' for main:Object (NameError)
	from compiler.rb:203:in `each'
	from compiler.rb:203:in `block in <top (required)>'
	from /Users/ryanoaks/.rbenv/versions/3.1.0/lib/ruby/gems/3.1.0/gems/parallel-1.22.1/lib/parallel.rb:587:in `call_with_index'
...
```

After this change, we can see which file triggered the failure:
```
E, [2024-03-05T19:37:44.467169 #14233] ERROR -- : Failed to compile using override ../../../../../../../../Users/ryanoaks/projects/magic-modules-private-overrides/products/compute/BackendService.yaml: undefined local variable or method `file_path' for main:Object

        || File.basename(file_path).include?('go_')
                         ^^^^^^^^^
...
compiler.rb:206:in `block (2 levels) in <top (required)>': undefined local variable or method `file_path' for main:Object (NameError)
	from compiler.rb:203:in `each'
	from compiler.rb:203:in `block in <top (required)>'
...
```

Note: my ruby is a bit rusty, so hopefully this is a suitable way to capture the error, print some more info, and then re-raise it.

```release-note:none

```

### DIFF
--- a/mmv1/compiler.rb
+++ b/mmv1/compiler.rb
@@ -192,6 +192,9 @@ products_for_version = Parallel.map(all_product_files, in_processes: 8) do |prod
       )
       resource.validate
       resources.push(resource)
+    rescue StandardError => e
+      Google::LOGGER.error "Failed to compile #{file_path}: #{e}"
+      raise e
     end
 
     if override_dir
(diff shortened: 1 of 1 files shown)


c2thorn (approved):

c2thorn:
noted for the Go rewrite

PR #10194, opened 2024-03-13
google_firebase_android_app should be replaced when package_name changes

rainshen49:
This is already the behavior in the API. It was forgotten in the initial PR. However, it's not a common operation for an app to change the package_name mid-way, so it probably has little real impact.

```release-note:bug
firebase: made the `google_firebase_android_app` resource's `package_name` field required and immutable. This prevents API errors encountered by users who attempted to update or leave that field unset in their configurations.
```

### DIFF
--- a/mmv1/products/firebase/AndroidApp.yaml
+++ b/mmv1/products/firebase/AndroidApp.yaml
@@ -122,8 +122,10 @@ properties:
       This identifier should be treated as an opaque token, as the data format is not specified.
   - !ruby/object:Api::Type::String
     name: packageName
+    immutable: true
+    required: true
     description: |
-      Immutable. The canonical package name of the Android app as would appear in the Google Play
+      The canonical package name of the Android app as would appear in the Google Play
       Developer Console.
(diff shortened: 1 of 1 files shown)


SarahFrench:
Hi @rainshen49 , thanks for this fix! I took a look at the docs and agree that this change is needed.

As this type of change (Optional=>Required) is typically a breaking change, I want to check: what would happen if a user wasn't setting `package_name` in `google_firebase_android_app` resource blocks in their configuration. Would the resource fail to be created, or would the API supply a default value?

rainshen49:
Thanks for double-checking. Yes, the immutable & required behavior is enforced at the API level since inception. If a current user doesn't specify a `package_name`, they will get an invalid argument error from the API. Similarly, if they try to change the package_name, they will also get an invalid argument error.

SarahFrench:
Thanks for confirming that! I'll address remaining check failures and then approve + merge

/gcbrun

SarahFrench (approved):
Approving as a non-breaking change due to reasons discussed above :shipit:

PR #10199, opened 2024-03-14
Modify issue labeler to skip linked issues

roaks3:
Context is in the comment: adding service labels to a linked issue creates ambiguity that is difficult for us to detect and address, so we decided it was better to skip them.

```release-note:none

```

### DIFF
--- a/tools/issue-labeler/labeler/backfill.go
+++ b/tools/issue-labeler/labeler/backfill.go
@@ -98,6 +98,14 @@ func ComputeIssueUpdates(issues []Issue, regexpLabels []RegexpLabel) []IssueUpda
 			continue
 		}
 
+		// Decision was made to no longer add new service labels to linked tickets, because it is
+		// more difficult to know which teams have received those tickets and which haven't.
+		// Forwarding a ticket to a different service team should involve removing the old service
+		// label and `linked` label.
+		if linked {
+			continue
+		}
+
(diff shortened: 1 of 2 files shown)


melinath (changes requested):
could you add a test for this behavior?

melinath:
actually this seems like it should have caused the "add missing service labels even if already linked" to start failing... but apparently it didn't? Are we running these tests automatically?

roaks3:
Ah yep, that thought crossed my mind here but I forgot to follow up. Let me look at added a test for this.

roaks3:
Confirmed that the test was not run in CI, and it did fail when running locally. I've updated the test to reflect the change, but we should update them to run in PRs.

cc @trodge is this a known limitation, or perhaps we removed this behavior by accident?

melinath (approved):
LGTM. It would be great to also run tests on PRs as a GHA - but that doesn't need to happen in this PR.

trodge:
> Confirmed that the test was not run in CI, and it did fail when running locally. I've updated the test to reflect the change, but we should update them to run in PRs.
> 
> cc @trodge is this a known limitation, or perhaps we removed this behavior by accident?

I don't think we ever ran these tests as part of CI.

trodge (approved):

roaks3:
Created https://github.com/hashicorp/terraform-provider-google/issues/17598

PR #10205, opened 2024-03-15
Change the dataflow zone from us-central1-f to us-east5-b

liferoad:
Based on https://github.com/hashicorp/terraform-provider-google/issues/17385, some Dataflow tests seem flakey.

This PR switches the tests to us-east5-b to check whether this can reduce the flakiness. 

```release-note:none

```


roaks3:
Note that I don't think this region is swept by default, but it seems reasonable if it fixes these tests

roaks3 (approved):

PR #10207, opened 2024-03-15
Only run mmv1 rake tests and yaml linting if relevant files have changed

melinath:
See https://docs.github.com/en/actions/using-workflows/triggering-a-workflow#using-filters-to-target-specific-paths-for-pull-request-or-push-events. This should reduce clutter on PRs that don't impact mmv1.

```release-note:none

```

### DIFF
--- a/.github/workflows/mmv1-rake-tests.yml
+++ b/.github/workflows/mmv1-rake-tests.yml
@@ -1,0 +1,39 @@
+name: mmv1-rake-tests
+
+permissions: read-all
+
+on:
+  pull_request:
+    paths:
+      - 'mmv1/**.rb'
+
+jobs:
+  rake-tests:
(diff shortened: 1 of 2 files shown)


.github/workflows/mmv1-rake-tests.yml
```diff
@@ -0,0 +1,39 @@
+name: mmv1-rake-tests
+
+permissions: read-all
+
+on:
+  pull_request:
+    paths:
+      - 'mmv1/**.rb'
```
SarahFrench, line 8: I think this may need to be `mmv1/**/*.rb`

```suggestion
      - 'mmv1/**/*.rb'
```

melinath: Done

PR #10221, opened 2024-03-18
artifactregistry: support custom remote repository uris

rafibarash:

```release-note:enhancement
artifactregistry: added `remote_repository_config.<facade>_repository.custom_repository.uri` field to `google_artifact_registry_repository` resource. This allows setting a remote repository upstream to any registry uri that supports Basic Authentication (such as setting a Jfrog Artifactory instance as the upstream).
```


mmv1/products/artifactregistry/Repository.yaml
```diff
@@ -369,10 +379,22 @@ properties:
                   - :UBUNTU
               - !ruby/object:Api::Type::String
                 name: 'repositoryPath'
-                required: true
```
rafibarash, line 372: Removing these `required` tags since these shouldn't be set if a custom_repository is chosen, and I believe keeping these will cause issues (I originally added `required` on the custom_repository uri field and this caused issues with public_repository tests). But if we can keep these `required` fields to only show up when `public_repository` is used and not `custom_repository`, I can re-add these!

rafibarash:
Sorry @hao-nan-li, let me make this a draft until tests are passing locally

hao-nan-li:
> Sorry @hao-nan-li, let me make this a draft until tests are passing locally

No worries, take your time.

rafibarash:
@hao-nan-li I'm running into an issue and was wondering if you could provide any guidance.

The nested object `remoteRepositoryConfig.dockerRepository` has a property `publicRepository` with a [default value set](https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/products/artifactregistry/Repository.yaml#L398). This PR is adding a `customRepository` property that can be set instead of `publicRepository`. However when I'm running my tests when setting `customRepository`, the default value for `publicRepository` is still being applied causing AR to return 400 and the test to fail.

Is there any way to keep the default value in `publicRepository`, but not apply it when `customRepository` is set? I saw that removing a default value is a breaking change so I want to see if there are any other options to support setting `customRepository` without removing this default value

hao-nan-li:
> @hao-nan-li I'm running into an issue and was wondering if you could provide any guidance.
> 
> The nested object `remoteRepositoryConfig.dockerRepository` has a property `publicRepository` with a [default value set](https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/products/artifactregistry/Repository.yaml#L398). This PR is adding a `customRepository` property that can be set instead of `publicRepository`. However when I'm running my tests when setting `customRepository`, the default value for `publicRepository` is still being applied causing AR to return 400 and the test to fail.
> 
> Is there any way to keep the default value in `publicRepository`, but not apply it when `customRepository` is set? I saw that removing a default value is a breaking change so I want to see if there are any other options to support setting `customRepository` without removing this default value

Is `publicRepository` created by default even if you don't have that field in your config?

rafibarash:
> > @hao-nan-li I'm running into an issue and was wondering if you could provide any guidance.
> > The nested object `remoteRepositoryConfig.dockerRepository` has a property `publicRepository` with a [default value set](https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/products/artifactregistry/Repository.yaml#L398). This PR is adding a `customRepository` property that can be set instead of `publicRepository`. However when I'm running my tests when setting `customRepository`, the default value for `publicRepository` is still being applied causing AR to return 400 and the test to fail.
> > Is there any way to keep the default value in `publicRepository`, but not apply it when `customRepository` is set? I saw that removing a default value is a breaking change so I want to see if there are any other options to support setting `customRepository` without removing this default value
> 
> Is `publicRepository` created by default even if you don't have that field in your config?

Yeah it appears that way from the test error message. The full error message is

```
Invalid value at 'repository.remote_repository_config.docker_repository' (oneof), oneof field 'upstream' is already set. Cannot set 'publicRepository'
```

I'm running the test on the commit that I just pushed. Do you think there are any [fields](https://googlecloudplatform.github.io/magic-modules/develop/field-reference/#mmv1-field-reference) I could add or remove (ideally not the default value field though) to fix this?

hao-nan-li:
> > > @hao-nan-li I'm running into an issue and was wondering if you could provide any guidance.
> > > The nested object `remoteRepositoryConfig.dockerRepository` has a property `publicRepository` with a [default value set](https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/products/artifactregistry/Repository.yaml#L398). This PR is adding a `customRepository` property that can be set instead of `publicRepository`. However when I'm running my tests when setting `customRepository`, the default value for `publicRepository` is still being applied causing AR to return 400 and the test to fail.
> > > Is there any way to keep the default value in `publicRepository`, but not apply it when `customRepository` is set? I saw that removing a default value is a breaking change so I want to see if there are any other options to support setting `customRepository` without removing this default value
> > 
> > 
> > Is `publicRepository` created by default even if you don't have that field in your config?
> 
> Yeah it appears that way from the test error message. The full error message is
> 
> ```
> Invalid value at 'repository.remote_repository_config.docker_repository' (oneof), oneof field 'upstream' is already set. Cannot set 'publicRepository'
> ```
> 
> I'm running the test on the commit that I just pushed. Do you think there are any [fields](https://googlecloudplatform.github.io/magic-modules/develop/field-reference/#mmv1-field-reference) I could add or remove (ideally not the default value field though) to fix this?

Have you tried this following property? https://github.com/GoogleCloudPlatform/magic-modules/blob/5fddaec1797fd46f5f57241dc878642b5bcf60f2/mmv1/api/type.rb#L103

mmv1/templates/terraform/examples/artifact_registry_repository_remote_custom_dockerhub_auth.tf.erb
```diff
@@ -0,0 +1,42 @@
...
+  description   = "<%= ctx[:vars]['description'] %>"
+  format        = "DOCKER"
+  mode          = "REMOTE_REPOSITORY"
+  remote_repository_config {
+    description = "custom uri docker hub with auth credentials"
+    disable_upstream_validation = true
+    docker_repository {
+      custom_repository {
+        uri = "<%= ctx[:vars]['custom_uri'] %>"
+      }
```
rafibarash, line 33: @hao-nan-li this is the test that is failing. As you can see I'm only specifying `custom_repository`, but the test is failing because the `public_repository` default is also being applied and both can't be applied

rafibarash:
Moving to https://github.com/GoogleCloudPlatform/magic-modules/pull/10242

PR #10227, opened 2024-03-19
remove `pattern` field from `.yaml` files

BBBmau:
Removes the `pattern` field from all `.yaml` fields since they aren't used in .erb or .rb files.

This was done by doing a regex search with `pattern:.*[^>|\.|\?]$` on all `.yaml` files in the repo

Fixes [#17269
](https://github.com/hashicorp/terraform-provider-google/issues/17269)

```release-note:none

```


BBBmau:
There's no make command that does a lint-fix 😞

melinath (approved):

PR #10231, opened 2024-03-19
Add DLP resource DiscoveryConfig 

patrickmoy:

```release-note:new-resource
google_data_loss_prevention_discovery_config
```


mmv1/products/dlp/DiscoveryConfig.yaml
```diff
@@ -0,0 +1,352 @@
...
+                  description: A developer-facing error message, which should be in English. Any user-facing error message should be localized and sent in the google.rpc.Status.details field, or localized by the client.
+                - !ruby/object:Api::Type::Array
+                  name: 'details'
+                  description: A list of messages that carry the error details.
+                  item_type: Api::Type::KeyValuePairs
+            - !ruby/object:Api::Type::String
+              name: timestamp
+              description: The times the error occurred. List includes the oldest timestamp and the last 9 timestamps.
+      - !ruby/object:Api::Type::String
+        name: createTime
```
trodge, line 339: These output only fields should be marked with `output: true`. We also usually omit create and update time fields from resources.

patrickmoy: Marked; I admit I just followed prior convention with existing DLP resources. I can remove the fields if that's preferred.

patrickmoy:
Looks like there was an additional layer of nesting in the .yaml that was causing fields to not be recognized correctly in tests which has now been removed.

shuyama1:
@patrickmoy There's an issue in our CI. Would you mind rebasing your PR to catch the latest fix? Sorry about it!

trodge:
> Tests are complaining about inclusion of a field that is an empty proto... is there a certain way to handle this? I looked for other test examples that use this setup for empty objects and many don't seem to have the field included in their tests.
> 
> This field is defined as follows in the .yaml:
> 
> ```
> - !ruby/object:Api::Type::NestedObject
>   name: otherTables
>   description: Catch-all. This should always be the last filter in the list because anything above it will apply first.
>   properties:
>     []  # Meant to be an empty object with no properties - see here : https://cloud.google.com/sensitive-data protection/docs/reference/rest/v2/organizations.locations.discoveryConfigs#allotherbigquerytables
>          # The fields below are necessary to include the "otherTables" filter in the payload
>   send_empty_value: true
>   allow_empty_object: true
> ```

It looks like in the yaml `otherTables` is nested under `tables` but in the example it's nested under `filter`.

patrickmoy:
Oh, whoops, thanks for catching that.... miss-indented there somehow.

patrickmoy:
Noticed that this check is queued still; does it need manual approval? [View more details on Google Cloud Build](https://console.cloud.google.com/cloud-build/builds/ccb15015-6d36-4e60-bfe2-c67671b2b32c?project=673497134629)

melinath:
@patrickmoy see cl/622290404 for information re: automatic test runs

patrickmoy:
@melinath Oh, should've figured there was something like that missing... thank you! Set to public

patrickmoy:
Consistently getting this error, and am not sure how to resolve:

> The requested URL <code>/v2/projects/ci-test-project-188019/discoveryConfigs?alt=json</code> was not found on this server.

mmv1/templates/terraform/examples/dlp_discovery_config_basic.tf.erb
```diff
@@ -0,0 +1,12 @@
...
+	parent = "projects/<%= ctx[:test_env_vars]['project'] %>/locations/us"
+
+    targets {
+        big_query_target {
+            filter {
+                other_tables {}
+            }
+        }
+    }
+    inspect_templates = ["projects/%{project}/locations/us/inspectTemplates/test"]
```
NickElliot, line 11: https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/templates/terraform/examples/dlp_inspect_template_basic.tf.erb example usage

NickElliot (dismissed):
Hi, the tests are failing due to this template not being accessible, and more importantly likely not existing. An inspect template should be put into the config for these tests using the existing dlp inspect template resource.

patrickmoy:
Changes made; some issues with parallelism of resource tests in local runs... is it possible to make the tests generated from examples run serially?

trodge:
> Changes made; some issues with parallelism of resource tests in local runs... is it possible to make the tests generated from examples run serially?

The test will need to be handwritten to do this. See [this test](https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/third_party/terraform/services/logging/resource_logging_billing_account_exclusion_test.go#L14-L33) for an example.

patrickmoy:
I've done that for the tests in mmv1/third_party/terraform/services/datalossprevention/resource_data_loss_prevention_discovery_config_test.go, and that seems to work fine, but the ones based on examples I'm not sure how to get to run in parallel. 

For example, TestAccDataLossPreventionDiscoveryConfig_dlpDiscoveryConfigActionsExample appears to be derived from mmv1/templates/terraform/examples/dlp_discovery_config_actions.tf.erb.

patrickmoy:
Additionally, I believe org-level tests will require permissions granted for the containing org for the service account.

hashicorp-test-runner@ci-test-project-188019.iam.gserviceaccount.com needs the role "dlp.jobTriggers.admin" on terraform-graphite-test.joonix.net (the equivalent service accounts for the nightly builds as well)

patrickmoy:
I'm not sure why it's flagging all these fields as missing tests... they're covered wholly in the example files as well as in resource_data_loss_prevention_discovery_config_test.go

patrickmoy:
Latest build log shows all tests passing https://storage.cloud.google.com/ci-vcr-logs/beta/refs/heads/auto-pr-10231/artifacts/75fcdbc2-71e8-4871-9d8f-c1aefd829a42/build-log/replaying_test.log

trodge (approved):

PR #10234, opened 2024-03-19
Upgraded version of actions related to scorecard

melinath:
Removed references to PAT token since we don't want to do that. This will hopefully resolve the error:

```
error signing scorecard json results: error signing payload: getting key from Fulcio: verifying SCT: updating local metadata and targets: error updating to TUF remote mirror: invalid key
```

If not, we should be able to follow up by stopping publication of results.

```release-note:none

```


zli82016 (approved):
LGTM

PR #10235, opened 2024-03-19
Move missing test detector into diff processor

trodge:

```release-note:none

```


.ci/magician/cmd/generate_comment.go
```diff
@@ -296,34 +296,25 @@ func execGenerateComment(prNumber int, ghTokenMagicModules, buildId, buildStep,
...
 
 	// Run missing test detector (currently only for beta)
-	missingTestsPath := mmLocalPath
 	for _, repo := range []source.Repo{tpgbRepo} {
 		if !repo.Cloned {
 			fmt.Println("Skipping missing tests; repo failed to clone: ", repo.Name)
 			continue
 		}
-		missingTests, err := detectMissingTests(missingTestsPath, repo.Path, oldBranch, rnr)
+		err = buildDiffProcessor(diffProcessorPath, repo.Path, diffProcessorEnv, rnr)
```
melinath, line 304: rather than rebuilding the diff processor, could you use the diff processor that's built above? (For now, by explicitly checking if the repo is beta before running missing test detection. Alternatively, we could run it for both and just discard one of them if both give responses, since we're not changing the formatting of the response as part of this.

melinath (changes requested):
Glanced through and this all seems reasonable - I'll take a deeper look next week. In the meantime - one small change request below. Could you also add a change to this PR that would trigger missing test detection?

.ci/magician/cmd/generate_comment.go
```diff
@@ -295,35 +305,6 @@ func execGenerateComment(prNumber int, ghTokenMagicModules, buildId, buildStep,
...
-		missingTests, err := detectMissingTests(missingTestsPath, repo.Path, oldBranch, rnr)
-		if err != nil {
-			fmt.Println("Error running missing test detector: ", err)
-			errors[repo.Title] = append(errors[repo.Title], "The missing test detector failed to run.")
-		}
-		data.MissingTests = missingTests
-	}
-
-	// Run unit tests for missing test detector
-	if err = runMissingTestUnitTests(
```
melinath, line 314: It looks like this PR will make us rely on the diff processor unit tests, which run using the main branch of TPG (rather than the current branch of TPGB.) However, we're still running the actual detector against the current branch.

We should probably run diff processor unit tests on TPGB as well. Do we also need to run missing test detector tests here still? That is, are there any issues that `TestReadAllTests` would catch that would _only_ be caught if it were run against the auto-pr branch?

.ci/magician/cmd/generate_comment.go
```diff
@@ -444,112 +425,17 @@ func changedSchemaLabels(prNumber int, currentLabels []string, diffProcessorPath
...
-
-// Run unit tests for the missing test detector.
-// Report results using Github API.
-func runMissingTestUnitTests(mmLocalPath, tpgbLocalPath, targetURL, commitSha string, prNumber int, gh GithubClient, rnr ExecRunner) error {
-	if err := rnr.PushDir(mmLocalPath); err != nil {
-		return err
-	}
-
-	diffs, err := rnr.Run("git", []string{"diff", "HEAD", "origin/main", "tools/missing-test-detector"}, nil)
+	output, err := rnr.Run("bin/diff-processor", []string{"detect-missing-tests", fmt.Sprintf("%s/google-beta/services", tpgbLocalPath)}, nil)
```
melinath, line 433:
```suggestion
	output, err := rnr.Run("bin/diff-processor", []string{"detect-missing-tests", fmt.Sprintf("%s/google/services", tpgbLocalPath)}, nil)
```

diff-processor renames the `google-beta` folder to `google` so that there's less customization needed for code interacting with the repo after build.

tools/diff-processor/cmd/detect_missing_tests.go
```diff
@@ -0,0 +1,75 @@
...
+			return o.run(args)
+		},
+	}
+}
+
+func (o *detectMissingTestsOptions) run(args []string) error {
+	allTests, errs := reader.ReadAllTests(args[0])
+	for path, err := range errs {
+		glog.Infof("error reading path: %s, err: %v", path, err)
+	}
```
melinath, line 44:
```suggestion
	}
	if len(errs) > 0 {
		return fmt.Errorf("reading tests")
	}
```

Should we return an error here? So that the root command can halt execution if the tests can't be read. Perhaps that would address the issue of not running the unit tests on this branch / making sure we catch newly-introduced unparseable tests?

melinath: Alternately we could return an error where the message includes the individual errors encountered (instead of logging them on separate lines)

tools/diff-processor/detector/detector.go
```diff
@@ -18,9 +19,44 @@ type MissingTestInfo struct {
 
 type FieldSet map[string]struct{}
 
+// ResourceChanges is a nested map with field names as keys and Field objects
+// as bottom-level values.
+// Fields are assumed not to be covered until detected in a test.
+type ResourceChanges map[string]any
```
melinath, line 25:
```suggestion
type ResourceChanges map[string]Field
```

Will the bottom-level value always be a field?

tools/diff-processor/detector/detector.go
```diff
@@ -53,16 +89,10 @@ func detectMissingTests(changedFields map[string]ResourceChanges, allTests []*re
 }
 
 func markCoverage(fieldCoverage ResourceChanges, config reader.Resource) error {
-	for fieldName, fieldValue := range config {
+	for fieldName := range config {
 		if coverage, ok := fieldCoverage[fieldName]; ok {
 			if field, ok := coverage.(*Field); ok {
 				field.Tested = true
-			} else if objectCoverage, ok := coverage.(ResourceChanges); ok {
```
melinath, line 60: why is this no longer needed?

tools/diff-processor/detector/detector.go
```diff
@@ -71,15 +101,13 @@ func markCoverage(fieldCoverage ResourceChanges, config reader.Resource) error {
...
+func untestedFields(fieldCoverage ResourceChanges) []string {
 	fields := make([]string, 0)
-	for fieldName, coverage := range fieldCoverage {
+	for key, coverage := range fieldCoverage {
 		if field, ok := coverage.(*Field); ok {
 			if !field.Tested {
-				fields = append(fields, strings.Join(append(path, fieldName), "."))
+				fields = append(fields, key)
 			}
-		} else if objectCoverage, ok := coverage.(ResourceChanges); ok {
```
melinath, line 81: why is this no longer necessary?

melinath (changes requested):
after making these changes, could you push changes that would trigger the missing test detector to detect missing tests? (And also a change that would cause ReadAllTests to fail.)

.ci/magician/cmd/generate_comment.go
trodge: This is using the original location of the tpgb repo, which doesn't get renamed.

tools/diff-processor/cmd/detect_missing_tests.go
trodge: The missing test detector can still produce useful output even if the reader encounters errors.

tools/diff-processor/detector/detector.go
trodge: The fields are now flattened, so there is no need to handle objects here.

melinath: ack, thanks!

tools/diff-processor/detector/detector.go
melinath: presumably the same reason as the other one.

.ci/magician/cmd/generate_comment.go
melinath: yep - thanks!

.ci/magician/cmd/generate_comment.go
```diff
@@ -338,38 +349,6 @@ func execGenerateComment(prNumber int, ghTokenMagicModules, buildId, buildStep,
...
-			fmt.Println("Error running missing test detector: ", err)
-			errors[repo.Title] = append(errors[repo.Title], "The missing test detector failed to run.")
-		}
-		data.MissingTests = missingTests
-	}
-
-	// Run unit tests for missing test detector (currently only for beta)
-	if pathChanged("tools/missing-test-detector", tpgbRepo.ChangedFiles) {
-		fmt.Printf("Found diffs in missing test detector:\n%s\nRunning tests.\n", diffs)
-		if err = runMissingTestUnitTests(
```
melinath, line 359: Are we losing anything by _not_ running these tests on this branch? The diff-processor unit tests only run on the main branch of the downstreams. Are there any types of problems that this test run would have caught that will no longer be caught?

melinath:
I've opened https://github.com/GoogleCloudPlatform/magic-modules/pull/10323 to add running diff-processor unit tests with tpgb main branch.

.ci/magician/cmd/generate_comment.go
melinath: Per discussion: the failures in ReadAll don't actually cause the _test_ to fail unless there's a panic. This matches the behavior of the missing test detector itself. So there is no case where a failure would previously have shown up which will no longer show up - and everything that would have been logged should still be getting logged.

However, we do need to (as part of this PR) make sure that the diff-processor unit tests are able to _run_ TestReadAllTests in a meaningful way (either by passing in SERVICES_DIR or by reworking TestReadAllTests to not require it, for example by using a relative dir to find the old / new directories locally.)

melinath (changes requested):

.github/workflows/unit-tests-diff-processor.yml
```diff
@@ -40,3 +40,5 @@ jobs:
         run: |
           cd tools/diff-processor
           go test -v ./...
+        env:
```
melinath, line 43: Could you add this to the TPG section as well? I know it's not strictly required for this change but it would be nice to do anyway (since it should pass in both cases.)

melinath (approved):

PR #9728, opened 2023-12-31
Add support for NFS and GCS mounts in Cloud Run v2 service

mvanholsteijn:
Adds support for NFS and GCS mounts in Cloud Run v2 service. The volume definitions are already in the REST API definition.

I also included the tcpSocket in the liveness_probe, as that was missing too.

PS: Updates were made using the [magic-module-scaffolder](https://github.com/binxio/magic-module-scaffolder) which automates the updates of Magic Module resource definitions.

* [#16880](https://github.com/hashicorp/terraform-provider-google/issues/16880)

```release-note:enhancement
cloudrunv2: added `nfs` field to `google_cloud_run_v2_service.template.volumes`
cloudrunv2: added `gcs` field to `google_cloud_run_v2_service.template.volumes`
cloudrunv2: adding `tcpSocket` field to `google_cloud_run_v2.template.containers.liveness_probe`
```

PR #10244, opened 2024-03-20
Add nfs support to cloudrun v1 services.

bskaplan:
Support for NFS in v2 services was added in https://github.com/GoogleCloudPlatform/magic-modules/pull/9728, this adds it to v1.

```release-note:enhancement
cloudrun: support nfs volumes (beta)

```

### DIFF
--- a/mmv1/products/cloudrun/Service.yaml
+++ b/mmv1/products/cloudrun/Service.yaml
@@ -875,6 +875,28 @@ properties:
                             Driver-specific attributes. The following options are supported for available drivers:
                               * gcsfuse.run.googleapis.com
                                 * bucketName: The name of the Cloud Storage Bucket that backs this volume. The Cloud Run Service identity must have access to this bucket.
+                    - !ruby/object:Api::Type::NestedObject
+                      name: nfs
+                      description: |-
+                        A filesystem backed by a Network File System share. This filesystem requires the
+                        run.googleapis.com/execution-environment annotation to be set to "gen2" and
+                        run.googleapis.com/launch-stage set to "BETA" or "ALPHA".
+                      min_version: beta
+                      properties:
+                        - !ruby/object:Api::Type::String
+                          name: server
+                          required: true
+                          description: |-
+                            IP address or hostname of the NFS server
+                        - !ruby/object:Api::Type::String
+                          name: path
+                          required: true
+                          description: |-
+                            Path exported by the NFS server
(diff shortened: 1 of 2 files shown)


mmv1/products/cloudrun/Service.yaml
```diff
@@ -875,6 +875,28 @@ properties:
                             Driver-specific attributes. The following options are supported for available drivers:
                               * gcsfuse.run.googleapis.com
                                 * bucketName: The name of the Cloud Storage Bucket that backs this volume. The Cloud Run Service identity must have access to this bucket.
+                    - !ruby/object:Api::Type::NestedObject
+                      name: nfs
```
BBBmau, line 879: We'll want to add a test that verifies that this works properly in the provider, a test configuration can be added [here](https://github.com/GoogleCloudPlatform/magic-modules/tree/77041bfd26766336f1c2a47818425c46712e5b7a/mmv1/templates/terraform/examples) as a as a `cloud_run_service_nfs.tf.erb` file.

bskaplan: I've been trying and I'm not sure it's possible to make a test for this. The test requires a VPC, and tearing down the VPC keeps failing because resource locks aren't immediately released when the Cloud Run Service and/or VPC Connector are deleted.

mmv1/products/cloudrun/Service.yaml
```diff
@@ -875,6 +875,28 @@ properties:
...
+                          required: true
+                          description: |-
+                            IP address or hostname of the NFS server
+                        - !ruby/object:Api::Type::String
+                          name: path
+                          required: true
+                          description: |-
+                            Path exported by the NFS server
+                        - !ruby/object:Api::Type::Boolean
+                          name: readOnly
```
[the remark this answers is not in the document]
bskaplan: Not needed since booleans default to false. https://googlecloudplatform.github.io/magic-modules/develop/field-reference/#default_value

bskaplan:
It looks like it's not feasible to make a test right now that uses a VPC (it would require a 1 hour sleep and I don't feel comfortable adding that to the magic-modules test suite). I've asked the team to help make this faster, but can we merge this in the mean time and we'll add the test later?

ian-mi (approved):

bskaplan:
@BBBmau can you approve? I don't think adding a 1 hour sleep in a test is worth it for this small of a change.

BBBmau (approved):

PR #10256, opened 2024-03-21
Pin workflow dependency versions

melinath:
This _should_ resolve a couple of alerts related to pinned versions in workflows.

```release-note:none

```


slevenick (approved):
Please fix lint issue before merging

PR #10259, opened 2024-03-22
Rewrite type functions

zli82016:
1. Rewrite functions of object Type with Go
2. Clean up some unused functions

```release-note:none

```

### DIFF
--- a/mmv1/api/type.go
+++ b/mmv1/api/type.go
@@ -16,7 +16,6 @@ package api
 import (
 	"fmt"
 	"log"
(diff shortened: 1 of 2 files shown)


c2thorn (approved):
Some of this we may end up cutting if we don't end up using them. I'm finding a lot is not actually useful anymore or there are better ways to do it. But for now, LGTM!

zli82016:
`community-membership-checker` failure is OK as it was triggered twice and the first one succeeded.

PR #9890, opened 2024-01-29
add stack_type attribute

ehorning:
Add `stack_type` attribute for `google_compute_interconnect_attachment` resource.

Note: Explicit test for this attribute cannot be added in examples because this attribute cannot be added for attachments of type `PARTNER`, attachments of type `DEDICATED` require `interconnect` to be specified, and there is no Terraform resource for creating an interconnect. Attachments of type `PARTNER_PROVIDER` cannot be created without being an approved partner. 

Fixes: https://github.com/hashicorp/terraform-provider-google/issues/17129

**Release Note Template for Downstream PRs (will be copied)**

```release-note:enhancement
compute: added `stack_type` attribute for `google_compute_interconnect_attachment` resource.
```

PR #10269, opened 2024-03-23
ci: add missing pathspec dependency for yamllint pipeline

srevinsaju:
pipelines for YAML lints are failing due to:
```
Traceback (most recent call last):
  File "/home/runner/.local/bin/yamllint", line 5, in <module>
    from yamllint.cli import run
  File "/home/runner/.local/lib/python3.10/site-packages/yamllint/cli.py", line 24, in <module>
    from yamllint.config import YamlLintConfig, YamlLintConfigError
  File "/home/runner/.local/lib/python3.10/site-packages/yamllint/config.py", line 19, in <module>
    import pathspec
ModuleNotFoundError: No module named 'pathspec'
```

A fresh install in a virtual environment using the following command:
```bash
$ pip install yamllint==1.32.0 pyyaml==6.0.1
$ pip freeze
pathspec==0.12.1
PyYAML==6.0.1
yamllint==1.32.0
```

Regression introduced by 60917c85e2f6645e21846fcf0197ff18244c747d

**Release Note Template for Downstream PRs (will be copied)**

```release-note:none

```

PR #10275, opened 2024-03-25
add ipv6 outputs

ehorning:
Add Ipv6 outputs to interconnect attachment resource. Fixes https://github.com/hashicorp/terraform-provider-google/issues/17671.

Followup to https://github.com/GoogleCloudPlatform/magic-modules/pull/9890

```release-note:enhancement
compute: added `cloud_router_ipv6_address`, `customer_router_ipv6_address` fields to `google_compute_interconnect_attachment` resource
```

### DIFF
--- a/mmv1/products/compute/InterconnectAttachment.yaml
+++ b/mmv1/products/compute/InterconnectAttachment.yaml
@@ -312,3 +312,15 @@ properties:
       - :IPV4_IPV6
       - :IPV4_ONLY
     default_from_api: true
+  - !ruby/object:Api::Type::String
+    name: 'cloudRouterIpv6Address'
+    description: |
+      IPv6 address + prefix length to be configured on Cloud Router
+      Interface for this interconnect attachment.
+    output: true
+  - !ruby/object:Api::Type::String
+    name: 'customerRouterIpv6Address'
+    description: |
+      IPv6 address + prefix length to be configured on the customer
+      router subinterface for this interconnect attachment.
+    output: true


ehorning:
I'm not sure why the lint check is failing here -- yamllint seems to run through fine when I run it against this file locally

rileykarson:
@trodge any idea what's up w/ the linter? Seems to be a missing dependency.

/gcbrun

mmv1/products/compute/InterconnectAttachment.yaml
```diff
@@ -312,3 +312,15 @@ properties:
       - :IPV4_IPV6
       - :IPV4_ONLY
     default_from_api: true
+  - !ruby/object:Api::Type::String
+    name: 'cloudRouterIpv6Address'
```
rileykarson, line 316: Would it be feasible to add a test that's IPV4_IPV6? If there's a significant setup cost we are fine without it, but both current tests for this resource are IPV4_ONLY.

ehorning: The attribute added here isn't easily testable for the reasons noted in the PR: https://github.com/GoogleCloudPlatform/magic-modules/pull/9890

While I didn't check whether these output-only attributes are present without that, my guess is that without setting the `stack_type` to be both, the outputs of these attributes will be empty, so there's not much point in testing them without having the `stack_type` attribute set to support ipv6.

rileykarson: That tracks, thanks!

rileykarson (approved):
LGTM- the lint issue appears to be resolved in https://github.com/GoogleCloudPlatform/magic-modules/pull/10269, so post-merge shouldn't cause an issue.

PR #10276, opened 2024-03-25
Update default TF version used in TeamCity to be `"1.8.0-rc1"`

SarahFrench:
This will be updated to 1.8.0 after the GA release

```release-note:none

```


SarahFrench:
Note: the `TeamCity Configuration Tests / teamcity-config-tests (pull_request)` check is failing due to dependency issues, not due to the configuration. I don't know how to address the missing dependency from TeamCity, but the change in this PR is very low risk so the plan is to merge despite the check failing.

PR #10278, opened 2024-03-25
Changelog linting

shuyama1:
This PR added a GHA for changelog unit tests and 3 more changelog linting rules:
1. Only one line is allowed per entry
2. changelog for new resource or datasource should be in format "`google_resource/datasource_name`"
3. changelog for enhancement or bug fix should be in format "package: change details"

No chnagelog error: https://github.com/GoogleCloudPlatform/magic-modules/actions/runs/8485433892/job/23250199809

Some example errors: https://github.com/GoogleCloudPlatform/magic-modules/actions/runs/8485475554/job/23250325859

```release-note:none
```

### DIFF
--- a/.github/workflows/changelog-checker.yml
+++ b/.github/workflows/changelog-checker.yml
@@ -14,6 +14,17 @@ jobs:
         uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.2
         with:
           path: repo
+
+      - name: Set up Go
+        uses: actions/setup-go@v4
+        with:
+          go-version: '^1.21.0'
+
+      - name: Build
+        run: |
+          cd repo/tools/go-changelog/cmd/changelog-pr-body-check
+          go build
+
       - name: Check Changelog
         env:
           GITHUB_OWNER: GoogleCloudPlatform
@@ -21,4 +32,4 @@ jobs:
           GITHUB_TOKEN: ${{github.token}}
         run: |
           cd repo/tools/go-changelog/cmd/changelog-pr-body-check
-          go run . ${{github.event.pull_request.number}}
+          ./changelog-pr-body-check ${{github.event.pull_request.number}}
--- a/.github/workflows/changelog-checker.yml
+++ b/.github/workflows/changelog-checker.yml
@@ -16,7 +16,7 @@ jobs:
           path: repo
 
       - name: Set up Go
-        uses: actions/setup-go@v4
+        uses: actions/setup-go@0c52d547c9bc32b1aa3301fd7a9cb496313a4491 # v5.0.0
         with:
           go-version: '^1.21.0'
 
--- a/.github/workflows/unit-test-go-changelog.yml
+++ b/.github/workflows/unit-test-go-changelog.yml
@@ -20,10 +20,10 @@ jobs:
       
       - name: Build
         run: |
-          cd repo/tools/go-changelog/cmd/changelog-pr-body-check
+          cd tools/go-changelog/cmd/changelog-pr-body-check
           go build
 
       - name: Test
         run: |
-          cd repo/tools/go-changelog
+          cd tools/go-changelog
           go test -v ./...
--- a/.github/workflows/unit-test-go-changelog.yml
+++ b/.github/workflows/unit-test-go-changelog.yml
@@ -1,0 +1,29 @@
+name: unit-tests-go-changelog
+
+permissions: read-all
+
+on:
+  pull_request:
+    paths:
+      - 'tools/go-changelog/**'
+
+jobs:
+  test:
+    runs-on: ubuntu-22.04
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Set up Go
+        uses: actions/setup-go@v4
+        with:
+          go-version: '^1.21.0'
+      
+      - name: Build
+        run: |
+          cd repo/tools/go-changelog/cmd/changelog-pr-body-check
+          go build
+
+      - name: Test
+        run: |
+          cd repo/tools/go-changelog
+          go test -v ./...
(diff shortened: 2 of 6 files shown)


tools/go-changelog/note.go
```diff
@@ -48,7 +64,7 @@ func NotesFromEntry(entry Entry) []Note {
 				}
 			}
 
-			note = strings.TrimSpace(note)
+			// note = strings.TrimSpace(note)
```
shuyama1, line 67: We need to skip this here so that we can detect changelogs that contain multiple lines

tools/go-changelog/cmd/changelog-pr-body-check/main.go
```diff
@@ -57,31 +58,23 @@ func main() {
...
-			}
-			body += " unknown release-note type"
-			if len(unknownTypes) > 1 {
-				body += "s"
-			}
-			body += " in your changelog entries:"
-			for _, t := range unknownTypes {
-				body += "\n* " + t
+	if errors := entry.Validate(); errors != nil {
+		log.Printf("error parsing changelog entry in %s: %s", entry.Issue, errors)
```
melinath, line 62: Since each of the errors will get printed out individually below, this output seems like it will just be a (less readable) duplicate... maybe it would be okay to leave out?

tools/go-changelog/cmd/changelog-pr-body-check/main.go
```diff
@@ -57,31 +58,23 @@ func main() {
...
+		body := "\nOops! Some errors are detected for your changelog entries:"
+		for i, err := range errors {
+			switch err.Code {
+			case changelog.EntryErrorNotFound:
+				body += fmt.Sprintf("\n\n Issue %d: It looks like no changelog entry is attached to this PR. Please include a release note block in the PR body, as described in https://googlecloudplatform.github.io/magic-modules/contribute/release-notes/", i+1)
+			case changelog.EntryErrorUnknownTypes:
+				body += fmt.Sprintf("\n\n Issue %d: unknown changelog types %v \nPlease only use the types listed in https://googlecloudplatform.github.io/magic-modules/contribute/release-notes/.", i+1, err.Details["type"].(string))
```
melinath, line 69: Rather than including each changelog inline here, what would you think about grouping the issues by changelog entry? So something like:

```
Changelog:
\`\`\`release-note:enhancement
compute engine: fixed an issue in google_resource resource 
\`\`\`

Issues:
  - invalid format. Please follow format in https://googlecloudplatform.github.io/magic-modules/contribute/release-notes/#type-specific-guidelines-and-examples.
```

After typing this I realized that we probably currently only support one error per entry... so this would only make sense if we want to support multiple per entry.

This isn't a blocker; let me know what you think!

tools/go-changelog/note.go
melinath:
```suggestion
```

melinath (changes requested):
Could you modify the workflow for this to build the checker in a separate step so that the build output doesn't show up as part of the failure?

tools/go-changelog/cmd/changelog-pr-body-check/main.go
shuyama1: yeah, this makes sense

tools/go-changelog/cmd/changelog-pr-body-check/main.go
shuyama1: yes, we currently only support one error per entry. Plus, based on the current number of rules, I don't think we're gonna see multiple errors per changelog entry (to clarify, we do support multiple entries per PR though). But I could reform how the errors are reported. I can follow the example you have and make it more clear.

shuyama1:
I updated the PR description to include new GHA runs.

.github/workflows/changelog-checker.yml
```diff
@@ -14,11 +14,22 @@ jobs:
         uses: actions/checkout@v4
         with:
           path: repo
+
+      - name: Set up Go
+        uses: actions/setup-go@v4
```
melinath, line 19: This should use a specific hash rather than just `v4`. I think it could be worth also rebasing or merging main, which already has actions/checkout pinned.

.github/workflows/unit-test-go-changelog.yml
```diff
@@ -0,0 +1,29 @@
...
+on:
+  pull_request:
+    paths:
+      - 'tools/go-changelog/**'
+
+jobs:
+  test:
+    runs-on: ubuntu-22.04
+    steps:
+      - uses: actions/checkout@v4
```
melinath, line 14: same comment here - these should be pinned to a specific hash.

melinath (approved):
LGTM apart from a couple minor changes to avoid scorecard violations

melinath (approved):
could you also make an invalid change just to double-check that the printing logic works as expected?

shuyama1:
> could you also make an invalid change just to double-check that the printing logic works as expected?

Tested the latest version against some invalid changelogs:

No chnagelog error: https://github.com/GoogleCloudPlatform/magic-modules/actions/runs/8485433892/job/23250199809

Some example errors: https://github.com/GoogleCloudPlatform/magic-modules/actions/runs/8485475554/job/23250325859
(with the following changelog entries)

```release-note:note
this changelog contains a newline

```

```release-note:feature
this is to add a feature (invalid type)
```

```release-note:new-resource
google_resource_without_backticks
```

```release-note:enhancement
added a new field to `google_resource` resource
```

```release-note:bug
compute engine: fixed an issue in `google_resource` resource
```

```release-note:new-datasource
`google_new_datasource_valid_format`
```

```release-note:enhancement
compute: added a new field to `google_resource` resource (valid changlog, no error's expected)
```

melinath (approved):
LGTM

PR #10279, opened 2024-03-25
Resource .go.tmpl conversion base

NickElliot:
Converts the top portion of the file up to `create()` minus schema fields. Next PR will be the schema fields.

Currently missing functionality for compiling custom code into the resource file.

Larger piece of logic has been removed at the equivalent within line 99/170 of resource.erb — removed due to only being used in one downstream resource and would require a separate custom function to parse within the `.go.tmpl` in the first place -- could simply add that small piece as relevant to those files as a customize diff function and custom schema entry.

### DIFF
--- a/mmv1/provider/template_data.go
+++ b/mmv1/provider/template_data.go
@@ -15,8 +15,6 @@ package provider
 
 import (
 	"bytes"
-	"fmt"
-	"go/format"
 	"log"
 	"os"
 	"path/filepath"
@@ -53,6 +51,7 @@ var TemplateFunctions = template.FuncMap{
 	"plural":     google.Plural,
 	"contains":   strings.Contains,
 	"join":       strings.Join,
+	"lower": 	  strings.ToLower,
 }
 
 var GA_VERSION = "ga"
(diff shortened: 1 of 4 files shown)


NickElliot:
cc @zli82016

mmv1/provider/template_data.go
```diff
@@ -105,13 +104,6 @@ func (td *TemplateData) GenerateFile(filePath, templatePath string, resource api
 		sourceByte = bytes.Replace(sourceByte, []byte("github.com/hashicorp/terraform-provider-google/google"), []byte(td.TerraformProviderModule+"/"+td.TerraformResourceDirectory), -1)
 	}
```
c2thorn, line 106: Why delete this part?

NickElliot: The file generation was failing due to the gofmt breaking (which is expected due to the file still incomplete and having spots of broken syntax) so I had to remove it for working down the file -- I'll restore it for the PRs or could swap it to a comment while these are worked through

c2thorn (approved):

PR #10288, opened 2024-03-25
Release provider-defined functions

SarahFrench:
**Release Note Template for Downstream PRs (will be copied)**

```release-note:note
provider: introduced support for provider-defined functions. This feature is in Terraform v1.8.0+.
```
```release-note:enhancement
provider: added provider-defined function `project_from_id` for retrieving the project id from a resource's self link or id
```
```release-note:enhancement
provider: added provider-defined functions `location_from_id`, `region_from_id`, and `zone_from_id` for retrieving the location/region/zone names from a resource's self link or id
```
```release-note:enhancement
provider: added provider-defined function `name_from_id` for retrieving the short-form name of a resource from its self link or id
```
```release-note:enhancement
provider: added provider-defined function `region_from_zone` for deriving a region from a zone's name
```

PR #10088, opened 2024-02-29
TeamCity: Add project for testing the provider functions feature branch

SarahFrench:
Partially addresses https://github.com/hashicorp/terraform-provider-google/issues/17451

This PR adds a new project in TeamCity to enable testing of provider-defined functions:

- Tests the FEATURE-BRANCH-provider-functions branch in the downstreams on a schedule
- Allows testing provider-defined functions from PRs in the MM upstream forks
- Runs those tests with an alpha release of TF 1.8.0 while leaving our nightly tests etc unaffected (using the default version of Terraform - 1.2.5)

Screenshot 2024-02-29 at 12 06 20

**Release Note Template for Downstream PRs (will be copied)**

```release-note:none

```

PR #10292, opened 2024-03-26
remove `FEATURE-branch-provider-functions` project from TeamCity folder

BBBmau:
This PR is meant to be merged once https://github.com/GoogleCloudPlatform/magic-modules/pull/10288 is merged into main.

This addresses the project that was added here https://github.com/GoogleCloudPlatform/magic-modules/pull/10088

```release-note:none

```


SarahFrench (approved):
Despite the problems using TF 1.8.0 in TeamCity I think we're still ok to go ahead with removing these resources. If we want to test the provider-defined functions in TeamCity we can run custom builds with TF 1.8.0.

PR #10297, opened 2024-03-26
Add check for version guards in erb templates

roaks3:
We sometimes break downstream providers when invalid version guards are introduced in .erb files. This happens because we have 3+ versions to consider, with each being a superset of the next:

"private" or "alpha" -> "beta" -> "ga"

The typical version guard would be `<% unless version == 'ga' -%>`, which means the feature should not be available in the GA provider, but should be in Beta and Private/Alpha. Sometimes users will assume `<% if version == 'beta' -%>` is equivalent, but it means the feature will NOT be available in Private/Alpha, which is almost always wrong.

In short, our features should never target only beta, because it would break the assumption of Private/Alpha being a superset. This logic also works in reverse (where we wouldn't want any GA-specific functionality to also be present in Private/Alpha).

```release-note:none

```

### DIFF
--- a/tools/template-check/go.mod
+++ b/tools/template-check/go.mod
@@ -1,0 +1,3 @@
+module github.com/GoogleCloudPlatform/magic-modules/tools/template-check
+
+go 1.21
--- a/tools/template-check/main.go
+++ b/tools/template-check/main.go
@@ -1,0 +1,66 @@
+package main
+
+import (
+	"bufio"
+	"flag"
+	"fmt"
+	"os"
+
+	"github.com/GoogleCloudPlatform/magic-modules/tools/template-check/ruby"
+)
+
+func isValidTemplate(filename string) (bool, error) {
+	results, err := ruby.CheckVersionGuardsForFile(filename)
+	if err != nil {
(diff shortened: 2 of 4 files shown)


roaks3:
@c2thorn this isn't complete yet, but I was wondering if you could give it a quick look to make sure the approach seems reasonable, and will not be too much work to change later (assuming we ultimately stop using erb templates). I'm planning to have it run with the other mm checks against all erb files that have changed in the submitted diff.

c2thorn:
This would run as a separate check on PR's?

Yeah I think this is a good approach. Seems easy enough to swap out the identifying text when we swap to Go `.tmpl` templates. It will eventually look something along the lines of `{{- if eq $.MinVersion "beta" }}`

roaks3:
Great, thanks!

Yea, it would be a separate PR check, something to the effect of `git diff --name-only ../../*.erb | go run main.go`

roaks3:
Note: this is ready for review now. The plan is to run this command in a check as a separate PR.

c2thorn (approved):

PR #10299, opened 2024-03-27
Fix client request only properties.

rohitjangid:
Fix client request only properties to ignore read from api. Update examples to match actual usage.

```release-note:none
```

### DIFF
--- a/mmv1/products/integrations/Client.yaml
+++ b/mmv1/products/integrations/Client.yaml
@@ -48,7 +48,7 @@ properties:
     description: |
       Cloud KMS config for AuthModule to encrypt/decrypt credentials.
     immutable: true
-    url_param_only: true
+    ignore_read: true
     properties:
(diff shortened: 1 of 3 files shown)


SarahFrench (commented):
Thanks @rohitjangid ! This PR looks ok to merge, but I wanted to check: is this fixing a user-facing issue?

I want to make sure the release note is accurate. If this PR just changes tests and internals of the resource in a way that's not noticeable to users I'd remove the release note. If there is a bug that's fixed by this PR I can update the release note to describe it

rohitjangid:
@SarahFrench This resource is not yet released in the latest terraform provide version. Since this is not being used by any users, there is no user impact.

SarahFrench (approved):
Looks good - I've set the release note to `none` as I understand this isn't fixing a user-facing issue

PR #10303, opened 2024-03-27
Fix crash when `parameter.enableStreamingEngine` is set on `google_dataflow_flex_template_job`, update resource's docs

SarahFrench:
Fixes https://github.com/hashicorp/terraform-provider-google/issues/16713

Potentially addresses some concerns in https://github.com/hashicorp/terraform-provider-google/issues/9291 via the documentation update.

**Note: all the relevant acceptance tests for this resource are skipped in VCR**

Also, this PR updates the docs for `google_dataflow_flex_template_job`, which had numerous missing fields.

---

```release-note:bug
dataflow: fixed an issue where the provider would crash when `enableStreamingEngine` is set as a `parameter` value in `google_dataflow_flex_template_job`
```

### DIFF
--- a/mmv1/third_party/terraform/services/dataflow/resource_dataflow_flex_template_job.go.erb
+++ b/mmv1/third_party/terraform/services/dataflow/resource_dataflow_flex_template_job.go.erb
@@ -366,11 +366,12 @@ func resourceDataflowFlexJobSetupEnv(d *schema.ResourceData, config *transport_t
 		delete(updatedParameters, "enableStreamingEngine")
 		e := strings.ToLower(p.(string))
 		switch e {
-		case "true", "1":
+		case "true":
 			enableStreamingEngine = true
-		case "false", "0":
+		case "false":
 			enableStreamingEngine = false
 		default:
+			log.Printf("[DEBUG] Map entry `parameters.enableStreamingEngine` has unexpected value `%s`. It should be either \"true\" or \"false\". Defaulting to false boolean value instead.", e)
 			enableStreamingEngine = false
 		}
 	} else {
--- a/mmv1/third_party/terraform/services/dataflow/resource_dataflow_flex_template_job.go.erb
+++ b/mmv1/third_party/terraform/services/dataflow/resource_dataflow_flex_template_job.go.erb
@@ -366,12 +366,12 @@ func resourceDataflowFlexJobSetupEnv(d *schema.ResourceData, config *transport_t
 		delete(updatedParameters, "enableStreamingEngine")
 		e := strings.ToLower(p.(string))
 		switch e {
-			case "true", "1"
-				enableStreamingEngine = true
-			case "false", "0"
-				enableStreamingEngine = false
-			default
-				enableStreamingEngine = false
+		case "true", "1":
+			enableStreamingEngine = true
+		case "false", "0":
+			enableStreamingEngine = false
+		default:
+			enableStreamingEngine = false
 		}
 	} else {
 		if v, ok := d.GetOk("enable_streaming_engine"); ok {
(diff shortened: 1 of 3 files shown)


SarahFrench:
I need to get some debug logs that show what format the data is in (true vs 1 etc). I'll run tests off this PR in TeamCity to get debug logs, as all tests for DataFlow are VCR-skipped

mmv1/third_party/terraform/services/dataflow/resource_dataflow_flex_template_job.go.erb
```diff
@@ -364,7 +364,16 @@ func resourceDataflowFlexJobSetupEnv(d *schema.ResourceData, config *transport_t
 	var enableStreamingEngine bool
 	if p, ok := d.GetOk("parameters.enableStreamingEngine"); ok {
 		delete(updatedParameters, "enableStreamingEngine")
-		enableStreamingEngine = p.(bool)
+		e := strings.ToLower(p.(string))
```
SarahFrench, line 367: My understanding of the original bug is that Map type fields are stored as string:string maps in state, so even if the value in the config is a boolean true/false it will be converted to "true"/"false" in state.

SarahFrench:
I've triggered the Dataflow tests to run here: https://hashicorp.teamcity.com/buildConfiguration/TerraformProviders_GoogleCloud_GOOGLE_BETA_MMUPSTREAMTESTS_GOOGLEBETA_PACKAGE_DATAFLOW/124616?buildTypeTab=overview&branch=refs%2Fheads%2Fauto-pr-10303&mode=builds&buildTab=overview&hideTestsFromDependencies=false&hideProblemsFromDependencies=false

mmv1/third_party/terraform/services/dataflow/resource_dataflow_flex_template_job.go.erb
```diff
@@ -364,7 +364,16 @@ func resourceDataflowFlexJobSetupEnv(d *schema.ResourceData, config *transport_t
...
+		switch e {
+		case "true":
+			enableStreamingEngine = true
+		case "false":
+			enableStreamingEngine = false
+		default:
+			log.Printf("[DEBUG] Map entry `parameters.enableStreamingEngine` has unexpected value `%s`. It should be either \"true\" or \"false\". Defaulting to false boolean value instead.", e)
```
NickElliot, line 374: Would it not be better to actually return this as a full failure (even if for debug purposes only) -- because these fields are immutable, if somehow a user triggered this (unlikely) scenario while intending the other outcome, they would then require a full recreation

NickElliot (approved):
LGTM -- one comment on a nuance of the error rules but if you don't believe it would be an issue then all is well to merge.

mmv1/third_party/terraform/services/dataflow/resource_dataflow_flex_template_job.go.erb
SarahFrench: [15dbe03](https://github.com/GoogleCloudPlatform/magic-modules/pull/10303/commits/15dbe0344d8df5fd3cffe122ab06413b210a2210)

SarahFrench:
Tests are currently running in TeamCity on this PR: https://hashicorp.teamcity.com/buildConfiguration/TerraformProviders_GoogleCloud_GOOGLE_BETA_MMUPSTREAMTESTS_GOOGLEBETA_PACKAGE_DATAFLOW/124617?buildTab=overview&hideTestsFromDependencies=false&hideProblemsFromDependencies=false&expandBuildTestsSection=true&expandBuildChangesSection=true

The tests that are failing have been failing in the nightly tests for a while so I'm not concerned about them

SarahFrench:
Tests in TeamCity finished and the failures are all prexisting issues - merging PR. Thanks for the speedy review Nick!

PR #10307, opened 2024-03-27
Test Fix: TestAccDataprocCluster_spotWithAuxiliaryNodeGroups

Harwayne:
Fixes: https://github.com/hashicorp/terraform-provider-google/issues/17534

Alter two properties of the Cluster created during testing so that it can be successfully created:

1. Set the `accelerator_type_uri` to `nvidia-tesla-t4`. 
    - This could be `nvidia-tesla-k80`, but that will no longer be usable as of May 1, 2024. See https://cloud.google.com/compute/docs/eol/k80-eol
3. Set the minimum CPU platform to `Intel Haswell`.

I have directly created a VM in GCE with these configurations and seen it run successfully. GCE VM:

```
{
  "cpuPlatform": "Intel Haswell",
  "guestAccelerators": [
    {
      "acceleratorCount": 1,
      "acceleratorType": "projects/REDACTED/zones/us-central1-b/acceleratorTypes/nvidia-tesla-t4"
    }
  ],
  "machineType": "projects/REDACTED/zones/us-central1-b/machineTypes/n1-standard-2",
  "status": "RUNNING",
  ...
}
```

**Release Note Template for Downstream PRs (will be copied)**

```release-note:none

```

PR #10309, opened 2024-03-27
Test Fix: Replace Dataproc's usage of K80 in tests with T4

Harwayne:
Replace Dataproc's usage of K80 in tests with T4.

GCE has declared K80 EOL and no longer usable as of May 1, 2024. See https://cloud.google.com/compute/docs/eol/k80-eol. 

```release-note:none

```


Harwayne:
The error in the test is being fixed by #10307.

NickElliot (approved):
LGTM

PR #10311, opened 2024-03-27
Return affected resources from diff-processor instead of labels

melinath:
We already need to convert affected resources to labels in generate_comment (because of changed file handling), so we can skip this step as part of the diff-processor command. This also lets us clean up / simplify the logic around labeling a bit more.

```release-note:none

```

### DIFF
--- a/.ci/magician/cmd/generate_comment.go
+++ b/.ci/magician/cmd/generate_comment.go
@@ -229,8 +229,8 @@ func execGenerateComment(prNumber int, ghTokenMagicModules, buildId, buildStep, 
 	data.Diffs = diffs
 
 	// The breaking changes are unique across both provider versions
+	uniqueAffectedResources := map[string]struct{}{}
 	uniqueBreakingChanges := map[string]struct{}{}
-	uniqueServiceLabels := map[string]struct{}{}
 	diffProcessorPath := filepath.Join(mmLocalPath, "tools", "diff-processor")
 	diffProcessorEnv := map[string]string{
 		"OLD_REF": oldBranch,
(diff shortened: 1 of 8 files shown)


tools/diff-processor/cmd/changed_schema_resources.go
```diff
@@ -10,51 +10,42 @@ import (
 	"os"
 
 	"github.com/GoogleCloudPlatform/magic-modules/tools/diff-processor/diff"
-	"github.com/GoogleCloudPlatform/magic-modules/tools/issue-labeler/labeler"
 	"github.com/spf13/cobra"
 	"golang.org/x/exp/maps"
 )
 
-const changedSchemaLabelsDesc = `Compute service labels based on schema changes.`
+const changedSchemaResourcesDesc = `Compute service labels based on schema changes.`
```
shuyama1, line 17: we may want to change the wording since we no longer compute the label?

.ci/magician/cmd/generate_comment.go
```diff
@@ -265,58 +265,68 @@ func execGenerateComment(prNumber int, ghTokenMagicModules, buildId, buildStep,
...
 	for _, repo := range []source.Repo{tpgRepo, tpgbRepo} {
 		if !repo.Cloned {
 			fmt.Println("Skipping changed file service labels; repo failed to clone: ", repo.Name)
 			continue
 		}
 		for _, path := range repo.ChangedFiles {
 			if r := fileToResource(path); r != "" {
-				affectedResources[r] = struct{}{}
+				uniqueAffectedResources[r] = struct{}{}
+				changedFilesAffectedResources[r] = struct{}{}
```
shuyama1, line 291: Just want to confirm my understanding is correct - we use `changedFilesAffectedResources` only to log the affected resources based on changed files, and  `uniqueAffectedResources` for converting affected resources(including detected from file changes or schema changes) - so we need to added the affected resource to  both sets here

.ci/magician/cmd/generate_comment.go
```diff
@@ -265,58 +265,68 @@ func execGenerateComment(prNumber int, ghTokenMagicModules, buildId, buildStep,
...
-			fmt.Printf("Error posting new service labels %q: %s", serviceLabelsSlice, err)
-			errors["Other"] = append(errors["Other"], "Failed to update service labels")
+		// short-circuit if service labels have already been added to the PR
+		hasServiceLabels := false
+		for _, label := range pullRequest.Labels {
+			if strings.HasPrefix(label.Name, "service/") {
+				hasServiceLabels = true
+			}
+		}
+		if !hasServiceLabels {
```
shuyama1, line 319: I might be misreading it. Does it prevent of adding new service labels if new affected resources are detected (potentially result in new service labels are needed) in any following commits of PRs

shuyama1 (commented):
Mostly LGTM. Left some comments + questions

tools/diff-processor/cmd/changed_schema_resources.go
melinath: Fixed in 79f8b5e6b

.ci/magician/cmd/generate_comment.go
melinath: yes - I added the second set only for printing debugging information.

.ci/magician/cmd/generate_comment.go
melinath: yes - we only add service labels if there aren't already service labels. The idea is that if something is incorrectly identified, we'd want to be able to manually set the service labels and not have them changed by automation. (This is the intended behavior, but it could be that we want to change it in the future if it ends up causing problems.)

shuyama1 (approved):

PR #10312, opened 2024-03-28
Add `project_number` field to google_storage_bucket resource and datasource, enable providing `project` argument to data source

SarahFrench:
Closes https://github.com/hashicorp/terraform-provider-google/issues/17166

This PR aims to address an edge case where users do not want to give the identity/service account used by Terraform permissions on the Compute API and want to provision GCS buckets. Previously if a project name was not supplied to the data source via provider defaults then the data source would:
- retrieve information about the bucket from the Storage API, including the project _number_ it's linked to
- call the Compute API to get a human-readable _project id_ for that project number, and use that to populate the `project` field

By letting users set the project field on the data source they can avoid that call to the Compute API. There is a chance that the project id they supply doesn't match the project number linked to the GCS bucket. This isn't ideal, but was [OK'd during triage as it matches the existing behaviour on import](https://github.com/hashicorp/terraform-provider-google/issues/17166#issuecomment-1927651371).

This PR also adds a new field, project_number, to surface the information returned from the Storage API which is guaranteed to be accurate. This enables users to perform any checks etc to verify the project id and number are a correct pairing.

```release-note:enhancement
storage: added `project_number` attribute to `google_storage_bucket` resource and data source
```

```release-note:enhancement
storage: added ability to provide `project` argument to `google_storage_bucket` data source. This will not impact reading the resource's data, instead this helps users avoid calls to the Compute API within the data source.
```

### DIFF
--- a/mmv1/third_party/terraform/services/storage/data_source_google_storage_bucket.go
+++ b/mmv1/third_party/terraform/services/storage/data_source_google_storage_bucket.go
@@ -12,6 +12,7 @@ func DataSourceGoogleStorageBucket() *schema.Resource {
 
 	dsSchema := tpgresource.DatasourceSchemaFromResourceSchema(ResourceStorageBucket().Schema)
 
+	tpgresource.AddOptionalFieldsToSchema(dsSchema, "project")
 	tpgresource.AddRequiredFieldsToSchema(dsSchema, "name")
 
 	return &schema.Resource{
--- a/mmv1/third_party/terraform/services/storage/data_source_google_storage_bucket_test.go
+++ b/mmv1/third_party/terraform/services/storage/data_source_google_storage_bucket_test.go
@@ -1,17 +1,19 @@
 package storage_test
 
 import (
-	"fmt"
 	"testing"
 
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
 	"github.com/hashicorp/terraform-provider-google/google/acctest"
+	"github.com/hashicorp/terraform-provider-google/google/envvar"
 )
 
 func TestAccDataSourceGoogleStorageBucket_basic(t *testing.T) {
 	t.Parallel()
 
-	bucket := "tf-bucket-" + acctest.RandString(t, 10)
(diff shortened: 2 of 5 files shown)


zli82016:
> By letting users set the project field on the data source they can avoid that call to the Compute API. There is a chance that the project id they supply doesn't match the project number linked to the GCS bucket. This isn't ideal, but was [OK'd during triage as it matches the existing behaviour on import](https://github.com/hashicorp/terraform-provider-google/issues/17166#issuecomment-1927651371).
> 

Instead of adding the project field to data source, maybe we can consider creating a new function for data source, similar to [setStorageBucket](https://github.com/hashicorp/terraform-provider-google/blob/32e645dbd0175d8baa4dc2b0a33eafc11877d745/google/services/storage/resource_storage_bucket.go#L1630), but without the project checking code. It will avoid the incorrect project provided by the users for the data source.

SarahFrench:
Yeah that's another option & I'm happy to pursue it. That function is trying to cover a lot of scenarios!

I'll probably pick that up next week, as I'm OOO Friday-Monday

SarahFrench:
Ah, I just thought. A change like that would be a breaking change for all users who are happy with the resource interacting with the Compute API. I'm opening this PR for a small minority of users who don't want Compute API to be used inside this data source but there may be users out there relying on this behaviour

zli82016:
> Ah, I just thought. A change like that would be a breaking change for all users who are happy with the resource interacting with the Compute API. I'm opening this PR for a small minority of users who don't want Compute API to be used inside this data source but there may be users out there relying on this behaviour

I see. I didn't realize that.  We can go with the current change. Can we also add the new field to the [data source documentation ](https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/third_party/terraform/website/docs/d/storage_bucket.html.markdown?plain=1) and add the note to provide the correct project? Thanks.

zli82016 (approved):
LGTM if all of the checks pass.

SarahFrench:
Thanks!

PR #10317, opened 2024-03-28
Set default values for Go properties

zli82016:
Set default values for Go properties

```release-note:none

```


mmv1/api/type.go
```diff
@@ -270,7 +275,47 @@ type Type struct {
 
 const MAX_NAME = 20
 
-// func (t *Type) validate() {
+func (t *Type) Validate(r *Resource) {
```
c2thorn, line 278: This isn't really a validation function anymore is it? It looks like it really just sets default values. What do you think?

zli82016: Yes, here it sets default values. We can have two functions to separate setting default values and validations (if the field is required or not).

c2thorn (approved):

PR #10319, opened 2024-03-28
Fix crash caused by nil value in BigQuery Table

slevenick:
Fixes: https://github.com/hashicorp/terraform-provider-google/issues/17708

I don't want to add a test because it's impossible to provision a table with an empty string as the column. This now just triggers the correct 400 from the API rather than crashing

```release-note:bug
bigquery: fixed a crash when `google_bigquery_table` had a `primary_key.columns` entry set to `""`
```


zli82016 (approved):
LGTM

PR #10321, opened 2024-03-28
Made it so that authors will never have their review requested on their own PR

melinath:
Fixed https://github.com/hashicorp/terraform-provider-google/issues/17722

Unfortunately it's not possible to avoid exiting because this build step doesn't currently have access to a commit sha, which is necessary for posting a build status (instead of just failing). I'll add it to the yaml config so we have the option later, but I think adding support for that is a lower priority.

```release-note:none

```


roaks3 (approved):
Makes sense, thanks for the quick fix!

melinath:
TGC failure looks unrelated - merging.

PR #10326, opened 2024-03-29
Add support for instance_ip_mode to appengine flexible in beta

gwendal-lecren:
Add support for instance_ip_mode in google_app_engine_flexible_app_version resource in the beta provider.
Not sure it is the right place to open a PR, but trying to address this issue:
https://github.com/hashicorp/terraform-provider-google/issues/17510

---

**Release note**

```release-note:enhancement
appengine: added field `instance_ip_mode` to resource `google_app_engine_flexible_app_version` resource (beta)
```


mmv1/products/appengine/FlexibleAppVersion.yaml
```diff
@@ -139,6 +139,16 @@ properties:
         description: |
           List of ports, or port pairs, to forward from the virtual machine to the application container.
         item_type: Api::Type::String
+      - !ruby/object:Api::Type::Enum
+        name: 'instanceIpMode'
+        description: |
+          Prevent instances from receiving an ephemeral external IP address.
+        min_version: beta
+        required: false
+        default_value: :external
```
rileykarson, line 148:
```suggestion
```

The API doesn't return the default explicitly for old resources, so we're unable to set a default clientside otherwise we'd break against existing resources!

mmv1/products/appengine/FlexibleAppVersion.yaml
```diff
@@ -139,6 +139,16 @@ properties:
         description: |
           List of ports, or port pairs, to forward from the virtual machine to the application container.
         item_type: Api::Type::String
+      - !ruby/object:Api::Type::Enum
+        name: 'instanceIpMode'
```
rileykarson, line 143: We'll want to use this field in at least one test! You can modify https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/third_party/terraform/services/appengine/resource_app_engine_flexible_app_version_test.go (and ideally we'd switch values between the two steps in update, to ensure that works)

mmv1/products/appengine/FlexibleAppVersion.yaml
```diff
@@ -139,6 +139,16 @@ properties:
...
+        description: |
+          Prevent instances from receiving an ephemeral external IP address.
+        min_version: beta
+        required: false
```
rileykarson, line 147:
```suggestion
```

nit: Omitted required implies optional

mmv1/products/appengine/FlexibleAppVersion.yaml
gwendal-lecren: Done

gwendal-lecren:
Hello @rileykarson - Does it look good to you or do you have any other feedback I should address? Thanks! :)

mmv1/products/appengine/FlexibleAppVersion.yaml
```diff
@@ -139,6 +139,14 @@ properties:
         description: |
           List of ports, or port pairs, to forward from the virtual machine to the application container.
         item_type: Api::Type::String
+      - !ruby/object:Api::Type::Enum
+        name: 'instanceIpMode'
+        description: |
+          Prevent instances from receiving an ephemeral external IP address.
+        min_version: beta
```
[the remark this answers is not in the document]
gwendal-lecren: This field is only available in the beta API, hence it should only be available in the beta provider (that's my understanding).
Hence I specified beta in `min_version`, but not sure it is the way I should do it?

cf https://cloud.google.com/appengine/docs/admin-api/reference/rest/v1beta/apps.services.versions#network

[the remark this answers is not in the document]
gwendal-lecren: Done

mmv1/products/appengine/product.yaml
```diff
@@ -18,5 +18,8 @@ versions:
   - !ruby/object:Api::Product::Version
     name: ga
     base_url: https://appengine.googleapis.com/v1/
+  - !ruby/object:Api::Product::Version
+    name: beta
+    base_url: https://appengine.googleapis.com/v1/
```
rileykarson, line 23:
```suggestion
    base_url: https://appengine.googleapis.com/v1beta/
```

mmv1/products/appengine/FlexibleAppVersion.yaml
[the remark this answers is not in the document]
rileykarson: Ah- since there's a beta API let's use that. Suggestion below.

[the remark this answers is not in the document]
gwendal-lecren: Indeed ^^ Thanks

gwendal-lecren:
Hello @rileykarson, hope you're doing well - Is there anything missing or does it look good on your side? Thanks! :)

rileykarson:
I didn't find any signal yesterday- opened https://github.com/GoogleCloudPlatform/magic-modules/pull/10419 to see if I can trigger the same failures with a diff. change.

rileykarson:
No failure- I am just gonna rerun these tests, the failure doesn't make much sense. It's a diff on `automatic_scaling`, which this doesn't touch.

/gcbrun

gwendal-lecren:
Hello @rileykarson - So it seems my changes bring a regression somehow. Would you have any details about the errors we are running into? Thanks

mmv1/products/appengine/FlexibleAppVersion.yaml
```diff
@@ -139,6 +139,14 @@ properties:
...
+        values:
+          - :external
```
rileykarson, line 148: EXTERNAL/INTERNAL per https://cloud.google.com/appengine/docs/admin-api/reference/rest/v1beta/apps.services.versions#instanceipmode

The API accepts these formats but returns the uppercase ones.

rileykarson (commented):
The main failure is this, probably due to the API version change:

```
          ~ update in-place
        
          # google_app_engine_standard_app_version.version will be updated in-place
          ~ resource "google_app_engine_standard_app_version" "version" {
                id                        = "apps/ci-test-project-188019/services/default/versions/fwl2j7401y"
                name                      = "apps/ci-test-project-188019/services/default/versions/fwl2j7401y"
                # (10 unchanged attributes hidden)
        
              - automatic_scaling {
                  - max_concurrent_requests = 0 -> null
                  - max_idle_instances      = 0 -> null
                  - min_idle_instances      = 0 -> null
                }
        
                # (3 unchanged blocks hidden)
            }
```

What's weird is that the API response doesn't seem to have changed. I'm not sure what's up there.

mmv1/products/appengine/FlexibleAppVersion.yaml
gwendal-lecren: OK let's try; will switch to uppercase.

gwendal-lecren: Updated

rileykarson:
These test failures are unrelated to your change it turns out, and are failing in nightlies when we always call live APIs. https://github.com/hashicorp/terraform-provider-google/issues/17531 and a bug to-be-filed for the IAP ones. My test of the tests was using the old results still- I'll resolve 'em out of band, our recording system will start using these results if I merge as-is.

rileykarson:
(If I haven't gotten back to this in a couple days, ping me)

mmv1/third_party/terraform/services/appengine/resource_app_engine_flexible_app_version_test.go
```diff
@@ -234,6 +235,7 @@ resource "google_app_engine_flexible_app_version" "foo" {
   network {
     name       = "default"
     subnetwork = "default"
+    instance_ip_mode = "INTERNAL"
```
rileykarson, line 238: `TestAccAppEngineFlexibleAppVersion_update`:

> Subnetwork 'default' does not have Private Google Access enabled. App Engine flexible instances using internal-only IP addresses require that Private Google Access be enabled in the target subnetwork. Have a Network Admin enable Private Google Access as described in: https://cloud.google.com/vpc/docs/configure-private-google-access#enabling-pga

At a glance, seems like we need to use a new network + subnetwork (instead of default) and configure https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/compute_subnetwork#private_ip_google_access on the subnetwork. We can modify the existing config(s) to do that.

rileykarson (commented):
`TestAccAppEngineDomainMapping_appEngineDomainMappingBasicExample`, `TestAccAppEngineDomainMapping_update`: Replace https://github.com/gwendal-lecren/magic-modules/blob/b805f0c9f713b2636866175898875be76adf6888/mmv1/products/appengine/DomainMapping.yaml#L81 with `default_from_api: true`, and delete https://github.com/GoogleCloudPlatform/magic-modules/blob/b27fe0a9d25ae29790edee20e14c93fa80bd3aa1/mmv1/templates/terraform/constants/domain_mapping.erb + https://github.com/GoogleCloudPlatform/magic-modules/blob/b27fe0a9d25ae29790edee20e14c93fa80bd3aa1/mmv1/templates/terraform/constants/go/domain_mapping.tmpl

mmv1/products/appengine/DomainMapping.yaml
```diff
@@ -78,7 +78,7 @@ properties:
     name: 'sslSettings'
     description: |
       SSL configuration for this domain. If unconfigured, this domain will not serve with SSL.
-    diff_suppress_func: 'sslSettingsDiffSuppress'
+    default_from_api: true
```
rileykarson, line 81: Agh, I missed a step- we need to remove L50/L51 in this file. We made the file redundant and deleted it, but still try to include the file. Sorry for the extra round trip!

The lines:

```diff
- custom_code: !ruby/object:Provider::Terraform::CustomCode
-  constants: templates/terraform/constants/domain_mapping.erb
```

gwendal-lecren: No worries :) Line removed.

rileykarson:
override-breaking-change: `Optional` + `Computed` is a superset of the previous DiffSuppressFunc.

rileykarson:
Ah, I misread the plan and AppEngineDomainMapping was doing the opposite of what I thought. Let me get back to you on Monday about what change we should make, sorry about that!

gwendal-lecren:
Hello @rileykarson, hope you're doing well - Would you have any idea about the fixes we should apply? Thanks :)

rileykarson:
Sorry! Busy couple of days on my end. Will address tomorrow.

mmv1/third_party/terraform/services/appengine/resource_app_engine_flexible_app_version_test.go
```diff
@@ -166,6 +167,19 @@ resource "google_project" "my_project" {
   billing_account = "%{billing_account}"
 }
 
+resource "google_compute_network" "network" {
+  name                    = "custom"
```
rileykarson, line 171: >       Error: Error updating FlexibleAppVersion "apps/tf-test-appeng-flexnr3ffe1qdr/services/default/versions/v1": googleapi: Error 400: App Engine flexible environment does not support changing networks (from 'default' to 'custom') with in place deployments. Please use the same network name, a different version name, or delete the existing version first.

We'll need to update the initial config in `testAccAppEngineFlexibleAppVersion_python` to include the same network as well!

mmv1/products/appengine/DomainMapping.yaml
```diff
@@ -78,7 +76,7 @@ properties:
...
```
rileykarson, line 79: OK- the issue we run into is that we get two different responses depending on the API version.

```
 "sslSettings": {
  "sslManagementType": "AUTOMATIC"
 }
```

is returned as the following in GA

```
  "sslSettings": {
    "sslManagementType": "AUTOMATIC",
    "pendingManagedCertificateId": "34199778"
  },
```

And the following in Beta

```
  "sslSettings": {
   "pendingManagedCertificateId": "34199778"
  },
```

I _suspect_ that `MANUAL` will still be correctly returned, but my test project was having issues so I wasn't able to verify. We may need to amend things if it doesn't get returned as well- the only test that covers it is the second stage of the _update test failing on the first stage so far.

---

To resolve this we'll want to introduce a decoder (https://googlecloudplatform.github.io/magic-modules/develop/custom-code/#decoder), similar to something like https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/templates/terraform/decoders/disk.erb. In decoders we have access to `res`, a `map[string]interface{}` representing the JSON response to Get.

In there you'll want to check if sslSettings is set. If it is, open its contents (Just like the `original` fields in the Disk sample I linked above). If sslManagementType is not set and pendingManagedCertificateId is, add the field sslManagementType with the value "AUTOMATIC". Then, just return the modified object.

Effectively, we're just rewriting the 2nd response into the first response in that code, which is what we already expect to handle.

mmv1/third_party/terraform/services/appengine/resource_app_engine_flexible_app_version_test.go
gwendal-lecren: Done :)

mmv1/products/appengine/DomainMapping.yaml
gwendal-lecren: I believe it's done ^^

rileykarson: Looks like it to me- running tests.

mmv1/products/appengine/DomainMapping.yaml
```diff
@@ -48,7 +48,7 @@ examples:
     name: 'app_engine_domain_mapping_basic'
     primary_resource_id: 'domain_mapping'
 custom_code: !ruby/object:Provider::Terraform::CustomCode
-  constants: templates/terraform/constants/domain_mapping.erb
+  decoder: templates/terraform/decoder/app_engine_domain_mapping.go.erb
```
rileykarson, line 51:
```suggestion
  decoder: templates/terraform/decoders/app_engine_domain_mapping.go.erb
```

rileykarson (commented):
(test results above are stale, the subsequent commit failed to build)

I'm gonna go ahead and apply this suggestion to save a round trip- if you're working locally you'll need to pull before making any future changes.

mmv1/templates/terraform/decoders/app_engine_domain_mapping.go.erb
```diff
@@ -0,0 +1,13 @@
...
+  // sslManagementType does not get returned with the beta endpoint when set to "AUTOMATIC"
+  // Hence if pendingManagedCertificateId is set and sslManagementType is not, enforce sslManagementType to "AUTOMATIC"
+  if _, ok := original["pendingManagedCertificateId"]; ok {
+    if _, ok := original["sslManagementType"]; !ok {
+      original["sslManagementType"] = "AUTOMATIC"
+    }
+  }
+  
+  res["sslSettings"] = original
+}
```
rileykarson, line 13:
```suggestion
}

return res, nil
```

rileykarson:
`terraform-google-conversion` failures appear to be a cache/timing issue

mmv1/templates/terraform/decoders/app_engine_domain_mapping.go.erb
```diff
@@ -0,0 +1,15 @@
+if v, ok := res["sslSettings"]; ok {
+  original := v.(map[string]interface{})
+
+  // sslManagementType does not get returned with the beta endpoint when set to "AUTOMATIC"
```
rileykarson, line 4:
```
              + ssl_settings {
                  + ssl_management_type = "MANUAL"
                }
```

OK- the beta API doesn't return it in either case. We'll want to update this comment and also set it to `MANUAL` when `pendingManagedCertificateId` is *not* present too.

mmv1/third_party/terraform/services/appengine/resource_app_engine_flexible_app_version_test.go
```diff
@@ -53,6 +53,19 @@ resource "google_project" "my_project" {
   billing_account = "%{billing_account}"
 }
 
+resource "google_compute_network" "network" {
```
rileykarson, line 56:
```suggestion
resource "google_compute_network" "network" {
  project = google_project.my_project.project_id
```

+ same thing for the subnetwork, and both subnetwork+network in the next step! We're creating these in the wrong project, I didn't catch it before because we were getting an error we couldn't update the network at all.

gwendal-lecren: Done

mmv1/templates/terraform/decoders/app_engine_domain_mapping.go.erb
gwendal-lecren: Thanks, updated as well.

mmv1/third_party/terraform/services/appengine/resource_app_engine_flexible_app_version_test.go
```diff
@@ -166,6 +182,21 @@ resource "google_project" "my_project" {
...
```
rileykarson, line 185: We'll need to use a `google_project_service` to enable `compute.googleapis.com` in the project & depend on that on this resource (in both configs)

mmv1/templates/terraform/decoders/app_engine_domain_mapping.go.erb
```diff
@@ -0,0 +1,18 @@
+if v, ok := res["sslSettings"]; ok {
```
rileykarson, line 1: The current test is failing because `sslSettings` isn't returned by the API. However, on further inspection, I think the logic I suggested is wrong, and it isn't always possible to determine whether a preexisting domain mapping is MANUAL or AUTOMATIC, as a manually configured certificate may also return a value in `pendingManagedCertificateId` (but none of our tests set it up to do so).

I think the logic we want to use instead is:

* if sslSettings is set
    * If sslManagementType is set, return that value
    * Otherwise, call `d.Get("ssl_settings.0.ssl_management_type")` and store that value in sslManagementType (carrying over the old value from state-  we do not have enough information to detect drift)
*  If ssl_settings was not set, call `d.Get("ssl_settings.0.ssl_management_type")`, create sslSettings, and store the retrieved value in sslManagementType

Sorry about the number of iterations here- this one's proving particularly gnarly to get right.

mmv1/third_party/terraform/services/appengine/resource_app_engine_flexible_app_version_test.go
gwendal-lecren: Done

mmv1/templates/terraform/decoders/app_engine_domain_mapping.go.erb
gwendal-lecren: No problem :) I believe it's done

mmv1/third_party/terraform/services/appengine/resource_app_engine_flexible_app_version_test.go
```diff
@@ -166,18 +187,38 @@ resource "google_project" "my_project" {
   billing_account = "%{billing_account}"
 }
 
-resource "google_app_engine_application" "app" {
-  project     = google_project.my_project.project_id
-  location_id = "us-central"
-}
-
 resource "google_project_service" "project" {
+  for_each = toset([
```
rileykarson, line 191:
```
        Error: Missing resource instance key
        
          on terraform_plugin_test.tf line 22, in resource "google_compute_network" "network":
          22:   project                 = google_project_service.project.project
        
        Because google_project_service.project has "for_each" set, its attributes
        must be accessed on specific instances.
        
        For example, to correlate with indices of a referring resource, use:
            google_project_service.project[each.key]
```

Let's just add separate blocks instead of using for_each here!

mmv1/products/appengine/DomainMapping.yaml
```diff
@@ -48,7 +48,7 @@ examples:
     name: 'app_engine_domain_mapping_basic'
```
rileykarson, line 48: For TestAccAppEngineDomainMapping_appEngineDomainMappingBasicExample we'll want to add an `ignore_read_extra` entry on `ssl_settings.0.ssl_management_type`, similar to https://github.com/GoogleCloudPlatform/magic-modules/blob/552d835ed5ba412770659dc926e07c5625ca1506/mmv1/products/apigateway/ApiConfig.yaml#L80-L81

rileykarson (commented):
DomainMapping is down to test-only failures! Comment inline for one of them. Failure is:

```
    vcr_utils.go:152: Step 2/2 error running import: ImportStateVerify attributes not equivalent. Difference is shown below. The - symbol indicates attributes missing after import.
        
          map[string]string{
        - 	"ssl_settings.0.ssl_management_type": "AUTOMATIC",
        + 	"ssl_settings.0.ssl_management_type": "",
          }
```

This is expected because Get against the Beta API doesn't return enough information to know the type.

For TestAccAppEngineDomainMapping_update, we'll want to update the `ImportStateVerifyIgnore` entries like https://github.com/GoogleCloudPlatform/magic-modules/blob/552d835ed5ba412770659dc926e07c5625ca1506/mmv1/third_party/terraform/services/appengine/resource_app_engine_domain_mapping_test.go#L28 to also include `ssl_settings.0.ssl_management_type`

mmv1/products/appengine/DomainMapping.yaml
gwendal-lecren: Done

mmv1/third_party/terraform/services/appengine/resource_app_engine_flexible_app_version_test.go
gwendal-lecren: Pfiou I missed that. Done

gwendal-lecren:
Hello @rileykarson, hope you're doing well - Is there anything missing to pass the test `TestAccAppEngineFlexibleAppVersion_update`? If you can provide some logs that would be nice. Thanks!

mmv1/third_party/terraform/services/appengine/resource_app_engine_flexible_app_version_test.go
```diff
@@ -166,20 +189,42 @@ resource "google_project" "my_project" {
...
+resource "google_compute_subnetwork" "subnetwork" {
+  project                  = google_project_service.compute.project
+  name                     = "custom"
+  region                   = "us-central1"
+  network                  = google_compute_network.network.id
+  ip_cidr_range            = "10.0.0.0/16"
+  private_ip_google_access = true
+}
+
+resource "google_app_engine_application" "app" {
```
rileykarson, line 221:
```
        Error: Error when reading or editing Subnetwork: googleapi: Error 400: The subnetwork resource 'projects/tf-test-appeng-flexnbak3l06x8/regions/us-central1/subnetworks/custom' is already being used by 'projects/tf-test-appeng-flexnbak3l06x8/zones/us-central1-c/instances/aef-default-v1-801x', resourceInUseByAnotherResource
```

On L291 there's `noop_on_destroy = true`, which leaves instances behind, preventing cleaning up the subnetwork. Let's change that to `delete_service_on_destroy = true` (https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/app_engine_flexible_app_version#delete_service_on_destroy)

That may fail with an error because we're in the default service. If it does, we probably just have to use a non-default on on L235 (I forget if there are more steps than that, though, so we probably just want to see if default causes issues first).

rileykarson (commented):
Apologies, thought I'd shared the error! Was underwater on some other work for a bit- suggestion inline.

mmv1/third_party/terraform/services/appengine/resource_app_engine_flexible_app_version_test.go
gwendal-lecren: Just updated! Yeah we might end up into an error since it's the default service actually. Let's see!

rileykarson: L32/L41 we'll need to change `noop_on_destroy` to `delete_service_on_destroy` in ImportStateVerifyIgnore. Out of range for me to apply a suggestion or else I'd do it myself!

gwendal-lecren: Done as well

mmv1/third_party/terraform/services/appengine/resource_app_engine_flexible_app_version_test.go
```diff
@@ -29,7 +29,7 @@ func TestAccAppEngineFlexibleAppVersion_update(t *testing.T) {
 				ResourceName:            "google_app_engine_flexible_app_version.foo",
 				ImportState:             true,
 				ImportStateVerify:       true,
-				ImportStateVerifyIgnore: []string{"env_variables", "deployment", "entrypoint", "service", "noop_on_destroy"},
+				ImportStateVerifyIgnore: []string{"env_variables", "deployment", "entrypoint", "service", "delete_service_on_destroy"},
```
rileykarson, line 32:
```suggestion
				ImportStateVerifyIgnore: []string{"env_variables", "deployment", "entrypoint", "service", "noop_on_destroy"},
```

Ah- we didn't change step 1, my bad.

rileykarson (commented):
OK- seems we can't delete the default. That'll require editing L99 and L235, we can pick an arbitrary service name.

> Error: Error when reading or editing Service: googleapi: Error 400: The default service (module) cannot be deleted.

gwendal-lecren:
> OK- seems we can't delete the default. That'll require editing L99 and L235, we can pick an arbitrary service name.
> 
> > Error: Error when reading or editing Service: googleapi: Error 400: The default service (module) cannot be deleted.

Thanks, done - Though I believe appengine requires a `default` service (before being able to add additional ones), shall we also add a dumb appengine standard `default` service?

rileykarson:
Ah, yep, we may need to do that!

rileykarson:
It's been a bit since I used App Engine- if we omit both `_on_destroy` options will we succesfully remove all versions from the service, but preserve the service? We may be able to use the default service that way and avoid some complexity- we need there to be zero instances on the subnetwork so we can delete it, but we're cleaning up the project anyways so we can leave a dangling default service.

rileykarson:
As you mentioned, we figured this would fail. Error for completeness:

```
        Error: Error creating FlexibleAppVersion: googleapi: Error 400: The first service (module) you upload to a new application must be the 'default' service (module). Please upload a version of the 'default' service (module) before uploading a version for the 'custom' service (module). See the documentation for more information. Python: (https://developers.google.com/appengine/docs/python/modules/#Python_Uploading%%20modules) Java: (https://developers.google.com/appengine/docs/java/modules/#Java_Uploading%%20modules)
```

gwendal-lecren:
> As you mentioned, we figured this would fail. Error for completeness:
> 
> ```
>         Error: Error creating FlexibleAppVersion: googleapi: Error 400: The first service (module) you upload to a new application must be the 'default' service (module). Please upload a version of the 'default' service (module) before uploading a version for the 'custom' service (module). See the documentation for more information. Python: (https://developers.google.com/appengine/docs/python/modules/#Python_Uploading%%20modules) Java: (https://developers.google.com/appengine/docs/java/modules/#Java_Uploading%%20modules)
> ```

Added a default appengine standard service based on https://github.com/GoogleCloudPlatform/magic-modules/blob/8937049ba82a104d32a13507980a66d50cab2ee4/mmv1/third_party/terraform/services/appengine/resource_app_engine_standard_app_version_test.go

rileykarson (approved):
LGTM

Thanks for sticking through this- using the beta API / getting the test to pass took a surprising amount of effort for what should have been a very quick change, if not for the various complications! 😅

gwendal-lecren:
> LGTM
> 
> Thanks for sticking through this- using the beta API / getting the test to pass took a surprising amount of effort for what should have been a very quick change, if not for the various complications! 😅

Thanks @rileykarson for all the guidance!

PR #10318, opened 2024-03-28
Remove ssh-keyscan usage

melinath:
This seems to not be necessary, since we don't use SSH to authenticate with GitHub.

I've built the image using this version of the dockerfile and am running generation / testing with it on this PR.

This PR only changes the image used by PR builds; if this rolls out without any issues I'll make a similar PR for the post-merge image.

**Release Note Template for Downstream PRs (will be copied)**

```release-note:none

```

PR #10328, opened 2024-03-29
Removed keyscan from build-environment

melinath:
This shouldn't be necessary since we don't use ssh to interact with GitHub. The build-environment image is used in PR builds so it's possible to test before merge.

Follow-up to https://github.com/GoogleCloudPlatform/magic-modules/pull/10318

```release-note:none

```

### DIFF
--- a/.ci/containers/build-environment/Dockerfile
+++ b/.ci/containers/build-environment/Dockerfile
@@ -36,10 +36,6 @@ RUN apt-get update && apt-get install -y --no-install-recommends git openssh-cli
 RUN git config --global user.name "Modular Magician"
 RUN git config --global user.email "magic-modules@google.com"
 
-# Set up Github SSH cloning.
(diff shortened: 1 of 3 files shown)


melinath:
Looks like downstream builds are unaffected even while using the testing image:
- building the magician binary: https://pantheon.corp.google.com/cloud-build/builds/4175ae76-b7b8-4721-87c8-f787e91992ca;step=8;tab=detail?e=-13802955&mods=logs_tg_staging&project=graphite-docker-images
- tpgb-head generation: https://pantheon.corp.google.com/cloud-build/builds/4175ae76-b7b8-4721-87c8-f787e91992ca;step=11;tab=detail?e=-13802955&mods=logs_tg_staging&project=graphite-docker-images

roaks3 (approved):

PR #10329, opened 2024-03-29
Updated instructions for how to build and test docker images

melinath:
Documented the process we actually use for testing images so that we don't have to reconstruct it every time. This is the process I used for https://github.com/GoogleCloudPlatform/magic-modules/pull/10328.

```release-note:none

```


shuyama1 (approved):
LGTM 💯 Thanks!

PR #10330, opened 2024-03-29
Ignore read on FlexibleAppVersion deployment

slevenick:
Possibly Fixes: https://github.com/hashicorp/terraform-provider-google/issues/17203

I am having trouble running tests due to billing setup, but this should prevent the panic at least

```release-note:bug
appengine: fixed a crash in `google_app_engine_flexible_app_version` due to the `deployment` field not being returned by the API
```


rileykarson (approved):
Did you find out if `deployment` is _never_ returned? The parent comment seems to have drifted from the PR's contents. LGTM if so.

slevenick:
> Did you find out if `deployment` is _never_ returned? The parent comment seems to have drifted from the PR's contents. LGTM if so.

It doesn't get returned for any config that we have in our tests. The `TestAccAppEngineFlexibleAppVersion_appEngineFlexibleAppVersionExample` test has been failing in CI due to the same panic though it sets the `zip` field in deployments where the `TestAccAppEngineFlexibleAppVersion_update` sets `files`. We don't have a test that sets `containers`, so I'm not sure about that one

rileykarson:
Seems reasonable enough. If you could use API explorer or console to test container that would be nice, but not necessarily required.

PR #10337, opened 2024-04-01
Set Validation field of Go Type object

zli82016:
Set Validation field of Go Type object

```release-note:none

```


NickElliot (approved):

NickElliot:
That's probably what confused me 😂

PR #10347, opened 2024-04-02
Run version guard checks on PRs

roaks3:
This adds the GitHub Action that checks version guards in erb files that were changed.

Confirmation of an invalid version guard being found in a previous commit here https://github.com/GoogleCloudPlatform/magic-modules/actions/runs/8525807220/job/23353642608?pr=10347

This PR also includes changes to all existing erbs so that they pass the checker (to avoid people touching files and then needing to fix them). You can see from the generated diffs that there are only whitespace differences.

```release-note:none

```

### DIFF
--- a/.github/workflows/mmv1-check-templates.yml
+++ b/.github/workflows/mmv1-check-templates.yml
@@ -12,7 +12,7 @@ jobs:
     runs-on: ubuntu-22.04
     steps:
       - name: Checkout Repository
-        uses: actions/checkout@v4
+        uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.2
         with:
           path: repo
           fetch-depth: 0
--- a/.github/workflows/mmv1-check-templates.yml
+++ b/.github/workflows/mmv1-check-templates.yml
@@ -1,0 +1,30 @@
+name: mmv1-check-templates
+
+permissions: read-all
+
+on:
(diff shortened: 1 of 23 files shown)


.github/workflows/mmv1-check-templates.yml
```diff
@@ -0,0 +1,30 @@
...
+  pull_request:
+    paths:
+      - 'mmv1/**/*.erb'
+
+jobs:
+  version-guard-check:
+    runs-on: ubuntu-22.04
+    steps:
+      - name: Checkout Repository
+        uses: actions/checkout@v4
```
shuyama1, line 15: I think we'll want to pin the version by a specific hash. Looks like that's how we set up for other workflows after https://github.com/GoogleCloudPlatform/magic-modules/pull/10239

shuyama1 (commented):
LGTM - only one small comment

.github/workflows/mmv1-check-templates.yml
roaks3: Nice catch! My copy-paste missed that change, should be fixed now.

shuyama1:
Hey @roaks3, looks like you'll need to rebase the PR due to the CI issue. Sorry about it!

roaks3:
No worries!

shuyama1 (approved):

PR #10352, opened 2024-04-02
docs: Document that the max allowed retention for Firestore backup schedules is 14 weeks

rwhogg:

```release-note:none

```


ZixuanLiuCanada (approved):

hao-nan-li (commented):
Could you do a rebase on main? I think this error happened before and the fix is included not long ago.

hao-nan-li (approved):

PR #10364, opened 2024-04-03
Support AlloyDB Maintenance Window flags

sony-c:
**Description**

This PR adds customized instance maintenance rollout support to AlloyDB.  
AlloyDB regularly rollout changes to the fleet instances to have bug fixed and to enhance or add new features. Maintenance window is a new feature added to AlloyDB to enhance this capability, the feature provides an option to users to specify the day and start time when they want to start the rollout instead of default schedule set by AlloyDB.  

Issue: 
- https://github.com/hashicorp/terraform-provider-google/issues/17757
 

```release-note:enhancement
alloydb: add `maintenance_update_policy`, `maintenance_update_policy.maintenance_windows.day` and `maintenance_update_policy.maintenance_windows.start_time` fields to `google_alloydb_cluster` resource.
```


sony-c:
Automated test result: 
```
=== RUN   TestAccAlloydbCluster_withMaintenanceWindows
=== PAUSE TestAccAlloydbCluster_withMaintenanceWindows
=== RUN   TestAccAlloydbCluster_withMaintenanceWindowsMissingFields
=== PAUSE TestAccAlloydbCluster_withMaintenanceWindowsMissingFields
=== CONT  TestAccAlloydbCluster_withMaintenanceWindows
=== CONT  TestAccAlloydbCluster_withMaintenanceWindowsMissingFields
--- PASS: TestAccAlloydbCluster_withMaintenanceWindowsMissingFields (145.94s)
--- PASS: TestAccAlloydbCluster_withMaintenanceWindows (176.38s)
PASS
ok      github.com/hashicorp/terraform-provider-google-beta/google-beta/services/alloydb        178.996s
```
Manual test result: 
```
google_compute_network.default: Creating...
google_compute_network.default: Still creating... [10s elapsed]
google_compute_network.default: Still creating... [20s elapsed]
google_compute_network.default: Still creating... [30s elapsed]
google_compute_network.default: Still creating... [40s elapsed]
google_compute_network.default: Still creating... [50s elapsed]
google_compute_network.default: Creation complete after 53s [id=projects/sonychiang-playground-stg/global/networks/alloydb-cluster]
google_alloydb_cluster.default: Creating...
google_alloydb_cluster.default: Still creating... [10s elapsed]
google_alloydb_cluster.default: Creation complete after 12s [id=projects/sonychiang-playground-stg/locations/us-central1/clusters/alloydb-cluster]
╷
│ Warning: Argument is deprecated
│
│   with google_alloydb_cluster.default,
│   on main.tf line 5, in resource "google_alloydb_cluster" "default":
│    5:   network    = google_compute_network.default.id
│
│ `network` is deprecated and will be removed in a future major release. Instead, use `network_config` to define the network configuration.
╵

Apply complete! Resources: 2 added, 0 changed, 0 destroyed.
```

mmv1/products/alloydb/Cluster.yaml
```diff
@@ -497,6 +497,55 @@ properties:
...
+                - :SUNDAY
+            - !ruby/object:Api::Type::NestedObject
+              name: 'startTime'
+              custom_expand: 'templates/terraform/custom_expand/alloydb_cluster_input_maintenance_window_start_time_expand.go.erb'
+              custom_flatten: 'templates/terraform/custom_flatten/alloydb_cluster_input_maintenance_window_start_time_flatten.go.erb'
+              description: |
+                Preferred time to start the maintenance operation on the specified day. Maintenance will start within 1 hour of this time.
+              properties:
+                - !ruby/object:Api::Type::Integer
+                  name: hours
```
slevenick, line 534: Should these properties be marked as required? Does it make sense to specify hours but not minutes, etc?

mmv1/templates/terraform/custom_expand/alloydb_cluster_input_maintenance_window_start_time_expand.go.erb
```diff
@@ -0,0 +1,130 @@
...
+	}
+	transformed["nanos"] = val
+	return transformed, nil 
+}
+
+func expandAlloydbClusterMaintenanceWindowStartTimeHours(v interface{}, d tpgresource.TerraformResourceData, config *transport_tpg.Config) (interface{}, error) {
+    // Handles the string fixed64 format
+	if strVal, ok := v.(string); ok {
+		if intVal, err := tpgresource.StringToFixed64(strVal); err == nil {
+            if intVal < 0 || intVal > 23 {
```
slevenick, line 50: We don't need to validate at this level. API restrictions on field values should throw a 400 on a request with an invalid value

mmv1/templates/terraform/custom_expand/alloydb_cluster_input_maintenance_window_start_time_expand.go.erb
```diff
@@ -0,0 +1,130 @@
...
+			return nil, fmt.Errorf("Error configuring hours: %d. Value has to be from 0 to 23", intVal)
+		}
+		return intVal, nil
+	}
+
+	return v, nil // let terraform core handle it otherwise
+}
+
+func expandAlloydbClusterMaintenanceWindowStartTimeMinutes(v interface{}, d tpgresource.TerraformResourceData, config *transport_tpg.Config) (interface{}, error) {
+	// Handles the string fixed64 format
```
slevenick, line 70: Why do we need alternative ways of handling the value based on if it's a string or float?

mmv1/templates/terraform/custom_flatten/alloydb_cluster_input_maintenance_window_start_time_flatten.go.erb
```diff
@@ -0,0 +1,95 @@
...
+        "hours": flattenAlloydbClusterMaintenanceWindowStartTimeHours(original["hours"], d, config),
+        "minutes": flattenAlloydbClusterMaintenanceWindowStartTimeMinutes(original["minutes"], d, config),
+        "seconds": flattenAlloydbClusterMaintenanceWindowStartTimeSeconds(original["seconds"], d, config),
+        "nanos": flattenAlloydbClusterMaintenanceWindowStartTimeNanos(original["nanos"], d, config),
+    }
+	return []interface{}{transformed}
+}
+
+func flattenAlloydbClusterMaintenanceWindowStartTimeHours(v interface{}, d *schema.ResourceData, config *transport_tpg.Config) interface{} {
+    // Handles the string fixed64 format
```
slevenick, line 30: Is the response from the API different for an int vs a float? Wouldn't parsing as a float always work?

mmv1/templates/terraform/pre_create/alloydb_cluster.go.erb
```diff
@@ -63,3 +62,20 @@ if secondaryConfig != nil {
...
+if val, ok := obj["maintenanceUpdatePolicy"]; ok {
+	maintenanceUpdatePolicy = val.(map[string]interface{})
+
+	// Validate maintenance window variables
+	if maintenanceUpdatePolicy["maintenanceWindows"] != nil {
+		mws := maintenanceUpdatePolicy["maintenanceWindows"].([]interface{})
+		for _, mw := range mws {
+			maintenanceWindow := mw.(map[string]interface{})
+			if maintenanceWindow["day"] == nil || maintenanceWindow["startTime"] == nil {
+				return fmt.Errorf("Error creating cluster. Both \"day\" and \"startTime\" are required if attempting to configure maintenance windows, otherwiese remove the maintenance_windows.")
```
slevenick, line 77: This can be represented by `required: true` in the YAML file for these fields. Then this can be removed

slevenick (changes requested):
I'm not sure why we need the custom flatten/expands, I don't think the conversion between types is necessary. Is there a reason it is?

mmv1/templates/terraform/custom_expand/alloydb_cluster_input_maintenance_window_start_time_expand.go.erb
sony-c: What do you mean by API restrictions? Are you saying the yaml setting or we should throw this from our service?

mmv1/templates/terraform/custom_expand/alloydb_cluster_input_maintenance_window_start_time_expand.go.erb
sony-c: I am not familiar with Terraform changes yet, I saw other function dealing in this way for similar field type and I supposed the value could be provided as string or float. 
Are you suggesting to remove the custom parsing ?

mmv1/templates/terraform/custom_flatten/alloydb_cluster_input_maintenance_window_start_time_flatten.go.erb
sony-c: From our API we should return int, float should always work.
I think you are saying the handling for string or float is not required here, right?

mmv1/templates/terraform/pre_create/alloydb_cluster.go.erb
sony-c: so I just set the required at day and start_time parameter? and this will not make the entire parent nested object required?

mmv1/templates/terraform/custom_flatten/alloydb_cluster_input_maintenance_window_start_time_flatten.go.erb
slevenick: Yeah, if it's always an int then our typical generated flatten methods should work. You should be able to remove the custom flatten entirely

mmv1/templates/terraform/custom_expand/alloydb_cluster_input_maintenance_window_start_time_expand.go.erb
slevenick: Yes, I would remove custom parsing unless you run into problems without it. If the field functions normally on the API side I would be surprised if you need any custom parsing

mmv1/templates/terraform/custom_expand/alloydb_cluster_input_maintenance_window_start_time_expand.go.erb
slevenick: I believe you should already throw validation errors from the service if the value sent is out of the expected range. Is that correct?

mmv1/templates/terraform/pre_create/alloydb_cluster.go.erb
slevenick: required only applies within the sub-object, so it means if the parent object is specified then `day` and `start_time` are required.

Marking the parent object as required is the only way to require that it exists at all

mmv1/templates/terraform/custom_expand/alloydb_cluster_input_maintenance_window_start_time_expand.go.erb
sony-c: Yes we do, adding validation here so that the request can be failed earlier without sending the API request. If this is not the recommended pattern, I will remove it.

mmv1/products/alloydb/Cluster.yaml
slevenick: No, it will not make the entire maintenance_update_policy or maintenance_windows required.

You can leave minutes or seconds not required

sony-c:
Thank you @slevenick for your comments, I removed all custom code and the tests succeeded as well.

sony-c:
Thank you Sam for the review. 
The feature is now in alpha/beta and not exposed to public now, I allowlisted the presubmit project and the change should take effect by the end of today(Apr. 5th) so that we can unblock the tests running in this PR. 
We will merge this PR only after we GA the feature (ETA 4/22).

slevenick:
Hm, my re-trigger command isn't working. Can you push a new commit to trigger a rebuild if it doesn't start soon?

mmv1/third_party/terraform/services/alloydb/resource_alloydb_cluster_test.go
```diff
@@ -1167,3 +1168,136 @@ resource "google_compute_global_address" "private_ip_alloc" {
...
+				ImportStateVerifyIgnore: []string{"initial_user", "cluster_id", "location"},
+			},
+		},
+	})
+}
+
+func testAccAlloydbCluster_withMaintenanceWindows(context map[string]interface{}) string {
+	return acctest.Nprintf(`
+resource "google_alloydb_cluster" "default" {
+  provider   = google-beta
```
slevenick, line 1200: Is the maintenance windows field beta-only? If so, it should be marked min_version: beta in the YAML file. Otherwise, remove this provider line

ScottSuarez: Note from office hours.
On launch this will be available both the ga and beta apis.

slevenick: Should it be available in the GA provider then? Or will it still be considered "beta"?

ScottSuarez: Correct. It should be in both providers. Not considered beta.

PR #10365, opened 2024-04-04
Set no prefix to fix "IP range no free block error" for test of looker enterprise

efeelaiho:
Set no prefix to fix IP range no free block error

Fixes https://github.com/hashicorp/terraform-provider-google/issues/17394

Error to resolve:

```
Error: Error waiting to create Instance: Error waiting for Creating Instance: Error code 9, message: 

Couldn't find free blocks in requested range '10.209.49.0/25'. 
Please allocate new ranges for this service provider.
```

```release-note:none

```


mmv1/products/looker/Instance.yaml
```diff
@@ -69,9 +69,9 @@ examples:
       client_id: 'my-client-id'
       client_secret: 'my-client-secret'
     test_vars_overrides:
-      address_name: 'acctest.BootstrapSharedTestGlobalAddress(t, "looker-vpc-network-1", acctest.AddressWithPrefixLength(20))'
+      address_name: 'acctest.BootstrapSharedTestGlobalAddress(t, "looker-vpc-network-1")'
```
shuyama1, line 72: We may need to change the name `looker-vpc-network-1` to something else to test if this works, as based on how we set up the bootstrapped resource, they won't get recreated without the name changed (we'll read the resource first and only create a new one if no existing one is found)

efeelaiho: @shuyama1 Okay, done!

shuyama1 (approved):
LGTM! Thanks!

PR #10367, opened 2024-04-04
Add DeletionProtectionEnabled to Redis Cluster

Harshita1603:
Adding deletion protection enabled field in the Cluster resource in redis.

1. Searched through the issue tracker for an open issue that this either resolves or contributes to, commented on it to claim it, and written "fixes {url}" or "part of {url}" in this PR description. If there were no relevant open issues, I opened one and commented that I would like to work on it - https://github.com/hashicorp/terraform-provider-google/issues/17763
2. Ensured that all new fields I added that can be set by a user appear in at least one example (for generated resources) or third_party test (for handwritten resources or update tests). 
3. Generated Terraform providers, and ran make test and make lint in the generated providers to ensure it passes unit and linter tests. 
4. Ran relevant acceptance tests using my own Google Cloud project and credentials. 
They show the following error -    
testing_new.go:85: Error running post-test destroy, there may be dangling resources: exit status 1 Error: Error waiting for Deleting Network: The network resource 'projects/harshig-test/global/networks/mynetworkpxlkaasq97' is already being used by 'projects/harshig-test/global/firewalls/mynetworkpxlkaasq97-vmvqhndc7vcvhkaszq37j4sh-4'4. 

5. Read/Write release notes before writing my release note below.

```release-note:enhancement
redis: added a `deletion_protection_enabled` field with a default value of `true` to the `google_redis_cluster` resource
```


NickElliot (commented):
It looks like whats happening is the API does not return the "deletion_protection_enabled" value in the "GET" request, so the plan assumes the value to be false and has a resulting permadiff. Is this intended API behavior? If so, an "ignore_read" will need to be set for the new field

Harshita1603:
Hey Nick, 
Could you please run the Google Cloud Build again?

Harshita1603:
The tests are now passing

Harshita1603:
Hey @NickElliot 
Could you please review the changes when you have time?

mmv1/products/redis/Cluster.yaml
```diff
@@ -231,6 +231,14 @@ properties:
...
       Required. Number of shards for the Redis cluster.
     required: true
+  - !ruby/object:Api::Type::Boolean
+    name: deletionProtectionEnabled
+    description: |
+      Optional. Indicates if the cluster is deletion protected or not.
+      If the value if set to true, any delete cluster operation will fail.
+      Default value is true.
+    required: false
+    default_value: true
```
NickElliot, line 241: I think this might be a better fit for `default_from_api` if this is also a server-side default -- if this config was manually overridden for a pre-existing resource by a user via cloud console or pantheon, adding a field with a default value could potentially introduce a breaking change for users' configs (i.e. it potentially planning to do `false -> true` for users who don't manually add `deletion_protection_enabled = false` to their config)

NickElliot (commented):
Let me know if you agree/disagree here!

mmv1/products/redis/Cluster.yaml
Harshita1603: Hey @NickElliot , it is not the default value. The server-side default is false. It will also not be backward compatible for existing terraform plans. Is my understanding correct?

Harshita1603:
> @Harshita1603, this PR is waiting for action from you. Please address any comments or change requests, or [re-request review](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/requesting-a-pull-request-review) from a core reviewer if no action is required.
> 
> 
> If no action is taken, this PR will be closed in 14 days.
> 

I need to work on the suggestions by the reviewer and will take it up after my vacation.

NickElliot (commented):
Hi, we're discussing on what the approach will be for adding these deletion protection fields will be in terms of a wider policy, and will hopefully comment soon with a decided approach

Harshita1603:
> Hi, we're discussing on what the approach will be for adding these deletion protection fields will be in terms of a wider policy, and will hopefully comment soon with a decided approach

Thanks @NickElliot.
1. Do we plan to add a new deletion protection field? If so, will these be specific to terraform?
2. When can we expect the change?

NickElliot (changes requested):
followed up in chat -- if the terraform-specific default is definitely desired, please update the pull request to be merging into the `FEATURE-BRANCH-major-release-6.0.0`

NickElliot:
details for breaking change:

this change does not qualify for a state_upgrader or other virtual field default read value ala the other deletion policy fields, as it is using an official `deletion_protection_enabled` field within the API using a Terraform-end opt-out default (rather than the server-side default of false), and as such this will produce a real, needs be visible, diff in user states

NickElliot (approved):
rebased onto 6.0.0 branch and merged in latest changes, resolved test failures resulting from the changes (which affected the non standard test template)

LGTM

mmv1/third_party/terraform/services/redis/resource_redis_cluster_test.go.erb
```diff
@@ -255,7 +285,8 @@ resource "google_redis_cluster" "test" {
 	replica_count = %d
 	shard_count = %d
 	node_type = "%s"
-	region         = "us-central1"
+	deletion_protection_enabled = %v
+  region         = "us-central1"
```
NickElliot, line 289:
```suggestion
	region         = "us-central1"
```

PR #9742, opened 2024-01-03
Add `google_dns_managed_zones` data source and acceptance test

SarahFrench:
**Release Note Template for Downstream PRs (will be copied)**

```release-note: new-datasource
`google_dns_managed_zones`
```

PR #10368, opened 2024-04-04
Migrate `dns` data sources back to the SDK to solve auth issue resulting from migration to plugin framework

SarahFrench:
# Description

Hopefully will fix https://github.com/hashicorp/terraform-provider-google/issues/16832

## What this PR does

This PR returns these data sources back to being implemented with the SDK:
- google_dns_keys
- google_dns_managed_zone
- google_record_set

The PR also migrates google_dns_managed_zone**s** to the SDK. It was originally implemented in the plugin framework.

When resources were migrated to the PF from SDK there were some small schema changes, namely adding `id` fields to schemas. These changes have been implemented in the new SDK code for the data sources, even though the original SDK code didn't contain them.

## Why this PR is happening

tl;dr of the timeline leading to this PR:
- Early 2023
    - The provider was muxed and resulted in a number of customer issues, which were not caught by acceptance tests.
    - Acceptance tests for identified issues were subsequently added
    - Important detail is that the SDK and PF sides of the provider contain parallel implementations of provider configuration logic 
- Spring 2023
    - It's identified that migrated data sources [fail acceptance tests in VCR](https://github.com/hashicorp/terraform-provider-google/issues/14158)
   - Exact details _why_ aren't relevant to this PR
- Jan 2024
    - I have a customer issue with a deadline to implement a plural version of the google_dns_managed_zone data source. I implement it in plugin-framework with the rationale that I'm about to start digging into the previous migration work and working through issues such as VCR test problem, so adding one more problematic data source is OK (spoiler: I was wrong).
    - I'm alerted to [an issue opened in Dec 2023](https://github.com/hashicorp/terraform-provider-google/issues/16832) that indicates there are **auth issues specific to the dns data sources** and not the dns resources, i.e specific to the plugin-framework migrated parts of that service.

I believe the issue [data.google_dns_managed_zone always fails with "Post "https://oauth2.googleapis.com/token": context canceled"](https://github.com/hashicorp/terraform-provider-google/issues/16832) shows there are problems with the provider configuration code implemented in the plugin-framework side of the provider. All our acceptance tests auth the provider in one way, so this bug wasn't identified until the first issue was opened.

Given that tweaking how the plugin-framework provider is authenticated is a non-trivial task, and that I will also need to make large changes to the central provider configuration code to address [the VCR issue](https://github.com/hashicorp/terraform-provider-google/issues/14158), I decided that the best way to resolve the problem for users is to revert the changes to the dns data sources and for me to do this work 'in my own time'. 

## Testing 

I've [manually tested](https://github.com/GoogleCloudPlatform/magic-modules/pull/10368#issuecomment-2039939863) and also have [some acc tests](https://github.com/GoogleCloudPlatform/magic-modules/pull/10368#issuecomment-2050186172) but they're highly dependent on the machine running them.

## Next steps

In future I'll fully investigate the auth issues in the plugin-framework implementation of the provider configuration logic. I suspect I'll need to make wider changes in that code to avoid having 2 parallel implementations, so I chose to do that work separately to addressing the issue linked to this this PR.

## Relevant issues and PRs
* https://github.com/hashicorp/terraform-provider-google/issues/16832
* https://github.com/hashicorp/terraform-provider-google/issues/14158
* https://github.com/GoogleCloudPlatform/magic-modules/pull/9742

---

```release-note:bug
dns: fixed bug where some methods of authentication didn't work when using `dns` data sources
```


SarahFrench:
Some tests failing due to checks, others because we need to demonstrate ownership of a domain. Will look into this tomorrow in work hours

SarahFrench:
TestAccDataSourceDNSKeys_basic

```
Error: Error creating ManagedZone: googleapi: Error 400: Please verify ownership of the 'dnssec.tf-test.club.' domain (or a parent) at http://www.google.com/webmasters/verification/ and try again, verifyManagedZoneDnsNameOwnership
```

I'll try address this with a different domain, and if not then I think I'll need to do some stuff internally

SarahFrench:
The current GitHub incident appears to have affected the automation, rerunning

/gcbrun

SarahFrench:
Manual tests

```tf
terraform {
  required_providers {
    google = {
      source = "hashicorp/google"
      # version = "~> 4.52.0" // pre-migration => can plan and apply
      # version = "~> 4.53.1" // muxed provider =>  can plan and apply
      # version = "~> 4.59.0" // succeeds
      version = "~> 4.60.0" // fails
      # version = "~> 4.65.0" // fails
      # version = "~> 4.84.0" // last 4.x.x provider =>  fails
      # version = "~> 5.9.0" // migrated provider => fails

      # using dev override to test revert back to SDK => SUCCESS
    }
  }
}


// Set up ADCs
//    gcloud auth application-default login
//
// Test provider with ADCs passed as credentials using:
//    terraform plan -var "credentials=$(cat  ~/.config/gcloud/application_default_credentials.json)"
//
// When changing version don't forget to use: terraform init -upgrade

provider "google" {
  credentials = var.credentials
}
variable "credentials" {
  type = string
}

resource "google_dns_managed_zone" "foo" {
  name     = "foobar"
  dns_name = "sarahtest.hashicorptest.com."

  dnssec_config {
    state         = "on"
    non_existence = "nsec3"
  }
}

data "google_dns_keys" "foo_dns_keys" {
  managed_zone = google_dns_managed_zone.foo.id
}

output "foo_dns_ds_record" {
  description = "DS record of the foo subdomain."
  value       = data.google_dns_keys.foo_dns_keys.key_signing_keys[0].ds_record
}

```

SarahFrench:
To look into - 4.60 is where client datasources added

SarahFrench:
>To look into - 4.60 is where client datasources added

**tl;dr this was a red herring and I've re-confirmed that the problem is the data source being in PF.**

Context about my comment quoted above : I see the problem in the `dns` data sources is present in 4.60.0 but not in 4.59.0

I wondered if the two datasources google_client_config & google_client_openid_userinfo ([migrated to PF in this commit](https://github.com/hashicorp/terraform-provider-google/commit/0601728cfe7acc3e440069a655d264d31036caa4)) were the actual source of the problem, as these were first released in v4.60.0. I investigated this by creating a version of the provider locally that contains the SDK versions of those two data sources but still has the PF versions of the `dns` data sources. I saw that the provider still failed to mange `dns` data sources when authenticated using ADCs, so google_client_* data sources aren't a factor in this problem.

BUT, I then realised that I was confused about which version(s) of the provider contain PF-related changes.

It looks like the `dns` data sources were implemented with PF in v4.53.0 when the provider was also muxed. The patch v4.53.1 to fix issues with the provider config reverted the `dns` data sources back to being implemented with SDK. They remain implemented in the SDK up until v4.60.0 when they're returned to being implemented with the PF.

When I try to reproduce the bug this PR is addressing by using v4.53.0, the first time the dns data sources were implemented with PF, I cannot reproduce the error due to another error [that was subsequently fixed](https://github.com/hashicorp/terraform-provider-google/issues/14298#issuecomment-1511326121).

As a result you only see the bug in v4.60.0+

SarahFrench:
It's not currently possible to create acceptance tests that use ADCs reliably. I've pushed 72063ef that includes a test that works if I supply GOOGLE_APPLICATION_CREDENTIALS and it points to the ~/.config/gcloud/application_default_credentials.json file made by `gcloud auth application-default login`.

I assume it'll fail, and then I could comment the test out... which feels gross but not sure what the other option is in the short term.

SarahFrench:
I skipped the new tests in VCR but running it in TeamCity to conclude that it'll fail due to lack of creds : https://hashicorp.teamcity.com/buildConfiguration/TerraformProviders_GoogleCloud_GOOGLE_MMUPSTREAMTESTS_GOOGLE_PACKAGE_DNS/131946?buildTab=overview&hideTestsFromDependencies=false&hideProblemsFromDependencies=false

SarahFrench:
Unit tests failing due to them running all tests, including TestAcc tests but they pass and skip due to TF_ACC not being set. I need to update my test to not trigger errors if TF_ACC isn't set, as normal. Currently I see:

```
    data_source_dns_key_test.go:106: GOOGLE_APPLICATION_CREDENTIALS must be set for acceptance tests that are dependent on ADCs

```

SarahFrench:
I've also confirmed that the issue in https://github.com/hashicorp/terraform-provider-google/issues/16832 also affects people using dynamic credentials in Terraform Cloud.

SarahFrench:
The new acceptance test `TestAccDataSourceDNSKeys_basic_AdcAuth` is skipped in VCR [but will fail in TeamCity](https://hashicorp.teamcity.com/buildConfiguration/TerraformProviders_GoogleCloud_GOOGLE_MMUPSTREAMTESTS_GOOGLE_PACKAGE_DNS/131946?buildTab=overview&hideTestsFromDependencies=false&hideProblemsFromDependencies=false&expandBuildDeploymentsSection=false&expandBuildTestsSection=true&expandBuildChangesSection=true). Due to this I'll comment out the test until running that test in automation is possible, which is beyond the scope of this PR.

mmv1/third_party/terraform/services/dns/data_source_dns_key_test.go.erb
```diff
@@ -3,91 +3,52 @@ package dns_test
...
 }
 
+
 func TestAccDataSourceDNSKeys_noDnsSec(t *testing.T) {
-	// TODO: https://github.com/hashicorp/terraform-provider-google/issues/14158
-	acctest.SkipIfVcr(t)
 	t.Parallel()
 
-	dnsZoneName := fmt.Sprintf("tf-test-dnskey-test-%s", acctest.RandString(t, 10))
+	dnsZoneName := fmt.Sprintf("data-dnskey-test-%s", acctest.RandString(t, 10))
```
rileykarson, line 43: Why do we change the names of things in a few places? Were we re-applying old code?

rileykarson (commented):
Unit test errors:

> Error: google-beta/services/dns/data_source_dns_key_test.go:7:2: "os" imported and not used
> Error: google-beta/services/dns/data_source_dns_key_test.go:8:2: "regexp" imported and not used

I've skimmed the resource code, but spent most of my effort on the test code, given the resource code was likely just copy-pasted or reverted using Git. No issues w/ either. Let me know if I should take a deeper look.

mmv1/third_party/terraform/services/dns/data_source_dns_key_test.go.erb
SarahFrench: Ah I think I initially changed some names to start with tf-test- to be more sweepable but then I was chasing a bunch of changes to those files that occurred in between the addition and reverting of the muxing work (e.g [this change](https://github.com/GoogleCloudPlatform/magic-modules/commit/260964926d27bfbad2dd964d937698262e8064ea)).

I'll look for other instances where tf-test- can be added

SarahFrench:
>I've skimmed the resource code, but spent most of my effort on the test code, given the resource code was likely just copy-pasted or reverted using Git. No issues w/ either. Let me know if I should take a deeper look.

I agree about focusing on tests - they were copied in from previous file versions as the plugin-framework's footprint on test files was pretty large and it was easiest to return to a prior version of the files, just like the data sources' files themselves.

The only thing that might need a deeper look is the google_dns_managed_zones data source, as that [was first implemented in plugin framework](https://github.com/GoogleCloudPlatform/magic-modules/pull/9742#issuecomment-1877373440), whereas all other data sources were implemented with SDKv2 before being migrated. The tests for the plural managed zone**s** data source [has fewer diffs](https://github.com/GoogleCloudPlatform/magic-modules/pull/10368/files#diff-2b33bfe3d86f43174f00731f2934c7d5d7ca209882c5f45c398108c35c1e0c76) due to this reason. I just realised that acc test was still skipped in VCR, so we'd need to let that run through.

Also, I added `id` to the schema of the google_dns_managed_zone singular data source, to enable me to reuse it when re-writing the google_dns_managed_zones plural data source in SDKv2. If I didn't do this the PF version of google_dns_managed_zones plural data source would have had `id` for each list item and the SDKv2 version would have been missing `id` for each list item, which would have been a breaking change.

rileykarson (approved):

PR #10376, opened 2024-04-05
Add New Resource resource_google_project_iam_member_remove

hao-nan-li:
Fixes https://github.com/hashicorp/terraform-provider-google/issues/11724

```release-note:new-resource
`google_project_iam_member_remove`
```

### DIFF
--- a/mmv1/third_party/terraform/services/resourcemanager/resource_google_project_iam_member_remove.go
+++ b/mmv1/third_party/terraform/services/resourcemanager/resource_google_project_iam_member_remove.go
@@ -46,10 +46,10 @@ func ResourceGoogleProjectIamMemberRemove() *schema.Resource {
 
 func resourceGoogleProjectIamMemberRemoveCreate(d *schema.ResourceData, meta interface{}) error {
 	config := meta.(*transport_tpg.Config)
-	
-	project:= d.Get("project").(string)
+
+	project := d.Get("project").(string)
 	role := d.Get("role").(string)
-	member:= d.Get("member").(string)
+	member := d.Get("member").(string)
 
 	iamPolicy, err := config.NewResourceManagerClient(config.UserAgent).Projects.GetIamPolicy(project,
 		&cloudresourcemanager.GetIamPolicyRequest{
@@ -87,8 +87,8 @@ func resourceGoogleProjectIamMemberRemoveCreate(d *schema.ResourceData, meta int
 func resourceGoogleProjectIamMemberRemoveRead(d *schema.ResourceData, meta interface{}) error {
 
 	project := d.Get("project").(string)
-	role:= d.Get("role").(string)
-	member:= d.Get("member").(string)
+	role := d.Get("role").(string)
+	member := d.Get("member").(string)
 
 	if err := d.Set("role", role); err != nil {
 		return fmt.Errorf("Error setting role: %s", err)
--- a/mmv1/third_party/terraform/services/resourcemanager/resource_google_project_iam_member_remove.go
+++ b/mmv1/third_party/terraform/services/resourcemanager/resource_google_project_iam_member_remove.go
@@ -6,7 +6,7 @@ import (
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 	"github.com/hashicorp/terraform-provider-google/google/tpgiamresource"
 	transport_tpg "github.com/hashicorp/terraform-provider-google/google/transport"
-	"google.golang.org/api/cloudresourcemanager/v1"
+	cloudresourcemanager "google.golang.org/api/cloudresourcemanager/v1"
 )
 
 func ResourceGoogleProjectIamMemberRemove() *schema.Resource {
@@ -16,22 +16,23 @@ func ResourceGoogleProjectIamMemberRemove() *schema.Resource {
 		Delete: resourceGoogleProjectIamMemberRemoveDelete,
 
 		Schema: map[string]*schema.Schema{
-			"role": {
-				Type:     schema.TypeString,
-				ForceNew: true,
-				Required: true,
-			},
 			"project": {
 				Type:        schema.TypeString,
 				ForceNew:    true,
 				Required:    true,
-				Description: `The project that the service account will be created in. Defaults to the provider project configuration.`,
+				Description: `The project id of the target project.`,
+			},
+			"role": {
+				Type:        schema.TypeString,
+				ForceNew:    true,
+				Required:    true,
+				Description: `The target role that should be removed.`,
 			},
 			"member": {
 				Type:        schema.TypeString,
 				ForceNew:    true,
 				Required:    true,
-				Description: `The Identity of the service account in the form 'serviceAccount:{email}'. This value is often used to refer to the service account in order to grant IAM permissions.`,
+				Description: `The IAM principal that should not have the target role.`,
 			},
 		},
 		UseJSONNumber: true,
@@ -44,6 +45,7 @@ func resourceGoogleProjectIamMemberRemoveCreate(d *schema.ResourceData, meta int
 	project := d.Get("project").(string)
 	role := d.Get("role").(string)
 	member := d.Get("member").(string)
+
 	found := false
 	iamPolicy, err := config.NewResourceManagerClient(config.UserAgent).Projects.GetIamPolicy(project,
 		&cloudresourcemanager.GetIamPolicyRequest{
@@ -51,20 +53,24 @@ func resourceGoogleProjectIamMemberRemoveCreate(d *schema.ResourceData, meta int
 				RequestedPolicyVersion: tpgiamresource.IamPolicyVersion,
 			},
 		}).Do()
+	if err != nil {
+		return transport_tpg.HandleNotFoundError(err, d, d.Id())
+	}
 
 	for i := 0; i < len(iamPolicy.Bindings); i++ {
-		for j := 0; j < len(iamPolicy.Bindings[i].Members); j++ {
-			if member == iamPolicy.Bindings[i].Members[j] {
-				if role == iamPolicy.Bindings[i].Role {
+		if role == iamPolicy.Bindings[i].Role {
+			for j := 0; j < len(iamPolicy.Bindings[i].Members); j++ {
+				if member == iamPolicy.Bindings[i].Members[j] {
 					found = true
 					iamPolicy.Bindings[i].Members = append(iamPolicy.Bindings[i].Members[:j], iamPolicy.Bindings[i].Members[j+1:]...)
+					break
 				}
 			}
 		}
 	}
 
 	if found == false {
-		fmt.Printf("Could not find Member %s with the corresponding role %s.", member, role)
+		fmt.Printf("[DEBUG] Could not find Member %s with the corresponding role %s. No removal necessary", member, role)
 	} else {
 		updateRequest := &cloudresourcemanager.SetIamPolicyRequest{
 			Policy:     iamPolicy,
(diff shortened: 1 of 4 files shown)


mmv1/third_party/terraform/services/resourcemanager/resource_google_project_iam_member_remove.go
```diff
@@ -0,0 +1,109 @@
...
+	iamPolicy, err := config.NewResourceManagerClient(config.UserAgent).Projects.GetIamPolicy(project,
+		&cloudresourcemanager.GetIamPolicyRequest{
+			Options: &cloudresourcemanager.GetPolicyOptions{
+				RequestedPolicyVersion: tpgiamresource.IamPolicyVersion,
+			},
+		}).Do()
+	for _, bind := range iamPolicy.Bindings {
+		for _, existingMember := range bind.Members {
+			if member == existingMember {
+				if role == bind.Role {
```
rileykarson, line 63: Can you test this with multiple roles assigned to the identity we're removing the single-role membership from?

mmv1/third_party/terraform/services/resourcemanager/resource_google_project_iam_member_remove.go
```diff
@@ -0,0 +1,109 @@
...
+					bind.Role = "role/remove"
```
rileykarson, line 64: This creates the right request, but I'm not entirely sure why it does- can you point to docs or code that makes this work? I expected an API error on first read.

mmv1/third_party/terraform/services/resourcemanager/resource_google_project_iam_member_remove_test.go
```diff
@@ -0,0 +1,60 @@
...
+	"testing"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
+)
+
+func TestAccProjectIamMemberRemove_basic(t *testing.T) {
+	t.Parallel()
+
+	org := envvar.GetTestOrgFromEnv(t)
+	members := "user:gterraformtest7@gmail.com"
```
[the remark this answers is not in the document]
rileykarson: (if not, make sure this is an OTA that is shared w/ the team)

mmv1/third_party/terraform/website/docs/r/google_project_iam_member_remove.html.markdown
```diff
@@ -0,0 +1,36 @@
+---
+subcategory: "Cloud Platform"
+description: |-
+ Allows removal of a binding policy.
```
rileykarson, line 4:
```suggestion
 Allows removal of a member:role pairing from an IAM policy.
```

rileykarson: Same wording below as well.

mmv1/third_party/terraform/website/docs/r/google_project_iam_member_remove.html.markdown
```diff
@@ -0,0 +1,36 @@
...
+* `member` - (Required) Identities that will be granted the privilege in `role`.
+  Each entry can have one of the following values:
+  * **user:{emailid}**: An email address that represents a specific Google account. For example, alice@gmail.com or joe@example.com.
+  * **serviceAccount:{emailid}**: An email address that represents a service account. For example, my-other-app@appspot.gserviceaccount.com.
+  * **group:{emailid}**: An email address that represents a Google group. For example, admins@example.com.
+  * **domain:{domain}**: A G Suite domain (primary, instead of alias) name that represents all the users of that domain. For example, google.com or example.com.
+
+* `role` - (Required) The role that should be removed. 
+
+* `project` - (Required) The project id of the target project.
```
rileykarson, line 36: Not consistent w/ current code. See other comment to correct.

mmv1/third_party/terraform/services/resourcemanager/resource_google_project_iam_member_remove.go
hao-nan-li: According to the logs, the role is removed. https://cloud.google.com/resource-manager/reference/rest/v1/projects/setIamPolicy. On the other hand it might not be the best message to send.

mmv1/third_party/terraform/services/resourcemanager/resource_google_project_iam_member_remove.go
```diff
@@ -0,0 +1,109 @@
...
+			tpgresource.DefaultProviderProject,
+		),
+
+		Schema: map[string]*schema.Schema{
+			"role": {
+				Type:     schema.TypeString,
+				ForceNew: true,
+				Required: true,
+			},
+			"project": {
```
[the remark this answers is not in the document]
hao-nan-li: Will do.

mmv1/third_party/terraform/website/docs/r/google_project_iam_member_remove.html.markdown
hao-nan-li: Will do.

mmv1/third_party/terraform/services/resourcemanager/resource_google_project_iam_member_remove_test.go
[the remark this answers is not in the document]
hao-nan-li: I'm not sure I fully understood this comment, what is a service account in-band?

mmv1/third_party/terraform/services/resourcemanager/resource_google_project_iam_member_remove.go
rileykarson: I don't see why we're adding role/remove on that page, though, rather than removing the role from the binding (or the binding altogether)

How would this work if there were more members in the same binding, as well?

mmv1/third_party/terraform/services/resourcemanager/resource_google_project_iam_member_remove_test.go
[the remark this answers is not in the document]
rileykarson: Creating a google_service_account resource within the test (in-band within the test) rather than pre-creating it out-of-band manually.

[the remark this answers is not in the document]
hao-nan-li: Could you give an example so I could refer to?

mmv1/third_party/terraform/services/resourcemanager/resource_google_project_iam_member_remove.go
hao-nan-li: remove role involves a GetIamPolicy request and then a SetIamPolicy request. https://yaqs.corp.google.com/eng/q/7281177105365204992. 

For this specific case, we only consider member as a string, not a set. So only one binding could be removed at a time.

mmv1/third_party/terraform/services/resourcemanager/resource_google_project_iam_member_remove_test.go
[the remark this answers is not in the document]
rileykarson: https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/google_service_account_iam#google_service_account_iam_member

Creating google_service_account.sa and then assigning it as a member in google_service_account_iam_member.gce-default-account-iam, **not** google_service_account_iam_member.admin-account-iam

mmv1/third_party/terraform/services/resourcemanager/resource_google_project_iam_member_remove.go
rileykarson: > So only one binding could be removed at a time.

That's correct- we can only remove one binding at a time, and only if we just removed the last member from the binding.

We should verify we are not removing the entire binding when we are removing only a single member in a binding with multiple members.

> remove role involves a GetIamPolicy request and then a SetIamPolicy request

Not sure I understand. We are doing a GetIamPolicy and then a SetIamPolicy request.

mmv1/third_party/terraform/services/resourcemanager/resource_google_project_iam_member_remove.go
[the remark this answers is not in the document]
hao-nan-li: I'm getting `Error: SetNew only operates on computed keys - project is not one` if I replace O+C to `required`. Not sure why this is happening.

mmv1/third_party/terraform/services/resourcemanager/resource_google_project_iam_member_remove_test.go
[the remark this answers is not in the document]
hao-nan-li: I'd ideally not intend to make the basic test too complicated. Can we simply follow what https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/third_party/terraform/services/resourcemanager/resource_google_project_iam_binding_test.go.erb#L342 is doing?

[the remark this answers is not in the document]
rileykarson: We can cover this with more test cases in that case.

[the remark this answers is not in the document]
rileykarson: Oh, this is the service account thread: We should ensure we're using an OTA, at least. That test should have used one but may not have.

mmv1/third_party/terraform/services/resourcemanager/resource_google_project_iam_member_remove.go
hao-nan-li: Good catch, looks like for each binding, there's one role matching to one member. I think this might cover it already but I need to verify.
https://stackoverflow.com/questions/61661116/want-to-assign-multiple-google-cloud-iam-roles-to-a-service-account-via-terrafor

mmv1/third_party/terraform/services/resourcemanager/resource_google_project_iam_member_remove_test.go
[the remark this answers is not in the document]
hao-nan-li: I'm looking to add another test with multiple members binding to the same role. What does OTA stand for? Is `"user:gterraformtest1@gmail.com"` part of the OTA? That was used in the test referred by the hyperlink above.

mmv1/third_party/terraform/services/resourcemanager/resource_google_project_iam_member_remove.go
hao-nan-li: removed this line as it caused some API errors.

mmv1/third_party/terraform/services/resourcemanager/resource_google_project_iam_member_remove.go
hao-nan-li: added a test for multiple members on a same role, test says passed, but it's not behaving correctly, will attempt to fix that.

hao-nan-li:
For both tests, the the member is removed from the setIamPolicy request, I end up with this error:

~ resource "google_project_iam_binding" "bar" {
  |         id      = "tf-test-7shlwzzaiz/roles/editor"
  |       ~ members = [
  |           + "user:gterraformtest1@gmail.com",
  |         ]
  |         # (3 unchanged attributes hidden)
  |     }

Looks like after removing the member, something goes wrong with the resource that creates them, will take a look tmr.

mmv1/third_party/terraform/services/resourcemanager/resource_google_project_iam_member_remove.go
```diff
@@ -0,0 +1,109 @@
...
+
+	d.SetId(fmt.Sprintf("%s/%s/%s", project, member, role))
+	if err != nil {
+		return transport_tpg.HandleNotFoundError(err, d, d.Id())
+	}
+
+	return resourceGoogleProjectIamMemberRemoveRead(d, meta)
+}
+
+func resourceGoogleProjectIamMemberRemoveRead(d *schema.ResourceData, meta interface{}) error {
```
rileykarson, line 87: Drift detection does not currently work- this step will currently never do anything. This should either remove the resource from state if the membership is present or we should manage a boolean on the resource that we use to determine whether we should call Update where we go ahead and remove the value.

rileykarson: Resolved in my commit

rileykarson:
@c2thorn I've picked this up, can you do a full re-review?

mmv1/third_party/terraform/services/resourcemanager/resource_google_project_iam_member_remove.go
```diff
@@ -0,0 +1,132 @@
...
+					found = true
+					break
+				}
+			}
+		}
+	}
+
+	if found {
+		fmt.Printf("[DEBUG] found policy for resource %v, removing from state", d.Id())
+		d.SetId("")
```
slevenick, line 121: So if the IAM member exists we consider this resource not present to trigger the recreation that removes the IAM member? I think that makes sense

mmv1/third_party/terraform/services/resourcemanager/resource_google_project_iam_member_remove.go
```diff
@@ -0,0 +1,132 @@
...
+				if member == iamPolicy.Bindings[i].Members[j] {
+					found = true
+					break
+				}
+			}
+		}
+	}
+
+	if found {
+		fmt.Printf("[DEBUG] found policy for resource %v, removing from state", d.Id())
```
slevenick, line 120: Change this log line to something like: "found IAM member matching ..."

mmv1/third_party/terraform/services/resourcemanager/resource_google_project_iam_member_remove.go
```diff
@@ -0,0 +1,132 @@
...
+				ForceNew:    true,
+				Required:    true,
+				Description: `The IAM principal that should not have the target role.`,
+			},
+		},
+		UseJSONNumber: true,
+	}
+}
+
+func resourceGoogleProjectIamMemberRemoveCreate(d *schema.ResourceData, meta interface{}) error {
```
slevenick, line 42: Do we expect any race conditions when creating this within a large Terraform config? For example, depending on the ordering of creates in a large config a binding may be applied before or after the member remove resource.

I guess there's not much we can do there and the recreate should appear on next plan time if it is applied in the wrong order

mmv1/third_party/terraform/services/resourcemanager/resource_google_project_iam_member_remove_test.go
```diff
@@ -0,0 +1,269 @@
...
+
+resource "time_sleep" "wait_20s" {
+  depends_on = [google_project_iam_binding.bar]
+  create_duration = "20s"
+}
+
+resource "google_project_iam_member_remove" "foo" {
+  role     = "roles/editor"
+  project  = google_project.project.project_id
+  member  = "user:gterraformtest1@gmail.com"
```
slevenick, line 113: Do we own the emails used in this test? If not, can we grant a lesser permission than "editor" on the project even though it's short-lived?

slevenick (approved):
Two minor changes requested, otherwise looks good

mmv1/third_party/terraform/services/resourcemanager/resource_google_project_iam_member_remove_test.go
rileykarson: They're OTAs held by our team, yep. I confirmed that earlier w/ Hao.

mmv1/third_party/terraform/services/resourcemanager/resource_google_project_iam_member_remove.go
rileykarson: Yep! From docs:

> If the membership is ever re-added, the next refresh will clear this
resource from state, proposing re-adding it to correct the membership.

rileykarson: I considered changing it to an output boolean we change in a CustomizeDiff, which would make adding import make sense, but figured it wasn't worth the extra effort on the initial implementation. We could always change that in a major version (and potentially even a minor depending how strict we are 🤷‍♀️ )

mmv1/third_party/terraform/services/resourcemanager/resource_google_project_iam_member_remove.go
rileykarson: Yep, exactly. I'll add some more details to the docs.

PR #10379, opened 2024-04-05
Go rewrite example templates (part 1)

zli82016:
1. (Part 1) Convert the example templates to Go templates with the [script](https://github.com/zli82016/magic-modules/blob/go-rewrite-example-templates-script/mmv1/example-template-converter.go)
2. The script doesn't work with a couple of example templates. I rewrite these templates to work with the script. 
  * mmv1/templates/terraform/examples/network_endpoints.tf.erb
  * mmv1/templates/terraform/examples/access_context_manager_access_policy_scoped.tf.erb
  * mmv1/templates/terraform/examples/bigquerydatatransfer_config_scheduled_query.tf.erb

```release-note:none

```

### DIFF
--- a/mmv1/templates/terraform/examples/bigquerydatatransfer_config_scheduled_query.tf.erb
+++ b/mmv1/templates/terraform/examples/bigquerydatatransfer_config_scheduled_query.tf.erb
@@ -25,7 +25,7 @@ resource "google_bigquery_data_transfer_config" "<%= ctx[:primary_resource_id] %
 resource "google_bigquery_dataset" "my_dataset" {
   depends_on = [google_project_iam_member.permissions]
 
-  dataset_id    = "<%= ctx[:vars]['dataset_id'].delete("-") %>"
+  dataset_id    = "<%= ctx[:vars]['dataset_id'] %>"
   friendly_name = "foo"
   description   = "bar"
   location      = "asia-northeast1"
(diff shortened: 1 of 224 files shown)


mmv1/templates/terraform/examples/bigquerydatatransfer_config_scheduled_query.tf.erb
```diff
@@ -25,7 +25,7 @@ resource "google_bigquery_data_transfer_config" "<%= ctx[:primary_resource_id] %
 resource "google_bigquery_dataset" "my_dataset" {
   depends_on = [google_project_iam_member.permissions]
 
-  dataset_id    = "<%= ctx[:vars]['dataset_id'].delete("-") %>"
+  dataset_id    = "<%= ctx[:vars]['dataset_id'] %>"
```
c2thorn, line 28: what did this do before? I guess it's not needed?

zli82016: It deletes the `-` from the value of dataset_id. Yes, it is not needed.   The value of `dataset_id` is `my_dataset` and doesn't have `-`

https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/products/bigquerydatatransfer/Config.yaml#L48

c2thorn (approved):

c2thorn:
@NickElliot fyi

PR #10383, opened 2024-04-05
Add enable_google_ml_integration setting to google_sql_database_insta…

rsm3171990:
This PR adds support for the new field enable_google_ml_integration to google_sql_database_instance settings. This new field is only available for google.golang.org/api v0.170.0 or after. Locally tested the change and it works. Added Acctests (unit tests) and they passed.

Fixes https://github.com/hashicorp/terraform-provider-google/issues/17785

```
sql: added `enable_google_ml_integration` field to `google_sql_database_instance` resource settings

```


melinath:
It looks like the SQL user failure is known and probably unrelated: https://github.com/hashicorp/terraform-provider-google/issues/16704

melinath (approved):
LGTM

PR #10386, opened 2024-04-05
Made TestAccLoggingProjectSink_default use a separate project

melinath:
Fixed https://github.com/hashicorp/terraform-provider-google/issues/16701.

```release-note:none

```


melinath:
Based on https://cloud.google.com/logging/docs/default-settings it sounds like the default log sink is either enabled or disabled in new projects based on a setting at a higher level, and maybe can't actually be controlled directly.

melinath:
@pengq-google FYI in case you want to take a look

c2thorn (approved):

pengq-google:
Thanks! ^_^

PR #10389, opened 2024-04-08
Minor fix to address ambiguity in error message.

spapi17:

```release-note:enhancement
composer: fixed validation on `google_composer_environment` resource so it will identify a disallowed upgrade to Composer 3 before attempting to provide feedback that's specific to using Composer 3
```


spapi17:
Hi @SarahFrench, could you please take a look?

SarahFrench (approved):

PR #10401, opened 2024-04-09
Show diff for computed nested labels fields when creating resources

zli82016:
fixes https://github.com/hashicorp/terraform-provider-google/issues/17756

Error
```
Error: Provider produced inconsistent final plan
│ 
│ When expanding the plan for module.cloud_run.google_cloud_run_service.main
│ to include new values learned so far during apply, provider
│ "registry.terraform.io/hashicorp/google-beta" produced an invalid new value
│ for .metadata[0].labels: was null, but now
```

Resource `google_cloud_run_service` and `google_cloud_run_domain_mapping` have nested labels fields `metadata.0.labels` and `metadata.0.annotations`. 

In the CustomizeDiff function [SetMetadataLabelsDiff](https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/third_party/terraform/tpgresource/labels.go#L134), `SetNew` is applied on "metadata" field with the object including `terraform_labels` and `effective_labels`. It introduced two bugs.
1. when creating resource, computed `metadata.0.labels` and `metadata.0.annotations` fields don't appear in terraform plan  
2. After creating the resource with the normal labels value, and then updating the labels fields with the computed value will fail, similar to root labels field  https://github.com/hashicorp/terraform-provider-google/issues/16217. 
This bug cannot be fixed with the similar way in https://github.com/GoogleCloudPlatform/magic-modules/pull/10182.
The reason is that `metadata.0.terraform_labels` and `metadat.0.effective_labels` fields cannot be set to computed due to a bug that SetNew doesn't work on nested fields
in terraform sdk https://github.com/hashicorp/terraform-plugin-sdk/issues/459

The solution to fix these two bugs in my mind is to move `terraform_labels` and `effective_labels` to root level in provider release 6.0.
Before the release 6.0, the first bug can be fixed in this PR and I don't find an easy way to fix the second bug.

```release-note:bug
cloudrun: fixed the bug that computed `metadata.0.labels` and `metadata.0.annotations` fields don't appear in terraform plan when creating resource `google_cloud_run_service` and `google_cloud_run_domain_mapping`
```

### DIFF
--- a/mmv1/third_party/terraform/tpgresource/annotations.go
+++ b/mmv1/third_party/terraform/tpgresource/annotations.go
@@ -54,8 +54,8 @@ func SetMetadataAnnotationsDiff(_ context.Context, d *schema.ResourceDiff, meta 
 
 	// Fix the bug that the computed and nested "annotations" field disappears from the terraform plan.
 	// https://github.com/hashicorp/terraform-provider-google/issues/17756
-	// The bug is introducted by SetNew on "metadata" field with the object including "effective_annotations".
-	// "effective_annotations" cannot be set directrly due to a bug that SetNew doesn't work on nested fields
+	// The bug is introduced by SetNew on "metadata" field with the object including "effective_annotations".
+	// "effective_annotations" cannot be set directly due to a bug that SetNew doesn't work on nested fields
 	// in terraform sdk.
 	// https://github.com/hashicorp/terraform-plugin-sdk/issues/459
 	if !d.GetRawPlan().GetAttr("metadata").AsValueSlice()[0].GetAttr("annotations").IsWhollyKnown() {
--- a/mmv1/third_party/terraform/tpgresource/annotations.go
+++ b/mmv1/third_party/terraform/tpgresource/annotations.go
@@ -52,6 +52,16 @@ func SetMetadataAnnotationsDiff(_ context.Context, d *schema.ResourceDiff, meta 
 		return nil
 	}
 
+	// Fix the bug that the computed and nested "annotations" field disappears from the terraform plan.
+	// https://github.com/hashicorp/terraform-provider-google/issues/17756
+	// The bug is introducted by SetNew on "metadata" field with the object including "effective_annotations".
+	// "effective_annotations" cannot be set directrly due to a bug that SetNew doesn't work on nested fields
+	// in terraform sdk.
+	// https://github.com/hashicorp/terraform-plugin-sdk/issues/459
+	if !d.GetRawPlan().GetAttr("metadata").AsValueSlice()[0].GetAttr("annotations").IsWhollyKnown() {
+		return nil
+	}
+
 	raw := d.Get("metadata.0.annotations")
 	if raw == nil {
 		return nil
--- a/mmv1/third_party/terraform/tpgresource/labels.go
+++ b/mmv1/third_party/terraform/tpgresource/labels.go
@@ -139,8 +139,8 @@ func SetMetadataLabelsDiff(_ context.Context, d *schema.ResourceDiff, meta inter
 
 	// Fix the bug that the computed and nested "labels" field disappears from the terraform plan.
 	// https://github.com/hashicorp/terraform-provider-google/issues/17756
-	// The bug is introducted by SetNew on "metadata" field with the object including terraform_labels and effective_labels.
-	// "terraform_labels" and "effective_labels" cannot be set directrly due to a bug that SetNew doesn't work on nested fields
+	// The bug is introduced by SetNew on "metadata" field with the object including terraform_labels and effective_labels.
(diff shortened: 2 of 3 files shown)


mmv1/third_party/terraform/tpgresource/annotations.go
```diff
@@ -52,6 +52,16 @@ func SetMetadataAnnotationsDiff(_ context.Context, d *schema.ResourceDiff, meta
 		return nil
 	}
 
+	// Fix the bug that the computed and nested "annotations" field disappears from the terraform plan.
+	// https://github.com/hashicorp/terraform-provider-google/issues/17756
+	// The bug is introducted by SetNew on "metadata" field with the object including "effective_annotations".
```
rafaeltello, line 57: nit: introduced

mmv1/third_party/terraform/tpgresource/annotations.go
```diff
@@ -52,6 +52,16 @@ func SetMetadataAnnotationsDiff(_ context.Context, d *schema.ResourceDiff, meta
...
+	// "effective_annotations" cannot be set directrly due to a bug that SetNew doesn't work on nested fields
```
rafaeltello, line 58: nit: directly

mmv1/third_party/terraform/tpgresource/labels.go
```diff
@@ -137,6 +137,16 @@ func SetMetadataLabelsDiff(_ context.Context, d *schema.ResourceDiff, meta inter
 		return nil
 	}
 
+	// Fix the bug that the computed and nested "labels" field disappears from the terraform plan.
+	// https://github.com/hashicorp/terraform-provider-google/issues/17756
+	// The bug is introducted by SetNew on "metadata" field with the object including terraform_labels and effective_labels.
```
rafaeltello, line 142: nit: introduced

mmv1/third_party/terraform/tpgresource/labels.go
```diff
@@ -137,6 +137,16 @@ func SetMetadataLabelsDiff(_ context.Context, d *schema.ResourceDiff, meta inter
...
+	// "terraform_labels" and "effective_labels" cannot be set directrly due to a bug that SetNew doesn't work on nested fields
```
rafaeltello, line 143: nit: directly

rafaeltello (approved):

zli82016:
The failed test is unrelated

mmv1/third_party/terraform/tpgresource/annotations.go
```diff
@@ -52,6 +52,16 @@ func SetMetadataAnnotationsDiff(_ context.Context, d *schema.ResourceDiff, meta
...
 	}
 
+	// Fix the bug that the computed and nested "annotations" field disappears from the terraform plan.
+	// https://github.com/hashicorp/terraform-provider-google/issues/17756
+	// The bug is introduced by SetNew on "metadata" field with the object including "effective_annotations".
+	// "effective_annotations" cannot be set directly due to a bug that SetNew doesn't work on nested fields
+	// in terraform sdk.
+	// https://github.com/hashicorp/terraform-plugin-sdk/issues/459
+	if !d.GetRawPlan().GetAttr("metadata").AsValueSlice()[0].GetAttr("annotations").IsWhollyKnown() {
+		return nil
```
rileykarson, line 62: At creation time this works because we'll keep the `effective_annotations` value unknown (/(known after apply) as it appears to end users) as a result of this, right?

rileykarson (approved):

mmv1/third_party/terraform/tpgresource/annotations.go
zli82016: At creation time this works because we'll keep the `annotations` value unknown (/(known after apply) as it appears to end users) as a result of this. Otherwise, `annotations` doesn't appear in the terraform plan.

I tested with the `labels` field
Before the fix, in terraform plan
```
  + resource "google_cloud_run_service" "default" {
      + metadata {
          + effective_annotations = (known after apply)
          + effective_labels      = (known after apply)
          + generation            = 0
          + namespace             = "terraform-dev-zhenhuali"
          + resource_version      = (known after apply)
          + self_link             = (known after apply)
          + terraform_labels      = (known after apply)
          + uid                   = (known after apply)
        }

```

With the fix, in terraform plan
```
  + resource "google_cloud_run_service" "default" {
      + metadata {
          + effective_annotations = (known after apply)
          + effective_labels      = (known after apply)
          + generation            = (known after apply)
          + labels                = (known after apply)
          + namespace             = "terraform-dev-zhenhuali"
          + resource_version      = (known after apply)
          + self_link             = (known after apply)
          + terraform_labels      = (known after apply)
          + uid                   = (known after apply)
        }
```

rileykarson: Oh, the failure was on labels itself. Got it, thanks!

PR #10402, opened 2024-04-09
Add support for intercept_children in sinks

Jbodeau:
Add support for recently created `intercept_children` property on log sinks, used for intercepting aggregated sinks. Docs for feature: https://cloud.google.com/logging/docs/export/aggregated_sinks

```release-note:enhancement
logging: added `intercept_children` field to `google_logging_organization_sink` and `google_logging_folder_sink` resources
```

### DIFF
--- a/mmv1/third_party/terraform/services/logging/resource_logging_folder_sink.go
+++ b/mmv1/third_party/terraform/services/logging/resource_logging_folder_sink.go
@@ -38,6 +38,13 @@ func ResourceLoggingFolderSink() *schema.Resource {
 		Default:     false,
 		Description: `Whether or not to include children folders in the sink export. If true, logs associated with child projects are also exported; otherwise only logs relating to the provided folder are included.`,
 	}
+	schm.Schema["intercept_children"] = &schema.Schema{
+		Type:        schema.TypeBool,
+		Optional:    true,
+		ForceNew:    true,
+		Default:     false,
+		Description: `Whether or not to intercept logs from child projects. If true, matching logs will not match with sinks in child resources, except _Required sinks. This sink will be visible to child resources when listing sinks.`,
+	}
 
 	return schm
 }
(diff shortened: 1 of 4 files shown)


mmv1/third_party/terraform/services/logging/resource_logging_folder_sink.go
```diff
@@ -38,6 +38,13 @@ func ResourceLoggingFolderSink() *schema.Resource {
 		Default:     false,
 		Description: `Whether or not to include children folders in the sink export. If true, logs associated with child projects are also exported; otherwise only logs relating to the provided folder are included.`,
 	}
+	schm.Schema["intercept_children"] = &schema.Schema{
+		Type:        schema.TypeBool,
+		Optional:    true,
+		ForceNew:    true,
+		Default:     false,
```
NickElliot, line 45: Is this field currently returned by current API get requests on existing resources not made with it? This is potentially a breaking change -- if at plan-time the API does not return the value this will result in terraform planning to recreate any existing instances of this resource.

NickElliot (commented):
Hi, one question for clarification

mmv1/third_party/terraform/services/logging/resource_logging_folder_sink.go
Jbodeau: Yes. I was actually planning to remove the ForceNew behavior for both IncludeChildren and this new InterceptChildren field in a subsequent PR, so I will just roll that into this PR. Will comment again once it's added.

Jbodeau: Ok, this is updated now with the other commit. ForceNew has been removed for the new field, and also the IncludeChildren field, which is supposed to have nearly identical behavior.

NickElliot (approved):
LGTM!

PR #10405, opened 2024-04-09
Allowed modifying request headers via pre_create, pre_update, or pre_delete custom code

melinath:
Resolved https://github.com/hashicorp/terraform-provider-google/issues/16331.

The specific resource issue this was blocking (https://github.com/hashicorp/terraform-provider-google/issues/15550) was resolved separately so no changes are needed there.

Given we don't have a clear idea of what features we might need to support in the future, and this does not currently seem to be a common use case, I don't want to implement any kind of special MM field at this time; however, this change would be necessary for any changes to MM fields in the future as well.

```release-note:none

```

### DIFF
--- a/mmv1/third_party/terraform/transport/transport.go
+++ b/mmv1/third_party/terraform/transport/transport.go
@@ -24,12 +24,13 @@ type SendRequestOptions struct {
 	UserAgent            string
 	Body                 map[string]any
 	Timeout              time.Duration
+	Headers              http.Header
 	ErrorRetryPredicates []RetryErrorPredicateFunc
 	ErrorAbortPredicates []RetryErrorPredicateFunc
 }
 
 func SendRequest(opt SendRequestOptions) (map[string]interface{}, error) {
-	reqHeaders := make(http.Header)
+	reqHeaders := opt.Headers
 	reqHeaders.Set("User-Agent", opt.UserAgent)
 	reqHeaders.Set("Content-Type", "application/json")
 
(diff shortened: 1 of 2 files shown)


melinath:
@c2thorn this should be ready for review now; the failing test is unrelated.

mmv1/third_party/terraform/transport/transport.go
```diff
@@ -24,12 +24,16 @@ type SendRequestOptions struct {
...
 	Timeout              time.Duration
+	Headers              http.Header
 	ErrorRetryPredicates []RetryErrorPredicateFunc
 	ErrorAbortPredicates []RetryErrorPredicateFunc
 }
 
 func SendRequest(opt SendRequestOptions) (map[string]interface{}, error) {
-	reqHeaders := make(http.Header)
+	reqHeaders := opt.Headers
+	if reqHeaders == nil {
```
c2thorn, line 34: is there an instance where we expect it to be nil here? Or just in case

c2thorn: nevermind, SendRequest is used a lot

c2thorn (approved):
LGTM, nil check seems to cover all cases

mmv1/third_party/terraform/transport/transport.go
melinath: The test failures in https://github.com/GoogleCloudPlatform/magic-modules/pull/10405#issuecomment-2046194626 were all because this nil handling wasn't present.

c2thorn:
FYI @NickElliot some lines added to resource.erb here, so we could close out a bug in time

PR #9910, opened 2024-02-02
Adding Parallelstore Beta Service Provider

unnatinadupalli:
Adding Parallelstore Beta Service Provider

Added Instance.yaml and product.yaml in mmv1/products/parallelstore/
Added Example in mmv1/templates/terraform/examples/
Added test in mmv1/third_party/terraform/services/parallelstore/ with basic and update 

```release-note:new-resource
`google_parallelstore_instance`
```

PR #10406, opened 2024-04-10
Fix issue in GA version of resource_parallelstore_instance_test.go

SarahFrench:
Addresses issue identified here : https://github.com/GoogleCloudPlatform/magic-modules/pull/9910#issuecomment-2047428662

The test file generated into the GA provider needs to have a package name, [like this example](https://github.com/hashicorp/magic-modules/blob/e93e811188d812d43d993ba8af9dcb351a630001/mmv1/third_party/terraform/services/apigateway/resource_api_gateway_api_config_test.go.erb#L1-L3). 

```release-note:none

```


c2thorn (approved):
Thanks for the fix!

PR #10407, opened 2024-04-10
Drop IAM resources from EkmConnection tests.

tdbhacks:
This should fix issue [#17535](https://github.com/hashicorp/terraform-provider-google/issues/17535), the IAM permissions are being reset after each run but the test passed when permissions were granted manually.

```release-note:none

```


tdbhacks:
@c2thorn FYI

c2thorn (approved):

PR #10409, opened 2024-04-10
Go rewrite constants templates

zli82016:
1. Add two fields `TargetVersionName` and `Compiler` to Resource object. These two fields are used in constants templates.
2. Convert constants templates with the [script](https://github.com/zli82016/magic-modules/blob/go-rewrite-example-templates-script/mmv1/template-converter.go#L84).

```release-note:none

```


zli82016:
It seems like there is a bug in mmv1-check-templates, as it should not fail when a template is removed.

c2thorn (approved):

zli82016:
Rebased the main branch to resolve the conflcts.

PR #10414, opened 2024-04-11
go rewrite - most of testing template

c2thorn:
checking in most of the testing template. Need to finish the DestroyProducer section and then fix an issue with test variables. You can see the current generated diff here: https://github.com/c2thorn/terraform-provider-google-beta/commit/c63f1ea080e1648568537b1bdd84943bbcfe9ab7

```release-note:none

```


mmv1/provider/terraform.go
```diff
@@ -144,6 +144,28 @@ func (t *Terraform) GenerateResource(object api.Resource, templateData TemplateD
 	}
 }
 
+func (t *Terraform) GenerateResourceTests(object api.Resource, templateData TemplateData, outputFolder string) {
+	eligibleExample := false
+	for _, example := range object.Examples {
+		if !example.SkipTest {
+			if object.ProductMetadata.VersionObjOrClosest(t.Version.Name).CompareTo(object.ProductMetadata.VersionObjOrClosest(example.MinVersion)) > 0 {
+				eligibleExample = true
```
zli82016, line 152: Should jump out the loop with `break` statement once `eligibleExample` is `true`?

mmv1/templates/terraform/examples/go/test_file.go.tmpl
```diff
@@ -0,0 +1,128 @@
+// Copyright (c) HashiCorp, Inc.
```
zli82016, line 1: I wonder if this file can be moved to `mmv1/templates/terraform` folder, which has resource and doc templates?

mmv1/provider/terraform.go
c2thorn: yes you are right! Added

mmv1/templates/terraform/examples/go/test_file.go.tmpl
c2thorn: I moved it to `mmv1/templates/terraform/examples/base_configs/` since that is where the `test_file.go.erb` is right now

zli82016 (approved):
Nice work. LGTM.

PR #10415, opened 2024-04-11
Add beta version to networkconnectivity.produc.yaml

simalcubuk:
Add beta api version to mmv1/products/networkconnectivity/product.yaml. This is needed to add RegionalEndpoint resource to magic-modules

Pbluc API doc for the new resource to be added: https://cloud.google.com/network-connectivity/docs/reference/networkconnectivity/rest/v1beta/projects.locations.regionalEndpoints

If your PR is still work in progress, please create it in draft mode

```release-note:none
This is not a breaking change, this change is only to enable adding a new resource in beta version.
```


trodge:
It looks like the endpoints for some existing resources aren't in the beta version, causing test failures. Is there a plan to address this?

PR #10421, opened 2024-04-11
Fix template-check to ignore files that have been deleted

roaks3:
Fixes issue where deleting an erb file would trigger the check on that file, which would fail because the file doesn't exist. The lowercase `d` means exclude "Delete" diffs.

```release-note:none

```


zli82016 (approved):
LGTM. Thanks for fixing it.

roaks3:
No problem, thanks for raising the issue!

PR #10410, opened 2024-04-10
Extracted config ordering logic into a helper function

melinath:
Related to https://github.com/hashicorp/terraform-provider-google/issues/16243 - if this looks good I'll do a follow-up PR to fix that & related problems.

**Release Note Template for Downstream PRs (will be copied)**

```release-note:none

```

PR #10429, opened 2024-04-12
Made apigee organization sort properties in state by config order

melinath:
Use the new function introduced in https://github.com/GoogleCloudPlatform/magic-modules/pull/10410 to order apigee properties according to the order in config. Resolved https://github.com/hashicorp/terraform-provider-google/issues/16243. Resolved https://github.com/hashicorp/terraform-provider-google/issues/13274.

```release-note:bug
apigee: fixed permadiff in ordering of `google_apigee_organization.properties.property`.
```

### DIFF
--- a/mmv1/products/apigee/Organization.yaml
+++ b/mmv1/products/apigee/Organization.yaml
@@ -60,8 +60,6 @@ examples:
     test_env_vars:
       org_id: :ORG_ID
       billing_account: :BILLING_ACCT
-    ignore_read_extra:
-      - properties
     skip_docs:
       true
       # Resource creation race
@@ -78,8 +76,6 @@ examples:
     test_env_vars:
       org_id: :ORG_ID
       billing_account: :BILLING_ACCT
-    ignore_read_extra:
-      - properties
     skip_docs:
       true
       # Resource creation race
@@ -103,8 +99,6 @@ examples:
       beta
       # Resource creation race
     skip_vcr: true
-    ignore_read_extra:
-      - properties
   - !ruby/object:Provider::Terraform::Examples
     name: 'apigee_organization_cloud_full_disable_vpc_peering'
     skip_test:
@@ -235,6 +229,7 @@ properties:
       - !ruby/object:Api::Type::Array
         name: 'property'
         description: List of all properties in the object.
+        custom_flatten: 'templates/terraform/custom_flatten/apigee_organization_property.go.erb'
         item_type: !ruby/object:Api::Type::NestedObject
           properties:
             - !ruby/object:Api::Type::String
--- a/mmv1/products/apigee/Organization.yaml
+++ b/mmv1/products/apigee/Organization.yaml
@@ -52,71 +52,79 @@ examples:
     name: 'apigee_organization_cloud_basic'
     skip_test:
       true
-      # This is a more verbose version of the above that creates all
-      # the resources needed for the acceptance test.
+  # This is a more verbose version of the above that creates all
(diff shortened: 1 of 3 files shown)


melinath:
Manual non-REPLAYING run: https://hashicorp.teamcity.com/buildConfiguration/TerraformProviders_GoogleCloud_GOOGLE_BETA_MMUPSTREAMTESTS_GOOGLEBETA_PACKAGE_APIGEE/132406?hideTestsFromDependencies=false&hideProblemsFromDependencies=false

melinath:
I forgot that import will still not necessarily work as we'd want since (I believe?) it doesn't have a config order to work with.

melinath:
Manual run of all ApigeeOrganization tests: https://hashicorp.teamcity.com/buildConfiguration/TerraformProviders_GoogleCloud_GOOGLE_BETA_MMUPSTREAMTESTS_GOOGLEBETA_PACKAGE_APIGEE/132408?hideProblemsFromDependencies=false&hideTestsFromDependencies=false

melinath:
okay... I think this should be ready for review now. :-)

mmv1/products/apigee/Organization.yaml
```diff
@@ -82,47 +82,49 @@ examples:
...
+  # resources in this test are in the GA API, we depend on a service
+  # identity resource which is only available in the beta provider.
   - !ruby/object:Provider::Terraform::Examples
     name: 'apigee_organization_cloud_full_test'
     primary_resource_id: 'org'
     test_env_vars:
       org_id: :ORG_ID
       billing_account: :BILLING_ACCT
+    ignore_read_extra:
+      - properties
```
SarahFrench, line 102: Is it still necessary to have `properties` included in `ImportStateVerifyIgnore` for acceptance tests? I assume the flatten function will perform the necessary reordering during the import test step, but it could be my mental model of the test framework is incorrect.

melinath: I tried taking these out and [one of the tests](https://hashicorp.teamcity.com/buildConfiguration/TerraformProviders_GoogleCloud_GOOGLE_BETA_MMUPSTREAMTESTS_GOOGLEBETA_PACKAGE_APIGEE/132406?buildTab=tests&expandedTest=build%3A%28id%3A132406%29%2Cid%3A2000000000) failed at the ImportStateVerify step. I believe that is expected (?) because import logic doesn't have a config to match the ordering of, so it'll always end up being in alphabetical order. The ordering logic should still be getting verified by the check for a diff after apply.

SarahFrench: That makes sense - my mental model was that import results in the Import function running and then the Read function running (so `flattenApigeeOrganizationPropertiesProperty` would be used), but even if there's a resource block containing an ordered list that's present at the time of import that information isn't necessarily available during that operation.

Thanks for checking!

SarahFrench (approved):
Approved - tests are passing with the new change 🎉  and my nit about `ignore_read_extra` was unnecessary.

It'll be good to see [TestAccApigeeOrganization_apigeeOrganizationCloudFullDisableVpcPeeringTestExample](https://hashicorp.teamcity.com/test/6707821277812550034?currentBuildTypeId=TerraformProviders_GoogleCloud_GOOGLE_BETA_NIGHTLYTESTS_GOOGLEBETA_PACKAGE_APIGEE&currentProjectId=TerraformProviders_GoogleCloud_GOOGLE_BETA_NIGHTLYTESTS&expandTestHistoryChartSection=true) going green soon!

PR #10434, opened 2024-04-12
Added new resource "Security Policy Rule"

matheusaleixo-cit:
Adds a new resource "google_compute_security_policy_rule", allowing an alternative to handling rules in the security policy resource itself. Functionally similar to [regional security policy rules](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/compute_region_security_policy_rule).

FIxes: https://github.com/hashicorp/terraform-provider-google/issues/5622

```release-note:new-resource
`google_compute_security_policy_rule`
```


mmv1/templates/terraform/examples/security_policy_rule_basic.tf.erb
```diff
@@ -0,0 +1,23 @@
+resource "google_compute_security_policy" "default" {
+  provider    = google-beta
```
shuyama1, line 2: I'm still in the middle of reviewing this PR. A quick suggestion is removing `provider    = google-beta` in the examples as this resource + its examples are no longer marked as beta-only in the latest commit

mmv1/templates/terraform/examples/security_policy_rule_default_rule.tf.erb
```diff
@@ -0,0 +1,37 @@
+resource "google_compute_security_policy" "default" {
+  name        = "<%= ctx[:vars]['sec_policy_name'] %>"
+  description = "basic global security policy"
+  type        = "CLOUD_ARMOR"
+}
+
+# A default rule is generated when creating the security_policy resource, import is needed to patch it
```
matheusaleixo-cit, line 7: The same issue exists in the regional security policy rule resource: https://github.com/hashicorp/terraform-provider-google/issues/15687

matheusaleixo-cit:
I commented out the fields inside the "network_match" block,  and  "title", "description" and "network" in the "expr" block.
Those fields are presented in the API docs, but doesn't seen to be supported. The "expr" fields don't update the values in the resource at all and "network_match" is only used for policies of type CLOUD_ARMOR_NETWORK, which the global security policy resource does not support.

mmv1/products/compute/SecurityPolicyRule.yaml
```diff
@@ -286,74 +287,75 @@ properties:
...
-        name: 'srcRegionCodes'
-        description: |
-          Two-letter ISO 3166-1 alpha-2 country code associated with the source IP address.
-        item_type: Api::Type::String
-      - !ruby/object:Api::Type::Array
-        name: 'srcAsns'
-        description: |
-          BGP Autonomous System Number associated with the source IP address.
-        item_type: Api::Type::Integer
+  # These fields are used for security policies of the type CLOUD_ARMOR_NETWORK, this type is not supported by the global security policy resource.
```
shuyama1, line 290: If the following fields won't be supported by the global security policy resource, I think we can just remove them.

PR #10435, opened 2024-04-12
Add data source for retrieving multiple objects from a GCS bucket

bestefreund:
Adds a new data source  "data_google_storage_bucket_objects", allowing to retrieve multiple objects from a GCS storage bucket.

Fixes: https://github.com/hashicorp/terraform-provider-google/issues/17842

```release-note:new-datasource
`google_storage_bucket_objects`
```

### DIFF
--- a/mmv1/third_party/terraform/services/storage/data_source_google_storage_bucket_objects.go
+++ b/mmv1/third_party/terraform/services/storage/data_source_google_storage_bucket_objects.go
@@ -1,0 +1,154 @@
+// Copyright (c) HashiCorp, Inc.
+// SPDX-License-Identifier: MPL-2.0
+package storage
+
+import (
+	"fmt"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"github.com/hashicorp/terraform-provider-google/google/tpgresource"
+	transport_tpg "github.com/hashicorp/terraform-provider-google/google/transport"
+)
+
+func DataSourceGoogleStorageBucketObjects() *schema.Resource {
+	return &schema.Resource{
+		Read: datasourceGoogleStorageBucketObjectsRead,
+		Schema: map[string]*schema.Schema{
+			"bucket": {
+				Type:     schema.TypeString,
+				Required: true,
+			},
+			"match_glob": {
+				Type:     schema.TypeString,
+				Optional: true,
+			},
+			"prefix": {
+				Type:     schema.TypeString,
+				Optional: true,
+			},
+			"bucket_objects": {
+				Type:     schema.TypeList,
+				Computed: true,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"content_type": {
+							Type:     schema.TypeString,
+							Computed: true,
+						},
+						"media_link": {
+							Type:     schema.TypeString,
+							Computed: true,
+						},
+						"name": {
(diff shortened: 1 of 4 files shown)


mmv1/third_party/terraform/services/storage/data_source_google_storage_bucket_objects.go
```diff
@@ -0,0 +1,160 @@
+// Copyright (c) HashiCorp, Inc.
+// SPDX-License-Identifier: MPL-2.0
+// Copyright (c) HashiCorp, Inc.
```
slevenick, line 3: Double license?

mmv1/third_party/terraform/website/docs/d/storage_bucket_objects.html.markdown
```diff
@@ -0,0 +1,45 @@
...
+
+## Attributes Reference
+
+The following attributes are exported:
+
+* `bucket_objects` - A list of retrieved objects contained in the provided GCS bucket. Structure is [defined below](#nested_bucket_objects).
+
+<a name="nested_bucket_objects"></a>The `bucket_objects` block supports:
+
+* `content_type` - (Computed) [Content-Type](https://tools.ietf.org/html/rfc7231#section-3.1.1.5) of the object data.
```
slevenick, line 41: These don't need (Computed). It's assumed because they're all attributes

mmv1/third_party/terraform/services/storage/data_source_google_storage_bucket_objects.go
bestefreund: This is a mistake.
I'll take care

mmv1/third_party/terraform/website/docs/d/storage_bucket_objects.html.markdown
bestefreund: This is a mistake.
I'll take care

JonasUebe (approved):

mmv1/third_party/terraform/services/storage/data_source_google_storage_bucket_objects.go
```diff
@@ -0,0 +1,158 @@
...
+
+func DataSourceGoogleStorageBucketObjects() *schema.Resource {
+	return &schema.Resource{
+		Read: datasourceGoogleStorageBucketObjectsRead,
+		Schema: map[string]*schema.Schema{
+			"bucket": {
+				Type:     schema.TypeString,
+				Required: true,
+			},
+			"delimiter": {
```
slevenick, line 21: Looks like this never gets used, should we remove it?

slevenick (changes requested):

mmv1/third_party/terraform/services/storage/data_source_google_storage_bucket_objects.go
bestefreund: @slevenick 
Thanks for the hint!

I'd like to add keep and use it because the change to use it wouldn't be a big deal and the `delimiter` seems to be a good optional parameter to query for bucket objects in regard to the Storage API Reference.

bestefreund: @slevenick 
How can I mark the requested change as done?

slevenick (approved):

PR #5989, opened 2022-04-27
Switch to using folder LIST over SEARCH

ScottSuarez:
closes https://github.com/hashicorp/terraform-provider-google/issues/10286

acc to b/228499462 
```
Search provides an eventually consistent view of the folders
http://cloud/resource-manager/reference/rest/v2/folders/search.

List is strongly consistent
http://cloud/resource-manager/reference/rest/v2/folders/list
```
**Release Note Template for Downstream PRs (will be copied)**

```release-note:none

```

PR #10439, opened 2024-04-12
Update active folder to support both LIST and SEARCH methods

ScottSuarez:
closes https://github.com/hashicorp/terraform-provider-google/issues/17677

We [switched](https://github.com/GoogleCloudPlatform/magic-modules/pull/5989) from `LIST` to `SEARCH` as `SEARCH` is [eventually consistent](https://cloud.google.com/resource-manager/reference/rest/v3/folders/search#:~:text=eventually%20consistent) while `LIST` is [strongly consistent](https://cloud.google.com/resource-manager/reference/rest/v3/folders/list#:~:text=list()%20provides%20a-,strongly%20consistent,-view%20of%20the).

Since there is a [use case](https://github.com/hashicorp/terraform-provider-google/issues/17677) for `SEARCH` I've re-added support for it and a way to switch between them.

```release-note:enhancement
resourcemanager: added the field `api_method` to datasource `google_active_folder` so you can use either `SEARCH` or `LIST` to find your folder
```


mmv1/third_party/terraform/website/docs/d/active_folder.html.markdown
```diff
@@ -25,6 +25,8 @@ The following arguments are supported:
 
 * `parent` - (Required) The resource name of the parent Folder or Organization.
 
+* `rest_method` - (Optional) The method to use to search for the folder. Both `LIST` and `SEARCH` are supported. The default value is `LIST` as SEARCH is [eventually consistent](https://cloud.google.com/resource-manager/reference/rest/v3/folders/search#:~:text=eventually%20consistent) while `LIST` is [strongly consistent](https://cloud.google.com/resource-manager/reference/rest/v3/folders/list#:~:text=list()%20provides%20a-,strongly%20consistent,-view%20of%20the).
```
melinath, line 28:
```suggestion
* `rest_method` - (Optional) The API method to use to search for the folder. Valid values are `LIST` and `SEARCH`. `LIST` is [strongly consistent](https://cloud.google.com/resource-manager/reference/rest/v3/folders/list#:~:text=list()%20provides%20a-,strongly%20consistent,-view%20of%20the) and requires `resourcemanager.folders.list` on the parent folder, while `SEARCH` is [eventually consistent](https://cloud.google.com/resource-manager/reference/rest/v3/folders/search#:~:text=eventually%20consistent) and only returns folders that the user has `resourcemanager.folders.get` permission on.
```

melinath (approved):
LGTM apart from a suggestion regarding the docs.

mmv1/third_party/terraform/website/docs/d/active_folder.html.markdown
ScottSuarez: done

mmv1/third_party/terraform/services/resourcemanager/data_source_google_active_folder.go
```diff
@@ -26,6 +27,13 @@ func DataSourceGoogleActiveFolder() *schema.Resource {
 				Type:     schema.TypeString,
 				Computed: true,
 			},
+			"rest_method": {
```
melinath, line 30: non-blocking: what would you think about calling this `api_method` rather than `rest_method`? Although it's called a REST API, SEARCH is not a standard REST method...

melinath (approved):

PR #10445, opened 2024-04-15
re-enable Dataproc Metastore Federation tests

NickElliot:
Fixes https://github.com/hashicorp/terraform-provider-google/issues/13710


SarahFrench:
I'm going to use your PR to test this (https://github.com/hashicorp/terraform-provider-google/issues/13710#issuecomment-2220367863) out, sorry for notification spam

SarahFrench:
Looks like TestAccDataprocMetastoreService_dataprocMetastoreServicePrivateServiceConnectExample is failing because that example includes:

```tf
  network_config {
    consumers {
      subnetwork = google_compute_subnetwork.subnet.id
    }
  }
```

which I assume is causing the address to be created with that subnet, resulting in failure to delete the subnet.

I think that's a new test failure, so not something to attempt to fix here (though maybe a handwritten test could remove the `network_config` block as it's last step to enable deletion?).

SarahFrench (commented):
I'd suggest we re-skip generating tests from the `dataproc_metastore_service_private_service_connect` example and merge this PR to close the original issue. I can open a new issue specific to the new test failure.

Does that sound ok?

NickElliot:
@SarahFrench that sounds good to me, feel free to merge this once those have been done!

mmv1/products/metastore/Service.yaml
```diff
@@ -101,7 +100,6 @@ examples:
       metastore_service_name: 'metastore-metadata'
   - !ruby/object:Provider::Terraform::Examples
     name: 'dataproc_metastore_service_private_service_connect'
-    skip_test: true
```
SarahFrench, line 104: Please resume this test being skipped here!

SarahFrench (approved):

PR #10455, opened 2024-04-15
Add resource tags to BigQuery Table

wj-chen:
Adds support for resource tags to BigQuery Table.

```release-note:enhancement
bigquery: added `resource_tags` field to `google_bigquery_table` resource
```

### DIFF
--- a/mmv1/third_party/terraform/services/bigquery/resource_bigquery_table.go.erb
+++ b/mmv1/third_party/terraform/services/bigquery/resource_bigquery_table.go.erb
@@ -1799,6 +1799,10 @@ func resourceBigQueryTableColumnDrop(config *transport_tpg.Config, userAgent str
 		return err
 	}
 
+	if table.Schema == nil {
+		return nil
+	}
+
 	newTableFields := map[string]bool{}
 	for _, field := range table.Schema.Fields {
 		newTableFields[field.Name] = true
@@ -1836,6 +1840,18 @@ func resourceBigQueryTableDelete(d *schema.ResourceData, meta interface{}) error
 	if d.Get("deletion_protection").(bool) {
 		return fmt.Errorf("cannot destroy instance without setting deletion_protection=false and running `terraform apply`")
 	}
+	<% unless version == 'ga' -%>
+	if v, ok := d.GetOk("resource_tags"); ok {
+		var resourceTags []string
+
+		for k, v := range v.(map[string]interface{}) {
+			resourceTags = append(resourceTags, fmt.Sprintf("%s:%s", k, v.(string)))
+		}
+
+		return fmt.Errorf("cannot destroy instance without clearing the following resource tags: %v", resourceTags)
+	}
+
+	<% end -%>
 	config := meta.(*transport_tpg.Config)
 	userAgent, err := tpgresource.GenerateUserAgentString(d, config.UserAgent)
 	if err != nil {
--- a/mmv1/third_party/terraform/services/bigquery/resource_bigquery_table.go.erb
+++ b/mmv1/third_party/terraform/services/bigquery/resource_bigquery_table.go.erb
@@ -1429,15 +1429,7 @@ func resourceTable(d *schema.ResourceData, meta interface{}) (*bigquery.Table, e
 	}
 
   <% unless version == 'ga' -%>
-	if v, ok := d.GetOk("resource_tags"); ok {
-		resourceTags := map[string]string{}
-
-		for k, v := range v.(map[string]interface{}) {
-			resourceTags[k] = v.(string)
-		}
-
-		table.ResourceTags = resourceTags
-	}
+	table.ResourceTags = tpgresource.ExpandStringMap(d, "resource_tags")
 
 	<% end -%>
 	return table, nil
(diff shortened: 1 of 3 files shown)


rileykarson (commented):
I verified the account has the appropriate permissions, so that error message is probably an effective 404 unless hierarchical permissions aren't getting respected here.

(I haven't taken a further look at the change itself yet)

wj-chen:
> I verified the account has the appropriate permissions, so that error message is probably an effective 404 unless hierarchical permissions aren't getting respected here.
> 
> (I haven't taken a further look at the change itself yet)

Thank you for checking the permissions! It turned out to be a race condition where resources were referenced before creation, which should be fixed by the newest commit.

mmv1/third_party/terraform/services/bigquery/resource_bigquery_table.go.erb
```diff
@@ -1419,6 +1428,18 @@ func resourceTable(d *schema.ResourceData, meta interface{}) (*bigquery.Table, e
 		table.TableConstraints = tableConstraints
 	}
 
+  <% unless version == 'ga' -%>
+	if v, ok := d.GetOk("resource_tags"); ok {
```
rileykarson, line 1432: non-blocking, but we could use `ExpandStringMap` here: https://github.com/GoogleCloudPlatform/magic-modules/blob/60a85738f569bbfa2731097270db8b19d5dcfebd/mmv1/third_party/terraform/tpgresource/utils.go#L231

(same result though 🤷 )

mmv1/third_party/terraform/services/bigquery/resource_bigquery_table.go.erb
```diff
@@ -1809,6 +1840,18 @@ func resourceBigQueryTableDelete(d *schema.ResourceData, meta interface{}) error
 	if d.Get("deletion_protection").(bool) {
 		return fmt.Errorf("cannot destroy instance without setting deletion_protection=false and running `terraform apply`")
 	}
+	<% unless version == 'ga' -%>
+	if v, ok := d.GetOk("resource_tags"); ok {
```
rileykarson, line 1844: Does the API have this check / does it take a long time to report a failure? Otherwise, if the API will reject the call there's not much value in doing this clientside first, since this will save 1s / one API call AIUI.

I'm a little hesitant about just checking state here- Terraform will sometimes short-circuit delete calls so Terraform may have drifted from the real config. I'm not sure we ever drift from state though, it's been a bit since I knew all the rules.

mmv1/third_party/terraform/services/bigquery/resource_bigquery_table.go.erb
```diff
@@ -1809,6 +1840,18 @@ func resourceBigQueryTableDelete(d *schema.ResourceData, meta interface{}) error
...
+		var resourceTags []string
+
+		for k, v := range v.(map[string]interface{}) {
+			resourceTags = append(resourceTags, fmt.Sprintf("%s:%s", k, v.(string)))
+		}
+
+		return fmt.Errorf("cannot destroy instance without clearing the following resource tags: %v", resourceTags)
```
rileykarson, line 1851:
```suggestion
		return fmt.Errorf("cannot destroy table %v without clearing the following resource tags: %v", d.Id(), resourceTags)
```

rileykarson (commented):
Generally LGTM- one question on the deletion precondition before approving/merging, and a couple nits inline.

mmv1/third_party/terraform/services/bigquery/resource_bigquery_table.go.erb
wj-chen: Done.

mmv1/third_party/terraform/services/bigquery/resource_bigquery_table.go.erb
wj-chen: Done.

mmv1/third_party/terraform/services/bigquery/resource_bigquery_table.go.erb
wj-chen: The deletion will go through actually, but what I discovered while debugging the VCR test failure is that the tag value bindings will remain if a table is deleted while having resource tags specified, and that makes it impossible to delete the tag value resource, failing at "it's still has an association". As a follow-up I'll reach out to our API team to discuss if the behavior should be fixed on the API-level.

wj-chen:
> Generally LGTM- one question on the deletion precondition before approving/merging, and a couple nits inline.

Thanks for the review! The new revision is out. Hoping to get this merged before the release cut today.

rileykarson (approved):

mmv1/third_party/terraform/services/bigquery/resource_bigquery_table.go.erb
```diff
@@ -1807,8 +1830,20 @@ func resourceBigQueryTableColumnDrop(config *transport_tpg.Config, userAgent str
 
 func resourceBigQueryTableDelete(d *schema.ResourceData, meta interface{}) error {
 	if d.Get("deletion_protection").(bool) {
-		return fmt.Errorf("cannot destroy instance without setting deletion_protection=false and running `terraform apply`")
+		return fmt.Errorf("cannot destroy table %v without setting deletion_protection=false and running `terraform apply`", d.Id())
```
rileykarson, line 1833: Good catch, thanks @wj-chen!

PR #10456, opened 2024-04-16
chore(ci): convert vcr_merge into golang command

iyabchen:
Convert vcr_merge into golang. 

Have not done integration tests, not sure how this works against a real scenario. 

```release-note:none

```

### DIFF
--- a/.ci/magician/cmd/vcr_merge.go
+++ b/.ci/magician/cmd/vcr_merge.go
@@ -83,7 +83,7 @@ func mergeCassettes(basePath, baseBranch, prPath string, runner source.Runner) {
 		runner,
 	)
 
-	rmCassettes(fmt.Sprintf("%s/%s/", basePath, prPath), runner)
+	// rmCassettes(fmt.Sprintf("%s/%s/", basePath, prPath), runner)
 }
 
 func listCassettes(path string, runner source.Runner) error {
--- a/.ci/magician/cmd/vcr_merge.go
+++ b/.ci/magician/cmd/vcr_merge.go
@@ -23,69 +23,67 @@ var vcrMergeCmd = &cobra.Command{
 	2. Run gsutil to list, copy, and remove the vcr cassettes fixtures.
 	`,
 	Args: cobra.ExactArgs(1),
-	Run: func(cmd *cobra.Command, args []string) {
+	RunE: func(cmd *cobra.Command, args []string) error {
 		reference := args[0]
 		fmt.Println("Reference commit SHA: ", reference)
+
 		githubToken, ok := os.LookupEnv("GITHUB_TOKEN_CLASSIC")
 		if !ok {
-			fmt.Println("Did not provide GITHUB_TOKEN_CLASSIC environment variable")
-			os.Exit(1)
-		}
-		rnr, err := exec.NewRunner()
-		if err != nil {
-			fmt.Println("Error creating Runner: ", err)
-			os.Exit(1)
+			return fmt.Errorf("did not provide GITHUB_TOKEN_CLASSIC environment variable")
 		}
 
 		baseBranch := os.Getenv("BASE_BRANCH")
 		if baseBranch == "" {
-			baseBranch = "main"
+			return fmt.Errorf("environment variable BASE_BRANCH is empty")
 		}
+
+		rnr, err := exec.NewRunner()
+		if err != nil {
+			return fmt.Errorf("error creating Runner: %w", err)
+		}
+
 		gh := github.NewClient(githubToken)
-		execVCRMerge(gh, reference, baseBranch, rnr)
+		return execVCRMerge(gh, reference, baseBranch, rnr)
 	},
 }
 
-func execVCRMerge(gh GithubClient, sha string, baseBranch string, runner source.Runner) {
+func execVCRMerge(gh GithubClient, sha string, baseBranch string, runner source.Runner) error {
 	arr, err := gh.GetPullRequests("closed", baseBranch, "updated", "desc")
 	if err != nil {
-		fmt.Println("Error getting pull requests: ", err)
-		os.Exit(1)
+		return fmt.Errorf("error getting pull requests: %w", err)
 	}
 	pr := findPRBySHA(sha, arr)
 	if pr == nil {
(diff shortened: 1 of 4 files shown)


.ci/magician/cmd/vcr_merge.go
```diff
@@ -0,0 +1,142 @@
...
+
+	The command expects the following as arguments:
+	1. Reference commit SHA
+
+	It then performs the following operations:
+	1. Get the latest closed PR matching the reference commit SHA.
+	2. Run gsutil to list, copy, and remove the vcr cassettes fixtures.
+	`,
+	Args: cobra.ExactArgs(1),
+	Run: func(cmd *cobra.Command, args []string) {
```
melinath, line 26: I'd like us to move towards using RunE instead of Run + os.Exit (https://github.com/hashicorp/terraform-provider-google/issues/17479 for more details.)

.ci/magician/cmd/vcr_merge.go
```diff
@@ -0,0 +1,142 @@
...
+		rnr, err := exec.NewRunner()
+		if err != nil {
+			fmt.Println("Error creating Runner: ", err)
+			os.Exit(1)
+		}
+
+		baseBranch := os.Getenv("BASE_BRANCH")
+		if baseBranch == "" {
+			baseBranch = "main"
+		}
```
melinath, line 43:
```suggestion
```

This fallback was added for backwards-compatibility when feature branches were first introduced - it should be fine to consider BASE_BRANCH missing as an error now.

.ci/magician/cmd/vcr_merge.go
```diff
@@ -0,0 +1,142 @@
...
+		fmt.Println("Error getting pull requests: ", err)
+		os.Exit(1)
+	}
+	pr := findPRBySHA(sha, arr)
+	if pr == nil {
+		fmt.Printf("Not finding any matching PR with commit SHA %s\n", sha)
+		return
+	}
+
+	mergeCassettes(false, baseBranch, pr.Number, runner)
```
melinath, line 61: could we pass the prefix here instead of an `isBeta` bool? Or possibly even a bucket name and a path?

.ci/magician/cmd/vcr_merge.go
```diff
@@ -0,0 +1,142 @@
...
+	mergeCassettes(true, baseBranch, pr.Number, runner)
+}
+
+func mergeCassettes(isBeta bool, baseBranch string, prNumber int, runner source.Runner) {
```
melinath, line 65: There are a lot of variables being created that are very similar to each other and only get used once. What would you think about building full paths sooner and then building single-use paths as part of the arguments to a function? For example,

```go
prPath = fmt.Sprintf("%s/refs/heads/auto-pr-%d", prefix, prNumber)
//...
err := listCassettes(prPath + "/fixtures", runner)
//...
cpCassettes(prPath + "/fixtures/*", branchPath + "/fixtures", runner)
//...
rmCassettes(prPath, runner)
```

.ci/magician/cmd/vcr_merge_test.go
```diff
@@ -0,0 +1,96 @@
...
+
+import (
+	"magician/github"
+	"strings"
+	"testing"
+
+	"github.com/google/go-cmp/cmp"
+)
+
+func TestExecVCRMerge(t *testing.T) {
```
melinath, line 11: After the RunE switch could you add some tests for error cases here?

melinath (changes requested):

melinath:
@trodge @shuyama1 FYI in case you want to take a look.

.ci/magician/cmd/vcr_merge.go
iyabchen: Done

.ci/magician/cmd/vcr_merge.go
iyabchen: Done

.ci/magician/cmd/vcr_merge.go
iyabchen: Done

.ci/magician/cmd/vcr_merge.go
iyabchen: Done

.ci/magician/cmd/vcr_merge_test.go
iyabchen: Added a test specific for RunE, not sure that's the way. It does not check argument number. 
It is strange to test a subcommand, since if calling Execute on the subcommand, it will find the parent to execute first. 
Or we move all the logic within RunE into the execVcrMerge function itself, then only testing that function.

melinath: sorry for being unclear - what I meant is that after converting to RunE it should be possible to test `execVCRMerge` error cases, because it will return an error instead of exiting.

I don't mean think that moving the arg parsing logic into execVcrMerge is necessary; keeping that in RunE makes sense.

melinath (changes requested):

.ci/magician/cmd/vcr_merge_test.go
iyabchen: Currently execVcrMerge does not return any error except at github get pull requests, to test that route to return error, will need to change the mock to make it return error upon some strange parameter?

melinath: hm, yeah, looking at this again I don't think there's necessarily a reason to add tests for those cases.

melinath:
@shuyama1 what do you think is the best way test this / roll this out?

shuyama1:
> @shuyama1 what do you think is the best way test this / roll this out?

I can't think of a good way to test this pre merge. We can skip the remove cassettes step for now, so it anything goes wrong, we'll still have the cassettes stored in the PR branch that we can copy over manually and we can restore it after we confirm it works (I wonder how much this will help though). 

Sorry I missed your previous comment. I can take a look at this PR now if needed.

shuyama1 (commented):
LGTM! 💯

melinath (approved):
LGTM - going ahead with merging this PR. We should probably wait until Wednesday or Thursday to test out the switchover (by editing the yaml to switch to this command.)

PR #10468, opened 2024-04-17
go rewrite - Finish test template

c2thorn:
- Differentiates Documentation examples vs Test examples
- Adds TestCheckDestroy to test template

Generates an exact copy of `google-beta/services/datafusion/resource_data_fusion_instance_generated_test.go` except for an extra import of 
```
	"github.com/hashicorp/terraform-provider-google-beta/google-beta/envvar"
```
Which I think can be resolved by running go imports elsewhere.

```release-note:none

```

### DIFF
--- a/mmv1/api/resource/examples.go
+++ b/mmv1/api/resource/examples.go
@@ -18,6 +18,7 @@ import (
 	"fmt"
 	"net/url"
 	"path/filepath"
+	"strings"
 	"text/template"
 
 	"github.com/GoogleCloudPlatform/magic-modules/mmv1/google"
@@ -157,7 +158,8 @@ type Examples struct {
 	// your test so avoid if you can.
 	PullExternal bool `yaml:"pull_external"`
 
-	HCLText string
+	DocumentationHCLText string
+	TestHCLText          string
 }
 
 func (e *Examples) UnmarshalYAML(n *yaml.Node) error {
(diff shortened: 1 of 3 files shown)


mmv1/provider/template_data.go
```diff
@@ -51,6 +51,7 @@ var TemplateFunctions = template.FuncMap{
 	"camelize":   google.Camelize,
 	"underscore": google.Underscore,
 	"contains":   strings.Contains,
+	"join":       strings.Join,
```
zli82016, line 54: Sorry, I just merged my [PR](https://github.com/GoogleCloudPlatform/magic-modules/pull/10460), which added this method. But no conflict is detected and reported in this PR.

mmv1/api/resource/examples.go
```diff
@@ -176,6 +178,32 @@ func (e *Examples) UnmarshalYAML(n *yaml.Node) error {
 }
 
 func (e *Examples) SetHCLText() {
+	e.DocumentationHCLText = ExecuteHCL(e)
+
+	copy := e
```
zli82016, line 183: It is better to keep the comments in https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/provider/terraform/examples.rb#L204-L220?

c2thorn: yes, added them in

mmv1/provider/template_data.go
c2thorn: No worries. I rebased just in case.

zli82016 (approved):
LGTM

PR #10473, opened 2024-04-17
Exclude output only and parent fields from missing test detection

trodge:
Parent fields are excluded because we check if the child fields are covered.

```release-note:none

```


melinath (approved):
LGTM

PR #10476, opened 2024-04-17
appengine: Suppress null automatic_scaling blocks during Read

rileykarson:
Fixes https://github.com/hashicorp/terraform-provider-google/issues/17531

I went with this over O+C or a DSF because the field exists in a oneof (`scaling` in proto) and those options both preserve the set-but-deeply-nil `automatic_scaling` block in state, meaning we'd eventually send it to the server even if the user was sending a different `scaling` entry. This approach is worse if they start returning values for existing fields because it will break users again, but it's at least mitigatable with lifecycle.ignore_changes if that happens.

```release-note:bug
appengine: added suppression for a diff in google_app_engine_standard_app_version.automatic_scaling when the block is unset in configuration
```

### DIFF
--- a/mmv1/templates/terraform/custom_flatten/appengine_standardappversion_automatic_scaling_handlenil.go.erb
+++ b/mmv1/templates/terraform/custom_flatten/appengine_standardappversion_automatic_scaling_handlenil.go.erb
@@ -1,0 +1,183 @@
+<%# The license inside this block applies to this file.
+	# Copyright 2024 Google Inc.
+	# Licensed under the Apache License, Version 2.0 (the "License");
+	# you may not use this file except in compliance with the License.
+	# You may obtain a copy of the License at
+	#
+	#     http://www.apache.org/licenses/LICENSE-2.0
+	#
+	# Unless required by applicable law or agreed to in writing, software
+	# distributed under the License is distributed on an "AS IS" BASIS,
+	# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+	# See the License for the specific language governing permissions and
+	# limitations under the License.
+-%>
+func flattenAppEngineStandardAppVersionAutomaticScaling(v interface{}, d *schema.ResourceData, config *transport_tpg.Config) interface{} {
+	if v == nil {
+		return nil
+	}
+	original := v.(map[string]interface{})
+	if len(original) == 0 {
+		return nil
+	}
+	transformed := make(map[string]interface{})
+	transformed["max_concurrent_requests"] =
+		flattenAppEngineStandardAppVersionAutomaticScalingMaxConcurrentRequests(original["maxConcurrentRequests"], d, config)
+	transformed["max_idle_instances"] =
+		flattenAppEngineStandardAppVersionAutomaticScalingMaxIdleInstances(original["maxIdleInstances"], d, config)
+	transformed["max_pending_latency"] =
+		flattenAppEngineStandardAppVersionAutomaticScalingMaxPendingLatency(original["maxPendingLatency"], d, config)
+	transformed["min_idle_instances"] =
+		flattenAppEngineStandardAppVersionAutomaticScalingMinIdleInstances(original["minIdleInstances"], d, config)
+	transformed["min_pending_latency"] =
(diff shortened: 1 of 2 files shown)


rileykarson:
I think TestAccAppEngineFlexibleAppVersion_update failed because it and TestAccAppEngineStandardAppVersion_update fight over the same singleton resources and were run literally in parallel here. I'll update them.

This should pass after /gcbrun

rileykarson:
Hmm- new error. /gcbrun

rileykarson:
Ah, I misread the initial failure. I misread an intended update call as the failing reason. The reason seems to be:

> ERROR: No matching distribution found for gunicorn==22.0.0 (from -r requirements.txt (line 2))

So unrelated to this change. Not sure why it's not failing in the nightlies then.

rileykarson:
Ah- it was https://github.com/GoogleCloudPlatform/magic-modules/pull/10469, that didn't run in VCR because it didn't modify a code file. https://github.com/GoogleCloudPlatform/magic-modules/pull/10477 to revert.

mmv1/templates/terraform/custom_flatten/appengine_standardappversion_automatic_scaling_handlenil.go.erb
```diff
@@ -0,0 +1,183 @@
...
+	transformed["max_pending_latency"] =
+		flattenAppEngineStandardAppVersionAutomaticScalingMaxPendingLatency(original["maxPendingLatency"], d, config)
+	transformed["min_idle_instances"] =
+		flattenAppEngineStandardAppVersionAutomaticScalingMinIdleInstances(original["minIdleInstances"], d, config)
+	transformed["min_pending_latency"] =
+		flattenAppEngineStandardAppVersionAutomaticScalingMinPendingLatency(original["minPendingLatency"], d, config)
+	transformed["standard_scheduler_settings"] =
+		flattenAppEngineStandardAppVersionAutomaticScalingStandardSchedulerSettings(original["standardSchedulerSettings"], d, config)
+
+	// solve for the following diff when no scaling settings are configured:
```
c2thorn, line 37: Can we note that this is the only custom/handwritten part while the rest of this flattener was copied from auto-generated code? Should give a hint to future maintainers on how to update this more easily in the future

c2thorn (approved):
Not in love with the downsides of this maintenance-wise, but it seems better for the user experience.

mmv1/templates/terraform/custom_flatten/appengine_standardappversion_automatic_scaling_handlenil.go.erb
rileykarson, line 37:
```suggestion
	// begin handwritten code (all other parts of this file are forked from generated code)
	// solve for the following diff when no scaling settings are configured:
```

mmv1/templates/terraform/custom_flatten/appengine_standardappversion_automatic_scaling_handlenil.go.erb
```diff
@@ -0,0 +1,183 @@
...
+	// if any new fields are added to the block or explicit defaults start getting
+	// returned, it will need to be updated
+	if transformed["max_concurrent_requests"] == nil && // even primitives are nil at this stage if they're not returned by the API
+		transformed["max_idle_instances"] == nil &&
+		transformed["max_pending_latency"] == nil &&
+		transformed["min_idle_instances"] == nil &&
+		transformed["min_pending_latency"] == nil &&
+		transformed["standard_scheduler_settings"] == nil {
+		return nil
+	}
```
rileykarson, line 61:
```suggestion
	}
	// end handwritten code
```

mmv1/templates/terraform/custom_flatten/appengine_standardappversion_automatic_scaling_handlenil.go.erb
rileykarson: Done

rileykarson:
> Not in love with the downsides of this maintenance-wise, but it seems better for the user experience.

Yeah- being in a oneof is the justification for taking this approach.

PR #10478, opened 2024-04-18
Change createSampleWorkflows field to createSampleIntegrations

rohitjangid:
Change `createSampleWorkflows` field to `createSampleIntegrations` in `google_integrations_client` to keep it consistent with terminology.

```release-note:enhancement
Change `createSampleWorkflows` field to `createSampleIntegrations` in `google_integrations_client` to keep it consistent with terminology.
```


rohitjangid:
> 
> ## Diff report
> 
> `google` provider: [Diff](https://github.com/modular-magician/terraform-provider-google/compare/auto-pr-10478-old..auto-pr-10478) ( 3 files changed, 17 insertions(+), 18 deletions(-)) `google-beta` provider: [Diff](https://github.com/modular-magician/terraform-provider-google-beta/compare/auto-pr-10478-old..auto-pr-10478) ( 3 files changed, 17 insertions(+), 18 deletions(-)) `terraform-google-conversion`: [Diff](https://github.com/modular-magician/terraform-google-conversion/compare/auto-pr-10478-old..auto-pr-10478) ( 1 file changed, 4 insertions(+), 4 deletions(-)) Open in Cloud Shell: [Diff](https://github.com/modular-magician/docs-examples/compare/auto-pr-10478-old..auto-pr-10478) ( 1 file changed, 1 insertion(+), 4 deletions(-))
> 
> ## Breaking Change(s) Detected
> The following breaking change(s) were detected within your pull request.
> 
> * Field `create_sample_workflows` within resource `google_integrations_client` was either removed or renamed - [reference](https://googlecloudplatform.github.io/magic-modules/develop/breaking-changes#resource-schema-field-removal-or-rename)
> 
> If you believe this detection to be incorrect please raise the concern with your reviewer. If you intend to make this change you will need to wait for a [major release](https://www.terraform.io/plugin/sdkv2/best-practices/versioning#example-major-number-increments) window. An `override-breaking-change` label can be added to allow merging.

@SarahFrench The change is expected. We do not have any usage for this resource and hence this change should be fine.

mmv1/templates/terraform/pre_create/integrations_client.go.erb
```diff
@@ -0,0 +1,5 @@
+// Translate `createSampleIntegrations` to `createSampleWorkflows`
```
SarahFrench, line 1: I see here that this is ensuring that the new field name that's used in the provider is 'translated back' to match the old field name that's still used in the API. Is the plan that the API will be changed soon, and this code will be removed at the same time?

SarahFrench (commented):
Hi there, thanks for your PR!

> @SarahFrench The change is expected. We do not have any usage for this resource and hence this change should be fine.

Renaming a field is typically a breaking change, that's [navigated via a deprecation process](https://googlecloudplatform.github.io/magic-modules/develop/breaking-changes/make-a-breaking-change/#add-deprecations-and-warnings-to-the-main-branch-of-magic-modules), because we don't know how many users may have a Terraform configuration that contains that field. I can see that this resource [was only added last month](https://github.com/GoogleCloudPlatform/magic-modules/commit/d874e11de524dddb932b70f4ef1bc7905fb41846) but how can you be sure that this doesn't break a user's existing configuration?

mmv1/templates/terraform/pre_create/integrations_client.go.erb
rohitjangid: The change in the API is not planned since we do not expect users to consume our api's directly. However, due to legacy field name, this is being referred as workflow instead integration in the api. Since terraform would allow users to directly use this field, we intend to fix the field name for terraform and later would plan to fix this field in api as well. We would remove this code once we mark the field in the api as deprecated.

rohitjangid:
> Hi there, thanks for your PR!
> 
> > @SarahFrench The change is expected. We do not have any usage for this resource and hence this change should be fine.
> 
> Renaming a field is typically a breaking change, that's [navigated via a deprecation process](https://googlecloudplatform.github.io/magic-modules/develop/breaking-changes/make-a-breaking-change/#add-deprecations-and-warnings-to-the-main-branch-of-magic-modules), because we don't know how many users may have a Terraform configuration that contains that field. I can see that this resource [was only added last month](https://github.com/GoogleCloudPlatform/magic-modules/commit/d874e11de524dddb932b70f4ef1bc7905fb41846) but how can you be sure that this doesn't break a user's existing configuration?

Added deprecation message

mmv1/products/integrations/Client.yaml
```diff
@@ -89,6 +90,13 @@ properties:
       Indicates if sample workflow should be created along with provisioning.
     immutable: true
     ignore_read: true
+    deprecation_message: "`create_sample_workflows` is deprecated and will be removed in a future major release. Use `create_sample_integrations` instead."
+  - !ruby/object:Api::Type::Boolean
+    name: 'createSampleIntegrations'
```
SarahFrench, line 95: Could you please [add a conflicts field](https://googlecloudplatform.github.io/magic-modules/develop/field-reference/#conflicts) to stop a user specifying both fields in the same resource block?

SarahFrench (commented):
Hi, thanks for adding the deprecation message! There's a 6.0.0 major release planned soon so the deprecated field can be removed then. Could you please open an issue about removing the field? That'll ensure the work is included in our planning.

Also, something I investigated was whether this PR should use [the `api_name` property](https://googlecloudplatform.github.io/magic-modules/develop/field-reference/#api_name), which we use to let a resource's fields have different names to API fields. I found that it's not currently usable because the deprecated field is present, so the existing solution with pre_create should be kept. However when the deprecated field is removed in future I think we should replace the use of the `pre_create` with adding `api_name: createSampleWorkflows` to the section describing createSampleIntegrations.


---

I'm currently not able to use this API in my GCP project, so could you please confirm via manual test or acceptance test that a resource can be provisioned using either the old or the new field? Thanks!

mmv1/products/integrations/Client.yaml
rohitjangid: Done

rohitjangid:
> Hi, thanks for adding the deprecation message! There's a 6.0.0 major release planned soon so the deprecated field can be removed then. Could you please open an issue about removing the field? That'll ensure the work is included in our planning.
> 
> Also, something I investigated was whether this PR should use [the `api_name` property](https://googlecloudplatform.github.io/magic-modules/develop/field-reference/#api_name), which we use to let a resource's fields have different names to API fields. I found that it's not currently usable because the deprecated field is present, so the existing solution with pre_create should be kept. However when the deprecated field is removed in future I think we should replace the use of the `pre_create` with adding `api_name: createSampleWorkflows` to the section describing createSampleIntegrations.
> 
> I'm currently not able to use this API in my GCP project, so could you please confirm via manual test or acceptance test that a resource can be provisioned using either the old or the new field? Thanks!

@SarahFrench Filed https://github.com/hashicorp/terraform-provider-google/issues/17928 as request.

I have also manually tested that resource could be provisioned with either fields.

mmv1/products/integrations/Client.yaml
```diff
@@ -83,18 +84,34 @@ properties:
...
+      Indicates if sample integrations should be created along with provisioning.
+    immutable: true
+    ignore_read: true
+    conflicts:
+      - create_sample_workflows
   - !ruby/object:Api::Type::Boolean
     name: 'provisionGmek'
     description: |
       Indicates provision with GMEK or CMEK.
+    deprecation_message: "`provision_gmek` is deprecated and will be removed in a future major release. Client would be provisioned as gmek if `cloud_kms_config` is not given."
```
rohitjangid, line 110: @SarahFrench The field is being marked as deprecated. The public docs are being updated however, discovery reflect the field deprecation.
https://integrations.googleapis.com/$discovery/rest?version=v1

rohitjangid:
Hi @SarahFrench
All issues are now resolved. PTAL.

mmv1/products/integrations/Client.yaml
SarahFrench: Sounds good, thanks for providing that link

SarahFrench (approved):
Approved, but here's a summary of the PR in case anyone looks back at it:

- `create_sample_integrations` added as a renamed version of `create_sample_workflows`
    - The API isn't updated yet; the `pre_create` code will allow the new, renamed version of the field to be matched to the relevant API field name.
    - Use of `pre_create` is necessary instead of `api_name` because of name collision between the deprecated and new versions of the field. [When the deprecated field is removed use of api_name would be preferred](https://github.com/GoogleCloudPlatform/magic-modules/pull/10478#pullrequestreview-2011231256), if its still necessary.
- [Tests confirm](https://github.com/GoogleCloudPlatform/magic-modules/pull/10478#issuecomment-2068482283) either field can be used currently
- [provision_gmek is deprecated](https://github.com/GoogleCloudPlatform/magic-modules/pull/10478#discussion_r1575946377) and not related to the create_sample_* fields above

PR #10481, opened 2024-04-18
Resource .go.tmpl conversion create method

NickElliot:
TODOs: custom code template inclusions, a method to verify if a property is present in returned `GetIdentity()`

```release-note:none

```


c2thorn (approved):
should be good after previous two PR's are merged

PR #10483, opened 2024-04-18
Run acceptance and unit tests when test fixtures change

trodge:
fixes https://github.com/hashicorp/terraform-provider-google/issues/17906

```release-note:none

```

### DIFF
--- a/.ci/magician/cmd/test_terraform_vcr.go
+++ b/.ci/magician/cmd/test_terraform_vcr.go
@@ -290,8 +290,12 @@ func modifiedPackages(newBranch, oldBranch string, rnr ExecRunner) (map[string]s
 	if err != nil {
 		return nil, false, err
 	}
+	return modifiedPackagesFromDiffs(strings.Split(diffs, "\n"))
+}
+
+func modifiedPackagesFromDiffs(diffs []string) (map[string]struct{}, bool, error) {
 	var goFiles []string
-	for _, line := range strings.Split(diffs, "\n") {
+	for _, line := range diffs {
 		if strings.HasSuffix(line, ".go") || strings.Contains(line, "test-fixtures") || line == "go.mod" || line == "go.sum" {
 			goFiles = append(goFiles, line)
 		}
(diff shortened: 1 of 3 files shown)


.ci/magician/cmd/test_terraform_vcr.go
```diff
@@ -292,7 +292,7 @@ func modifiedPackages(newBranch, oldBranch string, rnr ExecRunner) (map[string]s
 	}
 	var goFiles []string
```
melinath, line 293: Could you split the logic from here to the end of the function into a separate function and add thorough unit tests for it?

.ci/magician/cmd/test_terraform_vcr.go
```diff
@@ -292,7 +292,7 @@ func modifiedPackages(newBranch, oldBranch string, rnr ExecRunner) (map[string]s
...
 	for _, line := range strings.Split(diffs, "\n") {
-		if strings.HasSuffix(line, ".go") || line == "go.mod" || line == "go.sum" {
+		if strings.HasSuffix(line, ".go") || strings.Contains(line, "test-fixtures") || line == "go.mod" || line == "go.sum" {
```
melinath, line 295: I see that the bash version of this is matching against `go.mod$` - why is that? Should we use HasSuffix here?

melinath (changes requested):
makes sense - one refactor to be able to test better

.ci/magician/cmd/test_terraform_vcr.go
trodge: I don't think there's any particular reason other than that the other pattern ends in `$`. We don't have any other files that end with `go.mod` or `go.sum`.

trodge: Oh, never mind. I forgot the line contains the full path of the file.

.ci/magician/cmd/test_terraform_vcr_test.go
```diff
@@ -0,0 +1,54 @@
+package cmd
+
+import (
+	"reflect"
+	"testing"
+)
+
+func TestModifiedPackagesFromDiffs(t *testing.T) {
```
melinath, line 8: could you add `go.mod` and `go.sum` cases, and some other cases that shouldn't trigger tests (such as markdown changes, or hypothetical changes to other files in google-beta that wouldn't match?)

melinath (approved):

melinath (approved):

PR #10485, opened 2024-04-18
Made base transport layer force ipv4 usage if ipv6 is not supported

melinath:
Fixes https://github.com/hashicorp/terraform-provider-google/issues/6782

Workaround for https://github.com/golang/go/issues/25321

Note that while I was able to reproduce the Go issue, I was not able to reproduce the TF issue - there seems to be some amount of randomness in terms of when it presents. However, this seems to be the best option for working around the issue.

This PR also removes a lot of complexity that was previously implicitly part of our transport setup via [transport.NewHTTPClient](https://github.com/googleapis/google-api-go-client/blob/707b9b287dc37fb07c2553c0391f83995f1420a6/transport/http/dial.go#L32) and that was mostly unused. (I traced through the code to boil it down to the ~10 or so~ 6 lines we were actually using.)

yaqs/47302089738551296

```release-note:bug
provider: Forced HTTP requests to use ipv4 on systems with ipv6 disabled, such as Cloud Shell.
```

### DIFF
--- a/mmv1/third_party/terraform/transport/config.go.erb
+++ b/mmv1/third_party/terraform/transport/config.go.erb
@@ -7,6 +7,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"log"
+	"net"
 	"net/http"
 	"regexp"
 	"strconv"
@@ -28,6 +29,8 @@ import (
 
 	"github.com/hashicorp/terraform-provider-google/google/verify"
 
+	"golang.org/x/net/http2"
+	"golang.org/x/net/nettest"
 	"golang.org/x/oauth2"
 	"google.golang.org/grpc"
 	googleoauth "golang.org/x/oauth2/google"
@@ -82,6 +85,7 @@ import (
 	"google.golang.org/api/storage/v1"
 	"google.golang.org/api/storagetransfer/v1"
 	"google.golang.org/api/transport"
+	apihttp "google.golang.org/api/transport/http"
 )
 
 type ProviderMeta struct {
@@ -429,6 +433,39 @@ func SetEndpointDefaults(d *schema.ResourceData) error {
 	return nil
 }
 
+// baseTransport returns the base HTTP transport. It starts with the
+// default transport and makes some tweaks to match best practices
+// from google-api-go-client, as well as ensuring that IPv6 does
+// not get used in environments that don't support it.
+func baseTransport() (http.RoundTripper, error) {
+	trans := http.DefaultTransport.(*http.Transport).Clone()
+	// Increase MaxIdleConnsPerHost due to reported performance issues under load in the
+	// GCS client.
+	trans.MaxIdleConnsPerHost = 100
+
+	// Configure the ReadIdleTimeout HTTP/2 option for the
+	// transport. This allows broken idle connections to be pruned more quickly,
+	// preventing the client from attempting to re-use connections that will no
+	// longer work. http2Trans is discarded after configuration.
+	http2Trans, err := http2.ConfigureTransports(trans)
+	if err != nil {
+		return trans, err
+	}
+	http2Trans.ReadIdleTimeout = time.Second * 31
+
+	// Override dialer so that we don't try IPv6 if it's not supported.
+	// https://github.com/golang/go/issues/25321
+	trans.DialContext = func(ctx context.Context, network string, addr string) (net.Conn, error) {
+		d := &net.Dialer{}
+		if !nettest.SupportsIPv6() {
+			return d.DialContext(ctx, "tcp4", addr)
+		}
+		return d.DialContext(ctx, network, addr)
+	}
+
+	return trans, nil
+}
+
 func (c *Config) LoadAndValidate(ctx context.Context) error {
 	if len(c.Scopes) == 0 {
 		c.Scopes = DefaultClientScopes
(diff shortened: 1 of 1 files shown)


melinath:
Some real test runs, in case VCR bypasses this change: https://hashicorp.teamcity.com/buildConfiguration/TerraformProviders_GoogleCloud_GOOGLE_MMUPSTREAMTESTS_GOOGLE_PACKAGE_PUBSUB/135875?hideTestsFromDependencies=false&hideProblemsFromDependencies=false&expandBuildChangesSection=true

rileykarson:
Adding @ScottSuarez as a mandatory reviewer- we should ensure that MTLS is not broken as a result of this change.

mmv1/third_party/terraform/transport/config.go.erb
```diff
@@ -429,6 +433,39 @@ func SetEndpointDefaults(d *schema.ResourceData) error {
...
 
+// baseTransport returns the base HTTP transport. It starts with the
+// default transport and makes some tweaks to match best practices
+// from google-api-go-client, as well as ensuring that IPv6 does
+// not get used in environments that don't support it.
+func baseTransport() (http.RoundTripper, error) {
+	trans := http.DefaultTransport.(*http.Transport).Clone()
+	// Increase MaxIdleConnsPerHost due to reported performance issues under load in the
+	// GCS client.
+	trans.MaxIdleConnsPerHost = 100
```
c2thorn, line 444: If I'm understanding correctly, these tweaks are from [here](https://github.com/googleapis/google-api-go-client/blob/707b9b287dc37fb07c2553c0391f83995f1420a6/transport/http/dial.go#L174)?

mmv1/third_party/terraform/transport/config.go.erb
```diff
@@ -429,6 +433,39 @@ func SetEndpointDefaults(d *schema.ResourceData) error {
 	return nil
 }
 
+// baseTransport returns the base HTTP transport. It starts with the
+// default transport and makes some tweaks to match best practices
+// from google-api-go-client, as well as ensuring that IPv6 does
+// not get used in environments that don't support it.
+func baseTransport() (http.RoundTripper, error) {
```
c2thorn, line 440: I may be missing something here, but why can't we reuse the google-api-go-client and override the transport's dial context after?

c2thorn:
Do we want a provider-wide changelog note?

mmv1/third_party/terraform/transport/config.go.erb
melinath: http.Client's Transport is a RoundTripper rather than a Transport, so we would need to cast it back to a Transport in order to do that. I don't see a reason that wouldn't work, and that would've been a lot easier to implement.

I think, having dug through the code, I would slightly prefer this way because we know exactly what we're getting and don't have to worry about the intricacies of NewClient.

mmv1/third_party/terraform/transport/config.go.erb
c2thorn: Great, you did mention it in the comment, but I'll leave the link here just for posterity

mmv1/third_party/terraform/transport/config.go.erb
c2thorn: I think that's reasonable now that we've gone through the code. My only concerns are about missing something, but I've gone through it myself now and everything seems in order.

c2thorn (approved):
LGTM for non-MTLS stuff

c2thorn:
> > Do we want a provider-wide changelog note?
> 
> added

We are expecting some user-facing impact right? Specifically user's facing https://github.com/hashicorp/terraform-provider-google/issues/6782 will now have their ipv4 settings honored?

melinath:
> > > Do we want a provider-wide changelog note?
> > 
> > 
> > added
> 
> We are expecting some user-facing impact right? Specifically user's facing [hashicorp/terraform-provider-google#6782](https://github.com/hashicorp/terraform-provider-google/issues/6782) will now have their ipv4 settings honored?

🤦 yep. yep. we sure are.

ScottSuarez (commented):
As is, this pr will likely break mtls.

Mtls endpoint switch is detected here:
https://github.com/modular-magician/terraform-provider-google-beta/blob/5c35eba9eb80027ba7ca98ebc3dddbf3229500af/google-beta/provider/mtls_util.go#L19

https://github.com/googleapis/google-api-go-client/blob/81ccc7e5305f4913388c254b39144c9a144abb20/internal/cba.go#L289
https://github.com/google/s2a-go/blob/fcfd1dacc64be11cc3532ab6a7e1cdd1a3e4b69e/s2a.go#L361

The long term solution was the switch to the DCL (at the time).

They have a testing playbook for mtls here:
https://docs.google.com/document/d/1MeAoY-XQ2O1DmP_PZDYWS1PFx9vkZ4qWwnhSV0_kfPU/edit?tab=t.0#heading=h.r7dggrs9pw6r

rileykarson (changes requested):
Marking request changes so that we can't click the big green button

melinath:
Unfortunately the "just override the transport's dialcontext" approach is not feasible because the base transport is wrapped a few times and two of them use private fields to store the wrapped RoundTripper.

melinath:
yaqs/5034757202074664960

melinath:
I'm going to attempt to get this change made in google-api-go-client

melinath:
Opened https://github.com/googleapis/google-api-go-client/pull/2549

rileykarson:

PR #10486, opened 2024-04-18
Convert custom_expand and custom_flatten

zli82016:
Convert custom_expand and custom_flatten using the [script](https://github.com/GoogleCloudPlatform/magic-modules/compare/main...zli82016:magic-modules:go-rewrite-example-templates-script) 

1. Set `Prefix` as a field of `Type` struct instead of passing as a parameter of functions, so it will not need to pass in prefix when calling the functions (such as `build_expand_method`) or embed templates (it depends on the Go rewriting later) [here](https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/templates/terraform/resource.erb#L1165) 
2.  Add the function `wrapMultipleParams` to pass more parameters to the templates

```release-note:none

```


mmv1/provider/template_data.go
```diff
@@ -45,6 +46,23 @@ type TemplateData struct {
 	//     attr_accessor :env
 }
 
+// Build a map(map[string]interface{}) from a list of paramerter
+// The format of passed in parmeters are key1, value1, key2, value2 ...
+func wrapMultipleParams(params ...interface{}) (map[string]interface{}, error) {
```
c2thorn, line 51: helpful! I might be able to reuse this in my previous templates

c2thorn (approved):

c2thorn:
LGTM after rebase

zli82016:
cc @NickElliot

PR #10495, opened 2024-04-19
Copy TeamCity debug logs to GCS

shuyama1:

```release-note:none

```

### DIFF
--- a/mmv1/third_party/terraform/.teamcity/components/builds/build_steps.kt
+++ b/mmv1/third_party/terraform/.teamcity/components/builds/build_steps.kt
@@ -146,18 +146,20 @@ fun BuildSteps.saveArtifactsToGCS() {
             gcloud auth activate-service-account --key-file=google-account.json
 
             # Get current date for nightly tests
-            current_date=$(date +"%%Y-%%m-%%d") 
+            CURRENT_DATE=$(date +"%%Y-%%m-%%d") 
             // "%%" is used to escape "%" see details at https://www.jetbrains.com/help/teamcity/9.0/defining-and-using-build-parameters-in-build-configuration.html#using-build-parameters-in-build-configuration-settings
+
+            echo "Print out CURRENT_DATE: ${'$'}{CURRENT_DATE}"
 
             # Detect Trigger Method 
             TRIGGERED_BY_USERNAME=%teamcity.build.triggeredBy.username%
             BRANCH_NAME=%teamcity.build.branch%
             if [[ "${'$'}TRIGGERED_BY_USERNAME" = "n/a" ]] ; then
                 echo "Build was triggered as part of automated testing. We know this because the `triggeredBy.username` value was `n/a`, value: ${'$'}{TRIGGERED_BY_USERNAME}"
-                FOLDER="nightly/%teamcity.project.id%/${'$'}current_date"
+                FOLDER="nightly/%teamcity.project.id%/${'$'}{CURRENT_DATE}"
             else
                 echo "Build was triggered manually. We know this because `triggeredBy.username` has a non- `n/a` value: ${'$'}{TRIGGERED_BY_USERNAME}"
-                FOLDER="manual/%teamcity.project.id%/${'$'}BRANCH_NAME/${'$'}current_date"
+                FOLDER="manual/%teamcity.project.id%/${'$'}{BRANCH_NAME}/${'$'}{CURRENT_DATE}"
             fi
 
             # Copy logs to GCS
--- a/mmv1/third_party/terraform/.teamcity/components/builds/build_steps.kt
+++ b/mmv1/third_party/terraform/.teamcity/components/builds/build_steps.kt
@@ -155,10 +155,10 @@ fun BuildSteps.saveArtifactsToGCS() {
             TRIGGERED_BY_USERNAME=%teamcity.build.triggeredBy.username%
             BRANCH_NAME=%teamcity.build.branch%
             if [[ "${'$'}TRIGGERED_BY_USERNAME" = "n/a" ]] ; then
-                echo "Build was triggered as part of automated testing. We know this because the `triggeredBy.username` value was `n/a`, value: ${'$'}{TRIGGERED_BY_USERNAME}"
+                echo "Build was triggered as part of automated testing. We know this because the \`triggeredBy.username\` value was \`n/a\`, value: ${'$'}{TRIGGERED_BY_USERNAME}"
                 FOLDER="nightly/%teamcity.project.id%/${'$'}{CURRENT_DATE}"
             else
-                echo "Build was triggered manually. We know this because `triggeredBy.username` has a non- `n/a` value: ${'$'}{TRIGGERED_BY_USERNAME}"
+                echo "Build was triggered manually. We know this because \`triggeredBy.username\` has a non- \`n/a\` value: ${'$'}{TRIGGERED_BY_USERNAME}"
                 FOLDER="manual/%teamcity.project.id%/${'$'}{BRANCH_NAME}/${'$'}{CURRENT_DATE}"
             fi
 
--- a/mmv1/third_party/terraform/.teamcity/components/builds/build_steps.kt
+++ b/mmv1/third_party/terraform/.teamcity/components/builds/build_steps.kt
@@ -146,7 +146,7 @@ fun BuildSteps.saveArtifactsToGCS() {
             gcloud auth activate-service-account --key-file=google-account.json
 
             # Get current date
-            current_date=$(date+"%Y-%m-%d")
+            current_date=$(date +"%%Y-%%m-%%d")
 
             # Detect Trigger Method 
             TRIGGERED_BY_USERNAME=%teamcity.build.triggeredBy.username%
--- a/mmv1/third_party/terraform/.teamcity/components/builds/build_steps.kt
+++ b/mmv1/third_party/terraform/.teamcity/components/builds/build_steps.kt
@@ -45,11 +45,11 @@ fun BuildSteps.tagBuildToIndicateTriggerMethod() {
             TRIGGERED_BY_USERNAME=%teamcity.build.triggeredBy.username%
 
             if [[ "${'$'}TRIGGERED_BY_USERNAME" = "n/a" ]] ; then
-                echo "Build was triggered as part of automated testing. We know this because the `triggeredBy.username` value was `n/a`, value: ${'$'}{TRIGGERED_BY_USERNAME}"
+                echo "Build was triggered as part of automated testing. We know this because the \`triggeredBy.username\` value was \`n/a\`, value: ${'$'}{TRIGGERED_BY_USERNAME}"
                 TAG="cron-trigger"
                 echo "##teamcity[addBuildTag '${'$'}{TAG}']"
             else
-                echo "Build was triggered manually. We know this because `triggeredBy.username` has a non- `n/a` value: ${'$'}{TRIGGERED_BY_USERNAME}"
+                echo "Build was triggered manually. We know this because \`triggeredBy.username\` has a non- \`n/a\` value: ${'$'}{TRIGGERED_BY_USERNAME}"
                 TAG="manual-trigger"
                 echo "##teamcity[addBuildTag '${'$'}{TAG}']"
             fi
@@ -149,8 +149,6 @@ fun BuildSteps.saveArtifactsToGCS() {
             CURRENT_DATE=$(date +"%%Y-%%m-%%d") 
             // "%%" is used to escape "%" see details at https://www.jetbrains.com/help/teamcity/9.0/defining-and-using-build-parameters-in-build-configuration.html#using-build-parameters-in-build-configuration-settings
 
-            echo "Print out CURRENT_DATE: ${'$'}{CURRENT_DATE}"
-
             # Detect Trigger Method 
             TRIGGERED_BY_USERNAME=%teamcity.build.triggeredBy.username%
             BRANCH_NAME=%teamcity.build.branch%
@@ -159,7 +157,7 @@ fun BuildSteps.saveArtifactsToGCS() {
                 FOLDER="nightly/%teamcity.project.id%/${'$'}{CURRENT_DATE}"
             else
                 echo "Build was triggered manually. We know this because \`triggeredBy.username\` has a non- \`n/a\` value: ${'$'}{TRIGGERED_BY_USERNAME}"
-                FOLDER="manual/%teamcity.project.id%/${'$'}{BRANCH_NAME}/${'$'}{CURRENT_DATE}"
+                FOLDER="manual/%teamcity.project.id%/${'$'}{BRANCH_NAME}"
             fi
 
             # Copy logs to GCS
--- a/mmv1/third_party/terraform/.teamcity/components/builds/build_steps.kt
+++ b/mmv1/third_party/terraform/.teamcity/components/builds/build_steps.kt
@@ -132,3 +132,43 @@ fun BuildSteps.runAcceptanceTests() {
         })
     }
 }
+
+fun BuildSteps.saveArtifactsToGCS() {
+    step(ScriptBuildStep {
+        name = "Tasks after running nightly tests: push artifacts(debug logs) to GCS"
+        scriptContent = """
+            #!/bin/bash
+            echo "Post-test step - storge artifacts(debug logs) to GCS"
+
+            # Authenticate gcloud CLI
+            echo "${'$'}{GOOGLE_CREDENTIALS}" > google-account.json
+            chmod 600 google-account.json
+            gcloud auth activate-service-account --key-file=google-account.json
+
+            # Detect Trigger Method 
+            TRIGGERED_BY_USERNAME=%teamcity.build.triggeredBy.username%
+            BRANCH_NAME=%teamcity.build.branch%
+            if [[ "${'$'}TRIGGERED_BY_USERNAME" = "n/a" ]] ; then
+                echo "Build was triggered as part of automated testing. We know this because the `triggeredBy.username` value was `n/a`, value: ${'$'}{TRIGGERED_BY_USERNAME}"
+                FOLDER="nightly"
+            else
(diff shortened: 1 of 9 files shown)


mmv1/third_party/terraform/.teamcity/components/builds/build_steps.kt
```diff
@@ -131,4 +131,44 @@ fun BuildSteps.runAcceptanceTests() {
...
+                echo "Build was triggered as part of automated testing. We know this because the `triggeredBy.username` value was `n/a`, value: ${'$'}{TRIGGERED_BY_USERNAME}"
+                FOLDER="nightly"
+            else
+                echo "Build was triggered manually. We know this because `triggeredBy.username` has a non- `n/a` value: ${'$'}{TRIGGERED_BY_USERNAME}"
+                FOLDER="manual/${'$'}BRANCH_NAME"
+            fi
+
+
+            # Copy logs to GCS
+            gsutil -m cp %teamcity.build.checkoutDir%/debug* gs://teamcity-nightly-logs/%PROVIDER_NAME%/${'$'}{FOLDER}/%env.BUILD_NUMBER%/
```
SarahFrench, line 161: I wanted to check if files would have unique paths if someone ran tests manually on the same commit, and it looks like they wouldn't be unique.

[We set env.BUILD_NUMBER to the short commit hash](https://github.com/hashicorp/terraform-provider-google/blob/82eda93fed4a662be30ee6dc56e507322539e8a7/.teamcity/components/builds/build_steps.kt#L25-L38) (this is copied from how the old TeamCity projects worked) so this value isn't unique per build. For example, [the first artifact listed here](https://hashicorp.teamcity.com/buildConfiguration/TerraformProviders_GoogleCloud_GOOGLE_NIGHTLYTESTS_GOOGLE_PACKAGE_ACCESSAPPROVAL/138150?buildTab=artifacts) would have the URL `gs://teamcity-nightly-logs/google/nightly/2061d55/debug-google-2061d55-TestAccAccessApprovalSettings__folder.txt`. If that commit was the tip of main branch after the last PR is merged on a Friday evening it'll be the same commit used for Saturday, Sunday and Monday's early-morning tests and the bucket would only contain the Monday's artifact.

I took a look at the `build.counter` configuration parameter as an alternative. It looks like this value increments for a given build configuration, i.e. there is a past build of bigquery tests that has counter=100, and there'll be a past build of the pubsub tests that has counter=100. This could be used to make the GCS objects unique, either by extending the prefix or by changing [the artifact's filename](https://github.com/hashicorp/terraform-provider-google/blob/82eda93fed4a662be30ee6dc56e507322539e8a7/.teamcity/components/builds/build_parameters.kt#L247) to use `%teamcity.build.counter%` instead of `%env.BUILD_NUMBER%`

As a real world example, here are some past nightly tests for AlloyDB:
- [Saturday April 13th](https://hashicorp.teamcity.com/buildConfiguration/TerraformProviders_GoogleCloud_GOOGLE_NIGHTLYTESTS_GOOGLE_PACKAGE_ALLOYDB/132693?expandBuildDeploymentsSection=false&hideTestsFromDependencies=false&hideProblemsFromDependencies=false&expandBuildChangesSection=true), build number = 13a7e31, build count = 199
- [Sunday April 14th](https://hashicorp.teamcity.com/buildConfiguration/TerraformProviders_GoogleCloud_GOOGLE_NIGHTLYTESTS_GOOGLE_PACKAGE_ALLOYDB/133249?expandBuildDeploymentsSection=false&hideTestsFromDependencies=false&hideProblemsFromDependencies=false&expandBuildTestsSection=true), build number = 13a7e31, build count = 200
- [Monday April 15th](https://hashicorp.teamcity.com/buildConfiguration/TerraformProviders_GoogleCloud_GOOGLE_NIGHTLYTESTS_GOOGLE_PACKAGE_ALLOYDB/133922?expandBuildDeploymentsSection=false&hideTestsFromDependencies=false&hideProblemsFromDependencies=false), build number = 13a7e31, build count = 201

SarahFrench: That comment was a bit of a ramble, but to distill it:

- This problem would be encountered by every test every weekend, so needs addressing
- Solutions are to either change the prefix of the GCS object or change the original artifact's filename
    - the filenames of the artifacts can be changed without breaking anything, so I'd recommend that option

mmv1/third_party/terraform/.teamcity/components/builds/build_steps.kt
```diff
@@ -131,4 +131,44 @@ fun BuildSteps.runAcceptanceTests() {
...
+            if [[ "${'$'}TRIGGERED_BY_USERNAME" = "n/a" ]] ; then
+                echo "Build was triggered as part of automated testing. We know this because the `triggeredBy.username` value was `n/a`, value: ${'$'}{TRIGGERED_BY_USERNAME}"
+                FOLDER="nightly"
+            else
+                echo "Build was triggered manually. We know this because `triggeredBy.username` has a non- `n/a` value: ${'$'}{TRIGGERED_BY_USERNAME}"
+                FOLDER="manual/${'$'}BRANCH_NAME"
+            fi
+
+
+            # Copy logs to GCS
```
SarahFrench, line 160: Am I right assuming that the `teamcity-nightly-logs` bucket is intended to contain all the logs from cron- and manually-triggered tests that use the main branch of the downstream provider repos only? I.e. any tests run against the MM upstream forks using the VCR projects will be excluded?

If that's correct then we'd need to add some logic to make this build step be conditionally included only in the nightly projects.

SarahFrench (commented):
Left some comments - happy to discuss here async or in a chat in a more synchronous way!

mmv1/third_party/terraform/.teamcity/components/builds/build_steps.kt
SarahFrench: Some context: the code in `.teamcity/components/builds/build_configuration_per_package.kt` is used to create the build configurations in the `Upstream MM Testing` and `Nightly Tests` sub projects within Google and Google Beta projects in TeamCity. Inclusion of the new `saveArtifactsToGCS()` function would either need to be made conditional or the code restructured somehow

SarahFrench:
In terms of testing this PR, I'd recommend running the [make test](https://github.com/shuyama1/magic-modules/blob/6c99b6fc2b9c1beffabee01edb488efa15fbfb62/mmv1/third_party/terraform/.teamcity/Makefile#L6) target in the TeamCity Makefile. That'll generate files in a gitignored `.teamcity/target` folder that you can inspect for some initial 'manual' testing. Another approach is we can make a new project in TeamCity to pull config from [the branch in this PR's diff report](https://github.com/GoogleCloudPlatform/magic-modules/pull/10495#issuecomment-2067093159).

I can do the latter for you and/or we can pair on a call to do it. Let me know!

mmv1/third_party/terraform/.teamcity/components/builds/build_steps.kt
```diff
@@ -131,4 +131,44 @@ fun BuildSteps.runAcceptanceTests() {
...
+
+fun BuildSteps.saveArtifactsToGCS() {
+    step(ScriptBuildStep {
+        name = "Tasks after running nightly tests: push artifacts(debug logs) to GCS"
+        scriptContent = """
+            #!/bin/bash
+            echo "Post-test step - storge artifacts(debug logs) to GCS"
+
+            # Authenticate gcloud CLI
+            echo "${'$'}{GOOGLE_CREDENTIALS}" > google-account.json
```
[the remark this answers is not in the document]
shuyama1: Good point! I've created a dedicated service account specifically for this purpose.

mmv1/third_party/terraform/.teamcity/components/builds/build_steps.kt
shuyama1: Good catch! I think we could add `counter` to the artifact's filename. I also want to keep the build number in the filename as it'd clear to tell which commit the test is at.

I wonder if we can get the time/date when the build is triggered by %teamcity.build.startDate%, as I was thinking this might be useful to group logs in (cron-triggered) nightly test.

mmv1/third_party/terraform/.teamcity/components/builds/build_steps.kt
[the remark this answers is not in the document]
SarahFrench: In that case we can add a new ENV to hold that value, and replace use of `GOOGLE_CREDENTIALS` here.

The new credentials would be entered as a context parameter [on this page](https://hashicorp.teamcity.com/admin/editProject.html?projectId=TerraformProviders_GoogleCloud&tab=versionedSettings&subTab=contextParams) - I can help with this when we get to manual testing of these config changes!

There would need to be code to make that value be used in the config code. This file ([.teamcity/settings.kts](https://github.com/hashicorp/magic-modules/blob/main/mmv1/third_party/terraform/.teamcity/settings.kts)) is the entry point where the context parameters are obtained and passed into the config. There would need to be an update to this class ([AllContextParameters](https://github.com/hashicorp/magic-modules/blob/6681336af2f445057526fd050033546ee6b87502/mmv1/third_party/terraform/.teamcity/components/builds/build_parameters.kt#L22)) to enable passing that new value in. Look for the VCR specific values already handled in that class! There would also need to be some changes to ensure that the context parameters passed to the different build configs include the new value.

Finally, [this code](https://github.com/hashicorp/magic-modules/blob/6681336af2f445057526fd050033546ee6b87502/mmv1/third_party/terraform/.teamcity/components/builds/build_parameters.kt#L226-L235) is where the context parameter values are used to set VCR-specific ENVs on builds. For whatever the name of the new ENV is, e.g. `VCR_BUCKET_CREDENTIALS` there'd need to be a new line setting that ENV's value.

mmv1/third_party/terraform/.teamcity/components/builds/build_steps.kt
SarahFrench: Re: `%teamcity.build.startDate%` , I can't see that value available when I look at [the parameters etc for a random build](https://hashicorp.teamcity.com/buildConfiguration/TerraformProviders_GoogleCloud_GOOGLE_NIGHTLYTESTS_GOOGLE_PACKAGE_ACCESSAPPROVAL/139161?buildTab=buildParameters&buildParametersTab=parameters&buildParametersShowDependencies=false). When I google about it I see [this JetBrains issue thats >10 years old](https://youtrack.jetbrains.com/issue/TW-4080/Predefined-properties-to-denote-build-start-date-and-time) 🤯 requesting the ability to access the start times for builds. The existing workaround is to install a plugin which I don't think is an option for us.


Aside: We can set values on builds ourselves using [service messages](https://www.jetbrains.com/help/teamcity/service-messages.html), and this is how the [build number is changed to the git commit](https://github.com/hashicorp/magic-modules/blob/6681336af2f445057526fd050033546ee6b87502/mmv1/third_party/terraform/.teamcity/components/builds/build_steps.kt#L32) and how we [tag builds](https://github.com/hashicorp/magic-modules/blob/6681336af2f445057526fd050033546ee6b87502/mmv1/third_party/terraform/.teamcity/components/builds/build_steps.kt#L50). You could make a build's first action be to tag itself with the current time, but that would only reflect when the build's user-defined steps started running and not when the build overall started running. Thought I'd share that option if it's still useful that way!

SarahFrench: >Good catch! I think we could add counter to the artifact's filename. I also want to keep the build number in the filename as it'd clear to tell which commit the test is at.

This sounds like a good idea!

mmv1/third_party/terraform/.teamcity/components/builds/build_steps.kt
SarahFrench: > My original plan was to only store the logs from the cron-triggered nightly tests.

That could be achieved by inspecting [if a build is tagged to indicate it's cron triggered](https://github.com/hashicorp/magic-modules/blob/6681336af2f445057526fd050033546ee6b87502/mmv1/third_party/terraform/.teamcity/components/builds/build_steps.kt#L47-L51)? If that was done then you could have the build step present across `Upstream MM Testing` and `Nightly Tests` but it only performs the upload step when builds are cron triggered (and no non-sweepers are triggered by CRON in `Upstream MM Testing`). Conditionally skipping steps is [already present in the code base](https://github.com/hashicorp/magic-modules/blob/6681336af2f445057526fd050033546ee6b87502/mmv1/third_party/terraform/.teamcity/components/builds/build_steps.kt#L117-L127) in [a few places](https://github.com/hashicorp/magic-modules/blob/6681336af2f445057526fd050033546ee6b87502/mmv1/third_party/terraform/.teamcity/components/builds/vcr_build_steps.kt#L100-L103), so there is precedent, but I'd argue that any conditional behaviour has to be made explicit!

SarahFrench: > Now I see that I might not set up the GCS path correctly to differentiate between these various projects/purposes, I'll modify it to make sure that they'll be stored in dedicated locations 

Yeah, there's a bit of a trap of figuring what the project boundaries mean! How you've included the project id in the prefix in 0eeab5bc34cc9a2385f72cf250017ddf3000b0d3 looks good - the ID won't change and indicates the provider and purpose pretty well. Purpose gets a bit muddy when there are cron versus manual triggered builds in the nightly test projects though 😬 

>Perhaps renaming the bucket to something like teamcity-logs would be less confusing. What do you think?

I think that change would be good - especially if MM Upstream builds will end up storing logs there. The project id in the object prefix will help to distinguish nightly tests from MM Upstream stuff, if you decide to include them

mmv1/third_party/terraform/.teamcity/components/builds/build_steps.kt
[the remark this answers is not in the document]
shuyama1: oh, I see. It's actually a bit more complicated than I thought 😅 I thought I only needed to create a new ENV in the TC UI. I'll update the code accordingly. Thanks for all the pointers!

mmv1/third_party/terraform/.teamcity/components/builds/build_steps.kt
shuyama1: > Re: `%teamcity.build.startDate%` , I can't see that value available when I look at [the parameters etc for a random build](https://hashicorp.teamcity.com/buildConfiguration/TerraformProviders_GoogleCloud_GOOGLE_NIGHTLYTESTS_GOOGLE_PACKAGE_ACCESSAPPROVAL/139161?buildTab=buildParameters&buildParametersTab=parameters&buildParametersShowDependencies=false). When I google about it I see [this JetBrains issue thats >10 years old](https://youtrack.jetbrains.com/issue/TW-4080/Predefined-properties-to-denote-build-start-date-and-time) 🤯 requesting the ability to access the start times for builds. The existing workaround is to install a plugin which I don't think is an option for us.
> 
> Aside: We can set values on builds ourselves using [service messages](https://www.jetbrains.com/help/teamcity/service-messages.html), and this is how the [build number is changed to the git commit](https://github.com/hashicorp/magic-modules/blob/6681336af2f445057526fd050033546ee6b87502/mmv1/third_party/terraform/.teamcity/components/builds/build_steps.kt#L32) and how we [tag builds](https://github.com/hashicorp/magic-modules/blob/6681336af2f445057526fd050033546ee6b87502/mmv1/third_party/terraform/.teamcity/components/builds/build_steps.kt#L50). You could make a build's first action be to tag itself with the current time, but that would only reflect when the build's user-defined steps started running and not when the build overall started running. Thought I'd share that option if it's still useful that way!

yeah, I also realized that the build parameters page does not contain any date related info unfortunately 😢 . I wonder if we can just simply get the run date by `current_date=$(date+"%Y-%m-%d")` in the bash script and use that value. It'd be nice if we can use date to group nightly test logs and show that in the storage path, as it'd be clear to viewers, I think.

SarahFrench: Ah if you're just wanting the date then that's super easy - your suggestion will work. Just bear in mind that all agents etc use UTC!

mmv1/third_party/terraform/.teamcity/components/builds/build_triggers.kt
```diff
@@ -26,7 +26,8 @@ class NightlyTriggerConfiguration(
 fun Triggers.runNightly(config: NightlyTriggerConfiguration) {
 
     schedule{
-        enabled = config.nightlyTestsEnabled
+        // enabled = config.nightlyTestsEnabled
+        enabled = false
```
shuyama1, line 30: Will revert this change before merging the PR

shuyama1:
@SarahFrench This PR should be ready for the final review. I'll share more test results via Chat. Thanks!

melinath (approved):
Testing GitHub approval behavior - please ignore.

melinath (commented):
Testing GitHub approval behavior - please ignore.

melinath (changes requested):
Testing GitHub approval behavior - please ignore.

SarahFrench (approved):
Looks good to me!

The only remaining part I'd like to see is some kind of automated test that can protect against this step being removed from the builds (i.e. avoiding L74 of build_configuration_per_package.kt being deleted) and us only realising once logs stop being copied to GCS.

One way to do that would be to have tests like those already in [terraform/.teamcity/tests/build_configuration_features.kt](https://github.com/GoogleCloudPlatform/magic-modules/blob/0ae36a9440f5800d17baaf3aae118f181a28f660/mmv1/third_party/terraform/.teamcity/tests/build_configuration_features.kt) where we check all build configurations for X thing. However I'm not sure if the [BuildType class](https://teamcity.jetbrains.com/app/dsl-documentation/root/build-type/index.html) will make steps accessible to make assertions on. I'm happy to look into this myself if that helps?

An alternative could be to have alerting on the GCS bucket that triggers if no new objects are written to the bucket in a certain timeframe. What do you think?


---

Besides that I can see the main project is ready for this to be merged, but you need to:
- re-enable the CRON triggers in this PR
- test TeamCity after this PR is merged to make sure the release cut isn't affected by unexpected issues!

SarahFrench (approved):
Thanks for the test! I'm happy for this PR to be merged (after CRON trigger is enabled again).

PR #10497, opened 2024-04-19
Provision overlapping internal range in the correct order

rileykarson:
This succeeded in VCR, but failed in nightly when it provisioned the range first. Make sure it happens in order.

```release-note:none
```


slevenick (commented):
Does this not get implied by:
`  network = google_compute_network.default.id`?

rileykarson:
Nope- this seems to only work correctly if the range is already reserved by the _subnetwork_, not just if the network exists. The current graph is:

google_compute_network
|
google_compute_subnetwork | google_network_connectivity_internal_range

Which means the subnetwork and range can be created in any arbitrary order after the network. We need:

google_compute_network
|
google_compute_subnetwork
|
google_network_connectivity_internal_range

Which guarantees the subnetwork is created first.

slevenick (approved):
Ah, missed that it adds the dependency on the subnet!

PR #10498, opened 2024-04-19
Update alloydb sweeper to sweep secondary clusters

rileykarson:
We currently sweep us-central1, which only deletes primary clusters. However, primary clusters can only be deleted if their secondary is deleted. This won't guarantee we sweep completely in one pass, but will mean that we clean up any primary+secondary pairs in two passes.

Note that there a force parameter on delete- that deletes instances, which are children of clusters, and does not appear to delete secondary clusters.

I ran this locally to test it, with: `GOOGLE_USE_DEFAULT_CREDENTIALS=true GOOGLE_PROJECT=ci-test-project-nightly-ga make testacc TEST=./google-beta/sweeper TESTARGS='-sweep=us-central1 -sweep-run=AlloydbCluster'`

Here's the results: https://console.cloud.google.com/alloydb/clusters?e=-13802955&mods=logs_tg_staging&project=ci-test-project-nightly-ga

```release-note:none
```

### DIFF
--- a/mmv1/third_party/terraform/services/alloydb/resource_alloydb_cluster_sweeper.go.erb
+++ b/mmv1/third_party/terraform/services/alloydb/resource_alloydb_cluster_sweeper.go.erb
@@ -48,10 +48,11 @@ func testSweepAlloydbCluster(region string) error {
 		},
 	}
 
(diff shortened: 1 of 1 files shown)


rileykarson:
TGC failures are stale, an older commit failed to build. Not sure what unsticks them.

roaks3 (approved):

PR #10501, opened 2024-04-19
Added item_validation property to handle validation on array elements

melinath:
Fixed https://github.com/hashicorp/terraform-provider-google/issues/17918

Also simplified logic around whether to render validation functions. yaqs/6477316457716776960

Note: the removals of DiffSuppressFuncs are on read-only resource refs (so removing the diff suppress should be a no-op.) The ValidateBase64String addition is for testing & needs to be removed before merge.

No release note - generator-only change.

```release-note:none

```

### DIFF
--- a/docs/content/develop/field-reference.md
+++ b/docs/content/develop/field-reference.md
@@ -250,6 +250,9 @@ This property has two mutually exclusive child properties:
   String fields. It is equivalent to
   [`function: verify.ValidateRegexp(REGEX_STRING)`](https://github.com/hashicorp/terraform-provider-google-beta/blob/0ef51142a4dd1c1a4fc308c1eb09dce307ebe5f5/google-beta/verify/validation.go#L425).
 
+`validation` is not supported for Array fields (including sets); however, individual
+elements in the array can be validated using [`item_validation`]({{<ref "/develop/field-reference#item_validation"}}).
+
 Example: Provider-specific function
 
 ```yaml
@@ -334,6 +337,10 @@ item_type: !ruby/object:Api::Type::NestedObject
       description: |
         MULTI_LINE_FIELD_DESCRIPTION
 ```
+
+### `item_validation`
+Array only. Controls the [`ValidateFunc`](https://developer.hashicorp.com/terraform/plugin/sdkv2/schemas/schema-behaviors#validatefunc)
+used to validate individual items in the array. Behaves like [`validation`]({{<ref "/develop/field-reference#validation" >}}).
 
 ## `NestedObject` properties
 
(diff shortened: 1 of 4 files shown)


docs/content/develop/field-reference.md
```diff
@@ -335,6 +338,30 @@ item_type: !ruby/object:Api::Type::NestedObject
         MULTI_LINE_FIELD_DESCRIPTION
 ```
 
+### `item_validation`
+Array only. Controls the [`ValidateFunc`](https://developer.hashicorp.com/terraform/plugin/sdkv2/schemas/schema-behaviors#validatefunc)
```
rileykarson, line 342: We should add something like:

> For arrays of enums, this will override the enum validation. If you need additional validation on top of an enum, ensure that the supplied validation func also verifies the enum values are correct.

mmv1/products/pubsub/Topic.yaml
```diff
@@ -119,6 +119,8 @@ properties:
           allowed regions. An empty list means that no regions are allowed,
           and is not a valid configuration.
         item_type: Api::Type::String
+        item_validation: !ruby/object:Provider::Terraform::Validation
```
rileykarson, line 122: Dismiss once removed

mmv1/templates/terraform/schema_property.erb
```diff
@@ -111,20 +112,26 @@
...
-        Type: <%= tf_types[property.item_type_class] -%>,
-      },
-  <% elsif property.item_type.is_a?(Api::Type::Enum) -%>
-      Elem: &schema.Schema{
-        Type: <%= tf_types[property.item_type.class] -%>,
-        ValidateFunc: verify.ValidateEnum([]string{"<%= property.item_type.values.join '","' -%>"}),
-      },
   <% else # array of basic types -%>
       Elem: &schema.Schema{
+    <% if property.item_type.is_a?(String) # Basic type like Api::Type::String -%>
```
rileykarson, line 117: Am I missing something here, or are both results in this if/else block identical?

melinath: they're different - the first one is `item_type_class` and the second one is `item_type.class` (with a period between type and class instead of an underscore).

rileykarson: There it is, thanks!

docs/content/develop/field-reference.md
melinath: done

mmv1/products/pubsub/Topic.yaml
melinath: done

rileykarson (approved):

rileykarson:
@NickElliot fyi, we'll want to replicate this logic in Go

PR #10502, opened 2024-04-19
Lower timeout for dangling cluster creation to 10s

trodge:
part of https://github.com/hashicorp/terraform-provider-google/issues/17196

```release-note:none

```


melinath (commented):
could you elaborate on why this resolves the issue?

trodge:
> could you elaborate on why this resolves the issue?

I suspect that this will force the error message to always be the one the test predicts, instead of the one that is sometimes showing up.

melinath (changes requested):
Since forcing a timeout is the point, could we just make the create timeout 1s instead of 10s?

melinath (approved):
🤞

PR #10506, opened 2024-04-20
chore(ci): return jsonify breaking change list

iyabchen:
Return json output for breaking change tool output. 
 https://github.com/hashicorp/terraform-provider-google/issues/17635

```release-note:none

```

### DIFF
--- a/tools/diff-processor/rules/breaking_changes_test.go
+++ b/tools/diff-processor/rules/breaking_changes_test.go
@@ -289,7 +289,6 @@ func TestComputeBreakingChanges(t *testing.T) {
 					},
 				},
 			},
-			wantViolationCount: 1,
 			wantViolations: []*BreakingChange{
 				{
 					Resource:               "google-x",
@@ -340,7 +339,6 @@ func TestComputeBreakingChanges(t *testing.T) {
 					},
 				},
 			},
-			wantViolationCount: 1,
 			wantViolations: []*BreakingChange{
 				{
 					Resource:               "google-x",
@@ -391,7 +389,6 @@ func TestComputeBreakingChanges(t *testing.T) {
 					},
 				},
 			},
-			wantViolationCount: 1,
 			wantViolations: []*BreakingChange{
 				{
 					Resource:               "google-x",
@@ -428,7 +425,6 @@ func TestComputeBreakingChanges(t *testing.T) {
 					},
 				},
 			},
-			wantViolationCount: 1,
 			wantViolations: []*BreakingChange{
 				{
 					Resource:               "google-x",
--- a/tools/diff-processor/rules/breaking_changes_test.go
+++ b/tools/diff-processor/rules/breaking_changes_test.go
@@ -332,7 +332,7 @@ func TestComputeBreakingChanges(t *testing.T) {
 			schemaDiff := diff.ComputeSchemaDiff(tc.oldResourceMap, tc.newResourceMap)
 			violations := ComputeBreakingChanges(schemaDiff)
 			for _, v := range violations {
-				if strings.Contains(v, "{{") || strings.Contains(v, "}}") {
+				if strings.Contains(v.Message, "{{") || strings.Contains(v.Message, "}}") {
 					t.Errorf("Test `%s` failed: found unreplaced characters in string - %s", tc.name, v)
 				}
 			}
--- a/tools/diff-processor/rules/breaking_changes_test.go
+++ b/tools/diff-processor/rules/breaking_changes_test.go
@@ -1,19 +1,21 @@
 package rules
 
 import (
+	"sort"
 	"strings"
 	"testing"
 
 	"github.com/GoogleCloudPlatform/magic-modules/tools/diff-processor/diff"
+	"github.com/google/go-cmp/cmp"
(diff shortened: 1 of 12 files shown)


melinath (changes requested):
The goal of this change is to make it so that we have more control over the display of breaking changes in [DIFF_COMMENT.md](https://github.com/GoogleCloudPlatform/magic-modules/blob/main/.ci/magician/cmd/DIFF_COMMENT.md). Currently part of the formatting is happening as part of the `Message` method on rules - for example, [ResourceInventoryRule.Message](https://github.com/GoogleCloudPlatform/magic-modules/blob/210578d95111726e35616730a7d6e004bb536a1b/tools/diff-processor/rules/rules_resource_inventory.go#L62) makes the main message and then adds a documentation reference to the end.

Instead, it would be great to return metadata about the breaking change (for now, resource, message (or maybe even message template?), docs identifier (or docs link) - I'm not sure whether we will want to keep the name / definition fields long-term and there's not a clear need for them at the moment, but it probably wouldn't hurt to include them either) as a JSON structure and format the display of that data in the DIFF_COMMENT.md template.

iyabchen:
> The goal of this change is to make it so that we have more control over the display of breaking changes in [DIFF_COMMENT.md](https://github.com/GoogleCloudPlatform/magic-modules/blob/main/.ci/magician/cmd/DIFF_COMMENT.md). Currently part of the formatting is happening as part of the `Message` method on rules - for example, [ResourceInventoryRule.Message](https://github.com/GoogleCloudPlatform/magic-modules/blob/210578d95111726e35616730a7d6e004bb536a1b/tools/diff-processor/rules/rules_resource_inventory.go#L62) makes the main message and then adds a documentation reference to the end.
> 
> Instead, it would be great to return metadata about the breaking change (for now, resource, message (or maybe even message template?), docs identifier (or docs link) - I'm not sure whether we will want to keep the name / definition fields long-term and there's not a clear need for them at the moment, but it probably wouldn't hurt to include them either) as a JSON structure and format the display of that data in the DIFF_COMMENT.md template.

I refactor the code to return a struct array, but not sure what benefits it can gain. Some template requires fields like min, max etc, more than just resource and field name. Putting different kinds of template into DIFF_COMMENT.md seems difficult as well. 
I just put message and document reference field as the template into DIFF_COMMENT.md.

tools/diff-processor/rules/message_context.go
```diff
@@ -3,21 +3,34 @@ package rules
...
 	message = strings.ReplaceAll(message, "{{field}}", field)
-	return message + documentationReference(mc.identifier)
+	return &BreakingChange{
+		Resource:               mc.Resource,
+		Field:                  mc.Field,
+		Message:                message,
+		RuleTemplate:           template,
+		DocumentationReference: constants.GetFileUrl(mc.identifier),
+		RuleDefinition:         mc.definition,
+		RuleName:               mc.name,
```
melinath, line 34: Do these need to be populated in IsRuleBreak?

tools/diff-processor/rules/breaking_changes_test.go
```diff
@@ -332,7 +332,7 @@ func TestComputeBreakingChanges(t *testing.T) {
 			schemaDiff := diff.ComputeSchemaDiff(tc.oldResourceMap, tc.newResourceMap)
 			violations := ComputeBreakingChanges(schemaDiff)
```
melinath, line 333: Could you add tests here that the standard fields (resource, field, documentationreference, rule*) are set as expected? (This would allow us to refactor the internals later and easily verify the behavior hasn't changed.)

melinath (approved):
That makes sense - +1 that we don't need to move complex formatting for messages from individual rules to the DIFF_COMMENT. Even breaking it down by message & docs reference is an improvement, and makes it easier for us to add more data in the future if it's useful.

Reviewing this PR reminds me that it would be great to refactor this tool to split rule break detection from message formatting... but that's out of scope.

Could you add a breaking change temporarily just to smoke test that the changes work?

melinath (changes requested):
meant to make that a change request, whoops!

tools/diff-processor/rules/message_context.go
iyabchen: This function runs as part of the isRuleBreak. 
I included these because previously u said: "I'm not sure whether we will want to keep the name / definition fields long-term and there's not a clear need for them at the moment, but it probably wouldn't hurt to include them either"

tools/diff-processor/rules/breaking_changes_test.go
iyabchen: Done.

melinath (changes requested):
LGTM - the unit test improvements are great! Could you temporarily add a breaking change to this PR to verify that the generate_comment behavior still works as expected?

iyabchen:
> LGTM - the unit test improvements are great! Could you temporarily add a breaking change to this PR to verify that the generate_comment behavior still works as expected?

Added one space into the DIFF_COMMENT.md file. I generated a breaking change in https://github.com/GoogleCloudPlatform/magic-modules/pull/10692, is that what u meant?

melinath:
You could have done it in this PR, but that works as well.

melinath (approved):
LGTM

PR #10517, opened 2024-04-24
Fix universe_domain for ADC and access token auth cases

c2thorn:
https://github.com/GoogleCloudPlatform/magic-modules/pull/9799 was too heavy handed and completely broke ADC and auth token access when `universe_domain` was configured in Terraform.

This PR sets the universe domain based the following logic:
**Pre-auth**
- Set universe domain to configured value initially as a default

**Auth**
- access token: keep default UD
- credentials file: set to `universe_domain` in file, unset if not found
- ADC:
	- JSON exists: set to `universe_domain` in JSON, unset if not found
	- JSON doesn't exist: call creds.UniverseDomain()

**Post-auth**
- compare credentials UD to configured UD and throw an error if they don't match

```release-note:bug
provider: fixed application default credential and access token authorization when `universe_domain` is set
```

### DIFF
--- a/mmv1/third_party/terraform/provider/provider.go.erb
+++ b/mmv1/third_party/terraform/provider/provider.go.erb
@@ -383,7 +383,9 @@ func ProviderConfigure(ctx context.Context, d *schema.ResourceData, p *schema.Pr
 
 	// Verify that universe domains match between credentials and configuration
 	if v, ok := d.GetOk("universe_domain"); ok {
-		if v.(string) != config.UniverseDomain {
+		if config.UniverseDomain == "" {
+			return nil, diag.FromErr(fmt.Errorf("Universe domain mismatch: '%s' supplied directly to Terraform with no matching universe domain in credentials. Credentials with no 'universe_domain' set are assumed to be in the default universe.", v))
+		} else if v.(string) != config.UniverseDomain {
 			return nil, diag.FromErr(fmt.Errorf("Universe domain mismatch: '%s' does not match the universe domain '%s' supplied directly to Terraform. The 'universe_domain' provider configuration must match the universe domain supplied by credentials.", config.UniverseDomain, v))
 		}
 	} else if config.UniverseDomain != "" && config.UniverseDomain != "googleapis.com" {
--- a/mmv1/third_party/terraform/provider/provider.go.erb
+++ b/mmv1/third_party/terraform/provider/provider.go.erb
@@ -3,7 +3,6 @@ package provider
 
 import (
 	"context"
-	"encoding/json"
 	"fmt"
 	"os"
 	"strings"
--- a/mmv1/third_party/terraform/provider/provider.go.erb
+++ b/mmv1/third_party/terraform/provider/provider.go.erb
@@ -279,35 +279,11 @@ func ProviderConfigure(ctx context.Context, d *schema.ResourceData, p *schema.Pr
 		})
 	}
 	
-	// set universe_domain based on the service account key file.
-	if config.Credentials != "" {
-		contents, _, err := verify.PathOrContents(config.Credentials)
-		if err != nil {
-			return nil, diag.FromErr(fmt.Errorf("error loading service account credentials: %s", err))
-		}
-		var content map[string]any
-
-		if err := json.Unmarshal([]byte(contents), &content); err != nil {
-			return nil, diag.FromErr(err)
-		}
-
-		if content["universe_domain"] != nil {
-			config.UniverseDomain = content["universe_domain"].(string)
-		}
+	// Default the universe domain to the configured value if set
+	if v, ok := d.GetOk("universe_domain"); ok {
+		config.UniverseDomain = v.(string)
 	}
-	
-	// Check if the user provided a value from the universe_domain field other than the default
-	if v, ok := d.GetOk("universe_domain"); ok && v.(string) != "googleapis.com" {
-		if config.UniverseDomain == "" {
-			return nil, diag.FromErr(fmt.Errorf("Universe domain mismatch: '%s' supplied directly to Terraform with no matching universe domain in credentials. Credentials with no 'universe_domain' set are assumed to be in the default universe.", v))
-		} else if v.(string) != config.UniverseDomain {
-			if _, err := os.Stat(config.Credentials); err == nil {
-				return nil, diag.FromErr(fmt.Errorf("Universe domain mismatch: '%s' does not match the universe domain '%s' already set in the credential file '%s'. The 'universe_domain' provider configuration can not be used to override the universe domain that is defined in the active credential.  Set the 'universe_domain' provider configuration when universe domain information is not already available in the credential, e.g. when authenticating with a JWT token.", v, config.UniverseDomain, config.Credentials))
-			} else {
-				return nil, diag.FromErr(fmt.Errorf("Universe domain mismatch: '%s' does not match the universe domain '%s' supplied directly to Terraform. The 'universe_domain' provider configuration can not be used to override the universe domain that is defined in the active credential.  Set the 'universe_domain' provider configuration when universe domain information is not already available in the credential, e.g. when authenticating with a JWT token.", v, config.UniverseDomain))
-			}
-		}
-	}
+
 	// Configure DCL basePath
 	transport_tpg.ProviderDCLConfigure(d, &config)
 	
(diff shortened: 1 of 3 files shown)


rileykarson (commented):
As discussed offline, we want to retrieve information from https://pkg.go.dev/golang.org/x/oauth2/google#Credentials when present

mmv1/third_party/terraform/transport/config.go.erb
```diff
@@ -604,7 +604,15 @@ func (c *Config) getTokenSource(clientScopes []string, initialCredentialsOnly bo
 	creds, err := c.GetCredentials(clientScopes, initialCredentialsOnly)
 	if err != nil {
 		return nil, fmt.Errorf("%s", err)
+	}	
+
+	ud, err := creds.GetUniverseDomain()
+	if err != nil {
+		log.Printf("[WARN] Error retrieving universe domain: %s", err)
+	} else if ud != "googleapis.com" { // Only overwrite if the universe domain is different from the default
```
c2thorn, line 612: GetUniverseDomain() defaults to `googleapis.com` if not set.

Wanted to point out that we now will not differentiate between when the credentials explicitly set `universe_domain = googleapis.com` and the universe domain defaulting to `googleapis.com` because it is not set at all.

This means when TF configuration has `universe_domain = X` and credentials explicitly have `universe_domain = googleapis.com`, we will attempt to operate with `universe_domain = X`.

Not sure how important this case would be. To avoid it we'd have to go back to parsing the various authentication methods to check for set values instead of simply relying on GetUniverseDomain().

rileykarson: As discussed offline yesterday, this should cause an error if X != `googleapis.com`. Older credentials are assumed to default to `googleapis.com` (because they all do- any non-default credentials will have a universe_domain set) and newer credentials will all contain a universe_domain.

AFAIK access tokens are the only exception, as they have no associated domain. We use the provider-configured one in all cases for those.

c2thorn: Updated the logic to unset the universe domain if it is unset in the credentials. The logic is now:
**Pre-auth**
- Set universe domain to configured value initially as a default

**Auth**
- access token: keep default universe domain
- credentials file: set to `universe_domain` in file, unset if not found
- ADC:
	- JSON exists: set to `universe_domain` in JSON, unset if not found
	- JSON doesn't exist: call creds.UniverseDomain()

**Post-auth**
- compare credentials universe domain to configured universe domain and throw an error if they don't match

mmv1/third_party/terraform/provider/provider.go.erb
```diff
@@ -406,6 +381,17 @@ func ProviderConfigure(ctx context.Context, d *schema.ResourceData, p *schema.Pr
 		return nil, diag.FromErr(err)
 	}
 
+	// Verify that universe domains match between credentials and configuration
+	if v, ok := d.GetOk("universe_domain"); ok {
+		if config.UniverseDomain == "" {
```
rileykarson, line 386: If a user explicitly provides GDU this will throw an error incorrectly, right?

mmv1/third_party/terraform/provider/provider.go.erb
```diff
@@ -406,6 +381,17 @@ func ProviderConfigure(ctx context.Context, d *schema.ResourceData, p *schema.Pr
 		return nil, diag.FromErr(err)
 	}
 
+	// Verify that universe domains match between credentials and configuration
+	if v, ok := d.GetOk("universe_domain"); ok {
```
rileykarson, line 385: What does `ok` return on an explicit `""`? It's been a bit.

Provider values are weird compared to resource ones, and we may want to use `d.Get` for predictability.

mmv1/third_party/terraform/transport/config.go.erb
```diff
@@ -1159,12 +1160,25 @@ func (c *Config) GetCredentials(clientScopes []string, initialCredentialsOnly bo
 		}, nil
 	}
 
+	// UniverseDomain is set by the credential file's "universe_domain" field
```
rileykarson, line 1163: This appears in both service account creds and BYOID?

rileykarson: (not new to this PR, just double-checking)

mmv1/third_party/terraform/provider/provider.go.erb
c2thorn: it did, updated the logic to check for v.(string) == "googleapis.com"

mmv1/third_party/terraform/provider/provider.go.erb
c2thorn: ok returns false on explicit "", so it'd be like it wasn't set at all. If credentials have a universe domain, users will get the error 
```
Universe domain mismatch: Universe domain 'X' was found in credentials without a corresponding 'universe_domain' provider configuration set. Please set 'universe_domain' to 'X' or use different credentials.
```
which seems appropriate?

mmv1/third_party/terraform/transport/config.go.erb
c2thorn: Both headful and headless BYOID logins I tested ended up being JSON credential files that contained a `universe_domain`, so yes.

mmv1/third_party/terraform/provider/provider.go.erb
```diff
@@ -279,35 +278,11 @@ func ProviderConfigure(ctx context.Context, d *schema.ResourceData, p *schema.Pr
...
-		} else if v.(string) != config.UniverseDomain {
-			if _, err := os.Stat(config.Credentials); err == nil {
-				return nil, diag.FromErr(fmt.Errorf("Universe domain mismatch: '%s' does not match the universe domain '%s' already set in the credential file '%s'. The 'universe_domain' provider configuration can not be used to override the universe domain that is defined in the active credential.  Set the 'universe_domain' provider configuration when universe domain information is not already available in the credential, e.g. when authenticating with a JWT token.", v, config.UniverseDomain, config.Credentials))
-			} else {
-				return nil, diag.FromErr(fmt.Errorf("Universe domain mismatch: '%s' does not match the universe domain '%s' supplied directly to Terraform. The 'universe_domain' provider configuration can not be used to override the universe domain that is defined in the active credential.  Set the 'universe_domain' provider configuration when universe domain information is not already available in the credential, e.g. when authenticating with a JWT token.", v, config.UniverseDomain))
-			}
-		}
+	// Default the universe domain to the configured value if set
```
rileykarson, line 281:
```suggestion
	// Set the universe domain to the configured value, if any
```

mmv1/third_party/terraform/provider/provider.go.erb
```diff
@@ -406,6 +381,17 @@ func ProviderConfigure(ctx context.Context, d *schema.ResourceData, p *schema.Pr
 		return nil, diag.FromErr(err)
 	}
 
+	// Verify that universe domains match between credentials and configuration
+	if v, ok := d.GetOk("universe_domain"); ok {
+		if config.UniverseDomain == "" && v.(string) != "googleapis.com" {
```
rileykarson, line 386:
```suggestion
		if config.UniverseDomain == "" && v.(string) != "googleapis.com" { // v can't be "", as it wouldn't pass `ok` above
```

mmv1/third_party/terraform/transport/config.go.erb
```diff
@@ -1136,6 +1136,7 @@ type StaticTokenSource struct {
 // If initialCredentialsOnly is true, don't follow the impersonation settings and return the initial set of creds
 // instead.
 func (c *Config) GetCredentials(clientScopes []string, initialCredentialsOnly bool) (googleoauth.Credentials, error) {
+	// UniverseDomain defaults to the previously set provider-configured value for access tokens
```
rileykarson, line 1139:
```suggestion
	// UniverseDomain is assumed to be the previously set provider-configured value for access tokens
```

mmv1/third_party/terraform/transport/config.go.erb
```diff
@@ -1194,33 +1208,51 @@ func (c *Config) GetCredentials(clientScopes []string, initialCredentialsOnly bo
...
-	log.Printf("[INFO] Authenticating using DefaultClient...")
-	log.Printf("[INFO]   -- Scopes: %s", clientScopes)
-
-	if c.UniverseDomain != "" && c.UniverseDomain != "googleapis.com" {
-		log.Printf("[INFO]   -- Sending JwtWithScope option")
-		creds, err := transport.Creds(context.Background(), option.WithScopes(clientScopes...), internaloption.EnableJwtWithScope())
+	if creds.JSON != nil {
+		var content map[string]any
+		if err := json.Unmarshal([]byte(creds.JSON), &content); err != nil {
+			log.Printf("[WARN] error unmarshaling credentials, skipping Universe Domain detection")
```
rileykarson, line 1240: This will preserve the c.UniverseDomain value. Could that result in an issue?

c2thorn: c.UniverseDomain would default to the configured value in this case. In this unexpected case where we could not verify, would we want to lean towards going with the configured value in case it is actually valid?

c2thorn: I would think allowing false positives (the universe domain does not match) would be less disruptive for users than blocking false negatives (the universe domains match but we could not unmarshal the JSON)

rileykarson: > would we want to lean towards going with the configured value in case it is actually valid?

No- sending a request to the wrong universe_domain exposes credentials to an unknown server, which is what we need to guard against. This probably has to always fail (i.e. create a false positive), although I'm not sure how we verify it would be backwards-compatible to do so.

c2thorn: Understood, I will unset the c.UniverseDomain so it can fall through and eventually fail if there is a mismatch.

As far as backwards-compatibility, right now any case that is not a credentials file already leads to the mismatch error due to the bug from the previous PR. So at worse, we would carry forward the same error in this hopefully rare case.

rileykarson (approved):

PR #10519, opened 2024-04-24
Alloydb: add support for PSC based clusters

simonebruzzechesse:
As per the PR title this PR adds support for PSC based Alloydb clusters, this is only supported in the google-beta provider since APIs are still in beta.

```release-note:enhancement
alloydb: added output field block `psc_instance_config` to `google_alloydb_instance ` resource
alloydb: added field block `psc_config` to `google_alloydb_cluster ` resource
```

### DIFF
--- a/mmv1/third_party/terraform/services/alloydb/resource_alloydb_cluster_test.go.erb
+++ b/mmv1/third_party/terraform/services/alloydb/resource_alloydb_cluster_test.go.erb
@@ -1,3 +1,4 @@
+<% autogen_exception -%>
 package alloydb_test
 
 import (
@@ -1296,6 +1297,7 @@ data "google_project" "project" {}
 `, context)
 }
 
+<% unless version == 'ga' -%>
 func TestAccAlloydbCluster_withPSC(t *testing.T) {
 	t.Parallel()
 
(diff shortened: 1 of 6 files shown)


mmv1/third_party/terraform/services/alloydb/resource_alloydb_cluster_test.go.erb
```diff
@@ -1295,3 +1296,48 @@ resource "google_compute_network" "default" {
...
+				ResourceName:      "google_alloydb_cluster.default",
+				ImportState:       true,
+				ImportStateVerify: true,
+			},
+		},
+	})
+}
+<% end -%>
+
+func testAccAlloydbCluster_withPSC(context map[string]interface{}) string {
```
ScottSuarez, line 1329: Could this be included in unless verion==ga block?

otherwise this function is never called in GA provider and will cause a compilation error.

mmv1/third_party/terraform/services/alloydb/resource_alloydb_instance_test.go.erb
```diff
@@ -704,3 +705,53 @@ data "google_compute_network" "default" {
...
+					resource.TestCheckResourceAttr("google_alloydb_instance.default", "psc_instance_config.0.allowed_consumer_projects.#", "0"),
+					resource.TestCheckResourceAttrSet("google_alloydb_instance.default", "psc_instance_config.0.psc_dns_name"),
+				),
+			},
+		},
+	})
+}
+<% end -%>
+
+func testAccAlloydbInstance_createInstanceWithPSC(context map[string]interface{}) string {
```
ScottSuarez, line 735: Same as above

ScottSuarez (changes requested):
Small comments on the testcase

PR #10524, opened 2024-04-24
fix: update resource template to remove beta and remove egress

iennae:
as per @steren egress is not needed; additionally Direct VPC egress for Cloud Run services is now GA: https://cloud.google.com/run/docs/release-notes#April_24_2024

<!-- AUTOCHANGELOG for Downstream PRs.

```release-note:none

```

### DIFF
--- a/mmv1/templates/terraform/examples/cloudrunv2_service_directvpc.tf.erb
+++ b/mmv1/templates/terraform/examples/cloudrunv2_service_directvpc.tf.erb
@@ -1,7 +1,7 @@
 resource "google_cloud_run_v2_service" "<%= ctx[:primary_resource_id] %>" {
   name     = "<%= ctx[:vars]['cloud_run_service_name'] %>"
   location = "us-central1"
-  launch_stage = "BETA"
+  launch_stage = "GA"
   template {
     containers {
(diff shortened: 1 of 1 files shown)


mmv1/templates/terraform/examples/cloudrunv2_service_directvpc.tf.erb
```diff
@@ -1,7 +1,7 @@
 resource "google_cloud_run_v2_service" "<%= ctx[:primary_resource_id] %>" {
   name     = "<%= ctx[:vars]['cloud_run_service_name'] %>"
```
bobfuru, line 2: Double-checking that this name is valid (as opposed to just `cloudrun-service`).

rileykarson: These have associated config blocks as described at https://googlecloudplatform.github.io/magic-modules/develop/test/test/, you can see the rendered value in https://github.com/modular-magician/terraform-provider-google-beta/compare/auto-pr-10524-old..auto-pr-10524

paridhishah18 (approved):

PR #10530, opened 2024-04-25
Address lint warnings in products/documentaiwarehouse/Location.yaml

rileykarson:
Introduced in https://github.com/GoogleCloudPlatform/magic-modules/pull/10526

```release-note:none
```


roaks3 (approved):
Might want to also get line 306 of DocumentSchema.yaml (but really, all yaml lint failures should probably all be fixed at once, so that unrelated changes don't trigger the linter like this; could be a fixit item)

PR #10542, opened 2024-04-26
Rewrite the code to handle handwritten resources

zli82016:
1. Rewrite the logic to handle handwritten resources. Use Go code to replace `common~copy.yaml` and `common~compile.yaml` files.
2. Convert a couple of `.erb` handwritten files to `.tmpl` files.
3. Remove files that are not compiled to downstream.
    * mmv1/third_party/terraform/utils/provider_versionfive_upgrade_test.go.erb
    * mmv1/third_party/terraform/website/docs/r/compute_firewall_policy_rule.html.markdown

```release-note:none

```


mmv1/provider/template_data.go
```diff
@@ -161,10 +161,6 @@ func (td *TemplateData) GenerateFile(filePath, templatePath string, input any, g
 	}
 
 	sourceByte := contents.Bytes()
-	// Replace import path based on version (beta/alpha)
```
zli82016, line 164: The function [importpath](https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/templates/terraform/resource.go.tmpl#L56C19-L56C29) is called in the templates files to decide the import path based on version. With `importpath`, the replacement is not needed.

zli82016:
cc @NickElliot

mmv1/third_party/terraform/go.mod
```diff
@@ -0,0 +1,110 @@
+module github.com/hashicorp/terraform-provider-google
```
c2thorn, line 1: being able to remove the templating on this file is great

mmv1/third_party/terraform/.goreleaser.yml.tmpl
```diff
@@ -0,0 +1,83 @@
+archives:
+  - files:
+      # Include built binary and license files in archive
+      - 'LICENSE'
+    format: zip
+    name_template: '{{"{{ .ProjectName }}_{{ .Version }}_{{ .Os }}_{{ .Arch }}"}}'
```
c2thorn, line 6: wish there was a better way for these escapes... makes it very hard to read

c2thorn (approved):
LGTM

mmv1/third_party/terraform/.goreleaser.yml.tmpl
NickElliot: They're all over the resource.go.tmpl too 🙃

zli82016: The function [Delims](https://medium.com/@etiennerouzeaud/change-default-delimiters-templating-with-template-delims-857938a0b661) can change the default delimiter. If we change the delimiter, we need to make the change in all of templates?

c2thorn: oh wow. Yeah we should think about doing that. Doesn't have to be now.

The downside would be we'd be deviating from standard go templating, and I don't have a pulse on how annoying that would be to contributors

PR #10543, opened 2024-04-26
Update CustomTargetType resource to include googleCloudBuildRepo type

katiexzhang:
…sitories

Fixes https://github.com/hashicorp/terraform-provider-google/issues/17935

```release-note:enhancement
clouddeploy: added `google_cloud_build_repo` to `custom_target_type` resource
```

### DIFF
--- a/mmv1/templates/terraform/examples/clouddeploy_custom_target_type_gcb_repo_skaffold_modules.tf.erb
+++ b/mmv1/templates/terraform/examples/clouddeploy_custom_target_type_gcb_repo_skaffold_modules.tf.erb
@@ -8,7 +8,7 @@ resource "google_clouddeploy_custom_target_type" "<%= ctx[:primary_resource_id] 
       include_skaffold_modules {
         configs = ["my-config"]
         google_cloud_build_repo {
-            repository = "projects/example-project/locations/us-central1/connections/git/repositories/example-repo"
+            repository = "projects/example/locations/us-central1/connections/git/repositories/example-repo"
             path = "configs/skaffold.yaml"
             ref = "main"
         }


katiexzhang:
Hi @BBBmau, could you PTAL at this?

mmv1/third_party/terraform/services/clouddeploy/resource_clouddeploy_custom_target_type_test.go
```diff
@@ -80,6 +80,14 @@ resource "google_clouddeploy_custom_target_type" "custom-target-type" {
           ref = "main"
         }
       }
+      include_skaffold_modules {
+        configs = ["my-config3"]
+        google_cloud_build_repo {
+          repository = "projects/777/locations/us-central1/connections/git/repositories/example"
```
BBBmau, line 86: Where else is this repository being used? Or is this only being used for running the tests that you included in the PR?

katiexzhang: This repository does not exist and does not have to exist - it is only being used for running the tests in this PR.

mmv1/templates/terraform/examples/clouddeploy_custom_target_type_gcb_repo_skaffold_modules.tf.erb
```diff
@@ -0,0 +1,17 @@
...
+    location = "us-central1"
+    name = "<%= ctx[:vars]['custom_target_type_name'] %>"
+    description = "My custom target type"
+    custom_actions {
+      render_action = "renderAction"
+      deploy_action = "deployAction"
+      include_skaffold_modules {
+        configs = ["my-config"]
+        google_cloud_build_repo {
+            repository = "projects/777/locations/us-central1/connections/git/repositories/example"
```
BBBmau, line 11: only request I have is to update the value so that it doesn't look like placeholder value. This is only for examples since this will be shown in docs.

katiexzhang: Done.

katiexzhang:
Bumping this! @BBBmau

mmv1/third_party/terraform/services/clouddeploy/resource_clouddeploy_custom_target_type_test.go
```diff
@@ -80,6 +80,14 @@ resource "google_clouddeploy_custom_target_type" "custom-target-type" {
           ref = "main"
         }
       }
+      include_skaffold_modules {
+        configs = ["my-config3"]
+        google_cloud_build_repo {
+          repository = "projects/example/locations/us-central1/connections/git/repositories/example-repo"
```
BBBmau, line 86: After thinking about it more, although it works as is. In the past, the service team has gotten around to validating input values.

So as it stands this would cause 100% failures in the event that they add a validation which is likely to happen.

Because of this we'll want to have an actual valid repository to prevent this from happening in the future.

katiexzhang: What kind of validation does the service team do?

The referenced resource (the GCB repo) only needs to exist when a Release for a Target that references a CustomTargetType is used.

katiexzhang: Circling back on this @BBBmau

PR #10544, opened 2024-04-26
Request review after PR author comments

melinath:
go/terraform-pr-review-reminders

Note that this also makes it so that we request reviews for _all_ PRs, regardless of target branch. I think that's okay since we're no longer tying review requests to test runs.

```release-note:none

```

### DIFF
--- a/.github/workflows/request-reviewer.yml
+++ b/.github/workflows/request-reviewer.yml
@@ -9,13 +9,13 @@ on:
       - ready_for_review
       - reopened
       - synchronize
-    branches:
-      - 'main'
-      - 'FEATURE-BRANCH-*'
+  issue_comment:
+    types:
+      - created
 
 jobs:
   request-review:
-    if: github.event.pull_request.draft == false
+    if: github.event.pull_request && github.event.pull_request.draft == false && (github.event.sender.login == github.event.pull_request.user.login || github.event.action != 'created')
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


.github/workflows/request-reviewer.yml
```diff
@@ -9,13 +9,13 @@ on:
...
-      - 'main'
-      - 'FEATURE-BRANCH-*'
+  issue_comment:
+    types:
+      - created
 
 jobs:
   request-review:
-    if: github.event.pull_request.draft == false
+    if: github.event.pull_request && github.event.pull_request.draft == false && (github.event.sender.login == github.event.pull_request.user.login || github.event.action != 'created')
```
c2thorn, line 18: just curious did we not need the null check `github.event.pull_request &&` before?

c2thorn (approved):

.github/workflows/request-reviewer.yml
melinath: theoretically, this workflow would trigger for comments on issues (in which case `github.event.issue` would be populated instead). But issues are disabled on this repo so it shouldn't ever happen. But I figure it doesn't hurt to be explicit?

(Previously this wasn't a question because pull_request_target only triggers for pull requests.)

c2thorn: I see, yep it doesn't hurt. Thanks for the explanation

PR #10548, opened 2024-04-29
Add Asset Type run.googleapis.com/Job - b/336823687

dvfons:

```release-note:REPLACEME

```


trodge (approved):

trodge (changes requested):
There seems to be an error in the TGC build. Is that related to this change?

dvfons:
> There seems to be an error in the TGC build. Is that related to this change?

i don't know let me check it

PR #10549, opened 2024-04-29
Secure Web Proxy fields gatewaySecurityPolicy and certificateUrls supports updates.

Samir-Cit:
This PR change the fields _gatewaySecurityPolicy_ and _certificateUrls_ on **Secure Web Proxy** (`google_network_services_gateway`) resource so both can be updatable.

Fixes: https://github.com/hashicorp/terraform-provider-google/issues/14600

- It was needed a new _pre-update_ code to be able to update those fields.
- Also changed the _beta version_ of the product to use **v1beta1**.
- Tests were updated to match the new version.
- Other small indentation (nit) fixes on this PR.

```release-note:enhancement
networkservices: fixed `gateway_security_policy` and `certificate_urls` fields in `google_network_services_gateway` to be updatable
```


mmv1/products/networkservices/product.yaml
```diff
@@ -16,7 +16,7 @@ name: NetworkServices
 versions:
   - !ruby/object:Api::Product::Version
     name: beta
-    base_url: https://networkservices.googleapis.com/v1/
+    base_url: https://networkservices.googleapis.com/v1beta1/
```
roaks3, line 19: Given the test failures, I'm specifically concerned about this change. Have you tested locally? Are we sure this is not a breaking change?

roaks3 (changes requested):
I haven't given a full review yet, but it looks like there are a few different changes to existing behavior, and we would need to make sure this doesn't break users.

mmv1/products/networkservices/product.yaml
Samir-Cit: @roaks3 I ran some tests again in my local environment and this "beta1" made a few tests fail. I already made a push with the code removing it.

But I saw that most of the tests were failing because of those 2 errors:
```
Error: Error creating EdgeCacheKeyset: googleapi: Error 404: Method not found.
Error: Error creating EdgeCacheOrigin: googleapi: Error 404: Method not found.
```
I don't have access to the logs. Can you validate for me if this is the case?

roaks3: Yea, re-running the tests now, but an example error was:

`The requested URL /v1beta1/projects/ci-test-project-188019/locations/global/edgeCacheKeysets?alt=json&amp;edgeCacheKeysetId=tf-test-my-keysetn946pboxbs was not found on this server`

which is why I was wondering about the v1 -> v1beta1 change.

Samir-Cit: Weird, for some reason that fixed the tests here on the PR.
But my local environment was still failing with the messages I sent.

Well, thanks for re-running!
And now that the test passed you can do the full review.

mmv1/templates/terraform/pre_update/network_services_gateway.erb
```diff
@@ -0,0 +1,4 @@
+if d.Get("type") == "SECURE_WEB_GATEWAY" {
```
roaks3, line 1: Can you explain what this is doing and why it was added? From what I can tell, it was previously possible for a user to have a `google_network_services_gateway` with `type = "SECURE_WEB_GATEWAY"`, and then change a field like `server_tls_policy` or `description`, which would trigger this condition, and presumably produce different results than before. Why do these 2 fields need to be set, and what happened before when they weren't set?

roaks3 (commented):
This looks good to me, I'm just wondering about the pre_update, so we have a clear understanding of the previous behavior and the new behavior.

Also, I don't know if this addresses everything in the issue, and I don't know if we want this PR to close it. Specifically, the updates to field descriptions. Are those no longer needed, or perhaps coming in another PR?

mmv1/templates/terraform/pre_update/network_services_gateway.erb
Samir-Cit: I was just double checking with a terraform code using the actual provider and when trying to update the **description** of `google_network_services_gateway ` with `type = SECURE_WEB_GATEWAY` it throw the following error:
`Error: Error waiting for Updating Gateway: Error code 13, message: an internal error has occurred`
_The update send the **name**, the **ID** and the updated **description** fields only:_
```
# google_network_services_gateway.foobar will be updated in-place
  ~ resource "google_network_services_gateway" "foobar" {
      ~ description                          = "my description" -> "new description"
        id                                   = "projects/{PROJECT}/locations/us-east1/gateways/{NAME}"
        name                                 = {NAME}
        # (18 unchanged attributes hidden)
    }
```


When updating through **GCP console**, the request sends all the fields that the Gateway contains:
```
gateway: {
  addresses: [1]
  certificate_urls: [1]
  description: "new description"
  gateway_security_policy: {POLICY}
  name: "projects/{PROJECT}/locations/us-east1/gateways/{NAME}"
  network: {NETWORK}
  ports: [1]
  subnetwork: {SUBNETWORK}
  type: "SECURE_WEB_GATEWAY"
}
update_mask: {
  paths: [
    0: "certificate_urls"
    1: "description"
    2: "gateway_security_policy"
]}
```
_Also, **GCP Console** uses API v1beta1_


To update using the **API Explorer** first I used all fields and then I was removing one by one to check which fields are required and those 2 (`name` and `type`) were required to update the Gateway through **API Explorer**
```
PATCH https://networkservices.googleapis.com/v1/projects/{PROJECT}/locations/us-east1/gateways/{NAME}?updateMask=description&key=[YOUR_API_KEY] HTTP/1.1

Authorization: Bearer [YOUR_ACCESS_TOKEN]
Accept: application/json

{
  "description": "new description",
  "name": "{NAME}",
  "type": "SECURE_WEB_GATEWAY"
}
```

**_Conclusion:_ Since GCP Console sends all fields it's able to update the gateway. So this modification that I did will send the required fields (name and type) to be able to update as well.**

roaks3: Hmm ok, that's interesting behavior, but seems to match your logic. I suppose it's safer to stick with the condition you've included, but I do wonder if it might be cleaner to _always_ include name and type in the request.

Samir-Cit: At first I did that... But if the `type` is other than `SECURE_WEB_GATEWAY ` the request fail.
_That broke some tests when I was starting the development._

Samir-Cit: @roaks3 regarding the **fields description**, it's still tracked on our radar but it was was deprioritized together with ngprabhu@

roaks3: Awesome, thanks for those answers!

roaks3 (approved):

PR #10551, opened 2024-04-29
add Update command to go resource template

NickElliot:
Unfortunately some major TODOs:

Two functions in terraform.rb need to be converted to a go equivalent. Then logic in the template on one of these functions requires creating a map grouping fields by shared attributes.

* field_specific_update_methods
* properties_by_custom_update
* group_by / key[:]

due to update mask being a nested template, it is not rendering at the moment but should be a smaller addition than the others. there also appears to be an issue where $.UpdateMask is not flagging itself true when it should be based on comparing an output file with one currently in TPG (as dummy text within its block did not render).

```release-note:none

```


NickElliot:
cc @zli82016

PR #10561, opened 2024-04-30
Add KMS support for the google_bigquery_connection resource

daanheikens:
Fixes: https://github.com/hashicorp/terraform-provider-google/issues/17955

```release-note:enhancement
compute:  added the 'kms_key_name` field to the 'google_bigquery_connection` resource
```

### DIFF
--- a/mmv1/products/bigqueryconnection/Connection.yaml
+++ b/mmv1/products/bigqueryconnection/Connection.yaml
@@ -133,7 +133,7 @@ examples:
       database_instance_name: 'my-database-instance'
       username: 'user'
       deletion_protection: 'true'
-      kms_key_name: 'bq-key'
+      kms_key_name: 'projects/project/locations/us-central1/keyRings/us-central1/cryptoKeys/bq-key'
     test_vars_overrides:
       deletion_protection: 'false'
       kms_key_name: 'acctest.BootstrapKMSKey(t).CryptoKey.Name'
--- a/mmv1/products/bigqueryconnection/Connection.yaml
+++ b/mmv1/products/bigqueryconnection/Connection.yaml
@@ -125,6 +125,25 @@ examples:
     primary_resource_id: "connection"
     vars:
       connection_id: "my-connection"
+  - !ruby/object:Provider::Terraform::Examples
+    name: 'bigquery_connection_kms'
+    pull_external: true
+    primary_resource_id:
+      'connection'
+    # Random provider
+    skip_vcr: true
+    vars:
+      database_instance_name: 'my-database-instance'
+      username: 'user'
+      deletion_protection: 'true'
+      kms_key_ring: 'bq-connection'
+      kms_key_name: 'bq-key'
+    test_vars_overrides:
+      deletion_protection: 'false'
+    oics_vars_overrides:
+      deletion_protection: 'false'
+    ignore_read_extra:
+      - 'cloud_sql.0.credential'  # password removed
 properties:
   - !ruby/object:Api::Type::String
     name: name
@@ -164,6 +183,12 @@ properties:
     output: true
     description: |
       True if the connection has credential assigned.
+  - !ruby/object:Api::Type::String
+    name: 'kmsKeyName'
+    description: |
+      Optional. The Cloud KMS key that is used for encryption.
+
+      Example: projects/[kms_project_id]/locations/[region]/keyRings/[key_region]/cryptoKeys/[key]
   - !ruby/object:Api::Type::NestedObject
     name: 'cloudSql'
     description: Connection properties specific to the Cloud SQL.
--- a/mmv1/products/bigqueryconnection/Connection.yaml
+++ b/mmv1/products/bigqueryconnection/Connection.yaml
@@ -129,9 +129,7 @@ examples:
     name: 'bigquery_connection_kms'
     pull_external: true
     primary_resource_id:
-      'connection'
-    # Random provider
-    skip_vcr: true
+      'bq-connection-cmek'
     vars:
       database_instance_name: 'my-database-instance'
       username: 'user'
--- a/mmv1/products/bigqueryconnection/Connection.yaml
+++ b/mmv1/products/bigqueryconnection/Connection.yaml
@@ -127,7 +127,6 @@ examples:
       connection_id: "my-connection"
   - !ruby/object:Provider::Terraform::Examples
     name: 'bigquery_connection_kms'
-    pull_external: true
     primary_resource_id:
       'bq-connection-cmek'
     vars:
--- a/mmv1/products/bigqueryconnection/Connection.yaml
+++ b/mmv1/products/bigqueryconnection/Connection.yaml
@@ -134,10 +134,14 @@ examples:
       database_instance_name: 'my-database-instance'
       username: 'user'
       deletion_protection: 'true'
-      kms_key_ring: 'bq-connection'
       kms_key_name: 'bq-key'
     test_vars_overrides:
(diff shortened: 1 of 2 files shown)


mmv1/products/bigqueryconnection/Connection.yaml
```diff
@@ -125,6 +125,25 @@ examples:
     primary_resource_id: "connection"
     vars:
       connection_id: "my-connection"
+  - !ruby/object:Provider::Terraform::Examples
+    name: 'bigquery_connection_kms'
+    pull_external: true
+    primary_resource_id:
+      'connection'
+    # Random provider
+    skip_vcr: true
```
ScottSuarez, line 134: Can we avoid using the random provider here so these test can run in the VCR?

If you use a variable with `-` or `_` in the name we attach a random suffix to it. ie: `my-password`

ScottSuarez (dismissed):

mmv1/products/bigqueryconnection/Connection.yaml
daanheikens: @ScottSuarez Removed it

mmv1/templates/terraform/examples/bigquery_connection_kms.tf.erb
```diff
@@ -0,0 +1,61 @@
...
+        database    = google_sql_database.db.name
+        type        = "POSTGRES"
+        credential {
+          username = google_sql_user.user.name
+          password = google_sql_user.user.password
+        }
+    }
+}
+
+resource "google_kms_key_ring" "default" {
```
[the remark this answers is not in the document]
daanheikens: Hi @ScottSuarez , does this mean i have to create an extra acc test myself for this particular KMS case? Second, what should i do with the `*_generated_test.go` file generated for this example?

melinath:
Hi! I'm taking over this review from Scott.

mmv1/templates/terraform/examples/bigquery_connection_kms.tf.erb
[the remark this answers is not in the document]
melinath: Here's an example of how to pass in a bootstrapped KMS key in a generated test: https://github.com/GoogleCloudPlatform/magic-modules/blob/07c06ecd2a0d5b9a5d040102eb4dc26676e6152b/mmv1/products/secretmanager/Secret.yaml#L46-L53

mmv1/templates/terraform/examples/bigquery_connection_kms.tf.erb
```diff
@@ -0,0 +1,61 @@
...
+
+    deletion_protection  = "<%= ctx[:vars]['deletion_protection'] %>"
+}
+
+resource "google_sql_database" "db" {
+    instance = google_sql_database_instance.instance.name
+    name     = "db"
+}
+
+resource "random_password" "pwd" {
```
melinath, line 17: could you pass in a password (even if it's random) instead of using the random provider? I think this would make the test deterministic, similar to https://github.com/hashicorp/terraform-provider-google-beta/blob/a84b92da47122442524846aeab327f48bdd24743/google-beta/services/databasemigrationservice/resource_database_migration_service_connection_profile_generated_test.go#L33

mmv1/templates/terraform/examples/bigquery_connection_kms.tf.erb
```diff
@@ -0,0 +1,61 @@
...
+}
+
+data "google_project" "project" {
+}
+
+resource "google_kms_crypto_key_iam_member" "default" {
+    crypto_key_id = google_kms_crypto_key.default.id
+    member        = "serviceAccount:bq-${data.google_project.project.number}@bigquery-encryption.iam.gserviceaccount.com"
+    role          = "roles/cloudkms.cryptoKeyEncrypterDecrypter"
+}
```
melinath, line 61: Since this is modifying a project-wide service account, it should also be bootstrapped using BootstrapPSARole similar to:
https://github.com/GoogleCloudPlatform/magic-modules/blob/07c06ecd2a0d5b9a5d040102eb4dc26676e6152b/mmv1/products/cloudfunctions2/Function.yaml#L129-L132

(It doesn't need to be used in the template, just present in the test_vars_overrides.)

melinath: Specifically it would be:

```yaml
policyChanged: 
         "acctest.BootstrapPSARole(t, \"bq-\", \"bigquery-encryption\", 
         \"roles/cloudkms.cryptoKeyEncrypterDecrypter\"\ 
         )" 
```

melinath (changes requested):

daanheikens:
> Hi! I'm taking over this review from Scott.

Hi! Thanks for the review, i will see too it.

mmv1/templates/terraform/examples/bigquery_connection_kms.tf.erb
daanheikens: Added!

mmv1/templates/terraform/examples/bigquery_connection_kms.tf.erb
daanheikens: Done!

mmv1/templates/terraform/examples/bigquery_connection_kms.tf.erb
```diff
@@ -0,0 +1,46 @@
...
+
+data "google_project" "project" {
+}
+
+resource "google_kms_crypto_key_iam_member" "default" {
+    crypto_key_id = "<%= ctx[:vars]['kms_key_name'] %>"
+    member        = "serviceAccount:bq-${data.google_project.project.number}@bigquery-encryption.iam.gserviceaccount.com"
+    role          = "roles/cloudkms.cryptoKeyEncrypterDecrypter"
+}
```
melinath, line 46: This is no longer necessary since the permissions are bootstrapped.

```suggestion
```

mmv1/products/bigqueryconnection/Connection.yaml
```diff
@@ -125,6 +125,27 @@ examples:
...
+  - !ruby/object:Provider::Terraform::Examples
+    name: 'bigquery_connection_kms'
+    pull_external: true
+    primary_resource_id:
+      'bq-connection-cmek'
+    vars:
+      database_instance_name: 'my-database-instance'
+      username: 'user'
+      deletion_protection: 'true'
+      kms_key_name: 'bq-key'
```
melinath, line 137: This should be a full name since it's not a reference - `projects/[kms_project_id]/locations/[region]/keyRings/[key_region]/cryptoKeys/[key]`

mmv1/products/bigqueryconnection/Connection.yaml
```diff
@@ -125,6 +125,27 @@ examples:
     primary_resource_id: "connection"
     vars:
       connection_id: "my-connection"
+  - !ruby/object:Provider::Terraform::Examples
+    name: 'bigquery_connection_kms'
+    pull_external: true
```
melinath, line 130: I believe this is no longer necessary.

```suggestion
```

melinath (changes requested):
Looks like the tests passed successfully! 🎉  Just a few small cleanups and it should be good to go.

mmv1/products/bigqueryconnection/Connection.yaml
daanheikens: Looked at the other example. Hope this works...

melinath (approved):
LGTM

PR #10566, opened 2024-04-30
Set enable_control_plane_v2 field to immutable

naitianliu-google:
According to https://cloud.google.com/anthos/clusters/docs/on-prem/latest/how-to/user-cluster-configuration-file-1.29#enablecontrolplanev2-field, the field should be immutable 

```release-note:enhancement
gkeonprem: set `enable_control_plane_v2` field to immutable in `google_gkeonprem_vmware_cluster` resource
```

### DIFF
--- a/mmv1/products/gkeonprem/VmwareCluster.yaml
+++ b/mmv1/products/gkeonprem/VmwareCluster.yaml
@@ -535,6 +535,7 @@ properties:
     name: 'enableControlPlaneV2'
     description: Enable control plane V2. Default to false.
     immutable: true
+    default_value: true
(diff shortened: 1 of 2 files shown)


c2thorn:
Can you help me with one more level of confirmation that the field cannot be updated? It'd be breaking if we were not correct, and I just want to make sure as documentation is sometimes outdated/incorrect.

Could be a snippet of the resulting API error or an internal ping pointing to the relevant code.

c2thorn (changes requested):
Further information was requested in https://github.com/GoogleCloudPlatform/magic-modules/pull/10566#issuecomment-2089110437

PR #10568, opened 2024-04-30
Allow deletion of resource_bigquery_table when it still has associated resource tags

wj-chen:
Allow deletion of resource_bigquery_table when it still has associated resource tags.

```release-note:bug
bigquery: added `allow_resource_tags_on_deletion` to `google_bigquery_table` to allow deletion of table when it still has associated resource tags (beta)
```

### DIFF
--- a/mmv1/third_party/terraform/services/bigquery/resource_bigquery_table.go.erb
+++ b/mmv1/third_party/terraform/services/bigquery/resource_bigquery_table.go.erb
@@ -1832,18 +1832,6 @@ func resourceBigQueryTableDelete(d *schema.ResourceData, meta interface{}) error
 	if d.Get("deletion_protection").(bool) {
 		return fmt.Errorf("cannot destroy table %v without setting deletion_protection=false and running `terraform apply`", d.Id())
 	}
-	<% unless version == 'ga' -%>
-	if v, ok := d.GetOk("resource_tags"); ok {
-		var resourceTags []string
-
-		for k, v := range v.(map[string]interface{}) {
-			resourceTags = append(resourceTags, fmt.Sprintf("%s:%s", k, v.(string)))
-		}
-
-		return fmt.Errorf("cannot destroy table %v without clearing the following resource tags: %v", d.Id(), resourceTags)
-	}
-
-	<% end -%>
 	config := meta.(*transport_tpg.Config)
 	userAgent, err := tpgresource.GenerateUserAgentString(d, config.UserAgent)
 	if err != nil {
--- a/mmv1/third_party/terraform/services/bigquery/resource_bigquery_table.go.erb
+++ b/mmv1/third_party/terraform/services/bigquery/resource_bigquery_table.go.erb
@@ -2686,9 +2686,11 @@ func resourceBigQueryTableImport(d *schema.ResourceData, meta interface{}) ([]*s
 	if err := d.Set("deletion_protection", true); err != nil {
 		return nil, fmt.Errorf("Error setting deletion_protection: %s", err)
 	}
+	<% unless version == 'ga' -%>
 	if err := d.Set("allow_resource_tags_on_deletion", false); err != nil {
 		return nil, fmt.Errorf("Error setting allow_resource_tags_on_deletion: %s", err)
 	}
+	<% end -%>
 
 	// Replace import id for the resource id
 	id, err := tpgresource.ReplaceVars(d, config, "projects/{{project}}/datasets/{{dataset_id}}/tables/{{table_id}}")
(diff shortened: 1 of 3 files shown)


sachinpro (approved):

mmv1/third_party/terraform/services/bigquery/resource_bigquery_table.go.erb
```diff
@@ -1832,18 +1832,6 @@ func resourceBigQueryTableDelete(d *schema.ResourceData, meta interface{}) error
 	if d.Get("deletion_protection").(bool) {
 		return fmt.Errorf("cannot destroy table %v without setting deletion_protection=false and running `terraform apply`", d.Id())
 	}
-	<% unless version == 'ga' -%>
-	if v, ok := d.GetOk("resource_tags"); ok {
```
shuyama1, line 1836: I think this is a breaking change that should be avoided a minor release. This is `Changing resource deletion behavior` in https://googlecloudplatform.github.io/magic-modules/develop/breaking-changes/breaking-changes/#resource-level-breaking-changes.

We have a couple of options:

1. wait until the next major release, which is expected in a couple of months
2. introduce a virtual field to force delete the resource even when resource_tags are present and default it to false.  

Please let me know your thoughts and if you have any questions. Thanks!

wj-chen: Would the combination of the points below make it eligible for a minor version release?

- Feature was very newly added (released on 4/22)
- Feature is in the beta provider
- The new deletion behavior is more permissive instead of more restrictive

shuyama1: @wj-chen Thanks for the quick response.

> * Feature was very newly added (released on 4/22)
> * Feature is in the beta provider

The beta provider is also a stable surface that does not allow breaking changes outside of major releases. See go/terraform-launch-stages#google-beta

> * The new deletion behavior is more permissive instead of more restrictive

I'd still think it's breaking since it's changing the default deletion behavior and I'm a bit hesitant about changing it directly in a minor. However, if an urgent fix is needed in a minor, we could follow https://googlecloudplatform.github.io/magic-modules/develop/breaking-changes/make-a-breaking-change/#in-minor-releases - "this change can be made if the default behavior stays the same and new behavior can be enabled with a flag", such as adding a `force_delete` field, which is the option 2 I proposed in https://github.com/GoogleCloudPlatform/magic-modules/pull/10568#discussion_r1588026232. Please let me know what your thought on this approach and if it will work for your use case. Thanks!

mmv1/third_party/terraform/services/bigquery/resource_bigquery_table.go.erb
```diff
@@ -2675,6 +2686,9 @@ func resourceBigQueryTableImport(d *schema.ResourceData, meta interface{}) ([]*s
 	if err := d.Set("deletion_protection", true); err != nil {
 		return nil, fmt.Errorf("Error setting deletion_protection: %s", err)
 	}
+	if err := d.Set("allow_resource_tags_on_deletion", false); err != nil {
+		return nil, fmt.Errorf("Error setting allow_resource_tags_on_deletion: %s", err)
+	}
```
shuyama1, line 2691: I think we may also need to mark it as beta only.

wj-chen: Added

shuyama1 (approved):
Thanks!

mmv1/third_party/terraform/services/bigquery/resource_bigquery_table.go.erb
shuyama1: Will discuss internally regarding the migration path for the next major release

wj-chen: Hi @shuyama1, I'd like to follow up on this discussion again since the submission window for the next major release is now open. We want to promote this field to the GA provider, and make the default value `true` in the GA provider from the get-go if possible.

* Would it count as a breaking change if the promotion and default value update take place together? If not, we'll proceed with that since it's the most ideal. Updating the default value for the beta provider will be treated as a separate breaking change.
* Otherwise, we would most likely promote first then make a breaking change to update the default value for both providers as a follow-up. This way the GA promotion won't be blocked by the next major release timeline.

shuyama1: @wj-chen Thank you for following up on this issue.

Our policy requires that the beta provider be a superset of the GA provider. This means the features and behaviors in GA should always be consistent with those in beta.

If we want to promote the feature to GA before changing the default in beta in 6.0.0 (as you correctly noted, this default value swap is a breaking change and must wait until the 6.0.0 major release), we need to use the current default value and update the default values for both GA and beta later in the 6.0.0 release.

Hope this answers your question and let me know if you have further questions.

wj-chen: In this case we'll proceed with promoting the feature to GA first, then making a breaking change to change the default value for the field for both GA and beta providers in the 6.0.0 release. Thank you for the guidance!

shuyama1: Perfect. Sounds good!

PR #10577, opened 2024-05-01
Add Terraform definition and tests for Interconnect resources.

bopas2:
Add support for Interconnect resources.

```release-note:new-resource
`google_compute_interconnect`
```

### DIFF
--- a/mmv1/products/compute/Interconnect.yaml
+++ b/mmv1/products/compute/Interconnect.yaml
@@ -1,0 +1,385 @@
+# Copyright 2024 Google Inc.
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+--- !ruby/object:Api::Resource
+name: 'Interconnect'
+kind: 'compute#Interconnect'
+description: |
+  Represents an Interconnect resource. The Interconnect resource is a dedicated connection between
+  Google's network and your on-premises network.
+references: !ruby/object:Api::Resource::ReferenceLinks
+  guides:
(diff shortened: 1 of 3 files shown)


mmv1/products/compute/Interconnect.yaml
```diff
@@ -0,0 +1,385 @@
...
+      - 'PENDING'
+      - 'RUNNING'
+      - 'DONE'
+  error: !ruby/object:Api::OpAsync::Error
+    path: 'error/errors'
+    message: 'message'
+examples:
+  - !ruby/object:Provider::Terraform::Examples
+    name: 'compute_interconnect_basic'
+    skip_test: true
```
slevenick, line 47: What is the procedure for testing these resources? Typically interconnects are very difficult to test

mmv1/products/compute/Interconnect.yaml
```diff
@@ -0,0 +1,385 @@
...
+      - :LINK_TYPE_ETHERNET_100G_LR
+    required: true
+    immutable: true
+  - !ruby/object:Api::Type::Integer
+    name: 'requestedLinkCount'
+    description: |
+      Target number of physical links in the link bundle, as requested by the customer.
+    required: true
+    immutable: true
+  - !ruby/object:Api::Type::Enum
```
slevenick, line 107: This could be a string so we don't document the IT_PRIVATE value

slevenick (commented):
Is update supported on this API? If so, can you write an update test?

mmv1/products/compute/Interconnect.yaml
bopas2: The official api documentation uses an enum, so this will be okay as is. Thank you for raising the alternative solution.

mmv1/products/compute/Interconnect.yaml
bopas2: We are using an approach that only works for testing Interconnects on internal GCP projects. Resolved offline.

bopas2:
> Is update supported on this API? If so, can you write an update test?

Patch operations are supported, however, it is not possible to add an update test with our testing approach.  I will follow up offline.

bopas2:
> > Is update supported on this API? If so, can you write an update test?
> 
> Patch operations are supported, however, it is not possible to add an update test with our testing approach. I will follow up offline.

We will make Interconnects immutable until an acceptable test implementation is available.

slevenick:
/gcbrun there's an error that doesn't look related to your changes

slevenick (approved):
Thanks!

PR #10580, opened 2024-05-02
Add support for Managed zone resource of Integration Connectors.

balanaguharsha:
**Release Note Template for Downstream PRs (will be copied)**

```release-note:new-resource
`google_integration_connectors_managed_zone`
```

PR #10594, opened 2024-05-03
Skip the project field for missing test detection

trodge:

```release-note:none

```

### DIFF
--- a/tools/diff-processor/detector/detector.go
+++ b/tools/diff-processor/detector/detector.go
@@ -47,6 +47,10 @@ func getChangedFieldsFromSchemaDiff(schemaDiff diff.SchemaDiff) map[string]Resou
 	for resource, resourceDiff := range schemaDiff {
 		resourceChanges := make(ResourceChanges)
 		for field, fieldDiff := range resourceDiff.Fields {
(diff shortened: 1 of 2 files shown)


melinath (changes requested):
why do we need this change? Could you add a test?

trodge:
> why do we need this change? Could you add a test?

We generally don't include the project field in tests because the provider adds it automatically.

melinath:
> > why do we need this change? Could you add a test?
> 
> We generally don't include the project field in tests because the provider adds it automatically.

That makes sense, did this come up recently in a PR? Or is there a ticket tracking it?

melinath (changes requested):
test failure

trodge:
> > > why do we need this change? Could you add a test?
> > 
> > 
> > We generally don't include the project field in tests because the provider adds it automatically.
> 
> That makes sense, did this come up recently in a PR? Or is there a ticket tracking it?

Came up here: https://github.com/GoogleCloudPlatform/magic-modules/pull/10580#issuecomment-2093506889

melinath (approved):

PR #10595, opened 2024-05-03
PR review reminders

melinath:
Note: this is a large PR but most of it is tests. This will run as a dry run for a while so I can double-check the output and make sure everything's fine before implementing & moving to non-dry-run.

```release-note:none

```

### DIFF
--- a/.ci/magician/cmd/scheduled_pr_reminders.go
+++ b/.ci/magician/cmd/scheduled_pr_reminders.go
@@ -36,7 +36,7 @@ import (
 
 var (
 	// used for flags
-	dryRun   bool
+	dryRun bool
 
 	//go:embed SCHEDULED_PR_WAITING_FOR_CONTRIBUTOR.md.tmpl
 	waitingForContributorTemplate string
@@ -49,9 +49,9 @@ var (
 )
 
 type reminderCommentData struct {
-		PullRequest *github.PullRequest
-		State pullRequestReviewState
-		SinceDays int
+	PullRequest *github.PullRequest
+	State       pullRequestReviewState
+	SinceDays   int
 }
 
 // scheduledPrReminders sends automated PR notifications and closes stale PRs
@@ -100,7 +100,7 @@ func execScheduledPrReminders(gh *github.Client) error {
 		if *pr.Draft {
 			fmt.Printf(
 				"%d/%d: PR %d: Skipping draft pr\n",
-				index + 1,
+				index+1,
 				len(allPulls),
 				*pr.Number,
 			)
@@ -149,7 +149,7 @@ func execScheduledPrReminders(gh *github.Client) error {
 		if err != nil {
 			fmt.Printf(
 				"%d/%d: PR %d: error computing notification state: %s\n",
-				index + 1,
+				index+1,
 				len(allPulls),
 				*pr.Number,
 				err,
@@ -158,7 +158,7 @@ func execScheduledPrReminders(gh *github.Client) error {
 		}
 		fmt.Printf(
 			"%d/%d: PR %d: %s since %v\n",
-			index + 1,
+			index+1,
 			len(allPulls),
 			*pr.Number,
 			state,
@@ -168,12 +168,12 @@ func execScheduledPrReminders(gh *github.Client) error {
 		if shouldNotify(pr, state, sinceDays) {
 			comment, err := formatReminderComment(state, reminderCommentData{
 				PullRequest: pr,
-				SinceDays: sinceDays,
+				SinceDays:   sinceDays,
 			})
 			if err != nil {
 				fmt.Printf(
 					"%d/%d: PR %d: error rendering comment: %s\n",
-					index + 1,
+					index+1,
 					len(allPulls),
 					*pr.Number,
 					err,
@@ -183,7 +183,7 @@ func execScheduledPrReminders(gh *github.Client) error {
 			if dryRun {
 				fmt.Printf("DRY RUN: Would post comment: %s\n", comment)
 			} else {
-				
+
 			}
 		}
 
@@ -191,7 +191,7 @@ func execScheduledPrReminders(gh *github.Client) error {
 			if dryRun {
 				fmt.Printf("DRY RUN: Would close PR %d\n", *pr.Number)
 			} else {
-				
+
 			}
 		}
 	}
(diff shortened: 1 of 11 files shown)


.ci/magician/cmd/scheduled_pr_reminders.go
```diff
@@ -0,0 +1,419 @@
...
+		if _, ok := approvedBy[reviewer]; ok {
+			continue
+		}
+		switch *review.State {
+		case "DISMISSED":
+			// ignore dismissed reviews
+			continue
+		case "APPROVED":
+			approvedBy[reviewer] = struct{}{}
+			// only consider approvals that are the most recent review by the reviewer.
```
shuyama1, line 311: I think the overall system looks good to me. There's only one specific scenario that I'm not certain about - comments after approval from the same contributor. I wonder if we should consider sending reminders to reviewers instead of contributors in such cases, as if the following comment should block the PR merge, the reviewer should probably dismiss the previous approval instead. One common reason that I can think of for this case is reviewer leave comment to announce pause on merging PRs due to probably CI issues.

.github/workflows/scheduled-pr-reminders.yml
```diff
@@ -0,0 +1,30 @@
+name: scheduled-pr-reminders
+
+permissions: read-all
+
+on:
+  schedule:
+    - cron: '37 9 * * *'
```
shuyama1, line 7: Out of curiosity, why is this particular time chosen?

.ci/magician/cmd/scheduled_pr_reminders_test.go
```diff
@@ -0,0 +1,865 @@
...
+				},
+				&github.PullRequestReview{
+					User:        &github.User{Login: github.String(secondCoreReviewer)},
+					State:       github.String("COMMENTED"),
+					SubmittedAt: &github.Timestamp{time.Date(2024, 2, 2, 0, 0, 0, 0, time.UTC)},
+				},
+			},
+			expectState: waitingForContributor,
+			expectSince: time.Date(2024, 2, 2, 0, 0, 0, 0, time.UTC),
+		},
```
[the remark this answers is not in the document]
melinath: "Comment review but no approval" is correct because the "APPROVED" review is overwritten by the "COMMENTED" review (because they have the same reviewer).

For comments that don't signify a request for contributor action, reviewers can leave a normal comment instead of a review of type "COMMENTED". Reviews are the only mechanism we have to signal that there is expected action from contributors; in general, I would expect a "COMMENTED" review to mean that something is blocking the PR besides a specific change (for example, needing more information about the use case). For cases where we don't want to notify the contributor, we can apply the `disable-automatic-closure` label.

.ci/magician/cmd/scheduled_pr_reminders_test.go
```diff
@@ -0,0 +1,865 @@
...
+				},
+				&github.PullRequestReview{
+					User:        &github.User{Login: github.String(secondCoreReviewer)},
+					State:       github.String("COMMENTED"),
+					SubmittedAt: &github.Timestamp{time.Date(2024, 2, 2, 0, 0, 0, 0, time.UTC)},
+				},
+			},
+			expectState: waitingForMerge,
+			expectSince: time.Date(2024, 2, 1, 0, 0, 0, 0, time.UTC),
+		},
```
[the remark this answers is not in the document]
melinath: yes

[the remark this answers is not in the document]
melinath: specifically, a new review of type "COMMENTED" would dismiss a previous review of type "APPROVED" from the same contributor. New comments that are _not_ part of a review would have no impact.

.ci/magician/cmd/scheduled_pr_reminders.go
melinath: A new review of type "COMMENTED" would dismiss a previous review of type "APPROVED" from the same contributor. New comments that are not part of a review would have no impact.

I testing this on https://github.com/GoogleCloudPlatform/magic-modules/pull/10495 and GitHub's approval behavior is:

1. I approve. State: approved.
2. I add a COMMENTED review. State: approved.
3. I add a CHANGES_REQUESTED review. State: Changes requested.
4. I dismiss the CHANGES_REQUESTED review. State: Unreviewed.

This end state has 3 reviews which are APPROVED, COMMENTED, and DISMISSED: https://api.github.com/repos/GoogleCloudPlatform/magic-modules/pulls/10495/reviews

This behavior seems clearly inconsistent to me; the dismissed review should be ignored and return the PR to its approved state.

After thinking on it some more, I am ambivalent about this; making COMMENTED not override APPROVED seems like a reasonable compromise.

.github/workflows/scheduled-pr-reminders.yml
melinath: The time is in UTC; after 9am should mean that the action always runs on the same day in PDT as UTC (in case that could end up being confusing) while still running early in the day for folks in the UK. `37` is arbitrary just to avoid running on the hour in case that's crowded.

.ci/magician/cmd/scheduled_pr_reminders.go
melinath: Changed in a86c375ac

.ci/magician/cmd/scheduled_pr_reminders_test.go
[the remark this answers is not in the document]
melinath: Changed this behavior in a86c375ac - see the other thread about it.

.ci/magician/cmd/scheduled_pr_reminders_test.go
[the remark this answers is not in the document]
melinath: changed this behavior in a86c375ac - see the other thread about it.

.ci/magician/cmd/scheduled_pr_reminders_test.go
```diff
@@ -0,0 +1,897 @@
...
+	"testing"
+	"time"
+
+	membership "magician/github"
+
+	"github.com/google/go-github/v61/github"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestNotificationState(t *testing.T) {
```
shuyama1, line 13: ~are we running these unit tests anywhere?~

nvm, looks like the tests are included in membership-checker unit tests. The GHA is running all magician unit tests.

.github/workflows/scheduled-pr-reminders.yml
shuyama1: yea, makes sense. Thanks

shuyama1 (approved):

PR #10599, opened 2024-05-06
Add Suport for `google_cloud_run_Job` to TGC

dvfons:

```release-note:none
```


shuyama1 (commented):
Thanks for the work!

Review pending internal discussion

sushant00:
@dvfons Since this is a TGC-only change it should use release note type none and the content of the release note should be empty.

shuyama1 (approved):

PR #10601, opened 2024-05-06
Fixes storage bucket custom_time_before/noncurrent_time_before update issue

kautikdk:
Fixes hashicorp/terraform-provider-google/issues/12544

```release-note:bug
storage: fixed a bug where `google_storage_bucket.lifecycle_rule.condition.custom_time_before` and `noncurrent_time_before update issue` was not updating.
```


kautikdk:
Context: The lifecycle rule conditions are of TypeSet https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/third_party/terraform/services/storage/resource_storage_bucket.go.erb#L191 .
There is a Hash Function for conditions which doesn't have these two fields and that's why terraform is not showing any diff when updating these two fields. Reference: https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/third_party/terraform/services/storage/resource_storage_bucket.go.erb#L1615

kautikdk:
Hi @roaks3, Can you please review this PR!

mmv1/third_party/terraform/services/storage/resource_storage_bucket.go.erb
```diff
@@ -1640,6 +1640,14 @@ func resourceGCSBucketLifecycleRuleConditionHash(v interface{}) int {
 		buf.WriteString(fmt.Sprintf("%s-", v.(string)))
 	}
 
+	if v, ok := m["custom_time_before"]; ok {
+		buf.WriteString(fmt.Sprintf("%s", v.(string)))
```
roaks3, line 1644: I think the hyphen delimiter is important here (at least generally), to mark the start and end of a field. Am I missing something?

mmv1/third_party/terraform/services/storage/resource_storage_bucket_test.go.erb
```diff
@@ -323,6 +323,12 @@ func TestAccStorageBucket_lifecycleRulesMultiple(t *testing.T) {
 			},
 			{
 				Config: testAccStorageBucket_lifecycleRulesMultiple_update(bucketName),
+				Check: resource.ComposeTestCheckFunc(
```
roaks3, line 326: I think the `ImportStateVerify` covers this for you. Do you know if these checks fail without your fix in place?

roaks3 (commented):
LGTM with two comments. It seems like changes to only these two fields would have been ignored due to the hash detecting no change?

mmv1/third_party/terraform/services/storage/resource_storage_bucket.go.erb
kautikdk: Nope. I think, I missed putting hyphen there. Will update the PR.

mmv1/third_party/terraform/services/storage/resource_storage_bucket_test.go.erb
kautikdk: Well, I thought of not explicitly checking fields but currently we are having complex issues with lifecycle rules. From that issues I found that there are many problems underneath in terraform core that we can't control and sometimes it creates unexpected behaviors. That's why I put this checks explicitly for lifecycle rules. If we don't any problems with tests or we don't require any separate check for TypeSet then I will remove it!
If It is okay to remove them I can remove them from the PR.

roaks3: Hmm yea, the TypeSet does make it a bit weird, and I'm actually wondering if the indexes in these check would be useful in the way you are hoping. I think they could be removed, but like I said, it would be nice to have something in place that fails with the previous (broken) implementation.

mmv1/third_party/terraform/services/storage/resource_storage_bucket.go.erb
kautikdk: Updated PR.

mmv1/third_party/terraform/services/storage/resource_storage_bucket_test.go.erb
kautikdk: Yes, I modified one of the update config to change value of  `custom_time_before` field and added `noncurrent_time_before` fields in the tests configs. That should be enough to tests that field are being modified properly. The indexes were working, I tried changing field values and tests were failing correctly as expected.

kautikdk:
@roaks3 Updated PR with respect to comments.

roaks3 (approved):
Great, thanks for the responses/changes! LGTM

PR #10605, opened 2024-05-06
Add KUBELET and CADVISOR to monitoring components

TheSpiritXIII:

```release-note:enhancement
container: added `KUBELET` and `CADVISOR` options to `monitoring_config.enable_components` in `google_container_cluster` resource
```


shuyama1 (commented):
The test failed with error `Error: googleapi: Error 400: Request contains an invalid argument., badRequest`

Do we know if these values are already supported?

TheSpiritXIII:
@shuyama1 thanks for the reply!

> Do we know if these values are already supported?

They're only available on [GKE 1.29 and above](https://cloud.google.com/kubernetes-engine/docs/how-to/configure-metrics#enable-cadvisor-kubelet).

I removed them from the test and added a note about support on the HTML page. Let me know if you want me to do any further changes. Thanks again!

shuyama1 (approved):

PR #10611, opened 2024-05-07
fix: Verify `google_cloud_run_v2_{service,job}` 'timeout' field with regex

alexrjones:
Fixes https://github.com/hashicorp/terraform-provider-google/issues/18028.

```release-note:bug
cloudrunv2: added validation for `timeout` field to `google_cloud_run_v2_job`, `google_cloud_run_v2_service`
```

### DIFF
--- a/mmv1/products/cloudrunv2/Job.yaml
+++ b/mmv1/products/cloudrunv2/Job.yaml
@@ -523,6 +523,8 @@ properties:
 
               A duration in seconds with up to nine fractional digits, ending with 's'. Example: "3.5s".
             default_from_api: true
+            validation: !ruby/object:Provider::Terraform::Validation
+              regex: '^[0-9]+(?:\.[0-9]{1,9})?s$'
           - !ruby/object:Api::Type::String
             name: 'serviceAccount'
             description: |-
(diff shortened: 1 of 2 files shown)


bskaplan (approved):

alexrjones:
@trodge, would you mind taking a look at this PR? Thanks in advance.

trodge:
Is there any way to verify that there are no valid values for this field that wouldn't match this regex?

alexrjones:
The field is a `google.protobuf.Duration` type according to Cloud Run V2 API docs ([ref](https://cloud.google.com/run/docs/reference/rest/v2/projects.locations.services)). The [protobuf declaration](https://github.com/protocolbuffers/protobuf/blob/ee98ba2c18486938952a7748765f2e8e119e7fc4/src/google/protobuf/duration.proto#L92-L100) says the following about its JSON representation:

> In JSON format, the Duration type is encoded as a string rather than an object, where the string ends in the suffix "s" (indicating seconds) and is preceded by the number of seconds, with nanoseconds expressed as fractional seconds. For example, 3 seconds with 0 nanoseconds should be encoded in JSON format as "3s", while 3 seconds and 1 nanosecond should be expressed in JSON format as "3.000000001s", and 3 seconds and 1 microsecond should be expressed in JSON format as "3.000001s".

The Cloud Run API docs additionally say:
> A duration in seconds with up to nine fractional digits.

To me, that implies the following invariants:
- the string must end with `s`
- the number of seconds must always be present
- the fractional portion may be omitted when the nanosecond part of the duration is zero
- there must be between one and nine digits in the fractional portion of the string when it is present

To the best of my understanding, this regex enforces all of these invariants.

trodge (approved):

PR #10615, opened 2024-05-07
Adding Datasource: google_composer_user_workloads_secret

spapi17:

```release-note:new-datasource
`google_composer_user_workloads_secret` (beta)
```


mmv1/third_party/terraform/website/docs/d/composer_user_workloads_secret.html.markdown
```diff
@@ -0,0 +1,60 @@
+---
+subcategory: "Cloud Composer"
+description: |-
+  Kubernetes secret in Composer Environment workloads.
```
kosteev, line 4: Please make "secret" capfirst as "Secret" here and in other places of this file

kosteev (approved):

mmv1/third_party/terraform/website/docs/d/composer_user_workloads_secret.html.markdown
spapi17: fixed

spapi17:
Could you please re-run the tests? The recent API changes should resolve the previous failures.

mmv1/third_party/terraform/services/composer/data_source_google_composer_user_workloads_secret.go
```diff
@@ -0,0 +1,64 @@
+package composer
+
+import (
```
melinath, line 3:
```suggestion
<% unless version == 'ga' -%>
import (
```

mmv1/third_party/terraform/services/composer/data_source_google_composer_user_workloads_secret.go
```diff
@@ -0,0 +1,64 @@
...
+	if err := tpgresource.SetDataSourceLabels(d); err != nil {
+		return err
+	}
+
+	if d.Id() == "" {
+		return fmt.Errorf("%s not found", id)
+	}
+
+	return nil
+}
```
melinath, line 64:
```suggestion
}
<% end -%>
```

mmv1/third_party/terraform/services/composer/data_source_google_composer_user_workloads_secret_test.go
```diff
@@ -0,0 +1,97 @@
+package composer_test
+
+import (
```
melinath, line 3:
```suggestion
<% unless version == 'ga' -%>
import (
```

mmv1/third_party/terraform/services/composer/data_source_google_composer_user_workloads_secret_test.go
```diff
@@ -0,0 +1,97 @@
...
+    username: base64encode("username"),
+    password: base64encode("password"),
+  }
+}
+data "google_composer_user_workloads_secret" "test" {
+  name        = google_composer_user_workloads_secret.test.name
+  environment = google_composer_environment.test.name
+}
+`, context)
+}
```
melinath, line 97:
```suggestion
}
<% end -%>
```

mmv1/third_party/terraform/website/docs/d/composer_user_workloads_secret.html.markdown
```diff
@@ -0,0 +1,60 @@
+---
+subcategory: "Cloud Composer"
+description: |-
+  User workloads Secret used by Airflow tasks that run with Kubernetes Executor or KubernetesPodOperator.
+---
+
+# google\_composer\_user\_workloads\_secret
+
+Provides access to Kubernetes Secret configuration for a given project, region and Composer Environment.
+
```
melinath, line 10:
```suggestion

~> **Warning:** This data source is in beta, and should be used with the terraform-provider-google-beta provider.
See [Provider Versions](https://terraform.io/docs/providers/google/guides/provider_versions.html) for more details on beta resources.

```

melinath (changes requested):
A few changes since this is a beta-only data source. (These are why the `google` unit tests are failing to compile.) Pushing these changes will rerun tests.

mmv1/third_party/terraform/website/docs/d/composer_user_workloads_secret.html.markdown
```diff
@@ -0,0 +1,63 @@
...
+    environment = google_composer_environment.example.name
+    name = resource.google_composer_user_workloads_secret.example.name
+}
+
+output "debug" {
+    value = data.google_composer_user_workloads_secret.example
+}
+```
+
+## Argument Reference
```
melinath, line 45: this is missing documentation for the `data` field. you could copy the docs from the resource - but for data sources we often just refer to the resource docs instead of restating them. For example, see https://registry.terraform.io/providers/hashicorp/google/latest/docs/data-sources/bigquery_dataset#attributes-reference

mmv1/third_party/terraform/services/composer/data_source_google_composer_user_workloads_secret_test.go.erb
```diff
@@ -0,0 +1,100 @@
...
+
+		dsAttr := ds.Primary.Attributes
+		rsAttr := rs.Primary.Attributes
+		errMsg := ""
+
+		for k := range rsAttr {
+			if k == "%" {
+				continue
+			}
+			// ignore diff if it's due to secrets being masked.
```
melinath, line 57: why would the secrets be masked?

mmv1/third_party/terraform/services/composer/data_source_google_composer_user_workloads_secret.go.erb
```diff
@@ -0,0 +1,67 @@
...
+	}
+	d.SetId(id)
+
+	// retrieve "data" in advance, because Read function won't do it.
+	userAgent, err := tpgresource.GenerateUserAgentString(d, config.UserAgent)
+	if err != nil {
+		return err
+	}
+
+	res, err := config.NewComposerClient(userAgent).Projects.Locations.Environments.UserWorkloadsSecrets.Get(id).Do()
```
melinath, line 43: Normally data sources would correspond pretty closely to the Read function of the resource; I noticed that this line is present in the resource's Import function but not in Read. Why is that?

mmv1/third_party/terraform/services/composer/data_source_google_composer_user_workloads_secret.go.erb
```diff
@@ -0,0 +1,67 @@
...
+	}
+
+	err = resourceComposerUserWorkloadsSecretRead(d, meta)
+	if err != nil {
+		return err
+	}
+
+	if err := tpgresource.SetDataSourceLabels(d); err != nil {
+		return err
+	}
```
melinath, line 59: This is not necessary because this data source doesn't have a labels field.

```suggestion
```

melinath (changes requested):

mmv1/third_party/terraform/services/composer/data_source_google_composer_user_workloads_secret_test.go.erb
spapi17: The API is returning masked values in "data" field. But they are still stored in state in case it was set in Terraform configuration.

mmv1/third_party/terraform/services/composer/data_source_google_composer_user_workloads_secret.go.erb
melinath: What is the use case for this data source if the secrets are masked? Most similar data sources are useful because they provide access to the real secret value, which is then referenced from other resources. For example, see: https://registry.terraform.io/providers/hashicorp/google/latest/docs/data-sources/secret_manager_secret_version_access

melinath (commented):
Leaving a review to remove from my queue pending resolution of this discussion: https://github.com/GoogleCloudPlatform/magic-modules/pull/10615#discussion_r1617733669

melinath:
This test failure keeps happening so it's probably "real" in the sense that it would happen to anyone, but it's probably still unrelated to your change. Feel free to ignore it unless you think it _is_ related somehow

mmv1/third_party/terraform/services/composer/data_source_google_composer_user_workloads_secret.go.erb
spapi17: Only values in "data" filed's key-value pairs are masked. I think it would still be useful to manage this StringMap (update values for existing keys, add new key-value pairs and so on).

spapi17:
> This test failure keeps happening so it's probably "real" in the sense that it would happen to anyone, but it's probably still unrelated to your change. Feel free to ignore it unless you think it _is_ related somehow

I don't think it is related

mmv1/third_party/terraform/services/composer/data_source_google_composer_user_workloads_secret.go.erb
melinath: This call is specifically to get the "data" field, but the values are all masked. Users won't be able to update the fields using the data source, since it's read-only. Could you elaborate?

spapi17: Right, I missed the fact that it is read-only. The remaining use case is reading keys of "data" field in a secret that is defined outside of Terraform.

mmv1/third_party/terraform/services/composer/data_source_google_composer_user_workloads_secret_test.go.erb
```diff
@@ -0,0 +1,100 @@
...
+			if strings.HasPrefix(k, "data.") && k != "data.%" {
+				if dsAttr[k] == "**********" {
+					continue
+				}
+			}
+			if dsAttr[k] != rsAttr[k] {
+				errMsg += fmt.Sprintf("%s is %s; want %s\n", k, dsAttr[k], rsAttr[k])
+			}
```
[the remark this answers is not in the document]
spapi17: Added error message for missing key in datasource. Also removed `k != "data.%` as this case is filtered above.

melinath (approved):
LGTM

PR #10624, opened 2024-05-07
Update instances of `google_compute_forwarding_rule` in tests and examples to avoid setting `ip_address` with resource ids

SarahFrench:
The Compute API alters the value of `ip_address` when making forwarding rule resources. The `ip_address` value is always returned as an IP address number but the input from the user can be an IP address number or either a resource id or self link.

To discourage users from encountering this situation, where their config and the API have shared ownership over a field's value, I'm updating as many instances of this problem as possible. This will reduce the likelihood of users copying examples that trigger the problem with the API.

```release-note:none

```


zli82016 (approved):
LGTM. Thanks.

PR #10626, opened 2024-05-07
Add example to GCS bucket that explains the need for `no_age` 

NickElliot:
Stopgap documentation update for b/339089840 and https://github.com/hashicorp/terraform-provider-google/issues/17990

```release-note:none

```

### DIFF
--- a/mmv1/third_party/terraform/website/docs/r/storage_bucket.html.markdown
+++ b/mmv1/third_party/terraform/website/docs/r/storage_bucket.html.markdown
@@ -69,6 +69,29 @@ resource "google_storage_bucket" "auto-expire" {
 }
 ```
 
+## Example Usage - Life cycle settings for storage bucket objects with `no_age` enabled
+When creating a life cycle condition that does not also include an `age` field, it's important to include the `no_age` flag so your new bucket does not unintentionally create an extreanous condition that may interact incorrectly with your other options.
+
+```hcl
+resource "google_storage_bucket" "no-age-enabled" {
+  provider = google-beta
+  name          = "somename-test"
+  location      = "US"
+  force_destroy = true
+
+  lifecycle_rule {
+    action {
+      type = "Delete"
+    }
+    condition {
+      days_since_noncurrent_time = 3
+      no_age = true
+    }
+  }
+}
+```
+
+
 ## Example Usage - Enabling public access prevention
 
 ```hcl
--- a/mmv1/third_party/terraform/website/docs/r/storage_bucket.html.markdown
+++ b/mmv1/third_party/terraform/website/docs/r/storage_bucket.html.markdown
@@ -91,7 +91,6 @@ resource "google_storage_bucket" "no-age-enabled" {
 }
 ```
 
-
 ## Example Usage - Enabling public access prevention
 
 ```hcl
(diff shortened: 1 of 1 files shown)


mmv1/third_party/terraform/website/docs/r/storage_bucket.html.markdown
```diff
@@ -69,6 +69,28 @@ resource "google_storage_bucket" "auto-expire" {
...
+      days_since_noncurrent_time = 3
+      no_age = true
+    }
+  }
+}
+```
+
 ## Example Usage - Enabling public access prevention
 
 ```hcl
```
rileykarson, line 96: Can we also update the documentation for `age` below to explicitly mention the default of `0`? "In a lifecycle_rule block where this is not set, it defaults to 0 unless no_age is set to true" or similar

Actually, let's make the other fields explicit about zero values as well. "Due to a bug, this cannot currently be set to 0. When set to that value in configuration, it will be ignored and treated as if there was no condition instead."

mmv1/third_party/terraform/website/docs/r/storage_bucket.html.markdown
```diff
@@ -69,6 +69,28 @@ resource "google_storage_bucket" "auto-expire" {
 }
 ```
 
+## Example Usage - Life cycle settings for storage bucket objects with `no_age` enabled
+When creating a life cycle condition that does not also include an `age` field, it's important to include the `no_age` flag so your new bucket does not unintentionally create an extreanous condition that may interact incorrectly with your other options.
```
rileykarson, line 73:
```suggestion
When creating a lifecycle rule that does not include an `age` setting, a default `age` of `0` is set. Set `no_age` to `true` to avoid this. 
```

rileykarson: nit: Let's be more direct about why!

NickElliot: Wasn't sure how direct about the situation we wanted to be, added in much more direct details!

mmv1/third_party/terraform/website/docs/r/storage_bucket.html.markdown
```diff
@@ -149,9 +171,9 @@ The following arguments are supported:
 
 <a name="nested_condition"></a>The `condition` block supports the following elements, and requires at least one to be defined. If you specify multiple conditions in a rule, an object has to match all of the conditions for the action to be taken:
 
-* `age` - (Optional) Minimum age of an object in days to satisfy this condition.
+* `age` - (Optional) Minimum age of an object in days to satisfy this condition. If not supplied alongside another condition and without setting `no_age` to `true`, a default `age` of 0 will be set. 
 
-* `no_age` - (Optional) While set `true`, `age` value will be omitted. **Note** Required to set `true` when `age` is unset in the config file.  
+* `no_age` - (Optional) While set `true`, `age` value will be omitted. **Note** Required to set `true` when `age` is unset in the config file.  If ommitted alongside other numeric conditions set to `0` this can lead to a state where your bucket will automatically execute your configured Lifecycle Rule immediately.
```
rileykarson, line 176:
```suggestion
* `no_age` - (Optional) While set `true`, `age` value will be omitted from requests. This prevents a default age of `0` from being applied, and if you do not have an `age` value set, setting this to `true` is strongly recommended. When unset and other conditions are set to zero values, this can result in a rule that applies your action to all files in the bucket.
```

rileykarson (approved):

PR #9767, opened 2024-01-06
Add `vertex_ai_model_resource` :upload 

BBBmau:
Fixes https://github.com/hashicorp/terraform-provider-google/issues/15303

**Release Note Template for Downstream PRs (will be copied)**

```release-note:new-resource
`google_vertex_ai_model`
```

PR #10628, opened 2024-05-08
Add `vertex_ai_model_resource` :copy

BBBmau:
As the `vertex_ai_models` work continues to get larger, rather than treat it as one massive PR the goal is to treat each section of what comes with creating an ai model in google cloud. One of them being `:copy` and the other being `:upload` which is worked on more in this PR #9767 which will take much more time due to the different variables that come with creating a model (Tabular Models, image recognition models, speech models, etc.)

With `:copy` being completed with the ability to create and delete copied models within a project, what's left is creating tests. 

Simplest tfconfig to copy an already existing model in your project
```
resource "google_vertex_ai_models" "copy" {
	project = "hc-17ccbe3c280c434a87e301c7c95"
	source_model = "projects/hc-17ccbe3c280c434a87e301c7c95/locations/us-central1/models/5460298988349554688"
	region = "us-central1"
}
```

```release-note:new-resource
vertex_ai: added `vertex_ai_model_resource`
```


rileykarson:

BBBmau:
@rileykarson This PR can be reopened as I'll be finishing up the work for this.

@c2thorn can the script that converts `Ruby code to Go code` be applied here?

mmv1/products/vertexai/Models.yaml
```diff
@@ -0,0 +1,444 @@
...
+          name: id
+          output: true
+          description: |
+            The ID of the export format.
+        - type: Enum
+          name: exportableContent
+          immutable: true
+          description: |
+            The content of this Model that may be exported.
+          values:
```
SarahFrench, line 144:
```suggestion
          enum_values:
```

mmv1/products/vertexai/Models.yaml
```diff
@@ -0,0 +1,444 @@
...
+      - false
+  error:
+    path: 'error'
+    message: 'message'
+custom_code:
+  decoder: templates/terraform/decoders/vertex_ai_models.go.erb
+  pre_create: templates/terraform/pre_create/vertex_ai_models.go.erb
+examples:
+  - name: 'vertex_ai_model_source_basic'
+    skip_import_test: true
```
SarahFrench, line 55:
```suggestion
    exclude_import_test: true
```

SarahFrench (commented):
There are a few errors impacting the build - if you build the provider locally from this PR the error messages should tell you all you need to know. I've left some suggestions that help with some of the issues

mmv1/products/vertexai/Models.yaml
BBBmau: latest commit fixes the errors from running magic modules provider generation.

mmv1/templates/terraform/examples/vertex_ai_model_source_basic.tf.tmpl
```diff
@@ -0,0 +1,7 @@
+resource "google_vertex_ai_models" "{{$.PrimaryResourceId}}" {
+  display_name = "{{index $.Vars "display_name"}}"
+  project = "{{index $.Vars "project_name"}}"
+  source_model = "projects/{{index $.Vars "project_name"}}>/locations/us-central1/models/4469245519628878651"
```
SarahFrench, line 4: Is this model only present in our HashiCorp-owned testing project? If so, this test will eventually fail when running in automation on this PR as the source model will be inaccessible.

Once we get the test passing using the HashiCorp testing project we could make the acceptance test handwritten and try to navigate ensuring how the project used for testing has a model present at the start of the test.

SarahFrench (commented):
Here's a first pass review - there are some problems generating the acceptance test and once those problems are addressed there are issues with the fields being used for that create method.

Also there's a question about accessing the existing model in the acc test- for now we should iterate on the test to make it work in our GCP project and then we can navigate ensuring the Google-owned projects have a model present in them for the test to use.

BBBmau:
> @BBBmau As follow up to my last review, I've got some questions:
> 
>     1. Are we going to use a feature branch for this work or not? Either choice is fine but I don't think that was firmly decided either way.
> 
>     2. Is the 'copy' PR being done before the 'upload' PR?
> 
> 
> Re: number 2, could you please mark whichever PR should be done first as non-draft and ready for review? We should then ensure that the PRs only contains code relevant to that creation method (e.g. the pre_create in this PR includes code related to create via upload)

Yes the idea is to make `copy` as the base for adding support to the model resource. I'll be addressing these and get to work on separating `upload` into their appropriate PRs

BBBmau:
Closing in favor of https://github.com/GoogleCloudPlatform/magic-modules/pull/12074

PR #10629, opened 2024-05-08
docs: warn message on google_project_iam_binding

hgsgtk:
!-- Put a description of what this PR is for here, along with any references to issues that this resolves or contributes to -->

When a user wants to manage IAM users' role via Terraform, using the `google_project_iam_binding`` has a risk of breaking existing infrastructure because it can affect the resource outside of Terraform, although the official doc does not warn it.

Therefore, I added a warning message to the official doc.

It resolves the below issue:
hashicorp/terraform-provider-google#18042.

```release-note:none

```

### DIFF
--- a/mmv1/third_party/terraform/website/docs/r/google_project_iam.html.markdown
+++ b/mmv1/third_party/terraform/website/docs/r/google_project_iam.html.markdown
@@ -74,6 +74,13 @@ data "google_iam_policy" "admin" {
 
 ## google\_project\_iam\_binding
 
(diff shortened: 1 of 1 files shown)


rileykarson:
This is redundant with the warnings at the top of the page- this resource is authoritative over a role, and we spell out that fact in a couple different ways. Admittedly, it's not always effective- but I'd prefer we don't layer on more redundant messages as that increases the odds folks will read _none_ of them.

PR #15, opened 2018-03-07
Merge in The Magician

ndmckinley:
This PR merges in the full history of the `terraform-provider-google-ci` repo into a new `.ci` folder.  `terraform-provider-google-ci` recently evolved to be The Magician, a system for managing downstream pull requests on Magic Modules.  It's getting to be time to start managing that as part of this repository instead of separately.

This PR is a pure merge - it doesn't change the workflow to depend on `magic-modules` instead of `terraform-provider-google-ci` or make any of the other minor changes which will be necessary - those will come in a followup, once the transition is all done.

All the code in here has undergone review, (this is `terraform-provider-google-ci/master`), so **you don't need to review the code**.  The request is to review the merge instead.

I'd like this to be merged in with a merge strategy instead of a squash-and-merge.  First, the history is likely to be useful for future development.  Second, I've done a subtree merge here, so the few remaining PRs open against `terraform-provider-google-ci` can be resolved there and then merged into this repo, in parallel with new development being done here.   A merge instead of squash-and-merge makes that transition possible.  Third, the usual reason for doing a squash-and-merge, easy rollbacks, doesn't apply here because this code has no interactions with the existing code.

PR #10632, opened 2024-05-08
Introduce endpoint global access field in endpoint attachment resource

anjalisoni3655:
Adding support for endpoint global access field Endpoint Attachment resource of Integration connectors.

`google_integration_connectors_endpoint_attachment`

```release-note:enhancement
integrationconnectors: added `endpoint_global_access` field to `google_integration_connectors_endpoint_attachment` resource
```
Fixes hashicorp/terraform-provider-google#18030

### DIFF
--- a/mmv1/products/integrationconnectors/EndpointAttachment.yaml
+++ b/mmv1/products/integrationconnectors/EndpointAttachment.yaml
@@ -105,3 +105,7 @@ properties:
     description: |
       The Private Service Connect connection endpoint ip.
     output: true
+  - !ruby/object:Api::Type::boolean
+    name: "endpointGlobalAccess"
+    description: |
+      Enable global access for endpoint attachment.
--- a/mmv1/products/integrationconnectors/EndpointAttachment.yaml
+++ b/mmv1/products/integrationconnectors/EndpointAttachment.yaml
@@ -105,7 +105,7 @@ properties:
     description: |
       The Private Service Connect connection endpoint ip.
     output: true
-  - !ruby/object:Api::Type::boolean
+  - !ruby/object:Api::Type::Boolean
     name: "endpointGlobalAccess"
     description: |
       Enable global access for endpoint attachment.
--- a/mmv1/templates/terraform/examples/integration_connectors_endpoint_attachment.tf.erb
+++ b/mmv1/templates/terraform/examples/integration_connectors_endpoint_attachment.tf.erb
@@ -6,4 +6,5 @@ resource "google_integration_connectors_endpoint_attachment" "<%= ctx[:primary_r
   labels = {
     foo = "bar"
   }
+  endpoint_global_access = false
 }
--- a/mmv1/templates/terraform/examples/integration_connectors_endpoint_attachment.tf.erb
+++ b/mmv1/templates/terraform/examples/integration_connectors_endpoint_attachment.tf.erb
@@ -6,5 +6,4 @@ resource "google_integration_connectors_endpoint_attachment" "<%= ctx[:primary_r
   labels = {
     foo = "bar"
   }
-  endpoint_global_access = false
 }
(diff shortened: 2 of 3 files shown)


mmv1/products/integrationconnectors/EndpointAttachment.yaml
```diff
@@ -105,3 +105,7 @@ properties:
     description: |
       The Private Service Connect connection endpoint ip.
     output: true
+  - !ruby/object:Api::Type::boolean
+    name: "endpointGlobalAccess"
```
balanaguharsha, line 109: Does this need to be required? Is the default value false?

mmv1/templates/terraform/examples/integration_connectors_endpoint_attachment.tf.erb
```diff
@@ -6,4 +6,5 @@ resource "google_integration_connectors_endpoint_attachment" "<%= ctx[:primary_r
   labels = {
     foo = "bar"
   }
+  endpoint_global_access = false
```
roaks3, line 9: Since you mentioned the default is `false`, can this line be removed from the example?

mmv1/third_party/terraform/services/integrationconnectors/resource_integration_connectors_endpoint_attachment_test.go
```diff
@@ -74,6 +75,7 @@ resource "google_integration_connectors_endpoint_attachment" "sampleendpointatta
   labels = {
     bar = "foo"
   }
+  endpoint_global_access = false
```
roaks3, line 78: I would suggest changing this to `true`, to verify that a value other than false works, and to verify that the value can be updated in place. If it can't be updated in-place, then instead it should be configured in the yaml with `immutable: true`

roaks3 (commented):
You will also need to reformat your PR description to include a proper release note section: https://github.com/GoogleCloudPlatform/magic-modules/actions/runs/8999322060/job/24721244891?pr=10632. These are collected and used in the release notes we publish for each new release.

mmv1/templates/terraform/examples/integration_connectors_endpoint_attachment.tf.erb
anjalisoni3655: Done

mmv1/third_party/terraform/services/integrationconnectors/resource_integration_connectors_endpoint_attachment_test.go
anjalisoni3655: Done

roaks3:
The build error is:

```
Step #15 - "tf-oics-head": /usr/local/lib/ruby/3.1.0/psych/class_loader.rb:99:in `find': Tried to load unspecified class: Api::Type::boolean (Psych::DisallowedClass)
```

I think you just need to capitalize `Boolean`

anjalisoni3655:
> The build error is:
> 
> ```
> Step #15 - "tf-oics-head": /usr/local/lib/ruby/3.1.0/psych/class_loader.rb:99:in `find': Tried to load unspecified class: Api::Type::boolean (Psych::DisallowedClass)
> ```
> 
> I think you just need to capitalize `Boolean`

Fixed this

roaks3 (changes requested):
Looks like some sort of internal issue is causing the test to fail. If this ends up being an allowlist issue, or some sort of environmental circumstance we need to adjust in our test projects, just let me know and I can look into making that change.

anjalisoni3655:
> Looks like some sort of internal issue is causing the test to fail. If this ends up being an allowlist issue, or some sort of environmental circumstance we need to adjust in our test projects, just let me know and I can look into making that change.

I ran the tests locally and they were successful, not sure why the test cases status is still in progress here.

roaks3:
Ah I see, this looks like perhaps something that resolved itself, or is flakey.

Note for any future readers that might see this test fail: we did see an initial run fail that we couldn't explain.

roaks3 (approved):

PR #10634, opened 2024-05-08
Add external_providers to replace pull_external in example object

zli82016:
fixes https://github.com/hashicorp/terraform-provider-google/issues/15312

1. Add field `external_providers` to example object, so the specific external providers will be added to the generated tests. Only official Hashicorp providers are allowed to put into external_providers.
2. Iterate `external_providers` in test templates
3. Replace `pull_external` with `external_providers` in YAML files
4. `pull_external` will be cleaned up in a couple weeks.

```release-note:none

```


c2thorn (approved):
nicely done

PR #10635, opened 2024-05-08
Add support for `google_compute_node_group` to TGC

delimaneto:
adding support for compute.googleapis.com/NodeGroup
```release-note:none
```


melinath (changes requested):
Since this is a TGC-only change it should use release note type `none` and the content of the release note should be empty. Release notes are only used for provider changes.

sushant00:
@melinath I think we would need to merge https://github.com/GoogleCloudPlatform/magic-modules/pull/10602 first for NodeTemplate before we can merge this PR for NodeGroup.

Can you please confirm?

melinath (changes requested):
TGC converters shouldn't technically need to depend on each other (in most cases) but if there's a separate PR for node template then we should remove that from here for sure, and we could merge them in order if you prefer.

delimaneto:
> TGC converters shouldn't technically need to depend on each other (in most cases) but if there's a separate PR for node template then we should remove that from here for sure, and we could merge them in order if you prefer.

I just did let NodeTemplate here because is a resource dependency for NodeGroup resource. I think we should merge the other PR about NodeTemplate and after merge here. Make sense?

melinath (commented):
sounds good, removing from my queue for now - let me know when node template is done & I'll review this at that point.

delimaneto:
node template is done, thanks.

mmv1/templates/tgc/resource_converters.go.erb
```diff
@@ -92,6 +92,8 @@ func ResourceConverters() map[string][]cai.ResourceConverter {
 		"google_storage_bucket_iam_policy":                {resourceConverterStorageBucketIamPolicy()},
 		"google_storage_bucket_iam_binding":               {resourceConverterStorageBucketIamBinding()},
 		"google_storage_bucket_iam_member":                {resourceConverterStorageBucketIamMember()},
+		"google_compute_node_group":				   	           {compute.ResourceConverterComputeNodeGroup()},
+		"google_compute_node_template": 				           {compute.ResourceConverterComputeNodeTemplate()},
```
melinath, line 96:
```suggestion
```

mmv1/templates/tgc/resource_converters.go.erb
```diff
@@ -80,7 +80,7 @@ func ResourceConverters() map[string][]cai.ResourceConverter {
 		"google_spanner_instance_iam_binding":             {spanner.ResourceConverterSpannerInstanceIamBinding()},
 		"google_spanner_instance_iam_member":              {spanner.ResourceConverterSpannerInstanceIamMember()},
 		"google_project_service":                          {resourceConverterServiceUsage()},
-		"google_secret_manager_secret_version": 		   {secretmanager.ResourceConverterSecretManagerSecretVersion()},
+		"google_secret_manager_secret_version": 		       {secretmanager.ResourceConverterSecretManagerSecretVersion()},
```
melinath, line 83:
```suggestion
		"google_secret_manager_secret_version": 		   {secretmanager.ResourceConverterSecretManagerSecretVersion()},
```

melinath (changes requested):

mmv1/templates/tgc/resource_converters.go.erb
```diff
@@ -92,7 +92,9 @@ func ResourceConverters() map[string][]cai.ResourceConverter {
 		"google_storage_bucket_iam_policy":                {resourceConverterStorageBucketIamPolicy()},
 		"google_storage_bucket_iam_binding":               {resourceConverterStorageBucketIamBinding()},
 		"google_storage_bucket_iam_member":                {resourceConverterStorageBucketIamMember()},
-    "google_cloud_tasks_queue":                        {cloudtasks.ResourceConverterCloudTasksQueue()},
+		"google_compute_node_group":                       {compute.ResourceConverterComputeNodeGroup()},
+		"google_compute_node_template":                    {compute.ResourceConverterComputeNodeTemplate()},
```
melinath, line 96: This is still adding node_template. Please remove this line.

melinath (changes requested):
Please undo the whitespace changes in resource_converters (or do them in a separate PR) - that's going to make this keep running into merge conflicts.

melinath:
Also please resolve the merge conflicts

mmv1/templates/tgc/resource_converters.go.erb
delimaneto: if I remove google_compute_node_template, it will not test node template resource anymore. I can't remove it.

melinath: google_compute_node_template support was added in https://github.com/GoogleCloudPlatform/magic-modules/pull/10602, so it doesn't also need to be added here.

melinath (changes requested):

mmv1/templates/tgc/resource_converters.go.erb
delimaneto: ok, done. Thanks

mmv1/templates/tgc/resource_converters.go.erb
```diff
@@ -46,18 +46,18 @@ func ResourceConverters() map[string][]cai.ResourceConverter {
 		"google_compute_health_check":           {compute.ResourceConverterComputeHealthCheck()},
 		"google_compute_instance":                         {compute.ResourceConverterComputeInstance()},
 		"google_compute_network":                          {compute.ResourceConverterComputeNetwork()},
-		"google_compute_node_template": 				           {compute.ResourceConverterComputeNodeTemplate()},
+		"google_compute_resource_policy":                  {compute.ResourceConverterComputeResourcePolicy()},
```
melinath, line 49: please undo these stray changes.

melinath (changes requested):

mmv1/templates/tgc/resource_converters.go.erb
delimaneto: done and aligned lines correctly

melinath (changes requested):
In general, if you're making a bunch of changes to the same file (for example, to fix formatting or whitespace issues), you should make the changes in a separate PR either before you start or after you finish to avoid causing merge conflicts.

I went ahead and merged your other PR that made whitespace changes so that they're done for now. However, that means this PR now has merge conflicts that need to be resolved. But moving forward this will hopefully not be a problem again.

melinath (changes requested):
why are there more whitespace changes? Also, in addition to explaining why, please remove them.

delimaneto:
> why are there more whitespace changes? Also, in addition to explaining why, please remove them.

There are more whitespace changes because I'm removing those on edit file github. I'm resolving the conflicts and sometimes it comes from main branch and I need to fix all of them.

melinath (changes requested):
Please undo the whitespace changes

delimaneto:
> Please undo the whitespace changes

done, thanks

melinath (changes requested):
Taking a look at this, it's only exercising a few of the fields on node_group. Could you add a more complete set of fields? As many as possible, ideally. https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/compute_node_group lists the available fields.

delimaneto:
> Taking a look at this, it's only exercising a few of the fields on node_group. Could you add a more complete set of fields? As many as possible, ideally. https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/compute_node_group lists the available fields.

@melinath we are using the most simple group of fields that resource can once those are just to test the specific resource. As much more simpler, it'll be better as internal team said.

melinath (changes requested):
the only other resource that could be referenced from a node group is a project, which doesn't seem like an excessive thing to add to this configuration. Please add all fields to the test. Thanks!

delimaneto:
> the only other resource that could be referenced from a node group is a project, which doesn't seem like an excessive thing to add to this configuration. Please add all fields to the test. Thanks!

added more fields, thanks.

sushant00 (approved):

delimaneto:
@melinath could you please review again?

melinath (approved):
LGTM assuming tests pass

delimaneto:
@melinath could you please make the merge?

melinath:
yep, thanks for the ping!

PR #10636, opened 2024-05-08
Update google_project_service docs

rileykarson:
Prompted by yaqs/6237359040070942720, let's recommend that users do *not* disable services by default and make our example reflect that. Also use the formatting from https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/container_cluster for resources with many reference pages.

Also introduce "Google Cloud project" terminology, that's what pages like https://cloud.google.com/resource-manager/docs/creating-managing-projects use these days.

```release-note:none
```

### DIFF
--- a/mmv1/third_party/terraform/website/docs/r/google_project_service.html.markdown
+++ b/mmv1/third_party/terraform/website/docs/r/google_project_service.html.markdown
@@ -34,7 +34,7 @@ resource "google_project_service" "project" {
     update = "40m"
   }
 
-  disable_on_destroy = true
+  disable_on_destroy = false
 }
 ```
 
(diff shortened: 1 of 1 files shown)


mmv1/third_party/terraform/website/docs/r/google_project_service.html.markdown
```diff
@@ -45,15 +47,16 @@ The following arguments are supported:
 * `project` - (Optional) The project ID. If not provided, the provider project
 is used.
 
+* `disable_on_destroy` - (Optional) If `true` or unset, disable the service when the
+Terraform resource is destroyed. If `false`, the service will be left enabled when
+the Terraform resource is destroyed. Defaults to `true`. Most configurations should
```
c2thorn, line 52: If most configurations should set this to false, should the default be changed in the future?

c2thorn (approved):

mmv1/third_party/terraform/website/docs/r/google_project_service.html.markdown
rileykarson: We could consider it. Filed https://github.com/hashicorp/terraform-provider-google/issues/18058

PR #10639, opened 2024-05-08
Support ignore_read on nested fields 

SirGitsalot:
This changes the compiler to generate a flattener for nested fields automatically in the form that resource authors have been writing manually, a la:

```
func flattenFooBar(v interface{}, d *schema.ResourceData, config *transport_tpg.Config) interface{} {
    return d.Get("foo.0.bar")
}
```

This change also automatically adds nested `ignore_read` to `ImportStateVerifyIgnore`. This will generate a bunch of test diffs because `ImportStateVerifyIgnore` is reordered in some cases, and in others it's getting additional `ignore_read` fields (which presumably aren't necessary for the tests to pass, but the compiler doesn't know that).

It also updates `google_compute_backend_service` and `google_compute_region_backend_service` to verify that the same code is being generated with their no-longer-needed custom flatteners and `ignore_read_extra` configs.

Fixes https://github.com/hashicorp/terraform-provider-google/issues/12410

```release-note:none

```


mmv1/templates/terraform/examples/base_configs/test_file.go.erb
```diff
@@ -43,10 +43,11 @@ object.examples
     test_slug = "#{resource_name}_#{example.name.camelize(:lower)}Example"
 
     ignore_read = object.all_user_properties
```
rileykarson, line 45: nit: Could we add a sort in a separate PR and then rebase this change to include it? That'll capture the true delta from your change- it's hard to see what has/hasn't changed as-is! That'll guard against spurious diffs during other similar changes to the set in the future.

rileykarson (approved):
LGTM- Changes are test-only and I didn't see anything that looked off- but also didn't step through every delta piece-by-piece.

rileykarson (changes requested):
Oh- I think that is actually a regression in the resource as it has changed!

SirGitsalot:
I merged https://github.com/GoogleCloudPlatform/magic-modules/pull/10648 so there should be fewer diffs.

The newly failing test is indeed seeing a diff; the previous flattener for `resource_vertex_ai_index`'s `metadata.0.config.0.feature_norm_type` was:

```
func flattenVertexAIIndexMetadataConfigFeatureNormType(v interface{}, d *schema.ResourceData, config *transport_tpg.Config) interface{} {
    return v
}
```

And now it's:

```
func flattenVertexAIIndexMetadataConfigFeatureNormType(v interface{}, d *schema.ResourceData, config *transport_tpg.Config) interface{} {
    return d.Get("metadata.0.config.0.feature_norm_type")
}
```

The two failing tests are now getting a value where for that field that they didn't see before. I'm not sure what the correct fix is though - I could add a custom flattener to retain the previous behavior for this field (ironic for a PR whose purpose is to reduce the need for custom flatteners!).

rileykarson (approved):

melinath:
could you also update the contribution docs? https://googlecloudplatform.github.io/magic-modules/develop/field-reference/#ignore_read

PR #10641, opened 2024-05-09
Update suppress func for compute_disk.type and compute_instance.machineType.

harshulsoni:
Update supress func for compute_disk.type and compute_instance.machineType to account for type URLs. These URLs were treated differently due to terraform import and plan which caused forced destroy and creation of new resources. This pull request updates the suppress func so only resource names are used for terraform import and plan.

Fixes: https://github.com/hashicorp/terraform-provider-google/issues/10428

```release-note:bug
compute: update supress func for compute_disk.type and compute_instance.machineType to account for type URLs.
```

### DIFF
--- a/mmv1/products/compute/Disk.yaml
+++ b/mmv1/products/compute/Disk.yaml
@@ -393,6 +393,7 @@ properties:
     name: 'type'
     resource: 'DiskType'
     imports: 'selfLink'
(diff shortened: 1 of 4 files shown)


roaks3 (approved):
LGTM, and tests look sufficient

PR #10642, opened 2024-05-09
Add vacation for SarahFrench in May and July

SarahFrench:

```release-note:none

```

### DIFF
--- a/.ci/magician/github/membership.go
+++ b/.ci/magician/github/membership.go
@@ -77,21 +77,6 @@ var (
 			startDate: newDate(2024, 5, 10, bstLoc),
 			endDate:   newDate(2024, 5, 12, bstLoc),
 		},
-		{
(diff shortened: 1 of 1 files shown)


melinath (changes requested):
I think we don't currently handle multiple vacations per person - could you double-check that it will work as expected? Maybe add some unit tests?

We could consider doing this by making each user's block support multiple start/end dates

SarahFrench:
I can do that in future but I'm OOO tomorrow & I'm at the end of my day, so I'll update this PR to just reflect my upcoming PTO

melinath (approved):

melinath:
> I can do that in future but I'm OOO tomorrow & I'm at the end of my day, so I'll update this PR to just reflect my upcoming PTO

makes sense!

PR #10466, opened 2024-04-17
Add resource for userWorkloadsSecret in Composer environment.

spapi17:
**Release Note Template for Downstream PRs (will be copied)**

```release-note:enhancement
composer: Add resource for userWorkloadsSecret in Composer environment. Includes tests and documentation for this resource. Adding in 'beta' version for now.
```

PR #10644, opened 2024-05-09
corrections in google_composer_user_workloads_secret

spapi17:

```release-note:enhancement
composer: some corrections in google_composer_user_workloads_secret including changes in description, operation timeouts and so on.
```

### DIFF
--- a/mmv1/third_party/terraform/services/composer/resource_composer_user_workloads_secret.go.erb
+++ b/mmv1/third_party/terraform/services/composer/resource_composer_user_workloads_secret.go.erb
@@ -206,7 +206,7 @@ func resourceComposerUserWorkloadsSecretDelete(d *schema.ResourceData, meta inte
 
 func resourceComposerUserWorkloadsSecretImport(d *schema.ResourceData, meta interface{}) ([]*schema.ResourceData, error) {
 	config := meta.(*transport_tpg.Config)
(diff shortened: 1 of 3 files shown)


mmv1/third_party/terraform/services/composer/resource_composer_user_workloads_secret.go.erb
```diff
@@ -32,9 +32,9 @@ func ResourceComposerUserWorkloadsSecret() *schema.Resource {
 		},
 
 		Timeouts: &schema.ResourceTimeout{
-			Create: schema.DefaultTimeout(120 * time.Minute),
-			Update: schema.DefaultTimeout(120 * time.Minute),
-			Delete: schema.DefaultTimeout(30 * time.Minute),
+			Create: schema.DefaultTimeout(time.Minute),
```
BBBmau, line 35: what's the reason for changing the timeout time? Is this time change documented somewhere

spapi17: This is a follow up to https://github.com/GoogleCloudPlatform/magic-modules/pull/10466. Since these operations are very fast (taking less than a second on average), the previous timeout value was unnecessarily long.

BBBmau (approved):

PR #10645, opened 2024-05-09
Check if property type has prefix KeyValue in Go templates

zli82016:
1. Check the property types if they starts with `KeyValue` to add element (`Elem: &schema.Schema{Type: schema.TypeString},`) to the map.  In Ruby code, `KeyValuePairs` are the parent class of other `KeyValue` classes. In Go code, the inheritance relationship is lost as the property type changes to string from an object in Ruby. Go doesn't support the same inheritance as Ruby. 
2. Trim the note added to labels/annotations fields by the system (not the user) when converting Ruby YAML files to Go YAML files.

```release-note:none

```


c2thorn (approved):

zli82016:
cc @NickElliot

PR #10647, opened 2024-05-09
Make downstream builder no-op when commit already built

trodge:
fixes https://github.com/hashicorp/terraform-provider-google/issues/12360

```release-note:none

```

### DIFF
--- a/.ci/magician/cmd/generate_downstream.go
+++ b/.ci/magician/cmd/generate_downstream.go
@@ -98,6 +98,24 @@ func execGenerateDownstream(baseBranch, command, repo, version, ref string, gh G
 		baseBranch = "main"
 	}
 
+	var syncBranchPrefix string
+	if repo == "terraform" {
+		if version == "beta" {
+			syncBranchPrefix = "tpgb-sync"
+		} else if version == "ga" {
+			syncBranchPrefix = "tpg-sync"
+		}
+	} else if repo == "terraform-google-conversion" {
+		syncBranchPrefix = "tgc-sync"
+	} else if repo == "tf-oics" {
+		syncBranchPrefix = "tf-oics-sync"
+	}
+	syncBranch := getSyncBranch(syncBranchPrefix, baseBranch)
+	if syncBranchHasCommit(ref, syncBranch, rnr) {
+		fmt.Printf("Sync branch %s already has commit %s, skipping generation\n", syncBranch, ref)
+		os.Exit(0)
+	}
+
 	mmLocalPath := filepath.Join(rnr.GetCWD(), "..", "..")
 	mmCopyPath := filepath.Join(mmLocalPath, "..", fmt.Sprintf("mm-%s-%s-%s", repo, version, command))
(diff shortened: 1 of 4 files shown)


melinath (changes requested):
This is doing a bunch of things at once, which makes it feel significantly riskier. Could you split it out into a few stages:

- switch to Go for `sync` jobs
- switch to Go for `push` jobs
- (once both are using go) change behavior to no-op across the board if the commit was already handled

rileykarson:
Do we plan to move forward with this PR? @trodge can you assign yourself to https://github.com/hashicorp/terraform-provider-google/issues/12360 if so?

.ci/magician/cmd/generate_downstream.go
```diff
@@ -98,6 +98,24 @@ func execGenerateDownstream(baseBranch, command, repo, version, ref string, gh G
...
+		}
+	} else if repo == "terraform-google-conversion" {
+		syncBranchPrefix = "tgc-sync"
+	} else if repo == "tf-oics" {
+		syncBranchPrefix = "tf-oics-sync"
+	}
+	syncBranch := getSyncBranch(syncBranchPrefix, baseBranch)
+	if syncBranchHasCommit(ref, syncBranch, rnr) {
+		fmt.Printf("Sync branch %s already has commit %s, skipping generation\n", syncBranch, ref)
+		os.Exit(0)
```
melinath, line 116:
```suggestion
		return nil
```

melinath (changes requested):
one small thing; otherwise LGTM (though I will also want to take time to remind myself what the plan here was because it's been a while.)

melinath:
@shuyama1 could you take a look at this to make sure there are no issues in terms of the changes you made in https://github.com/GoogleCloudPlatform/magic-modules/pull/10939?

shuyama1:
> @shuyama1 could you take a look at this to make sure there are no issues in terms of the changes you made in #10939?

If I read the code correctly, I'm not sure if this change will fix the issue that we'd like to resolve --  no manual cleanup of the sync branches is needed before clicking the `retry` button on the failed cloudbuild job. 

The actual failure of the build happens in the `sync` step or in `wait-for-commit` command to be clear (https://github.com/GoogleCloudPlatform/magic-modules/blob/main/.ci/magician/cmd/wait_for_commit.go#L50), which is the step before magician generates the actual commit and push it to the downstream. `sync-branch` command is used to push the given commit to the given sync branch, which happens after the actual push.

melinath (changes requested):
+1 to Shuya's point - good catch! This also needs to update wait_for_commit to be a no-op.

melinath:
@trodge could you take a look at this today?

trodge:
> @trodge could you take a look at this today?

Were there more changes needed than the one I most recently pushed?

melinath:
I thought there might still be more you were planning to review since you hadn't re-requested review yet.

melinath (approved):

PR #10650, opened 2024-05-09
added functionality in mmv1 to support configurable sweepers

shuyama1:
fixes https://github.com/hashicorp/terraform-provider-google/issues/16859

```release-note:none

```


mmv1/provider/terraform/sweeper.rb
```diff
@@ -0,0 +1,34 @@
...
+require 'uri'
+require 'api/object'
+require 'compile/core'
+require 'google/golang_utils'
+
+module Provider
+  class Terraform
+    # Support for schema ValidateFunc functionality.
+    class Sweeper < Google::YamlValidator
+      # The field checked by seeper to determine
```
c2thorn, line 23:
```suggestion
      # The field checked by sweeper to determine
```

mmv1/provider/terraform/sweeper.rb
```diff
@@ -0,0 +1,34 @@
...
+require 'compile/core'
+require 'google/golang_utils'
+
+module Provider
+  class Terraform
+    # Support for schema ValidateFunc functionality.
+    class Sweeper < Google::YamlValidator
+      # The field checked by sweeper to determine
+      # eligibility for deletion for generated resources
+      attr_reader :tf_test_field
```
c2thorn, line 25: I was a little confused at what this was at first... can you think of a better name for this? I can't think of anything perfect, but maybe `sweepable_name_field` or something?

c2thorn (approved):
LGTM, just have an opinion on attribute name

c2thorn:
As far as Go rewrite, we haven't gotten to sweepers yet, so this should be good to add right now.

mmv1/provider/terraform/sweeper.rb
shuyama1: Yea, I was also struggled to find a better name for it. I'll update it with a more clear name, `sweepable_name_field` seems a good one

shuyama1: change it to `sweepable_identifier_field`. let me know if that makes sense

c2thorn: a lot clearer thanks!

c2thorn (approved):

PR #10651, opened 2024-05-10
Run unit tests in tools and rename magician unit tests

shuyama1:

```release-note:none

```

### DIFF
--- a/.github/workflows/unit-test-tools.yml
+++ b/.github/workflows/unit-test-tools.yml
@@ -1,22 +1,24 @@
-name: membership-checker
+name: unit-tests-tools
 
 permissions: read-all
 
 on:
   pull_request:
     paths:
-      - '.ci/magician/**'
+      - 'tools/**'
 
 jobs:
-  build-and-unit-tests:
+  test:
     runs-on: ubuntu-22.04
     steps:
       - uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.2
+
       - name: Set up Go
         uses: actions/setup-go@0c52d547c9bc32b1aa3301fd7a9cb496313a4491 # v5.0.0
         with:
-          go-version: '^1.21'
-      - name: Run membership checker unit tests
+          go-version: '^1.21.0'
+
+      - name: Run unit tests in tools/
         run: |
-          cd .ci/magician
-          go test ./... -v
+          cd tools
+          go test -v ./...
--- a/.github/workflows/unit-test-magician.yml
+++ b/.github/workflows/unit-test-magician.yml
@@ -1,29 +1,22 @@
-name: unit-tests-go-changelog
+name: unit-test-magician
 
 permissions: read-all
 
 on:
   pull_request:
     paths:
-      - 'tools/go-changelog/**'
+      - '.ci/magician/**'
 
 jobs:
-  test:
+  build-and-unit-tests:
     runs-on: ubuntu-22.04
     steps:
       - uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.2
-
       - name: Set up Go
         uses: actions/setup-go@0c52d547c9bc32b1aa3301fd7a9cb496313a4491 # v5.0.0
         with:
-          go-version: '^1.21.0'
-      
-      - name: Build
+          go-version: '^1.21'
+      - name: Run magician unit tests
         run: |
-          cd tools/go-changelog/cmd/changelog-pr-body-check
-          go build
-
-      - name: Test
-        run: |
-          cd tools/go-changelog
-          go test -v ./...
+          cd .ci/magician
+          go test ./... -v


shuyama1:
I assume changing GHA workflow file name will not interrupt ongoing PRs. Please correct me if my assumption is wrong

.github/workflows/unit-test-tools.yml
```diff
@@ -1,12 +1,12 @@
-name: unit-tests-diff-processor
+name: unit-tests-tools
```
melinath, line 1: tests -> test for consistency

```suggestion
name: unit-test-tools
```

.github/workflows/unit-test-tools.yml
```diff
@@ -1,12 +1,12 @@
...
 
 permissions: read-all
 
 on:
   pull_request:
     paths:
-      - 'tools/diff-processor/**'
-      - '.github/workflows/unit-tests-diff-processor.yml'
+      - 'tools/**'
+      - '.github/workflows/unit-tests-tools.yml'
```
melinath, line 9:
```suggestion
      - '.github/workflows/unit-test-tools.yml'
```

.github/workflows/unit-test-tools.yml
```diff
@@ -19,28 +19,48 @@ jobs:
...
 
-      - name: Test with TPGB
+      - name: Test diff-processor with TPGB
         run: |
           cd tools/diff-processor
           go test -v ./...
         env:
           SERVICES_DIR: tools/diff-processor/new/google/services
+
+      - name: Build go-changelog
```
melinath, line 48: What would you think about doing these as separate jobs (instead of just separate steps?)

melinath (changes requested):
Would it be easy to also add test steps for the `template-check` tool?

.github/workflows/unit-test-tools.yml
shuyama1: Oh, like one GHA job per tool? Yeah, totally. I think I misunderstood the comment https://github.com/hashicorp/terraform-provider-google/issues/17598#issuecomment-2086555014 and thought we want to includes all unit test in tools directory into one job. I can split them up and also add one for `template-check`.

melinath: I _think_ that comment was wondering if we could just run `go test` once and get the entire directory (so that we don't need to remember to add new tools to the unit test action in the future) but I'm not sure that would actually work.

shuyama1: Unfortunately, it's not gonna work with running `go test` in the root. We'll need to have separate jobs or steps. Agree that having separate jobs may be clearer and more efficient for testing future tool changes in this case. I'll revert the changes for diff-processor and go-changlog but add new ones for [issue-labeler](https://github.com/GoogleCloudPlatform/magic-modules/tree/main/tools/issue-labeler) and [template-check](https://github.com/GoogleCloudPlatform/magic-modules/tree/main/tools/template-check) which are the two that contain unit tests but not running at this point.

melinath: (To clarify, I was only suggesting new jobs within this file, not necessarily new files - although new files are necessary if we only want to run the tests for the specific modified tools.)

shuyama1: oh, I see. That makes senses. Split them into separate jobs.

.ci/magician/cmd/check_cassettes.go
```diff
@@ -42,7 +42,7 @@ var checkCassettesCmd = &cobra.Command{
 	The following environment variables are expected:
 ` + listCCEnvironmentVariables() + `
 
-	It prints a list of tests that failed in replaying mode along with all test output.`,
+	It prints a list of tests that failed in replaying mode along with all test output. Add this line to trigger unit tests`,
```
melinath, line 45: to revert

.github/workflows/unit-test-tools.yml
```diff
@@ -0,0 +1,106 @@
...
+permissions: read-all
+
+on:
+  pull_request:
+    paths:
+      - 'tools/**'
+      - '.github/workflows/unit-test-tools.yml'
+
+jobs:
+  unit-test-diff-processor:
```
melinath, line 12: the `unit-test` part of these job names is redundant because of how these are displayed in the UI:

Screenshot 2024-05-16 at 9 52 36 AM


```suggestion
  diff-processor:
```

tools/diff-processor/cmd/breaking_changes.go
```diff
@@ -13,7 +13,7 @@ import (
 	"github.com/spf13/cobra"
 )
 
-const breakingChangesDesc = `Check for breaking changes between the new / old Terraform provider versions.`
+const breakingChangesDesc = `Check for breaking changes between the new / old Terraform provider versions. Add this line to trigger unit tests`
```
melinath, line 16: to revert

melinath (approved):

melinath (approved):

PR #10653, opened 2024-05-10
Add support for `google_compute_ssl_certificate` to TGC

delimaneto:
adding support for compute.googleapis.com/SslCertificate
```release-note:none
```


slevenick (changes requested):
Test failing:
            │ Error: Invalid function argument
            │ 
            │   on example_google_compute_ssl_certificate.tf line 17, in resource "google_compute_ssl_certificate" "webserver_cert":
            │   17:   certificate = file("/etc/apache2/ssl/apache.crt")

delimaneto:
[files.zip](https://github.com/GoogleCloudPlatform/magic-modules/files/15276493/files.zip)

Could you please add the files on the test directory? Or I think we could try adding those on the repository and update the path. What do you think?

slevenick:
In the providers we have a test directory that we can add this sort of files to: https://github.com/GoogleCloudPlatform/magic-modules/tree/main/mmv1/third_party/terraform/services/compute/test-fixtures

I'm not sure how this works for TGC, @rileykarson do you know?

franciscolimaneto:
@slevenick I have just updated the files with a new version that should work now without need to upload the certificates once it create those one.

delimaneto:
@slevenick could you please review again?

slevenick (approved):

PR #10660, opened 2024-05-10
Promote google_cloud_asset_resources_search_all datasource to ga 

averbuks:
Promote google_cloud_asset_resources_search_all datasource to ga
https://github.com/hashicorp/terraform-provider-google/issues/18034

Note: CloudAsset V1 API: [resourcesSearchAll](https://cloud.google.com/asset-inventory/docs/reference/rest/v1/TopLevel/searchAllResources#resourcesearchresult) has multiple `dynamicType` fields, which is currently impossible to handle well in the provider. All suth fields and some useless are avoided in the datasource, including `additional_attributes` field. We still keep `additional_attributes` field in the `beta` datasource, since beta API has a fixed type for the field.  

```release-note:new-datasource
`google_cloud_asset_resources_search_all`
```


rileykarson (commented):
Do we have any guarantee that, other than `additional_attributes`, the GA API here is a superset of the beta one? We may want to add a separate resource if there aren't, `data.google_cloud_asset_search_all_resources`.

The tests for a datasource are definitely not complete enough to assert as much.

rileykarson (commented):
Code looks good, waiting on test results.

rileykarson (approved):

PR #10666, opened 2024-05-11
[CloudSQL] Handle CLOUD_IAM_GROUP username edgecases

jackweinbender:
Although some of the issues surrounding the use of CLOUD_IAM_GROUP users with cloudsql I believe are google's own problems, in this case, I think this PR will address some of the problems with TF not being able to locate CLOUD_IAM_GROUP type resources, as discussed in https://github.com/hashicorp/terraform-provider-google/issues/17040.

The issue, I believe, is that group users retain the domain as a part of their username while standard users and service accounts have the domain stripped. You can see on L369 of the PR that this fact is already accounted for in part, but the same exception wasn't included in the diffSuppress function (L21ff.).

I've also included comments to make some of this explicit, since it's a bit counterintuitive.

```release-note:bug
sql: fix `diffSuppressIamUserName` function to handle `CLOUD_IAM_GROUP` correctly
```

### DIFF
--- a/mmv1/third_party/terraform/services/sql/resource_sql_user.go
+++ b/mmv1/third_party/terraform/services/sql/resource_sql_user.go
@@ -3,6 +3,7 @@ package sql
 import (
 	"fmt"
 	"log"
+	"slices"
 	"strings"
 	"time"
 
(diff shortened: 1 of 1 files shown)


mmv1/third_party/terraform/services/sql/resource_sql_user.go
```diff
@@ -17,12 +17,25 @@ import (
...
 
-	if old == strippedName && strings.Contains(userType, "IAM") {
-		return true
+	if slices.Contains(untruncated_iam_types, userType) {
+		// We compare old and new directly for untruncated entries
+		return old == new
+	}
+
+	if slices.Contains(truncated_iam_types, userType) {
+		// For truncated types, We strip the domain from the new value use it for comparison
```
jackweinbender, line 36:
```suggestion
		// For truncated types, We strip the domain from the new value and use it for comparison
```

slevenick (changes requested):
Failing due to resource_sql_user.go:32:5: undefined: slices

Have you been able to test this change?

jackweinbender:
@slevenick I'm just an idiot and forgot to migrate the `slices` import into this repo when I moved it.

mmv1/third_party/terraform/services/sql/resource_sql_user.go
```diff
@@ -17,12 +18,25 @@ import (
 )
 
 func diffSuppressIamUserName(_, old, new string, d *schema.ResourceData) bool {
-	strippedName := strings.Split(new, "@")[0]
+	// IAM users of type `CLOUD_IAM_USER` and `CLOUD_IAM_SERVICE_ACCOUNT` are created based on
```
slevenick, line 21: Is this behavior documented somewhere? I want to make sure this is intended behavior and not something that gets changed later and leaves us in a bad spot

jackweinbender: I haven't been able to find this in any documentation—the docs on this are pretty poor overall.

But the (apparent) rule I'm making explicit here is one we already account for on L368: "If the current user type isn't pg or CLOUD_IAM_GROUP, split the domain off the supplied email address." The implication being that the domain is retained by CLOUD_IAM_GROUP. So, we have a precedent, this PR just brings this block into alignment with how we're treating CLOUD_IAM_GROUP elsewhere.

jackweinbender:
Is there anything else you need me to do here, @slevenick?

OscarVanL:
How timely, I was just having a world of pain trying to setup group IAM on a CloudSQL instance. Is there any chance this is likely to be merged/released soon?

slevenick:
I think this looks good. Can you add a test that uses a CLOUD_IAM_GROUP?

slevenick (changes requested):
Please add a test that uses CLOUD_IAM_GROUP

OscarVanL:
Please can someone fix this bug in a separate PR? I would really like to create a CLOUD_IAM_GROUP via terraform and have been unable to.

PR #10668, opened 2024-05-11
fix vcr message format

shuyama1:
fix vcr message format issue. example: https://github.com/GoogleCloudPlatform/magic-modules/pull/10650#issuecomment-2105305924

```release-note:none
```


melinath:
Could we fix it like this instead: https://github.com/GoogleCloudPlatform/magic-modules/pull/10681

melinath:
sorry, didn't see this until after making that

melinath:
let me know what you think!

shuyama1:
This will be fixed in https://github.com/GoogleCloudPlatform/magic-modules/pull/10681

PR #10679, opened 2024-05-13
Add support for AutokeyConfig and KeyHandle KMS resources

tdbhacks:
Add support for the new AutokeyConfig and KeyHandle KMS resources, used to auto-provision KMS keys ready for Customer Managed Encryption Key (CMEK) use.

```release-note:new-resource
`google_kms_autokey_config`
```
```release-note:new-resource
`google_kms_key_handle`
```


tdbhacks:
@roaks3 This is pretty much the same as what was reviewed internally a couple weeks ago. I've tried adding and tweaking a custom_check_destroy snippet, but I haven't been able to make VCR tests happy. Wondering if I should skip the AutokeyConfig test because of the singleton weirdness, the test is failing because it detects a dangling resource after destroy, which is expected.

If this can be fixed and my snippet logic is wrong, let me know! The tl;dr is that for AutokeyConfig, "destroy" == unset the `key_project` field. Thanks!

roaks3:
Note: this is a promotion of https://cloud-internal-review.git.corp.google.com/c/cloud-graphite-eng/magic-modules-private-overrides/+/46210

@tdbhacks to help with review, can you confirm that the implementation is the same, besides the changes needed for promotion like removing override_path? Or if not, could you point to areas that needed to change?

Regarding the destroy check, I think what you're doing makes sense. I'm wondering if the `key_project` value might be nil instead of "". The destroy check looks correct on the generated side of things, and conceptually this should all work with the VCR environment.

tdbhacks:
@roaks3 woo-hoo, thanks for catching that!

tdbhacks:
> Note: this is a promotion of https://cloud-internal-review.git.corp.google.com/c/cloud-graphite-eng/magic-modules-private-overrides/+/46210
> 
> @tdbhacks to help with review, can you confirm that the implementation is the same, besides the changes needed for promotion like removing override_path? Or if not, could you point to areas that needed to change?

This is correct, no changes other than promotion changes and the custom_check_destroy logic.

tdbhacks:
@roaks3  One thing to note here, before we merge, is that this fetaure is in public preview. IIUC, that means it should only be available in `google-beta` even if we don't expect changes to the API. Is that accurate? If so, lmk and I'll add the beta restriction, I don't think that should break anything because the example tests uses beta anyways (since we rely on a beta resource)

roaks3:
Yea, see go/terraform-launch-stages for details, but if the API is public preview, these should be changed to beta-only resources.

Also, hopefully you have links for docs by now? Otherwise we may want to remove those lines from the yaml actually, or pre-populate them.

Everything else LGTM

tdbhacks:
> Yea, see go/terraform-launch-stages for details, but if the API is public preview, these should be changed to beta-only resources.
 
Ack, will do!

> Also, hopefully you have links for docs by now? Otherwise we may want to remove those lines from the yaml actually, or pre-populate them.

Indeed, docs should be published tomorrow so I'll update this PR once ready.

> Everything else LGTM

Amazing, thanks for your help!

tdbhacks:
Added docs links as well as the `min_version: beta` restriction, let me know if you see anything else that needs fixing before this is merged. Thanks for your help!

roaks3 (approved):
LGTM, updates look good!

PR #10681, opened 2024-05-13
Fixed color formatting for VCR test results

melinath:
Also fixed an incorrect reference to RECORDING mode.

Example failure mode: https://github.com/GoogleCloudPlatform/magic-modules/pull/10664#issuecomment-2105384793

```release-note:none

```


shuyama1 (approved):
I think this should also work!

PR #10682, opened 2024-05-13
chore(ci): use wait for commit go version in downstream ci

iyabchen:
use wait for commit go version in downstream ci

```release-note:none

```

### DIFF
--- a/.ci/gcb-push-downstream.yml
+++ b/.ci/gcb-push-downstream.yml
@@ -30,10 +30,10 @@ steps:
       waitFor: ["checkout"]
 
     # TPG
-    - name: 'gcr.io/graphite-docker-images/bash-plus'
+    - name: 'gcr.io/graphite-docker-images/go-plus'
       entrypoint: '/workspace/.ci/scripts/go-plus/magician/exec.sh'
       id: tpg-sync
-      waitFor: ["checkout"]
+      waitFor: ["build-magician-binary", "checkout"]
       args:
           - wait-for-commit
           - 'tpg-sync'
@@ -41,10 +41,10 @@ steps:
           - $COMMIT_SHA
 
     # TPGB
-    - name: 'gcr.io/graphite-docker-images/bash-plus'
+    - name: 'gcr.io/graphite-docker-images/go-plus'
       entrypoint: '/workspace/.ci/scripts/go-plus/magician/exec.sh'
       id: tpgb-sync
-      waitFor: ["checkout"]
+      waitFor: ["build-magician-binary", "checkout"]
       args:
           - wait-for-commit
           - 'tpgb-sync'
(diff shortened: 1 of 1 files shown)


.ci/gcb-push-downstream.yml
```diff
@@ -31,40 +31,44 @@ steps:
 
     # TPG
     - name: 'gcr.io/graphite-docker-images/bash-plus'
```
melinath, line 33: Do these need to use `go-plus`? That is what the previous PR did.

```suggestion
    - name: 'gcr.io/graphite-docker-images/go-plus'
```

melinath (changes requested):
We [did this change on Friday](https://github.com/GoogleCloudPlatform/magic-modules/pull/10658) and ended up having to revert it in https://github.com/GoogleCloudPlatform/magic-modules/pull/10678. I think the only thing missing was that the sync steps need to depend on the magician build step (which also needs to be the case here.)

CC @trodge FYI

iyabchen:
Added the build-magician-binary into the wait for list for sync steps.

melinath (approved):
LGTM

melinath:
This LGTM - @roaks3 could you take a look as well & merge if this looks good?

.ci/gcb-push-downstream.yml
```diff
@@ -30,41 +30,45 @@ steps:
       waitFor: ["checkout"]
 
     # TPG
-    - name: 'gcr.io/graphite-docker-images/bash-plus'
-      entrypoint: '/workspace/.ci/scripts/bash-plus/downstream-waiter/wait_for_commit.sh'
+    - name: 'gcr.io/graphite-docker-images/go-plus'
+      entrypoint: '/workspace/.ci/scripts/go-plus/magician/exec.sh'
       id: tpg-sync
-      waitFor: ["checkout"]
+      waitFor: ["build-magician-binary", "checkout"]
```
melinath, line 36: nit: I think actually `checkout` isn't necessary since it's already a dependency via `build-magician-binary` - but it's not going to cause problems to have it here, so not a blocker.

iyabchen: Done, removed the unnecessary ones, also in the push steps.

melinath (approved):

melinath:
(we could also wait to merge on Wednesday at this point.)

roaks3 (approved):
Changes LGTM based on the issue we saw on Monday. Is there any way to test this out before merging?

@melinath I don't feel comfortable merging, as I'm not intimately involved in this effort, but I am happy to be available when we do for support. Who owns this change? I would strongly prefer that the owner merges this and is available if something goes wrong.

melinath:
Unfortunately it's not possible to test yml changes before merging (as far as I know.) @trodge since you were working on this as part of the fixit, would you like to own merging this & keeping an eye on it? Otherwise I can.

PR #10684, opened 2024-05-13
Add 'force' parameter to all Datastream resources

alexandrunagy:
Adds the 'force' parameter for Datastream resources. This parameter allows creating the resources without validations.

The API parameter definition for each resource:
- ConnectionProfile: https://cloud.google.com/datastream/docs/reference/rest/v1/projects.locations.connectionProfiles/create#query-parameters
- PrivateConnection: https://cloud.google.com/datastream/docs/reference/rest/v1/projects.locations.privateConnections/create#query-parameters
- Stream: https://cloud.google.com/datastream/docs/reference/rest/v1/projects.locations.streams/create#query-parameters

Fixes https://github.com/hashicorp/terraform-provider-google/issues/14022

```release-note:enhancement
datastream: added `create_without_validation` field to `google_datastream_connection_profile`, `google_datastream_private_connection` and `google_datastream_stream` resources
```

### DIFF
--- a/mmv1/products/datastream/ConnectionProfile.yaml
+++ b/mmv1/products/datastream/ConnectionProfile.yaml
@@ -14,7 +14,7 @@
 --- !ruby/object:Api::Resource
 name: 'ConnectionProfile'
 base_url: 'projects/{{project}}/locations/{{location}}/connectionProfiles'
-create_url: 'projects/{{project}}/locations/{{location}}/connectionProfiles?connectionProfileId={{connection_profile_id}}'
+create_url: 'projects/{{project}}/locations/{{location}}/connectionProfiles?connectionProfileId={{connection_profile_id}}&force={{force}}'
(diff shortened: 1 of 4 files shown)


slevenick (commented):
Can you add this field to some tests so we can validate that it works correctly?

alexandrunagy:
@slevenick added it to some, let me know if there other places where I should add it

slevenick:
I'm thinking we should rename this field. Usually `force` is used for a force-delete behavior, as described in https://google.aip.dev/165

Maybe force_create or create_without_validation would be more clear about what this actually does?

alexandrunagy:
> I'm thinking we should rename this field. Usually `force` is used for a force-delete behavior, as described in https://google.aip.dev/165
> 
> Maybe force_create or create_without_validation would be more clear about what this actually does?

@slevenick renamed to create_without_validation

slevenick (approved):

PR #10507, opened 2024-04-20
Add reservedInternalRange fields to the google_compute_subnetwork resource

daanheikens:
Fixes: https://github.com/hashicorp/terraform-provider-google/issues/17881

**Release Note Template for Downstream PRs (will be copied)**

```release-note:enhancement
compute: added fields `reserved_internal_range` and `secondary_ip_ranges[].reserved_internal_range` to `google_compute_subnetwork` resource
```

PR #10697, opened 2024-05-14
Revert "Add reservedInternalRange fields to the google_compute_subnetwork resource"

trodge:
Reverts GoogleCloudPlatform/magic-modules#10507

The field is treated by terraform as required despite being marked optional

```release-note:breaking-change
compute: removed `secondary_ip_range.reserved_internal_range` field from `google_compute_subnetwork`
```


slevenick (approved):
This will be included in a patch release made today. It does break existing configs that have updated to 5.29.0 and started using the added field, but we've deemed this the best option to fix the breakage shown in https://github.com/hashicorp/terraform-provider-google/issues/18115

PR #10714, opened 2024-05-15
Make region from zone functions support more zone names

trodge:
fixes b/339683598

```release-note:none

```

### DIFF
--- a/mmv1/third_party/terraform/functions/region_from_zone.go
+++ b/mmv1/third_party/terraform/functions/region_from_zone.go
@@ -3,6 +3,7 @@ package functions
 import (
 	"context"
 	"fmt"
+	"strings"
 
 	"github.com/hashicorp/terraform-plugin-framework/function"
 )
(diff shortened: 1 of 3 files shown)


mmv1/third_party/terraform/tpgresource/utils.go
```diff
@@ -58,15 +58,14 @@ type TerraformResourceDiff interface {
...
-// If there aren't enough characters in the input string, an empty string is returned
+// This is by removing the characters after the last '-'.
 // e.g. southamerica-west1-a => southamerica-west1
 func GetRegionFromZone(zone string) string {
-	if zone != "" && len(zone) > 2 {
-		region := zone[:len(zone)-2]
-		return region
+	zoneParts := strings.Split(zone, "-")
+	if len(zoneParts) < 3 {
+		return zone
```
SarahFrench, line 66: Previously if the input was not a valid zone an empty string was returned, but now the original argument is returned. Is that change intentional?

mmv1/third_party/terraform/functions/region_from_zone.go
```diff
@@ -35,23 +36,24 @@ func (f RegionFromZoneFunction) Definition(ctx context.Context, req function.Def
...
 		resp.Error = function.ConcatFuncErrors(err)
 		return
 	}
 
-	if arg0[len(arg0)-2] != '-' {
-		err := function.NewArgumentFuncError(0, fmt.Sprintf("The input string \"%s\" is not a valid zone name.", arg0))
+	zoneParts := strings.Split(zone, "-")
+
+	if len(zoneParts) < 3 {
+		err := function.NewArgumentFuncError(0, fmt.Sprintf("The input string \"%s\" is not a valid zone name.", zone))
```
SarahFrench, line 54: This is much nicer, thanks!

SarahFrench (commented):
One small comment but happy to approve after it's discussed

SarahFrench (approved):

PR #10717, opened 2024-05-16
Fixes LifecycleRules Conditions not being set issue.

kautikdk:
Fixes https://github.com/hashicorp/terraform-provider-google/issues/17990

```release-note:bug
storage: fixed a bug where `google_storage_bucket.lifecycle_rule.condition` block fields  `days_since_noncurrent_time` and `days_since_custom_time`  and `num_newer_versions` were not working for 0 value. You can now send these by including a corresponding field such as `send_days_since_noncurrent_time_if_zero` in your configuration that is set to true.
```

### DIFF
--- a/mmv1/third_party/terraform/services/storage/resource_storage_bucket.go.erb
+++ b/mmv1/third_party/terraform/services/storage/resource_storage_bucket.go.erb
@@ -262,6 +262,21 @@ func ResourceStorageBucket() *schema.Resource {
 										Elem:        &schema.Schema{Type: schema.TypeString},
 										Description: `One or more matching name suffixes to satisfy this condition.`,
 									},
+									"send_days_since_noncurrent_time_if_zero": {
+										Type:        schema.TypeBool,
+										Optional:    true,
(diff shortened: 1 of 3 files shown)


kautikdk:
For tests, We can not actually check value from the cloud infra because of the empty value limitation in go language. I have added all the virtual fields to cover no permadiff happing in any case and field modification is not causing state inconsistency.

NickElliot (approved):
LGTM -- I'm not certain we should mark this as "fixing" the issue, as we still want to update the behavior for the `age` condition down the line during the major release. May be appropriate to open a new ticket for that seperately, or keep the current one open.

kautikdk:
Well, We can do it either ways. I think with this PR merged and with the `no_age` field, Users can set lifecycle rule conditions as they want without any problem. So we can mark this one as fixed and open new ticket for changing `age` behavior. That's my opinion on this but I am also okay with keeping this ticket open.

NickElliot:
sounds good to me!

NickElliot (changes requested):
Ah sorry, these need to be included in the docs at `mmv1/third_party/terraform/website/docs/r/storage_bucket.html.markdown`

kautikdk:
Oops, I forgot to add these new fields in the markdown file. 
Added in the new commit.

NickElliot (approved):

PR #10720, opened 2024-05-16
Address acc tests failures with higher TF versions by removing unused output block from acctest configs

SarahFrench:
Fixes: https://github.com/hashicorp/terraform-provider-google/issues/18109

In the context of the IAM tests that are generated from this example config, this output block triggers a bug present in Terraform 1.4.0 up to 1.8.3 (latest release).

I've raised that bug internally, but this edit to the test will enable it to pass again. The output isn't being used by any tests, so removing it seems ok to me.

```release-note:none

```

### DIFF
--- a/mmv1/templates/terraform/examples/cloudfunctions2_basic.tf.erb
+++ b/mmv1/templates/terraform/examples/cloudfunctions2_basic.tf.erb
@@ -36,7 +36,3 @@ resource "google_cloudfunctions2_function" "<%= ctx[:primary_resource_id] %>" {
     timeout_seconds     = 60
   }
 }
-
-output "function_uri" { 
-  value = google_cloudfunctions2_function.function.service_config[0].uri
-}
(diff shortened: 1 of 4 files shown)


SarahFrench:
I've triggered all the cloudfunction2 IAM tests in TeamCity that use the GA provider and use TF 1.8.3 : https://hashicorp.teamcity.com/buildConfiguration/TerraformProviders_GoogleCloud_GOOGLE_MMUPSTREAMTESTS_GOOGLE_PACKAGE_CLOUDFUNCTIONS2/149883?buildTab=overview&branch=refs%2Fheads%2Fauto-pr-10720&focusLine=NaN&hideTestsFromDependencies=false&hideProblemsFromDependencies=false

SarahFrench:
👆 Those tests passed!

melinath (approved):
Weird! +1 that these seem unnecessary

PR #10722, opened 2024-05-16
Add the subnetLength field to the InterconnectAttachment resource

bopas2:
Add the subnetLength field to the InterconnectAttachment resource

```release-note:enhancement
compute: added `subnet_length` field to `google_compute_interconnect_attachment` resource
```


mmv1/templates/terraform/examples/interconnect_attachment_dedicated.tf.erb
```diff
@@ -0,0 +1,36 @@
...
+  location             = "https://www.googleapis.com/compute/v1/projects/${data.google_project.project.name}/global/interconnectLocations/z2z-us-east4-zone1-lciadl-a" # Special location only available for Google testing.
+}
+
+resource "google_compute_interconnect_attachment" "<%= ctx[:primary_resource_id] %>" {
+  name                     = "<%= ctx[:vars]['interconnect_attachment_name'] %>"
+  type                     = "DEDICATED"
+  interconnect             = google_compute_interconnect.foobar.id
+  router                   = google_compute_router.foobar.id
+  mtu                      = 1500
+  subnet_length            = 0
```
roaks3, line 18: I'm a little confused about this test. The description for the field says that 29 and 30 are valid values, but here we are testing 0. Because of how TF handles zero-values, this might not even be sent to the server (ie. the equivalent of not setting the field). Is it not possible to test with the suggested values of 29 or 30?

Also, is the default value set on the server? If so, you may need to add `default_from_api: true`.

bopas2: Added the `default_from_api` field.

Followed up about the 0 value offline.

roaks3: Ok, confirmed that this value is not being sent:

```
POST /compute/beta/projects/ci-test-project-188019/regions/us-east4/interconnectAttachments?alt=json HTTP/1.1
Host: compute.googleapis.com
User-Agent: Terraform/1.2.5 (+https://www.terraform.io) Terraform-Plugin-SDK/2.33.0 terraform-provider-google-beta/acc
Content-Length: 429
Accept-Encoding: gzip

{
 "adminEnabled": true,
 "encryption": "NONE",
 "interconnect": "projects/ci-test-project-188019/global/interconnects/tf-test-interconenct-18x24pgb1wd",
 "mtu": "1500",
 "name": "tf-test-on-prem-attachment8x24pgb1wd",
 "region": "projects/ci-test-project-188019/global/regions/us-east4",
 "router": "projects/ci-test-project-188019/regions/us-east4/routers/tf-test-router-18x24pgb1wd",
 "stackType": "IPV4_ONLY",
 "type": "DEDICATED",
 "vlanTag8021q": 1000
}
```

If the 0 truly holds value here (I'm still not 100% clear), you may want to use the config `send_empty_value: true` to allow it to be sent. Otherwise, I believe there isn't any value in the test, and we should just remove it. I know there are limitations with this particular resource, and it is only a single field that is not being tested. It would be nice to have some sort of future-looking plan to test this, if possible, but not blocking.

bopas2: Ok, that plan sounds good to me. I will file a bug for our team to go back and add a test for this field in the future. Thanks!

roaks3 (approved):
LGTM, just waiting on checks to complete before merge

roaks3 (approved):
LGTM, will merge on Monday morning (which will be well ahead of the release cut on Tuesday evening)

PR #10724, opened 2024-05-16
Change region field required to default_from_api

pawelJas:
Change region field for google_compute_region_ssl_policy to be optional and to be inferred from environment.
 
Fixes https://github.com/hashicorp/terraform-provider-google/issues/17142

- Searched through the [issue tracker](https://github.com/hashicorp/terraform-provider-google/issues) for an open issue that this either resolves or contributes.
- [Generated Terraform](https://github.com/GoogleCloudPlatform/magic-modules#generating-downstream-tools), and ran make test and make lint to ensure it passes unit and linter tests.
- Ensured that all new fields I added that can be set by a user appear in at least one [example](https://github.com/GoogleCloudPlatform/magic-modules/tree/master/templates/terraform/examples) (for generated resources) or [third_party test](https://github.com/GoogleCloudPlatform/magic-modules/tree/master/third_party/terraform/tests) (for handwritten resources or update tests). - read only field
- [Ran](https://github.com/hashicorp/terraform-provider-google/blob/master/.github/CONTRIBUTING.md#tests) relevant acceptance tests (If the acceptance tests do not yet pass or you are unable to run them, please let your reviewer know).
- Read the [Release Notes Guide](https://github.com/GoogleCloudPlatform/magic-modules/blob/master/.ci/RELEASE_NOTES_GUIDE.md) before writing my release note below.

```release-note:enhancement
compute: Allowed `region` field to `google_compute_region_ssl_policy` resource to be inferred from environment
```


mmv1/products/compute/RegionSslPolicy.yaml
```diff
@@ -53,10 +53,12 @@ parameters:
     name: 'region'
     resource: 'Region'
     imports: 'name'
-    required: true
+    required: false
```
BBBmau, line 56: this can be removed entirely since by default it'll be set to `required: false`

pawelJas: removed 👍

BBBmau (approved):

PR #10741, opened 2024-05-20
Adding support for Dataplex Aspect Type resource.

avijain25:
Adding Terraform support for a Dataplex Aspect Type resource.
Tracking issue: https://github.com/hashicorp/terraform-provider-google/issues/18147

```release-note:new-resource
`google_dataplex_aspect_type`
```


mmv1/products/dataplex/AspectType.yaml
```diff
@@ -0,0 +1,144 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+--- !ruby/object:Api::Resource
+name: 'AspectType'
+base_url: 'projects/{{project}}/locations/{{location}}/aspectTypes/{{aspect_type_id}}'
+self_link: 'projects/{{project}}/locations/{{location}}/aspectTypes/{{aspect_type_id}}'
```
roaks3, line 17: Nit: I think self_link defaults to base_url, and this line can be removed

mmv1/products/dataplex/AspectType.yaml
```diff
@@ -0,0 +1,144 @@
...
+create_url: 'projects/{{project}}/locations/{{location}}/aspectTypes?aspectTypeId={{aspect_type_id}}'
+update_url: 'projects/{{project}}/locations/{{location}}/aspectTypes/{{aspect_type_id}}'
```
roaks3, line 19: Nit: Same for update_url and delete_url, I think they default to self_link, and these lines can be removed

mmv1/products/dataplex/AspectType.yaml
```diff
@@ -0,0 +1,144 @@
...
+    description: |
+      MetadataTemplate of the Aspect.
+    custom_expand: 'templates/terraform/custom_expand/json_schema.erb'
+    custom_flatten: 'templates/terraform/custom_flatten/json_schema.erb'
+    state_func:
+      'func(v interface{}) string { s, _ := structure.NormalizeJsonString(v);
+      return s }'
+    validation: !ruby/object:Provider::Terraform::Validation
+      function: 'validation.StringIsJSON'
+  - !ruby/object:Api::Type::Enum
```
roaks3, line 114: Nit: Since this is output-only, it's usually better modeled as a String, because no validation is needed, and it better supports changing the set of options.

mmv1/third_party/terraform/services/dataplex/resource_dataplex_aspect_type_test.go
```diff
@@ -0,0 +1,226 @@
...
+}
+EOF
+}
+`, context)
+}
+
+func testAccDataplexAspectType_update(context map[string]interface{}) string {
+	return acctest.Nprintf(`
+resource "google_dataplex_aspect_type" "test_aspect_type_basic" {
+  aspect_type_id = "tf-test-aspect-type-basic%{random_suffix}"
```
roaks3, line 194: `aspect_type_id` is immutable, so I don't think it should be changed here. Effectively this means the test is recreating the resource, when we want to be testing that in-place update worked as expected.

roaks3 (changes requested):
LGTM except a few nits, and the test should be changed

mmv1/products/dataplex/AspectType.yaml
avijain25: I tried removing self_link. This was causing build failures. So added it back.

mmv1/products/dataplex/AspectType.yaml
avijain25: Our API supports enum for transferStatus. We want to be consistent with our API contract.

roaks3:
Closing in favor of https://github.com/GoogleCloudPlatform/magic-modules/pull/10751

PR #10746, opened 2024-05-21
Rewrite example_config_body with Go

zli82016:
Rewrite ExampleConfigBody with Go

```release-note:none

```

### DIFF
--- a/mmv1/api/resource.go
+++ b/mmv1/api/resource.go
@@ -1121,7 +1121,7 @@ func (r Resource) IamTerraformName() string {
 	return fmt.Sprintf("%s_iam", r.TerraformName())
 }
 
-func (r Resource) IamSelfLinkProperties() []*Type {
+func (r Resource) IamSelfLinkIdentifiers() []string {
 	var selfLink string
 	if r.IamPolicy != nil {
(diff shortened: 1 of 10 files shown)


zli82016:
The diffs are only [new lines ](https://github.com/hashicorp/terraform-provider-google-beta/compare/main...ExampleConfigBody-diff), which are added [here](https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/api/resource/examples.go#L236).

To remove the diffs, we can have a new function similar to `ExecuteTemplate`, without adding new line at end. Any thoughts?

c2thorn:
> To remove the diffs, we can have a new function similar to `ExecuteTemplate`, without adding new line at end. Any thoughts?

Ah, we probably need to have a flag to the function to control when the newline is added or not.

c2thorn:
@zli82016 would you mind adding that flag to reduce diffs?

zli82016:
> @zli82016 would you mind adding that flag to reduce diffs?

I will do that

zli82016:
The flag is added to the remove diffs.

c2thorn (approved):
thank you for adding the flag!

PR #10750, opened 2024-05-21
update the preUpdate process for vertex_ai_index

JoyceYingZhu:
This PR is change the preUpdate process for vertex_ai_index so that when metadata.0.contents_delta_uri has change only first layer mask will apply and no other mask will add.

```release-note:bug
Update the preupdate process to set update mask that supported by vertex_index API.

```

### DIFF
--- a/mmv1/products/securitycenterv2/OrganizationNotificationConfig.yaml
+++ b/mmv1/products/securitycenterv2/OrganizationNotificationConfig.yaml
@@ -1,0 +1,129 @@
+# Copyright 2024 Google Inc.
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+--- !ruby/object:Api::Resource
+name: 'OrganizationNotificationConfig'
+base_url: organizations/{{organization}}/locations/{{location}}/notificationConfigs
+self_link: '{{name}}'
+create_url: organizations/{{organization}}/locations/{{location}}/notificationConfigs?configId={{config_id}}
+update_verb: :PATCH
+update_mask: true
+description: |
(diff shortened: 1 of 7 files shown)


mmv1/third_party/terraform/services/vertexai/resource_vertex_ai_index_test.go
```diff
@@ -139,7 +139,7 @@ resource "google_vertex_ai_index" "index" {
   }
   region   = "us-central1"
   display_name = "tf-test-test-index%{random_suffix}"
-  description = "index for test (updated)"
+  description = "index for test"
```
slevenick, line 142: Why did you change this back to the unupdated text? Doesn't changing this exercise the update code?

JoyceYingZhu: Yes, this is according the change on preupdate process. When field contents_delta_uri update, no other field should update. So in this case field content_delta_uri is updated and description will not update. You can see that in preUpdate we rewrite updateMask when the field content_delta_uri change.

slevenick: But a user will likely attempt to update contents_delta_uri alongside other fields. We can't have the resource fail in that case

JoyceYingZhu: In that case we apply updateMask=metadata in preupdate so that the other fields will not be updated.

slevenick: But that will end up in a two-stage apply for the user, where they run `terraform apply` which succeeds but hasn't fully completed the update it said it would.

If they can't be updated simultaneously then we need to call the API twice, once to update the contents_delta_uri and once to update the rest of the fields.

Why can't we update the entire resource at the same time? Why does the API only allow updating this field by itself?

JoyceYingZhu: https://cloud.google.com/vertex-ai/docs/vector-search/update-rebuild-index#update-batch, we mark it in user guideline that don't update the field with other fields.

slevenick: Sure, but that's not an acceptable way for this resource to work in Terraform. We need to build the resource so a user can update multiple fields at once.

My question is around the design of the API. Why is it that users can't update the field with other fields?

slevenick (changes requested):
Previous review questions still apply

PR #10754, opened 2024-05-21
added required label to load balancing scheme for traffic extensions

maxi-cit:
Fixes: https://github.com/hashicorp/terraform-provider-google/issues/18196

```release-note:enhancement
networkservices: made field `load_balancing_scheme` required in resource `google_network_services_lb_traffic_extension`, as resource creation is always failing
```

**References**
* https://cloud.google.com/service-extensions/docs/reference/rest/v1beta1/projects.locations.lbTrafficExtensions#resource:-lbtrafficextension

### DIFF
--- a/mmv1/products/networkservices/LbTrafficExtension.yaml
+++ b/mmv1/products/networkservices/LbTrafficExtension.yaml
@@ -174,6 +174,7 @@ properties:
                 item_type: Api::Type::String
   - !ruby/object:Api::Type::Enum
     name: 'loadBalancingScheme'
+    required: true
     immutable: true
     description: |
       All backend services and forwarding rules referenced by this extension must share the same load balancing scheme.


maxi-cit:
API fails when omiting the 'load_balancing_scheme' field which is the expected behaviour according to the API docs in the references. So I am adding this flag to make it fail at the provider level.

I do not think this is a breaking change since all users are not able to actually create this resource without the 'load_balacing_scheme' field, but let me know if this not the case.

maxi-cit:
Hello @BBBmau, could you get sometime to check this?

BBBmau (approved):

maxi-cit:
Hello @BBBmau any new regarding this?

BBBmau:
Been meaning to get input on this. With this being marked as a breaking change would we include this into the 6.0.0 release @c2thorn ?

c2thorn:
This has missed the cutoff for 6.0, so I will close it to prevent it from accidentally getting merged into a minor release.

PR #10764, opened 2024-05-22
Support Delayed Deletion for Vmwareengine Private Cloud

swamitagupta:
1. Adds support to delay the Private Cloud deletion . More information about the intended behavior can be found [here](https://cloud.google.com/vmware-engine/docs/private-clouds/howto-delete-private-cloud#delete_a_private_cloud). 
2. Modifies the region for vmwareengine resource tests to region 'me-west1' with allocated quota.
3. Facilitates type change from `TIME_LIMITED` to `STANDARD` for multi-node Private Clouds

```release-note:enhancement
vmwareengine: added `deletion_delay_hours` field to `google_vmwareengine_private_cloud` resource
```
```release-note:enhancement
vmwareengine: support type change from `TIME_LIMITED` to `STANDARD` for multi-node `google_vmwareengine_private_cloud` resource
```

### DIFF
--- a/mmv1/products/vmwareengine/PrivateCloud.yaml
+++ b/mmv1/products/vmwareengine/PrivateCloud.yaml
@@ -15,7 +15,7 @@
 name: 'PrivateCloud'
 base_url: 'projects/{{project}}/locations/{{location}}/privateClouds'
 self_link: 'projects/{{project}}/locations/{{location}}/privateClouds/{{name}}'
-delete_url: 'projects/{{project}}/locations/{{location}}/privateClouds/{{name}}?delay_hours=0'
+delete_url: 'projects/{{project}}/locations/{{location}}/privateClouds/{{name}}'
 create_url: 'projects/{{project}}/locations/{{location}}/privateClouds?privateCloudId={{name}}'
 update_verb: :PATCH
 references: !ruby/object:Api::Resource::ReferenceLinks
@@ -45,11 +45,12 @@ async: !ruby/object:Api::OpAsync
 import_format: ["projects/{{project}}/locations/{{location}}/privateClouds/{{name}}"]
 autogen_async: true
 custom_code: !ruby/object:Provider::Terraform::CustomCode
-  post_delete: "templates/terraform/post_delete/private_cloud.go.erb"
+  constants: templates/terraform/constants/vmwareengine_private_cloud.go.erb
   decoder: "templates/terraform/decoders/private_cloud.go.erb"
+  pre_delete: templates/terraform/pre_delete/vmwareengine_private_cloud.go.erb
+  post_delete: "templates/terraform/post_delete/private_cloud.go.erb"
   post_update: "templates/terraform/post_update/private_cloud.go.erb"
   update_encoder: "templates/terraform/update_encoder/private_cloud.go.erb"
-  constants: templates/terraform/constants/vmwareengine_private_cloud_type.go.erb
 examples:
   - !ruby/object:Provider::Terraform::Examples
     name: "vmware_engine_private_cloud_basic"
@@ -89,6 +90,16 @@ parameters:
     url_param_only: true
     description: |
       The ID of the PrivateCloud.
+
+virtual_fields:
+  - !ruby/object:Api::Type::Integer
+    name: "deletion_delay_hours"
+    description: |
+      The number of hours to delay this request. You can set this value to an hour between 0 to 8, where setting it to 0 starts the deletion request immediately. If no value is set, a default value is set at the API Level.
+  - !ruby/object:Api::Type::Boolean
+    name: "send_deletion_delay_hours_if_zero"
+    description: |
+      While set true, deletion_delay_hours value will be sent in the request even for zero value of the field. This field is only useful for setting 0 value to the deletion_delay_hours field. It can be used both alone and together with deletion_delay_hours.
 
 properties:
   - !ruby/object:Api::Type::String
--- a/mmv1/templates/terraform/constants/vmwareengine_private_cloud.go.erb
+++ b/mmv1/templates/terraform/constants/vmwareengine_private_cloud.go.erb
@@ -1,0 +1,33 @@
+func vmwareenginePrivateCloudStandardTypeDiffSuppressFunc(_, old, new string, d *schema.ResourceData) bool {
+	if (old == "STANDARD" && new == "") || (old == "" && new == "STANDARD") {
+		return true
+	}
+    if (isMultiNodePrivateCloud(d) && old == "TIME_LIMITED" && new == "STANDARD") {
+        log.Printf("[DEBUG] Multinode Private Cloud found, facilitating TYPE change to STANDARD")
+        return true
+    }
+	return false
+}
+
+func isMultiNodePrivateCloud(d *schema.ResourceData) bool {
+    nodeConfigMap := d.Get("management_cluster.0.node_type_configs").(*schema.Set).List()
+    totalNodeCount := 0
+    for _, nodeConfig := range nodeConfigMap {
+        configMap, ok := nodeConfig.(map[string]interface{})
+        if !ok {
+            log.Printf("[DEBUG] Invalid node configuration format for private cloud.")
+            continue 
+        }
+        nodeCount, ok := configMap["node_count"].(int)
+        if !ok {
+            log.Printf("[DEBUG] Invalid node_count format for private cloud.")
+            continue 
+        }
+        totalNodeCount += nodeCount
+    }
+    log.Printf("[DEBUG] The node count of the private cloud is found to be %v nodes.", totalNodeCount)
+    if totalNodeCount > 2 {
+        return true
+    }
+    return false
+}
--- a/mmv1/templates/terraform/post_delete/private_cloud.go.erb
+++ b/mmv1/templates/terraform/post_delete/private_cloud.go.erb
@@ -29,6 +29,13 @@ privateCloudPollRead := func(d *schema.ResourceData, meta interface{}) transport
             if err != nil {
                 return res, err
             }
+            // if resource exists but is marked for deletion
+            log.Printf("[DEBUG] Fetching state of the private cloud.")
+            v, ok := res["state"]
+            if ok && v.(string) == "DELETED" {
+                log.Printf("[DEBUG] The Private cloud has been successfully marked for delayed deletion.")
+                return nil, nil
+            }
             return res, nil
         }
     }
--- a/mmv1/third_party/terraform/services/vmwareengine/resource_vmwareengine_private_cloud_test.go
+++ b/mmv1/third_party/terraform/services/vmwareengine/resource_vmwareengine_private_cloud_test.go
@@ -18,7 +18,7 @@ func TestAccVmwareenginePrivateCloud_vmwareEnginePrivateCloudUpdate(t *testing.T
 	t.Parallel()
 
 	context := map[string]interface{}{
-		"region":          "southamerica-west1",
+		"region":          "me-west1",
 		"random_suffix":   acctest.RandString(t, 10),
 		"org_id":          envvar.GetTestOrgFromEnv(t),
 		"billing_account": envvar.GetTestBillingAccountFromEnv(t),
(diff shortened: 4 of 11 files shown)


swamitagupta:
This PR intends to solve #10764

swamitagupta:
Hi @melinath, 
Wanted to let you know that I am in the process of testing various CUJs locally and will update you once the changes have successfully passed the tests and are ready to be reviewed for submission.

melinath (commented):
ah okay - in that case, marking as reviewed for now.

melinath (commented):
It looks like this might not be ready for review yet; re-request review if you do want me to take a look.

mmv1/templates/terraform/post_delete/private_cloud.go.erb
```diff
@@ -26,9 +26,19 @@ privateCloudPollRead := func(d *schema.ResourceData, meta interface{}) transport
                 RawURL: url,
                 UserAgent: userAgent,
             })
+            // if resource does not exist
             if err != nil {
+                log.Printf("[DEBUG] The Private cloud has been successfully deleted.")
```
melinath, line 31: this if statement actually indicates there was an error during deletion.

mmv1/templates/terraform/post_delete/private_cloud.go.erb
```diff
@@ -26,9 +26,19 @@ privateCloudPollRead := func(d *schema.ResourceData, meta interface{}) transport
...
                 return res, err
             }
+            // if resource exists but is marked for deletion
+            log.Printf("[DEBUG] Fetching state of the private cloud.")
+            v, ok := res["state"]
+            if ok && v.(string) == "DELETED" {
+                log.Printf("[DEBUG] The Private cloud has been successfully marked for delayed deletion.")
+                return res, errors.New("PrivateCloud is marked for deletion")
```
melinath, line 39: I think you might want to return nil, nil here to signify that the resource was successfully put in a deleted state.

melinath (changes requested):

melinath:
Put down some initial thoughts, but I'm handing off the review since it looks like I won't be able to wrap it up before I'm OOO for a bit.

mmv1/products/vmwareengine/PrivateCloud.yaml
```diff
@@ -89,6 +91,17 @@ parameters:
     url_param_only: true
     description: |
       The ID of the PrivateCloud.
+  - !ruby/object:Api::Type::NestedObject
+    name: 'deletionParameters'
```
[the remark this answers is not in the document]
swamitagupta: I initially planned to add only the `deletion_delay_hours` parameter. However, this made me unable to achieve the following behavior:
- if user enters no value, pass the delete API url without the delay hr mask. This would add the default delay hour which is set at the api level. We dont want to hardcode this value at the terraform level so that it can be easily modified and is at one place. 
- if the user adds any value (0 and non zero), use this value for the delay hr mask in the delete url.  

Since Terraform cant distinguish between empty and zero values, I created a nested object so that I can check its length to find out if the user did set the `deletion_delay_hours` value.  Please let me know if you feel there is any other better approach to handle this use case.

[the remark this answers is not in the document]
melinath: We don't have a strong pattern for this kind of situation but we usually use a second boolean field to control behavior in some way. For example:

- `send_deletion_delay_hours_if_zero`: similar to https://github.com/GoogleCloudPlatform/magic-modules/pull/10717. It would mean that people who want to send zero need to really really mean it!
- you could also do a field like `use_default_delay_hours` to indicate that the server should by default always send whatever is set (including zero) unless the field is set.

I'm guessing you'd prefer something like the former since folks are less likely to set `0` for delay_hours accidentally.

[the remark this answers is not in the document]
swamitagupta: I am concerned that introducing an additional boolean field alongside the delay hours field could potentially introduce user error, particularly given its absence from the VMware Engine API.

Creating a nested object looks like a cleaner and more adaptive approach. This would also hold additional delete API parameters which might be added in the near future.

Wdyt?

mmv1/templates/terraform/post_delete/private_cloud.go.erb
swamitagupta: Noted, thanks

mmv1/templates/terraform/pre_create/vmwareengine_private_cloud.go.erb
```diff
@@ -0,0 +1,9 @@
+// Check if the project exists in a deleted state
+pcMarkedForDeletion, err := isPrivateCloudInDeletedState(config, d, billingProject, userAgent)
+if err != nil {
+	return fmt.Errorf("Error checking if Private Cloud exists and is marked for deletion: %s", err)
+}
+if pcMarkedForDeletion {
+    log.Printf("[DEBUG] Private Cloud exists and is marked for deletion. Triggering UNDELETE of the Private Cloud.\n")
+    url = url + ":undelete"
```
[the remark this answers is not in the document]
swamitagupta: Noted, removed

mmv1/products/vmwareengine/PrivateCloud.yaml
```diff
@@ -89,6 +91,17 @@ parameters:
     url_param_only: true
     description: |
       The ID of the PrivateCloud.
+  - !ruby/object:Api::Type::NestedObject
```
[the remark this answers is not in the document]
swamitagupta: Modified, thanks

mmv1/products/vmwareengine/PrivateCloud.yaml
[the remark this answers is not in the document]
swamitagupta: Used `send_deletion_delay_hours_if_zero` since virtual fields dont support nested fields at the moment

mmv1/products/vmwareengine/PrivateCloud.yaml
```diff
@@ -90,6 +91,16 @@ parameters:
     description: |
       The ID of the PrivateCloud.
 
+virtual_fields:
+  - !ruby/object:Api::Type::Integer
+    name: "deletion_delay_hours"
+    description: |
+      The number of hours to delay this request. You can set this value to an hour between 0 to 8, where setting it to 0 starts the deletion request immediately. If no value is set, a default value is set at the API Level.
```
c2thorn, line 98: What is the API default value for deletion delay? Is it 0? How often do you think users will need to additionally set `send_deletion_delay_hours_if_zero`?

mmv1/templates/terraform/constants/vmwareengine_private_cloud.go.erb
```diff
@@ -0,0 +1,33 @@
+func vmwareenginePrivateCloudStandardTypeDiffSuppressFunc(_, old, new string, d *schema.ResourceData) bool {
+	if (old == "STANDARD" && new == "") || (old == "" && new == "STANDARD") {
+		return true
+	}
+    if (isMultiNodePrivateCloud(d) && old == "TIME_LIMITED" && new == "STANDARD") {
+        log.Printf("[DEBUG] Multinode Private Cloud found, facilitating TYPE change to STANDARD")
```
c2thorn, line 6: Can you give more context as to why a type diff from `TIME_LIMITED` -> `STANDARD` needs to be suppressed with Multinode? Also, is this related or separate to delayed deletion?

mmv1/third_party/terraform/services/vmwareengine/resource_vmwareengine_private_cloud_test.go
```diff
@@ -18,7 +18,7 @@ func TestAccVmwareenginePrivateCloud_vmwareEnginePrivateCloudUpdate(t *testing.T
 	t.Parallel()
 
 	context := map[string]interface{}{
-		"region":          "southamerica-west1",
+		"region":          "me-west1",
```
c2thorn, line 21: can you provide more context as to why this is being updated?

mmv1/third_party/terraform/services/vmwareengine/resource_vmwareengine_subnet_test.go
```diff
@@ -8,6 +8,8 @@ import (
 )
 
 func TestAccVmwareengineSubnet_vmwareEngineUserDefinedSubnetUpdate(t *testing.T) {
+	// Temporarily skipping so that this test does not run and consume resources during PR pushes. It is bound to fail and is being fixed by PR #10992
```
c2thorn, line 11: we will remove these before submitting right?

c2thorn (commented):
asked some clarifying questions for context, things seem okay at first glance however.
Lots of parts, but since changes are contained to a diff suppress or deletion functionality, it should be fairly safe.

mmv1/products/vmwareengine/PrivateCloud.yaml
swamitagupta: Currently, the default delay at the API level is set to 3 hours. However, it might be modified in the near future. 

Users would need to set `send_deletion_delay_hours_if_zero` if they want to send `delay_hours=0` through the API call and delete the PC immediately. I have added this boolean field after a discussion with Stephen in a previous thread since terraform does not differentiate between "no value set" and "0 value set".

mmv1/third_party/terraform/services/vmwareengine/resource_vmwareengine_private_cloud_test.go
swamitagupta: Some months back, we exhausted the number of nodes in the southamerca-west1 region which were being used for this test, leading to PC creation failures in multiple terraform tests. 

As a result, [go/gcve-resource-optimisation-for-tf-tests](http://goto.google.com/gcve-resource-optimisation-for-tf-tests) was created and additional nodes were reserved in me-west1 region for terraform test projects. This PR is part of its implementation efforts along with PR #10992.

mmv1/third_party/terraform/services/vmwareengine/resource_vmwareengine_subnet_test.go
swamitagupta: We can keep these commented till PR #10992 gets submitted. That way, the consumption of nodes for tests which are bound to fail would be prevented. 

These tests would be modified and combined by PR #10992 which is built on the design [go/gcve-resource-optimisation-for-tf-tests](http://goto.google.com/gcve-resource-optimisation-for-tf-tests)

mmv1/templates/terraform/constants/vmwareengine_private_cloud.go.erb
c2thorn: I also see that `type` is set to `ignore_read`, so at least it will remain consistent. Thanks for the context.

mmv1/third_party/terraform/services/vmwareengine/resource_vmwareengine_subnet_test.go
c2thorn: Understood, o the relevant test for these code changes are all in `mmv1/third_party/terraform/services/vmwareengine/resource_vmwareengine_private_cloud_test.go`

mmv1/products/vmwareengine/PrivateCloud.yaml
c2thorn: Just noting that I do not like the need for `send_deletion_delay_hours_if_zero` but cannot think of another way to force send 0 only when the user has specified due to TF limitations. I am sure this was discussed previously

c2thorn (approved):
LGTM

mmv1/products/vmwareengine/PrivateCloud.yaml
swamitagupta: Yes, [this](https://github.com/GoogleCloudPlatform/magic-modules/pull/10764#discussion_r1655305161) is the previous discussion thread for implementing this change. 

Another approach which I could think of was using `delay_hours` inside a nested object `deletion_parameters` but nested objects are not yet supported for `virtual_fields`.

mmv1/third_party/terraform/services/vmwareengine/resource_vmwareengine_subnet_test.go
swamitagupta: Yes, thats correct

PR #10767, opened 2024-05-22
Rework comments on deletion_protection

slevenick:
Make comments on deletion_protection more obvious about the field's behavior

```release-note:none

```

### DIFF
--- a/mmv1/products/privateca/CertificateAuthority.yaml
+++ b/mmv1/products/privateca/CertificateAuthority.yaml
@@ -113,8 +113,7 @@ examples:
       - 'deletion_protection'
   - !ruby/object:Provider::Terraform::Examples
     name: 'privateca_certificate_authority_custom_ski'
-    primary_resource_id: 'default'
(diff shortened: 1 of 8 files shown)


mmv1/products/privateca/CertificateAuthority.yaml
```diff
@@ -136,8 +136,9 @@ virtual_fields:
     name: 'deletion_protection'
     default_value: true
     description: |
-      Whether or not to allow Terraform to destroy the CertificateAuthority. Unless this field is set to false
-      in Terraform state, a `terraform destroy` or `terraform apply` that would delete the instance will fail.
+      Whether Terraform will be prevented from destroying the CertificateAuthority.
+      When a`terraform destroy` or `terraform apply` would delete the CertificateAuthority,
```
melinath, line 140: What would you think about stating what happens for both true and false, similar to https://developers.google.com/style/api-reference-comments#parameters?

melinath (changes requested):

melinath (approved):

PR #10771, opened 2024-05-23
Remove most calls to 'GetServiceAccount' in the create path

chasevedder:
Removes calls to GetServiceAccount in all cases except when `create_ignore_already_exists=true` and the Service Account already exists. This should eliminate most cases of the "Provider produced inconsistent result after apply" that occurs when creating new Service Accounts.

This bug occurs when a Service Account is created successfully but GetServiceAccount returns a 404 due to eventual consistency. We have all the information we need from the input config and the CreateServiceAccount response, so there's no need to make requests to GetServiceAccount at all.

```release-note:bug
iam: Fix most instances of the "Provider produced inconsistent result after apply" bug for `google_service_account` resources.
```

### DIFF
--- a/mmv1/third_party/terraform/services/resourcemanager/resource_google_service_account.go
+++ b/mmv1/third_party/terraform/services/resourcemanager/resource_google_service_account.go
@@ -118,6 +118,7 @@ func resourceGoogleServiceAccountCreate(d *schema.ResourceData, meta interface{}
 		AccountId:      aid,
 		ServiceAccount: sa,
 	}
+
 	d.SetId(fmt.Sprintf("projects/%s/serviceAccounts/%s@%s.iam.gserviceaccount.com", project, aid, project))
 
 	sa, err = config.NewIamClient(userAgent).Projects.ServiceAccounts.Create("projects/"+project, r).Do()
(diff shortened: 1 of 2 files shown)


mmv1/third_party/terraform/services/resourcemanager/resource_google_service_account.go
```diff
@@ -118,47 +118,50 @@ func resourceGoogleServiceAccountCreate(d *schema.ResourceData, meta interface{}
...
+		return fmt.Errorf("Error setting member: %s", err)
 	}
 
-	// We poll until the resource is found due to eventual consistency issue
-	// on part of the api https://cloud.google.com/iam/docs/overview#consistency
-	err = transport_tpg.PollingWaitTime(resourceServiceAccountPollRead(d, meta), transport_tpg.PollCheckForExistence, "Creating Service Account", d.Timeout(schema.TimeoutCreate), 1)
-
-	if err != nil {
-		return err
+	if d.Get("disabled") == nil {
```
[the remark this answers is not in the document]
chasevedder: The intention is to use the input value and set false as the default, but the test seems to be flaky so I'm not sure this is working as intended. I'll see if I can add some better test coverage and fix this.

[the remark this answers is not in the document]
chasevedder: I removed the unnecessary bits and made some changes in the test to reduce the flakiness.

benhxy (approved):

mmv1/third_party/terraform/services/resourcemanager/resource_google_service_account.go
```diff
@@ -119,48 +119,65 @@ func resourceGoogleServiceAccountCreate(d *schema.ResourceData, meta interface{}
...
-			_, saerr := config.NewIamClient(userAgent).Projects.ServiceAccounts.Get(d.Id()).Do()
-			return saerr
-		},
-		Timeout: d.Timeout(schema.TimeoutCreate),
-		ErrorRetryPredicates: []transport_tpg.RetryErrorPredicateFunc{
-			transport_tpg.IsNotFoundRetryableError("service account creation"),
-			transport_tpg.IsForbiddenIamServiceAccountRetryableError("service account creation"),
-		},
-	})
+	email := sa.Name[strings.LastIndex(sa.Name, "/")+1:]
```
benhxy, line 161: resourceServiceAccountPollRead() already calls populateResourceData(). Do we still need line 161-174?

chasevedder: resourceServiceAccountPollRead() actually does not call populateResourceData(). We could do that, but the intent here is just to poll as an existence check without actually updating the ResourceData so that even if we are unable to poll, the create flow will always succeed as long as CreateServiceAccount returns success.

benhxy: I see, I was confused by GitHub code folding. LGTM

benhxy (approved):

c2thorn:
Postponing this for broader discussion with the team and deferring to https://github.com/GoogleCloudPlatform/magic-modules/pull/10813 to solve the immediate problem

rileykarson:

PR #10775, opened 2024-05-23
`google_compute_instance` - add validation for `name`

AlexPykavy:
To fail during the plan phase if the name violates the following rules:
https://cloud.google.com/compute/docs/naming-resources#resource-name-format.

```release-note:enhancement
compute: added `regex` validation for the `name` field of the `google_compute_instance` resource
```

### DIFF
--- a/mmv1/products/compute/Instance.yaml
+++ b/mmv1/products/compute/Instance.yaml
@@ -378,6 +378,8 @@ properties:
       lowercase letter, and all following characters must be a dash,
       lowercase letter, or digit, except the last character, which cannot
       be a dash.
+    validation: !ruby/object:Provider::Terraform::Validation
+      regex: '^(?:[a-z](?:[-a-z0-9]{0,61}[a-z0-9])?)$'
   - !ruby/object:Api::Type::Array
     name: 'networkInterfaces'
     description: |


AlexPykavy:
Initially I opened a PR to the https://github.com/hashicorp/terraform-provider-google/pull/18227 but then found that it has to be directed here.

In that PR I wanted to use some human-friendly error message, like `The name can be 1-63 characters in length. The first character must be a lowercase letter, and all the following characters must be hyphens, lowercase letters, or digits, except the last character, which cannot be a hyphen.`. As far as I can [see](https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/provider/terraform/validation.rb) it's not possible with the current implementation of the magic-modules because `regex` attribute can't accept any inner arguments.  May we extend this logic to allow this?

c2thorn (changes requested):
Hi @AlexPykavy 
There are a few things with this change:

1. Adding stricter validation is overall a breaking change: https://googlecloudplatform.github.io/magic-modules/develop/breaking-changes/breaking-changes/#making-validation-more-strict
This would need to be saved for the 6.0.0 major release and have a github issue created to track it.
1. Compute instance is a handwritten file so your changes would have to be made directly against the go code in https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/third_party/terraform/services/compute/resource_compute_instance.go.erb

AlexPykavy:
Hi @c2thorn ,

I will update the handwritten file then.

Also I am not sure that it's a breaking change as this validation has been already done by the Compute Engine API. The only thing I do is repeating the same validation during the plan phase. So let me know where you are ready to accept this change: in the current version or in the next major.

c2thorn (commented):
Saving validation changes for major releases is a general rule, and I'd want to make sure we are careful when it comes to as important of a resource as `google_compute_instance` is. If you wouldn't mind creating a GH issue for it: https://github.com/hashicorp/terraform-provider-google/issues/new/choose

It would go to the triage team which would be able to make a decision on the time

PR #10778, opened 2024-05-23
Fix `content` in the `google_storage_bucket_object_content` data source which is incorrectly marked as a non-computed attribute

wafrelka:
Fixes https://github.com/hashicorp/terraform-provider-google/issues/15717

```release-note:bug
storage: fixed `content` in the `google_storage_bucket_object_content` data source which is incorrectly marked as a non-computed attribute
```


NickElliot:
Going to verify some of the behavior here locally to ensure this change isn't actually breaking due to the weirdness of how GCS buckets function as handwritten resources.

wafrelka:
Is there anything I should do to move this PR forward?

NickElliot (changes requested):
Sorry about that prolonged check on my end.

Can you undo the change removing the "computed" attribute from the resource? I don't believe it should affect the change made in `mmv1/third_party/terraform/services/storage/data_source_storage_bucket_object_content.go` that would resolve the `null` issue.

wafrelka:
I don't think this is a breaking change as `content` has never been a computed attribute since it was implemented.

Nevertheless, I reverted the change to the resource and force-pushed the branch.
Thank you for your review. :)

wafrelka:
Could you take a look?

NickElliot (approved):
LGTM!

NickElliot:
@wafrelka this change needs to be reverted because removing the ability to supply to the data source is breaking (despite this not having functional purpose), if you want to reopen the PR against the 6.0.0 release branch (per https://googlecloudplatform.github.io/magic-modules/develop/breaking-changes/make-a-breaking-change/) that would be appreciated!

I'll go ahead and handle adding the deprecation notice in the meantime.

wafrelka:
@NickElliot can we mark the `content` attribute as "computed" and "optional"?
I think it will fix https://github.com/hashicorp/terraform-provider-google/issues/15717 without breaking the ability to set the field.

PR #10783, opened 2024-05-24
[AlloyDB] Private Service Connect Support

vmiglani:
Description:
Changes for supporting Private Service Connect (PSC) in Terraform.

Issue - https://b.corp.google.com/issues/313702463

```release-note:enhancement
alloydb: added `psc_config` field to `google_alloydb_cluster` resource
```
```release-note:enhancement
alloydb: added `psc_instance_config` field to `google_alloydb_instance` resource
```
```release-note:none
alloydb: added`psc_config.0.psc_enabled` to the `exactly_one_of` field network restrictions to `google_alloydb_cluster` resource
```

### DIFF
--- a/mmv1/products/alloydb/Instance.yaml
+++ b/mmv1/products/alloydb/Instance.yaml
@@ -290,7 +290,7 @@ properties:
         output: true
         description: |
           The service attachment created when Private Service Connect (PSC) is enabled for the instance.
-          The name of the resource will be in the format of 
+          The name of the resource will be in the format of
           `projects/<alloydb-tenant-project-number>/regions/<region-name>/serviceAttachments/<service-attachment-name>`
       - !ruby/object:Api::Type::Array
         name: allowedConsumerProjects
--- a/mmv1/products/alloydb/Instance.yaml
+++ b/mmv1/products/alloydb/Instance.yaml
@@ -299,7 +299,7 @@ properties:
           List of consumer projects that are allowed to create PSC endpoints to service-attachments to this instance.
           These should be specified as project numbers only.
         item_validation: !ruby/object:Provider::Terraform::Validation
-          function: 'verify.ValidateProjectNumber()'
+          regex: '^\d+$'
       - !ruby/object:Api::Type::String
         name: 'pscDnsName'
         output: true
--- a/mmv1/products/alloydb/Instance.yaml
+++ b/mmv1/products/alloydb/Instance.yaml
@@ -298,6 +298,8 @@ properties:
         description: |
           List of consumer projects that are allowed to create PSC endpoints to service-attachments to this instance.
           These should be specified as project numbers only.
+        item_validation: !ruby/object:Provider::Terraform::Validation
+          function: 'verify.ValidateProjectNumber()'
       - !ruby/object:Api::Type::String
         name: 'pscDnsName'
         output: true
--- a/mmv1/products/alloydb/Instance.yaml
+++ b/mmv1/products/alloydb/Instance.yaml
@@ -281,7 +281,6 @@ properties:
               - :ALLOW_UNENCRYPTED_AND_ENCRYPTED
   - !ruby/object:Api::Type::NestedObject
     name: 'pscInstanceConfig'
-    default_from_api: true
     description: |
       Configuration for Private Service Connect (PSC) for the instance.
     properties:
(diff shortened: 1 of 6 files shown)


mmv1/third_party/terraform/verify/validation.go
```diff
@@ -248,6 +249,18 @@ func ValidateProjectID() schema.SchemaValidateFunc {
 	}
 }
 
+func ValidateProjectNumber() schema.SchemaValidateFunc {
```
melinath, line 252: This function should be added via a `custom_code.constants` file as described here: https://googlecloudplatform.github.io/magic-modules/develop/custom-code/#add-reusable-variables-and-functions

Or, since this is an extremely simple regex check, you could use the `regex` options for the `item_validation` field: https://googlecloudplatform.github.io/magic-modules/develop/field-reference/#item_validation

melinath (changes requested):
I'm not sure why you're getting so many failures in VCR; these changes don't seem like they should be impacting that many tests.

mmv1/third_party/terraform/verify/validation.go
vmiglani: Addressed, thanks Stephen!

vmiglani:
@melinath could you please take a look? The failing tests are not related to my change. Also, I have made the changes as we discussed.

mmv1/products/alloydb/Instance.yaml
```diff
@@ -279,6 +279,33 @@ properties:
             values:
               - :ENCRYPTED_ONLY
               - :ALLOW_UNENCRYPTED_AND_ENCRYPTED
+  - !ruby/object:Api::Type::NestedObject
+    name: 'pscInstanceConfig'
+    default_from_api: true
```
melinath, line 284: Is this necessary? I don't see this field getting returned by the API in our nightly builds... default_from_api causes field removal to behave in unexpected ways so we try to avoid it where possible. For more information see https://googlecloudplatform.github.io/magic-modules/develop/field-reference/#default_from_api

mmv1/third_party/terraform/services/alloydb/resource_alloydb_instance_test.go
```diff
@@ -704,3 +704,80 @@ data "google_compute_network" "default" {
...
+	return acctest.Nprintf(`
+resource "google_alloydb_instance" "default" {
+  cluster       = google_alloydb_cluster.default.name
+  instance_id   = "tf-test-alloydb-instance%{random_suffix}"
+  instance_type = "PRIMARY"
+  machine_config {
+    cpu_count = 2
+  }
+  psc_instance_config {
+	allowed_consumer_projects = ["${data.google_project.project.number}", "1044355742748"]
```
melinath, line 768: will this cause problems if the second project is ever deleted? If so, it would be great to set up a new project as part of these tests and reference its number instead.

mmv1/third_party/terraform/services/alloydb/resource_alloydb_cluster_test.go
```diff
@@ -1295,3 +1295,39 @@ resource "google_compute_network" "default" {
...
+
+	context := map[string]interface{}{
+		"random_suffix": acctest.RandString(t, 10),
+	}
+
+	acctest.VcrTest(t, resource.TestCase{
+		PreCheck:                 func() { acctest.AccTestPreCheck(t) },
+		ProtoV5ProviderFactories: acctest.ProtoV5ProviderFactories(t),
+		CheckDestroy:             testAccCheckAlloydbClusterDestroyProducer(t),
+		Steps: []resource.TestStep{
```
melinath, line 1311: could you add a step that disables psc as well?

melinath (changes requested):

mmv1/third_party/terraform/services/alloydb/resource_alloydb_cluster_test.go
vmiglani: PSC cannot be disabled, once it is enabled. That is, a cluster that has PSC enabled at the time of creation can never disable it in future.

mmv1/third_party/terraform/services/alloydb/resource_alloydb_instance_test.go
vmiglani: Possibly, but the project used here is the project used for our PSC probers, so it will not be deleted.

mmv1/products/alloydb/Instance.yaml
vmiglani: Removed

mmv1/third_party/terraform/services/alloydb/resource_alloydb_instance_test.go
melinath: ack, if you're okay with exposing that project number and relying on it then that's fine.

melinath (approved):
LGTM

PR #10670, opened 2024-05-11
Add Managed Folder Resource

hassannmoussaa:
Adding GCS Managed Folder support in terraform provider

Related Issue : https://github.com/hashicorp/terraform-provider-google/issues/18053

```release-note:new-resource 
`google_storage_managed_folder`
```
Terraform configuration 

```
resource "google_storage_managed_folder" "managed_folder" {
    bucket = "bucket here"
    name   = "folder_name"
}
```

PR #7318, opened 2023-02-17
fixed virtual field update issues

edwardmedia:
Fixes https://github.com/hashicorp/terraform-provider-google/issues/13749

If this PR is for Terraform, I acknowledge that I have:

- [x] Searched through the [issue tracker](https://github.com/hashicorp/terraform-provider-google/issues) for an open issue that this either resolves or contributes to, commented on it to claim it, and written "fixes {url}" or "part of {url}" in this PR description. If there were no relevant open issues, I opened one and commented that I would like to work on it (not necessary for very small changes).
- [ ] Ensured that all new fields I added that can be set by a user appear in at least one [example](https://github.com/GoogleCloudPlatform/magic-modules/tree/main/mmv1/templates/terraform/examples) (for generated resources) or [third_party test](https://github.com/GoogleCloudPlatform/magic-modules/tree/main/mmv1/third_party/terraform/tests) (for handwritten resources or update tests).
- [x] [Generated Terraform providers](https://github.com/GoogleCloudPlatform/magic-modules/blob/main/docs/content/docs/getting-started/generate-providers.md), and ran [`make test` and `make lint`](https://googlecloudplatform.github.io/magic-modules/docs/getting-started/run-provider-tests/#run-unit-tests) in the generated providers to ensure it passes unit and linter tests.
- [x] [Ran](https://github.com/GoogleCloudPlatform/magic-modules/blob/main/docs/content/docs/getting-started/run-provider-tests.md) relevant acceptance tests using my own Google Cloud project and credentials (If the acceptance tests do not yet pass or you are unable to run them, please let your reviewer know).
- [x] Read the [Release Notes Guide](https://github.com/GoogleCloudPlatform/magic-modules/blob/main/.ci/RELEASE_NOTES_GUIDE.md) before writing my release note below.

**Release Note Template for Downstream PRs (will be copied)**

```release-note:bug
spanner: fixed the error when updating `deletion_protection` on `google_spanner_database` 
```

```release-note:bug
spanner: fixed the error when updating `force_destroy` on `google_spanner_instance`
```

PR #10786, opened 2024-05-24
Add support for GCS managed folders

juliocc:
This PR adds support for GCS Managed Folders

Related Issue : hashicorp/terraform-provider-google#18053

```release-note:new-resource
`google_storage_managed_folder`
```

```release-note:new-resource
`google_storage_managed_folder_iam`
```


juliocc:
I just noticed that #10670 was also addressing the same issue but it seems inactive for the last few days.

mmv1/templates/terraform/examples/storage_managed_folder_nested.tf.erb
```diff
@@ -0,0 +1,15 @@
+resource "google_storage_bucket" "bucket" {
+  name                        = "<%= ctx[:vars]['bucket_name'] %>"
+  location                    = "EU"
+  uniform_bucket_level_access = true
```
trodge, line 4: Please add `force_destroy = true` to this bucket so that it will be deleted even if there are still objects in it.

juliocc: done

kautikdk:
Hi @trodge, Can you please prioritize this PR review?

melinath:
@trodge are there any blockers here? If this implementation is complete it would be great to get it merged; the other PR seems to have stalled out.

mmv1/products/storage/ManagedFolder.yaml
```diff
@@ -0,0 +1,104 @@
...
+  skip_import_test: true
+  custom_diff_suppress: 'templates/terraform/iam/storage_managed_folder_diff_suppress.go.erb'
+  example_config_body: 'templates/terraform/iam/example_config_body/storage_managed_folder.tf.erb'
+import_format:
+  - '{{bucket}}/managedFolders/{{%name}}'
+  - '{{bucket}}/{{%name}}'
+examples:
+  - !ruby/object:Provider::Terraform::Examples
+    name: 'storage_managed_folder_basic'
+    primary_resource_id: 'folder'
```
melinath, line 63: The first example is also used to generate IAM tests, so it needs to define a primary_resource_name.

```suggestion
    primary_resource_id: 'folder'
    primary_resource_name: 'fmt.Sprintf("tf-test-my-bucket%s", context["random_suffix"]), fmt.Sprintf("tf-test-managed/folder/name%s", context["random_suffix"])'
```

mmv1/products/storage/ManagedFolder.yaml
```diff
@@ -0,0 +1,104 @@
...
+    vars:
+      bucket_name: 'my-bucket'
```
melinath, line 65: You also need to control the managed folder name with a variable. For example:

```suggestion
      bucket_name: 'my-bucket'
      managed_folder_name: 'managed/folder/name/'
```

The vars here need to match what's in `primary_resource_name` for the IAM import tests to work correctly.

It could make sense to have this be the nested example so that nested folder behavior for IAM gets tested as well, since that's more complex.

mmv1/templates/terraform/examples/storage_managed_folder_nested.tf.erb
```diff
@@ -0,0 +1,16 @@
...
+  force_destroy               = true
+}
+
+resource "google_storage_managed_folder" "parent" {
```
melinath, line 8: According to https://cloud.google.com/storage/docs/managed-folders#considerations you don't need to make the parent folder managed to make the nested folder managed.

mmv1/products/storage/ManagedFolder.yaml
```diff
@@ -0,0 +1,104 @@
...
+  - !ruby/object:Provider::Terraform::Examples
+    name: 'storage_managed_folder_nested'
+    primary_resource_id: 'folder'
+    vars:
+      bucket_name: 'my-bucket'
+parameters:
+  - !ruby/object:Api::Type::ResourceRef
+    name: 'bucket'
+    resource: 'Bucket'
```
melinath, line 74: Looks like this is consistent with other storage resources. 👍

mmv1/products/storage/ManagedFolder.yaml
```diff
@@ -0,0 +1,104 @@
...
+    # API returns values with trailing slashes, even if not provided.
+    # Enforcing trailing slashes prevents diffs and ensures consistent
+    # output.
+    validation: !ruby/object:Provider::Terraform::Validation
+      regex: '/$'
+properties:
+  - !ruby/object:Api::Type::String
+    name: createTime
+    description: |
+      Output only. The timestamp at which this managed folder was created.
```
melinath, line 93: This will be automatically added.

```suggestion
      The timestamp at which this managed folder was created.
```

mmv1/products/storage/ManagedFolder.yaml
```diff
@@ -0,0 +1,104 @@
...
+    output: true
+  - !ruby/object:Api::Type::String
+    name: updateTime
+    description: |
+      Output only. The timestamp at which this managed folder was most recently updated.
```
melinath, line 98:
```suggestion
      The timestamp at which this managed folder was most recently updated.
```

mmv1/products/storage/ManagedFolder.yaml
```diff
@@ -0,0 +1,104 @@
...
+    output: true
+  - !ruby/object:Api::Type::String
+    name: metageneration
+    description: |
+      Output only. The metadata generation of the managed folder.
```
melinath, line 103:
```suggestion
      The metadata generation of the managed folder.
```

melinath (changes requested):
Requested some changes below.

melinath:
@kautikdk I don't think it needs to be a blocker for this PR but it might be worth adding a flag to this resource to allow deletion of nonempty managed folders, similar to [`google_storage_bucket.force_destroy`](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/storage_bucket#force_destroy). Looks like for managed folders the query param is called [`allowNonEmpty`](https://cloud.google.com/storage/docs/json_api/v1/managedFolder/delete) but that wouldn't really track as a resource field name - maybe either `delete_allow_non_empty` or `force_destroy` (to match `google_storage_bucket.force_destroy`)?

If the priority is getting this resource out we can ship it without that field and add it in a separate PR, since it may introduce some complications.

kautikdk:
Hi @melinath, Got your point. It make sense to add new field to force destroy managed folder even if it is non-empty. The filed would be more of like virtual field as we won't be sent field in the INSERT/PATCH methods and won't get it from the response as well. It will be only used in deletion of managed folder. I am not sure how virtual fields are being implemented in MMV1 resources but if It's not possible to add it in the MMV1 resource without modifying 2 or more CRUD methods then we might have to move this resource to the hand-written resources.  
As we know it's query parameter and not a blocker of delete method, We can track it in the separate Feature Request. We will prioritize this and add support ASAP.

melinath:
@kautikdk SGTM. We do have support for [virtual fields like `force_delete` in mmv1](https://github.com/GoogleCloudPlatform/magic-modules/blob/5e23d34e0912807f0b7ba28cb8102e4fcd667222/mmv1/products/monitoring/NotificationChannel.yaml#L60-L70) though it is not a well documented process. Definitely reach out if you have further questions while working on the new ticket; don't want to take up too much space here.

juliocc:
@melinath thanks for the comments. Will start working on them today.

melinath (changes requested):
see above

kautikdk:
Hi @juliocc, Thank you for this contribution. 
Also please prioritize this required PR changes so that we can complete review and go ahead with merge. Let me know if you need any support from our side.

kautikdk:
@melinath I don't know much about auto generation of MMV1 resource documentation. 
Just want to confirm whether auto generated resource samples will be enough or should we consider adding more samples ourselves to clearly show use cases of managed folders?

melinath:
We only require basic / full samples but you're welcome to add more if you think it would be useful. However, too many samples can end up making it difficult to read the docs. You can copy the generated docs into Hashicorp's [Doc Preview Tool](https://registry.terraform.io/tools/doc-preview) to see what it would look like.

Generally speaking, I'd recommend including more complete use cases on cloud.google.com alongside other samples - see go/terraform-on-cgc for more details.

kautikdk:
Hi @juliocc, Looks like there are still some pending required changes from Stephen. Can you please consider them as well?

juliocc:
> Hi @juliocc, Looks like there are still some pending required changes from Stephen. Can you please consider them as well?

Yeah, I just merged a few suggestions directly from the Github UI. I'm still working through all other points and will submit the code shortly

melinath (changes requested):
ack, removing from my queue in the meantime

mmv1/products/storage/ManagedFolder.yaml
```diff
@@ -0,0 +1,104 @@
...
+  bucket-level access enabled.
+references: !ruby/object:Api::Resource::ReferenceLinks
+  guides:
+    'Official Documentation': 'https://cloud.google.com/storage/docs/managed-folders'
+  api: 'https://cloud.google.com/storage/docs/json_api/v1/managedFolder'
+iam_policy: !ruby/object:Api::Resource::IamPolicy
+  allowed_iam_role: 'roles/storage.objectViewer'
+  admin_iam_role: 'roles/storage.admin'
+  parent_resource_attribute: 'managed_folder'
+  base_url: 'b/{{bucket}}/managedFolders/{{%managed_folder}}'
```
kautikdk, line 43: Hi @melinath, @trodge. Just wanted to confirm that Here path escape character(%) works or not? 
I looked at some generated IAM files and found that url is not being constructed with url encoding of path parameters. For example, [this code](https://github.com/hashicorp/terraform-provider-google/blob/main/google/services/sourcerepo/iam_sourcerepo_repository.go#L212), I can see that `urlTemplate` is being constructed directly from the parameters value without any url encoding. 
I do see all the generated tests passed but It might be because of "managed-folder-1/" resource name. We basically don't need any encoding here as there is only one slash here which is at the end. It may fail for nested managed folders like "a/b/".

mmv1/templates/terraform/iam/storage_managed_folder_diff_suppress.go.erb
```diff
@@ -0,0 +1,14 @@
+func <%= object.resource_name -%>DiffSuppress(_, old, new string, _ *schema.ResourceData) bool {
```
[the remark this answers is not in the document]
melinath: If this would be preferable please let me know and I'll implement it. In the meantime I've added https://github.com/hashicorp/terraform-provider-google/issues/18415 to track implementation of this feature.

mmv1/products/storage/ManagedFolder.yaml
melinath: +1 that we should test the IAM behavior with nested folders to be safe. The code [looks correct to me](https://github.com/hashicorp/terraform-provider-google/blob/a4438d4339a4386b018c63d01cd43a97dc4ab431/google/services/sourcerepo/iam_sourcerepo_repository.go#L212) but also I don't see any difference from the google_sourcerepo_repository_iam code you linked (and that [doesn't override the base_url for IAM at all](https://github.com/GoogleCloudPlatform/magic-modules/blob/c07e40b24de7aeed6faa445a659ea5ff28ff50f6/mmv1/products/sourcerepo/Repository.yaml#L29).) It could be that the % is a no-op here.

kautikdk: Actually I took this url constructor reference just to ask that I don't see any url encoding happening in the IAM resource. As you show that MMV1 yaml is not having any path escaper for IAM, it won't have any logic in generated resource but I tried manually with `%` character and No difference. It is generating urlTemplate like `{{StorageBasePath}}b/a/managedFolders/b//iam` which I think doesn't make any sense and results in 404 not found error.

juliocc: I just tested this locally and you're absolutely right.  The current approach doesn't work for nested folders. I don't see any hooks in [iam_policy.erb](https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/templates/terraform/iam_policy.go.erb#L307) to modify the URL, so it looks like we'll have to write it manually.

Is there any documentation on how create a handwritten IAM policy resource? Should I start from the generated code and update it manually?

melinath: So if I understand correctly, the problem is that the managed folder path has to end with a "/" (ideally, if we could set up validation for the parent resource field), but when you join that with the IAM methods it ends up having an extra slash? If that's correct, you could try adding setting the method separator to an empty string, since the trailing "/" would act as the separator on its own.

melinath: oh - no, the problem is that the managed folder path is supposed to be URL escaped.

melinath: Unfortunately, as far as I know, the `%` just indicates that the given part of the path can contain slashes; it doesn't mean that we urlencode that part of the path. If that's necessary here, then either users need to provide the path already escaped (for example using the Terraform [`urlencode`](https://developer.hashicorp.com/terraform/language/functions/urlencode) function might work?) or the IAM support for this resource needs to be handwritten.

Adding support for this as a feature of the generator would be a pretty significant lift, I think, so it is not something I'd want to commit to in the context of this PR.

juliocc: Asking the user to urlencode might break other things. Let me give it a try and I'll get back to you

kautikdk: +1. I think, we should do url encoding from provider side. As it appears This is the special case for IAM auto generation, We can solve this by moving this resource to Handwritten and applying custom encoding logic.

kautikdk: @melinath, Can it be fixed with MMV1 auto generation? 
We may require this url encoding(path escape) with upcoming new features. If it is possible to add path escape support in auto generated IAM resource then we won't require moving IAM resource to handwritten and no need to write custom logic.

mmv1/templates/terraform/examples/storage_managed_folder_nested.tf.erb
juliocc: The idea here was to create a managed folder inside another managed folder, while the first example is just creating a managed folder at a given path. 

Let me know if this makes sense, otherwise we can keep the first example only.

mmv1/templates/terraform/iam/storage_managed_folder_diff_suppress.go.erb
```diff
@@ -0,0 +1,14 @@
...
+	// Standard IAM policies use the expanded base URL for the
+	// resource's last parameter. Managed folders have a unique
+	// endpoint format
+	// ('b/{{bucket}}/managedFolders/{{%managed_folder}}'), so we
+	// adjust extraction to correctly identify the folder name.
+
+	oldParts := strings.SplitN(old, "/managedFolders/", 2)
+	if len(oldParts) == 2 {
+		oldFolderName := oldParts[1] + "/"
```
kautikdk, line 10: I think `managed_folder` will have trailing `/` at the end. Do we need to add "/" separately?

mmv1/products/storage/ManagedFolder.yaml
kautikdk: If It requires significant effort to modify generator then we should move IAM to handwritten by following guidelines given [here](https://googlecloudplatform.github.io/magic-modules/develop/resource/#add-iam-support). 
In HandWritten resource, I think we need to only change `qualifyManagedFolderUrl` function with path escaping logic. Probably Something like this should work,
`urlTemplate := fmt.Sprintf("{{StorageBasePath}}%s/%s", fmt.Sprintf("b/%s/managedFolders/%s", u.bucket, url.PathEscape(u.managedFolder)), methodIdentifier)`

mmv1/templates/terraform/iam/storage_managed_folder_diff_suppress.go.erb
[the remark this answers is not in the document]
kautikdk: Any Updates on this?

kautikdk:
Hi @juliocc, Left some comments. Also can you please prioritize this PR pending work?

mmv1/templates/terraform/iam/storage_managed_folder_diff_suppress.go.erb
melinath: See discussion at https://github.com/GoogleCloudPlatform/magic-modules/pull/10786#discussion_r1636809896

mmv1/products/storage/ManagedFolder.yaml
melinath: Yeah, we probably don't want to handle this specially for IAM right now. If there are additional resources that need it we should discuss that separately (so this resources isn't blocked) and you can look into converting back to generated later.

melinath (commented):
I'm not sure where to put this exactly but since a number of the managedFolder APIs seem to require the managedFolder path to be escaped, that should be possible to do using pre_read, pre_update, and pre_delete custom code. https://googlecloudplatform.github.io/magic-modules/develop/custom-code/#inject-code-before--after-crud-operations-and-import

mmv1/templates/terraform/examples/storage_managed_folder_nested.tf.erb
melinath: We should definitely have a test that creates a nested folder (and it should be the first example so that IAM tests use it as well). However, the nested folder doesn't need a "parent" folder. Users should be able to figure out how to do that if they want to; I don't think we need to document it separately, since it's not enforced by GCS.

mmv1/products/storage/ManagedFolder.yaml
juliocc: Added the extra vars, but with the manual IAM resources I don't think it matters too much

mmv1/templates/terraform/iam/storage_managed_folder_diff_suppress.go.erb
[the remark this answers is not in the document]
juliocc: With the handwritten IAM resources this is not necessary anymore.

mmv1/templates/terraform/iam/storage_managed_folder_diff_suppress.go.erb
juliocc: Ack. Removed

juliocc:
@melinath @kautikdk I pushed new code implementing what has been discussed here. To summarize:
- Moved IAM to a handwritten resource
- Added force_destroy virtual field mapped to the allowNonEmpty query parameter
- Minor documentation fixes

@melinath I'm having issues with the IAM tests. I'm trying a simple example such as this:
```
resource "google_storage_managed_folder" "folder" {
  bucket = google_storage_bucket.bucket.name
  name   = "managed/folder/name123"
}

resource "google_storage_managed_folder_iam_binding" "foo" {
  bucket         = google_storage_managed_folder.folder.bucket
  managed_folder = google_storage_managed_folder.folder.name
  role           = "roles/storage.objectViewer"
  members        = ["user:xxx@example.com"]
}
```

If I run this I get an inconsistent state error:
```
│ Error: Provider produced inconsistent final plan
│
│ When expanding the plan for google_storage_managed_folder_iam_binding.foo to include new values learned so far during apply, provider "registry.terraform.io/hashicorp/google" produced
│ an invalid new value for .managed_folder: was cty.StringVal("managed/folder/name123"), but now cty.StringVal("managed/folder/name123/").
│
│ This is a bug in the provider, which should be reported in the provider's own issue tracker.
╵
```
However, it succeeds if I create the folder first and then in a separate run I create the binding:
```
❯ tfa
╷
│
│  - hashicorp/google in REDACTED
│
╵
google_storage_bucket.bucket: Refreshing state... [id=tf-test-my-bucket123dd1]

  + create

  # google_storage_managed_folder.folder will be created
  + resource "google_storage_managed_folder" "folder" {
      + bucket         = "tf-test-my-bucket123dd1"
      + create_time    = (known after apply)
      + force_destroy  = false
      + id             = (known after apply)
      + metageneration = (known after apply)
      + name           = "managed/folder/name123"
      + self_link      = (known after apply)
      + update_time    = (known after apply)
    }

  Enter a value: yes

google_storage_managed_folder.folder: Creating...
google_storage_managed_folder.folder: Creation complete after 0s [id=tf-test-my-bucket123dd1/managed/folder/name123]

Apply complete! Resources: 1 added, 0 changed, 0 destroyed.



❯ tfa
╷
│
│  - hashicorp/google in REDACTED
│
╵
google_storage_bucket.bucket: Refreshing state... [id=tf-test-my-bucket123dd1]
google_storage_managed_folder.folder: Refreshing state... [id=tf-test-my-bucket123dd1/managed/folder/name123]

  + create

  # google_storage_managed_folder_iam_binding.foo will be created
  + resource "google_storage_managed_folder_iam_binding" "foo" {
      + bucket         = "tf-test-my-bucket123dd1"
      + etag           = (known after apply)
      + id             = (known after apply)
      + managed_folder = "managed/folder/name123/"
      + members        = [
          + "user:xxx@example.com",
        ]
      + role           = "roles/storage.objectViewer"
    }

  Enter a value: yes

google_storage_managed_folder_iam_binding.foo: Creating...
google_storage_managed_folder_iam_binding.foo: Creation complete after 4s [id=b/tf-test-my-bucket123dd1/managedFolders/managed/folder/name123//roles/storage.objectViewer]

Apply complete! Resources: 1 added, 0 changed, 0 destroyed.


```

This behavior is causing all the IAM tests to fail. Any ideas?

melinath:
@juliocc That error is most likely because you don't have a diff suppress func on the IAM resource's managed_folder field. However, I'd recommend using validation funcs to require the user to include the trailing slash rather than a diff suppress func. That should be possible now that the IAM resources are handwritten.

melinath (changes requested):
Please also add an [update test](https://googlecloudplatform.github.io/magic-modules/develop/test/test/#add-an-update-test)

I've opened https://github.com/hashicorp/terraform-provider-google/issues/18514 to track the IAM PathEscape issue.

mmv1/products/storage/ManagedFolder.yaml
```diff
@@ -0,0 +1,94 @@
...
+custom_code: !ruby/object:Provider::Terraform::CustomCode
+  constants: templates/terraform/constants/storage_managed_folder.go.erb
+import_format:
+  - '{{bucket}}/managedFolders/{{%name}}'
+  - '{{bucket}}/{{%name}}'
+examples:
+  - !ruby/object:Provider::Terraform::Examples
+    name: 'storage_managed_folder_basic'
+    primary_resource_id: 'folder'
+    # primary_resource_name: 'fmt.Sprintf("tf-test-my-bucket%s", context["random_suffix"]), fmt.Sprintf("tf-test-managed/folder/name%s", context["random_suffix"])'
```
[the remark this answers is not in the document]
juliocc: Done

juliocc:
> Please also add an [update test](https://googlecloudplatform.github.io/magic-modules/develop/test/test/#add-an-update-test)

The resource is immutable. Do we really need an update test?

melinath:
> > Please also add an [update test](https://googlecloudplatform.github.io/magic-modules/develop/test/test/#add-an-update-test)
> 
> The resource is immutable. Do we really need an update test?

no, my bad, I forgot it was immutable.

juliocc:
> > > Please also add an [update test](https://googlecloudplatform.github.io/magic-modules/develop/test/test/#add-an-update-test)
> > 
> > 
> > The resource is immutable. Do we really need an update test?
> 
> no, my bad, I forgot it was immutable.

Got it. In that case, this is ready for review

kautikdk:
LGTM,
@juliocc, Thank you for adding `allowNonEmpty` virtual field in this PR itself. 
@melinath You mentioned that this field may introduce complications. Can you point out any possible drift or any other issue that can arise So that we can mitigate it before hand? Otherwise we can go ahead review and merge!

melinath:
@kautikdk I will need to do a closer review, but basically I just meant that every additional complexity on the resource increases the chance that something will cause an unanticipated delay, so in general it's good to keep things as simple as possible (while still complex enough to catch issues caused by necessary complexity)

mmv1/third_party/terraform/services/storage/iam_storage_managed_folder.go
```diff
@@ -0,0 +1,182 @@
...
+	"bucket": {
+		Type:     schema.TypeString,
+		Required: true,
+		ForceNew: true,
+	},
+	"managed_folder": {
+		Type:             schema.TypeString,
+		Required:         true,
+		ForceNew:         true,
+		DiffSuppressFunc: tpgresource.CompareSelfLinkOrResourceName,
```
melinath, line 26: This should also have a validation func that has the same logic as the [name field on the parent resource](https://github.com/modular-magician/terraform-provider-google-beta/compare/auto-pr-10786-old..auto-pr-10786#diff-fe97e5996e9605a4951a42ccce2c50b9067a734e7e98ccbf98586099062f50a6R61):

```suggestion
		DiffSuppressFunc: tpgresource.CompareSelfLinkOrResourceName,
		ValidateFunc: verify.ValidateRegexp(`/$`),
```

This will require also importing `"github.com/hashicorp/terraform-provider-google-beta/google-beta/verify"`

melinath (changes requested):
need to take a more thorough look in the morning but for now:

mmv1/third_party/terraform/services/storage/iam_storage_managed_folder.go
juliocc: Done

melinath:
The TGC failure is unrelated.

kautikdk:
Then I think The PR is ready for final review. 
@melinath Can you please review it today itself so that It can be merged before EOD and feature will be released in TGP next Monday release.

melinath:
Manual test run for the non-VCR tests added: https://hashicorp.teamcity.com/buildConfiguration/TerraformProviders_GoogleCloud_GOOGLE_BETA_MMUPSTREAMTESTS_GOOGLEBETA_PACKAGE_STORAGE/175646?hideTestsFromDependencies=false&hideProblemsFromDependencies=false&expandBuildChangesSection=true

mmv1/products/storage/ManagedFolder.yaml
```diff
@@ -0,0 +1,94 @@
...
+    required: true
+    # The API returns values with trailing slashes, even if not
+    # provided. Enforcing trailing slashes prevents diffs and ensures
+    # consistent output.
+    validation: !ruby/object:Provider::Terraform::Validation
+      regex: '/$'
+virtual_fields:
+  - !ruby/object:Api::Type::Boolean
+    name: 'force_destroy'
+    immutable: true
```
melinath, line 73: This can't be immutable - otherwise it will not be possible to update to `true` (because that would cause the resource to be destroyed, which could fail because force_destroy is `false`.) Since this is a virtual field, I think the only workaround is to mark the resource as not immutable, and instead mark all writable non-virtual fields as immutable (just bucket and name).

It would be great to have an [update test](https://googlecloudplatform.github.io/magic-modules/develop/test/test/#add-an-update-test) that just verifies the behavior of this field, since virtual fields can be a bit unusual / since the resource is not actually updatable. I think you'll need to override the update method (with [`custom_code.custom_update`](https://googlecloudplatform.github.io/magic-modules/develop/custom-code/#replace-entire-crud-methods)) since otherwise the resource will try to make an invalid API request. For comparison, Spanner had a similar issue with virtual field updates causing invalid API requests (https://github.com/GoogleCloudPlatform/magic-modules/pull/7318).

```suggestion
```

mmv1/third_party/terraform/provider/provider_mmv1_resources.go.erb
```diff
@@ -252,6 +252,7 @@ var handwrittenIAMDatasources = map[string]*schema.Resource{
 	"google_kms_crypto_key_iam_policy":             tpgiamresource.DataSourceIamPolicy(kms.IamKmsCryptoKeySchema, kms.NewKmsCryptoKeyIamUpdater),
 	"google_spanner_instance_iam_policy":           tpgiamresource.DataSourceIamPolicy(spanner.IamSpannerInstanceSchema, spanner.NewSpannerInstanceIamUpdater),
 	"google_spanner_database_iam_policy":           tpgiamresource.DataSourceIamPolicy(spanner.IamSpannerDatabaseSchema, spanner.NewSpannerDatabaseIamUpdater),
+    "google_storage_managed_folder_iam_policy":     tpgiamresource.DataSourceIamPolicy(storage.StorageManagedFolderIamSchema, storage.StorageManagedFolderIamUpdaterProducer),
```
melinath, line 255: nit: please fix the indentation in this file to match.

mmv1/third_party/terraform/website/docs/r/storage_managed_folder_iam.html.markdown
```diff
@@ -0,0 +1,203 @@
...
+exported:
+
+* `etag` - (Computed) The etag of the IAM policy.
+
+## Import
+
+For all import syntaxes, the "resource in question" can take any of the following forms:
+
+* b/{{bucket}}/managedFolders/{{managed_folder}}
+* {{managed_folder}}
```
melinath, line 181:
```suggestion
* {{bucket}}/{{managed_folder}}
```

melinath (changes requested):
the force_destroy field won't work as implemented - please fix or remove. (I would recommend just removing for now, since it's lower risk to implement the field in a separate release.)

mmv1/third_party/terraform/provider/provider_mmv1_resources.go.erb
juliocc: Done

mmv1/third_party/terraform/website/docs/r/storage_managed_folder_iam.html.markdown
juliocc: Done

mmv1/products/storage/ManagedFolder.yaml
juliocc: Removing the whole field for now as suggested. Opened hashicorp/terraform-provider-google#18553 to track this.

mmv1/third_party/terraform/website/docs/r/storage_managed_folder_iam.html.markdown
```diff
@@ -178,7 +178,7 @@ exported:
 For all import syntaxes, the "resource in question" can take any of the following forms:
 
 * b/{{bucket}}/managedFolders/{{managed_folder}}
-* {{managed_folder}}
+* bucket/{{managed_folder}}
```
melinath, line 181:
```suggestion
* {{bucket}}/{{managed_folder}}
```

melinath (changes requested):

melinath (approved):
I believe this should be good to go.

kautikdk:
@juliocc Thank you very much for the contribution! Hope you will contribute again in future.
@melinath Thank you for your guidance throughout the process.

PR #10793, opened 2024-05-24
Add support for specifying storage pool when creating a disk.

sizzle0121:
Add field `storage_pool` to `google_compute_disk` resource to support creating disk with the storage pool in which it will reside.
Instead of adding a resource for storage pool and using resourceRef to let disk reference it, we add the field as a string which will contain the URL for a storage pool. The main reason for this decision is that if we add storage pool as a resource, and in the scenario that users want to destroy a storage pool, Terraform will plan to delete all disks first and then delete the storage pool, which we definitely want to prevent from happening.
Therefore, we eventually decided to not support creating storage pool via Terraform, and only support adding the field for disks to specify existing storage pool. The management of storage pool itself will remain through GCP APIs.

```release-note:enhancement
compute: added `storage_pool` field to `google_compute_disk` resource
```

### DIFF
--- a/mmv1/third_party/terraform/services/compute/resource_compute_disk_test.go.erb
+++ b/mmv1/third_party/terraform/services/compute/resource_compute_disk_test.go.erb
@@ -3,14 +3,17 @@ package compute_test
 
 import (
 	"fmt"
+	"net/http"
 	"os"
 	"testing"
+	"time"
 
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/terraform"
 	"github.com/hashicorp/terraform-provider-google/google/acctest"
 	"github.com/hashicorp/terraform-provider-google/google/envvar"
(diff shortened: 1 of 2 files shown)


sizzle0121:
Hi @c2thorn, I think the failed tests are not related to my changes, please review, thanks!

mmv1/third_party/terraform/services/compute/resource_compute_disk_test.go.erb
```diff
@@ -1516,3 +1519,109 @@ resource "google_compute_disk" "foobar" {
...
+				ImportState:       true,
+				ImportStateVerify: true,
+			},
+		},
+	})
+
+	cleanupTestingStoragePool(t, storagePoolName)
+}
+
+func setupTestingStoragePool(t *testing.T, storagePoolName string) func() {
```
c2thorn, line 1551: Should Compute storage pool be a Terraform resource eventually?

c2thorn:
re-running tests to hopefully avoid some of the flakey tests

PR #10798, opened 2024-05-25
Add secrets_discovery_target field in google_data_loss_prevention_discovery_config, as well as fields to support single-resource mode for big_query_target and cloud_sql_target

patrickmoy:
```release-note:enhancement
dlp: added `secrets_discovery_target` field to `google_data_loss_prevention_discovery_config` resource
```

```release-note:enhancement
dlp: added `cloud_sql_target.filter.database_resource_reference` and `big_query_target.filter.table_reference` fields to `google_data_loss_prevention_discovery_config` resource to support single-resource mode
```
Also disable some example tests, since DiscoveryConfigs are limited to 1-per-region-per-storage-type at the project level, and the parallelism applied to example tests causes flakiness and interferes with update tests.

### DIFF
--- a/mmv1/products/dlp/DiscoveryConfig.yaml
+++ b/mmv1/products/dlp/DiscoveryConfig.yaml
@@ -32,17 +32,20 @@ id_format: '{{parent}}/discoveryConfigs/{{name}}'
 examples:
   - !ruby/object:Provider::Terraform::Examples
     name: 'dlp_discovery_config_basic'
+    skip_test: true
     primary_resource_id: 'basic'
     test_env_vars:
       project: :PROJECT_NAME
       location: :REGION
   - !ruby/object:Provider::Terraform::Examples
     name: 'dlp_discovery_config_actions'
+    skip_test: true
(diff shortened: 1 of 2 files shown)


patrickmoy:
Commits squashed into one; tests are passing now.

NickElliot (changes requested):
LGTM for new changes -- but we need to have the old tests replicated on a non-parallel area for adequate coverage if that is causing problems with the new update tests. For example, the previous `dlp_discovery_config_actions` testing the `actions{}` block, a new test needs to exist so as to cover all of the scenarios covered within the previous examples (not necessarily a new handwritten test for all of them, but all fields need to be in a test).

patrickmoy:
I believe all of the example fields are covered in update tests, since when I originally wrote the first set of update tests, I based them off of examples (starting with a basic barebones config, and then updating to the example). 

Does that suffice?

NickElliot (approved):
Sorry about that, LGTM! Had some of the pages of diffs vs existing files mixed up.

PR #10818, opened 2024-05-29
Removed required tag on maxFindingsPerInfoType.infoType

patrickmoy:
Fixes https://github.com/hashicorp/terraform-provider-google/issues/18149
```release-note:bug
dlp: removed required on  `google_data_loss_prevention_inspect_template` to allow use of default `per_info_type` limit by not setting the `info_type` field
```


slevenick (changes requested):
Can you add a test that doesn't specify this field just so we can make sure it all works as expected?

patrickmoy:
Added test that excludes the formerly required field

slevenick (approved):

PR #8060, opened 2023-06-01
Remove default from optional `rule_visibility` field

smcgivern:
`google_compute_security_policy` -> `adaptive_protection_config` -> `layer_7_ddos_defense_config` -> `rule_visibility` is optional, but had a default value, and was required to be either PREMIUM or STANDARD.

This means that:

1. If you did not include this block at all, you get a persistent message about changing this from `false` (in state) to `null` (in config, due to the missing default).
2. If you included this block with `enable = false`, you'd get a plan attempting to set `rule_visibility = "STANDARD"`, which would then fail:

   > Invalid value for field
   > 'resource.adaptiveProtectionConfig.layer7DdosDefenseConfig.enable':
   > 'false'. Cannot set layer 7 DDoS options if it's not enabled.

This fixes the second problem, but not the first.

Part of https://github.com/hashicorp/terraform-provider-google/issues/12743

If this PR is for Terraform, I acknowledge that I have:

- [x] Searched through the [issue tracker](https://github.com/hashicorp/terraform-provider-google/issues) for an open issue that this either resolves or contributes to, commented on it to claim it, and written "fixes {url}" or "part of {url}" in this PR description. If there were no relevant open issues, I opened one and commented that I would like to work on it (not necessary for very small changes).
- [ ] Ensured that all new fields I added that can be set by a user appear in at least one [example](https://github.com/GoogleCloudPlatform/magic-modules/tree/main/mmv1/templates/terraform/examples) (for generated resources) or [third_party test](https://github.com/GoogleCloudPlatform/magic-modules/tree/main/mmv1/third_party/terraform/tests) (for handwritten resources or update tests).
- [x] [Generated Terraform providers](https://github.com/GoogleCloudPlatform/magic-modules/blob/main/docs/content/docs/getting-started/generate-providers.md), and ran [`make test` and `make lint`](https://googlecloudplatform.github.io/magic-modules/docs/getting-started/run-provider-tests/#run-unit-tests) in the generated providers to ensure it passes unit and linter tests.
- [ ] [Ran](https://github.com/GoogleCloudPlatform/magic-modules/blob/main/docs/content/docs/getting-started/run-provider-tests.md) relevant acceptance tests using my own Google Cloud project and credentials (If the acceptance tests do not yet pass or you are unable to run them, please let your reviewer know).
- [x] Read the [Release Notes Guide](https://github.com/GoogleCloudPlatform/magic-modules/blob/main/.ci/RELEASE_NOTES_GUIDE.md) before writing my release note below.

[...]

PR #10823, opened 2024-05-29
Make `google_compute_security_policy`'s `rule_visibility` optional + computed and remove default value

SarahFrench:
Supersedes https://github.com/GoogleCloudPlatform/magic-modules/pull/8060
Might fix https://github.com/hashicorp/terraform-provider-google/issues/12743 ?

I find that issue above confusing as it includes 2 issues:
1. Being unable to create a google_compute_security_policy resource when this is in the config:
```
adaptive_protection_config {
      layer_7_ddos_defense_config {
          enable = false
      }
}
```
2. Issues/expected behaviour when the adaptive_protection_config block is removed:
```diff
- adaptive_protection_config {
-      layer_7_ddos_defense_config {
-          enable = false
-      }
-}
```

This PR addresses issue 1.
[I explored issue 2](https://github.com/GoogleCloudPlatform/magic-modules/pull/10823#issuecomment-2142542094) but [decided that it'd be best to address that separately to 1.](https://github.com/GoogleCloudPlatform/magic-modules/pull/10823#issuecomment-2142696373)

```release-note:bug
compute: fixed issue where users could not create `google_compute_security_policy` resources with `layer_7_ddos_defense_config` explicitly disabled
```

### DIFF
--- a/mmv1/third_party/terraform/services/compute/resource_compute_security_policy.go.erb
+++ b/mmv1/third_party/terraform/services/compute/resource_compute_security_policy.go.erb
@@ -496,7 +496,7 @@ func ResourceComputeSecurityPolicy() *schema.Resource {
 									"rule_visibility": {
 										Type: schema.TypeString,
 										Optional: true,
-										Default: "STANDARD",
+										Computed: true,
 										ValidateFunc: validation.StringInSlice([]string{"STANDARD", "PREMIUM"}, false),
 										Description: `Rule visibility. Supported values include: "STANDARD", "PREMIUM".`,
 									},
--- a/mmv1/third_party/terraform/services/compute/resource_compute_security_policy_test.go.erb
+++ b/mmv1/third_party/terraform/services/compute/resource_compute_security_policy_test.go.erb
@@ -239,7 +239,7 @@ func TestAccComputeSecurityPolicy_withAdaptiveProtection(t *testing.T) {
 		CheckDestroy:             testAccCheckComputeSecurityPolicyDestroyProducer(t),
 		Steps: []resource.TestStep{
 			{
-				Config: testAccComputeSecurityPolicy_withAdaptiveProtection(spName),
+				Config: testAccComputeSecurityPolicy_withAdaptiveProtection_enabled(spName),
 			},
 			{
 				ResourceName:      "google_compute_security_policy.policy",
@@ -247,7 +247,7 @@ func TestAccComputeSecurityPolicy_withAdaptiveProtection(t *testing.T) {
 				ImportStateVerify: true,
 			},
 			{
-				Config: testAccComputeSecurityPolicy_withAdaptiveProtectionUpdate(spName),
+				Config: testAccComputeSecurityPolicy_withAdaptiveProtection_update(spName),
 			},
 			{
 				ResourceName:      "google_compute_security_policy.policy",
@@ -1192,7 +1192,31 @@ resource "google_compute_security_policy" "policy" {
 `, spName)
 }
 
-func testAccComputeSecurityPolicy_withAdaptiveProtection(spName string) string {
+func testAccComputeSecurityPolicy_withoutAdaptiveProtection(spName string) string {
+	return fmt.Sprintf(`
+resource "google_compute_security_policy" "policy" {
+  name        = "%s"
+  description = "updated description"
+}
+`, spName)
+}
+
+func testAccComputeSecurityPolicy_withAdaptiveProtection_disabled(spName string) string {
+	return fmt.Sprintf(`
+resource "google_compute_security_policy" "policy" {
+  name        = "%s"
+  description = "updated description"
+
+  adaptive_protection_config {
+    layer_7_ddos_defense_config {
+      enable = false
+    }
+  }
+}
+`, spName)
+}
+
+func testAccComputeSecurityPolicy_withAdaptiveProtection_enabled(spName string) string {
 	return fmt.Sprintf(`
 resource "google_compute_security_policy" "policy" {
   name        = "%s"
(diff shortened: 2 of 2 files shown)


SarahFrench:
Test failures are due to a quota limit in the VCR project:

```
        Error: Error waiting for Creating Target Pool: Quota 'TARGET_POOLS' exceeded.  Limit: 150.0 globally.
```

I'll run a sweeper from TeamCity

SarahFrench:
There isn't a sweeper for google_compute_target_instance

SarahFrench:
The failures in `TestAccComputeInstanceNetworkIntefaceWithSecurityPolicy` and `TestAccComputeInstance_guestAccelerator ` are unrelated to this PR.

The failure in `TestAccComputeSecurityPolicy_withoutAdaptiveProtection` is due to the final step:

```
			{
				// Can remove adaptive protection block
				Config: testAccComputeSecurityPolicy_withoutAdaptiveProtection(spName),
			},
```

In this step we make this change to the config from the previous step:

```diff
resource "google_compute_security_policy" "policy" {
  name        = "%s"
  description = "updated description"
-  adaptive_protection_config {
-    layer_7_ddos_defense_config {
-      enable = false
-    }
-  }
}
```

And the PATCH request from that update looks like:

```
PATCH /compute/beta/projects/PROJECT_ID/global/securityPolicies/tf-test-aaoujjl9qe?alt=json&prettyPrint=false&updateMask= HTTP/1.1
Host: compute.googleapis.com
User-Agent: google-api-go-client/0.5 Terraform/1.8.3 (+https://www.terraform.io) Terraform-Plugin-SDK/2.33.0 terraform-provider-google-beta/acc
Content-Length: 31
X-Goog-Api-Client: gl-go/1.21.10 gdcl/0.177.0
Accept-Encoding: gzip

{
 "fingerprint": "UQV3Yc2hZ-g="
}
```

SarahFrench:
After the resource has DDOS defence enabled, disabled, and then deleting the adaption_protection_config block when we read the resource's data from the API it will still contain this:

```
 "adaptiveProtectionConfig": {
  "layer7DdosDefenseConfig": {
   "enable": false,
   "ruleVisibility": "STANDARD"
  }
 },
```

Making the test step mentioned in this step pass will require a change to the flatten functions so that fields are selectively read from the API response based on the user's config. If that's a path we'd want to pursue that's ok but I'd like it to be in a PR separate to this one; this PR should only be about removing the default value that makes it impossible to create a resource where `enabled=false` is in the config

slevenick (commented):
Looks like the test is failing due to the security policy?

SarahFrench:
Oh yes it is, though it doesn't look immediately related to the change in this PR.

Here are the failing tests:

- TestAccComputeInstanceNetworkIntefaceWithSecurityPolicy/[two_nics_two_access_configs_update_remove_access_config](https://github.com/hashicorp/terraform-provider-google-beta/blob/132488240c845e929b78570944b13051988bc03c/google-beta/services/compute/resource_compute_instance_test.go#L2986)
  - Uses only [this config](https://github.com/hashicorp/terraform-provider-google-beta/blob/132488240c845e929b78570944b13051988bc03c/google-beta/services/compute/resource_compute_instance_test.go#L8482) that includes `google_compute_region_security_policy` but not `google_compute_security_policy`
 - TestAccComputeInstanceNetworkIntefaceWithSecurityPolicy/[access_config_update_access_config](https://github.com/hashicorp/terraform-provider-google-beta/blob/132488240c845e929b78570944b13051988bc03c/google-beta/services/compute/resource_compute_instance_test.go#L2988)
    - Uses a bunch of config ([1](https://github.com/hashicorp/terraform-provider-google-beta/blob/132488240c845e929b78570944b13051988bc03c/google-beta/services/compute/resource_compute_instance_test.go#L8303),[ 2](https://github.com/hashicorp/terraform-provider-google-beta/blob/132488240c845e929b78570944b13051988bc03c/google-beta/services/compute/resource_compute_instance_test.go#L8892), [3](https://github.com/hashicorp/terraform-provider-google-beta/blob/132488240c845e929b78570944b13051988bc03c/google-beta/services/compute/resource_compute_instance_test.go#L9000))  but all use `google_compute_region_security_policy` only


google_compute_region_security_policy is MMv1 generated while google_compute_security_policy is handwritten, so I don't think there's any schema re-used between them.

SarahFrench:
I'll re-run the tests and hopefully the test will pass in RECORDING mode : /gcbrun

SarahFrench:
TestAccComputeInstanceNetworkIntefaceWithSecurityPolicy has been failing in nightly tests: https://github.com/hashicorp/terraform-provider-google/issues/17838

TestAccComputeInstance_guestAccelerator has been failing in nightly tests: 
https://hashicorp.teamcity.com/test/-2086049782442135804?currentProjectId=TerraformProviders_GoogleCloud_GOOGLE_NIGHTLYTESTS&expandTestHistoryChartSection=true

SarahFrench:
> Are we okay with merging a PR that is failing VCR for valid reasons? I believe we make sure that these always pass

I think so - if a test is failing on main then there should be a test failure ticket open and a PR opened to address the test failure

SarahFrench:
I realised that in https://github.com/GoogleCloudPlatform/magic-modules/pull/10823/commits/ce34db40878383e105734450194fb731a56ed7a6 I intended to remove a step but I removed the whole test.

I intended to remove only:

```
{
	// Can remove adaptive protection block
	Config: testAccComputeSecurityPolicy_withoutAdaptiveProtection(spName),
},
```

I'm going to push a commit to re-add that test

SarahFrench:
The test I returned, TestAccComputeSecurityPolicy_withoutAdaptiveProtection, passed.

The failing test in VCR is failing on main: https://github.com/hashicorp/terraform-provider-google/issues/17838

I'll merge this later after the nightly tests finish

PR #10828, opened 2024-05-29
Add SCC New Management Service

vijaykanthm:
Was part of https://github.com/GoogleCloudPlatform/magic-modules/pull/10769 orignally

```release-note:REPLACEME

```


melinath:
I think this is not actually going to have the intended effect - see https://github.com/GoogleCloudPlatform/magic-modules/pull/10769#issuecomment-2138067760

PR #10834, opened 2024-05-30
go rewrite - fix some more documentation diffs

c2thorn:

```release-note:none

```

### DIFF
--- a/mmv1/api/resource.go
+++ b/mmv1/api/resource.go
@@ -1293,6 +1293,17 @@ func (r Resource) GetPropertyUpdateMasksGroups() map[string][]string {
 	return maskGroups
 }
 
+func FormatDocDescription(desc string) string {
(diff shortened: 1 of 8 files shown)


mmv1/api/resource.go
```diff
@@ -1293,6 +1293,17 @@ func (r Resource) GetPropertyUpdateMasksGroups() map[string][]string {
 	return maskGroups
 }
 
+func FormatDocDescription(desc string) string {
```
zli82016, line 1296: Will it remove all of diffs in the description field or it works for some use cases? 
A comment could be helpful.

c2thorn: It should be all diffs in the documentation description, the Ruby template has specific indentation that this copies.
Added a comment

zli82016 (approved):
LGTM.

PR #10836, opened 2024-05-30
Force MM Upstream nightly sweepers to use hashicorp vcs root

melinath:
I noticed that MM Upstream nightly sweepers were using the modular-magician vcs root, which means they were not getting new sweepers. This change should force the nightly builds to use Hashicorp repositories while still allowing user-initiated runs to default to modular-magician for ease of running sweepers based on PRs.

```release-note:none

```

### DIFF
--- a/mmv1/third_party/terraform/.teamcity/components/builds/build_triggers.kt
+++ b/mmv1/third_party/terraform/.teamcity/components/builds/build_triggers.kt
@@ -17,7 +17,6 @@ import jetbrains.buildServer.configs.kotlin.triggers.schedule
 import jetbrains.buildServer.configs.kotlin.vcs.GitVcsRoot
 
 class NightlyTriggerConfiguration(
-    val vcsRoot: GitVcsRoot,
     val branch: String = DefaultBranchName,
     val nightlyTestsEnabled: Boolean = true,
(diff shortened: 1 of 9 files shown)


SarahFrench (commented):
This changes looks good for setting the VCS root on the trigger, though the builds that the cron trigger is attached to will also have a VCS root [set here](https://github.com/GoogleCloudPlatform/magic-modules/blob/ebf018ad38d3905fa11e83c3d695b15e9b191d1d/mmv1/third_party/terraform/.teamcity/components/builds/build_configuration_sweepers.kt#L67-L70). I'm not sure which one would be used.

I'll create a test project in TeamCity based on the auto branch made for this PR, and we can test it out!

melinath:
I was looking into running tests for this and am now concerned that this overrides the default vcs root for the buildtype instead of only overriding it for the triggered builds 🤔  I'm not entirely sure how to fix it though - maybe I need to pass the override via buildParams somehow?

melinath:
oh you just said the same thing. Thanks for creating a test project!

SarahFrench (approved):
The code looks good to me, though I'll leave it to you to merge. From looking at the test project in TeamCity I can see service sweeper build names are as expected and they use the right VCR roots in their builds. 🚢

PR #10840, opened 2024-05-31
Add Secure Source Manager Repository resource

ericpang777:

```release-note:new-resource
`google_secure_source_manager_repository`
```

### DIFF
--- a/mmv1/products/securesourcemanager/Repository.yaml
+++ b/mmv1/products/securesourcemanager/Repository.yaml
@@ -1,0 +1,280 @@
+# # Copyright 2024 Google Inc.
+# # Licensed under the Apache License, Version 2.0 (the "License");
+# # you may not use this file except in compliance with the License.
+# # You may obtain a copy of the License at
+# #
+# #     http://www.apache.org/licenses/LICENSE-2.0
+# #
+# # Unless required by applicable law or agreed to in writing, software
+# # distributed under the License is distributed on an "AS IS" BASIS,
+# # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# # See the License for the specific language governing permissions and
+# # limitations under the License.
+
+--- !ruby/object:Api::Resource
+name: 'Repository'
+kind: 'securesourcemanager#repository'
+base_url: 'projects/{{project}}/locations/{{location}}/repositories?repository_id={{repository_id}}'
+self_link: 'projects/{{project}}/locations/{{location}}/repositories/{{repository_id}}'
+immutable: true
+description: 'Repositories store source code. It supports all Git SCM client commands and has built-in pull requests and issue tracking. Both HTTPS and SSH authentication are supported.'
+references: !ruby/object:Api::Resource::ReferenceLinks
+  guides:
+    'Official Documentation': 'https://cloud.google.com/secure-source-manager/docs/overview'
+import_format: ['projects/{{project}}/locations/{{location}}/repositories/{{repository_id}}', '{{repository_id}}']
+autogen_async: true
+async: !ruby/object:Api::OpAsync
+  actions: ['create', 'delete']
+  operation: !ruby/object:Api::OpAsync::Operation
+    path: 'name'
+    base_url: '{{op_id}}'
+    wait_ms: 100
+    timeouts: !ruby/object:Api::Timeouts
+      insert_minutes: 1
+      update_minutes: 1
+      delete_minutes: 1
+  result: !ruby/object:Api::OpAsync::Result
+    path: 'response'
+  status: !ruby/object:Api::OpAsync::Status
+    path: 'done'
+    complete: true
+    allowed:
+      - true
+      - false
+  error: !ruby/object:Api::OpAsync::Error
+    path: 'error'
+    message: 'message'
+iam_policy: !ruby/object:Api::Resource::IamPolicy
+  parent_resource_attribute: 'repository_id'
+  method_name_separator: ':'
+  allowed_iam_role: 'roles/securesourcemanager.repoAdmin'
+  import_format: ['projects/{{project}}/locations/{{location}}/repositories/{{repository_id}}']
+examples:
+  - !ruby/object:Provider::Terraform::Examples
+    name: 'secure_source_manager_repository_basic'
+    primary_resource_id: 'default'
+    primary_resource_name: "fmt.Sprintf(\"tf-test-my-repo%s\",
+      context[\"random_suffix\"\
+      ])"
+    vars:
+      instance_id: 'my-instance'
+parameters:
+  - !ruby/object:Api::Type::String
+    name: 'location'
+    description: |
+      The location for the Repository.
+    required: true
+    immutable: true
+    url_param_only: true
+  - !ruby/object:Api::Type::String
+    name: 'repository_id'
+    description: |
+      The name for the Repository.
+    required: true
+    immutable: true
+    url_param_only: true
+properties:
+  - !ruby/object:Api::Type::String
+    name: 'name'
+    description: |
+      The resource name for the Repository.
+    output: true
+    immutable: true
+  - !ruby/object:Api::Type::String
+    name: 'description'
+    description: |
+      Description of the repository, which cannot exceed 500 characters.
+  - !ruby/object:Api::Type::String
+    name: 'instance'
+    description: |
+      The name of the instance in which the repository is hosted.
+    diff_suppress_func: 'tpgresource.ProjectNumberDiffSuppress'
+  - !ruby/object:Api::Type::String
+    name: 'uid'
+    description: |
+      Unique identifier of the repository.
+    output: true
+  - !ruby/object:Api::Type::Time
+    name: 'createTime'
+    description: |
+      Time the repository was created in UTC.
+    output: true
+  - !ruby/object:Api::Type::Time
+    name: 'updateTime'
+    description: |
+      Time the repository was updated in UTC.
+    output: true
+  - !ruby/object:Api::Type::String
+    name: 'etag'
+    description: |
+      This checksum is computed by the server based on the value of other fields, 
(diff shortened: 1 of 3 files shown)


mmv1/templates/terraform/examples/secure_source_manager_repository_basic.tf.erb
```diff
@@ -0,0 +1,21 @@
+resource "google_secure_source_manager_instance" "instance" {
```
xingao267, line 1: Is the example here also used as tests?

ericpang777: Yep

xingao267: Can you add a couple more test cases? e.g. no init config and even failed cases. If these are also the examples displayed in the resource page like https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/secure_source_manager_instance, then failed cases might not be a good fit.

ericpang777: Sure will do. Yea, it seems like these are all shown as examples, so you can't add failed cases.

ericpang777: Added no initial config case

xingao267 (commented):
It's still in Draft. Should it be moved to some ready to review state?

ericpang777:
Changed to review state

mmv1/products/securesourcemanager/Repository.yaml
```diff
@@ -0,0 +1,170 @@
+# # Copyright 2024 Google Inc.
```
xingao267, line 1: Is this file auto-generated based on our API proto?

ericpang777: No I wrote it manually

ericpang777: But since I wrote the Instance.yaml, for this one I mostly just copied that one over

mmv1/products/securesourcemanager/Repository.yaml
```diff
@@ -0,0 +1,170 @@
...
+    'Official Documentation': 'https://cloud.google.com/secure-source-manager/docs/overview'
+import_format: ['projects/{{project}}/locations/{{location}}/repositories/{{repository_id}}', '{{repository_id}}']
+autogen_async: true
+async: !ruby/object:Api::OpAsync
+  actions: ['create', 'delete']
+  operation: !ruby/object:Api::OpAsync::Operation
+    path: 'name'
+    base_url: '{{op_id}}'
+    wait_ms: 1000
+    timeouts: !ruby/object:Api::Timeouts
```
umairidris, line 32: Can we just leave the default timeouts?

mmv1/products/securesourcemanager/Repository.yaml
```diff
@@ -0,0 +1,170 @@
...
+    name: 'location'
+    description: |
+      The location for the Repository.
+    required: true
+    immutable: true
+    url_param_only: true
+  - !ruby/object:Api::Type::String
+    name: 'repository_id'
+    description: |
+      The name for the Repository.
```
umairidris, line 82: ID of the Repository

mmv1/products/securesourcemanager/Repository.yaml
```diff
@@ -0,0 +1,170 @@
...
+        description:
+          git_https is the git HTTPS URI for git operations.
+        output: true
+      - !ruby/object:Api::Type::String
+        name: 'api'
+        description: |
+          API is the URI for API access.
+        output: true
+  - !ruby/object:Api::Type::NestedObject
+    name: 'initialConfig'
```
umairidris, line 146: Does this need to be immutable too or does ignore_read prevent updates to this field?

ericpang777: We don't return initialConfig in our Get response, so if you try a Terraform plan afterwards, it attempts to replace the repository, because it thinks you are adding an initial config.

mmv1/products/securesourcemanager/Repository.yaml
ericpang777: Done

mmv1/products/securesourcemanager/Repository.yaml
ericpang777: Done

mmv1/products/securesourcemanager/Repository.yaml
ericpang777: If you change the field, Terraform asks if you want to replace it

umairidris: Ah, got it. SG then

mmv1/products/securesourcemanager/Repository.yaml
umairidris: I still see the previous fields, maybe forgot to push the commit?

umairidris (approved):

mmv1/products/securesourcemanager/Repository.yaml
ericpang777: Pushed now oops.

mmv1/products/securesourcemanager/Repository.yaml
```diff
@@ -0,0 +1,166 @@
...
+# #
+# # Unless required by applicable law or agreed to in writing, software
+# # distributed under the License is distributed on an "AS IS" BASIS,
+# # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# # See the License for the specific language governing permissions and
+# # limitations under the License.
+
+--- !ruby/object:Api::Resource
+name: 'Repository'
+kind: 'securesourcemanager#repository'
```
roaks3, line 16: Nit: I don't think you need this field (I believe it is only used for `compute`)

mmv1/products/securesourcemanager/Repository.yaml
```diff
@@ -0,0 +1,166 @@
...
+    vars:
+      repository_id: 'my-repository'
+      instance_id: 'my-instance'
+parameters:
+  - !ruby/object:Api::Type::String
+    name: 'location'
+    description: |
+      The location for the Repository.
+    required: true
+    immutable: true
```
roaks3, line 73: I just verified from the [diff](https://github.com/modular-magician/terraform-provider-google/compare/auto-pr-10840-old..auto-pr-10840), but since this whole resource is immutable, all fields are made immutable, and this setting does nothing. You could leave it if you like, but if you do, I would recommend staying consistent across all of the fields (eg. `description`, etc).

mmv1/products/securesourcemanager/Repository.yaml
```diff
@@ -0,0 +1,166 @@
...
+    url_param_only: true
+properties:
+  - !ruby/object:Api::Type::String
+    name: 'name'
+    description: |
+      The resource name for the Repository.
+    output: true
+    immutable: true
```
roaks3, line 88: This line can be removed: output-only fields cannot be set by the user

mmv1/templates/terraform/examples/secure_source_manager_repository_initial_config.tf.erb
```diff
@@ -0,0 +1,18 @@
+resource "google_secure_source_manager_instance" "instance" {
+    location = "us-central1"
+    instance_id = "<%= ctx[:vars]['instance_id'] %>"
+}
+
+resource "google_secure_source_manager_repository" "<%= ctx[:primary_resource_id] %>" {
```
roaks3, line 6: Per the missing test message, it might be worth making this a "full" example, and include every possible field (like `etag`). Not strictly necessary if it doesn't make sense, but we do strongly prefer to have all configurable fields be used in at least one test.

roaks3 (changes requested):
Looks good, my main piece of feedback is regarding some of the field configs

mmv1/products/securesourcemanager/Repository.yaml
ericpang777: Done.

mmv1/products/securesourcemanager/Repository.yaml
ericpang777: Removed all field immutables. I removed it from the repository_id as well, I noticed you didn't comment there but I assume that should be safe.

mmv1/products/securesourcemanager/Repository.yaml
ericpang777: Removed the immutable

mmv1/templates/terraform/examples/secure_source_manager_repository_initial_config.tf.erb
ericpang777: Our API response currently doesn't have an etag field. So I think we can leave it for now unused. For the other missing test, we also don't support IAM conditions either.

roaks3: Yea, the iam conditions are fine to ignore.

For the `etag` though, it is a bit strange to be adding it here if it can't be used. I would definitely consider removing it from the PR, unless there are plans to add it in the near future. It can always be added later (in practice, it might not be a very useful field anyway).

ericpang777: Good points, I have removed that field. 👍

roaks3 (approved):
LGTM. I can't quite tell the state of one of the other reviews, so if you can confirm this looks ready to merge from your end, I can go ahead and do that.

umairidris:
LGTM!

ericpang777:
It is good, thank you!

roaks3 (approved):

PR #10853, opened 2024-05-31
Moved test reader to a separate directory

melinath:
This makes it easier to validate the tool's behavior separately from the diff processor

```release-note:none

```


melinath:
Removing review request since this is not passing tests. I'll request a new reviewer once I'm able to fix the tests.

melinath:
testing comment behavior

shuyama1 (commented):
Only one small comment

tools/test-reader/test-reader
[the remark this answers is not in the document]
melinath: done

melinath:
good catch, thanks!

tools/test-reader/cmd/read_tests.go
```diff
@@ -0,0 +1,63 @@
...
+)
+
+const readTestsDesc = "Run the missing test detector using the given services directory"
+
+type readTestsOptions struct {
+	rootOptions *rootOptions
+	testPrefix  string
+}
+
+func newReadTestsCmd(rootOptions *rootOptions) *cobra.Command {
```
shuyama1, line 18: Just to make sure I understand the purpose of this PR correctly. 
1. we moved the test reader away from the diff processor 
2. We implement the feature to run the missing test detector for the given services directory. This is not actually being called in our CI at this moment, is it?

shuyama1 (approved):
Sorry, forgot to hit the approve button the other day

tools/test-reader/cmd/read_tests.go
melinath: 1. yes
2. this is only implemented as a new command for the test reader. This behavior is used by the missing test detector, but that's still part of the diff-processor. The missing test detector uses the output from `reader.ReadAllTests`, then filters it by what fields / resources were modified according to the diff.

I wanted to split this out because it makes it a lot easier to run the test reader on its own and debug why it's having trouble with a specific test file.

shuyama1: Makes sense. Thanks for clarifying!

shuyama1 (approved):
💯

PR #10511, opened 2024-04-22
 feat(google_container_node_pool): support secondary boot disks

aston-github:
Fixes hashicorp/terraform-provider-google/issues/17655

This PR added support for secondary boot disks to google_container_node_pool resource. This allows users to improve workload startup latency by using secondary boot disks to preload data or container images on new nodes. See [Use secondary boot disks to preload data or container images](https://cloud.google.com/kubernetes-engine/docs/how-to/data-container-image-preloading) for further documentation. These code changes build off a [draft commit](https://github.com/elfinhe/magic-modules/commit/34660ed459f1887ebd0f633232e0b9db9d587a31) by @elfinhe, adding required configs, schema changes, and test validation.

If this PR is for Terraform, I acknowledge that I have:

- [x] Searched through the [issue tracker](https://github.com/hashicorp/terraform-provider-google/issues) for an open issue that this either resolves or contributes to, commented on it to claim it, and written "fixes {url}" or "part of {url}" in this PR description. If there were no relevant open issues, I opened one and commented that I would like to work on it (not necessary for very small changes).

- [x]  Ensured that all new fields I added that can be set by a user appear in at least one [example](https://github.com/GoogleCloudPlatform/magic-modules/tree/main/mmv1/templates/terraform/examples) (for generated resources) or [third_party test](https://github.com/GoogleCloudPlatform/magic-modules/tree/main/mmv1/third_party/terraform/tests) (for handwritten resources or update tests). **Added TestAccContainerNodePool_secondaryBootDisks to //third_party/terraform/services/container/resource_container_node_pool_test**

- [x]  [Generated Terraform providers](https://github.com/GoogleCloudPlatform/magic-modules/blob/main/docs/content/docs/getting-started/generate-providers.md), and ran [make test and make lint](https://googlecloudplatform.github.io/magic-modules/docs/getting-started/run-provider-tests/#run-unit-tests) in the generated providers to ensure it passes unit and linter tests.

[...]

PR #10854, opened 2024-05-31
feat(google_container_node_pool): Handle nil mode in secondary boot disk.

aston-github:
Fixes https://github.com/hashicorp/terraform-provider-google/issues/17655

This PR is a follow up to https://github.com/GoogleCloudPlatform/magic-modules/pull/10511, which added Terraform support for secondary boot disks in GKE. I discovered a bug when trying to create a node pool with "mode" not specified. Terraform throws a "panic: interface conversion: interface {} is nil, not map[string]interface {}" during the expandNodeConfig function. I pinpointed the error to the function's type assertion of the mode argument. This PR adds logic to bypass the type assertion when mode is nil.

```release-note:bug
container: modify `node_config.secondary_boot_disks` logic in `google_container_node_pool` to handle nil mode
```

### DIFF
--- a/mmv1/third_party/terraform/services/container/node_config.go.erb
+++ b/mmv1/third_party/terraform/services/container/node_config.go.erb
@@ -896,17 +896,23 @@ func expandNodeConfig(v interface{}) *container.NodeConfig {
 	}
 
 	if v, ok := nodeConfig["secondary_boot_disks"]; ok && len(v.([]interface{})) > 0 {
-		conf := v.([]interface{})[0].(map[string]interface{})
-		modeValue, modeOK := conf["mode"]
-		diskImage := conf["disk_image"]
-    	if modeOK && modeValue.(string) == "CONTAINER_IMAGE_CACHE" {
-			nc.SecondaryBootDisks = append(nc.SecondaryBootDisks, &container.SecondaryBootDisk{
-				DiskImage: diskImage.(string),
-				Mode:      modeValue.(string),
-			})
+		conf, confOK := v.([]interface{})[0].(map[string]interface{})
+		if confOK {
+			modeValue, modeOK := conf["mode"]
+			diskImage := conf["disk_image"].(string)
+			if modeOK && modeValue.(string) == "CONTAINER_IMAGE_CACHE" {
+				nc.SecondaryBootDisks = append(nc.SecondaryBootDisks, &container.SecondaryBootDisk{
+					DiskImage: diskImage,
+					Mode:      modeValue.(string),
+				})
+			} else {
+				nc.SecondaryBootDisks = append(nc.SecondaryBootDisks, &container.SecondaryBootDisk{
+					DiskImage: diskImage,
+				})
+			}
 		} else {
 			nc.SecondaryBootDisks = append(nc.SecondaryBootDisks, &container.SecondaryBootDisk{
-				DiskImage: diskImage.(string),
+				DiskImage: "",
 			})
 		}
 	}
@@ -1535,7 +1541,9 @@ func flattenSecondaryBootDisks(c []*container.SecondaryBootDisk) []map[string]in
 		for _, disk := range c {
 			secondaryBootDisk := map[string]interface{}{
 				"disk_image": disk.DiskImage,
-				"mode": disk.Mode,
+			}
+			if disk.Mode == "CONTAINER_IMAGE_CACHE" {
+				secondaryBootDisk["mode"] = disk.Mode
 			}
 			result = append(result, secondaryBootDisk)
 		}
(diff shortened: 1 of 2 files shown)


melinath (changes requested):

mmv1/third_party/terraform/services/container/node_config.go.erb
```diff
@@ -897,10 +897,18 @@ func expandNodeConfig(v interface{}) *container.NodeConfig {
 
 	if v, ok := nodeConfig["secondary_boot_disks"]; ok && len(v.([]interface{})) > 0 {
 		conf := v.([]interface{})[0].(map[string]interface{})
-		nc.SecondaryBootDisks = append(nc.SecondaryBootDisks, &container.SecondaryBootDisk{
-			DiskImage: conf["disk_image"].(string),
-			Mode:      conf["mode"].(string),
-		})
+		modeValue, modeOK := conf["mode"]
+		diskImage := conf["disk_image"]
+    	if modeOK && modeValue.(string) == "CONTAINER_IMAGE_CACHE" {
```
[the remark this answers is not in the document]
aston-github: Aligned with modeValue and diskImage lines above.

mmv1/third_party/terraform/services/container/node_config.go.erb
```diff
@@ -897,10 +897,18 @@ func expandNodeConfig(v interface{}) *container.NodeConfig {
 
 	if v, ok := nodeConfig["secondary_boot_disks"]; ok && len(v.([]interface{})) > 0 {
 		conf := v.([]interface{})[0].(map[string]interface{})
```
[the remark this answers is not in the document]
aston-github: Thanks for the insight! I figured the problem was with creating conf at line 884, but I thought it was the missing mode field causing the panic, hence adding the modeOK logic. I didn't know Terraform would treat the empty disk image as completely empty, but that would make sense with what we're seeing.

mmv1/third_party/terraform/services/container/node_config.go.erb
```diff
@@ -1527,7 +1541,9 @@ func flattenSecondaryBootDisks(c []*container.SecondaryBootDisk) []map[string]in
 		for _, disk := range c {
 			secondaryBootDisk := map[string]interface{}{
 				"disk_image": disk.DiskImage,
-				"mode": disk.Mode,
+			}
+			if disk.Mode == "CONTAINER_IMAGE_CACHE" {
```
melinath, line 1545: It looks like this is an enum field with only one potential value; however, if new values are added to that field in the future (which happens from time to time) this will not work for those cases. What would you think about checking for empty or `MODE_UNSPECIFIED` here?

mmv1/third_party/terraform/services/container/node_config.go.erb
```diff
@@ -896,11 +896,25 @@ func expandNodeConfig(v interface{}) *container.NodeConfig {
...
-		conf := v.([]interface{})[0].(map[string]interface{})
-		nc.SecondaryBootDisks = append(nc.SecondaryBootDisks, &container.SecondaryBootDisk{
-			DiskImage: conf["disk_image"].(string),
-			Mode:      conf["mode"].(string),
-		})
+		conf, confOK := v.([]interface{})[0].(map[string]interface{})
+		if confOK {
+			modeValue, modeOK := conf["mode"]
+			diskImage := conf["disk_image"].(string)
+			if modeOK && modeValue.(string) == "CONTAINER_IMAGE_CACHE" {
```
melinath, line 903: What would you think about checking that it's not empty or equal to `MODE_UNSPECIFIED` here? That would make the code forwards-compatible with other potential future enum values.

melinath (changes requested):

mmv1/third_party/terraform/services/container/node_config.go.erb
aston-github: Right now, if mode is not equal to CONTAINER_IMAGE_CACHE, gcloud automatically changes it into MODE_UNSPECIFIED. @elfinhe Do we want Terraform to reflect this behavior, or do we want it to reject modes that are not CONTAINER_IMAGE_CACHE?

aston-github: Discussing with @elfinhe, we want terraform to reject invalid modes instead of changing them to MODE_UNSPECIFIED. I've removed the mode changing logic, so the googleapi invalid mode error surfaces. This eliminates comparisons to CONTAINER_IMAGE_CACHE and associated logic. An example error is shown below with `mode="DATA" ` .

Screenshot 2024-06-04 at 1 24 23 PM

mmv1/third_party/terraform/services/container/node_config.go.erb
aston-github: After clarifying with @elfinhe about intended Terraform behavior, I removed mode changing logic. This check against CONTAINER_IMAGE_CACHE is no longer used. The new logic just checks for modeOK, meaning a mode exists, so type assertion to string does not panic.

melinath (approved):

PR #10856, opened 2024-06-03
Allow guest OS validation to be handled by GCE APIs

shruthi019:
Remove validation for guest OS feature types and allow validation to be handled by the GCE APIs.

```release-note:enhancement
compute: removed enum validation on `guest_os_features.type` in `google_compute_disk` to allow for new features to be used without provider update
```


slevenick (changes requested):
You'll also need to change the type of this field from:
        - !ruby/object:Api::Type::Enum
to

shruthi019:
Done.

slevenick (approved):

PR #10857, opened 2024-06-03
add project_number field to featureview

kkram01:
Add a new field "project_number" to featureview resource and update documentation changes for "feature_group" resource 

```release-note:enhancement
vertexai: added `project_number` field to `google_vertex_ai_feature_online_store_featureview` resource
```


rileykarson (commented):
Marking a review so the robot doesn't complain at me; we discussed the test failure offline, it's a preexisting issue. I suggested default_from_api: true to resolve and @kkram01 raised it didn't work for them in local testing.

mmv1/products/vertexai/FeatureOnlineStoreFeatureview.yaml
```diff
@@ -160,6 +160,10 @@ properties:
               description: |
                 Identifiers of features under the feature group.
               item_type: Api::Type::String
+      - !ruby/object:Api::Type::String
+        name: 'projectNumber'
```
rileykarson, line 164: Can we use this in a test? https://googlecloudplatform.github.io/magic-modules/develop/test/test/

I checked and current tests don't see it get returned- which I think we expect, since we're only using sources within the same project.

kkram01: yes, I'm not sure on how to add a test for this field as It requires two different GCP project

kkram01: Hi, @rileykarson how do we proceed forward here? Is there a way the magic-modules resource tests can access cross-project resources?

rileykarson: Apologies, missed that comment. You can create a new project within a test using the google_project resource: https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/google_project

kkram01:
hi @rileykarson I've added the test. however the creation of the resources is failing because of insufficient permissions error on the new project. Any inputs on how these permissions can be added to the underlying service_account that runs these tests?

rileykarson:
Oh interesting- are the appropriate permissions not granted by roles/owner on the new project?

kkram01:
@rileykarson 
image

The permissions are included in the Owner role to create the resources. The Owner role is automatically assigned to the Project creator, so it should work ideally. Not sure why it is failing. Any inputs?

mmv1/templates/terraform/examples/vertex_ai_featureonlinestore_featureview_cross_project.tf.erb
```diff
@@ -0,0 +1,132 @@
...
+  project = google_project.project.project_id
+  timeouts {
+    create = "30m"
+    update = "40m"
+  }
+  disable_on_destroy = false
+}
+
+resource "google_vertex_ai_feature_online_store" "featureonlinestore" {
+  name     = "<%= ctx[:vars]['name'] %>"
```
rileykarson, line 29: Should this be part of `google_project.project`? If not, should we remove the dependency on the `google_project_service` resource?

mmv1/templates/terraform/examples/vertex_ai_featureonlinestore_featureview_cross_project.tf.erb
```diff
@@ -0,0 +1,132 @@
...
+resource "google_project_service" "vertexai" {
+  service = "aiplatform.googleapis.com"
+  project = google_project.project.project_id
+  timeouts {
+    create = "30m"
+    update = "40m"
+  }
+  disable_on_destroy = false
+}
+
```
rileykarson, line 27:
```suggestion

//TODO: remove this before merging
data "google_project_iam_policy" "policy" {
  project  = google_project.project.project_id
}

```

Can you apply this suggestion so getIamPolicy gets run against the new project in the test? We'll remove it later, I just want to make sure I'm not making a bad assumption about the IAM roles assigned to this project.

mmv1/templates/terraform/examples/vertex_ai_featureonlinestore_featureview_cross_project.tf.erb
```diff
@@ -0,0 +1,132 @@
+resource "google_project" "project" {
```
rileykarson, line 1: Do you mind explaining in writing which resource(s) should be in project A (the standard test project) and project B (the project we're creating inline in the test)? I'm not familiar with the CUJ we're covering, so I'm just poking at things (like with my comment below) otherwise.

mmv1/templates/terraform/examples/vertex_ai_featureonlinestore_featureview_cross_project.tf.erb
kkram01: you're right, thanks for pointing it out. I removed the dependency now on featureonlinestore.

mmv1/templates/terraform/examples/vertex_ai_featureonlinestore_featureview_cross_project.tf.erb
kkram01: Sure, added

mmv1/templates/terraform/examples/vertex_ai_featureonlinestore_featureview_cross_project.tf.erb
kkram01: In the project B (inline created project), we create the BQ dataset & table, FeatureGroup and Feature. In project A, we create the FeatureOnlineStore and FeatureView. In the FeatureView, we point to the FeatureGroup and Features created in ProjectB as the source along with the project B's project_number. This is the usecase. 

Previously cross project FeatureGroup referencing wasn't allowed. We're trying to enable that in terraform with this change.

mmv1/templates/terraform/examples/vertex_ai_featureonlinestore_featureview_cross_project.tf.erb
kkram01: image

The serviceAccount has the owner role on the new project

mmv1/templates/terraform/examples/vertex_ai_featureonlinestore_featureview_cross_project.tf.erb
kkram01: I've tried to replicate this by creating a new serviceAccount with only owner role and impersonated it to create FeatureGroups. It works fine.

mmv1/templates/terraform/examples/vertex_ai_featureonlinestore_featureview_cross_project.tf.erb
rileykarson: I suspect it's just an eventual consistency/timing issue then. Can you chain a `time_sleep` of 60s between the project creation and resource use? https://github.com/GoogleCloudPlatform/magic-modules/blob/c96e8a0c51a81713f1244e24a4369e21f53b5e61/mmv1/templates/terraform/examples/iap_project.tf.erb#L7 is an example that does that. We'll also need to register the `time` external provider in the YAML with `external_providers`, similar to https://github.com/GoogleCloudPlatform/magic-modules/blob/c96e8a0c51a81713f1244e24a4369e21f53b5e61/mmv1/products/iap/Web.yaml#L35

mmv1/templates/terraform/examples/vertex_ai_featureonlinestore_featureview_cross_project.tf.erb
rileykarson: Thanks!

kkram01:
@rileykarson pls review. the tests have passed

rileykarson (approved):
LGTM, thanks!

PR #10859, opened 2024-06-03
Unset operation when reading instance_group_manager resource

askubis:
fixes https://github.com/hashicorp/terraform-provider-google/issues/16043
It is reverting a change done by refactoring in https://github.com/hashicorp/terraform-provider-google-beta/commit/79e0a2e37e138ec8ade75ac726e3de4dc9adfc88#diff-1cf6fdb65a48539ac4730b8920d0366d3a173b7e824b08502d3c3da73456d8f8

```release-note:bug
compute: fixed an issue where `google_compute_instance_group_manager` with a pending operation was incorrectly removed due to the operation no longer being present in the backend
```


c2thorn (changes requested):
I cannot find the context for this change. Is there a better link? The current one links to a seemingly random commit that is very large.

Also, is this related to a Github issue in https://github.com/hashicorp/terraform-provider-google/issues?q=is%3Aissue+is%3Aopen+sort%3Aupdated-desc ?

askubis:
> I cannot find the context for this change. Is there a better link? The current one links to a seemingly random commit that is very large.
> 
> Also, is this related to a Github issue in https://github.com/hashicorp/terraform-provider-google/issues?q=is%3Aissue+is%3Aopen+sort%3Aupdated-desc ?

the context is https://github.com/hashicorp/terraform-provider-google/issues/16043

what the commit is mentioning is  google-beta/resource_compute_instance_group_manager.go
```
-               d.Set("operation", "")
+		if err := d.Set("operation", op.Name); err != nil {
+			return fmt.Errorf("Error setting operation: %s", err)
+		}
```

which seems to be a copy-paste error, as the previous change in the same file is correct:
```
-			d.Set("operation", op.Name)
+			if err := d.Set("operation", op.Name); err != nil {
+				return fmt.Errorf("Error setting operation: %s", err)
+			}
```

c2thorn:
> hashicorp/terraform-provider-google#16043

I see, thank you.

c2thorn (approved):

PR #10860, opened 2024-06-03
Go Rewrite - resolve all current diffs

c2thorn:
all current diffs are trivial: https://github.com/c2thorn/terraform-provider-google-beta/commit/2615f922323cf2b6b9e06c9f8f5516e0b720098c

go_*.yaml files copied over using an experimental script with the converter script: https://github.com/c2thorn/magic-modules/pull/1

```release-note:none

```


mmv1/api/resource.go
```diff
@@ -1279,15 +1280,30 @@ func CompareByName(a, b *Type) int {
...
+	keys := []string{}
+	for _, prop := range properties {
+		if prop.FlattenObject {
+			k := r.GetPropertyUpdateMasksGroupKeys(prop.Properties)
+			keys = append(keys, k...)
+		} else if len(prop.UpdateMaskFields) > 0 {
+			keys = append(keys, google.Underscore(prop.Name))
+		} else {
+			keys = append(keys, google.Underscore(prop.Name))
+		}
```
zli82016, line 1293: These two cases can be combined to one?

zli82016 (changes requested):

mmv1/api/resource.go
c2thorn: yeah, they started off different but ended up being the same later

zli82016 (approved):
LGTM

PR #10865, opened 2024-06-03
Adding mmv1 generated Resource: google_composer_user_workloads_config_map

spapi17:

```release-note:new-resource
`google_composer_user_workloads_config_map` (beta)
```

### DIFF
--- a/mmv1/products/composer/UserWorkloadsConfigMap.yaml
+++ b/mmv1/products/composer/UserWorkloadsConfigMap.yaml
@@ -1,0 +1,147 @@
+# Copyright 2024 Google Inc.
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+--- !ruby/object:Api::Resource
+# API resource name
+name: 'UserWorkloadsConfigMap'
+# Resource description for the provider documentation.
+description: |
+  User workloads Secret used by Airflow tasks that run with Kubernetes Executor or KubernetesPodOperator. 
+  Intended for Composer 3 Environments.
+references: !ruby/object:Api::Resource::ReferenceLinks
+  guides:
+   # Link to quickstart in the API's Guides section. For example:
+   # 'Create and connect to a database': 'https://cloud.google.com/alloydb/docs/quickstart/create-and-connect'
+    'QUICKSTART_TITLE': 'QUICKSTART_URL'
+  # Link to the REST API reference for the resource. For example,
+  # https://cloud.google.com/alloydb/docs/reference/rest/v1/projects.locations.backups
+  api: 'API_REFERENCE_URL'
+# Marks the resource as beta-only. Ensure a beta version block is present in
+# provider.yaml.
+min_version: beta
+
+# Inserts styled markdown into the header of the resource's page in the
+# provider documentation.
+# docs: !ruby/object:Provider::Terraform::Docs
+#   warning: |
+#     MULTILINE_WARNING_MARKDOWN
+#   note: |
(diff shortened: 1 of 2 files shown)


spapi17:
This failing tests are unrelated, most likely due to the upgrade of Terraform version.

mmv1/products/composer/UserWorkloadsConfigMap.yaml
```diff
@@ -0,0 +1,76 @@
...
+
+examples:
+  - !ruby/object:Provider::Terraform::Examples
+    name: "composer_user_workloads_config_map_basic"
+    primary_resource_id: "config_map"
+    vars:
+      environment_name: "test-environment"
+      config_map_name: "test-config-map"
+
+parameters:
```
SarahFrench, line 42: There are `parameters` and `properties` sections of the YAML for storing information about user-supplied values. In general a resource's [fields should be nested under `properties`](https://googlecloudplatform.github.io/magic-modules/develop/resource/#add-fields) and only added to `parameters` when needed (i.e if url_param_only=true).

Please move all the fields that aren't url_param_only=true under a top-level `properties` field in the YAML.

SarahFrench: Also, could you also please add a new field `project` as a url_param_only=true parameter? Currently the code generator is identifying that a project value is needed in the URLs for managing the resource and, as a result, is adding a project field to the resource automatically. It'd be good to make the field explicitly defined in the YAML instead.

SarahFrench (changes requested):
Hi @spapi17, thanks for your PR!

I've left some feedback about some changes required to the YAML files, and a fix for a GA-specific build issue. Following these changes I would expect the new test you've added will continue to pass and I will run some tests on the GA provider too.

mmv1/products/composer/UserWorkloadsConfigMap.yaml
```diff
@@ -0,0 +1,76 @@
...
+    immutable: true
+    custom_flatten: 'templates/terraform/custom_flatten/name_from_self_link.erb'
+    custom_expand: 'templates/terraform/custom_expand/shortname_to_url.go.erb'
+    validation: !ruby/object:Provider::Terraform::Validation
+      function: 'verify.ValidateGCEName'
+    description: |
+      Name of the Kubernetes ConfigMap.
+  - !ruby/object:Api::Type::KeyValuePairs
+    name: 'data'
+    required: false
```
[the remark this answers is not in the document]
SarahFrench: When these fields aren't set they default to false, so our convention is to not include `required` in YAML unless we're setting it to true

mmv1/third_party/terraform/provider/provider_mmv1_resources.go.erb
```diff
@@ -8,7 +8,6 @@ import (
 	  "github.com/hashicorp/terraform-provider-google/google/services/<%= service -%>"
 	<% end -%>
 
-	"github.com/hashicorp/terraform-provider-google/google/services/composer"
```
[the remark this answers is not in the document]
spapi17: done

mmv1/products/composer/UserWorkloadsConfigMap.yaml
spapi17: When I defined project field, it was still generated and I got double definition in the final code. I'm getting this error as a result (on build) : 
```
google-beta/services/composer/resource_composer_user_workloads_config_map.go:91:4: duplicate key "project" in map literal
```

SarahFrench (approved):
Approved - the refactor I requested in the last review didn't cause the new acceptance  test to be re-run as it didn't impact the test, and [the new test passed](https://hashicorp.teamcity.com/buildConfiguration/TerraformProviders_GoogleCloud_GOOGLE_BETA_MMUPSTREAMTESTS_GOOGLEBETA_PACKAGE_COMPOSER/163299?buildTab=log&branch=refs%2Fheads%2Fauto-pr-10865&focusLine=1346&logView=linear&linesState=1290) when I ran it in our test environment. Also, I [ran a random Composer test against the GA provider code](https://hashicorp.teamcity.com/buildConfiguration/TerraformProviders_GoogleCloud_GOOGLE_MMUPSTREAMTESTS_GOOGLE_PACKAGE_COMPOSER/163298?buildTab=overview&hideTestsFromDependencies=false&hideProblemsFromDependencies=false&expandBuildChangesSection=true&expandBuildDeploymentsSection=false) resulting from this PR and it worked ok

PR #10866, opened 2024-06-04
Fix test for the NEG AppEngine

Cheriit:
Fix a TestAccComputeRegionNetworkEndpointGroup_regionNetworkEndpointGroupAppengineExample test
https://github.com/hashicorp/terraform-provider-google/issues/16170

```release-note:none

```

### DIFF
--- a/mmv1/templates/terraform/examples/region_network_endpoint_group_appengine.tf.erb
+++ b/mmv1/templates/terraform/examples/region_network_endpoint_group_appengine.tf.erb
@@ -101,8 +101,6 @@ resource "google_app_engine_flexible_app_version" "<%= ctx[:vars]['app_version_n
       target_utilization = 0.5
     }
   }
-
-  noop_on_destroy = true
 }
 
 resource "google_storage_bucket" "bucket" {
--- a/mmv1/templates/terraform/examples/region_network_endpoint_group_appengine.tf.erb
+++ b/mmv1/templates/terraform/examples/region_network_endpoint_group_appengine.tf.erb
@@ -105,7 +105,7 @@ resource "google_app_engine_flexible_app_version" "<%= ctx[:vars]['app_version_n
     }
   }
 
-  delete_service_on_destroy = true
+  delete_service_on_destroy = false
 }
 
 resource "google_storage_bucket" "bucket" {
--- a/mmv1/templates/terraform/examples/region_network_endpoint_group_appengine.tf.erb
+++ b/mmv1/templates/terraform/examples/region_network_endpoint_group_appengine.tf.erb
@@ -105,7 +105,7 @@ resource "google_app_engine_flexible_app_version" "<%= ctx[:vars]['app_version_n
     }
   }
 
-  noop_on_destroy = true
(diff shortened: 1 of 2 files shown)


NickElliot (commented):
The test is failing it seems due to the inability to delete the `google_app_engine_flexible_app_version` because it is the last instance of a `default` service, and I'm not sure there is a way around that due to the new version of this test creating a new project. Would it be possible to still use the default CI project for this testing?

Cheriit:
@NickElliot unfortunately, according to the `google_app_engine_application` documentation "App Engine applications cannot be deleted once they're created; you have to delete the entire project to delete the application. Terraform will report the application has been successfully deleted; this is a limitation of Terraform, and will go away in the future. Terraform is not able to delete App Engine applications."

To test this it looks like we have to create a separate project to clear it :(

Test previously failed because of the API change with the buckets, but after resolving this issue, the one with app engine has appeared. 

I'll try to add `delete_service_on_destroy` to the `google_app_engine_flexible_app_version`. Mabye it will work

NickElliot:
I'm thinking the approach will need to be adding a bootstrap function, so a separate app version with the same service is made in the new project agnostic of the ongoing test. I'll link some examples soon

mmv1/templates/terraform/examples/region_network_endpoint_group_appengine.tf.erb
```diff
@@ -58,13 +108,15 @@ resource "google_app_engine_flexible_app_version" "<%= ctx[:primary_resource_id]
   delete_service_on_destroy = true
```
NickElliot, line 108:
```suggestion
  delete_service_on_destroy = false
```
try setting this!

NickElliot (changes requested):
I looked into various things around this but this might be a simple fix

NickElliot (changes requested):
Ok, the following will need to be done:

1. Move the test to be handwritten
2. Add two bootstrap functions to the test:
*  one using [BootstrapProject](https://github.com/GoogleCloudPlatform/magic-modules/blob/def52fed28327b8f3e14c1894aeb21cba08da598/mmv1/third_party/terraform/acctest/bootstrap_test_utils.go#L625) 
* the other using a new bootstrap function you will need to implement for `google_app_engine_flexible_app_version` that will create a seperate version of the resource in the bootstrapped project. 
3. the test will need to be updated to then use a project data source that refers to the bootstrapped project.

Here's an example of a handwritten test using a bootstrap function: https://github.com/GoogleCloudPlatform/magic-modules/blob/70941b698762c42f979bcf5ab24c70e85dc1992e/mmv1/third_party/terraform/services/sql/resource_sql_database_instance_test.go#L182

Let me know if you have any questions along the way!

rileykarson:
Hey! I'm closing this PR as a part of a cleanup of older inactive PRs, using a threshold of PRs in "Draft" last updated over a month ago. This doesn't represent rejection of the change, and feel free to comment for me to reopen it if you plan to pick it back up, or feel free to start a new PR with the same changes in the future.

PR #10873, opened 2024-06-04
add log bucket resource

delimaneto:
add logging.googleapis.com/LogBucket resource

```release-note:none
```

### DIFF
--- a/mmv1/provider/terraform_tgc.rb
+++ b/mmv1/provider/terraform_tgc.rb
@@ -317,7 +317,8 @@ module Provider
                        ['converters/google/resources/dataproc_cluster.go',
                         'third_party/tgc/dataproc_cluster.go'],
                        ['converters/google/resources/resource_logging_project_bucket_config.go',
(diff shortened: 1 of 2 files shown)


mmv1/provider/terraform_tgc.rb
```diff
@@ -317,7 +317,10 @@ def copy_common_files(output_folder, generate_code, _generate_docs)
                        ['converters/google/resources/dataproc_cluster.go',
                         'third_party/tgc/dataproc_cluster.go'],
                        ['converters/google/resources/commitment.go',
-                        'third_party/tgc/commitment.go']
+                        'third_party/tgc/commitment.go'],
+                       ['converters/google/resources/resource_logging_project_bucket_config.go',
+                        'third_party/terraform/services/logging/' \
```
rileykarson, line 322: Terraform files can't be directly used here. as they're not valid TGC files. You'd need to create a new handwritten file based on other TGC resources to add support for this one.

The specific error you're hitting here is that the package is wrong- it should be `google`- but that's likely just the first error, and fixing it directly would just lead to another one.

PR #10877, opened 2024-06-04
Added update support to sslPolicy field in region_target_https_proxy

matheusaleixo-cit:
Adds support to update "ssl_policy" field in "compute_region_target_https_proxy" via the [PATCH](https://cloud.google.com/compute/docs/reference/rest/v1/regionTargetHttpsProxies/patch) endpoint, removing its ForceNew flag, as well as related tests;

Fixes: https://github.com/hashicorp/terraform-provider-google/issues/16963

```release-note:enhancement
compute: added update support to 'ssl_policy' field in 'compute_region_target_https_proxy' resource
```

### DIFF
--- a/mmv1/third_party/terraform/services/compute/resource_compute_region_target_https_proxy_test.go.erb
+++ b/mmv1/third_party/terraform/services/compute/resource_compute_region_target_https_proxy_test.go.erb
@@ -5,7 +5,7 @@ import (
 	"fmt"
 	"testing"
 	"github.com/hashicorp/terraform-provider-google/google/acctest"
-  "github.com/hashicorp/terraform-provider-google/google/envvar"
+	"github.com/hashicorp/terraform-provider-google/google/envvar"
 
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
 )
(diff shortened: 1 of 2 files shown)


mmv1/third_party/terraform/services/compute/resource_compute_region_target_https_proxy_test.go.erb
```diff
@@ -349,3 +350,290 @@ resource "google_compute_region_ssl_certificate" "foobar2" {
...
+}
+
+resource "google_compute_region_target_https_proxy" "default-https" {
+  project          = "%{project_id}"
+  region           = "us-central1"
+  name             = "https-proxy-%{resource_suffix}"
+  url_map          = google_compute_region_url_map.default-https.self_link
+  ssl_certificates = [google_compute_region_ssl_certificate.foobar0.self_link]
+
+  lifecycle {
```
slevenick, line 460: Why do we need create_before_destroy here?

slevenick (commented):
Just a small question in the test config. I believe the failing test is unrelated and is failing for all PRs, so it should be safe to ignore

mmv1/third_party/terraform/services/compute/resource_compute_region_target_https_proxy_test.go.erb
matheusaleixo-cit: After re-testing, those fields don't appear to be as necessary as I expected, I will be removing them.

slevenick (approved):

PR #10891, opened 2024-06-06
Make magician subcommands use RunE and return errors instead of calling os.Exit directly

iyabchen:
https://github.com/hashicorp/terraform-provider-google/issues/17479

```release-note:none

```

### DIFF
--- a/.ci/magician/cmd/test_tgc_integration.go
+++ b/.ci/magician/cmd/test_tgc_integration.go
@@ -20,22 +20,20 @@ var testTGCIntegrationCmd = &cobra.Command{
 	1. GOPATH
 	2. GITHUB_TOKEN_MAGIC_MODULES
 	`,
-	Run: func(cmd *cobra.Command, args []string) {
+	RunE: func(cmd *cobra.Command, args []string) error {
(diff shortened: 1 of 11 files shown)


.ci/magician/cmd/test_tgc_integration.go
```diff
@@ -20,52 +20,47 @@ var testTGCIntegrationCmd = &cobra.Command{
...
-			fmt.Println("Did not provide GITHUB_TOKEN_MAGIC_MODULES or GITHUB_TOKEN environment variables")
-			os.Exit(1)
+			return fmt.Errorf("did not provide GITHUB_TOKEN_MAGIC_MODULES or GITHUB_TOKEN environment variables")
 		}
 
 		rnr, err := exec.NewRunner()
 		if err != nil {
-			fmt.Println("Error creating runner: ", err)
+			return fmt.Errorf("error creating runner: %w", err)
 			os.Exit(1)
```
melinath, line 37:
```suggestion
```

melinath (approved):
LGTM - it looks like there are a few trailing os.Exit calls in non-root commands but they can probably be more easily addressed in a follow-up PR.

PR #10892, opened 2024-06-06
[AlloyDB] Support project IDs and project numbers for PSC instances 

vmiglani:
This PR addresses the issue raised: https://github.com/hashicorp/terraform-provider-google/issues/18327. 

With this change, PSC instances can specify the field `allowed_consumer_projects` as either project IDs or project numbers. 

Details about why this change was implemented this way can be found here - https://yaqs.corp.google.com/eng/q/1374949186138537984. Briefly, we want to support the customer specifying either project numbers or IDs, however, the API always returns project numbers. 

```release-note:enhancement 
alloydb: allowed `google_alloydb_instance.psc_instance_config.allowed_consumer_projects` to be specified as either project numbers or project IDs.  
```

### DIFF
--- a/mmv1/third_party/terraform/tpgresource/common_diff_suppress.go.erb
+++ b/mmv1/third_party/terraform/tpgresource/common_diff_suppress.go.erb
@@ -282,6 +282,23 @@ func LastSlashDiffSuppress(_, old, new string, _ *schema.ResourceData) bool {
 }
 
 // Suppress diffs when the value read from api
+// has the project number instead of the project name.
+// This function suppresses the difference when the project ID is not prefixed with the string "projects".
+func ProjectNumberDiffSuppressWithoutPrefix(k, old, new string, _ *schema.ResourceData) bool {
+	// Ignore list size changes
+	if strings.HasSuffix(k, ".#") {
+		return false
+	}
+	var a2, b2 string
+	reN := regexp.MustCompile("\\d+")
+	re := regexp.MustCompile("[^/]+")
+	replacement := []byte("equal")
+	a2 = string(reN.ReplaceAll([]byte(old), replacement))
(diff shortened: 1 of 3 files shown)


vmiglani:
@melinath Failing tests are not related to my changes

mmv1/third_party/terraform/tpgresource/common_diff_suppress.go.erb
```diff
@@ -281,6 +281,23 @@ func LastSlashDiffSuppress(_, old, new string, _ *schema.ResourceData) bool {
 	return new == old
 }
 
+// Suppress diffs when the value read from api
+// has the project number instead of the project name.
+// This function suppresses the difference when the project ID is not prefixed with the string "projects".
+func ProjectNumberDiffSuppressWithoutPrefix(k, old, new string, _ *schema.ResourceData) bool {
```
melinath, line 287: Could you implement this in services/alloydb (using a [`custom_code.constants`](https://googlecloudplatform.github.io/magic-modules/develop/custom-code/#add-reusable-variables-and-functions) file) instead of in common_diff_suppress? Although it is similar to ProjectNumberDiffSuppress it may not be as widely applicable to different services, so I'd rather not add it globally.

It would also be great to get some [unit tests](https://googlecloudplatform.github.io/magic-modules/develop/test/test/#add-unit-tests) for this function.

melinath (changes requested):
One small change request & it would be great to add some unit tests! Please let me know if you have any questions about how to do these.

melinath:
Reassigning due to upcoming OOO

NickElliot (commented):
Ping me when the previously requested changes are made! thanks

simonebruzzechesse:
Is it possible to re-open this pull request since lot of customers are complaining about using project numbers rather than IDs?

PR #10894, opened 2024-06-06
Migrate BigQuery Reservation Assignment from DCL to MMv1

obada-ab:
Fixes https://github.com/hashicorp/terraform-provider-google/issues/18338

```release-note:none
bigqueryreservation: `google_bigquery_reservation_assignment ` now uses mmv1 engine instead of DCL
```


shuyama1:
Adding `override-breaking-change` label, since the breaking change detected has been seen in DCL->MMv1 migrations and can be safely ignored. Context: b/329621750

shuyama1 (commented):
I've gone through the generated files in the downstreams and it looks good to me overall. 
@slevenick or @trodge would either of you take a second look incase I miss anything for this type of migration. Thank you.

slevenick:
Yeah, I'll take a look at this today.

slevenick:
The only potential issue that I see is that the name field is changing in state from being stored as the short form (123456789) to the long form (projects/my-project/locations/us-central1/reservations/example-reservation/assignments/12346789)

I think we should fix this by adding this flattener: `custom_flatten: 'templates/terraform/custom_flatten/name_from_self_link.erb'`

Then we can remove the logic to get the short name from the decoder as it's already done

obada-ab:
@slevenick

> I think we should fix this by adding this flattener: custom_flatten: 'templates/terraform/custom_flatten/name_from_self_link.erb'

Done

> Then we can remove the logic to get the short name from the decoder as it's already done

I removed the decoder entirely, and added `nested_query` since it's gonna do the needed job (flatten `name` and compare)

slevenick (approved):
LGTM, I'll let @shuyama1 make a final pass

shuyama1 (approved):
Looks good! Thank you!

PR #10897, opened 2024-06-06
Breaking change - Add integration for subnetworks with internal ranges API

daanheikens:
Fixes: https://github.com/hashicorp/terraform-provider-google/issues/17881

Second attempt, superseding: https://github.com/GoogleCloudPlatform/magic-modules/pull/10507

```release-note:enhancement
compute: added fields `reserved_internal_range` and `secondary_ip_ranges[].reserved_internal_range` to `google_compute_subnetwork` resource
```


daanheikens:
@trodge FYI! As discussed on: https://github.com/hashicorp/terraform-provider-google/issues/18300

@c2thorn +1 for visibility

roaks3:
@trodge passing this one to you, since it looks like you did the initial review. I'm not adding `disable-review-reminders` since you may want to take a look first, but otherwise it seems like we could add it here to keep the notifications quiet.

c2thorn:
I've updated the base branch to point to the major release branch. This should be good to continue.

I've forgotten what the exact breaking change is here - but please follow the [steps here](https://github.com/GoogleCloudPlatform/magic-modules/pull/11237#issuecomment-2246331651) to edit the upgrade guide so User's know how to handle the breaking change.

@daanheikens @trodge

daanheikens:
> I've updated the base branch to point to the major release branch. This should be good to continue.
> 
> I've forgotten what the exact breaking change is here - but please follow the [steps here](https://googlecloudplatform.github.io/magic-modules/develop/breaking-changes/make-a-breaking-change/#make-the-change-on-feature-branch-major-release-600) to edit the upgrade guide so User's know how to handle the breaking change.
> 
> @daanheikens @trodge

Thank you @c2thorn, the issue was that one of the fields was treated required while it was optional. See also: https://github.com/GoogleCloudPlatform/magic-modules/pull/10697

If I recall correctly, it was related to this issue:  https://github.com/hashicorp/terraform-provider-google/issues/18115 where @rileykarson explains the issue in detail. I am not sure if it's already fixed in this major release, if not, we indeed have to update the documentation and add more tests I think?

Happy to hear your advice.

c2thorn:
@daanheikens Thanks for the information. As I now understand, this is a breaking change due to the parent field having [schema_config_mode_attr](https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/products/compute/Subnetwork.yaml#L210).

I planned on testing the effect of [removing schema_config_mode_attr](https://github.com/hashicorp/terraform-provider-google/issues/12824) as part of 6.0, but have not gotten to it yet.

With schema_config_mode_attr, adding this new field is a breaking change because it becomes required.
Without schema_config_mode_attr, this should not be a breaking change, but it is unclear if removing schema_config_mode_attr itself is a breaking change. If you are interested in testing this out yourself as part of this PR that would be helpful, but otherwise I will have to report back in a few weeks after I've had time to test.

Assuming schema_config_mode_attr is removed outside of this PR, this will no longer be a breaking change and not need the associated breaking change documentation.

trodge:
Removing `schema_config_mode_attr` would be a breaking change in that a config could produce this error:
```
  error=
  | exit status 1
  | 
  | Error: Unsupported argument
  | 
  |   on terraform_plugin_test.tf line 12, in resource "google_compute_subnetwork" "network-with-private-secondary-ip-ranges":
  |   12:   secondary_ip_range = []
  | 
  | An argument named "secondary_ip_range" is not expected here. Did you mean to
  | define a block of type "secondary_ip_range"?
  ```
  
  If the user is attempting to use the current recommended method of specifying an empty block in order to remove secondary ip ranges from their config. (This being the workaround Riley mentions in https://github.com/hashicorp/terraform-provider-google/issues/12824 )
  
 If we're no longer concerned about maintaining 0.12 -> 0.13 forwards compatibility in the 6.0 version of the provider, could we remove `schema_config_mode_attr` and all tests/docs mentioning the `[]` syntax?

**Comment by daanheikens** [comment]
> Removing `schema_config_mode_attr` would be a breaking change in that a config could produce this error:
> 
> ```
>   error=
>   | exit status 1
>   | 
>   | Error: Unsupported argument
>   | 
>   |   on terraform_plugin_test.tf line 12, in resource "google_compute_subnetwork" "network-with-private-secondary-ip-ranges":
>   |   12:   secondary_ip_range = []
>   | 
>   | An argument named "secondary_ip_range" is not expected here. Did you mean to
>   | define a block of type "secondary_ip_range"?
> ```
> 
> If the user is attempting to use the current recommended method of specifying an empty block in order to remove secondary ip ranges from their config. (This being the workaround Riley mentions in [hashicorp/terraform-provider-google#12824](https://github.com/hashicorp/terraform-provider-google/issues/12824) )
> 
> If we're no longer concerned about maintaining 0.12 -> 0.13 forwards compatibility in the 6.0 version of the provider, could we remove `schema_config_mode_attr` and all tests/docs mentioning the `[]` syntax?

@trodge @c2thorn I can try some variations here and write some tests/examples with `schema_config_mode_attr` removed if you are OK with that? Could you please guide me a bit in where I can start? Let me know, happy to help.

**Comment by c2thorn** [comment]
> @trodge @c2thorn I can try some variations here and write some tests/examples with `schema_config_mode_attr` removed if you are OK with that? Could you please guide me a bit in where I can start? Let me know, happy to help.

I plan to get to testing other fields this week, but if you have time to start with this field, that would be helpful. With removing  `schema_config_mode_attr`, we'd essentially just need to manually verify if it is possible to remove the field `secondary_ip_ranges` or not. I don't think there is a specific test that needs to be added to MMv1, what do you think @trodge

**Comment by trodge** [comment]
> > @trodge @c2thorn I can try some variations here and write some tests/examples with `schema_config_mode_attr` removed if you are OK with that? Could you please guide me a bit in where I can start? Let me know, happy to help.
> 
> I plan to get to testing other fields this week, but if you have time to start with this field, that would be helpful. With removing `schema_config_mode_attr`, we'd essentially just need to manually verify if it is possible to remove the field `secondary_ip_ranges` or not. I don't think there is a specific test that needs to be added to MMv1, what do you think @trodge

We have a test already that removes secondary ip ranges using `secondary_ip_range = []` syntax. That would need to be updated to whatever the new syntax is along with the documentation.

**Comment by c2thorn** [comment]
Just updating that I'm getting to this now and should hopefully have an update tomorrow on the effect of removing `schema_config_mode_attr` and how to handle emptying fields without it.

**Comment by c2thorn** [comment]
After looking into this, this PR can be merged without removing `schema_config_mode_attr` from `secondaryIpRange`

I plan to remove `schema_config_mode_attr` separately after adding a 5.X workaround with https://github.com/GoogleCloudPlatform/magic-modules/pull/11410 

So once that's done this should not be a breaking change, but it does need to go into the 6.0 branch.

**Comment by daanheikens** [comment]
> After looking into this, this PR can be merged without removing `schema_config_mode_attr` from `secondaryIpRange`
> 
> 
> 
> I plan to remove `schema_config_mode_attr` separately after adding a 5.X workaround with https://github.com/GoogleCloudPlatform/magic-modules/pull/11410 
> 
> 
> 
> So once that's done this should not be a breaking change, but it does need to go into the 6.0 branch.

Ah great news @c2thorn. I do see the VCR test is still failing, anything we need to do for that?

**Comment by c2thorn** [comment]
> Ah great news @c2thorn. I do see the VCR test is still failing, anything we need to do for that?

yes it looks like the error for `TestAccComputeSubnetwork_subnetworkReservedInternalRangeExample` is real
```
        Error: "name" ("") doesn't match regexp "^(?:[a-z](?:[-a-z0-9]{0,61}[a-z0-9])?)$"
        
          with google_compute_network.default,
          on terraform_plugin_test.tf line 12, in resource "google_compute_network" "default":
          12:   name                    = ""
```
@trodge

File: mmv1/products/compute/Subnetwork.yaml [hunk 1]
```diff
@@ -123,6 +123,19 @@ examples:
     vars:
       subnetwork_name: 'subnet-cidr-overlap'
       network_name: 'net-cidr-overlap'
+  - !ruby/object:Provider::Terraform::Examples
+    name: 'subnetwork_reserved_internal_range'
+    min_version: beta
+    primary_resource_id: 'subnetwork-reserved-internal-range'
+    vars:
+      subnetwork_name: 'subnetwork-reserved-internal-range'
```
trodge, line 131:
```suggestion
      subnetwork_name: 'subnetwork-reserved-internal-range'
      network_name: 'network-reserved-internal-range'
```

trodge (changes requested):
This should fix the test error @c2thorn mentioned.

daanheikens:
> This should fix the test error @c2thorn mentioned.

Applied the suggestions!

trodge (approved):

PR #10899, opened 2024-06-06
Add Service Networking VPC Service Controls resource.

juliocc:
This PR add a new resource to manage service networking VPC-SC

Related Issue hashicorp/terraform-provider-google#9317

```release-note:new-resource
`google_service_networking_vpc_service_controls`
```

### DIFF
--- a/mmv1/templates/terraform/constants/service_networking_vpc_service_controls.go.erb
+++ b/mmv1/templates/terraform/constants/service_networking_vpc_service_controls.go.erb
@@ -1,0 +1,76 @@
+func resourceServiceNetworkingVPCServiceControlsSet(d *schema.ResourceData, meta interface{}, config *transport_tpg.Config) error {
+	userAgent, err := tpgresource.GenerateUserAgentString(d, config.UserAgent)
+	if err != nil {
+		return err
+	}
+
+	project, err := tpgresource.GetProject(d, config)
+	if err != nil {
+		return err
+	}
+	projectNumber, err := getProjectNumber(d, config, project, userAgent)
+	if err != nil {
+		return err
+	}
+
+	network := d.Get("network").(string)
+	enabled := d.Get("enabled").(bool)
+
+	obj := make(map[string]interface{})
+	obj["consumerNetwork"] = fmt.Sprintf("projects/%s/global/networks/%s", projectNumber, network)
+
+	url, err := tpgresource.ReplaceVars(d, config, "{{ServiceNetworkingBasePath}}services/{{service}}")
+	if err != nil {
+		return err
+	}
+
+	if enabled {
+		url = url + ":enableVpcServiceControls"
+	} else {
+		url = url + ":disableVpcServiceControls"
+	}
+
+	log.Printf("[DEBUG] Setting service networking VPC service controls: %#v", obj)
+	billingProject := ""
+
+	// err == nil indicates that the billing_project value was found
+	if bp, err := tpgresource.GetBillingProject(d, config); err == nil {
+		billingProject = bp
+	}
+
+	headers := make(http.Header)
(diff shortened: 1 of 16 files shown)


juliocc:
I think the remaining test failures are not related to this PR.

Regarding the the `terraform-google-conversion-*` checks, I'm not sure how to fix these. @c2thorn what should I do about these?

c2thorn:
> I think the remaining test failures are not related to this PR.
> 
> Regarding the the `terraform-google-conversion-*` checks, I'm not sure how to fix these. @c2thorn what should I do about these?

some of the functions in the `constants` file are not generated in TGC. You would either have to add them to https://github.com/GoogleCloudPlatform/magic-modules/tree/main/mmv1/third_party/tgc (not sure how reasonable this would be to do) or skip TGC generation completely with `exclude_tgc: true`

juliocc:
> some of the functions in the `constants` file are not generated in TGC. You would either have to add them to https://github.com/GoogleCloudPlatform/magic-modules/tree/main/mmv1/third_party/tgc (not sure how reasonable this would be to do) or skip TGC generation completely with `exclude_tgc: true`

Which one would you prefer?

c2thorn:
> > some of the functions in the `constants` file are not generated in TGC. You would either have to add them to [`main`/mmv1/third_party/tgc](https://github.com/GoogleCloudPlatform/magic-modules/tree/main/mmv1/third_party/tgc?rgh-link-date=2024-06-10T19%3A09%3A36Z) (not sure how reasonable this would be to do) or skip TGC generation completely with `exclude_tgc: true`
> 
> Which one would you prefer?

excluding with a comment explaining it is due to the constant custom_code

c2thorn:
rerunning one more time

mmv1/templates/terraform/constants/service_networking_vpc_service_controls.go.erb
```diff
@@ -0,0 +1,76 @@
...
+	obj := make(map[string]interface{})
+	obj["consumerNetwork"] = fmt.Sprintf("projects/%s/global/networks/%s", projectNumber, network)
+
+	url, err := tpgresource.ReplaceVars(d, config, "{{ServiceNetworkingBasePath}}services/{{service}}")
+	if err != nil {
+		return err
+	}
+
+	if enabled {
+		url = url + ":enableVpcServiceControls"
```
c2thorn, line 28: Precreate + Preread custom code is pretty atypical, and I want to make sure I'm understanding the reasoning behind their inclusion. Is it because of this enabled vs disabled logic?

juliocc: The preread is to translate the project ID into a project number as required by the service networking API. This translation is needed to build the correct URL. Note that this same issue has affected other [similar resources](https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/third_party/terraform/services/servicenetworking/resource_google_service_networking_peered_dns_domain.go#L243)

The custom_create/custom_update is needed, again, because of the atypical URLs used by this API. The API uses distinct URLs for [enabling](https://cloud.google.com/service-infrastructure/docs/service-networking/reference/rest/v1/services/enableVpcServiceControls) and [disabling](https://cloud.google.com/service-infrastructure/docs/service-networking/reference/rest/v1/services/disableVpcServiceControls) the controls, and doesn't involve creating or updating a traditional resource. Since the URL depends on the desired state, we need custom logic to construct the correct request.

c2thorn: Thanks for the explanation and additional context.

c2thorn (approved):

PR #10906, opened 2024-06-07
Add allowlistedCertificates field to TrustConfig

Hamzawy63:
This CL adds allowlistedCertificates field to TrustConfig resource.
Fixes https://github.com/hashicorp/terraform-provider-google/issues/18170 and https://github.com/hashicorp/terraform-provider-google/issues/18385

Tests that I run: 
`make testacc TEST=./google/services/certificatemanager TESTARGS='-run=TestAccCertificateManagerTrustConfig_certificateManagerTrustConfigAllowlistedCertificatesExample'`

`make testacc TEST=./google/services/certificatemanager TESTARGS='-run=TestAccCertificateManagerTrustConfig_update'`

Manual tests that I performed: 
- Create a resource with one allowlisted
- Add one more cert to the resource
- Remove one cert from the resoruce

```release-note:enhancement
certificatemanager: added `allowlisted_certificates` to `google_certificate_manager_trust_config`
```

### DIFF
--- a/mmv1/products/certificatemanager/TrustConfig.yaml
+++ b/mmv1/products/certificatemanager/TrustConfig.yaml
@@ -55,8 +55,6 @@ examples:
     primary_resource_id: 'default'
     vars:
       trust_config_name: 'trust-config'
-custom_code: !ruby/object:Provider::Terraform::CustomCode
-  pre_update: templates/terraform/pre_update/certificate_manager_trust_config.go.erb
 parameters:
   - !ruby/object:Api::Type::String
     name: 'name'
@@ -92,7 +90,6 @@ properties:
   - !ruby/object:Api::Type::KeyValueLabels
     name: 'labels'
     description: 'Set of label tags associated with the trust config.'
-    immutable: true
   - !ruby/object:Api::Type::String
     name: 'description'
     description: |
--- a/mmv1/products/certificatemanager/TrustConfig.yaml
+++ b/mmv1/products/certificatemanager/TrustConfig.yaml
@@ -50,6 +50,11 @@ examples:
     primary_resource_id: 'default'
     vars:
       trust_config_name: 'trust-config'
+  - !ruby/object:Provider::Terraform::Examples
+    name: 'certificate_manager_trust_config_allowlisted_certificates'
+    primary_resource_id: 'default'
+    vars:
+      trust_config_name: 'trust-config'
 custom_code: !ruby/object:Provider::Terraform::CustomCode
   pre_update: templates/terraform/pre_update/certificate_manager_trust_config.go.erb
 parameters:
@@ -124,3 +129,15 @@ properties:
                   PEM intermediate certificate used for building up paths for validation.
                   Each certificate provided in PEM format may occupy up to 5kB.
                 sensitive: true
+  - !ruby/object:Api::Type::Array
+    name: allowlistedCertificates
+    description: |
+       Allowlisted PEM-encoded certificates. A certificate matching an allowlisted certificate is always considered valid as long as
+       the certificate is parseable, proof of private key possession is established, and constraints on the certificate's SAN field are met.
+    item_type: !ruby/object:Api::Type::NestedObject
+      properties:
+        - !ruby/object:Api::Type::String
+          name: 'pemCertificate'
+          description: |
+            PEM certificate that is allowlisted. The certificate can be up to 5k bytes, and must be a parseable X.509 certificate.
+          required: true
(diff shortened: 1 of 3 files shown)


Hamzawy63:
During my manual testing, I faced a strange behavior: If I make two consecutive update to the resource, the first update will be in-place (as expected) and the second one will force replace the resource. 
That was my resource definition: 
```
resource "google_certificate_manager_trust_config" "default" {
  name        = "sample-trust-configs"
  description = "A sample trust config resource with allowlisted certificates"
  location = "global"

  allowlisted_certificates  {
    pem_certificate = file("cert.pem") 
  }
  labels = {
    foo = "bar"
  }
}
````
The first update, I changed cert.pem to cert2.pem and the update was in-place and succssful 
The second update, I changed cert2.pem back to cert.pem. In this case, terraform decided to force replace the resource because it detected a change in effective_labels variable

```
  # google_certificate_manager_trust_config.default must be replaced
-/+ resource "google_certificate_manager_trust_config" "default" {
      ~ create_time      = "2024-06-10T16:43:25.052728242Z" -> (known after apply)
      ~ effective_labels = { # forces replacement
          + "foo" = "bar"
    }
```

I can't determine the root cause for this issue, this resource has a pre-update custom code that changes the updateMask to `*`, but even when I removed it, the issue still remains.

**Comment by Hamzawy63** [comment]
Gentle reminder @trodge

**Comment by trodge** [comment]
> During my manual testing, I faced a strange behavior: If I make two consecutive update to the resource, the first update will be in-place (as expected) and the second one will force replace the resource. That was my resource definition:
> 
> ```
> resource "google_certificate_manager_trust_config" "default" {
>   name        = "sample-trust-configs"
>   description = "A sample trust config resource with allowlisted certificates"
>   location = "global"
> 
>   allowlisted_certificates  {
>     pem_certificate = file("cert.pem") 
>   }
>   labels = {
>     foo = "bar"
>   }
> }
> ```
> 
> The first update, I changed cert.pem to cert2.pem and the update was in-place and succssful The second update, I changed cert2.pem back to cert.pem. In this case, terraform decided to force replace the resource because it detected a change in effective_labels variable
> 
> ```
>   # google_certificate_manager_trust_config.default must be replaced
> -/+ resource "google_certificate_manager_trust_config" "default" {
>       ~ create_time      = "2024-06-10T16:43:25.052728242Z" -> (known after apply)
>       ~ effective_labels = { # forces replacement
>           + "foo" = "bar"
>     }
> ```
> 
> I can't determine the root cause for this issue, this resource has a pre-update custom code that changes the updateMask to `*`, but even when I removed it, the issue still remains.

It looks like in this resource the `effective_labels` field has `ForceNew`. I'm not sure why this would be since the field is computed. Removing it should fix this issue.

**Comment by Hamzawy63** [comment]
> It looks like in this resource the effective_labels field has ForceNew. I'm not sure why this would be since the field is computed. Removing it should fix this issue

Do you mean to make `labels` not immutable? as we don't have control over  `effective_labels` field (it does not belong to only TrustConfig resource).  See: https://github.com/GoogleCloudPlatform/magic-modules/pull/8950

I believe that making the `labels` not immutable will solve the issue, because `effective_labels` is immutable if `labels` is immutable. 

It's ok in this resource to make labels not immutable. There was an issue with updating labels before https://github.com/GoogleCloudPlatform/magic-modules/pull/8651#discussion_r1295239758, but it's no longer the case (I tested that manually). It's also ok to remove the `pre_update` custom code that was created earlier.

However, I want to point out two things:

- I'm surprised that `effective_labels` field changed during my experiment, I did not touch the labels during updating the resource and yet `effective_labels` changed for some reason. 
- I would guess that this is an issue that will face any resource that has an immutable `label` field (and hence an immutable `effective_label`).  Maybe it's rare to have a resource with immutable `labels` that's why this issue is not common.

**Comment by Hamzawy63** [comment]
@trodge Friendly reminder

File: mmv1/products/certificatemanager/TrustConfig.yaml [hunk 1]
```diff
@@ -50,8 +50,11 @@ examples:
     primary_resource_id: 'default'
     vars:
       trust_config_name: 'trust-config'
-custom_code: !ruby/object:Provider::Terraform::CustomCode
-  pre_update: templates/terraform/pre_update/certificate_manager_trust_config.go.erb
```
trodge, line 54: Is this pre-update no longer required? It looks like it adds "*" as the update mask, is that no longer needed?

mmv1/products/certificatemanager/TrustConfig.yaml
```diff
@@ -124,3 +126,15 @@ properties:
                   PEM intermediate certificate used for building up paths for validation.
                   Each certificate provided in PEM format may occupy up to 5kB.
                 sensitive: true
+  - !ruby/object:Api::Type::Array
+    name: allowlistedCertificates
```
trodge, line 130: Could you also add a test that updates this field if it is updateable?

trodge (changes requested):

mmv1/products/certificatemanager/TrustConfig.yaml
Hamzawy63: Done! added the field to the existing update test

trodge (approved):

PR #10510, opened 2024-04-22
Use data google_kms(_key_ring | _crypto_key) instead of resource

damondouglas:
This PR fixes https://github.com/hashicorp/terraform-provider-google/issues/17773 refactoring mmv1/third_party/terraform/services/dataflow/resource_dataflow_flex_template_job_test.go.erb to reference pre-existing google_kms_key_ring and google_kms_crypto_key, instead of creating/destroying new ones for each test.

To validate this PR, per [generate-providers](https://googlecloudplatform.github.io/magic-modules/get-started/generate-providers/) and [run-tests](https://googlecloudplatform.github.io/magic-modules/develop/test/run-tests/) documentation:

1. Generated the provider changes (google-beta only):

```
make provider VERSION=beta OUTPUT_PATH="$GOPATH/src/github.com/hashicorp/terraform-provider-google-beta" PRODUCT=dataflow
```

2. Create Google Cloud project environment

I created a clean, unused Google Cloud project, enabling the minimally permissive IAM roles to the default service account as well as turn on KMS, Dataflow, and other required API services.

3. Ran tests

Set the required environment variables.

```
export GOOGLE_USE_DEFAULT_CREDENTIALS=true
export GOOGLE_ZONE=us-west1-a
export GOOGLE_REGION=us-west1
export GOOGLE_PROJECT=<GCP project ID not shown>
```

Ran the tests.

```
cd $GOPATH/src/github.com/hashicorp/terraform-provider-google-beta
make testacc TEST=./google-beta/services/dataflow TESTARGS='-run=TestAccDataflowFlexTemplateJob_withKmsKey'
```

4. Validated output and Dataflow Job

```
==> Checking that code complies with gofmt requirements...
go vet
TF_ACC=1 TF_SCHEMA_PANIC_ON_ERROR=1 go test ./google-beta/services/dataflow -v -run=TestAccDataflowFlexTemplateJob_withKmsKey -timeout 240m -ldflags="-X=github.com/hashicorp/terraform-provider-google-beta/version.ProviderVersion=acc"
=== RUN   TestAccDataflowFlexTemplateJob_withKmsKey
=== PAUSE TestAccDataflowFlexTemplateJob_withKmsKey
=== CONT  TestAccDataflowFlexTemplateJob_withKmsKey
--- PASS: TestAccDataflowFlexTemplateJob_withKmsKey (354.30s)
PASS
ok      github.com/hashicorp/terraform-provider-google-beta/google-beta/services/dataflow       354.431s
```

[...]

PR #10908, opened 2024-06-07
Removed unused variable

melinath:
Follow-up to https://github.com/GoogleCloudPlatform/magic-modules/pull/10510

```release-note:none

```


melinath:
@damondouglas FYI if you're curious, this is what I was trying to convey - but the GitHub interface made it difficult because it touches lines too far away from the rest of the PR.

damondouglas:
> @damondouglas FYI if you're curious, this is what I was trying to convey - but the GitHub interface made it difficult because it touches lines too far away from the rest of the PR.

Thank you for the FYI. Why didn't the test/lint/vet locally in google-beta catch this?

melinath:
> Why didn't the test/lint/vet locally in google-beta catch this?

Go linting doesn't check whether variables in a function signature are unused (because in Go, it's common to have a framework that requires a particular signature, like Terraform's [DiffSuppressFunc](https://developer.hashicorp.com/terraform/plugin/sdkv2/schemas/schema-behaviors#diffsuppressfunc), but the function may not actually _use_ all of the vars. Some more discussion of this: https://stackoverflow.com/questions/22549228/why-does-go-allow-compilation-of-unused-function-parameters

c2thorn (approved):

PR #10915, opened 2024-06-07
Allow ENTERPRISE_PLUS as default Edition for POSTGRES_16

skysarthak:
Current default: `ENTERPRISE` for `settings.edition` is not consistent with the API which treats `EDITION_UNSPECIFIED` as the default (refer [link](https://github.com/googleapis/google-api-go-client/blob/v0.183.0/sqladmin/v1beta4/sqladmin-gen.go#L3580)) when customer doesn't provide any. With PostgreSQL version 16, the default Edition will be `ENTERPRISE_PLUS` as documented in the PostgreSQL June 07, 2024 [release notes](https://cloud.google.com/sql/docs/postgres/release-notes#June_07_2024). Thus, unset the default `edition` so that it can instead be computed in the API, allowing the API to use `ENTERPRISE_PLUS` as the default for POSTGRES_16.

b/338258663

I acknowledge that I have:

- [X] Searched through the [issue tracker](https://github.com/hashicorp/terraform-provider-google/issues) for an open issue that this either resolves or contributes to, commented on it to claim it, and written "fixes {url}" or "part of {url}" in this PR description. If there were no relevant open issues, I opened one and commented that I would like to work on it (not necessary for very small changes).
- [X] Ensured that all new fields I added that can be set by a user appear in at least one [example](https://github.com/GoogleCloudPlatform/magic-modules/tree/main/mmv1/templates/terraform/examples) (for generated resources) or [third_party test](https://github.com/GoogleCloudPlatform/magic-modules/tree/main/mmv1/third_party/terraform/tests) (for handwritten resources or update tests).
- [X] [Generated Terraform providers](https://github.com/GoogleCloudPlatform/magic-modules/blob/main/docs/content/docs/getting-started/generate-providers.md), and ran [`make test` and `make lint`](https://googlecloudplatform.github.io/magic-modules/docs/getting-started/run-provider-tests/#run-unit-tests) in the generated providers to ensure it passes unit and linter tests.
- [X] [Ran](https://github.com/GoogleCloudPlatform/magic-modules/blob/main/docs/content/docs/getting-started/run-provider-tests.md) relevant acceptance tests using my own Google Cloud project and credentials (If the acceptance tests do not yet pass or you are unable to run them, please let your reviewer know).
- [X] Read the [Release Notes Guide](https://github.com/GoogleCloudPlatform/magic-modules/blob/main/.ci/RELEASE_NOTES_GUIDE.md) before writing my release note below.

```release-note:enhancement
sql: Unset the default `edition` so that it can instead be computed in the API, allowing the API to use `ENTERPRISE_PLUS` as the default for POSTGRES_16.
```

### DIFF
--- a/mmv1/third_party/terraform/services/sql/resource_sql_database_instance.go.erb
+++ b/mmv1/third_party/terraform/services/sql/resource_sql_database_instance.go.erb
@@ -6,7 +6,6 @@ import (
 	"errors"
 	"fmt"
 	"log"
-	"reflect"
 	"strings"
 	"time"
 
@@ -178,7 +177,7 @@ func ResourceSqlDatabaseInstance() *schema.Resource {
 						"edition": {
 							Type:         schema.TypeString,
 							Optional:     true,
-							Default:      "ENTERPRISE",
+							Computed:     true,
 							ValidateFunc: validation.StringInSlice([]string{"ENTERPRISE", "ENTERPRISE_PLUS"}, false),
 							Description:  `The edition of the instance, can be ENTERPRISE or ENTERPRISE_PLUS.`,
 						},
@@ -2059,7 +2058,7 @@ func flattenSettings(settings *sqladmin.Settings, d *schema.ResourceData) []map[
 	data := map[string]interface{}{
 		"version":                      settings.SettingsVersion,
 		"tier":                         settings.Tier,
-		"edition":                      flattenEdition(settings.Edition),
+		"edition":                      settings.Edition,
 		"activation_policy":            settings.ActivationPolicy,
 		"availability_type":            settings.AvailabilityType,
 		"collation":                    settings.Collation,
@@ -2393,14 +2392,6 @@ func flattenPasswordValidationPolicy(passwordValidationPolicy *sqladmin.Password
 		"enable_password_policy":      passwordValidationPolicy.EnablePasswordPolicy,
 	}
 	return []map[string]interface{}{data}
-}
-
-func flattenEdition(v interface{}) string {
-	if v == nil || tpgresource.IsEmptyValue(reflect.ValueOf(v)) {
-		return "ENTERPRISE"
-	}
-
-	return v.(string)
 }
 
 func instanceMutexKey(project, instance_name string) string {
--- a/mmv1/third_party/terraform/services/sql/resource_sql_database_instance_test.go
+++ b/mmv1/third_party/terraform/services/sql/resource_sql_database_instance_test.go
@@ -1564,6 +1564,32 @@ func TestAccSQLDatabaseInstance_DenyMaintenancePeriod(t *testing.T) {
 	})
 }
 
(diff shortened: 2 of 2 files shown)


roaks3 (commented):
I have 2 main concerns on first look:
1. This appears to be a breaking change, because a user that does not set the `edition` in their config now has different behavior. Can you confirm that is the intention?
2. The default of `ENTERPRISE` might have been added for convenience, like if it is often desired by users. Are we sure we would want to remove that?

skysarthak:
Hi [roaks3](https://github.com/roaks3), thanks for taking a look.

I have updated the pull request to take care of the failing `TestAccSqlDatabaseInstance_Mysql_Edition_Upgrade` test, it should now succeed.

> 1. This appears to be a breaking change, because a user that does not set the edition in their config now has different behavior. Can you confirm that is the intention?

Yes it is. The earlier change where we defaulted Edition to "ENTERPRISE" was inconsistent with the CloudSQL API that actually uses default Edition as "EDITION_UNSPECIFIED" (refer: type [Settings](https://pkg.go.dev/google.golang.org/api/sqladmin/v1beta4#Settings)). Today a "ENTERPRISE" instance is configured similar to a "EDITION_UNSPECIFIED" instance but this could change within the CloudSQL API which will create inconsistent behavior for the customers when dealing directly with CloudSQL API vs Terraform provider.

> 2. The default of ENTERPRISE might have been added for convenience, like if it is often desired by users. Are we sure we would want to remove that?

Issue is that with PostgreSQL version 16, the default Edition will be ENTERPRISE_PLUS as documented in the PostgreSQL June 07, 2024 [release notes](https://cloud.google.com/sql/docs/postgres/release-notes#June_07_2024). Similar to this, if the CloudSQL API changes the default Edition for any database version going forward, Terraform provider will not be able to keep up. I wouldn't want to compromise correctness for convenience.

roaks3 (approved):
LGTM, marking as a breaking change and flagging @c2thorn (also re-triggering tests to confirm they pass)

rileykarson:
I think we missed this during the `6.0.0` window? Our primary unit of accounting were issues in the TPG repo rather than the `6.0.0` label in this repo. FYI @c2thorn we should cover that during the retro, unless I've missed other context.

We should consider whether we want to wait for 7.0.0 (and likely just update examples to specify the new default) or change it in a minor version. IMO it's on the line of whether we consider it acceptable or not- the revised rules in the API are that this is a default controlled by the resource's internal version (the SQL version in this case, we see the same volatile defaults happen with GKE's versions too). Existing configurations that have been applied will not see a diff, and existing configurations that specify an older version (and versions are required!) will adopt the same default.

That does mean that a replacement of a database instance with a newer version would see a changed value. So the scope is pretty small but it does observably change things for those users.

c2thorn:
Closing this PR since it is past the major release, there are conflicting files, and will be further broken by the upcoming Go migration.

this will be correctly tracked for 7.0.0 now that I have made a Github issue to track it: https://github.com/hashicorp/terraform-provider-google/issues/19606

PR #10916, opened 2024-06-07
Ignored review requests that were later removed

melinath:
For an example of the behavior this fixes, see https://github.com/GoogleCloudPlatform/magic-modules/pull/10853#issuecomment-2143089636

```release-note:none

```

### DIFF
--- a/.ci/magician/cmd/scheduled_pr_reminders.go
+++ b/.ci/magician/cmd/scheduled_pr_reminders.go
@@ -283,11 +283,21 @@ func notificationState(pr *github.PullRequest, issueEvents []*github.IssueEvent,
 	})
 
 	var latestReviewRequest *github.IssueEvent
+	removedRequests := map[string]struct{}{}
 	for _, event := range issueEvents {
+		if *event.Event == "review_request_removed" && event.RequestedReviewer != nil {
+			removedRequests[*event.RequestedReviewer.Login] = struct{}{}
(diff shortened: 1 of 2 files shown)


.ci/magician/cmd/scheduled_pr_reminders.go
```diff
@@ -283,13 +283,23 @@ func notificationState(pr *github.PullRequest, issueEvents []*github.IssueEvent,
...
+		}
 		if *event.Event != "review_requested" {
 			continue
 		}
+		// Ignore review requests for users who no longer exist.
 		if event.RequestedReviewer == nil {
 			continue
 		}
+		// Ignore review requests that were later removed.
+		if _, ok := removedRequests[*event.RequestedReviewer.Login]; ok {
```
roaks3, line 300: I see the test you have, but I'm confused why this works. It seems like you would want to remove the entry from the set here, in order to allow another subsequent review request to be honored. Am I missing something?

roaks3 (approved):
LGTM, but there was a part of the implementation I didn't understand

.ci/magician/cmd/scheduled_pr_reminders.go
melinath: We're iterating through the events newest -> oldest - so if we see a review request removal then we know that we should stop considering review requests for that user.

PR #10922, opened 2024-06-07
add redis self service update support

yangspirit:
add redis self service update support

```release-note:enhancement
redis: added `maintenance_version` field to `google_redis_instance`
```


slevenick (changes requested):
Can you add a test that sets and then updates this field?

yangspirit (commented):
Add selfservice update test

slevenick (changes requested):
Lint is failling with:
gofmt needs running on the following files:
./google-beta/services/redis/resource_redis_instance_test.go
You can use the command: `make fmt` to reformat code.

You'll need to get those changes into the magic-modules repo as you can't run the make fmt command there

slevenick (approved):

PR #10923, opened 2024-06-08
Fixed an issue which force optional field is_ca in Certificate to be set

gfxcc:
b/345539403

The semantic of is_ca for Certificate in terraform is

1. If `ca_options` is not specified at all (enabled by this PR) OR `ca_options {non_ca=true}` is specified
    -  a certificate will be issued with a basicConstraints ext(is_ca=false)
2. if `ca_options {}` is specified, a certificate will be issued without a basicConstraints ext.
4. if `ca_options {is_ca=true}` is specified, a certificate will be issued with a basicConstraints ext (is_ca=true).

```release-note: bug
privateca: fixed an issue which force optional field is_ca in Certificate to be set
```


shuyama1 (commented):
I think the failing test `TestAccPrivatecaCertificate_privatecaCertificateUpdate` is related. The test failed due to error:
```
vcr_utils.go:152: Step 3/6 error: Error running pre-apply refresh: exit status 1
        
        Error: Provider produced invalid plan
        
        Provider "registry.terraform.io/hashicorp/google" has indicated "requires
        replacement" on google_privateca_certificate.default for a non-existent
        attribute path cty.Path{cty.GetAttrStep{Name:"config"},
        cty.IndexStep{Key:cty.NumberIntVal(0)}, cty.GetAttrStep{Name:"x509_config"},
        cty.IndexStep{Key:cty.NumberIntVal(0)}, cty.GetAttrStep{Name:"ca_options"},
        cty.IndexStep{Key:cty.NumberIntVal(0)}, cty.GetAttrStep{Name:"is_ca"}}.
        
        issue tracker.
```

gfxcc:
> I think the failing test `TestAccPrivatecaCertificate_privatecaCertificateUpdate` is related. The test failed due to error:
> 
> ```
> vcr_utils.go:152: Step 3/6 error: Error running pre-apply refresh: exit status 1
>         
>         Error: Provider produced invalid plan
>         
>         Provider "registry.terraform.io/hashicorp/google" has indicated "requires
>         replacement" on google_privateca_certificate.default for a non-existent
>         attribute path cty.Path{cty.GetAttrStep{Name:"config"},
>         cty.IndexStep{Key:cty.NumberIntVal(0)}, cty.GetAttrStep{Name:"x509_config"},
>         cty.IndexStep{Key:cty.NumberIntVal(0)}, cty.GetAttrStep{Name:"ca_options"},
>         cty.IndexStep{Key:cty.NumberIntVal(0)}, cty.GetAttrStep{Name:"is_ca"}}.
>         
>         This is a bug in the provider, which should be reported in the provider's own
>         issue tracker.
> ```

The test passed when running it locally. Do you think it could due to the limitation from RECORDING mode?

shuyama1:
It shouldn't be affected by running in VCR, especially in recording mode, as it's just running the actual acceptance test. Let me re-run the test for now, and if it fails again, I'll need to look a bit closer into the failure.

/gcbrun

shuyama1 (commented):
Sorry for the late review. Do you think it's possible to add test coverage on this change?

shuyama1 (commented):
Just want to confirm I understand the cases correctly -  
* omitting the `ca_options` block in Terraform would be the same as setting `ca_options {non_ca=true}`, which means we will send `CaOptions {is_ca=false}` to the API. 
* omitting the `ca_options` block is different from explicitly set it as a empty block `ca_options {}` in Terraform, as the latter means CaOptions {} (is_ca unspecified) to the API?

Will these be expected behaviors for users?

gfxcc:
> Just want to confirm I understand the cases correctly -
> 
> * omitting the `ca_options` block in Terraform would be the same as setting `ca_options {non_ca=true}`, 
> which means we will send `CaOptions {is_ca=false}` to the API.

Correct. The requests are different (omitting `ca_options` vs `CaOptions {is_ca=false}`), but result to the same behavior.

> * omitting the `ca_options` block is different from explicitly set it as a empty block `ca_options {}` in Terraform, as the latter means CaOptions {} (is_ca unspecified) to the API?

Correct. Those 2 config has different meaning to the API.

The new behaviors are listed in the PR description.

shuyama1:
Sorry for the delay on review. I'll be taking a look later today.

shuyama1 (commented):
I've tested with this change a bit locally and find out that setting empty block `ca_options{}` in `google_privateca_ca_pool` or `google_privateca_certificate` will start to cause a [permadiff](https://googlecloudplatform.github.io/magic-modules/develop/permadiff/) after this change. perma-destroy for `google_privateca_certificate` actually as it will destroy and recreate the resource

```
~ resource "google_privateca_ca_pool" "default" {
                  ~ baseline_values {
        
        
                      + ca_options {}
```

```
-/+ resource "google_privateca_certificate" "default" {
                  ~ x509_config {
        
                      + ca_options { # forces replacement}
```

PR #10927, opened 2024-06-10
Updated description of connection_draining_timeout_sec, balancing_mode and  outlier_detection in backend_service and regional_backend_service before changing defaults

pawelJas:
Adds info in the description of `connection_draining_timeout_sec`, `balancing_mode` and `outlier_detection` about breaking change in the next major release. 

https://github.com/hashicorp/terraform-provider-google/issues/13478
https://github.com/hashicorp/terraform-provider-google/issues/12299
https://github.com/hashicorp/terraform-provider-google/issues/15210
https://github.com/hashicorp/terraform-provider-google/issues/17257

```release-note:note
compute: Updated field description of `connection_draining_timeout_sec`, `balancing_mode` and `outlier_detection` in `google_compute_region_backend_service` and `google_compute_backend_service`  to inform that default values will be changed
```

### DIFF
--- a/mmv1/products/compute/RegionBackendService.yaml
+++ b/mmv1/products/compute/RegionBackendService.yaml
@@ -164,6 +164,8 @@ properties:
 
             See the [Backend Services Overview](https://cloud.google.com/load-balancing/docs/backend-service#balancing-mode)
             for an explanation of load balancing modes.
+
+            From version 6.0.0 default value will be UTILIZATION to match default GCP value.
         - !ruby/object:Api::Type::Double
           name: 'capacityScaler'
           description: |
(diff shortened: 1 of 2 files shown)


mmv1/products/compute/RegionBackendService.yaml
```diff
@@ -658,6 +658,7 @@ properties:
         description: |
           Time for which instance will be drained (not accept new
           connections, but still work to finish started).
+        deprecation_message: "connection_draining_timeout_sec default value will be changed to 300 to match default GCP value"
```
c2thorn, line 661: Terraform deprecations messages are highly disruptive for users and should be reserved for fields that will actually be removed. I quickly conferred with the team, and we don't believe we should use it here. Can we just add an [upgrade guide entry](https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/third_party/terraform/website/docs/guides/version_6_upgrade.html.markdown) + maybe mention in documentation description?

c2thorn (changes requested):

mmv1/products/compute/RegionBackendService.yaml
pawelJas: Understood. I have removed the deprecation_message and added info to docs about the changes coming in 6.0.0.

c2thorn (approved):

PR #10933, opened 2024-06-10
Wbi resize disk

bcreddy-gcp:

```release-note:enhancement
workbench: Added support for resizing disk
```


mmv1/products/workbench/Instance.yaml
```diff
@@ -270,12 +270,12 @@ properties:
         name: bootDisk
         default_from_api: true
         description: The definition of a boot disk.
-        immutable: true
+        update_url: 'projects/{{project}}/locations/{{location}}/instances/{{name}}/resizeDisk'
```
rileykarson, line 273: This is currently generating a null update- I think we need to apply this to each of the leaf fields we intend to update.

bcreddy-gcp:
https://github.com/GoogleCloudPlatform/magic-modules/pull/10972

mmv1/products/workbench/Instance.yaml
zli82016: It looks like the update_url only works for root properties. https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/templates/terraform/resource.erb#L821

Custom code is used to update disks in another [PR](https://github.com/GoogleCloudPlatform/magic-modules/pull/10972/files), as updating disks needs a different method `ResizeDisk`.  @rileykarson , does custom code sound reasonable to handle this case?

zli82016:
It looks like the custom code works to resize disks in https://github.com/GoogleCloudPlatform/magic-modules/pull/10972. Closing this PR.

PR #10934, opened 2024-06-10
workbench instances support for static external IP

bcreddy-gcp:

```release-note:enhancement
workbench: added `access_configs` to `google_workbench_instance` resource
```


mmv1/products/workbench/Instance.yaml
```diff
@@ -385,6 +385,26 @@ properties:
...
+              description: |
+                Optional. An array of configurations for this interface. Currently, only one access
+                config, ONE_TO_ONE_NAT, is supported. If no accessConfigs specified, the
+                instance will have an external internet access through an ephemeral
+                external IP address.
+              item_type: !ruby/object:Api::Type::NestedObject
+                properties:
+                  - !ruby/object:Api::Type::String
+                    name: externalIp
+                    default_from_api: true
```
shuyama1, line 400: Can we make this field as required, since this is the only field in the block, we'll try to prevent users from sending empty blocks if possible.

shuyama1 (commented):
Are you able to get the tests pass locally? The tests failed due to `The zone 'projects/xxxx/zones/us-central1-a' does not have enough resources available to fulfill the request`, but it's not clear which resource is hitting the quota limit.

mmv1/products/workbench/Instance.yaml
bcreddy-gcp: Cannot make this required. If specified, this will be a static ip. If not, an ephemeral ip will be assigned.

shuyama1: Does it mean that sending `access_configs {}` is different from not sending the block to the API?

bcreddy-gcp: It is not different. Made the suggested change.

mmv1/products/workbench/Instance.yaml
```diff
@@ -384,6 +384,27 @@ properties:
...
```
shuyama1, line 399:
```suggestion
```
`default_from_api: true` will translate to `optional` + `computed` in the provider, and given it's required, we can't and don't need to set this flag

mmv1/products/workbench/Instance.yaml
```diff
@@ -384,6 +384,27 @@ properties:
               values:
                 - VIRTIO_NET
                 - GVNIC
+            - !ruby/object:Api::Type::Array
+              name: accessConfigs
```
shuyama1, line 388: we may need to add `default_from_api: true` to the parent level field though. It seems that the API will return values for this block even if it's explicitly specified and it will cause a permadiff in Terraform. Adding `default_from_api: true` should prevent it from happening. I think the current test if failing due to this issue

shuyama1 (changes requested):

shuyama1 (commented):
Looks like the test is failing due to a permadiff that access_configs is explicitly specified but the API does not return it back.

bcreddy-gcp:
Looks like the API rollout is not complete. I'll let you know once it is and you can look at rerunning the tests.

shuyama1 (commented):
Thanks @bcreddy-gcp for the update! Please re-request my review when the rollout is fully complete and I'll re-run the tests for the PR. Thanks :)

shuyama1 (commented):
still waiting on rollout to complete

shuyama1 (commented):
@bcreddy-gcp is the rollout finished? Also please let me know if you need help debug the test failure. Thanks

bcreddy-gcp:
Rollout is finished. Let me rerun the tests.

shuyama1 (approved):
Thank you!

PR #9452, opened 2023-11-10
Add dynamicGroupMetadata in cloud_identity_group

sqin2019:
Add dynamicGroupMetadata in cloud_identity_group

cann't test the change as I don't have a test cloud identity account.

```release-note:enhancement
cloudidentity: added support for `dynamic_group_metadata` block to `google_cloud_identity_group` resource
```

PR #10936, opened 2024-06-10
`google_storage_bucket`: fix `custom_placement_config` values not normalized

BBBmau:
Resolves https://github.com/hashicorp/terraform-provider-google/issues/18035

Solution was to use `StateFunc` which converts the values in `data_locations` to be set as uppercase before storing into state. However as stated in the linked issue, there's a known problem with SDK that when using `StateFunc` in a nested attribute such as `type.Set` it will add the new value instead of replacing.

But by ignoring values in `data_locations` when expanding that *aren't* all capitalized, we can ensure that the final state will only include what we specified in `StateFunc`

```go
for _, raw := range l {
		if raw.(string) == strings.ToUpper(raw.(string)) {
			req = append(req, raw.(string))
		}
	}
}
```	

test output:
`TestAccStorageBucket_dualLocation_lowercase` - newly added test
```bash
(base) ┌─(~/Dev/terraform-provider-google)───────────────────────────────────────(mau@mau-JKDT676NCP:s008)─┐
└─(14:40:36 on main ✹)──> envchain GCLOUD make testacc TEST=./google/services/storage TESTARGS='-run=TestAccStorageBucket_dualLocation_lowercase'                                                    ──(Mon,Jun10)─┘
TF_ACC=1 TF_SCHEMA_PANIC_ON_ERROR=1 go test ./google/services/storage -v -run=TestAccStorageBucket_dualLocation_lowercase -timeout 240m -ldflags="-X=github.com/hashicorp/terraform-provider-google/version.ProviderVersion=acc"
=== RUN   TestAccStorageBucket_dualLocation_lowercase
=== PAUSE TestAccStorageBucket_dualLocation_lowercase
=== CONT  TestAccStorageBucket_dualLocation_lowercase
--- PASS: TestAccStorageBucket_dualLocation_lowercase (20.90s)
PASS
ok      github.com/hashicorp/terraform-provider-google/google/services/storage  22.114s
```
`TestAccStorageBucket_dualLocation_versionChange` - newly added test to ensure that behavior for user doesn't change:
```bash
(base) ┌─(~/Dev/terraform-provider-google)───────────────────────────────────────(mau@mau-JKDT676NCP:s008)─┐
└─(12:10:48 on main ✹)──> envchain GCLOUD make testacc TEST=./google/services/storage TESTARGS='-run=TestAccStorageBucket_dualLocation_versionChange'
TF_ACC=1 TF_SCHEMA_PANIC_ON_ERROR=1 go test ./google/services/storage -v -run=TestAccStorageBucket_dualLocation_versionChange -timeout 240m -ldflags="-X=github.com/hashicorp/terraform-provider-google/version.ProviderVersion=acc"
=== RUN   TestAccStorageBucket_dualLocation_versionChange
--- PASS: TestAccStorageBucket_dualLocation_versionChange (48.35s)
PASS
```
`TestAccStorageBucket_dualLocation`:
```bash
(base) ┌─(~/Dev/terraform-provider-google)──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────(mau@mau-JKDT676NCP:s008)─┐
└─(14:48:58 on main ✹)──> envchain GCLOUD make testacc TEST=./google/services/storage TESTARGS='-run=TestAccStorageBucket_dualLocation'                                                              ──(Mon,Jun10)─┘
TF_ACC=1 TF_SCHEMA_PANIC_ON_ERROR=1 go test ./google/services/storage -v -run=TestAccStorageBucket_dualLocation -timeout 240m -ldflags="-X=github.com/hashicorp/terraform-provider-google/version.ProviderVersion=acc"
=== RUN   TestAccStorageBucket_dualLocation
=== PAUSE TestAccStorageBucket_dualLocation
=== RUN   TestAccStorageBucket_dualLocation_lowercase
=== PAUSE TestAccStorageBucket_dualLocation_lowercase
=== RUN   TestAccStorageBucket_dualLocation_rpo
=== PAUSE TestAccStorageBucket_dualLocation_rpo
=== CONT  TestAccStorageBucket_dualLocation
=== CONT  TestAccStorageBucket_dualLocation_rpo
=== CONT  TestAccStorageBucket_dualLocation_lowercase
--- PASS: TestAccStorageBucket_dualLocation_lowercase (21.61s)
--- PASS: TestAccStorageBucket_dualLocation (23.27s)
--- PASS: TestAccStorageBucket_dualLocation_rpo (49.14s)
PASS
ok      github.com/hashicorp/terraform-provider-google/google/services/storage  50.278s
```

```release-note:bug
storage: fixed `custom_placement_config` values not normalized in `google_storage_bucket`
```

### DIFF
--- a/mmv1/third_party/terraform/services/storage/resource_storage_bucket.go.erb
+++ b/mmv1/third_party/terraform/services/storage/resource_storage_bucket.go.erb
@@ -1172,6 +1172,10 @@ func flattenBucketCustomPlacementConfig(cfc *storage.BucketCustomPlacementConfig
 func expandBucketDataLocations(configured interface{}) []string {
 	l := configured.(*schema.Set).List()
 
+	// Since we only want uppercase values to prevent unnecessary diffs, we can do a comparison
+	// to determine whether or not to include the value as part of the request.
+
+	// This extra check comes from the limitations of both DiffStateFunc and StateFunc towards types of Sets,Lists, and Maps.
 	req := make([]string, 0, len(l))
 	for _, raw := range l {
 		if raw.(string) == strings.ToUpper(raw.(string)) {
--- a/mmv1/third_party/terraform/services/storage/resource_storage_bucket.go.erb
+++ b/mmv1/third_party/terraform/services/storage/resource_storage_bucket.go.erb
@@ -496,6 +496,9 @@ func ResourceStorageBucket() *schema.Resource {
 							MinItems: 2,
 							Elem: &schema.Schema{
 								Type: schema.TypeString,
+								StateFunc: func(s interface{}) string {
+									return strings.ToUpper(s.(string))
+								},
 							},
 							Description: `The list of individual regions that comprise a dual-region bucket. See the docs for a list of acceptable regions. Note: If any of the data_locations changes, it will recreate the bucket.`,
 						},
@@ -1171,7 +1174,9 @@ func expandBucketDataLocations(configured interface{}) []string {
 
 	req := make([]string, 0, len(l))
 	for _, raw := range l {
-		req = append(req, raw.(string))
+		if raw.(string) == strings.ToUpper(raw.(string)) {
+			req = append(req, raw.(string))
+		}
 	}
 	return req
 }
--- a/mmv1/third_party/terraform/services/storage/resource_storage_bucket_test.go.erb
+++ b/mmv1/third_party/terraform/services/storage/resource_storage_bucket_test.go.erb
@@ -272,29 +272,6 @@ func TestAccStorageBucket_dualLocation_lowercase(t *testing.T) {
 	})
 }
 
-func TestAccStorageBucket_dualLocation_lowercase(t *testing.T) {
-	t.Parallel()
-
-	bucketName := acctest.TestBucketName(t)
-
-	acctest.VcrTest(t, resource.TestCase{
-		PreCheck:                 func() { acctest.AccTestPreCheck(t) },
-		ProtoV5ProviderFactories: acctest.ProtoV5ProviderFactories(t),
-		CheckDestroy:             testAccStorageBucketDestroyProducer(t),
-		Steps: []resource.TestStep{
-			{
-				Config: testAccStorageBucket_dualLocation_lowercase(bucketName),
-			},
-			{
-				ResourceName:            "google_storage_bucket.bucket",
-				ImportState:             true,
-				ImportStateVerify:       true,
-				ImportStateVerifyIgnore: []string{"force_destroy"},
-			},
-		},
-	})
-}
-
 func TestAccStorageBucket_dualLocation_versionChange(t *testing.T) {
 	// Test is not parallel because ENVs are set.
 	// Need to skip VCR as this test downloads providers from the Terraform Registry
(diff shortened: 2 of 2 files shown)


SarahFrench (commented):
Could you avoid StateFunc and its limitations by making the flattener function downcase all strings before they're stored in state?

BBBmau:
> Could you avoid StateFunc and its limitations by making the flattener function downcase all strings before they're stored in state?

this would work for when the user inputs lowercase values but if users set as all caps they'll get diffs since state is set to `ASIA_SOUTHEAST1` by user but downcasing the flattener returns us `asia_southeast1`

BBBmau:
@roaks3 hi! let me know if there's anything else needed for this review 👍🏼

roaks3 (commented):
For the most part, this seems reasonable, and looks like it's working to me. But it is a bit unusual, so I'd like to understand the solution a bit better if possible:
1. I'm not quite understanding the trick to filter for uppercased values in the expander. Could you explain how it is working? It seems to handle the migration between provider versions well, but are there other types of changes to the config that might cause an issue?
2. Probably naive, but I can't help but wonder why a diff suppress doesn't work instead. It seems like the API accepts lower or uppercase, and returns uppercase?

BBBmau:
@roaks3 

> 1. I'm not quite understanding the trick to filter for uppercased values in the expander. Could you explain how it is working? It seems to handle the migration between provider versions well, but are there other types of changes to the config that might cause an issue?

Normally `StateFunc` would just replace the value but when using it with lists and sets it appends to it rather than replacing, causing `data_locations = ["asia-east1", "asia-southeast1"]` to be `["asia-east1", "asia-southeast1", "ASIA-EAST1", "ASIA-SOUTHEAST1"]` after `StateFunc`

The added logic filters the list that `StateFunc` produces in expand to only keep Upper strings. we can do this by comparing a before and after of applying `.ToUpper` on each string.
```
func expandBucketDataLocations(configured interface{}) []string {
	l := configured.(*schema.Set).List()

	req := make([]string, 0, len(l))
	for _, raw := range l {
		if raw.(string) == strings.ToUpper(raw.(string)) {
			req = append(req, raw.(string))
		}
	}
	return req
}
```

For the most part I wasn't able to come up with other configs that could potentially cause issues with this. It handles lowercase and uppercase list configs well. Let me know if there's still something in my explanation that wasn't clear with this approach. I agree that the solution is unusual and comes from needing to workaround the limitations of `StateFunc` and `DiffSuppressFunc` with `type.Set`.

> 2. Probably naive, but I can't help but wonder why a diff suppress doesn't work instead. It seems like the API accepts lower or uppercase, and returns uppercase?

`DiffSuppressFunc` isn't supported on types `Maps,Sets, and Lists`, I did see that a workaround exists for map that was used here: https://github.com/GoogleCloudPlatform/magic-modules/pull/9452

Look into implementing a workaround for using it on Sets but not coming across a solution yet, I can spend more time it next week but that's the main reason why `DiffSuppressFunc` isn't being implemented, related discussion on `DiffSuppresFunc` on sets: https://github.com/hashicorp/terraform-plugin-sdk/issues/477

roaks3 (approved):
Ah ok, thanks for explaining, that all makes sense. I personally don't think it requires more investigation into using `DiffSuppressFunc`, but it might be worth adding a comment or two so the reader of the code knows why this works.

PR #10937, opened 2024-06-10
Fixed review requests on issue comment from author

melinath:
According to https://docs.github.com/en/webhooks/webhook-events-and-payloads#issue_comment the data for comment creation should be in issue, not pull_request

Note that pull request comments (on specific lines of files) are separate and are currently intentionally not included here.

~This should be possible to test by having the reviewer leave a review, then having me leave a comment - I believe the issue comment portion of the logic should run based off of what is in this PR.~ This all runs off the main branch so it won't be possible to test before merge. However, I think the logic should be more clear now so I am more confident in it.

```release-note:none

```

### DIFF
--- a/.github/workflows/request-reviewer.yml
+++ b/.github/workflows/request-reviewer.yml
@@ -15,14 +15,13 @@ on:
 
 jobs:
   request-review:
-    # if: github.event.pull_request && github.event.pull_request.draft == false && (github.event.comment.user.login == github.event.issue.user.login || github.event.action != 'created')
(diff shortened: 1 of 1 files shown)


melinath:
leaving a comment

melinath:
leaving comment

melinath:
another comment

melinath:
🤦  nope, this also runs off the default branch.

melinath:
Based on https://stackoverflow.com/questions/59077079/how-to-get-pull-request-number-within-github-actions-workflow it looks like the `created` event will not set `pull_request` at all. Which does technically match the [documented payload](https://docs.github.com/en/webhooks/webhook-events-and-payloads#issue_comment); I think I just thought they might do something clever. (But this is probably for the best!)

roaks3 (approved):
LGTM

PR #10939, opened 2024-06-10
[Prep for 6.0.0 release branch] Make downstream waiter compatible with non-linear history for main

shuyama1:

```release-note:none

```

### DIFF
--- a/.ci/magician/cmd/wait_for_commit.go
+++ b/.ci/magician/cmd/wait_for_commit.go
@@ -52,39 +52,23 @@ func execWaitForCommit(syncBranchPrefix, baseBranch, sha string, runner source.R
 	}
 
 	for {
-		if baseBranch != "main" {
-			output, err := gitRevParse("origin/"+syncBranch, runner)
-			if err != nil {
-				return err
-			}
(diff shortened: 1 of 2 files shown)


c2thorn (approved):

melinath:
@shuyama1 could you explain why this was necessary so that we have it for the record?

shuyama1:
> @shuyama1 could you explain why this was necessary so that we have it for the record?

Sure. This change is needed for the final merge of the release branch -> main. Our downstream waiter was previously built based on the assumption that we're enforcing squash-merges-only. This is mostly true, especially for merging into the main branch. The only one exception is when we merge the release branch back to the main, we'll need commit merge to keep all the commit history. 

The previous `(git log --pretty=%H origin/$SYNC_BRANCH..origin/$BASE_BRANCH | tail -n 1)` returns the hash of the first commit that is present in origin/$BASE_BRANCH but not in origin/$SYNC_BRANCH. This will not work for commit merge, as it will give the first commit where the two branches diverged, which is the first commit merged into the release branch.

melinath:
Thanks!

PR #10941, opened 2024-06-11
Reararngement of google_composer_environment resource documentation

spapi17:

```release-note:none

```


BBBmau (approved):
approving after coming across this which confirms that these changes are valid:
https://cloud.google.com/composer/docs/composer-2/disaster-recovery-with-snapshots

PR #10942, opened 2024-06-11
Support absolute override dir for generator

roaks3:
Currently, absolute paths can't be used for the `override_dir` (ie. `make mmv1 OVERRIDES=...` or `-r`), because we concatenate paths in a few places in our templates. Instead, it must be relative to the magic-modules repo directory.

Example error when providing absolute path:
`/Users/ryanoaks/projects/magic-modules-private-overrides/artifacts/magic-modules/mmv1/compile/core.rb:141:in 'read': No such file or directory @ rb_sysopen - /Users/ryanoaks/projects/magic-modules-private-overrides/artifacts/magic-modules/mmv1//Users/ryanoaks/projects/magic-modules-private-overrides/examples/backend_service_ip_address_selection_policy.tf.erb (Errno::ENOENT)`

This PR allows absolute paths by immediately checking if the path is absolute, and if it is, changing it to be relative to magic-modules.

Testing: I've confirmed locally that with this change, relative override paths still work, and absolute paths now work as well.

```release-note:none

```

### DIFF
--- a/mmv1/compiler.rb
+++ b/mmv1/compiler.rb
@@ -119,6 +119,14 @@ end
 
 if override_dir
   Google::LOGGER.info "Using override directory '#{override_dir}'"
+
+  # Normalize override dir to a path that is relative to the magic-modules directory
+  # This is needed for templates that concatenate pwd + override dir + path
+  if Pathname.new(override_dir).absolute?
+    override_dir = Pathname.new(override_dir).relative_path_from(__dir__).to_s
+    Google::LOGGER.info "Override directory normalized to relative path '#{override_dir}'"
+  end
+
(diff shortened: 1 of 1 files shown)


roaks3:
@c2thorn I didn't notice a Go variant of this file, so wanted to check if I needed to make this change in any other places.

c2thorn (approved):
This doesn't immediately affect the current EAP build right?

roaks3:
Sorry, had to merge first because I didn't want to clear the review.

Correct, this doesn't affect EAP right now. However, it supports us changing EAP to use absolute paths, which makes it quite a bit easier to run the build in different environments.

PR #10946, opened 2024-06-11
Adding support for regional flex zones in netapp volume beta provider

roop2:
Adding support for regional flex zones in netapp volume beta provider

```release-note:enhancement
netapp: added `zone`, `replica_zone` field to `google_netapp_storage_pool` resource
```


mmv1/templates/terraform/pre_update/netapp_storagepool.go.erb
```diff
@@ -0,0 +1,65 @@
...
+					UserAgent: userAgent,
+					Timeout:   d.Timeout(schema.TimeoutUpdate),
+				})
+				if err != nil {
+					return fmt.Errorf("Error switching active zone for pool: %s, %v", d.Id(), err)
+				}
+
+				err = NetappOperationWaitTime(
+					config, reso, project, "Switching active pool zone", userAgent,
+					d.Timeout(schema.TimeoutDelete))
```
okrause, line 32: @roop2 please change it to d.Timeout(schema.TimeoutUpdate))

roop2: done

mmv1/products/netapp/storagePool.yaml
```diff
@@ -25,9 +25,21 @@ description: |
  The capacity of the pool can be split up and assigned to volumes within the pool. Storage pools are a billable
  component of NetApp Volumes. Billing is based on the location, service level, and capacity allocated to a pool
  independent of consumption at the volume level.
+
+ Storage pools of service level Flex are available as zonal (single zone) or regional (two zones in same region) pools.
```
roaks3, line 29: Nit: we may want to note some or all of this is beta-only

mmv1/products/netapp/storagePool.yaml
```diff
@@ -25,9 +25,21 @@ description: |
...
+ Zonal and regional pools are high-available within the zone. On top of that, regional pools have `replica_zone` as
+ hot standby zone. All volume access is served from the `zone`. If `zone` fails, `replica_zone`
+ automatically becomes the active zone. This will cause state drift in your configuration.
```
roaks3, line 32: I assume this is the best option at this point, but I'll still ask: is there any chance of this being modeled differently in the API, so that there is a separate "desired" zone and "actual" zone? This is what we usually recommend to prevent drift scenarios like this.

mmv1/templates/terraform/pre_update/netapp_storagepool.go.erb
```diff
@@ -0,0 +1,65 @@
...
+				url, err = tpgresource.ReplaceVars(d, config, "{{NetappBasePath}}projects/{{project}}/locations/{{location}}/storagePools/{{name}}")
+				if err != nil {
+					return err
+				}
+				url, err = transport_tpg.AddQueryParams(url, map[string]string{"updateMask": strings.Join(updateMask, ",")})
+				if err != nil {
+					return err
+				}
+			} else {
+				return fmt.Errorf("Incorrect zone change for pool: %s. Supported Zones are : %s, %s", d.Id(), oldZone, oldReplicaZone)
```
roaks3, line 61: Nit: will this be clear enough to a user? I'm wondering if we may want to mention the fact that `zone` and `replica_zone` are the fields involved, and they can only be swapped.

mmv1/third_party/terraform/services/netapp/resource_netapp_flex_storage_pool_test.go.erb
```diff
@@ -0,0 +1,200 @@
+<% autogen_exception -%>
```
roaks3, line 1: A few comments regarding this test file:
1. Nit: would you mind formatting the file? The indentation is a bit hard to read.
2. We would normally just include these tests in `resource_netapp_storage_pool_test.go.erb`, with the version guard around them. Is there a reason this wouldn't work?
3. I would suggest changing the strategy slightly. First, it might be good to have an example of the new fields being used in the documentation, which would most likely look like another (minimal) example file. Also, these tests shouldn't need to cover updating fields that are covered in `resource_netapp_storage_pool_test.go.erb`, so I would suggest focusing on minimal configs that test the switching behavior specifically.

mmv1/third_party/terraform/services/netapp/resource_netapp_flex_volume_test.go.erb
```diff
@@ -0,0 +1,104 @@
+<% autogen_exception -%>
```
roaks3, line 1: I suspect this file is largely redundant with `resource_netapp_volume_test.go`, so I would recommend dropping it, unless there is novel update behavior with FLEX level. Alternatively, you may want to consider a minimal example to show FLEX usage (which adds the documentation and test).

roaks3 (commented):
Mostly LGTM, but had some question/suggestions and nits

mmv1/products/netapp/storagePool.yaml
roop2: done, added a note

mmv1/third_party/terraform/services/netapp/resource_netapp_flex_volume_test.go.erb
roop2: removed the file

mmv1/third_party/terraform/services/netapp/resource_netapp_flex_storage_pool_test.go.erb
roop2: I have added the tests to a single file `resource_netapp_storage_pool_test.go.erb` which contains both beta and ga provider tests as it is done in other services and gated the tests with the macro to run in beta provider tests.

mmv1/templates/terraform/pre_update/netapp_storagepool.go.erb
roop2: Updated the error message.

mmv1/third_party/terraform/services/netapp/resource_netapp_storage_pool_test.go.erb
```diff
@@ -0,0 +1,254 @@
...
+data "google_compute_network" "default" {
+    provider = google-beta
+    name = "%{network_name}"
+}
+`, context)
+}
+
+func testAccNetappstoragePool_FlexRegionalStoragePoolCreateExample_switchBackZone(context map[string]interface{}) string {
+	// wait 5 minutes before executing the switchback due to api zone switch issues
+	time.Sleep(5 * time.Minute)
```
roaks3, line 229: I think this sleep is not run when we are hoping: it probably waits to start the entire test, and then still executes the step immediately. The `time_sleep` you have is more in the direction of what we will need, but I can't tell, does it do what you need? If not, we may need to tweak that part to ensure a sleep occurs before continuing with the update.

mmv1/third_party/terraform/services/netapp/resource_netapp_storage_pool_test.go.erb
```diff
@@ -0,0 +1,254 @@
...
+				ImportStateVerify:       true,
+				ImportStateVerifyIgnore: []string{"location", "name", "labels", "terraform_labels"},
+			},
+		},
+	})
+}
+
+func testAccNetappstoragePool_FlexRegionalStoragePoolCreateExample_full(context map[string]interface{}) string {
+	return acctest.Nprintf(`
+resource "google_netapp_storage_pool" "test_pool" {
```
roaks3, line 179: Just to confirm, per my prior comment, do we want to include an example in the documentation for FLEX?

roaks3 (changes requested):
LGTM, we just need to do something about the sleep in the test

mmv1/third_party/terraform/services/netapp/resource_netapp_storage_pool_test.go.erb
roop2: Yes this is correct, we do not want to add this to documentation examples, we only want to keep our most basic storage pool example there.

mmv1/third_party/terraform/services/netapp/resource_netapp_storage_pool_test.go.erb
roaks3: Ok, understood. What is confusing though is that this `time.Sleep` fixed your tests. This sleep will run when the configuration is set up, but not when the tests actually run. You can see in the [debug log](https://storage.cloud.google.com/ci-vcr-logs/beta/refs/heads/auto-pr-10946/artifacts/ec2f2cd2-3117-4cf5-aac8-19b5f773189f/recording/TestAccNetappstoragePool_FlexRegionalStoragePoolCreateExample_update.log) that there is no observable 5 min pause until 2024-06-27T11:52:01, when the deletion occurs.

Instead, I think you probably want one or two additional `"time_sleep"` resources in your config, to put artificial delays after creation (and perhaps after the first switch as well).

roop2: Thanks @roaks3 for clarifying this, I see that the sleep at the location is not having the intended effect as you mentioned.
I have moved it to a check function which gets executed after the first switch is done and verified in the logs that the wait is happening now.
I am not sure if doing it via a sleep resource would work as intended (between the 2 switch calls) as that might depend on terraform implicit create destroy order. So I am going with adding this to the check call which gets executed post first switch.

roaks3 (approved):

PR #10950, opened 2024-06-11
Fix firewall rule to support empty descritpion on update

r-teller:
bug: Fix firewall rule updates to support empty descriptions

```release-note:REPLACEME

```

### DIFF
--- a/mmv1/products/compute/Firewall.yaml
+++ b/mmv1/products/compute/Firewall.yaml
@@ -167,6 +167,7 @@ properties:
     description: |
       An optional description of this resource. Provide this property when
       you create the resource.
+    send_empty_value: true
   - !ruby/object:Api::Type::Array
     name: 'destinationRanges'
     description: |


slevenick (commented):
Can you add a test that would specifically update this field to empty and check that it works?

r-teller:
I have modify TestAccComputeFirewall_update to verify that description can be set to null/empty

slevenick:
Can you explain why setting the description to empty is desirable? I get that it's technically correct, but it may introduce unexpected behavior as most of our resources with descriptions don't behave like this

Are you running into a specific issue around the description field?

r-teller:
I noticed that for most resources within GCP, the description is an immutable field, but for firewall rules, it is not. Through the console or gcloud command, you can set this to empty. However, when using Terraform, it provides a different experience.

In my organization, I observed that you could create a firewall rule without a description, but once a rule has a description, you couldn't set it back to empty. The goal for this merge was to align Terraform functionality with the expected experience provided through other interactions

slevenick (changes requested):
This is causing a failure in TGC (which you don't need to know about) but I think the way to fix it is to add a description to this file: https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/third_party/tgc/tests/data/firewall.json

Can you try that and we'll see if the tests pass? I'm not entirely sure how to fix it, so I'm kind of guessing here

r-teller:
I believe i have updated all firewall rules that would have failed based on missing descriptions

slevenick (approved):

PR #10955, opened 2024-06-12
go rewrite custom update

NickElliot:
Verified that this is generating correctly post format via google_dns_policy -- this resource itself fails to format due to issues with the nested schema template, but the diffs are otherwise non existent: https://diff.googleplex.com/#key=PVL1tJM8BY6j

With key exception of a DNS wait time warning, however the logic is itself correct, so data is being mismanaged somewhere.

```release-note:none

```

### DIFF
--- a/mmv1/api/resource.go
+++ b/mmv1/api/resource.go
@@ -1372,3 +1372,70 @@ func (r Resource) FirstIdentityProp() *Type {
 
 	return idProps[0]
 }
+
+type UpdateGroup struct {
+	UpdateUrl string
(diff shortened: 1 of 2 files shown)


NickElliot:
cc @zli82016

c2thorn (approved):
I did not realize Custom Update was this involved. Looks good for now. I am interested in the current diffs from DNS, but probably won't be worth looking more into it until we actually get to that product

c2thorn:
I think you need an empty changelog with `none`

PR #10959, opened 2024-06-13
Fix flaky DiscoveryConfig.yaml tests

patrickmoy:
Fix flakiness in DiscoveryConfig.yaml tests; errors and timestamp fields should not be verified in update tests, as these are dynamic and will change across the lifespan of the resource as DLP performs scans.

```release-note:none
```


SarahFrench (approved):
Thanks for addressing this!

PR #10970, opened 2024-06-14
Remove redundant `resource.TestStep`

zli82016:
Remove redundant `resource.TestStep`, which is causing the diffs when converting the handwritten files, as the Go code uses [format library](https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/provider/template_data.go#L191), and the ruby code uses [format command](https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/provider/file_template.rb#L87).

```release-note:none

```


zli82016:
It looks like there are more changes.

c2thorn (approved):

PR #10974, opened 2024-06-17
add update support to big_query field on google_vertex_ai_feature_group resource

kkram01:
Add update_mask_fields to bigquery nestedfield on feature_group resource 

```release-note:enhancement
vertexai: added update support for `big_query.entity_id_columns` field on `google_vertex_ai_feature_group` resource
```

### DIFF
--- a/mmv1/products/vertexai/FeatureGroup.yaml
+++ b/mmv1/products/vertexai/FeatureGroup.yaml
@@ -90,6 +90,8 @@ properties:
   - !ruby/object:Api::Type::NestedObject
     name: bigQuery
     description: Indicates that features for this group come from BigQuery Table/View. By default treats the source as a sparse time series source, which is required to have an entityId and a feature_timestamp column in the source.
+    update_mask_fields:
+      - 'bigQuery.entityIdColumns'
     properties:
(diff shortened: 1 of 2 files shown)


mmv1/products/vertexai/FeatureGroup.yaml
```diff
@@ -90,6 +90,8 @@ properties:
   - !ruby/object:Api::Type::NestedObject
     name: bigQuery
     description: Indicates that features for this group come from BigQuery Table/View. By default treats the source as a sparse time series source, which is required to have an entityId and a feature_timestamp column in the source.
+    update_mask_fields:
+      - 'bigQuery.entityIdColumns'
```
slevenick, line 94: Do we have a test that updated this field before? If so, why was it not failing? If not, can you add one?

kkram01: hi @slevenick I've included the update to this field now in the update test. Pls check. I think previously updation of this field wasn't supported

slevenick (approved):
LGTM thanks! I updated the changelog note to match our format

PR #10975, opened 2024-06-17
go rewrite - sort context map in test file

c2thorn:
golang maps are sorted alphabetically, so let's make the ruby version match

```release-note:none

```


mmv1/templates/terraform/env_var_context.go.erb
```diff
@@ -1,4 +1,5 @@
-<% example.test_env_vars&.each do |var_name, var_type| -%>
+<% unless example.test_env_vars.nil? -%>
```
c2thorn, line 1: this replaces the `test_env_vars&` from before. Couldn't figure out how to add `.sort` while still including the `&`. Let me know if you know a better way

rileykarson (approved):

PR #10977, opened 2024-06-17
add support to compute.googleapis.com/SslCertificate for TGC

dvfons:

```release-note:none

```


rileykarson (changes requested):
This adds new fields to the provider and makes no TGC-specific changes. Please follow the contribution guide for the provider if you'd like to add these fields.

PR #10979, opened 2024-06-17
add support for compute.googleapis.com/Subnetwork to TGC

dvfons:

```release-note:none

```

### DIFF
--- a/mmv1/products/compute/Subnetwork.yaml
+++ b/mmv1/products/compute/Subnetwork.yaml
@@ -408,3 +408,11 @@ properties:
       existing resources are dropped and prevented from leaving the VPC.
       Setting this field to true will allow these packets to match dynamic routes injected
       via BGP even if their destinations match existing subnet ranges.
+  - !ruby/object:Api::Type::Boolean
+    name: 'enableFlowLogs'
+    description: |
+      Whether to enable flow logging for this subnetwork.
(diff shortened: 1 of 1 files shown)


mmv1/products/compute/Subnetwork.yaml
```diff
@@ -408,3 +408,11 @@ properties:
       existing resources are dropped and prevented from leaving the VPC.
       Setting this field to true will allow these packets to match dynamic routes injected
       via BGP even if their destinations match existing subnet ranges.
+  - !ruby/object:Api::Type::Boolean
```
c2thorn, line 411: Hi, I see the title as "add support for compute.googleapis.com/Subnetwork to TGC", but this seems to only add one field to the resource which may already exist for TGC?

c2thorn (changes requested):
the tests fail specifically due to this field

c2thorn (changes requested):
I'm seeing the following diff created in the tests
`- enable_flow_logs                 = true -> null # forces replacement`

That makes me think that this field need to be marked as `default_from_api: true`.

Was this specifically intended for TGC or would you like to add it to the providers as well? I'm not sure what the intention is here

c2thorn:
If these are to be added to the providers, it needs tests https://googlecloudplatform.github.io/magic-modules/develop/test/test/

PR #10982, opened 2024-06-17
Bigtable: Adds ignore_warning to gc policy resource

andre-sampaio:
Adds `ignore_warning` to `bigtable_gc_policy`, which allows relaxing the gc policy for replicated tables. This is the same approach currently used for [bigtable_app_profiles](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/bigtable_app_profile#ignore_warnings), and is preferable to always setting this to `true` on the backend request due to possible undesired [side effects](https://cloud.google.com/bigtable/docs/garbage-collection#increasing) of doing so, either accidentally or due to being unaware of these side effects.

. [x] Not linking an issue since I think the description provides enough context
. [x] `ignore_warnings` not in [ga](https://github.com/hashicorp/terraform-provider-google/blob/main/google/services/bigtable/resource_bigtable_gc_policy.go) nor in [beta](https://github.com/hashicorp/terraform-provider-google-beta/blob/main/google-beta/services/bigtable/resource_bigtable_gc_policy.go)
. [x] Added test covering new field, all resources are prefixed by `tf-test`
. [x] Manually added documentation

```release-note:enhancement
bigtable: added `ignore_warnings` field to `bigtable_gc_policy` resource
```

### DIFF
--- a/mmv1/third_party/terraform/services/bigtable/resource_bigtable_gc_policy.go
+++ b/mmv1/third_party/terraform/services/bigtable/resource_bigtable_gc_policy.go
@@ -216,6 +216,17 @@ func ResourceBigtableGCPolicy() *schema.Resource {
 				in a replicated instance. Possible values are: "ABANDON".`,
 				ValidateFunc: validation.StringInSlice([]string{"ABANDON", ""}, false),
 			},
+
+			"ignore_warnings": {
+				Type:     schema.TypeBool,
+				Optional: true,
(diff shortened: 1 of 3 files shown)


andre-sampaio:
@ron-gal Please take a look, thanks!

hao-nan-li (commented):
Looks like there's some issue with formatting, probably something not properly indented or missing EOD line. Could you follow the error message and fix it?

hao-nan-li (approved):

PR #10986, opened 2024-06-17
Make firebase_hosting_site upsert on creation

rainshen49:
This will provide a pathway for users to associate a web app with an existing hosting site without running `terraform import` in between. This behavior is necessary because the default Hosting site is currently provisioned outside of Terraform, which causes trouble for many users as indicated in https://github.com/hashicorp/terraform-provider-google/issues/12955#issuecomment-2172442248

```release-note:enhancement
firebasehosting: allowed associating a web app with an existing Site.
```


rainshen49:
Please hold off review, let me figure out why the test passed locally but not here.

rainshen49:
Okay the tests are passing now. Pls review.

BBBmau (approved):

PR #10987, opened 2024-06-17
Use `id` in getting started guide

rileykarson:
This content was pretty stale- we use `id` now and `self_link` is no longer general

Confirmed the test config applies stably:

```
$ terraform apply
google_compute_network.vpc_network: Refreshing state... [id=projects/graphite-test-rileykarson/global/networks/terraform-network]
google_compute_instance.vm_instance: Refreshing state... [id=projects/graphite-test-rileykarson/zones/us-central1-a/instances/terraform-instance]

```

```release-note:none
```

### DIFF
--- a/mmv1/third_party/terraform/website/docs/guides/getting_started.html.markdown
+++ b/mmv1/third_party/terraform/website/docs/guides/getting_started.html.markdown
@@ -100,23 +100,23 @@ are used as a short way to identify resources, and a resource's display name in
 the Cloud Console will be the one defined in the `name` field.
 
 When linking resources in a Terraform config though, you'll primarily want to
-use a different field, the `self_link` of a resource. Like `name`, nearly every
-resource has a `self_link`. They look like:
+use a different field, the `id` of a resource. Every Terraform resource has an
+`id`. In the Google provider they look like:
 
 ```
-{{API base url}}/projects/{{your project}}/{{location type}}/{{location}}/{{resource type}}/{{name}}
+projects/{{your project}}/{{location type}}/{{location}}/{{resource type}}/{{name}}
 ```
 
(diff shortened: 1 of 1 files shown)


mmv1/third_party/terraform/website/docs/guides/getting_started.html.markdown
```diff
@@ -100,23 +100,23 @@ are used as a short way to identify resources, and a resource's display name in
 the Cloud Console will be the one defined in the `name` field.
 
 When linking resources in a Terraform config though, you'll primarily want to
-use a different field, the `self_link` of a resource. Like `name`, nearly every
-resource has a `self_link`. They look like:
+use a different field, the `id` of a resource. Every Terraform resource has an
+`id`. In the Google provider they look like:
```
rileykarson, line 104:
```suggestion
`id`. In the Google provider they generally look like:
```

SarahFrench (approved):
Approving, though something to bear in mind in future is that the plugin-framework doesn't use an `id` field the same way as the SDK. 

As a quick summary:
* SDK adds an implicit `id` field to resource/data source schema if it's not explicitly included in schema. This is because `id` is special and used to control how things are added to/managed in/removed from state.
* The plugin-framework doesn't require `id` in this way, so it doesn't add an implicit `id` field to schemas like the SDK does.

Happy to discuss this further outside this PR. tl;dr given this PR's changes it looks like we might want to keep `id` fields intentionally in any future resources that are added to the provider in plugin-framework.

rileykarson:
Yeah, that's a decision we'll need to make as we roll out framework-native resources. This is true today / will be for conversions, so I kept it mirroring the current status!

PR #10993, opened 2024-06-18
add support to pubsub.googleapis.com/Topic for TGC

dvfons:

```release-note:none

```

### DIFF
--- a/mmv1/products/pubsub/Topic.yaml
+++ b/mmv1/products/pubsub/Topic.yaml
@@ -120,6 +120,14 @@ properties:
           and is not a valid configuration.
         item_type: Api::Type::String
         required: true
+      - !ruby/object:Api::Type::Boolean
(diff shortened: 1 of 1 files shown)


roaks3 (changes requested):
The code looks fine, but a few other things:
1. Is there an issue in GitHub that this is resolving?
2. Could we at least add a release note? I'm not sure how this is related to TGC, but it seems like the field would be usable in Terraform.
3. Could we add this field to one or more tests?

roaks3:
I'll need more info to help with that error, my guess is that the config is invalid. Could you share your TF config?

roaks3 (commented):
Clear review request

roaks3 (commented):
> I'll need more info to help with that error, my guess is that the config is invalid. Could you share your TF config?

PR #10994, opened 2024-06-18
Make sweeper for google_vmwareengine_network handwritten, fix how it accesses data from the list operation, add looping through locations

SarahFrench:
Fixes https://github.com/hashicorp/terraform-provider-google/issues/18471

The existing sweeper only sweeps in us-central1 (the value provided to the sweeper in TeamCity by default)/can only sweep a single location at once.

This PR:
* updates the sweeper to sweep a list of predetermined locations, in addition to the location passed into the sweeper.
* updates the sweeper to be able to use data returned from the API list method ([9214e48](https://github.com/GoogleCloudPlatform/magic-modules/pull/10994/commits/9214e4879d6f63312814948c5d4f9061cdc8acf0)) 
    *  Currently the sweeper cannot find the "networks" field in the response JSON and concludes there's nothing to delete. The linked commit should fix this.

```release-note:none

```

### DIFF
--- a/mmv1/third_party/terraform/services/vmwareengine/resource_vmwareengine_network_sweeper.go
+++ b/mmv1/third_party/terraform/services/vmwareengine/resource_vmwareengine_network_sweeper.go
@@ -73,7 +73,7 @@ func testSweepVmwareengineNetwork(region string) error {
 			return nil
 		}
 
-		resourceList, ok := res["networks"]
+		resourceList, ok := res["vmwareEngineNetworks"]
 		if !ok {
 			log.Printf("[INFO][SWEEPER_LOG] Nothing found in response.")
 			return nil
--- a/mmv1/third_party/terraform/services/vmwareengine/resource_vmwareengine_network_sweeper.go
+++ b/mmv1/third_party/terraform/services/vmwareengine/resource_vmwareengine_network_sweeper.go
@@ -40,7 +40,7 @@ func testSweepVmwareengineNetwork(region string) error {
 	//   * global location
 	//   * regions used for this resource type's acc tests in the past
 	//   * the 'region' passed to the sweeper
-	locations := []string{"global", "us-central1", "southamerica-west1", "me-west1", region}
+	locations := []string{region, "global", "southamerica-west1", "me-west1"}
 	for _, location := range locations {
 		log.Printf("[INFO][SWEEPER_LOG] Beginning the process of sweeping location '%s'.", location)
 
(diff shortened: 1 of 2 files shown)


SarahFrench:
Once the new generated code is present in the modular magician fork I'll test the sweeper manually in TeamCity

SarahFrench:
This appears to be working as intended, but we need to sweep away private clouds to let the delete action work: https://hashicorp.teamcity.com/buildConfiguration/TerraformProviders_GoogleCloud_GOOGLE_MMUPSTREAMTESTS_SERVICE_SWEEPER__MANUAL/169691?buildTab=log&focusLine=1576&logView=flowAware&linesState=1140

I'll open another PR tomorrow

mmv1/third_party/terraform/services/vmwareengine/resource_vmwareengine_network_sweeper.go
```diff
@@ -0,0 +1,131 @@
...
+	}
+
+	t := &testing.T{}
+	billingId := envvar.GetTestBillingAccountFromEnv(t)
+
+	// List of location values includes:
+	//   * global location
+	//   * regions used for this resource type's acc tests in the past
+	//   * the 'region' passed to the sweeper
+	locations := []string{"global", "us-central1", "southamerica-west1", "me-west1", region}
```
SarahFrench, line 43:
```suggestion
	locations := []string{region, "global", "southamerica-west1", "me-west1"}
```

Use user-provided region first, remove us-central1 as it's assumed that's the value of region most times

mmv1/third_party/terraform/services/vmwareengine/resource_vmwareengine_network_sweeper.go
```diff
@@ -0,0 +1,131 @@
...
+	}
+
+	t := &testing.T{}
+	billingId := envvar.GetTestBillingAccountFromEnv(t)
+
+	// List of location values includes:
+	//   * global location
+	//   * regions used for this resource type's acc tests in the past
+	//   * the 'region' passed to the sweeper
+	locations := []string{region, "global", "southamerica-west1", "me-west1"}
```
SarahFrench, line 43:
```suggestion
	locations := []string{region, "global", "southamerica-west1", "me-west1"}
	log.Printf("[INFO][SWEEPER_LOG] Sweeping will include these locations: %v.", locations)
```

melinath (approved):
LGTM - linking to https://github.com/hashicorp/terraform-provider-google/issues/17835 since we may want to revisit this once that work is done.

PR #10313, opened 2024-03-28
Fix permadiff on google_vpc_access_connector resource with breaking changes (min_instances, max_instances, min_throughput, max_throughput fields)

SarahFrench:
Fixes https://github.com/hashicorp/terraform-provider-google/issues/15278

When reviewing this PR please see:
- This comment for context about this PR : https://github.com/GoogleCloudPlatform/magic-modules/pull/10313#issuecomment-2159065620
- This comment about the decision to make the two fields conflict in the schema: https://github.com/GoogleCloudPlatform/magic-modules/pull/10313#issuecomment-2159101760
- Note to the reviewer when I marked this ready for review: https://github.com/GoogleCloudPlatform/magic-modules/pull/10313#issuecomment-2161335142

**Release Note Template for Downstream PRs (will be copied)**

```release-note:breaking-change
vpcaccess: removed default values for `min_throughput` and `min_instances` fields on `google_vpc_access_connector` and made them default to values returned from the API when not provided by users
```

```release-note:breaking-change
vpcaccess: added a conflicting fields restriction between `min_throughput` and `min_instances` fields on `google_vpc_access_connector`
```

```release-note:breaking-change
vpcaccess: added a conflicting fields restriction between `max_throughput` and `max_instances` fields on `google_vpc_access_connector`
```

PR #10995, opened 2024-06-18
Make `google_vpc_access_connector` acc tests use a bootstrapped VPC network

SarahFrench:
This is to address some CIDR related test failures: https://github.com/GoogleCloudPlatform/magic-modules/pull/10313#issuecomment-2175660015

```release-note:none

```

### DIFF
--- a/mmv1/products/vpcaccess/Connector.yaml
+++ b/mmv1/products/vpcaccess/Connector.yaml
@@ -46,11 +46,17 @@ examples:
     primary_resource_id: 'connector'
     vars:
       name: 'vpc-con'
+      network_name: 'default'
+    test_vars_overrides:
+      network_name: 'acctest.BootstrapSharedServiceNetworkingConnection(t, "vpc-access-connector")'
   - !ruby/object:Provider::Terraform::Examples
     name: 'vpc_access_connector_shared_vpc'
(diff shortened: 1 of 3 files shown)


mmv1/products/vpcaccess/Connector.yaml
```diff
@@ -46,11 +46,17 @@ examples:
     primary_resource_id: 'connector'
     vars:
       name: 'vpc-con'
+      network_name: 'default'
+    test_vars_overrides:
+      network_name: 'acctest.BootstrapSharedServiceNetworkingConnection(t, "vpc-access-connector")'
```
SarahFrench, line 51: Or, I could name this `"vpc-access"` and if any other vpcaccess acceptance tests need to use a special VPC they should use the same, service-wide bootstrapped VPC? I'm not sure what the intended usage of BootstrapSharedServiceNetworkingConnection is.

roaks3: I think what you have is probably fine. Unless things have changed significantly, anything using bootstrapped networks generally needs to use a per-test id.

roaks3 (approved):

SarahFrench:
Thanks!

PR #10996, opened 2024-06-18
Go rewrite handwritten templates script

zli82016:
1. Improve custom templates script to handle the cases for handwritten erb files
2. Add `go run . --handwritten` to convert handwritten erb files
3. Convert service handwritten `.erb` files to `.go.tmpl` or `.go` files and stored in `services/xxxx/go` folder. The original `.go` files are still in `services/xxxx` folder

Will submit another PR for the rest service handwritten files.

```release-note:none

```


mmv1/template-converter.go
```diff
@@ -60,252 +60,385 @@ func convertTemplate(folder string) int {
 			log.Fatalf("Cannot open the file: %v", file)
 		}
 
-		// Replace {{}}
-		r, err := regexp.Compile(`{{(.*?)}}`)
-		if err != nil {
-			log.Fatalf("Cannot compile the regular expression: %v", err)
-		}
-		data = r.ReplaceAll(data, []byte(`{{"{{"}}$1{{"}}"}}`))
+		data = replace(data)
```
zli82016, line 63: Move the replacing logic to the function `replace`

mmv1/main.go
```diff
@@ -33,6 +33,9 @@ var yamlMode = flag.Bool("yaml", false, "copy text over from ruby yaml to go yam
 // Example usage: --template
 var templateMode = flag.Bool("template", false, "copy templates over from .erb to go .tmpl")
 
+// Example usage: --handwritten
+var handwrittenMode = flag.Bool("handwritten", false, "copy handwritten files over from .erb to go .tmpl")
```
c2thorn, line 37: We should add this to the GNUMakefile command `make refresh-go`

c2thorn (approved):

mmv1/main.go
zli82016: Sure. I will add it there after the first round is done.

PR #10999, opened 2024-06-18
Add data_source_google_site_verification_token

SirGitsalot:
This is a start to implementing https://github.com/hashicorp/terraform-provider-google/issues/5698 which will ultimately allow TPG users to verify site ownership for use with the [Google Search Console](https://search.google.com/search-console/welcome).

The verification system works by providing you with a random-looking value (which is what this new data source obtains for you) which you then add to your site as a `meta` tag a la:

```html
<html>
  <head>
    <title>My Web Site</title>
    <meta name="google-site-verification" content="-token-you-got-from-this-new-data-source-" />
  </head>
  <body>
    ...
```

You can also use the token in a DNS record, or in a separate file on your web site, or in several other ways. The gist is that by making a modification to a web site that includes a Google supplied value you are verifying that you are an owner of the site.

The steps will look something like:

1. Use this data source to obtain the token
2. Add the token to your site (for instance, by adding a DNS record via `google_dns_record_set` or adding a `meta` tag as shown above)
3. Use the forthcoming `google_site_verification_web_resource` to  request confirmation of your site ownership by validating that step 2 has taken place.

```release-note:new-datasource
`google_site_verification_token`
```


shuyama1:
@SirGitsalot Is this PR ready for review. Are you able to figure out why the tests are failing. Test failures `TestAccLoggingBucketConfigOrganization_basic` and  `TestAccLoggingBucketConfigOrganization_basic ` are not related to your change, please ignore them.

SirGitsalot:
Not quite ready - the tests are failing because the API requires an additional OAuth scope that the test creds don't have. I need to update the test configs to add that.

shuyama1 (commented):
> Not quite ready - the tests are failing because the API requires an additional OAuth scope that the test creds don't have. I need to update the test configs to add that.

Cool, thanks! Just want to make sure in case it's waiting on me.

shuyama1 (commented):
I assume this PR is still WIP, so I added the `disable-review-reminders` label. Please let me know if this is ready for review.

SirGitsalot:
@shuyama1  - this is ready for review.

This API requires some specific OAuth scopes, which is the reason for the sub provider blocks in the tests. That part I copied from other TPG tests, but I couldn't find any other tests that use a sub provider to add scopes (which is not too surprising since this may be the first non-cloud API we're supporting). If there's a better way of authenticating with additional scopes please let me know.

Thanks!

shuyama1 (approved):
LGTM. The test failures are unrelated to this PR.

@SirGitsalot Please let me know if you have permission to merge the PR or if I should do it. Thank you.

PR #11001, opened 2024-06-19
Correct redisVersion behavior

AndreasBergmeier6176:
Especially in Terraform we encounter a different behavior. Try to clarify, what the actual behavior is.

Basically not provided version either defaults to recommended version (for creation) or is ignored.

```release-note:REPLACEME

```

### DIFF
--- a/mmv1/products/redis/Instance.yaml
+++ b/mmv1/products/redis/Instance.yaml
@@ -423,9 +423,9 @@ properties:
   - !ruby/object:Api::Type::String
     name: redisVersion
     description: |
-      The version of Redis software. If not provided, latest supported
-      version will be used. Please check the API documentation linked
-      at the top for the latest valid values.
+      The version of Redis software. If not provided, the current recommended
+      version will be used for creation. For updates, a not provided version has no effect.
+      Please check the API documentation linked at the top for the latest valid values.
     default_from_api: true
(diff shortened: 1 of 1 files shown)


mmv1/products/redis/Instance.yaml
```diff
@@ -423,9 +423,9 @@ properties:
   - !ruby/object:Api::Type::String
     name: redisVersion
     description: |
-      The version of Redis software. If not provided, latest supported
-      version will be used. Please check the API documentation linked
-      at the top for the latest valid values.
+      The version of Redis software. If not provided, the current recommended
+      version will be used for creation. For updates, a not provided version has no effect.
```
c2thorn, line 427:
```suggestion
      version will be used for creation. When updating, using a version that
      is not provided will have no effect.
```
is this an accurate rewording?

c2thorn (changes requested):

mmv1/products/redis/Instance.yaml
AndreasBergmeier6176: I am not a native English speaker, but I would assume the rewording sounds more like a version that is not "available". On the other hand, this might be a general inaccuracy with the docs - maybe it would be clearer to replace usages of _provided_ with _set_?

c2thorn: @AndreasBergmeier6176 I see, I think I misunderstood the wording myself. I think you are right, _set_ is clear.

PR #11007, opened 2024-06-20
Fix resource_compute_shared_reservation_update encoder

averbuks:
Fix resource_compute_shared_reservation_update encoder to avoid crashes on resource update
Fixes [hashicorp/terraform-provider-google/16208](https://github.com/hashicorp/terraform-provider-google/issues/16208)

```release-note:bug
compute: fixed a crash in `google_compute_reservation` resource where `share_settings` field has changes  
```

### DIFF
--- a/mmv1/products/compute/Reservation.yaml
+++ b/mmv1/products/compute/Reservation.yaml
@@ -141,6 +141,7 @@ properties:
         description: |
           Type of sharing for this shared-reservation
         default_from_api: true
+        immutable: true
       - !ruby/object:Api::Type::Map
         name: 'projectMap'
         description: |
--- a/mmv1/templates/terraform/update_encoder/go/reservation.go.tmpl
+++ b/mmv1/templates/terraform/update_encoder/go/reservation.go.tmpl
@@ -37,10 +37,27 @@
 		// Set project_map.
 		projectMap := make(map[string]interface{})
 		old, new := d.GetChange("share_settings")
-		oldMap := old.([]interface{})[0].(map[string]interface{})["project_map"]
-		newMap := new.([]interface{})[0].(map[string]interface{})["project_map"]
-		before := oldMap.(*schema.Set)
-		after := newMap.(*schema.Set)
+
+		var before *schema.Set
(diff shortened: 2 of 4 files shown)


slevenick (changes requested):
Can you add a test or update the existing test to where it would have triggered the bug?

This is described (here)[https://github.com/hashicorp/terraform-provider-google/issues/16208#issuecomment-1777760277] as going from a config with no `share_settings` specified to one where it is specified.

averbuks:
> Can you add a test or update the existing test to where it would have triggered the bug?
> 
> This is described (here)[https://github.com/[hashicorp/terraform-provider-google/issues/16208](https://github.com/hashicorp/terraform-provider-google/issues/16208)#issuecomment-1777760277] as going from a config with no `share_settings` specified to one where it is specified.

In fact it does not make sense, because: 
1. when `share_settings` is not specified it is created with `share_type = "LOCAL"` and this is immutable, it's not possible to change it. With the current fix it will return an error from the compute API, but not crash anymore. 
2. When `share_settings` is provided and `share_type = "SPECIFIC_PROJECTS"` it requires at least one `project_map`, it can't be empty. 
3. Given # 1 and # 2 the only scenario is to have share_settings with some project_map and then add more projects to it with an update. That's the test scenario I have uncommented [here](https://github.com/GoogleCloudPlatform/magic-modules/pull/11007/files#diff-db68d023d39de627a3fd1294198e72423b9bb31ab75cc790b1b704baf35fe80e), so it's covered now.

averbuks:
@slevenick I have updated the spec and marked the `share_type` fields as immutable (as per my comment above). Do you think it still makes sense to put a tests: create a reservation with no share setting and then add it?  it fact it will force resource recreation, which is optimal given that it's immutable on the API level

slevenick:
Interesting. I'm not sure how this actually works with the existing implementation.

The top-level field share_settings is marked as ignore_read *AND* default_from_api, which is really unusual to me. I think that would make marking share_type as immutable doesn't actually fix the issue, because adding a share_type may not trigger the recreate

slevenick (approved):

PR #11026, opened 2024-06-24
Custom Update ordering diffs

NickElliot:
Resolved diffs in order of custom update groups: previous issues resulted from go maps being unordered -- every generation would result in new random diffs. added a layer of sorted keys that serves as basis for template iteration.

https://diff.googleplex.com/#key=0bHXRhFA5GZv

```release-note:none

```

### DIFF
--- a/mmv1/api/resource.go
+++ b/mmv1/api/resource.go
@@ -1476,6 +1476,10 @@ func (r Resource) PropertiesByCustomUpdateGroups() []UpdateGroup {
 			UpdateVerb:      prop.UpdateVerb,
 			UpdateId:        prop.UpdateId,
 			FingerprintName: prop.FingerprintName}
+
+		if slices.Contains(updateGroups, groupedProperty){
+			continue
+		}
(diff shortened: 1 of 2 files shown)


mmv1/api/resource.go
```diff
@@ -1467,6 +1468,24 @@ func (r Resource) PropertiesByCustomUpdate() map[UpdateGroup][]*Type {
...
+			UpdateVerb:      prop.UpdateVerb,
+			UpdateId:        prop.UpdateId,
+			FingerprintName: prop.FingerprintName}
+
+		if slices.Contains(updateGroups, groupedProperty){
+			continue
+		}
+		updateGroups = append(updateGroups, groupedProperty)
+	}
+	sort.Slice(updateGroups, func(i, j int) bool { return updateGroups[i].UpdateId < updateGroups[i].UpdateId })
```
c2thorn, line 1485: very nice. wonder why it was sorted like this originally

PR #11032, opened 2024-06-24
New data source for google_gke_hub_membership_binding

sahsagar-google:
Fixes https://github.com/hashicorp/terraform-provider-google/issues/18538

```release-note:new-datasource
`google_gke_hub_membership_binding`
```


mmv1/third_party/terraform/services/gkehub2/data_source_google_gke_hub_membership_binding_test.go
```diff
@@ -0,0 +1,109 @@
...
+					acctest.CheckDataSourceStateMatchesResourceState("data.google_gke_hub_membership_binding.example", "google_gke_hub_membership_binding.example"),
+				),
+			},
+		},
+	})
+}
+
+func testAccDataSourceGoogleGKEHub2MembershipBinding_basic(context map[string]interface{}) string {
+	return acctest.Nprintf(`
+resource "google_project" "project" {
```
shuyama1, line 40: Is there any particular reason that we need a new project created for this test? I check the test for the resource `google_gke_hub_membership_binding` and I don't see project creation is needed.

sahsagar-google: Let me remove the project creation, will update you. Thank you for your comment.

mmv1/third_party/terraform/services/gkehub2/data_source_google_gke_hub_membership_binding_test.go
```diff
@@ -0,0 +1,96 @@
...
+}
+
+resource "time_sleep" "wait_for_gkehub_enablement" {
+  create_duration = "150s"
+  depends_on = [
+  google_project_service.gkehub, 
+  google_project_service.compute,
+  google_project_service.anthos
+  ]
+}
```
shuyama1, line 62: I'd assume that we don't need service enablement for our testing project as all these services should've been already enabled. Can you try removing it and see if it still works? If so, we can save the sleep time for the test. Thanks.

shuyama1 (commented):
Overall looks good, only one small comment regarding the test

PR #11035, opened 2024-06-25
Add example for using TimespanConfig successfully without row_limit_percent or row_limit set

patrickmoy:
TimespanConfig in JobTrigger cannot be set if sampleMethod is also set — this is an API-side restriction. Unfortunately, a previous magic-module contributor erroneously set a default value for sampleMethod, which caused requests to fail.

Add example detailing the correct workaround (removing the default value is a breaking change), which is to set sampleMethod to a blank string when setting TimespanConfig.

This should allow us to close https://github.com/hashicorp/terraform-provider-google/issues/14096. 

```release-note:none

```

### DIFF
--- a/mmv1/products/dlp/JobTrigger.yaml
+++ b/mmv1/products/dlp/JobTrigger.yaml
@@ -117,6 +117,13 @@ examples:
       trigger: 'trigger'
     test_env_vars:
       project: :PROJECT_NAME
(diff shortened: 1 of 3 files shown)


patrickmoy:
Unit tests seem to have failed, but it looks unrelated to my change
https://screenshot.googleplex.com/BmP7BVQLzk9VeJd — it's pubsub, not our service.

PR #11038, opened 2024-06-25
fix: Allow Pub/Sub Topic update to remove schema settings

kamalaboulhosn:
Remove default_from_api from Pub/Sub Topic schemaSettings so that schema settings can be removed from a topic. Fixes https://github.com/hashicorp/terraform-provider-google/issues/15925.

```release-note:breaking-change
pubsub: Allow `schema_settings` of `google_pubsub_topic` to be removed
```

### DIFF
--- a/mmv1/products/pubsub/Topic.yaml
+++ b/mmv1/products/pubsub/Topic.yaml
@@ -124,7 +124,6 @@ properties:
     name: 'schemaSettings'
     description: |
       Settings for validating messages published against a schema.
-    default_from_api: true
     properties:
       - !ruby/object:Api::Type::String
         name: 'schema'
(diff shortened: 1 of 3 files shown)


roaks3 (commented):
LGTM, but I think we might still want an entry in the [upgrade guide](https://github.com/GoogleCloudPlatform/magic-modules/blob/FEATURE-BRANCH-major-release-6.0.0/mmv1/third_party/terraform/website/docs/guides/version_6_upgrade.html.markdown) for this. Something like "a default value is not longer provided for schema_settings, instead an empty value means the setting should be cleared"?

kamalaboulhosn:
@roaks3 Added an entry as suggested.

roaks3 (approved):

PR #11039, opened 2024-06-25
`vertex_ai`: add `google_vertex_ai_index_endpoint_deployed_index` resource

BBBmau:
Resolves https://github.com/hashicorp/terraform-provider-google/issues/12818

Checklist:

- [x] YAML completion
  - [x] successful provider generation and build
- [x] manual testing of crud functions
  - [x] Create
  - [x] Read 
  - [x] Update
  - [x] Delete
- [x] *_test.go.erb files added

```release-note:new-resource
`google_vertex_ai_index_endpoint_deployed_index`
```


BBBmau:
currently it's able to send a request to GCP but returns a 404. This is due to the request wanting an index, which is then used as the `indexEndpoint`. An index can exist but that same index cannot exist as an endpoint. Resulting in this error:
```hcl
2024-06-27T16:54:14.486-0700 [DEBUG] provider.terraform-provider-google_9.9.9_darwin_arm64: POST /v1/projects/hc-terraform-testing/locations/us-central1/indexEndpoints/3623695656033452032:deployIndex?alt=json HTTP/1.1
2024-06-27T16:54:14.486-0700 [DEBUG] provider.terraform-provider-google_9.9.9_darwin_arm64: Host: us-central1-aiplatform.googleapis.com
2024-06-27T16:54:14.486-0700 [DEBUG] provider.terraform-provider-google_9.9.9_darwin_arm64: User-Agent: Terraform/1.8.0-alpha20240131 (+https://www.terraform.io) Terraform-Plugin-SDK/2.33.0 terraform-provider-google/dev
2024-06-27T16:54:14.486-0700 [DEBUG] provider.terraform-provider-google_9.9.9_darwin_arm64: Content-Length: 136
2024-06-27T16:54:14.486-0700 [DEBUG] provider.terraform-provider-google_9.9.9_darwin_arm64: Content-Type: application/json
2024-06-27T16:54:14.486-0700 [DEBUG] provider.terraform-provider-google_9.9.9_darwin_arm64: Accept-Encoding: gzip
2024-06-27T16:54:14.486-0700 [DEBUG] provider.terraform-provider-google_9.9.9_darwin_arm64
2024-06-27T16:54:14.486-0700 [DEBUG] provider.terraform-provider-google_9.9.9_darwin_arm64: {
2024-06-27T16:54:14.486-0700 [DEBUG] provider.terraform-provider-google_9.9.9_darwin_arm64:  "deployedIndex": {
2024-06-27T16:54:14.486-0700 [DEBUG] provider.terraform-provider-google_9.9.9_darwin_arm64:   "id": "test_index_public1",
2024-06-27T16:54:14.486-0700 [DEBUG] provider.terraform-provider-google_9.9.9_darwin_arm64:   "index": "projects/hc-terraform-testing/locations/us-central1/indexes/3623695656033452032"
2024-06-27T16:54:14.486-0700 [DEBUG] provider.terraform-provider-google_9.9.9_darwin_arm64:  }
2024-06-27T16:54:14.486-0700 [DEBUG] provider.terraform-provider-google_9.9.9_darwin_arm64: }
2024-06-27T16:54:14.486-0700 [DEBUG] provider.terraform-provider-google_9.9.9_darwin_arm64
2024-06-27T16:54:14.486-0700 [DEBUG] provider.terraform-provider-google_9.9.9_darwin_arm64: -----------------------------------------------------
2024-06-27T16:54:15.052-0700 [DEBUG] provider.terraform-provider-google_9.9.9_darwin_arm64: 2024/06/27 16:54:15 [DEBUG] Google API Response Details:
2024-06-27T16:54:15.052-0700 [DEBUG] provider.terraform-provider-google_9.9.9_darwin_arm64: ---[ RESPONSE ]--------------------------------------
2024-06-27T16:54:15.052-0700 [DEBUG] provider.terraform-provider-google_9.9.9_darwin_arm64: HTTP/2.0 404 Not Found
2024-06-27T16:54:15.052-0700 [DEBUG] provider.terraform-provider-google_9.9.9_darwin_arm64: Alt-Svc: h3=":443"; ma=2592000,h3-29=":443"; ma=2592000
2024-06-27T16:54:15.052-0700 [DEBUG] provider.terraform-provider-google_9.9.9_darwin_arm64: Cache-Control: private
2024-06-27T16:54:15.052-0700 [DEBUG] provider.terraform-provider-google_9.9.9_darwin_arm64: Content-Type: application/json; charset=UTF-8
2024-06-27T16:54:15.052-0700 [DEBUG] provider.terraform-provider-google_9.9.9_darwin_arm64: Date: Thu, 27 Jun 2024 23:54:14 GMT
2024-06-27T16:54:15.052-0700 [DEBUG] provider.terraform-provider-google_9.9.9_darwin_arm64: Server: ESF
2024-06-27T16:54:15.052-0700 [DEBUG] provider.terraform-provider-google_9.9.9_darwin_arm64: Vary: Origin
2024-06-27T16:54:15.052-0700 [DEBUG] provider.terraform-provider-google_9.9.9_darwin_arm64: Vary: X-Origin
2024-06-27T16:54:15.052-0700 [DEBUG] provider.terraform-provider-google_9.9.9_darwin_arm64: Vary: Referer
2024-06-27T16:54:15.052-0700 [DEBUG] provider.terraform-provider-google_9.9.9_darwin_arm64: X-Content-Type-Options: nosniff
2024-06-27T16:54:15.052-0700 [DEBUG] provider.terraform-provider-google_9.9.9_darwin_arm64: X-Frame-Options: SAMEORIGIN
2024-06-27T16:54:15.052-0700 [DEBUG] provider.terraform-provider-google_9.9.9_darwin_arm64: X-Xss-Protection: 0
2024-06-27T16:54:15.052-0700 [DEBUG] provider.terraform-provider-google_9.9.9_darwin_arm64
2024-06-27T16:54:15.052-0700 [DEBUG] provider.terraform-provider-google_9.9.9_darwin_arm64: {
2024-06-27T16:54:15.052-0700 [DEBUG] provider.terraform-provider-google_9.9.9_darwin_arm64:   "error": {
2024-06-27T16:54:15.052-0700 [DEBUG] provider.terraform-provider-google_9.9.9_darwin_arm64:     "code": 404,
2024-06-27T16:54:15.052-0700 [DEBUG] provider.terraform-provider-google_9.9.9_darwin_arm64:     "message": "IndexEndpoint `projects/hc-terraform-testing/locations/us-central1/indexEndpoints/3623695656033452032` is not found.",
2024-06-27T16:54:15.052-0700 [DEBUG] provider.terraform-provider-google_9.9.9_darwin_arm64:     "status": "NOT_FOUND"
2024-06-27T16:54:15.052-0700 [DEBUG] provider.terraform-provider-google_9.9.9_darwin_arm64:   }
2024-06-27T16:54:15.052-0700 [DEBUG] provider.terraform-provider-google_9.9.9_darwin_arm64: }
2024-06-27T16:54:15.052-0700 [DEBUG] provider.terraform-provider-google_9.9.9_darwin_arm64
2024-06-27T16:54:15.052-0700 [DEBUG] provider.terraform-provider-google_9.9.9_darwin_arm64: 
```

BBBmau:
latest commit supports interaction with GCP through after some url fixes.
image

seems like there's some issues in regards to index deployment. I don't think it should be taking this long for deploying an index on an endpoint.

BBBmau:
recent commit was work done in getting the Read method to work when it comes to how users will import already existing deployed indexes. The method of creating deployed index resources in terraform is a bit tricky since the url path is the same as the index endpoint with the only difference being `:deployIndex`

The import path may fall along the lines of `projects/hc-terraform-testing/locations/us-central1/indexEndpoints/7127777641104408576/deployedIndex/test_index_public1`

image


Some changes were also necessary in order to make the import work such as replacing `name` with `deploy_index_id`. The `name` will be a resource reference to the indexEndpoint being used.

BBBmau:
latest commit supports all CRUD methods for: 
`deployed_index` resource allowing users to `deploy an index onto an existing indexEndpoint - Create`
`perform GET requests and import already present deployedIndexes - Read`
`mutate existing deployedIndexes that involve changing deployed resouces - Update`
`undeploy an existing deployedIndex from an indexEndpoint - Delete`

Some notes worth mentioning:

- For importing, I found the ideal import format for users would be: `projects/{{project}}/locations/{{region}}/indexEndpoints/{{index_endpoint}}/deployedIndex/{{deployed_index_id}}` since this is a fine grained resource that involves the use of an already existing `indexEndpoint` it made sense to copy the id_format from it and add in the `deployedIndex` portion. `/deployedIndex/{{deployed_index_id}}`

- There were two fields that caused issues:
  - automaticResources
    - when importing (Read) the plan would return `automaticResources` to be nil. This wouldn't be an issue if it wasn't for the nested object containing values that are filled by API as default, forcing `automaticResources` to exist in state.
    - when mutating, it is only possible to perform a `mutateDeployIndex (update)` on values that are deploy resources which are found in `automaticResources`. The issue is the [API references](https://cloud.google.com/vertex-ai/docs/reference/rest/v1/AutomaticResources) both values in `automaticResources` as immutable which goes against the use of `mutateDeployIndex` from docs found [here](https://cloud.google.com/vertex-ai/docs/vector-search/deploy-index-public#mutate-deployed-index)
  - reservedIpRanges
    - when importing (read) the plan returns to be nil if not set in config, yet the resp from the api returns an empty array. The empty array makes sense in this case since the `reservedIpRanges` field is a list type. I would've expected it to handle it as empty rather than nil. Probably some extra logic for this.

Both fields were set to `ignore_read` to continue CRUD work. Tests are the next step which will include addressing the behavior of both fields from above before marking this as ready for review.

BBBmau:
>     * There were two fields that caused issues:
>       
>       * automaticResources
>         
>         * when importing (Read) the plan would return `automaticResources` to be nil. This wouldn't be an issue if it wasn't for the nested object containing values that are filled by API as default, forcing `automaticResources` to exist in state.
>         * when mutating, it is only possible to perform a `mutateDeployIndex (update)` on values that are deploy resources which are found in `automaticResources`. The issue is the [API references](https://cloud.google.com/vertex-ai/docs/reference/rest/v1/AutomaticResources) both values in `automaticResources` as immutable which goes against the use of `mutateDeployIndex` from docs found [here](https://cloud.google.com/vertex-ai/docs/vector-search/deploy-index-public#mutate-deployed-index)
>       * reservedIpRanges
>         
>         * when importing (read) the plan returns to be nil if not set in config, yet the resp from the api returns an empty array. The empty array makes sense in this case since the `reservedIpRanges` field is a list type. I would've expected it to handle it as empty rather than nil. Probably some extra logic for this.
> 
> 
> Both fields were set to `ignore_read` to continue CRUD work. Tests are the next step which will include addressing the behavior of both fields from above before marking this as ready for review.

After looking into `ignore_read`, the two fields are valid cases for using ignore _read on them since setting `ignore_read` to `true` results in the provider setting the field's value in the resource state based only
on the user's configuration. If false or unset, the provider sets the field's
value in the resource state based on the API response. This info comes from [`field-reference.md`](https://github.com/GoogleCloudPlatform/magic-modules/blob/main/docs/content/develop/field-reference.md)

`ignore_read` was also used to prevent diffs such as:
```hcl
~ index                 = "projects/1067888929963/locations/us-central1/indexes/8115499722661167104" -> "projects/ci-test-project-188019/locations/us-central1/indexes/8115499722661167104"
```

I've also commented out the fields `acceleratorType`, `acceleratorCount`, `tpuTopolgy` and `autoscalingMetricSpecs`. This is due to experiencing reads that don't produce the same state as what's set in config. After looking at what can be set in the Google Cloud console it's unclear whether these values are ever used for a deployed_index resource or where they're used in the first place. From Google Cloud console for deploying an index:
image

Marking as ready for review since this will allow tests to pass though I can imagine there will be changes due to the comment from above. Requesting @SarahFrench for a quick review. Let me know what doesn't make sense from the comments i left in this PR!

mmv1/products/vertexai/IndexEndpointDeployedIndex.yaml
```diff
@@ -0,0 +1,268 @@
...
+  - !ruby/object:Api::Type::ResourceRef
+    name: 'indexEndpoint'
+    description: |
+      Identifies the index endpoint. Must be in the format
+      'projects/{{project}}/locations/{{region}}/indexEndpoints/{{indexEndpoint}}'
+    required: true
+    immutable: true
+    resource: 'IndexEndpoint'
+    imports: 'name'
+    url_param_only: true
```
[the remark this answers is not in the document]
SarahFrench: What's your preference here? The options are:

- Schema includes project, region and index_endpoint where index_endpoint is a short name
- Schema doesn't include project or region, and only accepts index_endpoint of the format projects/{{project}}/locations/{{region}}/indexEndpoints/{{indexEndpoint}}, thereby providing the region and project values.
    - An example of this elsewhere in the provider is [google_kms_crypto_key](https://github.com/GoogleCloudPlatform/magic-modules/blob/24f92f9cd64e2829ef3e3fd4e7138594f0b5f56b/mmv1/products/kms/CryptoKey.yaml)

SarahFrench (commented):
Good job on this, looks like a lot of work has gone into navigating the difficulties of fine-grained resources! This is an initial review with some questions and nits, and I'm planning to follow up with another review soon after (e.g. questions on the commented out fields, which I need some more time to think on)

mmv1/products/vertexai/IndexEndpointDeployedIndex.yaml
```diff
@@ -0,0 +1,268 @@
...
+      display_name: "vertex-deployed-index"
+    skip_docs: true
+custom_code: !ruby/object:Provider::Terraform::CustomCode
+  encoder: templates/terraform/encoders/vertex_ai_index_endpoint_deployed_index.go.erb
+  update_encoder: templates/terraform/update_encoder/vertex_ai_index_endpoint_deployed_index.go.erb
+  decoder: templates/terraform/decoders/vertex_ai_index_endpoint_deployed_index.go.erb
+  pre_delete: templates/terraform/pre_delete/vertex_ai_index_endpoint_deployed_index.go.erb
+parameters:
+  - !ruby/object:Api::Type::String
+    name: region
```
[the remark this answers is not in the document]
SarahFrench: [This is the related comment](https://github.com/GoogleCloudPlatform/magic-modules/pull/11039#discussion_r1666841675) - if we decide the schema should contain region then it should be updated to include project too, to avoid the bug I described

mmv1/products/vertexai/IndexEndpointDeployedIndex.yaml
```diff
@@ -0,0 +1,268 @@
...
+    description: If this timestamp's value is at least the Index.update_time of the original Index, it means that this DeployedIndex and the original Index are in sync. If this timestamp is older, then to see which updates this DeployedIndex already contains (and which it does not), one must list the operations that are running on the original Index.
+  - !ruby/object:Api::Type::NestedObject
+    name: automaticResources
+    description: A description of resources that the DeployedIndex uses, which to large degree are decided by Vertex AI, and optionally allows only a modest additional configuration.
+    ignore_read: true
+    properties:
+      - !ruby/object:Api::Type::Integer
+        name: 'minReplicaCount'
+        description: The minimum number of replicas this DeployedModel will be always deployed on.
+        default_value: 2
```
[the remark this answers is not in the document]
BBBmau: Comes from this tutorial on how to [deploy an index](https://cloud.google.com/vertex-ai/docs/vector-search/deploy-index-public#deploy_index_autoscaling-drest)

> If both minReplicaCount and maxReplicaCount are not set, they are set to 2 by default.

mmv1/products/vertexai/IndexEndpointDeployedIndex.yaml
```diff
@@ -0,0 +1,268 @@
...
+      The name of the DeployedIndex resource.
+  - !ruby/object:Api::Type::String
+    name: 'deployedIndexId'
+    description: The user specified ID of the DeployedIndex. The ID can be up to 128 characters long and must start with a letter and only contain letters, numbers, and underscores. The ID must be unique within the project it is created in.
+    required: true
+  - !ruby/object:Api::Type::String
+    name: 'index'
+    description: The name of the Index this is the deployment of.
+    ignore_read: true
+    required: true
```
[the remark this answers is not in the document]
BBBmau: This was actually based off of `resource_alloydb_instance` and how they separated the instance_id and name to be separate in order to allow the `instance_id` to be used in the `self_link`

https://github.com/GoogleCloudPlatform/magic-modules/blob/aa0fe465572e7407d3352a34841f346916de786f/mmv1/products/alloydb/Instance.yaml#L123-L135

mmv1/products/vertexai/IndexEndpointDeployedIndex.yaml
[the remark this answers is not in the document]
BBBmau: After thinking through this it makes sense to go with the `projects/{{project}}/locations/{{region}}/indexEndpoints/{{indexEndpoint}}` format. Would make it easier for users since all they would need is to use the id of the `indexEndpoint` resource. Since this `deployed_index` resource depends on an index_endpoint existing already. It would use the `project` and `region` values previously set by `indexEndpoint`. We know that it depends on it since it requires a `indexEndpoint` link in order to `:deployIndex`.

BBBmau:
latest commit removes `project` and `region` and relies on the user to input the resource `index_endpoint.Id` as the value for `index_endpoint` when wanting to deploy:

`tfstate:`
```hcl
            "id": "projects/hc-terraform-testing/locations/us-central1/indexEndpoints/5760653679220752384/deployedIndex/index_endpoint_deployed_index_dedicated_resources_test",
            "index": null,
            "index_endpoint": "projects/hc-terraform-testing/locations/us-central1/indexEndpoints/5760653679220752384",
            "index_sync_time": "2024-07-06T02:40:57.481753Z",
```
`tfconfig:`
```hcl
  depends_on = [ google_vertex_ai_index_endpoint.vertex_endpoint ]
  index_endpoint = google_vertex_ai_index_endpoint.vertex_endpoint.id
  index = google_vertex_ai_index.index.id // this is the index that will be deployed onto an endpoint
  deployed_index_id = "index_endpoint_deployed_index_dedicated_resources_test" //deployed index id
  display_name = "dedicated_resources"
```

The import format remains the same: `import_format: ['projects/{{project}}/locations/{{region}}/indexEndpoints/{{index_endpoint}}/deployedIndex/{{deployed_index_id}}']
`

mmv1/templates/terraform/examples/vertex_ai_index_endpoint_deployed_index_basic.tf.erb
```diff
@@ -0,0 +1,72 @@
+resource "google_vertex_ai_index_endpoint_deployed_index" "<%= ctx[:primary_resource_id] %>" {
+  depends_on = [ google_vertex_ai_index_endpoint.vertex_index_endpoint_deployed ]
+  index_endpoint = google_vertex_ai_index_endpoint.vertex_index_endpoint_deployed.name
+  index = google_vertex_ai_index.index.id // this is the index that will be deployed onto an endpoint
+  deployed_index_id = "index_endpoint_deployed_index_basic_test" //deployed index id
```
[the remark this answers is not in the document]
BBBmau: latest commit addresses this by adding `<%= ctx[:vars]['deployed_index_id'] %>` on both tests

mmv1/products/vertexai/IndexEndpointDeployedIndex.yaml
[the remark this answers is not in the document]
SarahFrench: ok, sounds good

mmv1/products/vertexai/IndexEndpointDeployedIndex.yaml
```diff
@@ -0,0 +1,267 @@
...
+    description: |
+      The name of the DeployedIndex resource.
+  - !ruby/object:Api::Type::String
+    name: 'deployedIndexId'
+    description: The user specified ID of the DeployedIndex. The ID can be up to 128 characters long and must start with a letter and only contain letters, numbers, and underscores. The ID must be unique within the project it is created in.
+    required: true
+  - !ruby/object:Api::Type::String
+    name: 'index'
+    description: The name of the Index this is the deployment of.
+    ignore_read: true
```
SarahFrench, line 109: I saw in [this comment](https://github.com/GoogleCloudPlatform/magic-modules/pull/11039#issuecomment-2209316648) you ignore_read this field because there's a diff where the API returns the project number but the provider wants to set the project id. Have you considered using the [`tpgresource.ProjectNumberDiffSuppress` diff suppress func](https://github.com/GoogleCloudPlatform/magic-modules/blob/a72543a018c089c8579bce785ed34589af22e174/mmv1/third_party/terraform/tpgresource/go/common_diff_suppress.go#L94)?

mmv1/products/vertexai/IndexEndpointDeployedIndex.yaml
```diff
@@ -0,0 +1,267 @@
...
+            name: allowedIssuers
+            output: true
+            description: |
+              A list of allowed JWT issuers. Each entry must be a valid Google service account, in the following format: service-account-name@project-id.iam.gserviceaccount.com
+            item_type: Api::Type::String
+  - !ruby/object:Api::Type::Array
+    name: reservedIpRanges
+    description: A list of reserved ip ranges under the VPC network that can be used for this DeployedIndex.
+    item_type: Api::Type::String
+    ignore_read: true
```
SarahFrench, line 262: If this is `ignore_read: true` due to [this reason here](https://github.com/GoogleCloudPlatform/magic-modules/pull/11039#issuecomment-2209316648):

>    * when importing (read) the plan returns to be nil if not set in config, yet the resp from the api returns an empty array. The empty array makes sense in this case since the `reservedIpRanges` field is a list type. I would've expected it to handle it as empty rather than nil. Probably some extra logic for this.

Could the diff between null (from config) and an empty list (from API) be suppressed by a diff suppress func?

Or another way to think about it, what would a user do if they want to import a deployed index that has reserved ip ranges and reference those values elsewhere?

SarahFrench (commented):
Left some more comments - I'm going to have a play around running acceptance tests with a debugger to learn more about the commented out fields. If the API doesn't return data that's used when creating the deployed index that's fine, but I'm currently not 100% clear about where the problem lies

mmv1/products/vertexai/IndexEndpointDeployedIndex.yaml
BBBmau: DiffSuppressFuncs tend to work only with `string-like` attributes. Though I'm sure there's a work around that can suppress an empty list being set as null in state. We may have to do something similar like what was done here: https://github.com/GoogleCloudPlatform/magic-modules/pull/10936#issuecomment-2168954728

mmv1/products/vertexai/IndexEndpointDeployedIndex.yaml
BBBmau: latest commit replaces `ignore_read` eith the `ProjectNumberDiffSuppress` func

mmv1/products/vertexai/IndexEndpointDeployedIndex.yaml
SarahFrench: Or maybe you could have a custom flattener that converts an empty array for that field to null, and otherwise doesn't change the value?

BBBmau: had the same idea, trying it out right now.

BBBmau: i may have just been importing with the wrong resource. Works fine now.

BBBmau:
The only blocker this PR faces are attributes that are not being applied onto the deployed_index. They are accepted as valid inputs but when importing them they return empty values for each of them. I've commented them out for now and are the ones below:
```
      # - !ruby/object:Api::Type::Array
      #   name: 'autoscalingMetricSpecs'
      #   description: The metric specifications that overrides a resource utilization metric (CPU utilization, accelerator's duty cycle, and so on) target value (default to 60 if not set). At most one entry is allowed per metric.
      #   immutable: true
      #   item_type: !ruby/object:Api::Type::NestedObject
      #     properties:
      #       - !ruby/object:Api::Type::String
      #         name: metricName
      #         required: true
      #         description: |
      #           The resource metric name.
      #       - !ruby/object:Api::Type::String
      #         name: target
      #         default_value: "60"
      #         description: |
      #           The target resource utilization in percentage (1% - 100%) for the given metric; once the real usage deviates from the target by a certain percentage, the machine replicas change. The default value is 60 (representing 60%) if not provided.
 ```
 
 ```
           # - !ruby/object:Api::Type::Enum
          #   name: 'acceleratorType'
          #   description: The name of the service attachment resource. Populated if private service connect is enabled.
          #   immutable: true
          #   values:
          #     - NVIDIA_TESLA_K80
          #     - NVIDIA_TESLA_P100
          #     - NVIDIA_TESLA_V100
          #     - NVIDIA_TESLA_P4
          #     - NVIDIA_TESLA_T4
          #     - NVIDIA_TESLA_A100
          #     - NVIDIA_A100_80GB
          #     - NVIDIA_L4
          #     - NVIDIA_H100_80GB
          #     - TPU_V2
          #     - TPU_V3
          #     - TPU_V4_POD
          #     - TPU_V5_LITEPOD
          # - !ruby/object:Api::Type::Integer
          #   name: 'acceleratorCount'
          #   description: The number of accelerators to attach to the machine.
          # - !ruby/object:Api::Type::String
          #   name: 'tpuTopology'
          #   immutable: true
          #   description: |
          #     The topology of the TPUs. Corresponds to the TPU topologies available from GKE. (Example: tpuTopology: "2x2x1").
  ```
  
  It's unclear where exactly these attributes are used and may be an issue the service team will have to look into: I talked about this in a previous comment which can be seen [here](https://github.com/GoogleCloudPlatform/magic-modules/pull/11039#issuecomment-2209316648)

File: mmv1/products/vertexai/IndexEndpointDeployedIndex.yaml [hunk 8]
```diff
@@ -0,0 +1,224 @@
...
+            name: audiences
+            output: true
+            description: The list of JWT audiences. that are allowed to access. A JWT containing any of these audiences will be accepted.
+            item_type: Api::Type::String
+          - !ruby/object:Api::Type::Array
+            name: allowedIssuers
+            output: true
+            description: |
+              A list of allowed JWT issuers. Each entry must be a valid Google service account, in the following format: service-account-name@project-id.iam.gserviceaccount.com
+            item_type: Api::Type::String
```
SarahFrench, line 215: Are these both definitely outputs? [The docs](https://cloud.google.com/vertex-ai/docs/reference/rest/v1/projects.locations.indexEndpoints#AuthProvider) don't indicate that's the case, and I believe providing these values is how the user controls access.

mmv1/products/vertexai/IndexEndpointDeployedIndex.yaml
```diff
@@ -0,0 +1,224 @@
...
+  - !ruby/object:Api::Type::Array
+    name: reservedIpRanges
+    description: A list of reserved ip ranges under the VPC network that can be used for this DeployedIndex.
+    item_type: Api::Type::String
+  - !ruby/object:Api::Type::String
+    name: deploymentGroup
+    output: true
+    description: |
+      The deployment group can be no longer than 64 characters (eg: 'test', 'prod'). If not set, we will use the 'default' deployment group. Note: we only support up to 5 deployment groups(not including 'default').
```
SarahFrench, line 224: nit:

```suggestion
      The deployment group can be no longer than 64 characters (eg: 'test', 'prod'). If not set, we will use the 'default' deployment group. Note: we only support up to 5 deployment groups (not including 'default').
```

mmv1/products/vertexai/IndexEndpointDeployedIndex.yaml
```diff
@@ -0,0 +1,224 @@
...
+    ignore_read: true
+    properties:
+      - !ruby/object:Api::Type::Integer
+        name: 'minReplicaCount'
+        description: The minimum number of replicas this DeployedModel will be always deployed on.
+        default_value: 2
+      - !ruby/object:Api::Type::Integer
+        name: 'maxReplicaCount'
+        description: The maximum number of replicas this DeployedModel may be deployed on when the traffic against it increases.
+        custom_expand: templates/terraform/custom_expand/vertex_ai_endpoint_deployed_index.go.erb
```
SarahFrench, line 169: Please see other comment about maxReplicaCount defaults!

mmv1/templates/terraform/custom_expand/vertex_ai_endpoint_deployed_index.go.erb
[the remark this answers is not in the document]
SarahFrench: Question: is this also needed for dedicated_resources. max_replica_count?

SarahFrench (commented):
Here's some more feedback! Now that some of the bigger questions are resolved I've been looking through the schema in detail and trying to find any discrepancies.

mmv1/templates/terraform/custom_expand/vertex_ai_endpoint_deployed_index.go.erb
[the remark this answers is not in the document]
BBBmau: Yep this same logic should also be applied to dedicated_resources. Adding right now.

https://cloud.google.com/vertex-ai/docs/reference/rest/v1/DedicatedResources

mmv1/products/vertexai/IndexEndpointDeployedIndex.yaml
```diff
@@ -0,0 +1,224 @@
...
+            immutable: true
+      - !ruby/object:Api::Type::Integer
+        name: 'minReplicaCount'
+        description: The minimum number of machine replicas this DeployedModel will be always deployed on. This value must be greater than or equal to 1.
+        immutable: true
+        required: true
+      - !ruby/object:Api::Type::Integer
+        name: 'maxReplicaCount'
+        description: The maximum number of replicas this DeployedModel may be deployed on when the traffic against it increases.
+        immutable: true
```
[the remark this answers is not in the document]
BBBmau: i've updated the basic test to include this. The only issue I'm currently looking into is when the user doesn't specify the use of `automatic_resources` since by default both min and max will be set to 2 if neither are set such as when their is no automatic_resources block.

For dedicated resources their is no default value set since it is a required field assuming the user is using the dedicated resources block. It's expected that the user must at least input the value for `minReplicaCount`. max will just default to min if nothing is set.

BBBmau:
latest passes are after addressing the recent requested changes:

> I think that the description should mention the behaviour "If maxReplicaCount is not set, the default value is minReplicaCount".
> 
> Also, what happens when a resource is made with this field unset? I'm wondering if there's a potential for drift given the field isn't Optional+Computed (i.e default_from_api) Your examples both provide a value for automatic_resources.max_replica_count; could you please make an example where that field isn't present, or maybe add a multi-step test for the resource where that field is initially not set and then is added to the config later?
> 
> Edit: what about dedicated_resources.max_replica_count?

Both `dedicated_resources` and `automatic_resources` now have the same logic which checks whether maxReplica has been set or not and uses `minReplica` by default if max has not been set.

In the case of `automatic_resources` not being set, it will now set both `minReplicaCount` and `maxReplicaCount` to  two using `default_from_api`. This can be seen within `basic` test (I locally tested this with the tfconfig and state being shown [here](https://github.com/GoogleCloudPlatform/magic-modules/pull/11039#discussion_r1674588291)). The new `automatic_resource` test is for when the user manually sets the replica counts.

This isn't necessary for `dedicated_resources` since if the user intends on using the `dedicated_resources` block over `automatic_resources` they MUST set `min_replica_count`.


> Are these both definitely outputs? [The docs](https://cloud.google.com/vertex-ai/docs/reference/rest/v1/projects.locations.indexEndpoints#AuthProvider) don't indicate that's the case, and I believe providing these values is how the user controls access.

I'm unsure why I set them both to be outputs, I've addressed this by including this as part of the basic test since in order to use these fields the endpoint cannot be a public endpoint. basic tests were already testing the use of private_endpoints with the use `network` and `global address` resources.

@SarahFrench it's ready for another review. Let me know what else you find. The `audience` value i used come from the docs here: https://cloud.google.com/vertex-ai/docs/vector-search/jwt-auth#deploy-index we could use a different value that may be more appropriate. I mostly just went with what the docs used as an example.

I also took one more look at the schema and noticed that `deploymentGroup` was marked as output. This isn't true and should be set to optional. I've included it into the tests with the latest commit as well as the check that the default value should be `default`.

mmv1/templates/terraform/examples/vertex_ai_index_endpoint_deployed_index_basic_two.tf erb
SarahFrench, line 1: Looks like the name needs a fix here

```diff
- mmv1/templates/terraform/examples/vertex_ai_index_endpoint_deployed_index_basic_two.tf erb
+ mmv1/templates/terraform/examples/vertex_ai_index_endpoint_deployed_index_basic_two.tf.erb
```

SarahFrench:
@BBBmau I'm afraid as it's Friday and I'm going on PTO I'll need to pass review of this PR onto someone else

BBBmau:
friendly ping, let me know if you need help with the review! @roaks3

mmv1/products/vertexai/IndexEndpointDeployedIndex.yaml
```diff
@@ -0,0 +1,248 @@
...
+            name: 'machineType'
+            description: The type of the machine.
+            immutable: true
+      - !ruby/object:Api::Type::Integer
+        name: 'minReplicaCount'
+        description: The minimum number of machine replicas this DeployedModel will be always deployed on. This value must be greater than or equal to 1.
+        required: true
+      - !ruby/object:Api::Type::Integer
+        name: 'maxReplicaCount'
+        description: The maximum number of replicas this DeployedModel may be deployed on when the traffic against it increases. If maxReplicaCount is not set, the default value is minReplicaCount
```
roaks3, line 217: To confirm, are we implementing this default behavior client-side because it is not done on the server, or something similar? This case in particular looks involved enough that we would prefer to avoid duplicating what the server does, if that was the case.

roaks3 (commented):
This is a complicated API, but overall things look good.

In addition to my one comment, I was curious why there aren't any update tests here (I saw some fields that seem to be updatable). Is there a reason they aren't being used?

mmv1/products/vertexai/IndexEndpointDeployedIndex.yaml
BBBmau: did a test of settting `default_from_value` on both `maxReplicaCounts` and we get the same behavior from the server. No need for the custom expand.

I'll go ahead and add in an update test before requesting another review. Thanks!

BBBmau:
Thanks for your review @roaks3 i was able to address some issues regarding updates on the resource. Let me know if there's anything else missing here.

BBBmau:
It's worth noting that the `update` operation is only meant to be done on deployment resources. This is mentioned in the docs for [mutating a deployIndex](https://cloud.google.com/vertex-ai/docs/vector-search/deploy-index-vpc#mutate-deployed-index)

This is the reason for the update test having updates on only `minReplicaCount` and `maxReplicaCount`. Let me know if theirs anything else that's unclear. I appreciate the reviews on this!

roaks3 (commented):
> It's worth noting that the update operation is only meant to be done on deployment resources. This is mentioned in the docs for [mutating a deployIndex](https://cloud.google.com/vertex-ai/docs/vector-search/deploy-index-vpc#mutate-deployed-index)
> 
> This is the reason for the update test having updates on only minReplicaCount and maxReplicaCount. Let me know if theirs anything else that's unclear. I appreciate the reviews on this!

FWIW, everything else LGTM, but I'm still a little unclear on this point. For example, looking at `reservedIpRanges`, it looks like it is mutable and would be sent to the mutate url, but I'm not sure why then we aren't testing it in the update test. According to the gcloud command, it does seem like it should be a field that can be mutated https://cloud.google.com/sdk/gcloud/reference/ai/index-endpoints/mutate-deployed-index.

But I guess what I was expecting to see was either the field would be marked as immutable, or it would be included in an update test. Am I missing something?

BBBmau:
> For example, looking at `reservedIpRanges`, it looks like it is mutable and would be sent to the mutate url, but I'm not sure why then we aren't testing it in the update test. According to the gcloud command, it does seem like it should be a field that can be mutated https://cloud.google.com/sdk/gcloud/reference/ai/index-endpoints/mutate-deployed-index.

You made a great point with the values not being marked as immutable which therefore we'd expect these values to also be part of the update. I went ahead and did some testing myself and although they aren't explicitly set as being immutable in the [API Reference of deployed_index](https://cloud.google.com/vertex-ai/docs/reference/rest/v1/projects.locations.indexEndpoints#DeployedIndex), they aren't actually updated when attempting to do so in an apply.


making updates to the tfconfig like what's seen below:
```hcl
  deployed_index_id = "index_endpoint_deployed_index_dedicated_resources_test_10" //deployed index id
  display_name = "dedicated_resources_test"
  reserved_ip_ranges = [ "10.0.0.0", "10.0.0.1"]
  enable_access_logging = true
  dedicated_resources {
    machine_spec {
      machine_type      = "n1-standard-32"
    }
    max_replica_count = 6
    min_replica_count = 3
  }
```
isn't actual matching the tfstate, yet the dedicated/automatic resources do:
```hcl
"attributes": {
            "automatic_resources": [],
            "create_time": "2024-07-30T20:32:09.364359Z",
            "dedicated_resources": [
              {
                "machine_spec": [
                  {
                    "machine_type": "n1-standard-32"
                  }
                ],
                "max_replica_count": 6,
                "min_replica_count": 3
              }
            ],
            "deployed_index_auth_config": [],
            "deployed_index_id": "index_endpoint_deployed_index_dedicated_resources_test_10",
            "deployment_group": "default",
            "display_name": "dedicated_resources_test(updated)",
            "enable_access_logging": true,
            "id": "projects/hc-terraform-testing/locations/us-central1/indexEndpoints/5511433975622533120/deployedIndex/index_endpoint_deployed_index_dedicated_resources_test_10",
            "index": "projects/871647908372/locations/us-central1/indexes/6608904909817380864",
            "index_endpoint": "projects/hc-terraform-testing/locations/us-central1/indexEndpoints/5511433975622533120",
            "index_sync_time": "2024-07-30T21:43:22.075405Z",
            "name": "",
            "private_endpoints": [],
            "reserved_ip_ranges": [
              "10.0.0.0"
            ],
```

When you attempt to update values yourself through the console you're greeted with only resource values, below is an example of what automatic resources can be updated:
image

> But I guess what I was expecting to see was either the field would be marked as immutable, or it would be included in an update test. Am I missing something?

So while it doesn't mark it as immutable in the docs, it makes sense like you said to mark them as immutable if they aren't update-able, which in this case only the automatic/dedicated resources are.

I've updated the update test to also include automatic_resource updates since previously the test was only done on dedicated resources. Let me know your thoughts! @roaks3

mmv1/third_party/terraform/services/vertexai/resource_vertex_ai_index_endpoint_deployed_index_test.go
```diff
@@ -0,0 +1,365 @@
...
+		ProtoV5ProviderFactories: acctest.ProtoV5ProviderFactories(t),
+		Steps: []resource.TestStep{
+			{
+				Config: testAccVertexAIIndexEndpointDeployedIndex_vertexAiIndexEndpointDeployedIndexDedicatedResourcesBasicExample_full(context),
+			},
+			{
+				ResourceName:            "google_vertex_ai_index_endpoint_deployed_index.dedicated_resources_update",
+				ImportState:             true,
+				ImportStateVerify:       true,
+				ImportStateVerifyIgnore: []string{"index_endpoint"},
```
SarahFrench, line 29: Is this ImportStateVerifyIgnore due to project id vs project number issues?

mmv1/products/vertexai/IndexEndpointDeployedIndex.yaml
```diff
@@ -0,0 +1,256 @@
...
+            item_type: Api::Type::String
+          - !ruby/object:Api::Type::Array
+            name: allowedIssuers
+            immutable: true
+            description: |
+              A list of allowed JWT issuers. Each entry must be a valid Google service account, in the following format: service-account-name@project-id.iam.gserviceaccount.com
+            item_type: Api::Type::String
+  - !ruby/object:Api::Type::Array
+    name: reservedIpRanges
+    immutable: true
```
SarahFrench, line 248: You could put a comment here about how the API accepts new values without error but doesn't process the update  to the fields correctly. It'll be easier for people to read that versus reading through your PR (with all the automation messages to scroll through!)

mmv1/products/vertexai/IndexEndpointDeployedIndex.yaml
```diff
@@ -0,0 +1,256 @@
...
+    description: A list of reserved ip ranges under the VPC network that can be used for this DeployedIndex.
+    item_type: Api::Type::String
+  - !ruby/object:Api::Type::String
+    name: deploymentGroup
+    immutable: true
+    default_value: 'default'
+    description: |
+      The deployment group can be no longer than 64 characters (eg: 'test', 'prod'). If not set, we will use the 'default' deployment group. Note: we only support up to 5 deployment groups (not including 'default').
```
SarahFrench, line 256: It'd be worth including some info about how deploymentGroup interacts with the reservedIpRanges value, as [described in the docs](https://cloud.google.com/vertex-ai/docs/reference/rest/v1/projects.locations.indexEndpoints#DeployedIndex.FIELDS.deployment_group).

```suggestion
      The deployment group can be no longer than 64 characters (eg: 'test', 'prod'). If not set, we will use the 'default' deployment group.
      Creating deployment_groups with reserved_ip_ranges is a recommended practice when the peered network has multiple peering ranges. This creates your deployments from predictable IP spaces for easier traffic administration. Also, one deployment_group (except 'default') can only be used with the same reserved_ip_ranges which means if the deployment_group has been used with reserved_ip_ranges: [a, b, c], using it with [a, b] or [d, e] is disallowed. [See the official documentation here](https://cloud.google.com/vertex-ai/docs/reference/rest/v1/projects.locations.indexEndpoints#DeployedIndex.FIELDS.deployment_group).
      Note: we only support up to 5 deployment groups (not including 'default').
```

mmv1/products/vertexai/IndexEndpointDeployedIndex.yaml
```diff
@@ -0,0 +1,256 @@
...
+          - !ruby/object:Api::Type::Array
+            name: allowedIssuers
+            immutable: true
+            description: |
+              A list of allowed JWT issuers. Each entry must be a valid Google service account, in the following format: service-account-name@project-id.iam.gserviceaccount.com
+            item_type: Api::Type::String
+  - !ruby/object:Api::Type::Array
+    name: reservedIpRanges
+    immutable: true
+    description: A list of reserved ip ranges under the VPC network that can be used for this DeployedIndex.
```
SarahFrench, line 249: The docs have more info than this, which would be worth including:

```suggestion
    description: |
      A list of reserved ip ranges under the VPC network that can be used for this DeployedIndex.
      If set, we will deploy the index within the provided ip ranges. Otherwise, the index might be deployed to any ip ranges under the provided VPC network.

      The value should be the name of the address (https://cloud.google.com/compute/docs/reference/rest/v1/addresses) Example: ['vertex-ai-ip-range'].

      For more information about subnets and network IP ranges, please see https://cloud.google.com/vpc/docs/subnets#manually_created_subnet_ip_ranges.
```

SarahFrench (commented):
Some review comments to accompany discussions in Slack

mmv1/products/vertexai/IndexEndpointDeployedIndex.yaml
SarahFrench: (I'd want us to resolve Slack convo before committing to this being immutable)

mmv1/third_party/terraform/services/vertexai/resource_vertex_ai_index_endpoint_deployed_index_test.go
BBBmau: that's right!

mmv1/products/vertexai/IndexEndpointDeployedIndex.yaml
BBBmau: since we confirmed that this would be immutable the extra comment wouldn't be necessary correct?

SarahFrench: I think it'd be useful to provide context that the API accepts changed values but doesn't use them, versus returning an error saying that the field cannot be updated. That's the information that a newcomer wouldn't be able to learn from the API docs

mmv1/products/vertexai/IndexEndpointDeployedIndex.yaml
```diff
@@ -0,0 +1,264 @@
...
+    immutable: true
+    description: The display name of the Index. The name can be up to 128 characters long and can consist of any UTF-8 characters.
+  - !ruby/object:Api::Type::String
+    name: 'createTime'
+    output: true
+    description: The timestamp of when the Index was created in RFC3339 UTC "Zulu" format, with nanosecond resolution and up to nine fractional digits.
+  - !ruby/object:Api::Type::NestedObject
+    name: privateEndpoints
+    output: true
+    description: Provides paths for users to send requests directly to the deployed index services running on Cloud via private services access
```
SarahFrench, line 150:
```suggestion
    description: Provides paths for users to send requests directly to the deployed index services running on Cloud via private services access. This field is populated if [network](https://cloud.google.com/vertex-ai/docs/reference/rest/v1/projects.locations.indexEndpoints#IndexEndpoint.FIELDS.network) is configured.
```

mmv1/products/vertexai/IndexEndpointDeployedIndex.yaml
```diff
@@ -0,0 +1,264 @@
...
+                Corresponding network in pscAutomationConfigs.
+            - !ruby/object:Api::Type::String
+              name: 'matchAddress'
+              output: true
+              description: |
+                ip Address created by the automated forwarding rule.
+  - !ruby/object:Api::Type::String
+    name: 'indexSyncTime'
+    output: true
+    description: If this timestamp's value is at least the Index.update_time of the original Index, it means that this DeployedIndex and the original Index are in sync. If this timestamp is older, then to see which updates this DeployedIndex already contains (and which it does not), one must list the operations that are running on the original Index.
```
SarahFrench, line 185:
```suggestion
    description: |
        The DeployedIndex may depend on various data on its original Index. Additionally when certain changes to the original Index are being done (e.g. when what the Index contains is being changed) the DeployedIndex may be asynchronously updated in the background to reflect these changes. If this timestamp's value is at least the [Index.update_time](https://cloud.google.com/vertex-ai/docs/reference/rest/v1/projects.locations.indexes#Index.FIELDS.update_time) of the original Index, it means that this DeployedIndex and the original Index are in sync. If this timestamp is older, then to see which updates this DeployedIndex already contains (and which it does not), one must [list](https://cloud.google.com/vertex-ai/docs/reference/rest/v1beta1/projects.locations.operations/list#google.longrunning.Operations.ListOperations) the operations that are running on the original Index. Only the successfully completed Operations with updateTime equal or before this sync time are contained in this DeployedIndex.

        A timestamp in RFC3339 UTC "Zulu" format, with nanosecond resolution and up to nine fractional digits. Examples: "2014-10-02T15:01:23Z" and "2014-10-02T15:01:23.045123456Z".
```

mmv1/products/vertexai/IndexEndpointDeployedIndex.yaml
```diff
@@ -0,0 +1,264 @@
...
+  - !ruby/object:Api::Type::NestedObject
+    name: automaticResources
+    description: A description of resources that the DeployedIndex uses, which to large degree are decided by Vertex AI, and optionally allows only a modest additional configuration.
+    default_from_api: true
+    properties:
+      - !ruby/object:Api::Type::Integer
+        name: 'minReplicaCount'
+        description: The minimum number of replicas this DeployedModel will be always deployed on.
```
SarahFrench, line 193:
```suggestion
        description: |
                The minimum number of replicas this DeployedModel will be always deployed on. If minReplicaCount is not set, the default value is 2 (we don't provide SLA when minReplicaCount=1).
        
                If traffic against it increases, it may dynamically be deployed onto more replicas up to [maxReplicaCount](https://cloud.google.com/vertex-ai/docs/reference/rest/v1/AutomaticResources#FIELDS.max_replica_count), and as traffic decreases, some of these extra replicas may be freed. If the requested value is too large, the deployment will error.
```

Adding in info from [the description of the parent field](https://cloud.google.com/vertex-ai/docs/reference/rest/v1/projects.locations.indexEndpoints#DeployedIndex.FIELDS.automatic_resources) and [the field's description](https://cloud.google.com/vertex-ai/docs/reference/rest/v1/AutomaticResources)

mmv1/products/vertexai/IndexEndpointDeployedIndex.yaml
```diff
@@ -0,0 +1,264 @@
...
+        default_from_api: true
+      - !ruby/object:Api::Type::Integer
+        name: 'maxReplicaCount'
+        description: The maximum number of replicas this DeployedModel may be deployed on when the traffic against it increases. If maxReplicaCount is not set, the default value is minReplicaCount
```
SarahFrench, line 197:
```suggestion
        description: |
                The maximum number of replicas this DeployedModel may be deployed on when the traffic against it increases. If maxReplicaCount is not set, the default value is minReplicaCount. The max allowed replica count is 1000.
        
        The maximum number of replicas this DeployedModel may be deployed on when the traffic against it increases. If the requested value is too large, the deployment will error, but if deployment succeeds then the ability to scale the model to that many replicas is guaranteed (barring service outages). If traffic against the DeployedModel increases beyond what its replicas at maximum may handle, a portion of the traffic will be dropped. If this value is not provided, a no upper bound for scaling under heavy traffic will be assume, though Vertex AI may be unable to scale beyond certain replica number.
```

Adding in info from [the description of the parent field](https://cloud.google.com/vertex-ai/docs/reference/rest/v1/projects.locations.indexEndpoints#DeployedIndex.FIELDS.automatic_resources) and [the field's description](https://cloud.google.com/vertex-ai/docs/reference/rest/v1/AutomaticResources)

mmv1/products/vertexai/IndexEndpointDeployedIndex.yaml
```diff
@@ -0,0 +1,264 @@
...
+        default_from_api: true
+  - !ruby/object:Api::Type::NestedObject
+    name: dedicatedResources
+    description: A description of resources that are dedicated to the DeployedIndex, and that need a higher degree of manual configuration.
```
SarahFrench, line 201:
```suggestion
    description: |
        A description of resources that are dedicated to the DeployedIndex, and that need a higher degree of manual configuration. The field minReplicaCount must be set to a value strictly greater than 0, or else validation will fail. We don't provide SLA when minReplicaCount=1. If maxReplicaCount is not set, the default value is minReplicaCount. The max allowed replica count is 1000.

        Available machine types for SMALL shard: e2-standard-2 and all machine types available for MEDIUM and LARGE shard.

        Available machine types for MEDIUM shard: e2-standard-16 and all machine types available for LARGE shard.

        Available machine types for LARGE shard: e2-highmem-16, n2d-standard-32.

        n1-standard-16 and n1-standard-32 are still available, but we recommend e2-standard-16 and e2-highmem-16 for cost efficiency.
```

mmv1/products/vertexai/IndexEndpointDeployedIndex.yaml
```diff
@@ -0,0 +1,264 @@
...
+    properties:
+      - !ruby/object:Api::Type::NestedObject
+        name: 'machineSpec'
+        description: The minimum number of replicas this DeployedModel will be always deployed on.
+        immutable: true
+        required: true
+        properties:
+          - !ruby/object:Api::Type::String
+            name: 'machineType'
+            description: The type of the machine.
```
SarahFrench, line 211:
```suggestion
            description: |
                        The type of the machine.
                        
                        See the [list of machine types supported for prediction](https://cloud.google.com/vertex-ai/docs/predictions/configure-compute#machine-types)

                        See the [list of machine types supported for custom training](https://cloud.google.com/vertex-ai/docs/training/configure-compute#machine-types).

                        For [DeployedModel](https://cloud.google.com/vertex-ai/docs/reference/rest/v1/projects.locations.endpoints#DeployedModel) this field is optional, and the default value is n1-standard-2. For [BatchPredictionJob](https://cloud.google.com/vertex-ai/docs/reference/rest/v1/projects.locations.batchPredictionJobs#BatchPredictionJob) or as part of [WorkerPoolSpec](https://cloud.google.com/vertex-ai/docs/reference/rest/v1/CustomJobSpec#WorkerPoolSpec) this field is required.
```

mmv1/products/vertexai/IndexEndpointDeployedIndex.yaml
```diff
@@ -0,0 +1,264 @@
...
+      - !ruby/object:Api::Type::Integer
+        name: 'maxReplicaCount'
+        description: The maximum number of replicas this DeployedModel may be deployed on when the traffic against it increases. If maxReplicaCount is not set, the default value is minReplicaCount
+        default_from_api: true
+  - !ruby/object:Api::Type::NestedObject
+    name: dedicatedResources
+    description: A description of resources that are dedicated to the DeployedIndex, and that need a higher degree of manual configuration.
+    properties:
+      - !ruby/object:Api::Type::NestedObject
+        name: 'machineSpec'
```
SarahFrench, line 204: I currently see machineType under machineSpec, but these fields aren't present ([present in the docs](https://cloud.google.com/vertex-ai/docs/reference/rest/v1/MachineSpec)):

- acceleratorType
- acceleratorCount
- tpuTopology

Was this a conscious decision (i.e difficulty in acceptance tests, or making this PR more complex) or the fields overlooked?

mmv1/products/vertexai/IndexEndpointDeployedIndex.yaml
```diff
@@ -0,0 +1,264 @@
...
+        description: The minimum number of replicas this DeployedModel will be always deployed on.
+        immutable: true
+        required: true
+        properties:
+          - !ruby/object:Api::Type::String
+            name: 'machineType'
+            description: The type of the machine.
+            immutable: true
+      - !ruby/object:Api::Type::Integer
+        name: 'minReplicaCount'
```
SarahFrench, line 214: The [docs describe this field as immutable](https://cloud.google.com/vertex-ai/docs/reference/rest/v1/DedicatedResources) FYI!

mmv1/products/vertexai/IndexEndpointDeployedIndex.yaml
```diff
@@ -0,0 +1,264 @@
...
+        description: The minimum number of machine replicas this DeployedModel will be always deployed on. This value must be greater than or equal to 1.
+        required: true
+      - !ruby/object:Api::Type::Integer
+        name: 'maxReplicaCount'
```
SarahFrench, line 218: The docs also [describe this field as immutable](https://cloud.google.com/vertex-ai/docs/reference/rest/v1/DedicatedResources)

mmv1/products/vertexai/IndexEndpointDeployedIndex.yaml
```diff
@@ -0,0 +1,264 @@
...
+      - !ruby/object:Api::Type::Integer
+        name: 'minReplicaCount'
+        description: The minimum number of replicas this DeployedModel will be always deployed on.
+        default_from_api: true
+      - !ruby/object:Api::Type::Integer
+        name: 'maxReplicaCount'
+        description: The maximum number of replicas this DeployedModel may be deployed on when the traffic against it increases. If maxReplicaCount is not set, the default value is minReplicaCount
+        default_from_api: true
+  - !ruby/object:Api::Type::NestedObject
+    name: dedicatedResources
```
SarahFrench, line 200: There's a nested field called autoscalingMetricSpecs [described in the docs](https://cloud.google.com/vertex-ai/docs/reference/rest/v1/DedicatedResources) that isn't implemented here- is that intentional or an oversight?

SarahFrench (changes requested):
I've been looking through the docs and comparing to the schema- this review includes some additions to field descriptions based on stuff I read there and also some fields I realised are missing and appear to have missing immutable:true .

I stopped working through the docs and schema as the number of comments was getting too large. Could you please address the missing fields (implement or justify them being absent) and compare the docs and schema to look for any more discrepancies

mmv1/products/vertexai/IndexEndpointDeployedIndex.yaml
SarahFrench: TBH it'll also help with reviewing this PR. Comparing the schema to the API docs is a lot of work so it'd be useful to have a comment in the schema that explains why it doesn't match the API docs

mmv1/products/vertexai/IndexEndpointDeployedIndex.yaml
BBBmau: They mention minReplicaCounts and maxReplicaCounts as being the fields that are mutatable but don't specifiy which ones (automatic/dedicated), so in this case the truly only updatable fields are from `automaticResources`.

https://cloud.google.com/vertex-ai/docs/vector-search/deploy-index-vpc#mutate-deployed-index

BBBmau: This is actually something that was brought up with riley and is also mentioned in a previous comment in this PR, because it's a fine grained resource the docs won't really match with what actual process is. I took another look at the docs and they have both automatic/dedicated fields being immutable which would go against the ability to mutate the resource. This is also mentioned in this comment here: https://github.com/GoogleCloudPlatform/magic-modules/pull/11039#issuecomment-2209316648

mmv1/products/vertexai/IndexEndpointDeployedIndex.yaml
BBBmau: Refer to https://github.com/GoogleCloudPlatform/magic-modules/pull/11039#discussion_r1700394854

mmv1/products/vertexai/IndexEndpointDeployedIndex.yaml
BBBmau: Refer to https://github.com/GoogleCloudPlatform/magic-modules/pull/11039#discussion_r1700382022

mmv1/products/vertexai/IndexEndpointDeployedIndex.yaml
SarahFrench: Could you please add a comment to the yaml file linking to that comment for context about why they're missing?  Thanks

SarahFrench: Could you make it a comment and not a user-facing part of the description? The intended audience is a developer working on the provider

mmv1/products/vertexai/IndexEndpointDeployedIndex.yaml
```diff
@@ -0,0 +1,295 @@
...
+
+        A timestamp in RFC3339 UTC "Zulu" format, with nanosecond resolution and up to nine fractional digits. Examples: "2014-10-02T15:01:23Z" and "2014-10-02T15:01:23.045123456Z".
+  - !ruby/object:Api::Type::NestedObject
+    name: automaticResources
+    description: |
+      A description of resources that the DeployedIndex uses, which to large degree are decided by Vertex AI, and optionally allows only a modest additional configuration.
+
+      Having the fields within automaticResouces not being marked as immutable was done in order to support the ability to update such fields. This is discussed extensively [here](https://github.com/GoogleCloudPlatform/magic-modules/pull/11039#issuecomment-2209316648).
```
SarahFrench, line 194:
```suggestion
// Note: Having the fields within automaticResouces not being marked as immutable was done in order to support the ability to update such fields. See : https://github.com/GoogleCloudPlatform/magic-modules/pull/11039#issuecomment-2209316648
```

SarahFrench: Could be moved away from the description to avoid issues

mmv1/products/vertexai/IndexEndpointDeployedIndex.yaml
SarahFrench: It looks like this comment is from [a time in the past when a custom expand was used instead of default_from_api: true](https://github.com/GoogleCloudPlatform/magic-modules/pull/11039/commits/c655f1da00b93fc61f24ab4ba009f52e2eb929e0). Marking as resolved FYI @BBBmau

mmv1/products/vertexai/IndexEndpointDeployedIndex.yaml
```diff
@@ -0,0 +1,295 @@
...
+      A list of reserved ip ranges under the VPC network that can be used for this DeployedIndex.
+      If set, we will deploy the index within the provided ip ranges. Otherwise, the index might be deployed to any ip ranges under the provided VPC network.
+
+      The value should be the name of the address (https://cloud.google.com/compute/docs/reference/rest/v1/addresses) Example: ['vertex-ai-ip-range'].
+
+      For more information about subnets and network IP ranges, please see https://cloud.google.com/vpc/docs/subnets#manually_created_subnet_ip_ranges.
+    item_type: Api::Type::String
+  - !ruby/object:Api::Type::String
+    name: deploymentGroup
+    immutable: true
```
SarahFrench, line 290: Is deployment_group definitely immutable?

SarahFrench (commented):
Left some more comments re: ironing out the schema. Like I said before, I'd like any discrepancy between the docs and the schema to be noted in a comment for both easy of review (this is a large, complex resource) and for the benefit of people reading these files in the future

mmv1/products/vertexai/IndexEndpointDeployedIndex.yaml
SarahFrench: Resolved by https://github.com/GoogleCloudPlatform/magic-modules/pull/11039#discussion_r1707602322

mmv1/products/vertexai/IndexEndpointDeployedIndex.yaml
BBBmau: yes since when previously attempting an update on `deploymentGroup` it returns no update on the resource.

The ability to update a deployed_index is only possible on deployment_resources (i.e max_replica / min_replica)

mmv1/products/vertexai/IndexEndpointDeployedIndex.yaml
```diff
@@ -0,0 +1,295 @@
...
+            name: audiences
+            description: The list of JWT audiences. that are allowed to access. A JWT containing any of these audiences will be accepted.
+            immutable: true
+            item_type: Api::Type::String
+          - !ruby/object:Api::Type::Array
+            name: allowedIssuers
+            immutable: true
+            description: |
+              A list of allowed JWT issuers. Each entry must be a valid Google service account, in the following format: service-account-name@project-id.iam.gserviceaccount.com
+            item_type: Api::Type::String
```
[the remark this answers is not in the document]
BBBmau: This falls under the same reasoning for setting the other fields as `immutable` which is that `mutate` can only be applied on `deployment_resources`. The only way to attempt an update is to include changes to the deployment resources, otherwise we get this error:
```hcl
│ Error: Error updating IndexEndpointDeployedIndex "projects/hc-terraform-testing/locations/us-central1/indexEndpoints/5511433975622533120/deployedIndex/index_endpoint_deployed_index_dedicated_resources_test_10": googleapi: Error 400: There is no deployment resources change found for update.
```

And even when we do include the deployment_resource update along with another field (in this case `deployed_index_auth_config` it'll attempt to send a `POST` for an update:
```hcl
 ---[ REQUEST ]
 POST /v1/projects/hc-terraform-testing/locations/us-central1/indexEndpoints/5511433975622533120:mutateDeployedIndex?alt=json
 Host: us-central1-aiplatform.googleapis.com
 User-Agent: Terraform/1.8.0-alpha20240131 
 Content-Length: 527
 Accept-Encoding: gzip

 {
  "dedicatedResources": {
   "machineSpec": {
    "machineType": "n1-standard-32"
   },
   "maxReplicaCount": 8,
   "minReplicaCount": 3
  },
  "deployedIndexAuthConfig": {
   "authProvider": {
    "allowedIssuers": [
     "mauricio.alvarezleon@hashicorp.com"
    ],
    "audiences": [
     "123456-my-app"
    ]
   }
  },
  "deploymentGroup": "default",
  "displayName": "dedicated_resources_test
  "enableAccessLogging": true,
  "id": 
  "index": "projects/871647908372/locations/
  "reservedIpRanges": [
   "10.0.0.0"
  ]
 }
```

But will not perform an actual update on the resource itself. We can see this when it performs a `GET` request upon the operation returning done:
```hcl
 ---[ REQUEST ]
 GET /v1/projects/hc-terraform-testing/locations/us-central1/indexEndpoints/5511433975622533120?alt=json
 Host: us-central1-aiplatform.googleapis.com
 User-Agent: Terraform/1.8.0-alpha20240131 
 Accept-Encoding: gzip


 
 2024/08/07 15:38:54 [DEBUG] Google API 
 ---[ RESPONSE ]
 HTTP/2.0 200 OK
 Alt-Svc: h3=":443"; ma=2592000,h3-29=":443"; 
 Cache-Control: private
 Date: Wed, 07 Aug 2024 22:38:54 GMT
 Server: ESF
 Vary: Origin
 Vary: X-Origin
 Vary: Referer
 X-Frame-Options: SAMEORIGIN
 X-Xss-Protection: 0

 {
   "name": "projects/871647908372/locations/
   "displayName": "sample-endpoint",
   "description": "A sample vertex endpoint",
   "deployedIndexes": [
     {
       "id": 
       "index": "projects/871647908372/
       "displayName": 
       "createTime": "2024-07-30T20:32:09.
       "indexSyncTime": "2024-08-07T22:33:27.
       "enableAccessLogging": true,
       "reservedIpRanges": [
 "10.0.0.0"
       ],
       "deploymentGroup": "default",
       "dedicatedResources": {
         "machineSpec": {
           "machineType": "n1-standard-32"
         },
         "minReplicaCount": 3,
         "maxReplicaCount": 8
       }
     }
   ],
   "etag": 
   "labels": {
     "label-one": "value-one"
   },
   "createTime": "2024-07-30T20:31:47.
   "updateTime": "2024-07-30T20:32:04.
   "publicEndpointDomainName": "707325528.
   "encryptionSpec": {}
 }
```

It's worth noting that an update IS taking place, it just only applies to `deployment_resource` and we can see this in both requests where both `max_replica_count` and `min_replica_count` are updated where as `deployedIndexAuthConfig` is not found in the `GET` request

BBBmau:
> Left some more comments re: ironing out the schema. Like I said before, I'd like any discrepancy between the docs and the schema to be noted in a comment for both easy of review (this is a large, complex resource) and for the benefit of people reading these files in the future

A lot of the discrepancies found in this resource is due to it being a fine-grained resource, leading to documentation being marked as updatable when it isn't due to `mutateDeployIndex` being only applicable on `deployment_resources`, more info can be found here about it. https://cloud.google.com/vertex-ai/docs/vector-search/deploy-index-vpc#mutate-deployed-index.

Everything else should be marked as immutable since even if you attempted to update it it would have no affect on the resource itself. See for an example: https://github.com/GoogleCloudPlatform/magic-modules/pull/11039#discussion_r1708053308

SarahFrench (approved):
Thanks for explaining the reasoning for why some fields aren't updatable. My understanding is that the [mutateDeployedIndex](https://cloud.google.com/vertex-ai/docs/reference/rest/v1beta1/projects.locations.indexEndpoints/mutateDeployedIndex) method to edit a single deployed index (and therefore the only update method usable by this fine-grained resource) only expects and acts on changes to dedicatedResources and silently ignores attempts to update other fields (https://github.com/GoogleCloudPlatform/magic-modules/pull/11039#discussion_r1708053308). Those field may be updatable but only via the [patch method](https://cloud.google.com/vertex-ai/docs/reference/rest/v1beta1/projects.locations.indexEndpoints/patch), which is not relevant to this fine-grained resource.

I've added some final suggested comments based on this, to help newcomers to this resource gain context. Please either accept, edit or reject those comments before merging.

Besides that, I'm happy to get this merged now that the schema is 100% reviewed and as the tests cover all of the expected field updates

BBBmau:
Thanks for the help in reviewing this complicated resource! 🎉 @SarahFrench @roaks3

slocoro:
Is this module already available to use?

SarahFrench:
This new resource will be part of today's release; this PR was merged on Friday 9th, made the Wednesday release cut on the 14th, and will be released on the next Monday (today)

PR #11042, opened 2024-06-25
Added private origin authentication for cloud cdn 'compute_backend_service'

matheusaleixo-cit:
Adds support to 'aws_v4_authentication' and it's nested fields to 'compute_backend_service' resource.
Also updates the 'clientTlsPolicy' field of 'compute_backend_service' so it is set to reference the correct resource, and added relevant tests.

Fixes: https://github.com/hashicorp/terraform-provider-google/issues/18551

```release-note:enhancement
compute: added 'aws_v4_authentication' fields to 'google_compute_backend_service' resource;
```


mmv1/products/compute/BackendService.yaml
```diff
@@ -1259,16 +1259,43 @@ properties:
...
-        required: true
+        # required: true
         custom_expand: 'templates/terraform/custom_expand/resourceref_with_validation.go.erb'
       - !ruby/object:Api::Type::Array
         name: 'subjectAltNames'
         description: |
           A list of alternate names to verify the subject identity in the certificate.
           If specified, the client will verify that the server certificate's subject
           alt name matches one of the specified values.
-        required: true
```
BBBmau, line 1270: What's the reason for commenting both `required`?

BBBmau:
This test was most likely fixed in a recent commit. Can you rebase?

mmv1/products/compute/BackendService.yaml
matheusaleixo-cit: Both these fields are listed in the API docs as optional, and should not be required when using awsV4Authentication that sits in the same nested object.

BBBmau: confirmed that they are marked as optional in docs: https://cloud.google.com/compute/docs/reference/rest/v1/backendServices

In that case can you just remove the field entirely instead of commenting it out?

BBBmau:
All though these tests pass you are still missing two fields in the tests. This is seen from the recent VCR report here: https://github.com/GoogleCloudPlatform/magic-modules/pull/11042#issuecomment-2220959818

If you can include a test that also includes these two fields we should be able to get this approved and merged @matheusaleixo-cit

mmv1/third_party/terraform/services/compute/resource_compute_backend_service_test.go.erb
```diff
@@ -2010,3 +2042,63 @@ resource "google_cloud_run_service" "cloudrun_neg" {
...
+  network_endpoint_type = "INTERNET_IP_PORT"
+}
+`, randomSuffix, randomSuffix)
+}
+
+func testAccComputeBackendService_withPrivateOriginAuthUpdate(randomSuffix string) string {
+	return fmt.Sprintf(`
+resource "google_compute_backend_service" "foobar" {
+  name = "tf-test-backend-service-%s"
+  security_settings {
```
BBBmau, line 2080: Although tests are missing you are still missing two fields that must be included somewhere in the tests. This is seen by the recent VCR report: https://github.com/GoogleCloudPlatform/magic-modules/pull/11042#issuecomment-2231262931

matheusaleixo-cit: Sorry for the delay, I needed to fix some issues with the current implementation of the 'clientTlsPolicy' field. I created a new test for these fields since they cannot be used together with the 'aws_v4_authentication' fields.

BBBmau (approved):
This looks great, thanks!

PR #11044, opened 2024-06-25
Add a sleep  to reduce the likelihood of eventual consistency issue

LujieZ:
Attempt to fix https://github.com/hashicorp/terraform-provider-google/issues/17332.

```release-note:bug
resourcemanager: added a 10s wait after `google_service_account_key` creation to attempt to mitigate eventual concistency issues resulting in a `Root resource was present, but now absent` error.
```

### DIFF
--- a/mmv1/third_party/terraform/services/resourcemanager/resource_google_service_account_key.go
+++ b/mmv1/third_party/terraform/services/resourcemanager/resource_google_service_account_key.go
@@ -145,6 +145,12 @@ func resourceGoogleServiceAccountKeyCreate(d *schema.ResourceData, meta interfac
 	if err != nil {
 		return err
 	}
+
+	// We can't guarantee complete consistency even after waiting on
+	// the results, so sleep for some additional time to reduce the
+	// likelihood of eventual consistency failures.
+	time.Sleep(10 * time.Second)
+
 	return resourceGoogleServiceAccountKeyRead(d, meta)
 }
 


rileykarson (approved):
LGTM, will merge after tests assuming there are no big surprises (I don't expect any)

geekorous:
We were facing similar issues with the older versions of Terraform, and then I found this PR. Hopefully, upgrading our installer system would fix this, but just out of curiosity, what happens if one day, GCP takes 11 seconds to be eventually consistent rather than 10 seconds? The wait time seems "arbitrary" to me, no matter the value. If the consistency is only "eventual" then what is the guarantee that it takes less than 10 seconds or any number of seconds for that matter? Shouldn't we do explicit checks in the code instead? I am new to the concepts and sorry if my comment does not make sense.

rileykarson:
The issue with eventual consistency is that there is no strong guarantee of when a change will be propagated, and typically not a reliable way to determine if it has. In a strongly consistent system waiting on the operation to finish and/or reading back the result will indicate that it is effective globally. In an eventually consistent one reading back the result may not indicate it's effective or may indicate that it's partially effective.

Once a change is sent to the server it will become effective globally after some time and that time falls on a right-tailed distribution i.e. most changes will be effective relatively quickly, but upper percentiles like 99%/99.9% could be many many times the typical time- i.e. 90% of changes becoming effective in 10s, 99% in 10m, and 99.99% in 10h.

Partial effectiveness applies to users as well & not just reporting- a partially effective service account creation could mean that a service account will work in some services/regions and not others until the change has finished and become effective globally.

> Shouldn't we do explicit checks in the code instead?

Not sure about this exact case offhand, but we sometimes try to- as mentioned above, they can only indicate "maybe" and don't provide much extra signal, unfortunately!

PR #11053, opened 2024-06-26
Add plural data source `google_kms_crypto_keys`

SarahFrench:
This PR adds a new plural data source `google_kms_crypto_keys`.

Note: a weird part of this work is the need to explicitly add the `id` field to the schema for each individual key. The alternative would be to update the `google_kms_crypto_key` resource to have `id` in its schema (instead of implicitly added by the SDK) but this doesn't appear to be easily possible via MMv1.

```release-note:new-datasource
`google_kms_crypto_keys`
```

### DIFF
--- a/mmv1/third_party/terraform/services/kms/data_source_google_kms_crypto_keys.go
+++ b/mmv1/third_party/terraform/services/kms/data_source_google_kms_crypto_keys.go
@@ -65,6 +65,9 @@ func dataSourceGoogleKmsCryptoKeysRead(d *schema.ResourceData, meta interface{})
 	}
 
 	id := fmt.Sprintf("%s/cryptoKeys", keyRingId.KeyRingId())
+	if filter, ok := d.GetOk("filter"); ok {
+		id += "/filter=" + filter.(string)
+	}
 	d.SetId(id)
 
 	log.Printf("[DEBUG] Searching for keys in key ring %s", keyRingId.KeyRingId())
--- a/mmv1/third_party/terraform/services/kms/data_source_google_kms_crypto_keys.go
+++ b/mmv1/third_party/terraform/services/kms/data_source_google_kms_crypto_keys.go
@@ -141,6 +141,8 @@ func dataSourceKMSCryptoKeysList(d *schema.ResourceData, meta interface{}, keyRi
 			RawURL:    url,
 			UserAgent: userAgent,
 			Headers:   headers,
+			// ErrorRetryPredicates used to allow retrying if rate limits are hit when requesting multiple pages in a row
+			ErrorRetryPredicates: []transport_tpg.RetryErrorPredicateFunc{transport_tpg.IsRetryable429QuotaError},
 		})
 		if err != nil {
 			return nil, transport_tpg.HandleNotFoundError(err, d, fmt.Sprintf("KMSCryptoKeys %q", d.Id()))
--- a/mmv1/third_party/terraform/services/kms/data_source_google_kms_crypto_keys.go
+++ b/mmv1/third_party/terraform/services/kms/data_source_google_kms_crypto_keys.go
@@ -1,0 +1,171 @@
+package kms
+
+import (
+	"fmt"
+	"log"
+	"net/http"
+	"regexp"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"github.com/hashicorp/terraform-provider-google/google/tpgresource"
+	transport_tpg "github.com/hashicorp/terraform-provider-google/google/transport"
+)
+
+func DataSourceGoogleKmsCryptoKeys() *schema.Resource {
+	dsSchema := tpgresource.DatasourceSchemaFromResourceSchema(ResourceKMSCryptoKey().Schema)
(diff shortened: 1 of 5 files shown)


mmv1/third_party/terraform/services/kms/data_source_google_kms_crypto_keys.go
```diff
@@ -0,0 +1,178 @@
...
+		billingProject = parts[1]
+	}
+
+	// err == nil indicates that the billing_project value was found
+	if bp, err := tpgresource.GetBillingProject(d, config); err == nil {
+		billingProject = bp
+	}
+
+	headers := make(http.Header)
+	res, err := transport_tpg.SendRequest(transport_tpg.SendRequestOptions{
```
BBBmau, line 126: This all looks good, my only concern is that it's not using `pageToken` to handle large lists of crypto keys. I can understand why this wasn't added though since it handles a large amount before needing to use the `nextPageToken` param. Regardless it would be worth adding support for `pageToken` to prevent any future panics for users.

Example of it being used:
https://github.com/GoogleCloudPlatform/magic-modules/blob/7381c52e624297d4328b71d0a0948f44f9d87dec/mmv1/third_party/terraform/services/storage/data_source_google_storage_buckets.go#L68-L104

mmv1/third_party/terraform/services/kms/data_source_google_kms_crypto_keys.go
```diff
@@ -0,0 +1,178 @@
...
+
+func dataSourceGoogleKmsCryptoKeysRead(d *schema.ResourceData, meta interface{}) error {
+	config := meta.(*transport_tpg.Config)
+
+	keyRingId, err := parseKmsKeyRingId(d.Get("key_ring").(string), config)
+	if err != nil {
+		return err
+	}
+
+	id := fmt.Sprintf("%s/cryptoKeys", keyRingId.KeyRingId())
```
BBBmau, line 67: we may want to also include the filter that's used for cryptokeys also since i can imagine users using this data source multiple times but with different filters. An example of this can be seen here:
https://github.com/GoogleCloudPlatform/magic-modules/blob/7381c52e624297d4328b71d0a0948f44f9d87dec/mmv1/third_party/terraform/services/secretmanager/data_source_secret_manager_secrets.go#L122-L129

mmv1/third_party/terraform/services/kms/data_source_google_kms_crypto_keys.go
BBBmau: I'm unsure about this now since I ran into a quota limit when implementing on `key_rings`

https://github.com/GoogleCloudPlatform/magic-modules/blob/7381c52e624297d4328b71d0a0948f44f9d87dec/mmv1/third_party/terraform/services/secretmanager/data_source_secret_manager_secrets.go#L122-L129

mmv1/third_party/terraform/services/kms/data_source_google_kms_crypto_keys.go
SarahFrench: Ah good point! I'll add that now

SarahFrench: Done in [ae575c8](https://github.com/GoogleCloudPlatform/magic-modules/pull/11053/commits/ae575c88bf63251834faa0b8a2870978ace1fbe1)

mmv1/third_party/terraform/services/kms/data_source_google_kms_crypto_keys.go
BBBmau: PR is ready to be merged. I would want to know whether it's a good idea to include pageToken support or not. Their would be some work needed on the tests in order to support it. If we're fine with not adding support i can approve this PR.

mmv1/third_party/terraform/services/kms/data_source_google_kms_crypto_keys.go
```diff
@@ -0,0 +1,203 @@
...
+
+	// Always include the filter param, and optionally include the pageToken parameter for subsequent requests
+	var params = make(map[string]string, 0)
+	if filter, ok := d.GetOk("filter"); ok {
+		log.Printf("[DEBUG] Search for keys in key ring %s is using filter ?filter=%s", keyRingId, filter.(string))
+		params["filter"] = filter.(string)
+	}
+
+	cryptoKeys := make([]interface{}, 0)
+	for {
```
SarahFrench, line 129: Given there's nothing to slow down this loop, there could be quota issues with this implementation. Waiting to see if this (https://github.com/GoogleCloudPlatform/magic-modules/pull/11074/files#r1657874219) helps or not.

SarahFrench: See: https://github.com/GoogleCloudPlatform/magic-modules/pull/11074#issuecomment-2195862215

SarahFrench:
The new version of the 429 retry that checks the 429 error is due to a rate issue (versus number of resources type of quotas) works here (https://github.com/GoogleCloudPlatform/magic-modules/pull/11074#issuecomment-2197472900) so I'm happy to merge this PR.

PR #11057, opened 2024-06-26
Update destroy_scheduled_duration default value description

tdbhacks:
The field's default value has been updated as part of https://cloud.google.com/kms/docs/key-destruction-msa

Looking for guidance on wether we need some sort of release note or not.

```release-note:none

```


slevenick (approved):
We don't need a changelog note for a docs-only change

PR #11061, opened 2024-06-26
`KMS`: add plural data source for `resource_google_kms_key_ring`

BBBmau:
This PR adds a new plural data source `google_kms_key_rings`.

Note:
- It's necessary to explicitly add the `id` field to the schema for each individual key ring. Similar behavior also appears here: https://github.com/GoogleCloudPlatform/magic-modules/pull/11053
- One attribute is missing which is the `createTime`. Only reason why this wasn't ended in this PR is due the attribute not being included in `resource_google_kms_key_ring.go`

Running locally:
```
└─(16:10:49 on main ✹ ✭)──> envchain GCLOUD make testacc TEST=./google/services/kms TESTARGS='-run TestAccDataSourceGoogleKmsKeyRings_basic'                                                     2 ↵ ──(Wed,Jun26)─┘
TF_ACC=1 TF_SCHEMA_PANIC_ON_ERROR=1 go test ./google/services/kms -v -run TestAccDataSourceGoogleKmsKeyRings_basic -timeout 240m -ldflags="-X=github.com/hashicorp/terraform-provider-google/version.ProviderVersion=acc"
=== RUN   TestAccDataSourceGoogleKmsKeyRings_basic
2024/06/26 16:11:06 [INFO] Authenticating using configured Google JSON 'credentials'...
2024/06/26 16:11:06 [INFO]   -- Scopes: [https://www.googleapis.com/auth/cloud-platform https://www.googleapis.com/auth/userinfo.email]
2024/06/26 16:11:06 [INFO] Authenticating using configured Google JSON 'credentials'...
2024/06/26 16:11:06 [INFO]   -- Scopes: [https://www.googleapis.com/auth/cloud-platform https://www.googleapis.com/auth/userinfo.email]
2024/06/26 16:11:06 [DEBUG] Waiting for state to become: [success]
2024/06/26 16:11:06 [INFO] Terraform is using this identity: mauricio-alvarezleon@hc-terraform-testing.iam.gserviceaccount.com
2024/06/26 16:11:06 [INFO] Instantiating Google Cloud KMS client for path https://cloudkms.googleapis.com/
2024/06/26 16:11:06 [DEBUG] Retry Transport: starting RoundTrip retry loop
2024/06/26 16:11:06 [DEBUG] Retry Transport: request attempt 0
2024/06/26 16:11:07 [DEBUG] Retry Transport: Stopping retries, last request was successful
2024/06/26 16:11:07 [DEBUG] Retry Transport: Returning after 1 attempts
2024/06/26 16:11:07 [DEBUG] Retry Transport: starting RoundTrip retry loop
2024/06/26 16:11:07 [DEBUG] Retry Transport: request attempt 0
2024/06/26 16:11:07 [DEBUG] Retry Transport: Stopping retries, last request was successful
2024/06/26 16:11:07 [DEBUG] Retry Transport: Returning after 1 attempts
--- PASS: TestAccDataSourceGoogleKmsKeyRings_basic (25.72s)
PASS
ok      github.com/hashicorp/terraform-provider-google/google/services/kms      26.964s
```

```release-note:new-datasource
`google_kms_key_rings`
```

### DIFF
--- a/mmv1/third_party/terraform/services/kms/data_source_google_kms_key_rings.go
+++ b/mmv1/third_party/terraform/services/kms/data_source_google_kms_key_rings.go
@@ -1,5 +1,3 @@
-// Copyright (c) HashiCorp, Inc.
-// SPDX-License-Identifier: MPL-2.0
 package kms
 
 import (
--- a/mmv1/third_party/terraform/services/kms/data_source_google_kms_key_rings.go
+++ b/mmv1/third_party/terraform/services/kms/data_source_google_kms_key_rings.go
@@ -1,0 +1,171 @@
+// Copyright (c) HashiCorp, Inc.
+// SPDX-License-Identifier: MPL-2.0
+package kms
+
+import (
+	"fmt"
+	"log"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"github.com/hashicorp/terraform-provider-google/google/tpgresource"
+	transport_tpg "github.com/hashicorp/terraform-provider-google/google/transport"
+)
+
+func DataSourceGoogleKmsKeyRings() *schema.Resource {
+	return &schema.Resource{
+		Read: dataSourceGoogleKmsKeyRingsRead,
+		Schema: map[string]*schema.Schema{
+			"project": {
+				Type:        schema.TypeString,
+				Optional:    true,
+				Description: `Project ID of the project.`,
+			},
+			"location": {
+				Type:        schema.TypeString,
+				Required:    true,
+				Description: `The canonical id for the location. For example: "us-east1".`,
+			},
+			"filter": {
+				Type:     schema.TypeString,
+				Optional: true,
+				Description: `
+					The filter argument is used to add a filter query parameter that limits which keys are retrieved by the data source: ?filter={{filter}}.
+					Example values:
+					
+					* "name:my-key-" will retrieve key rings that contain "my-key-" anywhere in their name. Note: names take the form projects/{{project}}/locations/{{location}}/keyRings/{{keyRing}}.
+					* "name=projects/my-project/locations/global/keyRings/my-key-ring" will only retrieve a key ring with that exact name.
+					
+					[See the documentation about using filters](https://cloud.google.com/kms/docs/sorting-and-filtering)
+				`,
+			},
+			"key_rings": {
+				Type:        schema.TypeList,
+				Computed:    true,
+				Description: "A list of all the retrieved key rings",
+				Elem: &schema.Resource{
+					// schema isn't used from resource_kms_key_ring due to having project and location fields which are empty when grabbed in a list.
+					Schema: map[string]*schema.Schema{
+						"id": {
+							Type:     schema.TypeString,
+							Computed: true,
+						},
+						"name": {
+							Type:     schema.TypeString,
+							Computed: true,
+						},
+					},
+				},
+			},
+		},
+	}
+}
+
+func dataSourceGoogleKmsKeyRingsRead(d *schema.ResourceData, meta interface{}) error {
+	config := meta.(*transport_tpg.Config)
+
+	id, err := tpgresource.ReplaceVars(d, config, "projects/{{project}}/locations/{{location}}/keyRings")
+	if err != nil {
+		return err
+	}
+	d.SetId(id)
+
+	log.Printf("[DEBUG] Searching for keyrings")
+	userAgent, err := tpgresource.GenerateUserAgentString(d, config.UserAgent)
+	if err != nil {
+		return err
+	}
+
+	billingProject := ""
+
+	project, err := tpgresource.GetProject(d, config)
+	if err != nil {
+		return fmt.Errorf("Error fetching project for keyRings: %s", err)
+	}
+	billingProject = project
+
+	// err == nil indicates that the billing_project value was found
+	if bp, err := tpgresource.GetBillingProject(d, config); err == nil {
+		billingProject = bp
+	}
+
+	var keyRings []interface{}
+
+	params := make(map[string]string)
+	if filter, ok := d.GetOk("filter"); ok {
+		log.Printf("[DEBUG] Search for key rings using filter ?filter=%s", filter.(string))
+		params["filter"] = filter.(string)
+		if err != nil {
+			return err
+		}
+	}
+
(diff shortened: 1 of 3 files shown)


mmv1/third_party/terraform/services/kms/data_source_google_kms_key_rings.go
```diff
@@ -0,0 +1,174 @@
+// Copyright (c) HashiCorp, Inc.
+// SPDX-License-Identifier: MPL-2.0
```
SarahFrench, line 2:
```suggestion
```

mmv1/third_party/terraform/services/kms/data_source_google_kms_key_rings.go
```diff
@@ -0,0 +1,174 @@
...
+					
+					[See the documentation about using filters](https://cloud.google.com/kms/docs/sorting-and-filtering)
+				`,
+			},
+			"key_rings": {
+				Type:        schema.TypeList,
+				Computed:    true,
+				Description: "A list of all the retrieved key rings",
+				Elem: &schema.Resource{
+					// schema isn't used from resource_kms_key_ring due to having project and location fields which are empty when grabbed in a list.
```
SarahFrench, line 47: You can pass those (known) values into your flattener `flattenKMSKeyRingsList` and set the project and location on each item in the list, and that way you can have rich data for each key ring being returned. What do you think?

BBBmau: I attempted this but still got empty `project, location` values for each keyring. I can attempt this again assuming i did something wrong.

BBBmau: > You can pass those (known) values into your flattener `flattenKMSKeyRingsList` and set the project and location on each item in the list, and that way you can have rich data for each key ring being returned. What do you think?

I was thinking about this also but it seems like extra info that's not really useful since `project` and `location` must be set in order to use the data source. Would show a list of keyRings with all the same locations and projects that were initially set.

An example of what this would look like if a user were to set both `project` and `location` fields while also including them in each `keyRing`
```hcl
.....
          + {
              + id       = "projects/hc-terraform-testing/locations/global/keyRings/tf-test-keyring-examplekipwkpbmlh"
              + location = "global"
              + name     = "tf-test-keyring-examplekipwkpbmlh"
              + project  = "hc-terraform-testing"
            },
          + {
              + id       = "projects/hc-terraform-testing/locations/global/keyRings/tf-test-keyring-exampleksrf4q2e41"
              + location = "global"
              + name     = "tf-test-keyring-exampleksrf4q2e41"
              + project  = "hc-terraform-testing"
            },
          + {
              + id       = "projects/hc-terraform-testing/locations/global/keyRings/tf-test-keyring-exampleml7ke81gcw"
              + location = "global"
              + name     = "tf-test-keyring-exampleml7ke81gcw"
              + project  = "hc-terraform-testing"
            },
        ]
      + location  = "global"
      + project   = "hc-terraform-testing"
 ```

Comment (SarahFrench) [reply mmv1/third_party/terraform/services/kms/data_source_google_kms_key_rings.go:45]: @BBBmau Oh wait, I can see that the API response when listing crypto key rings [only includes names and createTime values](https://cloud.google.com/kms/docs/reference/rest/v1/projects.locations.keyRings#KeyRing), so there isn't other data to put in the schema of each list item. In that case keeping just name and createTime is fine!

Comment (SarahFrench) [reply mmv1/third_party/terraform/services/kms/data_source_google_kms_key_rings.go:45]: I just realised that the API only returns name and id, and the crypto key ring resource doesn't have additional fields to fill. In that case I think it's fine to only include id, name, and createTime using a 'custom schema'. I was thinking we should include project and location values if the crypto key ring resource's schema was more substantial

Comment (SarahFrench) [reply mmv1/third_party/terraform/services/kms/data_source_google_kms_key_rings.go:45]: Agreed that adding location and project to each entry doesn't make sense in that case.

**Comment by BBBmau** [comment]
latest commit adds `filter` into id and also sets the project into state. project wouldn't be set when left empty.

The following is the output as well as the tfconfig that produces it:

`tfconfig`
```hcl
data "google_kms_key_rings" "mau" {
  location = "global"
}
```
`output`:
```hcl
...
    {
      "id" = "projects/hc-terraform-testing/locations/global/keyRings/tf-test-keyring-exampleksrf4q2e41"
      "name" = "tf-test-keyring-exampleksrf4q2e41"
    },
    {
      "id" = "projects/hc-terraform-testing/locations/global/keyRings/tf-test-keyring-exampleml7ke81gcw"
      "name" = "tf-test-keyring-exampleml7ke81gcw"
    },
  ])
  "location" = "global"
  "project" = "hc-terraform-testing"
}
```

@SarahFrench should be ready after VCR tests pass 👍🏼

mmv1/third_party/terraform/services/kms/data_source_google_kms_key_rings.go
BBBmau: what's the process for this? I see some data sources have it and some dont. we should clean this up in the future.

https://github.com/GoogleCloudPlatform/magic-modules/blob/f514cc2e5b57129f3136b17ce602aa724fa390c4/mmv1/third_party/terraform/services/storage/data_source_google_storage_buckets.go#L1-L5

BBBmau:
recent commit adds the recent pushes from https://github.com/GoogleCloudPlatform/magic-modules/pull/11053 which allows the 429 error to be handled.

mmv1/third_party/terraform/services/kms/data_source_google_kms_key_rings.go
```diff
@@ -0,0 +1,175 @@
+// Copyright (c) HashiCorp, Inc.
+// SPDX-License-Identifier: MPL-2.0
```
SarahFrench, line 2:
```suggestion
```

mmv1/third_party/terraform/services/kms/data_source_google_kms_key_rings_test.go
```diff
@@ -0,0 +1,75 @@
...
+	return acctest.Nprintf(`
+data "google_kms_key_rings" "all_key_rings" {
+  location = "%{location}"
+  %{filter}
+}
+data "google_project" "project" {
+}
+output "project_number" {
+  value = data.google_project.project.number
+}
```
SarahFrench, line 73:
```suggestion
```

SarahFrench (approved):
Approved, but please accept the suggestions and allow tests to finish before merging!

PR #11065, opened 2024-06-27
Fix docs link to predefined metadata keys

serpro69:
The link to default metadata keys seems to have changed since the old one leads to 404.

```release-note:none

```


roaks3 (approved):
Thanks for catching this and fixing!

PR #11067, opened 2024-06-27
Add data clean rooms support 

obada-ab:
API: 
* https://cloud.google.com/bigquery/docs/reference/analytics-hub/rest/v1/projects.locations.dataExchanges
* https://cloud.google.com/bigquery/docs/reference/analytics-hub/rest/v1/projects.locations.dataExchanges.listings

Guide: https://cloud.google.com/bigquery/docs/data-clean-rooms

Reference: b/347844459

```release-note:enhancement
bigqueryanalyticshub: added `sharing_environment_config` to `google_bigquery_analytics_hub_data_exchange`
```

```release-note:enhancement
bigqueryanalyticshub: added `selected_resources`, and `restrict_direct_table_access` to `google_bigquery_analytics_hub_listing`
```

```release-note:breaking-change
bigqueryanalyticshub: made `bigquery_dataset` immutable in `google_bigquery_analytics_hub_listing`. Now modifying the field will recreate the resource rather than causing Terraform to report it would attempt an invalid update.
```


shuyama1 (commented):
Hey @obada-ab, I assume the PR is still WIP, so I added the `disable-review-reminders` label to avoid generating noise to this PR. Please feel free to let me know if you have any questions regarding the change in this PR/ when the PR is ready for review. Thanks!

mmv1/products/bigqueryanalyticshub/Listing.yaml
```diff
@@ -151,6 +164,7 @@ properties:
     name: bigqueryDataset
     description: Shared dataset i.e. BigQuery dataset source.
     required: true
+    immutable: true
```
shuyama1, line 167: Do you happen to know what's current behavior if a user is trying to update this field? Will it error out?

Making the field immutable may break users depends on how the update of the field is handled currently.

obada-ab: I tried updating the field before making this change and got an error:
```
Error: Error updating Listing "projects/X/locations/US/dataExchanges/Y/listings/Z": googleapi: Error 400: The field 'bigquery_dataset' cannot be updated on this listing.
```

shuyama1: Thanks for confirming! Then, it should be safe to make this change in a minor release as update is not currently supported for this field and will run into API errors in all cases, according to https://googlecloudplatform.github.io/magic-modules/develop/breaking-changes/make-a-breaking-change/#in-minor-releases

sachinpro (approved):

shuyama1 (commented):
Mostly LGTM. got one question though - Do any of these fields updatable? I notice some immutable flags are on the parent fields but not on the subfields, so I just want to confirm. If any of these are updatable, would you mind following https://googlecloudplatform.github.io/magic-modules/develop/test/test/#add-an-update-test to either add a new test or modify an existing one to cover these cases?Thanks!

obada-ab:
@shuyama1 
> Do any of these fields updatable? I notice some immutable flags are on the parent fields but not on the subfields, so I just want to confirm.

All the added fields are either immutable or output only. I believe it's enough to have the immutable flag on the parent fields, because all subfields will also be immutable ([reference](https://googlecloudplatform.github.io/magic-modules/develop/field-reference/#immutable))

shuyama1 (commented):
> All the added fields are either immutable or output only. I believe it's enough to have the immutable flag on the parent fields, because all subfields will also be immutable ([reference](https://googlecloudplatform.github.io/magic-modules/develop/field-reference/#immutable))

Thanks for confirm that all the fields are immutable. I just had a discussion with my team and we agree that the documentation needs to be updated - Each immutable subfield must have the immutable attribute explicitly added, even if the parent field is already marked as immutable. 

Would you mind adding `immutable:true` to all the immutable subfields as well? Apologize for the confusion in the doc and thanks for working on this.

shuyama1:
`terraform-google-conversion-build-and-unit-tests` failure is not related to the change in this PR. Please ignore it and sorry for the noise.

shuyama1 (approved):
Thank you!

PR #11081, opened 2024-06-28
Add custom update to update metadata in a separate call

JoyceYingZhu:
This pull request make a custom 'updateIndex' method. Since if index.metadata need to update, not other index fields can be also updated as part of the same update call. So we need to make a separate update call for index.metadata. 

```release-note:bug
index_update: fixed pre-update use unsupported nested updatemask "metadata.contentsDeltaUri". Use first layer update mask and update "metadata" field in a separate call.
```


JoyceYingZhu:
Please take a look review the change.

BBBmau (approved):
Passing since the existing tests are passing. All looks good!

PR #11084, opened 2024-06-28
Use sweepable name for vpc access connector

roaks3:
This should prevent these resources (and a variety of other dependent resources) from accumulating in the test instances

```release-note:none

```


zli82016 (approved):
LGTM

PR #11090, opened 2024-07-01
Add pre-read for importing log view from previous versions

trodge:
This fixes an issue with importing log views created with the DCL-based provider.

```release-note:none

```


slevenick (commented):
Is there a test that would have caught this? Can you add one if so?

trodge:
> Is there a test that would have caught this? Can you add one if so?

I can't think of any test that would catch this. `RELEASE_DIFF` doesn't catch it as far as I can tell, and the only way to test is to create the resource with the older version of the provider and then apply it with the new one.

slevenick (approved):
LGTM assuming you've tested the flow manually

PR #11098, opened 2024-07-02
Add send_age_if_zero field and deprecate no_age.

kautikdk:
This PR introduces new field `send_age_if_zero` which is similar to other LifecycleRuleCondition virtual fields. The intention of introducing new field is to have common virtual field logic for all the fields and deprecation of no_age field which will be removed in upcoming major release. 

This Breaking change is expected and will only effect customers who are using lifecycle rules in terraform configuration.

Reference Issue link:https://github.com/hashicorp/terraform-provider-google/issues/18232

```release-note:breaking-change
storage: added `send_age_if_zero` field with default value `true` and deprecated `no_age` field. 
```


kautikdk:
Hi @NickElliot, Created this PR to deprecate `no_age` attribute and introduce new `send_age_if_zero` attribute with default value of `true`. This will be first step towards removal of `no_age` attribute in upcoming major release 6.0.0 . 
The reason for introducing this attribute with default value of `true` is to make this change backward compatible as we won't be sending `age` 0 value without `send_age_if_zero` being true. 
cc: @rileykarson

kautikdk:
Making PR draft while fixing the test failure.

kautikdk:
Hi @NickElliot, Before we move to review, I have something to discuss. 
Usually Virtual Fields are being introduced at the top level of the resource schema but Here we are introducing field with default value inside nested blocks. Because of this special case, We can not set this field value to it's default in import function. This may result in a drift while importing bucket with lifecycle configuration like it will generate diff from `false->true`. I don't think It will be too problematic for user to replace lifecycleRuleCondition but still we can consider it before introducing this field. What are your views on this?

kautikdk:
Hi @NickElliot, Can you do a initial review of the PR? So that I can do require modifications before EOW. Thanks in advance!

NickElliot:
Hi, sorry for the delay, we were out of the office on Thursday and Friday.

I'm trying to figure out if there is an elegant solution for the import issue, and will aim to get in a full review by EOD tomorrow.

kautikdk:
@NickElliot Any Updates?

NickElliot (commented):
Think I found a solution: 
In flattenBucketLifecycleRuleCondition, for `send_age_if_zero`, set the value to be the inverse of `no_age`.

If the user has not set `no_age` or made it be false, then we want `send_age_if_zero` to be true. If they have set it to be true, then `send_age_if_zero` will be false. This aligns with the other functionality in that `no_age` has precedence, so if users have a conflicting value in their local state it will cause issues.

To that end, it would be worth adding an error to update if users *do* have conflicting values in their state on attempted create or update, i.e. double true or double false.

kautikdk:
Well, We might be prioritizing `no_age` to set value of `send_age_if_zero` incase of updating state file but I am having concerns about throwing error in case of same values for both attributes. 
We introduced `no_age` in last December because of an critical customer issue and after that many users started using`no_age = true` in their config. Now we are introducing a new attribute with a default and it will conflict with already existing attribute and make their config invalid. I think It will be too much friction from the customer side.

NickElliot (commented):
I didn't think about that situation fully, good catch. 

I think if the behavior on import is implemented with no hard-error, and we give precedence to `send_age_if_zero` things should work out. 

What we want to have covered is that users with intentional `no_age = true` are told to set `send_age_if_zero` to false and remove `no_age` from their config files. This will be adding an ultimately meaningless line to their files, but they can clear their deprecation message for the field that will be totally gone, and will then experience no issues through the 6.0.0 upgrade.

The hard part is users who have nothing in their states today need to be warned somehow: I think the best option is we can add a separate CLI output warning message on attempted create or update if the state is "no_age is false/empty, send age is true", along the lines of:

> in an upcoming major release the default behavior of this resource will be changed and the field `no_age` will be deprecated, to avoid interruptions in upgrade, add the following to your life cycle conditions `send_age_if_zero=true`. If you have already done so, please disregard this message.

This will add some noise to users, but will not interrupt most plan refreshes and unfortunately we have to add this noise on at least one end of the system, as at least one end-configuration involves a conflict between default/false/empty values whether we go with no default or yes default.

I'll verify tomorrow how the deprecation message is displayed to users and get back to you on the final decision for what path I'll say to go with.

NickElliot:
@kautikdk we can go ahead without implementing the seperate CLI warning message for users who have no values in their state, the warning in patch notes/documentation will suffice. we only need to worry about the deprecation notice for the people who explicitly declare a `no_age` value.

kautikdk:
@NickElliot, I believe this little fix should do the work. Checking weather lifecycle rule is present in the state file or not, If it is then we will directly take the VF values from the state file otherwise we will set it to the defaults. This fixes the import issue because I think there is no case where we are getting empty lifecycle condition when it is not actually there in the state file like primitive data types. Initially I thought TypeSet is also facing empty/null value issue like primitive data type but checked locally and it is working as expected.

NickElliot (approved):
alls well that ends well! thanks for the work

NickElliot (commented):
one thing -- could you add in a description to `mmv1/third_party/terraform/website/docs/guides/version_6_upgrade.html.markdown`?

kautikdk:
@NickElliot Done!

mmv1/third_party/terraform/website/docs/guides/version_6_upgrade.html.markdown
```diff
@@ -126,3 +126,11 @@ An empty value means the setting should be cleared.
...
 Cloud Run does not provide a default value for liveness probe. Now removing this field
 will remove the liveness probe from the Cloud Run service.
+
+## Resource: `google_storage_bucket`
+
+### `lifecycle_rule.condition.no_age` is now removed
+
+Previously `lifecycle_rule.condition.age` attirbute was being set zero value by default and `lifecycle_rule.condition.no_age` was introduced to prevent that.
+Now `lifecycle_rule.condition.no_age` is no longer supported and `lifecycle_rule.condition.age` won't set zero value.
+Removed in favor of the field `lifecycle_rule.condition.send_age_if_zero` which can be used to set zero value for `lifecycle_rule.condition.age` attribute.
```
NickElliot, line 136:
```suggestion
Previously `lifecycle_rule.condition.age` attirbute was being set zero value by default and `lifecycle_rule.condition.no_age` was introduced to prevent that.
Now `lifecycle_rule.condition.no_age` is no longer supported and `lifecycle_rule.condition.age` won't set a zero value by default.
Removed in favor of the field `lifecycle_rule.condition.send_age_if_zero` which can be used to set zero value for `lifecycle_rule.condition.age` attribute. 

For a seamless update, if your state today uses `no_age=true`, update it to remove `no_age` and set `send_age_if_zero=false`. If you do not use `no_age=true`, you will need to add `send_age_if_zero=true` to your state to avoid any changes after updating to 6.0.0. 
```

PR #11100, opened 2024-07-02
Add buildServiceAccount field to cloudfunctions

nifflets:
Adding support for the buildServiceAccount field in cloudfunctions as it has recently been [released to GA](https://cloud.google.com/functions/docs/release-notes#June_28_2024).

```release-note:enhancement
cloudfunctions: added `build_service_account` field to `google_cloudfunctions_function` resource
```

### DIFF
--- a/mmv1/products/cloudfunctions/CloudFunction.yaml
+++ b/mmv1/products/cloudfunctions/CloudFunction.yaml
@@ -119,6 +119,10 @@ properties:
     output: true
     description: 'The email of the service account for this function.'
   - !ruby/object:Api::Type::String
+    name: 'buildServiceAccount'
+    default_from_api: true
+    description: 'The fully-qualified name of the service account to be used for the build step of deploying this function'
+  - !ruby/object:Api::Type::String
     name: 'updateTime'
     output: true
     description: 'The last update timestamp of a Cloud Function'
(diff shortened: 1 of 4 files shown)


mmv1/products/cloudfunctions/CloudFunction.yaml
```diff
@@ -118,6 +118,10 @@ properties:
     name: 'serviceAccountEmail'
     output: true
     description: 'The email of the service account for this function.'
+  - !ruby/object:Api::Type::String
+    name: 'buildServiceAccount'
```
c2thorn, line 122: This YAML file is [excluded from generation](https://github.com/GoogleCloudPlatform/magic-modules/pull/11100/files#diff-6330766290dce11fbe6f233175b8cf216f13e9eafb28c75675d398b4e359c7d0R47), I believe it is just around for IAM generation. Either way, adding this doesn't actually do anything but also doesn't really hurt.

c2thorn (changes requested):
looks good, just need to add the field to https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/third_party/terraform/website/docs/r/cloudfunctions_function.html.markdown

mmv1/products/cloudfunctions/CloudFunction.yaml
nifflets: acked

nifflets:
> looks good, just need to add the field to https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/third_party/terraform/website/docs/r/cloudfunctions_function.html.markdown

Added!

c2thorn (approved):

PR #11101, opened 2024-07-02
Add validation to prevent BigQuery view from creation when input schema contains required fields

wj-chen:
Fixes https://github.com/hashicorp/terraform-provider-google/issues/14220.

Old behavior: Create API is called to first create the view without schema, then Update API is called to apply the schema to the view, at which point it will result in an error if the schema contains any required fields. The new table is left created with fields being nullable. Subsequent `apply` will result in permadiff.

New behavior: No new table is created if view is specified while schema contains required fields.

```release-note:breaking-change
bigquery: added validation to prevent table view creation if schema contains required fields for `google_bigquery_table` resource
```

### DIFF
--- a/mmv1/third_party/terraform/services/bigquery/resource_bigquery_table.go
+++ b/mmv1/third_party/terraform/services/bigquery/resource_bigquery_table.go
@@ -1591,6 +1591,9 @@ func resourceBigQueryTableCreate(d *schema.ResourceData, meta interface{}) error
 	}
 
 	if table.View != nil && table.Schema != nil {
+		if schemaHasRequiredFields(table.Schema) {
+			return errors.New("Schema cannot contain required fields when creating a view")
+		}
 
 		log.Printf("[INFO] Removing schema from table definition because BigQuery does not support setting schema on view creation")
 		schemaBack := table.Schema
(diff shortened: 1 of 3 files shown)


hao-nan-li:
@sachinpro Do you plan to do a first round of review on this PR?

hao-nan-li (commented):
Looks like there's a git conflict. Let's resolve it and start the review.

wj-chen:
Looks like the build error is for `google_vpc_access_connector`. I wonder if other PRs against this feature branch is seeing the same issue.

hao-nan-li:
> Looks like the build error is for `google_vpc_access_connector`. I wonder if other PRs against this feature branch are seeing the same issue.

I don't think it's related to your change, it was introduced when you tried to resolve the merge conflict. Will check on other PRs.

hao-nan-li:
The fix is merged.

wj-chen:
> The fix is merged.

I merged from upstream but I think that commit might still have issues...

hao-nan-li:
> > The fix is merged.
> 
> I merged from upstream but I think that commit might still have issues...

Yeah, looks like another TGC test failed... Will sync

hao-nan-li:
Could you solve the conflicts and let's try to run the test again?

wj-chen:
> Could you solve the conflicts and let's try to run the test again?

Resolved the conflict but build still seems to be failing for the same error. Can you advise?

hao-nan-li:
> > > Could you solve the conflicts and let's try to run the test again?
> > 
> > 
> > Resolved the conflict but build still seems to be failing for the same error. Can you advise?
> 
> It's the same failure in #11117, the latest commit in https://github.com/GoogleCloudPlatform/magic-modules/tree/FEATURE-BRANCH-major-release-6.0.0.

Correct, this has been escalated and we are looking for a solution to it.

hao-nan-li:
I double-checked with other PRs with similar issues, is your branch based on `FEATURE-BRANCH-major-release-6.0.0`? https://googlecloudplatform.github.io/magic-modules/develop/breaking-changes/make-a-breaking-change/#make-the-change-on-feature-branch-major-release-600

wj-chen:
> I double-checked with other PRs with similar issues, should rebase now as #11179 is merged.

Thanks for the update. Merged and updated the PR and it should be ready for review again.

hao-nan-li (approved):

PR #5162, opened 2021-09-01
Add support for ILB as next-hop with tags

juliocc:
Add Terraform Support for ILB as next hop with tags

Fixes https://github.com/hashicorp/terraform-provider-google/issues/9833

One test is failing in the GA provider ([TestAccComputeRoute_hopInstance](https://github.com/GoogleCloudPlatform/magic-modules/blob/master/mmv1/third_party/terraform/tests/resource_compute_route_test.go#L30)) but I'm not sure it's realated to this change.

If this PR is for Terraform, I acknowledge that I have:

- [X] Searched through the [issue tracker](https://github.com/hashicorp/terraform-provider-google/issues) for an open issue that this either resolves or contributes to, commented on it to claim it, and written "fixes {url}" or "part of {url}" in this PR description. If there were no relevant open issues, I opened one and commented that I would like to work on it (not necessary for very small changes).
- [X] [Generated Terraform](https://github.com/GoogleCloudPlatform/magic-modules#generating-downstream-tools), and ran `make test` and `make lint` to ensure it passes unit and linter tests.
- [X] Ensured that all new fields I added that can be set by a user appear in at least one [example](https://github.com/GoogleCloudPlatform/magic-modules/tree/master/templates/terraform/examples) (for generated resources) or [third_party test](https://github.com/GoogleCloudPlatform/magic-modules/tree/master/third_party/terraform/tests) (for handwritten resources or update tests).
- [x] [Ran](https://github.com/hashicorp/terraform-provider-google/blob/master/.github/CONTRIBUTING.md#tests) relevant acceptance tests (If the acceptance tests do not yet pass or you are unable to run them, please let your reviewer know).
- [X] Read the [Release Notes Guide](https://github.com/GoogleCloudPlatform/magic-modules/blob/master/.ci/RELEASE_NOTES_GUIDE.md) before writing my release note below.

**Release Note Template for Downstream PRs (will be copied)**

```release-note:enhancement
compute: allow passing an IP address to the `nextHopIlb` field of `google_compute_route` resource
```

PR #1772, opened 2019-05-15
Sql database port

tysen:
-----------------------------------------------------------------
# [all]

PR #11107, opened 2024-07-03
Add Service Attachment support for Secure Web Proxy

Samir-Cit:
This PR adds the support for **Secure Web Proxy** (Gateway) on the **Service Attachment** resource.
This is provided on the field _ targetService_ which now will have support for the _Forwarding Rule_ and the _Secure Web Proxy_ (Gateway).

**Note 1:** Test scenario was needed to verify if it is working with SWP.

**Note 2:** Diff Suppress function was needed because the fields maybe be slightly different when using **self_link** and this will generate a change on the resource and break the test. For example:
```
OLD = "https://networkservices.googleapis.com/beta/projects/{PROJECT_ID}/locations/us-east1/gateways/{SWP_ID}"
NEW = "https://networkservices.googleapis.com/v1alpha1/projects/{PROJECT_ID}/locations/us-east1/gateways/{SWP_ID}"
```

```release-note:enhancement
compute: changed `target_service` field on the `google_compute_service_attachment` resource to accept a `ForwardingRule` or `Gateway` URL.
```


Samir-Cit:
Hello @roaks3 
Did you had a chance to take a look on this PR?

roaks3:
@Samir-Cit this is interesting, I'll have to take a closer look at it tomorrow. Generally, I'm surprise the ResourceRefs need to be dropped, and I'm wondering if we've run into similar situations elsewhere.

I'll kick off the tests now to see how they do, and in the meantime, would you mind posting any related issues or prior discussion?

Samir-Cit:
> @Samir-Cit this is interesting, I'll have to take a closer look at it tomorrow. Generally, I'm surprise the ResourceRefs need to be dropped, and I'm wondering if we've run into similar situations elsewhere.

@roaks3 I checked for other situations like that and I was able to find one PR that does kind of the same change for Router.
You can check the PR https://github.com/GoogleCloudPlatform/magic-modules/pull/5162 for more details. In that case the change was to accept ForwardingRule and IP Address (string), so it's not the very same, but it's two diferent types of values. And that was the only way I found of doing that.

> I'll kick off the tests now to see how they do, and in the meantime, would you mind posting any related issues or prior discussion?

Regarding the test, thanks for runnig it. I did other change today, but I tested the function and the test should pass again!
The only test that is not working is one that is breaking on other PRs as well...

I don't have a prior discussion on this task but I found [this other PR](https://github.com/GoogleCloudPlatform/magic-modules/pull/1772/files#diff-706b96a4d29d1afef5f04308f0ea1b865be073b771a0767a8211e9cec91c739dL388-R389:~:text=%2D%20!ruby/object%3AApi%3A%3AType%3A%3AResourceRef,%2D%20!ruby/object%3AApi%3A%3AType%3A%3AString) with changes kind of the same that I did.

mmv1/products/compute/ServiceAttachment.yaml
```diff
@@ -156,16 +156,13 @@ properties:
...
             The status of the connection from the consumer forwarding rule to
             this service attachment.
-  - !ruby/object:Api::Type::ResourceRef
+  - !ruby/object:Api::Type::String
     name: targetService
     required: true
     immutable: true
-    resource: 'ForwardingRule'
-    imports: 'selfLink'
+    diff_suppress_func: 'TargetServiceDiffSuppress'
```
roaks3, line 163: With the most recent change, could this be `'tpgresource.CompareSelfLinkOrResourceName'` instead? Same below, and it might mean we can remove the custom diff suppress function.

roaks3 (commented):
Thanks for the additional context, that helped a lot. It looks like this does come up occasionally, and is a valid reason for ejecting from the ResourceRef.

I had one minor suggestion, but we should be able to run the test one last time and then merge once they are done.

mmv1/products/compute/ServiceAttachment.yaml
Samir-Cit: Sure, I saw yesterday that some other resources have this.
Soon I'll push a new commit with this change.
Thanks for the tip!

mmv1/products/compute/go_ServiceAttachment.yaml
```diff
@@ -157,15 +157,12 @@ properties:
...
     description: |
-      The URL of a forwarding rule that represents the service identified by
-      this service attachment.
+      The URL of a service serving the endpoint identified by this service attachment.
     required: true
     immutable: true
-    custom_expand: 'templates/terraform/custom_expand/go/resourceref_with_validation.go.tmpl'
-    resource: 'ForwardingRule'
-    imports: 'selfLink'
+    diff_suppress_func: 'TargetServiceDiffSuppress'
```
roaks3, line 165: I think you need `tpgresource.CompareSelfLinkOrResourceName` here as well now?

Samir-Cit: I forgot this file.
Now it is updated!

Samir-Cit:
Hello @roaks3 
Regarding the **go_ServiceAttachment.yaml** file:
On the file there is a [warning](https://github.com/GoogleCloudPlatform/magic-modules/blob/ea48f6ff65eb177b737777d0627f799a7c9b2397/mmv1/products/compute/go_ServiceAttachment.yaml#L14) to not edit directly.
Should I remove the change I did on this file or leave it there?
(I just saw this today)

roaks3:
Let's see how the tests do, but I think it is probably fine. I think the `go_` variants are probably generated from the originals, but updating both in lockstep like you've done seems ideal.

roaks3 (approved):

PR #11124, opened 2024-07-09
Fix wrong field value example in the iap client resource

gecko655:
The example in the doc for google_iap_client has a wrong argument field example. The `brand` field does not have `/identityAwareProxyClients/{client_id}` part, while the client has that field.

ref: https://cloud.google.com/iap/docs/reference/rest/v1/projects.brands.identityAwareProxyClients/create
> brand
> Required. Path to create the client in. In the following format: projects/{projectNumber/id}/brands/{brand}

```release-note:none

```


shuyama1 (approved):
Thanks for helping fix the documentation!

PR #11129, opened 2024-07-10
Fix duplicate attribute field in the iap client resource documentation

gecko655:
The `client_id` attribute in google_iap_client is described twice.

スクリーンショット 2024-07-10 13 00 01

```release-note:none

```


NickElliot (approved):
thanks for catching this!

PR #11139, opened 2024-07-11
Add cloud_storage_target field in google_data_loss_prevention_discovery_config

patrickmoy:
```release-note:enhancement
dlp: added `cloud_storage_target` field to `google_data_loss_prevention_discovery_config` resource
```

### DIFF
--- a/mmv1/products/dlp/DiscoveryConfig.yaml
+++ b/mmv1/products/dlp/DiscoveryConfig.yaml
@@ -74,6 +74,12 @@ examples:
     primary_resource_id: 'cloud_sql'
     test_env_vars:
       project: :PROJECT_NAME
+  - !ruby/object:Provider::Terraform::Examples
+    name: 'dlp_discovery_config_cloud_storage'
+    skip_test: true
+    primary_resource_id: 'cloud_storage'
+    test_env_vars:
+      project: :PROJECT_NAME
 custom_code: !ruby/object:Provider::Terraform::CustomCode
   encoder: templates/terraform/encoders/wrap_object.go.erb
   decoder: templates/terraform/decoders/unwrap_resource.go.erb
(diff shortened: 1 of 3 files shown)


mmv1/products/dlp/DiscoveryConfig.yaml
```diff
@@ -485,6 +491,128 @@ properties:
             # The fields below are necessary to include the "secretsDiscoveryTarget" target in the payload
           send_empty_value: true
           allow_empty_object: true
+        - !ruby/object:Api::Type::NestedObject
+          name: cloudStorageTarget
```
roaks3, line 495: To confirm, all of these field are mutable?

mmv1/products/dlp/DiscoveryConfig.yaml
```diff
@@ -485,6 +491,128 @@ properties:
...
+          description: Cloud Storage target for Discovery. The first target to match a bucket will be the one applied.
+          properties:
+            - !ruby/object:Api::Type::NestedObject
+              name: filter
+              description: The buckets the generation_cadence applies to. The first target with a matching filter will be the one to apply to a bucket.
+              required: true
+              properties:
+                - !ruby/object:Api::Type::NestedObject
+                  name: collection
```
roaks3, line 504: For some of these nested fields, you may want to make them required, specifically if there is only one field that can be used. This can help prevent users from accidentally using invalid config like an empty `{}`, but it also usually works best when you do not expect more fields to be added in the future.

mmv1/templates/terraform/examples/dlp_discovery_config_cloud_storage.tf.erb
```diff
@@ -0,0 +1,81 @@
+resource "google_data_loss_prevention_discovery_config" "<%= ctx[:primary_resource_id] %>" {
+	parent = "projects/<%= ctx[:test_env_vars]['project'] %>/locations/us"
```
roaks3, line 2: Nit: Spacing is a bit off here. This example will show up at the top of the documentation page for this resource, so inconsistencies in the formatting will be more visible

roaks3 (commented):
Overall looks good, just a few minor comments

mmv1/products/dlp/DiscoveryConfig.yaml
patrickmoy: Yes, they are mutable.

mmv1/products/dlp/DiscoveryConfig.yaml
patrickmoy: The filter is required — the nested fields are not required though. Our API-side validation will check for an invalid filter, but we're hesitant to add more restrictions on the terraform side because we might add more possible fields to the filter.

mmv1/templates/terraform/examples/dlp_discovery_config_cloud_storage.tf.erb
patrickmoy: Fixed with another commit and squashed into previous.

patrickmoy:
Comments addressed; I'd like to avoid adding those fields as required. 

Spacing fixed.

roaks3 (approved):

PR #11146, opened 2024-07-12
google_compute_public_advertised_prefix adding in output only shared_secret 

NA2047:
Adding in shared_secret 

fixes https://github.com/hashicorp/terraform-provider-google/issues/18171 

Please self-review your PR against the review checklist before creating it: https://googlecloudplatform.github.io/magic-modules/contribute/review-pr/

Completing the checklist will help speed up the review process, and we appreciate you spending time on them before sending
your code to be reviewed.

If your PR is still work in progress, please create it in draft mode
-->

```release-note:enhancement
compute: added `shared_secret` field to `google_compute_public_advertised_prefix` resource
```

### DIFF
--- a/mmv1/products/compute/PublicAdvertisedPrefix.yaml
+++ b/mmv1/products/compute/PublicAdvertisedPrefix.yaml
@@ -75,3 +75,8 @@ properties:
       The IPv4 address range, in CIDR format, represented by this public
       advertised prefix.
     required: true
+  - !ruby/object:Api::Type::String
+    name: 'sharedSecret'
+    output: true
+    description: |
+      Output Only. The shared secret to be used for reverse DNS verification.


roaks3 (approved):
LGTM, but wanted to double check, were you able to verify that the field was returned as expected during local testing?

NA2047:
> LGTM, but wanted to double check, were you able to verify that the field was returned as expected during local testing?

I wasn't able to deploy it out unfortunately, I tried to test with the example but I dont have the needed https://cloud.google.com/vpc/docs/create-pap#create-pap I would need to create a ROA request i believe as well. I figured since it was an output only field it was ok to just add it in.

roaks3 (approved):
Ok, sounds good. It does seem safe to change, but perhaps we can confirm with the user that this fixes the issue for them.

roaks3:
Also, could you check the CLA issue? I'm not sure if you need to make the commit with another account.

NA2047:
> Also, could you check the CLA issue? I'm not sure if you need to make the commit with another account.
I'll check with them and see if I can get the, to test 



> Also, could you check the CLA issue? I'm not sure if you need to make the commit with another account.

Also fixed the CLA issue should be good now I have the app installed

acharrett:
This looks really helpful, as per the note at the end of the issue would it be possible to also return dnsVerificationIp?

Thanks!

roaks3 (commented):
Just waiting on CLA fix to merge

PR #11148, opened 2024-07-12
Switch nightly ci to use the golang version of vcr_cassette_update

iyabchen:
https://github.com/hashicorp/terraform-provider-google/issues/18249

Part 1: https://github.com/GoogleCloudPlatform/magic-modules/pull/11147

This is tested in https://github.com/GoogleCloudPlatform/magic-modules/pull/11282

```release-note:none

```

### DIFF
--- a/.ci/gcb-vcr-nightly.yml
+++ b/.ci/gcb-vcr-nightly.yml
@@ -2,10 +2,15 @@
 steps:
     - name: 'gcr.io/graphite-docker-images/go-plus'
       id: gcb-vcr-nightly
-      entrypoint: '/workspace/.ci/scripts/go-plus/vcr-cassette-update/vcr_cassette_update.sh'
-      secretEnv: ["GOOGLE_BILLING_ACCOUNT", "GOOGLE_CUST_ID", "GOOGLE_IDENTITY_USER", "GOOGLE_MASTER_BILLING_ACCOUNT", "GOOGLE_ORG", "GOOGLE_ORG_2", "GOOGLE_ORG_DOMAIN", "GOOGLE_PROJECT", "GOOGLE_PROJECT_NUMBER", "GOOGLE_SERVICE_ACCOUNT", "SA_KEY", "GOOGLE_PUBLIC_AVERTISED_PREFIX_DESCRIPTION"]
+      entrypoint: '/workspace/.ci/scripts/go-plus/magician/exec.sh'
+      secretEnv: ["GOOGLE_BILLING_ACCOUNT", "GOOGLE_CUST_ID", "GOOGLE_IDENTITY_USER", "GOOGLE_MASTER_BILLING_ACCOUNT", "GOOGLE_ORG", "GOOGLE_ORG_2", "GOOGLE_ORG_DOMAIN", "GOOGLE_PROJECT", "GOOGLE_PROJECT_NUMBER", "GOOGLE_SERVICE_ACCOUNT", "SA_KEY", "GOOGLE_PUBLIC_AVERTISED_PREFIX_DESCRIPTION", "GITHUB_TOKEN_CLASSIC"]
+      env:
+        - "GOOGLE_REGION=us-central1"
+        - "GOOGLE_ZONE=us-central1-a"
+        - "USER=magician"
       args:
-          - $BUILD_ID
+        - 'vcr-cassette-update'
+        - $BUILD_ID
 
 # Long timeout to enable waiting on VCR test
 timeout: 20000s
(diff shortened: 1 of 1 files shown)


.ci/magician/cmd/vcr_cassette_update.go
```diff
@@ -0,0 +1,210 @@
+package cmd
+
+import (
```
shuyama1, line 3: @iyabchen qq: I assume this is the rewrite of https://github.com/GoogleCloudPlatform/magic-modules/blob/main/.ci/scripts/go-plus/vcr-cassette-update/vcr_cassette_update.sh. Where do we upload the cassettes to the GCS bucket after running in recording? Specifically [this step](https://github.com/GoogleCloudPlatform/magic-modules/blob/main/.ci/scripts/go-plus/vcr-cassette-update/vcr_cassette_update.sh#L130) in the original script.

iyabchen: Yes, this is for [vcr_cassette_update.sh](https://github.com/GoogleCloudPlatform/magic-modules/blob/main/.ci/scripts/go-plus/vcr-cassette-update/vcr_cassette_update.sh).

Good catch, I completely miss that, just updated the code and added a test for recording.

iyabchen:
Note that this PR depends on https://github.com/GoogleCloudPlatform/magic-modules/pull/11147, which reduces circular dependency and help to use mock for testing

.ci/gcb-vcr-nightly.yml
```diff
@@ -41,3 +46,5 @@ availableSecrets:
       env: SA_KEY
     - versionName: projects/673497134629/secrets/ci-test-public-advertised-prefix-description/versions/latest
       env: GOOGLE_PUBLIC_AVERTISED_PREFIX_DESCRIPTION
+    - versionName: projects/673497134629/secrets/github-classic--repo-workflow/versions/latest
```
shuyama1, line 49: we'll also need to add this to `secretEnv`.

shuyama1:
@iyabchen I've added the necessary permissions to the SA and the job successfully pulled the cassettes and run VCR in replaying and recording mode but I seems the job failed at uploading cassettes step. The `vcr-nightly-temp-testing` in https://github.com/GoogleCloudPlatform/magic-modules/pull/11171 should point to the latest run. Please let me know if you can't view the build log/ have any question/ need help debugging the issue. Thanks!

iyabchen:
> @iyabchen I've added the necessary permissions to the SA and the job successfully pulled the cassettes and run VCR in replaying and recording mode but I seems the job failed at uploading cassettes step. The `vcr-nightly-temp-testing` in #11171 should point to the latest run. Please let me know if you can't view the build log/ have any question/ need help debugging the issue. Thanks!

Hi @shuyama1 ,
I tried to run it locally, and since missing some env variable, so it doesn't proceed. I suspect the recording is not working properly for similar reasons, and I adjusted the upload log steps to upload building/testing log first, then upload the cassettes later, hoping the testing log can show some errors. 
Could you copy again and let me know the results?

.ci/magician/cmd/tester.go
```diff
@@ -1,9 +1,8 @@
-package vcr
+package cmd
```
melinath, line 1: Why does this need to be in the cmd package? It would be preferable to keep it separated so that it can be easily used as a library in other contexts if possible.

melinath: cc @trodge

iyabchen: The move is in this PR: https://github.com/GoogleCloudPlatform/magic-modules/pull/11147/files as part 1. 
It is to avoid circular dependency. 
I would like to have vcr Tester struct reference ExecRunner interface rather than the actual exec runner struct for tests.  
Since ExecRunner interface is defined in cmd package, and cmd package needs to reference tester struct, not moving will cause circular dependency. 
One alternative is to move the interface definition out of cmd.

melinath: I think that moving the interface definition out would be preferable (though I'm not sure how best to make that match best practices.)

shuyama1 (approved):
LGTM! I'll remove the temp testing trigger and monitor the VCR run tonight. Thanks!

shuyama1:
@iyabchen Looks like the VCR nightly run failed last night. See [logs](https://pantheon.corp.google.com/cloud-build/builds/e9676603-1717-4e3b-92df-fce83709a31d?project=graphite-docker-images&e=-13802955&mods=logs_tg_staging&invt=AbZZpQ&inv=1). Let's revert it for now?

PR #11149, opened 2024-07-12
Add new zone to VmwareenginePrivateCloud sweeper

SarahFrench:
When sweeping VMware Engine resources I've seen private clouds in this new zone. When running sweepers manually I can supply the zone as an input, but this PR ensures that scheduled sweeping will handle PCs in that zone

```release-note:none

```


zli82016 (approved):
LGTM.

PR #11156, opened 2024-07-14
add a subcommand into diff processor to detect missing docs

iyabchen:
https://github.com/hashicorp/terraform-provider-google/issues/14743
Part 1 - add missing docs command into diff-processor to detect new fields or new resource without doc. This only covers resource, NOT data sources.
- For all field diffs, it gets the resource doc
- Parse resource doc to get the list of existing arguments and attributes defined in the doc
  -  returned as a list of string, each string represented a flattend field, eg. a.b.c
  -  the parse might not be 100% accurate and not covered all corner cases. Whether a field is a nested object is determined by whether that field description contains a nested tag (some resource, eg. compute_instance `params` field does not have that in the description)
- If the field is argument and not int he argument list returned by doc parser, then it's included in the result; similary for attribute
  - however, i'm not sure about the argument/attribute condition written in the code is correct or not. 

```release-note:none

```


tools/diff-processor/detector/detector.go
```diff
@@ -152,3 +161,61 @@ func suggestedTest(resourceName string, untested []string) string {
...
+			}
+			argumentsInDoc = listToMap(parser.Arguments())
+			attributesInDoc = listToMap(parser.Attributes())
+		}
+
+		for field, fieldDiff := range resourceDiff.Fields {
+			if !isNewField(fieldDiff) {
+				continue
+			}
+			// presume all fields can be arguments
```
iyabchen, line 189: @shuyama1 I presume all fields can be arguments, and only some of them are attributes. Please let me know if this doesn't look right, or something else needs to be reported as missing doc. 

BTW, the high level general idea is described in the PR description.

shuyama1: Thanks. As we discussed offline, we only consider top-level computed field as attributes and the rest are arguments

iyabchen: sorry, the comment was not updated, i have cleaned it up now.

shuyama1:
Sorry for the delay! I just returned from vacation and am reviewing it now!

shuyama1:
> @iyabchen, this PR is waiting for action from you. If no action is taken, this PR will be closed in 14 days.
> 
> Please address any comments or change requests, or [re-request review](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/requesting-a-pull-request-review) from a core reviewer if no action is required.
> 
> 

I apologize for the delay in my review. I'm currently in another round of detailed review. Please ignore this message. Thank you for your patience!

tools/diff-processor/detector/detector.go
```diff
@@ -33,6 +37,16 @@ type Field struct {
...
 }
 
+type MissingDocDetails struct {
+	FilePath string
+	Fields   []MissingDocField
+}
+
+type MissingDocField struct {
+	Field   string
+	Section string
```
shuyama1, line 47: I assume section indicates if this field is in argument or attribute section, IIRC? Can you add a small comment for it?

tools/diff-processor/detector/detector.go
```diff
@@ -152,3 +166,67 @@ func suggestedTest(resourceName string, untested []string) string {
...
+			}
+		}
+		if len(details.Fields) > 0 {
+			ret[resource] = details
+		}
+	}
+	return ret, nil
+}
+
+func isAttribute(fieldDiff diff.FieldDiff) bool {
```
shuyama1, line 212: I might be reading this wrong - it seems that we still count the non-top level computed fields as attributes?

I’m starting to wonder that even if we only run the missing doc detector on newly added fields, inconsistencies in how attributes and arguments are divided in existing resource docs could still lead to issues(?), like for subfields added to existing objects. What are your thoughts on this?

tools/diff-processor/cmd/detect_missing_docs_test.go
```diff
@@ -0,0 +1,100 @@
...
+	"testing"
+
+	"github.com/GoogleCloudPlatform/magic-modules/tools/diff-processor/detector"
+	"github.com/GoogleCloudPlatform/magic-modules/tools/diff-processor/diff"
+	"github.com/google/go-cmp/cmp"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+)
+
+func TestDetectMissingDocs(t *testing.T) {
```
shuyama1, line 15: Can we add some more complex cases? Like nested objects?

shuyama1 (commented):
Mostly looks good, some minor comments asking for more clarify and unit test coverages.

tools/diff-processor/detector/detector.go
iyabchen: Decided to not differentiate between argument and attribute, since finding some fields not conforming to that rule, eg. terraform_labels.

tools/diff-processor/cmd/detect_missing_docs_test.go
iyabchen: I added the tests in detector_test.go, and leaving this to mostly verify the output is the expected json format.

tools/diff-processor/detector/detector.go
iyabchen: Section is removed since not differentiate argument and attribute now.

shuyama1 (approved):
LGTM overall! Thanks for putting in the work. Sorry for the delay in reviewing and merging—I’ve been a bit short on time to monitor CI changes. I’ll test it out this afternoon, and if everything goes well, I’ll merge it.

Next steps:
* add the missing doc detector logic for the datasource (PR in progress)
* surface this detector in the diff-report

shuyama1 (approved):
Thank you!

PR #11168, opened 2024-07-15
Add organization level mute config for SCC v2.

owenhuyn:
Adding new template for SCC API V2 Organization Mute Config
https://cloud.google.com/security-command-center/docs/reference/rest/v2/organizations.muteConfigs

`google_scc_v2_organization_mute_config`

```release-note:new-resource
`google_scc_v2_organization_mute_config`
```

### DIFF
--- a/mmv1/templates/terraform/examples/scc_v2_organization_mute_config_basic.tf.erb
+++ b/mmv1/templates/terraform/examples/scc_v2_organization_mute_config_basic.tf.erb
@@ -1,0 +1,8 @@
+resource "google_scc_v2_organization_mute_config" "<%= ctx[:primary_resource_id] %>" {
+  mute_config_id    = "<%= ctx[:vars]['mute_config_id'] %>"
+  organization = "<%= ctx[:test_env_vars]['org_id'] %>"
+  location     = "global"
+  description  = "My custom Cloud Security Command Center Finding Organization mute Configuration"
+  filter = "severity = \"HIGH\""
+  type = "STATIC"
+}
--- a/mmv1/third_party/terraform/services/securitycenterv2/resource_scc_v2_organization_mute_config_test.go
+++ b/mmv1/third_party/terraform/services/securitycenterv2/resource_scc_v2_organization_mute_config_test.go
@@ -8,7 +8,7 @@ import (
 	"github.com/hashicorp/terraform-provider-google/google/envvar"
 )
 
-func TestAccSecurityCenterV2OrganizationMuteConfig_basic(t *testing.T) {
+func TestAccSecurityCenterV2OrganizationMuteConfig_update(t *testing.T) {
 	t.Parallel()
 
 	context := map[string]interface{}{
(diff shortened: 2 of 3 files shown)


mmv1/templates/terraform/examples/scc_v2_organization_mute_config_basic.tf.erb
```diff
@@ -0,0 +1,8 @@
+resource "google_scc_v2_organization_mute_config" "<%= ctx[:primary_resource_id] %>" {
+  mute_config_id    = "<%= ctx[:vars]['config_id'] %>"
+  organization = "<%= ctx[:test_env_vars]['org_id'] %>"
+  location     = "global"
+  description  = "My custom Cloud Security Command Center Finding Organization mute Configuration"
+  filter = "category = \"OPEN_FIREWALL\" AND state = \"ACTIVE\""
```
rileykarson, line 6: This value looks invalid- not sure it's it, I noticed that `=` is the separator here and not `:` like was used in V1: https://github.com/hashicorp/terraform-provider-google-beta/blob/main/google-beta/services/securitycenter/resource_scc_mute_config_generated_test.go#L65

mmv1/third_party/terraform/services/securitycenterv2/resource_scc_v2_organization_mute_config_test.go
```diff
@@ -0,0 +1,73 @@
...
+
+	acctest.VcrTest(t, resource.TestCase{
+		PreCheck:                 func() { acctest.AccTestPreCheck(t) },
+		ProtoV5ProviderFactories: acctest.ProtoV5ProviderFactories(t),
+		Steps: []resource.TestStep{
+			{
+				Config: testAccSecurityCenterV2OrganizationMuteConfig_basic(context),
+			},
+			{
+				ResourceName:      "google_scc_v2_organization_mute_config.default",
```
rileykarson, line 27: Discussed offline, import is failing because we made bad assumptions about `self_link` (we append `name`) and an explicit `import_format` is required.

mmv1/templates/terraform/examples/scc_v2_organization_mute_config_basic.tf.erb
owenhuyn: Hmm it should work just fine with `=`. In fact I think`:` is not necessarily a correct separator for us:
https://cloud.google.com/security-command-center/docs/how-to-api-filter-notifications#create-filter

mmv1/third_party/terraform/services/securitycenterv2/resource_scc_v2_organization_mute_config_test.go
owenhuyn: Fixed.

rileykarson (approved):
LGTM- two minor nits, I'll apply the test case name change and merge. The other's just FYI.

PR #11173, opened 2024-07-16
Use the correct custom import for fields.

pcostell:
Use the correct custom import for fields.

Fixes https://github.com/hashicorp/terraform-provider-google/issues/17313.

Verified with a local `terraform import` followed by `terraform plan`.

```release-note:bug
firestore: fixed missing import of `field` for `google_firestore_field`.
```


roaks3 (changes requested):
My main concern is just that this wasn't caught with the tests, and it looks like we are using `ImportStateVerifyIgnore` to ignore fields like `field`. Would you mind removing the `ImportStateVerifyIgnore` line from the handwritten tests, so that this case is covered? Since you set these fields during import, we don't need to ignore them.

pcostell:
ImportVerify sections removed, thanks @roaks3 for pointing those out!

PR #11174, opened 2024-07-16
Firestore: Mark identifier fields as immutable.

pcostell:
This has the side effect that changing any of these fields will trigger the resource to be deleted and re-created, rather than executing a failed update.

Fixes https://github.com/hashicorp/terraform-provider-google/issues/12897.

```release-note:bug
firestore: fixes bug where fields database, collection, document_id, and field on resources could not be updated. 
```

### DIFF
--- a/mmv1/products/firestore/Database.yaml
+++ b/mmv1/products/firestore/Database.yaml
@@ -175,7 +175,6 @@ properties:
   - !ruby/object:Api::Type::Enum
     name: type
     required: true
-    immutable: true
     description: |
       The type of the database.
       See https://cloud.google.com/datastore/docs/firestore-or-datastore
--- a/mmv1/products/firestore/Database.yaml
+++ b/mmv1/products/firestore/Database.yaml
@@ -175,6 +175,7 @@ properties:
   - !ruby/object:Api::Type::Enum
     name: type
     required: true
+    immutable: true
     description: |
       The type of the database.
       See https://cloud.google.com/datastore/docs/firestore-or-datastore
--- a/mmv1/products/firestore/Document.yaml
+++ b/mmv1/products/firestore/Document.yaml
@@ -64,17 +64,20 @@ parameters:
     default_value: '(default)'
     description: |
       The Firestore database id. Defaults to `"(default)"`.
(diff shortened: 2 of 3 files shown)


pcostell:
@rwhogg for a Firestore reviewer.

mmv1/products/firestore/Database.yaml
```diff
@@ -175,6 +175,7 @@ properties:
   - !ruby/object:Api::Type::Enum
     name: type
     required: true
+    immutable: true
```
rwhogg, line 178: I agree with all the other changes, but I'm divided about this one. Doing this will prevent database mode changes via Terraform.

That's a very minor complaint, since you're only allowed to do that for empty databases anyways, so delete-then-recreate is just an annoyance rather than a real hindrance. If you think that's an acceptable tradeoff, feel free to resolve.

pcostell: Very good point, thanks Bob. Dropping this change.

rwhogg (approved):

pcostell:
Thanks for the feedback @roaks3. I did a little experimentation, described below.

Example starting resource:
```
resource "google_firestore_document" "test_doc" {
    collection = "test-collection"
    document_id = "foo"
}
```

| Description | File State | .tfstate | server state |
| --- | --- | --- | --- |
| Pre-Change, doc_id --> bar | bar | foo | foo |
| Post-Change, no change | bar | foo | foo |
| Post-Change, doc_id --> baz | baz | baz | (foo deleted), baz |

Importantly, providers picking up this change won't cause any behavior change. Users will only see the behavior difference the next time they change one of the name fields (and then the terraform plan/apply would show the change as a delete/create as expected.).

I think this means we could consider this as a non-breaking change?

roaks3 (approved):
Oh interesting, thanks for the breakdown. I think that the `url_param_only: true` config is what makes this a non-issue: these fields are not stored in state on Read, so a diff is only detected and sent in the request when the value changes in config. So like you said, upgrading the provider version with your fix should not have any negative impact for users.

For that reason, approving this as a non-breaking change.

PR #11175, opened 2024-07-16
Use new VPC networks in private connection tests

owaismawasy:
Fixes https://github.com/hashicorp/terraform-provider-google/issues/17384

The test currently uses a bootstrapped network for creating private connections. When a private connection creation fails, it leaves orphan resources linked to the peered subnet which prevents future runs from using that subnet. 

```release-note:none

```


NickElliot (approved):
LGTM!

PR #11177, opened 2024-07-16
google_firestore_index: Add reference to google_firestore_field

rwhogg:
It's not clear from the docs how to use SFIs from Terraform. We should fix that.

```release-note:none

```


rwhogg:
@pcostell can you take a look at this? Thanks!

pcostell (approved):

NickElliot (approved):

PR #11181, opened 2024-07-17
Need self_link instead of id in private_network

timwsuqld:
Move PR from https://github.com/hashicorp/terraform-provider-google/pull/18756 to here.

`google_compute_network.private_network.id` isn't a valid network name for `private_network`, it needs to be a `self_link`. This updates the docs to reflect that.

```release-note:none

```


NickElliot (approved):
Good catch, all of our tests are using self_link too. Thanks for the contribution!

PR #11182, opened 2024-07-17
add default values from api to smb_settings field in google cloud netapp volumes

Mehul3217:

```release-note:enhancement
netapp: added `default_from_api = true` attribute to `smb_settings` field of GoogleCloudNetappVolumes
```

### DIFF
--- a/mmv1/products/netapp/volume.yaml
+++ b/mmv1/products/netapp/volume.yaml
@@ -211,6 +211,7 @@ properties:
         - :SHOW_PREVIOUS_VERSIONS
         - :ACCESS_BASED_ENUMERATION
         - :CONTINUOUSLY_AVAILABLE
+    default_from_api: true
   - !ruby/object:Api::Type::String
     name: 'unixPermissions'
     description: |


mmv1/products/netapp/volume.yaml
```diff
@@ -211,6 +211,7 @@ properties:
         - :SHOW_PREVIOUS_VERSIONS
         - :ACCESS_BASED_ENUMERATION
         - :CONTINUOUSLY_AVAILABLE
+    default_from_api: true
```
BBBmau, line 214: Could I get some more context of this addition? Their isn't a default value for this based off of the REST reference: https://cloud.google.com/netapp/volumes/docs/reference/rest/v1/projects.locations.volumes#SMBSettings

Setting this could cause tests to fail in the future if the api suddenly changes.

Mehul3217: The execution plan is not stable if we do not set this, since while creating a smb_volume we do not explicitly set this field and it also does not have the default value so everytime we try to run plan without any change to TF config, we see this 
```
# google_netapp_volume.at_volume01 will be updated in-place
  ~ resource "google_netapp_volume" "at_volume01" {
        id                 = "projects/cvs-pm-host-1p/locations/us-west3/volumes/ok-at-volume01"
        name               = "ok-at-volume01"
      ~ smb_settings       = [
          - "OPLOCKS",
          - "CHANGE_NOTIFY",
          - "SHOW_PREVIOUS_VERSIONS",
          - "SHOW_SNAPSHOT",
          - "BROWSABLE",
        ]
        # (32 unchanged attributes hidden)

        # (1 unchanged block hidden)
    }
```

after we set it to true it will accept the empty value provided.

BBBmau: Thanks for the explanation!

BBBmau (approved):

PR #11183, opened 2024-07-17
Annotate immutable fields in ServerTlsPolicy.yaml

Stanwise:
Mark `ClientValidationTrustConfig` and `ClientValidationMode` immutable in Terraform.

Part of https://github.com/hashicorp/terraform-provider-google/issues/15453

```release-note:bug
networksecurity: added recreate functionality on update for `client_validation_mode` and `client_validation_trust_config` in  `google_network_security_server_tls_policy`
```

### DIFF
--- a/mmv1/products/networksecurity/ServerTlsPolicy.yaml
+++ b/mmv1/products/networksecurity/ServerTlsPolicy.yaml
@@ -150,6 +150,7 @@ properties:
     properties:
       - !ruby/object:Api::Type::Enum
         name: 'clientValidationMode'
+        immutable: true
         description: |
           When the client presents an invalid certificate or no certificate to the load balancer, the clientValidationMode specifies how the client connection is handled.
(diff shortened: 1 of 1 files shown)


slevenick (commented):
What happens if you try to update these currently?

Stanwise:
> What happens if you try to update these currently?

Just as described in the issue linked in the description: `terraform apply` tries to update in-place and fails with `Error: Error waiting for Updating ServerTlsPolicy: Error code 9, message: Update not supported for ServerTlsPolicy created for Load Balancer`

The PR is marked as "part of" not as "fixes" because it addresses the problem for  `ClientValidationTrustConfig` and `ClientValidationMode`, but not for `description` and `labels`. The latter ones cannot be marked as immutable, because in some cases they are mutable and in some they are not. This would require a fix on API level (to make them always mutable).

slevenick (approved):
LGTM

PR #11190, opened 2024-07-17
Remove `default_from_api` from `liveness_probe` field in Cloud Run v2 service

yanweiguo:
Fix https://github.com/hashicorp/terraform-provider-google/issues/17468.

Cloud Run doesn't set a default value for liveness probe. Marking `liveness_probe` field with `default_from_api: true` was a mistake.

```release-note:breaking-change
cloudrunv2: `liveness_probe` field in `google_cloud_run_v2_service` resource no longer has a default value from API side. Removing this field and applying the change will remove liveness probe from the Cloud Run service.
```


hao-nan-li:
@NickElliot Taking over this PR as I was reviewing the closed one.

hao-nan-li (approved):

trshafer (commented):
looks good

PR #11222, opened 2024-07-22
update cloudfunctionv2 test runtimes to nodejs18

dixuswe:
update cloudfunctionsV2 test case to use nodejs18 runtimes. The nodejs12 is end of lifecycle support

fixes #18638

```release-note:none

```


BBBmau (approved):
Thanks!

PR #11230, opened 2024-07-23
Add support of storage pool on boot disk.

sizzle0121:
Add `storage_pool` to the boot_disk of `google_compute_instance` resource to support creating boot disk with the storage pool specified.
Since the storage pool is not a resource in the terraform provider (due to some security concern, please reference https://github.com/GoogleCloudPlatform/magic-modules/pull/10793), it's only a URL string.

```release-note:enhancement
compute: added `storage_pool` under `boot_disk.initialize_params` to `google_compute_instance` resource
```

### DIFF
--- a/mmv1/products/compute/Instance.yaml
+++ b/mmv1/products/compute/Instance.yaml
@@ -232,6 +232,15 @@ properties:
                     customer-supplied encryption key that protects this
                     resource.
                   output: true
+            - !ruby/object:Api::Type::String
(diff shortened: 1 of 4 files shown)


sizzle0121:
Hi @trodge, we already have different test cases for the fields it mentioned, and the one I added I also added a test for it and made sure it passes, please review, thanks!

trodge (approved):

PR #11249, opened 2024-07-24
Add accessMode support for compute disk

chentaozgoogle:
```release-note:enhancement
compute: added `access_mode` field to `google_compute_disk` resource
```

### DIFF
--- a/mmv1/products/compute/Disk.yaml
+++ b/mmv1/products/compute/Disk.yaml
@@ -521,5 +521,6 @@ properties:
       * READ_WRITE_SINGLE
       * READ_WRITE_MANY
       * READ_ONLY_SINGLE
-    diff_suppress_func: 'tpgresource.CompareResourceNames'
-    custom_flatten: 'templates/terraform/custom_flatten/name_from_self_link.erb'
+    default_from_api: true
+    update_verb: :PATCH
+    update_url: 'projects/{{project}}/zones/{{zone}}/disks/{{name}}?paths=accessMode'
--- a/mmv1/products/compute/Disk.yaml
+++ b/mmv1/products/compute/Disk.yaml
@@ -511,3 +511,15 @@ properties:
       * /projects/{project}/zones/{zone}/storagePools/{storagePool}
     diff_suppress_func: 'tpgresource.CompareResourceNames'
     custom_flatten: 'templates/terraform/custom_flatten/name_from_self_link.erb'
+  - !ruby/object:Api::Type::String
+    name: 'accessMode'
+    required: false
+    immutable: false
+    description: |
+      The accessMode of the disk.
+      For example:
+      * READ_WRITE_SINGLE
+      * READ_WRITE_MANY
+      * READ_ONLY_SINGLE
+    diff_suppress_func: 'tpgresource.CompareResourceNames'
+    custom_flatten: 'templates/terraform/custom_flatten/name_from_self_link.erb'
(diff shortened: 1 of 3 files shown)


mmv1/third_party/terraform/services/compute/resource_compute_disk_test.go.erb
```diff
@@ -1720,3 +1720,40 @@ resource "google_compute_disk" "foobar" {
...
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr("google_compute_disk.foobar", "access_mode", accessMode),
+				),
+			},
+			{
+				ResourceName:      "google_compute_disk.foobar",
+				ImportState:       true,
+				ImportStateVerify: true,
+			},
+		},
```
zli82016, line 1745: Can you add more test steps for this new field to test if updating works or not?

Doc: [how to add a update test](https://googlecloudplatform.github.io/magic-modules/develop/test/test/#add-an-update-test)

zli82016 (changes requested):

zli82016:
Hello, @chentaozgoogle, do you know if there is a Github issue for this PR?

mmv1/third_party/terraform/services/compute/resource_compute_disk_test.go.erb
chentaozgoogle: added

chentaozgoogle:
> Hello, @chentaozgoogle, do you know if there is a Github issue for this PR?

hey @zli82016 I don't think there is a Github issue for this PR. Adding accessMode support is GAed for hyperdisk multi-writer GA.  Do i need to create a Github issue for it? so how can I create it?

mmv1/products/compute/Disk.yaml
```diff
@@ -511,3 +511,16 @@ properties:
...
+    name: 'accessMode'
+    required: false
+    immutable: false
+    description: |
+      The accessMode of the disk.
+      For example:
+      * READ_WRITE_SINGLE
+      * READ_WRITE_MANY
+      * READ_ONLY_SINGLE
+    default_from_api: true
```
zli82016, line 524: What is the reason to use `default_from_api: true` here? `default_from_api: true` is used when the field has a default value set the service.

zli82016:
> > Hello, @chentaozgoogle, do you know if there is a Github issue for this PR?
> 
> hey @zli82016 I don't think there is a Github issue for this PR. Adding accessMode support is GAed for hyperdisk multi-writer GA. Do i need to create a Github issue for it? so how can I create it?

I see. Then it should be fine without the Github issue.

mmv1/products/compute/Disk.yaml
chentaozgoogle: I was referring to enabledConfidentialCompute field https://github.com/chentaozgoogle/magic-modules/blob/main/mmv1/products/compute/Disk.yaml#L438 . They are similar case in regard with default value, e.g. some disk doesn't have the field at all, some disk has the field, and when the disk has  the field, the default is determine by the backend API server.   I though `default_from_api: true` is for the case. Am I wrong? if `default_from_api: tru` field is not meant for that, I can remove it.

zli82016: It meant that if the value of the field is not provided by the user, the API server will set a default value, then the default value is sent back to the API in the following requests if no provided value by the user. If a value is provided by the user, the user's value is sent to API.

When no value is provided by the user,  the server API will set a default value for it. Then `default_from_api: true` can be used. 

When no value is provided by the user,  will the server API set a default value for it?

chentaozgoogle: Some disk may not have field, so the API server will not return the field at all.
For disks that do have the field,  when the no value is provided by the user, the API server will return the default value.

zli82016 (approved):
LGTM. Thanks.

PR #11256, opened 2024-07-25
Add deletion_protection field to CloudAD domain resource

abd-goog:
Part of b/337047978

Part of  https://github.com/hashicorp/terraform-provider-google/issues/18854

```release-note:breaking-change
activeDirectory: added `deletion_protection` field to `google_active_directory_domain` to make deleting them require an explicit intent. `google_active_directory_domain` resources now cannot be destroyed unless `deletion_protection = false` is set for the resource.
```


abd-goog:
@slevenick could you please review this PR, since you would already have context from our previous email conversation?

slevenick (changes requested):
I'm not seeing this change actually do anything in the generated code. Have you seen this work locally?

abd-goog:
> I'm not seeing this change actually do anything in the generated code. Have you seen this work locally?

My bad, uncommited changes on local. Update PR.

mmv1/templates/terraform/custom_delete/active_directory_domain.erb
```diff
@@ -0,0 +1,52 @@
+if d.Get("deletion_protection").(bool) {
```
slevenick, line 1: This logic should be able to go in a pre_delete by itself to prevent having to copy over the entire delete functionality

mmv1/third_party/terraform/services/activedirectory/resource_active_directory_domain_trust_test.go
```diff
@@ -62,6 +62,7 @@ resource "google_active_directory_domain_trust" "ad-domain-trust" {
     trust_direction         = "OUTBOUND"
     trust_type              = "FOREST"
     trust_handshake_secret  = "Testing1!"
+	deletion_protection = false
```
slevenick, line 65: Spacing is a little wrong here. Maybe tabs were used

slevenick (changes requested):

mmv1/third_party/terraform/services/activedirectory/resource_active_directory_domain_trust_test.go
abd-goog: Updated. Thanks!

mmv1/templates/terraform/custom_delete/active_directory_domain.erb
abd-goog: Updated. Thanks!

slevenick (approved):

PR #11264, opened 2024-07-25
[18856] update access policy documentation

joeaawad:
Fixes https://github.com/hashicorp/terraform-provider-google/issues/18856

```release-note:none

```

### DIFF
--- a/mmv1/products/accesscontextmanager/AccessPolicy.yaml
+++ b/mmv1/products/accesscontextmanager/AccessPolicy.yaml
@@ -68,7 +68,7 @@ parameters:
     required: true
     description: |
       The parent of this AccessPolicy in the Cloud Resource Hierarchy.
-      Format: organizations/{organization_id}
+      Format: 'organizations/{{orgnumber}}'
   - !ruby/object:Api::Type::String
     name: title
     required: true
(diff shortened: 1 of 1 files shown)


roaks3 (approved):
Approving because I think this is technically ok, but would you mind making these variables a bit more consistent with the rest of the codebase?

- I believe `organization_id` is always a number, so can probably stick with that
- Similar with `folder_id`, I think the folder name is the user-specified value, and id is the number
- `project _number` seems marginally more common than `projectnumber`, but either is probably fine

joeaawad:
@roaks3 I was trying to follow the exact text of the error message, which says

```
Error: Error creating AccessPolicy: googleapi: Error 400: Invalid scope.  Must be of the form 'organizations/<orgnumber>', 'folders/<foldernumber>', or 'projects/<projectnumber>'.
```
If you would prefer me to use the other values instead, I am happy to do so, but wanted to first confirm you would prefer those new values even though they differ from the error message the Google API generates.

roaks3:
Yea, there is definitely some ambiguity. The values in the error message are controlled by the API (not the TF client), but you can see that those aren't even consistent with [the API documentation](https://cloud.google.com/access-context-manager/docs/reference/rest/v1/accessPolicies). I would suggest staying consistent with the TF surface (and majority of other services).

joeaawad:
@roaks3 updated per your comments

roaks3 (approved):

PR #11268, opened 2024-07-26
[Datastream] BigQuery data staleness defaults to '900s'

marblejenka:
For https://github.com/hashicorp/terraform-provider-google/issues/18693

```release-note:bug
datastream: BigQuery data staleness defaults to '900s'
```

### DIFF
--- a/mmv1/products/datastream/Stream.yaml
+++ b/mmv1/products/datastream/Stream.yaml
@@ -1080,7 +1080,7 @@ properties:
         properties:
           - !ruby/object:Api::Type::String
             name: 'dataFreshness'
-            default_value: 600
+            default_value: '900s'
             description: |
               The guaranteed data freshness (in seconds) when querying tables created by the stream.
               Editing this field will only affect new tables created in the future, but existing tables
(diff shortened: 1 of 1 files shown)


slevenick (changes requested):
Error: "Property 'default_value' is 'Integer' instead of 'String' (RuntimeError)"

slevenick:
Please see https://googlecloudplatform.github.io/magic-modules/develop/breaking-changes/breaking-changes/#field-level-breaking-changes

This could potentially be made in the 6.0.0 release which is coming up, but you will need to target that branch instead of main

slevenick (changes requested):
See previous comment

marblejenka:
@slevenick I think if we put  `override-breaking-change` for this PR, it can be merged.

slevenick:
Unfortunately we missed the window for breaking changes for 6.0.0. I was OOO when you switched to the 6.0 branch and the window was closed by the time I came back. Sorry about this! We won't be able to make this change until 7.0.0 now

PR #11270, opened 2024-07-26
go rewrite workbench-storageinsights

NickElliot:
Minor diff in `website/docs/r/workbench_instance.html.markdown` due to different default example storage account names

Removed non-functioning exactly_one_of tag in `workbench_instance`, which produced a meaningless syntactical diff in `google-beta/services/workbench/resource_workbench_instance.go` via a removed piece of non functional code (but did produce a real diff if untouched due to the go rewrite handling this situation of a nested exactly_one_of differently)

```
-                                                                               ExactlyOneOf: []string{},
                                                                        },
                                                                        "name": {
-                                                                               Type:         schema.TypeString,
-                                                                               Optional:     true,
-                                                                               ForceNew:     true,
-                                                                               Description:  `Optional. Use VM image name to find the image.`,
-                                                                               ExactlyOneOf: []string{},
+                                                                               Type:        schema.TypeString,
+                                                                               Optional:    true,
+                                                                               ForceNew:    true,
+                                                                               Description: `Optional. Use VM image name to find the image.`,
                                                                        },
```

```release-note:none

```


c2thorn (approved):
LGTM

PR #11272, opened 2024-07-27
container: add support for kubelet read only port

wyardley:
- Add `no_enable_insecure_kubelet_readonly_port` to google_container_cluster
- Allow setting `insecure_kubelet_readonly_port_enabled` for `container_node_pool` and friends

https://cloud.google.com/kubernetes-engine/docs/how-to/disable-kubelet-readonly-port

Fixes https://github.com/hashicorp/terraform-provider-google/issues/15208

Note: @trodge: may be good to get some feedback internally from Google folks about what the right default behavior should be and if this will change over time. I'm trying to do this in the least breaking (for the provider) way, but from my very quick reading of the announcements, it's possible that default behavior may change in the future (and that this may depend on new vs. existing clusters and / or based on cluster version).

```release-note:enhancement
container: added `insecure_kubelet_readonly_port_enabled` to `node_pool.node_config.kubelet_config` and `node_config.kubelet_config` in `google_container_node_pool` resource.
```
```release-note:enhancement
container: added `insecure_kubelet_readonly_port_enabled` to `node_pool_defaults.node_config_defaults`, `node_pool.node_config.kubelet_config`, and `node_config.kubelet_config` in `google_container_cluster` resource.
```

### DIFF
--- a/mmv1/third_party/terraform/services/container/node_config.go.erb
+++ b/mmv1/third_party/terraform/services/container/node_config.go.erb
@@ -85,7 +85,7 @@ func schemaInsecureKubeletReadonlyPortEnabled() *schema.Schema {
 		Type:         schema.TypeString,
 		Optional:     true,
 		Computed:     true,
-		Description:  "Controls whether the kubelet read-only port is enabled. It is strongly recommended to set this to FALSE.",
+		Description:  "Controls whether the kubelet read-only port is enabled. It is strongly recommended to set this to `FALSE`. Possible values: `TRUE`, `FALSE`.",
 		ValidateFunc: validation.StringInSlice([]string{"FALSE","TRUE"}, false),
 	}
 }
--- a/mmv1/third_party/terraform/services/container/node_config.go.erb
+++ b/mmv1/third_party/terraform/services/container/node_config.go.erb
@@ -78,6 +78,18 @@ func schemaContainerdConfig() *schema.Schema {
 	}
 }
 
+// Note: this is a bool internally, but implementing as an enum internally to
+// make it easier to accept API level defaults.
+func schemaInsecureKubeletReadonlyPortEnabled() *schema.Schema {
+	return &schema.Schema{
+		Type:         schema.TypeString,
+		Optional:     true,
+		Computed:     true,
+		Description:  "Controls whether the kubelet read-only port is enabled. It is strongly recommended to set this to FALSE.",
+		ValidateFunc: validation.StringInSlice([]string{"FALSE","TRUE"}, false),
+	}
+}
+
 func schemaLoggingVariant() *schema.Schema {
 	return &schema.Schema{
 		Type: schema.TypeString,
@@ -597,6 +609,7 @@ func schemaNodeConfig() *schema.Schema {
 								Optional: true,
 								Description: `Set the CPU CFS quota period value 'cpu.cfs_period_us'.`,
 							},
+							"insecure_kubelet_readonly_port_enabled": schemaInsecureKubeletReadonlyPortEnabled(),
 							"pod_pids_limit": {
 								Type:        schema.TypeInt,
 								Optional:    true,
@@ -769,7 +782,7 @@ func schemaNodeConfig() *schema.Schema {
 }
 
 func expandNodeConfigDefaults(configured interface{}) *container.NodeConfigDefaults {
-        configs := configured.([]interface{})
+	configs := configured.([]interface{})
 	if len(configs) == 0 || configs[0] == nil {
 		return nil
 	}
@@ -777,6 +790,12 @@ func expandNodeConfigDefaults(configured interface{}) *container.NodeConfigDefau
 
 	nodeConfigDefaults := &container.NodeConfigDefaults{}
 	nodeConfigDefaults.ContainerdConfig = expandContainerdConfig(config["containerd_config"])
+	if v, ok := config["insecure_kubelet_readonly_port_enabled"]; ok {
+		nodeConfigDefaults.NodeKubeletConfig = &container.NodeKubeletConfig{
+			InsecureKubeletReadonlyPortEnabled: expandInsecureKubeletReadonlyPortEnabled(v),
+			ForceSendFields: []string{"InsecureKubeletReadonlyPortEnabled"},
+		}
+	}
 	if variant, ok := config["logging_variant"]; ok {
 		nodeConfigDefaults.LoggingConfig = &container.NodePoolLoggingConfig{
 			VariantConfig: &container.LoggingVariantConfig{
@@ -785,14 +804,14 @@ func expandNodeConfigDefaults(configured interface{}) *container.NodeConfigDefau
 		}
 	}
 <% unless version == "ga" -%>
-        if v, ok := config["gcfs_config"]; ok && len(v.([]interface{})) > 0 {
-                gcfsConfig := v.([]interface{})[0].(map[string]interface{})
+	if v, ok := config["gcfs_config"]; ok && len(v.([]interface{})) > 0 {
+		gcfsConfig := v.([]interface{})[0].(map[string]interface{})
 		nodeConfigDefaults.GcfsConfig = &container.GcfsConfig{
 			Enabled: gcfsConfig["enabled"].(bool),
 		}
 	}
 <% end -%>
-        return nodeConfigDefaults
+	return nodeConfigDefaults
 }
 
 func expandNodeConfig(v interface{}) *container.NodeConfig {
@@ -1134,6 +1153,13 @@ func expandWorkloadMetadataConfig(v interface{}) *container.WorkloadMetadataConf
 	return wmc
 }
 
+func expandInsecureKubeletReadonlyPortEnabled(v interface{}) bool {
+	if v == "TRUE" {
+		return true
+	}
+	return false
+}
+
 func expandKubeletConfig(v interface{}) *container.NodeKubeletConfig {
 	if v == nil {
 		return nil
@@ -1153,6 +1179,10 @@ func expandKubeletConfig(v interface{}) *container.NodeKubeletConfig {
 	}
 	if cpuCfsQuotaPeriod, ok := cfg["cpu_cfs_quota_period"]; ok {
 		kConfig.CpuCfsQuotaPeriod = cpuCfsQuotaPeriod.(string)
+	}
+	if insecureKubeletReadonlyPortEnabled, ok := cfg["insecure_kubelet_readonly_port_enabled"]; ok {
+		kConfig.InsecureKubeletReadonlyPortEnabled = expandInsecureKubeletReadonlyPortEnabled(insecureKubeletReadonlyPortEnabled)
+		kConfig.ForceSendFields = append(kConfig.ForceSendFields, "InsecureKubeletReadonlyPortEnabled")
 	}
 	if podPidsLimit, ok := cfg["pod_pids_limit"]; ok {
 		kConfig.PodPidsLimit = int64(podPidsLimit.(int))
@@ -1362,6 +1392,8 @@ func flattenNodeConfigDefaults(c *container.NodeConfigDefaults) []map[string]int
 
 	result[0]["containerd_config"] = flattenContainerdConfig(c.ContainerdConfig)
 
+	result[0]["insecure_kubelet_readonly_port_enabled"] = flattenInsecureKubeletReadonlyPortEnabled(c.NodeKubeletConfig)
+
 	result[0]["logging_variant"] = flattenLoggingVariant(c.LoggingConfig)
 
 <% unless version == 'ga' -%>
@@ -1553,6 +1585,14 @@ func flattenSecondaryBootDisks(c []*container.SecondaryBootDisk) []map[string]in
 	return result
 }
 
+func flattenInsecureKubeletReadonlyPortEnabled(c *container.NodeKubeletConfig) string {
+	// Convert bool from the API to the enum values used internally
+	if c != nil && c.InsecureKubeletReadonlyPortEnabled {
+		return "TRUE"
+	}
+	return "FALSE"
+}
+
 func flattenLoggingVariant(c *container.NodePoolLoggingConfig) string {
 	variant := "DEFAULT"
 	if c != nil && c.VariantConfig != nil && c.VariantConfig.Variant != "" {
@@ -1702,10 +1742,11 @@ func flattenKubeletConfig(c *container.NodeKubeletConfig) []map[string]interface
 	result := []map[string]interface{}{}
 	if c != nil {
 		result = append(result, map[string]interface{}{
-			"cpu_cfs_quota":        c.CpuCfsQuota,
-			"cpu_cfs_quota_period": c.CpuCfsQuotaPeriod,
-			"cpu_manager_policy":   c.CpuManagerPolicy,
-			"pod_pids_limit":       c.PodPidsLimit,
+			"cpu_cfs_quota":                          c.CpuCfsQuota,
+			"cpu_cfs_quota_period":                   c.CpuCfsQuotaPeriod,
+			"cpu_manager_policy":                     c.CpuManagerPolicy,
+			"insecure_kubelet_readonly_port_enabled": flattenInsecureKubeletReadonlyPortEnabled(c),
+			"pod_pids_limit":                         c.PodPidsLimit,
 		})
 	}
 	return result
--- a/mmv1/third_party/terraform/services/container/resource_container_cluster.go.erb
+++ b/mmv1/third_party/terraform/services/container/resource_container_cluster.go.erb
@@ -3835,23 +3835,25 @@ func resourceContainerClusterUpdate(d *schema.ResourceData, meta interface{}) er
 			log.Printf("[INFO] GKE cluster %s: image type has been updated to %s", d.Id(), it)
 		}
 
-		// Note: probably long term this should be handled broadly for all the
-		// items in kubelet_config in a simpler / DRYer way.
-		// See b/361634104
-		if d.HasChange("node_config.0.kubelet_config.0.insecure_kubelet_readonly_port_enabled") {
-			if v, ok := d.GetOk("node_config.0.kubelet_config.0.insecure_kubelet_readonly_port_enabled"); ok {
-				insecureKubeletReadonlyPortEnabled := v.(string)
-				defaultPool := "default-pool"
+		if d.HasChange("node_config.0.kubelet_config") {
 
-				timeout := d.Timeout(schema.TimeoutCreate)
+			defaultPool := "default-pool"
 
-				nodePoolInfo, err := extractNodePoolInformationFromCluster(d, config, clusterName)
-				if err != nil {
-					return err
-				}
+			timeout := d.Timeout(schema.TimeoutCreate)
 
-				// Acquire write-lock on nodepool.
-				npLockKey := nodePoolInfo.nodePoolLockKey(defaultPool)
+			nodePoolInfo, err := extractNodePoolInformationFromCluster(d, config, clusterName)
+			if err != nil {
+				return err
+			}
+
+			// Acquire write-lock on nodepool.
+			npLockKey := nodePoolInfo.nodePoolLockKey(defaultPool)
+
+			// Note: probably long term this should be handled broadly for all the
+			// items in kubelet_config in a simpler / DRYer way.
+			// See b/361634104
+			if d.HasChange("node_config.0.kubelet_config.0.insecure_kubelet_readonly_port_enabled") {
+				it := d.Get("node_config.0.kubelet_config.0.insecure_kubelet_readonly_port_enabled").(string)
 
 				// While we're getting the value from the drepcated field in
 				// node_config.kubelet_config, the actual setting that needs to be updated
@@ -3859,7 +3861,7 @@ func resourceContainerClusterUpdate(d *schema.ResourceData, meta interface{}) er
 				req := &container.UpdateNodePoolRequest{
 					Name: defaultPool,
 					KubeletConfig: &container.NodeKubeletConfig{
-						InsecureKubeletReadonlyPortEnabled: expandInsecureKubeletReadonlyPortEnabled(insecureKubeletReadonlyPortEnabled),
+						InsecureKubeletReadonlyPortEnabled: expandInsecureKubeletReadonlyPortEnabled(it),
 						ForceSendFields:                    []string{"InsecureKubeletReadonlyPortEnabled"},
 					},
 				}
@@ -3875,19 +3877,16 @@ func resourceContainerClusterUpdate(d *schema.ResourceData, meta interface{}) er
 					}
 
 					// Wait until it's updated
-					return ContainerOperationWait(config, op,
-						nodePoolInfo.project,
-						nodePoolInfo.location,
-						"updating GKE node pool insecure_kubelet_readonly_port_enabled", userAgent,
-						timeout)
+					return ContainerOperationWait(config, op, nodePoolInfo.project, nodePoolInfo.location,
+						"updating GKE node pool insecure_kubelet_readonly_port_enabled", userAgent, timeout)
 				}
 
 				if err := retryWhileIncompatibleOperation(timeout, npLockKey, updateF); err != nil {
 					return err
 				}
 
-				log.Printf("[INFO] GKE cluster %s: default-pool setting for insecure_kubelet_readonly_port_enabled updated to %s", d.Id(), insecureKubeletReadonlyPortEnabled)
-			}
+				log.Printf("[INFO] GKE cluster %s: default-pool setting for insecure_kubelet_readonly_port_enabled updated to %s", d.Id(), it)
+      }
 		}
 	}
 
--- a/mmv1/third_party/terraform/services/container/resource_container_cluster.go.erb
+++ b/mmv1/third_party/terraform/services/container/resource_container_cluster.go.erb
@@ -162,6 +162,7 @@ func clusterSchemaNodePoolDefaults() *schema.Schema {
 <% unless version == 'ga' -%>
 							"gcfs_config": schemaGcfsConfig(false),
 <% end -%>
+							"insecure_kubelet_readonly_port_enabled": schemaInsecureKubeletReadonlyPortEnabled(),
 							"logging_variant": schemaLoggingVariant(),
 						},
 					},
@@ -4256,6 +4257,28 @@ func resourceContainerClusterUpdate(d *schema.ResourceData, meta interface{}) er
 			}
 
 			log.Printf("[INFO] GKE cluster %s enabled Kubernetes Beta APIs has been updated", d.Id())
+		}
+	}
+
+	if d.HasChange("node_pool_defaults") && d.HasChange("node_pool_defaults.0.node_config_defaults.0.insecure_kubelet_readonly_port_enabled") {
+		if v, ok := d.GetOk("node_pool_defaults.0.node_config_defaults.0.insecure_kubelet_readonly_port_enabled"); ok {
+			insecureKubeletReadonlyPortEnabled := v.(string)
+			req := &container.UpdateClusterRequest{
+				Update: &container.ClusterUpdate{
+					DesiredNodeKubeletConfig: &container.NodeKubeletConfig{
+						InsecureKubeletReadonlyPortEnabled: expandInsecureKubeletReadonlyPortEnabled(insecureKubeletReadonlyPortEnabled),
+						ForceSendFields:                    []string{"InsecureKubeletReadonlyPortEnabled"},
+					},
+				},
+			}
+
+			updateF := updateFunc(req, "updating GKE cluster desired node pool insecure kubelet readonly port configuration defaults.")
+			// Call update serially.
+			if err := transport_tpg.LockedCall(lockKey, updateF); err != nil {
+				return err
+			}
+
+			log.Printf("[INFO] GKE cluster %s node pool insecure_kubelet_readonly_port_enabled default has been updated", d.Id())
 		}
 	}
 
--- a/mmv1/third_party/terraform/services/container/resource_container_cluster.go.erb
+++ b/mmv1/third_party/terraform/services/container/resource_container_cluster.go.erb
@@ -3834,6 +3834,61 @@ func resourceContainerClusterUpdate(d *schema.ResourceData, meta interface{}) er
 
 			log.Printf("[INFO] GKE cluster %s: image type has been updated to %s", d.Id(), it)
 		}
+
+		// Note: probably long term this should be handled broadly for all the
+		// items in kubelet_config in a simpler / DRYer way.
+		// See b/361634104
+		if d.HasChange("node_config.0.kubelet_config.0.insecure_kubelet_readonly_port_enabled") {
+			if v, ok := d.GetOk("node_config.0.kubelet_config.0.insecure_kubelet_readonly_port_enabled"); ok {
+				insecureKubeletReadonlyPortEnabled := v.(string)
+				defaultPool := "default-pool"
+
+				timeout := d.Timeout(schema.TimeoutCreate)
+
+				nodePoolInfo, err := extractNodePoolInformationFromCluster(d, config, clusterName)
+				if err != nil {
+					return err
+				}
+
+				// Acquire write-lock on nodepool.
+				npLockKey := nodePoolInfo.nodePoolLockKey(defaultPool)
+
+				// While we're getting the value from the drepcated field in
+				// node_config.kubelet_config, the actual setting that needs to be updated
+				// is on the default nodepool.
+				req := &container.UpdateNodePoolRequest{
+					Name: defaultPool,
+					KubeletConfig: &container.NodeKubeletConfig{
+						InsecureKubeletReadonlyPortEnabled: expandInsecureKubeletReadonlyPortEnabled(insecureKubeletReadonlyPortEnabled),
+						ForceSendFields:                    []string{"InsecureKubeletReadonlyPortEnabled"},
+					},
+				}
+
+				updateF := func() error {
+					clusterNodePoolsUpdateCall := config.NewContainerClient(userAgent).Projects.Locations.Clusters.NodePools.Update(nodePoolInfo.fullyQualifiedName(defaultPool), req)
+					if config.UserProjectOverride {
+						clusterNodePoolsUpdateCall.Header().Add("X-Goog-User-Project", nodePoolInfo.project)
+					}
+					op, err := clusterNodePoolsUpdateCall.Do()
+					if err != nil {
+						return err
+					}
+
+					// Wait until it's updated
+					return ContainerOperationWait(config, op,
+						nodePoolInfo.project,
+						nodePoolInfo.location,
+						"updating GKE node pool insecure_kubelet_readonly_port_enabled", userAgent,
+						timeout)
+				}
+
+				if err := retryWhileIncompatibleOperation(timeout, npLockKey, updateF); err != nil {
+					return err
+				}
+
+				log.Printf("[INFO] GKE cluster %s: default-pool setting for insecure_kubelet_readonly_port_enabled updated to %s", d.Id(), insecureKubeletReadonlyPortEnabled)
+			}
+		}
 	}
 
 	if d.HasChange("notification_config") {
--- a/mmv1/third_party/terraform/services/container/resource_container_cluster_test.go.erb
+++ b/mmv1/third_party/terraform/services/container/resource_container_cluster_test.go.erb
@@ -6628,7 +6628,9 @@ resource "google_container_cluster" "with_insecure_kubelet_readonly_port_enabled
 
   node_config {
     kubelet_config {
-      cpu_manager_policy                     = "static"
+      # Must be set when kubelet_config is, but causes permadrift unless set to
+      # undocumented empty value
+      cpu_manager_policy                     = ""
       insecure_kubelet_readonly_port_enabled = "%s"
     }
   }
--- a/mmv1/third_party/terraform/services/container/resource_container_cluster_test.go.erb
+++ b/mmv1/third_party/terraform/services/container/resource_container_cluster_test.go.erb
@@ -1568,8 +1568,8 @@ func TestAccContainerCluster_withInsecureKubeletReadonlyPortEnabledInNodeConfigU
 			},
 			{
 				Config: testAccContainerCluster_withInsecureKubeletReadonlyPortEnabledInNodeConfig(clusterName, networkName, subnetworkName, "FALSE"),
-        // TODO: Add check for update; updates for all the resources here
-        // currently broken b/361634104
+				// TODO: Add check for update; updates for all the resources here
+				// currently broken b/361634104
 			},
 			{
 				ResourceName:            "google_container_cluster.with_insecure_kubelet_readonly_port_enabled_in_node_config",
--- a/mmv1/third_party/terraform/services/container/resource_container_cluster_test.go.erb
+++ b/mmv1/third_party/terraform/services/container/resource_container_cluster_test.go.erb
@@ -1555,9 +1555,14 @@ func TestAccContainerCluster_withInsecureKubeletReadonlyPortEnabledInNodeConfigU
 		Steps: []resource.TestStep{
 			{
 				Config: testAccContainerCluster_withInsecureKubeletReadonlyPortEnabledInNodeConfig(clusterName, networkName, subnetworkName, "TRUE"),
+				ConfigPlanChecks: resource.ConfigPlanChecks{
+					PreApply: []plancheck.PlanCheck{
+						acctest.ExpectNoDelete(),
+					},
+				},
 				Check: resource.ComposeTestCheckFunc(
 					resource.TestCheckResourceAttr("google_container_cluster.with_insecure_kubelet_readonly_port_enabled_in_node_config",
-						"node_pool_defaults.0.node_config_defaults.0.insecure_kubelet_readonly_port_enabled", "TRUE"),
+						"node_config.0.kubelet_config.0.insecure_kubelet_readonly_port_enabled", "TRUE"),
 				),
 			},
 			{
--- a/mmv1/third_party/terraform/services/container/resource_container_cluster_test.go.erb
+++ b/mmv1/third_party/terraform/services/container/resource_container_cluster_test.go.erb
@@ -8,6 +8,7 @@ import (
 	"regexp"
 
 	"github.com/hashicorp/terraform-plugin-testing/helper/resource"
+	"github.com/hashicorp/terraform-plugin-testing/plancheck"
 	"github.com/hashicorp/terraform-plugin-testing/terraform"
 	"github.com/hashicorp/terraform-provider-google/google/acctest"
 	"github.com/hashicorp/terraform-provider-google/google/envvar"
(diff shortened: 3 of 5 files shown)


mmv1/third_party/terraform/services/container/resource_container_cluster_test.go.erb
```diff
@@ -6422,6 +6514,65 @@ resource "google_container_cluster" "with_node_config" {
...
+func testAccContainerCluster_withInsecureKubeletReadonlyPortEnabledNodePoolDefaultBool(clusterName, insecureKubeletReadonlyPortEnabled, networkName, subnetworkName string) string {
+	return fmt.Sprintf(`
+resource "google_container_cluster" "with_insecure_kubelet_readonly_port_enabled_node_pool_default" {
+  name               = "%s"
+  location           = "us-central1-f"
+  initial_node_count = 1
+
+  node_pool_defaults {
+    node_config_defaults {
+      insecure_kubelet_readonly_port_enabled = %v
```
wyardley, line 6565: I think this is how it should be (flat at top level, like `logging_variant`, vs nested in a `kubelet_config` block?

Note: this part is not yet implemented in code.

michellepolte-optimizely:
Would be great to get this fixed asap depends on notification from GCP a lot of user will need this.

wyardley:
Just a note that it will be at least a week before I can dig into it much more, and it may be a tricky one to finish / get working. So if anyone has time to take it on, or if anyone from the team that works on this project wants to take over, I will not be at all unhappy.

michellepolte-optimizely:
Do we have any updates here? We have already next week without any progress...

wyardley:
@trodge I know this is in draft still, but any chance you could take a look at what I've got so far? I think there is still some code missing, but I want to make sure I'm going in the right direction (it at least seems to build and work to do a basic plan off of)

Also curious if there are any docs on how, e.g., `mmv1/third_party/terraform/services/container/go/node_config.go.tmpl` and `mmv1/third_party/terraform/services/container/node_config.go.erb` interact (assume this is part of the `tpgtools` transition), and if one generates the other... I asked in the GCP Slack too, but so far didn't hear anything.

Also, I _think_ the `google_container_node_pool` acceptance test should more or less cover the bit that @modular-magician is complaining about, but not sure if it needs to be a standalone test case, or if I'm just missing one of the variants.

trodge:
> @trodge I know this is in draft still, but any chance you could take a look at what I've got so far? I think there is still some code missing, but I want to make sure I'm going in the right direction (it at least seems to build and work to do a basic plan off of)
> 
> Also curious if there are any docs on how, e.g., `mmv1/third_party/terraform/services/container/go/node_config.go.tmpl` and `mmv1/third_party/terraform/services/container/node_config.go.erb` interact (assume this is part of the `tpgtools` transition), and if one generates the other... I asked in the GCP Slack too, but so far didn't hear anything.

They do not interact at the moment. The `.go.tmpl` file will in the future replace the `.go.erb` file when the mmv1 generator is fully migrated from ruby to go. This is also completely separate from the `tpgtools` transition.

> Also, I _think_ the `google_container_node_pool` acceptance test should more or less cover the bit that @modular-magician is complaining about, but not sure if it needs to be a standalone test case, or if I'm just missing one of the variants.

I believe the missing test detection is failing because the provider isn't building, and the provider isn't building because of this error:

```
Error: google-beta/services/container/resource_container_cluster_test.go:1519:13: undefined: testAccContainerCluster_withInsecureKubeletReadonlyPortEnabledInNodeConfig
Error: google-beta/services/container/resource_container_cluster_test.go:1544:117: cannot use false (untyped bool constant) as string value in argument to testAccContainerCluster_withInsecureKubeletReadonlyPortEnabledInNodePoolBool
Error: google-beta/services/container/resource_container_cluster_test.go:1568:108: cannot use true (untyped bool constant) as string value in argument to testAccContainerCluster_withInsecureKubeletReadonlyPortEnabledNodePoolDefaultBool
```
The files and line numbers here refer to the generated code in the downstream.

wyardley:
Hi, thanks for responding.
> They do not interact at the moment. The .go.tmpl file will in the future replace the .go.erb file when the mmv1 generator is fully migrated from ruby to go. This is also completely separate from the tpgtools transition.

Thanks -- this is all super helpful information. Are there any (non-Google / Hashicorp internal) docs that explain this at high level? Do the tests run against both sets, and is there anything I can / should do to help make sure they stay in sync while I'm working on stuff?

> and the provider isn't building because of this error:

Huh, it was building for me locally, though maybe the test code doesn't get generated / built when you run `PRODUCT=container`?  I just tried building ga and beta for all products off my branch, and I can successfully run `make` and a provider is built (`go vet` and `go install` run without error). I will take a look at the errors; is there an easy way to run that build and check for errors without actually running the integration tests locally?

Side note: with some recent change, I don't seem to be able to run the magic modules build on OS X without doing `ulimit -n unlimited` first, which has always worked in the past.

wyardley:
The updated version should fix the basic problem (at least those tests build now locally), though there may still be some issues with the provider code and / or tests.

mmv1/third_party/terraform/services/container/node_config.go.erb
```diff
@@ -1550,6 +1580,13 @@ func flattenSecondaryBootDisks(c []*container.SecondaryBootDisk) []map[string]in
 	return result
 }
 
+func flattenInsecureKubeletReadonlyPortEnabled(c *container.NodeKubeletConfig) bool {
```
wyardley, line 1583: I don't think this is right, and wonder if a flatten function is even necessary, but struggled to get it working properly without one either.

wyardley:
I've made some progress, and am now better setup to do functional and acceptance testing locally. Didn't realize there were all the manual update methods for some of these resources. May yet be a few missing parts.

```
2024-08-10T17:05:07.377-0700 [WARN]  Provider "provider[\"registry.opentofu.org/hashicorp/google\"]" produced an unexpected new value for google_container_cluster.test_insecure_kubelet_readonly_port, but we are tolerating it because it is using the legacy plugin SDK.
    The following problems may be the cause of any confusing errors from downstream operations:
      - .node_config[0].kubelet_config[0].insecure_kubelet_readonly_port_enabled: was cty.False, but now cty.True
```

I think the flatten method (`flattenInsecureKubeletReadonlyPortEnabled`) is still a problem, and I'm still not seeing updates work properly for either `node_config.kubelet_config` or `node_pool_defaults.node_config_defaults` (at least one of those issues lines up with the failing integration test).

mmv1/third_party/terraform/services/container/go/resource_container_cluster.go.tmpl
```diff
@@ -4252,6 +4253,27 @@ func resourceContainerClusterUpdate(d *schema.ResourceData, meta interface{}) er
 		}
 	}
 
+	if d.HasChange("node_pool_defaults") && d.HasChange("node_pool_defaults.0.node_config_defaults.0.insecure_kubelet_readonly_port_enabled") {
```
trodge, line 4256: Based on my reading of the logs for the remaining failing test, I don't think think this is making an update when the value in the config changes from `true` to `false`.

mmv1/third_party/terraform/services/container/node_config.go.erb
trodge: The flatten function probably isn't necessary for this field. What happens if you just use `c.InsecureKubeletReadonlyPortEnabled` instead of calling this function? It looks to me like the `c == nil` case is already covered.

trodge (changes requested):
It looks like there's still a bug when updating the field from `true` to `false`, but all the other tests are passing now.

mmv1/third_party/terraform/services/container/go/resource_container_cluster.go.tmpl
wyardley: Yes, agree, and that's what shows up in my functional testing as well (and, as mentioned below, I'm seeing a similar permadiff type problem with `node_config` as well, when changing the existing value, even though that test is passing):
```
      - .node_config[0].kubelet_config[0].insecure_kubelet_readonly_port_enabled: was cty.False, but now cty.True
```

Or am I missing a critical component or screwing up the data structure somewhere?

FWIW, this is how everything looks at the API level. I implemented it similarly to the logging variant one where it _doesn't_ totally mirror the structure (for `node_pool_defaults`, i.e., there's no `kubelet_config` block there).

```yaml
nodeConfig:
  # [...]
  kubeletConfig:
    cpuCfsQuota: false
    cpuManagerPolicy: static
    insecureKubeletReadonlyPortEnabled: true
# [...]
nodePoolDefaults:
  nodeConfigDefaults:
    loggingConfig:
      variantConfig:
        variant: DEFAULT
    nodeKubeletConfig:
      insecureKubeletReadonlyPortEnabled: true
```

One thing I did find was a duplicate bit for `insecure_kubelet_readonly_port_enabled` in `expandNodeConfig()`, where that should actually come from the nested `expandKubeletConfig()` within it - will push up that update and the other update.

mmv1/third_party/terraform/services/container/node_config.go.erb
wyardley: Thanks, yeah, updated! 👍

mmv1/third_party/terraform/services/container/go/resource_container_cluster.go.tmpl
wyardley: Btw, this is the simple test I'm using locally (switching between true and false, obviously) if it's at all helpful

```hcl
resource "google_container_cluster" "test_insecure_kubelet_readonly_port" {
  name               = "test-insecure-kubelet-readonly-port"
  location           = "us-central1-f"
  initial_node_count = 1
  
  node_config {
    kubelet_config {
      cpu_manager_policy                     = "static"
      insecure_kubelet_readonly_port_enabled = false
    }
  }

  node_pool_defaults {
    node_config_defaults {
      insecure_kubelet_readonly_port_enabled = false
    }
  }

  deletion_protection = false
}
```

wyardley: Confirmed that making a cluster with both set to false, and then toggling both to true, it _does_ work for `node_pool_defaults` (but not `node_config`), but then doesn't work going the other direction. Also confirmed that the value is unset in the API when creating a new cluster with the value set to `false`. So maybe we do need some kind of custom flatten function to handle the value when the API isn't returning those fields since they're nested?

```
2024-08-12T11:25:21.749-0700 [DEBUG] provider.terraform-provider-google: {
2024-08-12T11:25:21.749-0700 [DEBUG] provider.terraform-provider-google:  "update": {
2024-08-12T11:25:21.749-0700 [DEBUG] provider.terraform-provider-google:   "desiredNodeKubeletConfig": {
2024-08-12T11:25:21.749-0700 [DEBUG] provider.terraform-provider-google:    "insecureKubeletReadonlyPortEnabled": true
2024-08-12T11:25:21.749-0700 [DEBUG] provider.terraform-provider-google:   }
2024-08-12T11:25:21.749-0700 [DEBUG] provider.terraform-provider-google:  }
2024-08-12T11:25:21.749-0700 [DEBUG] provider.terraform-provider-google: }
```

mmv1/third_party/terraform/services/container/go/resource_container_node_pool.go.tmpl
```diff
@@ -1412,6 +1412,42 @@ func nodePoolUpdate(d *schema.ResourceData, meta interface{}, nodePoolInfo *Node
 
 	if d.HasChange(prefix + "node_config") {
 
+		if d.HasChange(prefix + "node_config.0.insecure_kubelet_readonly_port_enabled") {
```
wyardley, line 1415: this is another mistake - I think this is nested in `node_config.0.kubelet_config` already covered, and can be moved

mmv1/third_party/terraform/services/container/go/resource_container_cluster.go.tmpl
wyardley: ps - even though we know `NodeConfigDefaults` is not `nil` already, I don't think we know whether `NodeConfigDefaults.NodeKubeletConfig` is or not?

mmv1/third_party/terraform/services/container/go/node_config.go.tmpl
```diff
@@ -1362,6 +1384,8 @@ func flattenNodeConfigDefaults(c *container.NodeConfigDefaults) []map[string]int
 
 	result[0]["containerd_config"] = flattenContainerdConfig(c.ContainerdConfig)
 
+	result[0]["insecure_kubelet_readonly_port_enabled"] = c.NodeKubeletConfig.InsecureKubeletReadonlyPortEnabled
```
trodge, line 1387: Is `c.NodeKubeletConfig == nil` checked here? I saw this error in a build log from the latest run:

```
github.com/hashicorp/terraform-provider-google-beta/google-beta/services/container.flattenNodeConfigDefaults(0xc01eea6ea0)
	/go/src/github.com/modular-magician/terraform-provider-google-beta/google-beta/services/container/node_config.go:1367 +0xea
```

wyardley: Yes, see other comment, and was looking at this / fixing now. I guess it's most consistent to move it back to a `flatten` routine like the others vs. doing it a bit more tersely within here?

mmv1/third_party/terraform/services/container/node_config.go.erb
```diff
@@ -78,6 +78,15 @@ func schemaContainerdConfig() *schema.Schema {
 	}
 }
 
+func schemaInsecureKubeletReadonlyPortEnabled() *schema.Schema {
+	return &schema.Schema{
+		Type: schema.TypeBool,
+		Optional: true,
+		Computed: true,
```
wyardley, line 85: @trodge: do you have thoughts on whether this should be true (here and in the other spot(s) this is used)?

mmv1/third_party/terraform/services/container/go/node_config.go.tmpl
wyardley: Updated in new commit (though this will likely just get us back to the earlier issue).

mmv1/third_party/terraform/services/container/node_config.go.erb
wyardley: Got it. Will give that a shot. Thanks for the additional guidance.

Just to confirm, this mean the value in the config settings will be `ENABLED` or `DISABLED` as well, correct vs. just using this internally in the state? I think get the idea, though IMO, having it be a bool is a lot easier and more consistent from an end-user perspective, and possibly more consistent with other params in this resource. It's also going to be quite a big update to the PR, so I want to make sure I get that part right. It's also a bit ugly if we leave `_enabled` on the parameter name; should we go from
```hcl
insecure_kubelet_readonly_port_enabled = true
```
to
```hcl
insecure_kubelet_readonly_port = "ENABLED" ## or "DISABLED"
```

I had just been working on adding `ForceSendFields` in various spots, though hadn't gotten it fixed yet.

wyardley: ps - If you're able to rework it more easily than all the back and forth getting it updated to the above will entail, totally fine with that as well, though happy to make a pass at it if you want me to update it.

trodge: Leaving as a boolean will be problematic if the API default ever changes, because users will be unable to distinguish between leaving it unspecified and explicitly setting it to false. This is the approach we've taken to similar situations in the past, such as in [org_policy_policy](https://github.com/hashicorp/terraform-provider-google/blob/479c901bf023bfcca8648358e24d3a795828afb3/google/services/orgpolicy/resource_org_policy_policy.go#L137-L141). In that case, all three of `true`, `false`, and `nil` were valid values from the API's perspective. This situation is a little bit different in that the latest version of the API only supports `true` and `false`, but similar in the sense that we don't want to invalidate configs that don't specify a value. Using an enum is the only way to distinguish between users setting no value (as they would in versions where the resource doesn't exist) and setting it to `false`. The API can then decide which default is appropriate when the user indicates no preference.

I agree that

```
insecure_kubelet_readonly_port_enabled = "DISABLED"
```
is undesirable. Either of 
```
insecure_kubelet_readonly_port = "ENABLED" ## or "DISABLED"
```
or
```
insecure_kubelet_readonly_port_enabled = "TRUE" ## or "FALSE"
```

seem clear to me, but I'm not really a frequent user of the resource so I'm not sure which one is the most readable.

I would love for you to keep working on this, since you seem to have a better handle on the complexity of these hand-written resources. However, if you'd rather I pick it up from here I can give it a try.

wyardley:
```
insecure_kubelet_readonly_port_enabled = "TRUE" ## or "FALSE"
```

In that case, the enum would be `TRUE` / `FALSE` vs `ENABLED` / `DISABLED`, right?
I tend to prefer that, because it would hew a little closer to the underlyling APIs, even if it's not a true Terraform boolean (I think it would also simplify the adjustments to the code at this point, though either way there's a whole lot of stuff that needs to get updated). There seems to be some prior art in both directions in the codebase. Can you doublecheck with whoever on the team had an opinion on this to make sure they're good with using `TRUE` / `FALSE` here?


(looks like for the hand written resources, the enum looks like `ValidateFunc: validation.StringInSlice([]string{"DISABLED","ENABLED"}, false),`, correct?)

trodge: Double checked with my tech lead and we're in agreement that enum is the way to go and `"TRUE"`/`"FALSE"` make sense as values to align with the API.

And yes, using `validation.StringInSlice` as seen elsewhere will be ideal here.

wyardley: Great.
> Using an enum is the only way to distinguish between users setting no value (as they would in versions where the resource doesn't exist) and setting it to false.

So are we allowing `""` in the validation as well to represent the default, if I'm understanding you correctly, right?  Or just TRUE / FALSE only?

The flattener would then look like this?

```go
func flattenInsecureKubeletReadonlyPortEnabled(c *container.NodeKubeletConfig) string {
  insecureKubeletReadonlyPortEnabled := ""
  if c != nil && c.InsecureKubeletReadonlyPortEnabled {
    insecureKubeletReadonlyPortEnabled = "TRUE"
  }
  return insecureKubeletReadonlyPortEnabled
}
```

wyardley: > The flattener should return "ENABLED" when the API value is true and "DISABLED" otherwise.

Just want to make sure I'm not getting tangled up here about what's going in which direction... the flattener for, e.g., `KubeletConfig` is subclassed to `container.NodeKubeletConfig`, so if I try to do something like this:

```go
func flattenKubeletConfig(c *container.NodeKubeletConfig) []map[string]interface{} {
[...]
      "insecure_kubelet_readonly_port_enabled": flattenInsecureKubeletReadonlyPortEnabled(c.InsecureKubeletReadonlyPortEnabled),
```

(where `flattenInsecureKubeletReadonlyPortEnabled` returns a string), I'll get an error like:
```
cannot use c.InsecureKubeletReadonlyPortEnabled (variable of type bool) as *"google.golang.org/api/container/v1".NodeKubeletConfig value in argument to flattenInsecureKubeletReadonlyPortEnabled
```

I can push up my non-working updates if that's helpful.

wyardley:
https://github.com/GoogleCloudPlatform/magic-modules/pull/11272#discussion_r1714388043
https://github.com/GoogleCloudPlatform/magic-modules/pull/11272#discussion_r1716022249

Just linking to these in the main comments. High level, we're going to implement these as string values `"TRUE"` and `"FALSE"` within terraform, to avoid issues if / when the API default ever changes (or if new clusters get a different value from existing ones).

mmv1/third_party/terraform/services/container/node_config.go.erb
wyardley: Pushed up the current progress, even though there are obvious failures... can you let me know if things are going in the right direction?

wyardley: @trodge I have things now in a partially working(ish) state, still some issues with `node_config.kubelet_config` on updates, I think. I note with https://github.com/hashicorp/terraform-provider-google/issues/15767 that there are some existing quirks with some of those parameters, but not sure if there's any connection or if I've just got to play with the forcesendfield settings a bit more.

```
--- PASS: TestAccContainerCluster_withInsecureKubeletReadonlyPortEnabledInNodeConfig (554.66s)
--- PASS: TestAccContainerCluster_withInsecureKubeletReadonlyPortEnabledInNodePool (573.91s)
--- PASS: TestAccContainerCluster_withInsecureKubeletReadonlyPortEnabledUpdates (606.97s)
--- PASS: TestAccContainerNodePool_withKubeletConfig (1063.76s)
```

I've done my best to keep stuff updated, but would be great if, once we get all the other stuff squared away, you can also double-check that the go and ruby templates are still identical (including whitespace) and that the docs updates are correct etc.

mmv1/third_party/terraform/services/container/go/node_config.go.tmpl
```diff
@@ -1134,6 +1159,13 @@ func expandWorkloadMetadataConfig(v interface{}) *container.WorkloadMetadataConf
 	return wmc
 }
 
+func expandInsecureKubeletReadonlyPortEnabled(v interface{}) bool {
+	if v == "TRUE" {
```
wyardley, line 1163: Not technically "expanding" here, I guess, since it's coming in as a string? But I think probably best to keep it named this way for consistency?

mmv1/third_party/terraform/services/container/resource_container_cluster_test.go.erb
```diff
@@ -1539,6 +1539,108 @@ func TestAccContainerCluster_withNodeConfig(t *testing.T) {
...
+
+	acctest.VcrTest(t, resource.TestCase{
+		PreCheck:     func() { acctest.AccTestPreCheck(t) },
+		ProtoV5ProviderFactories: acctest.ProtoV5ProviderFactories(t),
+		CheckDestroy: testAccCheckContainerClusterDestroyProducer(t),
+		Steps: []resource.TestStep{
+			// Test API default (no value set in config) first
+			{
+				Config: testAccContainerCluster_withInsecureKubeletReadonlyPortEnabledNodePoolUpdateBaseline(clusterName, networkName, subnetworkName),
+			},
```
wyardley, line 1605: I made a separate config vs. doing some kind of complicated testing, so this one test uses two different configs. I think this is the only / best way (esp. without allowing `""` as an valid value for the string) to test the API default behavior?

michellepolte-optimizely:
Hey, is it maybe possible to finish up this PR? The topic is already open since multiple weeks and we run a bit out of time on this topic. This is an important topic for all GKE users which manage their infrastructure with terraform.

mmv1/third_party/terraform/services/container/go/node_config.go.tmpl
trodge: Yes, expand functions normally turn the config value into the API value.

mmv1/third_party/terraform/services/container/resource_container_cluster_test.go.erb
trodge: Yes, sounds reasonable to me.

wyardley:
Pulling in a comment from @hoskeri here: https://github.com/hashicorp/terraform-provider-google/issues/15208#issuecomment-2299669810

sounds like it should also be implemented in `node_pool_auto_config`

trodge:
> @trodge Pulling in a comment from @hoskeri here: [hashicorp/terraform-provider-google#15208 (comment)](https://github.com/hashicorp/terraform-provider-google/issues/15208#issuecomment-2299669810)
> 
> Basically, sounds like it should also be implemented in `node_pool_auto_config`. I can work on adding that; guessing shouldn't be too difficult. _edit_: see comment [here](https://github.com/hashicorp/terraform-provider-google/issues/15208#issuecomment-2299839033), but it's not actually that trivial, because `node_kubelet_config` doesn't yet seem to be implemented at all in `node_pool_auto_config` in the tf provider. So I'll take the guidance from the Google folks on this, but my suggestion is that it may be better to implement that as a separate followup?

+1 on doing that in a separate PR

> At this point, I could also use some help figuring out the remaining issue when updating the setting, if you or anyone else on the Google team that works on the tf provider are able to provide any guidance / help there.

I'm running these tests locally now as well as in CI :
TestAccContainerCluster_withInsecureKubeletReadonlyPortEnabledInNodeConfig
TestAccContainerCluster_withInsecureKubeletReadonlyPortEnabledInNodePool
TestAccContainerCluster_withInsecureKubeletReadonlyPortEnabledUpdates

The build is succeeding after rebasing on upstream main.

I think we can ignore the missing test report because node pool is a derived resource of cluster, as far as I understand.

wyardley:
> The build is succeeding after rebasing on upstream main.

Yeah, I haven't updated here yet, but rebased locally and will push it up at some point

> I think we can ignore the missing test report because node pool is a derived resource of cluster, as far as I understand.

There is a test for `TestAccContainerNodePool_withKubeletConfig`. I agree we probably don't need to be too perfect about testing redundant stuff in multiple places.

That said, technically, there isn't a test for the situation that's failing for me locally, and IMO, we should _probably_ try to figure that out (and, probably also add a test that covers it). Either way, that's the thing I need the most help sorting out. I could, I guess make a new set of tests, or add `node_config` to `TestAccContainerCluster_withInsecureKubeletReadonlyPortEnabledUpdates` which would have some pros and cons vs. testing separately.

With a regular config (posted above also)
```hcl
resource "google_container_cluster" "test_insecure_kubelet_readonly_port" {
  name               = "test-insecure-kubelet-readonly-port"
  location           = "us-central1-f"
  initial_node_count = 1

  ## this is the one that's not updating properly
  node_config {
    kubelet_config {
      cpu_manager_policy                     = "static"
      insecure_kubelet_readonly_port_enabled = "FALSE"
    }
  }
  node_pool_defaults {
    node_config_defaults {
      insecure_kubelet_readonly_port_enabled = "FALSE"
    }
  }
  deletion_protection = false
```

To review:
* I can make a cluster without the `node_config` or `node_pool_defaults` at all, and (as expected), the values are not set at all in the API or in terraform
* I can set both instances of `insecure_kubelet_readonly_port_enabled` to `"TRUE"` from not being set successfully
* I can make a new cluster with the fields set to either `"TRUE"` or `"FALSE"`, and the correct behavior is observed
* I can update `"FALSE"` to `"TRUE"` for `node_pool_defaults.node_config_defaults.insecure_kubelet_readonly_port_enabled` in place

I can play around with a few other scenarios, but I think the main remaining thing that doesn't work and probably does need to be fixed is the update going from `"FALSE"` to `"TRUE"` for `node_config.node_config_defaults.insecure_kubelet_readonly_port_enabled`.

wyardley:
I do notice that the top level `nodeConfig` schema has `forceNew` set, however, I _think_ this particular setting can be updated in place?

Also worth reading the docs here:
> The nodeConfig field is deprecated and applies only to the default node pool that GKE creates when you create a new Standard mode cluster. The status of the port in the deprecated nodeConfig field doesn't apply to other node pools in the cluster.

Here's the procedure on disabling it on an existing standard cluster:
https://cloud.google.com/kubernetes-engine/docs/how-to/disable-kubelet-readonly-port#disable-standard

I think obviously the main use cases here are
a) People creating new clusters, either directly with the resource or using the https://github.com/terraform-google-modules/terraform-google-kubernetes-engine module or similar
b) People with existing clusters where they've updated (or want to update) this setting

So may be worth stepping back and making sure that we want to expose the `node_config` field, and that we don't need to make any additional docs changes.

wyardley:
Locally, I'm getting this for `TestAccContainerCluster_privateRegistry`; not sure if that's the same error that's happening in CI
```
    vcr_utils.go:152: Step 1/7 error: Error running apply: exit status 1
        
        Error: googleapi: Error 400: Private_registry_access_config requires scope 'cloud-platform' in order to pull certificates. The following node-pools are lacking such scope: [default-pool].
        Details:
        [
          {
            "requestId": "0x2d99b1d1422a69d"
          }
        ]
        , badRequest
        
          with google_container_cluster.primary,
          on terraform_plugin_test.tf line 28, in resource "google_container_cluster" "primary":
          28: resource "google_container_cluster" "primary" {   
```

wyardley:
Getting this locally, both on `main` and on my branch after rebasing; not sure if this is the same error in CI:

```
    vcr_utils.go:152: Step 1/7 error: Error running apply: exit status 1
        
        Error: googleapi: Error 400: Private_registry_access_config requires scope 'cloud-platform' in order to pull certificates. The following node-pools are lacking such scope: [default-pool].
        Details:
        [
          {
            "requestId": "0x934e83a9b36bcd01"
          }
        ]
        , badRequest
        
          with google_container_cluster.primary,
          on terraform_plugin_test.tf line 28, in resource "google_container_cluster" "primary":
          28: resource "google_container_cluster" "primary" {
        
--- FAIL: TestAccContainerCluster_privateRegistry (19.26s)
```

wyardley:
I think #11514 should fix the unrelated failing test

c2thorn (commented):
Couple of things to note:

- `mmv1/third_party/terraform/services/container/go/*` files are inconsequential and will be automatically updated later anyway. The changes to them can be removed from this PR to simplify things, but it doesn't matter either way.
- The 6.0 branch is merging into `main` today, so there is a chance conflicts may appear afterwards.
- We'll still try to get this cherrypicked/backported into a 5.X release so it's not gated by a major upgrade.

wyardley:
> mmv1/third_party/terraform/services/container/go/* files are inconsequential and will be automatically updated later anyway. The changes to them can be removed from this PR to simplify things, but it doesn't matter either way.

@c2thorn thanks! I was asking about this earlier, but don't think I got a firm response. That does make things significantly easier, as it's been a big pain to try and keep them up to date

> The 6.0 branch is merging into main today, so there is a chance conflicts may appear afterwards.

😬 got it. I had been hoping the timing wouldn't work out that way. I'll do my best.

trodge:
> > mmv1/third_party/terraform/services/container/go/* files are inconsequential and will be automatically updated later anyway. The changes to them can be removed from this PR to simplify things, but it doesn't matter either way.
> 
> @c2thorn thanks! I was asking about this earlier, but don't think I got a firm response. That does make things significantly easier, as it's been a big pain to try and keep them up to date
> 
> > The 6.0 branch is merging into main today, so there is a chance conflicts may appear afterwards.
> 
> 😬 got it. I had been hoping the timing wouldn't work out that way. I'll do my best.

I would be happy to help resolve any conflicts, but would like to wait until the merge is complete to do that.

wyardley:
@trodge @c2thorn The good news is that, so far, seems to rebase Ok off of trunk after the 6.0 stuff, so hopefully that also bodes well for being able to cherry-pick it back to 5.x, assuming other big container resource changes (breaking or not) don't go in first🤞

Sorry in advance for all the noise / questions. My analysis, based on the above, is that we have a couple questions / decisions:
* Is the current way of updating this setting for the _default_ cluster node pool correct? Where we _see_ the setting _may_ be different from the way we'd actually update the setting, from what I can see (see above with CLI examples)
* Am I right in thinking that when `node_config.kubelet_config.insecure_kubelet_readonly_port_enabled` we need to trigger an operation on the default node-pool if it exists?
* Is there a good example within the codebase, and / or can someone who knows this resource well, of _how_ to handle this situation (where, from what I can see, we need to actually trigger a nodepool update based on the values changing
* I know you all are busy, but would it be possible for someone from Google to check the existing behavior with `node_config.kubelet_config.cpu_manager_policy` not being able to be updated to `"static"` and see if there is an existing issue there? I can create a GH issue against the provider if needed (side note, though: I can try to set it to `"none"` (what the docs say is the default) vs `"static"` explicitly to see if that avoids the permadrift?

While I don't see it trying to send a `POST` changing that setting at all, from what I can see (from the example above), I think actually updating the default node-pool is a bit more of complex operation than just sending that one API request

As far as the missing test, I know you said it wasn't a big deal, but I do actually think we should make sure to test the node pool variant, but from what I can see, it should be there (and it seems to be passing):
https://github.com/GoogleCloudPlatform/magic-modules/blob/79296d1deab5eee89e279d29c1088064db318f60/mmv1/third_party/terraform/services/container/resource_container_node_pool_test.go.erb#L3140
so do you know why it thinks it's missing? I think that's the right resource and parameters, right?

To see more what's happening under the hood, I created a test cluster again, but ran the operation with `--log-http` and `log_http_redact_token` set to `false`.

What I see mostly is a `PUT` to `uri: https://container.googleapis.com/v1/projects/xxxx/locations/us-central1-f/clusters/test-insecure-kubelet-readonly-port/nodePools/default-pool?alt=json` with the body
```
{"kubeletConfig": {"insecureKubeletReadonlyPortEnabled": false}, "name": "projects/xxxx/locations/us-central1-f/clusters/test-insecure-kubelet-readonly-port/nodePools/default-pool"}
```
[and then it doing the regular set of checks to check on the operation until it completes `https://container.googleapis.com/v1/projects/xxxx/locations/us-central1-f/operations/operation-1724348145235-e4498f66-55e2-4edd-a848-bfcdf06ad6c1?alt=json`]

So I _think_ we have a situation where the setting we're looking at is different from the actual set of things we have to call to update that value, if that makes sense.

wyardley:
Also, if you all want to call this "good enough" in a partially working state, but document that people might have to update the values out of band with the CLI, (or have your team work on a followup PR that fixes this behavior), I'm totally good with that as well. There's probably a balance here between making it perfect and getting it so that people can at least avoid permadiffs, and the GKE module can provide some level of support for adjusting this parameter. But, of course, it would be nicest / most satisfying to get it to work better from the start.

Feel free to reach out on GCP Community Slack if there's anything that's useful to discuss synchronously.

trodge:
From looking at the code for the cluster resource, I don't see any way `node_config.kubelet_config` gets updated, nor any field under `node_config` besides `image_type`. The `node_config` field is marked deprecated in the API documentation and it doesn't look like we support updating anything else under it without recreating the cluster. From my perspective it seems fine to leave this as-is for the purposes of this PR.

I'm not able to reproduce the issue you raised [here](https://github.com/GoogleCloudPlatform/magic-modules/pull/11272#issuecomment-2302483053) with `node_pool_defaults.node_config_defaults.insecure_kubelet_readonly_port_enabled` not updating from true to false when explicitly set.

When I built the provider with this change, I was able to use this config:

```
resource "google_container_cluster" "primary" {
  provider = google-beta
  name               = "test-insecure-kubelet-readonly-port"
  location           = "us-central1-f"
  initial_node_count = 1 

  node_pool_defaults {
    node_config_defaults {
      insecure_kubelet_readonly_port_enabled = "FALSE"
    }   
  }
  deletion_protection = false
}
```
to set `node_pool_defaults.node_config_defaults.insecure_kubelet_readonly_port_enabled` from `"FALSE"` to `"TRUE"` and back again, and at every step `gcloud container clusters describe` showed the field set to the correct value.

wyardley:
> to set node_pool_defaults.node_config_defaults.insecure_kubelet_readonly_port_enabled from "FALSE" to "TRUE" and back again, and at every step gcloud container clusters describe showed the field set to the correct value.

Yes, you're right - I think I misspoke there. That works properly, but `node_config.kubelet_config.insecure_kubelet_readonly_port_enabled` does not, in the case where the value is updated after the cluster and nodepool are created. It can be fixed by adjusting the value out of band outside of terraform, or by recreating the nodepool (but obviously the latter would be overkill, and too destructive to be the default behavior).

I put some more information [here](https://github.com/hashicorp/terraform-provider-google/issues/19225) in the ticket I linked earlier, but I think the same thing is true for other settings within that block as well. So if you are happy with this as-is, and without any docs fixes, you're welcome to ship it, but IMO, the behavior with the default node-pool should be addressed at some point.

trodge (approved):

wyardley:
> The node_config field is marked deprecated in the API documentation and it doesn't look like we support updating anything else under it without recreating the cluster. From my perspective it seems fine to leave this as-is for the purposes of this PR.

👍 I'm good with that if you all are. And I do still have the issue I created for `cpu_manager_policy` and friends open, which would cover this if resolved.

You're right that the API level field we're using to infer whether this is set on the default node-pool is deprecated, but the actual update we'd need to make for updates in that section to work would be a different field, and I think this is the _only_ way the provider currently lets people adjust some of those settings for the default node pool that is created along with the `google_container_cluster` resource. I could make a small docs PR later as a followup that mentions the caveat about updates not working for any of the bits in that section (and even a skipped failing test for it) if there's interest.

I do wonder if we should allow the `""` option as well, like the (undocumented?) `cpu_manager_policy` one, to disambiguate between it not being set at all and being explicitly set to the default setting.

trodge:
Based on my local testing, I don't think this will break anything that is not currently broken.

I was able to create a cluster with `node_config.kubelet_config.cpu_manager_policy = "static"` using the 5.42.0 provider, then refresh using a provider built with this change, see a change outside terraform in the `insecure_kubelet_readonly_port_enabled` field, then refresh again and see the diff go away.

While my testing was not exhaustive, I am reasonably confident that this field in particular will function as intended.

wyardley:
@c2thorn: since I believe @trodge may be on vacation, and you're also flagged as a reviewer, would you be able to take a look at this one now?

melinath:
yes, trodge is on vacation; I'm taking over reviewing this PR. Hi! taking a look through the history now...

mmv1/third_party/terraform/services/container/resource_container_cluster_test.go.erb
```diff
@@ -1539,6 +1539,108 @@ func TestAccContainerCluster_withNodeConfig(t *testing.T) {
...
+	networkName := acctest.BootstrapSharedTestNetwork(t, "gke-cluster")
+	subnetworkName := acctest.BootstrapSubnet(t, "gke-cluster", networkName)
+
+	acctest.VcrTest(t, resource.TestCase{
+		PreCheck:     func() { acctest.AccTestPreCheck(t) },
+		ProtoV5ProviderFactories: acctest.ProtoV5ProviderFactories(t),
+		CheckDestroy: testAccCheckContainerClusterDestroyProducer(t),
+		Steps: []resource.TestStep{
+			{
+				Config: testAccContainerCluster_withInsecureKubeletReadonlyPortEnabledInNodeConfig(clusterName, networkName, subnetworkName, "TRUE"),
```
melinath, line 1554: What would you think about adding steps to update this as well? It could be good to just make sure the bases are covered.

melinath (changes requested):
Thanks for all your work on this PR! Please bear with me as I may have missed some context in the discussion above. But hopefully we can get this in shortly!

Overall this looks good to me. I've left a couple questions & change requests below. I also wanted to ask if you could go ahead and revert the changes to the go files? That would make this PR clearer for anyone taking a look in the future & might make backporting easier. Sorry we didn't clarify this wasn't necessary to keep in sync earlier!

Thanks!

mmv1/third_party/terraform/services/container/resource_container_cluster.go.erb
```diff
@@ -4259,6 +4260,28 @@ func resourceContainerClusterUpdate(d *schema.ResourceData, meta interface{}) er
 		}
 	}
 
+	if d.HasChange("node_pool_defaults") && d.HasChange("node_pool_defaults.0.node_config_defaults.0.insecure_kubelet_readonly_port_enabled") {
```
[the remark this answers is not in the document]
wyardley: I think we got tangled up a bit between the different resources here, but this _should_ discuss the only thing that doesn't work as expected - basically, #19225, so as @trodge mentioned, I believe the bit that's not working is in an area that is already not working, and I think the node_pool_defaults setting does update as expected.

https://github.com/GoogleCloudPlatform/magic-modules/pull/11272#issuecomment-2305839332

mmv1/third_party/terraform/services/container/node_config.go.erb
```diff
@@ -597,6 +609,13 @@ func schemaNodeConfig() *schema.Schema {
 								Optional: true,
 								Description: `Set the CPU CFS quota period value 'cpu.cfs_period_us'.`,
 							},
+							"insecure_kubelet_readonly_port_enabled": {
```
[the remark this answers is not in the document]
wyardley: I went back and forth on this a bit. This setting is used in a bunch of different places, and I wasn't totally sure that the description should be the same for all of them?

At the time, I think it may not have been 100% obvious whether the description should be ultimately the same, as well as whether the `Optional`, `Computed`, etc. settings would be the same. I can update it to use that.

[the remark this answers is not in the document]
melinath: sounds good, thanks!

mmv1/third_party/terraform/services/container/resource_container_cluster_test.go.erb
```diff
@@ -1539,6 +1539,108 @@ func TestAccContainerCluster_withNodeConfig(t *testing.T) {
...
+				ResourceName:            "google_container_cluster.with_insecure_kubelet_readonly_port_enabled_in_node_pool",
+				ImportState:             true,
+				ImportStateVerify:       true,
+				ImportStateVerifyIgnore: []string{"deletion_protection"},
+			},
+		},
+	})
+}
+
+func TestAccContainerCluster_withInsecureKubeletReadonlyPortEnabledUpdates(t *testing.T) {
```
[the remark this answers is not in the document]
wyardley: I am pushing these up in the next update. I'll test it locally before I push up, but wasn't 100% sure whether you wanted this on all the updates after the first or just the second (or third?) one.

You can let me know if it looks right, and I can update. I don't _think_ it would be likely for something to get deleted because of this, but if it works, it's no problem to add it.

mmv1/third_party/terraform/services/container/resource_container_cluster_test.go.erb
wyardley: I've added at least a basic one going from one setting to the other (though not with it missing the setting entirely) - let me know how you think it looks.

wyardley: Ack, should have realized, but this is the thing that is already known to be broken / not yet working (#19225), causing the failure [here](https://github.com/GoogleCloudPlatform/magic-modules/pull/11272#issuecomment-2312981818).

I'd tested locally, but only before adding in the second check.

```
    vcr_utils.go:167: Step 1/4 error: Check failed: Check 1/1 error: google_container_cluster.with_insecure_kubelet_readonly_port_enabled_in_node_config: Attribute 'node_pool_defaults.0.node_config_defaults.0.insecure_kubelet_readonly_port_enabled' expected "TRUE", got "FALSE"
```
This did help me realize that the way I switched the name for the _other_ test was wrong (it's testing "node pool defaults" not the "default pool", which is what _this_ one does). - I'm going to go back to something closer to the original name for _that_ (the other) test (`TestAccContainerCluster_withInsecureKubeletReadonlyPortEnabledDefaultsUpdates`), and also add some short comments before each test.

I could either
* switch this back to how it was
* mark this one as skipped and add a simpler variant that's separate
* leave this test as it is, but temporarily remove the second "Check" (that it's false), which I think will make it pass, but leave us with an evergreen test 😞 

For now, I'm going to do #3 assuming it passes locally; let me know if that seems reasonable and if the comments I've added make sense. I'm squashing all of these down, but let me know if it's easier if I push separate commits going forward for easier review if there are additional changes.

The fact that #3 passes without the `Check` makes me think maybe there's a missing idempotency check that should be in some of these tests too?

Time-permitting, I'm happy to try to take on 19225 and / or the autopilot one after this ships if you've got time to help provide some guidance, though in the end, may be easier / faster for someone from Google to handle those.

mmv1/third_party/terraform/services/container/resource_container_cluster_test.go.erb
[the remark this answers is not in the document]
melinath: yes, it would be great to add these checks for every config step in the update tests you're adding. If it causes something to break, then that means the test wasn't actually exercising the update path.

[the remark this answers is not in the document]
melinath: yep, thanks!

melinath (changes requested):
It looks like you need to add handling for `insecure_kubelet_readonly_port_enabled` to the [flattenNodeConfig](https://github.com/modular-magician/terraform-provider-google/blob/d8c16c18831a0f068ca9484dd08331877019180c/google/services/container/node_config.go#L1304) function - this is what's causing the test failure. The API is working correctly, but that is not being stored in state. If this is fixed, perhaps the update tests will also start working?

melinath:
Small request: if possible, could you use new commits for changes instead of force pushing? That makes it easier to see what's changed.

wyardley:
> It looks like you need to add handling for insecure_kubelet_readonly_port_enabled to the [flattenNodeConfig](https://github.com/modular-magician/terraform-provider-google/blob/d8c16c18831a0f068ca9484dd08331877019180c/google/services/container/node_config.go#L1304) function

https://github.com/modular-magician/terraform-provider-google/blob/d8c16c18831a0f068ca9484dd08331877019180c/google/services/container/node_config.go#L1350

It’s neated so handled in `flattenKubeletConfig`

the issue with updates not working is, I think, unfortunately more complicated. Some more details in some earlier comments, but it’s literally a different endpoint / field that needs to be updated, so it needs to call the node pool update method I think.

melinath:
Yeah you're right, it should be getting flattened there. 🤔  but right now, it's not even getting set properly on Create - and that's not an API problem. The API responses all have the field set correctly, but it's not making its way into state (which is why the Check on the first step of TestAccContainerCluster_withInsecureKubeletReadonlyPortEnabledInNodeConfigUpdates is failing)

melinath:
or at least that's how I would understand the current failure.

wyardley:
>  but right now, it's not even getting set properly on Create

I just had a successful test earlier with essentially the same code. And it's been working functionally up til now as far as the specific situation of creating a new cluster with both settings set from the start. What _wouldn't_ work is if we did the test with a baseline that started with the parameter not being set at all.

> which is why the Check on the first step of TestAccContainerCluster_withInsecureKubeletReadonlyPortEnabledInNodeConfigUpdates is failing)

I was able to reproduce the test failure locally. I think it's maybe as simple as copy / pasta on the resource name (`node_pool_defaults.0.node_config_defaults.0.insecure_kubelet_readonly_port_enabled`) in the `Check` bit.

As far as the bigger issue with updates, [this comment](https://github.com/GoogleCloudPlatform/magic-modules/pull/11272#issuecomment-2302483053) has a bit more about what I think is going on with the bigger issue with updating this setting.

For sure, if I create a new cluster (using the built provider) with the following config:
```hcl
resource "google_container_cluster" "test_insecure_kubelet_readonly_port" {
  name               = "test-insecure-kubelet-readonly-port"
  location           = "us-central1-f"
  initial_node_count = 1

  node_config {
    kubelet_config {
      cpu_manager_policy                     = "static"
      insecure_kubelet_readonly_port_enabled = "TRUE"
    }
  }

  node_pool_defaults {
    node_config_defaults {
      insecure_kubelet_readonly_port_enabled = "TRUE"
    }
  }

  deletion_protection = false
}
```
I see the following:
```
% gcloud container clusters describe --location us-central1-f test-insecure-kubelet-readonly-port | yq '(.nodePoolDefaults, .nodeConfig.kubeletConfig)'
nodeConfigDefaults:
  loggingConfig:
    variantConfig:
      variant: DEFAULT
  nodeKubeletConfig:
    insecureKubeletReadonlyPortEnabled: true
cpuCfsQuota: false
cpuManagerPolicy: static
insecureKubeletReadonlyPortEnabled: true
```

wyardley:
> Small request: if possible, could you use new commits for changes instead of force pushing? That makes it easier to see what's changed.

Yeah, no worries, was asking that above. Happy to do it that way, just let me know if you want me to rebase and / or squash at some point.

mmv1/third_party/terraform/services/container/resource_container_cluster_test.go.erb
wyardley: And yeah, I'll test if it still bombs out on step 3/4 with the plancheck commented out -- if so, we might have to set the update test to skipped, and add separate create tests for true and / or false variants?

melinath: Okay, I see what you mean. I don't think we need tests for the separate true / false variants, but we do need this field to either support in-place update or be marked ForceNew (if it can't be updated in place - we shouldn't intentionally ship a bug.) That's tricky because it would mean every iteration of the field would become ForceNew (including the ones that _can_ be updated in-place.)

melinath: I think that at a minimum extracting the kubelet_config behavior from nodePoolUpdate & replicating it in the cluster Update method seems reasonable as a short-term solution here. There is precedent (in a way) with [changing node_version](https://github.com/GoogleCloudPlatform/magic-modules/blob/24542a8b0ad579c06980ab467ec694e58dff2ab5/mmv1/third_party/terraform/services/container/resource_container_cluster.go.erb#L3776-L3799) (though that uses cluster update instead of node pool update). (I'd also be fine with the full nodePoolUpdate change, though that would probably be best in a separate PR.)

wyardley: > but we do need this field to either support in-place update or be marked ForceNew (if it can't be updated in place - we shouldn't intentionally ship a bug.)

In practical terms, this is a bug that already exists in the code as it is now (https://github.com/hashicorp/terraform-provider-google/issues/19225), as the other fields in `kubelet_config` also cannot be successfully updated in place, which I think is why @trodge had previously approved shipping it with update broken for now. In other words, this update path doesn't work, but it doesn't introduce new broken behavior (other than introducing a new parameter for it). Updating it in place is probably not a common problem, but if someone does experience it, they'll just have permadrift, and could at least adjust the setting at the node-pool level via CLI or in the UI to reconcile the drift.

I had thought about doing `ForceNew`, but I'd argue that it is, actually, worse to ship an update that does that than shipping it as-is -- if someone accidentally does this while trying to disable the Kubelet read only port, they will potentially recreate their default nodepool, which is a much more time-consuming and potentially destructive operation. I think we'd have to force it on _any_ of the settings within the `kubelet_config` section.

> There is precedent (in a way) with [changing node_version](https://github.com/GoogleCloudPlatform/magic-modules/blob/24542a8b0ad579c06980ab467ec694e58dff2ab5/mmv1/third_party/terraform/services/container/resource_container_cluster.go.erb#L3776-L3799) (though that uses cluster update instead of node pool update). 

I'll take a look. That looks reasonable, however, I think there seems to be quite a lot going on in [nodePoolUpdate](https://github.com/GoogleCloudPlatform/magic-modules/blob/24542a8b0ad579c06980ab467ec694e58dff2ab5/mmv1/third_party/terraform/services/container/resource_container_node_pool.go.erb#L1349) including a whole section for `kubelet_config` that does exactly what we want (in fact, it _is_ exactly the same thing, since this is for a node_pool resource, and we're doing it for the `default-pool` nodepool resource created implicitly within the cluster resource):
https://github.com/GoogleCloudPlatform/magic-modules/blob/24542a8b0ad579c06980ab467ec694e58dff2ab5/mmv1/third_party/terraform/services/container/resource_container_node_pool.go.erb#L1778-L1802 -- pulling that plus the bit that handles waiting for the update etc., seems like quite a bit of code to extract.

So _if_ it's reasonable to call that directly, _and_ we can infer the reflected value in the (deprecated, in this case) fields directly within nodeConfig, it might not be too hard to "fix" this.

But I agree it would be better to fix that as a separate PR if at all possible, especially since it would very likely require fixing (and adding tests) for some of the other attributes as well, as well as maybe some other adjustments to the expand / flatten functions for one or more of them.

wyardley: Ok, so _if_ I were going to try to do that, would it be within here, right?:
https://github.com/GoogleCloudPlatform/magic-modules/blob/24542a8b0ad579c06980ab467ec694e58dff2ab5/mmv1/third_party/terraform/services/container/resource_container_cluster.go.erb#L3806-L3837

Would I just look for any changes within `node_config.0.kubelet_config` and run the update operation on that? Or take the narrow approach, and look for `node_config.0.kubelet_config.0.insecure_kubelet_readonly_port_enabled` and then call `updateNodePool` with at least some of the code around then waiting for that operation?

wyardley:
image
@melinath is the way that text displays a bug / known issue (if I copy the text, it all shows up with a single space between each letter, and no extra space between words, like `P l e a s e f i x t h e m t o c o m p l e t e y o u r P R .`, but visually, it shows up in italics). the behavior seems to come and go, but I couldn't sort out what was going on from just quickly reviewing the templates.

mmv1/third_party/terraform/services/container/resource_container_cluster_test.go.erb
wyardley: Ok, I think I've got something working which I've just pushed up.

It may not be perfect, but seems to work in my functional tests, as well as in the unit test, and it's not a _crazy_ amount of extra code. Happy to make any minor adjustments for consistent style, avoiding hard-coded values (if there's a better way to get the name of the default pool without hardcoding it, let me know, ditto for how I'm setting the timeout, etc.), but I would rather if a broader fix for kubelet_config updates gets handled separately later.

The `cpu_manager_policy` permadrift still shows up in my functional testing; that's also related to a different bug, but setting it to `""` in the test should resolve that issue for now.

wyardley: I'm also pushing a slight variation that I can revert if you prefer the other way -- I think this will keep certain things at the top level scope of `kubelet_config` that will _hopefully_ make it easier to fix some of the other stuff in this section later?

melinath:
> @melinath is the way that text displays a bug / known issue (if I copy the text, it all shows up with a single space between each letter, and no extra space between words, like `P l e a s e f i x t h e m t o c o m p l e t e y o u r P R .`, but visually, it shows up in italics). the behavior seems to come and go, but I couldn't sort out what was going on from just quickly reviewing the templates.

it's an issue I've seen before but I don't think there was an issue for it. Created https://github.com/hashicorp/terraform-provider-google/issues/19301 to track.

mmv1/third_party/terraform/services/container/resource_container_cluster_test.go.erb
wyardley: @melinath is that more of less what you were looking for, and do you have a sense which direction you prefer from the last 2 commits?

mmv1/third_party/terraform/website/docs/r/container_cluster.html.markdown
```diff
@@ -1105,6 +1105,8 @@ node_pool_auto_config {
 
 The `node_config_defaults` block supports:
 
+* `insecure_kubelet_readonly_port_enabled` (Optional) Controls whether the kubelet read-only port is enabled for newly created node pools in the cluster. It is strongly recommended to set this to FALSE.
```
melinath, line 1108:
```suggestion
* `insecure_kubelet_readonly_port_enabled` (Optional) Controls whether the kubelet read-only port is enabled for newly created node pools in the cluster. It is strongly recommended to set this to `FALSE`. Possible values: `TRUE`, `FALSE`.
```

mmv1/third_party/terraform/website/docs/r/container_cluster.html.markdown
```diff
@@ -1296,6 +1298,8 @@ value and accepts an invalid `default` value instead. While this remains true,
 not specifying the `kubelet_config` block should be the equivalent of specifying
 `none`.
 
+* `insecure_kubelet_readonly_port_enabled` - (Optional) Controls whether the kubelet read-only port is enabled. It is strongly recommended to set this to FALSE.
```
melinath, line 1301:
```suggestion
* `insecure_kubelet_readonly_port_enabled` - (Optional) Controls whether the kubelet read-only port is enabled. It is strongly recommended to set this to `FALSE`. Possible values: `TRUE`, `FALSE`.
```

mmv1/third_party/terraform/services/container/node_config.go.erb
```diff
@@ -78,6 +78,18 @@ func schemaContainerdConfig() *schema.Schema {
...
 }
 
+// Note: this is a bool internally, but implementing as an enum internally to
+// make it easier to accept API level defaults.
+func schemaInsecureKubeletReadonlyPortEnabled() *schema.Schema {
+	return &schema.Schema{
+		Type:         schema.TypeString,
+		Optional:     true,
+		Computed:     true,
+		Description:  "Controls whether the kubelet read-only port is enabled. It is strongly recommended to set this to FALSE.",
```
melinath, line 88:
```suggestion
		Description:  "Controls whether the kubelet read-only port is enabled. It is strongly recommended to set this to `FALSE`. Possible values: `TRUE`, `FALSE`.",
```

mmv1/third_party/terraform/services/container/resource_container_cluster_test.go.erb
melinath: I like the way you have it - seems like it could make the refactor more clear later. Thanks!

melinath (approved):
LGTM, just a couple minor changes to docs.

melinath (approved):

wyardley:
Thanks for all the feedback / help @melinath and @trodge! Really appreciate it. And thanks for pushing on getting the update thing to work as well, in the end, much happier that it's shipped with that.

Maybe double-check that the release notes as I had them are also good.

wyardley:
We may see some issues related to https://github.com/hashicorp/terraform-provider-google/issues/15767 once people start using this setting -- I believe `cpu_cfs_quota` should mostly be a noop for existing clusters when not set, but could cause some inconsistent behavior if a cluster is newly created with this setting, and without setting `cpu_cfs_quota` to the API default of `true` explicitly -- i.e., the newly created cluster would come up with `cpu_cfs_quota` set to `false` vs. the default behavior of `true`.

melinath:
@wyardley Thanks for sticking with it! Yes, we're planning to backport this into 5.X.

wyardley:
There's a soft tab instead of a hard tab I missed in one spot, not sure if it's worth fixing, especially if they'll get fixed with the move to go templates at some point? I can make a tiny PR to update it if you'd like though.

PR #11287, opened 2024-07-30
functions2 align storage object generation to other storage source attributes

gleichda:
Currently Cloud Functions v2 has a permanent state drift for storage generation due to different implementation between other strings in the storage source and generation.  This aligns generations between all storage source attributes and allows using the generation as a trigger for redeployment

Tests for cloudfunctions2 were already successfully executed locally

```release-note:bug
cloudfunctions2: fixed an issue with `google_cloudfunctions2_function` where `storage_source.generation` would always show a diff
```

Fixes hashicorp/terraform-provider-google/issues/15405


mmv1/products/cloudfunctions2/Function.yaml
```diff
@@ -423,6 +423,7 @@ properties:
               - !ruby/object:Api::Type::Integer
                 name: 'generation'
                 default_from_api: true
+                custom_flatten: 'templates/terraform/custom_flatten/cloudfunctions2_function_source_generation.go.erb'
```
slevenick, line 426: Can we add ignore_read: true to this field instead? From looking at the custom flatten function it seems to just read from state, which is pretty similar to just adding ignore_read

mmv1/templates/terraform/custom_flatten/cloudfunctions2_function_source_generation.go.erb
```diff
@@ -0,0 +1,40 @@
...
+	# limitations under the License.
+-%>
+func flatten<%= prefix -%><%= titlelize_property(property) -%>(v interface{}, d *schema.ResourceData, config *transport_tpg.Config) interface{} {
+	// This flatten function is shared between the resource and the datasource.
+	// TF Input will use the generation from the source object
+	// GET Response will use the generation from the automatically created object
+	// As TF Input and GET response values have different format,
+	// we will return TF Input value to prevent state drift.
+
+	if ObjVal, ok := d.GetOk("build_config.0.source.0.storage_source.0.generation"); ok {
```
slevenick, line 22:
```suggestion
	if objVal, ok := d.GetOk("build_config.0.source.0.storage_source.0.generation"); ok {
```

mmv1/templates/terraform/custom_flatten/cloudfunctions2_function_source_generation.go.erb
```diff
@@ -0,0 +1,40 @@
...
+		v = ObjVal
```
slevenick, line 23: What happens if this value is not set in state? We end up using the `v` passed into the function which doesn't actually work?

slevenick (changes requested):
Can you add a test that fails on the existing provider but will pass with this change?

slevenick (commented):
Can you add a test that fails on the existing provider but will pass with this change?

gleichda:
> Can you add a test that fails on the existing provider but will pass with this change?

Sorry oversaw that comment...

Testing is currently not possible due to hashicorp/terraform-provider-google/issues/14117 I'm unable to retrieve the generation for a GCS object via Terraform that I could use in the test case.

gleichda:
Nevermind. I talked to @prauc and he opened #11335 so I'll update the test once his PR is merged

slevenick (approved):

slevenick (changes requested):
Sorry, there's a lint error on resource_cloudfunctions2_function_test

Can you fix that? Other than that this looks good

slevenick (approved):

PR #11293, opened 2024-07-30
Add deletion_protection field to Folders resource

abd-goog:
Part of b/330143705

Fixes https://github.com/hashicorp/terraform-provider-google/issues/18904

```release-note:breaking-change
google_folders: added `deletion_protection` field to `folders` to make deleting them require an explicit intent. `folder` resources now cannot be destroyed unless `deletion_protection = false` is set for the resource.
```


melinath (changes requested):
Please make sure that deletion_protection gets set in the import function similar to https://github.com/hashicorp/terraform-provider-google-beta/blob/a4bdc631489c4974394ba18123ecf32baa30109a/google-beta/services/container/resource_container_cluster.go#L6488-L6490

abd-goog:
> Please make sure that deletion_protection gets set in the import function similar to https://github.com/hashicorp/terraform-provider-google-beta/blob/a4bdc631489c4974394ba18123ecf32baa30109a/google-beta/services/container/resource_container_cluster.go#L6488-L6490

Added. Thanks!

melinath (changes requested):
I ended up checking with some other folks on the team to reconcile some differences between how this was implemented compared to some other deletion_protection fields and have created https://github.com/GoogleCloudPlatform/magic-modules/pull/11330 to document the best practices.

Basically the changes for you to make (assuming that I got everything down right in that PR) are:

- Set deletion_protection in read (as described in the new docs)
- Add logic to skip API requests if only deletion_protection is being modified (not strictly necessary but good to do)
- Remove setting of deletion protection in import (it seems like it may not be necessary after all if it's getting set in read - sorry to ask you to remove something I just asked you to add!)

I wanted to let you know about these changes in guidance even though they're not fully merged yet so that you can get a head start - but feel free to wait until it's merged if you prefer. (FYI there's a chance that setting during Import actually _does_ need to be there due to some edge case that I wasn't able to figure out, but I think it probably ought to go.)

abd-goog:
@melinath could you please review? 
I don't see any failures in the VCR logs related to this change, lmk if I'm missing something.

melinath:
I don't see anything either - rerunning to see if it's a flake /gcbrun

melinath (changes requested):
It looks like the error is coming from the data source tests, because the [folder datasource](https://github.com/hashicorp/terraform-provider-google-beta/blob/main/google-beta/services/resourcemanager/data_source_google_folder.go) doesn't have `deletion_protection` in its schema. You'll need to add it to the schema (and add a few lines that set it to nil, because it's not meaningful in that context.)

Something like:

```go
# In schema
"deletion_protection": {
	Type:        schema.TypeBool,
	Computed:    true,
},

# In Read
	if err := d.Set("deletion_protection", nil); err != nil {
		return fmt.Errorf("Error setting deletion_protection: %s", err)
	}
```

abd-goog:
> It looks like the error is coming from the data source tests, because the [folder datasource](https://github.com/hashicorp/terraform-provider-google-beta/blob/main/google-beta/services/resourcemanager/data_source_google_folder.go) doesn't have `deletion_protection` in its schema. You'll need to add it to the schema (and add a few lines that set it to nil, because it's not meaningful in that context.)
> 
> Something like:
> 
> ```go
> # In schema
> "deletion_protection": {
> 	Type:        schema.TypeBool,
> 	Computed:    true,
> },
> 
> # In Read
> 	if err := d.Set("deletion_protection", nil); err != nil {
> 		return fmt.Errorf("Error setting deletion_protection: %s", err)
> 	}
> ```

Done, thanks!
Though I wonder why is the schema explicitly redeclared in data source instead of reusing the schema from the resource.

melinath:
> Though I wonder why is the schema explicitly redeclared in data source instead of reusing the schema from the resource.

The google_folder data source almost certainly pre-dates the tooling to reuse the resource schema... we could hypothetically switch it over but there probably hasn't been a reason to (and I'm not 100% sure it would Just Work without doing further investigation.)

melinath (approved):
LGTM, thanks!

PR #11295, opened 2024-07-30
Ignore `deletion_delay_hours` when importing

SarahFrench:

```release-note:none

```


swamitagupta (approved):
LGTM

Thanks for adding this fix Sarah!

Could you please elaborate why  a failed import fails the post delete check here and returns the Private Cloud in a 'found' state? 

```
testing_new.go:90: Error running post-test destroy, there may be dangling resources: exit status 1
        Error: Error waiting to delete PrivateCloud: got pending status "found"
 ```

SarahFrench:
@swamitagupta I think those are two separate problems with the test. First the test fails due to a failed assertion that the state of the imported resource should match its state from when it was being managed via the test framework (which this PR fixes), and then the test framework is reporting a failure because it's unable to delete the resources it provisioned.

I believe "got pending status "found"" is an error surfaced from hashicorp/terraform-plugin-testing, and is due to how the the testing framework is waiting for a 404 after deletion but is still 'finding' the resource when performing GET requests.

BBBmau (approved):

PR #11303, opened 2024-07-31
Add `force_destroy` field to `google_storage_managed_folder`

juliocc:
This PR introduces a `force_destroy` field for `google_storage_managed_folder` resources. 

Fixes hashicorp/terraform-provider-google#18553 which was left as a pending action from https://github.com/GoogleCloudPlatform/magic-modules/pull/10786#discussion_r1653524192 

```release-note:enhancement
storage: added `force_destroy` field to `google_storage_managed_folder` resource
```


mmv1/templates/terraform/custom_update/storage_managed_folder.go.erb
```diff
@@ -0,0 +1,11 @@
+_ = config
+
+// we can only get here if force_destroy was updated
```
SarahFrench, line 3: Thanks for the comments here!

mmv1/third_party/terraform/services/storage/storage_managed_folder_test.go
```diff
@@ -0,0 +1,54 @@
...
+			},
+			{
+				ResourceName:            "google_storage_managed_folder.folder",
+				ImportState:             true,
+				ImportStateVerify:       true,
+				ImportStateVerifyIgnore: []string{"bucket", "force_destroy"},
+			},
+			{
+				Config: testAccStorageManagedFolder_storageManagedFolderUpdate(name, true),
+			},
```
SarahFrench, line 29: This test currently confirms that the virtual field can be updated but it doesn't confirm that the field affects the deletion of the folder.

Ideally this test could contain 2 additional steps that:
- adds a object into the managed folder, making it not empty
- removes the managed folder from the config (triggering a deletion) but the object is still in the config and therefore is present in the folder when it's deleted.

Would you be ok to add these steps?

---

For what it's worth, I can see that the new virtual field is present when the folder is deleted by the testing framework when it cleans up all the resources it creates during the test:

>DELETE /storage/v1/b/tf-test-managed-folder-uwyqek1f9n/managedFolders/managed%2Ffolder%2Fname%2F?allowNonEmpty=true

because of this, if you don't have time we could skip the new test steps being added

SarahFrench: Something like this would be used:

```tf
resource "google_storage_bucket_object" "object" {
  name   = "managed/folder/name/my-file.txt"
  content = "This file will affect the folder being deleted if allowNonEmpty=false"
  bucket = google_storage_bucket.bucket.name
}
```

juliocc: I've added the two extra steps. For the second step used [`testAccCheckGoogleStorageObject`](https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/third_party/terraform/services/storage/resource_storage_bucket_object_test.go#L509) which checks that the object is indeed in the bucket and not just in the state. Let me know if this is OK or if checking the state directly is better/preferable

I've also renamed the file to follow the same convention as the other tests in the storage folder.

juliocc:
> Tests failed during RECORDING mode: `TestAccStorageManagedFolder_storageManagedFolderBasicExample`[[Error message](https://storage.cloud.google.com/ci-vcr-logs/beta/refs/heads/auto-pr-11303/artifacts/66ac3d5d-0493-456e-96aa-17dc5630a5aa/build-log/recording_build/TestAccStorageManagedFolder_storageManagedFolderBasicExample_recording_test.log)] [[Debug log](https://storage.cloud.google.com/ci-vcr-logs/beta/refs/heads/auto-pr-11303/artifacts/66ac3d5d-0493-456e-96aa-17dc5630a5aa/recording/TestAccStorageManagedFolder_storageManagedFolderBasicExample.log)]
> 
> Errors occurred during RECORDING mode. Please fix them to complete your PR.
> 
> View the [build log](https://storage.cloud.google.com/ci-vcr-logs/beta/refs/heads/auto-pr-11303/artifacts/66ac3d5d-0493-456e-96aa-17dc5630a5aa/build-log/recording_test.log) or the [debug log](https://console.cloud.google.com/storage/browser/ci-vcr-logs/beta/refs/heads/auto-pr-11303/artifacts/66ac3d5d-0493-456e-96aa-17dc5630a5aa/recording) for each test

Should I do something about this error? Not sure why it fails now.

mmv1/third_party/terraform/services/storage/storage_managed_folder_test.go
SarahFrench: Checking that the file is still present in GCP after the delete operation might not be necessary as we're just wanting to test that the folder is being successfully deleted by the provider. I assume if the force_destroy value wasn't used in the URL for delete actions by the provider then the API would return an error saying that the folder isn't empty and cannot be deleted. Given that the test isn't failing due to this I think we're good!

I'm happy to either leave in or remove  the use of testAccCheckGoogleStorageObject.

Happy to approve this PR without this but here's an idea: Another improvement to the test could be to make the object be present from the start and have force_destroy=false on the folder. Then attempt to delete the folder in the next step, triggering an API error, but make the test step expect an error with `ExpectError` ([example](https://github.com/hashicorp/terraform-provider-google/blob/abcccfe983b864f0b7390c830879592822d54011/google/services/compute/resource_compute_region_backend_service_test.go#L85)). Then the test would update to force_destroy=true and re-attempt the delete.

mmv1/templates/terraform/examples/storage_managed_folder_basic.tf.erb
```diff
@@ -5,6 +5,7 @@ resource "google_storage_bucket" "bucket" {
 }
 
 resource "google_storage_managed_folder" "<%= ctx[:primary_resource_id] %>" {
-  bucket = google_storage_bucket.bucket.name
-  name   = "managed/folder/name/"
+  bucket        = google_storage_bucket.bucket.name
+  name          = "managed/folder/name/"
+  force_destroy = true
```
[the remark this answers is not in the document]
juliocc: Done

mmv1/third_party/terraform/services/storage/storage_managed_folder_test.go
juliocc: Done. Added the additional test step as suggested and removed testAccCheckGoogleStorageObject call as it makes the code a bit simpler.

SarahFrench (approved):
Thanks for iterating on the tests with me, this looks good to merge now :shipit:

PR #11305, opened 2024-07-31
google_container_cluster: add Node auto-provisioning locations

prauc:
When Node auto-provisioning is enabled, the terraform provider currently cannot be used to define node auto-provisioning locations. This configuration is, as of today, presently limited to the UI, gcloud, or API.

Tests for google_container_cluster were already successfully executed locally

**Release Note Template for Downstream PRs (will be copied)**

```release-note:enhancement
container: added `auto_provisioning_locations` field to `google_container_cluster`
```

PR #11315, opened 2024-07-31
Remove change to state migrate function that was not touched

slevenick:
No changes, this isn't currently being used to generate. Follow-up to my miss of this being added in https://github.com/GoogleCloudPlatform/magic-modules/pull/11305

I think this should be removed, as we don't add new schema values to old state migrate functions

```release-note:none

```


c2thorn (approved):
There's a good chance this would be removed automatically in the next "go rewrite sync" someone does, but it doesn't hurt to move ahead with it now

PR #11318, opened 2024-07-31
Add deletion_protection to v2 cloud run service resource

avianap:
Add deletion_protection to v2 cloud run service resource
[b/317099814](https://b.corp.google.com/issues/317099814)

```release-note:breaking-change
cloudrunv2: added `deletion_protection` field to `google_cloudrunv2_service` to make deleting them require an explicit intent. `google_cloudrunv2_service` resources now cannot be destroyed unless `deletion_protection = false` is set for the resource.
```


trodge (changes requested):
Overall LGTM, but I think we can improve the test coverage to include updating `deletion_protection`. As long as it is `false` in the last step the tests should still be able to clean up properly.

trodge (approved):

emilecaron:
hello. not sure if this is the right place for such a question but i'll give it a try:
how is this not considered a breaking change ? we relied on the fact that deletion was possible by default and over the night this behaviour changes and breaks our pipelines.
I guess most people using features such as https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/cloud_run_v2_job#start_execution_token are also recreating cloud run jobs on every deployment and will have to manually run extra steps to add `deletion_protection=false` before running their regular pipelines again.
Is there any workaround that i would be missing ? 

i guess pinning the provider version was also not the best idea but the issue remains the same

Moumouls:
Hi @trodge, it's not clear if applying a template with protection set to false is sufficient, or if it needs first an apply then a delete ?

Moumouls:
@emilecaron do you have some feedbacks about your usage ? Does protection to false on destroy is sufficient ?

emilecaron:
You must first remove delete protection then delete it. Can't work in a
single apply, which is kinda breaking when some resources have to be
recreated on every deployment

Le mar. 17 déc. 2024, 19:12, Antoine Cormouls ***@***.***> a
écrit :

Moumouls:
thanks @emilecaron for your answer, we tested the change on staging on a old ressource, if we trigger the destroy with the new provider version (from a old state of old provider) it works.
It avoided us a huge pain !
I hope it could help developers landing here, about the behavior from a old version and old state.

PR #11321, opened 2024-08-01
Add retaining unchanged alias ip ranges and handle permadiff caused by the change

karolgorc:
References [issue #12286](https://github.com/hashicorp/terraform-provider-google/issues/12286)
- added checkForCommonAliasIp() function and condition to use it
- Added tests for this scenario
- Maintaining the unchanged ranges causes permadiff because of sorting so flattening function needed to be rewritten 
  - [docs](https://googlecloudplatform.github.io/magic-modules/develop/permadiff/#list-order)

```release-note:bug
compute: fixed maintaining unchanged `alias_ip_range` in `compute_instance` in between PATCH API calls and handled permadiff caused by the change
```

### DIFF
--- a/mmv1/third_party/terraform/services/compute/resource_compute_instance_test.go.erb
+++ b/mmv1/third_party/terraform/services/compute/resource_compute_instance_test.go.erb
@@ -21,6 +21,7 @@ import (
 	"github.com/hashicorp/terraform-plugin-testing/helper/resource"
 	"github.com/hashicorp/terraform-plugin-testing/terraform"
 
+	"github.com/stretchr/testify/assert"
 	"github.com/hashicorp/terraform-provider-google/google/acctest"
 	"github.com/hashicorp/terraform-provider-google/google/envvar"
 	tpgcompute "github.com/hashicorp/terraform-provider-google/google/services/compute"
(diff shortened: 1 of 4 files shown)


mmv1/third_party/terraform/services/compute/resource_compute_instance_test.go.erb
```diff
@@ -7176,6 +7220,162 @@ resource "google_compute_instance" "foobar" {
...
+  network_interface {
+    subnetwork = google_compute_subnetwork.inst-test-subnetwork.self_link
+    alias_ip_range {
+      subnetwork_range_name = "inst-test-secondary"  
+	  ip_cidr_range = "172.16.1.0/24"
+    }
+	alias_ip_range {
+	  subnetwork_range_name = "inst-test-tertiary"
+	  ip_cidr_range = "10.1.3.0/24"
+	}
```
melinath, line 7373:
```suggestion
    alias_ip_range {
      subnetwork_range_name = "inst-test-secondary"  
      ip_cidr_range = "172.16.1.0/24"
    }
    alias_ip_range {
      subnetwork_range_name = "inst-test-tertiary"
      ip_cidr_range = "10.1.3.0/24"
    }
```

mmv1/third_party/terraform/services/compute/resource_compute_instance_test.go.erb
```diff
@@ -1821,15 +1821,59 @@ func TestAccComputeInstance_secondaryAliasIpRange(t *testing.T) {
...
 					testAccCheckComputeInstanceExists(t, "google_compute_instance.foobar", &instance),
 					testAccCheckComputeInstanceHasAliasIpRange(&instance, "", "10.0.1.0/24"),
 				),
 			},
+			computeInstanceImportStep("us-east1-d", instanceName, []string{"network_interface.0.alias_ip_range.0.ip_cidr_range", "network_interface.0.alias_ip_range.0.subnetwork_range_name", "network_interface.0.alias_ip_range.1.ip_cidr_range", "network_interface.0.alias_ip_range.1.subnetwork_range_name"}),
+		},
+	})
+}
+
+func TestAccComputeInstance_aliasIpRangeCommonAddresses(t *testing.T) {
```
melinath, line 1837: Will this actually be able to verify that the mechanism used to make the change is correct? If so, how?

melinath (changes requested):
Overall this makes sense & LGTM - just have some questions & a few minor whitespace changes.

mmv1/third_party/terraform/services/compute/resource_compute_instance_test.go.erb
```diff
@@ -1821,15 +1821,59 @@ func TestAccComputeInstance_secondaryAliasIpRange(t *testing.T) {
 					testAccCheckComputeInstanceHasAliasIpRange(&instance, "inst-test-secondary", "172.16.0.0/24"),
 				),
 			},
-			computeInstanceImportStep("us-east1-d", instanceName, []string{}),
+			computeInstanceImportStep("us-east1-d", instanceName, []string{"network_interface.0.alias_ip_range.0.ip_cidr_range", "network_interface.0.alias_ip_range.0.subnetwork_range_name", "network_interface.0.alias_ip_range.1.ip_cidr_range", "network_interface.0.alias_ip_range.1.subnetwork_range_name"}),
```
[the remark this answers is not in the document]
karolgorc: The tpgresource.SortMapsByConfigOrder function breaks ImportStateVerify test step

[reference](https://googlecloudplatform.github.io/magic-modules/develop/permadiff/#list-order)
> This means that tests for the resource need to ignore the field’s import behavior via ignore_read_extra (for MMv1 examples) or [ImportStateVerifyIgnore](https://developer.hashicorp.com/terraform/plugin/sdkv2/resources/import#importstateverifyignore-1) (for handwritten tests).

All the ranges are still inside the API request but they are in the wrong order causing the test to fail. And the wrong order is causing the permadiff

mmv1/third_party/terraform/services/compute/resource_compute_instance_test.go.erb
karolgorc: These test cases are not verifying this mechanism per se, but rather the end result. See your point though. 

If there is any way to catch the first PATCH call to the API on the fly in the test case I'd be happy to code that in. Just didn't find anything similar

Maybe a unit test for the checkForCommonAliasIp() function will be sufficient to cover this

mmv1/third_party/terraform/services/compute/resource_compute_instance_test.go.erb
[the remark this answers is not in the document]
melinath: ah, I'd forgotten that was a side effect - thanks!

mmv1/third_party/terraform/services/compute/resource_compute_instance_test.go.erb
melinath: hmm... There's no (reasonable) way to catch the first PATCH call on the fly as far as I know, so it might just require verification of the calls in the debug logs.

but adding unit tests for checkForCommonAliasIp certainly wouldn't hurt!

karolgorc: Added the unit tests but haven't found a way to get the provider debug log from within an acc test other than using `stdout`. Which doesn't seem to be safe considering `t.Parallel()`. 

Are there any similar implementations? 
And also do the tests in GitHub Actions run with `TF_LOG=DEBUG` or `TRACE`? To catch any logs to begin with.

melinath: unit tests don't use the Terraform binary so `TF_LOG` wouldn't have any impact. using the `t.Errorf` and `t.Logf` should let you output anything that shouldn't halt the test.

melinath: sorry, you said the acc test. Just a sec.

melinath: yeah getting things to print inside acc tests can be tricky. I assume you're talking about logging resource behavior specifically? I think using the `"github.com/hashicorp/terraform-plugin-log/tflog"` package might be the canonical best practice (though we only use it in a few places - I think most resource use `log`, but I'm pretty sure that's just for historical reasons.

(Side note: Acc tests don't get run in GitHub Actions.)

melinath:
Failures are unrelated

melinath (approved):
LGTM

PR #11333, opened 2024-08-01
fix permadiff issue in google_cloudfunctions2_function service_config.environment_variables

daniel-cit:
fix permadiff issue in google_cloudfunctions2_function service_config.environment_variables https://github.com/hashicorp/terraform-provider-google/issues/18747

```release-note:bug
cloudfunctions2: fixed a "Provider produced inconsistent final plan" bug affecting the `service_config.environment_variables` field in `google_cloudfunctions2_function` resource
```

### DIFF
--- a/mmv1/products/cloudfunctions2/Function.yaml
+++ b/mmv1/products/cloudfunctions2/Function.yaml
@@ -560,6 +560,7 @@ properties:
         description:
           'Environment variables that shall be available during function
           execution.'
+        default_from_api: true
         diff_suppress_func: 'environmentVariablesDiffSuppress'
       - !ruby/object:Api::Type::Integer
         name: 'maxInstanceCount'
--- a/mmv1/products/cloudfunctions2/go_Function.yaml
+++ b/mmv1/products/cloudfunctions2/go_Function.yaml
@@ -511,6 +511,7 @@ properties:
         description:
           'Environment variables that shall be available during function
           execution.'
+        default_from_api: true
         diff_suppress_func: 'environmentVariablesDiffSuppress'
       - name: 'maxInstanceCount'
         type: Integer
--- a/mmv1/templates/terraform/examples/cloudfunctions2_full.tf.erb
+++ b/mmv1/templates/terraform/examples/cloudfunctions2_full.tf.erb
@@ -50,7 +50,8 @@ resource "google_cloudfunctions2_function" "<%= ctx[:primary_resource_id] %>" {
     max_instance_request_concurrency = 80
     available_cpu = "4"
     environment_variables = {
-        SERVICE_CONFIG_TEST = "config_test"
+        SERVICE_CONFIG_TEST      = "config_test"
+        SERVICE_CONFIG_DIFF_TEST = google_service_account.account.email
     }
     ingress_settings = "ALLOW_INTERNAL_ONLY"
     all_traffic_on_latest_revision = true


daniel-cit:
this is the diference from a case where the same changes works
https://github.com/GoogleCloudPlatform/magic-modules/blob/c0299ff1ecd4a6850fac7566591139004a10bbb9/mmv1/products/workbench/Instance.yaml#L428C7-L433C63

SarahFrench (commented):
I'm at the end of my day and I'm OOO Monday and Tuesday so a 'proper' review will be delayed, but I had some thoughts from a quick look at your PR and your discussion in the linked issue that might unblock you in the meantime:

First, do you want to make the changes that you described in [this comment](https://github.com/hashicorp/terraform-provider-google/issues/18747#issuecomment-2261890426) to the `environmentVariablesDiffSuppress` function? If you did want to make that change you'd need to edit this file: [cloudfunctions2_function.go.erb](https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/templates/terraform/constants/cloudfunctions2_function.go.erb). Was there a reason for the different approach in this PR?

Second, it'd be great to have a minimal reproduction of the issue to see what conditions lead to the problem. [Your discussion of the problem here](https://github.com/hashicorp/terraform-provider-google/issues/18747#issuecomment-2261827496) was really useful and it sounds like reproduction of the issue could be achieved with a handful of resources (instead of the module shared in the linked issue). If that's the case it'd be useful to add or update an existing acceptance test to trigger this issue and then use it to confirm your PR solves the issue.

I'll resume review on Wednesday!

daniel-cit:
> I'm at the end of my day and I'm OOO Monday and Tuesday so a 'proper' review will be delayed, but I had some thoughts from a quick look at your PR and your discussion in the linked issue that might unblock you in the meantime:
> 
> First, do you want to make the changes that you described in [this comment](https://github.com/hashicorp/terraform-provider-google/issues/18747#issuecomment-2261890426) to the `environmentVariablesDiffSuppress` function? If you did want to make that change you'd need to edit this file: [cloudfunctions2_function.go.erb](https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/templates/terraform/constants/cloudfunctions2_function.go.erb). Was there a reason for the different approach in this PR?
> 
Hi @SarahFrench Thanks for the review.

The approach for the fix in the discussion was based in a partial understanding of the root cause. 
 
Later, comparing the case for the Cloud Function v2 with other exemples of the same type of permadiff function of a Map attribute that does not have the issue like the [workbench metadata map](https://github.com/GoogleCloudPlatform/magic-modules/blob/c0299ff1ecd4a6850fac7566591139004a10bbb9/mmv1/products/workbench/Instance.yaml#L428C7-L433C63) helped to identify the real root cause.

The missing part was this instruction from the development guide regarding updating the [yaml](https://googlecloudplatform.github.io/magic-modules/develop/permadiff/#default).

This was the only difference form the other cases where the same permadiff function worked and the problem was fixed with only this change.

Should the new test case be added here?
https://github.com/GoogleCloudPlatform/magic-modules/blob/d8061c717111ee9310c85676fa37c0028cf77a4b/mmv1/templates/terraform/examples/cloudfunctions2_full.tf.erb#L52-L54

like

```
    environment_variables = {
        SERVICE_CONFIG_TEST      = "config_test"
        SERVICE_CONFIG_DIFF_TEST = google_storage_bucket.bucket.name
    }
```

SarahFrench (commented):
>Should the new test case be added here?
[magic-modules/mmv1/templates/terraform/examples/cloudfunctions2_full.tf.erb](https://github.com/GoogleCloudPlatform/magic-modules/blob/d8061c717111ee9310c85676fa37c0028cf77a4b/mmv1/templates/terraform/examples/cloudfunctions2_full.tf.erb#L52-L54)
>
>like
>
> ```
>     environment_variables = {
>         SERVICE_CONFIG_TEST      = "config_test"
>         SERVICE_CONFIG_DIFF_TEST = google_storage_bucket.bucket.name
>     }
> ```
>

Yes, that should work. I'm hoping that that change will cause the test(s) generated from that example file to fail when your changes in this PR aren't present 🤞

trodge:
I've opened a draft PR here: https://github.com/GoogleCloudPlatform/magic-modules/pull/11345/files to try out the change of adding `SERVICE_CONFIG_DIFF_TEST = google_storage_bucket.bucket.name`

If it fails as expected, feel free to pull it into this PR and we can confirm the fix.

daniel-cit:
Hi @SarahFrench could you PTAL?

@trodge helped creating a draft PR to [reproduce the error](https://github.com/GoogleCloudPlatform/magic-modules/pull/11345#issuecomment-2270123770) this PR is fixing

SarahFrench (approved):
Approving based on the fact the test fails when the test is updated but the schema isn't ([b65255d](https://github.com/GoogleCloudPlatform/magic-modules/pull/11345/commits/b65255d425ce90b2476b1eda2c479056591c9fab), https://github.com/GoogleCloudPlatform/magic-modules/pull/11345#issuecomment-2270123770), and the test passes here due to the schema changes

PR #11337, opened 2024-08-02
Add new data source `google_cloud_identity_transitive_group_memberships`

SarahFrench:
Fixes https://github.com/hashicorp/terraform-provider-google/issues/18128

This data source pulls data about memberships of a Google Group that are both direct and indirect (i.e via membership of a nested group) whereas the previous `google_cloud_identity_group_memberships` data source returns only direct members. It's also worth noting that the data describing a membership differs between the two endpoints used for these two datasources. Because of this I didn't reuse Schema from related resources.

```release-note:new-datasource
`google_cloud_identity_transitive_group_memberships`
```

### DIFF
--- a/mmv1/third_party/terraform/services/cloudidentity/data_source_cloud_identity_group_transitive_memberships.go.erb
+++ b/mmv1/third_party/terraform/services/cloudidentity/data_source_cloud_identity_group_transitive_memberships.go.erb
@@ -1,0 +1,179 @@
+<% autogen_exception -%>
+package cloudidentity
+
+import (
+	"fmt"
+	"time"
+
+	"github.com/hashicorp/terraform-provider-google/google/tpgresource"
+	transport_tpg "github.com/hashicorp/terraform-provider-google/google/transport"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+<% unless version == 'ga' -%>
+	cloudidentity "google.golang.org/api/cloudidentity/v1beta1"
+<% else -%>
+	"google.golang.org/api/cloudidentity/v1"
+<% end -%>
+)
+
+func DataSourceGoogleCloudIdentityGroupTransitiveMemberships() *schema.Resource {
+
+	return &schema.Resource{
+		Read: dataSourceGoogleCloudIdentityGroupTransitiveMembershipsRead,
+
+		// We don't reuse schemas from google_cloud_identity_group_membership because data returned about
+		// transative memberships is structured differently, with information like expiry missing.
+		Schema: map[string]*schema.Schema{
+			"group": {
+				Type:             schema.TypeString,
+				Required:         true,
+				ForceNew:         true,
+				DiffSuppressFunc: tpgresource.CompareSelfLinkOrResourceName,
+				Description:      `The name of the Group to get memberships from.`,
+			},
+			"memberships": {
+				Type:        schema.TypeList,
+				Computed:    true,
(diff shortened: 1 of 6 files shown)


mmv1/third_party/terraform/services/cloudidentity/data_source_cloud_identity_group_transitive_memberships.go.erb
```diff
@@ -0,0 +1,179 @@
...
+
+import (
+	"fmt"
+	"time"
+
+	"github.com/hashicorp/terraform-provider-google/google/tpgresource"
+	transport_tpg "github.com/hashicorp/terraform-provider-google/google/transport"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+<% unless version == 'ga' -%>
```
slevenick, line 12: Do we need to switch the client version? It looks like the schemas are the same between beta & GA, so we could just use the v1 client for both

mmv1/third_party/terraform/services/cloudidentity/data_source_cloud_identity_group_transitive_memberships.go.erb
```diff
@@ -0,0 +1,179 @@
...
+		return nil
+	})
+	if err != nil {
+		return transport_tpg.HandleDataSourceNotFoundError(err, d, fmt.Sprintf("CloudIdentityGroupMemberships %q", d.Id()), "")
+	}
+
+	if err := d.Set("memberships", result); err != nil {
+		return fmt.Errorf("Error setting memberships: %s", err)
+	}
+	d.SetId(time.Now().UTC().String())
```
slevenick, line 151: Is this a typical way to set an ID in a datasource? Do we not want it to be something related to this datasource in particular?

mmv1/third_party/terraform/services/cloudidentity/data_source_cloud_identity_group_transitive_memberships.go.erb
SarahFrench: Good spot, thanks! I'll remove

SarahFrench: Removed in [60da2c4](https://github.com/GoogleCloudPlatform/magic-modules/pull/11337/commits/60da2c4125361dfa836c24c0818114d53f2c0132)

mmv1/third_party/terraform/services/cloudidentity/data_source_cloud_identity_group_transitive_memberships.go.erb
SarahFrench: Id format changed in [c28db4e](https://github.com/GoogleCloudPlatform/magic-modules/pull/11337/commits/c28db4ed5e95535eae456a2ca2357e74d2c7eca5) to `groups/{group_id}/transitiveMemberships`

mmv1/third_party/terraform/services/cloudidentity/data_source_cloud_identity_group_transitive_memberships.go.erb
SarahFrench: Ah, this is actually needed. In the data source we use `config.NewCloudIdentityClient` and in the Beta version of the provider that function is declared using the beta version of client. I've rebased to remove the commit above

slevenick (approved):

SarahFrench:
Thanks for the review!

PR #11344, opened 2024-08-02
Don't send labels in POST request to create PSC forwarding rules

zli82016:
Don't send labels in POST request to create PSC forwarding rules. labels will be set in a separate request after the POST request is complete.

fixes https://github.com/hashicorp/terraform-provider-google/issues/17334

```release-note:bug
compute: fixed the bug that creation of PSC forwarding rules fails in `google_compute_forwarding_rule` resource when provider default labels are set
```

### DIFF
--- a/mmv1/third_party/terraform/services/compute/go/resource_compute_forwarding_rule_test.go.tmpl
+++ b/mmv1/third_party/terraform/services/compute/go/resource_compute_forwarding_rule_test.go.tmpl
@@ -191,6 +191,7 @@ func TestAccComputeForwardingRule_forwardingRuleVpcPscExampleUpdate(t *testing.T
 				ResourceName:      "google_compute_forwarding_rule.default",
 				ImportState:       true,
 				ImportStateVerify: true,
+				ImportStateVerifyIgnore: []string{"labels", "terraform_labels"},
 			},
 			{
 				Config: testAccComputeForwardingRule_forwardingRuleVpcPscExampleUpdate(context, false),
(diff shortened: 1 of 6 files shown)


mmv1/third_party/terraform/services/compute/go/resource_compute_forwarding_rule_test.go.tmpl
```diff
@@ -191,6 +191,7 @@ func TestAccComputeForwardingRule_forwardingRuleVpcPscExampleUpdate(t *testing.T
 				ResourceName:      "google_compute_forwarding_rule.default",
 				ImportState:       true,
 				ImportStateVerify: true,
+				ImportStateVerifyIgnore: []string{"labels", "terraform_labels"},
```
ScottSuarez, line 194: If labels are set in a post_create function why do we still need to ignore labels in the State verify? These should be set? I worry that this is a implication that they are not.

zli82016: The resource `google_compute_forwarding_rule` applies the new labels model. With the new labels model, the `labels` and `terraform_labels` fields are decided based on state. After importing, `labels` and `terraform_labels` will be empty as no value for these two fields in the state. 

https://github.com/hashicorp/terraform-provider-google/blob/main/google/services/compute/resource_compute_forwarding_rule.go#L1390

https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/third_party/terraform/website/docs/guides/version_5_upgrade.html.markdown?plain=1#L226

ScottSuarez (approved):

mmv1/third_party/terraform/services/compute/go/resource_compute_forwarding_rule_test.go.tmpl
ScottSuarez: Makes sense. Thank you for explaining.

PR #11348, opened 2024-08-02
Track what parts of a field changed on a fieldDiff object

melinath:
I think this was something you'd suggested a while ago. I'm implementing it now because it will make things a little easier to understand for cases like ExactlyOneOf - we can rely on whether it was marked as changed rather than having to re-check it.

```release-note:none

```

### DIFF
--- a/tools/diff-processor/diff/diff.go
+++ b/tools/diff-processor/diff/diff.go
@@ -1,6 +1,7 @@
 package diff
 
 import (
+	"fmt"
 	"reflect"
 
 	"github.com/google/go-cmp/cmp"
@@ -23,6 +24,7 @@ type ResourceConfigDiff struct {
 }
 
 type FieldDiff struct {
+	Changed map[string]bool
 	Old *schema.Schema
 	New *schema.Schema
 }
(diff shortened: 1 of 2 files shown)


trodge (commented):
I'm not sure if this is the approach we want to take. It might be less confusing to keep the logic for detecting changes at the rule level.

That being said, I'm not sure how awkward it would be to implement detection for `exactly_one_of` field changes based strictly on comparing `Old` and `New` for the parent fields.

melinath:
cc @shuyama1 who will be interacting with this shortly

If this seems too confusing I'm fine with us not doing it. FWIW the impetus was mostly that for slices (like ExactlyOneOf) the check for whether they're equal is relatively complicated:

```
(len(oldField.ConflictsWith) > 0 || len(newField.ConflictsWith) > 0) && !cmp.Equal(oldField.ConflictsWith, newField.ConflictsWith, cmpopts.SortSlices(less))
```

But there are also other ways we could handle that, like just duplicating that logic (since we'd need parts of it anyway) or factoring the logic out into a helper function.

melinath:
closing this as there is no urgency to this change and we're not aligned on the approach.

PR #10560, opened 2024-04-30
Added logic for merging current and new binding for google_bigquery_dataset_iam_member

sachinpro:
fix(hashicorp/terraform-provider-google#11218): Added logic for merging current and new binding for google_bigquery_dataset_iam_member

**Release Note Template for Downstream PRs (will be copied)**

```release-note:bug
Makes `google_bigquery_dataset_iam_member` non-authoritative
```

PR #11352, opened 2024-08-05
fix(hashicorp/terraform-provider-google#11218): Added logic for mergi…

sachinpro:
fix(hashicorp/terraform-provider-google#11218): Added logic for merging current and new binding for google_bigquery_dataset_iam_member

```release-note:bug
bigquery: made `google_bigquery_dataset_iam_member` non-authoritative. To remove a bigquery dataset iam member, use an authoritative resource like `google_bigquery_dataset_iam_policy`
```

### DIFF
--- a/mmv1/third_party/terraform/services/bigquery/iam_bigquery_member_dataset.go
+++ b/mmv1/third_party/terraform/services/bigquery/iam_bigquery_member_dataset.go
@@ -1,0 +1,326 @@
+// Copyright (c) HashiCorp, Inc.
+// SPDX-License-Identifier: MPL-2.0
+/*
+	This file is a copy of mmv1/third_party/terraform/services/bigquery/iam_bigquery_dataset.go
+	with new functions mergeAccess and GetCurrentResourceAccess
+*/
+package bigquery
+
+import (
+	"errors"
+	"fmt"
+	"strings"
+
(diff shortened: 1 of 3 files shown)


mmv1/third_party/terraform/services/bigquery/iam_bigquery_member_dataset.go
```diff
@@ -0,0 +1,326 @@
+// Copyright (c) HashiCorp, Inc.
+// SPDX-License-Identifier: MPL-2.0
+/*
```
sachinpro, line 3: This is added to fix https://github.com/GoogleCloudPlatform/magic-modules/pull/10560/files#r1586840421 .

c2thorn (approved):

oprudkyi:
> bigquery: made `google_bigquery_dataset_iam_member` non-authoritative. To remove a bigquery dataset iam member, use an authoritative resource like `google_bigquery_dataset_iam_policy`

we used next approach

google_bigquery_dataset_iam_member used to give permissions (we can't use authoritative, because there are hundreds of root modules)
removing permissions via deleting google_bigquery_dataset_iam_member
now we have huge security issue because deleting google_bigquery_dataset_iam_member doesn't remove real iam

PR #11357, opened 2024-08-05
Migrate google_compute_firewall_policy resource from DCL to MMv1

Samir-Cit:
Hello folks.
This PR is to move the Compute Firewall Policy (`google_compute_firewall_policy`) resource from DCL to MMv1. And this PR includes:
- Creation of the resource on MMv1.
- Creation of the examples and tests.
- Remove references to DCL resource.

_**Note:** This PR doesn't include the deletion of the DCL resource. This change will be done on the [Delarative Resource Client Library](https://github.com/GoogleCloudPlatform/declarative-resource-client-library/tree/main) repository._

```release-note:enhancement
compute: `google_compute_firewall_policy` now uses MMv1 engine instead of DCL.
```


mmv1/third_party/terraform/services/compute/resource_compute_firewall_policy_association_test.go
trodge, line 1: Do these test changes need to be included in this PR? They seem to be tests for a different resource.

mmv1/third_party/terraform/services/compute/resource_compute_firewall_policy_rule_test.go.tmpl
trodge, line 1: Why are there changes to these firewall policy rule tests when this PR is for migrating the firewall policy resource?

mmv1/third_party/terraform/services/compute/resource_compute_firewall_policy_test.go.tmpl
```diff
@@ -49,34 +49,34 @@ func TestAccComputeFirewallPolicy_update(t *testing.T) {
...
+	return fmt.Sprintf(`
 resource "google_folder" "folder" {
   display_name = "%s"
   parent       = "%s"
   deletion_protection = false
 }
 
 resource "google_compute_firewall_policy" "default" {
-  parent      = google_folder.folder.name
+  parent      = "%s"
```
trodge, line 61: Why is the parent of this resource changing from the folder to the org?

trodge (changes requested):
I'm worried that these test changes might be concealing bugs.

Samir-Cit:
@shuyama1, @trodge and @slevenick 
Can you do the Code Review in this Pull Request?

This is the last one regading move the Firewall Policy resources from DCL to MMv1.

Thanks.

mmv1/products/compute/FirewallPolicy.yaml
```diff
@@ -0,0 +1,108 @@
...
+  For more information see the [official documentation](https://cloud.google.com/vpc/docs/firewall-policies)
+min_version: 'beta'
+references:
+  guides:
+  api: 'https://cloud.google.com/compute/docs/reference/rest/v1/firewallPolicies'
+docs:
+base_url: 'locations/global/firewallPolicies'
+self_link: 'locations/global/firewallPolicies/{{name}}'
+create_url: 'locations/global/firewallPolicies?parentId={{parent}}'
+update_verb: 'PATCH'
```
trodge, line 31: Should this resource be using `legacy_long_form_project`?

Samir-Cit: This resource is not bind to the project.
I tryied to use this `legacy_long_form_project` and the resource fails to create because it remove the reference from the parent.

Eg.: You are using the parent as `organizations/<ORG_ID>`, the code removes the **organizations/** so it will try to use only the **<ORG_ID>** and then throws the following error
```
{
  "error": {
    "code": 404,
    "message": "The resource '<ORG_ID>' was not found",
    "errors": [
      {
        "message": "The resource '<ORG_ID>' was not found",
        "domain": "global",
        "reason": "notFound"
      }
    ]
  }
}
```

maxi-cit:
Hello @trodge, I updated this PR could you run tests again please?

trodge (approved):

PR #11255, opened 2024-07-25
Make deletion_policy field for Project resource `PREVENT` by default

abd-goog:
Make deletion_policy field 'PREVENT' by default to make deletion actions require an explicit intent.
Deprecate skip_delete field in favor of this change.
Part of b/330143705

Part of  https://github.com/hashicorp/terraform-provider-google/issues/18775

**Release Note Template for Downstream PRs (will be copied)**
```release-note:breaking-change
resourcemanager: Make `deletion-policy` in `google_project` 'PREVENT' by default. This makes deleting them require an explicit intent. `google_project` resources cannot be destroyed unless `deletion_policy` is set to 'ABANDON' or 'NONE' for the resource.
```

PR #11195, opened 2024-07-18
add affected test changes and deletion_protection field

abd-goog:
**Release Note Template for Downstream PRs (will be copied)**

```release-note:REPLACEME

```

PR #11359, opened 2024-08-05
Add deletion_policy field to Project resource

abd-goog:
Add deletion_policy field to make deletion actions require an explicit intent.
Part of b/330143705

Part of  https://github.com/hashicorp/terraform-provider-google/issues/18775

```release-note:enhancement
resourcemanager: added `deletion_policy` field to replace `skip_delete` field in `google_project` resource. Setting `deletion_policy` to `PREVENT` will protect the project against any destroy actions caused by a terraform apply or terraform destroy. Setting `deletion_policy` to `ABANDON` allows the resource to be abandoned rather than deleted and it behaves the same with `skip_delete = true`. Default value is `DELETE`.
```

### DIFF
--- a/mmv1/third_party/terraform/services/resourcemanager/resource_google_project.go
+++ b/mmv1/third_party/terraform/services/resourcemanager/resource_google_project.go
@@ -13,6 +13,7 @@ import (
 	"github.com/hashicorp/errwrap"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/customdiff"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/validation"
 
 	tpgcompute "github.com/hashicorp/terraform-provider-google/google/services/compute"
 	tpgserviceusage "github.com/hashicorp/terraform-provider-google/google/services/serviceusage"
@@ -72,6 +73,14 @@ func ResourceGoogleProject() *schema.Resource {
 				Optional:    true,
 				Computed:    true,
 				Description: `If true, the Terraform resource can be deleted without deleting the Project via the Google API.`,
+			},
+			"deletion_policy": {
+				Type:     schema.TypeString,
+				Optional: true,
+				Default:  "NONE",
+				Description: `The deletion policy for the Project. Setting PREVENT will protect the project against any destroy actions caused by a terraform apply or terraform destroy. Setting ABANDON allows the resource
+				to be abandoned rather than deleted. Possible values are: "PREVENT", "ABANDON", "NONE"`,
+				ValidateFunc: validation.StringInSlice([]string{"PREVENT", "ABANDON", "NONE"}, false),
 			},
 			"auto_create_network": {
 				Type:        schema.TypeBool,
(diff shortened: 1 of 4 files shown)


mmv1/third_party/terraform/services/resourcemanager/resource_google_project_test.go
```diff
@@ -350,6 +351,65 @@ func testAccCheckGoogleProjectHasNoLabels(t *testing.T, r, pid string) resource.
...
+				ImportStateVerify:       true,
+				ImportStateVerifyIgnore: []string{"skip_delete", "deletion_policy"},
+			},
+			{
+				Config:      testAccProject_noAllowDestroy(pid, org),
+				Destroy:     true,
+				ExpectError: regexp.MustCompile("deletion_policy"),
+			},
+			{
+				Config: testAccProject(pid, org),
```
zli82016, line 378: testAccProject is not defined.

abd-goog: Ahh missed it. Updated.

zli82016:
You need to add `deletion_policy = "NONE"` to all of the places with `google_project` as it is doing in https://github.com/GoogleCloudPlatform/magic-modules/pull/11255/files

abd-goog:
> You need to add `deletion_policy = "NONE"` to all of the places with `google_project` as it is doing in https://github.com/GoogleCloudPlatform/magic-modules/pull/11255/files

Could you please clarify why? The other PR requires it because it was PREVENT by default and hence it broke post test destroy process. In this PR on the other hand its NONE by default and therefore doesn't affect existing flows.

zli82016:
> > You need to add `deletion_policy = "NONE"` to all of the places with `google_project` as it is doing in https://github.com/GoogleCloudPlatform/magic-modules/pull/11255/files
> 
> Could you please clarify why? The other PR requires it because it was PREVENT by default and hence it broke post test destroy process. In this PR on the other hand its NONE by default and therefore doesn't affect existing flows.

Oh, right.

zli82016 (approved):
LGTM.

zli82016:
Added @slevenick as the secondary review to double check.

zli82016:
I forgot that one more change is needed in the read method. Otherwise, this will be a breaking change.  Please find more information on the PR https://github.com/GoogleCloudPlatform/magic-modules/pull/11330. 

```
	// Explicitly set client-side fields to default values if unset
	if _, ok := d.GetOkExists("deletion_policy"); !ok {
		if err := d.Set("deletion_policy", "NONE"); err != nil {
			return fmt.Errorf("Error setting deletion_policy: %s", err)
		}
	}
```

abd-goog:
@slevenick could you please approve? This is blocking the breaking change on major release branch.

zli82016 (approved):
LGTM. I tested the projects locally by upgrading the provider with this PR and no perma diffs in the plan after upgrading.

mmv1/third_party/terraform/services/resourcemanager/resource_google_project.go
```diff
@@ -73,6 +74,14 @@ func ResourceGoogleProject() *schema.Resource {
 				Computed:    true,
 				Description: `If true, the Terraform resource can be deleted without deleting the Project via the Google API.`,
 			},
+			"deletion_policy": {
+				Type:     schema.TypeString,
+				Optional: true,
+				Default:  "NONE",
+				Description: `The deletion policy for the Project. Setting PREVENT will protect the project against any destroy actions caused by a terraform apply or terraform destroy. Setting ABANDON allows the resource
+				to be abandoned rather than deleted. Possible values are: "PREVENT", "ABANDON", "NONE"`,
+				ValidateFunc: validation.StringInSlice([]string{"PREVENT", "ABANDON", "NONE"}, false),
```
zli82016, line 83: Can you change `NONE` to `DELETE` to match with other resources?

zli82016:
Added @melinath  as the secondary reviewer.

mmv1/third_party/terraform/services/resourcemanager/resource_google_project.go
abd-goog: You had suggested NONE earlier in https://github.com/GoogleCloudPlatform/magic-modules/pull/11195#issuecomment-2249018282 so could you please confirm it with the 'team' if DELETE is preferred over NONE?  I would appreciate not having a back and forth on this.

zli82016: The team prefer `DELETE` over `NONE`. Sorry about that.

abd-goog: Got it. Done.

melinath (approved):

zli82016 (approved):
LGTM. Thanks.

PR #11372, opened 2024-08-06
Removed cloud armor tests that modify the default test project

melinath:
Fixed https://github.com/hashicorp/terraform-provider-google/issues/18659.

It turns out that these tests modify the global project in a way that is incompatible with other tests (https://github.com/hashicorp/terraform-provider-google/issues/18950). We shouldn't actually need tests for the project being unset, since we're just trying to test the API behavior (which is already tested by the tests which set the project.)

```release-note:none

```


zli82016 (approved):
LGTM.

PR #11381, opened 2024-08-06
add `google_kms_crypto_key_version_latest`

BBBmau:
Currently their is no way for users to grab the latest crypto key version without needing to use a separate resource that can perform a get of `cryptoKeyVersions` in order to grab the latest version from a latest.

This PR addes a `crypto_key_version` data source that takes in a `crypto_key` id in order to output the latest cryptoVersion for the specified crypto_key

`simple tfconfig`:
```hcl
resource "google_kms_key_ring" "key_ring" {
  project  = "<INSERT-PROJECT>"
  name     = "<INSERT-KEY-RING>"
  location = "us-central1"
}

resource "google_kms_crypto_key" "crypto_key" {
  name     = "<INSERT-CRYPTO-KEY>"
  key_ring = google_kms_key_ring.key_ring.id
  purpose  = "ASYMMETRIC_SIGN"

  version_template {
    algorithm = "EC_SIGN_P256_SHA256"
  }
}

data "google_kms_crypto_key_version_latest" "latest_version_data_source" {
  crypto_key = google_kms_crypto_key.crypto_key.id
}

output "latest_version" {
  value = data.google_kms_crypto_key_version_latest.latest_version_data_source.version
}
```

The `GET` request of cryptoVersions:
```hcl
 GET /v1/projects/hc-terraform-testing/locations/us-central1/keyRings/mau-latest-test/cryptoKeys/mau-crypto-key-test/cryptoKeyVersions?alt=json HTTP/1.1
 Host: cloudkms.googleapis.com
 User-Agent: Terraform/1.8.0-alpha20240131 (+https://www.terraform.io) Terraform-Plugin-SDK/2.33.0 terraform-provider-google/dev
 Accept-Encoding: gzip


 2024/08/06 15:27:21 [DEBUG] Google API Response Details:
 HTTP/2.0 200 OK
 Alt-Svc: h3=":443"; ma=2592000,h3-29=":443"; ma=2592000
 Cache-Control: private
 Date: Tue, 06 Aug 2024 22:27:21 GMT
 Server: ESF
 Vary: Origin
 Vary: X-Origin
 Vary: Referer
 X-Frame-Options: SAMEORIGIN
 X-Xss-Protection: 0

 {
   "cryptoKeyVersions": [
     {
       "name": "projects/hc-terraform-testing/locations/us-central1/keyRings/mau-latest-test/cryptoKeys/mau-crypto-key-test/cryptoKeyVersions/1",
       "state": "DESTROY_SCHEDULED",
       "createTime": "2024-08-01T22:50:13.684360401Z",
       "destroyTime": "2024-08-31T23:07:49.100169Z",
       "protectionLevel": "SOFTWARE",
       "algorithm": "EC_SIGN_P256_SHA256",
       "generateTime": "2024-08-01T22:50:13.757083972Z"
     },
     {
       "name": "projects/hc-terraform-testing/locations/us-central1/keyRings/mau-latest-test/cryptoKeys/mau-crypto-key-test/cryptoKeyVersions/2",
       "state": "ENABLED",
       "createTime": "2024-08-06T20:20:32.503393705Z",
       "protectionLevel": "SOFTWARE",
       "algorithm": "EC_SIGN_P256_SHA256",
       "generateTime": "2024-08-06T20:20:32.539700895Z"
     },
     {
       "name": "projects/hc-terraform-testing/locations/us-central1/keyRings/mau-latest-test/cryptoKeys/mau-crypto-key-test/cryptoKeyVersions/3",
       "state": "ENABLED",
       "createTime": "2024-08-06T20:20:35.716831348Z",
       "protectionLevel": "SOFTWARE",
       "algorithm": "EC_SIGN_P256_SHA256",
       "generateTime": "2024-08-06T20:20:35.753066470Z"
     },
     {
       "name": "projects/hc-terraform-testing/locations/us-central1/keyRings/mau-latest-test/cryptoKeys/mau-crypto-key-test/cryptoKeyVersions/4",
       "state": "ENABLED",
       "createTime": "2024-08-06T20:20:38.344795186Z",
       "protectionLevel": "SOFTWARE",
       "algorithm": "EC_SIGN_P256_SHA256",
       "generateTime": "2024-08-06T20:20:38.376814820Z"
     }
   ],
   "totalSize": 4
 }
```

The output from the tfconfig confirms the latest version by outputting version 4 which can be found in the GET request above:
```hcl
Outputs:

latest_version = 4
latest_version_crypto_key = "projects/hc-terraform-testing/locations/us-central1/keyRings/mau-latest-test/cryptoKeys/mau-crypto-key-test"
latest_version_name = "projects/hc-terraform-testing/locations/us-central1/keyRings/mau-latest-test/cryptoKeys/mau-crypto-key-test/cryptoKeyVersions/4"
```

```release-note:new-datasource
`google_kms_crypto_key_version_latest`
```

### DIFF
--- a/mmv1/third_party/terraform/services/kms/data_source_google_kms_crypto_key_version_latest.go
+++ b/mmv1/third_party/terraform/services/kms/data_source_google_kms_crypto_key_version_latest.go
@@ -67,7 +67,7 @@ func dataSourceGoogleLatestKmsCryptoKeyVersionLatestRead(d *schema.ResourceData,
 		return err
 	}
 
-	url, err := tpgresource.ReplaceVars(d, config, "{{KMSBasePath}}{{crypto_key}}/cryptoKeyVersions")
+	url, err := tpgresource.ReplaceVars(d, config, "{{KMSBasePath}}{{crypto_key}}/cryptoKeyVersions?filter=state=ENABLED")
 	if err != nil {
 		return err
 	}
--- a/mmv1/third_party/terraform/services/kms/data_source_google_kms_crypto_key_version_latest.go
+++ b/mmv1/third_party/terraform/services/kms/data_source_google_kms_crypto_key_version_latest.go
@@ -1,0 +1,218 @@
+package kms
+
+import (
+	"fmt"
+	"log"
+	"strings"
+
+	"github.com/hashicorp/terraform-provider-google/google/tpgresource"
+	transport_tpg "github.com/hashicorp/terraform-provider-google/google/transport"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+)
+
+func DataSourceGoogleLatestKmsCryptoKeyVersionLatest() *schema.Resource {
+	return &schema.Resource{
+		Read: dataSourceGoogleLatestKmsCryptoKeyVersionLatestRead,
+		Schema: map[string]*schema.Schema{
+			"crypto_key": {
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+			},
+			"name": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+			"version": {
+				Type:     schema.TypeInt,
+				Computed: true,
+			},
+			"algorithm": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+			"protection_level": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+			"state": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+			"public_key": {
+				Type:     schema.TypeList,
+				Computed: true,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"algorithm": {
+							Type:     schema.TypeString,
+							Computed: true,
+						},
+						"pem": {
+							Type:     schema.TypeString,
+							Computed: true,
+						},
+					},
+				},
+			},
+		},
+	}
+}
+
+func dataSourceGoogleLatestKmsCryptoKeyVersionLatestRead(d *schema.ResourceData, meta interface{}) error {
+	config := meta.(*transport_tpg.Config)
+	userAgent, err := tpgresource.GenerateUserAgentString(d, config.UserAgent)
+	if err != nil {
+		return err
+	}
+
+	url, err := tpgresource.ReplaceVars(d, config, "{{KMSBasePath}}{{crypto_key}}/cryptoKeyVersions")
+	if err != nil {
+		return err
+	}
+
+	log.Printf("[DEBUG] Getting list of cryptoKeyVersions from cryptoKey: {{crypto_key}}")
+
+	cryptoKeyId, err := ParseKmsCryptoKeyId(d.Get("crypto_key").(string), config)
+	if err != nil {
+		return err
+	}
+	res, err := transport_tpg.SendRequest(transport_tpg.SendRequestOptions{
+		Config:    config,
+		Method:    "GET",
+		Project:   cryptoKeyId.KeyRingId.Project,
+		RawURL:    url,
+		UserAgent: userAgent,
+	})
+	if err != nil {
+		return transport_tpg.HandleDataSourceNotFoundError(err, d, fmt.Sprintf("KmsCryptoKeyVersion %q", d.Id()), url)
+	}
+
+	v, err := flattenKmsCryptoKeyVersionLatest(res["cryptoKeyVersions"].([]interface{}), d, config, cryptoKeyId.CryptoKeyId())
+	if err != nil {
+		return fmt.Errorf("Error getting latest CryptoKeyVersion from crypto key: %s", cryptoKeyId)
+	}
+	latestVersion := v.(map[string]interface{})
+
+	if err := d.Set("version", latestVersion["version"].(int64)); err != nil {
+		return fmt.Errorf("Error setting CryptoKeyVersion: %s", err)
+	}
+	if err := d.Set("name", latestVersion["name"].(string)); err != nil {
+		return fmt.Errorf("Error setting CryptoKeyVersion: %s", err)
+	}
+	if err := d.Set("state", latestVersion["state"]); err != nil {
+		return fmt.Errorf("Error setting CryptoKeyVersion: %s", err)
+	}
+	if err := d.Set("protection_level", latestVersion["protectionLevel"]); err != nil {
+		return fmt.Errorf("Error setting CryptoKeyVersion: %s", err)
+	}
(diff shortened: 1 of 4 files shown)


BBBmau:
looks like in an attempt to add a new `cryptoVersion` for this test (in order to check for the latest version in a cryptoKey in stead of just relying on checking for one version) it unfortunately led to the test failing even after reverting the change in the latest commits. Marking it as ready for review to get assistance on this.

ScottSuarez:
@slevenick, I'm out most of this week. Feel free to pass this back to me on Monday if it's not resolved by then. I want to make sure I don't hold this up with my schedule.

slevenick:
What's the use case for this resource? It looks like most resources that take a version can use "latest" as a special key to get the latest version

BBBmau:
> What's the use case for this resource? It looks like most resources that take a version can use "latest" as a special key to get the latest version

@slevenick This specific data source takes in a number for the version field. Their currently is no way to get the latest cryptoKeyVersion without using a separate curl resource.

```hcl
(base) ┌─(~/Dev/Scratch/latest_version_kms_crypto_version)───────────────────────────────────────────────────────────────────────────────────────────────────────────────(mau@mau-JKDT676NCP:s116)─┐
└─(10:37:48)──> envchain GCLOUD terraform apply                                                                                                                               ──(Wed,Aug07)─┘
╷
│ Error: Incorrect attribute value type
│ 
│   on main.tf line 23, in data "google_kms_crypto_key_version" "latest":
│   23:   version = "latest"
│ 
│ Inappropriate value for attribute "version": a number is required.
```

mmv1/third_party/terraform/services/kms/data_source_google_kms_crypto_key_version_latest.go
```diff
@@ -0,0 +1,218 @@
...
+func flattenKmsCryptoKeyVersionProtectionLevelLatest(v interface{}, d *schema.ResourceData) interface{} {
+	return v
+}
+
+func flattenKmsCryptoKeyVersionAlgorithmLatest(v interface{}, d *schema.ResourceData) interface{} {
+	return v
+}
+
+func flattenKmsCryptoKeyVersionLatest(versionsList []interface{}, d *schema.ResourceData, config *transport_tpg.Config, cryptoKeyId string) (interface{}, error) {
+	latestVersion := versionsList[len(versionsList)-1].(map[string]interface{})
```
slevenick, line 189: Does this always work? If the last key has been disabled or deleted does it show up here, or what would be considered "latest"?

mmv1/third_party/terraform/services/kms/data_source_google_kms_crypto_key_version_latest.go
```diff
@@ -0,0 +1,218 @@
...
+		if err != nil {
+			log.Printf("Error generating public key: %s", err)
+			return err
+		}
+
+		if err := d.Set("public_key", flattenKmsCryptoKeyVersionPublicKey(res, d)); err != nil {
+			return fmt.Errorf("Error setting CryptoKeyVersion public key: %s", err)
+		}
+	}
+	d.SetId(fmt.Sprintf("//cloudkms.googleapis.com/v1/%s/cryptoKeyVersions/%d", d.Get("crypto_key"), d.Get("version")))
```
slevenick, line 157: Why use this format for the ID? A more common format would not be prefixed with //cloudkms...

mmv1/third_party/terraform/services/kms/data_source_google_kms_crypto_key_version_latest.go
slevenick: And can you add a test for such a case?

slevenick (changes requested):
Test is failing

mmv1/third_party/terraform/services/kms/data_source_google_kms_crypto_key_version_latest.go
BBBmau: This was based off of how it's setup in `data_source_google_kms_crypto_key_version.go`:
https://github.com/GoogleCloudPlatform/magic-modules/blob/fdf0c088f9ad24207e61421c5810ebbbb1032b2a/mmv1/third_party/terraform/services/kms/data_source_google_kms_crypto_key_version.go#L148-L154

mmv1/third_party/terraform/services/kms/data_source_google_kms_crypto_key_version_latest.go
BBBmau: When marking versions as `Disabled` they are still present in the list but are not available through the API, the following error appears which is also the result of the failing vcr test (will be destroyed is treated the same as a `disabled` state):

`tf output`:
```hcl
 Error: googleapi: Error 400: projects/hc-terraform-testing/locations/us-central1/keyRings/mau-latest-test/cryptoKeys/mau-crypto-key-test/cryptoKeyVersions/4 is not enabled, current state is: DESTROY_SCHEDULED.
```

The version list in gcloud console:
image

So although it's marked as `disabled` it will still show up as the latest version. I'll look into adding a filter in order to get versions that have the state as `enabled`.

BBBmau: Upon working on this more, I'm starting to believe that having a data source where it's only purpose is to grab the latest version seems like not the best option especially when we consider that their is currently no existing `cryptoKeyVersions.list` data source.

This realization comes from the fact that I would also need to implement `pageTokens` to cover the case of a cryptoKey containing many versions where it requires pages.

It would be best to actually just focus on a plural `cryptoKeyVersions` data source where a boolean field can be added that grabs you the latest version.

example:
```hcl
data "google_kms_crypto_key_versions" "plural" {
    crypto_key = "mau-crypto-latest-test"
    latest_version = true
}
```

@slevenick let me know your thoughts on this.

slevenick: Hmmm, google_kms_crypto_key_versions implies that it returns a collection of versions, and latest_version making that only return 1 seems unusual. Maybe the google_kms_crypto_key_versions could have a `latest_version` field that contains the latest version though.

BBBmau: after a conversation with @SarahFrench through slack, we agreed that the best approach would be to first add support for a plural `crypto_key_versions` data source.

From their support for a singular `crypo_key_version_latest` data source can be then worked on where it uses the same logic from the plural data source for getting the latest version from a paginated list.

I'll be closing this and opening a PR for the plural one first. Thanks for your initial feedback on this! @slevenick

PR #11390, opened 2024-08-07
Added new resource "Router Nat Address"

matheusaleixo-cit:
Adds new resource `google_compute_router_nat_address`, allowing changes to the `nat_ips` field separately from the `google_compute_router_nat` resource.

(WiP) Also adds a new field `initial_nat_ip` for resource `google_compute_router_nat` as well, to be used for the required address during creation.

Fixes: https://github.com/hashicorp/terraform-provider-google/issues/6812

```release-note:new-resource
`google_compute_router_nat_address`
```
```release-note:enhancement
compute: added 'initial_nat_ip' field to 'google_compute_router_nat' resource;
```


mmv1/templates/terraform/examples/router_nat_address_count.tf.erb
```diff
@@ -0,0 +1,52 @@
...
+
+resource "google_compute_address" "address" {
+  count  = 3
+  name   = "<%= ctx[:vars]['address_name'] %>-${count.index}-${random_id.group-manager-suffix.hex}"
+  region = google_compute_subnetwork.subnet.region
+
+  lifecycle {
+    create_before_destroy = true
+  }
+}
```
SarahFrench, line 30: It's more repetitive, but could you please re-implement this test without using count? That seems to be the source of the non-determinism [reported in the tests here](https://github.com/GoogleCloudPlatform/magic-modules/pull/11390#issuecomment-2311915263).

matheusaleixo-cit: Sorry for the delay. I tested configuring each address manually (both with and without the random_id and the lifecycle argument) but that reintroduces the dependency problem shown in the issue. My team is investigating this issue at the moment, and this router_nat_address resource might have to be remade as handwritten.

Besides that, wouldn't the use of random_id itself be the cause of the test failure? Looking at other tests that make use of this (such as [managedSslCertificate](https://github.com/hashicorp/terraform-provider-google/blob/main/google/services/compute/resource_compute_managed_ssl_certificate_generated_test.go#L119) and [dnsManagedZone](https://github.com/hashicorp/terraform-provider-google/blob/main/google/services/dns/resource_dns_managed_zone_generated_test.go#L162)), they seem to be set to skip VCR.

SarahFrench: The [acctest.RandString function](https://github.com/hashicorp/terraform-provider-google/blob/b42543e589586d011caf4a7746a4893856a41582/google/acctest/test_utils.go#L96-L115) provides a random string when the test is run 'normally', but when we run in VCR mode the random string should always be `"abcdefghijklmnopqrstuvwxyz012346789"` to avoid non deterministic behaviour. It needs to be deterministic in the context of VCR as the same random string needs to be used when the test is re-run using recorded API interactions.

These are
```
    vcr_utils.go:167: Step 1/6 error: Error running apply: exit status 1
        
        Error: Error creating Address: Post "https://compute.googleapis.com/compute/beta/projects/PROJECT_ID/regions/us-east1/addresses?alt=json": Requested interaction not found
        
          with google_compute_address.foobar[0],
          on terraform_plugin_test.tf line 24, in resource "google_compute_address" "foobar":
          24: resource "google_compute_address" "foobar" {
        
        
        Error: Error creating Address: Post "https://compute.googleapis.com/compute/beta/projects/PROJECT_ID/regions/us-east1/addresses?alt=json": Requested interaction not found
        
          with google_compute_address.foobar[1],
          on terraform_plugin_test.tf line 24, in resource "google_compute_address" "foobar":
          24: resource "google_compute_address" "foobar" {
```

SarahFrench: Running the test locally with these ENVs set will help debug the issue:

First time running the test:
- TF_LOG=DEBUG
- VCR_MODE=RECORDING
- VCR_PATH=<a path to a place for files to be created>

After the test runs you should be able to see 2 files at that path, the yaml one describing all API interactions.

Second time running the test:
- TF_LOG=DEBUG
- VCR_MODE=REPLAYING
- VCR_PATH=<same path as before>

This should give the insight about what data is available during replay mode, and why the test is failing due to being unable to find matching data in the yaml file

matheusaleixo-cit: Removing the use of random_id did seem to fix the issue when replaying the tests locally, and since apparently only the lifecycle argument is necessary to prevent the dependency issue, I think is there is no point in keeping it.

Nevertheless, I'm having talks with my team about the demand for this resource, since it wont be needed for fixing this issue. I think this PR should be put in hold for now, as we are considering closing it.

SarahFrench: Ah sorry, I didn't spot that a random suffix was being generated in the acceptance test itself and appended. Yes, that would make the test non deterministic. Usually random suffixes are [provided like this](https://github.com/GoogleCloudPlatform/magic-modules/blob/d17814cf2e547c0b1010a46e76ea130822f412ac/mmv1/templates/terraform/examples/organization_security_policy_association_basic.tf.erb#L3), which [are sourced from](https://github.com/hashicorp/terraform-provider-google-beta/blob/d2a5254f2603f0ea89e7cd83e57dc5221676d673/google-beta/services/compute/resource_compute_organization_security_policy_association_generated_test.go#L37) the acctest.RandString function I mentioned. If you need any random strings in future PRs that's the method I'd suggest!

matheusaleixo-cit: > [...]
> Nevertheless, I'm having talks with my team about the demand for this resource, since it wont be needed for fixing this issue. I think this PR should be put in hold for now, as we are considering closing it.

We should be all good on this side, sorry for the delay.

SarahFrench:
I've got a review in progress but I'm asking some questions internally about expectations around fine-grained resources. I'll publish my review once I've got some input!

maxi-cit:
Hello @SarahFrench, as Matheus mentioned the purpose of creating this new resource `google_compute_router_nat_address`, was to manage manually the automatic creation and deletion of Address resources which caused the depencency Error.

however, we found we only need to attach a lifecycle block with create_before_detroy=true attribute.

Here is the PR which only adds this as a test and documentation note [PR](https://github.com/GoogleCloudPlatform/magic-modules/pull/11674)

Feel free to add this new resource in case you consider it you will need it in the future and to close this or the other PR since they have kind of overlapping tests.

SarahFrench:
@maxi-cit , thanks for the heads up - I'll trigger the automation on that PR and do a review.

@matheusaleixo-cit - I'll publish my review on this PR and we can decide next steps. If there's a desire for a fine grained resource from users (separate to fixing the delete a dependant resource error) then this PR can still go ahead

SarahFrench (commented):
Publishing review comments following https://github.com/GoogleCloudPlatform/magic-modules/pull/11390#issuecomment-2341450256

Comments are about:
- Diffsuppression only versus explicitly making the field Optional and Computed
- Updating acceptance tests to assert the fine-grained resource's effects on the parent, and highlighting the delay between the fine-grained resource's effects appearing in the parent resource's Terraform state.

mmv1/products/compute/RouterNatAddress.yaml
```diff
@@ -0,0 +1,136 @@
...
+delete_verb: :PATCH
+identity:
+  - routerNat
+collection_url_key: nats
+nested_query: !ruby/object:Api::Resource::NestedQuery
+  modify_by_patch: true
+  keys:
+    - nats
+description: |
+  A list of IP addresses to be used in a NAT service.
```
SarahFrench, line 32: Additional to my review just now, it would be good to have a callout in this description for the fine-grained resource to warn about how it should be used. For example [this notice block](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/bigquery_dataset_access#:~:text=If%20this%20resource,the%20same%20apply.) in [google_bigquery_dataset_access](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/bigquery_dataset_access) is [created in the yaml for that resource here](https://github.com/GoogleCloudPlatform/magic-modules/blob/c1bcaf1c7e243c7709cc9f289ef892343746fca5/mmv1/products/bigquery/DatasetAccess.yaml#L35-L43), and something similar could be included in RouterNatAddress.yaml.

mmv1/products/compute/RouterNat.yaml
```diff
@@ -164,6 +184,7 @@ properties:
       imports: 'selfLink'
       description: 'A reference to an address associated with this NAT'
     custom_expand: 'templates/terraform/custom_expand/array_resourceref_with_validation.go.erb'
+    diff_suppress_func: 'hasInitialNatIpsDiffSuppress'
```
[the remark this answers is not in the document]
matheusaleixo-cit: Using the `default_from_api: true` parameter did seen to achieve the needed behavior, even with the current use of `send_empty_value: true`.
Only a small issue showed up during testing in that when changing from a `nat_ip_allocate_option = "MANUAL_ONLY"` config with the use of `nat_ips` to a config using `nat_ip_allocate_option = "AUTO_ONLY"`, the user may need to explicitly set `nat_ips = []`, else the provider tries to patch the router nat using the previous nat_ips values.

mmv1/third_party/terraform/services/compute/resource_compute_router_nat_address_test.go.erb
```diff
@@ -0,0 +1,437 @@
...
+
+	testId := acctest.RandString(t, 10)
+	routerName := fmt.Sprintf("tf-test-router-nat-%s", testId)
+
+	acctest.VcrTest(t, resource.TestCase{
+		PreCheck:                 func() { acctest.AccTestPreCheck(t) },
+		ProtoV5ProviderFactories: acctest.ProtoV5ProviderFactories(t),
+		CheckDestroy:             testAccCheckComputeRouterNatDestroyProducer(t),
+		Steps: []resource.TestStep{
+			// (ERROR): Creation with drain nat IPs should fail
```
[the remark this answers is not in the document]
matheusaleixo-cit: I added some data source checks to some of the tests, but apparently the returned data is matching the resource state, only being updated after a second apply. I'm updating this to make the data source dependent on the router nat adddress resource so it gets updated on the same test step.

SarahFrench:
👋 Sorry, I've had a few days off on sick leave. Taking a look at the PR now!

mmv1/third_party/terraform/services/compute/resource_compute_router_nat_address_test.go.erb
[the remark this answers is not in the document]
SarahFrench: That seems to have solved the problem given [these tests pass](https://github.com/GoogleCloudPlatform/magic-modules/pull/11390#issuecomment-2352515495). Thanks!

mmv1/products/compute/RouterNat.yaml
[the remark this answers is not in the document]
SarahFrench: > Only a small issue showed up during testing in that when changing from a nat_ip_allocate_option = "MANUAL_ONLY" config with the use of nat_ips to a config using nat_ip_allocate_option = "AUTO_ONLY", the user may need to explicitly set nat_ips = [], else the provider tries to patch the router nat using the previous nat_ips values.

This is a concern to be honest. I think it'd be worth adding an acceptance test that shows a google_compute_router_nat resource being created with `nat_ip_allocate_option = "MANUAL_ONLY"`  and use of google_compute_router_nat_address, and then being changed to `nat_ip_allocate_option = "AUTO_ONLY"` and setting `nat_ips = []`. If that acceptance test shows that the swap isn't possible due to problems setting `nat_ips = []` then I think we'll need to revert back using diff suppress funcs, instead of default_from_api: true.

If you set up the acceptance test I can run it on this PR to see how/if it fails and then we can figure out if reverting to using diff suppress funcs is the route forward.

SarahFrench (commented):
Left a [comment here](https://github.com/GoogleCloudPlatform/magic-modules/pull/11390#discussion_r1767127906) about the need for a new acceptance test to help us decide between diff suppress funcs and using `default_from_api` when implementing these changes

mmv1/products/compute/RouterNat.yaml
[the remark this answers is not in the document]
matheusaleixo-cit: I made a change to `TestAccComputeRouterNat_update` so that the last two test steps shows this scenario, the `testAccComputeRouterNatUpdateToNatIPsName` config sets the router nat with `nat_ip_allocate_option = "MANUAL_ONLY"` and one value for `nat_ips`, while the `testAccComputeRouterNatBasicBeforeUpdate` config changes it for `nat_ip_allocate_option = "AUTO_ONLY"` and `nat_ips = []`.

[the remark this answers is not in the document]
matheusaleixo-cit: This issue only really pops up when using `nat_ips` from the router_nat directly, as when using a router_nat_address the router_nat is supposed to have only `initial_nat_ips` values configured.
The issue was showing up on the last two test steps of `TestAccComputeRouterNat_update` so I added `nat_ips = []` to the last step. Nevertheless I created the test TestAccComputeRouterNatAddress_withAutoAllocateAndAddressRemoved anyway just to be sure setting `nat_ips = []` does not cause any issues here.

SarahFrench:
Thanks! The new test passing shows that users can still change those fields when their use case changes, and from looking at the test debug logs I can see the same resource is being updated (versus being recreated). Given this I'm happy to approve

SarahFrench (approved):

PR #11392, opened 2024-08-08
Ignore prefix-based diffs on AutokeyConfig.folder parameter

tdbhacks:
Also add custom_expand to drop the 'folders/' prefix from API requests.

```release-note:enhancement
kms: added support for full resource names in folder attribute of `google_kms_autokey_config`
```

### DIFF
--- a/mmv1/products/kms/AutokeyConfig.yaml
+++ b/mmv1/products/kms/AutokeyConfig.yaml
@@ -40,6 +40,7 @@ id_format: 'folders/{{folder}}/autokeyConfig'
 import_format: ['folders/{{folder}}/autokeyConfig']
 min_version: beta
 custom_code: !ruby/object:Provider::Terraform::CustomCode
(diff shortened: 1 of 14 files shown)


mmv1/products/kms/AutokeyConfig.yaml
```diff
@@ -63,6 +64,8 @@ parameters:
     required: true
     immutable: true
     url_param_only: true
+    diff_suppress_func: 'folderPrefixSuppress'
+    custom_expand: 'templates/terraform/custom_expand/autokey_config_folder.go.erb'
```
zli82016, line 68:
```suggestion
    custom_expand: 'templates/terraform/custom_expand/kms_autokey_config_folder.go.erb'
```

tdbhacks:
Test is passing but I still see errors when looking at the debug logs, because the ID isn't being set correctly. I'll keep tweaking this PR and eventually close it in favor of a new, clean one when everything is fixed.

PR #11393, opened 2024-08-08
Remove allow_resource_tags_on_deletion from google_bigquery_table

wj-chen:
Remove `allow_resource_tags_on_deletion` from `google_bigquery_table`. A follow-up to https://github.com/GoogleCloudPlatform/magic-modules/pull/10568.

```release-note:breaking-change
bigquery: removed `allow_resource_tags_on_deletion` from `google_bigquery_table`. Resource tags are now always allowed on table deletion.
```

### DIFF
--- a/mmv1/products/bigquery/Table.yaml
+++ b/mmv1/products/bigquery/Table.yaml
@@ -43,16 +43,6 @@ examples:
     vars:
       dataset_id: 'dataset_id'
       table_id: 'table_id'
-virtual_fields:
-  - !ruby/object:Api::Type::Boolean
-    name: 'allow_resource_tags_on_deletion'
-    description: |
-      If set to true, it allows table deletion when there are still resource tags attached.
-    deprecation_message: |
-      `allow_resource_tags_on_deletion` is deprecated and will be removed in a future major
-      release. The default behavior will be allowing the presence of resource tags on
-      deletion after the next major release.
-    default_value: false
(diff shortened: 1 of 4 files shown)


shuyama1 (commented):
@wj-chen Thanks for making the change. LGTM overall! Just to confirm, after removing this field, we'll always allow resource_tags on deletion, and this won't cause any backend errors? Apologies, just wanted to refresh my memory on this. Thanks!

shuyama1:
/gcbrun 

A rerun should get the breaking change test passed with the override label.

wj-chen:
> @wj-chen Thanks for making the change. LGTM overall! Just to confirm, after removing this field, we'll always allow resource_tags on deletion, and this won't cause any backend errors? Apologies, just wanted to refresh my memory on this. Thanks!

Yes, it will be the same behavior as `allow_resource_tags_on_deletion = true` today with no backend error. Thanks for checking.

shuyama1 (approved):
> > @wj-chen Thanks for making the change. LGTM overall! Just to confirm, after removing this field, we'll always allow resource_tags on deletion, and this won't cause any backend errors? Apologies, just wanted to refresh my memory on this. Thanks!
> 
> Yes, it will be the same behavior as `allow_resource_tags_on_deletion = true` today with no backend error. Thanks for checking.

Thanks for confirming!

PR #11394, opened 2024-08-08
Fix Datastream private connection resource destruction

sagiyosefia:
Fix Datastream private connection resource destruction. The current private connection resource deletion fails as a result of sub resource existence.
This fix will always force the resource deletion, as the sub resources deletion is taking care the the API anyway.

Fixes https://github.com/hashicorp/terraform-provider-google/issues/17290 
https://github.com/hashicorp/terraform-provider-google/issues/13054

```release-note:bug
compute:  fixed issue where sub-resources managed by `google_compute_forwarding_rule` prevented resource deletion
```

### DIFF
--- a/mmv1/products/datastream/PrivateConnection.yaml
+++ b/mmv1/products/datastream/PrivateConnection.yaml
@@ -43,6 +43,7 @@ custom_code: !ruby/object:Provider::Terraform::CustomCode
   constants: templates/terraform/constants/private_connection.go.erb
   post_create: templates/terraform/post_create/private_connection.go.erb
   post_import: templates/terraform/post_import/private_connection.go.erb
+  pre_delete: templates/terraform/pre_delete/private_connection.go.erb
 parameters:
   - !ruby/object:Api::Type::String
     name: privateConnectionId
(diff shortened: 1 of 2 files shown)


mmv1/products/datastream/PrivateConnection.yaml
```diff
@@ -43,6 +43,7 @@ custom_code: !ruby/object:Provider::Terraform::CustomCode
   constants: templates/terraform/constants/private_connection.go.erb
   post_create: templates/terraform/post_create/private_connection.go.erb
   post_import: templates/terraform/post_import/private_connection.go.erb
+  pre_delete: templates/terraform/pre_delete/private_connection.go.erb
```
ScottSuarez, line 46: you could achieve the same through a  delete url. It might be more readable this way !!

https://github.com/GoogleCloudPlatform/magic-modules/blob/5043710da3cb65c093a0dd527c7d857b9bc6d697/mmv1/products/bigquery/go_Dataset.yaml#L32

:)

ScottSuarez (changes requested):
Are your routes not managed through terraform?

I can see why this behavior might be needed but if everything is within terraform this should be a non-issue. 

We could potentially make this opt in but let me understand your usecase before I suggest anything.

joshbautista:
> Are your routes not managed through terraform?

@ScottSuarez More details in https://github.com/hashicorp/terraform-provider-google/issues/17290 but the gist is:

The following supporting resources are not managed or created explicitly by Terraform when instantiating a `google_datastream_private_connection`:

- VPC Peering Connection with a Google owned VPC
- VPC Peering Routes
- Allocated IP Range for Private Services Access

From what I can tell, the Datastream API takes it upon itself to manage the lifecycle of these resources on your behalf and doesn't expect (or possibly want) the end user to have created these resources beforehand.

ScottSuarez (approved):

PR #11401, opened 2024-08-08
Fixes https://github.com/hashicorp/terraform-cdk/issues/3695 https://github.com/hashicorp/terraform-provider-google/issues/15456

eahrend:
When creating terraform compliant JSON, the IP Allocation policy block in the composer resource will incorrectly state that all attributes are required. 

This PR removes the ConfigMode of the ip_allocation_policy, and removes the atleastoneof fields from the elements in the ip_allocation_policy field. This brings it more in line with the container engine resource, and has the benefit of allowing terraform compliant JSON to work the same as HCL

```release-note:bug
composer: removed ConfigMode: SchemaConfigModeAttr ip_allocation_policy and AtLeastOneOf fields from ip_allocation_policy sub fields
```


eahrend:
Ran tests locally that confirm the existing terraform configurations still work, so this shouldn't break any existing terraform code. Also built a local version of the provider and confirmed the terraform compliant JSON that was failing before is working as intended.

eahrend:
Added test case with named alias ranges for ip_allocation_policy

eahrend:
@shuyama1 can I get a review on this please?

eahrend:
@shuyama1 apologies if this is coming off rude or pestering, just a small reminder if I can get a review on this

shuyama1:
> @shuyama1 apologies if this is coming off rude or pestering, just a small reminder if I can get a review on this

Apologies for the delay on the reviews, and thank you for the reminder. I'm working through my review list today and will get to this PR shortly.

mmv1/third_party/terraform/services/composer/resource_composer_environment.go.erb
```diff
@@ -338,39 +338,34 @@ func ResourceComposerEnvironment() *schema.Resource {
 										Optional:    true,
 										Computed:    true,
 										ForceNew:    true,
-										ConfigMode:  schema.SchemaConfigModeAttr,
```
shuyama1, line 341: I'll need to double check if removing it will be a breaking check.

mmv1/third_party/terraform/services/composer/resource_composer_environment.go.erb
```diff
@@ -338,39 +338,34 @@ func ResourceComposerEnvironment() *schema.Resource {
...
 										MaxItems:    1,
 										Description: `Configuration for controlling how IPs are allocated in the GKE cluster. Cannot be updated.`,
 										Elem: &schema.Resource{
 											Schema: map[string]*schema.Schema{
 												"use_ip_aliases": {
 													Type:         schema.TypeBool,
-													Optional:     true,
+													Optional:     true,													
 													ForceNew:     true,
-													AtLeastOneOf: composerIpAllocationPolicyKeys,
```
shuyama1, line 350: I believe this is originally added to avoid users from sending empty blocks, so it wouldn't be breaking to remove. However, I'm still a bit confused why this will cause the error you see in https://github.com/hashicorp/terraform-provider-google/issues/15456#issuecomment-2274414644, why it makes all subfields as required.

Is there any potential way that you can check locally if this will fix the issue you're facing in CDKTF  before merging the change to the providers?

shuyama1 (commented):
Sorry for the late review. Left some comments and I'll also need to look into the issue a bit.

mmv1/third_party/terraform/services/composer/resource_composer_environment.go.erb
eahrend: Yeah, I confirmed this fixes the issue with CDKTF/terraform compliant JSON. I'll be very honest, I'm not entirely certain why this fixes it, because it really seems like this _should_ work as intended, but yeah, this fixes it. 

I don't have any definitive proof, but I have a feeling this is probably due to the lack of a 1:1 translation between terraform compliant JSON and HCL, and something that may be a lower level terraform bug fix. 
https://developer.hashicorp.com/terraform/language/attr-as-blocks#in-json-syntax

Also, while I don't have any definitive proof as to why the GKE ip_allocation_policy resource is structured in a similar way to this PR, I wouldn't be surprised if it was due to this as well. 

https://github.com/hashicorp/terraform-provider-google/blob/main/google/services/container/resource_container_cluster.go#L1443-L1536

mmv1/third_party/terraform/services/composer/resource_composer_environment.go.erb
eahrend: No worries, I ran some of the existing tests as well when doing a bit of local development and at least confirmed from my end that the tests were still valid before creating this PR

eahrend:
@shuyama1 no worries on the late review, saw the github outage on [wednesday](https://www.githubstatus.com/incidents/kz4khcgdsfdv) and figured that would impact the ETA for this to get reviewed.

eahrend:
@shuyama1 just checking in on that

mmv1/third_party/terraform/services/composer/resource_composer_environment.go.erb
shuyama1: Removing SchemaConfigModeAttr is indeed a breaking change and can't be included in a minor release. We're currently at phase of working on 6.0.0 major release, and SchemaConfigModeAttr will be removed from all the resources. This work has already been done and 6.0.0 should be out in the next few weeks.

shuyama1 (commented):
> @shuyama1 just checking in on this

Thanks for the ping. I just replied in https://github.com/GoogleCloudPlatform/magic-modules/pull/11401#discussion_r1724127428. Let's wait until the 6.0.0 is out to continue with the following work/ evaluate if further changes are needed, as I wonder if removing SchemaConfigModeAttr itself will resolve your issues.

eahrend:
@shuyama1 any chance we can keep this open or prevent the stale bot from closing this until 6.0.0 drops or should I just keep a comment open until it does?

shuyama1:
> @shuyama1 any chance we can keep this open or prevent the stale bot from closing this until 6.0.0 drops or should I just keep a comment open until it does?

Sorry for the noise!
Sure, I added the label to disable the reminders from the bot and I'll also convert this PR to draft, but feel free to convert it back you'd like to continue working on it. Thank you!

mmv1/third_party/terraform/services/composer/resource_composer_environment.go.erb
eahrend: It looks like 6.1.0 fixes this, it attempts to recreate the cluster but it looks like there was something also done on GCP's backend that removed the ability to use `-`

PR #11408, opened 2024-08-09
Ignore `trace_id` in the request

lola98:
When updating any existing quota_preference object, ignore the associated `trace_id` .

```release-note:REPLACEME

```


trodge:
Is there an issue you can link to that has more context on why this change is needed?

lola98:
> Is there an issue you can link to that has more context on why this change is needed?

Yes https://buganizer.corp.google.com/issues/356922112#comment5 has more context on this. Basically we want to ignore the `trace_id` field when creating and updating preferences.

trodge (approved):

PR #11410, opened 2024-08-09
add send_secondary_ip_range_if_empty

c2thorn:
part of https://github.com/hashicorp/terraform-provider-google/issues/12824 and https://github.com/hashicorp/terraform-provider-google/issues/17881

adds a virtual field to replace the need to set `secondary_ip_range = []` to remove `secondary_ip_range` explicitly. The virtual field in combination with `GetRawConfig().GetAttr("secondary_ip_range")` can let us detect when the user has intent to set to an empty list vs intending to let the backend manage the field.

Normally, when removing an Optional+Computed field from config, Terraform will not produce a diff. The `sendSecondaryIpRangeIfEmptyDiff` checks if the new flag is set and then compares the config value with the state value using `GetRawConfig()` and `GetOk` respectively.

The actual update call is handled in a post_update custom code block.

This logic works **with and without SchemaConfigModeAttr**, allowing an upgrade path in 5.x before removing SchemaConfigModeAttr from `secondary_ip_range` in 6.0

```release-note:deprecation
compute: setting `google_compute_subnetwork.secondary_ip_range = []` to explicitly set a list of empty objects is deprecated and will produce an error in the upcoming major release. Use `send_secondary_ip_range_if_empty` while removing `secondary_ip_range` from config instead.
```

```release-note:enhancement
compute: added `send_secondary_ip_range_if_empty` to `google_compute_subnetwork`
```

### DIFF
--- a/mmv1/products/compute/Subnetwork.yaml
+++ b/mmv1/products/compute/Subnetwork.yaml
@@ -130,7 +130,7 @@ virtual_fields:
     name: 'send_secondary_ip_range_if_empty'
     description: |
       Controls the removal behavior of secondary_ip_range.
-      When false, removing secondary_ip_range from config will not produce a diff as 
+      When false, removing secondary_ip_range from config will not produce a diff as
       the provider will default to the API's value.
       When true, the provider will treat removing secondary_ip_range as sending an
       empty list of secondary IP ranges to the API.
--- a/mmv1/products/compute/Subnetwork.yaml
+++ b/mmv1/products/compute/Subnetwork.yaml
@@ -71,8 +71,10 @@ iam_policy: !ruby/object:Api::Resource::IamPolicy
 custom_code: !ruby/object:Provider::Terraform::CustomCode
   constants: templates/terraform/constants/subnetwork.erb
   extra_schema_entry: templates/terraform/extra_schema_entry/subnetwork.erb
+  post_update: templates/terraform/post_update/compute_subnetwork.go.erb
 custom_diff: [
   'customdiff.ForceNewIfChange("ip_cidr_range", IsShrinkageIpCidr)',
+  'sendSecondaryIpRangeIfEmptyDiff',
 ]
 examples:
   - !ruby/object:Provider::Terraform::Examples
@@ -123,6 +125,16 @@ examples:
     vars:
       subnetwork_name: 'subnet-cidr-overlap'
       network_name: 'net-cidr-overlap'
+virtual_fields:
+  - !ruby/object:Api::Type::Boolean
+    name: 'send_secondary_ip_range_if_empty'
+    description: |
+      Controls the removal behavior of secondary_ip_range.
+      When false, removing secondary_ip_range from config will not produce a diff as 
+      the provider will default to the API's value.
+      When true, the provider will treat removing secondary_ip_range as sending an
+      empty list of secondary IP ranges to the API.
+      Defaults to false.
 properties:
   - !ruby/object:Api::Type::Time
     name: 'creationTimestamp'
@@ -215,11 +227,7 @@ properties:
       to the primary ipCidrRange of the subnetwork. The alias IPs may belong
       to either primary or secondary ranges.
 
-      **Note**: This field uses [attr-as-block mode](https://www.terraform.io/docs/configuration/attr-as-blocks.html) to avoid
-      breaking users during the 0.12 upgrade. To explicitly send a list
-      of zero objects you must use the following syntax:
-      `example=[]`
-      For more details about this behavior, see [this section](https://www.terraform.io/docs/configuration/attr-as-blocks.html#defining-a-fixed-object-collection-value).
+      **Note**: To explicitly send a list of zero objects, set `send_secondary_ip_range_if_empty = true`
     update_verb: :PATCH
     update_url: projects/{{project}}/regions/{{region}}/subnetworks/{{name}}
     update_id: 'secondaryIpRanges'
--- a/mmv1/products/compute/Subnetwork.yaml
+++ b/mmv1/products/compute/Subnetwork.yaml
@@ -227,7 +227,9 @@ properties:
       to the primary ipCidrRange of the subnetwork. The alias IPs may belong
       to either primary or secondary ranges.
 
-      **Note**: To explicitly send a list of zero objects, set `send_secondary_ip_range_if_empty = true`
+      **Note**: This field uses [attr-as-block mode](https://www.terraform.io/docs/configuration/attr-as-blocks.html) to avoid
+      breaking users during the 0.12 upgrade. To explicitly send a list of zero objects,
+      set `send_secondary_ip_range_if_empty = true`
     update_verb: :PATCH
     update_url: projects/{{project}}/regions/{{region}}/subnetworks/{{name}}
     update_id: 'secondaryIpRanges'
--- a/mmv1/templates/terraform/constants/subnetwork.erb
+++ b/mmv1/templates/terraform/constants/subnetwork.erb
@@ -19,24 +19,26 @@ func IsShrinkageIpCidr(_ context.Context, old, new, _ interface{}) bool {
 }
 
 func sendSecondaryIpRangeIfEmptyDiff(_ context.Context, diff *schema.ResourceDiff, meta interface{}) error {
+	if diff.Id() == "" {
+		return nil
+	}
+
 	sendZero := diff.Get("send_secondary_ip_range_if_empty").(bool)
 	if !sendZero {
 		return nil
 	}
 
-	configValueIsEmpty := false
 	configSecondaryIpRange := diff.GetRawConfig().GetAttr("secondary_ip_range")
-
-	// IsNull check can be deleted when SchemaConfigModeAttr is removed
-	if configSecondaryIpRange.IsNull() {
-		configValueIsEmpty = true
-	} else {
-		configSlice := configSecondaryIpRange.AsValueSlice()
-		configValueIsEmpty = len(configSlice) == 0
+	if !configSecondaryIpRange.IsKnown() {
+		return nil
 	}
+	configValueIsEmpty := configSecondaryIpRange.IsNull() || configSecondaryIpRange.LengthInt() == 0
 
-	stateSecondaryIpRange, ok := diff.GetOk("secondary_ip_range")
-	stateValueIsEmpty := !ok || len(stateSecondaryIpRange.([]interface{})) == 0
+	stateSecondaryIpRange := diff.GetRawState().GetAttr("secondary_ip_range")
+	if !stateSecondaryIpRange.IsKnown() {
+		return nil
+	}
+	stateValueIsEmpty := stateSecondaryIpRange.IsNull() || stateSecondaryIpRange.LengthInt() == 0
 
 	if configValueIsEmpty && !stateValueIsEmpty {
 		log.Printf("[DEBUG] setting secondary_ip_range to newly empty")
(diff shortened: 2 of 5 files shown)


mmv1/templates/terraform/constants/subnetwork.erb
```diff
@@ -17,3 +17,31 @@ func IsShrinkageIpCidr(_ context.Context, old, new, _ interface{}) bool {
...
 }
+
+func sendSecondaryIpRangeIfEmptyDiff(_ context.Context, diff *schema.ResourceDiff, meta interface{}) error {
+	sendZero := diff.Get("send_secondary_ip_range_if_empty").(bool)
+	if !sendZero {
+		return nil
+	}
+
+	configValueIsEmpty := false
+	configSecondaryIpRange := diff.GetRawConfig().GetAttr("secondary_ip_range")
```
rileykarson, line 28: Have you tested a case where the value of the field is unknown? Specifically, is the value considered `Null` in that case? One easy way to simulate unknowns is `local_file + data.local_file` I think.

Just a quick refresher, CustomizeDiff runs twice; once during plan, and once during apply immediately before calling [the internal SDK function] Apply on the resource. Different information can be present each time because interpolations are not resolved at plan time but are resolved at apply time. Terraform gets annoyed if known values change, typically returning an error, although to what degree in a messy case like this I'm not sure.

Because the field is O+C there are two ways it can be unknown:

- No entries are defined
- Entries are defined, but at least one of their values are unknown and we're in the plan phase

So it's possible we may need to report the value as _unknown_ during plan, and then set the known value of explicit zero during apply. I'm not entirely sure we can tell the two phases apart, either. There are GetRawPlan and NewValueKnown that may help provide more information.

rileykarson: (we may also just decide to compromise and make it so that `send_secondary_ip_range_if_empty` overrides _all_ unknowns with a big 'ol warning)

mmv1/templates/terraform/post_update/compute_subnetwork.go.erb
```diff
@@ -0,0 +1,72 @@
+if v, ok := d.GetOk("send_secondary_ip_range_if_empty"); ok && v.(bool) {
+	if sv, ok := d.GetOk("secondary_ip_range"); ok {
+		configValue := d.GetRawConfig().GetAttr("secondary_ip_range").AsValueSlice()
```
rileykarson, line 3: What case does this handle, if we've presumably set a known zero value for the field in the CustomizeDiff? Theoretically wouldn't we always get the correct value from `d.Get`?

mmv1/templates/terraform/constants/subnetwork.erb
rileykarson: @SarahFrench icyi, we are pretty into the weeds on this stuff.

SarahFrench: I'm by no means an expert but there are methods for explicitly handling unknown values: [IsKnown](https://pkg.go.dev/github.com/hashicorp/go-cty@v1.4.1-0.20200414143053-d3edf31b6320/cty#Value.IsKnown), [IsWhollyKnown](https://pkg.go.dev/github.com/hashicorp/go-cty@v1.4.1-0.20200414143053-d3edf31b6320/cty#Value.IsWhollyKnown)

---

Sidebar that might be helpful: when I've looked into ways to allow unsetting of O+C fields I've looked at [this example](https://github.com/hashicorp/terraform-provider-aws/blob/b257779038f383ff3670f33a8637786ed1ce1673/internal/service/appstream/stack.go#L208-L240) from the AWS provider. This code detects if application_settings is in the state but isn't in the config, i.e. detects a scenario when the user has deleted the [application_settings block](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/appstream_stack#application_settings) completely from their config, and then sets a new 'zero' value for fields in that block. This code only works because application_settings is a top level field and is marked as Computed (requirements to use [schema#ResourceDiff.SetNew](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-sdk/v2@v2.33.0/helper/schema#ResourceDiff.SetNew)). I couldn't find a solution for fields that don't meet those requirements. Because of this, I think your approach of adding a new field is correct for solving this type of problem specifically with the SDK.

c2thorn: Thanks @SarahFrench, I'll take a look at that code - especially for handing unknown values

If I'm understanding it correctly, `application_settings` loses the ability to go from a Terraform configured value -> default to the API. It is assumed unsetting the field in an update means setting the empty value. The field can default to the API if initially created without being set in config.

@rileykarson what do you think of that model? Unsetting the field in an update is always assumed to be setting the empty value - which would eliminate the need for a virtual flag

rileykarson: I think that'd be reasonable. Can the field still be set to empty on creation?

rileykarson: If we ran `terraform apply` a second time after creation would we see a diff from the API default to empty?

Subnetwork is a multiactor case so it's less a priority but GKE cluster (at least, one of the fields) is a list of IPs allowed to access the cluster. For that case create-with-zero is a flow we pretty strongly want to support.

mmv1/templates/terraform/post_update/compute_subnetwork.go.erb
c2thorn: This is necessary when we remove ConfigModeAttr. After removing it, d.GetChange/d.HasChange loses the diff from the plan, so this ended up necessary to actually trigger the update. It works the same for with and without ConfigModeAttr

mmv1/templates/terraform/constants/subnetwork.erb
c2thorn: I've confirmed that sending an empty `secondary_ip_range` list during create does not make the API actually return an empty list after creation. (gpaste/6269886252974080)

 Given this and the multiactor case, we should not need to send an empty `secondary_ip_range` after create.
 
 I've modified this customDiff to exit early if `IsKnown()` is false, similar to the previous example @SarahFrench linked.

mmv1/products/compute/Subnetwork.yaml
```diff
@@ -215,11 +227,7 @@ properties:
       to the primary ipCidrRange of the subnetwork. The alias IPs may belong
       to either primary or secondary ranges.
 
-      **Note**: This field uses [attr-as-block mode](https://www.terraform.io/docs/configuration/attr-as-blocks.html) to avoid
```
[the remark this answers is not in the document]
c2thorn: added the warnings back

mmv1/templates/terraform/post_update/compute_subnetwork.go.erb
rileykarson: > d.GetChange/d.HasChange loses the diff from the plan

Just to make sure I have this right- `terraform plan` displays a diff, but those functions- the "safe" functions in the SDK- report that it did not?

mmv1/templates/terraform/constants/subnetwork.erb
rileykarson: > I've confirmed that sending an empty secondary_ip_range list during create does not make the API actually return an empty list after creation.

I think I'm missing somethere here it returns null i.e. empty for lists in that paste.

> If we ran terraform apply a second time after creation would we see a diff from the API default to empty?

Sorry, this was me grounding too much on GKE. Not as relevant here as a new subnetwork should default to empty.

mmv1/templates/terraform/constants/subnetwork.erb
```diff
@@ -17,3 +17,33 @@ func IsShrinkageIpCidr(_ context.Context, old, new, _ interface{}) bool {
 
 	return true
 }
+
+func sendSecondaryIpRangeIfEmptyDiff(_ context.Context, diff *schema.ResourceDiff, meta interface{}) error {
+	if diff.Id() == "" {
```
rileykarson, line 22:
```suggestion
	// on create, return immediately as we don't need to determine if the value is empty or not
	if diff.Id() == "" {
```

mmv1/templates/terraform/constants/subnetwork.erb
rileykarson: Also flagging that the unknown nonzero value case has not been covered or I have not realised it has been covered anywhere yet! (apologies for two topics in one thread)

c2thorn: > I think I'm missing somethere here it returns null i.e. empty for lists in that paste.

We send `"secondaryIpRanges": []` but don't get anything back during create. Not sending `secondaryIpRanges` will end up with the same effect on create. 

If something were to change `secondaryIpRanges` out of band and the user wanted to set it back to empty, the new flag will trigger an update on next apply.

mmv1/templates/terraform/post_update/compute_subnetwork.go.erb
c2thorn: yes this is a local run [without ConfigModeAttr](https://github.com/c2thorn/terraform-provider-google-beta/commit/fc830b85261a4e8f4bda19e487edd093b19e39fd#diff-0fb3f296c74e78df5ca33f93d3f3ada9f40d9fad606ebf4207149b1874bc441bR287) and some added debug logs: gpaste/5449355903631360

`setting secondary_ip_range to newly empty` line from the customDiff appears twice as expected

paste/5449355903631360#l=399 are the print logs I added. Here's [the code](https://github.com/c2thorn/terraform-provider-google-beta/commit/fc830b85261a4e8f4bda19e487edd093b19e39fd#diff-0fb3f296c74e78df5ca33f93d3f3ada9f40d9fad606ebf4207149b1874bc441bR1086)

For good measure, I even added another [debug log](https://github.com/c2thorn/terraform-provider-google-beta/commit/fc830b85261a4e8f4bda19e487edd093b19e39fd#diff-0fb3f296c74e78df5ca33f93d3f3ada9f40d9fad606ebf4207149b1874bc441bR1013) for if the previous MMv1-generated update code is even triggering. It does not appear

mmv1/templates/terraform/constants/subnetwork.erb
```diff
@@ -17,3 +17,33 @@ func IsShrinkageIpCidr(_ context.Context, old, new, _ interface{}) bool {
...
+		return nil
+	}
+
+	sendZero := diff.Get("send_secondary_ip_range_if_empty").(bool)
+	if !sendZero {
+		return nil
+	}
+
+	configSecondaryIpRange := diff.GetRawConfig().GetAttr("secondary_ip_range")
+	if !configSecondaryIpRange.IsKnown() {
```
c2thorn, line 32: > Also flagging that the unknown nonzero value case has not been covered or I have not realised it has been covered anywhere yet! (apologies for two topics in one thread)

@rileykarson Doesn't this covered all unknown cases? Exit early if the value is not known

mmv1/templates/terraform/constants/subnetwork.erb
c2thorn: marking as resolved an continuing the 2nd topic in another thread

mmv1/templates/terraform/post_update/compute_subnetwork.go.erb
rileykarson: Thanks!

mmv1/templates/terraform/constants/subnetwork.erb
rileykarson: Oh, yep! Could confirm the intended behaviour with a test (one-off manual would be fine) that we see the intended behavious with an unknown value for a secondary range?

c2thorn: yes here it is (with some added debug logs for `IsKnown` / `isWhollyKnown`: paste/5787900929376256#l=324

rileykarson (approved):

PR #11421, opened 2024-08-11
validate filter length doesn't exceed 256 bytes in subscription

omrikiei:
This PR validates that the filter length doesn't exceed 256 bytes in a PubSub subscription, on TF it should fail when planning

```release-note:enhancement
pusbub: added validation to `filter` field for `google_pubsub_subscription`
```

### DIFF
--- a/mmv1/products/pubsub/Subscription.yaml
+++ b/mmv1/products/pubsub/Subscription.yaml
@@ -387,6 +387,8 @@ properties:
   - !ruby/object:Api::Type::String
     name: 'filter'
     required: false
+    validation: !ruby/object:Provider::Terraform::Validation
+      regex: '^.{1,256}$'
     description: |
       The subscription only delivers the messages that match the filter.
       Pub/Sub automatically acknowledges the messages that don't match the filter. You can filter messages
--- a/mmv1/products/pubsub/go_Subscription.yaml
+++ b/mmv1/products/pubsub/go_Subscription.yaml
@@ -387,7 +387,7 @@ properties:
         diff_suppress_func: 'comparePubsubSubscriptionExpirationPolicy'
   - name: 'filter'
     type: String
-    validation:
+    validation: !ruby/object:Provider::Terraform::Validation
       function: 'verify.ValidateRegexp(`^.{1,256}$`)'
     description: |
       The subscription only delivers the messages that match the filter.
--- a/mmv1/products/pubsub/go_Subscription.yaml
+++ b/mmv1/products/pubsub/go_Subscription.yaml
@@ -387,6 +387,8 @@ properties:
         diff_suppress_func: 'comparePubsubSubscriptionExpirationPolicy'
   - name: 'filter'
     type: String
+    validation:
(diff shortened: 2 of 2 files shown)


hao-nan-li (commented):
Looks like this change didn't create properly create the validation function. Could you refer to https://github.com/GoogleCloudPlatform/magic-modules/blob/8bed26c9a024ada0fcdb1ddea3bb01df32028c1c/mmv1/products/securitycenter/FolderCustomModule.yaml#L82 and see if that works?

omrikiei:
> Looks like this change didn't create properly create the validation function. Could you refer to
> 
> https://github.com/GoogleCloudPlatform/magic-modules/blob/8bed26c9a024ada0fcdb1ddea3bb01df32028c1c/mmv1/products/securitycenter/FolderCustomModule.yaml#L82
> 
> and see if that works?

I've changed it to what I think would instrument a change, not really sure how to test this locally..

hao-nan-li:
> > Looks like this change didn't create properly create the validation function. Could you refer to
> > https://github.com/GoogleCloudPlatform/magic-modules/blob/8bed26c9a024ada0fcdb1ddea3bb01df32028c1c/mmv1/products/securitycenter/FolderCustomModule.yaml#L82
> > 
> > and see if that works?
> 
> I've changed it to what I think would instrument a change, not really sure how to test this locally..

This change still did not cause any diff. I think you should make the same change in subscription.yaml as well.

omrikiei:
> > > Looks like this change didn't create properly create the validation function. Could you refer to
> > > https://github.com/GoogleCloudPlatform/magic-modules/blob/8bed26c9a024ada0fcdb1ddea3bb01df32028c1c/mmv1/products/securitycenter/FolderCustomModule.yaml#L82
> > > 
> > > and see if that works?
> > 
> > 
> > I've changed it to what I think would instrument a change, not really sure how to test this locally..
> 
> This change still did not cause any diff. I think you should make the same change in subscription.yaml as well.

 I see, I changed the file, would it be generated now?

hao-nan-li (approved):

martinOrigin8Cares:
Why using a regex to validate the content, If the check is only length?

ndopj:
Hello, this change introduced a breaking behaviour which was not released with new major version.
https://github.com/hashicorp/terraform-provider-google/pull/19131#issuecomment-2324053553
Bellow is copy paste with additional info from link above:  

This is breaking change which was not released under new Major version. The API allows filters to contain new line characters. This is useful when Terraform users use [Heredoc strings](https://developer.hashicorp.com/terraform/language/expressions/strings#heredoc-strings) for attributes expecting string type.

Example for PubSub filter:
```hcl
resource "google_pubsub_subscription" "events" {
  name   = "events"
  topic  = google_pubsub_topic.events.name
  filter = <<-EOT
  (attributes.platform = "web" AND attributes.category = "shopping_cart")
  OR (attributes.platform = "app")
  EOT
}
```
The above filter used to work before the change introduced in this MR. Introduced regex, ^.{1,256}$, used for validation of the filter attribute does not allow new lines to be present in the provided string value.

PR #11429, opened 2024-08-12
Add handwritten sweeper for google_data_loss_prevention_discovery_config

patrickmoy:
Addresses https://github.com/hashicorp/terraform-provider-google/issues/19089.

This PR:

- Adds a new sweeper to remove DiscoveryConfigs in the location passed by the sweeper (us-central1)
- Creates a sweeper that **does not** check for unsweepable resources, as there should be only one DiscoveryConfig of each storage type per location per project, so we must delete all DiscoveryConfigs in this location and project to avoid impacting CI tests

```release-note:none

```

### DIFF
--- a/mmv1/third_party/terraform/services/datalossprevention/resource_data_loss_prevention_discovery_config_sweeper.go
+++ b/mmv1/third_party/terraform/services/datalossprevention/resource_data_loss_prevention_discovery_config_sweeper.go
@@ -1,0 +1,106 @@
+package datalossprevention
+
+import (
+	"context"
+	"log"
+	"strings"
+
+	"github.com/hashicorp/terraform-provider-google/google/sweeper"
+	"github.com/hashicorp/terraform-provider-google/google/tpgresource"
+	transport_tpg "github.com/hashicorp/terraform-provider-google/google/transport"
+)
+
+func init() {
+	sweeper.AddTestSweepers("DataLossPreventionDiscoveryConfig", testSweepDataLossPreventionDiscoveryConfig)
+}
+
+// At the time of writing, the CI only passes us-central1 as the region
+func testSweepDataLossPreventionDiscoveryConfig(region string) error {
+	resourceName := "DataLossPreventionDiscoveryConfig"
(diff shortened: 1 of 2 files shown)


mmv1/third_party/terraform/services/datalossprevention/resource_data_loss_prevention_discovery_config_sweeper.go
```diff
@@ -0,0 +1,110 @@
...
+	rl := resourceList.([]interface{})
+
+	log.Printf("[INFO][SWEEPER_LOG] Found %d items in %s list response.", len(rl), resourceName)
+	for _, ri := range rl {
+		obj := ri.(map[string]interface{})
+		if obj["name"] == nil {
+			log.Printf("[INFO][SWEEPER_LOG] %s resource name was nil", resourceName)
+			return nil
+		}
+
```
c2thorn, line 84: can we add a comment that we don't care about checking for a sweepable-prefix?

c2thorn (changes requested):
Were you able to test this locally? Would you happen to have a log of it deleting something

mmv1/third_party/terraform/services/datalossprevention/resource_data_loss_prevention_discovery_config_sweeper.go
patrickmoy: Done, lines 79-80. Also squashed into 1 commit.

patrickmoy:
I just tested successfully with logs — here's the breakdown.

1. Created a fresh [DiscoveryConfig](https://screenshot.googleplex.com/9QttN5asA2Wf2ew) in my test project, note the resource name 6564684080044951644 on the last line.
2. New discovery config is [visible](https://screenshot.googleplex.com/7WdbAJo9PiC9KGa) in pantheon UI in list of configs.
3. Ran sweeper locally, with attached [gpaste](https://paste.googleplex.com/6534070295330816#l=28) log and matching resource name. 
4. Discovery config is no longer seen in [pantheon](https://screenshot.googleplex.com/8qcmfWqmp3vEPiR) since it's been deleted

PR #11331, opened 2024-08-01
feat: support Cloud SQL's new field `server_ca_mode`.

feng-zhe:
This PR supports the new field `settings.ip_configuration.server_ca_mode` of Cloud SQL instances.

**Release Note Template for Downstream PRs (will be copied)**

```release-note:enhancement
sql: added `server_ca_mode` field to `google_sql_database_instance` resource
```

PR #11432, opened 2024-08-12
Revert "feat: support Cloud SQL's new field `server_ca_mode`. (#11331)"

shuyama1:
This reverts commit 4933234.

```release-note:none

```


melinath (commented):
Based on discussion - we think this probably doesn't need to be reverted after all. It is likely related to a quirk of how the test is written (interacting with recent changes to the bigtable client library) and probably won't break users.

shuyama1:
> Based on discussion - we think this probably doesn't need to be reverted after all. It is likely related to a quirk of how the test is written (interacting with recent changes to the bigtable client library) and probably won't break users.

yes, the change should be relatively safe to be released to users, given the test that failed in our test environment passed locally. I'm taking a closer look now and will close this PR after I figure out why the test is failing. Thanks

PR #11454, opened 2024-08-14
Adds the IdleTimeout parameter to the RouteAction section of the TcpRoute resource.

ellisonmarks:
https://cloud.google.com/service-mesh/docs/reference/network-services/rest/v1/projects.locations.tcpRoutes#routeaction

```release-note:enhancement
networkservices: added `idle_timeout` field to the `google_network_services_tcp_route` resource (beta)
```

### DIFF
--- a/mmv1/products/networkservices/TcpRoute.yaml
+++ b/mmv1/products/networkservices/TcpRoute.yaml
@@ -184,6 +184,8 @@ properties:
             - !ruby/object:Api::Type::String
               name: idleTimeout
               description: |
-                Specifies the idle timeout for the selected route. The idle timeout is defined as the period in which there are no bytes sent or received on either the upstream or downstream connection.
+                Specifies the idle timeout for the selected route. The idle timeout is defined as the period in which there are no bytes sent or received on either the upstream or downstream connection. If not set, the default idle timeout is 30 seconds. If set to 0s, the timeout will be disabled.
+
+                A duration in seconds with up to nine fractional digits, ending with 's'. Example: "3.5s".
               validation: !ruby/object:Provider::Terraform::Validation
                 regex: '^(0|[1-9][0-9]*)(\.[0-9]{1,9})?s$'
--- a/mmv1/products/networkservices/TcpRoute.yaml
+++ b/mmv1/products/networkservices/TcpRoute.yaml
@@ -181,3 +181,9 @@ properties:
               name: originalDestination
               description: |
                 If true, Router will use the destination IP and port of the original connection as the destination of the request.
+            - !ruby/object:Api::Type::String
+              name: idleTimeout
+              description: |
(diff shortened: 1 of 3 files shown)


mmv1/products/networkservices/TcpRoute.yaml
```diff
@@ -181,3 +181,9 @@ properties:
               name: originalDestination
               description: |
                 If true, Router will use the destination IP and port of the original connection as the destination of the request.
+            - !ruby/object:Api::Type::String
+              name: idleTimeout
+              description: |
+                Specifies the idle timeout for the selected route. The idle timeout is defined as the period in which there are no bytes sent or received on either the upstream or downstream connection.
```
SarahFrench, line 187: I figured it would be worth including the full field description so that users know about the validation when reading the provider docs, and know that a default value is provided by the API when unspecified.

```suggestion
                Specifies the idle timeout for the selected route. The idle timeout is defined as the period in which there are no bytes sent or received on either the upstream or downstream connection. If not set, the default idle timeout is 30 seconds. If set to 0s, the timeout will be disabled.

                A duration in seconds with up to nine fractional digits, ending with 's'. Example: "3.5s".
```

mmv1/products/networkservices/TcpRoute.yaml
```diff
@@ -181,3 +181,9 @@ properties:
               name: originalDestination
               description: |
                 If true, Router will use the destination IP and port of the original connection as the destination of the request.
+            - !ruby/object:Api::Type::String
+              name: idleTimeout
```
SarahFrench, line 185: Given the API provides a default value when this field isn't set by a user, should we add [default_from_api: true](https://googlecloudplatform.github.io/magic-modules/develop/field-reference/#default_from_api) to this field so that the API-side value is pulled down?

mmv1/third_party/terraform/services/networkservices/resource_network_services_tcp_route_test.go.erb
```diff
@@ -112,6 +113,7 @@ resource "google_network_services_tcp_route" "foobar" {
         weight = 1
       }
       original_destination = false
+      idle_timeout = "120s"
```
SarahFrench, line 116: Thanks for ensuring there's a test updating the field!

SarahFrench (commented):
👋 Hi! Just left a few comments

mmv1/products/networkservices/TcpRoute.yaml
ellisonmarks: I'm not sure exactly. From testing, I think the 30s default value is implicit. That is it's used if nothing is set, and not returned by the api.

SarahFrench (approved):
LGTM, thanks!

PR #11455, opened 2024-08-15
`KMS`: add `google_kms_crypto_key_versions` data source

BBBmau:
continuation of https://github.com/GoogleCloudPlatform/magic-modules/pull/11381

splitting up into two parts

This is the first step which addresses adding a plural data source for `cryptoKeyVersion`

~Keeping in draft until `purpose` is added in. All other fields are supported.~

```release-note:new-datasource
`google_kms_crypto_key_versions`
```


BBBmau:
Marked you as a reviewer since you have the most context due to looking at the previously closed PR @slevenick

mmv1/third_party/terraform/services/kms/data_source_google_kms_crypto_key_versions.go
```diff
@@ -0,0 +1,239 @@
...
+		}
+	}
+
+	return nil
+}
+
+func dataSourceKMSCryptoKeyVersionsList(d *schema.ResourceData, meta interface{}, cryptoKeyId string, userAgent string) ([]interface{}, error) {
+	config := meta.(*transport_tpg.Config)
+
+	url, err := tpgresource.ReplaceVars(d, config, "{{KMSBasePath}}{{crypto_key}}/cryptoKeyVersions?filter=state=ENABLED")
```
slevenick, line 146: Do we want to default the filter to only ENABLED keys? We should at least document that if it's what we want

I'm worried that a user would try to get *all* their keys and be confused when only enabled ones come back

mmv1/third_party/terraform/services/kms/data_source_google_kms_crypto_key_versions.go
```diff
@@ -0,0 +1,239 @@
...
+	// err == nil indicates that the billing_project value was found
+	if bp, err := tpgresource.GetBillingProject(d, config); err == nil {
+		billingProject = bp
+	}
+
+	// Always include the filter param, and optionally include the pageToken parameter for subsequent requests
+	var params = make(map[string]string, 0)
+	if filter, ok := d.GetOk("filter"); ok {
+		log.Printf("[DEBUG] Search for cryptoKeyVersions in crypto key %s is using filter ?filter=%s", cryptoKeyId, filter.(string))
+		params["filter"] = filter.(string)
```
slevenick, line 166: Will this overwrite the default filter=state=enabled?

mmv1/third_party/terraform/services/kms/data_source_google_kms_crypto_key_versions.go
slevenick: Is this the correct format for a filter also? filter=state=enabled seems like it should be escaped somehow

mmv1/third_party/terraform/services/kms/data_source_google_kms_crypto_key_versions_test.go
```diff
@@ -0,0 +1,37 @@
...
+func TestAccDataSourceGoogleKmsCryptoKeyVersions_basic(t *testing.T) {
+	asymSignKey := acctest.BootstrapKMSKeyWithPurpose(t, "ASYMMETRIC_SIGN")
+
+	acctest.VcrTest(t, resource.TestCase{
+		PreCheck:                 func() { acctest.AccTestPreCheck(t) },
+		ProtoV5ProviderFactories: acctest.ProtoV5ProviderFactories(t),
+		Steps: []resource.TestStep{
+			{
+				Config: testAccDataSourceGoogleKmsCryptoKeyVersions_basic(asymSignKey.CryptoKey.Name),
+				Check:  resource.TestCheckResourceAttr("data.google_kms_crypto_key_versions.versions", "versions.#", "2"),
```
slevenick, line 20: Can you set up a test where we would expect a different number of versions to exist than the default? Maybe by disabling a key and then making sure it doesn't come back

mmv1/third_party/terraform/services/kms/data_source_google_kms_crypto_key_versions_test.go
```diff
@@ -0,0 +1,37 @@
...
+			},
+			// Asymmetric keys should have a public key
+			{
+				Config: testAccDataSourceGoogleKmsCryptoKeyVersions_basic(asymSignKey.CryptoKey.Name),
+				Check:  resource.TestCheckResourceAttr("data.google_kms_crypto_key_versions.versions", "versions.0.state", "ENABLED"),
```
slevenick, line 25: This is always true based on the filter we use, right?

mmv1/third_party/terraform/services/kms/data_source_google_kms_crypto_key_versions_test.go
```diff
@@ -0,0 +1,37 @@
...
+			},
+		},
+	})
+}
+
+func testAccDataSourceGoogleKmsCryptoKeyVersions_basic(kmsKey string) string {
+	return fmt.Sprintf(`
+data "google_kms_crypto_key_versions" "versions" {
```
slevenick, line 33: Can you add a test that uses a filter to only grab a subset of versions?

slevenick (changes requested):

mmv1/third_party/terraform/services/kms/data_source_google_kms_crypto_key_versions.go
BBBmau: > Do we want to default the filter to only ENABLED keys? We should at least document that if it's what we want
> 
> I'm worried that a user would try to get _all_ their keys and be confused when only enabled ones come back

I think to avoid confusion we can have it include both enabled and disabled versions. My view point is that users would only be interested in ENABLED versions since you are unable to do anything with disable versions until they are enabled.

The format is also correct, i agree it's a bit weird with the current format of `filter=state=enabled`

BBBmau: latest commit includes the filter so that users can filter through the list of versions (such as name, state, etc.)

mmv1/third_party/terraform/services/kms/data_source_google_kms_crypto_key_versions.go
BBBmau: refer to https://github.com/GoogleCloudPlatform/magic-modules/pull/11455#discussion_r1725957893

mmv1/third_party/terraform/services/kms/data_source_google_kms_crypto_key_versions_test.go
BBBmau: I added more test cases that involve the use of filter so that we can get different amount of versions (whether it's based on name or `state:enabled`)

mmv1/third_party/terraform/services/kms/data_source_google_kms_crypto_key_versions_test.go
BBBmau: refer to https://github.com/GoogleCloudPlatform/magic-modules/pull/11455#discussion_r1725958366

mmv1/third_party/terraform/services/kms/data_source_google_kms_crypto_key_versions_test.go
BBBmau: this is test is changed due to the addition of filter: https://github.com/GoogleCloudPlatform/magic-modules/pull/11455#discussion_r1725958366

slevenick (approved):

PR #11465, opened 2024-08-15
Add trigger for downstream generation and test

ScottSuarez:
Apart of community checker removal

```release-note:none

```


melinath (approved):

melinath (changes requested):
apparently cloudbuild doesn't like secrets that are defined but unused. (which I guess makes sense)

ScottSuarez:
Added to the secret env list

melinath (approved):

melinath (changes requested):
sorry for the noise; looks like it's still failing but I don't see a clear error

ScottSuarez:
/gcbrun

I think I know what happened. Need to grant the SA access

melinath (approved):

PR #11466, opened 2024-08-16
fix overwriting of `name` field for IAM Group user in resourceSqlUserRead method

pratikgarg10:

```release-note:bug
sql: fix overwriting of `name` field for IAM Group user for resource `google_sql_user`
```

### DIFF
--- a/mmv1/third_party/terraform/services/sql/resource_sql_user.go
+++ b/mmv1/third_party/terraform/services/sql/resource_sql_user.go
@@ -348,10 +348,13 @@ func resourceSqlUserRead(d *schema.ResourceData, meta interface{}) error {
 	}
 
 	for _, currentUser := range users.Items {
+		var username string
 		if !(strings.Contains(databaseInstance.DatabaseVersion, "POSTGRES") || currentUser.Type == "CLOUD_IAM_GROUP") {
-			name = strings.Split(name, "@")[0]
+			username = strings.Split(name, "@")[0]
+		} else {
+			username = name
 		}
-		if currentUser.Name == name {
+		if currentUser.Name == username {
 			// Host can only be empty for postgres instances,
 			// so don't compare the host if the API host is empty.
 			if host == "" || currentUser.Host == host {


pratikgarg10:
Hi @BBBmau, Could you please take a look at this pull request for review?

ScottSuarez:
Is there a ticket for this issue or could you provide an explanation of the change? I am not familiar with this resource.

ScottSuarez:
Also if a testcase could be added showing exactly what was broken and is now fixed that would be great !

pratikgarg10:
> Is there a ticket for this issue or could you provide an explanation of the change? I am not familiar with this resource.
Here is the issue: https://github.com/hashicorp/terraform-provider-google/issues/17040

**Issue:**
Creating IAM group user resource fails with error "Root resource was present, │ but now absent. "

**Explanation:**
The code snippet in Condition 1 below truncates the SQL user `name` field and overwrites it if the condition is satisfied for any user in the users list, but later the comparison in Condition 2 will fail for valid SQL user `name` due to overwritten truncation of previous user in the list.

Condition 1:
```
if !(strings.Contains(databaseInstance.DatabaseVersion, "POSTGRES") || currentUser.Type == "CLOUD_IAM_GROUP") {
            name = strings.Split(name, "@")[0]
        }
```

Condition 2:
```
if currentUser.Name == name {
	// Host can only be empty for postgres instances,
	// so don't compare the host if the API host is empty.
	if host == "" || currentUser.Host == host {
		user = currentUser
		break
	}
}
```

During IAM group creation on the instance, the condition that triggers the issue is the naming convention used for the built-in users & IAM user/groups and corresponding alphabetical order. If the IAM group user comes last in the order, we'll see the issue.

pratikgarg10:
> Also if a testcase could be added showing exactly what was broken and is now fixed that would be great !

It was tested locally, but difficult to reproduce via unit test

ScottSuarez:
Got it!! This change seems reasonable. We are currently frozen for 6.0 release but I'll approve this once things move along. 

I'm going to leave this PR as unreviewed simply to maintain it in my review queue and that I don't lose track of it.

ScottSuarez (approved):
unfroze early.. merging now.

PR #11467, opened 2024-08-16
Adding manual sweepers for resources in netapp

Mehul3217:
…e regions

```release-note:bug
netapp: Adding sweepers for all resources due to tests being run in multiple regions
```


mmv1/third_party/terraform/services/netapp/resource_netapp_active_directory_sweeper.go.erb
```diff
@@ -0,0 +1,131 @@
+<% autogen_exception -%>
+package netapp
+<% unless version == 'ga' -%>
```
BBBmau, line 3: My only question is what's the reason for wrapping these in a beta wrapper when the resources are not beta? (they look to be GA)

This also allows us to mark these files as `.go` instead of `.go.erb`

Relevant Documentation that talks about version guards: https://googlecloudplatform.github.io/magic-modules/develop/promote-to-ga/#promote-fields-and-resources

Mehul3217: Done

mmv1/third_party/terraform/services/netapp/resource_netapp_active_directory_sweeper.go
```diff
@@ -0,0 +1,128 @@
+package netapp
+
```
BBBmau, line 2: looks like this file has some lint fails, could you address them in this file? Related gha: https://github.com/GoogleCloudPlatform/magic-modules/actions/runs/10535926639/job/29195530738

Mehul3217: Done

PR #11470, opened 2024-08-16
Fix how `data.google_client_config` handles errors from invalid credentials

SarahFrench:

Related to https://github.com/hashicorp/terraform-provider-google/issues/18774

```release-note:bug
resourcemanager: fixed a bug where data.google_client_config failed silently when inadequate credentials were used to configure the provider
```


SarahFrench:
The test in [eb570f7](https://github.com/GoogleCloudPlatform/magic-modules/pull/11470/commits/eb570f763c7305189b1ba4bfe18ab63d17d9423d) fails due to expecting access_token to be set when the credentials are 'bad'. In that scenario the access_token value is left as an empty string.

After 28434ad the test will fail due to an error being returned due to the bad credentials

After tests run on this PR I'll push a change to the test where it expects the error (so the test will be successful when the error is encountered).

SarahFrench:
Love that the other tests came along for the ride, but I'm only concerned with `TestAccDataSourceGoogleClientConfig_credentialsInProviderConfig`

melinath (approved):

roaks3 (approved):

PR #11431, opened 2024-08-12
Add send_age_if_zero field and deprecate no_age

NickElliot:
Redo of https://github.com/GoogleCloudPlatform/magic-modules/pull/11098

Further details: only changes from the above pr are within the `mmv1/third_party/terraform/services/storage/resource_storage_bucket_600_migration.go` file, and adding the schema version/state upgraders to the top of `mmv1/third_party/terraform/services/storage/resource_storage_bucket.go.erb` (i.e. [this commit](https://github.com/GoogleCloudPlatform/magic-modules/pull/11431/commits/a80468e9302f031f123b64f573f072dc9fbf07c3))

**Release Note Template for Downstream PRs (will be copied)**

```release-note:deprecation
storage: deprecated `lifecycle_rule.condition.no_age` field in `google_storage_bucket`. Use the new `lifecycle_rule.condition.send_age_if_zero` field instead.
```

PR #11471, opened 2024-08-16
add state upgrader to set send_age_if_zero virtual field default to false

NickElliot:
follow-up to https://github.com/GoogleCloudPlatform/magic-modules/pull/11431

fixes https://github.com/hashicorp/terraform-provider-google/issues/18232

```release-note:none

```

### DIFF
--- a/mmv1/third_party/terraform/services/storage/resource_storage_bucket.go.erb
+++ b/mmv1/third_party/terraform/services/storage/resource_storage_bucket.go.erb
@@ -48,7 +48,7 @@ func ResourceStorageBucket() *schema.Resource {
 			Read:   schema.DefaultTimeout(4 * time.Minute),
 		},
 
-		SchemaVersion: 2,
+		SchemaVersion: 3,
 		StateUpgraders: []schema.StateUpgrader{
 			{
 				Type:    resourceStorageBucketV0().CoreConfigSchema().ImpliedType(),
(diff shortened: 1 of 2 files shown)


NickElliot:
cc @melinath @kautikdk

c2thorn (commented):
this looks good for setting the virtual field to false. Just to make sure did we try a manual upgrade test?

Also looks like `no_age` has already been removed?

NickElliot:
@c2thorn yeah, I walked through all the scenarios we're looking for from earlier discussion i.e.

1. Users with neither field will see no diff, as the default was removed. age values of 0 are no longer sent, and the dangerous behavior is fixed
2. Users with send_age_if_zero set to false explicitly (the new default) see no diff. They can remove the explicit setting if desired.
3. Users with send_age_if_zero set to true will see a diff. Their next apply will correct the diff, locking the value into state, and will use the new setting in requests to the API (as we retrieve it from state with d.Get in Update)

and that's correct actually 😅 will update the pr details

c2thorn:
> 3. Users with send_age_if_zero set to true will see a diff of false-> true. Their next apply will correct the diff, locking the value into state, and will use the new setting in requests to the API (as we retrieve it from state with d.Get in Update)

Is this unavoidable or can we avoid it by checking if `send_age_if_zero` is already true in the state migration?

NickElliot:
`send_age_if_zero` is true for all users as a general rule in the migration -- as it is the default in 5.x due to preserving the old `no_age` behavior

c2thorn (approved):

NickElliot:
ultimately it is what will impact the fewest users as it has to require users to specifically *want* to add a condition that is always true (as age will always be greater than zero), which is what we're trying to get rid of anyway

NickElliot:
will merge once the preceding 5.x PR finishes review/once any changes potentially needed there are brought upstream to here @melinath

PR #11472, opened 2024-08-16
Remove validation on logging_config::enable_components.

yungdarek122:
The validation does not bring too much value. Many validation logic happens when the backend process the cluster create/patch API such that a valid enum would still be rejected at the backend. OTOH, this validation logic adds complexity in supporting new enum (log types).

This also aligns with monitoring_config::enable_components, which no longer has validation.

```release-note:note
container: remove validation on logging_config::enable_components to allow other log types in the future
```


melinath:
@yungdarek122 is there an open issue about this or a specific plan to update the allowed values here? Otherwise it would be preferable to leave the validation in place, since it does move the error earlier.

melinath (commented):
removing from my review queue pending response

PR #11484, opened 2024-08-19
Support creating Database Migration Service connection profiles connected to existing Cloud SQL and AlloyDB instances 

owaismawasy:
Support creating DMS connection profiles that link to existing Cloud SQL instances/AlloyDB clusters. 

This will allow using such connection profiles as destination to the following migration types:
- PostgreSQL to Cloud SQL (PostgreSQL)
- MySQL to Cloud SQL (MySQL)
- PostgreSQL to AllloyDB

```release-note:enhancement
Support creating DMS connection profiles that link to existing Cloud SQL instances/AlloyDB clusters. 
```


mmv1/products/databasemigrationservice/ConnectionProfile.yaml
```diff
@@ -266,25 +289,37 @@ properties:
     properties:
       - !ruby/object:Api::Type::String
         name: 'host'
-        required: true
+        required_with:
+        - 'postgresql.0.port'
```
BBBmau, line 293: There's some lint errors on this line, could you fix them?

owaismawasy: Done.

mmv1/products/databasemigrationservice/ConnectionProfile.yaml
```diff
@@ -266,25 +289,37 @@ properties:
     properties:
       - !ruby/object:Api::Type::String
         name: 'host'
-        required: true
+        required_with:
```
BBBmau, line 292: what's the reasoning behind replacing `required` with `required_with`?

owaismawasy: DMS started to support a new path, called existing-instance, for the following homogenous migration types:
1. MySQL to Cloud SQL (MySQL)
2. PostgreSQL to Cloud SQL (PostgreSQL)
3. PostgreSQL to AlloyDB

With this new existing-instance path, users can create the destination database instance outside of DMS and then create DMS connection profiles that point to these destination instances. In that case, users need to provide only the CloudSQL ID or AlloyDB ID, and they don't need to provide additional connection params like username, password, host, etc.

For example, a destination connection profile for MySQL -> MySQL migrations will look like this:
```
resource "google_database_migration_service_connection_profile" "source" {
  location              = "us-central1"
  …
  mysql {
    cloud_sql_id = "<csql_id>"
  }
}
```

This means that connection params (e.g. username, password, etc.) are no longer required in all cases. 

That said, for other migration types (e.g. Oracle -> PostgreSQL), it doesn't make sense to provide partial connection params, so we're configuring the resource so that if one connection param is provided (e.g. username), we'll need other params to be provided as well, and we're using the required_with to achieve that.

BBBmau (approved):
Thank you for the thorough explanation! I also came across this which matches what you just explained: https://cloud.google.com/database-migration/docs/mysql/configure-source-database#overview-mysql

PR #8460, opened 2023-07-27
Enhance BigQuery table schema input validation

wj-chen:
Fixes https://github.com/hashicorp/terraform-provider-google/issues/14659.

If this PR is for Terraform, I acknowledge that I have:

- [x] Searched through the [issue tracker](https://github.com/hashicorp/terraform-provider-google/issues) for an open issue that this either resolves or contributes to, commented on it to claim it, and written "fixes {url}" or "part of {url}" in this PR description. If there were no relevant open issues, I opened one and commented that I would like to work on it (not necessary for very small changes).
- [x] Ensured that all new fields I added that can be set by a user appear in at least one [example](https://github.com/GoogleCloudPlatform/magic-modules/tree/main/mmv1/templates/terraform/examples) (for generated resources) or [third_party test](https://github.com/GoogleCloudPlatform/magic-modules/tree/main/mmv1/third_party/terraform/tests) (for handwritten resources or update tests).
- [x] [Generated Terraform providers](https://github.com/GoogleCloudPlatform/magic-modules/blob/main/docs/content/docs/getting-started/generate-providers.md), and ran [`make test` and `make lint`](https://googlecloudplatform.github.io/magic-modules/docs/getting-started/run-provider-tests/#run-unit-tests) in the generated providers to ensure it passes unit and linter tests.
- [x] [Ran](https://github.com/GoogleCloudPlatform/magic-modules/blob/main/docs/content/develop/run-tests.md) relevant acceptance tests using my own Google Cloud project and credentials (If the acceptance tests do not yet pass or you are unable to run them, please let your reviewer know).
- [x] Read [Write release notes](https://googlecloudplatform.github.io/magic-modules/contribute/release-notes/) before writing my release note below.

**Release Note Template for Downstream PRs (will be copied)**

```release-note:breaking-change
added more input validations for BigQuery table schema
```

PR #11486, opened 2024-08-19
Add regex for time format to `resource_compute_resource_policies`

karolgorc:
closes [issue #8460](https://github.com/hashicorp/terraform-provider-google/issues/8460)

This is probably a breaking change because some `terraform apply` commands won't go through if they have a wrong format. The regex will throw an error. But it also ensures that it's the correct format for the API.

Will write a test case for this into the PR if it's an approved solution.

```release-note:breaking-change
compute: `resource_policies` now have a regex on their time format that will allow for correct calls to the API
```

### DIFF
--- a/mmv1/products/compute/ResourcePolicy.yaml
+++ b/mmv1/products/compute/ResourcePolicy.yaml
@@ -142,7 +142,7 @@ properties:
                   where HH : [00-23] and MM : [00] GMT. eg: 21:00
                 required: true
                 validation: !ruby/object:Provider::Terraform::Validation
-                  function: 'verify.ValidateHourlyOnly'
+                  regex: '^([01][0-9]|2[0-3]):00$'
           - !ruby/object:Api::Type::NestedObject
             name: 'dailySchedule'
             description: |
@@ -165,7 +165,7 @@ properties:
                   both 13:00-5 and 08:00 are valid.
                 required: true
                 validation: !ruby/object:Provider::Terraform::Validation
-                  function: 'verify.ValidateHourlyOnly'
+                  regex: '^([01][0-9]|2[0-3]):00$'
           - !ruby/object:Api::Type::NestedObject
             name: 'weeklySchedule'
             description: |


mmv1/products/compute/ResourcePolicy.yaml
```diff
@@ -142,7 +142,7 @@ properties:
                   where HH : [00-23] and MM : [00] GMT. eg: 21:00
                 required: true
                 validation: !ruby/object:Provider::Terraform::Validation
-                  function: 'verify.ValidateHourlyOnly'
+                  regex: '^([01][0-9]|2[0-3]):00$'
```
SarahFrench, line 145: Note for posterity:

I tested the regex and can see that these cases are all identified as invalid

- 0:00
- 0:01
- 1:00
- 10:10
- 24:00

and these are identified as valid:

- 01:00
- 10:00
- 20:00
- 23:00

SarahFrench (commented):
Hi, thanks for your PR!

My understanding of the problem is that currently if a user was to use "1:00" in their configuration Terraform will identify a permadiff because the API will return "01:00" and there's a mismatch as a result.

The new regex for validating input looks correct to me, but you're also correct that this would be a breaking change for some users - e.g. some might have used [ignore_changes](https://developer.hashicorp.com/terraform/language/meta-arguments/lifecycle#ignore_changes) on this field to navigate the permadiff.

Another approach is to add [diff suppression to the field](https://googlecloudplatform.github.io/magic-modules/develop/field-reference/#diff_suppress_func) so that the provider regards "1:00" and "01:00" (for example) as equivalent and doesn't raise a diff in this situation. Would you consider this alternative instead, as it's non-breaking?

karolgorc:
ok i wasn't aware that the Google API actually translated the given value from H:MM to HH:MM. If that happens then the suppress diff seems like a better solution

- Added a HourlyFormatSuppressDiff() to `start_time` under `hourly_schedule` and `daily_schedule`
- Added unit test for it

SarahFrench (commented):
>@SarahFrench This PR has been waiting for review for 3 weekdays. Please take a look!

I'm OOO and will review this on Tuesday

SarahFrench (commented):
This looks good to me, thanks for the testing of the diff suppression function! Before merging I'd like there to be an acceptance test that contains a `start_time` value in H:mm format and shows that there aren't errors.

An approach for this could be to update [this example file](https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/templates/terraform/examples/resource_policy_basic.tf.erb), which is referenced here in [ResourcePolicy.yaml ](https://github.com/GoogleCloudPlatform/magic-modules/blob/a6206ac9911804d06cf156ca028f33e5159a95bc/mmv1/products/compute/ResourcePolicy.yaml#L45-L49), so that the `start_time` value can be passed in from the YAML as a variable (similar to what you see for `name`). After doing this, you could add another entry to the `examples` list in ResourcePolicy.yaml but set the value for `start_time` in the H:mm format. You'd also want to include `skip_docs: true` in that block, so that the example is only used to generate a test and doesn't accidentally promote using H:mm format format in the official provider documentation

karolgorc:
Using the same `example` file generates 2 test functions that have the same name. If there is a way to generate 2 unique testing functions using the same `example` file please LMK! But for now i just created another file that stores the new config with `4:00` instead of `04:00`.

```
--- PASS: TestAccComputeResourcePolicy_resourcePolicyHourlyFormatExample (12.39s)
PASS
ok      github.com/hashicorp/terraform-provider-google/google/services/compute    12.417s
```

SarahFrench:
Ah, of course name collision would happen 🤦 . In that case I'm afraid that the route forward is to create a new example file with a different name that has the start_time value set in H:mm. Because of this having the time set via a variable wouldn't be necessary. Sorry for leading you down an incorrect path!

karolgorc:
>In that case I'm afraid that the route forward is to create a new example file with a different name that has the start_time value set in H:mm

Done [here](https://github.com/GoogleCloudPlatform/magic-modules/pull/11486/commits/ec4da7dde87d6fa9e498caa29f0742524d778312)

SarahFrench (approved):
Thanks, and sorry again! The new generated test passes so I'll merge. The `TestAccComputeInstanceTemplate_withNamePrefix` failure is unrelated to this PR.

PR #11490, opened 2024-08-19
Stop passing version as an argument when making a new PF provider

SarahFrench:
# Context

The version variable comes from the version package, here: [version.ProviderVersion](https://github.com/hashicorp/terraform-provider-google/blob/924af2bcae221ec3c65d027b617dbcd0b7b2605f/version/version.go#L7)

When we release a new version of the provider that string is overwritten via `ldflags` with the semver version of the new release ([via GoReleaser, here](https://github.com/hashicorp/terraform-provider-google/blob/924af2bcae221ec3c65d027b617dbcd0b7b2605f/.goreleaser.yml#L32-L33)), and when we run acceptance tests "dev" is replaced with "acc" ([see Makefile command here](https://github.com/hashicorp/terraform-provider-google/blob/924af2bcae221ec3c65d027b617dbcd0b7b2605f/GNUmakefile#L18)).

The provider only uses this version number to create a UserAgent string that includes the version of the provider.
- [SDK creation of UserAgent when configuring the provider](https://github.com/hashicorp/terraform-provider-google/blob/924af2bcae221ec3c65d027b617dbcd0b7b2605f/google/provider/provider.go#L828) - note that version.ProviderVersion is used directly
- [PF creation of UserAgent when configuring the provider](https://github.com/hashicorp/terraform-provider-google/blob/924af2bcae221ec3c65d027b617dbcd0b7b2605f/google/fwtransport/framework_config.go#L191)

# This PR

This PR refactors how the provider version is passed into the code that returns an instance of a plugin-framework implemented provider.

This PR is part of breaking [a large future PR](https://github.com/SarahFrench/terraform-provider-google/pull/9) into smaller chunks! There will be future changes to the `New` function, so removing this argument is a first step. When that work is complete we'll actually be removing the parallel user agent creation code completely and only have it implemented in one place.

```release-note:none

```

### DIFF
--- a/mmv1/third_party/terraform/fwprovider/framework_provider.go.erb
+++ b/mmv1/third_party/terraform/fwprovider/framework_provider.go.erb
@@ -19,6 +19,7 @@ import (
     "github.com/hashicorp/terraform-provider-google/google/fwmodels"
     "github.com/hashicorp/terraform-provider-google/google/fwtransport"
     "github.com/hashicorp/terraform-provider-google/google/services/resourcemanager"
+	"github.com/hashicorp/terraform-provider-google/version"
     <% unless version == 'ga' -%>
     "github.com/hashicorp/terraform-provider-google/google/services/firebase"
     <% end -%>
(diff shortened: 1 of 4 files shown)


zli82016 (approved):
LGTM

PR #11491, opened 2024-08-19
Add PUPI import and export to google_compute_network_peering_routes_config

benjamin-maynard:
This PR adds the `import_subnet_routes_with_public_ip` and `export_subnet_routes_with_public_ip` fields to the compute resource `google_compute_network_peering_routes_config`.

Fixes: https://github.com/hashicorp/terraform-provider-google/issues/11573

Although there is an API side default, I set these fields as required without defaults as people may have this resource already in state, and it is undesirable to change without explicitly choosing to. 

This mirrors the approach of `import_custom_routes` and `export_custom_routes`. Please let me know if there is a better approach.

```release-note:enhancement
compute: added `import_subnet_routes_with_public_ip` and `export_subnet_routes_with_public_ip` fields to `google_compute_network_peering_routes_config` resource
```

### DIFF
--- a/mmv1/products/compute/NetworkPeeringRoutesConfig.yaml
+++ b/mmv1/products/compute/NetworkPeeringRoutesConfig.yaml
@@ -118,7 +118,7 @@ properties:
       IPv4 special-use ranges are always exported to peers and
       are not controlled by this field.
     send_empty_value: true
-    required: true
+    default_from_api: true
   - !ruby/object:Api::Type::Boolean
     name: 'importSubnetRoutesWithPublicIp'
     description: |
@@ -126,4 +126,4 @@ properties:
       IPv4 special-use ranges are always imported from peers and
       are not controlled by this field.
     send_empty_value: true
-    required: true
+    default_from_api: true
--- a/mmv1/products/compute/NetworkPeeringRoutesConfig.yaml
+++ b/mmv1/products/compute/NetworkPeeringRoutesConfig.yaml
@@ -115,7 +115,7 @@ properties:
     name: 'exportSubnetRoutesWithPublicIp'
     description: |
       Whether subnet routes with public IP range are exported.
-      The default value is true, all subnet routes are exported.
+      All subnet routes are exported.
       IPv4 special-use ranges are always exported to peers and
       are not controlled by this field.
     send_empty_value: true
@@ -124,7 +124,6 @@ properties:
     name: 'importSubnetRoutesWithPublicIp'
     description: |
       Whether subnet routes with public IP range are imported.
-      The default value is false.
       IPv4 special-use ranges are always imported from peers and
       are not controlled by this field.
     send_empty_value: true
--- a/mmv1/products/compute/NetworkPeeringRoutesConfig.yaml
+++ b/mmv1/products/compute/NetworkPeeringRoutesConfig.yaml
@@ -115,7 +115,6 @@ properties:
     name: 'exportSubnetRoutesWithPublicIp'
     description: |
       Whether subnet routes with public IP range are exported.
-      All subnet routes are exported.
       IPv4 special-use ranges are always exported to peers and
       are not controlled by this field.
     send_empty_value: true
--- a/mmv1/products/compute/NetworkPeeringRoutesConfig.yaml
+++ b/mmv1/products/compute/NetworkPeeringRoutesConfig.yaml
@@ -111,3 +111,21 @@ properties:
     description: |
       Whether to import the custom routes to the peer network.
     send_empty_value: true
+  - !ruby/object:Api::Type::Boolean
+    name: 'exportSubnetRoutesWithPublicIp'
+    description: |
+      Whether subnet routes with public IP range are exported.
+      The default value is true, all subnet routes are exported.
+      IPv4 special-use ranges are always exported to peers and
+      are not controlled by this field.
+    send_empty_value: true
+    required: true
+  - !ruby/object:Api::Type::Boolean
+    name: 'importSubnetRoutesWithPublicIp'
+    description: |
+      Whether subnet routes with public IP range are imported.
+      The default value is false.
+      IPv4 special-use ranges are always imported from peers and
+      are not controlled by this field.
+    send_empty_value: true
+    required: true
--- a/mmv1/templates/terraform/examples/network_peering_routes_config_basic.tf.erb
+++ b/mmv1/templates/terraform/examples/network_peering_routes_config_basic.tf.erb
@@ -4,6 +4,8 @@ resource "google_compute_network_peering_routes_config" "<%= ctx[:primary_resour
 
   import_custom_routes = true
   export_custom_routes = true
+  import_subnet_routes_with_public_ip = true
+  export_subnet_routes_with_public_ip = true
 }
 
 resource "google_compute_network_peering" "peering_primary" {


mmv1/products/compute/NetworkPeeringRoutesConfig.yaml
```diff
@@ -111,3 +111,19 @@ properties:
...
+    send_empty_value: true
+    required: true
+  - !ruby/object:Api::Type::Boolean
+    name: 'importSubnetRoutesWithPublicIp'
+    description: |
+      Whether subnet routes with public IP range are imported.
+      IPv4 special-use ranges are always imported from peers and
+      are not controlled by this field.
+    send_empty_value: true
+    required: true
```
kevintchan-goog, line 129: I'm concerned that making these fields required is a breaking change to existing deployments. It's why this [testcase](https://storage.cloud.google.com/ci-vcr-logs/beta/refs/heads/auto-pr-11491/artifacts/afe31881-f44c-4460-aaaf-114e8b3efe11/build-log/recording_build/TestAccComputeNetworkPeeringRoutesConfig_networkPeeringRoutesConfigGkeExample_recording_test.log) is failing.

benjamin-maynard: My concern is that:

1) Private Services Access configures VPC Peering connections with values that differ from the defaults for for [networks.updatePeering](https://cloud.google.com/compute/docs/reference/rest/v1/networks/updatePeering).

   `exportSubnetRoutesWithPublicIp`, `importSubnetRoutesWithPublicIp`, `exportCustomRoutes`, and `importCustomRoutes` are all set to `false` when VPC Peering is created by PSA (at least in my testing with Memorystore for Redis). Whereas the API defaults are `true`, `false`, `false` and `false` respectively.

2) Users may already be managing this resource through Terraform, but have changed `exportSubnetRoutesWithPublicIp` and `importSubnetRoutesWithPublicIp` manually due to lack of TF support.

So I made this required to ensure that users are intentional about setting these values. The danger I can see with setting defaults and making this not required are that we risk users unintentionally changing the values if they do not notice the diff on apply.

The main danger being changing `importSubnetRoutesWithPublicIp` to the default of `false`, which might withdraw routes needed to access services in another VPC and cause production issues.

So I think it might be better to fail safely. But if there is another approach or you disagree LMK.

shuyama1: yes, making an optional field required or adding a new required field is strictly a breaking change in Terraform that needs to wait until the next major release.

benjamin-maynard: Ok, so I guess the main question is are we in agreement that these fields should be required, without a default based on the rationale above?

If no - what should the alternative be? Make the field default to the API default, and trust users to validate the diffs if they exist?

If yes - whats the process for queuing this PR for the next major release?

kevintchan-goog: Hmmm you raise good points ben.  Let me try to figure out whether any Producers in PSA relies on importSubnetRoutesWithPublicIp.  My hunch is that they wouldn't.

benjamin-maynard: @kevintchan-goog - I think it depends on the ranges that the customer is using in their VPC, rather than the design on the producer side.

If they are using non RFC-1918 PUPI subnets in their VPC (e.g. [in GKE](https://cloud.google.com/kubernetes-engine/docs/how-to/configuring-privately-used-public-ips-for-gke)) - then traffic can reach the producer VPC, but there are no returns routes exported to the producer VPC, so the traffic is blackholed and the service is inaccessible.

By default, PSA sets `exportSubnetRoutesWithPublicIp` and `importSubnetRoutesWithPublicIp` to `true` on the producer side of the peering, but not on the consumer side.

kevintchan-goog: ## if we make fields required:

- all users need to update their TF config. This is a mass inconvenience. **How frequent are these types of updates for terraform users?**
- customers risk mistakenly setting a field to wrong value and breaking network connectivity. True/false could both be considered wrong.

**In this case can we omit explicit terraform default values and rely on API defaults instad?**

I'm leaning towards required & either omitted defaults or API-defaults. This seems like the safest option to not get blamed for customer outages.

## If we make fields non-required
- all users need to check that version update doesn't make a secret update to their config. This is not an inconvenience
- customers risk missing an update breaking connectivity. This is an easier mistake to make and blame GCP for.

benjamin-maynard: Thanks @kevintchan-goog  - Agree on `default_from_api`. I thought this took the API default, but TIL it will default to the current value given the peering resource already exists. Updated the PR.

benjamin-maynard: Hmm - actually no. I am not sure this will work.

Setting `default_from_api` to `true` means that if this is already set to `true` by PSA or the user in the console, setting Terraform to `false` means the field is omitted from the PATCH request and it becomes impossible to set this to `false`.

I tried combining with `send_empty_value` which works well in my local testing but the test case fails and https://googlecloudplatform.github.io/magic-modules/develop/field-reference/#send_empty_value documents:

> send_empty_value and default_from_api cannot both be true on the same field.

So not sure on a path forward.

shuyama1:
Sorry for missing this PR on my review list. Taking a look now.

/gcbrun for triggering a new presubmit run

shuyama1 (commented):
The test failed due to error:
```
resource_compute_network_peering_routes_config_generated_test.go:35: Step 1/2 error: After applying this test step and performing a `terraform refresh`, the plan was not empty.
        stdout
        
        
        -/+ destroy and then create replacement
        
          # google_compute_network_peering.peering_primary must be replaced
        -/+ resource "google_compute_network_peering" "peering_primary" {
              ~ id                                  = "tf-test-primary-networkwgqeabv2ol/tf-test-primary-peeringwgqeabv2ol" -> (known after apply)
              - import_subnet_routes_with_public_ip = true -> null # forces replacement
                name                                = "tf-test-primary-peeringwgqeabv2ol"
              ~ peer_network                        = "https://www.googleapis.com/compute/beta/projects/ci-test-project-188019/global/networks/tf-test-secondary-networkwgqeabv2ol" -> "projects/ci-test-project-188019/global/networks/tf-test-secondary-networkwgqeabv2ol"
              ~ state                               = "ACTIVE" -> (known after apply)
              ~ state_details                       = "[2024-09-16T20:42:55.366-07:00]: Connected." -> (known after apply)
                # (5 unchanged attributes hidden)
            }
        
```
This is likely that setting `import_subnet_routes_with_public_ip` and `export_subnet_routes_with_public_ip` in `google_compute_network_peering_routes_config` causes a [permadiff](https://googlecloudplatform.github.io/magic-modules/develop/permadiff/#fix-a-permadiff) in `google_compute_network_peering` field in the `google_compute_network_peering` resource. You may likely need to make that field optional + computed to fix the issue

c2thorn:
Hi @benjamin-maynard 

Tomorrow, the Magic Modules repository is scheduled to undergo a language migration from Ruby to Go. You can view more details about this in our announcement here: https://github.com/hashicorp/terraform-provider-google/issues/19583 (or [go/mm-migration-announcement](go/mm-migration-announcement) if you are a Googler)

This open pull request may become incompatible due to most YAML and .erb files converting to Go-compatible files.

Our team (Magic Modules repository maintainers) has tooling to automatically convert changes to the new language, and we can prepare a new commit for this pull request that is compatible with the migration.

In order to push the new changes to your pull request, we would need to <ins>force push the commit to your fork's branch</ins>. Our tooling saves a backup branch before converting, so we could rollback or open a new pull request if needed. We would also work with you and the PR reviewer in the event additional changes are needed.

You also have the option to update the pull request yourself after the migration. You can view a [preview branch](https://github.com/GoogleCloudPlatform/magic-modules/tree/FEATURE-BRANCH-rewrite-switchover-preview) and [updated documentation](https://googlecloudplatform.github.io/magic-modules/reference/ruby-go-changes/) related to the migration changes.

We will take no action until we have your explicit permission to push changes to your fork's branch used for this pull request. Let me or your reviewer know if you have any further questions!

PR #11184, opened 2024-07-17
Add update support to serverTlsPolicy field in region_target_https_proxy

gservat:
Adds support to update "server_tls_policy" field in "compute_region_target_https_proxy" via the [PATCH](https://cloud.google.com/compute/docs/reference/rest/v1/regionTargetHttpsProxies/patch) endpoint, removing its ForceNew flag, as well as related tests.

Fixes: https://github.com/hashicorp/terraform-provider-google/issues/18757

**Release Note Template for Downstream PRs (will be copied)**

```release-note:enhancement
compute: added update support for the 'server_tls_policy' field in 'compute_region_target_https_proxy' resource
```

PR #11496, opened 2024-08-20
Improve handling when enabling/disabling server TLS policy in global target HTTPS proxies

gservat:
This PR implements similar behavior as in https://github.com/GoogleCloudPlatform/magic-modules/pull/11184 but for **global** target HTTPS proxies.

Prior to this PR, you could enable a server TLS policy on a global HTTPS proxy, but disabling it would not work as removing `server_tls_policy` removes it from the `PATCH` payload entirely, so Google simply leaves the server TLS policy as is. What we actually need to do is send `serverTlsPolicy = null` if it's not set, which is what this PR does.

We're also adding a note about having to add the `create_before_destroy = true` lifecycle on `google_network_security_server_tls_policy` as removing the resource would make Terraform destroy the resource first _before_ creating/updating other resources, and that would result in a "resource already in use" error. This lifecycle rule means Terraform will create/update resources (in this case it would update `google_compute_target_https_proxy` first and remove the association with the `google_network_security_server_tls_policy` resource) before destroying the resource.

I've updated the tests so that it creates the resources _without_ a server TLS policy, then it adds the `server_tls_policy` and related resources, and then it removes them again. This is to ensure that we can successfully enable & disable mTLS.

```release-note:enhancement
compute: allowed disabling 'server_tls_policy' during update in 'google_compute_target_https_proxy' resources
```

### DIFF
--- a/mmv1/products/compute/TargetHttpsProxy.yaml
+++ b/mmv1/products/compute/TargetHttpsProxy.yaml
@@ -221,7 +221,7 @@ properties:
       load balancer (classic), this option is not available publicly.
   - !ruby/object:Api::Type::ResourceRef
     name: 'serverTlsPolicy'
-    resource: 'ServerSslPolicy'
+    resource: 'ServerTlsPolicy'
     imports: 'selfLink'
     description: |
       A URL referring to a networksecurity.ServerTlsPolicy
@@ -233,6 +233,11 @@ properties:
       INTERNAL_SELF_MANAGED and which with EXTERNAL, EXTERNAL_MANAGED
       loadBalancingScheme consult ServerTlsPolicy documentation.
       If left blank, communications are not encrypted.
+
+      If you remove this field from your configuration at the same time as
+      deleting or recreating a referenced ServerTlsPolicy resource, you will
+      receive a resourceInUseByAnotherResource error. Use lifecycle.create_before_destroy
+      within the ServerTlsPolicy resource to avoid this.
     update_verb: :PATCH
     update_url: 'projects/{{project}}/global/targetHttpsProxies/{{name}}'
     fingerprint_name: 'fingerprint'
--- a/mmv1/templates/terraform/encoders/compute_target_https_proxy.go.erb
+++ b/mmv1/templates/terraform/encoders/compute_target_https_proxy.go.erb
@@ -12,7 +12,7 @@ if _, ok := obj["certificateManagerCertificates"]; ok {
 // in the "PATCH" payload so if you were to remove a server TLS policy from a target HTTPS proxy, it would NOT remove
 // the association.
 if _, ok := obj["serverTlsPolicy"]; !ok {
-        obj["serverTlsPolicy"] = nil
+	obj["serverTlsPolicy"] = nil
 }
 
 return obj, nil
--- a/mmv1/templates/terraform/encoders/compute_target_https_proxy.go.erb
+++ b/mmv1/templates/terraform/encoders/compute_target_https_proxy.go.erb
@@ -7,4 +7,12 @@ if _, ok := obj["certificateManagerCertificates"]; ok {
 	obj["sslCertificates"] = obj["certificateManagerCertificates"]
 	delete(obj, "certificateManagerCertificates")
 }
+
+// Send null if serverTlsPolicy is not set. Without this, Terraform would not send any value for `serverTlsPolicy`
+// in the "PATCH" payload so if you were to remove a server TLS policy from a target HTTPS proxy, it would NOT remove
+// the association.
+if _, ok := obj["serverTlsPolicy"]; !ok {
+        obj["serverTlsPolicy"] = nil
+}
+
 return obj, nil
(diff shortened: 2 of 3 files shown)


mmv1/products/compute/TargetHttpsProxy.yaml
```diff
@@ -221,7 +221,7 @@ properties:
       load balancer (classic), this option is not available publicly.
   - !ruby/object:Api::Type::ResourceRef
     name: 'serverTlsPolicy'
-    resource: 'ServerSslPolicy'
+    resource: 'ServerTlsPolicy'
```
ScottSuarez, line 224: I'm not sure I completely understand the rename here. Do you have any context to share with me?

mmv1/templates/terraform/encoders/compute_target_https_proxy.go.erb
```diff
@@ -7,4 +7,12 @@ if _, ok := obj["certificateManagerCertificates"]; ok {
 	obj["sslCertificates"] = obj["certificateManagerCertificates"]
 	delete(obj, "certificateManagerCertificates")
 }
-return obj, nil
\ No newline at end of file
+
+// Send null if serverTlsPolicy is not set. Without this, Terraform would not send any value for `serverTlsPolicy`
+// in the "PATCH" payload so if you were to remove a server TLS policy from a target HTTPS proxy, it would NOT remove
```
ScottSuarez, line 12: What is the behavior today. If we remove it when it's not set I would expect a persistent diff.

ScottSuarez (changes requested):
Have you looking to simply using the field property `send_empty_value`.

https://googlecloudplatform.github.io/magic-modules/develop/field-reference/#send_empty_value

mmv1/products/compute/TargetHttpsProxy.yaml
gservat: @ScottSuarez: this rename isn't actually necessary. I just noticed that it was called `ServerSslPolicy` but the field is actually server **TLS** policy and every other reference says TLS (not SSL) so I updated it, ran `make provider` and did not see a change in the code so I figured it was harmless. Is it? Found [this code](https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/api/type.rb#L624), which talks about a reference to another resource but I'm not sure what is being referenced here. If you think it should stay as is, happy to revert this change.

mmv1/templates/terraform/encoders/compute_target_https_proxy.go.erb
gservat: So the behavior today is that if you remove `server_tls_policy`, it is completely excluded from the payload in the `PATCH` request, so Google simply leaves the `serverTlsPolicy` set as it was. This change makes it so that if `server_tls_policy` is not set, then `serverTlsPolicy` will be `null` in the `PATCH` payload, which actually removes the association to the server TLS policy (the behavior we want).

gservat:
> Have you looking to simply using the field property `send_empty_value`.
> 
> https://googlecloudplatform.github.io/magic-modules/develop/field-reference/#send_empty_value

The issue with this option is that it sends an empty value (as the name suggests), but that's not what we want when the value is missing. Google expects either `null` or a non-empty value. I tried it and got the expected error:

```
TF_ACC=1 TF_SCHEMA_PANIC_ON_ERROR=1 go test ./google-beta/services/compute -v -run=TestAccComputeTargetHttpsProxyServerTlsPolicy_update -timeout 240m -ldflags="-X=github.com/hashicorp/terraform-provider-google-beta/version.ProviderVersion=acc"
=== RUN   TestAccComputeTargetHttpsProxyServerTlsPolicy_update
=== PAUSE TestAccComputeTargetHttpsProxyServerTlsPolicy_update
=== CONT  TestAccComputeTargetHttpsProxyServerTlsPolicy_update
    vcr_utils.go:152: Step 3/3 error: Error running apply: exit status 1

        Error: Error updating TargetHttpsProxy "projects/gservat-75cb0b26/global/targetHttpsProxies/tf-test-httpsproxy-1v3klw8hi2": googleapi: Error 400: Invalid value for field 'resource.serverTlsPolicy': ''. Must have between 1 and 1024 character(s), invalid

          with google_compute_target_https_proxy.foobar,
          on terraform_plugin_test.tf line 4, in resource "google_compute_target_https_proxy" "foobar":
           4: resource "google_compute_target_https_proxy" "foobar" {

    testing_new.go:90: Error running post-test destroy, there may be dangling resources: exit status 1

        Error: Error when reading or editing ServerTlsPolicy: googleapi: Error 400: Resource '"projects/gservat-75cb0b26/locations/global/serverTlsPolicies/tf-test-server-tls-policy-1v3klw8hi2"' is already being used by resource(s) '"//compute.googleapis.com/projects/529811265946/locations/global/targetHttpsProxies/tf-test-httpsproxy-1v3klw8hi2"'

--- FAIL: TestAccComputeTargetHttpsProxyServerTlsPolicy_update (221.43s)
FAIL
FAIL    github.com/hashicorp/terraform-provider-google-beta/google-beta/services/compute        221.768s
FAIL
make: *** [testacc] Error 1
```

Perhaps magic-modules could support `send_null_value` as an option?

ScottSuarez:
This LGTM but we are currently frozen due to 6.0 release. This will need to wait until next week to be addressed.

gservat:
@ScottSuarez: OK, no worries. Can the failed tests be ignored since they're completely unrelated to my changes?

ScottSuarez (approved):

ScottSuarez:
Yes no problem. We unfroze early. Merging now.

PR #11500, opened 2024-08-20
initial fix

maxi-cit:
Fixes: https://github.com/hashicorp/terraform-provider-google/issues/18793

Fixes an issue regarding GKE Resource Manager Tags which was causing in-place update due to not handling null values properly

```release-note:bug
container: fixed update inconsistency in `google_container_cluster`
```


BBBmau (approved):
looks good, thanks for adding these checks!

PR #11506, opened 2024-08-20
Remove SchemaConfigModeAttr

c2thorn:
fixes https://github.com/hashicorp/terraform-provider-google/issues/12824

Compute subnetwork preparation change: https://github.com/GoogleCloudPlatform/magic-modules/pull/11410
Guest Accelerator preparation change: https://github.com/GoogleCloudPlatform/magic-modules/pull/11425

Composer's `ip_allocation_policy` did not actually send an empty list when specifying an empty block, and it is create-only. Therefore removing configModeAttr doesn't make a difference

`google_compute_instance_from_template` and `google_compute_instance_from_machine_image` functionality for empty blocks is being removed to prevent accidental breaking changes for new subfields from the parent `google_compute_instance`

```release-note:breaking-change
composer: `ip_allocation_policy = []` in `google_composer_environment` is no longer valid configuration. Removing the field from configuration should not produce a diff.
```

```release-note:breaking-change
compute: `secondary_ip_ranges = []` in `google_compute_subnetwork` is no longer valid configuration. To set an explicitly empty list, use `send_secondary_ip_range_if_empty` and completely remove `secondary_ip_range` from config.
```

```release-note:breaking-change
compute: `guest_accelerator = []` is no longer valid configuration in `google_compute_instance`. To explicitly set an empty list of objects, set guest_accelerator.count = 0.
```

```release-note:breaking-change
compute: `google_compute_instance_from_template` and `google_compute_instance_from_machine_image` `network_interface.alias_ip_range, network_interface.access_config, attached_disk, guest_accelerator, service_account, scratch_disk` can no longer be set to an empty block `[]`. Removing the fields from configuration should not produce a diff.
```

```release-note:breaking-change
container: `guest_accelerator = []` is no longer valid configuration in `google_container_cluster` and `google_container_node_pool`. To explicitly set an empty list of objects, set guest_accelerator.count = 0.
```

```release-note:breaking-change
container: `guest_accelerator.gpu_driver_installation_config = []` and `guest_accelerator.gpu_sharing_config = []` are no longer valid configuration in `google_container_cluster` and `google_container_node_pool`. Removing the fields from configuration should not produce a diff.
```

### DIFF
--- a/mmv1/templates/terraform/schema_property.erb
+++ b/mmv1/templates/terraform/schema_property.erb
@@ -26,9 +26,6 @@
 <% if property.default_from_api -%>
 	Computed: true,
 	Optional: true,
-	<% if property.schema_config_mode_attr -%>
-	ConfigMode: schema.SchemaConfigModeAttr,
-	<% end -%>
 <% elsif property.required -%>
   Required: true,
 <% elsif property.output -%>
--- a/docs/content/develop/breaking-changes/breaking-changes.md
+++ b/docs/content/develop/breaking-changes/breaking-changes.md
@@ -84,9 +84,6 @@ For more information, see
 * <a name="field-removing-diff-suppress"></a> Removing diff suppression from a field.
   * For MMv1 resources, removing `diff_suppress_func` from a field.
   * For handwritten resources, removing `DiffSuppressFunc` from a field.
-* <a name="field-adding-subfield-to-config-mode-attr"></a> Adding a subfield to
-  a SchemaConfigModeAttr field.
-  * Subfields of SchemaConfigModeAttr fields are treated as required even if the schema says they are optional.
 * Removing update support from a field.
 
 ### Making validation more strict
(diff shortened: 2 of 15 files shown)


mmv1/templates/terraform/schema_property.erb
```diff
@@ -26,9 +26,6 @@
 <% if property.default_from_api -%>
 	Computed: true,
 	Optional: true,
-	<% if property.schema_config_mode_attr -%>
```
c2thorn, line 29: Quick way to stop accidental generation going forward. Will need to fully remove probably after/during Go Rewrite

c2thorn:
`terraform-google-conversion-build-and-unit-tests` is already broken and will be fixed after the full 6.0 merge

rileykarson (dismissed):

rileykarson (approved):
I think `TestAccComputeInstance_guestAcceleratorSkip` doesn't get re-recorded because the old cassette allows it to replay successfully

c2thorn:
> I think `TestAccComputeInstance_guestAcceleratorSkip` doesn't get re-recorded because the old cassette allows it to replay successfully

I also ran it manually before pushing

PR #11507, opened 2024-08-21
Add InspectTemplateModifiedCadence to BigQueryTarget and CloudSqlTarg…

patrickmoy:
…et in DiscoveryConfig. Add TagResources action to DiscoveryConfig.

Added new feature fields (InspectTemplateModifiedCadence and TagResources action) to DiscoveryConfig resource. Modify the update tests to support new fields. 

```release-note:enhancement
dlp: added `inspect_template_modified_cadence` field to `big_query_target` and `cloud_sql_target` in `google_data_loss_prevention_discovery_config`
```

```release-note:enhancement
dlp: added `tag_resources` field to `google_data_loss_prevention_discovery_config`.
```

### DIFF
--- a/mmv1/products/dlp/DiscoveryConfig.yaml
+++ b/mmv1/products/dlp/DiscoveryConfig.yaml
@@ -206,6 +206,49 @@ properties:
               values:
                 - :TABLE_PROFILE
                 - :RESOURCE_NAME
+        - !ruby/object:Api::Type::NestedObject
+          name: tagResources
+          description: Publish a message into the Pub/Sub topic.
+          properties:
+            - !ruby/object:Api::Type::Array
+              name: tagConditions
+              description: The tags to associate with different conditions.
+              item_type: !ruby/object:Api::Type::NestedObject
+                properties:
+                  - !ruby/object:Api::Type::NestedObject
+                    name: tag
+                    description: The tag value to attach to resources.
+                    properties:
+                      - !ruby/object:Api::Type::String
+                        name: namespacedValue
+                        description: The namespaced name for the tag value to attach to resources. Must be in the format `{parent_id}/{tag_key_short_name}/{short_name}`, for example, "123456/environment/prod".
+                  - !ruby/object:Api::Type::NestedObject
+                    name: sensitivityScore
+                    description: Conditions attaching the tag to a resource on its profile having this sensitivity score.
+                    properties:
+                      - !ruby/object:Api::Type::Enum
+                        name: 'score'
+                        required: true
+                        description: |
+                          The sensitivity score applied to the resource.
+                        values:
+                          - :SENSITIVITY_LOW
+                          - :SENSITIVITY_MODERATE
+                          - :SENSITIVITY_HIGH
+            - !ruby/object:Api::Type::Array
+              name: profileGenerationsToTag
+              description: The profile generations for which the tag should be attached to resources. If you attach a tag to only new profiles, then if the sensitivity score of a profile subsequently changes, its tag doesn't change. By default, this field includes only new profiles. To include both new and updated profiles for tagging, this field should explicitly include both `PROFILE_GENERATION_NEW` and `PROFILE_GENERATION_UPDATE`.
+              item_type: !ruby/object:Api::Type::Enum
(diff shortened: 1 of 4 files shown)


mmv1/products/dlp/DiscoveryConfig.yaml
```diff
@@ -475,6 +529,17 @@ properties:
                     - :UPDATE_FREQUENCY_NEVER
                     - :UPDATE_FREQUENCY_DAILY
                     - :UPDATE_FREQUENCY_MONTHLY
+                - !ruby/object:Api::Type::NestedObject
+                  name: inspectTemplateModifiedCadence
+                  description: Governs when to update data profiles when the inspection rules defined by the `InspectTemplate` change. If not set, changing the template will not cause a data profile to update.
+                  properties:
+                    - !ruby/object:Api::Type::Enum
+                      name: 'frequency'
```
slevenick, line 537: This should probably be marked as required?

mmv1/products/dlp/DiscoveryConfig.yaml
```diff
@@ -346,6 +389,17 @@ properties:
                         - :UPDATE_FREQUENCY_NEVER
                         - :UPDATE_FREQUENCY_DAILY
                         - :UPDATE_FREQUENCY_MONTHLY
+                - !ruby/object:Api::Type::NestedObject
+                  name: inspectTemplateModifiedCadence
+                  description: Governs when to update data profiles when the inspection rules defined by the `InspectTemplate` change. If not set, changing the template will not cause a data profile to update.
+                  properties:
+                    - !ruby/object:Api::Type::Enum
+                      name: 'frequency'
```
slevenick, line 397: This should probably be marked as required

mmv1/third_party/terraform/services/datalossprevention/resource_data_loss_prevention_discovery_config_test.go
```diff
@@ -116,7 +116,10 @@ func testAccDataLossPreventionDiscoveryConfig_ActionsUpdate(t *testing.T) {
 	acctest.VcrTest(t, resource.TestCase{
 		PreCheck:                 func() { acctest.AccTestPreCheck(t) },
 		ProtoV5ProviderFactories: acctest.ProtoV5ProviderFactories(t),
-		CheckDestroy:             testAccCheckDataLossPreventionDiscoveryConfigDestroyProducer(t),
+		ExternalProviders: map[string]resource.ExternalProvider{
+			"time": {},
```
slevenick, line 120: Why do we need the time provider here?

mmv1/products/dlp/DiscoveryConfig.yaml
patrickmoy: Why should this be required? It's not required in the API.

mmv1/third_party/terraform/services/datalossprevention/resource_data_loss_prevention_discovery_config_test.go
patrickmoy: Oops, forgot to remove that; attempted to make a test work by adding a wait, but it turned out using "depends_on" was better. Removed.

mmv1/products/dlp/DiscoveryConfig.yaml
patrickmoy: We'll handle an empty value in the API as if it was set to never, but we won't actually return a server-side value.

slevenick: Generally we prefer to prevent users from specifying empty objects in Terraform. It can have some unexpected behaviors, so marking at least one sub-field as required prevents that.

If there's not a valid reason for a user to specify 
```
inspect_template_modified_cadence {}
```

Then it's easier to just mark the sub-field as required. It only becomes required in their config *if* they specify the inspect_template_modified_cadence field in the first place

patrickmoy: Got it. Added 'required: true' to the frequency value.

slevenick (approved):
Thanks!

PR #11510, opened 2024-08-21
Remove ForceNew: true in GcfsConfig

dominykasn:
Gcfsconfig changes do not require to re-create node pools hence there is no point to ForceNew on each gcfsconfig change, gcfs can be enabled and disabled via GUI or gcloud CLI without recreating node pools, examples:
https://cloud.google.com/kubernetes-engine/docs/how-to/image-streaming#disable

https://cloud.google.com/kubernetes-engine/docs/how-to/image-streaming#enable_on_node_pools

Also, same issue noted here https://github.com/hashicorp/terraform-provider-google/issues/18417

```release-note:enhancement
container: remove unnecessary force replacement in gcfs_config
```


dominykasn:
@NickElliot Hi, may I have a review on this?

PR #11514, opened 2024-08-21
container: add oauth scope to private registry test

wyardley:
Add oauth scopes in `node_config` for `TestAccContainerCluster_privateRegistry` (required to avoid an API level error).

Test only change; doesn't affect behavior

Without this change:

```
2024/08/21 10:45:48 [DEBUG] Retry Transport: Returning after 1 attempts
    vcr_utils.go:152: Step 1/7 error: Error running apply: exit status 1
        
        Error: googleapi: Error 400: Private_registry_access_config requires scope 'cloud-platform' in order to pull certificates. The following node-pools are lacking such scope: [default-pool].
        Details:
        [
          {
            "requestId": "0x9bda853dba747e59"
          }
        ]
        , badRequest
        
          with google_container_cluster.primary,
          on terraform_plugin_test.tf line 28, in resource "google_container_cluster" "primary":
          28: resource "google_container_cluster" "primary" {
        
--- FAIL: TestAccContainerCluster_privateRegistry (19.16s)
```

With it:
```
--- PASS: TestAccContainerCluster_privateRegistry (1375.74s)
PASS
ok  	github.com/hashicorp/terraform-provider-google/google/services/container	1376.747s
```

```release-note:none

```

### DIFF
--- a/mmv1/third_party/terraform/services/container/go/resource_container_cluster_test.go.tmpl
+++ b/mmv1/third_party/terraform/services/container/go/resource_container_cluster_test.go.tmpl
@@ -10972,6 +10972,11 @@ resource "google_container_cluster" "primary" {
   network    = "%s"
   subnetwork    = "%s"
 
+  node_config {
+    oauth_scopes = [
+      "https://www.googleapis.com/auth/cloud-platform",
+    ]
+  }
   node_pool_defaults {
     node_config_defaults {
       containerd_config {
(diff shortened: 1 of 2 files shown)


slevenick (approved):
LGTM thanks! I can't merge this just yet, but should be able to soon

wyardley:
@slevenick thanks. Looks like the 6.x stuff shipped.
I won't rebase this unless you ask, but let me know if you want me to.

wyardley:
@slevenick what's odd is that it seems like this test wasn't consistently failing even without this change? So not sure if this was an API level brownout for some sort of planned change, or a bug at the API level that got fixed, or what. But either way, I imagine the updated test is safe enough to leave as is?

slevenick:
Yeah, it seems safe to make

PR #11515, opened 2024-08-21
Add initial support for diffing conflicts

trodge:
Helps with https://github.com/hashicorp/terraform-provider-google/issues/18902

```release-note:none
Additional data for comparing field conflicts in the diff processor
```


tools/diff-processor/diff/conflicts.go
```diff
@@ -0,0 +1,265 @@
...
+	ConflictsWiths *SetOfConflictSets
+	ExactlyOneOfs  *SetOfConflictSets
+	AtLeastOneOfs  *SetOfConflictSets
+	RequiredWiths  *SetOfConflictSets
+	Key            string // sorted string representation of all sets, used for quick comparison between fields
+}
+
+type FieldConflictSetsDiff struct {
+	Old *FieldConflictSets
+	New *FieldConflictSets
```
trodge, line 42:
```suggestion
	New *FieldConflictSets
	// TODO: add methods of filling Added, Removed, Expanded, and Shrank to be called after Merge
	Added *FieldConflictSets // Sets of sets that are only in New, and have no subsets in Old
	Removed *FieldConflictSets // Sets of sets that are only in Old, and have no subsets in New
	Expanded *FieldConflictSets // Sets of added fields to existing sets in Old
	Shrank *FieldConflictSets // Sets of removed fields from existing sets in Old 
```

melinath:
I think the automation detected a comment from a core team member even though it's from the PR author

melinath (changes requested):
Overall I think this ends up being a bit confusing because of the amount of nesting. Although it may prevent some duplication of code, I think it ends up being harder to read, especially because the type & func names end up being long and similar. Additionally, "conflicts" as a name for these sets is a little inaccurate, since only two of the four types of sets involve conflicting.

I'd recommend something like this:

```go
type ResourceDiff struct {
	ResourceConfig ResourceConfigDiff
	Fields         map[string]FieldDiff
	Conflicts      FieldSetDiff
	ExactlyOneOf   FieldSetDiff
	AtLeastOneOf   FieldSetDiff
	RequiredWith   FieldSetDiff
}

// Slices of FieldSets for the old & new versions of a resource
type FieldSetDiff struct {
	Old []map[string]struct{}
	New []map[string]struct{}
}
```

A couple things to note:

- Only stores the field sets on the resource level diff, because they're not meaningful/useful at the field-level. (If you need to do something at the field level, you could also always process the field-level attributes directly).
- Doesn't store keys for each FieldSet, because we only need the keys for deduplicating the fieldsets - meaningful analysis (like checking for breaking changes) will require comparing the actual sets
- This relies on the constraint that within a given type of field set, all sets should be non-overlapping.
  - It might make sense to add in detection of overlapping sets and log them. Longer-term, it should maybe be raised as an error, but we don't have a mechanism for that and it would likely be out of scope of this project.
- Let's leave out Added / Removed / Expanded / Shrank for now. It will keep the differ code (and this PR!) simpler. We can always add them later if we find out we need that breakdown for multiple purposes in consuming code and if the computation ends up being very expensive.

What would you think of taking an approach like this? Happy to discuss offline as well.

trodge:
1. Agree that "conflicts" is a slightly confusing name, but it can be argued that all four of `AtLeastOneOf` `RequiredWith` `ExactlyOneOf` and `Conflicts` are types of conflicts in that they all restrict how different fields can be set relative to each other.

1. I think only having `Old` and `New` slices will not allow us to do meaningful analysis, because each set will need to be compared with every other set. It's possible there could be some readability benefit to doing this separately from `diffFields`, but the benefit of doing all of it here is that we don't need to do additional work to detect if any of the conflict sets of a field have changed.

1. The main benefit to using sets all the way down is being 100% sure that we don't have duplicate fields within a set of fields, or duplicate sets of fields within a resource without needing to iterate all the sets at every step.

1. Checking for breaking changes will require analyzing the sets themselves, but the keys will still be useful in identifying which sets to compare. Once we have `Added` `Removed` `Expanded` and `Shrank`, we will be able to write rules looking at, for example, `Added` or `Expanded` `ExactlyOneOf` sets.

1. I'm not sure how we would detect breaking changes without a detailed view of *how* the conflict sets are changing. That being said, it might be better to put this in a different subpackage of the diff processor so that the interface is more separate from the implementation.

melinath:
> 1. Agree that "conflicts" is a slightly confusing name, but it can be argued that all four of `AtLeastOneOf` `RequiredWith` `ExactlyOneOf` and `Conflicts` are types of conflicts in that they all restrict how different fields can be set relative to each other.

I think that's a bit of a stretch - "conflicting" and "restricting" are not really the same.

> 2. I think only having `Old` and `New` slices will not allow us to do meaningful analysis, because each set will need to be compared with every other set. It's possible there could be some readability benefit to doing this separately from `diffFields`, but the benefit of doing all of it here is that we don't need to do additional work to detect if any of the conflict sets of a field have changed.

I believe that constructing an accurate list of sets that were added, removed, expanded, or shrunk requires comparing every set with every other set regardless - whether we do it here or in the breaking change detector. The benefit of doing it here would be that we could reuse the computation (but that's only relevant if we have multiple things that _need_ the computation.)

> 3. The main benefit to using sets all the way down is being 100% sure that we don't have duplicate fields within a set of fields, or duplicate sets of fields within a resource without needing to iterate all the sets at every step.

I don't think I understand this argument. If we construct the top-level sets as we iterate over fields, how could we end up missing fields? We do need to deduplicate fieldsets, but we don't need to store the keys _past_ the deduplication (because they aren't useful for comparison.)

> 4. Checking for breaking changes will require analyzing the sets themselves, but the keys will still be useful in identifying which sets to compare. Once we have `Added` `Removed` `Expanded` and `Shrank`, we will be able to write rules looking at, for example, `Added` or `Expanded` `ExactlyOneOf` sets.

I'm not convinced that the keys are useful for comparison. For example:

- Initial set is field1, field3. field2 is added. The key can't be used to determine that a field was added.
- Initial set is field1, field2, field3. field2 is removed. The key can't be used to determine that field2 was removed.

Hypothetically, we could try to use information about what fields in the set were added / removed (as part of the field diffing process) to build keys for comparison, but that has limitations. For example,

- Initial set is field1, field2. field3 and field4 are added. Neither new field can find the "old" version of the set because it changed by _two_ keys.
- Same thing happens if both are removed, or if one is added and one removed.

I believe that trying to fix those limitations will end up adding additional complexity and eventually just end up comparing all the sets with each other (but in a less clear way). And I don't think that we have evidence that the set comparison is prohibitively slow at this time. (And if it were, I think it would be easier to approach the optimization as a problem of efficiently comparing sets rather than trying to avoid comparing sets.)

> 5. I'm not sure how we would detect breaking changes without a detailed view of _how_ the conflict sets are changing. That being said, it might be better to put this in a different subpackage of the diff processor so that the interface is more separate from the implementation.

My proposal is that the breaking change detection would use the old / new fieldsets to build a detailed view of how the conflict sets are changing. We could also do it here. I would have a slight preference to put it in the breaking change detection code first, since we only have that one use case so far, but I could be convinced to put it in the differ instead.

---

Let me know what you think!

trodge:
I think you're right that this PR is potentially a premature optimization. I'll work on a simpler, more brute force solution when I have time.

PR #11516, opened 2024-08-21
feat(containerattached): Add security_posture_config

hdp617:
Adds support for configuring Security Posture.

```release-note:enhancement
containerattached: added `security_posture_config` field to `google_container_attached_cluster` resource
```


shuyama1 (commented):
Both tests failed with error:
```
vcr_utils.go:152: Step 1/10 error: Error running apply: exit status 1
        
        Error: Error creating Cluster: googleapi: Error 400: Invalid Security Posture config. Security Posture is not supported in this version., field: attached_cluster.security_posture_config
        Details:
        [
          {
            "fieldViolations": [
              {
                "description": "Invalid Security Posture config. Security Posture is not supported in this version.",
                "field": "attached_cluster.security_posture_config"
              }
            ]
          }
        ]
        
          with google_container_attached_cluster.primary,
          on terraform_plugin_test.tf line 10, in resource "google_container_attached_cluster" "primary":
          10: resource "google_container_attached_cluster" "primary" {
```
Are you able to get tests passed locally? Let me know if you need help with debugging the issue. Thanks!

hdp617:
> Are you able to get tests passed locally? Let me know if you need help with debugging the issue. Thanks!

This is expected as the version that supports the feature is not fully rolled out yet. I'll rerun the test and let you know once it's ready. Thanks!

mmv1/products/containerattached/Cluster.yaml
```diff
@@ -358,3 +358,17 @@ properties:
             required: true
             description: |
               Namespace of the kubernetes secret containing the proxy config.
+  - !ruby/object:Api::Type::NestedObject
+    name: securityPostureConfig
+    description: |
+      Enable/Disable Security Posture API features for the cluster.
+    allow_empty_object: true
```
shuyama1, line 365: Does explicitly sending an empty security_posture_config {} block to the API have any specific implications? In general, we'd avoid allowing Terraform users to send empty blocks.

mmv1/third_party/terraform/services/containerattached/resource_container_attached_cluster_update_test.go
```diff
@@ -118,6 +118,9 @@ resource "google_container_attached_cluster" "primary" {
       namespace = "default"
     }
   }
+  security_posture_config {
```
shuyama1, line 121: qq: Is it necessary to include this block in every configuration? Including it during creation and updating a value once in an update step should be sufficient for testing. We may also need to exclude it from certain tests to ensure that omitting the fields doesn’t cause any issues.

mmv1/products/containerattached/Cluster.yaml
hdp617: > Does explicitly sending an empty security_posture_config {} block to the API have any specific implications?

No, it doesn't. I was following the convention for this resource as other fields also have `allow_empty_object` set.

mmv1/third_party/terraform/services/containerattached/resource_container_attached_cluster_update_test.go
hdp617: Sounds good. I've removed this block from the other tests.

mmv1/products/containerattached/Cluster.yaml
shuyama1: Thanks for confirming! Then I'd suggest that we remove the `allow_empty_object: true` and mark the only subfield `vulnerabilityMode` as required to avoid allowing users to send empty blocks. Thanks!

shuyama1 (commented):
I'll convert the PR to draft for now to disable reviewer reminders. Feel free to convert it back when the backend is released and the PR is ready for review. Thanks!

shuyama1 (approved):

PR #11524, opened 2024-08-22
Add new compute-network-firewall-policy-with-rules resource

mihhalj:
Add a new resource to allow updating all rules in network firewall policy at once.

Fixes 
https://github.com/hashicorp/terraform-provider-google/issues/19195

```release-note:new-resource
`google_compute_network_firewall_policy_with_rules` (beta)
```


mihhalj:
Could I ask fro review?
I ran my tests locally and they pass - not sure why TestReadPlannedAssetsCoverage* tests and VCR-test are failing.

mmv1/templates/terraform/constants/resource_compute_network_firewall_policy_with_rules.go.erb
```diff
@@ -0,0 +1,54 @@
...
+
+  if err != nil {
+      return false, err
+    }
+
+  return priorityInt >= ReservedPriorityStart, nil
+
+}
+
+func  networkFirewallPolicyWithRulesSplitPredefinedRules(allRules []interface{}) ([]interface{}, []interface{}, error) {
```
NickElliot, line 35:
```suggestion
func networkFirewallPolicyWithRulesSplitPredefinedRules(allRules []interface{}) ([]interface{}, []interface{}, error) {
```

mmv1/templates/terraform/decoders/resource_compute_network_firewall_policy_with_rules.go.erb
```diff
@@ -0,0 +1,16 @@
...
+res["rules"] = rules
+res["predefinedRules"] = predefinedRules
+
+config := meta.(*transport_tpg.Config)
+
+if err := d.Set("predefined_rules", flattenComputeNetworkFirewallPolicyWithRulesPredefinedRules(predefinedRules, d, config)); err != nil {
+	return nil, fmt.Errorf("Error occurred while setting pre-defined rules: %s", err)
+}
+
+return res, nil
```
NickElliot, line 16:
```suggestion
return res, nil

```

mmv1/products/compute/NetworkFirewallPolicyWithRules.yaml
```diff
@@ -0,0 +1,570 @@
...
+      security_profile_name: 'tf-security-profile'
+    test_env_vars:
+      org_id: :ORG_ID
+custom_code: !ruby/object:Provider::Terraform::CustomCode
+  constants: templates/terraform/constants/resource_compute_network_firewall_policy_with_rules.go.erb
+  encoder: templates/terraform/encoders/resource_compute_network_firewall_policy_with_rules.go.erb
+  decoder: templates/terraform/decoders/resource_compute_network_firewall_policy_with_rules.go.erb
+  update_encoder: templates/terraform/update_encoder/resource_compute_network_firewall_policy_with_rules.go.erb
+  post_create: templates/terraform/post_create/resource_compute_network_firewall_policy_with_rules.go.erb
+
```
NickElliot, line 60:
```suggestion
```

NickElliot (changes requested):
hi, a couple formatting changes needed

could you link me some of the public documentation on the resource?

mihhalj:
> hi, a couple formatting changes needed
> 
> could you link me some of the public documentation on the resource?

Docs:
https://cloud.google.com/firewall/docs/network-firewall-policies
https://cloud.google.com/compute/docs/reference/rest/v1/networkFirewallPolicies

For firewall policy there's a terraform resource (not possible to set rules):
https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/compute_network_firewall_policy

And the rules in firewall policy can be configured using:
https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/compute_region_network_firewall_policy_rule

The new google_compute_network_firewall_policy_with_rules would allow a user to configure firewall policy with all rules (it was suggested by [rileykarson](https://github.com/rileykarson) to create a new resource to handle it).

NickElliot (approved):
LGTM

joekohlsdorf:
@mihhalj I have experimented with the new resource a little bit, it isn't great.

Adding a new rule between existing rules yields a plan which recreates all the rules following the new addition. The same happens when deleting a rule when multiple rules exist.
I haven't been able to verify it but when applying this it might even temporarily remove existing firewall while things are reshuffled rules causing outages.

mihhalj:
> @mihhalj I have experimented with the new resource a little bit, it isn't great.
> 
> Adding a new rule between existing rules yields a plan which recreates all the rules following the new addition. The same happens when deleting a rule when multiple rules exist. I haven't been able to verify it but when applying this it might even temporarily remove existing firewall while rules are reshuffled causing outages.

[joekohlsdorf](https://github.com/joekohlsdorf) Thanks for testing and You're right - using this resource when you update/remove/add even only one rule it results in updating all rules - the main goal for this resource is to be able to change many rules at once - it either succeeds and all the rules are updated or it fails and none of the rules are not updated - for example if one of the rule is invalid then the firewall policy won't be updated (there won't be a situation where valid rules are updated and invalid not).

Compare it to the case when you use [compute_network_firewall_policy_rule](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/compute_network_firewall_policy_rule) resources for your rules - valid rules will be configured and the invalid not.

PR #11527, opened 2024-08-22
Add disk resource policies field to the compute instance

karolgorc:
closes https://github.com/hashicorp/terraform-provider-google/issues/8963
closes https://github.com/hashicorp/terraform-provider-google/issues/9260

This adds boot_disk -> initialize_params -> resource_policies

- Added the field into `google_compute_instance`
- Added data_source support
- Added tests for the new field
- Doc changes
- Added translation in `instance_from_template` that will support this change and resolve [this issue](https://github.com/hashicorp/terraform-provider-google/issues/9260)

If there are any more files that i need to update please LMK

```release-note:enhancement
compute: added support for `boot_disk.initialize_params.resource_policies` in `google_compute_instance` and `google_instance_template`
```

### DIFF
--- a/mmv1/third_party/terraform/services/compute/resource_compute_instance.go.erb
+++ b/mmv1/third_party/terraform/services/compute/resource_compute_instance.go.erb
@@ -74,6 +74,7 @@ var (
 		"boot_disk.0.initialize_params.0.provisioned_throughput",
 		"boot_disk.0.initialize_params.0.enable_confidential_compute",
 		"boot_disk.0.initialize_params.0.storage_pool",
+		"boot_disk.0.initialize_params.0.resource_policies",
 	}
 
 	schedulingKeys = []string{
--- a/mmv1/third_party/terraform/services/compute/resource_compute_instance.go.erb
+++ b/mmv1/third_party/terraform/services/compute/resource_compute_instance.go.erb
@@ -280,6 +280,17 @@ func ResourceComputeInstance() *schema.Resource {
 										Description:  `A map of resource manager tags. Resource manager tag keys and values have the same definition as resource manager tags. Keys must be in the format tagKeys/{tag_key_id}, and values are in the format tagValues/456. The field is ignored (both PUT & PATCH) when empty.`,
 									},
 
+									"resource_policies": {
+										Type:             schema.TypeList,
+										Elem:             &schema.Schema{Type: schema.TypeString},
+										Optional:         true,
+										ForceNew:         true,
+										AtLeastOneOf:     initializeParamsKeys,
+										DiffSuppressFunc: tpgresource.CompareSelfLinkRelativePaths,
+										MaxItems:         1,
+										Description:      `A list of self_links of resource policies to attach to the instance's boot disk. Modifying this list will cause the instance to recreate. Currently a max of 1 resource policy is supported.`,
+									},
+
 									"provisioned_iops": {
 										Type:         schema.TypeInt,
 										Optional:     true,
@@ -3018,6 +3029,10 @@ func expandBootDisk(d *schema.ResourceData, config *transport_tpg.Config, projec
 			disk.InitializeParams.ResourceManagerTags = tpgresource.ExpandStringMap(d, "boot_disk.0.initialize_params.0.resource_manager_tags")
 		}
 
+		if _, ok := d.GetOk("boot_disk.0.initialize_params.0.resource_policies"); ok {
+			disk.InitializeParams.ResourcePolicies = tpgresource.ConvertStringArr(d.Get("boot_disk.0.initialize_params.0.resource_policies").([]interface{}))
+		}
+
 		if v, ok := d.GetOk("boot_disk.0.initialize_params.0.storage_pool"); ok {
 			disk.InitializeParams.StoragePool = v.(string)
 		}
@@ -3063,6 +3078,7 @@ func flattenBootDisk(d *schema.ResourceData, disk *compute.AttachedDisk, config 
 			"size":   diskDetails.SizeGb,
 			"labels": diskDetails.Labels,
 			"resource_manager_tags": d.Get("boot_disk.0.initialize_params.0.resource_manager_tags"),
+			"resource_policies": diskDetails.ResourcePolicies,
 			"provisioned_iops": diskDetails.ProvisionedIops,
 			"provisioned_throughput": diskDetails.ProvisionedThroughput,
 			"enable_confidential_compute": diskDetails.EnableConfidentialCompute,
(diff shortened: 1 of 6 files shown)


mmv1/third_party/terraform/services/compute/resource_compute_instance_from_template.go.erb
```diff
@@ -262,6 +262,15 @@ func adjustInstanceFromTemplateDisks(d *schema.ResourceData, config *transport_t
 						// only have the name (since they're global).
 						disk.InitializeParams.DiskType = fmt.Sprintf("zones/%s/diskTypes/%s", zone.Name, dt)
 					}
+					if rp := disk.InitializeParams.ResourcePolicies; len(rp) > 0 {
```
c2thorn, line 265: do we have a test for this logic?

c2thorn (changes requested):
a few questions

mmv1/third_party/terraform/services/compute/resource_compute_instance_from_template.go.erb
karolgorc: currently working on acctests for google_compute_instance_from_template and google_compute_instance_from_machine_image that were requested by the modular-magician. It'll be tested then

mmv1/third_party/terraform/services/compute/resource_compute_instance.go.erb
```diff
@@ -273,6 +273,17 @@ func ResourceComputeInstance() *schema.Resource {
 										Description:  `A map of resource manager tags. Resource manager tag keys and values have the same definition as resource manager tags. Keys must be in the format tagKeys/{tag_key_id}, and values are in the format tagValues/456. The field is ignored (both PUT & PATCH) when empty.`,
 									},
 
+									"resource_policies": {
+										Type:             schema.TypeList,
+										Elem:             &schema.Schema{Type: schema.TypeString},
+										Optional:         true,
+										ForceNew:         true,
+										AtLeastOneOf:     initializeParamsKeys,
```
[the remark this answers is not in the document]
karolgorc: agreed, will add that in

mmv1/third_party/terraform/services/compute/resource_compute_instance.go.erb
```diff
@@ -273,6 +273,17 @@ func ResourceComputeInstance() *schema.Resource {
...
+										DiffSuppressFunc: tpgresource.CompareSelfLinkRelativePaths,
```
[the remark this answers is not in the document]
karolgorc: yup, this does work on lists. done a small test where i had 2 local variables like this
```
locals {
  relative = "projects/iac-poc-krakow/regions/us-central1/resourcePolicies/test-policy"
  self_link = "https://www.googleapis.com/compute/v1/${local.relative}"
}
```

And commented the DiffSuppressFunc in code. Here is the result
image

After uncommenting the diff goes away

karolgorc:
- Added testing for the translation from `google_compute_instance_template` to `google_compute_instance_from_template`
- Added new field to `initializeParamsKeys`

There is no way to use this in `google_compute_instance_from_machine_image` because it uses the `SavedAttachedDisk` object which doesn't have `InitializeParams` saved anywhere. Only way to do that on this resource is through overrides but they are currently disabled. Or another solution are the specific objects made for that like `google_compute_disk_resource_policy_attachment`.

The @modular-magician asks for testing the overrides which are currently disabled
> *Due to API limitations, disk overrides are currently disabled. This includes the "boot_disk", "attached_disk", and "scratch_disk" fields.

So i guess the new field is supported only on instances and templates for now. Do we need any doc changes reflecting that?

mmv1/third_party/terraform/services/compute/resource_compute_instance.go.erb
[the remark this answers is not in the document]
c2thorn: thanks for confirming, that was unexpected!

mmv1/third_party/terraform/website/docs/d/compute_instance.html.markdown
```diff
@@ -131,6 +131,8 @@ The following arguments are supported:
 
 * `labels` - A set of key/value label pairs assigned to the disk.
 
+* `resource_policies` - A self_link of a resource policy attached to the disk
```
c2thorn, line 134: this would also be a list right?

c2thorn (commented):
> So i guess the new field is supported only on instances and templates for now. Do we need any doc changes reflecting that?

I don't see a reason not to briefly mention it at the same time we mention the other fields are disabled - https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/third_party/terraform/website/docs/r/compute_instance_from_machine_image.html.markdown?plain=1#L59

mmv1/third_party/terraform/website/docs/d/compute_instance.html.markdown
karolgorc: yup, changed that in the docs

karolgorc:
>I don't see a reason not to briefly mention it at the same time we mention the other fields are disabled
I think it's self-explanatory here because the new field is under `boot_disk` which is mentioned in the note.

karolgorc:
Done, the rebase. The tests requested here are already added
> do we have a test for this logic?

karolgorc:
don't think these 2 are related 
```
resource "google_compute_instance_from_template" "primary" {
  boot_disk {
    initialize_params {
      enable_confidential_compute = # value needed
      image                       = # value needed
      labels                      = # value needed
      provisioned_iops            = # value needed
      provisioned_throughput      = # value needed
      resource_manager_tags       = # value needed
      resource_policies           = # Except for this but cannot test because overrides are off
      size                        = # value needed
      storage_pool                = # value needed
      type                        = # value needed
    }
  }
}
```

```
resource "google_compute_instance" "primary" {
  boot_disk {
    initialize_params {
      labels = # value needed
    }
  }
}
```
And as stated earlier this cannot be tested because of the disabled overrides
```
resource "google_compute_instance_from_machine_image" "primary" {
  boot_disk {
    initialize_params {
      resource_policies = # value needed
    }
  }
}

```

c2thorn:
> don't think these 2 are related
> 
> ```
> resource "google_compute_instance_from_template" "primary" {
>   boot_disk {
>     initialize_params {
>       enable_confidential_compute = # value needed
>       image                       = # value needed
>       labels                      = # value needed
>       provisioned_iops            = # value needed
>       provisioned_throughput      = # value needed
>       resource_manager_tags       = # value needed
>       resource_policies           = # Except for this but cannot test because overrides are off
>       size                        = # value needed
>       storage_pool                = # value needed
>       type                        = # value needed
>     }
>   }
> }
> ```
> 
> ```
> resource "google_compute_instance" "primary" {
>   boot_disk {
>     initialize_params {
>       labels = # value needed
>     }
>   }
> }
> ```
> 
> And as stated earlier this cannot be tested because of the disabled overrides
> 
> ```
> resource "google_compute_instance_from_machine_image" "primary" {
>   boot_disk {
>     initialize_params {
>       resource_policies = # value needed
>     }
>   }
> }
> ```
> 
> Should i write these override tests and skip them in VCR?

No I think we should be good to go as the primary tests have passed for Instance and Instance template. Thanks!

c2thorn (approved):
LGTM

PR #11528, opened 2024-08-22
Fix documentation for data rentention configuration

Khaledmohamedrefaat:

```release-note:none

```

### DIFF
--- a/mmv1/third_party/terraform/website/docs/r/composer_environment.html.markdown
+++ b/mmv1/third_party/terraform/website/docs/r/composer_environment.html.markdown
@@ -879,6 +879,24 @@ The following arguments are supported:
   for high resilience and `STANDARD_RESILIENCE` for standard
   resilience.
 
+* `data_retention_config` -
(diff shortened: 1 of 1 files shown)


mmv1/third_party/terraform/website/docs/r/composer_environment.html.markdown
```diff
@@ -905,24 +923,6 @@ The following arguments are supported:
   (Required)
   `cidr_block` must be specified in CIDR notation.
 
-* `data_retention_config` -
```
melinath, line 908: why is this block moved?

Khaledmohamedrefaat: For Composer 2, It was mistakenly intended under cidr_blocks while it is part of config: https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/third_party/terraform/services/composer/resource_composer_environment.go.erb#L736

melinath (approved):

PR #11503, opened 2024-08-20
feat: support nfs and gcs in cloudrun and cloudrunv2 GA provider

bskaplan:
Move nfs, and gcs/csi to GA provider in cloudrun and cloudrunv2

**Release Note Template for Downstream PRs (will be copied)**

```release-note:enhancement
cloudrun: added support for nfs and csi volumes (for cloud storage fuse) for google_cloud_run_service (ga)
```

```release-note:enhancement
cloudrunv2: added support for nfs and gcs volumes for google_cloud_run_v2_job (ga)
```

PR #11531, opened 2024-08-22
Fix test TestAccCloudRunService_csiVolume

ScottSuarez:
This test is currently failing due to diff on import with an OUTPUT only field. Suppressing the diff here.

https://hashicorp.teamcity.com/buildConfiguration/TerraformProviders_GoogleCloud_GOOGLE_BETA_NIGHTLYTESTS_GOOGLEBETA_PACKAGE_CLOUDRUN/222026?hideTestsFromDependencies=false&hideProblemsFromDependencies=false&expandBuildDeploymentsSection=false&expandBuildTestsSection=true&expandBuildChangesSection=true

```release-note:none

```

### DIFF
--- a/mmv1/third_party/terraform/services/cloudrun/resource_cloud_run_service_test.go.erb
+++ b/mmv1/third_party/terraform/services/cloudrun/resource_cloud_run_service_test.go.erb
@@ -1333,7 +1333,7 @@ func TestAccCloudRunService_csiVolume(t *testing.T) {
 				ResourceName:            "google_cloud_run_service.default",
 				ImportState:             true,
 				ImportStateVerify:       true,
-				ImportStateVerifyIgnore: []string{"metadata.0.resource_version", "metadata.0.annotations", "metadata.0.labels", "metadata.0.terraform_labels", "status.0.conditions"},
+				ImportStateVerifyIgnore: []string{"metadata.0.resource_version", "metadata.0.annotations", "metadata.0.effective_annotations", "metadata.0.labels", "metadata.0.terraform_labels", "status.0.conditions"},
 			},
 			{
 				Config: testAccCloudRunService_cloudRunServiceUpdateWithGcsVolume(name, project,),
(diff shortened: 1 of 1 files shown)


rileykarson (commented):
The failing test you linked has a different kind of error than this addresses, ImportStateVerify affects the Import step but the failure is on a Config step. `effective_annotations` is a weird field in terms of diffs- fyi @zli82016 in case you recognise the issue.

(re: VCR success, the test is skipped in VCR)

zli82016:
The reason for the failure is that the `run.googleapis.com/launch-stage = BETA` exists in the configuration, but is not returned in the response by API because no beta feature is used.  We can try to remove it from the configuration of test [testAccCloudRunService_cloudRunServiceUpdateWithGcsVolume](https://github.com/GoogleCloudPlatform/magic-modules/blob/ea524af13b78b637072d1b05c165f0b641cf7574/mmv1/third_party/terraform/services/cloudrun/resource_cloud_run_service_test.go.erb#L1403).

zli82016:
I tested locally by removing `run.googleapis.com/launch-stage = BETA` from [testAccCloudRunService_cloudRunServiceUpdateWithGcsVolume](https://github.com/GoogleCloudPlatform/magic-modules/blob/ea524af13b78b637072d1b05c165f0b641cf7574/mmv1/third_party/terraform/services/cloudrun/resource_cloud_run_service_test.go.erb#L1403) and the test still failed.

Let me check if it is a bug in the annotations model.

zli82016:
Oh,  besides to remove 1run.googleapis.com/launch-stage = BETA1 from testAccCloudRunService_cloudRunServiceUpdateWithGcsVolume config, `ignore_changes` also needs to be removed from the test config to apply the changes in annotations.

zli82016:
The tests are modified in the PR https://github.com/GoogleCloudPlatform/magic-modules/pull/11503

PR #11537, opened 2024-08-23
Change doc for `google_compute_instance` to reflect the correct behaviour of `subnetwork_project` field

karolgorc:
closes

The `subnetwork_project` field is ignored in the API request when a self_link is given in `subnetwork`, but not in terraform which is correct behavior because `subnetwork_project` is still in the code. Suppressing a diff on this seems unreasonable because providing something like this isn't really a valid config
```
subnetwork_project = "projectA"
subnetwork = "...projectB..."
```

Maybe throw an error when this happens?

```release-note:note
compute: change doc for `google_compute_instance` `subnetwork_project`
```

### DIFF
--- a/mmv1/third_party/terraform/website/docs/r/compute_instance.html.markdown
+++ b/mmv1/third_party/terraform/website/docs/r/compute_instance.html.markdown
@@ -321,7 +321,7 @@ is desired, you will need to modify your state file manually using
 
 
 *  `subnetwork_project` - (Optional) The project in which the subnetwork belongs.
-   If the `subnetwork` is a self_link, this field is ignored in favor of the project
+   If the `subnetwork` is a self_link, this field is set to the project
    defined in the subnetwork self_link. If the `subnetwork` is a name and this
    field is not provided, the provider project is used.
 


karolgorc:
that unit-test fail doesn't seem related

zli82016:
> The `subnetwork_project` field is ignored in the API request when a self_link is given in `subnetwork`, but not in terraform which is correct behavior because `subnetwork_project` is still in the code. Suppressing a diff on this seems unreasonable because providing something like this isn't really a valid config
> 
> ```
> subnetwork_project = "projectA"
> subnetwork = "...projectB..."
> ```
> 
> Maybe throw an error when this happens?

Throwing an error is a good idea to let the user know that the provided subnetwork_project is not correct.

zli82016 (changes requested):
Can you please add the validation? Thanks.

karolgorc:
- Added GetProjectFromRegionalSelfLink to tpgresource and a unit test for it
- Added the new logic to _CustomizeDiff_

This isn't really a simple field validation to use _ValidateFunc_ because it depends on an external value which isn't really supported by the types used in _ValidateFunc_, _StateFunc_ or _DiffSuppressFunc_.

image

karolgorc:
Got the same test fails in 2 other PR's. Don't think they are related.

zli82016:
Thank you for adding the validation. 

Do you mind adding an acceptance test to check if the incorrect configuration for subnetwork_project will returns error in https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/third_party/terraform/services/compute/resource_compute_instance_test.go.erb?

Example: https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/third_party/terraform/services/alloydb/resource_alloydb_cluster_restore_test.go#L39-L42

zli82016 (approved):
LGTM. Thanks for adding the validation. The failed tests are unrelated.

PR #11540, opened 2024-08-23
Made virtual fields get generated with the same template as non-virtual fields

melinath:
This allows users to use more complex field types as virtual fields.

Fixes https://github.com/hashicorp/terraform-provider-google/issues/19245. See yaqs/986109097998811136 for additional context.

Note that I had to remove "Enum" status from all virtual fields, because they were not actually being honored previously (and adding new validatefuncs would be a breaking change.)

```release-note:none

```

### DIFF
--- a/mmv1/api/resource.go
+++ b/mmv1/api/resource.go
@@ -329,6 +329,12 @@ func (r *Resource) UnmarshalYAML(n *yaml.Node) error {
 		r.CollectionUrlKey = google.Camelize(google.Plural(r.Name), "lower")
 	}
 
+	if len(r.VirtualFields) > 0 {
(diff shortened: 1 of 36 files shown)


melinath:
Remaining VCR failure looks flakey; rerunning. https://hashicorp.teamcity.com/test/8506985094336897859?currentProjectId=TerraformProviders_GoogleCloud_GOOGLE_BETA_NIGHTLYTESTS&expandTestHistoryChartSection=true&expandedTest=build%3A%28id%3A224218%29%2Cid%3A2000000201

c2thorn (approved):
Thanks for making the Go changes here too. Looks good

PR #11552, opened 2024-08-27
Feat: Add new files for datasources google_bigquery_tables

anubhavsharma515:
Fixes https://github.com/hashicorp/terraform-provider-google/issues/17695

```release-note:new-datasource
`google_bigquery_tables`
```


anubhavsharma515:
G'day @SarahFrench! This is my first time working with Go and on trying to run
```
make testacc TEST=./google/services/bigquery TESTARGS='-run=TestAccDataSourceGoogleBigqueryTables_basic'
```,

I keep running into the following error: 
```
package github.com/hashicorp/terraform-provider-google/google/services/bigquery_test
	imports github.com/hashicorp/terraform-provider-google/google/acctest
	imports github.com/hashicorp/terraform-provider-google/google/provider
	imports github.com/hashicorp/terraform-provider-google/google/services/apigee
	imports github.com/hashicorp/terraform-provider-google/google/acctest: import cycle not allowed
FAIL	github.com/hashicorp/terraform-provider-google/google/services/bigquery [setup failed]
FAIL
make: *** [testacc] Error 1
```

Was hoping to request your advice on how to resolve this. I've essentially tried to mimic the settings from https://github.com/GoogleCloudPlatform/magic-modules/pull/9352 and can't pin down where the `import cycle` is coming from.

Many thanks in advance for any help!

**Review by SarahFrench — COMMENTED** [review]
Hi, could you please sign the CLA (https://github.com/GoogleCloudPlatform/magic-modules/pull/11552#issuecomment-2311609874) and then I can start reviewing, thanks!

**Comment by anubhavsharma515** [comment]
Hi @SarahFrench! Apologies for not pushing refresh commits after signing the CLA - I think it should be good to go now?

**Comment by SarahFrench** [comment]
Thanks!

File: mmv1/third_party/terraform/services/bigquery/data_source_google_bigquery_tables.go [hunk 1]
```diff
@@ -0,0 +1,86 @@
+// Copyright (c) HashiCorp, Inc.
+// SPDX-License-Identifier: MPL-2.0
+
```
SarahFrench, line 3:
```suggestion
```

mmv1/third_party/terraform/services/bigquery/data_source_google_bigquery_tables_test.go
```diff
@@ -0,0 +1,66 @@
+// Copyright (c) HashiCorp, Inc.
+// SPDX-License-Identifier: MPL-2.0
+
```
SarahFrench, line 3:
```suggestion
```

mmv1/third_party/terraform/services/bigquery/data_source_google_bigquery_tables.go
```diff
@@ -0,0 +1,86 @@
...
+      Type:        schema.TypeString,
+      Optional:    true,
+      Description: "The ID of the project in which the dataset is located. If it is not provided, the provider project is used.",
+    }, 
+    "tables": { 
+      Type:        schema.TypeList,
+      Computed:    true,
+      Elem:        &schema.Schema{Type: schema.TypeString},
+      Description: "A list of table names in the dataset.",
+    },
```
SarahFrench, line 37: Could the data source contain more information about the tables beyond their names?

One pattern we use in the provider currently is to take the schema from a resource and re-use it when implementing data sources, so that data sources' schemas don't repeat information unnecessarily. [Here's an example](https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/third_party/terraform/services/kms/data_source_google_kms_crypto_keys.go#L15-L17) where the schema is obtained from a resource (you would use [ResourceBigQueryTable](https://github.com/hashicorp/terraform-provider-google/blob/7a4da3410a94796cd12d05f12a77b793888c7e8f/google/services/bigquery/resource_bigquery_table.go#L436) instead) and is altered to make sure any Required fields are changed to optional.

There might be some complications with this but wanted to raise the idea in case you weren't aware this was possible!

SarahFrench (commented):
Thanks for your PR! There are some comments that can be removed, and I've asked if we can get richer info about each table returned.

mmv1/third_party/terraform/services/bigquery/data_source_google_bigquery_tables.go
anubhavsharma515: Hey Sarah! It definitely can. This is honestly one of my first OS contributions so I'm gonna appreciate your advice here:
1. The attribute/schema structure designed here (including only table names) was purely following the ask from the issue itself (which I did not create, but do have an immediate use case for). I'm not sure what exactly to include to enrich this datasource, and would appreciate your thoughts here. Happy to go back to the drawing board to make the refinements then.
2. I'm also not able to review the error logs here: https://github.com/GoogleCloudPlatform/magic-modules/pull/11552#issuecomment-2313168435, so not sure how to go about fixing any issues. Any help here would also be much appreciated.

mmv1/third_party/terraform/services/bigquery/data_source_google_bigquery_tables.go
```diff
@@ -0,0 +1,83 @@
...
+		table_name := table.TableReference.TableId
+		table_names = append(table_names, table_name)
+	}
+
+	if err := d.Set("tables", table_names); err != nil {
+		log.Printf("[ERROR] Failed to set 'tables' attribute: %s", err)
+		return fmt.Errorf("error setting 'tables' attribute: %w", err)
+	}
+
+	d.SetId(time.Now().UTC().String())
```
wj-chen, line 80: Should it be based off of the dataset ID like in this example? https://github.com/GoogleCloudPlatform/magic-modules/blob/12fcf34c6b26c11f4f8c7c111775ae412787afff/mmv1/third_party/terraform/services/kms/data_source_google_kms_crypto_keys.go#L67C47-L67C56

Full dataset name + "/tables" should work (e.g. `projects/.../locations/.../datasets/.../tables`)

wj-chen (commented):
About the test failure, here is the error log:
```
2024-08-27T17:44:38.696Z [ERROR] sdk.helper_resource: Unexpected error: test_terraform_path=/bin/terraform test_working_directory=/tmp/plugintest2304613468 test_step_number=1 error="Check failed: Check 1/2 error: data.google_bigquery_tables.example: Attribute 'tables.#' expected \"1\", got \"0\"" test_name=TestAccDataSourceGoogleBigqueryTables_basic
```
I saw API request and response logs for dataset creation and table creation but not table list, or any log statement you added. Did the test pass for you when running locally?

-----
Also unrelated but we'd probably want a more descriptive release note in the PR description. Maybe "bigquery: added the google_bigquery_tables data source"

mmv1/third_party/terraform/services/bigquery/data_source_google_bigquery_tables.go
SarahFrench: @anubhavsharma515 No worries, thanks for coming and contributing!

> I'm not sure what exactly to include to enrich this datasource

Instead of the `tables` field on this data source being a list of names it could instead be a list of objects that have multiple fields describing each field. You range over the `Tables` list in your code, and at that point you can access [any of the values in this struct](https://pkg.go.dev/google.golang.org/api/bigquery/v2#TableListTables) for each table. Your code would need to map the fields in the client library's struct(s) to the data source's schema, e.g. [this code](https://github.com/GoogleCloudPlatform/magic-modules/blob/049127c69be365f0538dc9877b4ffdb27d8083f7/mmv1/third_party/terraform/services/kms/data_source_google_kms_crypto_key_versions.go#L233-L255) shows how you'd make a map[string]interface{} per table where the map keys are field names, and collect those maps in an []interface{}.

However there's no guarantee that the data returned from the client library when listing resources is as comprehensive as if you were retrieving the details of a single table (hence why re-using the schema from `google_bigquery_table` might not be 100% appropriate). In that case it's a choice of what makes sense to implement vs not. **@/wj-chen's suggestion to just implement names to start with is be a good idea here!** However you'd still need to update the PR so that the list is a list of objects instead of a list of strings.



>I'm also not able to review the error logs here

If you're unable to run the tests locally then us reviewers can give feedback about what needs fixing


---

>Is it a convention to return the full resource in a data source list? If we only support names but want to later on expand to the full resource, would that be considered a breaking change?

@wj-chen I don't think there are very established conventions here as there aren't many plural data sources! In the past I've tried to ensure as much data is present as possible for each item listed, but that may be unnecessary. As a counter-example [google_storage_buckets](https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/third_party/terraform/services/storage/data_source_google_storage_buckets.go) includes only a [subset of the possible fields](https://github.com/GoogleCloudPlatform/magic-modules/blob/049127c69be365f0538dc9877b4ffdb27d8083f7/mmv1/third_party/terraform/services/storage/data_source_google_storage_buckets.go#L30-L50). I think implementing only the name field to start is valid, but it would require the swap from list of strings to list of maps.

mmv1/third_party/terraform/services/bigquery/data_source_google_bigquery_tables_test.go
```diff
@@ -0,0 +1,63 @@
...
+        "type": "STRING",
+        "mode": "NULLABLE"
+      }
+    ]
+    EOF
+  }
+
+  data "google_bigquery_tables" "example" {
+    dataset_id = google_bigquery_dataset.foo.dataset_id
+  }
```
SarahFrench, line 61:
```suggestion
  data "google_bigquery_tables" "example" {
    dataset_id = google_bigquery_table. test_table.dataset_id
  }
```
OR 

```suggestion
  data "google_bigquery_tables" "example" {
    dataset_id = google_bigquery_dataset.foo.dataset_id
    depends_on = google_bigquery_table. test_table
  }
```

This will help ensure the table is provisioned before the data source tries to retrieve tables from the dataset

mmv1/third_party/terraform/services/bigquery/data_source_google_bigquery_tables.go
wj-chen: I think to avoid potentially having to update the list from just name strings to table objects, which I believe would be a breaking change for users, it might be worth exposing table objects from the get-go. On the other hand, the BigQuery Table schema is fairly involved, and I haven't done any work with data sources in the past myself to know if there might be complications mapping the schemas. I will let @anubhavsharma515 explore a bit.

If it becomes too much of a headache, we can also maybe just name this data source `google_bigquery_table_names` and scope it down to just name strings. What do you both think?

anubhavsharma515: Hey folks! I appreciate the patient feedback here.

I'm not apposed to enriching the legwork of investigating what makes sense to go in here at all. However, and this may come with a hint of selfishness, I believe that the availability of this data source with even the table names is of immense value for several use-cases. That being said, I am fully aligned on the concerns with having this attribute be a `list` - especially since future enrichment will likely be cumbersome as a result of.

For the time-being, I believe a fair implementation involves incorporating @SarahFrench's recommendation to make the attribute a map. @wj-chen would you say it's fair to leave the name of this data source as `google_bigquery_tables` following this change?

mmv1/third_party/terraform/services/bigquery/data_source_google_bigquery_tables.go
```diff
@@ -0,0 +1,83 @@
...
+		return fmt.Errorf("Error fetching project: %s", err)
+	}
+
+	bigquery_service, err := bq.NewService(ctx)
+
+	if err != nil {
+		return fmt.Errorf("Error creating BigQuery service: %s", err)
+	}
+
+	tables, err := bq.NewTablesService(bigquery_service).List(project, dataset_id).Do()
```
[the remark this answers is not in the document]
anubhavsharma515: This is really helpful context, @wj-chen. Thank you for sharing 🫡.

I've tried to implement this with the most recent commits - seemed like something that would be worthwhile to have at first-pass just given the scale of things I'm personally working with at work atm 😅.

mmv1/third_party/terraform/services/bigquery/data_source_google_bigquery_tables.go
```diff
@@ -0,0 +1,102 @@
...
+	"log"
+
+	bq "google.golang.org/api/bigquery/v2"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"github.com/hashicorp/terraform-provider-google/google/tpgresource"
+	transport_tpg "github.com/hashicorp/terraform-provider-google/google/transport"
+)
+
+func DataSourceGoogleBigQueryTables() *schema.Resource {
```
wj-chen, line 15: Please make the indentations in this file consistent

mmv1/third_party/terraform/services/bigquery/data_source_google_bigquery_tables.go
```diff
@@ -0,0 +1,102 @@
...
+		}
+
+		tables, err := listCall.Do()
+		if err != nil {
+			return fmt.Errorf("Error listing tables: %s", err)
+		}
+
+		for _, table := range tables.Tables {
+			tableName := table.TableReference.TableId
+			log.Printf("[INFO] Found BigQuery table: %s", tableName)
```
wj-chen, line 82: Do we need to log here if we are returning the names already?

mmv1/third_party/terraform/services/bigquery/data_source_google_bigquery_tables.go
```diff
@@ -0,0 +1,102 @@
...
+
+			tableMap[tableName] = nil
```
wj-chen, line 84: More in the other comment thread, but if I understood the suggestion correctly, each map key should be a [`google_bigquery_table`](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/bigquery_table) attribute and the value the corresponding value, so in this case the key would be `table_id` and the value `tableName`. In the future we can expand this map to return more table attributes like `friendly_name`, `labels`, etc. Then for each map representation of a table, we add that to the `tables` list in this data source.

mmv1/third_party/terraform/services/bigquery/data_source_google_bigquery_tables.go
```diff
@@ -0,0 +1,102 @@
...
+			Required:    true,
+			Description: "The ID of the dataset containing the tables.",
+		},
+		"project": {
+			Type:        schema.TypeString,
+			Optional:    true,
+			Description: "The ID of the project in which the dataset is located. If it is not provided, the provider project is used.",
+		},
+    "tables": {
+			Type:         schema.TypeMap,
```
wj-chen, line 29: More in the other comment thread but I think this should still be a list.

mmv1/third_party/terraform/services/bigquery/data_source_google_bigquery_tables.go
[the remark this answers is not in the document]
wj-chen: Great, thanks a lot of taking care of pagination!

mmv1/third_party/terraform/services/bigquery/data_source_google_bigquery_tables.go
wj-chen: I misunderstood the suggestion earlier, per my current understanding, yes let's keep the data source name `google_bigquery_tables`. And if my current understanding is correct, `tables` will still be a list, but now a list of maps where each map is a representation of a [`google_bigquery_table`](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/bigquery_table), in the form of key-value pairs of a `google_biguqery_table` attributes to their values. For now we want to return only the table names, so each map will have only one entry looking like `table_id`=`<the table name>`. In the future we can add more attributes. @SarahFrench Let me know if my understanding is correct.

mmv1/third_party/terraform/services/bigquery/data_source_google_bigquery_tables.go
anubhavsharma515: Great point, this was essentially for testing. I can take it out!

mmv1/third_party/terraform/services/bigquery/data_source_google_bigquery_tables.go
anubhavsharma515: @wj-chen @SarahFrench my apologies, I completely misunderstood the structure of the changes requested. As advised, the following structure for `tables` attribute has been implemented:

```
tables = [
  {
    "table_id" = "<string>",
    "labels" = {
        <key/value pair>....
    },
  },
  ....
]
```

mmv1/third_party/terraform/services/bigquery/data_source_google_bigquery_tables_test.go
```diff
@@ -0,0 +1,70 @@
...
+    dataset_id                  = "tf_test_ds_%{random_suffix}"
+    friendly_name               = "testing"
+    description                 = "This is a test description"
+    location                    = "US"
+    default_table_expiration_ms = 3600000
+  }
+
+  resource "google_bigquery_table" "test" {
+    dataset_id        = google_bigquery_dataset.test.dataset_id
+    table_id          = "test_table"
```
SarahFrench, line 48:
```suggestion
    table_id          = "tf_test_table_%{random_suffix}"
```

We try to make sure resources provisioned by test start with tf-test-/tf_test_ so that our automation can identify them as safe to delete them from our testing projects. Even though this table would be deleted when the parent dataset is deleted we try to keep this consistent everywhere for simplicity!

(L26 would need updating to reflect this too)

mmv1/third_party/terraform/services/bigquery/data_source_google_bigquery_tables_test.go
```diff
@@ -0,0 +1,70 @@
...
+        "name": "name",
+        "type": "STRING",
+        "mode": "NULLABLE"
+      }
+    ]
+    EOF
+    # Default label for a tf-provisioned table (referenced in the test suite):
+    # labels = {
+    #   goog-terraform-provisioned = "true"
+    # }
```
SarahFrench, line 62: This can be deleted - the provider automatically adds this label to all resources now

SarahFrench (commented):
Thanks for implementing the changes to the `tables` attribute, looks good!

I've left some minor comments, and after those are addressed I think we'd be good to go ahead and merge

mmv1/third_party/terraform/services/bigquery/data_source_google_bigquery_tables_test.go
anubhavsharma515: Oo this is great context for future contributions, thanks for sharing Sarah! I'll go ahead and make the change 👍🏼

mmv1/third_party/terraform/services/bigquery/data_source_google_bigquery_tables_test.go
anubhavsharma515: Ah yeah this was essentially a contextual comment for anyone new reading the code not figuring out where the test [here](https://github.com/GoogleCloudPlatform/magic-modules/pull/11552/files#diff-45919cfdf826343c45a26d709c331f1a41b9a5170d1f26c9144e55041a7ad5a8R28) is showing up from

anubhavsharma515:
Hey @SarahFrench, apologies for the direct ping but just wanted to provide some context for the most recent commits.

I reviewed the issue again today post your almost-merge-ready message and noticed a the request for an equivalent "datasets" data source. Seeing that the implementation would go through the same considerations as the tables data source, I've added in (implementing all feedback from the 'tables' side of things).

Please let me know if this would be more appropriately addressed in another PR and I shall go ahead and revert the last two commits.

Thanks in advance!

mmv1/third_party/terraform/services/bigquery/data_source_google_bigquery_tables.go
```diff
@@ -0,0 +1,143 @@
...
+			Type:        schema.TypeString,
+			Optional:    true,
+			Description: "The ID of the project in which the dataset is located. If it is not provided, the provider project is used.",
+		},
+		"tables": {
+			Type:     schema.TypeList,
+			Computed: true,
+			Elem: &schema.Resource{
+				Schema: map[string]*schema.Schema{
+					"labels": {
```
wj-chen, line 29: We don't need to add this for now since there is no use case of it yet, I just mentioned it in a previous comment as an example. Let's remove it to scope down the PR but thanks for looking into it.

mmv1/third_party/terraform/services/bigquery/data_source_google_bigquery_tables.go
```diff
@@ -0,0 +1,143 @@
...
+		return fmt.Errorf("Error retrieving tables: %s", err)
+	}
+
+	id := fmt.Sprintf("projects/%s/datasets/%s/tables", project, datasetID)
+	d.SetId(id)
+
+	return nil
+}
+
+func flattenDatasourceGoogleBigQueryTablesList(v interface{}) []map[string]interface{} {
```
wj-chen, line 109: "DataSource"

mmv1/third_party/terraform/services/bigquery/data_source_google_bigquery_tables.go
```diff
@@ -0,0 +1,143 @@
...
+	if v == nil {
+		return make([]map[string]interface{}, 0)
+	}
+
+	ls := v.([]interface{})
+
+	tables := make([]map[string]interface{}, 0, len(ls))
+
+	for _, raw := range ls {
+		o := raw.(map[string]interface{})
```
wj-chen, line 120: Can we use a more descriptive variable name?

mmv1/third_party/terraform/services/bigquery/data_source_google_bigquery_tables_test.go
```diff
@@ -0,0 +1,67 @@
...
+        "type": "STRING",
+        "mode": "NULLABLE"
+      }
+    ]
+    EOF
+  }
+
+  data "google_bigquery_tables" "example" {
+    dataset_id = google_bigquery_table.test.dataset_id
+    depends_on = [google_bigquery_table.test]
```
wj-chen, line 64: Since you are referencing the table's dataset_id above I think you don't need to explicitly set up this dependency, can you remove and see?

wj-chen (changes requested):
~~The release notes in the description are not showing up correctly. Can you try e.g. "bigquery: added the `google_bigquery_tables` data source"?~~
Edit: Release note format was correct.

Also would recommend breaking up the two data sources into separate PRs so we keep each change isolated.

mmv1/third_party/terraform/services/bigquery/data_source_google_bigquery_datasets.go
```diff
@@ -0,0 +1,137 @@
...
+	"fmt"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"github.com/hashicorp/terraform-provider-google/google/tpgresource"
+	transport_tpg "github.com/hashicorp/terraform-provider-google/google/transport"
+)
+
+func DataSourceGoogleBigQueryDatasets() *schema.Resource {
+
+	dsSchema := map[string]*schema.Schema{
```
SarahFrench, line 13: Could you please implement a top-level field `filter` that will be used to set [the filter query parameter](https://cloud.google.com/bigquery/docs/reference/rest/v2/datasets/list#query-parameters)? That query parameter isn't available when listing tables, so I didn't mention it before, but we should support it for listing datasets.

Adding filtering will impact:
- The id of the data source - [see example here](https://github.com/BBBmau/magic-modules/blob/17a2e4db34706f557b7d864b3b20efe5a78381cf/mmv1/third_party/terraform/services/kms/data_source_google_kms_crypto_key_versions.go#L90-L93)
- The filter param will need to be included in the URL if filter is set - [example here
](https://github.com/BBBmau/magic-modules/blob/17a2e4db34706f557b7d864b3b20efe5a78381cf/mmv1/third_party/terraform/services/kms/data_source_google_kms_crypto_key_versions.go#L181-L193)

mmv1/third_party/terraform/services/bigquery/data_source_google_bigquery_tables.go
anubhavsharma515: Hey @wj-chen! Appreciate the feedback here.

I was hoping that we keep this particular piece in (if acceptable/reasonable to of course). Having this available OOB will be of immense value to my organization, as there's a lot of power in leveraging the labels applied to tables for conditionally performing operations.

Would this be possible to leave in by any chance?

mmv1/third_party/terraform/services/bigquery/data_source_google_bigquery_tables.go
anubhavsharma515: Absolutely, I'll rename this to `output`.

mmv1/third_party/terraform/services/bigquery/data_source_google_bigquery_datasets.go
anubhavsharma515: Hi @SarahFrench! Thank you so much for the feedback. Per @wj-chen's comment here: https://github.com/GoogleCloudPlatform/magic-modules/pull/11552#pullrequestreview-2278675204
I'll be reverting these changes and spinning up an identical PR soon. I'll address this feedback over there 👍🏼.

anubhavsharma515:
> The release notes in the description are not showing up correctly. Can you try e.g. "bigquery: added the `google_bigquery_tables` data source"?
> 
> Also would recommend breaking up the two data sources into separate PRs so we keep each change isolated.

Hey @wj-chen! Unfortunately if I make any changes to the release notes from how they've been set currently, this CI check fails: https://github.com/GoogleCloudPlatform/magic-modules/actions/runs/10685790058/job/29619454282?pr=11552.

Looking at best practices [here](https://googlecloudplatform.github.io/magic-modules/contribute/release-notes/#type-specific-guidelines-and-examples), I believe the way it's set is the only way the CI check passes.

wj-chen:
> > The release notes in the description are not showing up correctly. Can you try e.g. "bigquery: added the `google_bigquery_tables` data source"?
> > Also would recommend breaking up the two data sources into separate PRs so we keep each change isolated.
> 
> Hey @wj-chen! Unfortunately if I make any changes to the release notes from how they've been set currently, this CI check fails: https://github.com/GoogleCloudPlatform/magic-modules/actions/runs/10685790058/job/29619454282?pr=11552.
> 
> Looking at best practices [here](https://googlecloudplatform.github.io/magic-modules/contribute/release-notes/#type-specific-guidelines-and-examples), I believe the way it's set is the only way the CI check passes.

I meant this as the release note in markdown (remove the `\`s at the beginning; I'm just using them as escapes):
```
\```release-note:enhancement
\bigquery: added the `google_bigquery_tables` data source
\```
```

Currently in the PR description I'm not seeing a full sentence as the release note. It just says
```
`google_bigquery_tables`
```

mmv1/third_party/terraform/services/bigquery/data_source_google_bigquery_tables.go
```diff
@@ -0,0 +1,143 @@
...
+
+	if err != nil {
+		return fmt.Errorf("Error fetching project: %s", err)
+	}
+
+	params := make(map[string]string)
+	tables := make([]map[string]interface{}, 0)
+
+	for {
+		url := fmt.Sprintf("https://bigquery.googleapis.com/bigquery/v2/projects/%s/datasets/%s/tables", project, datasetID)
```
wj-chen, line 71: Can you construct the `url` using base path like this instead of hardcoding the API endpoint? https://github.com/GoogleCloudPlatform/magic-modules/blob/40b53825878a3eaaf56d8dbcc49835c69723a31a/mmv1/third_party/terraform/services/bigquery/resource_bigquery_table.go#L1805

wj-chen (commented):
Update the PR title? Otherwise aside from the one comment everything else looks good

mmv1/third_party/terraform/services/bigquery/data_source_google_bigquery_tables.go
anubhavsharma515: Sounds good, lots of learning on my end with these suggestions so I really appreciate your patience with providing them.

Addressed!

wj-chen: No problem, thanks for addressing the feedback. For this block let's also check `err` and return it if it's present. This error handling is based on golang's style guide: https://go.dev/blog/error-handling-and-go. Also the indentation here is different from the rest of the block.

SarahFrench:
The original release note layout was correct according to our contribution guides: https://googlecloudplatform.github.io/magic-modules/contribute/release-notes/#type-specific-guidelines-and-examples

We'll need 2 entries due to the PR adding 2 data sources. I'll update them

wj-chen:
> The original release note layout was correct according to our contribution guides for new resources/data sources: https://googlecloudplatform.github.io/magic-modules/contribute/release-notes/#type-specific-guidelines-and-examples
> 
> ~We'll need 2 entries due to the PR adding 2 data sources.~ I'll update them. Edit:Just saw the PR is being split

Thanks for the correction. Didn't realize data source release note has its own format.

wj-chen (approved):
Thank you very much for this contribution!

SarahFrench (approved):
>Hey @SarahFrench! Just popping by to ask what the release process looks like once all checks have passed?

When a PR is merged it joins a release cut, made on Wednesdays, that's determined by what's merged into main by the end of Tuesday. That release cut is checked for bugs and then released the following Monday. So by merging this now it should get into next weeks release cut, and be released on Mon September 16th

billinghamj:
Is there a new PR for the datasets change? https://github.com/GoogleCloudPlatform/magic-modules/pull/11552#pullrequestreview-2278675204

PR #11553, opened 2024-08-27
feat: Remove force replacement from gcfs_config

dominykasn:
Gcfsconfig changes do not require to re-create node pools hence there is no point to ForceNew on each gcfsconfig change, gcfs can be enabled and disabled via GUI or gcloud CLI without recreating node pools, examples:
https://cloud.google.com/kubernetes-engine/docs/how-to/image-streaming#disable

https://cloud.google.com/kubernetes-engine/docs/how-to/image-streaming#enable_on_node_pools

Also, same issue noted here https://github.com/hashicorp/terraform-provider-google/issues/18417

```release-note:enhancement
container: removed unnecessary force replacement in `gcfs_config`
```

### DIFF
--- a/mmv1/third_party/terraform/services/container/go/node_config.go.tmpl
+++ b/mmv1/third_party/terraform/services/container/go/node_config.go.tmpl
@@ -87,13 +87,12 @@ func schemaLoggingVariant() *schema.Schema {
 	}
 }
 
-func schemaGcfsConfig(forceNew bool) *schema.Schema {
+func schemaGcfsConfig() *schema.Schema {
         return &schema.Schema{
                 Type:     schema.TypeList,
 		Optional: true,
 		MaxItems: 1,
 		Description: `GCFS configuration for this node.`,
-		ForceNew: forceNew,
 		Elem: &schema.Resource{
 			Schema: map[string]*schema.Schema{
 				"enabled": {
@@ -332,7 +331,7 @@ func schemaNodeConfig() *schema.Schema {
 					},
 				},
 
-				"gcfs_config": schemaGcfsConfig(true),
+				"gcfs_config": schemaGcfsConfig(),
 
 				"gvnic": {
 					Type:     schema.TypeList,
--- a/mmv1/third_party/terraform/services/container/go/resource_container_cluster.go.tmpl
+++ b/mmv1/third_party/terraform/services/container/go/resource_container_cluster.go.tmpl
@@ -159,7 +159,7 @@ func clusterSchemaNodePoolDefaults() *schema.Schema {
 						Schema: map[string]*schema.Schema{
 							"containerd_config": schemaContainerdConfig(),
 {{- if ne $.TargetVersionName "ga" }}
-							"gcfs_config": schemaGcfsConfig(false),
+							"gcfs_config": schemaGcfsConfig(),
 {{- end }}
 							"logging_variant": schemaLoggingVariant(),
 						},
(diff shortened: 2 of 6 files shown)


hao-nan-li (commented):
I don't think `contributor-membership-checker (graphite-docker-images)` is part of the CI nowadays, could you rebase on main and re-commit?

dominykasn:
Rebased and re-commited from main branch, however, I see other branches also have the same checks

hao-nan-li:
> Rebased and re-commited from main branch, however, I see other branches also have the same checks

Thanks for the rebase, thanks for correcting, `contributor-membership-checker` is indeed part of the check, just that it no longer requires manual approval.

wyardley:
Similar note to: https://github.com/GoogleCloudPlatform/magic-modules/pull/11541#pullrequestreview-2269095446 - it may be worth adding the suggested test of this parameter when it's used in `resource "google_container_cluster"`, and specifically testing updates as well. It's a known issue that many of these parameters don't actually support updating when called this way.

dominykasn:
Sure, will check, however, is there a possibility to check why my tests are failing?

dominykasn:
Also, please correct me if I'm wrong but there are already tests written [here](https://github.com/GoogleCloudPlatform/magic-modules/blob/e1694376f735ab3c6c587baff0551b51a28a0904/mmv1/third_party/terraform/services/container/resource_container_cluster_test.go.erb#L1796) for cluster and gcfs config changes?

wyardley:
Oh yeah, looks like it!

dominykasn:
What do I need to do more in this case for this to get approved? 🤔

hao-nan-li:
> What do I need to do more in this case for this to get approved? 🤔

I think the VCR test needs to be fixed. `resource_container_node_pool_test.go:1707: Step 3/4 error: Check failed: Check 1/1 error: google_container_node_pool.np: Attribute 'node_config.0.gcfs_config.0.enabled' expected "false", got "true"
`

wyardley:
> Also, please correct me if I'm wrong but there are already tests written [here](https://github.com/GoogleCloudPlatform/magic-modules/blob/e1694376f735ab3c6c587baff0551b51a28a0904/mmv1/third_party/terraform/services/container/resource_container_cluster_test.go.erb#L1796) for cluster and gcfs config changes?

[Just to be clear, I'm not a Google employee, and am not reviewing this PR, just was commenting in case it's helpful since I did some work around similar changes recently.]

Reading those a bit more carefully, from what I can see, that's only `node_pool_defaults.node_config_defaults`, which influence the default behavior of newly created node pools. This only requires a cluster update to be performed. 
The comment is also requesting a test for `node_config.gcfs_config`, which I don't think is tested anyhere in the case where it's configured for the default pool created within the `google_container_cluster` resource, vs. a separately created nodepool..

My prediction is that if you add or extend the test to test `node_config.gcfs_config`, it will fail, because I don't think there's anything related to `gcfs_config` [here](https://github.com/GoogleCloudPlatform/magic-modules/blob/e1694376f735ab3c6c587baff0551b51a28a0904/mmv1/third_party/terraform/services/container/resource_container_cluster.go.erb#L3801).

That is more complicated to update, and as yet only implemented for one or two parameters. See https://github.com/hashicorp/terraform-provider-google/issues/19225 for a high level description of the issue, and some of the comments in #11272 for more context

hao-nan-li (approved):

PR #11556, opened 2024-08-27
`pubsub`: allow empty filter definition

prauc:
This PR allows to define an empty filter definition, as we do for our API. The implementation is due to PR #11421, which prohibits this.

Tests added and passed.

Fixes [#19269](https://github.com/hashicorp/terraform-provider-google/issues/19269)

```release-note:bug
pubsub: fixed a validation bug that didn't allow empty filter definitions for `google_pubsub_subscription` resources
```


SarahFrench (approved):
This looks good to me, thanks! I'm going to suggest and apply an change to remove the copyright notice before merging though

PR #11548, opened 2024-08-26
Update docs for `google_compute_firewall` resource - Remove quotes for numbers in the list

air3ijai:
This is a follow up of the closed PR (https://github.com/hashicorp/terraform-provider-google/pull/19251) and it adds cosmetic change to reflect more appropriately Terraform type [list](https://developer.hashicorp.com/terraform/language/expressions/types#lists-tuples) for [`google_compute_firewall`](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/compute_firewall) resource
```diff
+ [22], [80, 443]
- ["22"], ["80","443"]
```

**Release Note Template for Downstream PRs (will be copied)**

```release-note:none

```

PR #11561, opened 2024-08-28
Update docs and tests for google_compute_firewall resource - Remove quotes for numbers in the list

air3ijai:
This is a follow up of the #11548 and adds similar changes to the lists in the tests and templates, to reflect more appropriately Terraform type [list](https://developer.hashicorp.com/terraform/language/expressions/types#lists-tuples) for [`google_compute_firewall`](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/compute_firewall) resource
```diff
+ [80, 8080, "1000-2000"]
- ["80", "8080", "1000-2000"]
```

```release-note:none

```

### DIFF
--- a/mmv1/templates/terraform/examples/firewall_basic.tf.erb
+++ b/mmv1/templates/terraform/examples/firewall_basic.tf.erb
@@ -8,7 +8,7 @@ resource "google_compute_firewall" "default" {
 
   allow {
     protocol = "tcp"
-    ports    = ["80", "8080", "1000-2000"]
+    ports    = [80, 8080, "1000-2000"]
   }
 
   source_tags = ["web"]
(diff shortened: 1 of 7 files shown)


rileykarson:
Hmm- I'd lean towards keeping it this way as the field is a list of strings: https://github.com/hashicorp/terraform-provider-google-beta/blob/f590ae668a68f3c004f621b13cb72ca3ddc5b919/google-beta/services/compute/resource_compute_firewall.go#L398-L411

Terraform does do fairly aggressive [type conversion](https://developer.hashicorp.com/terraform/language/expressions/types#type-conversion) that means that your examples work, at least in HCL, but I'd mildly prefer that the official docs follow the schema types strictly so that it's clear in cases that that may not be the case like modules/CDK (maybe?)/checks.

air3ijai:
Isn't it a number or range (string)?

That story is confusing when all your code has unquoted [numbers](https://developer.hashicorp.com/terraform/language/expressions/types#number).

Screenshot 2024-08-30 at 22 10 01

rileykarson:
They're valid for the reason I linked above, but I think we should stick more to the strict string type.

PR #11562, opened 2024-08-28
Update resource_container_cluster.go.erb

maci0:
Fix autopilot diff suppression func.
We only need a diff if a cluster already exists.

This will fix https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/issues/2042 and other similar issues 

```release-note:bug
Fix autopilot diff suppression func.
We only need a diff if a cluster already exists.

This will fix https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/issues/2042 and other similar issues 
```

### DIFF
--- a/mmv1/third_party/terraform/services/container/resource_container_cluster.go.erb
+++ b/mmv1/third_party/terraform/services/container/resource_container_cluster.go.erb
@@ -113,7 +113,7 @@ var (
 
 	suppressDiffForAutopilot = schema.SchemaDiffSuppressFunc(func(k, oldValue, newValue string, d *schema.ResourceData) bool {
 		if v, _ := d.Get("enable_autopilot").(bool); v {
-			/ If there is an ID we can safely assume the cluster already exists and return true.
+			// If there is an ID we can safely assume the cluster already exists and return true.
 			// If there is no ID we don't need to supress any diffs and return false.
 			return d.Id() != ""
 		}
--- a/mmv1/third_party/terraform/services/container/resource_container_cluster.go.erb
+++ b/mmv1/third_party/terraform/services/container/resource_container_cluster.go.erb
@@ -113,7 +113,9 @@ var (
 
 	suppressDiffForAutopilot = schema.SchemaDiffSuppressFunc(func(k, oldValue, newValue string, d *schema.ResourceData) bool {
 		if v, _ := d.Get("enable_autopilot").(bool); v {
-			return true
+			/ If there is an ID we can safely assume the cluster already exists and return true.
+			// If there is no ID we don't need to supress any diffs and return false.
+			return d.Id() != ""
 		}
 		return false
 	})


ajinkya101:
Hi Team, Can you take a look at this Request? It is one of the core requirement for the enterprise customer

cc - @maci0 @slevenick

slevenick (commented):
We've had some discussions about this and I'm not sure this is the correct fix.

For one, this function is being used in 6 places across this file. Are we sure that we should be adding those fields to the initial plan for autopilot clusters? Is specifying things like disk size valid for autopilot clusters?

Second, if we're giving users the ability to set these fields on create, why would they not be able to update them? If a user can set `additive_vpc_scope_dns_domain` on an autopilot cluster, if they modify that value it should show a diff.

If we decide to move forward with this change we will also need tests for setting the fields on an autopilot cluster on create

maci0:
> For one, this function is being used in 6 places across this file. Are we sure that we should be adding those fields to the initial plan for autopilot clusters? Is specifying things like disk size valid for autopilot clusters?
> 
This I am not 100% sure about and would need to be verified/tested.

> Second, if we're giving users the ability to set these fields on create, why would they not be able to update them? If a user can set `additive_vpc_scope_dns_domain` on an autopilot cluster, if they modify that value it should show a diff.
> 

Autopilot doesn't support changing this block after the initial cluster creation. Standard GKE does though. So effectively it's immutable for Autopilot only, if this will change in the future I don't know but that is what autopilot currently supports.
I'm sure there's another more proper way to implement this edge case but personally I am not too familiar with the codebase here.

slevenick:
> > For one, this function is being used in 6 places across this file. Are we sure that we should be adding those fields to the initial plan for autopilot clusters? Is specifying things like disk size valid for autopilot clusters?
> 
> This I am not 100% sure about and would need to be verified/tested.
> 
> > Second, if we're giving users the ability to set these fields on create, why would they not be able to update them? If a user can set `additive_vpc_scope_dns_domain` on an autopilot cluster, if they modify that value it should show a diff.
> 
> Autopilot doesn't support changing this block after the initial cluster creation. Standard GKE does though. So effectively it's immutable for Autopilot only, if this will change in the future I don't know but that is what autopilot currently supports. I'm sure there's another more proper way to implement this edge case but personally I am not too familiar with the codebase here.

Ok, if it's immutable for Autopilot then we should probably remove the diff suppress func entirely for this field and instead add a customize diff that will trigger replacement if this field in particular changes *and* autopilot is set to true. You can see an example of similar behavior on `containerClusterAutopilotCustomizeDiff`, and this could potentially be added to that function. 

Also, I don't think we want to change behavior on the fields that are sharing this diff suppress func unless we know their behavior.

slevenick (changes requested):
See most recent comment

maci0:
close in favor of https://github.com/GoogleCloudPlatform/magic-modules/pull/11744

PR #11563, opened 2024-08-28
Refactor: move validator tests into their own file, replace units tests for how provider handles credentials with acceptance tests

SarahFrench:
This PR:
- Splits up framework_provider_internal_test.go to make a file specifically for testing validators
- Deletes TestFrameworkProvider_LoadAndValidateFramework_credentials
- Re-implements the test cases from that unit test as acceptance tests

```release-note:none

```


mmv1/third_party/terraform/fwprovider/framework_provider_credentials_test.go
```diff
@@ -0,0 +1,278 @@
...
+	"regexp"
+	"testing"
+
+	"github.com/hashicorp/terraform-plugin-testing/helper/resource"
+	"github.com/hashicorp/terraform-provider-google/google/acctest"
+	"github.com/hashicorp/terraform-provider-google/google/envvar"
+	transport_tpg "github.com/hashicorp/terraform-provider-google/google/transport"
+)
+
+func TestAccFwProvider_credentials(t *testing.T) {
```
SarahFrench, line 13: For comparison, the test cases in this PR's deletions are:

- `credentials can be configured as a path to a credentials JSON file`
- `configuring credentials as a path to a non-existent file results in an error`
- `credentials set in the config are not overridden by environment variables`
- `when credentials is unset in the config, environment variables are used: GOOGLE_CREDENTIALS used first`
- `when credentials is unset in the config, environment variables are used: GOOGLE_CLOUD_KEYFILE_JSON used second`
- `when credentials is unset in the config, environment variables are used: GCLOUD_KEYFILE_JSON used third`
- `when credentials is unset in the config (and access_token unset), GOOGLE_APPLICATION_CREDENTIALS is used...` (used after the ENVs above)
- `when credentials is set to an empty string in the config the value isn't ignored and results in an error`

mmv1/third_party/terraform/fwprovider/framework_provider_credentials_test.go
```diff
@@ -0,0 +1,278 @@
...
+
+// testAccFwProvider_credentialsInEnvsOnly allows testing when the credentials argument
+// is only supplied via ENVs
+func testAccFwProvider_credentialsInEnvsOnly(context map[string]interface{}) string {
+	return acctest.Nprintf(`
+data "google_client_config" "default" {}
+
+output "token" {
+  value = data.google_client_config.default.access_token
+}
```
SarahFrench, line 276:
```suggestion
output "token" {
  value = data.google_client_config.default.access_token
  sensitive = true
}
`, context)
}

// testAccFwProvider_credentialsInEnvsOnly allows testing when the credentials argument
// is only supplied via ENVs
func testAccFwProvider_credentialsInEnvsOnly(context map[string]interface{}) string {
	return acctest.Nprintf(`
data "google_client_config" "default" {}

output "token" {
  value = data.google_client_config.default.access_token
  sensitive = true
}
```

mmv1/third_party/terraform/provider/provider_credentials_test.go
```diff
@@ -0,0 +1,263 @@
...
+	"github.com/hashicorp/terraform-plugin-testing/helper/resource"
+	"github.com/hashicorp/terraform-provider-google/google/acctest"
+	"github.com/hashicorp/terraform-provider-google/google/envvar"
+	transport_tpg "github.com/hashicorp/terraform-provider-google/google/transport"
+)
+
+// TestAccSdkProvider_credentials is a series of acc tests asserting how the SDK provider handles credentials arguments
+// It is SDK specific because the HCL used provisions SDK-implemented resources
+// It is a counterpart to TestAccFwProvider_credentials
+func TestAccSdkProvider_credentials(t *testing.T) {
```
SarahFrench, line 16: This file was made by copy-pasting `framework_provider_credentials_test.go` and updating the HCL to be SDK specific. We could reduce this duplication by combining in one file, but it would be really easy to accidentally make tests that pass that don't actually test the SDK and the PF (that all depends on the HCL used). I'm open to ideas!

mmv1/third_party/terraform/fwprovider/framework_provider_credentials_test.go
```diff
@@ -0,0 +1,283 @@
...
+				PlanOnly:    true,
+				ExpectError: regexp.MustCompile("expected a non-empty string"),
+			},
+		},
+	})
+}
+
+// testAccFwProvider_credentialsInProviderBlock_provisionSdkResource is used when we want to make a plan to test
+// the credentials JSON path is valid (ie file exists) but we want to avoid a test failure cause by trying to use
+// those credentials, as they're fake.
```
SarahFrench, line 240: If we do a PlanOnly test step with the config using google_client_config then we get an error because the credentials are actively used to generate an access token, whereas a plan using a resource like google_service_account doesn't active try to use the credentials until apply time.

The alternative to this would be to migrate a data source (this is more conventional than google_client_config) back to the plugin-framework to be used here in this test. However we know that would be a breaking change until the muxing fixes are finished.

This currently doesn't test the PF provider but my intention is to update this to use a PF-implemented data source in the future.

SarahFrench (commented):
Self-review to help reviewer

SarahFrench:
Wondering if the above could be a race condition, as the tests passed ok during development... /gcbrun

SarahFrench:
I realised that I don't actually know if the provider is guaranteed to be re-configured between each step. The failure above looks like it's because the provider configured in step 1 is still used in step 2 (bad creds carry over), and that also means that the test as a whole isn't reliable...?

SarahFrench:
Ah, the error from attempting to use fake creds doesn't include any data about email so if I want to inspect the (fake) creds contents I'll need to surface it via a data source.

At this point my options are to split the `		"when credentials is unset in the config, environment variables are used in a given order": testAccSdkProvider_credentials_precedenceOrderEnvironmentVariables, // GOOGLE_CREDENTIALS, GOOGLE_CLOUD_KEYFILE_JSON, GCLOUD_KEYFILE_JSON, GOOGLE_APPLICATION_CREDENTIALS` test into multiple acc tests (vs multiple steps in one test) or using a data source to expose the creds value. I'm leaning towards the latter as that's the approach I'll take for other acc tests to check how things like User Agent are set.

SarahFrench:
This PR has got too much noise in it now, so I'll split it into:
- A PR for moving the validator tests into their own file
- A PR for acc tests checking behaviour of the credentials argument but using a data source

PR #11567, opened 2024-08-28
Terraform redis tags

SamanthaMathews:
This PR is to add tags field to Redis Instaance

release-note:enhancement


mmv1/third_party/terraform/services/redis/resource_redis_instance_test.go
```diff
@@ -319,6 +319,42 @@ func TestAccRedisInstance_downgradeRedisVersion(t *testing.T) {
 	})
 }
 
+func TestAccRedisInstance_tags(t *testing.T) {
+	t.Skip()
```
zli82016, line 323: What is the reason to skip this test?

PR #3760, opened 2020-07-21
Upstream support for Kubelet and Linux Node configurations in GKE

rileykarson:
Upstreams https://github.com/terraform-providers/terraform-provider-google-beta/pull/2279, fixes https://github.com/terraform-providers/terraform-provider-google/issues/6773

I took some liberty with upstreaming the docs- I added documentation for the nested types, which I'd missed were absent in the original PR.

**Release Note Template for Downstream PRs (will be copied)**

```release-note:enhancement
container: added support for `kubelet_config` and `linux_node_config` to GKE node pools (beta)
```

PR #11572, opened 2024-08-29
container: make `cpu_manager_policy` optional in `kubelet_config`

wyardley:
Part of https://github.com/hashicorp/terraform-provider-google/issues/19225

This should resolve some confusing behavior with `cpu_manager_policy`

* It frequently will show a permadrift when it can't be set
* It also doesn't seem to accept the documented value of "none" as an empty value, though the previously undocumented empty string (`""`) seems to work.

https://github.com/GoogleCloudPlatform/magic-modules/pull/3760/commits/efb71a903fc26775d568b8129ef18ce707dfbdc6#r458238583
https://github.com/GoogleCloudPlatform/magic-modules/pull/3760/commits/efb71a903fc26775d568b8129ef18ce707dfbdc6#r473173480
☝️ context on when it was originally marked `Required`

This doesn't resolve all of the issues, but resolves other issues where it must be set where it's not actually needed (for example, if `insecure_kubelet_readonly_port_enabled` is set).

It appears that it was marked as `Required` somewhat arbitrarily (see above), and it's also possible that some of what's in place is tied to an API level bug that may have since been resolved. Maybe we could require at last one instead -- happy to do that if there's a good example to follow.

I did come across https://github.com/hashicorp/terraform-provider-google/issues/15767 while testing this, but I think this is neutral as far as that goes.

```release-note:bug
container: make `cpu_manager_policy` optional in `kubelet_config`
```

### DIFF
--- a/mmv1/third_party/terraform/services/container/node_config.go.erb
+++ b/mmv1/third_party/terraform/services/container/node_config.go.erb
@@ -601,9 +601,9 @@ func schemaNodeConfig() *schema.Schema {
 						Schema: map[string]*schema.Schema{
 							"cpu_manager_policy": {
 								Type:         schema.TypeString,
-								Required:     true,
+								Optional:     true,
 								ValidateFunc: validation.StringInSlice([]string{"static", "none", ""}, false),
-								Description: `Control the CPU management policy on the node.`,
+								Description:  `Control the CPU management policy on the node.`,
 							},
 							"cpu_cfs_quota": {
 								Type:     schema.TypeBool,
@@ -1204,6 +1204,7 @@ func expandKubeletConfig(v interface{}) *container.NodeKubeletConfig {
 	kConfig := &container.NodeKubeletConfig{}
 	if cpuManagerPolicy, ok := cfg["cpu_manager_policy"]; ok {
 		kConfig.CpuManagerPolicy = cpuManagerPolicy.(string)
+		kConfig.ForceSendFields = append(kConfig.ForceSendFields, "CpuManagerPolicy")
 	}
 	if cpuCfsQuota, ok := cfg["cpu_cfs_quota"]; ok {
 		kConfig.CpuCfsQuota = cpuCfsQuota.(bool)
--- a/mmv1/third_party/terraform/services/container/node_config.go.erb
+++ b/mmv1/third_party/terraform/services/container/node_config.go.erb
@@ -1204,7 +1204,6 @@ func expandKubeletConfig(v interface{}) *container.NodeKubeletConfig {
 	kConfig := &container.NodeKubeletConfig{}
 	if cpuManagerPolicy, ok := cfg["cpu_manager_policy"]; ok {
 		kConfig.CpuManagerPolicy = cpuManagerPolicy.(string)
-		kConfig.ForceSendFields = append(kConfig.ForceSendFields, "CpuManagerPolicy")
 	}
 	if cpuCfsQuota, ok := cfg["cpu_cfs_quota"]; ok {
 		kConfig.CpuCfsQuota = cpuCfsQuota.(bool)
--- a/mmv1/third_party/terraform/website/docs/r/container_cluster.html.markdown
+++ b/mmv1/third_party/terraform/website/docs/r/container_cluster.html.markdown
@@ -1292,8 +1292,7 @@ Enables monitoring and attestation of the boot integrity of the instance. The at
 
 * `cpu_manager_policy` - (Optional) The CPU management policy on the node. See
 [K8S CPU Management Policies](https://kubernetes.io/docs/tasks/administer-cluster/cpu-management-policies/).
-One of `"none"` or `"static"` (empty string `""` is also allowed, which is the
-API default). Default is unset (`"none"`).
+One of `"none"` or `"static"`. If unset (or set to the empty string `""`), the API will treat the field as if set to "none".
 
 * `cpu_cfs_quota` - (Optional) If true, enables CPU CFS quota enforcement for
 containers that specify CPU limits.
--- a/mmv1/third_party/terraform/website/docs/r/container_cluster.html.markdown
+++ b/mmv1/third_party/terraform/website/docs/r/container_cluster.html.markdown
@@ -1290,9 +1290,10 @@ Enables monitoring and attestation of the boot integrity of the instance. The at
 
 <a name="nested_kubelet_config"></a>The `kubelet_config` block supports:
 
-* `cpu_manager_policy` - (Required) The CPU management policy on the node. See
+* `cpu_manager_policy` - (Optional) The CPU management policy on the node. See
 [K8S CPU Management Policies](https://kubernetes.io/docs/tasks/administer-cluster/cpu-management-policies/).
-One of `"none"` or `"static"`. Defaults to `none` when `kubelet_config` is unset.
+One of `"none"` or `"static"` (empty string `""` is also allowed, which is the
+API default). Default is unset (`"none"`).
 
 * `cpu_cfs_quota` - (Optional) If true, enables CPU CFS quota enforcement for
 containers that specify CPU limits.
@@ -1301,11 +1302,6 @@ containers that specify CPU limits.
 as a sequence of decimal numbers, each with optional fraction and a unit suffix,
 such as `"300ms"`. Valid time units are "ns", "us" (or "µs"), "ms", "s", "m",
 "h". The value must be a positive duration.
-
--> Note: At the time of writing (2020/08/18) the GKE API rejects the `none`
-value and accepts an invalid `default` value instead. While this remains true,
-not specifying the `kubelet_config` block should be the equivalent of specifying
-`none`.
 
 * `insecure_kubelet_readonly_port_enabled` - (Optional) Controls whether the kubelet read-only port is enabled. It is strongly recommended to set this to `FALSE`. Possible values: `TRUE`, `FALSE`.
 
--- a/mmv1/third_party/terraform/services/container/resource_container_cluster_test.go.erb
+++ b/mmv1/third_party/terraform/services/container/resource_container_cluster_test.go.erb
@@ -1536,6 +1536,40 @@ func TestAccContainerCluster_withNodeConfig(t *testing.T) {
 	})
 }
 
+// Note: Updates for these are currently known to be broken (b/361634104), and
+// so are not tested here.
+// They can probably be made similar to, or consolidated with,
+// TestAccContainerCluster_withInsecureKubeletReadonlyPortEnabledInNodeConfigUpdates
+// after that's resolved.
+func TestAccContainerCluster_withNodeConfigKubeletConfigSettings(t *testing.T) {
+	t.Parallel()
+	clusterName := fmt.Sprintf("tf-test-cluster-%s", acctest.RandString(t, 10))
+	networkName := acctest.BootstrapSharedTestNetwork(t, "gke-cluster")
+	subnetworkName := acctest.BootstrapSubnet(t, "gke-cluster", networkName)
+
+	acctest.VcrTest(t, resource.TestCase{
+		PreCheck:     func() { acctest.AccTestPreCheck(t) },
+		ProtoV5ProviderFactories: acctest.ProtoV5ProviderFactories(t),
+		CheckDestroy: testAccCheckContainerClusterDestroyProducer(t),
+		Steps: []resource.TestStep{
+			{
+				Config: testAccContainerCluster_withNodeConfigKubeletConfigSettings(clusterName, networkName, subnetworkName),
+				ConfigPlanChecks: resource.ConfigPlanChecks{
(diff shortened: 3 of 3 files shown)


melinath:
Here's the full PR where it was originally marked required: https://github.com/GoogleCloudPlatform/magic-modules/pull/3760

melinath (changes requested):
If I'm understanding this correctly, the API has an API-side default of "none", but if you send `""` or `"none"` it returns an empty value? If so, this might fall under https://googlecloudplatform.github.io/magic-modules/develop/permadiff/#default_if_empty

melinath:
@hoskeri could you confirm the API behavior?

wyardley:
I’m not sure if there’s a permadiff in this case; I imagine a real fix to the linked issue may involve some more code changes.

This is only intended to be a _partial_ fix in that I don’t believe this parameter is actually required? But I think the behavior may be a net neutral or slight improvement as it is, and also lets people set unrelated parameters when needed. If the value is unset, I believe the behavior should already be default by setting it to an empty string or `”none”`

wyardley:
> TestAccContainerCluster_withInsecureKubeletReadonlyPortEnabledInNodeConfigUpdates

note: with this change, we could technically remove `cpu_manager_policy` from the test if we wanted it to be (and left in the broader `TestAccContainerNodePool_withKubeletConfig` one.

wyardley:
I can't comment on how it's _supposed_ to behave, obviously, so hopefully we'll get a response on that, but:

https://pkg.go.dev/google.golang.org/api/container/v1#NodeKubeletConfig
> Control the CPU management policy on the node. See https://kubernetes.io/docs/tasks/administer-cluster/cpu-management-policies/
> The following values are allowed. * "none": the default, which represents the existing scheduling behavior. * "static": allows pods with certain resource characteristics to be granted increased CPU affinity and exclusivity on the node. The default value is 'none' if unspecified.
> CpuManagerPolicy [string](https://pkg.go.dev/builtin#string) `json:"cpuManagerPolicy,omitempty"`

Even as-is, though, `"static"` and `""` (which is allowed) seem to work as expected, other than the update variant.

What I see is if I create a new cluster with the current 6.0.1 (unpatched) provider, it sends `"none"`, and the API does seem to return it vs. returning empty for default. So maybe even though they have identical results, seemingly, setting `none` does not behave the same way as not setting it?

```hcl
  node_config {
    kubelet_config {
      cpu_manager_policy = "none"
    }
  }
```

Call to API during initial cluster creation:
```
2024-08-29T15:30:52.742-0700 [DEBUG] provider.terraform-provider-google_v6.0.1_x5:   "nodeConfig": {
2024-08-29T15:30:52.742-0700 [DEBUG] provider.terraform-provider-google_v6.0.1_x5:    "kubeletConfig": {
2024-08-29T15:30:52.742-0700 [DEBUG] provider.terraform-provider-google_v6.0.1_x5:     "cpuCfsQuota": false,
2024-08-29T15:30:52.742-0700 [DEBUG] provider.terraform-provider-google_v6.0.1_x5:     "cpuManagerPolicy": "none"
2024-08-29T15:30:52.742-0700 [DEBUG] provider.terraform-provider-google_v6.0.1_x5:    },
```

Interestingly, the API does seem to return a value vs. no value in that case: maybe @hoskeri can clarify. I'm not sure what that old comment about `"default"` as a value means or if that's still the case, but I haven't tried sending that, and I don't think the provider does either. 

```
% gcloud container clusters describe --location us-central1-f xxxx | yq .nodeConfig.kubeletConfig
cpuCfsQuota: false
cpuManagerPolicy: none
```

(`cpuCfsQuota` being set to false is due to the existing known issues we're looking at separately)

This is probably good from the standpoint of the tf provider - there is no permadrift when setting the value to `"none"` explicitly, so I am guessing _most_ of the issues we have with this parameter are around updates not working.

wyardley:
Side note: I confirmed with someone that this was (at the time this was added) common practice at one point to require one attribute to avoid a situation where an empty block was defined and cause issues, and apparently there are much better checks for this in place now.

wyardley:
@melinath not sure if you heard back from @hoskeri, but a question I have is whether if we can establish that this PR is relatively safe / neutral as-is in terms of behavior, whether you and the team would feel comfortable allowing this as-is before a proper resolution to https://github.com/hashicorp/terraform-provider-google/issues/19225.

I could do some more testing if there are specific scenarios you're thinking of, but IMO, this will improve the existing behavior in a relatively safe way.

I do absolutely think the behavior and the docs around the confusing should _eventually_ be resolved and made more consistent, but I think that this change as-is should help avoid confusing diffs / behavior to users who need to set other values within `kubelet_config`.

Once we get into coercing `""` and `"none"` to the same value, I think we have a greater risk that planning may show confusing diffs to the user, even if it can be done in a way that behaves the same.

melinath:
apologies for the delayed review. I've been working through a backlog. I should be able to take a look tomorrow. In general this does look like a good change; I just want to do some manual testing to make sure I understand how it works.

mmv1/third_party/terraform/services/container/node_config.go.erb
```diff
@@ -1172,6 +1172,7 @@ func expandKubeletConfig(v interface{}) *container.NodeKubeletConfig {
 	kConfig := &container.NodeKubeletConfig{}
 	if cpuManagerPolicy, ok := cfg["cpu_manager_policy"]; ok {
 		kConfig.CpuManagerPolicy = cpuManagerPolicy.(string)
+		kConfig.ForceSendFields = append(kConfig.ForceSendFields, "CpuManagerPolicy")
```
melinath, line 1175: This will cause the field to be sent even if unset - per https://github.com/hashicorp/terraform-provider-google/issues/15767 that might not be necessary / desired?

mmv1/third_party/terraform/website/docs/r/container_cluster.html.markdown
```diff
@@ -1281,9 +1281,10 @@ Enables monitoring and attestation of the boot integrity of the instance. The at
 
 <a name="nested_kubelet_config"></a>The `kubelet_config` block supports:
 
-* `cpu_manager_policy` - (Required) The CPU management policy on the node. See
+* `cpu_manager_policy` - (Optional) The CPU management policy on the node. See
 [K8S CPU Management Policies](https://kubernetes.io/docs/tasks/administer-cluster/cpu-management-policies/).
-One of `"none"` or `"static"`. Defaults to `none` when `kubelet_config` is unset.
+One of `"none"` or `"static"` (empty string `""` is also allowed, which is the
+API default). Default is unset (`"none"`).
```
melinath, line 1287:
```suggestion
One of `"none"` or `"static"`. If unset (or set to the empty string `""`), the API will treat the field as if set to "none".
```

melinath (commented):
I'm having some trouble reproducing the described permadiff (unless I try to update a cluster, but that's due to https://github.com/hashicorp/terraform-provider-google/issues/19225, which I agree is orthogonal to this issue and doesn't need to be fixed at the same time.) https://github.com/hashicorp/terraform-provider-google/issues/19225 so they don't need to be fixed in the same PR.

My main concern is that treating `""` (unset) and `"none"` as distinct values even though they both behave the same may cause confusion for users, since the server could silently change the behavior for that case (similar to what will happen with `insecure_kubelet_readonly_port_enabled` eventually).

However, this block may also have problems with sending values to the API when they're unset - ref: https://github.com/hashicorp/terraform-provider-google/issues/15767 and https://github.com/hashicorp/terraform-provider-google/issues/19428 (for another block on the same resource.)

I'm currently inclined to go with your proposed implementation and just make the field optional with no additional changes.

mmv1/third_party/terraform/services/container/node_config.go.erb
wyardley: It's been a few, so I may have to run through some different scenarios to remember exactly why this ended up seemingly being needed.

IIRC, there was some kind of issue when I didn't have it set -- at the least, I did try it first without that line added.

If you're already setup to test it, you could try building the provider with this line removed and step through some of the tests that I did. If I've got time to dig into it, I will try to see as well.

wyardley: I did at least a basic test with this line removed again, and a config with other `kubelet_config` stuff set seemed to do what I'd expect.

Another scenario to look at is when the nodepool is managed separately (where nodepool updates _do_ work)... I think there's a case where removing the attribute when it was set might cause an issue when the attribute doesn't get set in the API call. That said, I haven't been able to induce that issue thus far with this example, and `TestAccContainerNodePool_withKubeletConfig` passes, so I'm pushing up an update that removes this again. I didn't test `""` but since that's not a documented value, probably doesn't matter too much even if that did break somehow.

```hcl
resource "google_container_cluster" "test_cfs_quota" {
  name               = "test-cfs-quota"
  location           = "us-central1-f"
  initial_node_count = 1

  node_config {
    kubelet_config {
      cpu_cfs_quota                          = true
      insecure_kubelet_readonly_port_enabled = "FALSE"
    }
  }

  deletion_protection = false
}

resource "google_container_node_pool" "test" {
  name       = "secondary-test"
  cluster    = google_container_cluster.test_cfs_quota.id
  node_count = 1

  node_config {
    kubelet_config {
      cpu_cfs_quota                          = true
      cpu_manager_policy                     = "none"
      insecure_kubelet_readonly_port_enabled = "FALSE"
    }
  }
}
```

wyardley:
> My main concern is that treating "" (unset) and "none" as distinct values even though they both behave the same may cause confusion for users, since the server could silently change the behavior for that case (similar to what will happen with insecure_kubelet_readonly_port_enabled eventually).

I totally agree, and I think this needs a real fix - am just trying to document the existing behavior that's been there (and has been in the tests) but isn't documented, especially since using `""` seems to work when `"none"` does not work properly. I'm just hoping to kind of sidestep that issue for now, but it probably will be unavoidable to fix it. And, it's probably not really "breaking" if you did decide to drop support for `""` (unless we're making that worse by documenting that here 🫠).

The API also doesn't behave quite the way I would have been expected (since , as mentioned in [this comment](https://github.com/GoogleCloudPlatform/magic-modules/pull/11572#issuecomment-2319188257), so IMO it is best if your team figures out internally both how the API should work, and how you think the provider should track default and / or unset values internally.

Anyway, I rebased and did one force-push just because this branch was kind of stale, but I left all the commit history atomic still, so hopefully that's still easy enough to review.

wyardley:
https://github.com/GoogleCloudPlatform/magic-modules/blob/29a1163236a8c87f3a9bde0ee4d95d1d2909bcb4/mmv1/third_party/terraform/services/container/resource_container_node_pool_test.go.erb#L3142-L3143

Also not 100% sure about the accuracy of these conditions, and whether the comments and / or test need any additional adjustments (or whether we need to do more provider level validation if the user tries to set disallowed combinations, or just let the API calls fail if something is not valid).

melinath (changes requested):
The more I think about this the more I agree with you that just making the field optional is safest for now. We may want to make other changes in the future but this fixes the immediate issue.

It would be best practice to make sure the `google_container_cluster.node_config.kubelet_config.cpu_manager_policy` field is used in at least one test, though.

wyardley:
Sure - I can add it. Do you want it in an existing test or a standalone new test case?

only caveat is that, since we already know that update is broken when it’s used that way, it probably can’t be a test where the resource is updated.

wyardley:
> It would be best practice to make sure the google_container_cluster.node_config.kubelet_config.cpu_manager_policy field is used in at least one test, though.

It is still present here, right?

https://github.com/GoogleCloudPlatform/magic-modules/blob/4ca0795882d087ae140ceaef479cc9e00c54f82b/mmv1/third_party/terraform/services/container/resource_container_cluster_test.go.erb#L6692

melinath:
> Sure - I can add it. Do you want it in an existing test or a standalone new test case?
> 
> only caveat is that, since we already know that update is broken when it’s used that way, it probably can’t be a test where the resource is updated.

An existing test would be ideal (but it doesn't have to be one of the ones already modified in this PR.

> > It would be best practice to make sure the google_container_cluster.node_config.kubelet_config.cpu_manager_policy field is used in at least one test, though.
> 
> It is still present here, right?
> 
> https://github.com/GoogleCloudPlatform/magic-modules/blob/4ca0795882d087ae140ceaef479cc9e00c54f82b/mmv1/third_party/terraform/services/container/resource_container_cluster_test.go.erb#L6692

That's `node_pool.node_config.kubelet_config` (as opposed to just `node_config.kubelet_config`).

wyardley:
> An existing test would be ideal (but it doesn't have to be one of the ones already modified in this PR).

@melinath I didn't see any real obvious choice, and the `insecure_...` one won't work because updates, so I made a new one that includes all the other `node_config.kubelet_config` settings, with a note? It could be folded into the insecure kubelet one later if someone were so inclined? Checking that it passes now, but let me know if this approach looks Ok to you.

melinath (approved):

PR #11573, opened 2024-08-29
container: Add `node_kublet_config` support for autopilot clusters

wyardley:
Add support for `node_kubelet_config` in `node_pool_auto_config`.

See:
https://github.com/hashicorp/terraform-provider-google/issues/15208#issuecomment-2299669810

Per:
https://pkg.go.dev/google.golang.org/api/container/v1#NodePoolAutoConfig Currently only `insecure_kubelet_readonly_port_enabled` can be set here. I'm not sure what the future plans are, and whether the provider should allow the other set of flags to be set; otherwise, we may have to adjust some of this so that only supported parameters are allowed.

Fixes https://github.com/hashicorp/terraform-provider-google/issues/19236
Fixes https://github.com/hashicorp/terraform-provider-google/issues/19153

```release-note:enhancement
container: Add `node_kublet_config` support for autopilot clusters.
```

### DIFF
--- a/mmv1/third_party/terraform/services/container/resource_container_cluster.go.erb
+++ b/mmv1/third_party/terraform/services/container/resource_container_cluster.go.erb
@@ -1505,7 +1505,7 @@ func ResourceContainerCluster() *schema.Resource {
 				Description: `Node pool configs that apply to all auto-provisioned node pools in autopilot clusters and node auto-provisioning enabled clusters.`,
 				Elem: &schema.Resource{
 					Schema: map[string]*schema.Schema{
-						"node_kubelet_config": schemaKubeletConfig(),
+						"node_kubelet_config": schemaNodePoolAutoConfigNodeKubeletConfig(),
 						"network_tags": {
 							Type:        schema.TypeList,
 							Optional:    true,
--- a/mmv1/third_party/terraform/services/container/resource_container_cluster.go.erb
+++ b/mmv1/third_party/terraform/services/container/resource_container_cluster.go.erb
@@ -1505,6 +1505,7 @@ func ResourceContainerCluster() *schema.Resource {
 				Description: `Node pool configs that apply to all auto-provisioned node pools in autopilot clusters and node auto-provisioning enabled clusters.`,
 				Elem: &schema.Resource{
 					Schema: map[string]*schema.Schema{
+						"node_kubelet_config": schemaKubeletConfig(),
 						"network_tags": {
 							Type:        schema.TypeList,
 							Optional:    true,
@@ -4403,6 +4404,24 @@ func resourceContainerClusterUpdate(d *schema.ResourceData, meta interface{}) er
 		}
 	}
 
+	if d.HasChange("node_pool_auto_config.0.node_kubelet_config") {
+		req := &container.UpdateClusterRequest{
+			Update: &container.ClusterUpdate{
+				DesiredNodePoolAutoConfigKubeletConfig: expandKubeletConfig(
+					d.Get("node_pool_auto_config.0.node_kubelet_config"),
+				),
+			},
+		}
+
+		updateF := updateFunc(req, "updating GKE cluster node pool auto config node_kubelet_config parameters")
+		// Call update serially.
+		if err := transport_tpg.LockedCall(lockKey, updateF); err != nil {
+			return err
+		}
+
+		log.Printf("[INFO] GKE cluster %s node pool auto config node_kubelet_config parameters have been updated", d.Id())
+	}
+
 	if d.HasChange("node_pool_auto_config.0.network_tags.0.tags") {
 		tags := d.Get("node_pool_auto_config.0.network_tags.0.tags").([]interface{})
 
@@ -5737,6 +5756,10 @@ func expandNodePoolAutoConfig(configured interface{}) *container.NodePoolAutoCon
 	npac := &container.NodePoolAutoConfig{}
 	config := l[0].(map[string]interface{})
 
+	if v, ok := config["node_kubelet_config"]; ok {
+		npac.NodeKubeletConfig = expandKubeletConfig(v)
+	}
+
 	if v, ok := config["network_tags"]; ok && len(v.([]interface{})) > 0 {
 		npac.NetworkTags = expandNodePoolAutoConfigNetworkTags(v)
 	}
@@ -6575,6 +6598,9 @@ func flattenNodePoolAutoConfig(c *container.NodePoolAutoConfig) []map[string]int
 	}
 
 	result := make(map[string]interface{})
+	if c.NodeKubeletConfig != nil {
+		result["node_kubelet_config"] = flattenKubeletConfig(c.NodeKubeletConfig)
+	}
 	if c.NetworkTags != nil {
 		result["network_tags"] = flattenNodePoolAutoConfigNetworkTags(c.NetworkTags)
 	}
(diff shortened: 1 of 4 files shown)


mmv1/third_party/terraform/services/container/resource_container_cluster_test.go.erb
```diff
@@ -10517,6 +10563,41 @@ func testAccContainerCluster_withWorkloadALTSConfigAutopilot(projectID, name str
...
+  return fmt.Sprintf(`
+  resource "google_container_cluster" "with_autopilot_kubelet_config" {
+    name               = "%s"
+    location           = "us-central1"
+    initial_node_count = 1
+
+    node_pool_auto_config {
+      node_kubelet_config {
+        # Needed to work around being required but not actually needed
+        cpu_manager_policy = ""
```
wyardley, line 10588: Not needed if #11572 is shipped; if we can't do that, we may have to add some docs around this or not use the same `kubelet_config` schema.

mmv1/third_party/terraform/services/container/resource_container_cluster.go.erb
```diff
@@ -1505,6 +1505,7 @@ func ResourceContainerCluster() *schema.Resource {
 				Description: `Node pool configs that apply to all auto-provisioned node pools in autopilot clusters and node auto-provisioning enabled clusters.`,
 				Elem: &schema.Resource{
 					Schema: map[string]*schema.Schema{
+						"node_kubelet_config": schemaKubeletConfig(),
```
wyardley, line 1508: Ok, so revert the change breaking out `schemaKubeletConfig` and then create a new schema (`schemaNodePoolAutoConfigNodeKubeletConfig`? in `node_config.go.erb`? -- I think `schemaNodeKubeletConfig` would be potentially confusing since `nodeKubeletConfig` is what `kubeletConfig` is in some places) that just includes `insecure_kubelet_readonly_port_enabled`, right?

melinath (changes requested):

mmv1/third_party/terraform/services/container/resource_container_cluster.go.erb
wyardley: Ok, updated, pushed the changes so far, and running the test locally as well.

mmv1/third_party/terraform/services/container/resource_container_cluster.go.erb
```diff
@@ -6575,6 +6598,9 @@ func flattenNodePoolAutoConfig(c *container.NodePoolAutoConfig) []map[string]int
 	}
 
 	result := make(map[string]interface{})
+	if c.NodeKubeletConfig != nil {
+		result["node_kubelet_config"] = flattenKubeletConfig(c.NodeKubeletConfig)
```
wyardley, line 6602: this needs to be updated to a separate function, or we get a kernel panic here... will push up a fix shortly.

wyardley: With that update:
```
--- PASS: TestAccContainerCluster_withAutopilotKubeletConfig (1921.53s)
PASS
ok  	github.com/hashicorp/terraform-provider-google/google/services/container	1922.494s
```

mmv1/third_party/terraform/website/docs/r/container_cluster.html.markdown
```diff
@@ -1084,6 +1084,13 @@ workload_identity_config {
 
 <a name="nested_node_pool_auto_config"></a>The `node_pool_auto_config` block supports:
 
+* `node_kubelet_config` - (Optional) Kubelet configuration for Autopilot clusters. Currently, only `insecure_kubelet_readonly_port_enabled` is supported here.
+Structure is [documented below](#nested_node_kubelet_config).
+
+<a name="nested_node_kubelet_config"></a>The `node_kubelet_config` block supports:
+
+* `insecure_kubelet_readonly_port_enabled` - (Optional) Controls whether the kubelet read-only port is enabled. It is strongly recommended to set this to `FALSE`. Possible values: `TRUE`, `FALSE`.
```
wyardley, line 1092: @melinath: is this the right way / right place?

Other than docs, I think should be close now, hopefully?

mmv1/third_party/terraform/website/docs/r/container_cluster.html.markdown
```diff
@@ -1084,6 +1084,13 @@ workload_identity_config {
...
+
 * `resource_manager_tags` - (Optional) A map of resource manager tag keys and values to be attached to the nodes for managing Compute Engine firewalls using Network Firewall Policies. Tags must be according to specifications found [here](https://cloud.google.com/vpc/docs/tags-firewalls-overview#specifications). A maximum of 5 tag key-value pairs can be specified. Existing tags will be replaced with new values. Tags must be in one of the following formats ([KEY]=[VALUE]) 1. `tagKeys/{tag_key_id}=tagValues/{tag_value_id}` 2. `{org_id}/{tag_key_name}={tag_value_name}` 3. `{project_id}/{tag_key_name}={tag_value_name}`.
 
 * `network_tags` (Optional) - The network tag config for the cluster's automatically provisioned node pools.
```
melinath, line 1096: Nested blocks need to come _after_ the end of the current block being documented; otherwise, it's ambiguous whether the remaining tags are part of the _current_ block or the _nested_ block. For example, `resource_manager_tags` is a sibling of `node_kubelet_config`, but this makes it look like it's a _child_ of `node_kubelet_config.

```suggestion
* `resource_manager_tags` - (Optional) A map of resource manager tag keys and values to be attached to the nodes for managing Compute Engine firewalls using Network Firewall Policies. Tags must be according to specifications found [here](https://cloud.google.com/vpc/docs/tags-firewalls-overview#specifications). A maximum of 5 tag key-value pairs can be specified. Existing tags will be replaced with new values. Tags must be in one of the following formats ([KEY]=[VALUE]) 1. `tagKeys/{tag_key_id}=tagValues/{tag_value_id}` 2. `{org_id}/{tag_key_name}={tag_value_name}` 3. `{project_id}/{tag_key_name}={tag_value_name}`.

* `network_tags` (Optional) - The network tag config for the cluster's automatically provisioned node pools.

<a name="nested_node_kubelet_config"></a>The `node_kubelet_config` block supports:

* `insecure_kubelet_readonly_port_enabled` - (Optional) Controls whether the kubelet read-only port is enabled. It is strongly recommended to set this to `FALSE`. Possible values: `TRUE`, `FALSE`.
```

mmv1/third_party/terraform/website/docs/r/container_cluster.html.markdown
```diff
@@ -1084,10 +1084,17 @@ workload_identity_config {
...
 
+<a name="nested_node_kubelet_config"></a>The `node_kubelet_config` block supports:
+
+* `insecure_kubelet_readonly_port_enabled` - (Optional) Controls whether the kubelet read-only port is enabled. It is strongly recommended to set this to `FALSE`. Possible values: `TRUE`, `FALSE`.
+
 The `network_tags` block supports:
```
melinath, line 1098:
```suggestion
* `network_tags` (Optional) - The network tag config for the cluster's automatically provisioned node pools. Structure is [documented below](#nested_network_tags).

<a name="nested_node_kubelet_config"></a>The `node_kubelet_config` block supports:

* `insecure_kubelet_readonly_port_enabled` - (Optional) Controls whether the kubelet read-only port is enabled. It is strongly recommended to set this to `FALSE`. Possible values: `TRUE`, `FALSE`.

<a name="nested_network_tags"></a>The `network_tags` block supports:
```

melinath (approved):
LGTM

PR #11575, opened 2024-08-29
Making network field optional in Backupdr MS

niharika-98:

```release-note:enhancement
backupdr: make `network` field optional in `google_backup_dr_management_server` resource
```

### DIFF
--- a/mmv1/products/networkservices/Gateway.yaml
+++ b/mmv1/products/networkservices/Gateway.yaml
@@ -54,7 +54,7 @@ async:
 custom_code:
   constants: 'templates/terraform/constants/network_services_gateway.go.tmpl'
   post_delete: 'templates/terraform/post_delete/network_services_gateway.go.tmpl'
-  pre_update: 'templates/terraform/pre_update/network_services_gateway.tmpl'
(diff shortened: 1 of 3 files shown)


roaks3 (changes requested):
Apologies if there's a limitation I've missed, but could you add a test (or probably better to modify an existing test) to not include these fields? This will help verify that they are in fact not required.

roaks3 (commented):
It looks like there is still a test failure, potentially for another test. But to be clear, I don't think every existing test should be changed (unless you are doing that for another reason, like performance?). I expected there to be one test with this field removed, to prove it is not required.

c2thorn:
@niharika-98 

Tomorrow, the Magic Modules repository is scheduled to undergo a language migration from Ruby to Go. You can view more details about this in our announcement here: https://github.com/hashicorp/terraform-provider-google/issues/19583 (or [go/mm-migration-announcement](go/mm-migration-announcement) if you are a Googler)

This open pull request may become incompatible due to most YAML and .erb files converting to Go-compatible files.

However, the changes seem minor enough that editing the converted files should not be too difficult. The main difficulty will be rebasing after the migration, but I suggest simply opening a new pull request. That should be easier overall.

Let us know if you have any questions or issues.

PR #11577, opened 2024-08-29
Refactor muxing 1 : Re-use same config to configure the SDK and PF providers, fix VCR testing

SarahFrench:
# Summary

This PR:
- Change how the plugin-framework implementation of the provider is configured
- Changes code defining the VCR system to accommodate the change above
    - This also fixes https://github.com/hashicorp/terraform-provider-google/issues/14158
- Makes changes to the data sources currently implemented with the plugin-framework, as the change in the core configuration struct leaks through to each data source currently (a future TODO I guess!)

# Description

Previously when the provider was muxed (i.e. a SDKv2-implemented provider and plugin framework-implemented implemented provider in a trench coat) those two provider implementations were kept very parallel to each other. All the logic for interpreting user inputs to the provider block was duplicated, and that contained errors. Also this introduced the idea of two parallel, independent HTTP clients being used, which means all API interactions for SDKv2-implemented resources/data sources and plugin framework-implemented resources/data sources go via different clients internally. This had the consequence of breaking VCR tests, but only if HTTP traffic was split between the two clients.

This PR avoids the pitfalls of re-implementing core provider logic by re-using the old logic that's been in the provider for years. We use the 'SDKv2 flavoured' code to configure the plugin-framework provider.

This has the advantages of:
- ensuring that the provider's behaviour is largely similar to the SDK provider* 
- resuming all HTTP traffic via a single client, enabling VCR testing to reuse working fully
- matching the approach used in https://github.com/hashicorp/terraform-provider-aws, so we can navigate muxing together 🙏  

This has the disadvantages of:
- The 'SDKv2 flavour' type system enters the plugin-framework code. This is addressed by converting the types uses in the SDKv2 provider `Config` struct into the plugin-frameworks type system as needed.
- In future when the provider becomes 100% plugin-framework implemented that core configuration logic will need to be refactored to use the plugin-frameworks type system (when that time comes, advantages of doing this include better Null/Unknown handling, ability to remove all resulting type conversions in the provider).

# Things to consider

These tests are likely to need attention: https://github.com/GoogleCloudPlatform/magic-modules/pull/11884

```release-note:note
provider: refactored how the provider configuration is handled internally
```


SarahFrench:
There are lots of code deletions I can do as a result of this change, but I'll split that into a separate PR for clarity.

SarahFrench:
Rebased to include Go rewrtie

SarahFrench:
Just rebased to pull in the latest changes from main, including new acc tests for how the provider is configured

SarahFrench:
I'm going to re-make this PR, as it was a bit messy to start but rebases to pull in Go rewrite and new acc tests have made the PR history too long for good review. I'll re-open a new PR with a cleaner commit history

SarahFrench:
See PR description for replacement PR!

PR #11578, opened 2024-08-29
Refactor muxing 2 : Remove unused code, resulting from #11903

SarahFrench:
Follow up to https://github.com/GoogleCloudPlatform/magic-modules/pull/11903. ~Most commits in this PR are in that PR, and will 'disappear' once this PR is unblocked and ready for review.~ This PR has been rebased to address this!

This PR deletes all the parallel implementations of provider configuration logic. The PR https://github.com/GoogleCloudPlatform/magic-modules/pull/11903 stops the provider using this duplicated logic, and this PR removes that redundant code.

In future the team may want to revisit this deleted code to inform work to pulls out the SDKv2 from the codebase completely and the provider becomes 100% plugin-framework. As that is a long way in the future I think it's best to delete this code now for clarity. It will still be in version control for reference in the future.

When looking at the deleted code be wary of:
- Functions with the same name as SDK code that are functions instead of methods on a struct
   - This can limit reusability of functions, e.g: https://github.com/GoogleCloudPlatform/magic-modules/pull/12141#discussion_r1844342397
-  Functions with the same name as SDK code but don't include the same logic

```release-note:none

```


SarahFrench:
Rebased to pull in Go rewrite

SarahFrench:
Thanks, I'll go ahead and merge as I don't think there's a need for multiple approvals here vs the first PR

PR #11579, opened 2024-08-29
chore: ignore `mmv1/vendor` path in `.gitignore`

wyardley:
From what I can see, the setup process for now still involves installing ruby libs that potentially get installed in `mmv1/vendor`. Feel free to close if my setup is outdated and this is not needed, but I think it should be harmless.

Add these to `.gitignore`. It doesn't appear that a trailing `*` is needed the way `mmv1/build/*` is, so I left that, but didn't add it here

```release-note:none

```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -42,6 +42,9 @@ final_api.yaml
 # Used by OiCS tutorial
 mmv1/build/*
 
+# Ignore vendored Ruby libs as well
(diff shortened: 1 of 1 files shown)


c2thorn:
We are planning to switch over to Go from Ruby within the next two weeks, so hopefully this will no longer be needed.

PR #11581, opened 2024-08-29
Fix post-6.0.0 TGC issues

shuyama1:

```release-note:none

```


shuyama1:
@melinath @c2thorn Sorry, this PR is still WIP.  I'm only adding reviewers to prevent it from assigning reviewers from service teams. I'll mark it as ready for review once all TGC tests have passed.

shuyama1:
@melinath It seems that once a service label is added, it remains even if the changes in that service package are reverted. I'm not sure if this is the intended behavior.

c2thorn (approved):
thank you!

melinath:
> @melinath It seems that once a service label is added, it remains even if the changes in that service package are reverted. I'm not sure if this is the intended behavior, just want to flag.

yes, this is the intended behavior. More specifically, if any service label is present, the CI pipeline will not modify the service labels. This allows us to do things like manual overrides if necessary.

PR #11558, opened 2024-08-27
Add "goog-terraform-provisioned": "true" to all TGC JSON files that already have labels in them

SarahFrench:
**Release Note Template for Downstream PRs (will be copied)**

```release-note:none

```

PR #11584, opened 2024-08-29
Made tgc test data change trigger tgc test in CI

shuyama1:
Currently, changes made solely in the [mmv1/third_party/tgc/tests/data](https://github.com/GoogleCloudPlatform/magic-modules/tree/main/mmv1/third_party/tgc/tests/data) will result in TGC tests being skipped during PR checks. This could lead to false positives in TGC-specific test fix PRs, as seen in https://github.com/GoogleCloudPlatform/magic-modules/pull/11558

```release-note:none

```

### DIFF
--- a/.github/workflows/test-tgc.yml
+++ b/.github/workflows/test-tgc.yml
@@ -67,7 +67,7 @@ jobs:
     - name: Check for Code Changes
       id: pull_request
       run: |
-        gofiles=$(git diff --name-only HEAD~1 | { grep -e "\.go$" -e "go.mod$" -e "go.sum$" || test $? = 1; })
+        gofiles=$(git diff --name-only HEAD~1 | { grep -e "\.go$" -e "go.mod$" -e "go.sum$" -e "tfplan2cai/testdata/templates/" || test $? = 1; })
         if [ -z "$gofiles" ]; then
           echo "has_changes=false" >> $GITHUB_OUTPUT
         else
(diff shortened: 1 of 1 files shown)


.github/workflows/test-tgc.yml
```diff
@@ -67,7 +67,7 @@ jobs:
     - name: Check for Code Changes
       id: pull_request
       run: |
-        gofiles=$(git diff --name-only HEAD~1 | { grep -e "\.go$" -e "go.mod$" -e "go.sum$" || test $? = 1; })
+        gofiles=$(git diff --name-only HEAD~1 | { grep -e "\.go$" -e "go.mod$" -e "go.sum$" -e "tfplan2cai/testdata/templates/" || test $? = 1; })
```
melinath, line 70: possibly anything in testdata?

```suggestion
        gofiles=$(git diff --name-only HEAD~1 | { grep -e "\.go$" -e "go.mod$" -e "go.sum$" -e "tfplan2cai/testdata/" || test $? = 1; })
```

melinath (changes requested):

melinath (approved):
meant to approve, whoops!

.github/workflows/test-tgc.yml
shuyama1: yeah, that makes sense! Thanks for the quick review

shuyama1:
The TGC failure is expected as it's already failed at head. The fix of the failure is in https://github.com/GoogleCloudPlatform/magic-modules/pull/11581

PR #11589, opened 2024-08-30
continue sweeper run for other resources if no resource found

Mehul3217:

```release-note:none
netapp: modified manual sweepers to continue or other locations if no resource found
```


melinath (changes requested):
All of the other `return nil` calls in all of these sweepers should also use `continue` instead

Mehul3217:
> All of the other `return nil` calls inside `for` loops in all of these sweepers should also use `continue` instead

Done

PR #11591, opened 2024-08-30
Add tags field to Alloydb Cluster for TagsR2401

SamanthaMathews:
Add tags field to cluster resource to allow setting tags on cluster resources at creation time.
b/365030846

```
alloydb: added `tags` field to `alloydb_cluster` to allow setting tags for clusters at creation time


roaks3 (commented):
@SamanthaMathews it looks like this is still in draft mode, so just let me know when you are ready for a review

c2thorn:

PR #11592, opened 2024-08-30
Dataproc Cluster: add more nil handling following API-side change

SarahFrench:

```release-note:bug
dataproc: updated `google_dataproc_cluster` to protect against handling nil KerberosConfig values
```

### DIFF
--- a/mmv1/third_party/terraform/services/dataproc/resource_dataproc_cluster.go
+++ b/mmv1/third_party/terraform/services/dataproc/resource_dataproc_cluster.go
@@ -2883,6 +2883,10 @@ func flattenSecurityConfig(d *schema.ResourceData, sc *dataproc.SecurityConfig) 
 	if sc == nil {
 		return nil
 	}
+	// API side change means that sc can be non-nil but contain only nils
+	if sc.IdentityConfig == nil && sc.KerberosConfig == nil {
+		return nil
+	}
 	data := map[string]interface{}{
 		"kerberos_config": flattenKerberosConfig(d, sc.KerberosConfig),
 	}
--- a/mmv1/third_party/terraform/services/dataproc/resource_dataproc_cluster.go
+++ b/mmv1/third_party/terraform/services/dataproc/resource_dataproc_cluster.go
@@ -2884,7 +2884,7 @@ func flattenSecurityConfig(d *schema.ResourceData, sc *dataproc.SecurityConfig) 
 		return nil
 	}
 	// API side change means that sc can be non-nil but contain only nils
-	if sc.IdentityConfig == nil && sc.KerberosConfig == nil {
+	if sc.KerberosConfig == nil {
 		return nil
 	}
 	data := map[string]interface{}{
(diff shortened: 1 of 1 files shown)


mmv1/third_party/terraform/services/dataproc/resource_dataproc_cluster.go
```diff
@@ -2883,6 +2883,10 @@ func flattenSecurityConfig(d *schema.ResourceData, sc *dataproc.SecurityConfig)
 	if sc == nil {
 		return nil
 	}
+	// API side change means that sc can be non-nil but contain only nils
+	if sc.IdentityConfig == nil && sc.KerberosConfig == nil {
+		return nil
+	}
```
SarahFrench, line 2889: Might be better focusing on only sc.KerberosConfig == nil, as that's the only one used. Might fall through if IdentityConfig is non nil but sc.KerberosConfig is

SarahFrench:
When I've run `TestAccDataprocCluster_basic` locally it failed due to the panic.

When I've run the test with this PR's changes I see this check error as a cause of test failure:


```
    resource_dataproc_cluster_test.go:65: Step 1/1 error: Check failed: Check 3/20 error: google_dataproc_cluster.basic: Attribute 'cluster_config.0.gce_cluster_config.0.internal_ip_only' expected "false", got "true"
```

mmv1/third_party/terraform/services/dataproc/resource_dataproc_cluster.go
roaks3: Yea, I think we may want to just make `flattenKerberosConfig` handle nil values passed in. Looking at the API logic, there seems to be multiple combinations that could be sent down (including a new field that hasn't been made visible yet), and it might be easier to maintain one level down.

SarahFrench:
VCR cassettes likely are from before the API change - I've triggered a build running all dataproc tests on this PR here:https://hashicorp.teamcity.com/buildConfiguration/TerraformProviders_GoogleCloud_GOOGLE_BETA_MMUPSTREAMTESTS_GOOGLEBETA_PACKAGE_DATAPROC/226840?buildTab=tests&branch=refs%2Fheads%2Fauto-pr-11592

SarahFrench:
TestAccDataprocCluster_basic failing for a similar reason in TeamCity, and looking at the TC history I can see that failure is newly introduced by this PR, which requires some investigation. Yet to see how other tests look.

mmv1/third_party/terraform/services/dataproc/resource_dataproc_cluster.go
SarahFrench: Given the tests have stopped running on the PR I'll push a fix, and once the current TC build ends we can trigger a more scoped-down build that runs only the relevant tests

SarahFrench:
`flattenSoftwareConfig` could be updated in a similar way, and there are other flatteners that aren't written defensively. However it's hard to test how the provider would behave if those flatteners were updated.

roaks3 (approved):

shuyama1:
Is it worth backporting this change? I assume the service team will re-release API-side change, and after that, the provider might start crashing for versions 5.x.

PR #11593, opened 2024-08-30
Final* go rewrite services

NickElliot:
All remaining services

a good number of hand written files needed minor syncing for services that were updated since the handwritten templates were batch converted. but since a lot of handwritten files have been minorly updated for whitespace, odds are we will need to re-sync them by hand for ones that need to be re-updated

I think the large yaml regeneration is the next step here, but hopefully would not result in any "new" diffs from the minor ones we accepted as inevitable

```release-note:none

```


c2thorn (approved):
LGTM assuming tests pass

NickElliot:
While waiting on the tests I wanted to call attention to a specific change: https://github.com/GoogleCloudPlatform/magic-modules/pull/11593/files#diff-d4257ecf68b40cde4cd0e6f0b557a06b872641c57dbda2bcc459c1a6691013ff

This appeared to satisfy the new issue I ran into from your original change (https://github.com/GoogleCloudPlatform/magic-modules/commit/bd00206ccbcd06c9e26091a946b8e116863c7a3c#diff-d4257ecf68b40cde4cd0e6f0b557a06b872641c57dbda2bcc459c1a6691013ff) that was a deviation from the original Ruby logic, but because your version caused a change in a resource I went over, it potentially also introduced diffs in other resources. Hopefully the way I altered it solved the fundamental edge case that necessitated it, but something to keep in mind if we see new diffs after syncing it all

NickElliot:
test failure looks to be a problem of all the netapp tests being ran together

c2thorn:
This should be good to go

PR #11594, opened 2024-08-30
Update the documentation for gkehub_feature_membership

haiyanmeng:

```release-note:enhancement
gkehub: added documentation for the `configmanagement.management` and `configmanagement.config_sync.enabled` fields to `google_gke_hub_feature_membership` resource
```


rileykarson (approved):
LGTM

We don't normally call out specific versions in these docs (there's a global version selector in the Registry), but this seems fine given that we're adding the docs late + only a few 5.X versions close to the end of the 5.X series have the feature.

haiyanmeng:
@rileykarson , thanks for the explanation. I didn't realize that I need to manually update the doc for `gkehub_feature_membership`. In the future I will make sure the code change and doc change are in the same PR.

PR #11596, opened 2024-08-30
Dataproc - `kerberosConfig` is optional, not required.

Harwayne:
`kerberosConfig` is optional, not required.

Add a test for the private `flattenSecurityConfig` function to ensure a missing `kerberosConfig` does not lead to a nil-pointer error. Because the test is in a different package, create `export_test.go`, which exports private functions from the main `dataproc` package to any tests within the folder.

```release-note:bug
dataproc: `KerberosConfig` is optional within the `SecurityConfig` message.
```

### DIFF
--- a/mmv1/third_party/terraform/services/dataproc/resource_dataproc_cluster.go
+++ b/mmv1/third_party/terraform/services/dataproc/resource_dataproc_cluster.go
@@ -1243,7 +1243,7 @@ func ResourceDataprocCluster() *schema.Resource {
 								Schema: map[string]*schema.Schema{
 									"kerberos_config": {
 										Type:        schema.TypeList,
-										Required:    true,
+										Optional:    true,
 										Description: "Kerberos related configuration",
 										MaxItems:    1,
 										Elem: &schema.Resource{
@@ -2883,8 +2883,10 @@ func flattenSecurityConfig(d *schema.ResourceData, sc *dataproc.SecurityConfig) 
 	if sc == nil {
 		return nil
 	}
-	data := map[string]interface{}{
-		"kerberos_config": flattenKerberosConfig(d, sc.KerberosConfig),
+	data := map[string]interface{}{}
+
+	if kfg := sc.KerberosConfig; kfg != nil {
+		data["kerberos_config"] = flattenKerberosConfig(d, kfg)
 	}
 
 	return []map[string]interface{}{data}
--- a/mmv1/third_party/terraform/services/dataproc/export_test.go.erb
+++ b/mmv1/third_party/terraform/services/dataproc/export_test.go.erb
@@ -1,0 +1,8 @@
+<% autogen_exception -%>
+package dataproc
+
+// This file is used to export private functions purely for the unit tests, which are in a
+// different package (dataproc vs dataproc_test).
+
+var FlattenSecurityConfig = flattenSecurityConfig
+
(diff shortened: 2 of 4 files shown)


mmv1/third_party/terraform/services/dataproc/resource_dataproc_cluster.go
```diff
@@ -1243,7 +1243,7 @@ func ResourceDataprocCluster() *schema.Resource {
 								Schema: map[string]*schema.Schema{
 									"kerberos_config": {
 										Type:        schema.TypeList,
-										Required:    true,
+										Optional:    true,
```
shuyama1, line 1246: I believe `kerberos_config` was marked as required initially since this is the only subfield under `security_config`. We'd generally want to prevent sending empty blocks, like `security_config{}`, unless explicitly sending `security_config{}` to the API is different from not sending the field at all. 

Do we want to allow users to explicitly specify `security_config{}` in this case?

Harwayne: > I believe kerberos_config was marked as required initially since this is the only subfield under security_config

That is true historically. But other fields have been added since then. 

> We'd generally want to prevent sending empty blocks, like security_config{}, unless explicitly sending security_config{} to the API is different from not sending the field at all.
>
> Do we want to allow users to explicitly specify security_config{} in this case?

I don't think Dataproc's service will differentiate between an empty and an absent `security_config`. 

The flatten side is what prompted this PR, as there are in-development hidden fields being added to `security_config`. Dataproc would return `security_config{}`, which terraform would try to flatten, resulting in errors. I made the change to the schema mostly for consistency.

shuyama1 (commented):
The test failed because the API currently does not return security_config{}. The resource configuration explicitly sends an empty block to the API, which results in a  [permadiff](https://googlecloudplatform.github.io/magic-modules/develop/permadiff/). This is likely why the subfield was marked as required in the first place, to prevent users from sending the empty block. I'm still doing some local testing to determine the best approach to proceed with this fix based on our offline discussion.

roaks3:
Note: I hadn't seen this PR unfortunately, but https://github.com/GoogleCloudPlatform/magic-modules/pull/11592 was just merged to guard against the nil error from the crash. Could you please rebase and check that it works with your changes, and revert as superseded by this change if needed? I suspect that it could replace your `flattenSecurityConfig` changes, by using a more standard flattener paradigm (specifically, that most flatteners check for nil being passed in, and return nil in that case).

Harwayne:
> Note: I hadn't seen this PR unfortunately, but #11592 was just merged to guard against the nil error from the crash. Could you please rebase and check that it works with your changes, and revert as superseded by this change if needed? I suspect that it could replace your `flattenSecurityConfig` changes, by using a more standard flattener paradigm (specifically, that most flatteners check for nil being passed in, and return nil in that case).

That looks like it will work for our immediate fix, superseding this PR.

However, iIn the medium-term future, we intend to add more fields to `securityConfig` and expect users will fill in those new fields, omitting `kerberosConfig`. Is it possible to change the schema to say 'at least one of these fields must be set'?

Harwayne:
> yes, we can always change required -> optional/ at least of if we add more fields to the schema in the future. This is non-breaking and can happen in minor release.

Great, I think that was my only remaining concern. I'll close the PR.

PR #11598, opened 2024-08-30
Bootstrap storage pool in tests

shuyama1:
Given the [limitations of storage pools per project](https://cloud.google.com/compute/docs/disks/storage-pools#sp_limitations), we'll need to bootstrap the storage pool resource in our testing project so it can be shared and reused across multiple tests to avoid hitting the quota limit.

This should help https://github.com/hashicorp/terraform-provider-google/issues/19199 and https://github.com/GoogleCloudPlatform/magic-modules/pull/11391

```release-note:none

```

### DIFF
--- a/mmv1/third_party/terraform/acctest/bootstrap_test_utils.go
+++ b/mmv1/third_party/terraform/acctest/bootstrap_test_utils.go
@@ -1113,7 +1113,7 @@ func BootstrapComputeStoragePool(t *testing.T, storagePoolName, storagePoolType 
 	projectID := envvar.GetTestProjectFromEnv()
 	zone := envvar.GetTestZoneFromEnv()
 
-	storagePoolName = SharedStoragePoolPrefix + storagePoolName
+	storagePoolName = SharedStoragePoolPrefix + storagePoolType + "-" + storagePoolName
 
 	config := BootstrapConfig(t)
 	if config == nil {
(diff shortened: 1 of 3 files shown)


shuyama1:
failing test `TestAccComputeInstanceTemplate_withNamePrefix` is not related to this PR

mmv1/third_party/terraform/acctest/bootstrap_test_utils.go
```diff
@@ -1107,6 +1107,78 @@ func BootstrapFirewallForDataprocSharedNetwork(t *testing.T, firewallName string
 	return firewall.Name
 }
 
+const SharedStoragePoolPrefix = "tf-bootstrap-storage-pool-"
+
+func BootstrapComputeStoragePool(t *testing.T, storagePoolName, storagePoolType string) string {
+	projectID := envvar.GetTestProjectFromEnv()
+	zone := envvar.GetTestZoneFromEnv()
+
+	storagePoolName = SharedStoragePoolPrefix + storagePoolName
```
melinath, line 1116: I think it would make sense to include the storage pool type here - or only allow storagePoolType and remove storagePoolName if we only need one of each type per region.

```suggestion
func BootstrapComputeStoragePool(t *testing.T, storagePoolType string) string {
	projectID := envvar.GetTestProjectFromEnv()
	zone := envvar.GetTestZoneFromEnv()

	storagePoolName := SharedStoragePoolPrefix + storagePoolType
```

melinath (approved):

mmv1/third_party/terraform/acctest/bootstrap_test_utils.go
shuyama1: I'll update it to `storagePoolName = SharedStoragePoolPrefix + storagePoolType + storagePoolName` to make it easier to identify the type by reading the resource name. I'll also keep the storagePoolName, as we might need more than one storage pool per type as test cases expand.

melinath (approved):

PR #11599, opened 2024-08-31
Use acceptance tests to test handling of the credentials provider configuration argument, add new data source that surfaces configuration of SDK and plugin-framework providers to facilitate acctests

SarahFrench:
# Description

This PR is the first of several PRs that remove old unit tests that define how the provider behaves when it's configured in different ways and replaces them with acceptance tests.

This PR is the first, so it also adds 2 new data sources:
- `google_provider_config_sdk`
- `google_provider_config_plugin_framework`

These new data sources are necessary for a number of reasons:

- **Enabling direct testing**: Some parts of provider configuration are easier to test indirectly than others, and some are impossible to test indirectly. For example, https://github.com/GoogleCloudPlatform/magic-modules/pull/11563 shows issues of testing `credentials` indirectly. Another example, the user agent set on the provider isn't able to be tested directly.  By creating a data source that exposes the values in the provider we can test them directly.
- **Providing access to info about both the SDK and PF parts of the codebase**: There is an existing data source that exposes some of these values already (google_client_config) but this is insufficient for implementing all acceptance tests because we need access to a data sources implemented with both the SDK and plugin-framework.

We ideally do not want users to start using these data sources because they would be removed in a future version of the provider when the SDK is wholly removed (a long way in the future). To enforce this, I have made the data sources only be registered on the provider in the context of acceptance tests ([a8c3757](https://github.com/GoogleCloudPlatform/magic-modules/pull/11599/commits/a8c375712b0404a2fd409943d82c92cd0aabbe14), [506c2a4](https://github.com/GoogleCloudPlatform/magic-modules/pull/11599/commits/506c2a4c95e3fcdc6c9b8831f65863f10c039834)). This makes them inaccessible to users.

## Test changes

This PR deletes existing tests, and replaces them with new acceptance tests that make the same assertions about the provider's behaviour.

## Motivation

These changes are made so that the way we test the provider's configuration behaviour is independent of how they're implemented. This will allow us to have a test suite in place before review and merging of this PR (https://github.com/GoogleCloudPlatform/magic-modules/pull/11577) that fixes a lot of issues related to muxing in the Google provider.

```release-note:none
```


SarahFrench:
TeamCity builds for VCR-skipped tests:
- [SDK Provider - Acceptance Tests](https://hashicorp.teamcity.com/buildConfiguration/TerraformProviders_GoogleCloud_GOOGLE_BETA_MMUPSTREAMTESTS_GOOGLEBETA_PACKAGE_PROVIDER/227554?hideTestsFromDependencies=false&hideProblemsFromDependencies=false&expandBuildChangesSection=true)
- [Framework Provider - Acceptance Tests](https://hashicorp.teamcity.com/buildConfiguration/TerraformProviders_GoogleCloud_GOOGLE_BETA_MMUPSTREAMTESTS_GOOGLEBETA_PACKAGE_FWPROVIDER/227555?mode=branches&buildTab=overview&hideTestsFromDependencies=false&hideProblemsFromDependencies=false&expandBuildChangesSection=true)

SarahFrench:
Hi @c2thorn & @NickElliot - requesting you as reviewers because I was told that you'll be doing some plugin-framework related work in future. This PR is meant to enable testing the provider configuration logic while allowing that logic to be refactored, so figured it'd be appropriate to ask you both to be reviewers.

mmv1/third_party/terraform/acctest/vcr_utils.go
```diff
@@ -436,6 +438,14 @@ func (p *frameworkTestProvider) Configure(ctx context.Context, req provider.Conf
 	}
 }
 
+// DataSources overrides the provider's DataSources function so that we can append test-specific data sources to the list of data sources on the provider.
+// This makes the data source(s) usable only in the context of acctests, and isn't available to users
+func (p *frameworkTestProvider) DataSources(ctx context.Context) []func() datasource.DataSource {
+	ds := p.FrameworkProvider.DataSources(ctx)
+	ds = append(ds, fwprovider.NewGoogleProviderConfigPluginFrameworkDataSource) // google_provider_config_plugin_framework
```
c2thorn, line 445: Pretty neat way to do this. Just to confirm, attempting to reference the datasource outside of a test will just show it as not available/expected?

mmv1/third_party/terraform/fwprovider/data_source_provider_config_plugin_framework.go
```diff
@@ -0,0 +1,231 @@
...
+
+type GoogleProviderConfigPluginFrameworkDataSource struct {
+	providerConfig *fwtransport.FrameworkProviderConfig
+}
+
+type GoogleProviderConfigPluginFrameworkModel struct {
+	// Currently this reflects the FrameworkProviderConfig struct and ProviderModel in google/fwmodels/provider_model.go
+	// which means it uses the plugin-framework type system where values can be explicitly Null or Unknown.
+	//
+	// As part of future muxing fixes/refactoring we'll change this struct to reflect structs used in the SDK code, and will move to
```
c2thorn, line 35: So this is saying one of the anticipated fixes for existing pf/sdk muxing issues is to rewrite the model in `google/fwmodels/provider_model.go`, which means we will also need to change it here too ?

c2thorn (approved):
LGTM - assuming the acceptance tests have passed locally since they are VCR skipped

SarahFrench:
I ran [the tests in TeamCity previously](https://github.com/GoogleCloudPlatform/magic-modules/pull/11599#issuecomment-2322896148) and they were ok - I'll re run them before merging to double check the latest version of the PR

mmv1/third_party/terraform/acctest/vcr_utils.go
SarahFrench: I'll manually check, but this code is only reached in the context of testing, with the entrypoint being ProtoV5ProviderFactories, which calls MuxedProviders, which calls the vcr_utils code

mmv1/third_party/terraform/fwprovider/data_source_provider_config_plugin_framework.go
SarahFrench: The fixes for the muxing issues is to replace use of FrameworkProviderConfig with using the Config struct that is used in the SDK code. That means eventually the google_provider_config_plugin_framework data source will receive all the input data from a Config struct and that's implemented with the SDK type system (string, vs types.StringType). But before I do that I need to get these tests in place using the old struct.

Basically this comment is me saying I'm implementing this all one way, and will immediately switch it up as soon as all the acceptance tests are in place and make it easier for someone to review the muxing fixes.

SarahFrench:
Confirmed that users won't be able to access these data sources:

Here's what I see if I build the provider from f2ce3e6 and attempt a plan using this config

```tf
data "google_provider_config_plugin_framework" "test" {}

data "google_provider_config_sdk" "test" {}
```

Screenshot 2024-09-05 at 10 01 23

and when I use the latest commit on this PR after the provider was made inaccessible except in acctests this is the result instead:

Screenshot 2024-09-05 at 10 09 18

SarahFrench:
From doing the above I realised the id of the SDK-implemented datasource has an encoding problem so I'm going to change it to something UTF-8 friendly

SarahFrench:
OK going ahead with merge 🎉

PR #11602, opened 2024-09-01
Adds support for Filestore instance new feature: deletion protection.

boarik:
Users will be able to enable **deletion protection** on their new or existing Filestore instances by providing an instance's `deletion_protection_enabled` boolean field and optionally also `deletion_protection_reason` string field specifying the reason for enabling the protection.

Fixes https://github.com/hashicorp/terraform-provider-google/issues/19336

```release-note:enhancement
filestore: added `deletion_protection_enabled` and `deletion_protection_reason` fields to `google_filestore_instance`
```

### DIFF
--- a/mmv1/third_party/terraform/services/filestore/resource_filestore_instance_test.go
+++ b/mmv1/third_party/terraform/services/filestore/resource_filestore_instance_test.go
@@ -212,10 +212,16 @@ func TestAccFilestoreInstance_deletionProtection_update(t *testing.T) {
 				ResourceName:            "google_filestore_instance.instance",
 				ImportState:             true,
 				ImportStateVerify:       true,
-				ImportStateVerifyIgnore: []string{"zone", "location"},
+				ImportStateVerifyIgnore: []string{"zone"},
 			},
 			{
 				Config: testAccFilestoreInstance_deletionProtection_update(name, location, tier, deletionProtection),
+			},
+			{
+				ResourceName:            "google_filestore_instance.instance",
+				ImportState:             true,
+				ImportStateVerify:       true,
+				ImportStateVerifyIgnore: []string{"zone"},
 			},
 			{
 				Config: testAccFilestoreInstance_deletionProtection_update(name, location, tier, !deletionProtection),
@@ -224,7 +230,7 @@ func TestAccFilestoreInstance_deletionProtection_update(t *testing.T) {
 				ResourceName:            "google_filestore_instance.instance",
 				ImportState:             true,
 				ImportStateVerify:       true,
-				ImportStateVerifyIgnore: []string{"zone", "location"},
+				ImportStateVerifyIgnore: []string{"zone"},
 			},
 		},
 	})
--- a/mmv1/third_party/terraform/services/filestore/resource_filestore_instance_test.go
+++ b/mmv1/third_party/terraform/services/filestore/resource_filestore_instance_test.go
@@ -190,3 +190,92 @@ resource "google_filestore_instance" "instance" {
 }
 `, name)
 }
+
+func TestAccFilestoreInstance_deletionProtection_update(t *testing.T) {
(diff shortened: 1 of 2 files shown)


boarik:
The failed VCR-test is due to API rollout changes that are in progress (should be finished very soon) and due to experiments not yet being fully ramped up yet. I'd appreciate a review regardless to move things forward in the meantime.
Thanks

boarik:
@rileykarson The experiments + API rollout dependencies that caused the test failure should be now resolved. Can you assist re-running the checks to validate all works?

mmv1/third_party/terraform/services/filestore/resource_filestore_instance_test.go
```diff
@@ -190,3 +190,92 @@ resource "google_filestore_instance" "instance" {
...
+			},
+			{
+				ResourceName:            "google_filestore_instance.instance",
+				ImportState:             true,
+				ImportStateVerify:       true,
+				ImportStateVerifyIgnore: []string{"zone", "location"},
+			},
+			{
+				Config: testAccFilestoreInstance_deletionProtection_update(name, location, tier, deletionProtection),
+			},
```
rileykarson, line 219:
```suggestion
			},
			{
				ResourceName:            "google_filestore_instance.instance",
				ImportState:             true,
				ImportStateVerify:       true,
				ImportStateVerifyIgnore: []string{"zone", "location"},
			},
```

We should generally use `ImportState` steps between every step unless there's a specific reason we're avoiding doing so!

mmv1/third_party/terraform/services/filestore/resource_filestore_instance_test.go
```diff
@@ -190,3 +190,92 @@ resource "google_filestore_instance" "instance" {
...
+			{
```
rileykarson, line 220: optional: We _could_ add an empty config here as a TestStep with ExpectError that we fail on deletion due to deletion protection, but that's really just exercising the API feature and we'll confirm that the value is set with the ImportState suggested above already. Feel free to dismiss, though.

rileykarson (commented):
Mostly LGTM- one small test change to add an ImportState step after one of the Config steps.

I did want to flag that if we add a clientside deletion protection in the future we'll need to reconcile the fields [[our thoughts on deletion_protection]](https://googlecloudplatform.github.io/magic-modules/best-practices/#deletion_protection). Given we don't have one I think this is still the right change to make anyways.

mmv1/third_party/terraform/services/filestore/resource_filestore_instance_test.go
boarik: Thanks for the input. Can I please ask why is that so? It seems we don't test the state itself rather the fact that it exists so I don't fully understand why this is done.

mmv1/third_party/terraform/services/filestore/resource_filestore_instance_test.go
boarik: Added the importState suggested above.

boarik (commented):
@rileykarson back to you. Thanks

mmv1/third_party/terraform/services/filestore/resource_filestore_instance_test.go
rileykarson: `ImportStateVerify` does a full deep comparison of the state after the apply and a fresh state created during import. It catches pretty much all non-adversarial issues with reading or setting values (for example, hardcoding a specific value wouldn't cause an issue- but would be caught during review).

In the past we wrote "TestCheck" checks for specific fields, but they tended to be brittle / took extra code to write (in particular, needing a distinct API fetch per resource, including a fair bit of flexibility in auth). Import-driven checks superseded them.

rileykarson (approved):

PR #11606, opened 2024-09-02
Add diff suppress for storage_config.bucket field

spapi17:

```release-note:enhancement
composer: added diff suppress to `storage_config.bucket` field (users should be able to specify bucket name with or without "gs://" prefix). 
```

### DIFF
--- a/mmv1/third_party/terraform/services/composer/resource_composer_environment.go.erb
+++ b/mmv1/third_party/terraform/services/composer/resource_composer_environment.go.erb
@@ -1078,7 +1078,7 @@ func ResourceComposerEnvironment() *schema.Resource {
 							Type:         schema.TypeString,
 							Required:     true,
 							ForceNew:     true,
-							DiffSuppressFunc: tpgresource.CompareResourceNames,
+							DiffSuppressFunc: gscBucketNameDiffSuppress,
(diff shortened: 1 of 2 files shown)


mmv1/third_party/terraform/services/composer/resource_composer_environment.go.erb
```diff
@@ -1078,6 +1078,7 @@ func ResourceComposerEnvironment() *schema.Resource {
 							Type:         schema.TypeString,
 							Required:     true,
 							ForceNew:     true,
+							DiffSuppressFunc: tpgresource.CompareResourceNames,
```
rileykarson, line 1081: Would you be able to add a dedicated function to deal with the gcs prefix to tpgresource? This would work but mostly by coincidence- because we strip after the last `/`.

rileykarson:
@hao-nan-li do you mind picking this review up? Thanks!

hao-nan-li:
> @hao-nan-li do you mind picking this review up? Thanks!

sure

hao-nan-li (commented):
Could you add this field to an existing test or add a new test to cover the `bucket name with "gs://" prefix`?

hao-nan-li (approved):

PR #11607, opened 2024-09-02
Add acceptance tests for how provider handles `project` arguments

SarahFrench:
This PR adds acc tests to test how the provider handles the project argument, using the new datasources `google_provider_config_plugin_framework` and `google_provider_config_sdk`.

In the existing unit tests there is a test case in the `TestFrameworkProvider_LoadAndValidateFramework_project` test called `"when project is an unknown value, the provider treats it as if it's unset and uses an environment variable instead"`.

I added a test case called `"when project is unknown in the config, environment variables are used"` in the new SDK and PF acc tests. However the test case is flakey and has different outcomes on different runs. Because of this I'm going to leave testing unknown handling to unit tests only.

```release-note:none

```


SarahFrench:
Weird that TestAccFwProvider_project never ran

SarahFrench:
[This failure is weird](https://github.com/GoogleCloudPlatform/magic-modules/pull/11607#issuecomment-2325045670), as I can get the original version of the acctest's checks to pass locally. I assume that there's a difference in the CI environment and my laptop that affects the data source being refreshed or not? I assume the plugin-framework version won't be as flaky as unknown handling is more explicit and less of "lucky timing" where a data source refreshes before you try to use its attributes in dependent resources

mmv1/third_party/terraform/fwprovider/framework_provider_project_test.go
```diff
@@ -0,0 +1,257 @@
...
+				// Unset all ENVs for project
+				PreConfig: func() {
+					for _, v := range envvar.ProjectEnvVars {
+						t.Setenv(v, "")
+					}
+				},
+				Config: testAccFwProvider_projectInEnvsOnly(),
+				Check: resource.ComposeTestCheckFunc(
+					// Differing behavior between SDK and PF; the attribute is NOT found here.
+					// This reflects the different type systems used in the SDKv2 and the plugin-framework
```
SarahFrench, line 85: Differing behaviour  = In PF tests the field is missing when there is no value for it, but in the SDK the field is set to an empty string.

mmv1/third_party/terraform/provider/provider_project_test.go
```diff
@@ -0,0 +1,257 @@
...
+				// Unset all ENVs for project
+				PreConfig: func() {
+					for _, v := range envvar.ProjectEnvVars {
+						t.Setenv(v, "")
+					}
+				},
+				Config: testAccSdkProvider_projectUnknownHandling(context),
+				Check: resource.ComposeAggregateTestCheckFunc(
+					// Matches project id output from google_project resource
+					resource.TestMatchResourceAttr("data.google_provider_config_sdk.default", "project", regexp.MustCompile(fmt.Sprintf("tf-test-[0-9a-z]{16}|%s", project))),
```
SarahFrench, line 204: There's some flakiness I see when running this locally vs on the PR

SarahFrench:
Opening as ready for review, but I think the unknown value handling test case will require some discussion

mmv1/third_party/terraform/provider/provider_project_test.go
```diff
@@ -0,0 +1,257 @@
...
+		ProtoV5ProviderFactories: acctest.ProtoV5ProviderFactories(t),
+		ExternalProviders: map[string]resource.ExternalProvider{
+			"random": {},
+		},
+		Steps: []resource.TestStep{
+			{
+				Config: testAccSdkProvider_projectUnknownHandling(context),
+				Check: resource.ComposeAggregateTestCheckFunc(
+					// Matches ENV instead of the project id output from google_project resource
+					resource.TestCheckResourceAttr("data.google_provider_config_sdk.default", "project", project),
```
zli82016, line 191: Sorry, I am lost somewhere. 
Can you help me understand why `data.google_provider_config_sdk.default.project` should be project here?
 
My understanding is that `data.google_provider_config_sdk.default.project` should be the project in the provider `google`.
The project in the provider `google` should be the value of `google_project.project.project_id`, which should be always be `tf-test-${random_id.project_name.hex}`. 


I also don't understand how the project is unknown in the provider `google`.

mmv1/third_party/terraform/provider/provider_project_test.go
zli82016: What kind of flakiness do you see?
Does the test fail locally?

SarahFrench: The flakiness was that locally I saw `tf-test-XXXXXX` was used here, but when I pushed to GitHub the tests running in the CI would have the value coming from environment variables. That's why the regex here can accept two alternate outcomes.

We'll have to see whether the test changes in 54eacde pass on this PR or not!

SarahFrench:
>    provider_project_test.go:182: Step 1/4 error: Check failed: 1 error occurred:
        	* Check 1/1 error: data.google_provider_config_sdk.default-1: Attribute 'project' didn't match "tf-test-[0-9a-z]{10}-[0-9]{1,}", got "<PROJECT ID OF CI TEST PROJECT>"

[That's the flakiness](https://github.com/GoogleCloudPlatform/magic-modules/pull/11607#issuecomment-2341954750) I was seeing before - when I run this test locally I get the value starting tf-test-, not the value from ENVs.

It could be a timing problem: the provider can receive multiple ConfigureProvider calls during an apply, so the timing of those  could impact the data source picking up a value or not. That is, it's not a local vs CI issue, really.

It could also be a just reflection of the SDK being worse at handling unknown string values versus the PF, as the PF type system can differentiate Unknown, Null and empty string values, but to the SDK those 3 states are all `""`, so the SDK test is inconsistent but the PF one is consistent.

I could also change the test so that we don't use the data sources as a read-out; instead I could make the test provision a resource in the project and then check which project that resource is provisioned in:

```diff
// Note that this is the unaliased provider, and is used in the data source below
provider "google" {
	project = google_project.project.project_id
}
-data "google_provider_config_sdk" "default-%{test_step}" {
- }
+ resource "google_service_account" "default" {
+  account_id = "tf-test-abcdefghijk"
+}
```

mmv1/third_party/terraform/provider/provider_project_test.go
zli82016: > * If the `project` config value in the `provider` block is unknown the provider will fall back to the known value coming from ENVs
> * If we remove all project-related ENVs then we force the provider to use the value from  `google_project.project.project_id` once it stops being unknown
> 

Thanks for the information. 
What makes the `project` in the `provider` block is unknown is that `project` needs to be calculated ?
`project = google_project.project.project_id`

mmv1/third_party/terraform/fwprovider/framework_provider_project_test.go
```diff
@@ -0,0 +1,284 @@
...
+	}
+
+	acctest.VcrTest(t, resource.TestCase{
+		PreCheck:                 func() { acctest.AccTestPreCheck(t) },
+		ProtoV5ProviderFactories: acctest.ProtoV5ProviderFactories(t),
+		Steps: []resource.TestStep{
+			{
+				Config: testAccFwProvider_projectUnknownHandling(context, "1"),
+				Check: resource.ComposeAggregateTestCheckFunc(
+					resource.TestMatchResourceAttr("data.google_provider_config_plugin_framework.default-1", "project", regexp.MustCompile("tf-test-[0-9a-z]{10}-[0-9]{1,}")),
```
zli82016, line 187: The project is the data source looks like using the value tf-test-%{random_suffix}-%{test_step} instead of the environment variable. Is it because the project value is known already?

zli82016:
> This PR removes unit tests for how the provider handles `project`. Instead there are acc tests that use the new datasources `google_provider_config_plugin_framework` and `google_provider_config_sdk`
> 

The acceptance tests added in this PR are integration tests.  Do we want to keep the unit tests to have both integration tests and unit tests?

mmv1/third_party/terraform/provider/provider_project_test.go
SarahFrench: Yes that's right, if the value is dependent on another resource being provisioned (or a data source fetching a value) then the value is unknown initially in the apply process. Here's an[ analogous example from a user](https://github.com/hashicorp/terraform-provider-google/issues/14255#issuecomment-1523928110) where the credentials value might be supplied from a resource and be temporarily unknown for the same reason ([related GH issue](https://github.com/hashicorp/terraform-provider-google/issues/14444)).

I think that unknown value handling might be best tested at the unit test level, given how it can be impacted by things like dependencies, sleeps, whether a user is applying something in a single apply or not etc. The [flakiness of the test case](https://github.com/GoogleCloudPlatform/magic-modules/pull/11607#issuecomment-2341976030) suggests it's not going to be useful to us in future!

mmv1/third_party/terraform/fwprovider/framework_provider_project_test.go
SarahFrench: To be honest, I'm confused too. The plugin framework-implemented code [explicitly checks for whether the project value is unknown](https://github.com/GoogleCloudPlatform/magic-modules/blob/477e5c94b4011b01a7f18ccd14fbcecd6d1073eb/mmv1/third_party/terraform/fwtransport/framework_config.go.erb#L151) during the decision to use environment variables, so it would use the ENVs if the value is unknown.

I think, given that we cannot reliably trigger behaviours were the handling of unknown values impacts the outputs of the test, it might be best to remove this test case and instead keep the unit tests around unknown value handling.

SarahFrench:
>The acceptance tests added in this PR are integration tests. Do we want to keep the unit tests to have both integration tests and unit tests?

@zli82016 I'm making PRs such as this one so that we can test the provider's configuration behaviour in a way that's independent to how it's implemented. Mainly this is to support the review process for [this change](https://github.com/GoogleCloudPlatform/magic-modules/pull/11577), which fixes a lot of issues with the muxing and includes removal of the FrameworkProviderConfig struct.

I could limit which unit tests I'm deleting to just those related to FrameworkProviderConfig, and only delete them when I remove FrameworkProviderConfig?

Given the confusing behaviour of the test cases above that focus on unknown value handling I think unit tests should be kept around for that, at the very least

zli82016:
> I could limit which unit tests I'm deleting to just those related to FrameworkProviderConfig, and only delete them when I remove FrameworkProviderConfig?
> 

It works to only delete the unit tests related to FrameworkProviderConfig when removing FrameworkProviderConfig if it doesn't block https://github.com/GoogleCloudPlatform/magic-modules/pull/11577. And we want to keep the unit test for unknown value handling.

mmv1/third_party/terraform/fwtransport/go/framework_config_test.go
```diff
@@ -14,168 +14,6 @@ import (
 	transport_tpg "github.com/hashicorp/terraform-provider-google/google/transport"
 )
 
-func TestFrameworkProvider_LoadAndValidateFramework_project(t *testing.T) {
```
zli82016, line 17: Do you mind adding this function back? Thanks.

SarahFrench:
👆 and in that test failure the ENV is used instead of the config! Definitely flakey. I'll remove the test case later

mmv1/third_party/terraform/fwtransport/go/framework_config_test.go
SarahFrench: Added in [f66e3ad](https://github.com/GoogleCloudPlatform/magic-modules/pull/11607/commits/f66e3ad34a8fbe8efcb891364cc0893b1d4e743f)

zli82016 (approved):
LGTM. The failed test looks like unrelated.

mmv1/third_party/terraform/fwtransport/go/framework_config_test.go
SarahFrench: Ah sorry, this is the go rewrite file - I'll push it back

SarahFrench:
I realised that [I was confused before](https://github.com/GoogleCloudPlatform/magic-modules/pull/11607#discussion_r1759445411) - once that latest push completes it's automation I'll merge the PR.

Thank you for the review & bearing with me on the weird Unknown behaviour!

SarahFrench:
Merging, as the [test failure](https://github.com/GoogleCloudPlatform/magic-modules/pull/11607#issuecomment-2350356597) here is unrelated + addressed in another PR

SarahFrench:
That resolves the missing test, merging now!

PR #11613, opened 2024-09-03
fix: Router advertised-route-priority undefined behavior

harshithpatte-g:
Bug: b/356353797
Issue: advertised-route-priority is an optional field and if not present, the value in the GCP is considered to be 100.
However, TF would set advertised-route-priority to be 0 even if the user has not explicitly set the value to be 0 when there is an update to the resource.

Repro steps:
1. Create a Router peer resource without advertised route priority value set.
resource "google_compute_router_peer" "foobar" {
  name                      = "tf-my-router"
  router                    = google_compute_router.foobar.name
  region                    = google_compute_router.foobar.region
  peer_asn                  = 65515
  interface                 = google_compute_router_interface.foobar.name
  advertise_mode            = "DEFAULT"
}

Query gcloud to check the value of advertised_route_priority, it will be empty
`gcloud compute routers describe {router-name}
`

2. Update the router peer resource (ex: enableIpv6 = true) without advertised route priority. However, TF would add advertised route priority = 0 in the update api call. 
resource "google_compute_router_peer" "foobar" {
  name                      = "tf-my-router"
  router                    = google_compute_router.foobar.name
  region                    = google_compute_router.foobar.region
  peer_asn                  = 65515
  interface                 = google_compute_router_interface.foobar.name
  advertise_mode            = "DEFAULT"
  enable_ipv6           = true
}

Query gcloud to check the value of advertised_route_priority, it will be 0 even though it is not set by the user.
`gcloud compute routers describe {router-name}
`

```release-note:bug
compute: fixed advertised_route_priority set to 0 without user update in 'google_compute_router_peer'
```

### DIFF
--- a/mmv1/third_party/terraform/services/compute/resource_compute_router_peer.go.erb
+++ b/mmv1/third_party/terraform/services/compute/resource_compute_router_peer.go.erb
@@ -420,7 +420,7 @@ func resourceComputeRouterBgpPeerCreate(d *schema.ResourceData, meta interface{}
 	advertisedRoutePriorityProp, err := expandNestedComputeRouterBgpPeerAdvertisedRoutePriority(d.Get("advertised_route_priority"), d, config)
 	if err != nil {
 		return err
-	} else if v, ok := d.GetOkExists("advertised_route_priority"); ok || !reflect.DeepEqual(v, advertisedRoutePriorityProp) {
+	} else if v, ok := d.GetOk("advertised_route_priority"); ok || !reflect.DeepEqual(v, advertisedRoutePriorityProp) {
 		obj["advertisedRoutePriority"] = advertisedRoutePriorityProp
 	}
 	advertiseModeProp, err := expandNestedComputeRouterBgpPeerAdvertiseMode(d.Get("advertise_mode"), d, config)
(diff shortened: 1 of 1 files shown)


c2thorn (changes requested):
Does this have an associated Github issue that this closes?

How did you test this change? Can it be written as a new testing or into an existing test?

Are we sure this won't be a breaking change for existing configurations?

harshithpatte-g:
> Does this have an associated Github issue that this closes?
> No, this is a customer issue. I don't see any github issue. It happens only when there is an update to the existing terraform bgp peer resource.
> How did you test this change? Can it be written as a new testing or into an existing test?
> I tested it manually by doing a gcloud query for the repro steps. We can't write a test or add into an existing test.
> Are we sure this won't be a breaking change for existing configurations?
> Yes, I have tried different permutations like, 
update the existing resources without advertised_route_priority with the fix.
update the existing resources with advertised_route_priority  with the fix.
Create a new resource with a fix.
Create a new resource with advertised_route_priority with a fix.
Remove advertised_route_priority from an existing resource with advertised_route_priority
and so on.

c2thorn (approved):
Confirming that the contributor has provided manual tests of the changes and walked through the logic. LGTM

PR #11620, opened 2024-09-03
Go Rewrite Doc edits

NickElliot:
cc @zli82016 

```release-note:none

```

### DIFF
--- a/docs/content/develop/resource-reference.md
+++ b/docs/content/develop/resource-reference.md
@@ -302,7 +302,7 @@ mutex: 'alloydb/instance/{{name}}'
 
 ### `virtual_fields`
 
-Contains a list of [virtual_fields]({{< ref "/develop/client-side-fields/" >}}). By convention,
+Contains a list of [virtual_fields]({{< ref "/develop/client-side-fields" >}}). By convention,
 these should be fields that do not get sent to the API, and are instead used to modify
 the behavior of a Terraform resource such as `deletion_protection`.
 
--- a/docs/content/develop/resource-reference.md
+++ b/docs/content/develop/resource-reference.md
@@ -26,7 +26,9 @@ Example:
 ```yaml
 description: |
   This is a multi-line description
-  of a resource.
+  of a resource. All multi-line descriptions must follow
+  this format of using a vertical bar followed by a line-break,
+  with the remaining description being indented.
 ```
 
 ### `references`
@@ -39,10 +41,10 @@ Links to reference documentation for a resource. Contains two attributes:
 Example:
 
 ```yaml
-references: !ruby/object:Api::Resource::ReferenceLinks
-     guides:
-      'Create and connect to a database': 'https://cloud.google.com/alloydb/docs/quickstart/create-and-connect'
-     api: 'https://cloud.google.com/alloydb/docs/reference/rest/v1/projects.locations.backups'
+references:
+  guides:
+    'Create and connect to a database': 'https://cloud.google.com/alloydb/docs/quickstart/create-and-connect'
+  api: 'https://cloud.google.com/alloydb/docs/reference/rest/v1/projects.locations.backups'
 ```
 
 ### `min_version: beta`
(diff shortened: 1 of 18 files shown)


docs/content/develop/resource-reference.md
```diff
@@ -26,7 +26,9 @@ Example:
 ```yaml
 description: |
   This is a multi-line description
-  of a resource.
+  of a resource. All multi-line descriptions must follow
```
c2thorn, line 29: Is this a new limitation of the Go-format YAMLs? I guess the other way to make multi-line descriptions would be to add newlines into the text?

docs/content/get-started/generate-providers.md
```diff
@@ -63,8 +58,6 @@ If you are familiar with Docker or Podman, you may want to use the experimental
    Expected output if everything is installed properly:
  
    ```
```
c2thorn, line 60: Just noting that we'll need to modify the doctor script when we flip

docs/content/reference/make-commands.md
```diff
@@ -17,26 +17,21 @@ Generates the code for the downstream `google` and `google-beta` providers.
 Examples:
 
 ```bash
-make provider VERSION=ga OUTPUT_PATH="$GOPATH/src/github.com/hashicorp/terraform-provider-google"
-make provider VERSION=beta OUTPUT_PATH="$GOPATH/src/github.com/hashicorp/terraform-provider-google-beta"
+go run . --version ga --output $GOPATH/src/github.com/hashicorp/terraform-provider-google
```
c2thorn, line 20: I think this command will actually stay the same. It'll just call the `go run` command under the hood. 

Basically we just want to modify any references to `bundle exec`

docs/content/reference/make-commands.md
```diff
@@ -17,26 +17,21 @@ Generates the code for the downstream `google` and `google-beta` providers.
...
 ```
 
 #### Arguments
 
-- `OUTPUT_PATH`: Required. The location you are generating provider code into.
-- `VERSION`: Required. The version of the provider you are building into. Valid values are `ga` and `beta`.
-- `PRODUCT`: Limits generations to the specified folder within `mmv1/products` or `tpgtools/api`. Handwritten files from `mmv1/third_party/terraform` are always generated into the downstream regardless of this setting, so you can provide a non-existant product name to generate only handwritten code. Required if `RESOURCE` is specified.
-- `RESOURCE`: Limits generation to the specified resource within a particular product. For `mmv1` resources, matches the resource's `name` field (set in its configuration file).For `tpgtools` resources, matches the terraform resource name.
-- `ENGINE`: Modifies `make provider` to only generate code using the specified engine. Valid values are `mmv1` or `tpgtools`. (Providing `tpgtools` will still generate any prerequisite mmv1 files required for tpgtools.)
+- `output`: Required. The location you are generating provider code into.
```
c2thorn, line 32: Se above, I think all of these arguments will remain

docs/content/develop/resource-reference.md
NickElliot: I was under the impression we had some description diffs that were impossible to remove until the yaml conversion template was modified to use this approach, but I suppose it's possible that `\n`s in a standard quotation wrapped string would still work? I would still prefer to recommend this approach as it would be easier to read.

c2thorn: Ah I see. Yes we wanted to use the | format to just reduce the number of diffs generated. Without using it, the generator was producing a unreviewable amount of diffs.

So I guess it's a recommendation rather than a necessity.

NickElliot: I'll replace it with a "should" to lower the intensity

NickElliot:
A couple todos:
there are references to `templates/terraform/iam/go/<file>`, `templates/terraform/examples/go/<file>`, and `templates/terraform/custom_*/go/<resource_specific_file>` in a bunch of files, these need modifying

some example file references have permalinks, these need to be updated once we have the main-branch commit hash post-update

c2thorn (approved):

PR #11621, opened 2024-09-03
Add go yaml validations (part 1)

zli82016:
1. Replaced yaml v3 with yaml v2 library, as yaml v3 doesn't have UnmarshalStrict method. UnmarshalStrict is used to check extraneous fields in the yaml files.
2. Added the validation for most objects except `type`

The remaining validation to be added is on the type object:
1. Add the same validation with the ruby object for the fields, like required fields, conflicted fields if any.
2. Add the validation for nested types, like array, object to make sure no extraneous fields. This validation could be difficult as all of fields are in the same Go Type object.

```release-note:none

```

### DIFF
--- a/mmv1/go.mod
+++ b/mmv1/go.mod
@@ -4,7 +4,7 @@ go 1.21
 
 require (
 	golang.org/x/exp v0.0.0-20240222234643-814bf88cf225
-	gopkg.in/yaml.v3 v3.0.1
+	gopkg.in/yaml.v2 v2.4.0
 )
 
 require github.com/golang/glog v1.2.0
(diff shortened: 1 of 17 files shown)


zli82016:
cc @NickElliot

mmv1/go.mod
```diff
@@ -4,7 +4,7 @@ go 1.21
 
 require (
 	golang.org/x/exp v0.0.0-20240222234643-814bf88cf225
-	gopkg.in/yaml.v3 v3.0.1
```
c2thorn, line 7: From what you can tell, do we lose anything by downgrading?

c2thorn (approved):
thank you!

mmv1/go.mod
zli82016: I don't see diffs after downgrading.

PR #11622, opened 2024-09-03
Fix google_dns_record_set with same name throw an error

Samir-Cit:
This PR fixes the issue https://github.com/hashicorp/terraform-provider-google/issues/10661 by removing the piece of code that prevents the error when creating a resource with a name that already exists.

```release-note:bug
dns: added a mutex to `google_dns_record_set` to prevent conflicts when multiple resources attempt to operate on the same record set
```


slevenick (changes requested):
After discussion with the team on this, I think the best course of action is to add a mutex to this resource to prevent the 409 conflict situations described in the linked bug.

We don't want to remove this code as it will make the resource unauthoritative, which some users may currently depend on. If we're convinced we want to make it unauthoritative we could consider that in a major version, but until then we should limit the change to adding a mutex

Samir-Cit (commented):
Suggestion applied to use mutex instead of removing the code.

slevenick:
/gcbrun I think I messed something up

slevenick (commented):
Compilation errors

slevenick (approved):
Thanks!

PR #11624, opened 2024-09-03
Add google_site_verification_web_resource

SirGitsalot:
The adds support for the site verification API [WebResource](https://developers.google.com/site-verification/v1/getting_started#web-resource). It's step two out of three in implementing [issue 5698](https://github.com/hashicorp/terraform-provider-google/issues/5698). Step one was #10999 and step 3 will add the `owners` field as a fine grained resource attached to this resource.

There's two basic ways to verify ownership of a site: creating a DNS record or modifying the site's content. I've included tests for both, but the former is disabled under VCR as it requires dynamically creating a real, resolvable DNS record as part of the test. I know we have `gcp.tfacc.hashicorptest.com` for use in other tests, but I couldn't find an example of creating records in that zone.

```release-note:new-resource
`google_site_verification_web_resource`
```


mmv1/products/siteverification/WebResource.yaml
```diff
@@ -0,0 +1,101 @@
...
+      - :DNS_CNAME
+      - :DNS_TXT
+      - :FILE
+      - :META
+      - :TAG_MANAGER
+properties:
+  - !ruby/object:Api::Type::String
+    name: 'id'
+    output: true
+    immutable: true
```
slevenick, line 71: immutable is not applicable for an output field

mmv1/products/siteverification/WebResource.yaml
```diff
@@ -0,0 +1,101 @@
...
+        required: true
+        description: |
+          The site identifier. If the type is set to SITE, the identifier is a URL. If the type is
+          set to INET_DOMAIN, the identifier is a domain name.
+  - !ruby/object:Api::Type::Array
+    name: 'owners'
+    description: |
+      The email addresses of all direct, verified owners of this exact property. Indirect owners —
+      for example verified owners of the containing domain—are not included in this list.
+    immutable: true
```
slevenick, line 99: immutable not needed

mmv1/products/siteverification/WebResource.yaml
```diff
@@ -0,0 +1,101 @@
...
+  the web resource is destroyed. The Google Site Verification API checks that the
+  verification data exists at creation time and does not exist at destruction time
+  and will fail if the required condition is not met.
+base_url: webResource
+self_link: webResource/{{id}}
+create_url: webResource?verificationMethod={{verification_method}}
+delete_url: webResource/{{id}}
+id_format: '{{id}}'
+immutable: true
+identity:
```
slevenick, line 32: Do we need `identity` here? I think that's only for fine-grained resources

mmv1/products/siteverification/WebResource.yaml
```diff
@@ -0,0 +1,101 @@
...
+  ~> **Note:** The verification data (DNS `TXT` record, HTML file, `meta` tag, etc.)
+  must already exist before the web resource is created, and must be deleted before
+  the web resource is destroyed. The Google Site Verification API checks that the
+  verification data exists at creation time and does not exist at destruction time
+  and will fail if the required condition is not met.
+base_url: webResource
+self_link: webResource/{{id}}
+create_url: webResource?verificationMethod={{verification_method}}
+delete_url: webResource/{{id}}
+id_format: '{{id}}'
```
slevenick, line 30: Should this be `webResource/{{id}}`? Usually we keep this in a selfLink-ish form so we can use .id from other resources. In this case that would probably never happen, but it would keep the ID formats standardized

mmv1/third_party/terraform/services/siteverification/resource_site_verification_web_resource_test.go
```diff
@@ -0,0 +1,277 @@
...
+
+	"github.com/hashicorp/terraform-plugin-testing/helper/resource"
+	"github.com/hashicorp/terraform-plugin-testing/terraform"
+
+	"github.com/hashicorp/terraform-provider-google/google/acctest"
+	"github.com/hashicorp/terraform-provider-google/google/tpgresource"
+	transport_tpg "github.com/hashicorp/terraform-provider-google/google/transport"
+)
+
+func TestAccSiteVerificationWebResource_siteverificationDomain(t *testing.T) {
```
slevenick, line 16:
```suggestion
func TestAccSiteVerificationWebResource_siteVerificationDomain(t *testing.T) {
```

mmv1/third_party/terraform/transport/error_retry_predicates.go
```diff
@@ -572,3 +572,13 @@ func ExternalIpServiceNotActive(err error) (bool, string) {
 	}
 	return false, ""
 }
+
+// Site verification may return a 400 error while waiting for DNS propagation.
+func IsSiteValidationRetryableError(err error) (bool, string) {
```
slevenick, line 577: Site validation vs verification?

mmv1/products/siteverification/WebResource.yaml
SirGitsalot: `identity` and `id_format` were arrived at through trial and error - omitting `identity` results in the post-create read requesting `/siteVerification/v1/webResource/` instead of `/siteVerification/v1/webResource/{{id}}`. There may be another combination of `identity`, `id_format`, `base_url`, `self_link`, etc. that'll work but this is the one I found to work.

mmv1/products/siteverification/WebResource.yaml
SirGitsalot: In this case, changing `id_format` results in a post-create read to `/siteVerification/v1/webResource/webResource/{{id}}`.

FWIW I have a follow-up PR to add a fine grained resource to this resource, and the test resource looks like:

```
resource "google_site_verification_owner" "example" {
  web_resource_id = google_site_verification_web_resource.example.id
  email           = "%{owner}"
}
```

It resolves OK 🤷

mmv1/products/siteverification/WebResource.yaml
SirGitsalot: Done

mmv1/products/siteverification/WebResource.yaml
SirGitsalot: Done

mmv1/third_party/terraform/services/siteverification/resource_site_verification_web_resource_test.go
SirGitsalot: Done

mmv1/third_party/terraform/transport/error_retry_predicates.go
SirGitsalot: Done

slevenick (changes requested):

SirGitsalot:
Ah, that all makes sense!

I did end up leaving in `identity` (changing it from `id` to `web_resource_id`). Without it the create doesn't extract `web_resource_id` from the API response. That is, `identity` causes this code to be generated post-create:

```
if err := d.Set("web_resource_id", flattenSiteVerificationWebResourceWebResourceId(res["id"], d, config)); err != nil {
    return fmt.Errorf(`Error setting computed identity field "web_resource_id": %s`, err)
}
```

slevenick (approved):

PR #11626, opened 2024-09-03
go rewrite - general refresh and diffs 9/3

c2thorn:
separated the changes out by commit

- full service refresh
- There was a bug in the versioning where many beta-only test files were not being generated.
- other minor diffs

the diff generated after running `go run . --version beta --output /Users/camthornton/terraform-provider-google-beta`
https://github.com/c2thorn/terraform-provider-google-beta/commit/468a149676ccf3166e97b423d0e43851a2d532f2

```release-note:none

```


NickElliot (approved):
diffs look good, I should've caught that change on the redis examples where I updated the new templates to use the correct logic but not the old

PR #11627, opened 2024-09-03
go rewrite - initial concurrency

c2thorn:
adds a quick attempt at running the product generation concurrently

Emulates how we currently do parallelization in Ruby. See https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/compiler.rb#L155. I did not try to improve upon this

reduces full generation time from 1 minute 45 seconds -> 53 seconds after https://github.com/GoogleCloudPlatform/magic-modules/pull/11626

```release-note:none

```

### DIFF
--- a/mmv1/main.go
+++ b/mmv1/main.go
@@ -9,6 +9,7 @@ import (
 	"path/filepath"
 	"sort"
 	"strings"
+	"sync"
 	"time"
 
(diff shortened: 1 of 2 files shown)


zli82016 (approved):
LGTM. Great work.

PR #11628, opened 2024-09-03
Make vcr tester Run and RunParallel use options struct

trodge:
Adding an options struct for these methods so that their parameters can be changed without breaking current usage.

```release-note:none

```

### DIFF
--- a/.ci/magician/cmd/test_terraform_vcr.go
+++ b/.ci/magician/cmd/test_terraform_vcr.go
@@ -403,7 +403,7 @@ func runReplaying(runFullVCR bool, services map[string]struct{}, vt *vcr.Tester)
 	var testDirs []string
 	var replayingErr error
 	if runFullVCR {
-		fmt.Println("run full VCR tests")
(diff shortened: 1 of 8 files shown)


.ci/magician/cmd/test_terraform_vcr.go
```diff
@@ -393,15 +403,22 @@ func runReplaying(runFullVCR bool, services map[string]struct{}, vt *vcr.Tester)
 	var testDirs []string
 	var replayingErr error
 	if runFullVCR {
-		fmt.Println("runReplaying: full VCR tests")
-		result, replayingErr = vt.Run(vcr.Replaying, provider.Beta, nil)
+		fmt.Println("run full VCR tests")
```
melinath, line 406: nit: I think this got lost in the merge conflict resolution

```suggestion
		fmt.Println("runReplaying: full VCR tests")
```

melinath (approved):
LGTM - could you push a commit to force test runs just to be extra safe?

PR #11633, opened 2024-09-04
Fixed potential panics during VCR tests

melinath:
Fixes errors like https://pantheon.corp.google.com/cloud-build/builds/fd3758bb-622f-42d9-9e15-13ef09651e83;step=23?inv=1&invt=AbbPDQ&project=graphite-docker-images

```release-note:none

```

### DIFF
--- a/.ci/magician/cmd/test_terraform_vcr.go
+++ b/.ci/magician/cmd/test_terraform_vcr.go
@@ -56,7 +56,7 @@ var ttvEnvironmentVariables = [...]string{
 }
 
 type analytics struct {
-	ReplayingResult  *vcr.Result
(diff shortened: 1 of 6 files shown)


trodge:
If we want the result to never be `nil`, let's just make it a plain struct instead of a pointer.

melinath:
If there's no specific reason it needs to be a pointer, I'd rather just have us use a struct - missing a nil check is too easy. I'll make that change.

trodge (changes requested):
If we want to always have a `Result` any time we call `Run`, let's make the return values of `Run` `(Result, error)` instead of `(*Result, error)`.

trodge (approved):

PR #11636, opened 2024-09-04
Fix deep_merge function to correctly handle nil arrays

roaks3:
This was discovered on https://cloud-internal-review.git.corp.google.com/c/cloud-graphite-eng/magic-modules-private-overrides/+/47770.

Previously, we could have a state where a new array field was added in the overrides (like `conflicts`), and this would cause a build failure. This occurred because `string_array?` assumes the array is not nil.

This fix checks both arrays for nil, and instead of failing when the array from magic-modules is nil, it returns the other array, which is what the caller likely expects.

```release-note:none

```

### DIFF
--- a/mmv1/api/object.go
+++ b/mmv1/api/object.go
@@ -34,11 +34,18 @@ type NamedObject struct {
 // }
 
 // func (n *Named) deep_merge(arr1, arr2) {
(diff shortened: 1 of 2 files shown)


c2thorn (approved):
LGTM

PR #11640, opened 2024-09-05
Marked customer_name field as optional for interconnect resource 

abhijeetkjha-google:
```release-note:enhancement
compute: marked `customer_name` and `location` fields as optional in `google_compute_interconnect` resource to support cross cloud interconnect
```

Bug Fix: https://github.com/hashicorp/terraform-provider-google/issues/18264
The proto which defines the interconnect resource has customer_name field as an optional, so we should be providing this same parity between the terraform and the provided api. For Cross cloud interconnect, We should not be providing customer_name, else error is being thrown.


mmv1/templates/terraform/examples/compute_interconnect_basic.tf.erb
```diff
@@ -2,7 +2,6 @@ data "google_project" "project" {}
```
shuyama1, line 2: Looks like this example is skipped from test generation. It will update the examples in the web documentation though. Would you mind also removing the field from [this example](https://github.com/abhijeetkjha-google/magic-modules/blob/bugFix-361094483/mmv1/templates/terraform/examples/compute_interconnect_basic_test.tf.erb) to have a test coverage on this. Plus, can you also remove this line https://github.com/abhijeetkjha-google/magic-modules/blob/bugFix-361094483/mmv1/products/compute/Interconnect.yaml#L52, given the field is removed from the example. Thanks

abhijeetkjha-google: > Would you mind also removing the field from [this example](https://github.com/abhijeetkjha-google/magic-modules/blob/bugFix-361094483/mmv1/templates/terraform/examples/compute_interconnect_basic_test.tf.erb) to have a test coverage on this. 

Since we are marking a field as optional wouldn't it be a good idea to have one test example with the field and another without the field. The compute_interconnect_basic.tf.erb contains the required field only and compute_interconnect_basic_test.tf.erb includes the optional field as well.

> can you also remove this line https://github.com/abhijeetkjha-google/magic-modules/blob/bugFix-361094483/mmv1/products/compute/Interconnect.yaml#L52, given the field is removed from the example.

Sure, Done.

harshithpatte-g:
@abhijeetkjha-google Please update the release note.

abhijeetkjha-google:
> @abhijeetkjha-google Please update the release note. Otherwise, LGTM.

Done.

mmv1/templates/terraform/examples/compute_interconnect_basic.tf.erb
abhijeetkjha-google: Yes yes, I have verified that customer_name is an optional field and is not required for resource creation.

shuyama1:
I've changed the change type from `bug` to `enhancement` in the release note as I think it's more accurate.

shuyama1 (commented):
I've removed `customer_name` from the test case and run tests in a test PR https://github.com/GoogleCloudPlatform/magic-modules/pull/11668. Looks like the test failed with API error `googleapi: Error 400: Required field 'resource.customerName' not specified, required`. I wonder if the field is still required?

abhijeetkjha-google:
> I've removed `customer_name` from the test case and run tests in a test PR #11668. Looks like the test failed with API error `googleapi: Error 400: Required field 'resource.customerName' not specified, required`. I wonder if the field is still required?

The error is rightly thrown, So I investigated this. So this is the issue.

We have three interconnects: Dedicated, Partner and Cross cloud.
So out of these three, Dedicated and Partner Interconnect requires customer_name as a field to create interconnect resource. And for **Cross cloud interconnect we do no pass the customer_name else we will throw an error, incase a customer_name is detected.** 

without this change, If a person creates an cross cloud interconnect resource, They have to pass the customer_name field as its marked "required" for now but this will throw an error stating ""Customer name should not be present for Cross Cloud Interconnect".

So the test which are failing are for dedicated interconnect, which was a miss. So thats why we are getting this error.
So what I propose is that we should mark "customer_name" field as optional and not make any changes in these test. Instead I will try to write some test for "cross cloud interconnect" scenario.

shuyama1 (commented):
`TestAccComputeInterconnect_computeInterconnectCrossCloudBasicExample` failed with 
```
resource_compute_interconnect_generated_test.go:85: Step 1/2 error: Error running pre-apply refresh: exit status 1
        
        Error: Missing required argument
        
          on terraform_plugin_test.tf line 4, in resource "google_compute_interconnect" "example-interconnect":
           4: resource "google_compute_interconnect" "example-interconnect" {
        
        The argument "location" is required, but no definition was found.
```

abhijeetkjha-google:
Hi @shuyama1, 
I tried adding/updating create tests, But it keeps failing as we can't use z2z location which are the locations available for testing. Also we need to add remote_location field which identifies this request as a cross_cloud instance. But as per the implementation we can't have location field in such cases so marking this field as optional as well. Since I am not able to add any function test around this scenario till we get some sort of test remote_location. I have updated this pr just with the field change. Let me know if anything can be done around this. Thanks.

mmv1/products/compute/Interconnect.yaml
```diff
@@ -89,7 +89,6 @@ properties:
       URL of the InterconnectLocation object that represents where this connection is to be provisioned.
     resource: 'InterconnectLocations'
     imports: 'selfLink'
-    required: true
```
shuyama1, line 92: Is making `location` optional in this resource intended? Sorry, I'm not fully understand what remote_location field is. Is location field not needed for resource creation in some cases?

mmv1/products/compute/Interconnect.yaml
```diff
@@ -145,7 +144,6 @@ properties:
       Customer name, to put in the Letter of Authorization as the party authorized to request a
       crossconnect.
     immutable: true
-    required: true
```
shuyama1, line 148: Could you please update the description of this field to clarify when this field is required and when it is not? Thanks

shuyama1 (commented):
Would you be able to set up a local environment and test it locally? We may rely on local testing in this case if our testing environment is not compatible with feature. Thank you! https://googlecloudplatform.github.io/magic-modules/develop/test/run-tests/#run-tests

mmv1/products/compute/Interconnect.yaml
abhijeetkjha-google: Yes, Incase of cross_cloud interconnect we don't need to specify location field as location field implies for google owned location, Incase of cross_cloud interconnect we have to replace it with remote_location field.

mmv1/products/compute/Interconnect.yaml
abhijeetkjha-google: Done.

mmv1/products/compute/Interconnect.yaml
shuyama1: Got it. Would you mind also clarifying it in the description of the field? Thanks!

shuyama1 (commented):
left a small comment in https://github.com/GoogleCloudPlatform/magic-modules/pull/11640#discussion_r1771950326 regarding a documentation improvement. Thanks!

abhijeetkjha-google:
> left a small comment in [#11640 (comment)](https://github.com/GoogleCloudPlatform/magic-modules/pull/11640#discussion_r1771950326) regarding a documentation improvement. Thanks!

Updated the description. Thanks.

mmv1/products/compute/Interconnect.yaml
abhijeetkjha-google: Done

shuyama1 (approved):
The contributor has confirmed the changes work as expected with local testing.

PR #11645, opened 2024-09-05
add AutoTier support for netapp storagePools and Volumes

Mehul3217:

```release-note:enhancement
netapp: added `allow_auto_tiering` field to StoragePool and  `tiering_policy` field to Volume resource
```


mmv1/third_party/terraform/services/netapp/resource_netapp_storage_pool_test.go.erb
```diff
@@ -41,6 +41,15 @@ func TestAccNetappStoragePool_storagePoolCreateExample_update(t *testing.T) {
 				ImportStateVerify:       true,
 				ImportStateVerifyIgnore: []string{"location", "name", "labels", "terraform_labels"},
 			},
+			{
+				Config: testAccNetappStoragePool_storagePoolCreateExample_autoTieredPool(context),
```
rileykarson, line 45: Since this is immutable, it should get covered in single-step test, probably a new one!

mmv1/third_party/terraform/services/netapp/resource_netapp_volume_test.go
```diff
@@ -659,3 +668,31 @@ func testAccNetappVolume_volumeBasicExample_cleanupScheduledBackup(t *testing.T,
...
+
+func testAccNetappVolume_volumeBasicExample_autoTier(context map[string]interface{}) string {
+	return acctest.Nprintf(`
+resource "google_netapp_storage_pool" "default" {
+    name = "tf-test-pool%{random_suffix}"
+    location = "us-west2"
+    service_level = "PREMIUM"
+    capacity_gib = "2048"
+    network = data.google_compute_network.default.id
+    allow_auto_tiering = true
```
rileykarson, line 680: Same here, we probably want a dedicated test. Since the new fields in `google_netapp_volume` we'll want to test a couple of different values from them in different `Config` steps, though.

mmv1/third_party/terraform/services/netapp/resource_netapp_volume_test.go
```diff
@@ -659,3 +668,31 @@ func testAccNetappVolume_volumeBasicExample_cleanupScheduledBackup(t *testing.T,
 		return nil
 	}
 }
+
+func testAccNetappVolume_volumeBasicExample_autoTier(context map[string]interface{}) string {
+	return acctest.Nprintf(`
+resource "google_netapp_storage_pool" "default" {
```
rileykarson, line 674: note: the actual failure reason is because we're recreating the `google_netapp_storage_pool` with both the delete and create in parallel. That's happening because in this config the resource is named (in Terraform) `google_netapp_storage_pool.default2` before and `google_netapp_storage_pool.default` now, so Terraform doesn't recognise it should wait for the deletion before creating the new resource.

We should obviate that with the other proposed changes.

Mehul3217: Done

mmv1/third_party/terraform/services/netapp/resource_netapp_volume_test.go
Mehul3217: Done

mmv1/third_party/terraform/services/netapp/resource_netapp_storage_pool_test.go.erb
Mehul3217: Done

mmv1/third_party/terraform/services/netapp/resource_netapp_storage_pool_test.go.erb
```diff
@@ -129,6 +129,75 @@ resource "google_netapp_storage_pool" "test_pool" {
...
+
+func TestAccNetappStoragePool_autoTieredStoragePoolCreateExample_update(t *testing.T) {
+	context := map[string]interface{}{
+		"network_name":  acctest.BootstrapSharedServiceNetworkingConnection(t, "gcnv-network-config-1", acctest.ServiceNetworkWithParentService("netapp.servicenetworking.goog")),
+		"random_suffix": acctest.RandString(t, 10),
+	}
+
+	acctest.VcrTest(t, resource.TestCase{
+		PreCheck:                 func() { acctest.AccTestPreCheck(t) },
+		ProtoV5ProviderFactories: acctest.ProtoV5ProviderBetaFactories(t),
```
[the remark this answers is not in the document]
Mehul3217: Done

.terraform.lock.hcl
```diff
@@ -0,0 +1,21 @@
+# This file is maintained automatically by "terraform init".
```
rileykarson, line 1: Can you remote these files under .terraform? I think they're the source of the issue.

Mehul3217: Done

mmv1/third_party/terraform/services/netapp/resource_netapp_volume_test.go
```diff
@@ -659,3 +658,60 @@ func testAccNetappVolume_volumeBasicExample_cleanupScheduledBackup(t *testing.T,
 		return nil
 	}
 }
+
+func TestAccNetappVolume_autoTieredNetappVolumeExample_update(t *testing.T) {
```
rileykarson, line 662:
```suggestion
func TestAccNetappVolume_autoTieredNetappVolume(t *testing.T) {
```

rileykarson: Applying this since we're not doing an update / this is a handwritten test rather than an example-based one.

mmv1/third_party/terraform/services/netapp/resource_netapp_volume_test.go
rileykarson: Ah, this was missing an important word, whoops. Since the new fields are updatable we'll need to add an update test (i.e. a test with multiple `Config` steps that exercise different values), as is covered at https://googlecloudplatform.github.io/magic-modules/develop/test/test/#add-an-update-test

I renamed the test for this resource based on it having a single `Config` step below; you probably want to undo that and add more steps.

rileykarson (commented):
We seem to have hit an API bug on update- the update mask is getting rejected, but looks well-formed.

rileykarson (changes requested):
In addition to removing the update step we also need to remove update support by marking `tieringPolicy` and its children immutable. We should not ship a broken field to end users.

Mehul3217:
> In addition to removing the update step we also need to remove update support by marking `tieringPolicy` and its children immutable. We should not ship a broken field to end users.

We can update the tierringPolicy children `tier action` between paused to enable, and update `cooling threshold days`. We can't remove tiering policy

So we cannot make them immutable

rileykarson:
Yes, we can make them mutable later when update works- in the meantime we need to mark them as immutable because Terraform will fail to update them.

mmv1/products/netapp/Volume.yaml
```diff
@@ -503,3 +503,29 @@ properties:
...
+  - !ruby/object:Api::Type::String
+    name: 'coldTierSizeGib'
+    description: |
+      Output only. Size of the volume cold tier data in GiB.
+    output: true
+  - !ruby/object:Api::Type::NestedObject
+    name: tieringPolicy
+    description: |-
+      Tiering policy for the volume.
+    update_mask_fields:
```
rileykarson, line 515: non-actionable: I'll flag that requiring this means an AIP is being broken, as the top-level field should have been sufficient: https://google.aip.dev/161#specifying-specific-fields

rileykarson (approved):

PR #11647, opened 2024-09-05
Further refactoring of magician VCR

trodge:
Options struct for `UploadLogs`

`logBucket` and `cassetteBucket` moved into Tester object

`safeToLog` map added, environment variables must be marked as safe to be logged

`FetchCassettes` will now return errors from gsutil instead of suppressing them

```release-note:none

```


.ci/magician/cmd/check_cassettes.go
```diff
@@ -64,7 +64,7 @@ var checkCassettesCmd = &cobra.Command{
 
 		ctlr := source.NewController(env["GOPATH"], "modular-magician", githubToken, rnr)
 
-		vt, err := vcr.NewTester(env, rnr)
+		vt, err := vcr.NewTester(env, "ci-vcr-logs", "ci-vcr-cassettes", rnr)
```
melinath, line 67: it looks like previously this was using `"vcr-check-cassettes"` - just wanted to double-check that this is intentionally changing?

.ci/magician/cmd/test_terraform_vcr.go
```diff
@@ -472,21 +505,21 @@ func formatComment(fileName string, tmplText string, data any) (string, error) {
 }
 
 func formatTestsAnalytics(data analytics) (string, error) {
-	return formatComment("test_terraform_vcr_test_analytics.tmpl", testsAnalyticsTmplText, data)
+	return formatComment("test_analytics.tmpl", testsAnalyticsTmplText, data)
 }
 
 func formatNonExercisedTests(data nonExercisedTests) (string, error) {
-	return formatComment("test_terraform_vcr_recording_mode_results.tmpl", nonExercisedTestsTmplText, data)
+	return formatComment("recording_mode_results.tmpl", nonExercisedTestsTmplText, data)
```
melinath, line 512: nit: This isn't actually "used" but to be accurate... (it was already wrong in main.)

```suggestion
	return formatComment("non_exercised_tests.tmpl", nonExercisedTestsTmplText, data)
```

.ci/magician/cmd/vcr_cassette_update.go
```diff
@@ -90,7 +90,7 @@ var vcrCassetteUpdateCmd = &cobra.Command{
 		}
 		ctlr := source.NewController(env["GOPATH"], "hashicorp", env["GITHUB_TOKEN_CLASSIC"], rnr)
 
-		vt, err := vcr.NewTester(env, rnr)
+		vt, err := vcr.NewTester(env, "ci-vcr-logs", "ci-vcr-cassettes", rnr)
```
melinath, line 93: Based on the code, I'm not sure these are correct... but I'm also not sure they're being used here? We're not currently using this go file & there could be related bugs there already? attn @iyabchen @shuyama1

.ci/magician/exec/runner.go
```diff
@@ -100,6 +100,14 @@ func (ar *Runner) ReadFile(name string) (string, error) {
 	return string(data), nil
 }
 
+func (ar *Runner) AppendFile(name, data string) error {
```
melinath, line 103: Rather than reading & then writing the file, we may want to use append modes - more efficient: https://pkg.go.dev/os#example-OpenFile-Append

.ci/magician/vcr/tester.go
```diff
@@ -89,33 +131,37 @@ func (vt *Tester) SetRepoPath(version provider.Version, repoPath string) {
...
 // Should be run from the base dir.
-func (vt *Tester) FetchCassettes(version provider.Version, baseBranch, prNumber string) error {
+func (vt *Tester) FetchCassettes(version provider.Version, baseBranch, head string) error {
 	_, ok := vt.cassettePaths[version]
 	if ok {
 		return nil
 	}
 	cassettePath := filepath.Join(vt.baseDir, "cassettes", version.String())
 	vt.rnr.Mkdir(cassettePath)
+	fetchErrors := make([]error, 0, 3)
```
melinath, line 141:
```suggestion
	var fetchErrors []error
```

.ci/magician/vcr/tester.go
```diff
@@ -89,33 +131,37 @@ func (vt *Tester) SetRepoPath(version provider.Version, repoPath string) {
...
+	if head != "" {
+		bucketPath := fmt.Sprintf("gs://%s/%srefs/heads/%s/fixtures/*", vt.cassetteBucket, version.BucketPath(), head)
 		if err := vt.fetchBucketPath(bucketPath, cassettePath); err != nil {
-			fmt.Println("Error fetching cassettes: ", err)
+			fetchErrors = append(fetchErrors, fmt.Errorf("error fetching bucket path %s: %v", bucketPath, err))
 		}
 	}
 	vt.cassettePaths[version] = cassettePath
+	if len(fetchErrors) > 0 {
+		return fmt.Errorf("errors fetching cassettes: %v", fetchErrors)
```
melinath, line 163: I think the changes to the error handling here are incorrect as written. I'm fairly certain the fetches can fail in valid circumstances - for example, if a PR is opened there won't be any cassettes, so the cassettes for that PR will "fail" to be fetched. See https://pantheon.corp.google.com/cloud-build/builds;region=global/fadf17a3-a053-400a-871a-95e1d1b6251f;step=23?inv=1&invt=AbbbyA&project=graphite-docker-images line 12-16 for example.

Why do we need these changes? What would you think about holding them for a later PR?

melinath (changes requested):
a few change requests below. This is a lot of unrelated changes at once so I'm a little concerned I may have missed issues. Please be sure to exercise the VCR test logic (including record after replay) to make sure it's all working as expected.

melinath:
It could be worth splitting this into multiple smaller PRs.

.ci/magician/vcr/tester.go
trodge: We don't need to change the error handling now. I'll hold off for a later PR.

.ci/magician/cmd/check_cassettes.go
trodge: Unintentional change, reverting to the correct bucket now.

.ci/magician/cmd/vcr_cassette_update.go
trodge: `"ci-vcr-cassettes"` was previously hard-coded into the `FetchCassettes` function. `"ci-vcr-logs"` would only be used if `vt.UploadLogs` were called.

.ci/magician/exec/runner.go
trodge: Does this look correct to you?

This function doesn't get called yet, so I'll save testing it for the next PR.

melinath: That looks about right; I'm not super familiar with doing this though.

roaks3 (commented):
Generally speaking I don't see anything wrong here, but I would second what Stephen suggested about splitting this up into a few smaller PRs. The difficulty in reading changes like this all together is the context switching that I need to do as I'm reading line-by-line, paired with a things like signature changes, so it is hard for me to feel confident that I haven't missed something. FWIW I think it would save you effort by having each individual review move faster.

That said, pending enough testing that this feels safe, 👍  from me.

.ci/magician/cmd/test_terraform_vcr.go
```diff
@@ -120,7 +126,7 @@ var testTerraformVCRCmd = &cobra.Command{
 		}
 		ctlr := source.NewController(env["GOPATH"], "modular-magician", env["GITHUB_TOKEN_DOWNSTREAMS"], rnr)
 
-		vt, err := vcr.NewTester(env, rnr)
+		vt, err := vcr.NewTester(env, "ci-vcr-logs", "vcr-check-cassettes", rnr)
```
melinath, line 129: this is inconsistent with the previous usage of vcr-check-cassettes (and it wasn't present in this file previously) - is this intentional?

melinath (changes requested):
Could you split this into multiple PRs?

- one for all the trivial changes like moving the template files
- one per major set of changes (for example, one for changes to make uploadlogs take a struct, one for moving logBucket and cassetteBucket into the tester, one for replacing prnumber with head, etc.)

.ci/magician/cmd/test_terraform_vcr.go
melinath: if it's an error - this should be easier to check after the PRs are split out

trodge: It is an error, and I'm reverting it now.

Going to work on the trivial changes PR first and then squash and rebase this PR onto it.

.ci/magician/vcr/tester.go
```diff
@@ -156,34 +194,34 @@ type RunOptions struct {
 
 // Run the vcr tests in the given mode and provider version and return the result.
 // This will overwrite any existing logs for the given mode and version.
-func (vt *Tester) Run(opt RunOptions) (Result, error) {
-	logPath, err := vt.getLogPath(opt.Mode, opt.Version)
+func (vt *Tester) Run(opts RunOptions) (Result, error) {
```
melinath, line 197:
```suggestion
func (vt *Tester) Run(opt RunOptions) (Result, error) {
```

Could this rename be excluded for clarity?

.ci/magician/vcr/tester.go
```diff
@@ -80,10 +118,10 @@ func NewTester(env map[string]string, logBucket, cassetteBucket string, rnr Exec
 	return &Tester{
 		env:            env,
 		rnr:            rnr,
-		logBucket:      logBucket,
-		cassetteBucket: cassetteBucket,
```
melinath, line 84: seems like maybe this move was included unintentionally?

.ci/magician/vcr/tester.go
```diff
@@ -50,10 +50,10 @@ type logKey struct {
 type Tester struct {
 	env            map[string]string           // shared environment variables for running tests
 	rnr            ExecRunner                  // for running commands and manipulating files
-	logBucket      string                      // GCS bucket name to store logs
-	cassetteBucket string                      // GCS bucket name to store cassettes
```
melinath, line 54: seems like maybe this move was included unintentionally?

.ci/magician/cmd/vcr_cassette_update_test.go
```diff
@@ -389,7 +389,7 @@ func TestExecVCRCassetteUpdate(t *testing.T) {
 			ctlr := source.NewController("gopath", "hashicorp", "token", rnr)
 			vt, err := vcr.NewTester(map[string]string{
 				"SA_KEY": "sa_key",
-			}, "", "ci-vcr-cassettes", rnr)
+			}, "ci-vcr-logs", "ci-vcr-cassettes", rnr)
```
melinath, line 392: seems like maybe this was included unintentionally? If not, why is it changing?

.ci/magician/cmd/test_terraform_vcr.go
```diff
@@ -73,6 +73,9 @@ type withReplayFailedTests struct {
 type withoutReplayFailedTests struct {
 	ReplayingErr error
 	PRNumber     string
+	Head         string
```
melinath, line 76: If I'm reading right, these changes seem like they aren't necessary for this round of changes - they change what goes into the template, but the templates aren't changing to use the new data. Could these be excluded for now?

.ci/magician/cmd/test_terraform_vcr.go
```diff
@@ -491,21 +506,21 @@ func formatComment(fileName string, tmplText string, data any) (string, error) {
 }
 
 func formatTestsAnalytics(data analytics) (string, error) {
-	return formatComment("test_terraform_vcr_test_analytics.tmpl", testsAnalyticsTmplText, data)
+	return formatComment("test_analytics.tmpl", testsAnalyticsTmplText, data)
```
melinath, line 509: Happy to leave these renames since they're trivial and just in this one spot. I thought we already did them for some reason...

melinath (changes requested):

.ci/magician/cmd/test_terraform_vcr.go
trodge: We renamed the files, but didn't rename the `fileName` which gets passed to `template.New`. Nothing broke because the files are embedded by their correct names.

.ci/magician/cmd/vcr_cassette_update_test.go
trodge: I thought we might be calling `UploadLogs` somewhere using this tester, but it appears we are not.

I'm reverting back to `""` now.

.ci/magician/cmd/test_terraform_vcr.go
trodge: Done.

.ci/magician/vcr/tester.go
trodge: I wanted these to be in a logical order. I'm flipping `cassetteBucket` with `logBucket` now to be consistent with `cassettePaths` and `logPaths` and moving them back up so that `baseDir` is together with all the `*Path(s)`.

I'm also switching the order of parameters for `NewTester` to put cassettes before logs, since cassettes come first.

.ci/magician/vcr/tester.go
trodge: Done

.ci/magician/vcr/tester.go
melinath: wfm

.ci/magician/cmd/test_terraform_vcr.go
```diff
@@ -261,16 +263,17 @@ func execTestTerraformVCR(prNumber, mmCommitSha, buildID, projectID, buildStep,
...
 		if err := vt.UploadLogs(vcr.UploadLogsOptions{
-			PRNumber: prNumber,
-			BuildID:  buildID,
-			Parallel: true,
-			Mode:     vcr.Recording,
-			Version:  provider.Beta,
+			Head:           newBranch,
+			BuildID:        buildID,
+			AfterRecording: true,
+			Parallel:       false,
```
melinath, line 274:
```suggestion
			Parallel:       true,
```

.ci/magician/cmd/test_terraform_vcr.go
```diff
@@ -196,10 +196,12 @@ func execTestTerraformVCR(prNumber, mmCommitSha, buildID, projectID, buildStep,
...
 
 	if err := vt.UploadLogs(vcr.UploadLogsOptions{
-		PRNumber: prNumber,
-		BuildID:  buildID,
-		Mode:     vcr.Replaying,
-		Version:  provider.Beta,
+		Head:           newBranch,
+		BuildID:        buildID,
+		AfterRecording: false,
+		Parallel:       false,
```
melinath, line 202:
```suggestion
```

melinath (approved):
LGTM apart from a couple things that seem incorrect / unnecessary.

.ci/magician/cmd/test_terraform_vcr.go
```diff
@@ -261,16 +261,16 @@ func execTestTerraformVCR(prNumber, mmCommitSha, buildID, projectID, buildStep,
...
 
 		if err := vt.UploadLogs(vcr.UploadLogsOptions{
-			PRNumber: prNumber,
-			BuildID:  buildID,
-			Parallel: true,
-			Mode:     vcr.Recording,
-			Version:  provider.Beta,
+			Head:           newBranch,
+			BuildID:        buildID,
+			AfterRecording: true,
```
melinath, line 271: It looks like this unintentionally changed?

```suggestion
			Parallel: true,
```

melinath (approved):

magician-vcr-eap
shuyama1, line 1: Is this intentional? It might be causing some of the check failures in the main section.

melinath: https://github.com/GoogleCloudPlatform/magic-modules/pull/11846 fixes this.

PR #11652, opened 2024-09-06
Add `host_error_timeout_seconds` field to `compute_instance` on beta provider

karolgorc:
-Added field to `google_compute_instance` and `google_compute_instance_template`
-Added tests
-Added doc changes
-Added data_source support

Do you want me to introduce plan-time validation of this field or should the API handle errors on this field?

```release-note:enhancement
compute: added `host_error_timeout_seconds` to `google_compute_instance` (beta)
```

### DIFF
--- a/mmv1/third_party/terraform/services/compute/resource_compute_instance_template_test.go.erb
+++ b/mmv1/third_party/terraform/services/compute/resource_compute_instance_template_test.go.erb
@@ -158,6 +158,44 @@ func TestAccComputeInstanceTemplate_maintenance_interval(t *testing.T) {
 }
 <% end -%>
 
+<% unless version == "ga" -%>
+func TestAccComputeInstanceTemplate_hostErrorTimeoutSeconds(t *testing.T) {
+	t.Parallel()
+
+	var instanceTemplate compute.InstanceTemplate
+
+	acctest.VcrTest(t, resource.TestCase{
+		PreCheck:                 func() { acctest.AccTestPreCheck(t) },
+		ProtoV5ProviderFactories: acctest.ProtoV5ProviderFactories(t),
+		CheckDestroy:             testAccCheckComputeInstanceTemplateDestroyProducer(t),
+		Steps: []resource.TestStep{
+			{
+				Config: testAccComputeInstanceTemplate_hostErrorTimeoutSeconds(acctest.RandString(t, 10)),
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckComputeInstanceTemplateExists(
+						t, "google_compute_instance_template.foobar", &instanceTemplate),
+					resource.TestCheckResourceAttr("google_compute_instance_template.foobar", "scheduling.0.host_error_timeout_seconds", "120"),
+				),
+			},
+			{
+				ResourceName:      "google_compute_instance_template.foobar",
+				ImportState:       true,
+				ImportStateVerify: true,
+				ImportStateVerifyIgnore: []string{"labels", "terraform_labels"},
+			},
+			{
+				Config: testAccComputeInstanceTemplate_basic(acctest.RandString(t, 10)),
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckComputeInstanceTemplateExists(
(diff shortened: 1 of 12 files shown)


mmv1/third_party/terraform/website/docs/d/compute_instance.html.markdown
```diff
@@ -186,6 +186,8 @@ The following arguments are supported:
 
 <a name="nested_scheduling"></a>The `scheduling` block supports:
 
+* `host_error_timeout_seconds` - [Beta](https://terraform.io/docs/providers/google/guides/provider_versions.html) Time in seconds for host error detection
```
NickElliot, line 189:
```suggestion
* `host_error_timeout_seconds` - [Beta](https://terraform.io/docs/providers/google/guides/provider_versions.html) Time in seconds for host error detection.
```

mmv1/third_party/terraform/website/docs/d/compute_instance_template.html.markdown
```diff
@@ -276,7 +276,9 @@ The `disk_encryption_key` block supports:
    groups will use as host systems. Read more on sole-tenant node creation
    [here](https://cloud.google.com/compute/docs/nodes/create-nodes).
    Structure [documented below](#nested_node_affinities).
-   
+
+* `host_error_timeout_seconds` - [Beta](https://terraform.io/docs/providers/google/guides/provider_versions.html) Time in seconds for host error detection
```
NickElliot, line 280:
```suggestion
* `host_error_timeout_seconds` - [Beta](https://terraform.io/docs/providers/google/guides/provider_versions.html) Time in seconds for host error detection.
```

mmv1/third_party/terraform/website/docs/d/compute_region_instance_template.html.markdown
```diff
@@ -260,7 +260,9 @@ The `disk_encryption_key` block supports:
    groups will use as host systems. Read more on sole-tenant node creation
    [here](https://cloud.google.com/compute/docs/nodes/create-nodes).
    Structure [documented below](#nested_node_affinities).
-   
+
+* `host_error_timeout_seconds` - [Beta](https://terraform.io/docs/providers/google/guides/provider_versions.html) Time in seconds for host error detection
```
NickElliot, line 264:
```suggestion
* `host_error_timeout_seconds` - [Beta](https://terraform.io/docs/providers/google/guides/provider_versions.html) Time in seconds for host error detection.
```

NickElliot (approved):
if the API receiving an improper value aborts the entire operation, then it should be fine to allow the API to handle it. The concern with Terraform validation would be adding a maintenance cost should the values involved change. The reason to have it handled in Terraform is if the error could result in a potential permadiff or mismatch in state between local and remote.

NickElliot (changes requested):
misclick

karolgorc:
Just to clarify. Yes the error-handling is fully supported on the API side. Any incorrect input will fail the apply so there is no risk of having a permadiff.

NickElliot (commented):
Need to ask one last question: for the "default behavior" is `host_error_timeout_seconds` still being returned in GET requests, or is the field not returned at all?

karolgorc:
While testing your question i found a bug that probably has something to do with the additional function for Diff on `scheduling` block. When removing the `host_error_timeout_seconds` field, terraform doesn't see a change in the config. Will work on a fix for this and write additional test

NickElliot (approved):
Given it's part of behavior for the full block I think it's fine to approve as is, and introduce a new PR for fixing the wider bug. Thanks for the contribution!

mmv1/third_party/terraform/services/compute/resource_compute_instance_template_test.go.erb
```diff
@@ -158,6 +158,44 @@ func TestAccComputeInstanceTemplate_maintenance_interval(t *testing.T) {
 }
 <% end -%>
 
+<% unless version == "ga" -%>
```
NickElliot, line 161:
```suggestion
<% unless version == 'ga' -%>
```

PR #11672, opened 2024-09-10
Add state "SUSPENDED" to `desired_status` field on `google_compute_instance`

karolgorc:
closes https://github.com/hashicorp/terraform-provider-google/issues/11374

- Added the field
- Added tests for it
- Added doc changes

There is an additional piece of logic that uses Resume instead of Start method when necessary to simplify the UX when using this. So there is no need to add a "RESUMED" state.

```release-note:enhancement
compute: add "SUSPENDED" as a possible value to `desired_state` on `google_compute_instance`
```

### DIFF
--- a/mmv1/third_party/terraform/services/compute/resource_compute_instance.go.erb
+++ b/mmv1/third_party/terraform/services/compute/resource_compute_instance.go.erb
@@ -1131,8 +1131,8 @@ be from 0 to 999,999,999 inclusive.`,
 			"desired_status": {
 				Type:         schema.TypeString,
 				Optional:     true,
-				ValidateFunc: validation.StringInSlice([]string{"RUNNING", "TERMINATED"}, false),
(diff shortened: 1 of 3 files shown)


roaks3 (commented):
LGTM pending tests

roaks3:
Looks like a compiler error:
```
Error: google-beta/services/compute/resource_compute_instance.go:2442:83: not enough arguments in call to config.NewComputeClient(userAgent).Instances.Resume
	have (string, string, string)
	want (string, string, string, *"google.golang.org/api/compute/v0.beta".InstancesResumeRequest)
make: *** [GNUmakefile:29: vet] Error 1
```

karolgorc:
The compiler error was due to the fact that the resume function takes different arguments in the beta API go package. Fixed now

roaks3 (approved):

PR #11674, opened 2024-09-10
Address google_compute_router_nat nat_ips dependency issue with updates to docs

maxi-cit:
Fixes: https://github.com/hashicorp/terraform-provider-google/issues/6812

This PR is based on work done by @matheusaleixo-cit [here](https://github.com/GoogleCloudPlatform/magic-modules/pull/11390) but focuses on reducing the impact from the fix

```release-note:bug
compute: added guidance in documentation to resolve dependency issues that block deletion of `google_compute_address` resources when they're referenced in `google_compute_router_nat` resources, specifically  when using `count` references
```

### DIFF
--- a/mmv1/products/compute/RouterNat.yaml
+++ b/mmv1/products/compute/RouterNat.yaml
@@ -158,8 +158,7 @@ properties:
     description: |
       Self-links of NAT IPs. Only valid if natIpAllocateOption
       is set to MANUAL_ONLY.
-
-      ~> **Note:** If this field is used alongside with a count created list of address resources `google_compute_address.foobar.*.self_link`,
+      If this field is used alongside with a count created list of address resources `google_compute_address.foobar.*.self_link`,
       the access level resource for the address resource must have a `lifecycle` block with `create_before_destroy = true` so
       the number of resources can be increased/decreased without triggering the `resourceInUseByAnotherResource` error.
     send_empty_value: true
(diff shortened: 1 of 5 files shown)


SarahFrench (commented):
Looks good! Another change that would be good to include in this PR is updating [terraform/examples/router_nat_manual_ips.tf.erb](https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/templates/terraform/examples/router_nat_manual_ips.tf.erb) to use the lifecycle block. This file is used to add [this example to the documentation](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/compute_router_nat#example-usage---router-nat-manual-ips), so updating it will help users who copy-paste examples without reading the docs fully.

SarahFrench:
The TestAccComputeInstance_confidentialHyperDiskBootDisk test failure [here](https://github.com/GoogleCloudPlatform/magic-modules/pull/11674#issuecomment-2344325830) is unrelated to this PR

SarahFrench (approved):
Thanks for this PR! [This test](https://github.com/GoogleCloudPlatform/magic-modules/pull/11674#issuecomment-2341519549) shows create_before_destroy is sufficient to fix the issue, and the changes to the docs will help users navigate the issue (either by reading the guidance or copy/pasting an example with the necessary protections already in it)

PR #11675, opened 2024-09-11
Public outbound ip 

prabhav2302:
This PR adds the enableOutboundPublicIp boolean field and outboundPublicIpAddresses (output only field) to the AlloyDB instance resource. Setting this field to true enables public outbound IP within the AlloyDB instance. PR addresses (b/353749975). It also adds testing which covers:
1. Creating an instance with the enableOutboundPublicIp set to true
2. Updating an instance with the enableOutboundPublicIp set to false
3. Updating an instance with the enableOutboundPublicIp set to true

```release-note:enhancement
alloyDB: added `enableOutboundPublicIp` field to `google_alloydb_instance` resource
```


hao-nan-li (approved):
`output:true` fields is optional, but it's fine to have it here.

PR #11676, opened 2024-09-11
fix: Match all whitespace and non-whitespace for Pub/Sub subscription filter validation

michaelpri10:
This change the regular expression validation for Pub/Sub subscriptions to use `[\s\S]` instead of `.` when matching characters to match both whitespace and non-whitespace characters. Because we allow empty filters (see [this issue](https://github.com/hashicorp/terraform-provider-google/issues/19269)), having only whitespace characters (which would be matched by this filter) is valid as well. This will not introduce any breaking changes as it only expands the character set for subscription filters.

This fixes [this issue](https://github.com/hashicorp/terraform-provider-google/issues/19296).

```release-note:bug
Allows Pub/Sub subscription filters to contain line breaks
```

### DIFF
--- a/mmv1/products/pubsub/Subscription.yaml
+++ b/mmv1/products/pubsub/Subscription.yaml
@@ -396,7 +396,7 @@ properties:
     name: 'filter'
     required: false
     validation: !ruby/object:Provider::Terraform::Validation
-      regex: '^.{0,256}$'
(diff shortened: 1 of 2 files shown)


slevenick (commented):
Can you add a test that uses a whitespace character so we can ensure this use case works in the future?

slevenick (approved):

PR #11679, opened 2024-09-11
FixDoc Update subnetwork purpose description

tferazzi:
Update the description of subnetwork purpose parameter.
[API docs](https://cloud.google.com/compute/docs/reference/rest/v1/subnetworks) reference the purpose `PRIVATE` while the [terraform-provider](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/compute_subnetwork#purpose) still mentions `PRIVATE_RFC_1918`. Specifying `PRIVATE_RFC_1918` does not result in an error but causes the subnetwork to be re-created on every apply.

Fixes https://github.com/hashicorp/terraform-provider-google/issues/16539

```release-note:none

```

### DIFF
--- a/mmv1/products/compute/Subnetwork.yaml
+++ b/mmv1/products/compute/Subnetwork.yaml
@@ -214,13 +214,13 @@ properties:
     name: 'purpose'
     immutable: true
     description: |
-      The purpose of the resource. This field can be either `PRIVATE_RFC_1918`, `REGIONAL_MANAGED_PROXY`, `GLOBAL_MANAGED_PROXY`, `PRIVATE_SERVICE_CONNECT` or `PRIVATE_NAT`([Beta](https://terraform.io/docs/providers/google/guides/provider_versions.html)).
+      The purpose of the resource. This field can be either `PRIVATE`, `REGIONAL_MANAGED_PROXY`, `GLOBAL_MANAGED_PROXY`, `PRIVATE_SERVICE_CONNECT` or `PRIVATE_NAT`([Beta](https://terraform.io/docs/providers/google/guides/provider_versions.html)).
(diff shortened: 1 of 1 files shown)


mmv1/products/compute/Subnetwork.yaml
```diff
@@ -210,17 +210,23 @@ properties:
...
-  - !ruby/object:Api::Type::String
+  - !ruby/object:Api::Type::Enum
     name: 'purpose'
     immutable: true
+    values:
+      - :PRIVATE
+      - :REGIONAL_MANAGED_PROXY
+      - :GLOBAL_MANAGED_PROXY
+      - :PRIVATE_SERVICE_CONNECT
+      - :PRIVATE_NAT
```
NickElliot, line 221:
```suggestion
```

mmv1/products/compute/Subnetwork.yaml
```diff
@@ -210,17 +210,23 @@ properties:
     immutable: true
     required: true
     custom_expand: 'templates/terraform/custom_expand/resourceref_with_validation.go.erb'
-  - !ruby/object:Api::Type::String
+  - !ruby/object:Api::Type::Enum
```
NickElliot, line 213:
```suggestion
  - !ruby/object:Api::Type::String
```

NickElliot (changes requested):
Hi, in this case we don't want to alter the actual field behavior because that is adding a maintenance cost for a field which has proven it will change accepted values. Just updating our documentation of what is accepted/not is fine. Thanks for making the update!

mmv1/products/compute/Subnetwork.yaml
tferazzi: Hi, thanks for your review. I removed the commit containing the var change and only kept the documentation updates.

NickElliot (approved):
LGTM, thanks!

PR #11683, opened 2024-09-11
Add includeImportRanges fields to NCC Spoke resource

benjamin-maynard:
This PR adds the `include_import_ranges` to the `google_network_connectivity_spoke` resource. Added for `linked_vpn_tunnels`, `linked_interconnect_attachments` and `linked_router_appliance_instances`

```release-note:enhancement
networkconnectivity: added `include_import_ranges` field to `google_network_connectivity_spoke` resource for `linked_vpn_tunnels`, `linked_interconnect_attachments` and `linked_router_appliance_instances`
```

### DIFF
--- a/mmv1/products/networkconnectivity/Spoke.yaml
+++ b/mmv1/products/networkconnectivity/Spoke.yaml
@@ -114,6 +114,13 @@ properties:
         min_version: ga
         required: true
         immutable: true
+      - !ruby/object:Api::Type::Array
+        name: includeImportRanges
+        description: |
+          IP ranges allowed to be included during import from hub (does not control transit connectivity).
+          The only allowed value for now is "ALL_IPV4_RANGES".
+        item_type: Api::Type::String
+        default_from_api: true
   - !ruby/object:Api::Type::NestedObject
     name: linkedInterconnectAttachments
     description: A collection of VLAN attachment resources. These resources should be redundant attachments that all advertise the same prefixes to Google Cloud. Alternatively, in active/passive configurations, all attachments should be capable of advertising the same prefixes.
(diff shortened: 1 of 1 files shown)


benjamin-maynard:
Will take a look at tests, there does not appear to be any existing tests that cover `linked_interconnect_attachments` and `linked_vpn_tunnels` at the moment. These former is also not trivial to create.

mmv1/products/networkconnectivity/Spoke.yaml
```diff
@@ -114,6 +141,13 @@ properties:
         min_version: ga
         required: true
         immutable: true
+      - !ruby/object:Api::Type::Array
+        name: includeImportRanges
+        description: |
+          IP ranges allowed to be included during import from hub (does not control transit connectivity).
+          The only allowed value for now is "ALL_IPV4_RANGES".
+        item_type: Api::Type::String
+        default_from_api: true
```
SarahFrench, line 150: Thanks for adding the test coverage for the different `linked_x` blocks!

A consequence of making this field default from the API is that users will have trouble setting this field to an empty value once the field is set. Is there any scenario where a user might want/need to set this to an empty array?

benjamin-maynard: Fair point - I added incase customers were already managing this by manually mutating the resource due to no TF support, but there are scenarios where customers may want to set this to an empty array. Will update.

SarahFrench: Thanks!

benjamin-maynard:
Hey @SarahFrench,

Do I need to worry about the test passing? Seems intermittent:

Failed: https://github.com/GoogleCloudPlatform/magic-modules/pull/11683#issuecomment-2352819160
But passed when re-run without changes: https://github.com/GoogleCloudPlatform/magic-modules/pull/11683#issuecomment-2355004370

SarahFrench (approved):
Thanks for updating that field to not default from the API!

The test failure you mentioned looks like a transient problem with how we record and re-run tests and not a problem with the test itself. I'm re-running the test in our other test environment to confirm this and will merge once it completes, assuming all's ok

SarahFrench:
Test re-runs look good, merging

PR #11688, opened 2024-09-12
Do not force send insecure_kubelet_readonly_port_enabled during creation

shuyama1:
potentially fixes https://github.com/hashicorp/terraform-provider-google/issues/19428

```release-note:bug
container: fixed a bug where specifying `node_pool_defaults.node_config_defaults` with `enable_autopilot = true` will cause `google_container_cluster` resource creation failure
```

### DIFF
--- a/mmv1/third_party/terraform/services/container/node_config.go.erb
+++ b/mmv1/third_party/terraform/services/container/node_config.go.erb
@@ -816,7 +816,6 @@ func expandNodeConfigDefaults(configured interface{}) *container.NodeConfigDefau
 	if v, ok := config["insecure_kubelet_readonly_port_enabled"]; ok {
 		nodeConfigDefaults.NodeKubeletConfig = &container.NodeKubeletConfig{
 			InsecureKubeletReadonlyPortEnabled: expandInsecureKubeletReadonlyPortEnabled(v),
-			ForceSendFields: []string{"InsecureKubeletReadonlyPortEnabled"},
 		}
 	}
 	if variant, ok := config["logging_variant"]; ok {
--- a/mmv1/third_party/terraform/services/container/node_config.go.erb
+++ b/mmv1/third_party/terraform/services/container/node_config.go.erb
@@ -101,18 +101,11 @@ func schemaLoggingVariant() *schema.Schema {
 }
 
 func schemaGcfsConfig() *schema.Schema {
-<<<<<<< HEAD
 	return &schema.Schema{
 		Type:        schema.TypeList,
 		Optional:    true,
 		Computed:    true,
 		MaxItems:    1,
-=======
-        return &schema.Schema{
-                Type:     schema.TypeList,
-		Optional: true,
-		MaxItems: 1,
->>>>>>> 8b0ff8b92 (revert change for GcfsConfig)
 		Description: `GCFS configuration for this node.`,
 		Elem:        &schema.Resource{
 			Schema: map[string]*schema.Schema{
--- a/mmv1/third_party/terraform/services/container/node_config.go.erb
+++ b/mmv1/third_party/terraform/services/container/node_config.go.erb
@@ -101,11 +101,18 @@ func schemaLoggingVariant() *schema.Schema {
 }
 
 func schemaGcfsConfig() *schema.Schema {
+<<<<<<< HEAD
 	return &schema.Schema{
 		Type:        schema.TypeList,
 		Optional:    true,
 		Computed:    true,
 		MaxItems:    1,
+=======
+        return &schema.Schema{
+                Type:     schema.TypeList,
+		Optional: true,
+		MaxItems: 1,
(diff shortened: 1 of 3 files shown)


wyardley:
@shuyama1 sorry if I introduced the issue described in https://github.com/hashicorp/terraform-provider-google/issues/19428 in #11272 or #11573... from what I remember, forcesendfields may have been needed in some situation, (which might explain the failing test?), so maybe some additional code changes will be needed to make this work vs. just removing it outright? Could be wrong, though.

Maybe the setting can just be disallowed or ignored when the cluster is an autopilot cluster?

Side note: I think the release note is not fully filled out above, unless that's something that will get updated later?

shuyama1:
> @shuyama1 sorry if I introduced the issue described in [hashicorp/terraform-provider-google#19428](https://github.com/hashicorp/terraform-provider-google/issues/19428) in #11272 or #11573... from what I remember, forcesendfields may have been needed in some situation, (which might explain the failing test?), so maybe some additional code changes will be needed to make this work vs. just removing it outright? Could be wrong, though.
> 
> Maybe the setting can just be disallowed or ignored when the cluster is an autopilot cluster?


@wyardley Hey, thanks for working on these feature implementations! GKE cluster resource may have some edge cases that that're a bit tricky to spot in advance - this seems like one. I'm testing locally o determine if forcesendfields is necessary here - removing it from the creation seems to resolve the issue described in the ticket. The test failure in this PR is due to a permadiff caused by another field. I'm investigating the cause of it and working on a fix as well. 
```
  ~ node_pool_defaults {
      ~ node_config_defaults {
            # (2 unchanged attributes hidden)

          - gcfs_config {
              - enabled = true -> null
            }
        }
    }
```

> 
> Side note: I think the release note is not fully filled out above, unless that's something that will get updated later?

Thanks for catching it! Yes, I was planning to update it before merging the PR

wyardley:
> I'm investigating the cause of it and working on a fix as well.

@shuyama1: #11717 (which you're also assigned to) should fix it, I believe

wyardley:
@shuyama1 it's possible it should also be removed in https://github.com/GoogleCloudPlatform/magic-modules/blob/1cee8f692420fc1a7fa4fc35affe34c0cd34614b/mmv1/third_party/terraform/services/container/node_config.go.erb#L818 as well, if the tests pass.

shuyama1:
> @shuyama1 it's possible it should also be removed in
> 
> https://github.com/GoogleCloudPlatform/magic-modules/blob/1cee8f692420fc1a7fa4fc35affe34c0cd34614b/mmv1/third_party/terraform/services/container/node_config.go.erb#L818
> 
> as well, if the tests pass.

Hey @wyardley Yes, I believe that removing the forced send of `InsecureKubeletReadonlyPortEnabled` should resolve the issue listed in the description of this PR. Additionally, your PR [#11717](https://github.com/GoogleCloudPlatform/magic-modules/pull/11717) should address the premadiff discovered in the test failure for this PR. Thank you for your work on the fix!

wyardley:
> Additionally, your PR https://github.com/GoogleCloudPlatform/magic-modules/pull/11717 should address the premadiff discovered in the test failure for this PR. Thank you for your work on the fix!

Thanks! Please also be aware of #11697

Hopefully, these 3 can all go out together in 6.4.0 or 6.5.0; hopefully they can get squeezed in

wyardley:
> Yes, I believe that removing the forced send of InsecureKubeletReadonlyPortEnabled should resolve the issue listed in the description of this PR

Oh, I get it now, this is for `node_config_defaults.node_kubelet_config` only. `node_config.kubelet_config` is the one I was thinking was originally getting updated here 🤦 

https://github.com/GoogleCloudPlatform/magic-modules/blob/20ac9f9559b3c12d6d637ff4e5f46f029d86997d/mmv1/third_party/terraform/services/container/node_config.go.erb#L1218

wyardley (approved):

ScottSuarez:
@shuyama1, wouldn't this not allow the field to be unset? I notice that this field is marked as Computed in the schema?

https://github.com/shuyama1/magic-modules/blob/e48cdbdb59cf71edfd96fe0e4ada0dc413dc2157/mmv1/third_party/terraform/services/container/node_config.go.erb#L87

shuyama1:
> @shuyama1, wouldn't this not allow the field to be unset? I notice that this field is marked as Computed in the schema?
> 
> [shuyama1/magic-modules@`e48cdbd`/mmv1/third_party/terraform/services/container/node_config.go.erb#L87](https://github.com/shuyama1/magic-modules/blob/e48cdbdb59cf71edfd96fe0e4ada0dc413dc2157/mmv1/third_party/terraform/services/container/node_config.go.erb#L87)

I've only removed it from the ForceSendFields in create but not update, so I think it should be fine, but I can test it locally to confirm

ScottSuarez (approved):

PR #11690, opened 2024-09-12
Add deletion_protection field to Metastore Federation

SamanthaMathews:
Part of b/366209903

Part of https://github.com/hashicorp/terraform-provider-google/issues/18854

```
Metastore: added `deletion_protection` field to `dataproc_metastore_service` to make deleting them require an explicit intent. `google_active_directory_domain` resources now cannot be destroyed unless `deletion_protection = false` is set for the resource.
```


c2thorn:

abd-goog (approved):

PR #11691, opened 2024-09-12
Add lifecycle prevent destroy to Secure Source Manager examples

ericpang777:

```release-note:none
add lifecycle prevent destroy to Secure Source Manager instance and repository examples
```

### DIFF
--- a/mmv1/templates/terraform/examples/secure_source_manager_instance_basic.tf.erb
+++ b/mmv1/templates/terraform/examples/secure_source_manager_instance_basic.tf.erb
@@ -5,7 +5,7 @@ resource "google_secure_source_manager_instance" "<%= ctx[:primary_resource_id] 
       "foo" = "bar"
     }
 
-    # For preventing accidental deletions
+    # Prevent accidental deletions.
     lifecycle {
       prevent_destroy = "<%= ctx[:vars]['prevent_destroy'] %>"
     }
--- a/mmv1/templates/terraform/examples/secure_source_manager_instance_basic.tf.erb
+++ b/mmv1/templates/terraform/examples/secure_source_manager_instance_basic.tf.erb
@@ -4,4 +4,8 @@ resource "google_secure_source_manager_instance" "<%= ctx[:primary_resource_id] 
     labels = {
       "foo" = "bar"
     }
+
+    lifecycle {
+      prevent_destroy = "<%= ctx[:vars]['prevent_destroy'] %>"
+    }
 }
--- a/mmv1/templates/terraform/examples/secure_source_manager_instance_basic.tf.erb
+++ b/mmv1/templates/terraform/examples/secure_source_manager_instance_basic.tf.erb
@@ -5,6 +5,7 @@ resource "google_secure_source_manager_instance" "<%= ctx[:primary_resource_id] 
       "foo" = "bar"
     }
 
+    # For preventing accidental deletions
     lifecycle {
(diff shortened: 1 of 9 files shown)


ericpang777:
Is there any need to add oics_var_overrides? I can't tell what it is used for but some examples seem to use it...

mmv1/templates/terraform/examples/secure_source_manager_instance_basic.tf.erb
```diff
@@ -4,4 +4,8 @@ resource "google_secure_source_manager_instance" "<%= ctx[:primary_resource_id]
     labels = {
       "foo" = "bar"
     }
+
+    lifecycle {
```
umairidris, line 8: Should we add a comment explaining why we are adding this block?

ericpang777: Added.

slevenick (approved):
Seems fine to me as a docs-only change

slevenick:
@umairidris ping me when you are good with this and I'll merge it

ericpang777:
> Seems fine to me as a docs-only change

Hi Sam, do I need to add `oics_var_overrides`? I notice some tests that have `test_var_overrides` also use it, but I have no idea what it does.

mmv1/templates/terraform/examples/secure_source_manager_instance_cmek.tf.erb
```diff
@@ -23,6 +23,11 @@ resource "google_secure_source_manager_instance" "<%= ctx[:primary_resource_id]
     depends_on = [
       google_kms_crypto_key_iam_member.crypto_key_binding
     ]
+
+    # For preventing accidental deletions
```
umairidris, line 27: Nit:

Prevent accidental deletions.

(full sentence ending with period)

ericpang777: Done

slevenick:
> > Seems fine to me as a docs-only change
> 
> Hi @slevenick, do I need to add `oics_var_overrides`? I notice some tests that have `test_var_overrides` also use it, but I have no idea what it does.

Hmmm, probably not? oics is open in cloud shell, but I don't think you would need specific overrides for that?

slevenick:
> > > Seems fine to me as a docs-only change
> > 
> > 
> > Hi @slevenick, do I need to add `oics_var_overrides`? I notice some tests that have `test_var_overrides` also use it, but I have no idea what it does.
> 
> Hmmm, probably not? oics is open in cloud shell, but I don't think you would need specific overrides for that?

Well, actually thinking about the use case if someone creates one of these in cloud shell it's probably just for testing purposes and they'll want to delete it. So it may be good to include the same overrides for oics as are used in tests

umairidris (approved):

ericpang777:
> > > > Seems fine to me as a docs-only change
> > > 
> > > 
> > > Hi @slevenick, do I need to add `oics_var_overrides`? I notice some tests that have `test_var_overrides` also use it, but I have no idea what it does.
> > 
> > 
> > Hmmm, probably not? oics is open in cloud shell, but I don't think you would need specific overrides for that?
> 
> Well, actually thinking about the use case if someone creates one of these in cloud shell it's probably just for testing purposes and they'll want to delete it. So it may be good to include the same overrides for oics as are used in tests

Mhm that's a good point, added 👍

umairidris (approved):

PR #11696, opened 2024-09-12
Renaming all remaining skip_* attributes

NickElliot:
all attributes except `skip_vcr` renamed. There are a few in tpgtools override files, but tpgtools isnt being touched in go rewrite so it doesn't seem like it needs to happen with this batch

```release-note:none

```

### DIFF
--- a/mmv1/products/developerconnect/go_Connection.yaml
+++ b/mmv1/products/developerconnect/go_Connection.yaml
@@ -105,7 +105,7 @@ properties:
             type: String
             description: |
               Required. A SecretManager resource containing the OAuth token
-              that authorizes the connection. 
+              that authorizes the connection.
               Format: `projects/*/secrets/*/versions/*`.
             min_version: 'beta'
             required: true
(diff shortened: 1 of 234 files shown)


mmv1/products/developerconnect/go_Connection.yaml
```diff
@@ -0,0 +1,200 @@
...
+    description:
+      "Describes stage and necessary actions to be taken by the\nuser to
+      complete the installation. Used for GitHub and GitHub Enterprise\nbased connections. "
+    min_version: 'beta'
+    output: true
+    properties:
+      - name: 'stage'
+        type: String
+        description:
+          "Output only. Current step of the installation process. \n Possible
```
c2thorn, line 156: This line needs to be edited to not have `values:` be the first term in the line: https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/products/developerconnect/Connection.yaml#L143C21-L143C110

then I think this can be regenerated

c2thorn (approved):

PR #11700, opened 2024-09-13
Fix interconnect immutability issue causing macsec failure

harshithpatte-g:
fixes https://github.com/hashicorp/terraform-provider-google/issues/19410

Customer issue: b/365795490
When Customer tries to deploy an interconnect with macsec enabled. The deployment fails. The circuit needs to be deployed first before enabling macsec, however, if macsec is enabled after, it causes the circuit to be destroyed and recreated. 
Interconnect can't created with macsec enabled.

Fix: remove immutability from the interconnect resource and add it at field level.

```release-note: enhancement
compute: added in-place update in `google_compute_interconnect` resource except for `remote_location ` and `requested_features` fields
```


zli82016 (changes requested):
Can you please fix the failed test `TestAccComputeInterconnect_computeInterconnectMacsecTest`? Thanks.

zli82016 (approved):
LGTM. Thanks.

PR #11701, opened 2024-09-13
Update error codes in TestAccFrameworkProviderBasePath_setInvalidBasePath, TestAccProviderBasePath_setInvalidBasePath

SarahFrench:
Fixes https://github.com/hashicorp/terraform-provider-google/issues/19468

In response to a test failure here: https://hashicorp.teamcity.com/buildConfiguration/TerraformProviders_GoogleCloud_GOOGLE_BETA_NIGHTLYTESTS_GOOGLEBETA_PACKAGE_FWPROVIDER/233695?hideTestsFromDependencies=false&hideProblemsFromDependencies=false&expandBuildDeploymentsSection=false&expandBuildTestsSection=true&expandBuildChangesSection=true

It might be worth waiting to see if that failure repeats in future, as it's only failed once so far.

```release-note:none

```

### DIFF
--- a/mmv1/third_party/terraform/fwprovider/framework_provider_test.go.erb
+++ b/mmv1/third_party/terraform/fwprovider/framework_provider_test.go.erb
@@ -52,12 +52,12 @@ func TestAccFrameworkProviderBasePath_setInvalidBasePath(t *testing.T) {
 					},
 				},
 				Config:      testAccProviderBasePath_setBasePath("https://www.example.com/compute/beta/", acctest.RandString(t, 10)),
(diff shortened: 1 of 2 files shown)


SirGitsalot (commented):
Drive by code review (this failure just bit me too!): since this is a regex it could catch any future changes with something like:

```
"got HTTP response code [4|5][0-9][0-9] with body"
```

(or something close to that, I didn't actually try it)

ScottSuarez (approved):
non-blocking but the review from the other googler isn't a bad suggestion. Making the regex more flexible could be useful?

SarahFrench:
Thanks for the input, both! I'll merge once the automation completes

PR #11707, opened 2024-09-13
Fix api resource names for mmv1

roaks3:
@melinath really just sharing this for visibility. These are the names that need to be manually overridden to get from a TF resource to the API resource, so you can see roughly how many we have (and perhaps notice any patterns we could use to automate). At this point, I do expect we will need a new field like I have here to specify these name overrides.

This does not include some or all resources for the following APIs:
* apigee.googleapis.com
* appengine.googleapis.com
* bigquery.googleapis.com
* compute.googleapis.com
* containeranalysis.googleapis.com
* dns.googleapis.com
* documentai.googleapis.com
* firestore.googleapis.com
* iam.googleapis.com
* iap.googleapis.com
* identitytoolkit.googleapis.com
* integrations.googleapis.com
* ml.googleapis.com
* monitoring.googleapis.com
* notebooks.googleapis.com
* servicenetworking.googleapis.com
* sourcerepo.googleapis.com
* sqladmin.googleapis.com
* storage.googleapis.com

```release-note:none

```

### DIFF
--- a/mmv1/products/artifactregistry/go_VPCSCConfig.yaml
+++ b/mmv1/products/artifactregistry/go_VPCSCConfig.yaml
@@ -14,6 +14,7 @@
 # Warning: This is a temporary file, and should not be edited directly
 ---
 name: 'VPCSCConfig'
+api_resource_type: VpcscConfig
 description: |-
   The Artifact Registry VPC SC config that applies to a Project.
 min_version: 'beta'
--- a/mmv1/products/blockchainnodeengine/go_BlockchainNodes.yaml
+++ b/mmv1/products/blockchainnodeengine/go_BlockchainNodes.yaml
@@ -14,6 +14,7 @@
 # Warning: This is a temporary file, and should not be edited directly
 ---
 name: 'BlockchainNodes'
+api_resource_type: BlockchainNode
 description: |
   A representation of a blockchain node.
 references:
--- a/mmv1/api/resource.go
+++ b/mmv1/api/resource.go
@@ -312,6 +312,8 @@ type Resource struct {
 	Compiler string
 
 	ImportPath string
+
+	ApiResourceType string `yaml:"api_resource_type"`
 }
 
 func (r *Resource) UnmarshalYAML(unmarshal func(any) error) error {
--- a/mmv1/products/accessapproval/go_FolderSettings.yaml
+++ b/mmv1/products/accessapproval/go_FolderSettings.yaml
@@ -14,6 +14,7 @@
 # Warning: This is a temporary file, and should not be edited directly
 ---
 name: 'FolderSettings'
+api_resource_type: AccessApprovalSettings
 legacy_name: 'google_folder_access_approval_settings'
 description: |
   Access Approval enables you to require your explicit approval whenever Google support and engineering need to access your customer content.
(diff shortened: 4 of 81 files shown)


melinath:
thanks!

mmv1/products/artifactregistry/go_VPCSCConfig.yaml
```diff
@@ -14,6 +14,7 @@
 # Warning: This is a temporary file, and should not be edited directly
 ---
 name: 'VPCSCConfig'
```
melinath, line 16: Could we change this name to VpcscConfig instead? I think it might be equivalent in terms of the output?

mmv1/products/blockchainnodeengine/go_BlockchainNodes.yaml
```diff
@@ -14,6 +14,7 @@
 # Warning: This is a temporary file, and should not be edited directly
 ---
 name: 'BlockchainNodes'
```
melinath, line 16: uff, this shouldn't be pluralized... but it's been around long enough it probably doesn't make sense to rename :-/

melinath:
Looks like there are a few categories:

- Fine-grained resources
- Resources that appear multiple times with different parents (project, folder, and org variants)
- Resources that duplicate part of the product name (so the contributor left it out) - like google_bigquery_data_transfer_config (which technically should be google_bigquery_data_transfer_transfer_config)
- Resources where the contributor added _more_ information into the TF resource name (for example, google_cloudidentity_group_membership should technically just be google_cloudidentity_membership)

melinath:
We could _maybe_ try to automate this but tbh I think having an override field seems sufficient. What would you think about _defaulting_ to the API resource type and having the override be if you want to force a different TF resource name? That is, name would essentially be renamed to (or treated as) api_resource_type, and all the cases that don't map correctly would need an override (which would state, for example, the explicit terraform resource name or similar.)

roaks3:
> What would you think about defaulting to the API resource type and having the override be if you want to force a different TF resource name?

Ah yea, I see. That should be fine, I can only think of some minor impacts (but also maybe worth considering if there is an advantage to doing that semantically):
- Might be mildly more disruptive to the generator code, but hopefully not too big of a deal
- Slightly less direct / unexpected to fix an incorrect resource by changing a name and adding an override vs adding an override
- Name no longer maps to a lot of the generated go functions (or we make large-scale changes to them)

melinath:
yeah... it feels more "correct" to me, but I'm not entirely sure it's worth the churn (or at least, not now?)

mmv1/products/artifactregistry/go_VPCSCConfig.yaml
roaks3: It changes all of the function names used by the resource, which could hypothetically be used by other resources. I do think it is most likely safe to change, but I would advocate for just using the override for now, and then handling this later (there may be others we want to do the same for).

roaks3:
Yea, I think I'm going to move in the direction of keeping `api_resource_type`, mainly to reduce churn with the function names. We can always change it later with a more focused PR.

But, perhaps we want to rename the new field to `resource_type`, since the yaml should be describing the API?

melinath:
I'd maybe lean towards keeping `api` similar to `api_name` for fields. Maybe it should just be `api_name` at the resource level too? Unless we need to specifically override the domain... which we definitely do because cai_base_url exists for that exact reason...

roaks3:
Superseded by https://github.com/GoogleCloudPlatform/magic-modules/pull/12019

PR #11744, opened 2024-09-18
Autopilot Suppressed Diff Prevents using `additive_vpc_scope_dns_domain`

EmileHofsink:
Currently when creating an Autopilot Cluster in Terraform it is not possible to use `additive_vpc_scope_dns_domain` within `dns_config`, while this feature is supported by the console and API currently. This looks like it is due to the suppressed diff configuration for Autopilot. 

To adjust for this I have moved the suppressed diff configuration from the top level `dns_config` block to just apply to the elements that are unchangeable for Autopilot. 

```release-note:REPLACEME

```


maci0:
we had a conversation regarding this in this PR also: https://github.com/GoogleCloudPlatform/magic-modules/pull/11562

maci0:
```
  ~ update in-place

  # google_container_cluster.default will be updated in-place
  ~ resource "google_container_cluster" "default" {
        id                                       = "projects/enablement-404602/locations/us-central1/clusters/autopilot"
        name                                     = "autopilot"
        # (35 unchanged attributes hidden)

      ~ dns_config {
          ~ additive_vpc_scope_dns_domain = "foo.bar" -> "foo.baz"
        }

        # (31 unchanged blocks hidden)
    }

  Enter a value: yes

google_container_cluster.default: Modifying... [id=projects/enablement-404602/locations/us-central1/clusters/autopilot]
╷
│ Error: googleapi: Error 400: Changing DNSConfig is not supported in Autopilot clusters.
│ Details:
│ [
│   {
│     "@type": "type.googleapis.com/google.rpc.RequestInfo",
│     "requestId": "0xc3f07455c848394d"
│   }
│ ]
│ , badRequest
│ 
│   with google_container_cluster.default,
│   on main.tf line 15, in resource "google_container_cluster" "default":
│   15: resource "google_container_cluster" "default" {
│ 
```

EmileHofsink:
Thanks for taking a look @maci0 ! Previous discussion looks good and all makes sense. What I've done (and hopefully haven't overstepped anything here) is keep my previous change of moving the other settings to be suppressed except for `additive_vpc_dns_scope_domain`, but, have now included a component of `containerClusterAutopilotCustomizeDiff` that checks to see if the cluster is an autopilot and if `additive_vpc_dns_scope_domain` has changed and if so request recreation of the cluster.

maci0:
@slevenick could you take a look at this one?

rileykarson (commented):
Running tests- I think this change looks reasonable, will LGTM+merge after success.

rileykarson:
/gcbrun

Hmm- we shouldn't have hit quotas there. Weird. Maybe other runs going on at the same time.

EmileHofsink:
Thanks @rileykarson, I'm not sure if this would be considered a breaking change as it's an optional field and the diff suppress only affected autopilot clusters. But I could be missing something.

mmv1/third_party/terraform/services/container/resource_container_cluster.go.erb
```diff
@@ -2129,33 +2129,35 @@ func ResourceContainerCluster() *schema.Resource {
 				Type:        schema.TypeList,
 				Optional:    true,
 				MaxItems:    1,
-				DiffSuppressFunc: suppressDiffForAutopilot,
```
rileykarson, line 2132: Ah, this ends up being a breaking change in practice. Autopilot clusters that don't set a value at all get this back in the response:

```
  "dnsConfig": {
   "clusterDns": "CLOUD_DNS",
   "clusterDnsScope": "CLUSTER_SCOPE",
   "clusterDnsDomain": "cluster.local"
  },
```

Terraform will see a diff on that by removing the DSF on the top-level field, even though the subfields will be supressed. If there was not a default value for the block this would be safe.

rileykarson (commented):
Yeah, I was looking into that in old test logs. Forgot I had to retrigger the tests (which also would have caught it).

See specific reason I don't think we can make this change as-is inline.

hadleyrich:
I've just run into this issue and am unable to create an autopilot cluster with additive_vpc_scope_dns_domain. Is there another issue or PR I'm missing?

maci0:
There is no other PR for this currently. This one was closed due to inactivity and breaking some of the tests.

PR #11750, opened 2024-09-18
Add resource Dataproc batch

zli82016:
fixes https://github.com/hashicorp/terraform-provider-google/issues/11775

```release-note:new-resource
`google_dataproc_batch`
```

```release-note:enhancement
dataproc: switched to the v1 API for `google_dataproc_autoscaling_policy` resource (beta)
```


zli82016:
> ## Missing test report
> 
> Resource: `google_dataproc_batch` (5 total tests) Please add an acceptance test which includes these fields. The test should include the following:
> 
> ```hcl
> resource "google_dataproc_batch" "primary" {
>   runtime_config {
>     container_image = # value needed
>   }
>   spark_batch {
>     archive_uris      = # value needed
>     file_uris         = # value needed
>     main_jar_file_uri = # value needed
>   }
>   spark_r_batch {
>     archive_uris = # value needed
>     file_uris    = # value needed
>   }
> }
> ```

Hello, @melinath, I added the test for most fields. There are still a couple of fields missing the test. I talked with the service team, and they may need about 2 days to set up. They want to check if we are OK to go without those fields in the test.

mmv1/products/dataproc/product.yaml
```diff
@@ -11,13 +11,12 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+# Warning: This is a temporary file, and should not be edited directly
```
melinath, line 14: Why is this temporary?

melinath:
What are they setting up? Could we just leave those fields out of the resource for now?

zli82016:
> What are they setting up? Could we just leave those fields out of the resource for now?

They need to create a public GAR image, write spark jobs for these fields and create respective files and archives. The setup is just for testing.

mmv1/products/dataproc/product.yaml
zli82016: The comment should not be here. I removed it.

melinath:
If they can get it set up before merge that would definitely be preferable, but I think we can go ahead otherwise if the overall resource works. There's just a higher risk of bugs.

melinath (approved):
LGTM - Adding the tests would be great if it works out before the next cut, otherwise it's probably fine to go ahead with this.

melinath:
Actually it would be great to merge this Monday if possible.

zli82016:
I am going to merge this PR now and add the missing test later.

PR #11751, opened 2024-09-19
Add PlanCheck to update test creation

trodge:
This additional step will ensure that new update tests always update the resource in-place rather than delete and recreate.

```release-note:none

```

### DIFF
--- a/docs/content/develop/test/test.md
+++ b/docs/content/develop/test/test.md
@@ -178,6 +178,7 @@ An update test is a test that creates the target resource and then makes updates
    - Change the suffix of the test function to `_update`.
    - Copy the 2 `TestStep` blocks and paste them immediately after, so that there are 4 total test steps.
    - Change the suffix of the second `Config` value to `_update`.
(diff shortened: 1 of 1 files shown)


NickElliot (approved):
I'm seeing a lot of the current implementations in other tests are as follows: 
```
						acctest.ExpectNoDelete(),
					},
				},
```
Should this be suggested or mentioned as an option for update-steps that update more than one resource at once?

Also is there plans to update most other existing tests to use this check, or will we just be recommending it going forward?

trodge:
> I'm seeing a lot of the current implementations in other tests are as follows:
> 
> ```
> ConfigPlanChecks: resource.ConfigPlanChecks{
> 					PreApply: []plancheck.PlanCheck{
> 						acctest.ExpectNoDelete(),
> 					},
> 				},
> ```
> 
> Should this be suggested or mentioned as an option for update-steps that update more than one resource at once?
> 
> Also is there plans to update most other existing tests to use this check, or will we just be recommending it going forward?

The plan is to eventually update existing tests to use `ExpectResourceAction`, and this is preferred over `ExpectNoDelete`

PR #11752, opened 2024-09-19
Add deletion_protection field to Memcache Instance

SamanthaMathews:
Part of b/368232860

Part of https://github.com/hashicorp/terraform-provider-google/issues/18854

```

Memcache: added `deletion_protection` field to `memcache_instance` to make deleting them require an explicit intent. `memcache_instance` resources now cannot be destroyed unless `deletion_protection = false` is set for the resource.
```

### DIFF
--- a/mmv1/products/memcache/Instance.yaml
+++ b/mmv1/products/memcache/Instance.yaml
@@ -43,7 +43,8 @@ async:
   error:
     path: 'error'
     message: 'message'
-custom_code:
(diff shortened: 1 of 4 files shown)


c2thorn:

SamanthaMathews:

rileykarson (commented):
We probably want to add these as part of a major version. @c2thorn I'm gonna hand this and another PR off to you.

PR #11778, opened 2024-09-20
Fix securityprofiles link in doc

rileykarson:
Not sure why we reference projects as it doesn't seem to exist.

```release-note:none
```


zli82016 (approved):
LGTM

PR #11788, opened 2024-09-23
rewrite - private override changes

c2thorn:
adds `--override` flag to the go engine for private overrides

needs to be combined with the changes in the private overrides internal repo to test, but produces only minor diffs.

```release-note:none

```

### DIFF
--- a/mmv1/api/product.go
+++ b/mmv1/api/product.go
@@ -15,6 +15,7 @@ package api
 
 import (
 	"log"
+	"reflect"
 	"strings"
 	"unicode"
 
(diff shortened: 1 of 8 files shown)


mmv1/api/product.go
```diff
@@ -231,3 +232,93 @@ func (p *Product) TerraformName() string {
 func (p Product) Lineage() string {
 	return p.Name
 }
+
+func Merge(self, otherObj reflect.Value) {
```
c2thorn, line 236: See the equivalent ruby merge functions here: https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/api/object.rb#L32

zli82016 (approved):
LGTM

PR #11791, opened 2024-09-23
Dataproc: Make `internal_ip_only` a computed, not defaulted, field. Always send it when specified.

Harwayne:
Make `internal_ip_only` of `google_dataproc_cluster` a computed, not defaulted, field. Always send it when specified by the user.

The 'default' value for this field is determined on Dataproc's servers, based on the image version used. Users can still provide a value to override the default, but if the user does not provide the value, then the value generated by Dataproc's servers should be used.

This is being done because the server-side default changed between Dataproc image versions < 2.1 and >= 2.2, from `true` to `false` respectively. 

Fixes https://github.com/hashicorp/terraform-provider-google/issues/19522.

```release-note:bug
dataproc: Changed `google_dataproc_cluster` `internal_ip_only` field to computed and always sent if specified.
```

### DIFF
--- a/mmv1/third_party/terraform/acctest/bootstrap_test_utils.go
+++ b/mmv1/third_party/terraform/acctest/bootstrap_test_utils.go
@@ -915,7 +915,7 @@ func BootstrapSubnetForDataprocBatches(t *testing.T, subnetName string, networkN
 	subnetOptions := map[string]interface{}{
 		"privateIpGoogleAccess": true,
 	}
-	return BootstrapSubnetWithOverrides(t, subnetName, networkName, subnetOptions)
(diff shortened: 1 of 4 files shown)


slevenick:
Just approved tests, but I think this will fail all the existing tests as they don't specify internal_ip_only = false

slevenick (changes requested):
The tests continue to fail. I think we need to go back and specify internal_ip_only = false on almost all the old tests, as they depend on that behavior.

Alternatively, I have an open PR that fixes this issue by keeping the existing default of false and always sending the value: https://github.com/GoogleCloudPlatform/magic-modules/pull/11923

slevenick (commented):
Dismissing review for now, seems like the clusters are not successfully coming up in many cases

slevenick (changes requested):
Looks like tests continue to fail.

PR #11792, opened 2024-09-23
add: flatten / expand biglake_configuration in bigquery_table

nevzheng:
The prior PR #11724, added the configuration to "table" but omitted the flatten and expand methods. This PR adds them.

```release-note:enhancement
bigquery: add: flatten / expand biglake_configuration in bigquery_table
```


slevenick (changes requested):
Test failing with: Error 400: Access denied to validate storage location:

nevzheng:
Fixed: "Error 400: Access denied to validate storage location"
+ Fix: Use `objectAdmin` instead of `objectViewer`

slevenick:
> Fixed: "Error 400: Access denied to validate storage location"
> 
> * Fix: Use `objectAdmin` instead of `objectViewer`
> * Fix: Add 'data/' to bucket to ensure path is valid

Why did changing the resource code make the test fail? It didn't look like there was a change to the test configuration prior to the failure. Does that mean we could break existing configurations by adding this code?

nevzheng:
the prior CL did not add flatten / expand so the requests never made it to the server to be run. Once we added flatten/expand, I saw the permission issue.

We also need the objectAdmin permission, which is only applied to the service account and connection created in this test case. This requirement is specific to Big Lake Managed Tables. (we need the object create permission). The path also needs to exist (adding "data/").

mmv1/third_party/terraform/services/bigquery/resource_bigquery_table_test.go
```diff
@@ -304,6 +304,11 @@ func TestAccBigQueryBigLakeManagedTable(t *testing.T) {
 			{
 				Config: testAccBigLakeManagedTable(bucketName, connectionID, datasetID, tableID, TEST_SIMPLE_CSV_SCHEMA),
 			},
+			{
+				ResourceName:      "google_bigquery_table.test",
+				ImportState:       true,
+				ImportStateVerify: true,
```
slevenick, line 310: Looks like we need an ignore here on deletion_protection

slevenick (changes requested):
Ah, makes sense. Just need a fix on the test and we should be good

mmv1/third_party/terraform/services/bigquery/resource_bigquery_table_test.go
nevzheng: done.

slevenick (approved):

slevenick:
I removed the release note because this just fixes a bug in an unreleased version of the provider

PR #11799, opened 2024-09-24
rewrite - more convert script changes

c2thorn:

```release-note:none

```

### DIFF
--- a/mmv1/provider/file_template.rb
+++ b/mmv1/provider/file_template.rb
@@ -36,7 +36,6 @@ module Provider
     # Once the file's contents are written, set the proper [chmod] mode and
     # format the file with a language-appropriate formatter.
     def generate(pwd, template, path, provider)
-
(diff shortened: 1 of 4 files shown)


mmv1/provider/file_template.rb
```diff
@@ -36,12 +36,6 @@ class FileTemplate
     # Once the file's contents are written, set the proper [chmod] mode and
     # format the file with a language-appropriate formatter.
     def generate(pwd, template, path, provider)
-      # If we've modified a file since starting an MM run, it's a reasonable
-      # assumption that it was this run that modified it.
-      if File.exist?(path) && File.mtime(path) > @env[:start_time]
```
c2thorn, line 41: Gets raised when trying to convert a `product.yaml` file with `product.yaml.temp`
I think we are close enough to the migration to turn this off

NickElliot (approved):

PR #11809, opened 2024-09-25
go rewrite - final preparations before migration

c2thorn:
refresh + various tweaks to get in before we fully migrate

changes separated by commit

```release-note:none

```


mmv1/template-converter.go
```diff
@@ -172,12 +172,7 @@ func convertHandwrittenFiles(folder, handwrittenTempFiles string) int {
 			log.Fatalf("Cannot open the file: %v", file)
 		}
 		data = replace(data)
-		goTemplate := ""
-		if strings.Contains(string(data), "{{") {
-			goTemplate = strings.Replace(file, ".erb", ".tmpl", 1)
-		} else {
-			goTemplate = strings.Replace(file, ".erb", "", 1)
-		}
+		goTemplate := strings.Replace(file, ".erb", ".tmpl", 1)
```
c2thorn, line 175: making all files in templates/ a .tmpl file whether they are .erb or not. This simplifies the migration steps. We can always rename the .go.tmpl files back to a .go file later

mmv1/main.go
```diff
@@ -158,20 +158,27 @@ func main() {
...
+	close(productFileChannel)
+	close(productsForVersionChannel)
+
+	var productsForVersion []*api.Product
+
+	for p := range productsForVersionChannel {
+		productsForVersion = append(productsForVersion, p)
+	}
 
 	slices.SortFunc(productsForVersion, func(p1, p2 *api.Product) int {
```
c2thorn, line 183: When testing other changes, I had one occurrence of the Go engine crashing here with a nil pointer reference to `productsForVersion`. I immediately retried and it did not occur again. I believe this is due to hitting a rare race condition with the goroutines.

Using a channel instead of passing the array pointer directly to the goroutines should solve for this, but it is something to keep an eye out for.

mmv1/provider/terraform.go
```diff
@@ -328,7 +328,11 @@ func (t Terraform) getCommonCopyFiles(versionName string, generateCode, generate
 func (t Terraform) getCopyFilesInFolder(folderPath, targetDir string) map[string]string {
 	m := make(map[string]string, 0)
 	filepath.WalkDir(folderPath, func(path string, di fs.DirEntry, err error) error {
-		if !di.IsDir() && !strings.HasSuffix(di.Name(), ".tmpl") && !strings.HasSuffix(di.Name(), ".erb") {
+		if !di.IsDir() && !strings.HasSuffix(di.Name(), ".tmpl") && !strings.HasSuffix(di.Name(), ".erb") { // Exception files
+			if di.Name() == "gha-branch-renaming.png" || di.Name() == "clock-timings-of-branch-making-and-usage.png" {
```
c2thorn, line 332: We had new files pushed here yesterday: https://github.com/GoogleCloudPlatform/magic-modules/tree/main/mmv1/third_party/terraform/.teamcity/docs/images

The go engine copies these files over in the GA provider, where the Ruby engine does not.

We don't have logic that explicitly recreates https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/provider/terraform/common~copy.yaml#L179 so I'm opting to just start an exception list

mmv1/provider/terraform.go
```diff
@@ -365,7 +365,14 @@ func (t Terraform) CopyFileList(outputFolder string, files map[string]string, ge
 			log.Fatalf("Cannot read source file %s while copying: %s", source, err)
 		}
 
-		err = os.WriteFile(targetFile, sourceByte, 0644)
+		var permission fs.FileMode
+		if strings.HasSuffix(targetDir, "scripts") {
+			permission = 0755
```
c2thorn, line 370: This resolves a diff for the script files that Github picks up on. May as well remove that diff.

mmv1/main.go
zli82016: Good catch.

mmv1/provider/terraform.go
zli82016: Good catch.

zli82016 (approved):
LGTM

mmv1/main.go
NickElliot: Do we have a plan for if other like, not easily reproducible build engine errors occur within the following weeks-months?

NickElliot (approved):
LGTM

mmv1/provider/terraform.go
NickElliot: we should consider a to-do to have a more organized file list here rather than a line of OR switches

mmv1/main.go
c2thorn: If we have more build engine errors relating to this block, we can remove the parallelization and let it run serially until we can fix the issue.

mmv1/provider/terraform.go
zli82016: Yeah, I am going to do that after the switchover.

PR #11811, opened 2024-09-25
Fix Service Account creation by ignoring 403 errors on read polling

sbocinec:
Fix issues with `google_service_account` resource creation due to eventual consistency of GCP IAM API. Fixes https://github.com/hashicorp/terraform-provider-google/issues/19624 and potentially other bugs.

```release-note:bug
iam: addressed `google_service_account` creation issues caused by the eventual consistency of the GCP IAM API by ignoring 403 errors returned on polling the service account after creation.
```

### DIFF
--- a/mmv1/third_party/terraform/services/resourcemanager/resource_google_service_account.go
+++ b/mmv1/third_party/terraform/services/resourcemanager/resource_google_service_account.go
@@ -152,7 +152,8 @@ func resourceGoogleServiceAccountCreate(d *schema.ResourceData, meta interface{}
 
 	// We poll until the resource is found due to eventual consistency issue
 	// on part of the api https://cloud.google.com/iam/docs/overview#consistency
-	err = transport_tpg.PollingWaitTime(resourceServiceAccountPollRead(d, meta), transport_tpg.PollCheckForExistence, "Creating Service Account", d.Timeout(schema.TimeoutCreate), 1)
+	// IAM API returns 403 when the queried SA is not found, so we must ignore both 404 & 403 errors
+	err = transport_tpg.PollingWaitTime(resourceServiceAccountPollRead(d, meta), transport_tpg.PollCheckForExistenceWith403, "Creating Service Account", d.Timeout(schema.TimeoutCreate), 1)
 
 	if err != nil {
 		return err


mmv1/third_party/terraform/services/resourcemanager/resource_google_service_account.go
```diff
@@ -152,7 +152,8 @@ func resourceGoogleServiceAccountCreate(d *schema.ResourceData, meta interface{}
 
 	// We poll until the resource is found due to eventual consistency issue
 	// on part of the api https://cloud.google.com/iam/docs/overview#consistency
-	err = transport_tpg.PollingWaitTime(resourceServiceAccountPollRead(d, meta), transport_tpg.PollCheckForExistence, "Creating Service Account", d.Timeout(schema.TimeoutCreate), 1)
+	// IAM API returns 403 when the queried SA is not found, so we must ignore both 404 & 403 errors
+	err = transport_tpg.PollingWaitTime(resourceServiceAccountPollRead(d, meta), transport_tpg.PollCheckForExistenceWith403, "Creating Service Account", d.Timeout(schema.TimeoutCreate), 1)
```
c2thorn, line 156: We already have eventual consistency checks for 403's in https://github.com/GoogleCloudPlatform/magic-modules/pull/11811/files#diff-0541a83ba5cbaa8bd8a0cd9128218cc5766b18aeaa97015ccaa06b301511e08cR145

Wouldn't this just add even more polling for 403's? 

In your testing, is the resource failing at this specific line? If so, what if we just moved the existing sleep already present in  https://github.com/GoogleCloudPlatform/magic-modules/pull/11811/files#diff-0541a83ba5cbaa8bd8a0cd9128218cc5766b18aeaa97015ccaa06b301511e08cR165 to before this line?

c2thorn (changes requested):

mmv1/third_party/terraform/services/resourcemanager/resource_google_service_account.go
sbocinec: >We already have eventual consistency checks for 403'
Yes, the addition was actually introduced by me. The problem how I understand it is, it only handles the first Read after initial creation of the account. When the initial Create + Read succeeds, the resource is added to the TF state.

The poller then polls to check if the account indeed exists to be sure the SA has been eventually created. But here is the problem - as the IAM API is eventually consistent, it returns 403 when the SA is not found (yes, it's not a mistake, not 404, but 403 is returned here, it's a bit misleading as it masks not found behind IAM_PERMISSION_DENIED for seurity reasons). As the poller immediatelly exits on any other error than 404, all the other defensive logic that attempts to deal with the EC is ineffective as the provider immediately fails. 

Also, the 10s sleep that was also meant to deal with that is actually never reached https://github.com/GoogleCloudPlatform/magic-modules/pull/11811/files#diff-0541a83ba5cbaa8bd8a0cd9128218cc5766b18aeaa97015ccaa06b301511e08cR162-R165

You can see it in the log output in the issue I 've submitted https://github.com/hashicorp/terraform-provider-google/issues/19624 - notice it fails within 1s  as the poll read returns 403 and the whole provider fails.
image

I personally think, the 10s sleep could even be removed iif the poller would handle 403 as I'm proposing.

c2thorn: Okay, thank you for providing the logs. I think this makes sense, and looks like the previous sleep did not actually address this issue.

Considering we are just swapping out the PollCheck used, I think this should be good to merge.

c2thorn (approved):

c2thorn:
running build first to confirm no issues

c2thorn (approved):

PR #11820, opened 2024-09-25
Support prodcuer port in network endpoint group

ericlin0101:

```release-note:REPLACEME

```

### DIFF
--- a/mmv1/products/compute/RegionNetworkEndpointGroup.yaml
+++ b/mmv1/products/compute/RegionNetworkEndpointGroup.yaml
@@ -168,6 +168,20 @@ properties:
     custom_expand: 'templates/terraform/custom_expand/resourceref_with_validation.go.tmpl'
     resource: 'Subnetwork'
     imports: 'selfLink'
+  - !ruby/object:Api::Type::NestedObject
+    name: 'pscData'
+    description: |
+      This field is only used for PSC NEGs.
+
+    properties:
+      - !ruby/object:Api::Type::String
+        name: 'producerPort'
+        ignore_read: true
+        description: |
+          The PSC producer port to use when consumer PSC NEG connects to a producer. If
+          this flag isn't specified for a PSC NEG with endpoint type
+          private-service-connect, then PSC NEG will be connected to a first port in the
+          available PSC producer port range.
   - name: 'cloudRun'
     type: NestedObject
     description: |
--- a/mmv1/templates/terraform/examples/region_network_endpoint_group_psc_service_attachment.tf.tmpl
+++ b/mmv1/templates/terraform/examples/region_network_endpoint_group_psc_service_attachment.tf.tmpl
@@ -39,7 +39,7 @@ resource "google_compute_forwarding_rule" "default" {
 
   load_balancing_scheme = "INTERNAL"
   backend_service       = google_compute_region_backend_service.default.id
-  all_ports             = true
+  ports                 = ["80", "88", "443"]
   network               = google_compute_network.default.name
   subnetwork            = google_compute_subnetwork.default.name
 }
@@ -61,7 +61,9 @@ resource "google_compute_region_network_endpoint_group" "{{$.PrimaryResourceId}}
 
   network_endpoint_type = "PRIVATE_SERVICE_CONNECT"
   psc_target_service    = google_compute_service_attachment.default.self_link
-
+  psc_data {
+    producer_port = "88"
+  }
   network               = google_compute_network.default.self_link
   subnetwork            = google_compute_subnetwork.default.self_link
 }


mmv1/products/compute/RegionNetworkEndpointGroup.yaml
```diff
@@ -168,6 +168,20 @@ properties:
     custom_expand: 'templates/terraform/custom_expand/resourceref_with_validation.go.tmpl'
     resource: 'Subnetwork'
     imports: 'selfLink'
+  - !ruby/object:Api::Type::NestedObject
+    name: 'pscData'
```
SarahFrench, line 172: Just in the last day we have changed our code generator in a way that impacts the contents of YAML files. To meet the new  code generator's expectations you should only need to remove these Ruby-specific annotations from your PR and instead put the type of a field as a property like below.

Currently the acceptance tests on this PR are failing as a consequence of this (looks like any fields missing the type property are generated as String fields by default.

```suggestion
   - name: 'pscData'
    type: String
    type: NestedObject
```

mmv1/products/compute/RegionNetworkEndpointGroup.yaml
```diff
@@ -168,6 +168,20 @@ properties:
...
+    description: |
+      This field is only used for PSC NEGs.
+
+    properties:
+      - !ruby/object:Api::Type::String
+        name: 'producerPort'
+        ignore_read: true
```
SarahFrench, line 179:
```suggestion
      - name: 'producerPort'
        type: String
        ignore_read: true
```

SarahFrench (commented):
👋 Hi, thanks for your PR!

We just changed our code generator ([details here](https://googlecloudplatform.github.io/magic-modules/reference/ruby-go-changes/)) and currently this PR is doing some things the old way, impacting your changes being generated correctly.

I've left some suggested changes about how to fix them, and once that's sorted I'll do a wider review

mmv1/products/compute/RegionNetworkEndpointGroup.yaml
SarahFrench: Note: the indenting might be incorrect here, please don't just accept the suggestion!

ericlin0101: Thanks, the branch was created days ago so did not catch the new code generator.

SarahFrench: No worries, just unlucky timing!

mmv1/products/compute/RegionNetworkEndpointGroup.yaml
```diff
@@ -168,6 +168,20 @@ properties:
...
     imports: 'selfLink'
+  - name: 'pscData'
+    type: NestedObject
+    description: |
+      This field is only used for PSC NEGs.
+
+    properties:
+      - name: 'producerPort'
+        type: String
+        ignore_read: true
```
SarahFrench, line 179: Given what I said in my other review comment, and from looking at [the logs for the acceptance test here](https://github.com/GoogleCloudPlatform/magic-modules/pull/11820#issuecomment-2377316553), I assume that currently the API can accept and use the producerPort value but does not return that value in any responses. If that's the case then ignore_read makes sense here, but I don't have much info to go on when reviewing.

If you could confirm the above I'm happy to LGTM the PR.

SarahFrench (commented):
Hi @ericlin0101 - Thanks for your PR & bearing with the code generator changes!

I took a look at the [API docs](https://cloud.google.com/compute/docs/reference/rest/v1/regionNetworkEndpointGroups#NetworkEndpointGroup.FIELDS.psc_data) for this resource and I can't see the producerPort field present in the public documentation, instead there's only:

```
  "pscData": {
    "consumerPscAddress": string,
    "pscConnectionId": string,
    "pscConnectionStatus": enum
  }
```

From your profile I see Google is listed as your employer, so I assume you are a Googler and this is a field that's being added to the API soon, and the public docs are lagging a bit.

Could you please confirm the above? I'm a HashiCorp employee so we have a bit of an asymmetry about what information is available to each us. In these situations I can only sanity check and LGTM PRs so I want to be explicit about that!

mmv1/products/compute/RegionNetworkEndpointGroup.yaml
ericlin0101: @SarahFrench, you are correct—the API can accept the new field, but it cannot return the value yet. We are currently working on this, and it may take a few weeks to roll out.

Additionally, the public documentation is slightly behind. It will be updated once everything is fully implemented.

SarahFrench: In that case, do you want to go ahead with this PR being merged? Or to wait until the value is also returned?

SarahFrench (approved):
LGTM, waiting for confirmation to merge

mmv1/products/compute/RegionNetworkEndpointGroup.yaml
ericlin0101: Thanks for checking, let us wait until it is been confirmed and tested before merge.

ericlin0101:
@SarahFrench, I believe this is ready to merge. Please proceed with the merge at your convenience. Thank you!

SarahFrench:
Just realised I hadn't approved the build - I'll let that complete before merging

SarahFrench:
That test is failing on multiple PRs and isn't specific to this PT. Reported that here: https://github.com/hashicorp/terraform-provider-google/issues/17838#issuecomment-2494512276

Approving & merging!

ericlin0101:
@SarahFrench, Thank you. We are all set.

RomanPavelkoMGM:
After switching to 6.13.0 I started to get constant recreation of my NEGs. I tried to recreate them once with 6.13.0 version but it keeps showing the "recreate" action on new plans. There was no code changes on my side and I also didn't add psc_data field since it's optional and I don't need it. I had to stick to 6.12.0 version to get this issue disappear. Is this a bug of the optional psc_data field?

image

SarahFrench:
Hi @RomanPavelkoMGM - comments on closed PRs will not go through our triage process. Please open an issue referencing this PR if you're having a problem. Thanks!

PR #11805, opened 2024-09-24
Add GCS ingestion settings and platform log settings to `google_pubsub_topic`

andreigurau:
Add [GCS ingestion settings](https://github.com/googleapis/googleapis/blob/master/google/pubsub/v1/pubsub.proto#L241-L324) and [Platform Log Settings](https://github.com/googleapis/googleapis/blob/master/google/pubsub/v1/pubsub.proto#L335-L366) to `google_pubsub_topic`

**Release Note Template for Downstream PRs (will be copied)**

```release-note:new-resource
pubsub: added GCS ingestion settings and platform log settings to `google_pubsub_topic` resource
```

PR #11823, opened 2024-09-25
Add GCS ingestion settings and platform log settings to `google_pubsub_topic`

andreigurau:
Add [GCS ingestion settings](https://github.com/googleapis/googleapis/blob/master/google/pubsub/v1/pubsub.proto#L241-L324) and [Platform Log Settings](https://github.com/googleapis/googleapis/blob/master/google/pubsub/v1/pubsub.proto#L335-L366) to `google_pubsub_topic`

```release-note:new-resource
pubsub: added GCS ingestion settings and platform log settings to `google_pubsub_topic` resource
```

### DIFF
--- a/mmv1/products/pubsub/Topic.yaml
+++ b/mmv1/products/pubsub/Topic.yaml
@@ -83,6 +83,10 @@ examples:
     primary_resource_id: 'example'
     vars:
       topic_name: 'example-topic'
+  - name: 'pubsub_topic_ingestion_cloud_storage'
+    primary_resource_id: 'example'
+    vars:
+      topic_name: 'example-topic'
 parameters:
 properties:
   - name: 'name'
@@ -166,6 +170,9 @@ properties:
         type: NestedObject
         description: |
           Settings for ingestion from Amazon Kinesis Data Streams.
+        exactly_one_of:
+          - 'aws_kinesis'
+          - 'cloud_storage'
         properties:
           - name: 'streamArn'
             type: String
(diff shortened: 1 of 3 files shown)


mmv1/products/pubsub/Topic.yaml
```diff
@@ -194,3 +201,100 @@ properties:
...
+              - 'pubsub_avro_format'
+            properties:
+              - name: 'delimiter'
+                type: String
+                description: |
+                  The delimiter to use when using the 'text' format. Each line of text as
+                  specified by the delimiter will be set to the 'data' field of a Pub/Sub
+                  message. When unset, '\n' is used.
+                required: false
+                default_value: "\n"
```
c2thorn, line 236:
```suggestion
                default_value: "\\n"
```

c2thorn:
This is remake of a previous PR

Posting context that this PR is waiting for end of week:
https://github.com/GoogleCloudPlatform/magic-modules/pull/11805#issuecomment-2374161099

andreigurau:
Our rollout looks to be finished - could we try rerunning the VCR tests? Not sure how to do that without adding another commit

mmv1/products/pubsub/Topic.yaml
```diff
@@ -166,6 +170,9 @@ properties:
         type: NestedObject
         description: |
           Settings for ingestion from Amazon Kinesis Data Streams.
+        exactly_one_of:
```
trodge, line 173: This is potentially a breaking change if `aws_kinesis` was previously optional. Could we use `conflicts` instead? That will make the behavior "at most one of" instead of "exactly one of".

trodge (changes requested):

mmv1/products/pubsub/Topic.yaml
andreigurau: Changed it to `conflicts`

trodge (approved):

PR #11826, opened 2024-09-26
container: refactor `node_config` node pool update

wyardley:
This sets the stage for consolidating the logic for updating nodepools between the `container_cluster` and `container_node_pool` resources, by creating a new `nodePoolNodeConfigUpdate()` function.

Part of https://github.com/hashicorp/terraform-provider-google/issues/19225

```release-note:none

```


mmv1/third_party/terraform/services/container/node_config.go.tmpl
```diff
@@ -1984,3 +1987,530 @@ func flattenHostMaintenancePolicy(c *container.HostMaintenancePolicy) []map[stri
...
+				if err != nil {
+					return err
+				}
+
+				// Wait until it's updated
+				return ContainerOperationWait(config, op,
+					nodePoolInfo.project,
+					nodePoolInfo.location,
+					"updating GKE node pool disk_size_gb/disk_type/machine_type/storage_pools", userAgent,
+					timeout)
```
wyardley, line 2079: A lot of opportunity to DRY up these parts with a function at some point too, but trying to keep it pretty self-contained here for now.

NickElliot (approved):
sorry for the delays! Didn't realize the scope of this specific change in the larger issue was deceptively small and kept pushing back the review to put more time into it.

LGTM!

wyardley:
Hah, yes, the actual change is one line plus moving a big block of code. It should set the stage for sharing the code with the default node pool update. Can check out the version before the second commit if you want to see which you prefer.

PR #11829, opened 2024-09-26
Convert database resource from automated to handwritten resource

jharshika:
This pull request addresses issue [hashicorp/terraform-provider-google#7810](https://github.com/hashicorp/terraform-provider-google/issues/7810) by migrating the database resource from an autogenerated model to a handwritten implementation. This change allows us to enhance error handling and implement more flexible execution conditions.

Migration to Handwritten Implementation: The database resource, previously autogenerated, has been refactored to a handwritten version. This change enables us to:
- Implement custom logic to handle error messages effectively.
- Introduce conditions during execution that were not feasible with the autogenerated version.

```release-note:none
sql: google_sql_database_instance can break refresh when activation_policy is set to NEVER
```


SarahFrench (changes requested):
Hi @jharshika - Thanks for your PR!

To be upfront, I don't think we'll accept this resource no longer being generated. If there are code changes needed in the resource we should use the code generator's features that allow custom code to be inserted into generated code. If that isn't sufficient we'd then think about next steps, potentially adding more flexibility to the code generator.

Could you please push some commits with the code changes you'd like to see added to the resource to address the issue you linked? Currently I can see (from the [diff links in this message](https://github.com/GoogleCloudPlatform/magic-modules/pull/11829#issuecomment-2376990049)) that the contents of the new services/sql/resource_sql_database.go file in this PR doesn't differ from the existing file in the google/google-beta repos.

Once I can see what code change you want to make I can suggest how to achieve them via the generator. For now, here is the page in our contribution guide about adding custom code: https://googlecloudplatform.github.io/magic-modules/develop/custom-code/

jharshika:
Hi @SarahFrench thanks for the explanation. Closing this PR and creating a new one with suggested changes

PR #11832, opened 2024-09-26
container: fix storage pool resource names in test

wyardley:
Came across this when working on #11826

Tests such as
`TestAccContainerNodePool_withDiskMachineAndStoragePoolUpdate` seem to be failing, possibly due to being promoted from beta to v1

Rather than update the regex in extractSP, use
`tpgresource.GetRelativePath()` and return that from the bootstrap method, as mentioned here:
https://github.com/GoogleCloudPlatform/magic-modules/pull/11391#discussion_r1750588141

See #11391 and #11598

```release-note:none

```

### DIFF
--- a/mmv1/third_party/terraform/services/container/resource_container_cluster_test.go.tmpl
+++ b/mmv1/third_party/terraform/services/container/resource_container_cluster_test.go.tmpl
@@ -11697,11 +11697,7 @@ func TestAccContainerCluster_storagePoolsWithNodePool(t *testing.T) {
 	subnetworkName := acctest.BootstrapSubnet(t, "gke-cluster", networkName)
 	location := envvar.GetTestZoneFromEnv()
 
-	storagePoolNameURL := acctest.BootstrapComputeStoragePool(t, "basic-1", "hyperdisk-balanced")
-	storagePoolResourceName, err := extractSPName(storagePoolNameURL)
-	if err != nil {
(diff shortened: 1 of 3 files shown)


mmv1/third_party/terraform/services/container/resource_container_cluster_test.go.tmpl
```diff
@@ -11800,17 +11792,6 @@ resource "google_container_cluster" "storage_pools_with_node_config" {
 `, cluster, location, networkName, subnetworkName, storagePoolResourceName)
 }
 
-func extractSPName(url string) (string, error) {
-    re := regexp.MustCompile(`https://www\.googleapis\.com/compute/beta/(projects/[^"]+)`)
```
wyardley, line 11804: for sure, updating this to
`/(beta|v1)` worked, but reading through the past issues / comments, thinking something like what's in here is closer to what you'd want?

wyardley:
cc: @shuyama1

mmv1/third_party/terraform/services/container/resource_container_cluster_test.go.tmpl
amacaskill: Yea I think the tpgresource.GetRelativePath() is a good idea

shuyama1 (approved):
Thank you!

PR #11833, opened 2024-09-26
Default output to true for Fingerprint

zli82016:

```release-note:none

```

### DIFF
--- a/mmv1/api/type.go
+++ b/mmv1/api/type.go
@@ -335,6 +335,11 @@ func (t *Type) SetDefault(r *Resource) {
 		if t.Description == "" {
 			t.Description = fmt.Sprintf("A reference to %s resource", t.Resource)
 		}
+	case t.IsA("Fingerprint"):
+		// Represents a fingerprint.  A fingerprint is an output-only
+		// field used for optimistic locking during updates.
+		// They are fetched from the GCP response.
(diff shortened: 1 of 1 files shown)


mmv1/api/type.go
```diff
@@ -335,6 +335,11 @@ func (t *Type) SetDefault(r *Resource) {
 		if t.Description == "" {
 			t.Description = fmt.Sprintf("A reference to %s resource", t.Resource)
 		}
+	case t.IsA("Fingerprint"):
+		// Represents a fingerprint.  A fingerprint is an output-only
+		// field used for optimistic locking during updates.
+		// They are fetched from the GCP response.
+		t.Output = true
```
zli82016, line 342: Should output be set true only if t.Output is nil ?

c2thorn: Probably. I can't imagine why someone would explicitly set Output: false, but I guess it could happen

c2thorn (commented):
~~Does this produce a diff?~~

Nevermind, I forgot I can see it now :)

c2thorn (approved):
LGTM, I'll leave the nil check decision to you

zli82016:
I will not check nil now. If there is an issue, we will add it later.

PR #11834, opened 2024-09-26
pubsub: fix permadiff with configuring an empty retry_policy.

anavada:
Fixes https://github.com/hashicorp/terraform-provider-google/issues/18016

The permadiff arises when users specify an empty retry policy as `retry_policy: {}` in their config, but the empty value doesn't get sent to the API and the API subsequently doesn't return a `retry_policy` in the response. By setting `send_empty_value: true` on the `retry_policy` property, we ensure that the API recognizes the user-configured value.

Similar fixes for other resource properties will be investigated and sent in separate PRs.

```release-note:bug
pubsub: fix permadiff with configuring an empty `retry_policy`.
```

### DIFF
--- a/mmv1/products/pubsub/Subscription.yaml
+++ b/mmv1/products/pubsub/Subscription.yaml
@@ -471,6 +471,8 @@ properties:
           A duration in seconds with up to nine fractional digits, terminated by 's'. Example: "3.5s".
         default_from_api: true
         diff_suppress_func: 'tpgresource.DurationDiffSuppress'
+    send_empty_value: true
+    allow_empty_object: true
   - name: 'enableMessageOrdering'
     type: Boolean
     description: |
--- a/mmv1/third_party/terraform/services/pubsub/resource_pubsub_subscription_test.go
+++ b/mmv1/third_party/terraform/services/pubsub/resource_pubsub_subscription_test.go
@@ -36,6 +36,30 @@ func TestAccPubsubSubscription_emptyTTL(t *testing.T) {
 	})
 }
 
+func TestAccPubsubSubscription_emptyRetryPolicy(t *testing.T) {
+	t.Parallel()
+
+	topic := fmt.Sprintf("tf-test-topic-%s", acctest.RandString(t, 10))
+	subscription := fmt.Sprintf("tf-test-sub-%s", acctest.RandString(t, 10))
+
+	acctest.VcrTest(t, resource.TestCase{
+		PreCheck:                 func() { acctest.AccTestPreCheck(t) },
+		ProtoV5ProviderFactories: acctest.ProtoV5ProviderFactories(t),
+		CheckDestroy:             testAccCheckPubsubSubscriptionDestroyProducer(t),
+		Steps: []resource.TestStep{
+			{
+				Config: testAccPubsubSubscription_emptyRetryPolicy(topic, subscription),
+			},
+			{
+				ResourceName:      "google_pubsub_subscription.foo",
+				ImportStateId:     subscription,
+				ImportState:       true,
+				ImportStateVerify: true,
+			},
+		},
+	})
+}
+
 func TestAccPubsubSubscription_basic(t *testing.T) {
 	t.Parallel()
 
@@ -490,6 +514,22 @@ resource "google_pubsub_subscription" "foo" {
     ttl = ""
   }
   enable_message_ordering    = false
+}
+`, topic, subscription)
+}
+
+func testAccPubsubSubscription_emptyRetryPolicy(topic, subscription string) string {
+	return fmt.Sprintf(`
+resource "google_pubsub_topic" "foo" {
+  name = "%s"
+}
+
+resource "google_pubsub_subscription" "foo" {
+  name  = "%s"
+  topic = google_pubsub_topic.foo.id
+
+  retry_policy {
+  }
 }
 `, topic, subscription)
 }
--- a/mmv1/third_party/tgc/tests/data/example_pubsub_subscription.json
+++ b/mmv1/third_party/tgc/tests/data/example_pubsub_subscription.json
@@ -48,6 +48,7 @@
         "pushConfig": {
           "pushEndpoint": "https://example.com/push"
         },
+        "retryPolicy": null,
         "topic": "projects/{{.Provider.project}}/topics/example-pubsub-topic"
       }
     }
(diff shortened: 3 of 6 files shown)


mmv1/products/pubsub/Subscription.yaml
```diff
@@ -471,6 +471,7 @@ properties:
           A duration in seconds with up to nine fractional digits, terminated by 's'. Example: "3.5s".
         default_from_api: true
         diff_suppress_func: 'tpgresource.DurationDiffSuppress'
+    send_empty_value: true
```
melinath, line 474: I think you'll want allow_empty_object here as well, though it looks like the tests are passing, so maybe it's fine? Did you happen to try that out?

```suggestion
    send_empty_value: true
    allow_empty_object: true
```

anavada: Yeah the earlier BQ-related test failures seem unrelated and are passing now. Thanks for opening https://github.com/hashicorp/terraform-provider-google/issues/19683.

Those tests pass locally as well:
```
$ make testacc TEST=./google-beta/services/pubsub TESTARGS='-run=TestAccPubsubSubscription_pubsubSubscriptionPushBqExample'
go vet
TF_ACC=1 TF_SCHEMA_PANIC_ON_ERROR=1 go test ./google-beta/services/pubsub -v -run=TestAccPubsubSubscription_pubsubSubscriptionPushBqExample -timeout 240m -ldflags="-X=github.com/hashicorp/terraform-provider-google-beta/version.ProviderVersion=acc"
=== RUN   TestAccPubsubSubscription_pubsubSubscriptionPushBqExample
=== PAUSE TestAccPubsubSubscription_pubsubSubscriptionPushBqExample
=== CONT  TestAccPubsubSubscription_pubsubSubscriptionPushBqExample
--- PASS: TestAccPubsubSubscription_pubsubSubscriptionPushBqExample (30.70s)
PASS
ok      github.com/hashicorp/terraform-provider-google-beta/google-beta/services/pubsub 31.317s

$ make testacc TEST=./google-beta/services/pubsub TESTARGS='-run=TestAccPubsubSubscription_pubsubSubscriptionPushBqTableSchemaExample'
go vet
TF_ACC=1 TF_SCHEMA_PANIC_ON_ERROR=1 go test ./google-beta/services/pubsub -v -run=TestAccPubsubSubscription_pubsubSubscriptionPushBqTableSchemaExample -timeout 240m -ldflags="-X=github.com/hashicorp/terraform-provider-google-beta/version.ProviderVersion=acc"
=== RUN   TestAccPubsubSubscription_pubsubSubscriptionPushBqTableSchemaExample
=== PAUSE TestAccPubsubSubscription_pubsubSubscriptionPushBqTableSchemaExample
=== CONT  TestAccPubsubSubscription_pubsubSubscriptionPushBqTableSchemaExample
--- PASS: TestAccPubsubSubscription_pubsubSubscriptionPushBqTableSchemaExample (30.34s)
PASS
ok      github.com/hashicorp/terraform-provider-google-beta/google-beta/services/pubsub 30.958s
```

Similarly, the "VCR-test" was flaky as well but is passing now. However, the `terraform-google-conversion-build-and-unit-tests` failures show a permadiff with `"retryPolicy": nil,` ([example test case](https://github.com/GoogleCloudPlatform/terraform-google-conversion/blob/f8ac0b1b7915cd283393b1854d4d15d3f5d3876f/tfplan2cai/testdata/templates/example_pubsub_subscription.json#L30-L55) and [test code throwing the error](https://github.com/GoogleCloudPlatform/terraform-google-conversion/blob/7c35fd7478bb9702ba2c4e1ef325bc2f66488de8/tfplan2cai/test/read_test.go#L245)). I think that I'd need to submit a separate PR to fix those tests since they are in a different repo.

> I think you'll want allow_empty_object here as well, though it looks like the tests are passing, so maybe it's fine? Did you happen to try that out?

I didn't see that in the [field reference](https://github.com/GoogleCloudPlatform/magic-modules/blob/main/docs/content/develop/field-reference.md) but I see it [here](https://github.com/GoogleCloudPlatform/magic-modules/blob/2d34af8bcf47b390ede98ace828c98e80cb3acd0/mmv1/api/type.rb#L93) and used elsewhere in `Subscription.yaml`. But yeah, the tests are passing with or without `allow_empty_object` and I'm not sure why, but it doesn't seem like adding `allow_empty_object` will hurt so I've added it.

melinath (changes requested):
It looks like `send_empty_value` means that the nil value also needs to be included in TGC tests. Basically that means adding it to the data for both suscriptions in https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/third_party/tgc/tests/data/example_pubsub_subscription.json, and also adding it to the subscription data in the example_pubsub_subscription_iam_*.json files in the same directory.

anavada:
> It looks like send_empty_value means that the nil value also needs to be included in TGC tests.

Ah right, thanks. Updated and all tests are passing.

I'm not 100% clear on why we needed to change these tests but not the existing tests [here](https://github.com/GoogleCloudPlatform/magic-modules/tree/main/mmv1/third_party/terraform/services/pubsub) however. [`send_empty_value`](https://github.com/GoogleCloudPlatform/magic-modules/blob/main/docs/content/develop/field-reference.md#send_empty_value) is only supposed to send empty values to the API if they are explicitly set in the user config but they were not in the affected TGC tests.

melinath (commented):
> I'm not 100% clear on why we needed to change these tests but not the existing tests [here](https://github.com/GoogleCloudPlatform/magic-modules/tree/main/mmv1/third_party/terraform/services/pubsub) however. [`send_empty_value`](https://github.com/GoogleCloudPlatform/magic-modules/blob/main/docs/content/develop/field-reference.md#send_empty_value) is only supposed to send empty values to the API if they are explicitly set in the user config but they were not in the affected TGC tests.

It looks like the send_empty_value docs are incorrect - the resource will also send the value if the field isn't set. This is probably why all the tests had to be re-recorded here: https://github.com/GoogleCloudPlatform/magic-modules/pull/11834#issuecomment-2377627262 (because they all started sending the field as `null`, which didn't match the previous requests.)

As long as that's the same for the API as not sending the field, this should be fine. Let me know!

anavada:
> As long as that's the same for the API as not sending the field, this should be fine. Let me know!

Yeah it shouldn't be an issue for the API (as evidenced by the existing acceptance tests passing). I was just concerned about potentially breaking folks' Terraform code, but I don't think that will be a problem either.

PR #11841, opened 2024-09-26
Fixes for eap in the go engine

c2thorn:

```release-note:none

```

### DIFF
--- a/mmv1/api/product.go
+++ b/mmv1/api/product.go
@@ -240,7 +240,8 @@ func Merge(self, otherObj reflect.Value) {
 
 		// skip if the override is the "empty" value
 		emptyOverrideValue := reflect.DeepEqual(reflect.Zero(otherObj.Field(i).Type()).Interface(), otherObj.Field(i).Interface())
-		if emptyOverrideValue {
+
+		if emptyOverrideValue && selfObj.Type().Field(i).Name != "Required" {
 			continue
 		}
 
--- a/mmv1/api/resource.go
+++ b/mmv1/api/resource.go
@@ -1618,9 +1618,13 @@ func (r Resource) TestExamples() []resource.Examples {
 }
 
 func (r Resource) VersionedProvider(exampleVersion string) bool {
-	vp := r.MinVersion
+	var vp string
 	if exampleVersion != "" {
 		vp = exampleVersion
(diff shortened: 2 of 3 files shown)


mmv1/api/product.go
```diff
@@ -240,7 +240,8 @@ func Merge(self, otherObj reflect.Value) {
 
 		// skip if the override is the "empty" value
 		emptyOverrideValue := reflect.DeepEqual(reflect.Zero(otherObj.Field(i).Type()).Interface(), otherObj.Field(i).Interface())
-		if emptyOverrideValue {
+
+		if emptyOverrideValue && selfObj.Type().Field(i).Name != "Required" {
```
c2thorn, line 244: One of the private overrides actually has `Required: false`, which means we can't just treat it as an empty boolean value.

mmv1/api/resource.go
```diff
@@ -1618,9 +1618,13 @@ func (r Resource) TestExamples() []resource.Examples {
 }
 
 func (r Resource) VersionedProvider(exampleVersion string) bool {
-	vp := r.MinVersion
+	var vp string
 	if exampleVersion != "" {
 		vp = exampleVersion
+	} else if r.MinVersion == "" {
```
c2thorn, line 1624: follows the old ruby logic more closely. It didn't matter for GA/Beta, but it makes diffs in private. See the old logic spread out between:
https://github.com/GoogleCloudPlatform/magic-modules/blob/legacy-ruby/mmv1/api/resource.rb#L606
and
https://github.com/GoogleCloudPlatform/magic-modules/blob/legacy-ruby/mmv1/templates/terraform/examples/base_configs/test_file.go.erb#L55
and
https://github.com/GoogleCloudPlatform/magic-modules/blob/legacy-ruby/mmv1/templates/terraform/examples/base_configs/test_file.go.erb#L77

mmv1/api/product.go
roaks3: Yea, I remember this one. With the ruby implementation, you couldn't otherwise change a required field to optional, so yea, we'd need to keep this unless the Go version handles it differently.

roaks3 (commented):
LGTM

roaks3 (approved):

roaks3 (approved):

PR #11847, opened 2024-09-26
Add TeamCity testing project for ephemeral resources feature branch, clean up imports

SarahFrench:
## Adding projects in our TeamCity environment for testing ephemeral resources

This project will:
- Use a more recent Terraform binary version that contains the ephemeral feature
    - See `EphemeralResourcesTfCoreVersion`
- Have builds only for services were we're adding ephemeral resources
- Have builds that allow using the hashicorp and modular-magician owned repos (can test PRs pre-merge)
- CRON triggers to schedule builds (that use the hashicorp repo) every night

## Preview of the changes

**Here is a test project where this PR's config is deployed: https://hashicorp.teamcity.com/project/TerraformProviders_TerraformProviderGoogleSarahCheckingEphemeralTesting?mode=builds**

This is what the new addition to TeamCity will look like - we only need to test the resourcemanager package as all ephemeral resources we want to add are in that package:

Screenshot 2024-10-03 at 21 15 35

Here's a look at the trigger set on the builds using the HashiCorp-owned repos:

Screenshot 2024-10-03 at 21 18 53

```release-note:none

```

### DIFF
--- a/mmv1/third_party/terraform/.teamcity/components/builds/build_triggers.kt
+++ b/mmv1/third_party/terraform/.teamcity/components/builds/build_triggers.kt
@@ -36,7 +36,8 @@ class NightlyTriggerConfiguration(
 fun Triggers.runNightly(config: NightlyTriggerConfiguration) {
 
     schedule{
-        enabled = config.nightlyTestsEnabled
+        // enabled = config.nightlyTestsEnabled
+        enabled = false
         branchFilter = "+:" + config.branch // returns "+:/refs/heads/nightly-test" if default
         triggerBuild = always() // Run build even if no new commits/pending changes
         withPendingChangesOnly = false
--- a/mmv1/third_party/terraform/.teamcity/components/builds/build_triggers.kt
+++ b/mmv1/third_party/terraform/.teamcity/components/builds/build_triggers.kt
@@ -36,8 +36,7 @@ class NightlyTriggerConfiguration(
 fun Triggers.runNightly(config: NightlyTriggerConfiguration) {
 
     schedule{
-        // enabled = config.nightlyTestsEnabled
-        enabled = false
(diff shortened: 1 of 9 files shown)


mmv1/third_party/terraform/.teamcity/components/builds/build_triggers.kt
```diff
@@ -36,7 +36,8 @@ class NightlyTriggerConfiguration(
 fun Triggers.runNightly(config: NightlyTriggerConfiguration) {
 
     schedule{
-        enabled = config.nightlyTestsEnabled
+        // enabled = config.nightlyTestsEnabled
+        enabled = false
```
BBBmau, line 40: what's the reasoning for changing the default schedule trigger from true to false? Wouldn't we want to change this for the feature branch on these lines: https://github.com/GoogleCloudPlatform/magic-modules/pull/11847/files#diff-88a2ec67d77fa2a705a15f250a637d99cd7e6fd8f9ca524c2fddb25c9fa5ed3dR43-R45 ?

This would apply it to all projects disabling the nightly-test from being triggered, if my understanding is correct.

SarahFrench: The commit message explains- while the test deployment of this PR is up (link in the PR description) I've got the code disabling crons so we don't run nightly tests in duplicate.

BBBmau (approved):
Looks all good!

mmv1/third_party/terraform/.teamcity/components/builds/build_triggers.kt
SarahFrench, line 40: I'm pulling down the test deployment, so reverting this now

```suggestion
        enabled = config.nightlyTestsEnabled
```

PR #11852, opened 2024-09-27
feat: add HNS support for storage bucket

gurusai-voleti:

```release-note:REPLACEME

```

### DIFF
--- a/mmv1/third_party/terraform/services/storage/resource_storage_bucket.go.tmpl
+++ b/mmv1/third_party/terraform/services/storage/resource_storage_bucket.go.tmpl
@@ -549,6 +549,23 @@ func ResourceStorageBucket() *schema.Resource {
 					},
 				},
 			},
+			"hierarchical_namespace": {
+				Type:        schema.TypeList,
+				MaxItems:    1,
+				Optional:    true,
+				Description: `The bucket's HNS support, which defines bucket can organize folders in logical file system structure`,
+				Elem : &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"enabled": {
+							Type:        schema.TypeBool,
+							Default:     false,
+							Optional:    true,
+							ForceNew:    true,
+							Description: `Set this enabled flag to true when folders with logical files structure. Default value is false.`,
+						},
+					},
+				},
+			},
 		},
 		UseJSONNumber: true,
 	}
--- a/mmv1/third_party/terraform/services/storage/resource_storage_bucket_test.go.tmpl
+++ b/mmv1/third_party/terraform/services/storage/resource_storage_bucket_test.go.tmpl
@@ -18,6 +18,46 @@ import (
 	"google.golang.org/api/storage/v1"
 )
 
+func TestAccStorageBucket_basic_hns(t *testing.T) {
+	t.Parallel()
+
+	bucketName := acctest.TestBucketName(t)
+
+	acctest.VcrTest(t, resource.TestCase{
+		PreCheck:                 func() { acctest.AccTestPreCheck(t) },
+		ProtoV5ProviderFactories: acctest.ProtoV5ProviderFactories(t),
+		CheckDestroy:             testAccStorageBucketDestroyProducer(t),
+		Steps: []resource.TestStep{
+			{
+				Config: testAccStorageBucket_basic_hns(bucketName),
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr(
+						"google_storage_bucket.bucket", "hierarchical_namespace.0.enabled", "true"),
+				),
+			},
+			{
+				ResourceName:            "google_storage_bucket.bucket",
+				ImportState:             true,
+				ImportStateVerify:       true,
+				ImportStateVerifyIgnore: []string{"force_destroy"},
+			},
+		},
+	})
+}
+
+func testAccStorageBucket_basic_hns(bucketName string) string {
+	return fmt.Sprintf(`
+resource "google_storage_bucket" "bucket" {
+  name     = "%s"
+  location = "US"
+  uniform_bucket_level_access = true
+  hierarchical_namespace {
+	enabled = true
+  }
+}
+`, bucketName)
+}
+
 func TestAccStorageBucket_basic(t *testing.T) {
 	t.Parallel()
 


mmv1/third_party/terraform/services/storage/resource_storage_bucket.go.tmpl
```diff
@@ -549,6 +549,23 @@ func ResourceStorageBucket() *schema.Resource {
 					},
 				},
 			},
+			"hierarchical_namespace": {
+				Type:        schema.TypeList,
+				MaxItems:    1,
+				Optional:    true,
```
kautikdk, line 555: I think the whole block should be configured with ForceNew behavior. This block can not be changes or remove from the config once it is configured, Right?

mmv1/third_party/terraform/services/storage/resource_storage_bucket.go.tmpl
```diff
@@ -549,6 +549,23 @@ func ResourceStorageBucket() *schema.Resource {
...
+				Description: `The bucket's HNS support, which defines bucket can organize folders in logical file system structure`,
+				Elem : &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"enabled": {
+							Type:        schema.TypeBool,
+							Default:     false,
```
kautikdk, line 561: No need to set Default: false, If it is not API default. I believe this field should marked with Required=true as I can see that empty "hierarchical_namespace" won't be returned in the API response.

mmv1/third_party/terraform/services/storage/resource_storage_bucket.go.tmpl
```diff
@@ -549,6 +549,23 @@ func ResourceStorageBucket() *schema.Resource {
...
+							Optional:    true,
```
kautikdk, line 562: Can remove this in favor of `Required: true,`

kautikdk:
Apart from the field behavior comments, Please check bucket deletion. If the bucket is configured with `force_destroy=true` then while deleting bucket, it should delete all the folders and objects within the it. This may require changing logic in deletion operation. 
Also please add documentation of the newly added fields and PR description as well.

mmv1/third_party/terraform/services/storage/resource_storage_bucket.go.tmpl
gurusai-voleti: Yes updated block with ForceNew

mmv1/third_party/terraform/services/storage/resource_storage_bucket.go.tmpl
gurusai-voleti: removed Default

mmv1/third_party/terraform/services/storage/resource_storage_bucket.go.tmpl
gurusai-voleti: removed Optional and added required true

gurusai-voleti:
> Apart from the field behavior comments, Please check bucket deletion. If the bucket is configured with `force_destroy=true` then while deleting bucket, it should delete all the folders and objects within the it. This may require changing logic in deletion operation. Also please add documentation of the newly added fields and PR description as well.

->> ran manual test case locally as below
1) create HNS enabled bucket with force_destroy=true and upload files in bucket
2) check bucket is created with contents
3) when tried to delete bucket, terraform deleted the all the files in bucket and deleted the bucket

so no code changes needed for force_destroy for HNS bucket

mmv1/third_party/terraform/services/storage/resource_storage_bucket.go.tmpl
```diff
@@ -549,6 +549,22 @@ func ResourceStorageBucket() *schema.Resource {
...
 			},
+			"hierarchical_namespace": {
+				Type:        schema.TypeList,
+				MaxItems:    1,
+				Optional:    true,
+				ForceNew:    true,
+				Description: `The bucket's HNS support, which defines bucket can organize folders in logical file system structure`,
+				Elem : &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"enabled": {
```
kautikdk, line 560: Should we mark this field with ForceNew as well since it is not mutable?

mmv1/third_party/terraform/services/storage/resource_storage_bucket.go.tmpl
```diff
@@ -880,6 +900,12 @@ func resourceStorageBucketUpdate(d *schema.ResourceData, meta interface{}) error
 		}
 	}
 
+	if d.HasChange("hierarchical_namespace") {
+		if v, ok := d.GetOk("hierarchical_namespace"); ok {
+			sb.HierarchicalNamespace = expandBucketHierachicalNamespace(v.([]interface{}))
```
kautikdk, line 905: I believe, There is no need for checking this block in update as the block is marked with ForceNew which will never allow having updates for the block.

mmv1/third_party/terraform/services/storage/resource_storage_bucket_test.go.tmpl
```diff
@@ -18,6 +18,80 @@ import (
...
+	acctest.VcrTest(t, resource.TestCase{
+		PreCheck:                 func() { acctest.AccTestPreCheck(t) },
+		ProtoV5ProviderFactories: acctest.ProtoV5ProviderFactories(t),
+		CheckDestroy:             testAccStorageBucketDestroyProducer(t),
+		Steps: []resource.TestStep{
+			{
+				Config: testAccStorageBucket_basic_hns(bucketName),
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr(
+						"google_storage_bucket.bucket", "hierarchical_namespace.0.enabled", "true"),
```
kautikdk, line 35: Since we are not adding update tests, It would be worth adding test for `false` value as well.

mmv1/third_party/terraform/services/storage/resource_storage_bucket_test.go.tmpl
```diff
@@ -18,6 +18,80 @@ import (
...
+  location = "US"
+  uniform_bucket_level_access = true
+  hierarchical_namespace {
+	enabled = true
+  }
+}
+`, bucketName)
+}
+
+func testAccStorageBucket_basic_hns_with_data(bucketName string, enabled bool, forceDestroy bool) string {
```
kautikdk, line 81: Not sure where are we adding data here?

gurusai-voleti: in config we are creating the bucket and in check we are calling testAccCheckStorageBucketPutItem where a file is uploaded to bucket, 

I debugged the test case with TF_LOG=DEBUG and it executed as below  
1) bucket created
2) object uploaded to bucket
3) after test is successful, in infra cleaning step, as we enforced force_destroy true, objects got deleted and then bucket deleted

mmv1/third_party/terraform/services/storage/resource_storage_bucket.go.tmpl
gurusai-voleti: removed this block

mmv1/third_party/terraform/services/storage/resource_storage_bucket.go.tmpl
gurusai-voleti: currently this block contains only one resource I see no need of adding again here as the whole block is enforce with ForceNew true,

please suggest otherwise I will add ForceNew also at enabled field level @kautikdk

mmv1/third_party/terraform/services/storage/resource_storage_bucket_test.go.tmpl
gurusai-voleti: added false test case

mmv1/third_party/terraform/services/storage/resource_storage_bucket.go.tmpl
kautikdk: So actually, I asked for the ForceNew because users can not update this field. If we don't add this schema behavior then it will start generating diff every time after updating but won't have any effect. I would leave this to you for further investigation and decision.

mmv1/third_party/terraform/services/storage/resource_storage_bucket_test.go.tmpl
kautikdk: Okay, Got it!

mmv1/third_party/terraform/services/storage/resource_storage_bucket.go.tmpl
kautikdk: I see, So if the Resource level attributes can directly pass ForceNew behavior to it's nested fields then currently it makes sense to keep it only at top level now but in future if there is requirement, we can remove it from the top level and directly add it to the nested individual fields.

kautikdk:
Hi @rileykarson, Can you please trigger tests for this PR?

mmv1/third_party/terraform/services/storage/resource_storage_bucket.go.tmpl
rileykarson: This level will also need ForceNew set! I don't think we'd recreate in this case or the inverse as is, only when we removed the block:

```
  hierarchical_namespace = {
    enabled = false
  }
```

to

```
  hierarchical_namespace = {
    enabled = true
  }
```

mmv1/third_party/terraform/services/storage/resource_storage_bucket_test.go.tmpl
```diff
@@ -18,6 +18,103 @@ import (
...
+	return fmt.Sprintf(`
+resource "google_storage_bucket" "bucket" {
+  name     = "%s"
+  location = "US"
+  uniform_bucket_level_access = true
+  hierarchical_namespace {
+	enabled = %t
```
rileykarson, line 111:
```suggestion
    enabled = %t
```

mmv1/third_party/terraform/services/storage/resource_storage_bucket_test.go.tmpl
```diff
@@ -18,6 +18,103 @@ import (
...
+}
+
+func testAccStorageBucket_basic_hns(bucketName string, enabled bool) string {
+	return fmt.Sprintf(`
+resource "google_storage_bucket" "bucket" {
+  name     = "%s"
+  location = "US"
+  uniform_bucket_level_access = true
+  hierarchical_namespace {
+	enabled = %t
```
rileykarson, line 98:
```suggestion
    enabled = %t
```

mmv1/third_party/terraform/services/storage/resource_storage_bucket_test.go.tmpl
```diff
@@ -18,6 +18,103 @@ import (
...
+			{
+				ResourceName:            "google_storage_bucket.bucket",
+				ImportState:             true,
+				ImportStateVerify:       true,
+				ImportStateVerifyIgnore: []string{"force_destroy"},
+			},
+		},
+	})
+
+	// test case with hns disabled
```
rileykarson, line 47: Can you structure this as a separate TestAcc entry? I think that's the cause of the VCR replacing error, we don't expect multiple cases inside the same `t` scope.

gurusai-voleti: moved to a separate testcase

mmv1/third_party/terraform/services/storage/resource_storage_bucket.go.tmpl
gurusai-voleti: Added ForceNew at enabled field level as well

mmv1/third_party/terraform/services/storage/resource_storage_bucket_test.go.tmpl
kautikdk: What is need of separate test for `false` value when you already wrote one test with the same step? I think because of ForceNew behavior, The above update test will create new bucket resource so this test will be redundant.

mmv1/third_party/terraform/services/storage/resource_storage_bucket_test.go.tmpl
```diff
@@ -18,6 +18,162 @@ import (
...
+				ResourceName:            "google_storage_bucket.bucket",
+				ImportState:             true,
+				ImportStateVerify:       true,
+				ImportStateVerifyIgnore: []string{"force_destroy"},
+			},
+		},
+	})
+}
+
+// testcase to force_destroy HNS bucket
```
kautikdk, line 129: nit: test config

mmv1/third_party/terraform/website/docs/r/storage_bucket.html.markdown
```diff
@@ -269,6 +285,11 @@ The following arguments are supported:
 
 * `effective_time` - (Computed) Server-determined value that indicates the time from which the policy, or one with a greater retention, was effective. This value is in RFC 3339 format.
 
+<a name="nested_hierarchical_namespace"></a>The `hierarchical_namespace` block supports:
+
+* `enabled` - (Optional) The flag enables bucket folder management operations provide reliability and management capabilities, including creating, deleting, listing, and renaming folders.
```
kautikdk, line 290: This can be improved by breaking a long sentence into smaller and more clear sentences.

kautikdk (approved):
LGTM with few comments.

kautikdk:
Also after all the changes, Please squash your all commits into one clear commit.

mmv1/third_party/terraform/services/storage/resource_storage_bucket_test.go.tmpl
gurusai-voleti: agrees removed the false test case

mmv1/third_party/terraform/services/storage/resource_storage_bucket.go.tmpl
```diff
@@ -549,6 +549,23 @@ func ResourceStorageBucket() *schema.Resource {
 					},
 				},
 			},
+			"hierarchical_namespace": {
```
rileykarson, line 552: Will this object ever get returned by default in the future? Right now it appears not to, and only appears after getting explicitly set to true/false in its enabled field. 

If it does (even with false internal values, such as if a new field is added that causes the block to no longer serialize to null), that will break existing users and create a plan that attempts to recreate their buckets. We can mitigate, it will just cause unsetting the field to preserve the old value rather than remove the feature, i.e. turn it off.

mmv1/third_party/terraform/services/storage/resource_storage_bucket_test.go.tmpl
```diff
@@ -18,6 +18,134 @@ import (
 	"google.golang.org/api/storage/v1"
 )
 
+func TestAccStorageBucket_basic_hns(t *testing.T) {
```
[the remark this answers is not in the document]
gurusai-voleti: moved the testcases to last

mmv1/third_party/terraform/services/storage/resource_storage_bucket_test.go.tmpl
```diff
@@ -1545,6 +1545,134 @@ func TestAccStorageBucket_SoftDeletePolicy(t *testing.T) {
...
+
+	acctest.VcrTest(t, resource.TestCase{
+		PreCheck:                 func() { acctest.AccTestPreCheck(t) },
+		ProtoV5ProviderFactories: acctest.ProtoV5ProviderFactories(t),
+		CheckDestroy:             testAccStorageBucketDestroyProducer(t),
+		Steps: []resource.TestStep{
+			{
+				Config: testAccStorageBucket_basic_hns_with_data(bucketName, true),
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckStorageBucketPutItem(t, bucketName),
```
rileykarson, line 1642: Thinking about this some more, shouldn't we actually add hierarchical data here? That way we'd also be testing if force_destroy can properly clear hierarchical data from the bucket.

mmv1/third_party/terraform/services/storage/resource_storage_bucket_test.go.tmpl
```diff
@@ -1545,6 +1545,134 @@ func TestAccStorageBucket_SoftDeletePolicy(t *testing.T) {
...
+}
+
+func testAccStorageBucket_basic_hns_with_data(bucketName string, enabled bool) string {
+	return fmt.Sprintf(`
+resource "google_storage_bucket" "bucket" {
+  name     = "%s"
+  location = "US"
+  uniform_bucket_level_access = true
+  hierarchical_namespace {
+	enabled = %t
```
rileykarson, line 1669:
```suggestion
    enabled = %t
```

mmv1/third_party/terraform/services/storage/resource_storage_bucket_test.go.tmpl
```diff
@@ -1545,6 +1545,134 @@ func TestAccStorageBucket_SoftDeletePolicy(t *testing.T) {
...
+}
+
+func testAccStorageBucket_basic_hns(bucketName string, enabled bool) string {
+	return fmt.Sprintf(`
+resource "google_storage_bucket" "bucket" {
+  name     = "%s"
+  location = "US"
+  uniform_bucket_level_access = true
+  hierarchical_namespace {
+	enabled = %t
```
rileykarson, line 1656:
```suggestion
    enabled = %t
```

mmv1/third_party/terraform/services/storage/resource_storage_bucket_test.go.tmpl
gurusai-voleti: we don't have any terraform support yet to create folders in HNS enabled buckets

kautikdk: Agree!
Even though we don't have support for HNS folders in terraform yet, We can add some hierarchical data through storage client and test force_destroy accordingly.

mmv1/third_party/terraform/services/storage/resource_storage_bucket_test.go.tmpl
```diff
@@ -1545,6 +1545,134 @@ func TestAccStorageBucket_SoftDeletePolicy(t *testing.T) {
...
+	t.Parallel()
+
+	bucketName := acctest.TestBucketName(t)
+	acctest.VcrTest(t, resource.TestCase{
+		PreCheck:                 func() { acctest.AccTestPreCheck(t) },
+		ProtoV5ProviderFactories: acctest.ProtoV5ProviderFactories(t),
+		CheckDestroy:             testAccStorageBucketDestroyProducer(t),
+		Steps: []resource.TestStep{
+			{
+				Config: testAccStorageBucket_basic_hns(bucketName, false),
```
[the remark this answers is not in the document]
gurusai-voleti: I think this is fine as one test case we have enabled HNS first and disabled and one more test case to disabled and enable HNS again which recreates buckets in both test cases, this ensures both cases recreates bucket

mmv1/third_party/terraform/services/storage/resource_storage_bucket_test.go.tmpl
gurusai-voleti: I verified a manual testcase for  this,
1) create a bucket with HNS enabled and force_destroy=true with terraform
2) create folders and sub-folders which are HNS type content and upload objects in those folders manually on GCP console
3) now remove the resource in terraform and run terraform apply
4) all the files and objects got deleted and HNS folders left empty
5) The bucket failed to destroy as the bucket is not empty (empty folders are still there)

I verified there is no direct way to delete the empty folders recursively and list folders API also doesn't support order 
how to handle this case?

mmv1/third_party/terraform/services/storage/resource_storage_bucket.go.tmpl
```diff
@@ -1058,6 +1058,19 @@ func resourceStorageBucketDelete(d *schema.ResourceData, meta interface{}) error
...
 		wp.StopWait()
 	}
+	// get all folders and subfolders using List API folder
+	folder, _ := config.NewStorageClient(userAgent).Folders.List(bucket).Do()
+	var folderItems []*storage.Folder
+	if folder.Items != nil && len(folder.Items) > 0 {
+		folderItems = folder.Items
+	}
+	// delete folders from bottom to top to avoid no empty folder error
+	for j := len(folderItems) - 1; j >= 0; j-- {
```
pavandagagoogle: Why would you not use the similar mechanism like above of workerpool?
Retention policy check, listing error, etc?

gurusai-voleti: this is not needed, I removed this logic, there is some problem with the manually created folders on GCP console and creating folders through API,

when manually created there are some invisible files and so the bucket is not empty and the bucket not deleting.

I added a testcase with nested folders and uploading an object, the testcase is working fine and I verified through API's also

pavandagagoogle: Log a separate item. The UI creates a placeholder object i.e. an object and an object ending with a '/'. That should be handled outside the scope of the current feature is it is an existing integration.

Do you have tests which have a HNS bucket with only objects in a hierarchy and they get deleted? You can create objects in a hierarchy by creating an object ending in / and then adding objects under it for a non-folders use case and ensure that they do get all cleaned up.

mmv1/third_party/terraform/services/storage/resource_storage_bucket_test.go.tmpl
gurusai-voleti: added testcase for folders

mmv1/third_party/terraform/services/storage/resource_storage_bucket.go.tmpl
rileykarson: fyi @pavandagagoogle

rileykarson (commented):
We need to answer the unresolved comment on the resource; otherwise we can clean up the redundant test case and I think this is ready to merge.

mmv1/third_party/terraform/services/storage/resource_storage_bucket_test.go.tmpl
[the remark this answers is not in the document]
gurusai-voleti: removed the testcase

mmv1/third_party/terraform/services/storage/resource_storage_bucket.go.tmpl
pavandagagoogle: Riley this would be considered a breaking change on the API so this is not expected for the API.

mmv1/third_party/terraform/website/docs/r/storage_bucket.html.markdown
```diff
@@ -269,6 +285,14 @@ The following arguments are supported:
 
 * `effective_time` - (Computed) Server-determined value that indicates the time from which the policy, or one with a greater retention, was effective. This value is in RFC 3339 format.
 
+<a name="nested_hierarchical_namespace"></a>The `hierarchical_namespace` block supports:
+
+* `enabled` - (Optional) The flag enables folder management operations on storage bcuket
```
kautikdk, line 290: nit: "bucket"

gurusai-voleti: updated

mmv1/third_party/terraform/services/storage/resource_storage_bucket.go.tmpl
pavandagagoogle: @rileykarson - Does it need to be factored in at present or is it feasible to treat  hierarchical_namespace object not being present and config with enabled=false as equivalent? This could avoid recreating the bucket which could be a valid scenario.

rileykarson: As written we'd consider there to be several distinct states:

```
# 1
hierarchical_namespace = null # also unset
```

```
# 2 
hierarchical_namespace {
  enabled = false
}
```

```
# 3
hierarchical_namespace {
  enabled = true
}
```

If we can't transition from false -> null by nulling the object in an update request, we either need to apply the mitigation to consider null "preserve previous value" or add a custom one to map it onto the `enabled = false` state.

Because this field is immutable this should work as-is, but would need to be changed if:
  * update support was added in the future (to this field or any future field in the block)
  * the API changed to return a default when hierarchicalNamespace is not sent in the request at all

pavandagagoogle: In future we could have an update support for this field or any future field in the block.

So #1 & #2 should be considered as the same behavior.

pavandagagoogle: @rileykarson -  If this is something that would not be a breaking change then we can approve it else let us know what all tests we should add and what scenarios to handle.

rileykarson: Discussed offline; I think the safest change is to consider null to be "preserve old value"

mmv1/third_party/terraform/services/storage/resource_storage_bucket.go.tmpl
rileykarson, line 555:
```suggestion
				Optional:    true,
				Computed:    true,
```

mmv1/third_party/terraform/services/storage/resource_storage_bucket_test.go.tmpl
```diff
@@ -1545,6 +1545,142 @@ func TestAccStorageBucket_SoftDeletePolicy(t *testing.T) {
...
+			},
+			{
+				Config: testAccStorageBucket_uniformBucketAccessOnly(bucketName, true),
+			},
+			{
+				ResourceName:            "google_storage_bucket.bucket",
+				ImportState:             true,
+				ImportStateVerify:       true,
+				ImportStateVerifyIgnore: []string{"force_destroy"},
+			},
```
rileykarson, line 1589:
```suggestion
```

These test steps are not testing anything interesting as a result of the change adding `Computed` above so we can remove them!

mmv1/third_party/terraform/services/storage/resource_storage_bucket.go.tmpl
pavandagagoogle: Concluded to handle: null = hierachical_namespace {enabled = false } for a better UX

mmv1/third_party/terraform/services/storage/resource_storage_bucket_test.go.tmpl
rileykarson: obsolete

mmv1/third_party/terraform/services/storage/resource_storage_bucket.go.tmpl
rileykarson: obsolete

mmv1/third_party/terraform/services/storage/resource_storage_bucket.go.tmpl
rileykarson: Yep, let me grab a good reference for making that change!

mmv1/third_party/terraform/services/storage/resource_storage_bucket.go.tmpl
```diff
@@ -1294,6 +1315,31 @@ func flattenBucketSoftDeletePolicy(softDeletePolicy *storage.BucketSoftDeletePol
...
+		Enabled: (configuredHierachicalNamespacePolicy["enabled"].(bool)),
+	}
+	hierachicalNamespacePolicy.ForceSendFields = append(hierachicalNamespacePolicy.ForceSendFields, "Enabled")
+	return hierachicalNamespacePolicy
+}
+
+func flattenBucketHierarchicalNamespacePolicy(hierachicalNamespacePolicy *storage.BucketHierarchicalNamespace) []map[string]interface{} {
+	policies := make([]map[string]interface{}, 0, 1)
+	if hierachicalNamespacePolicy == nil {
+		return policies
```
rileykarson, line 1334:
```suggestion
		// a null object returned from the API is equivalent to a block with enabled = false
		// to handle this consistently, always write a null response as a hydrated block with false
		defaultPolicy :=map[string]interface{}{
			"enabled": false,
		}
		
		policies = append(policies, defaultPolicy)
		return policies
```

mmv1/third_party/terraform/services/storage/resource_storage_bucket.go.tmpl
rileykarson, line 555: We'll want to add a DiffSuppressFunc to this top-level field similar to this one: https://github.com/hashicorp/terraform-provider-google-beta/blob/804148d2559f8c6628258cfa232cd81c72ed6233/google-beta/services/container/resource_container_cluster.go#L6814-L6826

We're always storing the full ("hydrated") value in state in our flatten function where null is mapped to an item being present with `enabled = false`, so if we see a diff from `.#` (length) of 1 -> 0 and `enabled`  is `false`, we can return `true` from the function to suppress the diff.

mmv1/third_party/terraform/services/storage/resource_storage_bucket_test.go.tmpl
```diff
@@ -1545,6 +1545,142 @@ func TestAccStorageBucket_SoftDeletePolicy(t *testing.T) {
...
+				Config: testAccStorageBucket_basic_hns_with_data(bucketName, false),
+			},
+			{
+				ResourceName:            "google_storage_bucket.bucket",
+				ImportState:             true,
+				ImportStateVerify:       true,
+				ImportStateVerifyIgnore: []string{"force_destroy"},
+			},
+			{
+				Config: testAccStorageBucket_uniformBucketAccessOnly(bucketName, true),
```
rileykarson, line 1582: After the above changes we should add the `ExpectEmptyPlan` PlanCheck here: https://developer.hashicorp.com/terraform/plugin/testing/acceptance-tests/plan-checks#general-plan-checks

rileykarson (commented):
Suggestions / comments to add and test the final proposed mitigation inline

mmv1/third_party/terraform/website/docs/r/storage_bucket.html.markdown
```diff
@@ -269,6 +285,14 @@ The following arguments are supported:
 
 * `effective_time` - (Computed) Server-determined value that indicates the time from which the policy, or one with a greater retention, was effective. This value is in RFC 3339 format.
 
+<a name="nested_hierarchical_namespace"></a>The `hierarchical_namespace` block supports:
+
+* `enabled` - (Optional) The flag enables folder management operations on storage bucket
```
pavandagagoogle, line 290: Nit: Use the same language as here - https://cloud.google.com/sdk/gcloud/reference/storage/buckets/create#--enable-hierarchical-namespace

mmv1/third_party/terraform/services/storage/resource_storage_bucket.go.tmpl
gurusai-voleti: added the suggested change for DiffSuppress func

mmv1/third_party/terraform/services/storage/resource_storage_bucket_test.go.tmpl
gurusai-voleti: updated an existing test case

rileykarson (approved):
LGTM, thanks!

PR #11857, opened 2024-09-27
Handle location tag bindings of removed resources

mKeRix:
Currently, the `google_tags_location_tag_binding` resource will cause any plans to fail whenever the resource (`parent`) it was for is deleted outside of Terraform. Specifically, it results in errors like these during the read operation:

```
│ Error: Error when reading or editing TagsLocationTagBinding "<some-id>": googleapi: Error 403: The caller does not have permission
│ Details:
│ [
│   {
│     "@type": "type.googleapis.com/google.rpc.ResourceInfo",
│     "description": "permission [storage.buckets.listTagBindings] required (or the resource may not exist in this location)",
│     "resourceName": "//storage.googleapis.com/projects/_/buckets/<bucket-name>"
│   }
│ ]
│
│   with google_tags_location_tag_binding.binding,
│   on main.tf line 24, in resource "google_tags_location_tag_binding" "binding":
│   24: resource "google_tags_location_tag_binding" "binding" {
│
╵
```

The correct Terraform behavior would be that it recognizes this resource has been removed and drops it from the state. This PR implements that behavior, allowing plan operations to succeed even when the related resource was removed externally.

I tested the change successfully against my own test files, as well as the [reproduction snippet](https://gist.github.com/alexjmoore/0bfb62718f872e6fb72e73dedda2ca3e) from the linked issue. 

Fixes hashicorp/terraform-provider-google#15928.

```release-note:bug
tags:  removed `google_tags_location_tag_binding` resource from the Terraform state when its parent resource has been removed outside of Terraform
```

### DIFF
--- a/mmv1/third_party/terraform/services/tags/resource_tags_location_tag_bindings.go.tmpl
+++ b/mmv1/third_party/terraform/services/tags/resource_tags_location_tag_bindings.go.tmpl
@@ -7,8 +7,10 @@ import (
 	"strings"
 	"time"
 
+    "github.com/hashicorp/errwrap"
 	"github.com/hashicorp/terraform-provider-google/google/tpgresource"
 	transport_tpg "github.com/hashicorp/terraform-provider-google/google/transport"
+	"google.golang.org/api/googleapi"
 
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 )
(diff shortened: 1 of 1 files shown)


zli82016 (approved):
LGTM. I tested locally and the resource `google_tags_location_tag_binding` is removed from the Terraform state when its parent is deleted outside of Terraform

PR #11860, opened 2024-09-27
Add acceptance tests for how provider handles `scopes` argument

SarahFrench:
This PR adds acceptance tests for usage of scopes that demonstrate:

- how the provider behaves when provider configuration arguments come from different sources ( config vs ENVs)
    - There are no ENVs for scopes
    - There is a default set of scopes used if the argument isn't supplied
- schema-level validation that's in place
    - There isn't any validation about empty lists, instead the empty list is ignored
- use cases: how does this argument impact the providers behaviour in plan/apply
    - I've implemented a test case where setting the scopes to incorrect values causes provisioning to fail.

```release-note:none

```

### DIFF
--- a/mmv1/third_party/terraform/provider/provider_scopes_test.go
+++ b/mmv1/third_party/terraform/provider/provider_scopes_test.go
@@ -1,0 +1,161 @@
+package provider_test
+
+import (
+	"fmt"
+	"regexp"
+	"testing"
+
+	"github.com/hashicorp/terraform-plugin-testing/helper/resource"
+	"github.com/hashicorp/terraform-provider-google/google/acctest"
+	"github.com/hashicorp/terraform-provider-google/google/transport"
+)
+
+// TestAccSdkProvider_scopes is a series of acc tests asserting how the SDK provider handles scopes arguments
+// It is SDK specific because the HCL used provisions SDK-implemented resources
+// It is a counterpart to TestAccFwProvider_scopes
+func TestAccSdkProvider_scopes(t *testing.T) {
+	testCases := map[string]func(t *testing.T){
+		// Configuring the provider using inputs
+		"default scopes are used when there are no user inputs": testAccSdkProvider_scopes_providerDefault,
+		"scopes can be set in config":                           testAccSdkProvider_scopes_setInConfig,
+		//no ENVs to test
+
+		// Schema-level validation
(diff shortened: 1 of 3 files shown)


SarahFrench:
TeamCity tests:

- [SDK](https://hashicorp.teamcity.com/buildConfiguration/TerraformProviders_GoogleCloud_GOOGLE_BETA_MMUPSTREAMTESTS_GOOGLEBETA_PACKAGE_PROVIDER/240887?hideTestsFromDependencies=false&hideProblemsFromDependencies=false&expandBuildChangesSection=true)
- [PF](https://hashicorp.teamcity.com/buildConfiguration/TerraformProviders_GoogleCloud_GOOGLE_BETA_MMUPSTREAMTESTS_GOOGLEBETA_PACKAGE_FWPROVIDER/240888?hideTestsFromDependencies=false&hideProblemsFromDependencies=false&expandBuildChangesSection=true)

SarahFrench:
I'm not sure what's going on in the failure above, very hard to search in the logs

mmv1/third_party/terraform/provider/provider_scopes_test.go
```diff
@@ -0,0 +1,161 @@
...
+// It is a counterpart to TestAccFwProvider_scopes
+func TestAccSdkProvider_scopes(t *testing.T) {
+	testCases := map[string]func(t *testing.T){
+		// Configuring the provider using inputs
+		"default scopes are used when there are no user inputs": testAccSdkProvider_scopes_providerDefault,
+		"scopes can be set in config":                           testAccSdkProvider_scopes_setInConfig,
+		//no ENVs to test
+
+		// Schema-level validation
+		"when scopes is set to an empty array in the config the value is ignored and default scopes are used": testAccSdkProvider_scopes_emptyArray,
```
c2thorn, line 24: nice capturing this case!

c2thorn (approved):

mmv1/third_party/terraform/provider/provider_scopes_test.go
SarahFrench: Thanks 😁 We could add some validation in future to make the scenario more explicit, i.e. block empty arrays so users aren't lulled into a false sense of using no scopes, but that'd be a breaking change. And I don't even know if setting zero scopes is something users would attempt to do, but I can imagine there's someone out there doing it!

PR #11861, opened 2024-09-27
Switched to emoji for color-based emphasis

melinath:
Also updated template formatting tests to be less prescriptive about the exact formatting of the returned message.

Fixed https://github.com/hashicorp/terraform-provider-google/issues/19301

```release-note:none

```

### DIFF
--- a/.ci/magician/cmd/templates.go
+++ b/.ci/magician/cmd/templates.go
@@ -17,7 +17,7 @@ func color(color, text string) string {
 	case "green":
 		emoji = "🟢"
 	default:
-		emoji = ""
+		return text
 	}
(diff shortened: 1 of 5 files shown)


roaks3 (approved):
Nice solution!

(Personal preference: I would probably vote for just the preceding emoji, instead of one before AND after)

melinath:
> (Personal preference: I would probably vote for just the preceding emoji, instead of one before AND after)

Yeah, it's a bit much - made that change in [7612963](https://github.com/GoogleCloudPlatform/magic-modules/pull/11861/commits/7612963a9376bf35de7365232710f75e2d6fe58b) and added a space after the emoji.

PR #11866, opened 2024-09-28
Update `google_sql_database` and `google_sql_user` to not do Read actions when instance is not active

jharshika:
This pull request addresses issue [hashicorp/terraform-provider-google#7810](https://github.com/hashicorp/terraform-provider-google/issues/7810). If the activation policy is not ALWAYS i.e. the instance is not active, return from the get resource method of "user" and "database". In the API layer if the activation policy is not set the ALWAYS then get resource will throw an error. To avoid breaking `terraform plan`, updating the getter method to avoid the error during refresh.

With these changes, `terraform plan` and `terraform apply` will succeed. If the customer try to update/delete the resource, it will still throw exception as it is not supported for inactive resources.

```release-note:none
sql: updated `google_sql_database` and `google_sql_user` so they don't perform Read operations if their associated `google_sql_database_instance` has `activation_policy` set to "NEVER". This avoids triggering API errors while the instance is unavailable.
```


SarahFrench (commented):
Hi @jharshika, thanks for pivoting from your previous approach to solve this issue!

I believe I understand the problem and the solution in this PR is option 5 described in the linked GitHub issue:

>Option 5: Check the activation_policy before refreshing

Given the [existing tests all pass here](https://github.com/GoogleCloudPlatform/magic-modules/pull/11866#issuecomment-2383652918) it looks like this change doesn't impact provisioning google_sql_* resources. However it looks like there are no tests covering the scenario that triggers the API errors in the linked GitHub issue.

**Could you please add an acceptance test that would have triggered the API error prior to your PR?** I'm imagining an [update test](https://googlecloudplatform.github.io/magic-modules/develop/test/test/#add-an-update-test) where the first step provisions a google_sql_database_instance with activation_policy="ALWAYS", a google_sql_database, and a google_sql_user resource, and then in a second step activation_policy is updated to "NEVER"? There might be a better approach that better resembles how a user might trigger this API error, but either way I think we should add a test to prove the issue is solved.

jharshika:
Hi @SarahFrench Added the tests and verified it locally as well to verify the functionality.

mmv1/third_party/terraform/services/sql/resource_sql_user_test.go
```diff
@@ -324,6 +324,64 @@ func TestAccSqlUser_mysqlPasswordPolicy(t *testing.T) {
...
+					testAccCheckGoogleSqlUserExists(t, "google_sql_user.user"),
+				),
+			},
+		},
+	})
+}
+
+func testGoogleSqlUser_instanceWithActivationPolicy(instance, activationPolicy string) string {
+	return fmt.Sprintf(`
+resource "google_sql_database_instance" "test_instance" {
```
SarahFrench, line 365:
```suggestion
resource "google_sql_database_instance" "instance" {
```

SarahFrench (commented):
Thanks for adding the test!

I ran the new code and tests from this PR in our test environment and I found a few things that needed a fix. However they're mostly to do with how the testing framework works instead of the config, so I figured I'd suggest the changes in a review. Also, I confirmed that the test fails when it's run using the code from main.

I hope it's ok but I'll go ahead and merge the suggested changes, run the tests, and we should be in a position to merge.

mmv1/third_party/terraform/services/sql/resource_sql_user_test.go
```diff
@@ -324,6 +324,64 @@ func TestAccSqlUser_mysqlPasswordPolicy(t *testing.T) {
...
+					testAccCheckGoogleSqlUserExists(t, "google_sql_user.user"),
+				),
+			},
+			// Step 3: Refresh to verify no errors
+			{
+				Config: testGoogleSqlUser_instanceWithActivationPolicy(instance, "NEVER"),
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckGoogleSqlUserExists(t, "google_sql_user.user"),
+				),
+			},
```
SarahFrench, line 358: Creating a single suggestion that combines the two previous suggestions affecting the same code

```suggestion
			// Step 2: Update activation_policy to NEVER
			{
				Config: testGoogleSqlUser_instanceWithActivationPolicy(instance, "NEVER"),
			},
			// Step 3: Refresh to verify no errors
			{
				Config: testGoogleSqlUser_instanceWithActivationPolicy(instance, "NEVER"),
			},
			// Step 4: Update activation_policy to ALWAYS so that post-test destroy code is able to delete the google_sql_user resource
			{
				Config: testGoogleSqlUser_instanceWithActivationPolicy(instance, "ALWAYS"),
				Check: resource.ComposeTestCheckFunc(
					testAccCheckGoogleSqlUserExists(t, "google_sql_user.user"),
				),
			},
```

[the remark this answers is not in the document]
SarahFrench: Applied in 3a6dd99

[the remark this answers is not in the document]
SarahFrench: Applied in 3a6dd99

SarahFrench:
For future reference by maintainers, here's a build running that acc tess with the latest code: [BUILD](https://hashicorp.teamcity.com/buildConfiguration/TerraformProviders_GoogleCloud_GOOGLE_MMUPSTREAMTESTS_GOOGLE_PACKAGE_SQL/247384?expandBuildChangesSection=true&hideTestsFromDependencies=false&hideProblemsFromDependencies=false)

mmv1/products/sql/Database.yaml
```diff
@@ -48,6 +48,7 @@ async:
 collection_url_key: 'items'
 custom_code:
   pre_delete: 'templates/terraform/pre_delete/sql_database_deletion_policy.tmpl'
+  pre_read: 'templates/terraform/pre_read/sql_database_activation_policy.tmpl'
```
SarahFrench, line 51: With the other test you've made as an example, could you please make a similar acceptance test that ensures that the Read action on the database resource is disabled as expected when the instance is availability_type = "NEVER"? It would essentially be a copy/paste of the test you've already added but with the database resource present in the config instead.

SarahFrench (commented):
Ok, the TestAccSqlUser_instanceWithActivationPolicy test is passing! 🎉 
Just to make sure nothing is overlooked, could you add a similar test for the database resource?

jharshika:
Thanks @SarahFrench for the suggested code changes. Updated the unit test for both user and database.

SarahFrench:
I'm just [running the skipped test](https://hashicorp.teamcity.com/buildConfiguration/TerraformProviders_GoogleCloud_GOOGLE_MMUPSTREAMTESTS_GOOGLE_PACKAGE_SQL/251041?hideTestsFromDependencies=false&hideProblemsFromDependencies=false&expandBuildChangesSection=true) again to confirm everything's ok.  Assuming it passes I'll approve and merge. Thank you for the additional test!

SarahFrench (approved):
[The test mentioned here](https://github.com/GoogleCloudPlatform/magic-modules/pull/11866#issuecomment-2420289280) passed, so approving & merging. Thanks for such a smooth PR process!

SarahFrench:
Final note on this PR for any future reference:

I've set the release note to 

>sql: updated `google_sql_database` and `google_sql_user` so they don't perform Read operations if their associated `google_sql_database_instance` has `activation_policy` set to "NEVER". This avoids triggering API errors while the instance is unavailable.

This PR doesn't prevent create/update/delete actions as the API error is useful in those cases. We need to skip Read when the instance isn't available, as the instance isn't queryable then. There may be Terraform configurations that are valid and contain a `google_sql_database_instance` with  other child resources, and then `activation_policy` is temporarily set to "NEVER". In those cases refresh actions will trigger API errors. When the instance is returned to being available then the child resources will resume as normal.

PR #11477, opened 2024-08-17
Handle email casing in google_bigquery_dataset_access

wj-chen:
Fixes https://github.com/hashicorp/terraform-provider-google/issues/17994.

**Release Note Template for Downstream PRs (will be copied)**

```release-note:bug
bigquery: fixed an error which could occur with email field values containing non-lower-case characters in `google_bigquery_dataset_access`
```

PR #11875, opened 2024-09-30
Handle service account casing in google_bigquery_dataset_access

wj-chen:
Fixes https://github.com/hashicorp/terraform-provider-google/issues/19547.

A follow-up to https://github.com/GoogleCloudPlatform/magic-modules/pull/11477. Service account names come back from the `GET` API in their original capitalization instead of being normalized to lower-case so need to be specifically handled.

There isn't a good way to add a unit test organically since `google_service_account` only allows lower-case names and the error came from customers creating upper-case service accounts outside of Terraform.

```release-note:bug
bigquery: fixed an error which could occur with service account field values containing non-lower-case characters in `google_bigquery_dataset_access`
```


sachinpro (approved):

slevenick (approved):
Seems reasonable to me!

PR #11880, opened 2024-10-01
Set `config_sync.enabled` field to `true` to fix broken tests

haiyanmeng:
Terraform google provider v5.41.0 (released on Aug 12, 2024) introduced a new field `config_sync.enabled` to the google_gke_hub_feature_membership resource. The default value of this field is false. However, when the field is omitted from the Terraform resource, Terraform sets the field to false in the ACM Hub API explicitly. This issue causes the Terraform users to fail to install Config Sync unless they set the new field to true explicitly.

To mitigate this issue, we enhanced the Hub CLH to throw an error if Config Sync is explicitly disabled and the Git or OCI configuration is set.

Our Terraform users need to set the new field to true explicitly to install Config Sync from v5.41.0. For existing CS installations, upgrading from Terraform versions prior to v5.41.0 to versions >= v5.41.0 does not break users, and does not uninstall Config Sync.

part of https://github.com/hashicorp/terraform-provider-google/issues/14591. 

```release-note:none

```

### DIFF
--- a/mmv1/third_party/terraform/services/gkehub/resource_gke_hub_feature_membership_test.go.tmpl
+++ b/mmv1/third_party/terraform/services/gkehub/resource_gke_hub_feature_membership_test.go.tmpl
@@ -108,6 +108,7 @@ resource "google_gke_hub_feature_membership" "feature_member_1" {
   configmanagement {
     version = "1.18.2"
     config_sync {
+      enabled = true
       source_format = "hierarchy"
       git {
         sync_repo   = "https://github.com/GoogleCloudPlatform/magic-modules"
@@ -125,6 +126,7 @@ resource "google_gke_hub_feature_membership" "feature_member_2" {
   configmanagement {
     version = "1.18.2"
     config_sync {
+      enabled = true
       source_format = "hierarchy"
       git {
         sync_repo   = "https://github.com/terraform-providers/terraform-provider-google"
@@ -176,6 +178,7 @@ resource "google_gke_hub_feature_membership" "feature_member_2" {
   configmanagement {
     version = "1.18.2"
     config_sync {
+      enabled = true
       source_format = "hierarchy"
       git {
         sync_repo   = "https://github.com/terraform-providers/terraform-provider-google-beta"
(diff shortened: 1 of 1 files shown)


mmv1/third_party/terraform/services/gkehub/resource_gke_hub_feature_membership_test.go.tmpl
```diff
@@ -157,6 +159,7 @@ resource "google_gke_hub_feature_membership" "feature_member_1" {
   configmanagement {
     version = "1.18.2"
     config_sync {
+      enabled = true
```
zli82016, line 162: It is already set on line 164.

zli82016 (changes requested):

mmv1/third_party/terraform/services/gkehub/resource_gke_hub_feature_membership_test.go.tmpl
haiyanmeng: > It is already set on line 164.

Removed.

zli82016:
These two tests [passed](https://hashicorp.teamcity.com/buildConfiguration/TerraformProviders_GoogleCloud_GOOGLE_BETA_MMUPSTREAMTESTS_GOOGLEBETA_PACKAGE_GKEHUB/242922?buildTab=tests) with the fix in this PR.
`TestAccGKEHubFeatureMembership_gkehubFeatureAcmOci` `TestAccGKEHubFeatureMembership_gkehubFeatureAcmAllFields`

The error for configSync.enabled is addressed for `TestAccGKEHubFeatureMembership_gkehubFeatureAcmUpdate`. It fails with a [perma-diff.](https://hashicorp.teamcity.com/buildConfiguration/TerraformProviders_GoogleCloud_GOOGLE_BETA_MMUPSTREAMTESTS_GOOGLEBETA_PACKAGE_GKEHUB/242929?hideTestsFromDependencies=false&hideProblemsFromDependencies=false&expandBuildTestsSection=true&expandBuildDeploymentsSection=false), which can be fixed in a separate PR.

PR #11882, opened 2024-10-01
Update documentation to list `global` as the only supported location for logging resource `LogScope`

tyh2333:
Only updating the public doc to make `global` the only supported location for logging resource LogScope.

We only support global for logging resource LogScope, so updating the public doc to reflect that.
The error handling will be added in a follow up PR since a test is broken by a different service and I'm not able to test it.

Error: 
google-beta/services/compute/data_source_google_compute_region_instance_group_manager.go:28:30: undefined: tpgresource.ParseRegionalInstanceGroupManagersFieldValue
make: *** [GNUmakefile:29: vet] Error 1

```release-note:none
```

### DIFF
--- a/mmv1/products/logging/LogScope.yaml
+++ b/mmv1/products/logging/LogScope.yaml
@@ -52,8 +52,7 @@ parameters:
   - name: 'location'
     type: String
     description:
-      'The location of the resource. The supported locations are: global,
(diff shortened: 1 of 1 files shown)


SarahFrench (approved):
Thanks for making this fix!

PR #11884, opened 2024-10-01
Add acceptance tests for provider configuration behaviour using `add_terraform_attribution_label` and `terraform_attribution_label_addition_strategy` 

SarahFrench:
This PR adds acceptance tests checking how `add_terraform_attribution_label` and `terraform_attribution_label_addition_strategy` are used when configuring the SDK and PF providers.

This PR doesn't include any tests of how the fields impact provisioning resources, as I assume there is lots of test coverage for that elsewhere.

These tests show that there's a bit of disparity between the SDK and PF when the default value is used - the SDK provider stores the default value in the meta struct but the PF provider does not. I'm not sure how the default value is used downstream from there, if at all.

```release-note:none

```


SarahFrench:
[SDK test in TeamCity](https://hashicorp.teamcity.com/buildConfiguration/TerraformProviders_GoogleCloud_GOOGLE_BETA_MMUPSTREAMTESTS_GOOGLEBETA_PACKAGE_PROVIDER/242926?hideTestsFromDependencies=false&hideProblemsFromDependencies=false&expandBuildChangesSection=true)

[PF test in TeamCity](https://hashicorp.teamcity.com/buildConfiguration/TerraformProviders_GoogleCloud_GOOGLE_BETA_MMUPSTREAMTESTS_GOOGLEBETA_PACKAGE_FWPROVIDER/242927?hideTestsFromDependencies=false&hideProblemsFromDependencies=false&expandBuildChangesSection=true)

mmv1/third_party/terraform/fwprovider/framework_provider_terraform_attribution_label_addition_strategy_test.go
```diff
@@ -0,0 +1,98 @@
...
+)
+
+// TestAccFwProvider_terraform_attribution_label_addition_strategy is a series of acc tests asserting how the plugin-framework provider handles terraform_attribution_label_addition_strategy arguments
+// It is plugin-framework specific because the HCL used provisions plugin-framework-implemented resources
+// It is a counterpart to TestAccSdkProvider_terraform_attribution_label_addition_strategy
+func TestAccFwProvider_terraform_attribution_label_addition_strategy(t *testing.T) {
+	testCases := map[string]func(t *testing.T){
+		// Configuring the provider using inputs
+		"config sets terraform_attribution_label_addition_strategy values":                                                                                 testAccFwProvider_terraform_attribution_label_addition_strategy_configUsed,
+		"when terraform_attribution_label_addition_strategy is unset in the config, the default value'CREATION_ONLY' is NOT set in the provider meta data": testAccFwProvider_terraform_attribution_label_addition_strategy_defaultValue,
```
c2thorn, line 17: Wouldn't we want the default value here? From what I understand `CREATION_ONLY` should be behavior that we would want to transfer to framework resources

c2thorn: Does this mean that we need to explicitly set the default value somewhere?

c2thorn (approved):
test themselves are fine, but I have a question about a default value that I may have missed the explanation for.

mmv1/third_party/terraform/fwprovider/framework_provider_terraform_attribution_label_addition_strategy_test.go
SarahFrench: From a quick look, I think that [only the SDK logic currently handles the two attribution-related arguments](https://github.com/hashicorp/terraform-provider-google/blob/738e5b8246db5495fa66f8fd46f3ff59b133bde5/google/provider/provider.go#L924-L935). For example if I do a global search for `transport_tpg.CreateOnlyAttributionStrategy` I cannot see it in the PF code.

However, once we start using the SDK configuration logic on the PF side of the provider this disparity should go away, I hope. Unless there are other bits of labels-related code that needs to be implemented on resources/data sources, but I'm not familiar with that work.

SarahFrench:
Because this test shows SDK/PF differences we'll be able to revisit this topic on the muxing fixes PR. Once we change how the PF provider in the mux server is configured the PF test cases will start failing, and we can plan any related changes that are needed (i.e. new GH issues to open).

Given that I'll go ahead and merge, but will put a note on the muxing fix PR

PR #11885, opened 2024-10-01
Update docs to specify workaround for required field:

slevenick:
Fixes: https://github.com/hashicorp/terraform-provider-google/issues/19630

Docs-only change to include workaround as we don't want to backport this

```release-note:none

```


mmv1/third_party/terraform/website/docs/r/container_cluster.html.markdown
```diff
@@ -1293,6 +1293,8 @@ Enables monitoring and attestation of the boot integrity of the instance. The at
 * `cpu_manager_policy` - (Optional) The CPU management policy on the node. See
 [K8S CPU Management Policies](https://kubernetes.io/docs/tasks/administer-cluster/cpu-management-policies/).
 One of `"none"` or `"static"`. If unset (or set to the empty string `""`), the API will treat the field as if set to "none".
+Prior to the 6.4.0 this field was marked as required. The workaround for the required field 
+is setting the empty string `""` will function identically to not setting this field.
```
melinath, line 1297:
```suggestion
is setting the empty string `""`, which will function identically to not setting this field.
```

melinath (approved):

PR #11886, opened 2024-10-02
Add plan-time validation of `name` on `google_compute_instance`

karolgorc:
closes https://github.com/hashicorp/terraform-provider-google/issues/18265
- Added name validation to `google_compute_instance` 

The regexes are wrong in the validation file. Our verification functions expected 2-63 characters when in reality it's 1-63. Basing on this documentation: https://cloud.google.com/compute/docs/naming-resources#resource-name-format and manual testing

Some other tests might fail due to these changes in the validate file so pls paste the test errors after the CICD run 

```release-note:enhancement
compute: added plan-time validation to `name` on `google_compute_instance`
```

### DIFF
--- a/mmv1/third_party/terraform/verify/validation.go
+++ b/mmv1/third_party/terraform/verify/validation.go
@@ -22,7 +22,7 @@ const (
 
 	SubnetworkLinkRegex = "projects/(" + ProjectRegex + ")/regions/(" + RegionRegex + ")/subnetworks/(" + SubnetworkRegex + ")$"
 
-	RFC1035NameTemplate = "[a-z]([-a-z0-9]{%v,%v}[a-z0-9])?"
+	RFC1035NameTemplate = "[a-z]([-a-z0-9]{%d,%d}[a-z0-9])?"
 	CloudIoTIdRegex     = "^[a-zA-Z][-a-zA-Z0-9._+~%]{2,254}$"
 
 	// Format of default Compute service accounts created by Google
--- a/mmv1/third_party/terraform/verify/validation.go
+++ b/mmv1/third_party/terraform/verify/validation.go
@@ -22,7 +22,7 @@ const (
 
 	SubnetworkLinkRegex = "projects/(" + ProjectRegex + ")/regions/(" + RegionRegex + ")/subnetworks/(" + SubnetworkRegex + ")$"
 
-	RFC1035NameTemplate = "[a-z]([-a-z0-9]{%d,%d}[a-z0-9])?"
+	RFC1035NameTemplate = "[a-z]([-a-z0-9]%v[a-z0-9])?"
 	CloudIoTIdRegex     = "^[a-zA-Z][-a-zA-Z0-9._+~%]{2,254}$"
 
 	// Format of default Compute service accounts created by Google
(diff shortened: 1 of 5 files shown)


mmv1/third_party/terraform/verify/validation.go
```diff
@@ -22,7 +22,7 @@ const (
 
 	SubnetworkLinkRegex = "projects/(" + ProjectRegex + ")/regions/(" + RegionRegex + ")/subnetworks/(" + SubnetworkRegex + ")$"
 
-	RFC1035NameTemplate = "[a-z](?:[-a-z0-9]{%d,%d}[a-z0-9])"
+	RFC1035NameTemplate = "[a-z]([-a-z0-9]{%d,%d}[a-z0-9])?"
```
rileykarson, line 25: This allows the string "a" when any input is supplied. For example, `1,61` (source input to the fn of `2,63`). Added some extra test cases as constraints.

I suspect we need to rework the regex substantially to go below a minimum of 2 while properly validating the format. We may want to mix procedural code and regex or just write this as entirely procedural code.

mmv1/third_party/terraform/verify/validation_test.go
```diff
@@ -174,7 +174,7 @@ func TestValidateRFC1035Name(t *testing.T) {
 		{TestName: "cannot end with a dash", Min: 6, Max: 10, Value: "invalid-", ExpectError: true},
 		{TestName: "too short", Min: 6, Max: 10, Value: "short", ExpectError: true},
 		{TestName: "too long", Min: 6, Max: 10, Value: "toolooooong", ExpectError: true},
-		{TestName: "min too small", Min: 1, Max: 10, Value: "", ExpectError: true},
+		{TestName: "min too small", Min: 0, Max: 10, Value: "", ExpectError: true},
 		{TestName: "min < max", Min: 6, Max: 5, Value: "", ExpectError: true},
```
rileykarson, line 178:
```suggestion
		{TestName: "min < max", Min: 6, Max: 5, Value: "", ExpectError: true},
		{TestName: "invalid smallest possible w/ higher limit", Min: 2, Max: 63, Value: "a", ExpectError: true},
		{TestName: "invalid smallest possible hyphen", Min: 1, Max: 1, Value: "-", ExpectError: true},
		{TestName: "invalid smallest possible ends with hyphen", Min: 2, Max: 63, Value: "a-", ExpectError: true},
```

mmv1/third_party/terraform/verify/validation.go
karolgorc: agreed. After checking this regex fails on one new test case that you've added. This is mainly due to length which should be procedural imo.

The only other think that this regex is affecting are Service Account names so will look out for errors there as well.

karolgorc:
sorry. Pushed a rebased branch on accident.

What was done.
- added new test cases
- The function now validates the length outside of the regex.
- The variable with the regex takes in one string value instead of 2 ints
```
RFC1035NameTemplate = "[a-z]([-a-z0-9]%v[a-z0-9])?"
```
This allows to input a `"*"` wildcard for our case but also can take `"{4,28}"` as input not to break the existing ServiceAccount regexes. Because they specify length in the regex while we do it through the function.

Here are 2 unit tests affected
```
=== RUN   TestValidateRFC1035Name
--- PASS: TestValidateRFC1035Name (0.00s)

=== RUN   TestValidateServiceAccountLink
--- PASS: TestValidateServiceAccountLink (0.00s)
```

karolgorc:
```
gofmt needs running on the following files:
./google/services/sql/resource_sql_database_test.go
You can use the command: `make fmt` to reformat code.
make: *** [GNUmakefile:26: fmtcheck] Error 1
```

doesn't seem related

rileykarson (approved):

PR #11889, opened 2024-10-02
Added dialogflow_agent_to_link field to the google_discovery_engine_chat_engine resource

averbuks:
Added `chat_engine_config.dialogflow_agent_to_link` field to the `google_discovery_engine_chat_engine` resource. 
Field exists in the [API](https://cloud.google.com/generative-ai-app-builder/docs/reference/rest/v1/projects.locations.collections.engines#chatengineconfig) and allows users to make use of an existing DialogFlow Agent (for ex. deployed by terraform), hence agent management can be done by terraform, including security settings, page configurations and many more.

Also fixes https://github.com/hashicorp/terraform-provider-google/issues/19678 

```release-note:enhancement
discoveryengine: added `chat_engine_config.dialogflow_agent_to_link` field to `google_discovery_engine_chat_engine` resource
```


mmv1/products/discoveryengine/ChatEngine.yaml
```diff
@@ -125,7 +129,12 @@ properties:
         type: NestedObject
         description: |
           The configuration to generate the Dialogflow agent that is associated to this Engine.
-        required: true
+          Exactly one of `agent_creation_config` or `dialogflow_agent_to_link` must be set.
+        immutable: true
+        ignore_read: true
```
c2thorn, line 134: why are we adding `ignore_read` to this one here? Did it not correctly read before?

averbuks: ignore_read is set on the parent field here https://github.com/GoogleCloudPlatform/magic-modules/blob/bbf9f4086880429a90c5272e117d24541698ac68/mmv1/products/discoveryengine/ChatEngine.yaml#L126
So I decided to repeat it for both nested fields to have it set more explicit in the code, but you are right, it can be removed since parent field has it set already

c2thorn: thanks for the context

c2thorn (approved):

PR #11891, opened 2024-10-02
Add unit tests for the VCR request matcher function

SarahFrench:
Part of https://github.com/hashicorp/terraform-provider-google/issues/19685

This PR adds tests for the logic used to match a request a provider wants to make to the request data recorded in cassettes.

From the tests, it looks like upgrading the version of Terraform used for VCR tests (TF version impacts User-Agent values) does not impact matching. This is because the User-Agent header doesn't appear to be used at all in the matching logic.

```release-note:note
provider: minor refactoring of how we run acceptance tests in VCR mode
```

### DIFF
--- a/mmv1/third_party/terraform/acctest/vcr_utils_test.go
+++ b/mmv1/third_party/terraform/acctest/vcr_utils_test.go
@@ -219,6 +219,105 @@ func TestNewVcrMatcherFunc_canDetectMismatches(t *testing.T) {
 	}
 }
 
+// Currently there is no code to actively force the matcher to overlook differing User-Agent values.
+// It isn't checked at any point in the matcher logic.
+func TestNewVcrMatcherFunc_ignoresDifferentUserAgents(t *testing.T) {
+
+	cases := map[string]struct {
+		httpRequest     requestDescription
+		cassetteRequest requestDescription
+		expectMatch     bool
+	}{
+		"GET requests with different useragents are matched": {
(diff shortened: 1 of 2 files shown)


mmv1/third_party/terraform/acctest/vcr_utils_test.go
```diff
@@ -0,0 +1,379 @@
...
+}
+
+// Currently there is no code to actively force the matcher to overlook differing User-Agent values.
+// It isn't checked at any point in the matcher logic.
+func TestNewVcrMatcherFunc_ignoresDifferentUserAgents(t *testing.T) {
+
+	cases := map[string]struct {
+		httpRequest     requestDescription
+		cassetteRequest requestDescription
+		expectMatch     bool
```
SarahFrench, line 229: I just realised that this is no longer needed, but I'll save pushing a commit to remove it until after the automation completes and/or review happens

SarahFrench:
I'm going to push a commit to make the automation think that it's the reviewer's turn. I've confused it by doing a self review

mmv1/third_party/terraform/acctest/vcr_utils_test.go
SarahFrench: 1959f8b824f0d16ec3aa43c9e25d344c83e2a43b

slevenick (approved):

PR #11893, opened 2024-10-02
StorageTransfer Service: Support FederatedIdentifyConfig for Azure Blob Storage

venky999:

Support FederatedIdentifyConfig in StorageTransfer Service for Azure Blob Storage 
Fixes https://github.com/hashicorp/terraform-provider-google/issues/19526

Note: Will work after merging https://github.com/googleapis/google-api-go-client/pull/2810


slevenick (commented):
Can you add a test that specifies this?

venky999:
> Can you add a test that specifies this?

I dont see any tests for azure blob storage  for other usecases, so I have not added thinking that the test might fail ..

slevenick (commented):
I'm unable to test this as well. Asking someone with credentials to follow-up here

slevenick:
Ah, it looks like we need to wait on the go client to be merged to be able to test

venky999:
> Ah, it looks like we need to wait on the go client to be merged to be able to test

Yes I raised a PR but its closed as its a copy. looks like it needs to be done from google side

PR #11896, opened 2024-10-02
Speed up goimports usage

SirGitsalot:
Now that the go compiler is live I've started toying with bazel, and the extra time that `goimports` adds to each compile run is irking me. This PR speeds things up by running `goimports` once as a last compilation step rather than running it for each generated source file.

`goimports` performs two functions for us: removing imports that are declared but not used, and adding imports that are used but not declared. The former is fast, the latter is slow - like, a couple of orders of magnitude slower than the former. The difference exists because removing imports requires parsing only the go file that's being fixed, while adding imports potentially parses every go file in the project, the standard library, and in your `GOPATH`.  `goimports` has an internal cache that avoids some repetitive parsing, and by using a single execution that runs over all of the generated files at once we're able to leverage that cache.

Here's some example timings on my M1 Macbook Pro. Before:

```
$ time make provider VERSION=beta OUTPUT_PATH="../terraform-provider-google-beta"
real	1m1.976s
user	1m33.130s
sys	2m37.034s

$ time make provider VERSION=ga OUTPUT_PATH="../terraform-provider-google"
real	1m0.521s
user	1m23.280s
sys	2m25.612s
```

After:

```
$ time make provider VERSION=beta OUTPUT_PATH="../terraform-provider-google-beta"
real	0m24.659s
user	0m38.855s
sys	0m18.785s

$ time make provider VERSION=ga OUTPUT_PATH="../terraform-provider-google"
real	0m22.059s
user	0m36.664s
sys	0m19.448s
```

YMMV of course, and full builds benefit a lot more than single product builds.

Ultimately the best way to speed up the compiler will be to include all imports that we're using, and the change to `resource.go.tmpl` picks the lowest hanging fruit (for context, 10 seconds of the speed up on my machine came from just those five lines). The new `--show-import-diffs` flag is useful for identifying the files that need import fixes.

```release-note:none
```

### DIFF
--- a/mmv1/main.go
+++ b/mmv1/main.go
@@ -56,6 +56,8 @@ var yamlTempMode = flag.Bool("yaml-temp", false, "copy text over from ruby yaml 
 var handwrittenTempFiles = flag.String("handwritten-temp", "", "copy specific handwritten files over from .erb to go .tmpl.temp comma separated")
 var templateTempFiles = flag.String("template-temp", "", "copy specific templates over from .erb to go .tmpl.temp comma separated")
 
+var showImportDiffs = flag.Bool("show-import-diffs", false, "write go import diffs to stdout")
+
 func main() {
 
 	flag.Parse()
@@ -193,6 +195,8 @@ func main() {
 	if generateCode {
 		providerToGenerate.CompileCommonFiles(*outputPath, productsForVersion, "")
 	}
+
+	provider.FixImports(*outputPath, *showImportDiffs)
 }
 
 func GenerateProduct(productChannel chan string, providerToGenerate provider.Provider, productsForVersionChannel chan *api.Product, startTime time.Time, productsToGenerate []string, resourceToGenerate, overrideDirectory string, generateCode, generateDocs bool) {
--- a/mmv1/provider/template_data.go
+++ b/mmv1/provider/template_data.go
@@ -15,6 +15,7 @@ package provider
 
 import (
 	"bytes"
+	"errors"
 	"fmt"
 	"go/format"
 	"log"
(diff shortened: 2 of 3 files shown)


c2thorn:
about to be OOO, can you take a look over this @zli82016  ?

mmv1/main.go
```diff
@@ -56,6 +56,8 @@ var yamlTempMode = flag.Bool("yaml-temp", false, "copy text over from ruby yaml
 var handwrittenTempFiles = flag.String("handwritten-temp", "", "copy specific handwritten files over from .erb to go .tmpl.temp comma separated")
 var templateTempFiles = flag.String("template-temp", "", "copy specific templates over from .erb to go .tmpl.temp comma separated")
 
+var showImportDiffs = flag.Bool("show-import-diffs", false, "write go import diffs to stdout")
```
zli82016, line 59: Do you plan to add `show-import-diffs` flag to the CLI in the make file?
https://github.com/GoogleCloudPlatform/magic-modules/blob/main/GNUmakefile#L61-L68

zli82016 (approved):
LGTM. Amazing improvement.

mmv1/main.go
SirGitsalot: No, for a couple of reasons:

- It dumps a **lot** of info into the console (and it's info that most developers don't care about)
- It runs a lot slower (since it has to run `goimports` twice)

I added the flag to make it easy for the core team to add future import fixes, and temporarily adding it to the makefile is how I found the `resource.go.tmpl` imports :-)

zli82016: Sounds good.

PR #11897, opened 2024-10-02
Adding replication field for Filestore Instance

NA2047:
Adding replication field for resource_filestore_instance

```release-note:enhancement
filestore: added `initial_replication` field for peer instance configuration and `effective_replication` output for replication configuration output to `google_filestore_instance`
```

### DIFF
--- a/mmv1/products/filestore/Instance.yaml
+++ b/mmv1/products/filestore/Instance.yaml
@@ -336,20 +336,6 @@ properties:
       resource, see the `google_tags_tag_value` resource.
     immutable: true
     ignore_read: true
-  - name: 'tags'
-    type: KeyValuePairs
-    description: |
-      A map of resource manager tags. Resource manager tag keys
-      and values have the same definition as resource manager
-      tags. Keys must be in the format tagKeys/{tag_key_id},
-      and values are in the format tagValues/456. The field is
-      ignored when empty. The field is immutable and causes
-      resource replacement when mutated. This field is only set
-      at create time and modifying this field after creation
-      will trigger recreation. To apply tags to an existing
-      resource, see the `google_tags_tag_value` resource.
-    immutable: true
-    ignore_read: true
   - name: 'initialReplication'
     api_name: replication
     type: NestedObject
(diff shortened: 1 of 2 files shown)


mmv1/products/filestore/Instance.yaml
```diff
@@ -294,3 +294,48 @@ properties:
     type: String
     description: |
       The reason for enabling deletion protection.
+  - name: 'replication'
```
rileykarson, line 297: My understanding here is that we set this value on creation and then never set it again from a client. However, can the server modify the (non-output) values under the user?

Depending on the answer we'll mix `immutable` and `ignore_read`, probably. Potentially a CustomizeDiff.

rileykarson:
[internal discussion / context](https://docs.google.com/document/d/1Tmxsvl1mJr_EJDJSGzXhMmcL32VJx3RAXPqtugOOaaE/edit?resourcekey=0-8vxjIWDF8uh7hdDVPO_sFQ&tab=t.0#bookmark=id.u4q5qlrakafc)

NickElliot:
from office hours meeting -- running tests again to see if quota increase is necessary or if quota was hit from how many builds got ran at once. quota increase would be possible, but our us-central1 region is at the default max quota of 6tb, so an increase would require approval

mmv1/products/filestore/Instance.yaml
```diff
@@ -336,3 +336,65 @@ properties:
...
+          The replication role.
+        default_value: "STANDBY"
+        enum_values:
+          - 'ROLE_UNSPECIFIED'
+          - 'ACTIVE'
+          - 'STANDBY'
+      - name: 'replicas'
+        type: Array
+        description: |
+          The replication role.
```
rileykarson, line 359: Could we clarify usage in the description here? i.e. "This should be specified on the replica, marking the active as the `peer_instance`" or similar.

mmv1/third_party/terraform/services/filestore/resource_filestore_instance_test.go
```diff
@@ -460,3 +460,74 @@ tags = {`, name)
...
+    network         = "default"
+    modes           = ["MODE_IPV4"]
+  }
+
+  initial_replication {
+	replicas {
+	  peer_instance = google_filestore_instance.instance.id
+	     }
+	}
+}
```
rileykarson, line 531:
```suggestion
  initial_replication {
    replicas {
      peer_instance = google_filestore_instance.instance.id
    }
  }
}
```

rileykarson (approved):

PR #11898, opened 2024-10-02
Fix the regression of handling deleted IAM members in google_bigquery_dataset_iam*

wj-chen:
After https://github.com/GoogleCloudPlatform/magic-modules/pull/9948, the following logic in `accessToIamMember` changed from:
```
if member, ok := access["iamMember"]; ok {
	return member.(string), nil
}
```
to
```
if member, ok := access["iamMember"]; ok {
	return fmt.Sprintf("iamMember:%s", member.(string)), nil
}
```
Since deleted service accounts show up in `dataset.access` as an `iamMember`:
```
{
    "iamMember": "deleted:serviceAccount:sa-tf-deletion-test-a@wjchen-osconfig-test.iam.gserviceaccount.com?uid=110999291001447756607",
    "role": "READER"
}
```
It means that the new `access["iamMember"]` value changed from `deleted:serviceAccount:...` to `iamMember:deleted:serviceAccount:...`.

```release-note:bug
bigquery: fixed a regression that caused `google_bigquery_dataset_iam_*` resources to attempt to set deleted IAM members, thereby triggering an API error 
```

### DIFF
--- a/mmv1/third_party/terraform/services/bigquery/iam_bigquery_dataset.go
+++ b/mmv1/third_party/terraform/services/bigquery/iam_bigquery_dataset.go
@@ -189,7 +189,7 @@ func policyToAccess(policy *cloudresourcemanager.Policy) ([]map[string]interface
 		}
 		for _, member := range binding.Members {
 			// Do not append any deleted members
-			if strings.HasPrefix(member, "deleted:") {
+			if strings.HasPrefix(member, "iamMember:deleted:") {
 				continue
 			}
 			access := map[string]interface{}{
@@ -211,7 +211,7 @@ func policyToAccess(policy *cloudresourcemanager.Policy) ([]map[string]interface
 // Dataset access uses different member types to identify groups, domains, etc.
 // these types are used as keys in the access JSON payload
 func iamMemberToAccess(member string) (string, string, error) {
-	if strings.HasPrefix(member, "deleted:") {
+	if strings.HasPrefix(member, "iamMember:deleted:") {
 		return "", "", fmt.Errorf("BigQuery Dataset IAM member is deleted: %s", member)
 	}
 	pieces := strings.SplitN(member, ":", 2)
--- a/mmv1/third_party/terraform/services/bigquery/iam_bigquery_member_dataset.go
+++ b/mmv1/third_party/terraform/services/bigquery/iam_bigquery_member_dataset.go
@@ -231,7 +231,7 @@ func policyToAccessForIamMember(policy *cloudresourcemanager.Policy) ([]map[stri
 		}
 		for _, member := range binding.Members {
 			// Do not append any deleted members
-			if strings.HasPrefix(member, "deleted:") {
+			if strings.HasPrefix(member, "iamMember:deleted:") {
 				continue
 			}
 			access := map[string]interface{}{
@@ -253,7 +253,7 @@ func policyToAccessForIamMember(policy *cloudresourcemanager.Policy) ([]map[stri
 // Dataset access uses different member types to identify groups, domains, etc.
 // these types are used as keys in the access JSON payload
 func iamMemberToAccessForIamMember(member string) (string, string, error) {
-	if strings.HasPrefix(member, "deleted:") {
+	if strings.HasPrefix(member, "iamMember:deleted:") {
 		return "", "", fmt.Errorf("BigQuery Dataset IAM member is deleted: %s", member)
 	}
 	pieces := strings.SplitN(member, ":", 2)
--- a/mmv1/third_party/terraform/services/bigquery/resource_bigquery_dataset_iam_member_test.go
+++ b/mmv1/third_party/terraform/services/bigquery/resource_bigquery_dataset_iam_member_test.go
@@ -3,7 +3,7 @@ package bigquery_test
 import (
 	"fmt"
 	"reflect"
-	"strings"
+	"regexp"
 	"testing"
 
 	"github.com/hashicorp/terraform-plugin-testing/helper/resource"
@@ -114,6 +114,48 @@ func TestAccBigqueryDatasetIamMember_iamMember(t *testing.T) {
 	})
 }
 
+func TestAccBigqueryDatasetIamMember_withDeletedServiceAccount(t *testing.T) {
+	t.Parallel()
+
+	datasetID := fmt.Sprintf("tf_test_%s", acctest.RandString(t, 10))
+	serviceAccountID := fmt.Sprintf("tf-test-%s", acctest.RandString(t, 10))
+	serviceAccountEmail := fmt.Sprintf("%s@%s.iam.gserviceaccount.com", serviceAccountID, envvar.GetTestProjectFromEnv())
+
+	expected := map[string]interface{}{
+		"role":        "roles/viewer",
+		"userByEmail": serviceAccountEmail,
+	}
+
+	acctest.VcrTest(t, resource.TestCase{
+		PreCheck:                 func() { acctest.AccTestPreCheck(t) },
+		ProtoV5ProviderFactories: acctest.ProtoV5ProviderFactories(t),
+		Steps: []resource.TestStep{
+			{
+				Config: testAccBigqueryDatasetIamMember_withServiceAccount(datasetID, serviceAccountID),
+				Check:  testAccCheckBigQueryDatasetIamMemberPresent(t, "google_bigquery_dataset.dataset", expected),
+			},
+			{
+				// No change from Step 1 except for hard-coding the service account email to remove the explicit resource reference in preparation for the service accoutn deletion.
+				Config: testAccBigqueryDatasetIamMember_withServiceAccountEmail(datasetID, serviceAccountID, serviceAccountEmail),
+				Check:  testAccCheckBigQueryDatasetIamMemberPresent(t, "google_bigquery_dataset.dataset", expected),
+			},
+			{
+				// Delete the service account but keep the rest the same.
+				// The service account will show up with a "deleted:serviceAccount:" prefix and a "?uid=" suffix when getting dataset access from now on.
+				// The plan should show an attempt to recreate the IAM member resource because the dataset access no longer contains the original account email so there is a diff.
+				Config:             testAccBigqueryDatasetIamMember_withServiceAccountDeleted(datasetID, serviceAccountEmail),
+				ExpectNonEmptyPlan: true,
+			},
+			{
+				// Apply the same plan as the previous step.
+				// It should fail because the specified account has now been deleted, so the account email is invalid.
+				Config:      testAccBigqueryDatasetIamMember_withServiceAccountDeleted(datasetID, serviceAccountEmail),
+				ExpectError: regexp.MustCompile(fmt.Sprintf(".+Service account %s does not exist.+", serviceAccountEmail)),
+			},
+		},
+	})
+}
+
 func testAccCheckBigQueryDatasetIamMemberPresent(t *testing.T, n string, expected map[string]interface{}) resource.TestCheckFunc {
 	return testAccCheckBigQueryDatasetIamMember(t, n, expected, true)
 }
@@ -146,15 +188,6 @@ func testAccCheckBigQueryDatasetIamMember(t *testing.T, n string, expected map[s
 		}
 		access := ds["access"].([]interface{})
 		for _, a := range access {
-			if aMap, ok := a.(map[string]interface{}); ok {
-				if iamMember, ok := aMap["iamMember"].(string); ok {
-					// The iam account may have been deleted but the binding may be present for the dataset.
-					// This case is supposed to always throw an error.
-					if strings.HasPrefix(iamMember, "deleted:") {
-						return fmt.Errorf("Found deleted service account: %s", iamMember)
-					}
-				}
-			}
 			if reflect.DeepEqual(a, expected) {
 				if !expectPresent {
 					return fmt.Errorf("Found access %+v, expected not present", expected)
@@ -257,4 +290,54 @@ resource "google_iam_workload_identity_pool_provider" "wif_provider" {
   }
 }
 `, datasetID, wifIDs, wifIDs)
+}
+
+func testAccBigqueryDatasetIamMember_withServiceAccount(datasetID, serviceAccountID string) string {
+	return fmt.Sprintf(`
+resource "google_bigquery_dataset" "dataset" {
+  dataset_id = "%s"
+}
+
+resource "google_service_account" "sa" {
+  account_id = "%s"
+}
+
+resource "google_bigquery_dataset_iam_member" "access" {
+  dataset_id = google_bigquery_dataset.dataset.dataset_id
+  role       = "roles/viewer"
+  member     = "serviceAccount:${google_service_account.sa.email}"
+}
+`, datasetID, serviceAccountID)
+}
+
+func testAccBigqueryDatasetIamMember_withServiceAccountEmail(datasetID, serviceAccountID, serviceAccountEmail string) string {
+	return fmt.Sprintf(`
+resource "google_bigquery_dataset" "dataset" {
+  dataset_id = "%s"
+}
+
+resource "google_service_account" "sa" {
+  account_id = "%s"
+}
+
+resource "google_bigquery_dataset_iam_member" "access" {
+  dataset_id = google_bigquery_dataset.dataset.dataset_id
+  role       = "roles/viewer"
+  member     = "serviceAccount:%s"
+}
+`, datasetID, serviceAccountID, serviceAccountEmail)
+}
+
+func testAccBigqueryDatasetIamMember_withServiceAccountDeleted(datasetID, serviceAccountEmail string) string {
+	return fmt.Sprintf(`
+resource "google_bigquery_dataset" "dataset" {
+  dataset_id = "%s"
+}
+
+resource "google_bigquery_dataset_iam_member" "access" {
+  dataset_id = google_bigquery_dataset.dataset.dataset_id
+  role       = "roles/viewer"
+  member     = "serviceAccount:%s"
+}
+`, datasetID, serviceAccountEmail)
 }


SarahFrench (commented):
HI @wj-chen, thanks for this bug fix! From reading through the code changes in https://github.com/GoogleCloudPlatform/magic-modules/pull/9948 and this PR I'm fairly confident about what the problem is and how your proposed changes would fix it.

Is there a way to write an acceptance test that confirms deleted members are handled correctly? Maybe one that 1) provisions a Bigquery dataset with X service accounts as members in the policy and 2) in the next step deletes one of the service accounts. Would this trigger the error or would an additional plan/apply step be needed to trigger the issue prior to your fix?

Let me know your thoughts!

wj-chen:
> HI @wj-chen, thanks for this bug fix! From reading through the code changes in #9948 and this PR I'm fairly confident about what the problem is and how your proposed changes would fix it.
> 
> Is there a way to write an acceptance test that confirms deleted members are handled correctly? Maybe one that 1) provisions a Bigquery dataset with X service accounts as members in the policy and 2) in the next step deletes one of the service accounts. Would this trigger the error or would an additional plan/apply step be needed to trigger the issue prior to your fix?
> 
> Let me know your thoughts!

I experimented a bit. The test has to be set up in three steps:
1) Create one service_account and assign it a role in an iam_member
```
resource "google_bigquery_dataset" "dataset" {
  dataset_id = "id"
}

resource "google_service_account" "sa" {
  account_id = "sa-1"
}

resource "google_bigquery_dataset_iam_member" "access" {
  dataset_id = google_bigquery_dataset.dataset.dataset_id
  role       = "roles/viewer"
  member     = "serviceAccount:${google_service_account.sa.email}"
}
```
2) Remove that one service_account
```
resource "google_bigquery_dataset" "dataset" {
  dataset_id = "id"
}

resource "google_bigquery_dataset_iam_member" "access" {
  <haven't figured out how to set this up properly>
}
```
3) Create another service_account and assign that a role in the iam_member
```
resource "google_bigquery_dataset" "dataset" {
  dataset_id = "id"
}

resource "google_service_account" "sa" {
  account_id = "sa-2"
}

resource "google_bigquery_dataset_iam_member" "access" {
  dataset_id = google_bigquery_dataset.dataset.dataset_id
  role       = "roles/viewer"
  member     = "serviceAccount:${google_service_account.sa.email}"
}
```

If we skip 2), test passes but Terraform always seems to update `iam_member` to remove that service account member first before deleting the service account itself, which makes sense in terms of dependency management, but doesn't cause `deleted:serviceAccount:...` to surface. So we need the service account deletion as a prerequisite.

If we include 2), there will be some dataset `GET` responses with `iamMember:deleted:`, but the content of `iam_member` in that config needs to update, because the member is no longer the first service account email, but `deleted:serviceAccount:...` instead. However, since the deleted service account email now always comes with a `?uid=<some number>` suffix, Terraform plan fails with a diff, even if I don't include a `Check:` step. As the UID is dynamic, I don't know if there is a way to code that in at plan time.

This is tricky so let me know if you have other ideas of how to set up this test to make step 2) pass.

wj-chen:
> > HI @wj-chen, thanks for this bug fix! From reading through the code changes in #9948 and this PR I'm fairly confident about what the problem is and how your proposed changes would fix it.
> > Is there a way to write an acceptance test that confirms deleted members are handled correctly? Maybe one that 1) provisions a Bigquery dataset with X service accounts as members in the policy and 2) in the next step deletes one of the service accounts. Would this trigger the error or would an additional plan/apply step be needed to trigger the issue prior to your fix?
> > Let me know your thoughts!
> 
> I experimented a bit. The test has to be set up in three steps:
> 
> 1. Create one service_account and assign it a role in an iam_member
> 
> ```
> resource "google_bigquery_dataset" "dataset" {
>   dataset_id = "id"
> }
> 
> resource "google_service_account" "sa" {
>   account_id = "sa-1"
> }
> 
> resource "google_bigquery_dataset_iam_member" "access" {
>   dataset_id = google_bigquery_dataset.dataset.dataset_id
>   role       = "roles/viewer"
>   member     = "serviceAccount:${google_service_account.sa.email}"
> }
> ```
> 
> 2. Remove that one service_account
> 
> ```
> resource "google_bigquery_dataset" "dataset" {
>   dataset_id = "id"
> }
> 
> resource "google_bigquery_dataset_iam_member" "access" {
>   <haven't figured out how to set this up properly>
> }
> ```
> 
> 3. Create another service_account and assign that a role in the iam_member
> 
> ```
> resource "google_bigquery_dataset" "dataset" {
>   dataset_id = "id"
> }
> 
> resource "google_service_account" "sa" {
>   account_id = "sa-2"
> }
> 
> resource "google_bigquery_dataset_iam_member" "access" {
>   dataset_id = google_bigquery_dataset.dataset.dataset_id
>   role       = "roles/viewer"
>   member     = "serviceAccount:${google_service_account.sa.email}"
> }
> ```
> 
> If we skip 2), test passes but Terraform always seems to update `iam_member` to remove that service account member first before deleting the service account itself, which makes sense in terms of dependency management, but doesn't cause `deleted:serviceAccount:...` to surface. So we need the service account deletion as a prerequisite.
> 
> If we include 2), there will be some dataset `GET` responses with `iamMember:deleted:`, but the content of `iam_member` in that config needs to update, because the member is no longer the first service account email, but `deleted:serviceAccount:...` instead. However, since the deleted service account email now always comes with a `?uid=<some number>` suffix, Terraform plan fails with a diff, even if I don't include a `Check:` step. As the UID is dynamic, I don't know if there is a way to code that in at plan time.
> 
> This is tricky so let me know if you have other ideas of how to set up this test to make step 2) pass.

@SarahFrench Looking for your guidance here ^. For more context - the original issue was raised for a customer with two Terraform directories and the they saw the failure after 1) creating two service accounts in the first directory, 2) creating an iam_member with one service account in the second directory, 3) removing that service account in the first directory, and 4) updating iam_member to the other account in the second directory. I don't imagine there is a way to set up a test like that but please let me know otherwise.

SarahFrench:
👋🏻 Hi @wj-chen , sorry for the delay reviewing! I'll take a look at this tomorrow and will experiment to see the problems you described and see if there are any solutions I can propose.

SarahFrench:
I've been taking a look at this and using the config you've provided above and want to double check I understand the problem correctly. Here's something I've been trying:

**Step 1 - Setup**
Apply works fine
```tf
resource "google_bigquery_dataset" "dataset" {
  dataset_id = "tf_test_my_dataset"
}

resource "google_service_account" "sa" {
  account_id = "tf-test-my-sa"
}

resource "google_bigquery_dataset_iam_member" "access" {
  dataset_id = google_bigquery_dataset.dataset.dataset_id
  role       = "roles/viewer"
  member     = "serviceAccount:${google_service_account.sa.email}"
}
```

**Step 2 - prepare for deleting the service account by hardcoding the `member` value**
This doesn't cause any diffs/is a no-op
```diff
resource "google_bigquery_dataset" "dataset" {
  dataset_id = "tf_test_my_dataset"
}

resource "google_service_account" "sa" {
  account_id = "tf-test-my-sa"
}

resource "google_bigquery_dataset_iam_member" "access" {
  dataset_id = google_bigquery_dataset.dataset.dataset_id
  role       = "roles/viewer"
-  member     = "serviceAccount:${google_service_account.sa.email}"
+  member     = "serviceAccount:tf-test-my-sa@<PROJECT_ID>.iam.gserviceaccount.com"
}
```

**Step 3 - Delete the service account**
Apply works fine, and the plan only includes removing the google_service_account resource
```tf
resource "google_bigquery_dataset" "dataset" {
  dataset_id = "tf_test_my_dataset"
}

resource "google_bigquery_dataset_iam_member" "access" {
  dataset_id = google_bigquery_dataset.dataset.dataset_id
  role       = "roles/viewer"
  member     = "serviceAccount:tf-test-my-sa@PROEJCT-ID.iam.gserviceaccount.com"
}
```

**Step 4 - run `terraform plan`**
The plan indicates that Terraform cannot see the google_bigquery_dataset_iam_member's member in the IAM binding anymore (as the entry includes `deleted` and the `uuid` now) so Terraform believes this resource needs to be recreated.
```
  # google_bigquery_dataset_iam_member.access will be created
  + resource "google_bigquery_dataset_iam_member" "access" {
      + dataset_id = "tf_test_my_dataset"
      + etag       = (known after apply)
      + id         = (known after apply)
      + member     = "serviceAccount:tf-test-my-sa@<PROJECT_ID>.iam.gserviceaccount.com"
      + project    = (known after apply)
      + role       = "roles/viewer"
    }
```

**Step 5 - run `terraform apply`**
The apply fails with an error, but which error happens depends on the provider version in use.

If I use the latest Google provider release (v6.6.0) then this error is returned that mentions unknown prefix types, which I believe is the regression you are fixing:

```
│ Error: Error applying IAM policy for Bigquery Dataset <PROJECT_ID>/tf_test_sarah:
Error creating DatasetAccess: googleapi: Error 400:
IAM setPolicy failed for Dataset <PROJECT_ID>:tf_test_sarah:
The member deleted:serviceaccount:tf-test-oob-e0ybkvrluu@<PROJECT_ID>.iam.gserviceaccount.com?uid=112611438796804060811 is of an unknown type. Please set a valid type prefix for the member., invalid
│ 
│   with google_bigquery_dataset_iam_member.deleted,
│   on main.tf line 15, in resource "google_bigquery_dataset_iam_member" "deleted":
│   15: resource "google_bigquery_dataset_iam_member" "deleted" {
│ 
```

If I change which version of the Google provider I use in step 6 I see a different error.
I see the error below if I use:
* The code from your PR
* The provider version that predates https://github.com/GoogleCloudPlatform/magic-modules/pull/9948, which is v5.17.0

```
│ Error: Error applying IAM policy for Bigquery Dataset <PROJECT_ID>/tf_test_my_dataset:
Error creating DatasetAccess: googleapi: Error 400:
IAM setPolicy failed for Dataset <PROJECT_ID>:tf_test_my_dataset:
Service account tf-test-my-sa@<PROJECT_ID>.iam.gserviceaccount.com does not exist., invalid
│ 
│   with google_bigquery_dataset_iam_member.access,
│   on main.tf line 17, in resource "google_bigquery_dataset_iam_member" "access":
│   17: resource "google_bigquery_dataset_iam_member" "access" {
```

---

Do those reproduction steps sound correct, and is the fix you're doing mainly to return to that correct error message? If that's the case I can give some feedback on how to write the test to achieve this

wj-chen:
> I've been taking a look at this and using the config you've provided above and want to double check I understand the problem correctly. Here's something I've been trying:
> 
> **Step 1 - Setup** Apply works fine
> 
> ```hcl
> resource "google_bigquery_dataset" "dataset" {
>   dataset_id = "tf_test_my_dataset"
> }
> 
> resource "google_service_account" "sa" {
>   account_id = "tf-test-my-sa"
> }
> 
> resource "google_bigquery_dataset_iam_member" "access" {
>   dataset_id = google_bigquery_dataset.dataset.dataset_id
>   role       = "roles/viewer"
>   member     = "serviceAccount:${google_service_account.sa.email}"
> }
> ```
> 
> **Step 2 - prepare for deleting the service account by hardcoding the `member` value** This doesn't cause any diffs/is a no-op
> 
> ```diff
> resource "google_bigquery_dataset" "dataset" {
>   dataset_id = "tf_test_my_dataset"
> }
> 
> resource "google_service_account" "sa" {
>   account_id = "tf-test-my-sa"
> }
> 
> resource "google_bigquery_dataset_iam_member" "access" {
>   dataset_id = google_bigquery_dataset.dataset.dataset_id
>   role       = "roles/viewer"
> -  member     = "serviceAccount:${google_service_account.sa.email}"
> +  member     = "serviceAccount:tf-test-my-sa@<PROJECT_ID>.iam.gserviceaccount.com"
> }
> ```
> 
> **Step 3 - Delete the service account** Apply works fine, and the plan only includes removing the google_service_account resource
> 
> ```hcl
> resource "google_bigquery_dataset" "dataset" {
>   dataset_id = "tf_test_my_dataset"
> }
> 
> resource "google_bigquery_dataset_iam_member" "access" {
>   dataset_id = google_bigquery_dataset.dataset.dataset_id
>   role       = "roles/viewer"
>   member     = "serviceAccount:tf-test-my-sa@PROEJCT-ID.iam.gserviceaccount.com"
> }
> ```
> 
> **Step 4 - run `terraform plan`** The plan indicates that Terraform cannot see the google_bigquery_dataset_iam_member's member in the IAM binding anymore (as the entry includes `deleted` and the `uuid` now) so Terraform believes this resource needs to be recreated.
> 
> ```
>   # google_bigquery_dataset_iam_member.access will be created
>   + resource "google_bigquery_dataset_iam_member" "access" {
>       + dataset_id = "tf_test_my_dataset"
>       + etag       = (known after apply)
>       + id         = (known after apply)
>       + member     = "serviceAccount:tf-test-my-sa@<PROJECT_ID>.iam.gserviceaccount.com"
>       + project    = (known after apply)
>       + role       = "roles/viewer"
>     }
> ```
> 
> **Step 5 - run `terraform apply`** The apply fails with an error, but which error happens depends on the provider version in use.
> 
> If I use the latest Google provider release (v6.6.0) then this error is returned that mentions unknown prefix types, which I believe is the regression you are fixing:
> 
> ```
> │ Error: Error applying IAM policy for Bigquery Dataset <PROJECT_ID>/tf_test_sarah:
> Error creating DatasetAccess: googleapi: Error 400:
> IAM setPolicy failed for Dataset <PROJECT_ID>:tf_test_sarah:
> The member deleted:serviceaccount:tf-test-oob-e0ybkvrluu@<PROJECT_ID>.iam.gserviceaccount.com?uid=112611438796804060811 is of an unknown type. Please set a valid type prefix for the member., invalid
> │ 
> │   with google_bigquery_dataset_iam_member.deleted,
> │   on main.tf line 15, in resource "google_bigquery_dataset_iam_member" "deleted":
> │   15: resource "google_bigquery_dataset_iam_member" "deleted" {
> │ 
> ```
> 
> If I change which version of the Google provider I use in step 6 I see a different error. I see the error below if I use:
> 
> * The code from your PR
> * The provider version that predates [[#16607] Fix: support iamMember (i.e. for WIF ids) in big query datasets #9948](https://github.com/GoogleCloudPlatform/magic-modules/pull/9948), which is v5.17.0
> 
> ```
> │ Error: Error applying IAM policy for Bigquery Dataset <PROJECT_ID>/tf_test_my_dataset:
> Error creating DatasetAccess: googleapi: Error 400:
> IAM setPolicy failed for Dataset <PROJECT_ID>:tf_test_my_dataset:
> Service account tf-test-my-sa@<PROJECT_ID>.iam.gserviceaccount.com does not exist., invalid
> │ 
> │   with google_bigquery_dataset_iam_member.access,
> │   on main.tf line 17, in resource "google_bigquery_dataset_iam_member" "access":
> │   17: resource "google_bigquery_dataset_iam_member" "access" {
> ```
> 
> Do those reproduction steps sound correct, and is the fix you're doing mainly to return to that correct error message? If that's the case I can give some feedback on how to write the test to achieve this

Much appreciated your thorough testing here!

Yes this fix is to get rid of the "invalid prefix" error and return to the old state. I was previously trying in Step 4 to instead also update the member reference to a new service account, that way instead of the error, the validation would be done to make sure the new service account could be added correctly without the "invalid prefix" error. Looks like my previous attempt missed Step 2. If you think it's easier to understand by validating the error I could do `ExpectError` instead of `Check`.

wj-chen:
Somehow my Step 3 is showing a non-empty plan:
```
2024-10-11T19:01:29.091Z [ERROR] sdk.helper_resource: Unexpected error: test_terraform_path=/usr/local/google/home/wjchen/.terraform/bin/terraform test_working_directory=/tmp/plugintest1401489590 test_step_number=3
  error=
  | After applying this test step and performing a `terraform refresh`, the plan was not empty.
  | stdout
  | 
  | 
  |   + create
  | 
  | 
  |   # google_bigquery_dataset_iam_member.access will be created
  |   + resource "google_bigquery_dataset_iam_member" "access" {
  |       + dataset_id = "tf_test_xry4gi9iwz"
  |       + etag       = (known after apply)
  |       + id         = (known after apply)
  |       + member     = "serviceAccount:tf-test-l1wbzhry1w@wjchen-osconfig-test.iam.gserviceaccount.com"
  |       + project    = (known after apply)
  |       + role       = "roles/viewer"
  |     }
  | 
  | Plan: 1 to add, 0 to change, 0 to destroy.
```

I haven't looked closely but have added my test code to this PR for you to take a look.

SarahFrench:
> Somehow my Step 3 is showing a non-empty plan

This makes sense - the steps I described above was for using Terraform manually, so the non empty plan doesn't show up until you explicitly run terraform plan again (my step 4). In the test framework it always checks if a step resulting in a non empty plan. I'd suggest putting a `ExpectNonEmptyPlan: true` on that step!

> If you think it's easier to understand by validating the error I could do ExpectError instead of Check.

I think that's a good way to simplify the test, seems like that's the most direct way to test that you've addressed the regression

wj-chen:
> > Somehow my Step 3 is showing a non-empty plan
> 
> This makes sense - the steps I described above was for using Terraform manually, so the non empty plan doesn't show up until you explicitly run terraform plan again (my step 4). In the test framework it always checks if a step resulting in a non empty plan after it finishes applying that step. I'd suggest putting a `ExpectNonEmptyPlan: true` on that step!
> 
> > If you think it's easier to understand by validating the error I could do ExpectError instead of Check.
> 
> I think that's a good way to simplify the test, seems like that's the most direct way to test that you've addressed the regression

Got it, thank you! The latest commit has a version of the test that works. Please take a look to see if it's sufficient to cover this code change.

One part I didn't understand was why I couldn't specify the `ExpectError` in Step 3 (it would say expect an error but was none), since the apply for that step should result in an error if the IAM member update happens after the service account deletion, so I'm guessing Terraform tries to update IAM member first before deleting the service account in practice?

SarahFrench:
Woo, test passed!


>One part I didn't understand was why I couldn't specify the ExpectError in Step 3 (it would say expect an error but was none), since the apply for that step should result in an error if the IAM member update happens after the service account deletion, so I'm guessing Terraform tries to update IAM member first before deleting the service account in practice?

I believe the reason is that, once the IAM member resource is updated so the email address is hardcoded, Terraform no longer knows that there is a link between the service account resource and the IAM resource. I dont know whether Terraform first identifies that the IAM member no longer exists during the apply that deletes the service account or in the next Terraform command, but running Terraform with TF_LOG=DEBUG would show that info!

SarahFrench (approved):

wj-chen:
> Woo, test passed!
> 
> > One part I didn't understand was why I couldn't specify the ExpectError in Step 3 (it would say expect an error but was none), since the apply for that step should result in an error if the IAM member update happens after the service account deletion, so I'm guessing Terraform tries to update IAM member first before deleting the service account in practice?
> 
> I believe the reason is that, once the IAM member resource is updated so the email address is hardcoded, Terraform no longer knows that there is a link between the service account resource and the IAM resource. I dont know whether Terraform first identifies that the IAM member no longer exists during the apply that deletes the service account or in the next Terraform command, but running Terraform with TF_LOG=DEBUG would show that info!

I ran the test with `TF_LOG=DEBUG`, and according to the API logs, the only API request to modify IAM member after the service account deletion happened in Step 4, so seems like although there was a non-empty plan in Step 3, it was never applied in practice.

Moreover, many thanks to your help with putting together this test case!

PR #11900, opened 2024-10-02
Fix whitespace diffs on logging_sink exclusion filters

lyoung-confluent:
The top-level `filter` resource already has `OptionalSurroundingSpacesSuppress`, but the filter on `exclusions` does not causing unnecessary diffs when there only difference is leading/trailing whitespace, common because of the use of HEREDOC on this attribute.

```release-note:bug
logging: fixed the whitespace permadiff on `exclusions.filter` field in `google_logging_billing_account_sink`, `google_logging_folder_sink`, `google_logging_organization_sink` and `google_logging_project_sink` resources
```


zli82016 (approved):
LGTM. I added the release note for this PR. Please remember to add the release note for PRs in the future.

PR #11903, opened 2024-10-03
Refactor muxing 1 : Re-use same config to configure the SDK and PF providers, fix VCR testing

SarahFrench:
Fixes https://github.com/hashicorp/terraform-provider-google/issues/19197
Fixes https://github.com/hashicorp/terraform-provider-google/issues/18774
Fixes https://github.com/hashicorp/terraform-provider-google/issues/20280

# Summary

This PR:
- **Changes how the plugin-framework implementation of the provider is configured**
    - _Before_: Parallel implementation of configuration logic that parses user inputs and populates a `fwtransport.FrameworkProviderConfig` struct
    - _After_: The configure function obtains an already populated `transport_tpg.Config` struct from the SDK provider, with no duplication of logic to process user inputs.
- **Changes code defining the VCR system to accommodate the change above**
    - Now the PF provider uses the same Config struct as the SDK provider, both providers can use the same cached configs.
    - This fixes https://github.com/hashicorp/terraform-provider-google/issues/14158 , because a single test will use a single cached client again.
- **Makes necessary changes to the data sources currently implemented with the plugin-framework**
    - These now expect to receive a `transport_tpg.Config` struct
    - Any provider-level values are now in the old type system, and require thought about where we convert between those type systems.

# Things to pay attention to during review/testing

- What acceptance tests need to be updated now that the PF provider is configured in a SDK-like way?
- Are only changes just to the test required, or do the failures indicate user-facing changes in behaviour?
- **Tests**
    - 🔴  Acceptance tests for the provider's configuration behaviour are mostly VCR skipped. Testing this PR will require use of TeamCity!
    - Manual testing encouraged
    - We should look at known bugs linked to the muxing and confirm they're gone before merge
        - https://github.com/hashicorp/terraform-provider-google/issues/16832#issuecomment-2441432304
    - After merge, we should closely monitor nightly tests and not rely on the Release Shepherd picking up any failures linked to this PR. Newer tests will provide more signal.

---

```release-note:note
provider: refactored how the provider configuration is handled internally
```


SarahFrench:
Initial acc tests for the SDK and PF provider code. These tests aren't final, as we're waiting for some more acc tests to be added to the code base and then pulled into this PR.

GA Provider:
- SDK - [all tests](https://hashicorp.teamcity.com/buildConfiguration/TerraformProviders_GoogleCloud_GOOGLE_MMUPSTREAMTESTS_GOOGLE_PACKAGE_PROVIDER/243826?hideTestsFromDependencies=false&hideProblemsFromDependencies=false&expandBuildChangesSection=true)
- Framework - [all tests](https://hashicorp.teamcity.com/buildConfiguration/TerraformProviders_GoogleCloud_GOOGLE_MMUPSTREAMTESTS_GOOGLE_PACKAGE_FWPROVIDER/243827?hideTestsFromDependencies=false&hideProblemsFromDependencies=false&expandBuildChangesSection=true)

Beta Provider:
- SDK - [all tests](https://hashicorp.teamcity.com/buildConfiguration/TerraformProviders_GoogleCloud_GOOGLE_BETA_MMUPSTREAMTESTS_GOOGLEBETA_PACKAGE_PROVIDER/243829?mode=branches&buildTab=overview&hideTestsFromDependencies=false&hideProblemsFromDependencies=false)
- Framework - [all tests](https://hashicorp.teamcity.com/buildConfiguration/TerraformProviders_GoogleCloud_GOOGLE_BETA_MMUPSTREAMTESTS_GOOGLEBETA_PACKAGE_FWPROVIDER/243828?mode=branches&buildTab=overview&hideTestsFromDependencies=false&hideProblemsFromDependencies=false&expandBuildChangesSection=true)
- Firebase - [affected tests only](https://hashicorp.teamcity.com/buildConfiguration/TerraformProviders_GoogleCloud_GOOGLE_BETA_MMUPSTREAMTESTS_GOOGLEBETA_PACKAGE_FIREBASE/243830?mode=branches&hideTestsFromDependencies=false&hideProblemsFromDependencies=false&buildTab=overview&expandBuildChangesSection=true)

SarahFrench:
Latest commits address some of the test failures described in [this comment](https://github.com/GoogleCloudPlatform/magic-modules/pull/11903#issuecomment-2390938783).

rainshen49 (approved):
Approval for Firebase since the changes are mechanical in nature and Firebase tests are passing.

SarahFrench:
Force pushed to pull in the final acceptance tests related to provider configuration logic

SarahFrench:
- [GA SDK tests](https://hashicorp.teamcity.com/buildConfiguration/TerraformProviders_GoogleCloud_GOOGLE_MMUPSTREAMTESTS_GOOGLE_PACKAGE_PROVIDER/255598?hideTestsFromDependencies=false&hideProblemsFromDependencies=false&expandBuildChangesSection=true)
- [GA Framework tests](https://hashicorp.teamcity.com/buildConfiguration/TerraformProviders_GoogleCloud_GOOGLE_MMUPSTREAMTESTS_GOOGLE_PACKAGE_FWPROVIDER/255599?hideTestsFromDependencies=false&hideProblemsFromDependencies=false&expandBuildChangesSection=true)
- [Beta SDK tests](https://hashicorp.teamcity.com/buildConfiguration/TerraformProviders_GoogleCloud_GOOGLE_BETA_MMUPSTREAMTESTS_GOOGLEBETA_PACKAGE_PROVIDER/255600?hideTestsFromDependencies=false&hideProblemsFromDependencies=false&expandBuildChangesSection=true)
- [Beta Framework tests](https://hashicorp.teamcity.com/buildConfiguration/TerraformProviders_GoogleCloud_GOOGLE_BETA_MMUPSTREAMTESTS_GOOGLEBETA_PACKAGE_FWPROVIDER/255601?hideTestsFromDependencies=false&hideProblemsFromDependencies=false&expandBuildChangesSection=true)

SarahFrench:
The last few commits are acting on my comments in https://github.com/GoogleCloudPlatform/magic-modules/pull/11903#issuecomment-2437457030

SarahFrench:
After the previous commits' changes to tests to address the discussion in https://github.com/GoogleCloudPlatform/magic-modules/pull/11903#issuecomment-2437457030 I've re-run acceptance tests in TeamCity:

SDK tests
- [GA](https://github.com/GoogleCloudPlatform/magic-modules/pull/11903#issuecomment-2437457030) 18/18 pass
- [Beta](https://hashicorp.teamcity.com/buildConfiguration/TerraformProviders_GoogleCloud_GOOGLE_BETA_MMUPSTREAMTESTS_GOOGLEBETA_PACKAGE_PROVIDER/257288?focusLine=NaN&hideProblemsFromDependencies=false&hideTestsFromDependencies=false&expandBuildChangesSection=true&expandBuildDeploymentsSection=false) 18/18 pass

Framework tests
- [GA](https://hashicorp.teamcity.com/buildConfiguration/TerraformProviders_GoogleCloud_GOOGLE_MMUPSTREAMTESTS_GOOGLE_PACKAGE_FWPROVIDER/257289?focusLine=NaN&hideTestsFromDependencies=false&hideProblemsFromDependencies=false&expandBuildChangesSection=true&expandBuildDeploymentsSection=false) 17/17 pass
- [Beta](https://hashicorp.teamcity.com/buildConfiguration/TerraformProviders_GoogleCloud_GOOGLE_BETA_MMUPSTREAMTESTS_GOOGLEBETA_PACKAGE_FWPROVIDER/257290?expandBuildDeploymentsSection=false&hideTestsFromDependencies=false&expandBuildTestsSection=true&hideProblemsFromDependencies=false&expandBuildChangesSection=true) 16/17 pass
    - The failure in TestAccFwProvider_user_project_override looks like a timing issue when enabling a service API in the test, so is something that could be investigated separately to _this_ PR.

SarahFrench:
Also, today I did some manual testing to confirm that this PR fixes some authentication bugs that were introduced by muxing: https://github.com/hashicorp/terraform-provider-google/issues/16832#issuecomment-2441432304

That comment has information about how to reproduce the bug and how to use my fork of the TPG repo to test how this PR's changes fix the issue.

c2thorn:
Wanting to follow up on https://github.com/GoogleCloudPlatform/magic-modules/pull/11903#issuecomment-2437457030

For `TestAccSdkProvider_request_timeout /when_request_timeout_is_set_to_an_empty_string_in_the_config_the_value_fails_validation,_as_it_is_not_a_duration` it looks like we've moved forward with solution 2 of keeping the same level of validation: [`aca5d33` (#11903)](https://github.com/GoogleCloudPlatform/magic-modules/pull/11903/commits/aca5d333d80b76d263426269c6855d223a880182) correct? I think that's the most straightforward approach as it doesn't make the provider configuration more or less restrictive, but I wanted to confirm.

The rest of the test cases from this seem to be handled as described in the comment

mmv1/third_party/terraform/services/resourcemanager/data_source_google_client_config.go
```diff
@@ -38,14 +38,14 @@ type GoogleClientConfigModel struct {
...
-func (m *GoogleClientConfigModel) GetLocationDescription(providerConfig *fwtransport.FrameworkProviderConfig) fwresource.LocationDescription {
+func (m *GoogleClientConfigModel) GetLocationDescription(providerConfig *transport_tpg.Config) fwresource.LocationDescription {
 	return fwresource.LocationDescription{
 		RegionSchemaField: types.StringValue("region"),
 		ZoneSchemaField:   types.StringValue("zone"),
 		ResourceRegion:    m.Region,
 		ResourceZone:      m.Zone,
-		ProviderRegion:    providerConfig.Region,
-		ProviderZone:      providerConfig.Zone,
+		ProviderRegion:    types.StringValue(providerConfig.Region),
```
c2thorn, line 47: Adding `types.StringValue()` is necessary because we're now directly accessing the SDK value instead of a pre-converted framework value?

Would we get a build error without adding `types.StringValue()` here? Is that how we know we've wrapped all of the appropriate fields that would need this?

c2thorn (commented):
Think I have a good high-level understanding of the changes made here, but I'll need some time to grab this locally, manually test, and then go through the previous muxing issues to confirm that they are actually gone.

Sorry this will probably have to pause a bit longer while I'm out next week but I'll leave some more comments tomorrow and catch up on the conversation when I'm back.

c2thorn:
side note: does anyone else have to run `make clean-provider` to actually get the `mmv1/third_party/terraform/acctest/framework_test_utils.go.tmpl` removal to show when building locally? I don't remember if Ruby mmv1 needed to do that for handwritten removals.

NickElliot:
> side note: does anyone else have to run `make clean-provider` to actually get the `mmv1/third_party/terraform/acctest/framework_test_utils.go.tmpl` removal to show when building locally? I don't remember if Ruby mmv1 needed to do that for handwritten removals.

Same here testing now, I'm not sure if it's a new issue, but isn't necessarily a problem as long as we add it to docs (as full handwritten file removals are relatively uncommon)

mmv1/third_party/terraform/fwprovider/framework_provider_access_token_test.go.tmpl
```diff
@@ -191,7 +191,7 @@ credentials = "%s"
 				Check: resource.ComposeTestCheckFunc(
 					resource.TestCheckResourceAttr("data.google_provider_config_plugin_framework.default", "access_token", accessToken),
 					// not set as ENV not used
-					resource.TestCheckNoResourceAttr("data.google_provider_config_plugin_framework.default", "credentials"),
+					resource.TestCheckResourceAttr("data.google_provider_config_plugin_framework.default", "credentials", ""),
```
c2thorn, line 194: Going from not present to "" doesn't impact users here because google_provider_config_plugin_framework isn't user-facing at all right?

Is the change in behavior coming from this addition: https://github.com/GoogleCloudPlatform/magic-modules/pull/11903/files#diff-faca97bd3809d437d4db599612e3d9eceae4102e172e3016097d0cfec117689aR206 ?

Can you help me understand fully why it was important enough to verify that it had no value before in a test, but we actually want to ensure it has an empty value now? Is it less of a "this is how the value should be" and more of a "this is how it is represented now, let's test it to make sure it doesn't change accidentally later"?

c2thorn (commented):
successfully ran several affected tests locally and feeling pretty good about the changes so far. Still want to recreate the issues like described in https://github.com/hashicorp/terraform-provider-google/issues/16832#issuecomment-2441432304 but just as a double-check

Just looking ahead at the timing after I get back - It would be nice to get this in by the cut 11/12 cut date so that it releases 11/18 and not during the 11/25 US holiday week (if there will even be a release that week). What do you all think?

mmv1/third_party/terraform/fwprovider/framework_provider_access_token_test.go.tmpl
SarahFrench: > Going from not present to "" doesn't impact users here because google_provider_config_plugin_framework isn't user-facing at all right?

That's right, google_provider_config_plugin_framework is not user facing so we don't need to worry about that data source specifically. This data source just exposes that 'meta' Config struct so the SDK-like behaviour is fully exposed here.

**I'll look more into `google_client_config` though, as this data source exposes the provider's config to users. It is analogous to `google_provider_config_*` data sources but is user-facing. As google_client_config is PF-implemented it'll be effectively like google_provider_config_plugin_framework.**

Also, I don't believe we will expose users to this SDK-like behaviour post muxing fixes because:

1.  We stop empty strings being passed from the PF provider to PF-implemented data sources
    a. For example currently there are helper functions like [GetProjectFramework](https://github.com/GoogleCloudPlatform/magic-modules/blob/1e3b6fed59b9905d4ea14e55c9e76b0aaeb7a074/mmv1/third_party/terraform/fwresource/field_helpers.go#L23-L34) that pass provider-level default values to resources but are written to explicitly treat provider-level arguments as unset if the value is `""`.
2. Datasource/resource-level fields that can use provider-level defaults cannot be empty/unset in valid resources
    a. Provider-level defaults are used for fields that _have_ to be set for something to be valid. They're Optional in the schema but are required fields for the API.
    b. I don't believe we'll encounter similar "Going from not present to "" " changes for pre-existing resources in a users config as a valid pre-existing config would need to have provider defaults that aren't `""`
    c. `google_client_config` is the exception to this
3. The existing migrated firebase data sources : uses GetProjectFramework and requires a value for `project` so are covered by point above

SarahFrench: > Can you help me understand fully why it was important enough to verify that it had no value before in a test, but we actually want to ensure it has an empty value now? Is it less of a "this is how the value should be" and more of a "this is how it is represented now, let's test it to make sure it doesn't change accidentally later"?

My intention with those checks was to show how the provider behaves now, and ensure it was the same between SDK and PF. I wanted to explore how conflicting fields were handled when one is set via config and the other is set via ENVs.

Currently [credentials and access_token are set as conflicting in the provider schema](https://github.com/GoogleCloudPlatform/magic-modules/blob/8a6f281bad55b80c3e9f6d739729cb5d78646ab3/mmv1/third_party/terraform/provider/provider.go.tmpl#L33-L45), which restricts what user inputs can be supplied via config. However as the provider configuration logic pulls in values from ENVs there is a chance of those conflicting fields both being set in a way that the schema-level validation cannot detect and prevent.

In these tests I wanted to define the behaviour that if one of the conflicting fields was set in the config and the other was provided via ENV:

1. No user-facing validation logic was triggered
2. The field set in the config is the one that's used, and the conflicting field is ignored.
    a. Specifically, ignored to the point of ENVs not being used even if they're present.

Does that help?

mmv1/third_party/terraform/fwprovider/data_source_provider_config_plugin_framework.go
```diff
@@ -202,23 +203,52 @@ func (d *GoogleProviderConfigPluginFrameworkDataSource) Read(ctx context.Context
...
-	data.Zone = d.providerConfig.Zone
-	data.UniverseDomain = d.providerConfig.UniverseDomain
-	data.Scopes = d.providerConfig.Scopes
-	data.UserProjectOverride = d.providerConfig.UserProjectOverride
-	data.RequestReason = d.providerConfig.RequestReason
-	data.RequestTimeout = d.providerConfig.RequestTimeout
-	data.DefaultLabels = d.providerConfig.DefaultLabels
-	data.AddTerraformAttributionLabel = d.providerConfig.AddTerraformAttributionLabel
-	data.TerraformAttributionLabelAdditionStrategy = d.providerConfig.TerraformAttributionLabelAdditionStrategy
+	delegateAttrs := make([]attr.Value, len(d.providerConfig.ImpersonateServiceAccountDelegates))
```
NickElliot, line 210: to double check, this extended logic is from the need to apply the string type assertions to every item within a list (as well as the list type on the list itself)?

seperately I think it would be good to have a comment somewhere around here summarizing the need for these assertions and the resulting mix of SDK behavior in this data source (as was mentioned [here](https://github.com/GoogleCloudPlatform/magic-modules/pull/11903#discussion_r1829448872), albeit not user-facing

mmv1/third_party/terraform/fwprovider/framework_provider.go.tmpl
```diff
@@ -259,16 +268,17 @@ func (p *FrameworkProvider) Configure(ctx context.Context, req provider.Configur
...
 
-    // Configuration values are now available.
-    p.LoadAndValidateFramework(ctx, &data, req.TerraformVersion, &resp.Diagnostics, p.Version)
-    if resp.Diagnostics.HasError() {
-        return
-    }
+	// Configuration values are now available.
+    // However we don't use them; for the plugin-framework implementation of the provider
+	// we take the configuration values from the SDK implementation of the provider. This avoids duplicated logic and inconsistencies in implementation.
+	// The trade off is that we don't benefit from the new type system that differentiates Null and Unknown values, which is especially useful for strings.
```
NickElliot, line 274: I think it would also be helpful to reference that this trade-off has some workarounds downstream on a resource level via the field helper functions like [GetProjectFramework](https://github.com/GoogleCloudPlatform/magic-modules/blob/1e3b6fed59b9905d4ea14e55c9e76b0aaeb7a074/mmv1/third_party/terraform/fwresource/field_helpers.go#L23-L34) that prevent passing empty provider values to resource schemas.

NickElliot (commented):
Sorry for the delay on my end, I think I'm understanding the majority of what the changes are doing here now and how/why things are happening in it

I commented in a couple particular areas but perhaps in a more central spot it would be useful to add some comments to explain the main quirks with the muxed framework provider maintaining use of the SDK config, and briefly explain how these have been mitigated/worked through without as much hopping between files, and PR/issue comments.

mmv1/third_party/terraform/fwprovider/data_source_provider_config_plugin_framework.go
SarahFrench: Here's a preview of that: https://github.com/SarahFrench/magic-modules/pull/10

mmv1/third_party/terraform/services/resourcemanager/data_source_google_client_config.go
SarahFrench: It's only necessary here because [the struct expects all inputs to be in the new type system](https://github.com/GoogleCloudPlatform/magic-modules/blob/27812e087aaf4250c076b5d572b3934c1a013e2e/mmv1/third_party/terraform/fwresource/framework_location.go#L16-L30). This is arbitrary - we could change `type LocationDescription struct` to expect type `string` values for ProviderRegion and ProviderZone and update the logic inside to check if the values `!= ""` etc.

This is an example of where we decide where and how the two type systems should meet. Should we enforce that all inputs to the `type LocationDescriber interface` logic are in the new type system and the responsibility for knowing about both type systems is with the calling code? Or do we write `type LocationDescriber interface` logic to handle both type systems?

mmv1/third_party/terraform/fwprovider/data_source_provider_config_plugin_framework.go
SarahFrench: Done in https://github.com/GoogleCloudPlatform/magic-modules/pull/11903/commits/2ef48bbd94b82ccd1bdb2b5da3ac9c5431fbbeaf

mmv1/third_party/terraform/fwprovider/framework_provider.go.tmpl
SarahFrench: Thanks I'll look into updating this and work out an appropriate centralised place for this knowledge to be recorded

SarahFrench: I'll update the comment here, but I'm unsure where an appropriate central spot would be to record this knowledge. Please let me know if you have ideas. A Google doc maybe?

mmv1/third_party/terraform/services/resourcemanager/data_source_google_client_config.go
c2thorn: Got it, since that's the expected type system in the struct itself, the provider wouldn't build if a `string` value was attempted for those fields. 

> Should we enforce that all inputs to the type LocationDescriber interface logic are in the new type system and the responsibility for knowing about both type systems is with the calling code? Or do we write type LocationDescriber interface logic to handle both type systems?

I like keeping enforcing all inputs to be the new type system so that the writer has to be aware of the new type system, and it's not hidden in the interface logic. At least for now.

NickElliot (approved):
LGTM, I think the comment updates achieve what I was looking for

c2thorn (approved):
I think we should be good to merge this. LGTM, thanks @SarahFrench

SarahFrench:
Thank you both! I'm just going to do one last test run before merging, partly so that there's an autogen pr branch for this PR for reference in future (the old one was deleted recently).

PR #11906, opened 2024-10-03
Make dataform tests use bootstrapped KMS keys, use non-authoritative IAM resources

SarahFrench:
Fixes https://github.com/hashicorp/terraform-provider-google/issues/19734

```release-note:none

```

### DIFF
--- a/mmv1/products/dataform/Repository.yaml
+++ b/mmv1/products/dataform/Repository.yaml
@@ -48,6 +48,9 @@ examples:
       secret_name: 'my-secret'
       key_ring_name: 'example-key-ring'
       crypto_key_name: 'example-crypto-key-name'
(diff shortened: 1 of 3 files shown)


SarahFrench:
Looks like a problem with the IAM tests is they're run in parallel.

SarahFrench:
I've tried changing the tests to use non-default service accounts for the google_dataform_repository resources, but despite there being [a field to set a service account](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/dataform_repository#service_account) it doesn't seem like that's the service account that needs the permissions on KMS Key resources at time of creating the repository.

SarahFrench:
Closing as this is more involved than I thought

PR #11907, opened 2024-10-03
Allow sending `false` for `enabled` field under IAP message for resource google_compute_backend_service and resource google_compute_region_backend_service 

arnabadg-google:
fixes https://github.com/hashicorp/terraform-provider-google/issues/19596

```release-note:bug
compute: fixed an issue in `google_compute_backend_service` and `google_compute_region_backend_service` to allow sending `false` for `iap.enabled`
```


arnabadg-google:
> Can you add a test that uses this? It would show up if you have a test where you set enabled = true and then have another step that sets enabled = false. Switching from true --> false forces sending the false value for everything to pass

Added update test

PR #11917, opened 2024-10-03
Stop Dataform tests failing on PRs

SarahFrench:
Trying to stop these tests failing on PRs 🤷 

```release-note:none

```

### DIFF
--- a/mmv1/products/dataform/Repository.yaml
+++ b/mmv1/products/dataform/Repository.yaml
@@ -70,6 +70,7 @@ examples:
     exclude_test: true
   - name: 'dataform_repository_with_cloudsource_repo_and_ssh'
     primary_resource_id: 'dataform_repository'
+    primary_resource_name: 'fmt.Sprintf("tf_test_dataform_repository%s", context["random_suffix"])'
     min_version: 'beta'
     vars:
       git_repository_name: 'my/repository'
--- a/mmv1/products/dataform/Repository.yaml
+++ b/mmv1/products/dataform/Repository.yaml
@@ -80,8 +80,6 @@ examples:
     # See : https://github.com/hashicorp/terraform-provider-google/issues/17335
     # See : https://issuetracker.google.com/issues/287850319
     exclude_docs: true
-    # These tests create a new KMS key ring and key with each run, which cannot be deleted.
-    exclude_test: true
 parameters:
   - name: 'region'
     type: String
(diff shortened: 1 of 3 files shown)


trodge (approved):

mmv1/products/dataform/Repository.yaml
```diff
@@ -78,6 +79,7 @@ examples:
     # See : https://github.com/hashicorp/terraform-provider-google/issues/17335
     # See : https://issuetracker.google.com/issues/287850319
     exclude_docs: true
+    exclude_test: true
```
trodge, line 82:
```suggestion
    # These tests create a new KMS key ring and key with each run, which cannot be deleted.
    exclude_test: true
```

mmv1/products/dataform/Repository.yaml
```diff
@@ -66,6 +66,7 @@ examples:
...
```
trodge, line 69:
```suggestion
    # These tests create a new KMS key ring and key with each run, which cannot be deleted.
    exclude_test: true
```

mmv1/products/dataform/Repository.yaml
```diff
@@ -78,6 +80,8 @@ examples:
     # See : https://github.com/hashicorp/terraform-provider-google/issues/17335
     # See : https://issuetracker.google.com/issues/287850319
     exclude_docs: true
+    # These tests create a new KMS key ring and key with each run, which cannot be deleted.
+    exclude_test: true
```
trodge, line 84:
```suggestion
```
Can't exclude every test or the build fails, so let's just exclude the one that's creating new keys for now

SarahFrench:
I believe the remaining tests are failing because the default Dataflow CMEK key hasn't been clearer from the test project:

>        Error: Error creating Repository: googleapi: Error 403: Permission 'cloudkms.cryptoKeyVersions.useToEncrypt' denied on resource 'projects/PROJECT_ID/locations/us-central1/**keyRings/tf-test-example-key-ringifpfzqngwj/cryptoKeys/tf-test-example-crypto-key-nameifpfzqngwj**' (or it may not exist).

mmv1/products/dataform/Repository.yaml
```diff
@@ -66,8 +66,11 @@ examples:
...
   - name: 'dataform_repository_with_cloudsource_repo_and_ssh'
     primary_resource_id: 'dataform_repository'
+    primary_resource_name: 'fmt.Sprintf("tf_test_dataform_repository%s", context["random_suffix"])'
     min_version: 'beta'
     vars:
       git_repository_name: 'my/repository'
```
SarahFrench, line 76: If we skip this example generating a test we should be able to stop all the failing tests being created. However we could [check for and remove the default CMEK key](https://cloud.google.com/dataform/docs/cmek#check-default-key) in the test project(s) to try resolve the failures in tests that don't provision keys themselves.

PR #11919, opened 2024-10-04
fix: fix slice init length

cuishuang:
The intention here should be to initialize a slice with a capacity of length rather than initializing the length of this slice. 

The online demo: https://go.dev/play/p/q1BcVCmvidW

```release-note:none
fix: fix slice init length
```


ScottSuarez:
ooo nice catch !

ScottSuarez (approved):

PR #11924, opened 2024-10-04
Proposal: use embedded structs to minimise repetitive Configure functions on PF-implemented resources & data sources

SarahFrench:
**Release Note Template for Downstream PRs (will be copied)**

```release-note:none

```

PR #11925, opened 2024-10-04
Proposal: use embedded structs to add GetProject methods to PF-implemented resources & data sources

SarahFrench:
Blocked by muxing fixes PR (this PR contains that PR's commits)

# Background

Dating from the time of the original muxing in 2023 there is [a function called **GetProjectFramework**](https://github.com/hashicorp/terraform-provider-google-beta/blob/c56e8bfea58d081d0e236e9630f5b2baa4820e80/google-beta/fwresource/field_helpers.go#L21) that's used to determine whether to take the `project` value from a resource argument or from the provider default.

```go
// GetProject reads the "project" field from the given resource and falls
// back to the provider's value if not given. If the provider's value is not
// given, an error is returned.
func GetProjectFramework(rVal, pVal types.String, diags *diag.Diagnostics) types.String {
	return getProjectFromFrameworkSchema("project", rVal, pVal, diags)
}

func getProjectFromFrameworkSchema(projectSchemaField string, rVal, pVal types.String, diags *diag.Diagnostics) types.String {
	if !rVal.IsNull() && rVal.ValueString() != "" {
		return rVal
	}

	if !pVal.IsNull() && pVal.ValueString() != "" {
		return pVal
	}

	diags.AddError("required field is not set", fmt.Sprintf("%s is not set", projectSchemaField))
	return types.String{}
}
```

This was written to heavily mimic [the original, SDKv2 equivalent **GetProject**](https://github.com/hashicorp/terraform-provider-google-beta/blob/c56e8bfea58d081d0e236e9630f5b2baa4820e80/google-beta/tpgresource/utils.go#L86) function. This function needed a 'schema field' argument because the first argument was `*schema.ResourceData`, the entire data for a resource. The function needed instructions on what key-value pair to look for and retrieve from that data structure: 

```go
// GetProject reads the "project" field from the given resource data and falls
// back to the provider's value if not given. If the provider's value is not
// given, an error is returned.
func GetProject(d TerraformResourceData, config *transport_tpg.Config) (string, error) {
	return GetProjectFromSchema("project", d, config)
}

func GetProjectFromSchema(projectSchemaField string, d TerraformResourceData, config *transport_tpg.Config) (string, error) {
	res, ok := d.GetOk(projectSchemaField)
	if ok && projectSchemaField != "" {
		return res.(string), nil
	}
	if config.Project != "" {
		return config.Project, nil
	}
	return "", fmt.Errorf("%s: required field is not set", projectSchemaField)
}
```

# My opinion

Mimicking the original SDK code when implementing a plugin-framework version of this logic doesn't make sense, as a resource's data is no longer stored as a map[string]interface{} matching a schema which can be queried with string keys. Instead the data is read into a struct annotated with struct tags that link fields to config values. For example:

```go
type GoogleFirebaseAppleAppConfigModel struct {
	Id                 types.String `tfsdk:"id"`
	AppId              types.String `tfsdk:"app_id"`
	ConfigFilename     types.String `tfsdk:"config_filename"`
	ConfigFileContents types.String `tfsdk:"config_file_contents"`
	Project            types.String `tfsdk:"project"`
}
```

Given this, there isn't any sense passing in the string `"project"` to `getProjectFromFrameworkSchema` above. And if you remove that part of a logic you end up with a generic function that compares two arguments using the new type system and then return the first one that isn't unknown, null, or `""`.

I don't think we want a generic 'compare two values' function as we want the function to be able to provide feedback to the user via errors and warnings if no suitable value is found. A generic function cannot tell the user what value is missing.

This led me to thinking that we should implement logic specifically for retrieving the `project` value

# This PR

In this PR I propose that we add a new embeddable struct that allows adding a new method `GetProject` to resources and data sources. This method will return either the resource's project argument, the provider-default project, or an error as feedback to the user.

The calling code that defines the custom model for that resource (e.g. GoogleFirebaseAppleAppConfigModel above) has the responsibility of needing to provide the correct field from the resource model struct as an argument. This replaces the use of `"project"` in the SDK version.

Because the method is specific to `project`, we can include feedback about which field has missing information.

An open question is whether we have any resources/data sources currently where a project ID is supplied by a field called something other than `project`. If not, we can hardcode the function to raise an error about `project` being unset. If there are special resources that use other fields for project we can update the logic so that the field name is stored in the resource model struct and accessed from there. However, setting the custom project schema field name in that struct is something that would need to happen in a non-standardised (see https://github.com/GoogleCloudPlatform/magic-modules/pull/11924) Configure function.

```release-note:none

```

### DIFF
--- a/mmv1/third_party/terraform/fwresource/with_get_project.go
+++ b/mmv1/third_party/terraform/fwresource/with_get_project.go
@@ -1,0 +1,39 @@
+package fwresource
+
+import (
+	"github.com/hashicorp/terraform-plugin-framework/diag"
+	"github.com/hashicorp/terraform-plugin-framework/path"
+	"github.com/hashicorp/terraform-plugin-framework/types"
+	transport_tpg "github.com/hashicorp/terraform-provider-google/google/transport"
+)
+
+type ProjectGetter interface {
+	GetProject(types.String, *transport_tpg.Config, *diag.Diagnostics) types.String
+}
+
+type WithGetProject struct {
+}
+
+// GetProject determines whether to use a project value from the resource/data source config or the provider-default project
+func (w *WithGetProject) GetProject(resourceProject types.String, config *transport_tpg.Config, diags *diag.Diagnostics) types.String {
+	// Prevent panic if nil pointer
+	if config == nil {
+		diags.AddError("nil pointer encountered in fwresource.GetProject", "Please report this issue to the provider developers")
+		return types.String{}
+	}
+
+	if !(resourceProject.IsNull() || resourceProject.IsUnknown() || resourceProject.ValueString() == "") {
+		return resourceProject
+	}
+
+	if config.Project != "" {
+		return types.StringValue(config.Project)
+	}
+
+	diags.AddAttributeError(
+		path.Root("project"),
+		"missing required field project",
+		"this resource/data source requires a project argument set either in the resource/data block or provided as a provider-level default",
+	)
+	return types.String{}
+}
--- a/mmv1/third_party/terraform/fwresource/data_source_with_configure.go
+++ b/mmv1/third_party/terraform/fwresource/data_source_with_configure.go
@@ -1,0 +1,34 @@
+package fwresource
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/hashicorp/terraform-plugin-framework/datasource"
+	transport_tpg "github.com/hashicorp/terraform-provider-google/google/transport"
+)
+
+type dataSourceConfigurer interface {
+	Configure(context.Context, datasource.ConfigureRequest, *datasource.ConfigureResponse)
+}
+
+type DataSourceWithConfigure struct {
+	Config *transport_tpg.Config
+}
+
+func (ds *DataSourceWithConfigure) Configure(ctx context.Context, req datasource.ConfigureRequest, resp *datasource.ConfigureResponse) {
+	// Prevent panic if the provider has not been configured.
+	if req.ProviderData == nil {
+		return
+	}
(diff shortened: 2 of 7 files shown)


mmv1/third_party/terraform/fwresource/with_get_project.go
```diff
@@ -0,0 +1,39 @@
...
+
+type ProjectGetter interface {
+	GetProject(types.String, *transport_tpg.Config, *diag.Diagnostics) types.String
+}
+
+type WithGetProject struct {
+}
+
+// GetProject determines whether to use a project value from the resource/data source config or the provider-default project
+func (w *WithGetProject) GetProject(resourceProject types.String, config *transport_tpg.Config, diags *diag.Diagnostics) types.String {
```
SarahFrench, line 18: This could be refactored so that WithGetProject expects to have access to Config:

```diff
type WithGetProject struct {
+ *transport_tpg.Config
}
```

and then the GetProject method doesn't need config as an argument:

```diff
// GetProject determines whether to use a project value from the resource/data source config or the provider-default project
-func (w *WithGetProject) GetProject(resourceProject types.String, config *transport_tpg.Config, diags *diag.Diagnostics) types.String {
+func (w *WithGetProject) GetProject(resourceProject types.String, diags *diag.Diagnostics) types.String {
    
-    ...do stuff using config...
+    ...do stuff using w.Config...

}
```

though this might obscure what the method is doing to users.

SarahFrench:
Just rebased so that this PR only includes new commits, and none of the muxing fix commits

NickElliot:
Implemented via modified `getProviderDefaultFromFrameworkSchema` function

PR #11928, opened 2024-10-04
Use subnets with `privateIpGoogleAccess: true` in tests, as is required when u…

Harwayne:
Use subnets with `privateIpGoogleAccess: true` in tests, as is required when using the now default `internal_ip_only: true` (as of Dataproc 2.2).

Note that this is a Test only chang.e

```release-note:none

```


melinath (changes requested):
It looks like there's a merge conflict - could you resolve it?

melinath:
@slevenick FYI - would you have any concern about us moving forward with this PR?

slevenick:
> @slevenick FYI - would you have any concern about us moving forward with this PR?

I would just make sure tests pass after my PR from last week changed several of them. I fixed most of the failing dataproc tests, though I think there was another server-side change that is causing some to fail on actually setting up dataproc

PR #11930, opened 2024-10-04
Update forward_ssh_connectivity.private_key to be mutable

ari-hacks:
This prevents the connection profile from failing permanently (needing to be re-created) when the forward ssh connectivity private key is updated. 
Fixes https://github.com/hashicorp/terraform-provider-google/issues/18999

```release-note: enhancement
datastream: updated `private_key`to be mutable in `google_datastream_connection_profile` resource.
```

### DIFF
--- a/mmv1/third_party/terraform/services/datastream/resource_datastream_connection_profile_test.go
+++ b/mmv1/third_party/terraform/services/datastream/resource_datastream_connection_profile_test.go
@@ -19,6 +19,9 @@ func TestAccDatastreamConnectionProfile_update(t *testing.T) {
 	random_pass_1 := acctest.RandString(t, 10)
 	random_pass_2 := acctest.RandString(t, 10)
 
+	random_key_1 := acctest.RandString(t, 10)
+	random_key_2 := acctest.RandString(t, 10)
+
 	acctest.VcrTest(t, resource.TestCase{
 		PreCheck:                 func() { acctest.AccTestPreCheck(t) },
 		ProtoV5ProviderFactories: acctest.ProtoV5ProviderFactories(t),
@@ -71,6 +74,28 @@ func TestAccDatastreamConnectionProfile_update(t *testing.T) {
 			{
 				// Disable prevent_destroy
 				Config: testAccDatastreamConnectionProfile_mySQLUpdate(context, false, random_pass_2),
+			},
+			{
+				Config: testAccDatastreamConnectionProfile_SSHKey_Update(context, true, random_key_1),
+			},
+			{
+				ResourceName:            "google_datastream_connection_profile.ssh_connectivity_profile",
+				ImportState:             true,
+				ImportStateVerify:       true,
+				ImportStateVerifyIgnore: []string{"connection_profile_id", "location", "forward_ssh_connectivity.0.private_key"},
+			},
+			{
+
+				Config: testAccDatastreamConnectionProfile_SSHKey_Update(context, true, random_key_2),
+			},
+			{
+				ResourceName:            "google_datastream_connection_profile.ssh_connectivity_profile",
+				ImportState:             true,
+				ImportStateVerify:       true,
+				ImportStateVerifyIgnore: []string{"connection_profile_id", "location", "forward_ssh_connectivity.0.private_key"},
+			},
+			{
+				Config: testAccDatastreamConnectionProfile_SSHKey_Update(context, false, random_key_2),
 			},
 		},
 	})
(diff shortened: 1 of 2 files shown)


mmv1/third_party/terraform/services/datastream/resource_datastream_connection_profile_test.go
```diff
@@ -19,6 +19,9 @@ func TestAccDatastreamConnectionProfile_update(t *testing.T) {
 	random_pass_1 := acctest.RandString(t, 10)
 	random_pass_2 := acctest.RandString(t, 10)
 
+	random_key_1 := acctest.RandString(t, 10)
+	random_key_2 := acctest.RandString(t, 10)
```
ari-hacks, line 23: I'll need to update this to pass in a valid key format

mmv1/third_party/terraform/services/datastream/resource_datastream_connection_profile_test.go
```diff
@@ -72,6 +75,28 @@ func TestAccDatastreamConnectionProfile_update(t *testing.T) {
 				// Disable prevent_destroy
 				Config: testAccDatastreamConnectionProfile_mySQLUpdate(context, false, random_pass_2),
 			},
+			{
+				Config: testAccDatastreamConnectionProfile_SSHKey_Update(context, true, random_key_1),
```
roaks3, line 79: This is a little strange, because destroy gets run after all steps are complete, and with the same instance name is being used, so I think it will try to update from mysql to postgres. I would create a new test if you need to use a postgres instance.

mmv1/third_party/terraform/services/datastream/resource_datastream_connection_profile_test.go
```diff
@@ -247,3 +272,88 @@ resource "google_datastream_connection_profile" "mysql_con_profile" {
...
+    
+        deletion_protection  = "false"
+    }
+    
+    resource "google_sql_database" "db" {
+        instance = google_sql_database_instance.instance.name
+        name     = "db"
+    }
+    
+    resource "random_password" "pwd" {
```
roaks3, line 327: It looks like the test is skipped in VCR (meaning we don't run it in PR checks) because the random_password provider is used, but I don't think you need it. You can probably pass in a random value instead, and then we can make sure the tests pass in the PR.

mmv1/third_party/terraform/services/datastream/resource_datastream_connection_profile_test.go
```diff
@@ -247,3 +272,88 @@ resource "google_datastream_connection_profile" "mysql_con_profile" {
...
+        username = google_sql_user.user.name
+        password = google_sql_user.user.password
+        database = google_sql_database.db.name
+    }
+
+    forward_ssh_connectivity {
+        hostname = google_sql_database_instance.instance.public_ip_address
+        username = google_sql_user.user.name
+        port     = 22
+        private_key = "%{private_key}"
```
roaks3, line 354: I'm no real familiar with the use case, but just in case it helps, this test only needs to do enough to use the private_key. If the other config here is extraneous, it could be removed to make things more focused, and possibly faster to deploy.

roaks3 (changes requested):
I can see you've been pretty thorough with your change, but the test still needs some work. I would really suggest creating a new test instead of adding to `TestAccDatastreamConnectionProfile_update`, but I think it could still technically work.

Here's what likely needs to change:
- `acctest.SkipIfVcr(t)` would need to be removed so the test is run in the PR
- `RandSSHKeyPair()` won't work because it will be random on every run (see mmv1/third_party/terraform/acctest/test_utils.go.tmpl), so you would need to wrap it. I would suggest just using `RandString` (or hard-coded keys) unless that causes issues for some reason.

ari-hacks:
> I can see you've been pretty thorough with your change, but the test still needs some work. I would really suggest creating a new test instead of adding to `TestAccDatastreamConnectionProfile_update`, but I think it could still technically work.
> 
> Here's what likely needs to change:
> 
> * `acctest.SkipIfVcr(t)` would need to be removed so the test is run in the PR
> * `RandSSHKeyPair()` won't work because it will be random on every run (see mmv1/third_party/terraform/acctest/test_utils.go.tmpl), so you would need to wrap it. I would suggest just using `RandString` (or hard-coded keys) unless that causes issues for some reason.

Thanks @roaks3! Sorry for the delay on this. 
I see, `SkipIfVcr(t)` enables VCR so test are skipped in `TestAccDatastreamConnectionProfile_update`? 
To decouple this a bit I can create a separate test file, `RandString()` doesn't support the key format for SSH - I'll hardcode these

ari-hacks:
> I can see you've been pretty thorough with your change, but the test still needs some work. I would really suggest creating a new test instead of adding to `TestAccDatastreamConnectionProfile_update`, but I think it could still technically work.
> 
> Here's what likely needs to change:
> 
> * `acctest.SkipIfVcr(t)` would need to be removed so the test is run in the PR
> * `RandSSHKeyPair()` won't work because it will be random on every run (see mmv1/third_party/terraform/acctest/test_utils.go.tmpl), so you would need to wrap it. I would suggest just using `RandString` (or hard-coded keys) unless that causes issues for some reason.

Thanks @roaks3! Sorry for the delay on this.
Can you explain how this works a bit more, I'm not too familiar with go test/how it runs in the PR. I see all the test steps are wrapped in `VcrTest`. Does that means without `SkipIfVcr(t)` those test should run? 

To get this working without too much of a lift what if we: 
- decouple this test form the other ones with your suggestion of creating a new test - I believe this would need a separate file? `resource_datastream_connection_profile_frwd_ssh_test.go`
- RandString() doesn't support the key format for SSH so I can hard code a random private/public key pair

roaks3 (commented):
> I see all the test steps are wrapped in VcrTest. Does that means without SkipIfVcr(t) those test should run?

Yea, remove that line and the test will run in a PR check, which helps us ensure it passes before we merge.

> decouple this test from the other ones with your suggestion of creating a new test - I believe this would need a separate file? resource_datastream_connection_profile_frwd_ssh_test.go

I would actually suggest using the existing file, but copying an existing `testAcc...` function, and then modifying as needed.

> RandString() doesn't support the key format for SSH so I can hard code a random private/public key pair

Yep, hardcoding seems reasonable to me here.

PR #11936, opened 2024-10-07
Add example of a Workload Identity Pool configuration for GitHub Actions

b-r-o-c-k:
A security check was recently enabled in GCP which prevents [insecure workload identity pools](https://cloud.google.com/iam/docs/best-practices-for-using-workload-identity-federation#multi-tenant-attribute-conditions) from being created for certain providers. This check requires an `attribute_condition` to specified when creating an OIDC provider for GitHub Actions. [GCP documentation](https://cloud.google.com/iam/docs/workload-identity-federation-with-deployment-pipelines#github-actions_2) already explains how to configure a provider for GitHub Actions, but it would also be helpful to provide an example for Terraform users.

```release-note:none

```

### DIFF
--- a/mmv1/products/iambeta/WorkloadIdentityPoolProvider.yaml
+++ b/mmv1/products/iambeta/WorkloadIdentityPoolProvider.yaml
@@ -59,6 +59,11 @@ examples:
     vars:
       workload_identity_pool_id: 'example-pool'
       workload_identity_pool_provider_id: 'example-prvdr'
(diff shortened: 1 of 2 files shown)


zli82016 (changes requested):
The new example needs to be added to https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/products/iambeta/WorkloadIdentityPoolProvider.yaml#L51

PR #11938, opened 2024-10-07
Add major release number to `dev` ua

rileykarson:
Based on the acc replacement in https://github.com/hashicorp/terraform-provider-google/blob/270d18f71bc6dd721d14235d4dd30693c483d033/GNUmakefile#L17-L18 the release system should be resilient, but we should check it (I think you'll have to?)

A large number of `4.X` releases were misattributed to `dev`, which makes it difficult to tell private build and Registry provider useragents apart. `hashicorp/terraform` uses a VERSION file updated with every release (https://github.com/hashicorp/terraform/blob/main/version/VERSION) while other providers use `dev`, like ourselves. Using just the major version should give enough fidelity to classify private builds while also not requiring too frequent of updates.

```release-note:none
```

### DIFF
--- a/mmv1/third_party/terraform/version/version.go
+++ b/mmv1/third_party/terraform/version/version.go
@@ -2,5 +2,5 @@ package version
 
 var (
 	// ProviderVersion is set during the release process to the release version of the binary
-	ProviderVersion = "dev"
+	ProviderVersion = "dev6"
 )


SarahFrench (approved):
This looks like a good way to get better signal if the release process fails 👍 

---

>the release system should be resilient, but we should check it (I think you'll have to?)

The testing I used when [I fixed the release process](https://github.com/GoogleCloudPlatform/magic-modules/pull/6929) was like this, and can be done by anyone:

1. Checkout a release branch in the TPG/TPGB repo : `git checkout release-6.6.0`
    a. This is necessary to control which version of the provider we test releasing; that's what is released, and release branches are where our release tags are.
2. Run this command in the repo: `goreleaser build --single-target --debug --rm-dist --skip-validate`
    a. This just builds a binary on your machine, so no worries about impacting anything user-facing.
    b. A binary is made in a new `dist` directory in the repo
    c. The debug output shows how ProviderVersion is overwritten and what the resulting command is, e.g:

```
go build -trimpath -ldflags=-s -w -X github.com/hashicorp/terraform-provider-google/version.ProviderVersion=6.6.0 -o /Users/sarahfrench/go/src/github.com/hashicorp/terraform-provider-google/dist/terraform-provider-google_darwin_arm64/terraform-provider-google_v6.6.0_x5 .
```

3. You can then move that binary to ~/go/bin and use [development overrides](https://developer.hashicorp.com/terraform/cli/config/config-file#development-overrides-for-provider-developers) to make Terraform use that binary as the Google provider. If you run a terraform command with that override and TF_LOG=DEBUG you can see the UserAgent value there.

---

We could make some automated tests to check on this:
- In CI, checkout the latest release branch in the TPG/TPGB repo and run the `goreleaser` command above
- Run a very simple acceptance test with debug logging enabled, write that debug log output to file
- Grep the file for the User-Agent value and assert that the UA matches the checked out release branch

SarahFrench:
👆 I made an issue to capture the idea about automation, which we can discuss in triage

hao-nan-li:
re-run as these tests look to be passing on nightly builds

rileykarson:
Given https://github.com/GoogleCloudPlatform/magic-modules/pull/11917 and https://github.com/GoogleCloudPlatform/magic-modules/pull/11923 will resolve the failing VCR tests, considering them non-blocking.

PR #11940, opened 2024-10-08
Add logging_settings and speech_settings to dialogflow_cx_agent/flow

averbuks:
Add `advanced_settings.logging_settings` and `advanced_settings.speech_settings` to `dialogflowcx_agent` and `dialogflowcx_flow` resources

Fixes https://github.com/hashicorp/terraform-provider-google/issues/15980

-->

```release-note:enhancement
dialogflowcx: added `advanced_settings.logging_settings` and `advanced_settings.speech_settings` to `google_dialogflow_cx_agent` and `google_dialogflow_cx_flow`
```

### DIFF
--- a/mmv1/third_party/terraform/services/dialogflowcx/resource_dialogflowcx_flow_test.go
+++ b/mmv1/third_party/terraform/services/dialogflowcx/resource_dialogflowcx_flow_test.go
@@ -25,17 +25,19 @@ func TestAccDialogflowCXFlow_update(t *testing.T) {
 				Config: testAccDialogflowCXFlow_basic(context),
 			},
 			{
-				ResourceName:      "google_dialogflow_cx_flow.my_flow",
-				ImportState:       true,
-				ImportStateVerify: true,
+				ResourceName:            "google_dialogflow_cx_flow.my_flow",
+				ImportState:             true,
+				ImportStateVerify:       true,
+				ImportStateVerifyIgnore: []string{"advanced_settings.0.logging_settings"},
 			},
 			{
 				Config: testAccDialogflowCXFlow_full(context),
 			},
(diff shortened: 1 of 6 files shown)


mmv1/third_party/terraform/services/dialogflowcx/resource_dialogflowcx_flow_test.go
```diff
@@ -341,11 +343,26 @@ func testAccDialogflowCXFlow_full(context map[string]interface{}) string {
...
+          count : "3"
+        }
+      }
       dtmf_settings {
         enabled      = true
         max_digits   = 1
         finish_digit = "#"
       }
+      logging_settings {
+        enable_stackdriver_logging     = true
```
slevenick, line 362: Can you set some of these to `false` in a separate step? Sometimes we see issues setting booleans to false values

slevenick: Or is that impossible to verify based on the ignore_read functionality? Have you tested setting the false value and seeing that it correctly sets it in the API?

averbuks: Hi @slevenick, I tested setting all values to true/false and it all works. The only issue is that one of the  fields `enableInteractionLogging ` automatically set to true if you completely remove `logging_settings {}` block which triggers permadiff. I have internal bug open and will keep an eye on it.
Indeed given that we have `ignore_read`, testing change to one of the fields does not make sense, but I tested it by hands back and force with curl and gcloud to make sure fields change after change to the TF config

slevenick (approved):

PR #11944, opened 2024-10-08
Updating documentation for AlertPolicy import_format

brokenthumbs:
- Updating the documentation to properly reflect the import requirements for `google_monitoring_alert_policy` (https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/monitoring_alert_policy#import)
- Per https://github.com/hashicorp/terraform-provider-google/issues/7693#issuecomment-721256403, the `google_monitoring_alert_policy` import expects `projects/{{project_id}}/alertPolicies/{{alert_policy_id}}`.

```release-note:none

```

### DIFF
--- a/mmv1/products/monitoring/AlertPolicy.yaml
+++ b/mmv1/products/monitoring/AlertPolicy.yaml
@@ -29,7 +29,7 @@ update_verb: 'PATCH'
 update_mask: true
 mutex: 'alertPolicy/{{project}}'
 import_format:
-  - '{{name}}'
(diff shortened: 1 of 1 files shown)


roaks3:
@brokenthumbs I'm not sure this change is necessary, since the `name` has this same format. Was there a specific issue you experienced? Per the closed issue that you linked, it seems like there was an issue with the project ids not matching, but I can't tell if that resulted in a problem (like, can the `project_id` on the resource be different from the project id used in the `name`?).

roaks3 (commented):
Mark review pass

PR #11945, opened 2024-10-08
accesscontextmanager - Update ingress / egress array ordering semantics

Charlesleonius:
Fixes a permadiff in ingress / egress fine grained resources in cases where the API returns a different ordering for lists.

```release-note:bug
accesscontextmanager: fixed a permadiff in ingress / egress fine grained resources in cases where the API returns a different ordering for lists
```


trodge:
This field doesn't seem to have any effect on the generated code. Was the intention to make these fields into sets? If so, that may be a breaking change.

Charlesleonius:
@trodge I was under the impression that we could ignore the ordering of fields without making them sets to avoid a breaking change but that only works for top level fields. Im looking for other work arounds because this is causing a permadiff and waiting till the next major release is not an option for some customers

PR #11867, opened 2024-09-28
Data source compute security policy file and function renames

jbuck:
I was looking into fixing a Cloud Armor perma-diff when I found this filename and function not matching the resource/API name.

This PR renames the files (`secutity` to `security`) and one function (`Comput` to `Compute`)

**Release Note Template for Downstream PRs (will be copied)**

```release-note:none

```

PR #11947, opened 2024-10-08
add iamcredentialsbasepath support for plugin framework

BBBmau:
This is temporary and will be resolved once the muxing issues are addressed.

```release-note:none

```

### DIFF
--- a/mmv1/third_party/terraform/fwtransport/framework_config.go.tmpl
+++ b/mmv1/third_party/terraform/fwtransport/framework_config.go.tmpl
@@ -24,7 +24,6 @@ import (
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/logging"
 
 	"github.com/hashicorp/terraform-provider-google/google/fwmodels"
-	iamcredentials "google.golang.org/api/iamcredentials/v1"
 	transport_tpg "github.com/hashicorp/terraform-provider-google/google/transport"
 	"github.com/hashicorp/terraform-provider-google/google/verify"
 
@@ -521,12 +520,6 @@ func (p *FrameworkProviderConfig) SetupGrpcLogging() {
 		option.WithGRPCDialOption(grpc.WithStreamInterceptor(
 			grpc_logrus.PayloadStreamClientInterceptor(logrus.NewEntry(logger), alwaysLoggingDeciderClient))),
 	)
-}
-
-// Remove the `/{{version}}/` from a base path if present.
-func RemoveBasePathVersion(url string) string {
-	re := regexp.MustCompile(`(?P<base>http[s]://.*)(?P<version>/[^/]+?/$)`)
-	return re.ReplaceAllString(url, "$1/")
 }
 
 func (p *FrameworkProviderConfig) logGoogleIdentities(ctx context.Context, data fwmodels.ProviderModel, diags *diag.Diagnostics) {
--- a/mmv1/third_party/terraform/fwtransport/framework_config.go.tmpl
+++ b/mmv1/third_party/terraform/fwtransport/framework_config.go.tmpl
@@ -24,6 +24,7 @@ import (
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/logging"
 
 	"github.com/hashicorp/terraform-provider-google/google/fwmodels"
+	iamcredentials "google.golang.org/api/iamcredentials/v1"
 	transport_tpg "github.com/hashicorp/terraform-provider-google/google/transport"
 	"github.com/hashicorp/terraform-provider-google/google/verify"
 
@@ -57,6 +58,7 @@ type FrameworkProviderConfig struct {
 	DefaultLabels              types.Map
 
 	// paths for client setup
+	IAMCredentialsBasePath           string // TODO: This will be removed once we resove the muxing issues
 	{{- range $product := $.Products }}
 	{{ $product.Name }}BasePath string
 	{{- end }}
@@ -99,6 +101,7 @@ func (p *FrameworkProviderConfig) LoadAndValidateFramework(ctx context.Context, 
 
 	// Setup Base Paths for clients
 	// Generated products
+	p.IAMCredentialsBasePath = data.IamCredentialsCustomEndpoint.ValueString() // TODO: This will be removed once we resove the muxing issues
 	{{- range $product := $.Products }}
 	p.{{ $product.Name }}BasePath = data.{{ $product.Name }}CustomEndpoint.ValueString()
 	{{- end }}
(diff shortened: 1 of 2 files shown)


mmv1/third_party/terraform/fwtransport/framework_config.go.tmpl
```diff
@@ -520,6 +523,26 @@ func (p *FrameworkProviderConfig) SetupGrpcLogging() {
...
+	if err != nil {
+		log.Printf("[WARN] Error creating client iam credentials: %s", err)
+		return nil
+	}
+	clientIamCredentials.UserAgent = userAgent
+	clientIamCredentials.BasePath = iamCredentialsClientBasePath
+
+	return clientIamCredentials
+}
+
```
SarahFrench, line 545: Please move this into fwtransport/framework_provider_clients.go alongside `func (p *FrameworkProviderConfig) NewDnsClient`!

mmv1/third_party/terraform/fwtransport/framework_config.go.tmpl
```diff
@@ -520,6 +523,26 @@ func (p *FrameworkProviderConfig) SetupGrpcLogging() {
 	)
 }
 
+// Remove the `/{{version}}/` from a base path if present.
+func RemoveBasePathVersion(url string) string {
```
SarahFrench, line 527: Could we just import RemoveBasePathVersion from google/transport/config.go?

SarahFrench (changes requested):
Here's some initial feedback - I'll check the code locally on my next pass!

Currently there's an issue with generating the beta provider:

```
error parsing framework_config.go.tmpl for filepath /go/src/github.com/hashicorp/terraform-provider-google-beta/google-beta/fwtransport/framework_config.go template: framework_config.go.tmpl:526: function "version" not defined
```

BBBmau:
Latest changes allow you to build successfully on both ga and beta providers.

The rebase from `upstream/main` was necessary in order to address some compute errors when attempting to build the provider:
```hcl
└─(09:25:40 on main ✹ ✭)──> devbuild                                                                                                                                                          1 ↵ ──(Wed,Oct23)─┘
# github.com/hashicorp/terraform-provider-google/google/services/compute
google/services/compute/data_source_compute_secutity_policy.go:14:6: DataSourceGoogleComputeSecurityPolicy redeclared in this block
        google/services/compute/data_source_compute_security_policy.go:14:6: other declaration of DataSourceGoogleComputeSecurityPolicy
```

Wasn't sure if to rebase on the feature-branch or not so i just decided to do it on here. Let me know if that's fine, i can also rebase on the feature branch itself.

I also tried to manually test `IAMCredentialsBasePath` by testing with the [poc ephemeral resource](https://github.com/BBBmau/magic-modules/blob/44839ef426db9ed84fd03efc13560e5c713a05c1/mmv1/third_party/terraform/ephemerals/ephemeral_service_account_access_token.go) that uses it.

Unfortunately was running into the following error despite it being configured correctly. I can create the PR that includes the code that would add `ephemeral google_service_account_access_token`

image

BBBmau:
resolved the previous comment. It was due to only running `go get github.com/hashicorp/terraform-plugin-framework/ephemeral@627a49bf3060aa8d1c681847ff3c682681ea3cd2` which just focuses on the single framework/ephemeral package and not on the entire plugin-framework. 

Currently running into issues where the `p.providerConfig.UserAgent` returns us `""` which causes errors. Currently investigating how the `userAgent` is set on SDK and how we could replicate this on `plugin-framework`. A reproduction can be seen https://github.com/BBBmau/terraform-provider-google/tree/ephemeral-iamcredentialspath-test

SarahFrench (changes requested):
Please rebase this PR so it's based on the feature branch.

[The PR that caused the problem you're seeing](https://github.com/GoogleCloudPlatform/magic-modules/pull/11867) was merged to main only 3 days ago, whereas the last sync into the feature branch included only commits from Sept 26th and earlier, so shouldn't affect us? I believe the problem is with your local environment as described here: https://github.com/GoogleCloudPlatform/magic-modules/pull/11947#issuecomment-2436357182

BBBmau:
> Please rebase this PR so it's based on the feature branch.
> 
> [The PR that caused the problem you're seeing](https://github.com/GoogleCloudPlatform/magic-modules/pull/11867) was merged to main only 3 days ago, whereas the last sync into the feature branch included only commits from Sept 26th and earlier, so shouldn't affect us? I believe the problem is with your local environment as described here: [#11947 (comment)](https://github.com/GoogleCloudPlatform/magic-modules/pull/11947#issuecomment-2436357182)

rebased back to `FEATURE-BRANCH`-ephemeral-resource`

SarahFrench:
Thanks this makes review a lot easier!

In future what you can do is checkout `FEATURE-BRANCH-ephemeral-resource` in your local copy of the `terraform-provider-google` repo and then make magic-modules generated to that repo (i.e. set [OUTPUT_PATH](https://github.com/GoogleCloudPlatform/magic-modules/blob/34dd74df5c9291629ca1b215636294ca56aaa6ee/GNUmakefile#L64) from the makefile to the terraform-provider-google where the downstream feature branch is checked out).

SarahFrench:
> Currently running into issues where the p.providerConfig.UserAgent returns us "" which causes errors. Currently investigating how the userAgent is set on SDK and how we could replicate this on plugin-framework

[This is where the User-Agent should get a value set during configuration of the PF-implemented provider](https://github.com/hashicorp/terraform-provider-google/blob/5bdb0d346a8b19da16ca021babb3059bfd33ab36/google/fwtransport/framework_config.go#L204-L210):

```
	// Handle User Agent string
	p.UserAgent = CompileUserAgentString(ctx, "terraform-provider-google", tfVersion, providerversion)
	// opt in extension for adding to the User-Agent header
	if ext := os.Getenv("GOOGLE_TERRAFORM_USERAGENT_EXTENSION"); ext != "" {
		ua := p.UserAgent
		p.UserAgent = fmt.Sprintf("%s %s", ua, ext)
	}
```

I'm confused why you'd see that problem, so maybe for now we should review this PR and then tackle any bugs in a subsequent PR where we have an ephemeral resource implemented + some acceptance tests. We should be able to write acc tests and run them locally now that there's an alpha TF binary with the necessary stuff.

SarahFrench:
I used some of your reproduction info and I think there's a problem with provider configuration (i.e. the meta) being passed to the ephemeral resources, I'll follow up with you in Slack. I think that is a problem 'downstream' of this PR though so isn't a blocker for merging this

SarahFrench (approved):
Approved - after we worked out what made our PoC demo not work I could see the new function was usable and didn't cause any obvious problems.

PR #11950, opened 2024-10-08
Update Instance.yaml delete_url for Looker

efeelaiho:
Fixes https://github.com/hashicorp/terraform-provider-google/issues/19467

```release-note:enhancement
looker: added `deletion_policy` to `google_looker_instance` to allow force-destroying instances with nested resources by setting `deletion_policy = FORCE` 
```


shuyama1 (commented):
We'll likely need to hide this change behind a flag as changing resource default deletion behavior is a breaking change according https://googlecloudplatform.github.io/magic-modules/develop/breaking-changes/breaking-changes/#resource-level-breaking-changes and can't be made in a minor release. A similar example change: https://github.com/GoogleCloudPlatform/magic-modules/pull/6474

efeelaiho:
@shuyama1 Added a deletion policy and all tests are passing now 👍

shuyama1 (approved):
Thank you!

PR #11953, opened 2024-10-09
Update CMEK usage in Bigquery tests

rileykarson:
Handful of custodial changes here in cleaning up some conflicting global IAM grants:

* Move towards key-level permissions as is done in this guide: https://cloud.google.com/bigquery/docs/customer-managed-encryption#assign_role
    * Remove bootstrapping of a project-level role as well
* Rename the `google_bigquery_connection` `kms` example to `sql_with_cmek` to better reflect what it actually showcases. Source page is https://cloud.google.com/bigquery/docs/connect-to-sql#create-sql-connection
* Cleaned up some inconsistent whitespace that was either 4-space or tabs (to 2-space) and moved around some fields that were not ordered as count > identity > other settings

```release-note:none
```


rileykarson:
This should work right now (we're no longer fighting over the permission) but still has a global static resource for the kms/sql_with_cmek test case. I may take a different approach- I'm undecided because it doesn't really matter _right now_ while there's only one user (the other two tests use their own KMS keys and not the global static bootstrapped one).

mmv1/templates/terraform/examples/bigquery_job_copy.tf.tmpl
```diff
@@ -68,7 +69,7 @@ EOF
     kms_key_name = google_kms_crypto_key.crypto_key.id
```
trodge, line 69: Where is this crypto key resource defined?

It doesn't get created with every test run, does it?

mmv1/templates/terraform/examples/bigquery_job_copy.tf.tmpl
```diff
@@ -92,8 +93,8 @@ data "google_project" "project" {
   project_id = "{{index $.TestEnvVars "project"}}"
 }
 
-resource "google_project_iam_member" "encrypt_role" {
-  project = data.google_project.project.project_id
+resource "google_kms_crypto_key_iam_member" "encrypt_role" {
```
trodge, line 96: Can we put this iam member resource in `.ci/infra/terraform/main.tf` instead of here? If it's part of the test run, it will get cleaned up when it finishes and potentailly cause simultaneous runs to fail.

trodge (changes requested):
`BootstrapPSARole` applies the permission at the project level. Is it necessary to apply it at the key level?

mmv1/templates/terraform/examples/bigquery_job_copy.tf.tmpl
rileykarson: Yeah, they are. These tests aren't the ones that are failing and I'm not addressing that in this change.

mmv1/templates/terraform/examples/bigquery_job_copy.tf.tmpl
rileykarson: That's fine, for now, imo- we want to record the permission grants in our examples. This is the exclusive user so far.

rileykarson:
> BootstrapPSARole applies the permission at the project level. Is it necessary to apply it at the key level?

Not if we applied it at the project level, but I removed that grant so now we need to apply it at the key level.

mmv1/templates/terraform/examples/bigquery_job_copy.tf.tmpl
rileykarson: (if we have a method to both apply permissions in configs *and* add the permission there, glad to follow that instead though- just don't want to take on the larger correction in place of the immediate problem of fixing the test)

mmv1/templates/terraform/examples/bigquery_connection_sql_with_cmek.tf.tmpl
trodge, line 1: Is this file based on the above deleted file?

If so, could you use `git mv` so that the history is preserved?

mmv1/templates/terraform/examples/bigquery_job_copy.tf.tmpl
trodge: What ensures that the key used in this test is not also used in other tests?

trodge (changes requested):

mmv1/templates/terraform/examples/bigquery_connection_sql_with_cmek.tf.tmpl
rileykarson: Do I need to break that up over several commits? I see the `mv` go away:

```
rileykarson@rileykarson-macbookpro ~/magic-modules
$ git status
On branch rileykarson-patch-4
Your branch is behind 'upstream/rileykarson-patch-4' by 1 commit, and can be fast-forwarded.
  (use "git pull" to update your local branch)

Changes to be committed:
  (use "git restore --staged <file>..." to unstage)
	renamed:    mmv1/templates/terraform/examples/bigquery_connection_kms.tf.tmpl -> mmv1/templates/terraform/examples/bigquery_connection_sql_with_cmek.tf.tmpl

Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
	modified:   mmv1/products/bigqueryconnection/Connection.yaml
	modified:   mmv1/templates/terraform/examples/bigquery_connection_sql_with_cmek.tf.tmpl
	modified:   mmv1/templates/terraform/examples/bigquery_job_copy.tf.tmpl
	modified:   mmv1/templates/terraform/examples/bigquery_job_copy_table_reference.tf.tmpl


rileykarson@rileykarson-macbookpro ~/magic-modules
$ git add .

rileykarson@rileykarson-macbookpro ~/magic-modules
$ git status
On branch rileykarson-patch-4
Your branch is behind 'upstream/rileykarson-patch-4' by 1 commit, and can be fast-forwarded.
  (use "git pull" to update your local branch)

Changes to be committed:
  (use "git restore --staged <file>..." to unstage)
	modified:   mmv1/products/bigqueryconnection/Connection.yaml
	deleted:    mmv1/templates/terraform/examples/bigquery_connection_kms.tf.tmpl
	new file:   mmv1/templates/terraform/examples/bigquery_connection_sql_with_cmek.tf.tmpl
	modified:   mmv1/templates/terraform/examples/bigquery_job_copy.tf.tmpl
	modified:   mmv1/templates/terraform/examples/bigquery_job_copy_table_reference.tf.tmpl
```

mmv1/templates/terraform/examples/bigquery_job_copy.tf.tmpl
rileykarson: Nothing structurally. I'm not _super_ worried about it happening because it's a relatively uncommon feature, and this is currently the sole user. If this were not an example test I think it would worth correcting now, but I think we need to solve for a way to show permission grants in examples while actually handling them OOB.

trodge (approved):

mmv1/templates/terraform/examples/bigquery_connection_sql_with_cmek.tf.tmpl
trodge: Looks like we actually needed to make two separate PRs to keep the history. TIL.

PR #11955, opened 2024-10-09
Add `creation_timestamp` field to `google_compute_instance` and template resources

karolgorc:
closes https://github.com/hashicorp/terraform-provider-google/issues/15663

- added field `creation_timestamp` to
  - google_compute_instance
  - google_compute_instance_template
  - google_compute_region_instance_template
- added data source support (templates didn't need a change because they use the Read function from the resource)
- added doc changes

This is an output only field. Please LMK if you want it tested somehow.

```release-note:enhacement
compute: added `creation_timestamp` field to `google_compute_instance`, `google_compute_instance_template`, `google_compute_region_instance_template`
```

### DIFF
--- a/mmv1/third_party/terraform/services/compute/data_source_google_compute_instance.go.tmpl
+++ b/mmv1/third_party/terraform/services/compute/data_source_google_compute_instance.go.tmpl
@@ -3,9 +3,9 @@ package compute
 import (
 	"fmt"
 
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 	"github.com/hashicorp/terraform-provider-google/google/tpgresource"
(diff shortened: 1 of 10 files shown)


hao-nan-li (changes requested):
Could you add this field to an existing test? Thanks

hao-nan-li:
I'm not sure if this field is tested. Add it to any existing test should be fine.

karolgorc:
It's an output only field. You cannot do anything else than read it. Which is done here
```go
resource.TestCheckResourceAttrSet("google_compute_instance_template.foobar", "creation_timestamp"),
```

I cannot really put a `creation_timestamp = X` in the test because it's `Computed: true` only. Same as self_link, etc.

hao-nan-li (approved):

PR #11956, opened 2024-10-09
Clarify data source attached_disk documentation

Asiderr:
closes: https://github.com/hashicorp/terraform-provider-google/issues/15718

This patch clarifies that the `source` field of `attached_disk` data source in the `google_compute_instance` which always is `self_link` and cannot be name.

```release-note:none
Clarifies that the `source` field of `attached_disk` data source in the `google_compute_instance` which always is `self_link` and cannot be name.
```


trodge:
Thanks for the change, I just made the release note type `none` since we don't use release notes for docs-only changes.

trodge (approved):

PR #11959, opened 2024-10-09
Removing all .rb and .erb files

NickElliot:
Step 2 of removing ruby files

All actual ruby files removed except those pertaining to autogen

@slevenick would it be fine to proceed with removing those too, with them being accessible via the legacy-ruby branch, or would it be preferable to retain them for the time being?

```release-note:none

```


NickElliot:
cc @c2thorn

.ci/containers/build-environment/Dockerfile
```diff
@@ -39,8 +39,8 @@ RUN git config --global user.email "magic-modules@google.com"
 RUN go install golang.org/x/tools/cmd/goimports@d088b475e3360caabc032aaee1dc66351d4e729a
 RUN go install github.com/github/hub@v2.11.2+incompatible
 
-ADD "https://raw.githubusercontent.com/GoogleCloudPlatform/magic-modules/main/mmv1/Gemfile" Gemfile
-ADD "https://raw.githubusercontent.com/GoogleCloudPlatform/magic-modules/main/mmv1/Gemfile.lock" Gemfile.lock
+ADD "https://raw.githubusercontent.com/GoogleCloudPlatform/magic-modules/refs/heads/legacy-ruby/mmv1/Gemfile" Gemfile
```
slevenick, line 42: We can probably remove ruby & the gemfiles from the build environment now?

slevenick (approved):
I'm fine with removing the ruby code at this point. It'll make it more obvious if people run `bundle exec ...` that they need to change their command

.ci/containers/build-environment/Dockerfile
NickElliot: Correct, but breaking it up into multiple steps (the part 1 is actually updating main to use these legacy references, and this part 2 was layered on that)

c2thorn (changes requested):
conflict on `main.go`

c2thorn (approved):

PR #11960, opened 2024-10-09
Feature/resource oracle database autonomous database

tulika-aakriti:
Part of: https://github.com/hashicorp/terraform-provider-google/issues/19701

```release-note:new-resource
`google_oracle_database_autonomous_database`
```


mmv1/templates/terraform/examples/oracledatabase_autonomous_database_basic.tf.tmpl
```diff
@@ -0,0 +1,22 @@
...
+    compute_count = "2"
+    data_storage_size_tb="1"
+    db_version = "19c"
+    db_workload = "OLTP"
+    is_auto_scaling_enabled= "true"
+    license_type = "LICENSE_INCLUDED"
+    }
+}
+
+data "google_compute_network" "default" {
```
rileykarson, line 20: I assume the network needs to live in the same project as the resource; we'll want to specify `project` on this. If we want to create the project per test run we'll want to use the resource instead of the datasource (https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/compute_network). If there's an already-created network we should be attaching to, we'd reference it here in the datasource.

tulika-aakriti:
> 
> ## Diff report
> 
> `google` provider: [Diff](https://github.com/modular-magician/terraform-provider-google/compare/auto-pr-11960-old..auto-pr-11960) ( 5 files changed, 4672 insertions(+), 2 deletions(-)) `google-beta` provider: [Diff](https://github.com/modular-magician/terraform-provider-google-beta/compare/auto-pr-11960-old..auto-pr-11960) ( 5 files changed, 4672 insertions(+), 2 deletions(-)) `terraform-google-conversion`: [Diff](https://github.com/modular-magician/terraform-google-conversion/compare/auto-pr-11960-old..auto-pr-11960) ( 1 file changed, 1355 insertions(+)) Open in Cloud Shell: [Diff](https://github.com/modular-magician/docs-examples/compare/auto-pr-11960-old..auto-pr-11960) ( 8 files changed, 261 insertions(+))
> 
> ## Missing test report
> 
> Resource: `google_oracle_database_autonomous_database` (2 total tests) Please add an acceptance test which includes these fields. The test should include the following:
> 
> ```hcl
> resource "google_oracle_database_autonomous_database" "primary" {
>   properties {
>     cpu_core_count         = # value needed
>     data_storage_size_gb   = # value needed
>     db_edition             = # value needed
>     private_endpoint_ip    = # value needed
>     private_endpoint_label = # value needed
>     secret_id              = # value needed
>     vault_id               = # value needed
>   }
> }
> ```
We can not add all these fields as these are optional (should be given if other field are not given) otherwise it will fail in prepare call.
example - ` InvalidParameter: dataStorageSizeInGBs parameter cannot be used with the dataStorageSizeInTBs parameter" `
https://docs.oracle.com/en-us/iaas/api/#/en/database/20160918/datatypes/CreateAutonomousDatabaseBase:~:text=This%20cannot%20be%20used%20in%20conjunction%20with%20adminPassword.

mmv1/templates/terraform/examples/oracledatabase_autonomous_database_basic.tf.tmpl
tulika-aakriti: Thanks added project field in network datasource worked.

mmv1/products/oracledatabase/AutonomousDatabase.yaml
```diff
@@ -0,0 +1,622 @@
...
+      update_minutes: 120
+      delete_minutes: 120
+  result:
+    path: 'response'
+    resource_inside_response: true
+  error:
+    path: 'error'
+    message: 'message'
+custom_code:
+examples:
```
rileykarson, line 51: Can we include an example/test case that doesn't set `properties`? We'd be fine to omit it from docs if you'd prefer.

mmv1/templates/terraform/examples/oracledatabase_autonomous_database_basic.tf.tmpl
```diff
@@ -0,0 +1,23 @@
...
+  display_name = "autonomousDatabase displayname"
+  autonomous_database_id = "my-adb-instance"
+  location = "us-east4"
+  project = "{{index $.Vars "project"}}"
+  database = "testADBdb"
+  admin_password = "123Abpassword"
+  network = data.google_compute_network.default.id
+  cidr = "10.5.0.0/24"
+  properties {
+    compute_count = "2"
```
rileykarson, line 11: Can we omit any `properties` subfields that are not required here?

mmv1/templates/terraform/examples/oracledatabase_autonomous_database_basic.tf.tmpl
```diff
@@ -0,0 +1,23 @@
+resource "google_oracle_database_autonomous_database" "{{$.PrimaryResourceId}}"{
+  display_name = "autonomousDatabase displayname"
+  autonomous_database_id = "my-adb-instance"
+  location = "us-east4"
+  project = "{{index $.Vars "project"}}"
+  database = "testADBdb"
```
rileykarson, line 6: Can we omit the `optional` values here?

mmv1/products/oracledatabase/AutonomousDatabase.yaml
tulika-aakriti: Sorry, I couldn't understand it. If we didn't set the properties the create API will fail as it has required field. And we definitely prefer not to have that in docs.

mmv1/products/oracledatabase/AutonomousDatabase.yaml
```diff
@@ -0,0 +1,622 @@
...
+    url_param_only: true
+    required: true
+    immutable: true
+properties:
+  - name: 'name'
+    type: String
+    description: "Identifier. The name of the Autonomous Database resource in the
+     following format:\nprojects/{project}/locations/{region}/autonomousDatabases/{autonomous_database} "
+    output: true
+  - name: 'database'
```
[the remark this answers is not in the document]
tulika-aakriti: displayName is a name which will be displayed on the UI 
Database is like a suffix and need to unique, it doesn't have to be same as autonomousDatabaseId 
Our API will call the oracle control plane so we want to be compatible to their field. Attaching oci api spec for reference -  https://docs.oracle.com/en-us/iaas/api/#/en/database/20160918/AutonomousDatabase/

mmv1/templates/terraform/examples/oracledatabase_autonomous_database_basic.tf.tmpl
tulika-aakriti: removed displayName, we need to pass project as we want to override the CI's project with our GCP project.

mmv1/templates/terraform/examples/oracledatabase_autonomous_database_basic.tf.tmpl
tulika-aakriti: The fields are optional but there are condition for example - 
1. user has to provide one of dataStorageSizeTb or dataStorageSizeGb.
2. user has to provide  computeCount or cpuCoreCount
Otherwise the prepare operation will itself fail

tulika-aakriti:
Hi Riley, Please review the PR.
Thanks

mmv1/products/oracledatabase/AutonomousDatabase.yaml
```diff
@@ -0,0 +1,618 @@
...
+    type: String
+    description: "The ID of the subscription entitlement associated with
+    the Autonomous\nDatabase. "
+    output: true
+  - name: 'adminPassword'
+    type: String
+    description: 'The password for the default ADMIN user. '
+    ignore_read: true
+  - name: 'properties'
+    type: NestedObject
```
rileykarson, line 112:
```suggestion
    type: NestedObject
    required: true
```

Per https://github.com/GoogleCloudPlatform/magic-modules/pull/11960/files#r1798802833, this field is required since some of its fields are required at creation time.

mmv1/templates/terraform/examples/oracledatabase_autonomous_database_basic.tf.tmpl
rileykarson: We could consider using exactly_one_of for those pairs: https://googlecloudplatform.github.io/magic-modules/develop/field-reference/#exactly_one_of

Non blocking, though.

tulika-aakriti: Will do as a follow up. Thanks for suggestions.

rileykarson (approved):

PR #11962, opened 2024-10-09
Break best practices pages out into standalone pages

rileykarson:
Mainly for discovery, as this puts the titles of the sections in the sidebar.

```release-note:none
```


docs/content/best-practices/immutable-fields.md
```diff
@@ -0,0 +1,22 @@
+---
+title: "Immutable fields"
+weight: 10
```
melinath, line 3: We should preserve the old redirect & add a new one. I'd suggest here because it's the first new page.

```suggestion
weight: 10
aliases:
  - /docs/best-practices
  - /best-practices
```

melinath (approved):
LGTM - this has been bugging me for a while, thanks!

docs/content/best-practices/immutable-fields.md
rileykarson: Oh, I'd tried just /docs/best-practices and it didn't seem to work so I removed it. That explains it!

PR #11981, opened 2024-10-10
Fix deletion_protection fields specified as strings

rileykarson:
Part of https://github.com/hashicorp/terraform-provider-google/issues/19747

These happen to work because of Terraform's particularly aggressive type coercion, but our docs should specify the true types where possible. This will turn some entries from `""` to `` (blank). Working on correcting for those in a followup change, I'll probably gate submission of this on finishing those (and merging both PRs independently will have no merge conflicts, but combine to correct all the pages due to a helpful-for-once cross merge).

I thought these might have been introduced during the Go rewrite, but turns out they werent! They were like that in Ruby too.

```release-note:none
```

### DIFF
--- a/mmv1/templates/terraform/examples/bigquery_connection_basic.tf.tmpl
+++ b/mmv1/templates/terraform/examples/bigquery_connection_basic.tf.tmpl
@@ -6,7 +6,7 @@ resource "google_sql_database_instance" "instance" {
 		tier = "db-f1-micro"
 	}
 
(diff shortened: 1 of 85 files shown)


NickElliot (approved):
resulting diffs look good, funny situation

rileykarson:
TestAccSqlDatabaseInstance_Edition_Downgrade was broken in another test. Will resolve in a separate PR.

I have the other PR staged; it'll get merged by the next website publish, so gonna merge.

PR #11987, opened 2024-10-11
Add cai_base_url fields for resources that need them

roaks3:
I believe these services all need the `cai_base_url` field to be specified, because of how their url is formatted. However, I don't know enough about this mechanism to know if it is actually correct (and what will change as a result).

Within the context of API matching, these were discovered to be services where we did not have a matching API name, which is what prompted the correction (we use `cai_base_url` as an override for matching if it is specified).

```release-note:none
fix cai_base_url values for some resources that have base urls that do not match their service domain name
```

### DIFF
--- a/mmv1/products/deploymentmanager/product.yaml
+++ b/mmv1/products/deploymentmanager/product.yaml
@@ -17,5 +17,6 @@ display_name: 'Cloud Deployment Manager'
 versions:
   - name: 'ga'
     base_url: 'https://www.googleapis.com/deploymentmanager/v2/'
+    cai_base_url: 'https://deploymentmanager.googleapis.com/v2/'
 scopes:
   - 'https://www.googleapis.com/auth/cloud-platform'
--- a/mmv1/products/secretmanagerregional/product.yaml
+++ b/mmv1/products/secretmanagerregional/product.yaml
@@ -18,7 +18,9 @@ display_name: 'Secret Manager'
 versions:
   - name: 'ga'
     base_url: 'https://secretmanager.{{location}}.rep.googleapis.com/v1/'
+    cai_base_url: 'https://secretmanager.googleapis.com/v1/'
   - name: 'beta'
     base_url: 'https://secretmanager.{{location}}.rep.googleapis.com/v1/'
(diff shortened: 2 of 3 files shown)


roaks3:
@melinath not sure your thoughts here. From looking over the diff, I think these values are more correct, but I'm wondering if it could be misleading since these resources are not actually all present in the asset inventory. If that's the case, we would potentially need another field to override the url.

melinath:
yeah, cai_base_url should be renamed to something else. Could be combined into an api_resource_type override field maybe?

melinath (approved):
This LGTM if you want to go ahead with merge & figure out the rest later.

mmv1/products/deploymentmanager/product.yaml
```diff
@@ -17,5 +17,6 @@ display_name: 'Cloud Deployment Manager'
 versions:
   - name: 'ga'
     base_url: 'https://www.googleapis.com/deploymentmanager/v2/'
+    cai_base_url: 'https://deploymentmanager.googleapis.com/v2/'
```
melinath, line 20: Oh actually one thing - I _think_ the CAI base url shouldn't contain the version.

melinath (changes requested):

mmv1/products/deploymentmanager/product.yaml
melinath: see https://cloud.google.com/asset-inventory/docs/supported-asset-types

roaks3: True, but all of the other instances of `cai_base_url` in the repo appear to include the version. We parse it for the metadata generation, and looks like we do the same for TGC, so I think we want to keep it.

roaks3:
I believe this might actually be good to merge as-is. @melinath could you double check?

Renaming of field will be address separately.

mmv1/products/deploymentmanager/product.yaml
melinath: SGTM - we can fix later

melinath (approved):

PR #11992, opened 2024-10-14
compute: Validate stopping for update in plan stage

Asiderr:
This patch moves the `allowUpdate` check from the apply stage to the plan stage. Covers all configuration changes that require an instance to be stopped. It also adds new helper functions for checking network configuration changes and service account changes.

This patch is covered by existing acceptance tests: `TestAccComputeInstance_updateRunning_desiredStatusRunning_notAllowStoppingForUpdate` `TestAccComputeInstance_updateRunning_desiredStatusNotSet_notAllowStoppingForUpdate`

* Add new customdiff `validateStoppingForUpdate`
* Add new helper functions
  * `hasServiceAccountChanged`
  * `hasNetworkInterfaceChanged`
* Delete `allowUpdate` check from apply stage
* Refactor scopeChange check using the new helper function in apply stage

Closes: https://github.com/hashicorp/terraform-provider-google/issues/4076

```release-note:enhancement
compute: validate stopping for update in plan stage
```

### DIFF
--- a/mmv1/third_party/terraform/services/compute/resource_compute_instance_test.go.tmpl
+++ b/mmv1/third_party/terraform/services/compute/resource_compute_instance_test.go.tmpl
@@ -1253,7 +1253,7 @@ func TestAccComputeInstance_serviceAccount(t *testing.T) {
 						"https://www.googleapis.com/auth/userinfo.email"),
 				),
 			},
-			computeInstanceImportStep("us-central1-a", instanceName, []string{}),
+			computeInstanceImportStep("us-central1-a", instanceName, []string{"allow_stopping_for_update"}),
 		},
 	})
 }
@@ -1277,7 +1277,7 @@ func TestAccComputeInstance_noServiceAccount(t *testing.T) {
 					testAccCheckComputeInstanceNoServiceAccount(&instance),
 				),
 			},
-			computeInstanceImportStep("us-central1-a", instanceName, []string{}),
+			computeInstanceImportStep("us-central1-a", instanceName, []string{"allow_stopping_for_update"}),
 		},
 	})
 }
@@ -6848,6 +6848,7 @@ resource "google_compute_instance" "foobar" {
   name         = "%s"
   machine_type = "e2-medium"
   zone         = "us-central1-a"
+  allow_stopping_for_update = true
 
   boot_disk {
     initialize_params {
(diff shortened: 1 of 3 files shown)


ScottSuarez (commented):
Waiting for test fixes

mmv1/third_party/terraform/services/compute/resource_compute_instance_test.go.tmpl
```diff
@@ -1253,7 +1253,7 @@ func TestAccComputeInstance_serviceAccount(t *testing.T) {
 						"https://www.googleapis.com/auth/userinfo.email"),
 				),
 			},
-			computeInstanceImportStep("us-central1-a", instanceName, []string{}),
```
ScottSuarez, line 1256: Could you help me understand why we are changing the tests? This tends to imply a change in behavior which is not what we want for this resource.

Asiderr: When we import an instance during a test and do not create a new one, we need to add an allow_stopping_for_update resource. And IMHO this is the correct behaviour - we have an existing instance that has some configuration and we want to change some resources, which requires us to stop the instance.
In the case of testAccComputeInstance_standardVM_maxRunDurationUpdated, there was no import of the instance, but the max_run_duration for the existing instance was changed, which also requires the instance to be stopped, so the test configuration should include allow_stopping_for_update.

```
"Changing the machine_type, min_cpu_platform, service_account, enable_display, shielded_instance_config, scheduling.node_affinities, scheduling.max_run_duration " +
"or network_interface.[#d].(network/subnetwork/subnetwork_project) or advanced_machine_features on a started instance requires stopping it. "
```

mmv1/third_party/terraform/services/compute/resource_compute_instance_test.go.tmpl
```diff
@@ -6848,6 +6848,7 @@ resource "google_compute_instance" "foobar" {
   name         = "%s"
   machine_type = "e2-medium"
   zone         = "us-central1-a"
+  allow_stopping_for_update = true
```
ScottSuarez, line 6851: My conclusion after looking is that these tests should not be changed. allow_stopping_for_update should not be needed on these tests because these tests do not update. If there is no way to check that this will effect only update plans we should avoid making this change.

This would break existing customer terraform deployments where `allow_stopping_for_update` is not specified.

ScottSuarez (changes requested):

PR #11993, opened 2024-10-14
Make the appengine service name sweepable in the test compute region network endpoint

zli82016:
part of https://github.com/hashicorp/terraform-provider-google/issues/16170

Make the appengine service name sweepable in the test TestAccComputeRegionNetworkEndpointGroup_regionNetworkEndpointGroupAppengineExample

```release-note:none

```


rileykarson (approved):
LGTM- version does have `delete_service_on_destroy = true` so we'll clean up by default, and sweepers should catch leaks.

PR #11995, opened 2024-10-15
Fix import for google_essential_contacts_contact

ttewari:
Fixes https://github.com/hashicorp/terraform-provider-google/issues/11124

Added custom import to include required parent attribute when contact is imported. [Argument ref](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/essential_contacts_contact#argument-reference) for essential contacts.

```release-note:bug
essentialcontacts: fixed `google_essential_contacts_contact` import to include required parent field.
```


slevenick (changes requested):
Can you add a test that uses this import format?

ttewari:
> Can you add a test that uses this import format?

Updated the tests to not ignore the parent attribute while verifying import state.

slevenick (approved):
Thanks!

PR #11996, opened 2024-10-15
Feat: Add google_oracle_database_cloud_exadata_infrastructure datasource

tulika-aakriti:
Part of- https://github.com/hashicorp/terraform-provider-google/issues/19849

```release-note:new-datasource
`google_oracle_database_cloud_exadata_infrastructure`
```

### DIFF
--- a/mmv1/third_party/terraform/services/oracledatabase/data_source_oracle_database_cloud_exadata_infrastructure_test.go
+++ b/mmv1/third_party/terraform/services/oracledatabase/data_source_oracle_database_cloud_exadata_infrastructure_test.go
@@ -1,0 +1,41 @@
+package oracledatabase_test
+
+import (
+	"fmt"
+	"github.com/hashicorp/terraform-plugin-testing/helper/resource"
+	"github.com/hashicorp/terraform-provider-google/google/acctest"
+	"testing"
+)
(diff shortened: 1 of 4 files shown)


mmv1/third_party/terraform/services/oracledatabase/data_source_oracle_database_cloud_exadata_infrastructure_test.go
```diff
@@ -0,0 +1,42 @@
...
+			},
+		},
+	})
+}
+
+func testAccOracleDatabaseCloudExadataInfrastructure_basic() string {
+	return fmt.Sprintf(`
+data "google_oracle_database_cloud_exadata_infrastructure" "my-exadata"{
+  cloud_exadata_infrastructure_id = "ofake-do-not-delete-tf-exadata"
+  project = "oci-terraform-testing"
```
slevenick, line 38: We don't want to hard code this project here. There should be functions used in the other tests to pass through this project via environment variables. Can you use those in this test? You'll need to call them from the above function and pass the value through as an argument to this one.

slevenick (changes requested):

slevenick (approved):

PR #12006, opened 2024-10-15
Doc: add some missing description for ssl_mode and server_ca_mode.

feng-zhe:
Add some missing description for ssl_mode and server_ca_mode. Related issue https://github.com/hashicorp/terraform-provider-google/issues/19711.

```release-note:none

```

### DIFF
--- a/mmv1/third_party/terraform/website/docs/r/sql_database_instance.html.markdown
+++ b/mmv1/third_party/terraform/website/docs/r/sql_database_instance.html.markdown
@@ -377,7 +377,7 @@ Specifying a network enables private IP.
 At least `ipv4_enabled` must be enabled or a `private_network` must be configured.
 This setting can be updated, but it cannot be removed after it is set.
 
-* `ssl_mode` - (Optional) Specify how SSL connection should be enforced in DB connections. Supported values are `ALLOW_UNENCRYPTED_AND_ENCRYPTED`, `ENCRYPTED_ONLY`, and `TRUSTED_CLIENT_CERTIFICATE_REQUIRED` (not supported for SQL Server). See [API reference doc](https://cloud.google.com/sql/docs/postgres/admin-api/rest/v1beta4/instances#ipconfiguration) for details.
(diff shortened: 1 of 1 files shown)


mmv1/third_party/terraform/website/docs/r/sql_database_instance.html.markdown
```diff
@@ -377,9 +377,9 @@ Specifying a network enables private IP.
 At least `ipv4_enabled` must be enabled or a `private_network` must be configured.
 This setting can be updated, but it cannot be removed after it is set.
 
-* `ssl_mode` - (Optional) Specify how SSL connection should be enforced in DB connections.
+* `ssl_mode` - (Optional) Specify how SSL connection should be enforced in DB connections. Supported values are `ALLOW_UNENCRYPTED_AND_ENCRYPTED`, `ENCRYPTED_ONLY`, and `TRUSTED_CLIENT_CERTIFICATE_REQUIRED` (not supported for SQL Server). See [API reference doc](https://cloud.google.com/sql/docs/postgres/admin-api/rest/v1beta4/instances#ipconfiguration) for details.
```
SarahFrench, line 380: Are any of the values here Beta-only? If not we should link to the GA version of the API documentation to avoid any potential confusion

SarahFrench (commented):
Hi, thanks for your PR!

In the original issue the author wants [the detail present here in the old docs](https://registry.terraform.io/providers/hashicorp/google/5.43.1/docs/resources/sql_database_instance#ssl_mode) to be reintroduced. Could you please add those details into your PR too?

mmv1/third_party/terraform/website/docs/r/sql_database_instance.html.markdown
feng-zhe: We cannot reintroduce the previous details because the `require_ssl` field has been deprecated in 6.0.0. So we only need to describe the valid values of `ssl_mode`.

I've updated the API version to v1. Thanks.

SarahFrench: Ah, thanks, I missed that that was a factor!

SarahFrench (approved):

PR #12009, opened 2024-10-15
Remove parallel runs for vertex_endpoint_iam tests

hao-nan-li:
There are 3 flaky different vertex endpoint iam tests failing for the same reason. Looks like for each endpoint only one network is permitted. Remove parallel runs for the tests and see if it fixes the problem.

Note: These tests always pass in VCR.
fixes https://github.com/hashicorp/terraform-provider-google/issues/18932

```release-note:none

```

### DIFF
--- a/mmv1/third_party/terraform/services/vertexai/iam_vertex_endpoint_test.go.tmpl
+++ b/mmv1/third_party/terraform/services/vertexai/iam_vertex_endpoint_test.go.tmpl
@@ -52,7 +52,7 @@ func TestAccVertexAIEndpointIamMember(t *testing.T) {
 	t.Parallel()
 
 	context := map[string]interface{}{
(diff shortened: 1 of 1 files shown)


mmv1/third_party/terraform/services/vertexai/iam_vertex_endpoint_test.go.tmpl
```diff
@@ -13,7 +13,7 @@ import (
 )
 
 func TestAccVertexAIEndpointIamBinding(t *testing.T) {
-	t.Parallel()
+	// t.Parallel()
```
trodge, line 16:
```suggestion
	// t.Parallel()
	// See https://github.com/hashicorp/terraform-provider-google/issues/18932
```

We could link the issue here to explain why these tests shouldn't be run in parallel.

trodge (approved):

PR #11697, opened 2024-09-13
container: fix `node_config.kubelet_config` updates in `google_container_cluster`

wyardley:
This is still shy of a fix for updates for everything within `node_config`, and doesn't consolidate where we're updating the `google_container_node_pool` resource vs. the `default-pool` that can be created directly within `google_container_cluster`, but it does fix a subset of the fields that don't work, and also simplifies some of the code and consolidates some tests, so maybe still worth considering as another step in the right direction. 

- Fix updates for fields within `node_config.kubelet_config` where updates didn't function properly (basically all except for `insecure_kubelet_readonly_port_enabled`.
- Consolidate test cases for items under `node_config.kubelet_config` with the one for `node_config.kubelet_config.insecure_kubelet_readonly_port_enabled`, including testing updates.

Part of https://github.com/hashicorp/terraform-provider-google/issues/19225

**Release Note Template for Downstream PRs (will be copied)**

```release-note:bug
container: fixed `node_config.kubelet_config` updates in `google_container_cluster` resource
```

PR #12014, opened 2024-10-16
container: fix missing updates for `google_container_cluster.node_config` subfields

wyardley:
This resolves an issue where many subfields of `node_config` in a cluster (which affects the default node-pool "default-pool" when `remove_default_node_pool` is not set to `false`) don't support updates properly, and also allows 3 subfields which had been previously set to force recreation of the default node pool (because updates were non-functional) to be updated in-place.

Some acceptance tests are added, and some existing tests have been adjusted to confirm that the behavior is the same between `google_container_cluster.node_config` and `google_container_node_pool.node_config`.

Fixes hashicorp/terraform-provider-google#19225
Fixes hashicorp/terraform-provider-google#18208
Fixes hashicorp/terraform-provider-google#16054
Fixes hashicorp/terraform-provider-google#13872

Possible fix for hashicorp/terraform-provider-google#17522 **roaks3:** This helps but I don't think it's a fix per se
Partial / possible fix for hashicorp/terraform-provider-google#12966 **roaks3:** Agreed, might fix, but we shouldn't close it yet

Followup to #11826 where the code used by the regular node pool update code was broken out.

```release-note:bug
container: fixed missing in-place updates for some `google_container_cluster.node_config` subfields
```

```release-note:enhancement
container: added in-place update support for `labels`, `resource_manager_tags` and `workload_metadata_config` in `google_container_cluster.node_config` 
```

# References

- https://github.com/hashicorp/terraform-provider-google/issues/19225
- #11272
- #11717
- #11826

See above for some other issues that I believe may be fully or partially resolved by this.

b/361634104

### DIFF
--- a/mmv1/third_party/terraform/services/container/resource_container_cluster.go.tmpl
+++ b/mmv1/third_party/terraform/services/container/resource_container_cluster.go.tmpl
@@ -3812,133 +3812,15 @@ func resourceContainerClusterUpdate(d *schema.ResourceData, meta interface{}) er
 	}
 
 	if d.HasChange("node_config") {
-		if d.HasChange("node_config.0.image_type") {
-			it := d.Get("node_config.0.image_type").(string)
-			req := &container.UpdateClusterRequest{
-				Update: &container.ClusterUpdate{
-					DesiredImageType: it,
-				},
-			}
 
-			updateF := func() error {
-				name := containerClusterFullName(project, location, clusterName)
-				clusterUpdateCall := config.NewContainerClient(userAgent).Projects.Locations.Clusters.Update(name, req)
-				if config.UserProjectOverride {
-					clusterUpdateCall.Header().Add("X-Goog-User-Project", project)
-				}
-				op, err := clusterUpdateCall.Do()
-				if err != nil {
-					return err
-				}
+		defaultPool := "default-pool"
 
-				// Wait until it's updated
-				return ContainerOperationWait(config, op, project, location, "updating GKE image type", userAgent, d.Timeout(schema.TimeoutUpdate))
-			}
-
-			// Call update serially.
-			if err := transport_tpg.LockedCall(lockKey, updateF); err != nil {
-				return err
-			}
-
-			log.Printf("[INFO] GKE cluster %s: image type has been updated to %s", d.Id(), it)
+		nodePoolInfo, err := extractNodePoolInformationFromCluster(d, config, clusterName)
+		if err != nil {
+			return err
 		}
 
-		if d.HasChange("node_config.0.kubelet_config") {
-
-			defaultPool := "default-pool"
-
-			timeout := d.Timeout(schema.TimeoutCreate)
-
-			nodePoolInfo, err := extractNodePoolInformationFromCluster(d, config, clusterName)
-			if err != nil {
-				return err
-			}
-
-			// Acquire write-lock on nodepool.
-			npLockKey := nodePoolInfo.nodePoolLockKey(defaultPool)
-
-			// Still should be further consolidated / DRYed up
-			// See b/361634104
-			it := d.Get("node_config.0.kubelet_config")
-
-			// While we're getting the value from fields in
-			// node_config.kubelet_config, the actual setting that needs to be
-			// updated is on the default nodepool.
-			req := &container.UpdateNodePoolRequest{
-				Name: defaultPool,
-				KubeletConfig: expandKubeletConfig(it),
-			}
-
-			updateF := func() error {
-				clusterNodePoolsUpdateCall := config.NewContainerClient(userAgent).Projects.Locations.Clusters.NodePools.Update(nodePoolInfo.fullyQualifiedName(defaultPool), req)
-				if config.UserProjectOverride {
-					clusterNodePoolsUpdateCall.Header().Add("X-Goog-User-Project", nodePoolInfo.project)
-				}
-				op, err := clusterNodePoolsUpdateCall.Do()
-				if err != nil {
-					return err
-				}
-
-				// Wait until it's updated
-				return ContainerOperationWait(config, op, nodePoolInfo.project, nodePoolInfo.location,
-					"updating GKE node pool kubelet_config", userAgent, timeout)
-			}
-
-			if err := retryWhileIncompatibleOperation(timeout, npLockKey, updateF); err != nil {
-				return err
-			}
-
-			log.Printf("[INFO] GKE cluster %s: kubelet_config updated", d.Id())
-		}
-
-		if d.HasChange("node_config.0.gcfs_config") {
-
-			defaultPool := "default-pool"
-
-			timeout := d.Timeout(schema.TimeoutCreate)
-
-			nodePoolInfo, err := extractNodePoolInformationFromCluster(d, config, clusterName)
-			if err != nil {
-				return err
-			}
-
-			// Acquire write-lock on nodepool.
-			npLockKey := nodePoolInfo.nodePoolLockKey(defaultPool)
-
-			gcfsEnabled := d.Get("node_config.0.gcfs_config.0.enabled").(bool)
-
-			// While we're getting the value from the drepcated field in
-			// node_config.kubelet_config, the actual setting that needs to be updated
-			// is on the default nodepool.
-			req := &container.UpdateNodePoolRequest{
-				Name: defaultPool,
-				GcfsConfig: &container.GcfsConfig{
-					Enabled: gcfsEnabled,
-				},
-			}
-
-			updateF := func() error {
-				clusterNodePoolsUpdateCall := config.NewContainerClient(userAgent).Projects.Locations.Clusters.NodePools.Update(nodePoolInfo.fullyQualifiedName(defaultPool), req)
-				if config.UserProjectOverride {
-					clusterNodePoolsUpdateCall.Header().Add("X-Goog-User-Project", nodePoolInfo.project)
-				}
-				op, err := clusterNodePoolsUpdateCall.Do()
-				if err != nil {
-					return err
-				}
-
-				// Wait until it's updated
-				return ContainerOperationWait(config, op, nodePoolInfo.project, nodePoolInfo.location,
-					"updating GKE node pool gcfs_config", userAgent, timeout)
-			}
-
-			if err := retryWhileIncompatibleOperation(timeout, npLockKey, updateF); err != nil {
-				return err
-			}
-
-			log.Printf("[INFO] GKE cluster %s: %s setting for gcfs_config updated to %t", d.Id(), defaultPool, gcfsEnabled)
-		}
-
+		nodePoolNodeConfigUpdate(d, config, nodePoolInfo, "", defaultPool, d.Timeout(schema.TimeoutUpdate))
 	}
 
 	if d.HasChange("notification_config") {
--- a/mmv1/third_party/terraform/services/container/resource_container_cluster.go.tmpl
+++ b/mmv1/third_party/terraform/services/container/resource_container_cluster.go.tmpl
@@ -104,12 +104,6 @@ var (
 		"private_cluster_config.0.master_global_access_config",
 	}
 
-	forceNewClusterNodeConfigFields = []string{
-		"labels",
-		"workload_metadata_config",
-		"resource_manager_tags",
-	}
-
 	suppressDiffForAutopilot = schema.SchemaDiffSuppressFunc(func(k, oldValue, newValue string, d *schema.ResourceData) bool {
 		if v, _ := d.Get("enable_autopilot").(bool); v {
 			return true
@@ -125,19 +119,6 @@ var (
 		return false
 	})
 )
-
-// This uses the node pool nodeConfig schema but sets
-// node-pool-only updatable fields to ForceNew
-func clusterSchemaNodeConfig() *schema.Schema {
-	nodeConfigSch := schemaNodeConfig()
-	schemaMap := nodeConfigSch.Elem.(*schema.Resource).Schema
-	for _, k := range forceNewClusterNodeConfigFields {
-		if sch, ok := schemaMap[k]; ok {
-			tpgresource.ChangeFieldSchemaToForceNew(sch)
-		}
-	}
-	return nodeConfigSch
-}
 
 // Defines default nodel pool settings for the entire cluster. These settings are
 // overridden if specified on the specific NodePool object.
@@ -1477,7 +1458,7 @@ func ResourceContainerCluster() *schema.Resource {
 				},
 			},
 
-			"node_config": clusterSchemaNodeConfig(),
+			"node_config": schemaNodeConfig(),
 
 			"node_pool": {
 				Type:     schema.TypeList,
--- a/mmv1/third_party/terraform/services/container/node_config.go.tmpl
+++ b/mmv1/third_party/terraform/services/container/node_config.go.tmpl
@@ -2161,7 +2161,7 @@ func nodePoolNodeConfigUpdate(d *schema.ResourceData, config *transport_tpg.Conf
 				req.Tags = ntags
 			}
 
-		    // sets tags to the empty list when user removes a previously defined list of tags entriely
+			// sets tags to the empty list when user removes a previously defined list of tags entriely
 			// aka the node pool goes from having tags to no longer having any
 			if req.Tags == nil {
 				tags := []string{}
--- a/mmv1/third_party/terraform/services/container/resource_container_cluster_migratev1.go.tmpl
+++ b/mmv1/third_party/terraform/services/container/resource_container_cluster_migratev1.go.tmpl
@@ -1198,7 +1198,7 @@ func resourceContainerClusterResourceV1() *schema.Resource {
 				},
 			},
 
-			"node_config": clusterSchemaNodeConfig(),
+			"node_config": schemaNodeConfig(),
 
 			"node_pool": {
 				Type:     schema.TypeList,
--- a/mmv1/third_party/terraform/services/container/resource_container_cluster_test.go.tmpl
+++ b/mmv1/third_party/terraform/services/container/resource_container_cluster_test.go.tmpl
@@ -11071,7 +11071,6 @@ resource "google_container_cluster" "primary" {
   initial_node_count = 1
 
   node_config {
-    machine_type = "n1-standard-1"  // can't be e2 because of local-ssd
     disk_size_gb = 15
 
     resource_manager_tags = {
--- a/mmv1/third_party/terraform/services/container/resource_container_cluster_test.go.tmpl
+++ b/mmv1/third_party/terraform/services/container/resource_container_cluster_test.go.tmpl
@@ -68,6 +68,10 @@ func TestAccContainerCluster_resourceManagerTags(t *testing.T) {
 	networkName := acctest.BootstrapSharedTestNetwork(t, "gke-cluster")
 	subnetworkName := acctest.BootstrapSubnet(t, "gke-cluster", networkName)
 
+	if acctest.BootstrapPSARole(t, "service-", "container-engine-robot", "roles/resourcemanager.tagHoldAdmin") {
+		t.Fatal("Stopping the test because a role was added to the policy.")
+	}
+
 	acctest.VcrTest(t, resource.TestCase{
 		PreCheck:                 func() { acctest.AccTestPreCheck(t) },
 		ProtoV5ProviderFactories: acctest.ProtoV5ProviderFactories(t),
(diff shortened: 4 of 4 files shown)


roaks3 (commented):
I've kicked off tests, but @wyardley this is a little tough to understand out of context, and seems potentially dangerous. Could you possibly either:
1. Describe what you've actually changed in the code, and why that solves the issue, or
2. Loop in someone else who may have prior context on these changes

wyardley:
@roaks3 Sure!

1. The `node_config` subfield of `google_container_cluster` controls the settings for the default node-pool when remove_default_node_pool is not set to false. The linked issue mentions how many of these fields previously didn't support updates when they should, causing an inability to control those settings (which would be updatable if they were on a standalone nodepool). It's actually the exact same resource. As suggested, I laid the foundation for this change in #11826 by moving the code for when `node_cofig` in the `container_node_pool` resource has changes into `node_config` into a standalone function. #11272 also has some background / context. In theory, the tests for updates on some of these fields in the node pool resource should behave the same way, and some of the existing tests (including `TestAccContainerCluster_withNodeConfigKubeletConfigSettingsUpdates and, I believe, `TestAccContainerCluster_withNodeConfigKubeletConfigSettingsUpdates` should also cover this scenario), but I'm happy to add or update some existing tests (within reason) if that helps create more confidence.
2. Have you read through the linked issues / PRs? Those have got a lot of discussion that hopefully outlines both what the issue is, and some other PRs that have addressed part of this issue. Some of the folks at Google who have been involved in previous discussions / PRs related to this are @rileykarson and @melinath, who probably have the most context here. I've linked to some related issues directly in the description now as well under "References".

So, tl;dr this should make the behavior of the default node pool within the `container_cluster` resource more consistent with the behavior of independently defined or additional node pools, and fix some cases where an existing setting for `default-pool` could not be updated in-place with failure / permadiff.

A bit broader context: `google_container_cluster` having the ability to define a node pool resource within the cluster resource (or outside of it) has been a persistent cause of headaches with this resource (one reason that many people choose to delete the default pool and manage nodepools separately).

mmv1/third_party/terraform/services/container/resource_container_cluster.go.tmpl
```diff
@@ -3796,133 +3796,15 @@ func resourceContainerClusterUpdate(d *schema.ResourceData, meta interface{}) er
...
-				return err
-			}
-
-			log.Printf("[INFO] GKE cluster %s: %s setting for gcfs_config updated to %t", d.Id(), defaultPool, gcfsEnabled)
+		nodePoolInfo, err := extractNodePoolInformationFromCluster(d, config, clusterName)
+		if err != nil {
+			return err
 		}
 
+		nodePoolNodeConfigUpdate(d, config, nodePoolInfo, "", defaultPool, d.Timeout(schema.TimeoutUpdate))
```
wyardley, line 3807: happy to adjust this to either define `timeout` above and / or hard-code `"default-pool"` vs defining it above

roaks3:
@wyardley thanks for the response!

Yea, I think I understand the intent to update this config, and some of the nuance where this is shared between resources. But I am still wondering if you could give a brief description of the actual code changes here. (I could probably deduce the answer by reading all of the related references, but it would take quite a bit longer for me to wrap my head around everything for this review). For example, why are all of these lines being removed? Are they effectively already copied under the function you are replacing them with, or do I need to review those functions as well to determine what behavior has changed? In short, what could break?

Acknowledged on the testing. Once I know a bit more about what is changing that will help me weigh in, but I'm tempted to say this is the type of change that would be covered well enough by existing tests. We may want to add some of these fields to existing tests though, if they are any that aren't already included.

wyardley:
> Yea, I think I understand the intent to update this config, and some of the nuance where this is shared between resources. But I am still wondering if you could give a brief description of the actual code changes here.

So the bug here is that not all of the fields that support update in place were being updated when defined for the default pool via `node_config`. _Some_ of the code, such as the fields I added in #11272 had the update implemented and had tests for it (because @melinath asked me to add them). In those cases, they would have been separately implemented in both places already, but in other cases, the code was already missing from one place.

The plan therefore was to first extract out the node_config updates in `container_node_pool` to a separate function as one step, and then call that same function for `node_config` within `container_cluster`.

So the simple answer to your question is that the code being pulled out should be a functionally equivalent subset of what's now being called. Some of the code should be identical or close to identical (other than the semantics around `prefix` being a little different and so on) to the code in `nodePoolNodeConfigUpdate()` that's now being called (e.g., `node_config.0.image_type`, `node_config.0.kubelet_config`, which I fixed in #11697 after adding part of it for #11272, and `gcfs_config`, fixed in #11717) , whereas some of the code (essentially, missing code to do the remainder of the updates) was simply not present here previously (e.g., `node_config.0.tags`, `node_config.0.resource_manager_tags`, the block with `disk_size_gb`, `disk_type`, etc.), which is the part that's being fixed. But the actual "thing" being operated on (a node pool) is the same in both cases.

I checked and when the default node pool is defined via `node_pool {}` within `container_cluster`, that already functions as it should.

`linux_node_config` and the disk size etc. ones may need to be double-checked (I can play with it a little), though I think this code should handle it correctly too - in my understanding, if updating it for externally defined nodepools works, it should also work for the default pool. And, to whatever extent these things are tested under `container_node_pool`, my understanding is that those tests should more or less cover their use here.

However, some of these resources may not have tests in either place that test the path where the resource is updated in-place.

melinath:
Drive-by comment: @wyardley _will_ existing tests catch this error? or is this fixing bugs that were not caught by the existing tests?

wyardley:
@melinath: There were presumably no (update) tests for the things that were broken before, at least not directly within `container_cluster.node_config`, since they’d be failing. I’m not sure how comprehensive the tests under `container_node_pool.node_config` are - I’d guess there are missing ones there as well. But in any cases where they do exist, presumably testing one should test the other once they’re using the same code.

I ran through the tests that do have update tests, but agree there are probably gaps.

wyardley:
Ok @roaks3 @melinath: tried to do a quick audit of existing test coverage. It's kind of a patchwork, though actually a reasonable amount of coverage already, esp. if we assume the same behavior between the default pool and other nodepools once this change is in place.

I'm adding a simple tweak to an existing test that will already cover `tags` and `labels` for `container_cluster.node_config`.

Here's what I found (may not be perfect)
* `logging_variant` update tests: `TestAccContainerCluster_withLoggingVariantUpdates` and `TestAccContainerNodePool_withLoggingVariantUpdates`
* `disk_size_gb`, `disk_type`, `machine_type`, `storage_pools`: Seems like `TestAccContainerNodePool_withMachineDiskStoragePoolsUpdate` tests this? And `TestAccContainerCluster_storagePoolsWithNodePool` and `TestAccContainerNodePool_storagePools` have `storage_pools` parameterized but only called once. Added a change here to `TestAccContainerCluster_withNodeConfig` for `disk_size_gb` in another commit, and probably would be pretty easy to change one or two of the other ones here too).
* `taint`: Seems to be tested in `TestAccContainerCluster_withNodeConfig` (calls config `testAccContainerCluster_withNodeConfig` and then `testAccContainerCluster_withNodeConfigUpdate`), and `testAccContainerNodePool_withTaintsUpdate`.
* `tags`: Not sure updates are tested in either place, but could be missing it? Adding a change in `TestAccContainerCluster_withNodeConfig`
* `resource_manager_tags`: `TestAccContainerCluster_resourceManagerTags` exists but doesn't test updates; `TestAccContainerNodePool_resourceManagerTags` does test updates
* `resource_labels: Looks like this is tested in `TestAccContainerCluster_misc` and `testAccContainerNodePool_withNodeConfigUpdate` 
* `labels`: `TestAccContainerCluster_withNodeConfig` (updates in config `testAccContainerCluster_withNodeConfigUpdate`) - I think labels didn't change, but I'm pushing a commit so that it does change. This is a pretty good test where we could test some other small / generic stuff pretty easily since it's just two different configs vs. being parameterized.
* `image_type`: Update covered in `testAccContainerNodePool_withNodeConfigUpdate` and `TestAccContainerCluster_withNodeConfig`
* `workload_metadata_config`: Update tested in `TestAccContainerNodePool_withWorkloadIdentityConfig`. Don't believe update is / was tested in `container_cluster`
* `gcfs_config`:  (already covered, see above)
* `kubelet_config`: (already covered, see above)
* `linux_node_config`: Updates tested in `TestAccContainerNodePool_withLinuxNodeConfig`
* `fast_socket`: Updates tested in `TestAccContainerNodePool_fastSocket`

wyardley:
I was able to extend `TestAccContainerCluster_withNodeConfig` to include a bunch of fields that previously would have failed on update before like `taint`, `machine_type`, `disk_type`, `disk_size_gb`, `tags`, and `labels`, which dramatically reduces any testing gaps with only minimal added / modified code. I also added a `ConfigPlanChecks` block.

With the changes, things pass, and, even with all the separate node pool updates, it's not horribly fast:
```
--- PASS: TestAccContainerCluster_withNodeConfig (808.25s)
PASS
ok  	github.com/hashicorp/terraform-provider-google/google/services/container	809.408s
```

The same test fails as expected without the code changes.

Some of the fields (`labels`, `workload_metadata_config`, `resource_manager_tags`) were originally marked as fields to force recreation, presumably _because_ of the lack of code to update those fields; I thought about trying to remove that separately, either before or after, but I think it's all intertwined, so I suggest we pull this stuff out as well as part of these changes, and then maybe add a second release note if necessary:
https://github.com/GoogleCloudPlatform/magic-modules/blob/3f06b2ed38e474ba08cca2d8553d29a741ec1810/mmv1/third_party/terraform/services/container/resource_container_cluster.go.tmpl#L107C2-L110

This is a good example where the tests directly for `container_cluster` and adding the plancheck were useful and (assuming this change is safe and makes sense) will also help make everything more consistent.

wyardley:
I've added / adjusted tests for almost all of the fields directly in `google_container_cluster` (see above for more detailed notes on where tests were added / adjusted). Since the gvnic setting is required, and since there are so many different ones all checked in the same test for `TestAccContainerCluster_withNodeConfig`, I did the `fast_socket` tests in a new test.

I can add `linux_node_config` tests if desired as well, though the existing tests do seem to confirm that the behavior is consistent between node pool and cluster defined node pool.

All of these passed when I ran them manually; would be great if someone can kick off the full suite again Monday.

I left the commits mostly atomic for easier review, but let me know if I should squash something.

```
--- PASS: TestAccContainerCluster_resourceManagerTags (763.72s)
--- PASS: TestAccContainerCluster_withNodeConfig (776.55s)
--- PASS: TestAccContainerCluster_withWorkloadMetadataConfig (963.77s)
--- PASS: TestAccContainerCluster_withLoggingVariantUpdates (570.99s)
--- PASS: TestAccContainerCluster_withNodeConfigFastSocket (846.52s)
--- PASS: TestAccContainerCluster_withNodeConfigKubeletConfigSettingsUpdates (1280.76s)
--- PASS: TestAccContainerCluster_misc (1257.44s)
```

mmv1/third_party/terraform/services/container/resource_container_cluster.go.tmpl
```diff
@@ -104,12 +104,6 @@ var (
 		"private_cluster_config.0.master_global_access_config",
 	}
 
-	forceNewClusterNodeConfigFields = []string{
-		"labels",
-		"workload_metadata_config",
-		"resource_manager_tags",
```
wyardley, line 110: If this list had been perfectly maintained / up to date, I think this "bug" wouldn't have existed [because the default pool would have gotten recreated if any of those fields changed), but also, it seems better to update the fields in-place if we can, which it seems like we can, and I'm pretty confident at least some fields were missing from this list.

wyardley:
@roaks3 @rileykarson @melinath: I added `Fixes xxx` (or partial / possible fix for cases I wasn't sure about) for some additional issues that I believe will be fixed by this; feel free to add / remove in there as makes sense.

roaks3:
Thanks @wyardley . I still have one last question, but I think I'm understanding things:

1. Previously mutable fields

- `image_type`
- `kubelet_config`
- `gcfs_config`

Tests seem good (looks like they were all covered already), but when I looked at the implementation it didn't look 1-to-1. For example, it seems like `kubelet_config` would now use ForceSendFields when it is nil. Sorry if I missed it, but are we accounting for potential changes to the behavior of these fields?

2. Previously mutable fields that didn't work

- `logging_variant`
- `disk_size_gb`
- `disk_type`
- `machine_type`
- `storage_pools`
- `taint`
- `tags`
- `resource_labels`
- `linux_node_config`
- `fast_socket`

It seems to me like tests are good and you've filled any gaps. Behavior was previously broken, and this is the focus of the bug fix.

3. Previously immutable fields

- `labels`
- `workload_metadata_config`
- `resource_manager_tags`

Tests seem good. These are enhancements that are included because "why not", since we are already adding in-place update behavior.

rileykarson:
> For example, it seems like kubelet_config would now use ForceSendFields when it is nil.

Oh, that may be integrating https://github.com/GoogleCloudPlatform/magic-modules/pull/11978 FWIW

wyardley:
Couple other things:

* can someone review the list of fixed bugs in the description to make sure it looks good?
* I think there were some recent changes to tests. There’s no conflicts, so will those get pulled in if tests are rerun via an extra merge step, or should I rebase?

roaks3:
Awesome, thanks!

I also just double checked the tests with what you stated in a previous comment, and everything looked good except the missing `linux_node_config` that you mentioned. The existing node pool test is probably sufficient honestly, but since we've already got all of the others tested explicitly for the container cluster, I would actually recommend adding that one too. It might be a bit surprising/confusing otherwise, and the explicit tests are nice should anything need to diverge.

roaks3:
Oh and yes, I'm reviewing the bugs you listed next, and I don't think you need to rebase.

wyardley:
Is there a specific subfield I should test, or should I combine? If you look at the comparable tests in `container_node_pool`, there are several, and they’re not super trivial to replicate; though I can take a pass.

roaks3:
I would recommend a minimal `linux_node_config` update test to verify it can be updated successfully. So maybe just a single subfield.

Reviewed related issues and added some notes, but overall they look good.

I wanted to ask about https://github.com/hashicorp/terraform-provider-google/issues/19056, which mentions `containerd_config`. Is that related to your fix here? I couldn't find update logic for this field being added or altered with `nodePoolNodeConfigUpdate()`.

wyardley:
> I wanted to ask about https://github.com/hashicorp/terraform-provider-google/issues/19056, which mentions containerd_config. Is that related to your fix here? I couldn't find update logic for this field being added or altered with nodePoolNodeConfigUpdate().

@roaks3 interesting - you're right. I'll comment in that issue.

wyardley:
Any context on the missing resource test warning? There is a test that tests this in the PR (interestingly, it's not complaining about `linux_node_config`, which _isn't_ tested).

IIRC, that test passed for me, and I _think_ the VCR tests getting as far as they did means the build passed, but sometimes this warning is the result of something that is failing with the build.

wyardley:
I am working on the `linux_node_config` test and testing that locally, and will also test `TestAccContainerCluster_resourceManagerTags` again to see what's going on there, but I guess that explains the test coverage error.

wyardley:
@roaks3 do you think maybe it needs something like this in the `TestAccContainerCluster_resourceManagerTags` test, especially since the terraform code is invoked multiple times?

I don't totally get how this works still, because in my experience, the test seems to clean up the binding when it's done, so wouldn't this fail every time? I see it used in other tests, though, so assuming it will work?

```diff
diff --git a/mmv1/third_party/terraform/services/container/resource_container_cluster_test.go.tmpl b/mmv1/third_party/terraform/services/container/resource_container_cluster_test.go.tmpl
index 9b54afebd..482695b3d 100644
--- a/mmv1/third_party/terraform/services/container/resource_container_cluster_test.go.tmpl
+++ b/mmv1/third_party/terraform/services/container/resource_container_cluster_test.go.tmpl
@@ -68,6 +68,10 @@ func TestAccContainerCluster_resourceManagerTags(t *testing.T) {
 	networkName := acctest.BootstrapSharedTestNetwork(t, "gke-cluster")
 	subnetworkName := acctest.BootstrapSubnet(t, "gke-cluster", networkName)
 
+	if acctest.BootstrapPSARole(t, "service-", "container-engine-robot", "roles/resourcemanager.tagHoldAdmin") {
+		t.Fatal("Stopping the test because a role was added to the policy.")
+	}
+
 	acctest.VcrTest(t, resource.TestCase{
 		PreCheck:                 func() { acctest.AccTestPreCheck(t) },
 		ProtoV5ProviderFactories: acctest.ProtoV5ProviderFactories(t),
```

Or was the failure in CI something else besides a permissions issue?

wyardley:
```
--- PASS: TestAccContainerCluster_resourceManagerTags (829.99s)
PASS
ok  	github.com/hashicorp/terraform-provider-google/google/services/container	831.174s
```

My runs that didn't hit a stockout condition seem to pass, so I guess we could do some combination of these things, depending on what you're seeing:

1. Switch to a different instance type than `n1`
2. Switch to a different zone (`-f`, let's say) vs `-a`
3. Add the bootstrap function above, which I think is a good idea anyway? I've pushed that up, but let me know if you want me to remove it.

roaks3:
Ok, let's see what the tests do this time. The error I saw looked like it might have to do with tag values changing, but we'll see.

The new test LGTM, and once we get the tests to pass I think this will be ready to merge.

wyardley:
> The error I saw looked like it might have to do with tag values changing, but we'll see.

Got it. Interesting... other than stockout issues, has passed for me locally pretty consistently. If you see it again, can you post the exact error here?

wyardley:
Ah, I see... it's passing in recording mode, but failing in replaying mode. Maybe something about going from:
```
google_tags_tag_key.key1.id) = google_tags_tag_value.value1.id
```
to
```
google_tags_tag_key.key2.id) = google_tags_tag_value.value2.id
```
(which I just did to avoid having to do more complicated stuff to update the tags) is causing an issue?

I'm going to run `TestAccContainerCluster_withNodeConfigLinuxNodeConfig` one more time locally as well, but let me know the errors you're seeing for both of those.

roaks3:
Well, it looks like `TestAccContainerCluster_resourceManagerTags` is having trouble because it is using the bootstrapped permission now, which I think behaves weird on the first run or two.

The original error was this (which might be intermittent):
```
    resource_container_cluster_test.go:73: Step 3/4 error: After applying this test step, the plan was not empty.
        stdout:
        
        
          ~ update in-place
        
          # google_container_cluster.primary will be updated in-place
          ~ resource "google_container_cluster" "primary" {
                id                                       = "projects/ci-test-project-188019/locations/us-central1-a/clusters/tf-test-cluster-38yqn0ooat"
                name                                     = "tf-test-cluster-38yqn0ooat"
                # (36 unchanged attributes hidden)
        
              ~ node_config {
                  ~ resource_manager_tags       = {
                      + "tagKeys/281482499338282" = "tagValues/281480833535742"
                      - "tagKeys/281484326142515" = "tagValues/281477968952549" -> null
                    }
                    tags                        = []
                    # (19 unchanged attributes hidden)
        
                    # (1 unchanged block hidden)
                }
        
                # (21 unchanged blocks hidden)
            }
        
```

The `TestAccContainerCluster_withNodeConfigLinuxNodeConfig` error is:
```
resource_container_cluster_test.go:1573: Step 3/6 error: Check failed: Check 1/1 error: Not found: google_container_cluster.with_node_pool in [root]
```

wyardley:
> resource_container_cluster_test.go:1573: Step 3/6 error: Check failed: Check 1/1 error: Not found: google_container_cluster.with_node_pool

Thanks, that's a typo that I thought I'd fixed... will fix that one. Not sure how it was working for me.

wyardley:
* Tested and then pushed up a fix for `TestAccContainerCluster_withNodeConfigLinuxNodeConfig`. Also, `UNSPECIFIED` seems to be the default so I went from unset => v2 => v1, and checked the latter two explicitly.
* Tried a run in `RECORDING` mode and a couple of runs in `REPLAYING` mode locally (and then another run in `RECORDING` mode) for `TestAccContainerCluster_resourceManagerTags`, and seems to be Ok.

wyardley:
@roaks3 any idea why it didn’t re-run `TestAccContainerCluster_resourceManagerTags` this time or why it’s still complaining about the missing test?

roaks3:
Yea, `TestAccContainerCluster_resourceManagerTags` was recorded properly, but didn't pass replay earlier because it was waiting for the bootstrap permissions to be fully available. On subsequent replay, the permissions were correct and it succeeded.

I'm not sure on the missing test detector. It is reporting a correct requirement, but we also clearly have it covered in a test. It might be hitting some edge case it isn't expecting.

roaks3 (approved):

PR #12016, opened 2024-10-16
Feat add oracledatabase_autonomous_database datasource

tulika-aakriti:
Part of - https://github.com/hashicorp/terraform-provider-google/issues/19851

```release-note:new-datasource
`google_oracle_database_autonomous_database`
```


mmv1/third_party/terraform/services/oracledatabase/data_source_oracle_database_autonomous_database.go
```diff
@@ -0,0 +1,36 @@
...
+}
+
+func dataSourceOracleDatabaseAutonomousDatabaseRead(d *schema.ResourceData, meta interface{}) error {
+	config := meta.(*transport_tpg.Config)
+
+	id, err := tpgresource.ReplaceVars(d, config, "projects/{{project}}/locations/{{location}}/autonomousDatabases/{{autonomous_database_id}}")
+	if err != nil {
+		return fmt.Errorf("Error constructing id: %s", err)
+	}
+	err = resourceOracleDatabaseAutonomousDatabaseRead(d, meta)
```
rileykarson, line 29: `SetId` before read, similar to https://github.com/GoogleCloudPlatform/magic-modules/blob/72b43deb6fe6a0678ebbc5bc189a6733bdf4e1f2/mmv1/third_party/terraform/services/compute/data_source_google_compute_region_ssl_certificate.go

mmv1/third_party/terraform/services/oracledatabase/data_source_oracle_database_autonomous_database.go
```diff
@@ -0,0 +1,36 @@
...
+	if err != nil {
+		return err
+	}
+	d.SetId(id)
```
rileykarson, line 33: After Read, return an error if `id` is null. That indicates we've hit a not found error- Read will return an error if we hit an unexpected 403 or 5XX, but a 404 (or 403 for some resources) will succeed, returning a resource with no id, flagging that Terraform should clear it from state.

This block is standard: https://github.com/GoogleCloudPlatform/magic-modules/blob/72b43deb6fe6a0678ebbc5bc189a6733bdf4e1f2/mmv1/third_party/terraform/services/compute/data_source_google_compute_region_ssl_certificate.go#L44-L46

rileykarson (commented):
What template or existing datasource did you reference? I think we need to restructure this a bit to account for https://github.com/hashicorp/terraform-provider-google/issues/12873 (and also correct the source used here, so we don't proliferate more variations that aren't consistent with our intended behaviour)

rileykarson (approved):

rileykarson:
Merging is blocked temporarily- if I don't merge this by EOD today please feel free to ping me

tulika-aakriti:
> What template or existing datasource did you reference? I think we need to restructure this a bit to account for [hashicorp/terraform-provider-google#12873](https://github.com/hashicorp/terraform-provider-google/issues/12873) (and also correct the source used here, so we don't proliferate more variations that aren't consistent with our intended behaviour)

I can't remember that but I can mostly all list API i.e datasources have similar thing.

PR #12017, opened 2024-10-16
Add support for global application in Apphub

eshagoel06:
Add support for global application and validation for location and scope type
FIXES https://github.com/hashicorp/terraform-provider-google/issues/19252

```release-note:enhancement
apphub: added `GLOBAL` enum value to `scope.type` field in `google_apphub_application` resource
```


mmv1/templates/terraform/constants/apphub_application.go.tmpl
```diff
@@ -0,0 +1,17 @@
+func apphubApplicationCustomizeDiff(_ context.Context, diff *schema.ResourceDiff, meta interface{}) error {
+    if diff.HasChange("location") || diff.HasChange("scope.0.type") {
+        location := diff.Get("location")
+        scope_type := diff.Get("scope.0.type")
+
+        if scope_type == "GLOBAL" {
+            if location != "global" {
```
c2thorn, line 7: Would the API give an appropriate error in this case? If so, is there a particular pressure to duplicate this validation in the Terraform provider?

Adding additional validation at this level can lead to instances where the API requirements change but the Terraform providers lag behind.

mmv1/templates/terraform/constants/apphub_application.go.tmpl
```diff
@@ -0,0 +1,17 @@
...
+                return fmt.Errorf("Error validating location %s with %s scope type", location, scope_type)
+            }
+        } else {
+            if location == "global" {
+                return fmt.Errorf("Error validating location %s with %s scope type", location, scope_type)
+            }
+        }
+    }
+    return nil
+}
```
c2thorn, line 17: we should cover this logic somewhere in a test. If we are going to keep it (see my previous comment), let's add an invalid config in https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/third_party/terraform/services/apphub/resource_apphub_application_test.go and use [ExpectError](https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/third_party/terraform/services/servicenetworking/resource_service_networking_connection_test.go#L83)

mmv1/products/apphub/Application.yaml
```diff
@@ -43,12 +43,30 @@ async:
...
     primary_resource_id: 'example'
     vars:
       application_id: 'example-application'
+      location: 'us-east1'
+      scope_type: 'REGIONAL'
+    test_vars_overrides:
+      'location': '"us-east1"'
+      'scope_type': '"REGIONAL"'
+  - name: 'application_global_basic'
+    config_path: 'templates/terraform/examples/apphub_application_basic.tf.tmpl'
```
c2thorn, line 61: is config_path actually necessary here? This is usually just used in for private overrides. I'm assuming other examples of using this were just accidentally copied over from the private repo

c2thorn (changes requested):

mmv1/products/apphub/Application.yaml
eshagoel06: Have updated basic and full examples to remove config_path. But for *_global_basic , we were intending to use the same config as basic with different values.

mmv1/templates/terraform/constants/apphub_application.go.tmpl
eshagoel06: Done

c2thorn (approved):

PR #12019, opened 2024-10-16
Add api_resource_type for mmv1 resources that need the name to be overridden

roaks3:
This PR adds support for a new resource field called `api_resource_type` that specifies the canonical name of the resource being covered (when it does not already match `name`). It contains only name overrides that could be confirmed with our primary internal tool.

No behavior is being changed in the provider, so the value in review is primarily for spot-checking that the name overrides look reasonable (in addition to any other feedback about the approach).

```release-note:none
provider: added `api_resource_type` overrides to mmv1 resources
```


mmv1/api/resource.go
```diff
@@ -328,6 +328,8 @@ type Resource struct {
 	Compiler string
 
 	ImportPath string
+
+	ApiResourceType string `yaml:"api_resource_type"`
```
melinath, line 332: Double-checking language used other places... we could consider calling this `api_resource_type_kind` if it's just the final piece of the name (since `api_resource_type` could be confused with the full type that includes the domain)

melinath (approved):
One note below but happy to have it merged as-is if you prefer the current name.

mmv1/api/resource.go
roaks3: Agreed, will change

PR #12024, opened 2024-10-16
Magician vcr eap cmd

trodge:
This PR adds the command for running VCR in eap using the magician.

Merge after https://github.com/GoogleCloudPlatform/magic-modules/pull/11874

```release-note:none

```

### DIFF
--- a/.ci/magician/cmd/test_eap_vcr.go
+++ b/.ci/magician/cmd/test_eap_vcr.go
@@ -1,232 +1,112 @@
 package cmd
 
 import (
+	_ "embed"
 	"fmt"
+	"magician/exec"
+	"magician/provider"
+	"magician/vcr"
 	"os"
 	"path/filepath"
-	"regexp"
 	"sort"
 	"strings"
-	"text/template"
 
 	"github.com/spf13/cobra"
-
-	"magician/exec"
-	"magician/github"
-	"magician/provider"
-	"magician/source"
-	"magician/vcr"
-
-	_ "embed"
 )
 
-var (
-	//go:embed templates/vcr/test_analytics.tmpl
-	testsAnalyticsTmplText string
-	//go:embed templates/vcr/non_exercised_tests.tmpl
-	nonExercisedTestsTmplText string
-	//go:embed templates/vcr/with_replay_failed_tests.tmpl
-	withReplayFailedTestsTmplText string
-	//go:embed templates/vcr/without_replay_failed_tests.tmpl
-	withoutReplayFailedTestsTmplText string
-	//go:embed templates/vcr/record_replay.tmpl
-	recordReplayTmplText string
-)
-
-var ttvEnvironmentVariables = [...]string{
+var tevEnvironmentVariables = [...]string{
+	"GEN_PATH",
 	"GOCACHE",
 	"GOPATH",
-	"GOOGLE_BILLING_ACCOUNT",
-	"GOOGLE_CUST_ID",
-	"GOOGLE_IDENTITY_USER",
-	"GOOGLE_MASTER_BILLING_ACCOUNT",
-	"GOOGLE_ORG",
-	"GOOGLE_ORG_2",
-	"GOOGLE_ORG_DOMAIN",
-	"GOOGLE_PROJECT",
-	"GOOGLE_PROJECT_NUMBER",
 	"GOOGLE_REGION",
-	"GOOGLE_SERVICE_ACCOUNT",
-	"GOOGLE_PUBLIC_AVERTISED_PREFIX_DESCRIPTION",
 	"GOOGLE_ZONE",
+	"ORG_ID",
+	"GOOGLE_PROJECT",
+	"GOOGLE_BILLING_ACCOUNT",
+	"GOOGLE_ORG",
+	"GOOGLE_ORG_DOMAIN",
+	"GOOGLE_PROJECT_NUMBER",
+	"GOOGLE_USE_DEFAULT_CREDENTIALS",
+	"GOOGLE_IMPERSONATE_SERVICE_ACCOUNT",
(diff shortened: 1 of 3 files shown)


.ci/magician/cmd/test_eap_vcr.go
```diff
@@ -0,0 +1,250 @@
...
+	"magician/vcr"
+	"os"
+	"path/filepath"
+	"sort"
+	"strings"
+
+	"github.com/spf13/cobra"
+)
+
+var tevEnvironmentVariables = [...]string{
```
roaks3, line 17: I think we want to keep all of the env vars, even though they aren't set up yet for EAP, so that we don't need to update this later. Namely:
- "GOOGLE_CUST_ID"
- "GOOGLE_IDENTITY_USER"
- "GOOGLE_MASTER_BILLING_ACCOUNT"
- "GOOGLE_ORG_2"
- and maybe "GOOGLE_PUBLIC_AVERTISED_PREFIX_DESCRIPTION"

.ci/magician/cmd/test_eap_vcr.go
```diff
@@ -0,0 +1,250 @@
...
+
+func listTEVEnvironmentVariables() string {
+	var result string
+	for i, ev := range tevEnvironmentVariables {
+		result += fmt.Sprintf("\t%2d. %s\n", i+1, ev)
+	}
+	return result
+}
+
+func execTestEAPVCR(changeNumber, genPath, kokoroArtifactsDir, modifiedFilePath string, rnr ExecRunner, vt *vcr.Tester) error {
```
roaks3, line 78: Don't we need to upload cassettes somewhere? Or maybe I'm misunderstanding something.

.ci/magician/cmd/test_eap_vcr.go
```diff
@@ -0,0 +1,250 @@
...
+	}
+
+	services, runFullVCR := modifiedPackages(strings.Split(changedFiles, "\n"), provider.Private)
+	if len(services) == 0 && !runFullVCR {
+		fmt.Println("Skipping tests: No go files or test fixtures changed")
+		return nil
+	}
+	fmt.Println("Running tests: Go files or test fixtures changed")
+
+	head := "auto-cl-" + changeNumber
```
roaks3, line 96: Maybe just a note, but if we're matching the public providers, are we planning to automatically create branches with this name?

.ci/magician/cmd/test_eap_vcr.go
```diff
@@ -0,0 +1,250 @@
...
+	if err := vt.FetchCassettes(provider.Private, "main", head); err != nil {
+		return fmt.Errorf("error fetching cassettes: %w", err)
+	}
+	replayingResult, testDirs, replayingErr := runReplaying(runFullVCR, provider.Private, services, vt)
+	if err := vt.UploadLogs(vcr.UploadLogsOptions{
```
roaks3, line 101: Since the EAP jobs run within an environment where we have logging accessible the way we want, do we still need to upload logs the same way? I could definitely see individual debug logs per test being useful, but was not sure about other logging.

.ci/magician/cmd/test_eap_vcr.go
```diff
@@ -0,0 +1,250 @@
...
+		}
+		comment := strings.Join([]string{testsAnalyticsComment, withoutReplayFailedTestsComment}, "\n")
+		if err := postGerritComment(kokoroArtifactsDir, modifiedFilePath, comment, rnr); err != nil {
+			return fmt.Errorf("error posting comment: %w", err)
+		}
+	}
+	return nil
+}
+
+var allowedAlphaServices = map[string]struct{}{
```
roaks3, line 206: How/why were these services chosen? Also, we may actually want a disallowed list, so that we don't need to update this every time a new service is added to EAP.

roaks3 (commented):
Took a first pass and overall looked good (thanks for splitting out the commits)

.ci/magician/cmd/test_eap_vcr.go
trodge: I wasn't planning on it, just keeping the naming scheme to associate cassettes and logs with CLs.

.ci/magician/cmd/test_eap_vcr.go
trodge: Must have gotten removed by mistake. I'll add it back in.

.ci/magician/cmd/test_eap_vcr.go
trodge: I'd prefer to have the same upload logs function for both environments in addition to any logging done by Kokoro.

.ci/magician/cmd/test_eap_vcr.go
trodge: These are the services that have private overrides.

My plan was to remove this restriction altogether once the EAP environment is fully set up. I could also make a disallowed list, but it would be much longer.

roaks3: Yea, a significant percentage of the contributions in EAP use new services, so I think we need to support testing for those out-of-the-box. I would actually tweak this to be a disallow list for the services that we know or suspect have problems due to the test environment not being set up. But even then, I would be curious if we could just omit this entirely and let the tests run without the env vars, and that way when we update the test environment we don't need to make additional changes.

.ci/magician/cmd/test_eap_vcr.go
trodge: I'll add these as optional variables for now so that the command will still read them when they're available and warn when they're not.

roaks3 (approved):
LGTM. What was your plan for testing? Do we need to merge this first to be able to give it a try in EAP?

PR #12025, opened 2024-10-16
Add support for ControlPlaneEndpointsConfig to GKE

gontech:
Add support for ControlPlaneEndpointsConfig to GKE.

Supports launch of unified cluster experience, including consolidation of control plane related settings, allowing toggling private nodes which previously required cluster recreation, and enablement of a new DNS based endpoint for easier cluster access.

```release-note:enhancement
container: added `control_plane_endpoints_config`, `master_authorized_networks_config.private_endpoint_enforcement_enabled`, and `network_config` fields to `cluster` resource.
```

### DIFF
--- a/mmv1/third_party/terraform/services/container/resource_container_cluster.go.tmpl
+++ b/mmv1/third_party/terraform/services/container/resource_container_cluster.go.tmpl
@@ -54,7 +54,12 @@ var (
 				Type: schema.TypeBool,
 				Optional: true,
 				Computed: true,
-				Description: `Whether Kubernetes master is accessible via Google Compute Engine Public IPs.`,
+				Description: `Whether Kubernetes master is accessible via Google Compute Engine Public IPs. Defaults to false`,
+			},
+			"private_endpoint_enforcement_enabled": {
+				Type:        schema.TypeBool,
+				Optional:    true,
+				Description: `Whether master authorized networks is enforced on the private endpoint or not. Defaults to false.`,
 			},
 		},
 	}
@@ -1421,6 +1426,8 @@ func ResourceContainerCluster() *schema.Resource {
 				Type:        schema.TypeList,
 				Optional:    true,
 				Computed:    true,
+				Deprecated:    "use control_plane_endpoints_config.ip_endpoints_config.authorized_networks_config instead",
+				ConflictsWith: []string{"control_plane_endpoints_config.0.ip_endpoints_config.0.authorized_networks_config"},
 				MaxItems:    1,
 				Elem:        masterAuthorizedNetworksConfig,
 				Description: `The desired configuration options for master authorized networks. Omit the nested cidr_blocks attribute to disallow external access (except the cluster node IPs, which GKE automatically whitelists).`,
@@ -1722,13 +1729,128 @@ func ResourceContainerCluster() *schema.Resource {
 				ConflictsWith: []string{"enable_autopilot"},
 			},
 
+			"control_plane_endpoints_config": {
+				Type:        schema.TypeList,
+				MaxItems:    1,
+				Optional:    true,
+				Computed:    true,
+				ConflictsWith: []string{
+					"private_cluster_config.0.enable_private_endpoint",
+					"private_cluster_config.0.private_endpoint_subnetwork",
+					"private_cluster_config.0.master_global_access_config",
+				},
+				Description: `Configuration for all of the cluster's control plane endpoints. If this is specified, then respective configured values from private_cluster_config are ignored.`,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"dns_endpoint_config": {
+							Type:        schema.TypeList,
+							MaxItems:    1,
+							Optional:    true,
+							Computed:    true,
+							Description: `DNS endpoint configuration.`,
+							Elem: &schema.Resource{
+								Schema: map[string]*schema.Schema{
+									"endpoint": {
+										Type:        schema.TypeString,
+										Optional:    true,
+										Computed:    true,
+										Description: `The cluster's DNS endpoint.`,
+									},
+									"allow_external_traffic": {
+										Type:        schema.TypeBool,
+										Optional:    true,
+										Description: `Controls whether user traffic is allowed over this endpoint. Note that GCP-managed services may still use the endpoint even if this is false.`,
+									},
+								},
+							},
+						},
+						"ip_endpoints_config": {
+							Type:        schema.TypeList,
+							MaxItems:    1,
+							Optional:    true,
+							Computed:    true,
+							Description: `IP endpoints configuration.`,
+							Elem: &schema.Resource{
+								Schema: map[string]*schema.Schema{
+									"enabled": {
+										Type:        schema.TypeBool,
+										Optional:    true,
+										Description: `Controls whether to allow direct IP access. When this is false, all other ip_endpoints_config values are ignored.`,
+									},
+									"enable_public_endpoint": {
+										Type:             schema.TypeBool,
+										Optional:         true,
+										DiffSuppressFunc: suppressWhenIpEndpointsDisabled,
+										ConflictsWith:    []string{"private_cluster_config.0.enable_private_endpoint"},
+										Description:      `Controls whether the control plane allows access through a public IP.`,
+									},
+									"global_access": {
+										Type:             schema.TypeBool,
+										Optional:         true,
+										DiffSuppressFunc: suppressWhenIpEndpointsDisabled,
+										ConflictsWith:    []string{"private_cluster_config.0.master_global_access_config"},
+										Description:      `Controls whether the control plane's private endpoint is accessible from sources in other regions.`,
+									},
+									"authorized_networks_config": {
+										Type:             schema.TypeList,
+										Optional:         true,
+										DiffSuppressFunc: suppressWhenIpEndpointsDisabled,
+										ConflictsWith:    []string{"master_authorized_networks_config"},
+										MaxItems:         1,
+										Description:      `Configuration of authorized networks. If enabled, restricts access to the control plane based on source IP.`,
+										Elem:             masterAuthorizedNetworksConfig,
+									},
+									"public_endpoint": {
+										Type:        schema.TypeString,
+										Optional:    true,
+										Computed:    true,
+										Description: `The external IP address of this cluster's control plane. Only populated if enabled.`,
+									},
+									"private_endpoint": {
+										Type:        schema.TypeString,
+										Optional:    true,
+										Computed:    true,
+										Description: `The internal IP address of this cluster's control plane. Only populated if enabled.`,
+									},
+									"private_endpoint_subnetwork": {
+										Type:             schema.TypeString,
+										Optional:         true,
+										DiffSuppressFunc: suppressWhenIpEndpointsDisabled,
+										ConflictsWith:    []string{"private_cluster_config.0.private_endpoint_subnetwork"},
+										Description:      `Subnet to provision the master's private endpoint during cluster. Specified in projects/*/regions/*/subnetworks/* format.`,
+									},
+								},
+							},
+						},
+					},
+				},
+			},
+
+			"network_config": {
+				Type:        schema.TypeList,
+				MaxItems:    1,
+				Optional:    true,
+				Computed:    true,
+				Description: `Configuration for cluster networking.`,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"default_enable_private_nodes": {
+							Type:          schema.TypeBool,
+							Optional:      true,
+							ConflictsWith: []string{"private_cluster_config.0.enable_private_nodes"},
+							Description:   `Controls whether by default nodes have private IP addresses only.`,
+						},
+					},
+				},
+			},
+
 			"private_cluster_config": {
 				Type:             schema.TypeList,
 				MaxItems:         1,
 				Optional:         true,
 				Computed:         true,
 				DiffSuppressFunc: containerClusterPrivateClusterConfigSuppress,
-				Description:      `Configuration for private clusters, clusters with private nodes.`,
+				Description:      `Configuration for private clusters, clusters with private nodes. It's recommended that new configurations use control_plane_endpoints_config for any settings that are available there.`,
 				Elem: &schema.Resource{
 					Schema: map[string]*schema.Schema{
 						// enable_private_endpoint is orthogonal to private_endpoint_subnetwork.
@@ -1739,6 +1861,8 @@ func ResourceContainerCluster() *schema.Resource {
 						"enable_private_endpoint": {
 							Type:             schema.TypeBool,
 							Optional:         true,
+							Deprecated:       "use control_plane_endpoints_config.ip_endpoints_config.enable_public_endpoint instead, and note the flipped semantics",
+							ConflictsWith:    []string{"control_plane_endpoints_config"},
 							AtLeastOneOf:     privateClusterConfigKeys,
 							DiffSuppressFunc: containerClusterPrivateClusterConfigSuppress,
 							Description:      `When true, the cluster's private endpoint is used as the cluster endpoint and access through the public endpoint is disabled. When false, either endpoint can be used.`,
@@ -1746,7 +1870,8 @@ func ResourceContainerCluster() *schema.Resource {
 						"enable_private_nodes": {
 							Type:             schema.TypeBool,
 							Optional:         true,
-							ForceNew:         true,
+							Deprecated:       "use network_config.default_enable_private_nodes instead",
+							ConflictsWith:    []string{"network_config.0.default_enable_private_nodes"},
 							AtLeastOneOf:     privateClusterConfigKeys,
 							DiffSuppressFunc: containerClusterPrivateClusterConfigSuppress,
 							Description:      `Enables the private cluster feature, creating a private endpoint on the cluster. In a private cluster, nodes only have RFC 1918 private addresses and communicate with the master's private endpoint via private networking.`,
@@ -1763,20 +1888,22 @@ func ResourceContainerCluster() *schema.Resource {
 						"peering_name": {
 							Type:         schema.TypeString,
 							Computed:     true,
-							Description: `The name of the peering between this cluster and the Google owned VPC.`,
+							Description:  `The name of the peering between this cluster and the Google owned VPC.`,
 						},
 						"private_endpoint": {
 							Type:         schema.TypeString,
 							Computed:     true,
-							Description: `The internal IP address of this cluster's master endpoint.`,
+							Description:  `The internal IP address of this cluster's master endpoint.`,
 						},
 						"private_endpoint_subnetwork": {
-							Type:         schema.TypeString,
-							Optional:     true,
-							ForceNew:     true,
-							AtLeastOneOf: privateClusterConfigKeys,
+							Type:             schema.TypeString,
+							Optional:         true,
+							Deprecated:       "use control_plane_endpoints_config.ip_endpoints_config.private_endpoint_subnetwork instead",
+							ConflictsWith:    []string{"control_plane_endpoints_config"},
+							ForceNew:         true,
+							AtLeastOneOf:     privateClusterConfigKeys,
 							DiffSuppressFunc: containerClusterPrivateClusterConfigSuppress,
-							Description: `Subnetwork in cluster's network where master's endpoint will be provisioned.`,
+							Description:      `Subnetwork in cluster's network where master's endpoint will be provisioned.`,
 						},
 						"public_endpoint": {
 							Type:         schema.TypeString,
@@ -1784,12 +1911,14 @@ func ResourceContainerCluster() *schema.Resource {
 							Description: `The external IP address of this cluster's master endpoint.`,
 						},
 						"master_global_access_config": {
-							Type:         schema.TypeList,
-							MaxItems:     1,
-							Optional:     true,
-							Computed:     true,
-							AtLeastOneOf: privateClusterConfigKeys,
-							Description: "Controls cluster master global access settings.",
+							Type:          schema.TypeList,
+							MaxItems:      1,
+							Optional:      true,
+							Deprecated:    "use control_plane_endpoints_config.ip_endpoints_config.global_access instead",
+							ConflictsWith: []string{"control_plane_endpoints_config"},
+							Computed:      true,
+							AtLeastOneOf:  privateClusterConfigKeys,
+							Description:   "Controls cluster master global access settings.",
 							Elem: &schema.Resource{
 								Schema: map[string]*schema.Schema{
 									"enabled": {
@@ -2345,13 +2474,13 @@ func resourceContainerClusterCreate(d *schema.ResourceData, meta interface{}) er
 	}
 
 	cluster := &container.Cluster{
-		Name:                           clusterName,
-		InitialNodeCount:               int64(d.Get("initial_node_count").(int)),
-		MaintenancePolicy:              expandMaintenancePolicy(d, meta),
-		MasterAuthorizedNetworksConfig: expandMasterAuthorizedNetworksConfig(d.Get("master_authorized_networks_config"), d),
-		InitialClusterVersion:          d.Get("min_master_version").(string),
-		ClusterIpv4Cidr:                d.Get("cluster_ipv4_cidr").(string),
-		Description:                    d.Get("description").(string),
+		Name:                        clusterName,
+		InitialNodeCount:            int64(d.Get("initial_node_count").(int)),
+		MaintenancePolicy:           expandMaintenancePolicy(d, meta),
+		ControlPlaneEndpointsConfig: expandControlPlaneEndpointsConfig(d),
+		InitialClusterVersion:       d.Get("min_master_version").(string),
+		ClusterIpv4Cidr:             d.Get("cluster_ipv4_cidr").(string),
+		Description:                 d.Get("description").(string),
 		LegacyAbac: &container.LegacyAbac{
 			Enabled:         d.Get("enable_legacy_abac").(bool),
 			ForceSendFields: []string{"Enabled"},
@@ -2388,6 +2517,7 @@ func resourceContainerClusterCreate(d *schema.ResourceData, meta interface{}) er
 			DnsConfig:                            expandDnsConfig(d.Get("dns_config")),
 			GatewayApiConfig:                     expandGatewayApiConfig(d.Get("gateway_api_config")),
 			EnableMultiNetworking:                d.Get("enable_multi_networking").(bool),
+			DefaultEnablePrivateNodes:	          expandDefaultEnablePrivateNodes(d),
 {{- if ne $.TargetVersionName "ga" }}
 			EnableFqdnNetworkPolicy:              d.Get("enable_fqdn_network_policy").(bool),
 {{- end }}
@@ -2486,8 +2616,8 @@ func resourceContainerClusterCreate(d *schema.ResourceData, meta interface{}) er
 		cluster.AuthenticatorGroupsConfig = expandAuthenticatorGroupsConfig(v)
 	}
 
-	if v, ok := d.GetOk("private_cluster_config"); ok {
-		cluster.PrivateClusterConfig = expandPrivateClusterConfig(v)
+	if v, ok := d.GetOk("private_cluster_config.0.master_ipv4_cidr_block"); ok {
+		cluster.PrivateClusterConfig = expandPrivateClusterConfigMasterIpv4CidrBlock(v, cluster)
 	}
 
 	if v, ok := d.GetOk("vertical_pod_autoscaling"); ok {
@@ -2537,10 +2667,6 @@ func resourceContainerClusterCreate(d *schema.ResourceData, meta interface{}) er
 	}
 
 	if err := validateNodePoolAutoConfig(cluster); err != nil {
-		return err
-	}
-
-	if err := validatePrivateClusterConfig(cluster); err != nil {
 		return err
 	}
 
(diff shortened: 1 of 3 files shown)


rileykarson (commented):
Sorry- thought I clicked send on this yesterday.

mmv1/third_party/terraform/go.mod
```diff
@@ -32,7 +32,7 @@ require (
 	golang.org/x/exp v0.0.0-20240409090435-93d18d7e34b8
 	golang.org/x/net v0.28.0
 	golang.org/x/oauth2 v0.22.0
-	google.golang.org/api v0.193.0
+	google.golang.org/api v0.201.0
```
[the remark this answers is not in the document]
gontech: Sure, I've created https://github.com/GoogleCloudPlatform/magic-modules/pull/12042

mmv1/third_party/terraform/services/container/resource_container_cluster.go.tmpl
```diff
@@ -56,6 +56,11 @@ var (
 				Computed: true,
 				Description: `Whether Kubernetes master is accessible via Google Compute Engine Public IPs.`,
 			},
+			"private_endpoint_enforcement_enabled": {
```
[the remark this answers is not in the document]
gontech: The REST API is live, so the documenation on the raw REST calls was supposed to have already been updated. I will look into that. We will be publishing additional documentation along with the cloud console changes next week, too. https://github.com/googleapis/googleapis/blob/master/google/container/v1/cluster_service.proto has already been updated with the protos, so you can look at it there if that helps.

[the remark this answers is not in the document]
gontech: Oh I totally missed that, thank you. Updated.

rileykarson, line 59: This will default to `false` (or null which means `false`), right?

mmv1/third_party/terraform/services/container/resource_container_cluster.go.tmpl
```diff
@@ -1420,7 +1425,8 @@ func ResourceContainerCluster() *schema.Resource {
 			"master_authorized_networks_config": {
 				Type:        schema.TypeList,
 				Optional:    true,
-				Computed:    true,
+				Deprecated:    "use control_plane_endpoints_config.ip_endpoints_config.authorized_networks_config instead",
```
rileykarson, line 1428: Could you add an update test that performs the migration required by this deprecation from one step to another? Ideally backwards from the new to old option as well- we should support both directions in `6.X`.

mmv1/third_party/terraform/services/container/resource_container_cluster.go.tmpl
```diff
@@ -1722,13 +1728,120 @@ func ResourceContainerCluster() *schema.Resource {
 				ConflictsWith: []string{"enable_autopilot"},
 			},
 
+			"control_plane_endpoints_config": {
```
rileykarson, line 1731: Is a user able to adopt this field for `master_authorized_networks_config` and the group of `private_cluster_config` subfields independently, or do they need to modify both all at once?

mmv1/third_party/terraform/services/container/resource_container_cluster.go.tmpl
```diff
@@ -1739,14 +1852,17 @@ func ResourceContainerCluster() *schema.Resource {
 						"enable_private_endpoint": {
 							Type:             schema.TypeBool,
 							Optional:         true,
+							Deprecated:       "use control_plane_endpoints_config.ip_endpoints_config.enable_public_endpoint instead, and note the flipped semantics",
```
rileykarson, line 1855: Similar to above, can we add a test case that migrates between these settings? Ideally in a separate case that does not have a value for master_authorized_networks_config rather than combined into one sample.

mmv1/third_party/terraform/services/container/resource_container_cluster.go.tmpl
```diff
@@ -1722,13 +1728,120 @@ func ResourceContainerCluster() *schema.Resource {
...
+			},
+
 			"private_cluster_config": {
 				Type:             schema.TypeList,
 				MaxItems:         1,
 				Optional:         true,
 				Computed:         true,
 				DiffSuppressFunc: containerClusterPrivateClusterConfigSuppress,
-				Description:      `Configuration for private clusters, clusters with private nodes.`,
+				Description:      `Configuration for private clusters, clusters with private nodes. Beware that most of the fields have been deprecated in favor of control_plane_endpoints_config.`,
```
rileykarson, line 1844: for here + docs, "It's recommended that new configurations use control_plane_endpoints_config for any settings that are available there"?

mmv1/third_party/terraform/services/container/resource_container_cluster.go.tmpl
```diff
@@ -1763,38 +1879,42 @@ func ResourceContainerCluster() *schema.Resource {
...
+							ConflictsWith: []string{"control_plane_endpoints_config.0.ip_endpoints_config.0.global_access"},
+							Computed:      true,
+							AtLeastOneOf:  privateClusterConfigKeys,
+							Description:   "Controls cluster master global access settings.",
 							Elem: &schema.Resource{
 								Schema: map[string]*schema.Schema{
 									"enabled": {
 										Type:        schema.TypeBool,
-										Required:    true,
+										Optional:    true,
```
rileykarson, line 1917: Can you explain the situation a user would want to set `master_global_access_config {}` in configuration? In case you are unaware, `Required` only applies in the scope of the parent block, so this specifically presents `master_global_access_config {}` from being specified, but does not force specifying a value when it is not present.

rileykarson: Similarly, we may want to consider making subfields of some new fields `Required: true`, unless it is valid to specify an empty block.

mmv1/third_party/terraform/services/container/resource_container_cluster.go.tmpl
```diff
@@ -2388,6 +2508,7 @@ func resourceContainerClusterCreate(d *schema.ResourceData, meta interface{}) er
 			DnsConfig:                            expandDnsConfig(d.Get("dns_config")),
 			GatewayApiConfig:                     expandGatewayApiConfig(d.Get("gateway_api_config")),
 			EnableMultiNetworking:                d.Get("enable_multi_networking").(bool),
+			DefaultEnablePrivateNodes:		      expandDefaultEnablePrivateNodes(d),
```
rileykarson, line 2511:
```suggestion
			EnableMultiNetworking:                d.Get("enable_multi_networking").(bool),
			DefaultEnablePrivateNodes:            expandDefaultEnablePrivateNodes(d),
```

tabs -> spaces (not sure if I got it to line up quite right...)

mmv1/third_party/terraform/services/container/resource_container_cluster.go.tmpl
```diff
@@ -3122,19 +3240,19 @@ func resourceContainerClusterUpdate(d *schema.ResourceData, meta interface{}) er
...
 	// so we have to make separate calls for each field that we want to update. The order here is fairly arbitrary-
 	// if the order of updating fields does matter, it is called out explicitly.
-	if d.HasChange("master_authorized_networks_config") {
-		c := d.Get("master_authorized_networks_config")
+
+	if d.HasChange("control_plane_endpoints_config") || d.HasChange("master_authorized_networks_config") {
 		req := &container.UpdateClusterRequest{
 			Update: &container.ClusterUpdate{
-				DesiredMasterAuthorizedNetworksConfig: expandMasterAuthorizedNetworksConfig(c, d),
+				DesiredControlPlaneEndpointsConfig: expandControlPlaneEndpointsConfig(d),
```
rileykarson, line 3247: Will this cause an issue for a config with both `master_authorized_networks_config` and `private_cluster_config` set? Do any test cases do that already so we can be sure there aren't regressions?

mmv1/third_party/terraform/services/container/resource_container_cluster_test.go.tmpl
```diff
@@ -11843,3 +11843,103 @@ resource "google_container_cluster" "with_autopilot_gcp_filestore" {
...
+	t.Parallel()
+
+	clusterName := fmt.Sprintf("tf-test-cluster-%s", acctest.RandString(t, 10))
+	acctest.VcrTest(t, resource.TestCase{
+		PreCheck:                 func() { acctest.AccTestPreCheck(t) },
+		ProtoV5ProviderFactories: acctest.ProtoV5ProviderFactories(t),
+		CheckDestroy:             testAccCheckContainerClusterDestroyProducer(t),
+		Steps: []resource.TestStep{
+			{
+				Config: fmt.Sprintf(`
```
rileykarson, line 11857: Could I get you to break these into functions, similar to other tests in the file / the structure in https://googlecloudplatform.github.io/magic-modules/develop/test/test/#add-an-update-test? Apologies, I see why this makes sense in the context of a single change, but it's a big codebase so we want to stay consistent.

mmv1/third_party/terraform/services/container/resource_container_cluster_test.go.tmpl
```diff
@@ -11843,3 +11843,103 @@ resource "google_container_cluster" "with_autopilot_gcp_filestore" {
...
+				),
+			},
+			{
+				ResourceName:            "google_container_cluster.primary",
+				ImportState:             true,
+				ImportStateVerify:       true,
+				ImportStateVerifyIgnore: []string{"deletion_protection"},
+			},
+			{
+				Config: fmt.Sprintf(`
```
rileykarson, line 11902: We should use the `ConfigPlanChecks` from https://googlecloudplatform.github.io/magic-modules/develop/test/test/#add-an-update-test to assert this updates and does not recreate the resource. This file predates that option being available, and we haven't widely backfilled it yet!

rileykarson (commented):
Note that this is just a partial review pass- I have not unraveled all the update code yet, but have a bunch of tests / test changes that should help a considerable amount if you've made them before I finish this pass.

rileykarson:
Some of those failures aren't real, we don't actually have the quota to run all those tests in one go in the presubmit environment & full GKE rerecords take a few iterations! Kicking off the next one...

/gcbrun

mmv1/third_party/terraform/services/container/resource_container_cluster.go.tmpl
```diff
@@ -1722,13 +1728,120 @@ func ResourceContainerCluster() *schema.Resource {
 				ConflictsWith: []string{"enable_autopilot"},
 			},
 
+			"control_plane_endpoints_config": {
+				Type:        schema.TypeList,
+				MaxItems:    1,
+				Optional:    true,
```
rileykarson, line 1734:
```suggestion
				Computed:    true,
				Optional:    true,
```

rileykarson: Since this gets returned back when the settings are set through the old mechanisms, this *must* be set.

mmv1/third_party/terraform/services/container/resource_container_cluster.go.tmpl
gontech: I've added separate tests for private nodes (network_config field), master_authorized_networks_config, misc private_cluster_config fields, and DNS enablement. I haven't run it yet, since it takes a while to run, I'll double check that soon.

mmv1/third_party/terraform/services/container/resource_container_cluster.go.tmpl
gontech: The docs at https://developer.hashicorp.com/terraform/plugin/sdkv2/schemas/schema-behaviors#computed were very unclear as to what Computed actually does. I gathered that it seems to have an effect on computed diffs, but that is not mentioned at all in the docs. I was largely cargo-culting this attribute and applying it to OUTPUT_ONLY fields. Does this mean dns_endpoint_config needs to be Computed as well, since the DNS endpoint field is always present? I will assume yes for now.

mmv1/third_party/terraform/services/container/resource_container_cluster.go.tmpl
gontech: That sounds better.

mmv1/third_party/terraform/services/container/resource_container_cluster.go.tmpl
gontech: Done as part of TestAccContainerCluster_migratePrivateClusterConfig

mmv1/third_party/terraform/services/container/resource_container_cluster.go.tmpl
gontech: This was just to match the underlying API, which was built before proto3 added the optional keyword back. So we only used this to get presence information for the enabled bool, so we could send private_cluster_config patches without having to do a full read-modify-write loop. It's unnecessary here, so I'll just go back to using Required. I only did this to make sure we can do the migration; I didn't grok the scoping.

mmv1/third_party/terraform/services/container/resource_container_cluster.go.tmpl
gontech: Done.

mmv1/third_party/terraform/services/container/resource_container_cluster_test.go.tmpl
gontech: Sure, consistency is a good thing.

mmv1/third_party/terraform/services/container/resource_container_cluster.go.tmpl
gontech: Yes, I've updated the descriptions.

mmv1/third_party/terraform/services/container/resource_container_cluster.go.tmpl
gontech: It can migrated incrementally/independently. See expandControlPlaneEndpointsConfig line 5619 where it first checks for d.GetOk("master_authorized_networks_config") when filling in the ip_endpoints_config to the server. This way, it'll always use the new fields in the API request to GKE, but users can migrate their TF config as slowly as they want.

gontech: Although, the way it's written now, the block of private_cluster_config fields needs to be migrated as one unit. So you can't, for instance, migrating enable_public_endpoints to control_plane_endpoints_config, but leave global_access in private_cluster_config. I'm not sure if we'd need to have additional validation to prevent that, or if it's better to just rewrite it to look for fallbacks for each and every field.

mmv1/third_party/terraform/services/container/resource_container_cluster.go.tmpl
rileykarson: We talk about the option here in our docs, in case it helps! https://googlecloudplatform.github.io/magic-modules/develop/permadiff/#default. HashiCorp under-covers Optional+Computed, as it is required much more often in GCP which has individually larger resources with more moving parts than AWS, Azure, or most other providers.

We use this if the API will return a value when `null` (i.e. nothing, although explicit nulls do the same) is sent and it's not a known fixed default for a primitive.

> Does this mean dns_endpoint_config needs to be Computed as well, since the DNS endpoint field is always present? I will assume yes for now.

Probably yes!

rileykarson (commented):
See error inline in the GH checks!

mmv1/third_party/terraform/services/container/resource_container_cluster_test.go.tmpl
gontech: Done.

gontech: ...and now with the right checks. I had it checking for updates in the test for ensuring that the migration to the new fields was a no-op 🤦‍♂️ .

mmv1/third_party/terraform/services/container/resource_container_cluster.go.tmpl
rileykarson: > I'm not sure if we'd need to have additional validation to prevent that, or if it's better to just rewrite it to look for fallbacks for each and every field.

Either approach can work, I don't have a strong opinion on which we do as long as we end up in the right final state for any given configuration.

mmv1/third_party/terraform/services/container/resource_container_cluster.go.tmpl
rileykarson: Oh man, that's a whole yak shave on it's own Proto3 optional is a slightly different thing than optional. That's equivalent to turning a bool into a *bool, which we don't have the types to express (and would be breaking anyways, as it would for an API to retrofit as it changes the meaning of the null value).

mmv1/third_party/terraform/services/container/resource_container_cluster.go.tmpl
gontech: FYI, after going through more of the tests, I found another implication of the use of Computed on the top level fields like `control_plane_endpoints_config` and `ip_endpoints_config`. I had assumed that since the leaf fields were not marked Computed, that I would be following https://developer.hashicorp.com/terraform/plugin/sdkv2/best-practices/deprecations#example-renaming-of-an-optional-attribute. But, apparently TF records the state of all the subfields, and so by only setting one of the fields in the flatten function (corresponding to whichever the user configured), the other unset fields would record their default values. This caused problems with unexpected diffs. So I changed to setting all the fields all the time in flatten. It seems like things are working now, I'm running the tests one last time locally to be sure it all passes.

mmv1/third_party/terraform/services/container/resource_container_cluster.go.tmpl
rileykarson: Thanks! Yeah, I was concerned about potentially sending a partial DesiredControlPlaneEndpointsConfig and overwriting previously set values for the other field.

mmv1/third_party/terraform/services/container/resource_container_cluster.go.tmpl
gontech: OK, I've updated the ConflictsWith attributes to ensure that users can't configure `private_cluster_config` in a way that `expandControlPlaneEndpointsConfig` doesn't handle.

gontech:
Bizarre, all the tests pass locally except for TestAccContainerCluster_migrateAuthorizedNetworks. It says there's a planned update action during step 3 where we switch to the newer config, but if I try it manually by copying the configs from that test to a file and trying `terraform plan`/`terraform apply`, there's no planned change.

rileykarson:
Some of these are still quotas- this should be the last run with those, at least.

/gcbrun

gontech:
> Bizarre, all the tests pass locally except for TestAccContainerCluster_migrateAuthorizedNetworks. It says there's a planned update action during step 3 where we switch to the newer config, but if I try it manually by copying the configs from that test to a file and trying `terraform plan`/`terraform apply`, there's no planned change.

OK it's all passing locally now, I must've missed a `make provider` between iterations. Let's see what the presubmit says, too.

gontech:
Huh, the last remaining failing check is the VCR test, but when I look at the details, it's successful.

rileykarson:
Something (that I haven't been able to figure out) is causing those tests to not get executed in VCR. Looked at some logs last week, but gonna poke at these newer ones.

mmv1/third_party/terraform/services/container/resource_container_cluster.go.tmpl
```diff
@@ -2486,7 +2616,7 @@ func resourceContainerClusterCreate(d *schema.ResourceData, meta interface{}) er
 		cluster.AuthenticatorGroupsConfig = expandAuthenticatorGroupsConfig(v)
 	}
 
-	if v, ok := d.GetOk("private_cluster_config"); ok {
+	if v, ok := d.GetOk("private_cluster_config.0.master_ipv4_cidr_block"); ok {
```
rileykarson, line 2619: We need to call this with the same value as before, or modify the expand function to expect the new value- we're currently running into an issue because the types have changed:

> panic: interface conversion: interface {} is string, not []interface {}
> 
> goroutine 22369 [running]:
> github.com/hashicorp/terraform-provider-google-beta/google-beta/services/container.expandPrivateClusterConfig(...)
> 	/go/src/github.com/modular-magician/terraform-provider-google-beta/google-beta/services/container/resource_container_cluster.go:5335

Exact line #s in this case correspond to the diff at https://github.com/modular-magician/terraform-provider-google-beta/compare/auto-pr-12025-old..auto-pr-12025, and will be erased on a push. L5536 is the corresponding line in this file.

gontech:
Ah I see there are a bunch of failures with the message `master_ipv4_cidr_block can only be set if enable_private_nodes is true`. This is because we moved the location of `private_cluster_config.enable_private_nodes` to `network_config.default_enable_private_nodes`. I don't see any server side validation of this condition, I don't think it is relevant. In fact, the only time `master_ipv4_cidr_block` is meaningless to set is for legacy public clusters, and it is no longer possible to make those clusters. We should just remove the client side validation here.

gontech:
Most of the rest of the failures appear to be quota issues, with the exception of `withEnablePrivateEndpoint` having an unexpected non-empty plan after changing enable_private_endpoint from true to false. This should be taken care of by the flatten function, so I'll try to reproduce this manually tomorrow morning. I'm still hopeful that I can get this in in time for the next release.

rileykarson:
I think some of the earlier VCR results were misleading- most tests were failing to run due to crashes. We're getting through them all now, and hitting some more failure cases. I'm seeing one in particular where the the API is reading `null` as `false` for `enable_private_nodes` (which is reasonable for the API to do IMO! just requires crafting the request message carefully):

>  "networkConfig": {
>    "defaultEnablePrivateNodes": true,
>    "enableIntraNodeVisibility": true
>   },
>   "privateClusterConfig": {
>    "masterIpv4CidrBlock": "172.16.0.0/28"
>   },

We get an error message of `Error: googleapi: Error 400: Cluster.private_cluster_config.enable_private_nodes cannot have different value from cluster.network_config.default_enable_private_nodes.`

rileykarson:
As a procedural note, if you could structure changes as new commits and only rebase / force push if we run into conflicts with `main` during CI runs that'd help reviews. The force push view can be hard to read as a reviewer and this is a big / complex change!

gontech:
I tracked the `withEnablePrivateEndpointToggle` test failures down to https://github.com/GoogleCloudPlatform/magic-modules/blob/8cdc67d5fc52526ea788ddafdf4cc720df29979a/mmv1/third_party/terraform/services/container/resource_container_cluster.go.tmpl#L5565 not working as I had hoped. I believe that as soon as TF creates the cluster in the first step, it records values in the `control_plane_endpoints_config` field for the DNS endpoint. Since the expand function branches based in the presence of `control_plane_endpoints_config`, it would never read from legacy `private_cluster_config` fields after the initial creation. I had originally written it thinking that `d.GetOk` only pulled from the requested config, not the complete stored state merged with the new config. I'm not sure how to get that to play nicely, I might need to reconsider unilaterally writing both deprecated and new fields in the flatten function, despite them being marked Computed.

gontech:
I'm stuck now, unfortunately I think that means this will miss the next release window. I refactored it to follow https://developer.hashicorp.com/terraform/plugin/sdkv2/best-practices/deprecations#example-renaming-of-an-optional-attribute, i.e. during flatten, only write to the field that's used by the config. Somehow, I see both values in the config when I run `terraform show`.
```
resource "google_container_cluster" "default" {
  name = "tf-cluster"
  location = "us-central1"
  initial_node_count = 1
  private_cluster_config {
    enable_private_endpoint = false
  }
  master_authorized_networks_config {
    gcp_public_cidrs_access_enabled = false
  }
  deletion_protection = false
}
```
and the state:
```
    control_plane_endpoints_config {
        dns_endpoint_config {
            allow_external_traffic = false
            endpoint               = <redacted>
        }
        ip_endpoints_config {
            enable_public_endpoint      = false
            enabled                     = true
            global_access               = false
            private_endpoint            = <redacted>
            private_endpoint_subnetwork = null
            public_endpoint             = null
        }
    }
...
    private_cluster_config {
        enable_private_endpoint     = false
        enable_private_nodes        = false
        master_ipv4_cidr_block      = null
        peering_name                = null
        private_endpoint            = "<redacted>"
        private_endpoint_subnetwork = null
        public_endpoint             = null
    }
```
It's hallucinating the value of `enable_public_endpoint`, since that is semantically opposite of `enable_private_endpoint`, so if it's not `null`, it should be `true` in this case. Additionally, when I try out a migration from the old config to the new config, there's an unexpected diff in the new field, as it thinks `enable_public_endpoint` is changing from `false` to `true`. Did I design myself into a corner with semantics that TF doesn't model well? (i.e. migrating a boolean to a field with reversed semantics)

rileykarson (commented):
This is a _very_ complex change and more complex than I initially thought it would be- we spoke about this offline, but I'd recommend we break up the migration from old fields -> new fields into a separate change, and introduce the wholly new fields in a PR first, switching our internal calls to use `controlPlaneEndpointsConfig` for the now-deprecated values.

---

We still need to solve for the migration case eventually, however.  I was playing with some of the advanced options available in CustomizeDiff and think we can identify the cases where they appear in users' configs by using a newer, advanced option called `GetRawConfig`. Specifically, I played with master_authorized_networks_config and think that switching will trigger a case where the value is known but has a length of zero that will set us set the value to unknown or empty in the plan, making it unambiguous in the expand function while the correct source of values is. 

> Additionally, when I try out a migration from the old config to the new config, there's an unexpected diff in the new field, as it thinks enable_public_endpoint is changing from false to true

Not sure why that happens, the flatten code looks right. Do you have the GET response from the API we flattened to create the state?

---
```
  /*
  private_cluster_config {
    enable_private_nodes    = true
    master_ipv4_cidr_block  = "10.42.0.0/28"
    enable_private_endpoint = true
    master_global_access_config {
      enabled = true
    }
  }

  master_authorized_networks_config {
    gcp_public_cidrs_access_enabled = false
  }
  */

  private_cluster_config {
    master_ipv4_cidr_block = "10.42.0.0/28"
  }

  network_config {
    default_enable_private_nodes = true
  }

  control_plane_endpoints_config {
    ip_endpoints_config {
      enabled                = true
      enable_public_endpoint = false
      global_access          = true
      authorized_networks_config {
        gcp_public_cidrs_access_enabled = false
      }
    }
```

```
	field := "master_authorized_networks_config"
	if d.NewValueKnown(field) {
		of, nf := d.GetChange(field)
		return fmt.Errorf("value was known, %v, %v, %v, %v, %v, %v", of, nf, d.GetRawConfig().GetAttr(field), d.GetRawConfig().GetAttr(field).IsWhollyKnown(), d.GetRawConfig().GetAttr(field).IsNull(), d.GetRawConfig().GetAttr(field).AsValueSlice())
	}
```

```
Error: value was known, [map[cidr_blocks:0x14001cf0aa0 gcp_public_cidrs_access_enabled:false private_endpoint_enforcement_enabled:false]], [map[cidr_blocks:0x14001cf0a00 gcp_public_cidrs_access_enabled:false private_endpoint_enforcement_enabled:false]], {{{{} {{{} map[cidr_blocks:{{{} {{{} map[cidr_block:{{{} 83}} display_name:{{{} 83}}]}}}} gcp_public_cidrs_access_enabled:{{{} 66}} private_endpoint_enforcement_enabled:{{{} 66}}]}}}} []}, true, false, []
```

gontech:
> This is a _very_ complex change and more complex than I initially thought it would be- we spoke about this offline, but I'd recommend we break up the migration from old fields -> new fields into a separate change, and introduce the wholly new fields in a PR first, switching our internal calls to use `controlPlaneEndpointsConfig` for the now-deprecated values.

I've created https://github.com/GoogleCloudPlatform/magic-modules/pull/12195 for just the new dns_endpoint_config field + using the new API fields in the internal calls. This should get the main new functionality out sooner.

jackweinbender:
FWIW, we'd be VERY happy to see the core functionality come out sooner, even without the migration stuff.

rileykarson:
New functionality was split out to https://github.com/GoogleCloudPlatform/magic-modules/pull/12195 and was merged.

philipsabri:
Any update on this?

gontech:
> Any update on this?

Unfortunately I have not had a chance to return to this, and the earliest I expect to be able to do so is mid-February. The core functionality of being able to enable/disable/view the DNS endpoint for a cluster has been released. The missing pieces are:
1) Moving the location of the old `private_cluster_config` related fields to their new location under `control_plane_endpoints_config` in a backwards-compatible way. This allows consistency with the underlying [HTTP API](https://cloud.google.com/kubernetes-engine/docs/reference/rest/v1/projects.locations.clusters#Cluster.IPEndpointsConfig) as well as a more logical grouping of fields, however this migration is proving to be difficult to ensure that nothing breaks.
2) Adding the new `enabled` field under `ip_endpoints_config` to allow disabling IP access for those who want to ensure that DNS access is the only way to get to their cluster.

philipsabri:
> > Any update on this?
> 
> Unfortunately I have not had a chance to return to this, and the earliest I expect to be able to do so is mid-February. The core functionality of being able to enable/disable/view the DNS endpoint for a cluster has been released. The missing pieces are:
> 
> 1. Moving the location of the old `private_cluster_config` related fields to their new location under `control_plane_endpoints_config` in a backwards-compatible way. This allows consistency with the underlying [HTTP API](https://cloud.google.com/kubernetes-engine/docs/reference/rest/v1/projects.locations.clusters#Cluster.IPEndpointsConfig) as well as a more logical grouping of fields, however this migration is proving to be difficult to ensure that nothing breaks.
> 2. Adding the new `enabled` field under `ip_endpoints_config` to allow disabling IP access for those who want to ensure that DNS access is the only way to get to their cluster.

@gontech 

Thank you for the update! 

We need the 2nd point `Adding the new enabled field under ip_endpoints_config to allow disabling IP access`. Would it be possible to break that out into another PR? I wouldn’t mind working on that if it’s possible.

rileykarson (commented):
(just marking a no-op review pass to pause the notifications- once there's a push they'll start again)

edit: actually, I'll move this to draft

PR #12026, opened 2024-10-16
fix TestAccAccessApprovalSettings

shuyama1:
potentially fixes https://github.com/hashicorp/terraform-provider-google/issues/13716

```release-note:none

```


melinath (approved):

mmv1/products/accessapproval/FolderSettings.yaml
```diff
@@ -36,6 +36,8 @@ custom_code:
   constants: 'templates/terraform/constants/access_approval.go.tmpl'
   pre_create: 'templates/terraform/update_mask.go.tmpl'
   custom_delete: 'templates/terraform/custom_delete/clear_folder_access_approval_settings.go.tmpl'
+  post_create: 'templates/terraform/post_create/sleep_1_min.go.tmpl'
+  post_update: 'templates/terraform/post_create/sleep_1_min.go.tmpl'
```
melinath, line 40:
```suggestion
  post_create: 'templates/terraform/post_create/sleep_2_min.go.tmpl'
  post_update: 'templates/terraform/post_create/sleep_2_min.go.tmpl'
```

mmv1/products/accessapproval/ProjectSettings.yaml
```diff
@@ -35,6 +35,8 @@ timeouts:
 custom_code:
   pre_create: 'templates/terraform/update_mask.go.tmpl'
   custom_delete: 'templates/terraform/custom_delete/clear_project_access_approval_settings.go.tmpl'
+  post_create: 'templates/terraform/post_create/sleep_1_min.go.tmpl'
+  post_update: 'templates/terraform/post_create/sleep_1_min.go.tmpl'
```
melinath, line 39:
```suggestion
  post_create: 'templates/terraform/post_create/sleep_2_min.go.tmpl'
  post_update: 'templates/terraform/post_create/sleep_2_min.go.tmpl'
```

melinath (approved):

melinath (approved):

shuyama1:
Run tests on [TC](https://hashicorp.teamcity.com/buildConfiguration/TerraformProviders_GoogleCloud_GOOGLE_BETA_MMUPSTREAMTESTS_GOOGLEBETA_PACKAGE_ACCESSAPPROVAL/285641) and it failed with the following error:
```
------- Stdout: -------
=== RUN   TestAccAccessApprovalSettings
=== RUN   TestAccAccessApprovalSettings/folder
=== RUN   TestAccAccessApprovalSettings/project
=== RUN   TestAccAccessApprovalSettings/organization
    resource_access_approval_organization_settings_test.go:46: Step 5/6 error: Error running apply: exit status 1
        Error: Provider produced inconsistent result after apply
        When applying changes to
        google_organization_access_approval_settings.organization_access_approval,
        provider "provider[\"registry.terraform.io/hashicorp/google\"]" produced an
        unexpected new value: Root object was present, but now absent.
        issue tracker.
--- FAIL: TestAccAccessApprovalSettings (1335.90s)
    --- PASS: TestAccAccessApprovalSettings/folder (543.34s)
    --- PASS: TestAccAccessApprovalSettings/project (399.34s)
    --- FAIL: TestAccAccessApprovalSettings/organization (393.21s)
FAIL
```
Looks like 2 minutes might not always be sufficient. In earlier attempts, I tested several wait times, and the tests only passed with a 5-minute wait. But adding such a long wait to the resource itself doesn’t feel right since it would unnecessarily slow down every creation and update for users. I’m thinking of switching approaches by adding a sleep delay in the test itself instead of the resource.

melinath:
I don't think we should add the sleep delay to the test itself - this is a real issue that could impact users, since we read immediately after apply. If 5min worked we should just do that. It looks like the test is able to get a passing result, then still fail later, so we can't rely on a successful GET request to determine whether enough time has passed.

sampcr-google (approved):

shuyama1:
Sorry, forgot to update the test status here: Run test on [TC](https://hashicorp.teamcity.com/buildConfiguration/TerraformProviders_GoogleCloud_GOOGLE_BETA_MMUPSTREAMTESTS_GOOGLEBETA_PACKAGE_ACCESSAPPROVAL/287276) and it passed only with a 5-min wait time. So let’s set it to 5-min for now, and we can reduce it in the future if the backend update time decreases.

melinath (approved):
SGTM

PR #12028, opened 2024-10-17
fix: remove retry in read as retry for create/update is already handled in respective methods

gurusai-voleti:
remove retry in read as retry for create/update is already handled in respective methods

Fixes: https://github.com/hashicorp/terraform-provider-google/issues/13464

```release-note: bug
storage: removed retry on 404s during refresh for `google_storage_bucket`, preventing hanging when refreshing deleted buckets
```


rileykarson:
What bug(s) is this addressing / what's the context?

gurusai-voleti:
> What bug(s) is this addressing / what's the context?

http://b/275607155

rileykarson (approved):
LGTM, I think this makes sense. We added this to several spots in https://github.com/GoogleCloudPlatform/magic-modules/pull/5288 and I'm not confident we needed this one + that PR has no extra context on why we would need it here specifically.

Waiting on approval from @kautikdk to merge

kautikdk (approved):
Considering public documentation of cloud storage consistency:https://cloud.google.com/storage/docs/consistency#strongly_consistent_operations, Bucket Listing is strongly consistent operation so we don't need retry logic for bucket read operation.

PR #12029, opened 2024-10-17
fix: Handle tpgtools block when provider version is 'private'

sana-google:
I was working on adding a few fields to the google-private repo, for an existing resource. I was running into errors when executing the `make provider ...` command. By digging through the code, it seems that the `mmv1/main.go` file looks for the VERSION flag to be `private` but the `tpgtools/main.go` file expects it to be `alpha`.  I was able to generate the private provider by updating the `tpgtools` block.


hao-nan-li (commented):
I don't think `private` version provider should be visible on git. There should be another flow to handle your use case.

PR #12031, opened 2024-10-17
convert output parameter enum to string

Mehul3217:

```release-note:none
```

### DIFF
--- a/mmv1/products/netapp/ActiveDirectory.yaml
+++ b/mmv1/products/netapp/ActiveDirectory.yaml
@@ -73,18 +73,10 @@ properties:
       Create time of the active directory. A timestamp in RFC3339 UTC "Zulu" format. Examples: "2023-06-22T09:13:01.617Z".
     output: true
   - name: 'state'
-    type: Enum
+    type: String
     description: |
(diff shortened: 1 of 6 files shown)


c2thorn:
@Mehul3217 it looks like since these fields are all output-only, this change has no actual code change in the downstream providers. Seems that we already omit any validation for output-only enums.

We can go ahead and merge this change to the yaml files for clarity/readability, but I wanted to make sure you were aware of this.

I've also removed the changelog note since there is no downstream change.

Mehul3217:
> @Mehul3217 it looks like since these fields are all output-only, this change has no actual code change in the downstream providers. Seems that we already omit any validation for output-only enums.
> 
> We can go ahead and merge this change to the yaml files for clarity/readability, but I wanted to make sure you were aware of this first.
> 
> I've also removed the changelog note since there is no downstream change.


Yes that's correct we understand there is actually no code change, we are just making this change for more clarity

c2thorn:
> Yes that's correct we understand there is actually no code change, we are just making this change for more clarity

Awesome. LGTM

c2thorn (approved):

PR #12036, opened 2024-10-17
Added google_apigee_api resource

apichick:
This is a PR to add the google_apigee_api resource

```release-note:new-resource
`google_apigee_api`
```


mmv1/third_party/terraform/services/apigee/resource_apigee_api.go
```diff
@@ -0,0 +1,403 @@
...
+	}
+	d.SetId(id)
+	log.Printf("[DEBUG] create d.SetId done, id = %s", id)
+
+	log.Printf("[DEBUG] Finished creating API proxy %q: %#v", d.Id(), res)
+
+	if resourceApigeeApiRead(d, meta) != nil {
+		return fmt.Errorf("Error reading API proxy at end of Create: %s", err)
+	}
+	d.Set("md5hash", localMd5Hash)
```
slevenick, line 201: Why do we set these *after* the read method is called? If they were set during read they would be overwritten here

mmv1/third_party/terraform/services/apigee/resource_apigee_api.go
```diff
@@ -0,0 +1,403 @@
...
+	d.Set("detect_md5hash", localMd5Hash)
+	return nil
+}
+
+func resourceApigeeApiUpdate(d *schema.ResourceData, meta interface{}) error {
+	//For how API proxy api is implemented, just treat an update as create, when the name is same, it will create a new revision
```
slevenick, line 207: If calling the create method creates a new revision, how do we make sure we are reading the right revision when we call the read method? If someone creates a new revision of an API outside of Terraform would we immediately see a diff?

mmv1/third_party/terraform/services/apigee/resource_apigee_api.go
```diff
@@ -0,0 +1,403 @@
...
+	}
+	lastModifiedAt := metaData["last_modified_at"].(string)
+	if lastModifiedAt == "" {
+		return "UNKNOWN"
+	}
+	return lastModifiedAt
+}
+
+func resourceApigeeApiDelete(d *schema.ResourceData, meta interface{}) error {
+	log.Printf("[DEBUG] resourceApigeeApiDelete")
```
slevenick, line 285: We can cleanup some of these lines.

mmv1/third_party/terraform/website/docs/r/apigee_api.html.markdown
```diff
@@ -0,0 +1,105 @@
...
+  A list of revisions of this API proxy.
+
+* `latest_revision_id` -
+  The id of the most recently created revision for this API proxy.
+
+* `md5hash` -
+  (Computed) Base 64 MD5 hash of the uploaded data. It is speculative as remote does not return hash of the bundle. Remote changes are detected using returned last_modified timestamp.
+
+* `detect_md5hash` -
+  (Optional) Detect changes to local config bundle file or changes made outside of Terraform. MD5 hash of the data, encoded using base64. Hash is automatically computed without need for user input.
```
slevenick, line 60: I don't think this would be under `Attributes` if it allows the user to set it. If the user can't set it, it should be not marked as (Optional)

mmv1/third_party/terraform/services/apigee/resource_apigee_api.go
```diff
@@ -0,0 +1,403 @@
...
+			Update: schema.DefaultTimeout(20 * time.Minute),
+			Delete: schema.DefaultTimeout(20 * time.Minute),
+		},
+
+		Schema: map[string]*schema.Schema{
+			"name": {
+				Type:        schema.TypeString,
+				Required:    true,
+				ForceNew:    true,
+				Description: `Name of the API proxy. Restrict the characters used to: A-Za-z0-9._-.`,
```
slevenick, line 58: The description here should probably be:
The name of the API proxy. This field only accepts the following characters: ...

mmv1/third_party/terraform/services/apigee/resource_apigee_api.go
```diff
@@ -0,0 +1,403 @@
...
+			},
+			"md5hash": {
+				Type:        schema.TypeString,
+				Computed:    true,
+				Description: `Base 64 MD5 hash of the uploaded config bundle.`,
+			},
+			"detect_md5hash": {
+				Type:        schema.TypeString,
+				Optional:    true,
+				Default:     "Different Hash",
```
slevenick, line 116: Why do we need a default value here?

mmv1/third_party/terraform/services/apigee/resource_apigee_api.go
```diff
@@ -0,0 +1,403 @@
...
+	}
+	if err := d.Set("revision", flattenApigeeApiRevision(res["revision"], d, config)); err != nil {
+		return fmt.Errorf("Error reading API proxy: %s", err)
+	}
+	if err := d.Set("latest_revision_id", flattenApigeeApiLatestRevisionId(res["latestRevisionId"], d, config)); err != nil {
+		return fmt.Errorf("Error reading API proxy: %s", err)
+	}
+
+	//setting hash to suggest update
+	if previousLastModifiedAt != currentLastModifiedAt {
```
slevenick, line 259: Why do we go through the whole md5hash detection for this resource?

It seems to me that we don't read the hash from the server-side, so we're only calculating it locally. But we only calculate it locally during create, and just use it here to trigger recreate if the last modified time changes. Can we just have a customize diff on last modified at to trigger this?

mmv1/third_party/terraform/services/apigee/resource_apigee_api.go
apichick: Are you referring to the DEBUG log statements? Do you want to remove all of the in the resource?

slevenick: Not all of them, but the logging of where we are in the method can be removed. I would keep one or two like "finished creating resource with ID: xxx"

mmv1/third_party/terraform/services/apigee/resource_apigee_api.go
apichick: Yes, that should trigger a diff and that is expected the proxy would have changed outside the local copy is different, so the expectation would be to have to re-apply to bring back the version locally.

mmv1/third_party/terraform/services/apigee/resource_apigee_api.go
apichick: I understand that the hash is calculated and stored in the state so it can be determined whether we locally have a change from what was previously deployed using terraform from what we currently have locally. As the revision is established remotely and not locally there would not be way to understand if there was been a changed locally without the hash.

mmv1/third_party/terraform/services/apigee/resource_apigee_api.go
apichick: Tests past without it so I removed it

mmv1/third_party/terraform/website/docs/r/apigee_api.html.markdown
apichick: Thay are part of the schema, so we should keep them

slevenick (approved):
Thanks!

mahek-cldcvr:
There is some documentation issue , Example is provided for google_apigee_sharedflow instead of [google_apigee_api](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/apigee_api)
image

shawnsavour:
md5 not working, have bundle same name, md5 different, but google_apigee_api not trigger update. md5hash base64 seems wrong
"terraform_version": "1.9.0"
"google-beta": "6.23.0"

```json
  {
      "index_key": "test-to-<REDACTED>-prd",
      "schema_version": 0,
      "attributes": {
        "config_bundle": "./temp/test-to-<REDACTED>-prd.zip",
        "detect_md5hash": "UVMpfEhJnZ06kkhywXKymg==",
        "id": "organizations/<REDACTED>/apis/test-to-<REDACTED>-prd",
        "latest_revision_id": "",
        "md5hash": "UVMpfEhJnZ06kkhywXKymg==",
        "meta_data": [{
          "created_at": "1740712271425",
          "last_modified_at": "1740902593940",
          "sub_type": "Proxy"
        }],
        "name": "<REDACTED>",
        "org_id": "<REDACTED>",
        "revision": ["1", "2", "3", "4"],
        "timeouts": null
      },
      "sensitive_attributes": [],
      "private": "<REDACTED>",
      "dependencies": ["archive_file.proxy_folders", "null_resource.always_run"]
    }
```

slevenick:
@shawnsavour please file an issue against https://github.com/terraform-providers/terraform-provider-google

PR #12038, opened 2024-10-17
Autogen tool in Go

slevenick:
Adds autogen tool with Go, may not have full parity with ruby version, but it works for some products

```release-note:none

```


c2thorn (changes requested):
https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/openapi_generate/parser.go#L165-L169 needs to be deleted

slevenick:
> https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/openapi_generate/parser.go#L165-L169 needs to be deleted

Done, also added copyright header writing, version stripping and updated our YAML lint to allow the spacing used by Go YAML marshaling

c2thorn (approved):

PR #12040, opened 2024-10-17
style: update formatting for go code and add CI check

wyardley:
Some go files in the project were not formatted. Add a CI check which should fail when files are not formatted; a second commit will then pass with the `gofmt -w` fixes.

Failing check without all files formatted:
https://github.com/GoogleCloudPlatform/magic-modules/actions/runs/11391921639/job/31696780814?pr=12040

```release-note:none

```


.github/workflows/gofmt.yml
```diff
@@ -0,0 +1,33 @@
...
+          go-version: '^1.21'
+
+      - name: gofmt
+        run: |
+          GOFMT_OUTPUT="$(gofmt -l .)"
+          # Currently no simple way to get a non-0 exit code directly from
+          # gofmt
+          if [ -n "$GOFMT_OUTPUT" ]; then 
+            echo "The following files are not formatted properly:" >&2
+            echo "$GOFMT_OUTPUT" >&2
```
wyardley, line 30: Happy to do this a different way if you have a better way to do this.

Some loose inspiration from https://github.com/Jerome1337/gofmt-action (MIT licensed), but figured this is simple enough to do that it's overkill to include a third party action, and I've made some adjustments anyway.

.github/workflows/gofmt.yml
```diff
@@ -0,0 +1,33 @@
...
+on:
+  pull_request:
+    paths:
+      - "**/*.go"
+
+jobs:
+  run_gofmt:
+    runs-on: ubuntu-22.04
+    steps:
+      - uses: actions/checkout@eef61447b9ff4aafe5dcd4e0bbf5d482be7e7871 # v4.2.1
```
ScottSuarez, line 15: Can we merge with the base branch before running this check?
https://github.com/GoogleCloudPlatform/magic-modules/blob/862a4608eae0216ab1a5300af9638d51e0b12ae4/.github/workflows/mmv1-check-templates.yml#L19-L27

Also can you pull the latest changes in from main?

ScottSuarez (changes requested):

.github/workflows/gofmt.yml
wyardley: I added that in a new commit, and also rebased. Let me know if that looks right to you.

wyardley:
I assume `contributor-membership-checker` being cancelled (and thus causing the failed status) is expected?

.github/workflows/gofmt.yml
wyardley: This would basically have the effect of making sure it runs against new changes in `main` when the base is stale, but only in the case where a new commit is pushed, right?

_presumably_ if the check is already running on all other changes, I'd think there shouldn't be (m)any cases where this would be at issue, since the go files would already be have failed the check in their own branch if they were formatted poorly (and note that this check will only run if `.go` (vs. `.go.tmpl`) files are modified in a given PR. But it's not hard to add, and seems to work 👍

ScottSuarez: Yeah correct.

It's more of an edge case where if we check in a change and this check is broken on a certain commit it's easy for use to fix forward for existing PRs without needing them to rebase.

.github/workflows/gofmt.yml
```diff
@@ -0,0 +1,41 @@
...
+permissions:
+  contents: read
+
+on:
+  pull_request:
+    paths:
+      - "**/*.go"
+
+jobs:
+  run_gofmt:
```
ScottSuarez, line 12:
```suggestion
  run-gofmt:
```

ScottSuarez:
I needed to make some changes to allow this to go through. This should run now

/gcbrun

ScottSuarez (approved):

PR #12049, opened 2024-10-17
Add a magician command for reassigning reviewers

trodge:
This command will allow reassigning a reviewer either to a random core contributor or to a specific one.

Currently can be invoked with
```
cd .ci/magician
go build .
./magician reassign-reviewer PR_NUMBER [REVIEWER]
```
(REVIEWER is optional, omitting will assign to a random core reviewer)

helps with https://github.com/hashicorp/terraform-provider-google/issues/19691

Remaining TODOs:
* Add a workflow that runs `reassign-reviewer` on `@modular-magician reassign-reviewer [reviewer]`
* Make the re-request reviewer workflow check the current reviewer using `FindReviewerComment`

```release-note:none

```

### DIFF
--- a/.ci/magician/cmd/reassign_reviewer.go
+++ b/.ci/magician/cmd/reassign_reviewer.go
@@ -16,6 +16,7 @@
 package cmd
 
 import (
+	"errors"
 	"fmt"
 	"magician/github"
 
@@ -24,7 +25,7 @@ import (
 
 // reassignReviewerCmd represents the reassignReviewer command
 var reassignReviewerCmd = &cobra.Command{
-	Use:   "reassign-reviewer PR_NUMBER",
+	Use:   "reassign-reviewer PR_NUMBER [REVIEWER]",
 	Short: "Reassigns primary reviewer to the given reviewer or a random reviewer if none given",
 	Long: `This command reassigns reviewers when invoked via a comment on a pull request.
 
@@ -35,6 +36,7 @@ var reassignReviewerCmd = &cobra.Command{
 
 	It then performs the following operations:
 	1. Updates the reviewer comment to reflect the new primary reviewer.
+	2. Requests a review from the new primary reviewer.
 	`,
 	Args: cobra.MinimumNArgs(1),
 	RunE: func(cmd *cobra.Command, args []string) error {
@@ -63,20 +65,17 @@ func execReassignReviewer(prNumber, newPrimaryReviewer string, gh GithubClient) 
 	reviewerComment, currentReviewer := github.FindReviewerComment(comments)
 
 	if currentReviewer == "" {
-		return fmt.Errorf("no reviewer comment found in PR %s", prNumber)
-	}
-
-	if newPrimaryReviewer == "" {
-		newPrimaryReviewer = github.GetNewRandomReviewer(currentReviewer)
-	}
-
-	if currentReviewer == newPrimaryReviewer {
-		return fmt.Errorf("primary reviewer is already %s", newPrimaryReviewer)
-	}
-
-	err = gh.UpdateComment(prNumber, github.FormatReviewerComment(newPrimaryReviewer), reviewerComment.ID)
-	if err != nil {
-		return err
+		newPrimaryReviewer, err = createReviewComment(prNumber, newPrimaryReviewer, gh)
+		if err != nil {
+			return err
+		}
+		fmt.Println("New primary reviewer is ", newPrimaryReviewer)
+	} else {
+		newPrimaryReviewer, err = updateReviewComment(prNumber, currentReviewer, newPrimaryReviewer, reviewerComment.ID, gh)
+		if err != nil {
+			return err
+		}
+		fmt.Println("New primary reviewer is ", newPrimaryReviewer)
 	}
 
 	err = gh.RequestPullRequestReviewers(prNumber, []string{newPrimaryReviewer})
@@ -85,6 +84,40 @@ func execReassignReviewer(prNumber, newPrimaryReviewer string, gh GithubClient) 
 	}
 
 	return nil
+}
+
+func createReviewComment(prNumber, newPrimaryReviewer string, gh GithubClient) (string, error) {
+	fmt.Println("No reviewer comment found, creating one")
+	if newPrimaryReviewer == "" {
+		newPrimaryReviewer = github.GetRandomReviewer()
+	}
+
+	if newPrimaryReviewer == "" {
+		return "", errors.New("no primary reviewer found")
+	}
+
+	err := gh.PostComment(prNumber, github.FormatReviewerComment(newPrimaryReviewer))
+	if err != nil {
+		return "", err
+	}
+	return newPrimaryReviewer, nil
+}
+
+func updateReviewComment(prNumber, currentReviewer, newPrimaryReviewer string, reviewerCommentID int, gh GithubClient) (string, error) {
+	fmt.Println("Reassigning to random reviewer")
+	if newPrimaryReviewer == "" {
+		newPrimaryReviewer = github.GetNewRandomReviewer(currentReviewer)
+	}
+
+	if currentReviewer == newPrimaryReviewer {
+		return newPrimaryReviewer, fmt.Errorf("primary reviewer is already %s", newPrimaryReviewer)
+	}
+
+	err := gh.UpdateComment(prNumber, github.FormatReviewerComment(newPrimaryReviewer), reviewerCommentID)
+	if err != nil {
+		return "", err
+	}
+	return newPrimaryReviewer, nil
 }
 
 func init() {
--- a/.ci/magician/cmd/reassign_reviewer.go
+++ b/.ci/magician/cmd/reassign_reviewer.go
@@ -57,7 +57,7 @@ var reassignReviewerCmd = &cobra.Command{
 }
 
 func execReassignReviewer(prNumber, newPrimaryReviewer string, gh GithubClient) error {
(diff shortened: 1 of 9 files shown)


.ci/magician/github/reviewer_assignment.go
```diff
@@ -66,3 +67,24 @@ func FormatReviewerComment(newPrimaryReviewer string) string {
...
 	return sb.String()
 }
+
+var reviewerCommentRegex = regexp.MustCompile("@(?P<reviewer>[^,]*), a repository maintainer, has been assigned to review your changes.")
+
+// FindReviewerComment returns the comment which mentions the current primary reviewer and the reviewer's login,
+// or an empty comment and empty string if no such comment is found.
+// comments should only include comments by the magician in the current PR.
+func FindReviewerComment(comments []PullRequestComment) (PullRequestComment, string) {
+	for _, comment := range comments {
```
c2thorn, line 77: what order are the comments returned in? first -> last?

If so, should it be reversed so that the latest reviewer comment is used, and not the first

c2thorn (approved):
it'd be great to get this into a PR-command like gcbrun

LGTM, left minor comment about reviewer comment order

.ci/magician/github/reviewer_assignment.go
trodge: Not sure if it's returned in any particular order. Probably the best thing would be to compare the creation times.

c2thorn: Just minor detail, small chance the user may have to run the command twice if there's been multiple reviewers

trodge:
> it'd be great to get this into a PR-command like gcbrun
> 
> LGTM, left minor comment about reviewer comment order

The plan is to use a workflow that triggers on PR comments to invoke this command.

.ci/magician/cmd/reassign_reviewer.go
```diff
@@ -0,0 +1,125 @@
...
+	It then performs the following operations:
+	1. Updates the reviewer comment to reflect the new primary reviewer.
+	2. Requests a review from the new primary reviewer.
+	`,
+	Args: cobra.MinimumNArgs(1),
+	RunE: func(cmd *cobra.Command, args []string) error {
+		prNumber := args[0]
+		fmt.Println("PR Number: ", prNumber)
+
+		githubToken, ok := lookupGithubTokenOrFallback("GITHUB_TOKEN_MAGIC_MODULES")
```
melinath, line 46: This shouldn't need the fallback

.ci/magician/cmd/reassign_reviewer.go
```diff
@@ -0,0 +1,125 @@
...
+		var newPrimaryReviewer string
+		if len(args) > 1 {
+			newPrimaryReviewer = args[1]
+		}
+		return execReassignReviewer(prNumber, newPrimaryReviewer, gh)
+	},
+}
+
+func execReassignReviewer(prNumber, newPrimaryReviewer string, gh GithubClient) error {
+	comments, err := gh.GetPullRequestCommentsByUser(prNumber, "modular-magician")
```
melinath, line 60:
```suggestion
	comments, err := gh.GetPullRequestCommentsByUser(prNumber, "github-actions")
```

.ci/magician/github/get.go
```diff
@@ -98,6 +104,33 @@ func (gh *Client) GetPullRequestPreviousReviewers(prNumber string) ([]User, erro
...
+
+	var comments []PullRequestComment
+	err := utils.RequestCall(url, "GET", gh.token, &comments, nil)
+	if err != nil {
+		return nil, err
+	}
+	return comments, nil
+}
+
+func (gh *Client) GetPullRequestCommentsByUser(prNumber, login string) ([]PullRequestComment, error) {
```
melinath, line 118: Are we expecting to reuse this? What would you think about just keeping it to GetPullRequestComments (which will be easier to port over to the standard GH client once we get around to that conversion) and handling the rest as part of FindReviewerComment?

.ci/magician/github/reviewer_assignment.go
```diff
@@ -66,3 +67,24 @@ func FormatReviewerComment(newPrimaryReviewer string) string {
 	})
 	return sb.String()
 }
+
+var reviewerCommentRegex = regexp.MustCompile("@(?P<reviewer>[^,]*), a repository maintainer, has been assigned to review your changes.")
```
melinath, line 71: We should definitely have a test that verifies that this matches the output of the function that formats this comment.

.ci/magician/github/reviewer_assignment.go
melinath: @c2thorn there should only ever be one comment that contains the reviewer - the proposed implementation is to update that comment when the reviewer is reassigned (so that we're not creating more noise in the PR). I don't _think_ there's a way for us to end up with more than one, but FWIW if it ends up being a common issue I'd prefer to delete or hide the duplicates.

melinath (changes requested):

.ci/magician/cmd/reassign_reviewer.go
trodge: Going to remove the filter altogether because the comment could come from modular magician if we have to make a new comment because none was found (as is the case for PRs from core contributors).

.ci/magician/github/get.go
trodge: Removing this altogether since reviewer comment doesn't necessarily come from a specific user.

.ci/magician/github/reviewer_assignment.go
trodge: Done

.ci/magician/github/reviewer_assignment.go
trodge: Just in case, I've made it return the most recent reviewer comment.

.ci/magician/cmd/reassign_reviewer.go
melinath: FWIW I believe that (since this is going to run in a GHA) the comment will always come from `github-actions`. But skipping the check is also probably fine.

.ci/magician/cmd/reassign_reviewer.go
```diff
@@ -0,0 +1,126 @@
...
+	It then performs the following operations:
+	1. Updates the reviewer comment to reflect the new primary reviewer.
+	2. Requests a review from the new primary reviewer.
+	`,
+	Args: cobra.MinimumNArgs(1),
+	RunE: func(cmd *cobra.Command, args []string) error {
+		prNumber := args[0]
+		fmt.Println("PR Number: ", prNumber)
+
+		githubToken, ok := os.LookupEnv("GITHUB_TOKEN_MAGIC_MODULES")
```
melinath, line 47: I think this should just use GITHUB_TOKEN since it'll run in the context of a GHA.

```suggestion
		githubToken, ok := os.LookupEnv("GITHUB_TOKEN")
```

.ci/magician/cmd/reassign_reviewer.go
```diff
@@ -0,0 +1,126 @@
...
+	err = gh.RequestPullRequestReviewers(prNumber, []string{newPrimaryReviewer})
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func createReviewComment(prNumber, newPrimaryReviewer string, gh GithubClient) (string, error) {
+	fmt.Println("No reviewer comment found, creating one")
```
melinath, line 91: nit: this kind of logging should happen in the caller, since that's where the decision about why this was called was made.

.ci/magician/github/reviewer_assignment.go
melinath: IMO It should take the oldest reviewer comment rather than the most recent. But since there should only be one it's not a blocker for me at this time.

.ci/magician/github/reviewer_assignment.go
```diff
@@ -66,3 +67,33 @@ func FormatReviewerComment(newPrimaryReviewer string) string {
...
+		if len(matches) < len(names) {
+			// Skip comments that don't match regex.
+			continue
+		}
+		for i, name := range names {
+			if name == "reviewer" {
+				newestComment = comment
+				currentReviewer = matches[i]
+			}
+		}
```
melinath, line 96: I think this could be simplified a bit.

```suggestion
		matches := reviewerCommentRegex.FindStringSubmatch(comment.Body)
		fmt.Println(comment, matches)
		if matches == nil {
			// Skip comments that don't match regex.
			continue
		}
		newestComment = comment
		currentReviewer = matches[1]
```

melinath (approved):

.ci/magician/cmd/reassign_reviewer.go
trodge: Done

.ci/magician/github/reviewer_assignment.go
melinath: I hear you, and I wish golang provided an easier way to access the matches by group name, but this implementation is currently a bit convoluted considering there's only one subgroup in the regex. It would be a bit more clear if you did something like [make a map of all the named groups and then extracted the value from that](https://stackoverflow.com/questions/20750843/using-named-matches-from-go-regex) but that also feels pretty heavy for what's happening here.

melinath: We could factor the map logic out into a separate util library for handling regex but this particular case doesn't provide a strong justification for it.

melinath: Just to close this discussion out - I'm not blocking the merge based on this.

PR #12050, opened 2024-10-18
Allow proper changing of google_compute_region_instance_group_manager.distribution_policy_target_shape

askubis:
Allow proper changing of google_compute_region_instance_group_manager.distribution_policy_target_shape. Effectively - don't send the zones distribution when ONLY target shape changes.

fixes [#19891](https://github.com/hashicorp/terraform-provider-google/issues/19891)

```release-note:bug
compute: fixed an issue where immutable `distribution_zones` was incorrectly sent to the API when updating `distribution_policy_target_shape` in `google_compute_region_instance_group_manager` resource
```

### DIFF
--- a/mmv1/third_party/terraform/services/compute/resource_compute_region_instance_group_manager.go.tmpl
+++ b/mmv1/third_party/terraform/services/compute/resource_compute_region_instance_group_manager.go.tmpl
@@ -1164,7 +1164,7 @@ func expandDistributionPolicy(d *schema.ResourceData) *compute.DistributionPolic
     }
 	  return &compute.DistributionPolicy{Zones: distributionPolicyZoneConfigs, TargetShape: dpts}
 	}
-	return &compute.DistributionPolicy{TargetShape: dpts}
+    return &compute.DistributionPolicy{TargetShape: dpts}
 }
 
 func flattenDistributionPolicy(distributionPolicy *compute.DistributionPolicy) []string {
--- a/mmv1/third_party/terraform/services/compute/resource_compute_region_instance_group_manager.go.tmpl
+++ b/mmv1/third_party/terraform/services/compute/resource_compute_region_instance_group_manager.go.tmpl
@@ -633,7 +633,7 @@ func resourceComputeRegionInstanceGroupManagerCreate(d *schema.ResourceData, met
 		UpdatePolicy:                expandRegionUpdatePolicy(d.Get("update_policy").([]interface{})),
 		InstanceLifecyclePolicy:     expandInstanceLifecyclePolicy(d.Get("instance_lifecycle_policy").([]interface{})),
 		AllInstancesConfig:          expandAllInstancesConfig(nil, d.Get("all_instances_config").([]interface{})),
-		DistributionPolicy:          expandDistributionPolicy(d),
+		DistributionPolicy:          expandDistributionPolicyForCreate(d),
 		StatefulPolicy:              expandStatefulPolicy(d),
 		{{- if ne $.TargetVersionName "ga" }}
 		Params:                      expandInstanceGroupManagerParams(d),
@@ -907,7 +907,7 @@ func resourceComputeRegionInstanceGroupManagerUpdate(d *schema.ResourceData, met
 	}
 
 	if d.HasChange("distribution_policy_target_shape") {
-		updatedManager.DistributionPolicy = expandDistributionPolicy(d)
+		updatedManager.DistributionPolicy = expandDistributionPolicyForUpdate(d)
 		change = true
 	}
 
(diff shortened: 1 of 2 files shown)


mmv1/third_party/terraform/services/compute/resource_compute_region_instance_group_manager.go.tmpl
```diff
@@ -1152,17 +1152,19 @@ func expandDistributionPolicy(d *schema.ResourceData) *compute.DistributionPolic
...
+    distributionPolicyZoneConfigs := make([]*compute.DistributionPolicyZoneConfiguration, 0, dpz.Len())
+    for _, raw := range dpz.List() {
+      data := raw.(string)
+      distributionPolicyZoneConfig := compute.DistributionPolicyZoneConfiguration{
+        Zone: "zones/" + data,
+      }
+
+      distributionPolicyZoneConfigs = append(distributionPolicyZoneConfigs, &distributionPolicyZoneConfig)
+    }
+	  return &compute.DistributionPolicy{Zones: distributionPolicyZoneConfigs, TargetShape: dpts}
```
pjotrekk, line 1165: The indentation doesn't seem right.

mmv1/third_party/terraform/services/compute/resource_compute_region_instance_group_manager_test.go.tmpl
```diff
@@ -329,6 +338,15 @@ func TestAccRegionInstanceGroupManager_distributionPolicy(t *testing.T) {
 				ImportStateVerify:       true,
 				ImportStateVerifyIgnore: []string{"status"},
 			},
+			{
+				Config: testAccRegionInstanceGroupManager_distributionPolicyEmpty(template, igm),
```
pjotrekk, line 342: I feel stupid that I don't know this but does this test tests the updates or only creation?

pjotrekk (approved):

mmv1/third_party/terraform/services/compute/resource_compute_region_instance_group_manager_test.go.tmpl
askubis: I think we should generally follow the pattern (in tests):
1. empty/unspecified field
2. some value
3. other value
4. empty/unspecified (to check if we're not overwriting/correctly deleting)
Generally we're applying all the configurations specified in testAccRegionInstanceGroupManager_distributionPolicy (including the one you commented on), and checking if nothing seems to be failing.

mmv1/third_party/terraform/services/compute/resource_compute_region_instance_group_manager.go.tmpl
askubis: thank you, fixed!

pjotrekk (approved):

pjotrekk (approved):

mmv1/third_party/terraform/services/compute/resource_compute_region_instance_group_manager.go.tmpl
```diff
@@ -1152,17 +1152,19 @@ func expandDistributionPolicy(d *schema.ResourceData) *compute.DistributionPolic
...
-		data := raw.(string)
-		distributionPolicyZoneConfig := compute.DistributionPolicyZoneConfiguration{
-			Zone: "zones/" + data,
-		}
-
-		distributionPolicyZoneConfigs = append(distributionPolicyZoneConfigs, &distributionPolicyZoneConfig)
-	}
-
-	return &compute.DistributionPolicy{Zones: distributionPolicyZoneConfigs, TargetShape: dpts}
+  if d.HasChange("distribution_policy_zones") {
```
zli82016, line 1155: It is better to not include the logic checking changes inside the function `func expandDistributionPolicy`. We prefer to separate those.

The field change checking for `distribution_policy_zones` should be put into `func resourceComputeRegionInstanceGroupManagerUpdate` around the lines [if d.HasChange("distribution_policy_target_shape")
](https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/third_party/terraform/services/compute/resource_compute_region_instance_group_manager.go.tmpl#L909-L912)

askubis: PTAL at my changes.
I think I understand your request - we use the same method in Create (where having a change seems weird), and in update.
My idea for the change is that since distribution_policy.zones are immutable, then we will never want to send them in the update request. The field is also marked as ForceNew, so if there is any change in the distribution_policy_zones, then we will destroy the resource and create it again.

PR #12051, opened 2024-10-18
Modified Internal Range acceptance tests for Network Connectivity to check for expected fields

dorianverna:
This PR aims to modify the Internal Range acceptance test as follows:

1. For the basic example we should be checking whether all the fields in the expected Internal Range object have the desired values.
2. For the External Range example we should be checking whether all the fields in the expected Internal Range object have the desired values
3. The description field should be checked that it was updated.

Note that:
- The initial purpose of this PR was to modify the tests as to be able to verify that the "description" property of an Internal Range is editable
- The Internal Range source code change is not fully rolled out yet, so the tests here are not expected to pass for now
- The PR is to remain on hold until the Internal Range description editability change is fully rolled out

```release-note:none
```

+CC @sebkalis

### DIFF
--- a/mmv1/third_party/terraform/services/networkconnectivity/resource_network_connectivity_internal_range_test.go
+++ b/mmv1/third_party/terraform/services/networkconnectivity/resource_network_connectivity_internal_range_test.go
@@ -1,9 +1,10 @@
 package networkconnectivity_test
 
 import (
+	"testing"
+
 	"github.com/hashicorp/terraform-plugin-testing/helper/resource"
 	"github.com/hashicorp/terraform-provider-google/google/acctest"
-	"testing"
 )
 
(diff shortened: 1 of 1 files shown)


c2thorn:
Hi @dorianverna,
Can you ping me here when the feature has been rolled out? I'll start the tests and build then.

dorianverna:
Hey, @c2thorn!

The change is rolled out. I tested this PR and everything seems to be okay. Can you review it?
Thanks!

c2thorn:
> Hey, @c2thorn!
> 
> The change is rolled out. I tested this PR and everything seems to be okay. Can you review it? Thanks!

great, I'll start the CI build

c2thorn (approved):
LGTM, tests have passed as well. Thanks!

PR #12054, opened 2024-10-18
Fixed SecurityPolicyRule and RegionSecurityPolicyRule resources being unable to manage the policy default rule

matheusaleixo-cit:
Adds custom code to allow the `google_compute_region_security_policy_rule` and `google_compute_security_policy_rule` resources to override the default rule of their respective security policy without the need of a multi-apply import process.

Fixes: https://github.com/hashicorp/terraform-provider-google/issues/15687

```release-note:bug
compute: fixed unable to create default rule when using `google_compute_region_security_policy_rule` resource (beta)
```
```release-note:bug
compute: fixed unable to create default rule when using `google_compute_security_policy_rule` resource
```

### DIFF
--- a/mmv1/products/compute/RegionSecurityPolicyRule.yaml
+++ b/mmv1/products/compute/RegionSecurityPolicyRule.yaml
@@ -67,6 +67,7 @@ examples:
       sec_policy_name: 'policywithmultiplerules'
   - name: 'region_security_policy_rule_default_rule'
     primary_resource_id: 'policy_rule'
+    min_version: 'beta'
     vars:
       sec_policy_name: 'policywithdefaultrule'
   - name: 'region_security_policy_rule_with_preconfigured_waf_config'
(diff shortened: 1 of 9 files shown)


rileykarson:
Sorry, been behind on reviews. Rolled a new reviewer.

zli82016:
@melinath , it looks like both PR, https://github.com/GoogleCloudPlatform/magic-modules/pull/12054 and  https://github.com/GoogleCloudPlatform/magic-modules/pull/11894 are trying to fix the same Github issue with different ways. They both look reasonable to me. What is your thought? Thanks.

melinath:
FWIW to me this feels like a reasonable situation for acquire-on-create behavior. That said, I don't think these PRs necessarily conflict & having both seems like it's probably good for users.

zli82016 (approved):
LGTM. Thanks.

PR #12055, opened 2024-10-18
Add gemini_enabled field to google_looker_instance.

cornmander:
Adds gemini_enabled attribute to google_looker_instance.

Fixes https://github.com/hashicorp/terraform-provider-google/issues/19913.

```release-note:enhancement
looker: added `gemini_enabled` field to `google_looker_instance` resource
```


roaks3 (commented):
@cornmander would it be possible to add this to one of the existing tests, to confirm it works as expected? It also looks like you've defined it to be mutable, so ideally we could also test that it can be changed on/off.

PR #12056, opened 2024-10-18
Fix issue with google_compute_(region_)per_instance_config.stateful.ips.ipAddress.address always having diff due to relative vs full path

### LINKED ISSUE #19265: Specifying IP address as preserved state in `google_compute_region_per_instance_config` uses fully qualified URL
### Community Note

Terraform v1.7.5
on linux_amd64
+ provider registry.terraform.io/hashicorp/google v5.23.0

### Affected Resource(s)

`google_compute_region_per_instance_config`

### Terraform Configuration

```tf
resource "google_compute_address" "pgbouncer_instance_ip_address" {
  name         = "${var.service_name}-pgbouncer-instance-ip-address"
  subnetwork   = "default"
  address_type = "INTERNAL"
  address      = var.instance_ip_addresses[0]
  region       = var.region
}

resource "google_compute_region_per_instance_config" "instance" {
  region_instance_group_manager = google_compute_region_instance_group_manager.pgbouncer_managed_instance_group.name
  region                        = google_compute_region_instance_group_manager.pgbouncer_managed_instance_group.region
  name                          = "${var.service_name}-pgbouncer-instance"
  preserved_state {
    internal_ip {
      interface_name = "nic0"
      auto_delete    = "NEVER"
      ip_address {
        address = google_compute_address.pgbouncer_instance_ip_address.id
      }
    }

    metadata = {
      instance_template = google_compute_instance_template.pgbouncer_template.self_link
    }
  }
}
```

### Debug Output

_No response_

### Expected Behavior

Following the example in the docs [here](https://cloud.google.com/compute/docs/instance-groups/configuring-stateful-ip-addresses-in-migs#configuring_static_ip_addresses_on_vm_creation_in_a_mig), I used `google_compute_address.pgbouncer_instance_ip_address.id` to set the `address` field in the preserved state. I expect this to be a valid configuration.

### Actual Behavior

I'm able to apply the configuration and it works as expected, but now on every subsequent `terraform apply`, Terraform applies the following action:
```
  # module.db_connection_pooler.google_compute_region_per_instance_config.instance_1 will be updated in-place
  ~ resource "google_compute_region_per_instance_config" "instance" {
        id                               = "service-id"
        name                             = "service-name"
        # (7 unchanged attributes hidden)

      ~ preserved_state {

          - internal_ip {
              - auto_delete    = "NEVER" -> null
              - interface_name = "nic0" -> null

              - ip_address {
                  - address = "https://www.googleapis.com/compute/v1/projects/my-project/regions/us-east4/addresses/service-pgbouncer-instance-ip-address" -> null
                }
            }
          + internal_ip {
              + auto_delete    = "NEVER"
              + interface_name = "nic0"

              + ip_address {
                  + address = "projects/my-project/regions/us-east4/addresses/service-pgbouncer-instance-ip-address"
                }
            }
        }
    }
```
The resource is being updated in-place because locally I am supplying `address = "projects/my-project/regions/us-east4/addresses/service-pgbouncer-instance-ip-address"`, but the remote state has `address = "https://www.googleapis.com/compute/v1/projects/my-project/regions/us-east4/addresses/service-pgbouncer-instance-ip-address"`.

I expect that the configuration state remains stable and I don't need to apply this change every time. Am I supposed to manually construct the full URL above to prevent this from happening?

### Steps to reproduce

1. `terraform apply` with the provided config
2. Make some other change to the environment
3. `terraform apply` again

### Important Factoids

_No response_

### References

_No response_

askubis:
Convert the full path to 

fixes https://github.com/hashicorp/terraform-provider-google/issues/19265

```release-note:bug
compute: fixed permadiffs on stateful ip address in `google_compute_region_per_instance_config` and `google_compute_per_instance_config` resources
```

### DIFF
--- a/mmv1/products/compute/PerInstanceConfig.yaml
+++ b/mmv1/products/compute/PerInstanceConfig.yaml
@@ -15,7 +15,7 @@
 name: 'PerInstanceConfig'
 description: |
   A config defined for a single managed instance that belongs to an instance group manager. It preserves the instance name
-  across instance group manager operations and can define stateful disks or metadata that are unique to the instance.
+  across instance group manager operations and can define stateful disks, ips or metadata that are unique to the instance.
 references:
   guides:
     'Official Documentation': 'https://cloud.google.com/compute/docs/instance-groups/stateful-migs#per-instance_configs'
@@ -213,6 +213,7 @@ properties:
                   type: ResourceRef
                   description: |
                     The URL of the reservation for this IP address.
+                  custom_flatten: 'templates/terraform/custom_flatten/full_to_relative_path.tmpl'
                   custom_expand: 'templates/terraform/custom_expand/resourceref_with_validation.go.tmpl'
                   resource: 'Address'
                   imports: 'selfLink'
@@ -243,6 +244,7 @@ properties:
                   type: ResourceRef
                   description: |
                     The URL of the reservation for this IP address.
+                  custom_flatten: 'templates/terraform/custom_flatten/full_to_relative_path.tmpl'
                   custom_expand: 'templates/terraform/custom_expand/resourceref_with_validation.go.tmpl'
                   resource: 'Address'
                   imports: 'selfLink'
--- a/mmv1/products/compute/PerInstanceConfig.yaml
+++ b/mmv1/products/compute/PerInstanceConfig.yaml
@@ -213,8 +213,8 @@ properties:
                   type: ResourceRef
                   description: |
                     The URL of the reservation for this IP address.
-                  custom_flatten: 'templates/terraform/custom_flatten/full_to_relative_path.tmpl'
-                  custom_expand: 'templates/terraform/custom_expand/resourceref_with_validation.go.tmpl'
+                  diff_suppress_func: 'tpgresource.CompareSelfLinkRelativePaths'
+                  #custom_expand: 'templates/terraform/custom_expand/resourceref_with_validation.go.tmpl'
                   resource: 'Address'
                   imports: 'selfLink'
       - name: 'externalIp'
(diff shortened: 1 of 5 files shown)


pjotrekk (approved):

askubis:
@shuyama1 is this considered a breaking change? Since we had to modify tests I would imagine it is, but please share your take on this :)

shuyama1 (commented):
> @shuyama1 is this considered a breaking change? Since we had to modify tests I would imagine it is, but please share your take on this :)

Thanks for making the change. Let me confirm if I understand the case correctly - 
- Before the change:
	we can send both full and relative path to the API, but since the API only returns the full URL, we'll run into permadiffs for cases where relative path is specified.

- After the change: 
  Since we add a custom flattener, it will always write the relative path to the state, which will result in permadiff for cases where full path is specified.
  

If I understand the case correctly, yes, this change will break current configs. We'll need to find a solution ensure that sending either full or relative path should result in no permadiffs in Terraform. I'm thinking that adding a [diff suppress function](https://googlecloudplatform.github.io/magic-modules/develop/permadiff/#normalized-value) might be working in this situation, likely you can use the existing function [`CompareSelfLinkRelativePaths`](https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/third_party/terraform/tpgresource/self_link_helpers.go#L)

Please let me know if you have any questions.

shuyama1 (commented):
> I think I need some help with that change. I modified the tests first - both .self_link and .id should work (because why not, one is the current state, the other is what we recommend in the documentation) I did what was recommended - added the diff_supress_func. I even added tests to it, so that it would be clear that it does what expected ( "partial path, same": { Old: "[googleapis.com/compute/v1/projects/your-project/global/networks/a-network](https://www.googleapis.com/compute/v1/projects/your-project/global/networks/a-network)", New: "projects/your-project/global/networks/a-network", Expect: true, },) And yet, still we have the same diff when running the test. ` - internal_ip { - auto_delete = "NEVER" -> null - interface_name = "nic0" -> null
> 
> ```
>                   - ip_address {
>                       - address = "https://www.googleapis.com/compute/v1/projects/REDACTED/regions/us-central1/addresses/tf-test-rigm-address97pzzqv6th" -> null
>                     }
>                 }
>               + internal_ip {
>                   + auto_delete    = "NEVER"
>                   + interface_name = "nic0"
>     
>                   + ip_address {
>                       + address = "projects/REDACTED/regions/us-central1/addresses/tf-test-rigm-address97pzzqv6th"
>                     }
>                 }
> ```
> 
> `
> 
> any idea what could I be doing wrong here?

Okay, I’ve looked into this issue a bit and found that the DSF not working is likely due to a known problem with the SDK https://github.com/hashicorp/terraform-plugin-sdk/issues/477. It seems we may need to write a custom DSF for its parent set field. I’m trying out some solutions locally and will share updates as I progress.

PR #12061, opened 2024-10-18
Add mmv1 metadata generation

roaks3:
This adds generator code for producing metadata yaml files, which will contain information about how the TF resource maps to an API resource. These files will be committed directly to the downstream providers.

Some notes:
- `cai_base_url` is known to be a confusing name, but we're keeping it for now to minimize changes
- API versions require parsing the urls, and while they are most often specified on the service `base_url`, there are some resources that specify them on the _resource_ `base_url` (to support multiple versions within the service)

```release-note:none
provider: added mmv1 metadata files
```

### DIFF
--- a/mmv1/api/product.go
+++ b/mmv1/api/product.go
@@ -16,6 +16,7 @@ package api
 import (
 	"log"
 	"reflect"
+	"regexp"
 	"strings"
 	"unicode"
 
@@ -51,6 +52,10 @@ type Product struct {
 	// The base URL for the service API endpoint
 	// For example: `https://www.googleapis.com/compute/v1/`
 	BaseUrl string `yaml:"base_url,omitempty"`
+
+	// The validator "relative URI" of a resource, relative to the product
+	// base URL. Specific to defining the resource as a CAI asset.
+	CaiBaseUrl string
 
 	// A function reference designed for the rare case where you
 	// need to use retries in operation calls. Used for the service api
@@ -213,6 +218,7 @@ func (p *Product) ExistsAtVersion(name string) bool {
 
 func (p *Product) SetPropertiesBasedOnVersion(version *product.Version) {
 	p.BaseUrl = version.BaseUrl
+	p.CaiBaseUrl = version.CaiBaseUrl
 }
 
 func (p *Product) TerraformName() string {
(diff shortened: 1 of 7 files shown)


mmv1/templates/terraform/metadata.yaml.tmpl
```diff
@@ -0,0 +1,5 @@
+resource: '{{ $.TerraformName }}'
```
roaks3, line 1: One thing that just crossed my mind: I wonder if we should take any steps to remove these files from the provider before it is released and/or built. I wouldn't expect them to impact the resulting build, but maybe we don't want them in the release source code archives.

mmv1/templates/terraform/metadata.yaml.tmpl
```diff
@@ -0,0 +1,5 @@
...
+generation_type: 'mmv1'
+api_name: '{{ $.ProductMetadata.ServiceName }}'
```
roaks3, line 3: Should we change this to `api_service_name` to match go/unified-resource-type?

mmv1/templates/terraform/metadata.yaml.tmpl
melinath: yes, we should remove them automatically if a TF resource is deleted, for example. I'm not sure offhand if this would happen automatically already.

mmv1/templates/terraform/metadata.yaml.tmpl
melinath: yeah that seems reasonable to me

mmv1/api/product.go
```diff
@@ -222,6 +228,38 @@ func (p *Product) TerraformName() string {
...
 }
 
+func (p *Product) ServiceBaseUrl() string {
+	if p.CaiBaseUrl != "" {
+		return p.CaiBaseUrl
+	}
+	return p.BaseUrl
+}
+
+func (p *Product) ServiceName() string {
```
melinath, line 238: could you add some tests for this?

mmv1/api/product.go
```diff
@@ -222,6 +228,38 @@ func (p *Product) TerraformName() string {
...
+	parts := strings.Split(p.ServiceBaseUrl(), "/")
+	// remove location prefix if present
+	trimmed, _ := strings.CutPrefix(parts[2], "{{location}}-")
+	return trimmed
+}
+
+var versionRegexp = regexp.MustCompile(`^v[0-9]+|beta`)
+
+func (p *Product) ServiceVersion() string {
```
melinath, line 247: could you add tests for this as well?

mmv1/api/resource.go
```diff
@@ -492,6 +492,24 @@ func (r Resource) UserParameters() []*Type {
...
 }
 
+func (r Resource) ServiceVersion() string {
+	if r.CaiBaseUrl != "" {
+		return extractVersionFromBaseUrl(r.CaiBaseUrl)
+	}
+	return extractVersionFromBaseUrl(r.BaseUrl)
+}
+
+func extractVersionFromBaseUrl(baseUrl string) string {
```
melinath, line 502: Could you add tests for this? (Possibly it could be combined with the other version extraction function?)

melinath (approved):
a few comments but overall lgtm

mmv1/templates/terraform/metadata.yaml.tmpl
roaks3: I was talking about the release artifacts here, but after another look, I don't think it's an issue to keep the yaml files.

For deleting these yaml files when the resources are removed, I confirmed this is covered under our current process (basically, `make clean-provider`)

PR #12062, opened 2024-10-18
Remove references to deleted compute images in multiple tests

SirGitsalot:
There are multiple tests that refer to preconfigured images in the `bosh-gce-raw-stemcells` storage bucket. That bucket was deleted recently, so the tests began failing. That bucket was at least five years old and we never controlled it (`bosh` is a Cloud Foundry tool), so it was bound to happen.

Replacing the missing bucket with another preconfigured bucket will just kick the can down the road, so the tests are updated to no longer require one. There's three different fixes applied in this PR:

- Where possible, the dependency on the `google_compute_image` resource was removed. These are tests that are not testing `google_compute_image` at all (and were likely copied and pasted from other tests that do).
- Uploading and referencing a `raw-disk-image.tar.gz` file, which is a valid but small disk image (it's mostly zeros so it compresses down to about 3KB). The tests using this method are replicating their earlier behavior, but without the preconfigured bucket.
- For generated tests, a new disk is created and the `google_compute_image` references that disk. This is a bit slower than the second method but doesn't rely on a test fixture so the tests can continue to be generated.

Fixes https://github.com/hashicorp/terraform-provider-google/issues/19885
Fixes https://github.com/hashicorp/terraform-provider-google/issues/19729

```release-note:none

```


NickElliot (approved):
changes LGTM!  how long is the actual slowdown in the generated tests with deploying new disks?

SirGitsalot:
Looking at the logs, it's about 10 seconds at the start of the test to create the disk, and another 10 seconds at the end to delete it.

PR #12064, opened 2024-10-21
Remove security_posture_config from example

hdp617:

```release-note:none

```


ScottSuarez (approved):

ScottSuarez:
This is being removed because the feature was rolled back on the API. Intermittently this has caused test failures for us. We expect this to be re-added once the feature is enabled again.

PR #12071, opened 2024-10-21
fix oracleDatabase datasource ref links

tulika-aakriti:
fix: https://github.com/hashicorp/terraform-provider-google/issues/19946


tulika-aakriti:
Hi @BBBmau, Can you get this merged before tomorrow's branch ? I know there is 2 business day rule but it would be nice get this one fix before next release.
Thanks.

BBBmau (approved):
only change i made was removing the release notes from the description since release notes are only meant to be added when the change is user facing. website updates fall under this. Thanks for adding this!

PR #12074, opened 2024-10-21
`vertex-ai`: add `google_vertex_ai_models` resource

BBBmau:
This initial PR will begin with supporting `:COPY`

replaces https://github.com/GoogleCloudPlatform/magic-modules/pull/10628

Will include PRs to support `:UPLOAD` in separate PRs.

- [ ] `:COPY`
- [ ] `:UPLOAD`
    - [ ] `explanation_spec_examples`
    - [ ] `explanation_gradients`
    - [ ] `explanation_xrai`
    - [ ] `explanation_shapley`

```release-note:new-resource
`google_vertex_ai_models`
```


BBBmau:
latest commit address some fields that would return a diff due to fields not being in the terraform configuration. We now get the following error which is due to attempting to use a source_model that does not exist.
```hcl
  | exit status 1
  | 
  | Error: Error waiting to create Models: Error waiting for Creating Models: Error code 5, message: NOT_FOUND
  | 
  |   with google_vertex_ai_models.model,
  |   on terraform_plugin_test.tf line 2, in resource "google_vertex_ai_models" "model":
  |    2: resource "google_vertex_ai_models" "model" {
```

Marking `:COPY` as completed from testing locally.

mmv1/templates/terraform/pre_create/vertex_ai_models.tmpl
```diff
@@ -0,0 +1,9 @@
+var postRequestType string
+if _, ok := d.GetOk("model"); ok {
+	postRequestType = ":upload"
+}
```
SarahFrench, line 4: Is there a way you can write this pre_create in a way that's specific to create-via-copy and doesn't have anything related to upload?

mmv1/products/vertexai/Models.yaml
```diff
@@ -0,0 +1,425 @@
...
+      The resource name of the Model to copy. That Model must be in the same Project
+  - type: ResourceRef
+    name: 'model'
+    resource: 'Model'
+    imports: 'displayName'
+    at_least_one_of:
+      - sourceModel
+      - model
+    description: |
+      The Model to create.
```
SarahFrench, line 420: I believe this field is only used for create-via-upload. Could you please remove this field (and any other fields specific to create-via-upload) from this PR and include them in your next PR that adds the new creation method?

mmv1/products/vertexai/Models.yaml
```diff
@@ -0,0 +1,425 @@
...
+  decoder: templates/terraform/decoders/vertex_ai_models.tmpl
+  pre_create: templates/terraform/pre_create/vertex_ai_models.tmpl
+examples:
+  - name: 'vertex_ai_model_source_basic'
+    exclude_import_test: true
+    primary_resource_id: 'model'
+    vars:
+      model_id: 'test-model'
+    test_env_vars:
+      project_name: 'PROJECT_NAME'
```
SarahFrench, line 54: Please check how this test is being generated- currently the config for the test includes:

```
resource "google_vertex_ai_models" "model" {
  display_name = ""
  project = ""
```

SarahFrench (commented):
Here's an initial review pass, mainly my feedback is to please ensure that this PR contains the minimal fields required for managing a model made via copy.

mmv1/templates/terraform/pre_create/vertex_ai_models.tmpl
BBBmau: removed

mmv1/products/vertexai/Models.yaml
BBBmau: you're referring to the generated test? I checked and getting the following generation:
```hcl
func testAccVertexAIModels_vertexAiModelSourceBasicExample(context map[string]interface{}) string {
	return acctest.Nprintf(`
resource "google_vertex_ai_models" "model" {
  model_id = "tf-test-test-model%{random_suffix}"
  project = "%{project_name}"
  source_model = "projects/%{project_name}/locations/us-central1/models/4469245519628878651"

  region       = "us-central1"
}
`, context)
}
```

mmv1/products/vertexai/Models.yaml
BBBmau: I removed as much as I could but source_model does depend on several fields since it needs a place to store the incoming fields that it it's copying from another model.

The response when a copy method is sent:
```hcl
{
    "name": "projects/871647908372/locations/us-central1/models/4741205189644517376",
    "displayName": "sampled_shapley_upload_container_spec",
    "predictSchemata": {},
    "containerSpec": {
        "imageUri": "us-docker.pkg.dev/vertex-ai/prediction/tf2-cpu.2-12:latest",
        "env": [
            {
                "name": "TF_CPP_MIN_LOG_LEVEL",
                "value": "2"
            },
            {
                "name": "AIP_STORAGE_URI",
                "value": "gs://your-bucket/model"
            }
        ],
        "ports": [
            {
                "containerPort": 8080
            }
        ],
        "predictRoute": "/v1/models/model/predict",
        "healthRoute": "/health",
        "grpcPorts": [
            {
                "containerPort": 8500
            }
        ],
        "deploymentTimeout": "300s",
        "sharedMemorySizeMb": "1024",
        "startupProbe": {
            "exec": {
                "command": [
                    "curl",
                    "http://localhost:8080/health"
                ]
            },
            "periodSeconds": 30,
            "timeoutSeconds": 10
        },
        "healthProbe": {
            "exec": {
                "command": [
                    "curl",
                    "http://localhost:8080/health"
                ]
            },
            "periodSeconds": 30,
            "timeoutSeconds": 5
        }
    },
    "supportedDeploymentResourcesTypes": [
        "DEDICATED_RESOURCES",
        "SHARED_RESOURCES"
    ],
    "supportedInputStorageFormats": [
        "jsonl",
        "bigquery",
        "csv",
        "tf-record",
        "tf-record-gzip",
        "file-list"
    ],
    "supportedOutputStorageFormats": [
        "jsonl",
        "bigquery"
    ],
    "createTime": "2024-10-31T01:08:20.493644Z",
    "updateTime": "2024-10-31T01:08:27.098115Z",
    "etag": "AMEw9yPNOVVwS7j2Pyg7Fo-1eJ6fAuJz24zjmF5JqdeXBAPXM8Ysr8nZtbIXW5bjoIc=",
    "supportedExportFormats": [
        {
            "id": "custom-trained",
            "exportableContents": [
                "ARTIFACT"
            ]
        }
    ],
    "explanationSpec": {
        "parameters": {
            "sampledShapleyAttribution": {
                "pathCount": 25
            }
        },
        "metadata": {}
    },
    "artifactUri": "gs://cloud-samples-data/ai-platform/mnist_tfrecord/pretrained",
    "versionId": "1",
    "versionAliases": [
        "default"
    ],
    "versionCreateTime": "2024-10-31T01:08:20.493644Z",
    "versionUpdateTime": "2024-10-31T01:08:27.053530Z",
    "originalModelInfo": {
        "model": "projects/871647908372/locations/us-central1/models/7131490691871408128@1"
    },
    "modelSourceInfo": {
        "sourceType": "CUSTOM",
        "copy": true
    }
}
```

mmv1/products/vertexai/Models.yaml
SarahFrench: Ah, I think I had a stale version of your branch pulled locally, sorry!

mmv1/templates/terraform/decoders/vertex_ai_models.go.tmpl
```diff
@@ -0,0 +1,16 @@
+if res["model"] != "" {
+	res["name"] = res["model"]
+}
```
SarahFrench, line 3: Why do we need to over-write the value of name here?

SarahFrench (changes requested):
Here's another review pass, with a bunch of suggested changes or comments that may lead to changes.

In addition to responding to those comments, please create 2 handwritten acceptance tests:

1. Provisioning a `google_vertex_ai_model` resource where `model_id` is set in the config
2. Provisioning a `google_vertex_ai_model` resource where `model_id` is NOT set in the config

mmv1/products/vertexai/Models.yaml
```diff
@@ -0,0 +1,419 @@
...
+    description:
+      The name of the Model. This value may be up to 60 characters, and
+      valid characters are [a-z0-9_]. The first character cannot be a number.
+    output: true
+    # url_param_only: true
+    custom_flatten: templates/terraform/custom_flatten/vertex_ai_name_from_model.go.tmpl
+  - type: String
+    name: 'description'
+    description: Description of the Model.
+    immutable: true
```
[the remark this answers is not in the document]
BBBmau: in this case for copy it would be output but since this resource is intending to support both copy and upload it's best to remove `output: true` as well as removing `immutable: true` since users can edit name and description on existing models:

image

mmv1/products/vertexai/Models.yaml
```diff
@@ -0,0 +1,419 @@
...
+  - type: Array
+    name: 'versionAliases'
+    default_from_api: true
+    description:
+      user provided version aliases so that a model version can be referenced via alias.
+    item_type:
+      type: String
+  - type: KeyValueLabels
+    name: 'labels'
+    description: The labels with user-defined metadata to organize your Models.
```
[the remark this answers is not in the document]
BBBmau: This makes since `labels` is meant to represent labels that are explicitly set in the terraform configuration.

We see the labels preset in effective labels and terraform labels since

- `effective_labels` represent all labels found on the resource
- `terraform_labels` represent labels that terraform is managing

So in the test configs we don't set labels (and aren't able to anyways). I'll remove it but keep in mind this will be in `:UPLOAD` since it's supported there.

mmv1/products/vertexai/Models.yaml
```diff
@@ -0,0 +1,419 @@
...
+  - type: String
+    name: 'sourceModel'
+    default_from_api: true
+    at_least_one_of:
+      - sourceModel
+    description: |
+      The resource name of the Model to copy. That Model must be in the same Project
+  - type: String
+    name: 'modelId'
+    default_from_api: true
```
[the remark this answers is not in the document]
BBBmau: Do we need to still include `default_from_api`? having ignore_read should be enough. From the docs:

`ignore_read`: If true, the provider sets the field’s value in the resource state based only on the user’s configuration. If false or unset, the provider sets the field’s value in the resource state based on the API response

I also included `immutable: true` since users are unable to edit the model_id

post_create has been included.

mmv1/templates/terraform/decoders/vertex_ai_models.go.tmpl
BBBmau: This is needed in order to save the `"model"` from opRes in order for the GET request to be processed properly.

```hcl
 GET /v1/projects/871647908372/locations/us-central1/models/source-model-test-post-create-2/operations/52692617292414976?alt=json HTTP/1.1

 2024/11/04 13:29:46 [DEBUG] Google API Response Details:
 HTTP/2.0 200 OK

 {
   "name": "projects/871647908372/locations/us-central1/models/source-model-test-post-create-2/operations/52692617292414976",
   "metadata": {
     "@type": "type.googleapis.com/google.cloud.aiplatform.v1.CopyModelOperationMetadata",
     "genericMetadata": {
       "createTime": "2024-11-04T21:29:36.239553Z",
       "updateTime": "2024-11-04T21:29:42.608155Z"
     }
   },
   "done": true,
   "response": {
     "@type": "type.googleapis.com/google.cloud.aiplatform.v1.CopyModelResponse",
<- this is used for the GET that follows ->
     "model": "projects/871647908372/locations/us-central1/models/source-model-test-post-create-2", \
     "modelVersionId": "1"
   }
 }

<-- {{name}} is used here. name is overwritten to be the output of model, if not we will get a 404 -->
 GET /v1/projects/871647908372/locations/us-central1/models/source-model-test-post-create-2?alt=json HTTP/1.1 
```

if we were to remove the block you highlighted it would set the name as `""` resulting in a GET being performed with no resourceReference.

Another solution would be to update the links so that it uses `model` instead of `name`. Example:

`self_link: '{{name}}'` -> `self_link: '{{model}}'

But after working on it seemed to be more complex than applying than over-writing the name value with `opRes["model"]`

mmv1/products/vertexai/Models.yaml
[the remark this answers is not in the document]
SarahFrench: I accidentally posted about labels [here](https://github.com/GoogleCloudPlatform/magic-modules/pull/12074#discussion_r1829883529), but should have posted here instead.


I realised that the labels field can be updated on a model created via copy, but they cannot be set at create time.

This means we're stuck between copy and update time:
- At create time the labels field would ideally be Computed, to pull down labels copied from the source model. The user cannot specify labels at this point.
- At update time, the user can specify labels to update the model. However they'd need to be careful to copy labels into their config that were originally stored in the API only.

I guess the solution here is the field is default_from_api:true, and we add an update after the create action (post_create) to set the labels. Or maybe we can add validation that stops users setting `labels` during create.

I think our options are to:
1) Remove labels from this PR, and let it be a FR added to the resource after it's initially released
2) Remove labels from this PR, and then add them in a subsequent PR on the feature branch. Maybe adding them after both copy and update are done?
3) Tackle the problem in this PR.

@BBBmau what's your preference here? I'm leaning towards 1 or 2

mmv1/templates/terraform/decoders/vertex_ai_models.go.tmpl
SarahFrench: As we discussed in Slack, I think the solution here would be to ensure that the `modelId` is grabbed in a post_create so that instead of :

```yaml
name: Model
base_url: projects/{{project}}/locations/{{region}}/models
create_url: 'projects/{{project}}/locations/{{region}}/modelsPRE_CREATE_REPLACE_ME'
delete_url: '{{name}}'
self_link: '{{name}}'
```


we can use


```yaml
name: Model
base_url: projects/{{project}}/locations/{{region}}/models
create_url: 'projects/{{project}}/locations/{{region}}/modelsPRE_CREATE_REPLACE_ME'
delete_url: 'projects/{{project}}/locations/{{region}}/models/{{model_id}}'
self_link: 'projects/{{project}}/locations/{{region}}/models/{{model_id}}'
```


I'll play with this and push to this PR if I get something working

SarahFrench:
ℹ️ [These tests](https://github.com/GoogleCloudPlatform/magic-modules/pull/12074#issuecomment-2459586634) only fail because the copied model is not in the CI test project

mmv1/templates/terraform/decoders/vertex_ai_models.go.tmpl
SarahFrench: Addressed in [b32aafa](https://github.com/GoogleCloudPlatform/magic-modules/pull/12074/commits/b32aafa4db1229fab6f4f5fdae66d6f910a3ce2b)

mmv1/products/vertexai/Models.yaml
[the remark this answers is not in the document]
SarahFrench: Addressed in [b32aafa](https://github.com/GoogleCloudPlatform/magic-modules/pull/12074/commits/b32aafa4db1229fab6f4f5fdae66d6f910a3ce2b)

mmv1/templates/terraform/decoders/vertex_ai_models.go.tmpl
```diff
@@ -0,0 +1,10 @@
+// Handle the name field not being set in the API response after Create (is set after Read)
+if _, ok := res["name"]; !ok {
+	if v, ok := res["model"]; ok {
+		res["name"] = v.(string)
+	}
+}
+
+res["modelName"] = res["displayName"]
```
SarahFrench, line 8: Why do we need to rename displayName to modelName? Why can't mmv1/products/vertexai/Model.yaml contain a field called modelName?

mmv1/templates/terraform/encoders/vertex_ai_models.go.tmpl
```diff
@@ -0,0 +1,6 @@
+if obj["sourceModel"] != nil {
+    delete(obj, "labels")
+    delete(obj, "metadata")
+}
```
SarahFrench, line 4: Can this be removed? Do we need a custom encoder?

It looks like this might be something created in the past to manage how some fields can only be sent on update and cannot be set on create?

mmv1/templates/terraform/examples/vertex_ai_model_source_basic_2.tf.tmpl
```diff
@@ -0,0 +1,7 @@
+resource "google_vertex_ai_model" "{{$.PrimaryResourceId}}" {
+  model_id = "{{index $.Vars "model_id"}}"
+  project = "{{index $.TestEnvVars "project_name"}}"
+  source_model = "projects/{{index $.TestEnvVars "project_name"}}/locations/us-central1/models/4469245519628878651"
```
SarahFrench, line 4:
```suggestion
resource "google_vertex_ai_model" "{{$.PrimaryResourceId}}" {
  model_id = "{{index $.Vars "model_id"}}"
  project = "{{index $.TestEnvVars "project_name"}}"
  source_model = "projects/{{index $.TestEnvVars "project_name"}}/locations/us-central1/models/1234567890"
```

mmv1/templates/terraform/examples/vertex_ai_model_source_basic_1.tf.tmpl
```diff
@@ -0,0 +1,6 @@
+resource "google_vertex_ai_model" "{{$.PrimaryResourceId}}" {
+  project = "{{index $.TestEnvVars "project_name"}}"
+  source_model = "projects/{{index $.TestEnvVars "project_name"}}/locations/us-central1/models/4469245519628878651"
```
SarahFrench, line 3:
```suggestion
  source_model = "projects/{{index $.TestEnvVars "project_name"}}/locations/us-central1/models/1234567890"
```

mmv1/templates/terraform/custom_flatten/vertex_ai_name_from_model.go.tmpl
SarahFrench, line 1: Is this custom flatten needed?

mmv1/templates/terraform/decoders/vertex_ai_models.go.tmpl
```diff
@@ -0,0 +1,10 @@
+// Handle the name field not being set in the API response after Create (is set after Read)
```
SarahFrench, line 1: Note: This is necessary in the decoder because the generated code for the Create method forces the resource's id to be set using a name field:


```go
	if err := d.Set("name", flattenVertexAIModelName(opRes["name"], d, config)); err != nil {
		return err
	}

	// This may have caused the ID to update - update it if so.
	id, err = tpgresource.ReplaceVars(d, config, "{{name}}")
	if err != nil {
		return fmt.Errorf("Error constructing id: %s", err)
	}
	d.SetId(id)
```

We could set `api_name: 'model'` in the Model.yaml file for the name field, but that would instead mean that the decoder would need to use the name field to fake the presence of the model field on Read; we're stuck between how `name` is not present after Create but is present after Read, and `model` is present after Create but not present after Read.

SarahFrench (changes requested):
Please see existing threads about:
- [Handling fields that cannot be sent on create but can be sent on update](https://github.com/GoogleCloudPlatform/magic-modules/pull/12074#discussion_r1829907452)
- [Related thread about how to handle labels](https://github.com/GoogleCloudPlatform/magic-modules/pull/12074#discussion_r1829895393), a more complex version of the problem above

---

In this review I left new comments about how I think we can:
- Rename displayName to modelName?
- Remove the custom encoder?

---

In addition to the above, we need to get some Vertex AI models present in the 3 test projects (VCR, GA nightly, Beta nightly) as static resources we can reference in this PRs acceptance tests. The references would need to be done in a way where the test can retrieve the correct model id based on the GCP project being used for the test.

@BBBmau as you have created models in our HashiCorp-owned project can you please [comment in this thread](https://chat.google.com/room/AAAA49vo02s/mB5eZMHVLQM/leeMfbFA2CQ?cls=10) to get the ball moving on this, as that's a blocker for this PR being merged.

mmv1/templates/terraform/decoders/vertex_ai_models.go.tmpl
SarahFrench: This could be addressed by setting `id_format: 'projects/{{project}}/locations/{{region}}/models/{{model_id}}'` in the YAML, but then the model_id is not available at the right time in the Create method.

[We could make the Create method entirely custom code](https://googlecloudplatform.github.io/magic-modules/develop/custom-code/#replace-entire-crud-methods) to overcome this. What do you think @BBBmau?

mmv1/templates/terraform/custom_flatten/vertex_ai_name_from_model.go.tmpl
BBBmau: initially i saw that this was necessary since even after decoding it was still not setting the name field with the model res after creation.

However I found that this was due to logic which allowed `res["name" = res["model"]` even though the `res["model"]` is nonexistent forcing name to be empty despite being already present in the res

Below is the old logic, you can see we only checked if it's not `""`. we never check if the field actually exists.

```hcl
if res["model"] != "" {
	res["name"] = res["model"]
}
```

The recent commit fixes this and removes the flattener with the new logic below:
```hcl
if modelName, ok := res["model"].(string); ok && modelName != "" {
	res["name"] = modelName
}
```

mmv1/templates/terraform/encoders/vertex_ai_models.go.tmpl
BBBmau: after testing without you're right to point out that we don't need this in this PR. I believe this makes sense to include when involving the `:UPLOAD` method. removing this file for this PR only. It can be looked at once more in the `UPLOAD` PR.

[Where it's referenced in UPLOAD PR](https://github.com/GoogleCloudPlatform/magic-modules/pull/12075/files#diff-50cd81d9be3f9dff6ce50d2aafd8bc3ef02dbca3aa9f3615d2b946a7a440a75bR2-R18)

mmv1/templates/terraform/decoders/vertex_ai_models.go.tmpl
BBBmau: the `modelName` field is non-existent in the API.

At first I was using displayName but found that it would be best to rename it to be `modelName` in order for it to be more clear for users that are viewing the state.

It also makes it easier to understand that `source_model` is for COPYING and `modelName` is for UPLOADING since `modelName` is required when uploading a model. It is set as required in the API.
`Required. The display name of the Model. The name can be up to 128 characters long and can consist of any UTF-8 characters.`

https://cloud.google.com/vertex-ai/docs/reference/rest/v1/projects.locations.models

In the case of COPYING, we grab the existing modelName from the model that's being copied and set the `modelId` if applied in tfconfig. If not applied it will be autogenerated.

mmv1/templates/terraform/decoders/vertex_ai_models.go.tmpl
BBBmau: I don't think it's worth committing to writing the custom_create just to resolve this only because of the overhead that it comes from it especially when dealing with what is to come when needing to support UPLOAD and all it's other fields. 

I think it's a fine trade-off of needing to address the missing name field on Create, a well-written comment should be more than enough so that any future reviewer can quickly understand why this is implemented the way it is.

mmv1/products/vertexai/Models.yaml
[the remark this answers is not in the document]
BBBmau: Just tried this out and it works great! Just a heads-up that I need need to add back in the encoder since the `modelName` field was being used in the patch instead of `displayName`. The code would check whether to switch modelName for displayName when going into the encoder.

```hcl
	if modelName, ok := obj["modelName"].(string); ok && modelName != "" {
		obj["displayName"] = modelName
		delete(obj, "modelName")
	}
	return obj, nil
```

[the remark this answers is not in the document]
BBBmau: I had to do quite a bit to resolve the labels diff and even then im not very satisfied with it.

Although the customize diff worked for descriptions and labels. We run into an issue where terraform applies it's own label despite the user never setting it. This causes us to see the following on apply. (This is after adding in the labels field once more since we are addressing the issue in this request thread)

```hcl
  + create

  # google_vertex_ai_model.name will be created
  + resource "google_vertex_ai_model" "name" {
      + artifact_uri             = (known after apply)
      + create_time              = (known after apply)
      + effective_labels         = {
          + "goog-terraform-provisioned" = "true"
        }
      + etag                     = (known after apply)
      + id                       = (known after apply)
      + model_id                 = (known after apply)
      + model_name               = (known after apply)
      + name                     = (known after apply)
      + project                  = "hc-terraform-testing"
      + region                   = (known after apply)
      + source_model             = "projects/871647908372/locations/us-central1/models/6033738282699849728"
      + supported_export_formats = (known after apply)
      + terraform_labels         = {
          + "goog-terraform-provisioned" = "true"
        }
      + trainer_pipeline         = (known after apply)
      + update_time              = (known after apply)
      + version_aliases          = (known after apply)
      + version_create_time      = (known after apply)
      + version_description      = (known after apply)
      + version_id               = (known after apply)
      + version_update_time      = (known after apply)
    }

2024-11-13T17:24:26.439-0800 [DEBUG] command: asking for input: "\nDo you want to perform these actions?"

  Enter a value: yes
```

You'll notice that we have both effective_labels and terraform_labels. We don't apply any labels in config but these are applied automatically. Because of this we get the following error.

```hcl
│ Error: Error creating Model: googleapi: Error 400: Invalid JSON payload received. Unknown name "labels": Cannot find field.
│ Details:
│ [
│   {
│     "@type": "type.googleapis.com/google.rpc.BadRequest",
│     "fieldViolations": [
│       {
│         "description": "Invalid JSON payload received. Unknown name \"labels\": Cannot find field."
│       }
│     ]
│   }
│ ]
│ 
│   with google_vertex_ai_model.name,
│   on main.tf line 7, in resource "google_vertex_ai_model" "name":
│    7: resource "google_vertex_ai_model" "name" {
│ 
```

This is due to           `"goog-terraform-provisioned" = "true"` being include in `labels`.

I was able to resolve this by deleting `labels` from the `obj` request in `pre_create`. however we still get on plan both `effective_labels` and `terraform_labels`

This seems fine since we are expecting these fields to be filled out by api but it also seems incorrect the implementation of it all. Let me know your thoughts! @SarahFrench

BBBmau:
Commenting here since it's very easy to miss old threads @SarahFrench https://github.com/GoogleCloudPlatform/magic-modules/pull/12074#discussion_r1841401745

mmv1/products/vertexai/Models.yaml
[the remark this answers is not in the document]
SarahFrench: I think we should just omit labels from this PR for simplicity.

mmv1/templates/terraform/decoders/vertex_ai_models.go.tmpl
SarahFrench: @BBBmau sorry, I mistyped my first message and I've updated it. My question was why we would need to rename a field from the API.

Could we avoid renaming the field in this PR, as this PR is only concerned with create-via-copy? If we rename the field in this PR we give the false impression that the rename is required for a technical reason/limitation. If it's change chosen for just a user experience decision I think that should be discussed on the PR for create via upload

SarahFrench (commented):
Replied in threads

mmv1/products/vertexai/Models.yaml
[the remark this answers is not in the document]
BBBmau: Would lean towards 1, thanks for bringing this up!

mmv1/templates/terraform/decoders/vertex_ai_models.go.tmpl
BBBmau: ah in this case I'll revert it to use displayName since you're correct that it's not a requirement for this PR to rename the field from the API.

But it is worth knowing noting that the use of modelName is meant for :UPLOAD and will be renamed once :UPLOAD PR is merged into here. I can add this in the description of this PR.

mmv1/products/vertexai/Models.yaml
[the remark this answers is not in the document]
BBBmau: sounds good, I'll remove the support for labels then.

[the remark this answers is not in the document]
BBBmau: When it came to testing I found that when wanting to copy a model with an already set description led to diffs occurring on the second apply. Terraform would attempt to update the description to what is set in config (which would be null):

```hcl
2024/11/21 23:57:08 [DEBUG] [transport] [server-transport 0xc000173040] loopyWriter exiting with error: transport closed by client 
    resource_vertex_ai_model_test.go:59: Step 1/1 error: After applying this test step, the plan was not empty.
        stdout:
        
        
          ~ update in-place
        
          # google_vertex_ai_model.model will be updated in-place
          ~ resource "google_vertex_ai_model" "model" {
              - description              = "Manually-uploaded model for testing" -> null
                id                       = "projects/1067888929963/locations/us-central1/models/2861444926426578944"
                name                     = "projects/1067888929963/locations/us-central1/models/2861444926426578944"
                # (17 unchanged attributes hidden)
        
                # (1 unchanged block hidden)
            }
        
```

I went ahead and marked it `default_from_api: true` to resolve this. It led to all tests passing after providing it with a provisioned model_id used just for testing here. We'll be providing a model via upload once we start merging the two PRs together.

[Related Magician Output](https://github.com/GoogleCloudPlatform/magic-modules/pull/12074#issuecomment-2492591913)

BBBmau:
Just so it doesn't get lost: https://github.com/GoogleCloudPlatform/magic-modules/pull/12074#discussion_r1853271879

once this is approved I can merge it into UPLOAD and include UPLOAD as part of the copy tests for initial setup prior to actually testing the copy method by itself.

mmv1/templates/terraform/pre_create/vertex_ai_models.go.tmpl
```diff
@@ -0,0 +1,9 @@
+var postRequestType string
+
+if _, ok := d.GetOk("source_model"); ok {
+	postRequestType = ":copy"
+
+	// These are not valid fields for the copy request on create
+	delete(obj, "labels")
```
SarahFrench, line 7: If labels isn't in the schema (at least for this PR) let's remove any reference to it in the code

```suggestion
```

mmv1/templates/terraform/decoders/vertex_ai_models.go.tmpl
```diff
@@ -0,0 +1,7 @@
+if modelName, ok := res["model"].(string); ok && modelName != "" {
+	res["name"] = modelName
+}
+
+res["modelName"] = res["displayName"]
```
SarahFrench, line 5: As the schema on this PR contain's the API's displayName field (https://github.com/GoogleCloudPlatform/magic-modules/pull/12074/commits/f1124737c6c9bba113958949f79f51ee608a90d0) can we remove this decoder that translates the API's displayName field into the made up modelName field?

```suggestion
```

mmv1/templates/terraform/encoders/vertex_ai_models.go.tmpl
```diff
@@ -0,0 +1,6 @@
+if modelName, ok := obj["modelName"].(string); ok && modelName != "" {
+	obj["displayName"] = modelName
+	delete(obj, "modelName")
+}
```
SarahFrench, line 4: Is this required given that this PR has displayName in the schema (https://github.com/GoogleCloudPlatform/magic-modules/commit/f1124737c6c9bba113958949f79f51ee608a90d0)?

SarahFrench (commented):
Some last comments to get this PR 100% about create via copy and to allow all code changes necessary for upload to be added in the separate upload PR

mmv1/templates/terraform/decoders/vertex_ai_models.go.tmpl
BBBmau: yes this should've been removed. This is meant for the upload method PR

mmv1/templates/terraform/encoders/vertex_ai_models.go.tmpl
BBBmau: no this would be implemented for when we include `:UPLOAD`. removing right now.

BBBmau:
addressed the last couple changes, should be good for another review and then I could look into upload support. @SarahFrench

SarahFrench (approved):
Approving, but just a heads up that this PR is currently due to go into main branch. If you merge it as-is then it'll be in the December 16th release, despite a lack of test coverage etc.

PR #12078, opened 2024-10-22
fix an argument in data.google_artifact_registry_docker_image example

dullest:
```release-note:bug
artifactregistry: fixed an argument in the `google_artifact_registry_docker_image` data source example
```


SirGitsalot (approved):
Thanks for the contribution!

PR #12084, opened 2024-10-22
container: simplify terraform code for `google_container_node_pool` tests

wyardley:
Let me know if this is too much of a nuisance, but I think this should help with readability. I didn't format everything, but updated most of the stuff with hard tabs or bad spacing problems.

- Use direct references where possible for resource manager tags
- Remove some other unnecessary quote interpolation in values
- Apply terraform formatting to some test cases with big formatting issues

```release-note:none

```

### DIFF
--- a/mmv1/third_party/terraform/services/container/resource_container_node_pool_test.go.tmpl
+++ b/mmv1/third_party/terraform/services/container/resource_container_node_pool_test.go.tmpl
@@ -3674,7 +3674,7 @@ resource "google_container_cluster" "cluster" {
 resource "google_container_node_pool" "with_upgrade_settings" {
   name = "%s"
   location = "us-central1"
-  cluster = "${google_container_cluster.cluster.name}"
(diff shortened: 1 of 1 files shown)


melinath:
thanks for keeping this in a separate PR! Seems reasonable assuming the tests still pass.

wyardley:
> thanks for keeping this in a separate PR!

Ahah, yes, this won't directly tie into the cases for #12014 since that's all in `container_cluster`, but tangentially related (trying to keep those resource manager tag changes separate).

I made some much smaller updates in there, but only within the resources I was directly working with. Could potentially do a similar PR for `container_cluster` tests separate from #12014 (including the same resource manager tag changes) if it's helpful and not annoying.

wyardley:
I am getting local failures on `TestAccContainerCluster_resourceManagerTags`, but I'm curious if it's just availability / quota related.

melinath (approved):

PR #12093, opened 2024-10-22
Remove stale GKE 1.27 version references

rileykarson:
Addresses test failures that include `Error: googleapi: Error 400: No valid versions with the prefix "1.27" found.`

I've assumed that all of these can just have their versions stripped. If they fail for new reasons I am not going to fix them here- I'd prefer we fail for the real reason in the nightly.

Also rearrange some fields I didn't like, where `network`/`subnetwork` were after `deletion_protection`.

```release-note:none
```


SarahFrench (approved):
Looks good to me given the tests running and passing on this PR

PR #12095, opened 2024-10-22
pubsub: use boostrapped IAM for `TestAccPubsubSubscription` tests

wyardley:
Fixes hashicorp/terraform-provider-google#19683
Fixes hashicorp/terraform-provider-google#19085

Note: same test and same issue (I think), so presumably 19085 should be fixed by this as well

```release-note:none

```

### DIFF
--- a/mmv1/products/pubsub/Subscription.yaml
+++ b/mmv1/products/pubsub/Subscription.yaml
@@ -72,6 +72,8 @@ examples:
       subscription_name: 'example-subscription'
       dataset_id: 'example_dataset'
       table_id: 'example_table'
+    test_vars_overrides:
+      policy_changed: 'acctest.BootstrapPSARoles(t, "service-", "gcp-sa-pubsub", []string{"roles/bigquery.dataEditor", "roles/bigquery.metadataViewer"})'
   - name: 'pubsub_subscription_push_bq_table_schema'
     primary_resource_id: 'example'
     vars:
@@ -79,6 +81,8 @@ examples:
       subscription_name: 'example-subscription'
       dataset_id: 'example_dataset'
       table_id: 'example_table'
+    test_vars_overrides:
+      policy_changed: 'acctest.BootstrapPSARoles(t, "service-", "gcp-sa-pubsub", []string{"roles/bigquery.dataEditor", "roles/bigquery.metadataViewer"})'
   - name: 'pubsub_subscription_push_bq_service_account'
     primary_resource_id: 'example'
     vars:
--- a/mmv1/third_party/terraform/services/pubsub/resource_pubsub_subscription_test.go
+++ b/mmv1/third_party/terraform/services/pubsub/resource_pubsub_subscription_test.go
@@ -246,6 +246,9 @@ func TestAccPubsubSubscriptionBigQuery_serviceAccount(t *testing.T) {
 		PreCheck:                 func() { acctest.AccTestPreCheck(t) },
 		ProtoV5ProviderFactories: acctest.ProtoV5ProviderFactories(t),
 		CheckDestroy:             testAccCheckPubsubSubscriptionDestroyProducer(t),
+		ExternalProviders: map[string]resource.ExternalProvider{
+			"time": {},
+		},
 		Steps: []resource.TestStep{
 			{
 				Config: testAccPubsubSubscriptionBigQuery_basic(dataset, table, topic, subscriptionShort, false, "bq-test-sa"),
@@ -715,6 +718,11 @@ resource "google_project_iam_member" "editor" {
 	return fmt.Sprintf(`
 data "google_project" "project" {}
 
+resource "time_sleep" "wait_30_seconds" {
+  create_duration = "30s"
+}
+
+
 %s
 
 resource "google_bigquery_dataset" "test" {
@@ -754,7 +762,8 @@ resource "google_pubsub_subscription" "foo" {
 
   depends_on = [
     google_project_iam_member.viewer,
-    google_project_iam_member.editor
+    google_project_iam_member.editor,
+    time_sleep.wait_30_seconds,
   ]
 }
 	`, serviceAccountResource, dataset, table, topic, subscription, useTableSchema, serviceAccountEmailField)
--- a/mmv1/third_party/terraform/services/pubsub/resource_pubsub_subscription_test.go
+++ b/mmv1/third_party/terraform/services/pubsub/resource_pubsub_subscription_test.go
@@ -207,6 +207,9 @@ func TestAccPubsubSubscriptionBigQuery_update(t *testing.T) {
 		PreCheck:                 func() { acctest.AccTestPreCheck(t) },
 		ProtoV5ProviderFactories: acctest.ProtoV5ProviderFactories(t),
 		CheckDestroy:             testAccCheckPubsubSubscriptionDestroyProducer(t),
+		ExternalProviders: map[string]resource.ExternalProvider{
+			"time": {},
+		},
 		Steps: []resource.TestStep{
 			{
 				Config: testAccPubsubSubscriptionBigQuery_basic(dataset, table, topic, subscriptionShort, false, ""),
--- a/mmv1/third_party/terraform/services/pubsub/resource_pubsub_subscription_test.go
+++ b/mmv1/third_party/terraform/services/pubsub/resource_pubsub_subscription_test.go
@@ -722,7 +722,6 @@ resource "time_sleep" "wait_30_seconds" {
   create_duration = "30s"
 }
 
-
 %s
 
 resource "google_bigquery_dataset" "test" {
--- a/mmv1/third_party/terraform/services/pubsub/resource_pubsub_subscription_test.go
+++ b/mmv1/third_party/terraform/services/pubsub/resource_pubsub_subscription_test.go
@@ -238,6 +238,10 @@ func TestAccPubsubSubscriptionBigQuery_serviceAccount(t *testing.T) {
 	topic := fmt.Sprintf("tf-test-topic-%s", acctest.RandString(t, 10))
 	subscriptionShort := fmt.Sprintf("tf-test-sub-%s", acctest.RandString(t, 10))
 
+	if acctest.BootstrapPSARoles(t, "service-", "gcp-sa-pubsub", []string{"roles/bigquery.dataEditor", "roles/bigquery.metadataViewer"}) {
+		t.Fatal("Stopping the test because roles were added to IAM policy.")
+	}
+
 	acctest.VcrTest(t, resource.TestCase{
 		PreCheck:                 func() { acctest.AccTestPreCheck(t) },
 		ProtoV5ProviderFactories: acctest.ProtoV5ProviderFactories(t),
--- a/mmv1/third_party/terraform/services/pubsub/resource_pubsub_subscription_test.go
+++ b/mmv1/third_party/terraform/services/pubsub/resource_pubsub_subscription_test.go
@@ -685,40 +685,40 @@ resource "google_service_account" "bq_write_service_account" {
 }
 
 resource "google_project_iam_member" "viewer" {
-	project = data.google_project.project.project_id
-	role   = "roles/bigquery.metadataViewer"
-	member = "serviceAccount:${google_service_account.bq_write_service_account.email}"
+  project = data.google_project.project.project_id
+  role    = "roles/bigquery.metadataViewer"
+  member  = "serviceAccount:${google_service_account.bq_write_service_account.email}"
 }
 
 resource "google_project_iam_member" "editor" {
-	project = data.google_project.project.project_id
-	role   = "roles/bigquery.dataEditor"
-	member = "serviceAccount:${google_service_account.bq_write_service_account.email}"
+  project = data.google_project.project.project_id
+  role    = "roles/bigquery.dataEditor"
+  member  = "serviceAccount:${google_service_account.bq_write_service_account.email}"
 }`, serviceAccountId)
 		serviceAccountEmailField = "service_account_email = google_service_account.bq_write_service_account.email"
 	} else {
 		serviceAccountResource = fmt.Sprintf(`
 resource "google_project_iam_member" "viewer" {
-	project = data.google_project.project.project_id
-	role   = "roles/bigquery.metadataViewer"
-	member = "serviceAccount:service-${data.google_project.project.number}@gcp-sa-pubsub.iam.gserviceaccount.com"
+  project = data.google_project.project.project_id
+  role    = "roles/bigquery.metadataViewer"
+  member  = "serviceAccount:service-${data.google_project.project.number}@gcp-sa-pubsub.iam.gserviceaccount.com"
 }
 
 resource "google_project_iam_member" "editor" {
-	project = data.google_project.project.project_id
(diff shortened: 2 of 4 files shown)


wyardley:
Is this what @melinath had in mind here?

Locally, running just the individual tests, using the same project, I'm sometimes seeing it trying to add the roles (and test failing) even after I can see that the roles already exist. But feel free to see if this passes consistently in CI.

mmv1/third_party/terraform/services/pubsub/resource_pubsub_subscription_test.go
```diff
@@ -238,6 +238,10 @@ func TestAccPubsubSubscriptionBigQuery_serviceAccount(t *testing.T) {
 	topic := fmt.Sprintf("tf-test-topic-%s", acctest.RandString(t, 10))
 	subscriptionShort := fmt.Sprintf("tf-test-sub-%s", acctest.RandString(t, 10))
 
+	if acctest.BootstrapPSARoles(t, "service-", "gcp-sa-pubsub", []string{"roles/editor", "roles/viewer"}) {
```
wyardley, line 241: looks like these two are project editor / viewer vs. the ones needed by the two generated ones. Not sure what's needed to get the tests to show up as affected tests?

There is still a service account that's created and has the same roles assigned in the test in 2 of the steps, so this step won't solve the issue for that case.... maybe a sleep still needs to be added for the case where the service account is created and has roles assigned in the test?

melinath:
> Is this what @melinath had in mind here?

Drive-by comment: I don't remember the specific conversation but in general we do currently recommend bootstrapping IAM permissions.

wyardley:
> Drive-by comment: I don't remember the specific conversation but in general we do currently recommend bootstrapping IAM permissions.

@melinath sorry, was referring to your description in the issue itself: https://github.com/hashicorp/terraform-provider-google/issues/19683

wyardley:
I think with the hand-written test especially, there might still need to be a sleep, because there are 2 iterations of running that test function where the permissions are added to a different service account that's created _in_ the test step.

I have not had a ton of luck running the tests individually locally, because the bootstrapped permissions don't "stick" between runs, but I think when running the whole suite, it may work better?

zli82016:
@trodge , are you OK to take on this PR? Thanks.

mmv1/products/pubsub/Subscription.yaml
```diff
@@ -72,13 +72,17 @@ examples:
       subscription_name: 'example-subscription'
       dataset_id: 'example_dataset'
       table_id: 'example_table'
+    test_vars_overrides:
+      policy_changed: 'acctest.BootstrapPSARoles(t, "service-", "gcp-sa-pubsub", []string{"roles/bigquery.dataEditor", "roles/bigquery.metadataViewer"})'
```
zli82016, line 76: `resource "google_project_iam_member" "viewer"` and `resource "google_project_iam_member" "editor" `should be removed from the test file with the bootstrapped iam https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/templates/terraform/examples/pubsub_subscription_push_bq.tf.tmpl#L19

mmv1/products/pubsub/Subscription.yaml
```diff
@@ -72,13 +72,17 @@ examples:
...
   - name: 'pubsub_subscription_push_bq_table_schema'
     primary_resource_id: 'example'
     vars:
       topic_name: 'example-topic'
       subscription_name: 'example-subscription'
       dataset_id: 'example_dataset'
       table_id: 'example_table'
+    test_vars_overrides:
+      policy_changed: 'acctest.BootstrapPSARoles(t, "service-", "gcp-sa-pubsub", []string{"roles/bigquery.dataEditor", "roles/bigquery.metadataViewer"})'
```
zli82016, line 85: The same with above

mmv1/products/pubsub/Subscription.yaml
wyardley: Thanks. Updated.

mmv1/products/pubsub/Subscription.yaml
wyardley: Updated

mmv1/third_party/terraform/services/pubsub/resource_pubsub_subscription_test.go
```diff
@@ -238,6 +238,10 @@ func TestAccPubsubSubscriptionBigQuery_serviceAccount(t *testing.T) {
 	topic := fmt.Sprintf("tf-test-topic-%s", acctest.RandString(t, 10))
 	subscriptionShort := fmt.Sprintf("tf-test-sub-%s", acctest.RandString(t, 10))
 
+	if acctest.BootstrapPSARoles(t, "service-", "gcp-sa-pubsub", []string{"roles/bigquery.dataEditor", "roles/bigquery.metadataViewer"}) {
```
zli82016, line 241: @melinath or @NickElliot , do you know the reason to stop the test because roles were added to IAM policy in other examples?

mmv1/third_party/terraform/services/pubsub/resource_pubsub_subscription_test.go
```diff
@@ -681,40 +685,40 @@ resource "google_service_account" "bq_write_service_account" {
...
-	role   = "roles/bigquery.dataEditor"
-	member = "serviceAccount:${google_service_account.bq_write_service_account.email}"
+  project = data.google_project.project.project_id
+  role    = "roles/bigquery.dataEditor"
+  member  = "serviceAccount:${google_service_account.bq_write_service_account.email}"
 }`, serviceAccountId)
 		serviceAccountEmailField = "service_account_email = google_service_account.bq_write_service_account.email"
 	} else {
 		serviceAccountResource = fmt.Sprintf(`
 resource "google_project_iam_member" "viewer" {
```
zli82016, line 701: @melinath or @NickElliot , do you know if this resource should be removed with the bootstrapped IAM?

mmv1/third_party/terraform/services/pubsub/resource_pubsub_subscription_test.go
wyardley: I've been wondering about this a little too, because in my limited testing, the bindings seem to get removed again when the test is over? But people generally talk about this only needing to fail on the first run, so maybe it is "sticky" (or supposed to be)?

From the comments, it seems to be because the binding may not show up quickly enough and the test might fail downstream? I couldn't really find any examples where it didn't do this.

mmv1/third_party/terraform/services/pubsub/resource_pubsub_subscription_test.go
wyardley: I think in this case, it can't be, because there are a couple test cases where the service account is created _within_ the test case, so in those cases, the permission still needs to be added. 

For example
https://github.com/GoogleCloudPlatform/magic-modules/blob/01a6cac159ae042b73ec49c4ef80eb29ffd5098f/mmv1/third_party/terraform/services/pubsub/resource_pubsub_subscription_test.go#L251
where it creates `bq-test-sa` (which will later be destroyed), and then `bq-test-sa2` on the 3rd test case.

Since the account will be created and destroyed within the test, it wouldn't be possible to boostrap its permissions.

However, if the test is flaky (or still flaky), it would be possible to add a sleep in there.

From what I can see, in the case where the permissions already exist from bootstrapping, Terraform doesn't seem to complain (maybe the test framework imports them when they exist?)

wyardley:
Side note: I didn't change it here, but `dataset_id = google_bigquery_dataset.test.dataset_id` I think should work from a quick test, and would be simpler and more readable than the way it is in the two examples.

zli82016:
> Side note: I didn't change it here, but `dataset_id = google_bigquery_dataset.test.dataset_id` I think should work from a quick test, and would be simpler and more readable than the way it is in the two examples.

`dataset_id = google_bigquery_dataset.test.dataset_id`  is the current way to be used in the two examples, for example https://github.com/GoogleCloudPlatform/magic-modules/pull/12095/files#diff-91baab7a5938e551176155af3786e17d85205fda7d62ced4fd37f6dd6da66316R22

Or I miss something.

wyardley:
> Or I miss something.

Ah, sorry for lack of context. I meant in `bigquery_config {}` in `google_pubsub_subscription`, e.g.:
https://github.com/GoogleCloudPlatform/magic-modules/blob/7e0082000730f2a8fc972383a4cbb8b09286d5ba/mmv1/templates/terraform/examples/pubsub_subscription_push_bq.tf.tmpl#L9-L10

mmv1/third_party/terraform/services/pubsub/resource_pubsub_subscription_test.go
zli82016: I see.  `testAccPubsubSubscriptionBigQuery_basic` handles two cases, no-empty service account and empty service account.

I think it is OK to not add a sleep right now. If it is still flaky, we can add the sleep later.

mmv1/third_party/terraform/services/pubsub/resource_pubsub_subscription_test.go
wyardley: Also, we could easily enough not define the resources when the service account isn't created... but there's an explicit `depends` later that I think would be harder to get rid of....

mmv1/third_party/terraform/services/pubsub/resource_pubsub_subscription_test.go
wyardley: Right. We could easily enough not define the resources when the service account isn't created... but there's an explicit `depends` later that I think would be harder to get rid of....

I could reproduce flakiness / permissions issue with some of the test cases locally. This is what the sleep would look like FWIW.

```diff
diff --git a/mmv1/third_party/terraform/services/pubsub/resource_pubsub_subscription_test.go b/mmv1/third_party/terraform/services/pubsub/resource_pubsub_subscription_test.go
index 6e2447ff4..5f2fe23db 100644
--- a/mmv1/third_party/terraform/services/pubsub/resource_pubsub_subscription_test.go
+++ b/mmv1/third_party/terraform/services/pubsub/resource_pubsub_subscription_test.go
@@ -246,6 +246,9 @@ func TestAccPubsubSubscriptionBigQuery_serviceAccount(t *testing.T) {
 		PreCheck:                 func() { acctest.AccTestPreCheck(t) },
 		ProtoV5ProviderFactories: acctest.ProtoV5ProviderFactories(t),
 		CheckDestroy:             testAccCheckPubsubSubscriptionDestroyProducer(t),
+		ExternalProviders: map[string]resource.ExternalProvider{
+			"time": {},
+		},
 		Steps: []resource.TestStep{
 			{
 				Config: testAccPubsubSubscriptionBigQuery_basic(dataset, table, topic, subscriptionShort, false, "bq-test-sa"),
@@ -715,6 +718,11 @@ resource "google_project_iam_member" "editor" {
 	return fmt.Sprintf(`
 data "google_project" "project" {}
 
+resource "time_sleep" "wait_30_seconds" {
+  create_duration = "30s"
+}
+
+
 %s
 
 resource "google_bigquery_dataset" "test" {
@@ -754,7 +762,8 @@ resource "google_pubsub_subscription" "foo" {
 
   depends_on = [
     google_project_iam_member.viewer,
-    google_project_iam_member.editor
+    google_project_iam_member.editor,
+    time_sleep.wait_30_seconds,
   ]
 }
 	`, serviceAccountResource, dataset, table, topic, subscription, useTableSchema, serviceAccountEmailField)
```

wyardley: But I don't see any open tickets / issues for this test failing / being flaky in GitHub.

zli82016: I checked the test history. The test failed twice recently, Aug 20 and Aug 03. 
Do you mind adding the sleep? Thanks for working on it.

zli82016: > Right. We could easily enough not define the resources when the service account isn't created... but there's an explicit `depends` later that I think would be harder to get rid of....
> 

I think we can split the two cases (empty service account and non-empty service account) into its own functions . It is also OK to keep it like this now.

wyardley: pushed that up.

wyardley: > I think we can split the two cases (empty service account and non-empty service account) into its own functions . It is also OK to keep it like this now.

Yeah, I'm going to leave it for now, but looks like it should be pretty easy to adjust the conditional (just might be hard to read), or, as you say, split it into two separate functions.

I'm guessing it won't affect execution time or success rate much, though?

mmv1/third_party/terraform/services/pubsub/resource_pubsub_subscription_test.go
wyardley: Also, does it need to add the role when `VCR_MODE=REPLAYING` is set also? I got an exit locally running this with `VCR_MODE=REPLAYING`

mmv1/third_party/terraform/services/pubsub/resource_pubsub_subscription_test.go
zli82016: > I'm guessing it won't affect execution time or success rate much, though?

I don't think it will affect those.

mmv1/third_party/terraform/services/pubsub/resource_pubsub_subscription_test.go
zli82016: I guess the second run will success, as the role has been added in the first time.

mmv1/third_party/terraform/services/pubsub/resource_pubsub_subscription_test.go
```diff
@@ -238,10 +238,17 @@ func TestAccPubsubSubscriptionBigQuery_serviceAccount(t *testing.T) {
...
 
+	if acctest.BootstrapPSARoles(t, "service-", "gcp-sa-pubsub", []string{"roles/bigquery.dataEditor", "roles/bigquery.metadataViewer"}) {
+		t.Fatal("Stopping the test because roles were added to IAM policy.")
+	}
+
 	acctest.VcrTest(t, resource.TestCase{
 		PreCheck:                 func() { acctest.AccTestPreCheck(t) },
 		ProtoV5ProviderFactories: acctest.ProtoV5ProviderFactories(t),
 		CheckDestroy:             testAccCheckPubsubSubscriptionDestroyProducer(t),
+		ExternalProviders: map[string]resource.ExternalProvider{
```
zli82016, line 249: Can you please also add `time` provider to the test [TestAccPubsubSubscriptionBigQuery_update](https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/third_party/terraform/services/pubsub/resource_pubsub_subscription_test.go#L198), where `testAccPubsubSubscriptionBigQuery_basic` is also called? Thanks.

wyardley: Oh thanks, good call. updated.

zli82016 (approved):
LGTM. Thanks for the fix.

PR #2755, opened 2019-11-27
try deleting storage buckets before clearing them

danawillow:
If we create a bucket IAM policy and then delete it (as we do in tests), we end up removing the ability to list items in the bucket. Because being able to delete buckets is a project-wide permission, we can still delete the bucket, as long as it is empty. This change fixes `TestAccStorageBucketIamPolicy`, which has been getting skipped in CI due to a missing environment variable, but consistently fails locally.

**Release Note Template for Downstream PRs (will be copied)**

```release-note:bug
`storage`: fixed bug where users without storage.objects.list permissions couldn't delete empty buckets
```

PR #12101, opened 2024-10-23
feat(HNS folders): add new resources to support HNS folders

gurusai-voleti:

```release-note:new-resource
`google_storage_folder`
```


kautikdk:
Please add tests for nested folders and objects inside folder.

mmv1/products/storage/Folder.yaml
```diff
@@ -0,0 +1,82 @@
...
+      - 'recursive'
+description: |
+  A Google Cloud Storage Folder.
+
+  The Folder resource represents a folder in a Cloud Storage bucket with hierarchical namespace enabled
+references:
+  guides:
+    'Official Documentation': 'https://cloud.google.com/storage/docs/folders-overview'
+  api: 'https://cloud.google.com/storage/docs/json_api/v1/folders'
+# iam_policy: handwritten in mmv1/third_party/terraform/services/storage/iam_storage_managed_folder.go
```
kautikdk, line 45: This can be removed.

mmv1/products/storage/Folder.yaml
```diff
@@ -0,0 +1,82 @@
...
+    description: |
+      The name of the folder expressed as a path. Must include
+      trailing '/'. For example, `example_dir/example_dir2/`.
+    required: true
+    # The API returns values with trailing slashes, even if not
+    # provided. Enforcing trailing slashes prevents diffs and ensures
+    # consistent output.
+    validation:
+      regex: '/$'
+  - name: 'recursive'
```
kautikdk, line 64: We are not planning to add this field, Right?

mmv1/templates/terraform/examples/storage_folder_basic.tf.tmpl
```diff
@@ -0,0 +1,16 @@
...
+    enabled = true
+  }
+  force_destroy = true
+}
+
+resource "google_storage_folder" "{{$.PrimaryResourceId}}" {
+  bucket        = google_storage_bucket.bucket.name
+  name          = "folder/name/"
+  force_destroy = true
+  recursive     = true
```
kautikdk, line 15: this can be removed as of now!

mmv1/third_party/terraform/services/storage/resource_storage_folder.go.tmpl
```diff
@@ -0,0 +1,358 @@
...
+		return err
+	}
+
+	bucket := d.Get("bucket").(string)
+	name := d.Get("name").(string)
+
+	folderName := strings.Split(name, "/")
+	var listError, deleteObjectError error
+	for deleteObjectError == nil {
+		res, err := config.NewStorageClient(userAgent).Objects.List(bucket).Prefix(folderName[0] + "/").Do()
```
kautikdk, line 221: Why to list objects of the top level folder when we are trying to delete nested folder?

mmv1/third_party/terraform/services/storage/resource_storage_folder.go.tmpl
```diff
@@ -0,0 +1,358 @@
...
+		return fmt.Errorf("could not delete non-empty folder due to error when listing contents: %v", listError)
+	}
+
+	if !d.Get("force_destroy").(bool) {
+		deleteErr := fmt.Errorf("Error trying to delete folder without force_destroy set to true")
+		log.Printf("Error! %s : %s\n\n", name, deleteErr)
+		return deleteErr
+	} else {
+		log.Printf("[DEBUG] folder names to delete: %#v", folderName)
+		foldersList, err := config.NewStorageClient(userAgent).Folders.List(bucket).Prefix(folderName[0] + "/").Do()
```
kautikdk, line 283: same as above comment!

mmv1/third_party/terraform/services/storage/resource_storage_folder.go.tmpl
```diff
@@ -0,0 +1,358 @@
...
+		if err != nil {
+			return err
+		}
+		items := foldersList.Items
+		for i := len(items) - 1; i >= 0; i-- {
+			err = transport_tpg.Retry(transport_tpg.RetryOptions{
+				RetryFunc: func() error {
+					err = config.NewStorageClient(userAgent).Folders.Delete(bucket, items[i].Name).Do()
```
kautikdk, line 291: What if there are objects inside nested folder?

mmv1/third_party/terraform/services/storage/resource_storage_folder.go.tmpl
gurusai-voleti: to delete folders first we need to delete all objects as non empty folders cannot be deleted, getting all the objects with top level folder prefix and delete all objects and in next step delete the sub folders and finally delete the main folder, this will clear up all the objects, subfolders and folders

mmv1/third_party/terraform/services/storage/resource_storage_folder.go.tmpl
gurusai-voleti: we are getting all objects with a top level folder prefix this will list all objects, here in this step only folders will remain so deleting only subfolders and folder here

mmv1/third_party/terraform/services/storage/resource_storage_folder.go.tmpl
kautikdk: But the scenario that I am referring is, If the folder resource is `a/b/c/d` then we try to delete it with `force_destroy` = true then as per this logic, it will try to delete everything inside folder `a/` which is not expected. we should only delete everything that is nested in `a/b/c/d` path not anything that is in upper level of path or in different path tree like `a/b/c` or `a/b/c/e/f`.

mmv1/third_party/terraform/services/storage/resource_storage_folder.go.tmpl
kautikdk: Ohh, I see!
I thought it won't list object that are deeply nested within sub-folders(more than 1 level). 
Got it!

mmv1/products/storage/Folder.yaml
gurusai-voleti: removed

mmv1/products/storage/Folder.yaml
gurusai-voleti: removed

mmv1/templates/terraform/examples/storage_folder_basic.tf.tmpl
gurusai-voleti: removed

mmv1/third_party/terraform/services/storage/resource_storage_folder.go.tmpl
gurusai-voleti: updated the logic

mmv1/third_party/terraform/services/storage/resource_storage_folder.go.tmpl
```diff
@@ -0,0 +1,343 @@
...
+	if err := d.Set("self_link", tpgresource.ConvertSelfLinkToV1(res.SelfLink)); err != nil {
+		return fmt.Errorf("Error reading Folder: %s", err)
+	}
+
+	return nil
+}
+
+func resourceStorageFolderUpdate(d *schema.ResourceData, meta interface{}) error {
+	config := meta.(*transport_tpg.Config)
+	_ = config
```
kautikdk, line 189: What is the need of this?

mmv1/third_party/terraform/services/storage/resource_storage_folder.go.tmpl
```diff
@@ -0,0 +1,343 @@
...
+
+	// we can only get here if force_destroy was updated
+	if d.Get("force_destroy") != nil {
+		if err := d.Set("force_destroy", d.Get("force_destroy")); err != nil {
+			return fmt.Errorf("Error updating force_destroy: %s", err)
+		}
```
kautikdk, line 195: Not sure but if the field is not being pass down to the API request then do we need to explicitly set in updater?

mmv1/third_party/terraform/services/storage/resource_storage_folder.go.tmpl
```diff
@@ -0,0 +1,343 @@
...
+
+		// Wait for everything to finish.
+		wp.StopWait()
+	}
+
+	if listError != nil {
+		return fmt.Errorf("could not delete non-empty folder due to error when listing contents: %v", listError)
+	}
+	log.Printf("[DEBUG] force_destroy value: %#v", d.Get("force_destroy").(bool))
+	if !d.Get("force_destroy").(bool) {
```
kautikdk, line 270: If the folder with `force_destroy=false` that doesn't contain any sub-folder then it should get deleted without this error being thrown. 
We should do this validation after getting list of sub-folders. This was not caught in tests because there is no test for `force_destroy=false`. I recommend adding one test for the false value. 
Also It will same deletion logic for two different resources(objects and folders) so it would be good to write common logic for both and utilize that everywhere.

mmv1/third_party/terraform/services/storage/resource_storage_folder.go.tmpl
```diff
@@ -0,0 +1,343 @@
...
+
+func resourceStorageFolderImport(d *schema.ResourceData, meta interface{}) ([]*schema.ResourceData, error) {
+	config := meta.(*transport_tpg.Config)
+	if err := tpgresource.ParseImportId([]string{
+		"^(?P<bucket>[^/]+)/folders/(?P<name>.+)$",
+		"^(?P<bucket>[^/]+)/(?P<name>.+)$",
+	}, d, config); err != nil {
+		return nil, err
+	}
+
```
kautikdk, line 308: I believe we need to replace resource id from the import id using `tpgresource.ReplaceVars` function?

mmv1/third_party/terraform/services/storage/resource_storage_folder_test.go
```diff
@@ -0,0 +1,134 @@
...
+		Steps: []resource.TestStep{
+			{
+				Config:      testAccStorageFolder_storageBucket(bucketName, false, true) + testAccStorageFolder_storageFolder(true),
+				ExpectError: regexp.MustCompile("Error creating Folder: googleapi: Error 409: The bucket does not support hierarchical namespace., conflict"),
+			},
+		},
+	})
+}
+
+func TestAccStorageFolder_FolderForceDestroy(t *testing.T) {
```
kautikdk, line 55: Just a intuition, I believe this test won't check the actual logic of force destroy deeply? The reason is we are creating all nested resources using terraform config only which will be part of terraform dependency graph. The deletion will happen by individual resource's DELETE method so there won't any object or sub-folder which will be deleted by force destroy. 
It would be good if we can some how add object in folder outside of terraform and then try to test force destroy functionality.

mmv1/third_party/terraform/services/storage/resource_storage_folder.go.tmpl
gurusai-voleti: not passing to API but need to manage in tf state file, customers may switch from false to true and vice versa, based on this flag we manage the deletion logic when destroying the resource

mmv1/third_party/terraform/services/storage/resource_storage_folder.go.tmpl
```diff
@@ -0,0 +1,342 @@
...
+				Description: `The metadata generation of the folder.`,
+			},
+			"update_time": {
+				Type:        schema.TypeString,
+				Computed:    true,
+				Description: `The timestamp at which this folder was most recently updated.`,
+			},
+			"force_destroy": {
+				Type:        schema.TypeBool,
+				Optional:    true,
```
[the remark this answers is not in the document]
gurusai-voleti: will add false as default value

mmv1/third_party/terraform/services/storage/resource_storage_folder_test.go
gurusai-voleti: objects are deleted only when force_destroy is set to true otherwise objects will not be deleted and will not  continue the delete operation

mmv1/third_party/terraform/services/storage/resource_storage_folder.go.tmpl
gurusai-voleti: updated the logic, if no subfolders are present and force_destroy set to false, allows to delete folder

mmv1/third_party/terraform/services/storage/resource_storage_folder.go.tmpl
gurusai-voleti: removed

roaks3 (commented):
I just have a couple of high-level questions for context to start here, because this looks like a relatively complicated handwritten resource:
1. Is this different from managed_folder?
2. I didn't see any tickets requesting this, which would give some context if one exists, but also this is something I would have expected to get a lot of usage.
3. Could you clarify why this needs to be handwritten? It seems like the part where the folder needs to be empty would be one of the reasons.

kautikdk:
Hi @roaks3,

> Is this different from managed_folder? 
Yes, This is different. Managed folder is specifically to set IAM on name prefixes. Managed Folder can not be treated as file structure folders. Meanwhile HNS folders are actual folders that simulates file structure hierarchy for users. Users can use these folders to organize data in various folders and sub-folders.

> I didn't see any tickets requesting this, which would give some context if one exists, but also this is something I would have expected to get a lot of usage.
@gurusai-voleti please create a GitHub ticket. Yes, This resource will have a lot of usage. 

>Could you clarify why this needs to be handwritten? It seems like the part where the folder needs to be empty would be one of the reasons.
So currently there are some complicated issues that we are seeing and that would require this resource to be hand-written. Currently this PR doesn't have that part of code. @gurusai-voleti can you please mark this PR as a draft while we are working offline?

kautikdk:
@gurusai-voleti Can you please check if we can auto generate this resource using MMV1? Since we are not going for any complex operations, I believe we should be able to auto generate it. Not sure about path escaping for import operation, Please check if it working or not.

gurusai-voleti:
> @gurusai-voleti Can you please check if we can auto generate this resource using MMV1? Since we are not going for any complex operations, I believe we should be able to auto generate it. Not sure about path escaping for import operation, Please check if it working or not.

DELETE and UPDATE are two actions which require custom code and will generate code using tmpl

kautikdk:
You can remove hand-written files if MMV1 source code is enough for the feature support.

gurusai-voleti:
> You can remove hand-written files if MMV1 source code is enough for the feature support.

Delete, import, update are customised operations so I kept the handwritten tmpl file and updated the yaml with custom_code attribute

mmv1/products/storage/Folder.yaml
```diff
@@ -0,0 +1,88 @@
...
+    url_param_only: true
+  - name: 'name'
+    description: |
+      The name of the folder expressed as a path. Must include
+      trailing '/'. For example, `example_dir/example_dir2/`.
+    required: true
+    # The API returns values with trailing slashes, even if not
+    # provided. Enforcing trailing slashes prevents diffs and ensures
+    # consistent output.
+    validation:
```
roaks3, line 71: I think you may want some more rigorous validation here, mainly because it is used on the path. Do you know what the user will see if they provide a name like `/example_dir/`? If it results in a cryptic error, then validation might be better, and then you may want to do other things like ensure alphanumeric characters or similar.

mmv1/templates/terraform/custom_update/storage_folder_update.go.tmpl
```diff
@@ -0,0 +1,10 @@
+_ = config
+// we can only get here if force_destroy was updated
+if d.Get("force_destroy") != nil {
+    if err := d.Set("force_destroy", d.Get("force_destroy")); err != nil {
+        return fmt.Errorf("Error updating force_destroy: %s", err)
+    }
+}
```
roaks3, line 7: NIt: Since you're already adding this, you may want to go ahead and add the function that checks if this is the only field being changed (in case other fields are added later).

mmv1/templates/terraform/custom_delete/storage_folder_delete.go.tmpl
```diff
@@ -0,0 +1,88 @@
...
+        log.Printf("Error listing contents of bucket %s: %v", bucket, err)
+        listError = err
+        break
+    }
+
+    if len(res.Items) == 0 {
+        break // 0 items, folder empty
+    }
+
+    if !d.Get("force_destroy").(bool) {
```
roaks3, line 17: I would recommend moving this to the beginning, since you should be able to escape early

mmv1/products/storage/Folder.yaml
gurusai-voleti: if the path starts with /, the API validates the request body and rejects it as it cannot be started with /

mmv1/templates/terraform/custom_delete/storage_folder_delete.go.tmpl
gurusai-voleti: moved the logic to starting

gurusai-voleti: we first check if the folder contains any objects, if no objects we delete the folder even force_destory set to false, if any objects are there force_destory should be set to true then we delete the objects and folders
 
if force_destroy is true non-empty folders will be deleted, if force_destory is false empty folders are deleted so first we need to check if any objects present in the folders

mmv1/templates/terraform/custom_update/storage_folder_update.go.tmpl
gurusai-voleti: for now this should be fine

mmv1/templates/terraform/custom_delete/storage_folder_delete.go.tmpl
roaks3: Ah, thanks for clarifying, I was mixing this up with how `prevent_destroy` would be working.

mmv1/products/storage/Folder.yaml
roaks3: Right, but what I'm asking about is how exactly it responds when that value is rejected. For example, a 404 would be difficult for a user to understand when they actually need to remove a leading slash here.

gurusai-voleti: below is the response if folder name starts with /
"code": 409,
"message": "The parent folder does not exist.",

mmv1/products/storage/Folder.yaml
```diff
@@ -0,0 +1,88 @@
...
+  - name: 'name'
+    description: |
+      The name of the folder expressed as a path. Must include
+      trailing '/'. For example, `example_dir/example_dir2/`, `example@#/`, `a-b/d-f/`.
+    required: true
+    # The API returns values with trailing slashes, even if not
+    # provided. Enforcing trailing slashes prevents diffs and ensures
+    # consistent output.
+    validation:
+      regex: '^[^/\s][^\s/]*(/[^\s/]+)*\/$'
```
kautikdk, line 72: I think, We should match only trailing slash?

mmv1/templates/terraform/custom_delete/storage_folder_delete.go.tmpl
```diff
@@ -0,0 +1,88 @@
+bucket := d.Get("bucket").(string)
+name := d.Get("name").(string)
+
+var listError, deleteObjectError error
+for deleteObjectError == nil {
+    res, err := config.NewStorageClient(userAgent).Objects.List(bucket).Prefix(name).Do()
+    if err != nil {
+        log.Printf("Error listing contents of bucket %s: %v", bucket, err)
```
kautikdk, line 8: Nit: Folder not bucket

mmv1/products/storage/Folder.yaml
gurusai-voleti: yes will update to only trailing /, but as Ryan has raised concern updated the regex but as per naming conventions only trailing / is fine as the info provided in another comment

mmv1/templates/terraform/custom_delete/storage_folder_delete.go.tmpl
gurusai-voleti: updated

mmv1/products/storage/Folder.yaml
```diff
@@ -0,0 +1,88 @@
...
+    'Official Documentation': 'https://cloud.google.com/storage/docs/folders-overview'
+  api: 'https://cloud.google.com/storage/docs/json_api/v1/folders'
+custom_code:
+  custom_import: templates/terraform/custom_import/storage_folder.go.tmpl
+  custom_update: templates/terraform/custom_update/storage_folder_update.go.tmpl
+  custom_delete: templates/terraform/custom_delete/storage_folder_delete.go.tmpl
+virtual_fields:
+  - name: 'force_destroy'
+    description:
+      If set to true, folder will be force destroyed.
```
roaks3, line 52: Nit: I would mention that items within the folder will be destroyed

mmv1/products/storage/Folder.yaml
```diff
@@ -0,0 +1,88 @@
...
+    default_value: false
+parameters:
+  - name: 'bucket'
+    resource: 'Bucket'
+    imports: 'name'
+    description: 'The name of the bucket that contains the folder.'
+    required: true
+    immutable: true
+    url_param_only: true
+  - name: 'name'
```
roaks3, line 63: Should name be immutable, or can it be changed?

mmv1/products/storage/Folder.yaml
roaks3: Oh ok, that's a very permissive spec. Seems ok to leave it for now, and we can always add more client-side validation later if needed (all it would really improve is the error message, which I do think could be an issue for users thinking they need a leading slash).

mmv1/templates/terraform/custom_delete/storage_folder_delete.go.tmpl
```diff
@@ -0,0 +1,88 @@
...
+            }
+        })
+    }
+
+    // Wait for everything to finish.
+    wp.StopWait()
+}
+
+if listError != nil {
+    return fmt.Errorf("could not delete non-empty folder due to error when listing contents: %v", listError)
```
roaks3, line 58: Sorry if I missed this: the bucket implementation appears to attempt the bucket deletion even if there is a listError, and then only surface the error if the deletion was not successful. Was there a reason we didn't do the same here?

mmv1/templates/terraform/custom_delete/storage_folder_delete.go.tmpl
```diff
@@ -0,0 +1,88 @@
...
+}
+log.Printf("[DEBUG] force_destroy value: %#v", d.Get("force_destroy").(bool))
```
roaks3, line 60: Nit: probably don't need this?

mmv1/templates/terraform/custom_delete/storage_folder_delete.go.tmpl
```diff
@@ -0,0 +1,88 @@
...
+            ErrorRetryPredicates: []transport_tpg.RetryErrorPredicateFunc{transport_tpg.Is429RetryableQuotaError},
+        })
+        if err != nil {
+            return err
+        }
+    }
+
+    log.Printf("[DEBUG] Finished deleting Folder %q: %#v", d.Id(), name)
+} else {
+    deleteErr := fmt.Errorf("Error trying to delete folder without force_destroy set to true")
```
roaks3, line 84: Nit: "... X subfolders still exist within %name, use force_destroy = true to delete all subfolders", or similar, to provide more info?

mmv1/templates/terraform/examples/storage_folder_basic.tf.tmpl
```diff
@@ -0,0 +1,21 @@
+resource "google_storage_bucket" "bucket" {
+  name                        = "{{index $.Vars "bucket_name"}}"
+  location                    = "EU"
+  uniform_bucket_level_access = true
+  hierarchical_namespace {
+    enabled = true
+  }
+  force_destroy = true
```
roaks3, line 8: To be more concise and clear about the functionality, would it make sense to remove `force_destroy = true` from the bucket and subfolder? I don't believe they are needed.

mmv1/third_party/terraform/services/storage/resource_storage_folder_test.go
```diff
@@ -0,0 +1,238 @@
...
+
+	acctest.VcrTest(t, resource.TestCase{
+		PreCheck:                 func() { acctest.AccTestPreCheck(t) },
+		ProtoV5ProviderFactories: acctest.ProtoV5ProviderFactories(t),
+		Steps: []resource.TestStep{
+			{
+				Config: testAccStorageFolder_storageBucket(bucketName, true, true) + testAccStorageFolder_storageOneFolder(false),
+				Check: resource.ComposeTestCheckFunc(
+					testAccStorageCreatSubFolder(t, bucketName, folderName),
+					testAccStorageDeleteFolder(t, bucketName, folderName),
```
roaks3, line 115: I'm not sure I'm understanding this part. Why are we deleting `folder/` here? Shouldn't we instead do that with Terraform config (by removing it)?

roaks3 (commented):
Mostly minor items. Overall code and tests look good 👍

mmv1/templates/terraform/custom_delete/storage_folder_delete.go.tmpl
gurusai-voleti: in bucket deletion, we are not attempting to delete the bucket if the listError occurs

mmv1/templates/terraform/custom_delete/storage_folder_delete.go.tmpl
gurusai-voleti: yes removed

mmv1/templates/terraform/custom_delete/storage_folder_delete.go.tmpl
gurusai-voleti: yes even if subfolders exist and force_destroy set to true we can delete the sub folders  else don't delete

mmv1/products/storage/Folder.yaml
kautikdk: It should be immutable.
cc: @gurusai-voleti

mmv1/templates/terraform/custom_delete/storage_folder_delete.go.tmpl
kautikdk: I think we do try to delete bucket even if the listError occurs,
[Here](https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/third_party/terraform/services/storage/resource_storage_bucket.go.tmpl#L993) we are only breaking object deletion loop if error occurs. We try to delete bucket regardless of object deletion, [Reference](https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/third_party/terraform/services/storage/resource_storage_bucket.go.tmpl#L1055).

mmv1/third_party/terraform/services/storage/resource_storage_folder_test.go
gurusai-voleti: here we are creating a bucket and a folder with force_destroy false in the config step and later created a sub-folder inside a folder and now tried to delete parent folder, as force_destory set to false on parent folder we expect an error as parent folder is not empty (sub-folder exists),  seems this is the best way to produce this test case

mmv1/products/storage/Folder.yaml
gurusai-voleti: yes updated to immutable

mmv1/products/storage/Folder.yaml
gurusai-voleti: updated the description

mmv1/templates/terraform/custom_delete/storage_folder_delete.go.tmpl
gurusai-voleti: updated logic, attempt to delete the folder after listError

mmv1/templates/terraform/examples/storage_folder_basic.tf.tmpl
gurusai-voleti: removed from bucket and subfolder

mmv1/templates/terraform/custom_delete/storage_folder_delete.go.tmpl
gurusai-voleti: updated message

mmv1/templates/terraform/custom_delete/storage_folder_delete.go.tmpl
roaks3: It looks like this was added to the bucket implementation here: https://github.com/GoogleCloudPlatform/magic-modules/pull/2755. You can read the PR description, but it appears this was added to get around permissions (a user may be able to delete the bucket, but not list buckets).

I would defer to the two of you if the same issue will be present for folders.

mmv1/third_party/terraform/services/storage/resource_storage_folder_test.go
```diff
@@ -0,0 +1,249 @@
...
+}
+resource "google_storage_folder" "folder" {
+  bucket        = google_storage_bucket.bucket.name
+  name          = "folder/"
+  force_destroy = %t
+}
+resource "google_storage_folder" "subfolder" {
+  bucket        = google_storage_bucket.bucket.name
+  name          = "${google_storage_folder.folder.name}subfolder/"
+  force_destroy = %t
```
roaks3, line 210: Nit: it would be nice to remove this line, as I don't think it has any influence on the behavior, and actually implies that the user would need to set `force_destroy` in all subfolders for deletion to work.

roaks3 (commented):
One nit, and only open question IMO is around the "list" permission

mmv1/third_party/terraform/services/storage/resource_storage_folder_test.go
gurusai-voleti: updated

mmv1/templates/terraform/custom_delete/storage_folder_delete.go.tmpl
gurusai-voleti: updated the logic to attempt delete folder

mmv1/templates/terraform/custom_delete/storage_folder_delete.go.tmpl
```diff
@@ -54,14 +54,33 @@ for deleteObjectError == nil {
...
 }
-if deleteObjectError != nil {
+if gerr, ok := err.(*googleapi.Error); ok && gerr.Code == 409 && strings.Contains(gerr.Message, "not empty") && deleteObjectError != nil {
     return fmt.Errorf("could not delete non-empty folder due to error when deleting contents: %v", deleteObjectError)
 }
+if gerr, ok := err.(*googleapi.Error); ok && gerr.Code == 409 && strings.Contains(gerr.Message, "not empty") && !d.Get("force_destroy").(bool) {
+    return fmt.Errorf("Sub folders or items may exist within folder, use force_destroy to true to delete all subfolders: %v", err)
+}
+
+if err == nil {
```
roaks3, line 78: Nit: we might want to still log the err if it is not nil and we get to this point. While it wouldn't change behavior, it might help when debugging, since the error is otherwise swallowed.

roaks3 (approved):
One nit that isn't critical, everything else LGTM

I'll also wait for a 👍 from @kautikdk before merging

mmv1/templates/terraform/custom_delete/storage_folder_delete.go.tmpl
gurusai-voleti: added log for the err

mmv1/templates/terraform/custom_delete/storage_folder_delete.go.tmpl
kautikdk: Discussed offline, LGTM!

kautikdk (approved):
LGTM,
Please change release notes to make it more clear about new resource.

roaks3 (approved):
I just updated the release note for you. We actually don't include messages for new resources, just the resource name: https://googlecloudplatform.github.io/magic-modules/code-review/release-notes/

gurusai-voleti:
> I just updated the release note for you. We actually don't include messages for new resources, just the resource name: https://googlecloudplatform.github.io/magic-modules/code-review/release-notes/

could you merge the PR if ok, I don't have permission to do a merge

PR #12103, opened 2024-10-23
Add VPC Flow Logs Config API to terraform-provider-google-beta.

alkobi-google:
Add VPC Flow Logs Config API to terraform-provider-google-beta.
The API YAML was created using terraform-autogen.


hao-nan-li (changes requested):
Looks like there's some fields missing in the tests. Could you please add them in?

alkobi-google:
> Looks like there's some fields missing in the tests. Could you please add them in?

I added almost all of the missing fields in the new commit.
The ones currently missing are
* metadata-fields - which are absent because I set metadata to "exclude" and not "custom"
* interconnect-attachment - because the VPC Flow Logs Config applies to one of [interconnect attachment, VPN gateway], and this tests only the VPN gateway (and updates it's fields accordingly)

PR #12107, opened 2024-10-23
Promoting Memorystore to GA

himanikh:
Promote Memorystore Instance to GA

```release-note:new-resource
`google_memorystore_Instance`
```


mmv1/products/memorystore/Instance.yaml
```diff
@@ -12,75 +12,71 @@
 # limitations under the License.
 
 ---
-name: 'Instance'
```
melinath, line 15: Why are there so many formatting changes?

melinath (changes requested):
Could you resolve the conflict?

melinath (changes requested):
This seems to have gotten rebased and ended up with a bunch of extraneous commits / changes, plus there are still merge conflicts.

PR #12110, opened 2024-10-23
Adding missing fields to vertexai.Endpoint resource

mmurakowski-verily:
Fixes https://github.com/hashicorp/terraform-provider-google/issues/19995

We were missing a few fields so I added them. `enable_private_service_connect` is deprecated and seems to always return "false" now, so I'm skipping it.

`traffic_split` can't actually be set on resource creation, you need to "[deploy](https://cloud.google.com/vertex-ai/docs/reference/rest/v1beta1/projects.locations.endpoints/deployModel)" a model associated with the endpoint, and you need to have deployed a model in order to get the keys in the map. I think `traffic_split` only applies to the new config I'm enabling here, so it's hard to test anyway.

```release-note:enhancement
vertexai: added `traffic_split`, `private_service_connect_config`, `predict_request_response_logging_config`, `dedicated_endpoint_enabled`, and `dedicated_endpoint_dns` fields to `google_vertex_ai_endpoint` resource
```


mmurakowski-verily:
Not sure what's going on with the VCR-test step 😅; I can fix stuff but it looks like logs aren't being uploaded? IDK.

ScottSuarez (commented):
Need to discuss with contributor on traffic_split functionality

ScottSuarez (approved):

PR #12116, opened 2024-10-24
Add deletion_protection field to workflow resource

BENY4M1N:
```release-note:enhancement
workflows: added `deletion_protection` field to `google_workflows_workflow` resource
```

### DIFF
--- a/mmv1/products/workflows/Workflow.yaml
+++ b/mmv1/products/workflows/Workflow.yaml
@@ -60,6 +60,19 @@ examples:
       name: 'workflow'
       account_id: 'my-account'
     exclude_import_test: true
+    ignore_read_extra:
+      - 'deletion_protection'
+virtual_fields:
+  - name: 'deletion_protection'
+    description: |
+      Whether Terraform will be prevented from destroying the workflow. Defaults to true.
+      When a`terraform destroy` or `terraform apply` would delete the workflow,
+      the command will fail if this field is not set to false in Terraform state.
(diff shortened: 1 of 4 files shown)


SirGitsalot:
Mostly LGTM, but you'll need to implement the [pre-delete custom code](https://googlecloudplatform.github.io/magic-modules/develop/client-side-fields/#implement-logic).

Since this is breaking change (albeit one [that's allowed](https://googlecloudplatform.github.io/magic-modules/best-practices/deletion-behaviors/#deletion_protection)) it should have accompanying tests that update the `deletion_protection` field (and only the `deletion_protection` field). You can pretty much just copy and paste the `TestAccWorkflowsWorkflow_Update` test, but change it so that it's only updating `deletion_protection`.

Create one test with three steps that confirm going from `false` to `true` works as expected:

1. Create the resource with `deletion_protection` set to `false`
2. Update `deletion_protection` to `true`
3. Update `deletion_protection` again to be `false`

Steps one and two will simulate a workflow created with an earlier version of TPG being updated with a version of TPG containing this change; the third step just allows the workflow to be deleted after the test completes.

And add another test with two steps that go from `true` to `false`. It's a total of 5 test steps but only two config generator functions are necessary since the only difference is the value of `deletion_protection` .

BENY4M1N:
> SirGitsalot

Thanks for your thorough review. I have added the pre-delete custom code and additional VCR tests as you mentioned.

SirGitsalot (approved):

PR #12118, opened 2024-10-24
refactor: update resource manager tag references

wyardley:
Update the code for handwritten / templated go tests to use `foo.id` vs `"tagKeys/${foo.name}"` or `"tagValues/${foo.name}"`, and `foo.namespaced_name` vs. `"${data.google_project.project.project_id}/${foo.short_name}"`

This would be part 1: can update the resources for generated resources separately, and container_cluster will be covered in a separate PR.

Wrap map keys in parens where needed.

I tried to adjust spacing only in the affected blocks vs. doing any other terraform reformatting, except in one or two extreme cases.

```release-note:none

```

### DIFF
--- a/mmv1/third_party/terraform/services/bigquery/resource_bigquery_table_test.go
+++ b/mmv1/third_party/terraform/services/bigquery/resource_bigquery_table_test.go
@@ -4274,7 +4274,7 @@ resource "google_tags_tag_key" "key1" {
 }
 
 resource "google_tags_tag_value" "value1" {
-  parent = "tagKeys/${google_tags_tag_key.key1.name}"
+  parent     = google_tags_tag_key.key1.id
   short_name = "%{tag_value_name1}"
 }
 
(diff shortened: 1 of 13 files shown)


mmv1/third_party/terraform/services/bigquery/resource_bigquery_table_test.go
```diff
@@ -4324,8 +4324,8 @@ resource "google_bigquery_table" "test" {
   dataset_id = "${google_bigquery_dataset.test.dataset_id}"
   table_id   = "%{table_id}"
   resource_tags = {
-    "%{project_id}/${google_tags_tag_key.key1.short_name}" = "${google_tags_tag_value.value1.short_name}"
-    "%{project_id}/${google_tags_tag_key.key2.short_name}" = "${google_tags_tag_value.value2.short_name}"
+    (google_tags_tag_key.key1.namespaced_name) = google_tags_tag_value.value1.short_name
```
wyardley, line 4327: note: for these `namespaced_name` / ` short_name` pairs, I _suspect` they would work with `.id` on both side like in the other examples, but this makes the adjustments more minimal, as I believe it should be exactly equivalent with the old value.

If someone has a strong opinion about which is correct, they could easily be updated.

slevenick (approved):

PR #12121, opened 2024-10-24
Allow for single field indexes in 'google_firestore_index' resources to support __name__ DESC indexes

NickChittle:
Allow single field indexes in the google_firestore_index resource, this is to support having a descending index on the __name__ of the document.

```release-note:enhancement
firestore: allowed single field indexes to support `__name__ DESC` indexes in `google_firestore_index` resources
```

### DIFF
--- a/mmv1/products/firestore/Index.yaml
+++ b/mmv1/products/firestore/Index.yaml
@@ -81,6 +81,12 @@ examples:
       database_id: 'database-id-vector'
     test_env_vars:
       project_id: 'PROJECT_NAME'
+  - name: 'firestore_index_name_descending'
+    primary_resource_id: 'my-index'
+    vars:
+      database_id: 'database-id'
+    test_env_vars:
+      project_id: 'PROJECT_NAME'
 parameters:
 properties:
   - name: 'name'
@@ -175,5 +181,6 @@ properties:
               properties:
  # Meant to be an empty object with no properties.
                 []
-    # Single field indexes _exist_, but the API only lets us manage composite ones.
-    min_size: 2
+    # Most composite indexes require at least two fields, but it is possible
+    # for a user to require a single field index such as `__name__ DESC`.
+    min_size: 1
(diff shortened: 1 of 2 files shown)


rwhogg (approved):
Thanks!

shuyama1 (commented):
The change looks great—thank you for adding test coverage. Is there a GitHub issue or internal ticket associated with this change?

pcostell:
This was escalated from a customer directly to the engineering team. For internal users (b/374762870).

NickChittle:
Friendly Ping on this!

shuyama1 (commented):
Thank you for the reminder! I'm not very familiar with the resource, but I want to confirm that the case `__name__ DESC` cannot be achieved using google_firestore_field, which is designed for single field configuration. Is that correct?

NickChittle:
> Thank you for the reminder! I'm not very familiar with the resource, but I want to confirm that the case `__name__ DESC` cannot be achieved using `google_firestore_field`, which is designed for single field configuration. Is that correct?

It's a bit confusing for sure. Firestore has 'Automatic Single Field Indexes', for most fields these have a default configuration. The 'google_firestore_field' resource is for changing the default: https://firebase.google.com/docs/firestore/query-data/index-overview#automatic_indexing

The __name__ field is a special case field which is not handled under the normal 'Automatic Single Field Index' case, and must be handled as a 'Composite Index'

shuyama1:
> > Thank you for the reminder! I'm not very familiar with the resource, but I want to confirm that the case `__name__ DESC` cannot be achieved using `google_firestore_field`, which is designed for single field configuration. Is that correct?
> 
> It's a bit confusing for sure. Firestore has 'Automatic Single Field Indexes', for most fields these have a default configuration. The 'google_firestore_field' resource is for changing the default: [firebase.google.com/docs/firestore/query-data/index-overview#automatic_indexing](https://firebase.google.com/docs/firestore/query-data/index-overview#automatic_indexing)
> 
> The **name** field is a special case field which is not handled under the normal 'Automatic Single Field Index' case, and must be handled as a 'Composite Index'

I see, thanks for the clarification!

shuyama1 (approved):

PR #12132, opened 2024-10-25
refactor: update resource manager tag references

wyardley:
Part 2:

Update resource manager tag references in terraform code templates for generated tests (also some minor formatting changes)

- Update the code to use `foo.id` vs `"tagKeys/${foo.name}"` or `"tagValues/${foo.name}"`
- Update some project data source references to `project.id` vs `projects/.....number`
- Wrap map keys in parens where needed.

Followup to #12118

```release-note:none

```

### DIFF
--- a/mmv1/templates/terraform/examples/bigquery_dataset_resource_tags.tf.tmpl
+++ b/mmv1/templates/terraform/examples/bigquery_dataset_resource_tags.tf.tmpl
@@ -27,7 +27,7 @@ resource "google_bigquery_dataset" "{{$.PrimaryResourceId}}" {
   location      = "EU"
 
   resource_tags = {
-    (google_tags_tag_key.tag_key1.id) = google_tags_tag_value.tag_value1.id
-    (google_tags_tag_key.tag_key2.id) = google_tags_tag_value.tag_value2.id
+    (google_tags_tag_key.tag_key1.namespaced_name) = google_tags_tag_value.tag_value1.short_name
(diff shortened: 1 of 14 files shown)


slevenick (approved):

wyardley:
I’ll double check the changes, and try to run some of the failed ones locally - do the errors make it clear if the failures are due to this change?

slevenick:
At least the BQ one seems to be
        Error: Error creating Dataset: googleapi: Error 400: Invalid tagKey. Please use a valid tagKey., invalid

This one seems like a bad config, missing a quote or something: TestAccComputeRegionNetworkFirewallPolicyWithRules_computeRegionNetworkFirewallPolicyWithRulesFullExample

The others may be unrelated

wyardley:
@slevenick I have a fix for the BQ one that I'm pushing up (works locally). I couldn't figure out how to run the generated test locally, but I found the errant quote, so updating that as well.

wyardley:
https://github.com/hashicorp/terraform-provider-google/issues/19885 << I think this is still not fixed and that may be the issue with `TestAccComputeInstanceConfidentialInstanceConfigMain` and probably also `TestAccComputeInstanceFromMachineImage_confidentialInstanceConfigMain`.

wyardley:
I don't know if it's the right fix, but opened #12149 on the off chance that it is.

wyardley:
@slevenick can you try re-running those tests now that the test fix is merged?

slevenick (approved):

PR #12135, opened 2024-10-25
container: fix in-place updates for `node_config.containerd_config`

wyardley:
It appears that in-place updates were never implemented for `node_config.containerd_config`, when contained within either `google_container_cluster` or `google_container_node_pool`

Fixes hashicorp/terraform-provider-google#19056
Related to hashicorp/terraform-provider-google#19225 (already closed / fixed)

The actual implementation is pretty simple, and is basically copy  / paste from other fields.

The adjustments to existing tests are a little trickier to parse out.

Basically, the existing tests for both `google_container_cluster` _and_ `google_container_node_pool`  _mostly_ covered `node_config_defaults` (maybe because that's the only one updates worked for), and didn't cover updates in `node_config`. In at least one case, the same test was used for all of `node_config_defaults`, `node_pool.node_config`, and `node_config`, and I've mostly preserved that, just adjusting the order. Note that in at least one case, that will likely require recreating the entire cluster and / or its node-pools; I've moved that test (named `node_pool {}` block in `google_container_cluster`) to be the last test, and attempted to add some comments.

Another approach would be to split out the tests of `node_config_defaults`, `node_pool.node_config`, and `node_config`.

```release-note:bug
container: fixed in-place updates for `node_config.containerd_config` in `google_container_cluster` and `google_container_node_pool`
```

### DIFF
--- a/mmv1/third_party/terraform/services/container/resource_container_node_pool_test.go.tmpl
+++ b/mmv1/third_party/terraform/services/container/resource_container_node_pool_test.go.tmpl
@@ -4786,7 +4786,7 @@ func TestAccContainerNodePool_privateRegistry(t *testing.T) {
 		CheckDestroy:             testAccCheckContainerNodePoolDestroyProducer(t),
 		Steps: []resource.TestStep{
 			{
-				Config: testAccContainerNodePool_privateRegistryEnabled(secretID, cluster, nodepool, networkName, subnetworkName),
+				Config: testAccContainerNodePool_privateRegistryEnabled(secretID, cluster, nodepool, networkName, subnetworkName, "custom.example.com"),
 				Check: resource.ComposeAggregateTestCheckFunc(
 					resource.TestCheckResourceAttr(
 						"google_container_node_pool.np",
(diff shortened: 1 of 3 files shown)


mmv1/third_party/terraform/services/container/resource_container_node_pool_test.go.tmpl
```diff
@@ -4862,13 +4886,12 @@ resource "google_container_node_pool" "np" {
     oauth_scopes = [
       "https://www.googleapis.com/auth/cloud-platform",
     ]
-    machine_type = "n1-standard-8"
```
wyardley, line 4865: pulling all these out because a) was seeing a lot of stockout issues with n1, b) n1 is super old, c) n1 isn't needed in any of these for a reason, and d) 

I imagine it probably would be good to (separately) adjust to the n1 stuff in these tests to e2 or n2/n3? There are some comments about e2 not working with `pd-ssd`, but afaict, it works  fine now

SirGitsalot: Yeah, we've got other ticking time bombs like this scattered throughout our tests. The challenge is that APIs require that we specify a machine type/disk type/Linux distro version/etc. and it's often the case that the best we can do is use the latest available to give us the maximum useful lifetime for the test and then fix it when it inevitably breaks in a few years.

SirGitsalot (commented):
LGTM, I've kicked off the tests and will approve and merge once they're done.

mmv1/third_party/terraform/services/container/resource_container_node_pool_test.go.tmpl
wyardley: @SirGitsalot some of the comments imply it's because e2 (which I think is the default?) is required for certain tests (SSD is one example given). I think in some cases, those limitations have changed and / or the code was just copy / pasted.

Taking it out here seemed to work in my test, but happy to adjust it to a different explicit type if needed.

PR #12144, opened 2024-10-25
fix import documentation for string separated values

ScottSuarez:
fix import documentation for string separated values.
b/368130892#comment14

closes https://github.com/hashicorp/terraform-provider-google/issues/19453#issuecomment-2438737436

```release-note:none

```

### DIFF
--- a/mmv1/products/securitycenter/NotificationConfig.yaml
+++ b/mmv1/products/securitycenter/NotificationConfig.yaml
@@ -31,6 +31,8 @@ self_link: '{{name}}'
 create_url: 'organizations/{{organization}}/notificationConfigs?configId={{config_id}}'
 update_verb: 'PATCH'
 update_mask: true
+import_format:
+  - '{{name}}'
 timeouts:
   insert_minutes: 20
   update_minutes: 20
--- a/mmv1/products/securitycenter/ProjectNotificationConfig.yaml
+++ b/mmv1/products/securitycenter/ProjectNotificationConfig.yaml
@@ -32,6 +32,8 @@ self_link: '{{name}}'
 create_url: 'projects/{{project}}/notificationConfigs?configId={{config_id}}'
 update_verb: 'PATCH'
 update_mask: true
+import_format:
+  - '{{name}}'
 timeouts:
   insert_minutes: 20
   update_minutes: 20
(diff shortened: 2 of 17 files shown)


mmv1/products/securitycenter/NotificationConfig.yaml
```diff
@@ -31,6 +31,8 @@ self_link: '{{name}}'
 create_url: 'organizations/{{organization}}/notificationConfigs?configId={{config_id}}'
 update_verb: 'PATCH'
 update_mask: true
+import_format:
```
melinath, line 34: Why is this one necessary?

mmv1/products/securitycenter/ProjectNotificationConfig.yaml
```diff
@@ -32,6 +32,8 @@ self_link: '{{name}}'
 create_url: 'projects/{{project}}/notificationConfigs?configId={{config_id}}'
 update_verb: 'PATCH'
 update_mask: true
+import_format:
```
melinath, line 35: Same question

mmv1/products/securitycenterv2/OrganizationNotificationConfig.yaml
```diff
@@ -32,6 +32,9 @@ self_link: '{{name}}'
 create_url: 'organizations/{{organization}}/locations/{{location}}/notificationConfigs?configId={{config_id}}'
 update_verb: 'PATCH'
 update_mask: true
+import_format:
+  - '{{project}} {{name}}'
```
melinath, line 36: why does this one need `{{project}}` when the other two similar ones didn't?

tpgtools/main.go
```diff
@@ -562,6 +562,7 @@ func generateProductsFile(fileName string, products []*ProductMetadata) {
 
 var TemplateFunctions = template.FuncMap{
 	"title":                           strings.Title,
+	"containz":                        strings.Contains,
```
melinath, line 565: I'm a little unclear why this seems to be working as-is?

```suggestion
	"contains":                        strings.Contains,
```

melinath (commented):
This seems to add a bunch of import formats that are like `{{project}}/{{name}}`, which I wasn't expecting - just want to double-check that those also seem to work?

tpgtools/main.go
ScottSuarez: loool yeah this is not needed

mmv1/products/securitycenterv2/OrganizationNotificationConfig.yaml
ScottSuarez: This one is wrong

melinath (approved):
This seems to add a bunch of import formats that are like `{{project}}/{{name}}`, which I wasn't expecting - just want to double-check that those also seem to work?

PR #12146, opened 2024-10-25
Unbreak TestAccBigQueryDataset_withProvider5

SirGitsalot:
The test was using a config shared with v6+ tests and was setting the attribution label, which doesn't work pre v6.

```release-note:none
```


wj-chen (approved):
Thank you for the fix!

c2thorn (approved):

PR #4849, opened 2021-06-07
Add `advanced_machine_features` to GCE Instances

upodroid:
Fixes: https://github.com/hashicorp/terraform-provider-google/issues/9251

If this PR is for Terraform, I acknowledge that I have:

- [X] Searched through the [issue tracker](https://github.com/hashicorp/terraform-provider-google/issues) for an open issue that this either resolves or contributes to, commented on it to claim it, and written "fixes {url}" or "part of {url}" in this PR description. If there were no relevant open issues, I opened one and commented that I would like to work on it (not necessary for very small changes).
- [X] [Generated Terraform](https://github.com/GoogleCloudPlatform/magic-modules#generating-downstream-tools), and ran `make test` and `make lint` to ensure it passes unit and linter tests.
- [X] Ensured that all new fields I added that can be set by a user appear in at least one [example](https://github.com/GoogleCloudPlatform/magic-modules/tree/master/templates/terraform/examples) (for generated resources) or [third_party test](https://github.com/GoogleCloudPlatform/magic-modules/tree/master/third_party/terraform/tests) (for handwritten resources or update tests).
- [X] [Ran](https://github.com/hashicorp/terraform-provider-google/blob/master/.github/CONTRIBUTING.md#tests) relevant acceptance tests (If the acceptance tests do not yet pass or you are unable to run them, please let your reviewer know).
- [x] Read the [Release Notes Guide](https://github.com/GoogleCloudPlatform/magic-modules/blob/master/.ci/RELEASE_NOTES_GUIDE.md) before writing my release note below.

**Release Note Template for Downstream PRs (will be copied)**

```release-note:enhancement
compute: added `advanced_machine_features` to `google_compute_instance`
```

PR #12148, opened 2024-10-26
compute: added support for `advanced_machine_features.turbo_mode`

wyardley:
- Added support for `advanced_machine_features.turbo_mode` in `google_compute_instance`, `google_compute_instance_template`, and `google_compute_region_instance_template`
- Adjusted tests to use c4 instance types for `advanced_machine_features` tests
- Minor docs and formatting fixes

Fixes hashicorp/terraform-provider-google#20023

I'm guessing this may be missing something

I looked at adding it in `google_container_cluster` as well, but seems like it's not implemented yet in regular or beta container API.

Side note: `EnableUefiNetworking` and `PerformanceMonitoringUnit` also seem to be available [here](https://pkg.go.dev/google.golang.org/api@v0.203.0/compute/v1#AdvancedMachineFeatures) but not implemented in the provider yet.

Should I add anything in tgc as well (like #10848)?

```release-note:enhancement
compute: added support for `advanced_machine_features.turbo_mode` to `google_compute_instance`, `google_compute_instance_template`, and `google_compute_region_instance_template`
```

### DIFF
--- a/mmv1/third_party/terraform/services/compute/resource_compute_instance.go.tmpl
+++ b/mmv1/third_party/terraform/services/compute/resource_compute_instance.go.tmpl
@@ -54,6 +54,13 @@ func IpCidrRangeDiffSuppress(k, old, new string, d *schema.ResourceData) bool {
 }
 
 var (
+	advancedMachineFeaturesKeys = []string{
+		"advanced_machine_features.0.enable_nested_virtualization",
+		"advanced_machine_features.0.threads_per_core",
+		"advanced_machine_features.0.turbo_mode",
+		"advanced_machine_features.0.visible_core_count",
+	}
+
 	bootDiskKeys = []string{
 		"boot_disk.0.auto_delete",
 		"boot_disk.0.device_name",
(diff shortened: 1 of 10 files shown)


wyardley:
Note: these tests pass for me locally, but only if I explicitly set `subnetwork = default` in `network_interface`; given the number of other tests that are using the default network (and no explicitly set subnetwork) vs. a bootstrapped created network, I assume it'll be Ok.

mmv1/third_party/terraform/services/compute/resource_compute_instance.go.tmpl
```diff
@@ -1091,19 +1098,26 @@ be from 0 to 999,999,999 inclusive.`,
 						"enable_nested_virtualization": {
 							Type:        schema.TypeBool,
 							Optional:    true,
-							AtLeastOneOf: []string{"advanced_machine_features.0.enable_nested_virtualization","advanced_machine_features.0.threads_per_core"},
+							AtLeastOneOf: advancedMachineFeaturesKeys,
```
wyardley, line 1101: I'm assuming it's Ok to require just at least one of these (or even an empty block might be Ok?

I didn't see obvious evidence of one of them requiring another, but it's possible the rules could be a little more complex here?

mmv1/third_party/terraform/services/compute/resource_compute_instance_template.go.tmpl
```diff
@@ -964,6 +964,12 @@ be from 0 to 999,999,999 inclusive.`,
 							ForceNew:    true,
 							Description: `The number of threads per physical core. To disable simultaneous multithreading (SMT) set this to 1. If unset, the maximum number of threads supported per core by the underlying processor is assumed.`,
 						},
+						"turbo_mode": {
+							Type:        schema.TypeString,
+							Optional:    true,
```
wyardley, line 969: not sure if any of these need to be computed?

mmv1/third_party/terraform/services/compute/resource_compute_instance.go.tmpl
roaks3: I would keep it to be safe. It looks like this was only added to prevent an empty block from being specified (see https://github.com/GoogleCloudPlatform/magic-modules/pull/4849#discussion_r659056546), and perhaps we don't technically need it, but if the behavior of `advanced_machine_features: {}` is no different than not specifying it, then it's better to enforce the more clear option (not specifying).

mmv1/third_party/terraform/services/compute/resource_compute_instance_template.go.tmpl
roaks3: Probably not. If it did, we would expect to see some users who don't specify the value getting permadiffs. We can also verify by looking at the debug logs once the CI tests run.

roaks3 (approved):
LGTM, just need to verify tests

wyardley:
btw, if the TDX / confidential image tests for compute fail, that may be a known / pre-existing issue (https://github.com/hashicorp/terraform-provider-google/issues/19885). I have #12149 up as a possible fix for those.

wyardley:
@roaks3 that test PR is merged if you want to re-run the tests now.

roaks3:
Yea, I think the tests you have are sufficient, and the test failures look unrelated.

PR #12154, opened 2024-10-26
spanner: added `google_spanner_database` data source

wyardley:
Add a `google_spanner_database` data source.

Fixes hashicorp/terraform-provider-google#13470

```release-note:new-datasource
`google_spanner_database`
```

### DIFF
--- a/mmv1/third_party/terraform/provider/provider_mmv1_resources.go.tmpl
+++ b/mmv1/third_party/terraform/provider/provider_mmv1_resources.go.tmpl
@@ -205,6 +205,7 @@ var handwrittenDatasources = map[string]*schema.Resource{
 	"google_service_accounts":                          resourcemanager.DataSourceGoogleServiceAccounts(),
 	"google_site_verification_token":                   siteverification.DataSourceSiteVerificationToken(),
 	"google_sourcerepo_repository":                     sourcerepo.DataSourceGoogleSourceRepoRepository(),
+	"google_spanner_database":                          spanner.DataSourceSpannerDatabase(),
 	"google_spanner_instance":                          spanner.DataSourceSpannerInstance(),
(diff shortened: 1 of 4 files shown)


wyardley:
This basically seems to work well, except for ddl showing up as `list(null)`; not sure if that's intentional (or a known issue) or not

It's ignored for import state verify here:
https://github.com/GoogleCloudPlatform/magic-modules/blob/551f48015dbed65b65a5610cbb41b02f7f801bc6/mmv1/third_party/terraform/services/spanner/resource_spanner_database_test.go.tmpl#L175

I'm guessing it's not the most likely thing people will want to get with this data source; if I ignore `ddl.#` with `acctest.CheckDataSourceStateMatchesResourceStateWithIgnores()`, the test stops complaining about the # of entries not matching, but somehow, ignoring `ddl.%` doesn't seem to resolve the error with DDL not matching.  I can also make the test pass by not defining DDL at the database level, but I'd like to get a better understanding of whether we want the ddl entries to show up in the data source or not, and why that ignore isn't working (and should probably have a docs update if it isn't supported).

SirGitsalot (approved):
It looks like there's a [separate API that can be called to get the DDL](https://cloud.google.com/spanner/docs/reference/rest/v1/projects.instances.databases/getDdl) (why though?!). Since this PR solves the problem in the linked issue and adding the DDL later if the need arises should not be a breaking change I'll go ahead and approve. Thanks!

wyardley:
> It looks like there's a [separate API that can be called to get the DDL](https://cloud.google.com/spanner/docs/reference/rest/v1/projects.instances.databases/getDdl) (why though?!).

Yeah, I wasn't sure if it was intentional that the attribute was marked `ignore_read`, though not `sensitive` (and if size of the DDL or potential sensitive info in there was a consideration)...

PR #12156, opened 2024-10-27
Google dataform config

Tuseeq1:
Adding a new resource Config to Google Cloud Dataform.
Internal tracking bug: b/366417072

Product Resource API Documentation:
https://cloud.google.com/dataform/reference/rest/v1beta1/projects.locations

```release-note:new-resource
`google_dataform_config` (beta)
```


mmv1/products/dataform/Config.yaml
```diff
@@ -0,0 +1,68 @@
...
+name: 'Config'
+description: |-
+  A singleton resource represents a Dataform configuration for project location.
+min_version: 'beta'
+references:
+  guides:
+    'Official Documentation': 'https://cloud.google.com/dataform/docs/'
+  api: 'https://cloud.google.com/dataform/reference/rest/v1beta1/projects.locations'
+
+id_format: 'projects/{{project}}/locations/{{region}}/config'
```
roaks3, line 24: I think you can remove this, it should default to the value in `self_link`. Same with `create_url` and `update_url` (default value should work).

mmv1/products/dataform/Config.yaml
```diff
@@ -0,0 +1,68 @@
...
+parameters:
+  - name: 'region'
+    type: String
+    required: true
+    min_version: 'beta'
+    description: 'A reference to the region'
+    immutable: true
+    url_param_only: true
+properties:
+  - name: 'name'
```
roaks3, line 57: This `name` attribute is a little weird. From what I see in the API documentation, it would just contain redundant information like project and region, so should it just be `output: true`? I'm also not sure why it would need a name since it is a singleton.

mmv1/templates/terraform/examples/dataform_config.tf.tmpl
```diff
@@ -0,0 +1,40 @@
+data "google_project" "project" {
```
roaks3, line 1: In our tests, we share a single project (per environment). Because this is a singleton on that project, and all that's being changed is the kms key, I think you will need to instead create a dedicate project per-test run. If you don't the kms key would need to be bootstrapped (otherwise they will accumulate), and then the test will not change anything on each run.

PR #12158, opened 2024-10-28
Bp tf resource

niharika-98:

```release-note:new-resource
`google_backup_dr_backup_plan`
```
```release-note:new-datasource
`google_backup_dr_backup_plan`
```


niharika-98:
\gcbrun

mmv1/products/backupdr/BackupPlan.yaml
```diff
@@ -0,0 +1,209 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+name: 'BackupPlan'
+base_url: projects/{{project}}/locations/{{location}}/backupPlans
+create_url: projects/{{project}}/locations/{{location}}/backupPlans/?backup_plan_id={{backup_plan_id}}
+self_link: projects/{{project}}/locations/{{location}}/backupPlans/{{backup_plan_id}}
+immutable: true
+delete_url: projects/{{project}}/locations/{{location}}/backupPlans/{{backup_plan_id}}
```
roaks3, line 19: You should be able to remove this line, I think this is the default value used (`self_link`)

mmv1/products/backupdr/BackupPlan.yaml
```diff
@@ -0,0 +1,209 @@
...
+timeouts:
+  insert_minutes: 60
+  delete_minutes: 60
+examples:
+  - min_version: beta
+    name: 'backup_dr_backup_plan'
+    primary_resource_id: 'my-backup-plan'
+    test_env_vars:
+      project: :PROJECT_NAME
+    exclude_test: true
```
roaks3, line 35: Why is this test excluded? If nothing else, I would suggest a basic test if possible, like you use for the datasource test.

mmv1/products/backupdr/BackupPlan.yaml
```diff
@@ -0,0 +1,209 @@
...
+            - name: 'backupWindow'
+              type: NestedObject
+              description: |
+                A BackupWindow defines the window of the day during which backup jobs will run. Jobs are queued at the beginning of the window and will be marked as
+                `NOT_RUN` if they do not start by the end of the window.
+              properties:
+                - name: 'startHourOfDay'
+                  type: Integer
+                  description: |
+                    The hour of the day (0-23) when the window starts, for example, if the value of the start hour of the day is 6, that means the backup window starts at 6:00.
```
roaks3, line 203: Since it looks like 0 is a valid option, you may need `send_empty_value: true` here

mmv1/templates/terraform/examples/backup_dr_backup_plan.tf.tmpl
```diff
@@ -0,0 +1,160 @@
...
+    days_of_week              = list(string)  # Optional for WEEKLY
+    days_of_month             = list(number)  # Optional for MONTHLY
+    week_day_of_month         = object({      # Optional for MONTHLY
+      week_of_month = string
+      day_of_week   = string
+    })
+    months                    = list(string)  # Optional for YEARLY
+  }))
+}
+default = [
```
roaks3, line 19: We don't normally see testing like this, and I believe it is just because it is harder to isolate why the test failed vs having separate examples. These also shows up in the docs, and for acceptance tests, we are normally more interested in the API interaction vs the full behavior of the endpoint. So that's all to say, I would maybe simplify this into 2 separate examples: 1 with a minimal config, and another with all fields specified.

roaks3 (commented):
Initial review looked good, main thing is that I think the tests need a few changes. Once those look good we can confirm they are passing, and I can do a final review.

mmv1/templates/terraform/examples/backup_dr_backup_plan.tf.tmpl
```diff
@@ -0,0 +1,163 @@
...
+    }
+  ]
+}
+
+resource "google_backup_dr_backup_vault" "my_backup_vault" {
+  provider = google-beta
+  location = "us-central1"
+  backup_vault_id = "bv-test1"
+  description     = "This is a backup vault built by Terraform."
+  backup_minimum_enforced_retention_duration = "100000s"
```
roaks3, line 109: For this resource, it looks like only `location`, `backup_vault_id`, and `backup_minimum_enforced_retention_duration` are required. Could we removed the other fields (unless they are necessary for the test? Same for you other example.

roaks3 (commented):
Caught up over chat, but it sounds like the complex ("full") test should probably be changed to remove the variable, and we should not exclude it for testing.

Otherwise all LGTM, just these final small testing pieces.

roaks3 (approved):

PR #12174, opened 2024-10-28
Remove unnecessary update API call for client-side-only field changes in google_bigquery_table

wj-chen:
Fixes https://github.com/hashicorp/terraform-provider-google/issues/19485, following instructions added in https://github.com/GoogleCloudPlatform/magic-modules/pull/11330.

Tested locally with `TF_LOG=DEBUG terraform apply` for client-side-only field changes and mixed changes.

```release-note:none
```

### DIFF
--- a/mmv1/third_party/terraform/services/bigquery/resource_bigquery_table.go
+++ b/mmv1/third_party/terraform/services/bigquery/resource_bigquery_table.go
@@ -1903,6 +1903,19 @@ type TableReference struct {
 }
 
 func resourceBigQueryTableUpdate(d *schema.ResourceData, meta interface{}) error {
+	// If only client-side fields were modified, short-circuit the Update function to avoid sending an update API request.
+	clientSideFields := map[string]bool{"deletion_protection": true}
+	clientSideOnly := true
+	for field := range ResourceBigQueryTable().Schema {
+		if d.HasChange(field) && !clientSideFields[field] {
+			clientSideOnly = false
+			break
+		}
+	}
+	if clientSideOnly {
+		return resourceBigQueryTableRead(d, meta)
+	}
+
 	config := meta.(*transport_tpg.Config)
 	userAgent, err := tpgresource.GenerateUserAgentString(d, config.UserAgent)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


ScottSuarez (changes requested):
Can you add an update test for this ?
https://googlecloudplatform.github.io/magic-modules/develop/test/test/#add-an-update-test

Should be relatively simple. Just want to make sure that this works going forward. My only concern is that we don't `return resourceBigQueryTableRead(d, meta)` like the original function does. I see you've tested locally and I don't foresee any problem but it would be great to capture this behavior in a test!

wj-chen:
> Can you add an update test for this ? https://googlecloudplatform.github.io/magic-modules/develop/test/test/#add-an-update-test
> 
> Should be relatively simple. Just want to make sure that this works going forward. My only concern is that we don't `return resourceBigQueryTableRead(d, meta)` like the original function does. I see you've tested locally and I don't foresee any problem but it would be great to capture this behavior in a test!

Added a test. And good point about `return resourceBigQueryTableRead(d, meta)`. I added it back since if the table was deleted outside of Terraform before the update, we should detect that correctly and treat it as a creation again. Thanks for the suggestion.

mmv1/third_party/terraform/services/bigquery/resource_bigquery_table.go
```diff
@@ -1903,6 +1903,19 @@ type TableReference struct {
 }
 
 func resourceBigQueryTableUpdate(d *schema.ResourceData, meta interface{}) error {
+	// If only client-side fields were modified, short-circuit the Update function to avoid sending an update API request.
+	clientSideFields := map[string]bool{"deletion_protection": true}
+	clientSideOnly := true
+	for field := range ResourceBigQueryTable().Schema {
+		if d.HasChange(field) && !clientSideFields[field] {
```
ScottSuarez, line 1910: I assume nested changes will be caught by `d.HasChange(field)`. Would be possible adding a test only changing a nested subfield?

wj-chen: Sure, added.

ScottSuarez (approved):

ScottSuarez:
Thanks for the work!

PR #11479, opened 2024-08-17
apigee: Added in-place update for consumer_accept_list field in google_apigee_instance

entertvl:
Fixes https://github.com/hashicorp/terraform-provider-google/issues/14237.

**Release Note Template for Downstream PRs (will be copied)**

```release-note:enhancement
apigee: Added in-place update for `consumer_accept_list` field in `google_apigee_instance`.
```

PR #12178, opened 2024-10-29
fix(apigee): Added in-place update of `google_apigee_environment` resource

entertvl:
- fix https://github.com/hashicorp/terraform-provider-google/issues/19820

```release-note:bug
apigee: added in-place update of `google_apigee_environment` resource
```

### DIFF
--- a/mmv1/products/apigee/Environment.yaml
+++ b/mmv1/products/apigee/Environment.yaml
@@ -114,13 +114,13 @@ properties:
     description: |
       Display name of the environment.
     required: false
-    immutable: true
+    immutable: false
   - name: 'description'
     type: String
     description: |
       Description of the environment.
     required: false
-    immutable: true
+    immutable: false
   - name: 'deploymentType'
     type: Enum
     description: |
--- a/mmv1/third_party/terraform/services/apigee/resource_apigee_environment_update_test.go
+++ b/mmv1/third_party/terraform/services/apigee/resource_apigee_environment_update_test.go
@@ -10,7 +10,6 @@ import (
 )
 
 func TestAccApigeeEnvironment_apigeeEnvironmentUpdateTest(t *testing.T) {
-	acctest.SkipIfVcr(t)
 	t.Parallel()
 
 	context := map[string]interface{}{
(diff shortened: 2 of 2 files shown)


mmv1/third_party/terraform/services/apigee/resource_apigee_environment_update_test.go
```diff
@@ -0,0 +1,115 @@
...
+import (
+	"testing"
+
+	"github.com/hashicorp/terraform-plugin-testing/helper/resource"
+
+	"github.com/hashicorp/terraform-provider-google/google/acctest"
+	"github.com/hashicorp/terraform-provider-google/google/envvar"
+)
+
+func TestAccApigeeEnvironment_apigeeEnvironmentUpdateTest(t *testing.T) {
```
hao-nan-li, line 12: What exactly is this test updating? Can you make this a 2 steps test to show the update?

entertvl: Do you mean splitting `resource.TestStep` within one test case, or separating it into a different `func`?
Just wanted to confirm, thanks!

hao-nan-li: Looks like I missed `testAccApigeeEnvironment_apigeeEnvironmentBasicTestExample` in the test first step. Is there any particular reason to skip the vcr test here?

entertvl: To be honest, I don’t fully understand the exact reason for skipping VCR test.
In the past [PR](https://github.com/GoogleCloudPlatform/magic-modules/pull/11479), we had a same situation.

I’ll remove the VCR skip for now and see if it works.

mmv1/products/apigee/Environment.yaml
```diff
@@ -114,13 +114,13 @@ properties:
     description: |
       Display name of the environment.
     required: false
-    immutable: true
+    immutable: false
```
hao-nan-li, line 117: I noticed that this resource has multiple fields with `immutable` set to true. Is this PR change intended to make these 2 fields updatable, or the entire resource updatable?

entertvl: This change makes only `DisplayName` and `Description`.

Although the following issue refers only to Description, DisplayName is also included as updatable:
https://github.com/hashicorp/terraform-provider-google/issues/19820

hao-nan-li (approved):

PR #12180, opened 2024-10-29
Partly rollback changes from #11321 to fix an `alias_ip_range` bug

### LINKED ISSUE #19765: alias_ip_range in google_compute_instance not maintaining IPs in proper order as of 5.41
### Community Note

Terraform v1.9.7
on windows_amd64
+ provider registry.terraform.io/hashicorp/google v6.4.0

### Affected Resource(s)

google_compute_instance

### Terraform Configuration

```tf
resource "google_compute_instance" "vm" {

  name        = "testvm"
  hostname    = "testvm"
  machine_type = "n1-standard-1"
  zone                = "us-central1-a"
  project             = "testproject"

  boot_disk {
    initialize_params {
       image = "debian-cloud/debian-11"
    }
  }


  network_interface {
    subnetwork         = "mysubnet" #10.10.10.0/24
    subnetwork_project = "mysubnet-project"
    alias_ip_range {
        ip_cidr_range = "/32"
    }
     alias_ip_range {
        ip_cidr_range = "10.10.10.2/32"
    }
  }

}
```

### Debug Output

_No response_

### Expected Behavior

terraform plan right after the apply does not show any changes

### Actual Behavior

terraform plan wants to modify one of the IPs and makes things worse by deleting/mixing up IPs if allowed to proceed

### Steps to reproduce

1. `terraform apply`
2. `terraform plan`

### Important Factoids

Previous to v5.41 everything was fine. Starting in v5.41 this issue occurs and seems related to the fix for alias_ip_range in [19015](https://github.com/hashicorp/terraform-provider-google/pull/19015). Seems it might be the call to tpgresource.SortMapsByConfigOrder in compute_instance_helpers.go that is causing the issue because it's changing the order of the IPs?

I depend on the order of the IPs in state matching the order of the code shows (given there is no unique key to access them) and them being flipped is an issue (and it's even an issue for terraform given it can't create the vm without wanting to change it immediately after)

### References

_No response_

karolgorc:
rollback from https://github.com/GoogleCloudPlatform/magic-modules/pull/11321
closes https://github.com/hashicorp/terraform-provider-google/issues/19765 (more context in the comments of this issue)

This issue will get reintroduced https://github.com/hashicorp/terraform-provider-google/issues/12286

This cannot be easily fixed without introducing new edge cases or crazy overhead for this field hence the rollback. The problems are:
- alias_ip's cannot be simultaneously removed and added in the same request. THIS IS AN API LIMITATION
- alias_ip's are a list of blocks so it cannot be validated and checked against other values or have a customized diff because you cannot access data from another block in customdiff, diffsuppress or validate functions
- Because alias_ip's are a list of blocks their order in the code matters. So if a user changes the order of them without changing the IP's this is a diff that we have to account for
- This field is getting auto sorted on the API's side (causing diff in config) and also has computed values when given a syntax like "/32". This makes it very complicated to cover all of the different cases.

Overall, i'd say that we should accept the fact that the array of alias_ip ranges get's cleared in between requests and treat it as an API limitation until it gets fixed. I don't think we should build additional UX features on top of an API method that doesn't work according to Google's standards (it should support deletion and creation at the same time)

partly rollback because i'm leaving the tests introduced in the previous PR. They could be useful if this gets fixed in the future

```release-note:bug
compute: `alias_ip_range` on `google_compute_instance` will no longer cause ordering issues when updating with "/xx" syntax
```

### DIFF
--- a/mmv1/third_party/terraform/services/compute/compute_instance_helpers.go.tmpl
+++ b/mmv1/third_party/terraform/services/compute/compute_instance_helpers.go.tmpl
@@ -53,28 +53,17 @@ func expandAliasIpRanges(ranges []interface{}) []*compute.AliasIpRange {
 	return ipRanges
 }
 
-func flattenAliasIpRange(d *schema.ResourceData, ranges []*compute.AliasIpRange, i int) []map[string]interface{} {
-	prefix := fmt.Sprintf("network_interface.%d", i)
+func flattenAliasIpRange(ranges []*compute.AliasIpRange) []map[string]interface{} {
+	rangesSchema := make([]map[string]interface{}, 0, len(ranges))
 
-	configData := []map[string]interface{}{}
-	for _, item := range d.Get(prefix + ".alias_ip_range").([]interface{}) {
-		configData = append(configData, item.(map[string]interface{}))
-	}
-
-	apiData := make([]map[string]interface{}, 0, len(ranges))
 	for _, ipRange := range ranges {
-		apiData = append(apiData, map[string]interface{}{
+		rangesSchema = append(rangesSchema, map[string]interface{}{
 			"ip_cidr_range":         ipRange.IpCidrRange,
 			"subnetwork_range_name": ipRange.SubnetworkRangeName,
 		})
 	}
 
-	//permadiff fix
-	sorted, err := tpgresource.SortMapsByConfigOrder(configData, apiData, "ip_cidr_range")
-	if err != nil {
-		return apiData
-	}
-	return sorted
+	return rangesSchema
 }
 
 func expandScheduling(v interface{}) (*compute.Scheduling, error) {
@@ -414,7 +403,7 @@ func flattenNetworkInterfaces(d *schema.ResourceData, config *transport_tpg.Conf
 			"subnetwork":          tpgresource.ConvertSelfLinkToV1(iface.Subnetwork),
 			"subnetwork_project":  subnet.Project,
 			"access_config":       ac,
-			"alias_ip_range":      flattenAliasIpRange(d, iface.AliasIpRanges, i),
+			"alias_ip_range":      flattenAliasIpRange(iface.AliasIpRanges),
 			"nic_type":            iface.NicType,
 			"stack_type":          iface.StackType,
 			"ipv6_access_config": flattenIpv6AccessConfigs(iface.Ipv6AccessConfigs),
--- a/mmv1/third_party/terraform/services/compute/resource_compute_instance.go.tmpl
+++ b/mmv1/third_party/terraform/services/compute/resource_compute_instance.go.tmpl
@@ -2270,9 +2270,6 @@ func resourceComputeInstanceUpdate(d *schema.ResourceData, meta interface{}) err
 					Fingerprint:     instNetworkInterface.Fingerprint,
 					ForceSendFields: []string{"AliasIpRanges"},
 				}
-				if commonAliasIpRanges := CheckForCommonAliasIp(instNetworkInterface, networkInterface); len(commonAliasIpRanges) > 0 {
-					ni.AliasIpRanges = commonAliasIpRanges
-				}
 				op, err := config.NewComputeClient(userAgent).Instances.UpdateNetworkInterface(project, zone, instance.Name, networkName, ni).Do()
 				if err != nil {
 					return errwrap.Wrapf("Error removing alias_ip_range: {{"{{"}}err{{"}}"}}", err)
@@ -3319,21 +3316,4 @@ func isEmptyServiceAccountBlock(d *schema.ResourceData) bool {
 		return true
 	}
 	return false
-}
-
-// Alias ip ranges cannot be removed and created at the same time. This checks if there are any unchanged alias ip ranges
-// to be kept in between the PATCH operations on Network Interface
-func CheckForCommonAliasIp(old, new *compute.NetworkInterface) []*compute.AliasIpRange {
-	newAliasIpMap := make(map[string]bool)
-	for _, ipRange := range new.AliasIpRanges {
-		newAliasIpMap[ipRange.IpCidrRange] = true
-	}
-
-	resultAliasIpRanges := make([]*compute.AliasIpRange, 0)
-	for _, val := range old.AliasIpRanges {
-		if newAliasIpMap[val.IpCidrRange] {
-			resultAliasIpRanges = append(resultAliasIpRanges, val)
-		}
-	}
-	return resultAliasIpRanges
 }
--- a/mmv1/third_party/terraform/services/compute/resource_compute_instance_test.go.tmpl
+++ b/mmv1/third_party/terraform/services/compute/resource_compute_instance_test.go.tmpl
@@ -19,7 +19,6 @@ import (
 	"github.com/hashicorp/terraform-plugin-testing/helper/resource"
 	"github.com/hashicorp/terraform-plugin-testing/terraform"
 
-	"github.com/stretchr/testify/assert"
 	"github.com/hashicorp/terraform-provider-google/google/acctest"
 	"github.com/hashicorp/terraform-provider-google/google/envvar"
 	tpgcompute "github.com/hashicorp/terraform-provider-google/google/services/compute"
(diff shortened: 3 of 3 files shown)


karolgorc:
Something changed with allow_stopping_for_update since this was merged i'm guessing.

```
resource_compute_instance_test.go:2730: Step 2/6 error running import: ImportStateVerify attributes not equivalent. Difference is shown below. The - symbol indicates attributes missing after import.
        
          map[string]string{
        -       "allow_stopping_for_update": "true",
          }
```

Putting it in as VerifyIgnore.

karolgorc:
Is this failed check a part of some migration? Because version guards are not changed here

image

shuyama1 (commented):
The rollback seems reasonable to me, as the original fix is potentially a breaking change(?) since it changes the order of `alias_ip_range` blocks in state. The change has led to another issue that breaks existing configs, as discussed in https://github.com/hashicorp/terraform-provider-google/issues/19765.

I'm inviting @melinath, the reviewer of the original PR https://github.com/GoogleCloudPlatform/magic-modules/pull/11321, to weigh in on this rollback, to ensure I haven’t overlooked any potential issues that could impact users.

shuyama1:
> Is this failed check a part of some migration? Because version guards are not changed here
> 
> image

I believe this should be resolved by https://github.com/GoogleCloudPlatform/magic-modules/pull/12208. If you rebase your PR, the check should pass on the next run.

melinath:
yeah, sorting by config order assumes that all the blocks are explicitly defined in the config; since the API can add computed values into the mix that we can't tie back to a specific block in the config, sorting by config order is insufficient. Rolling this back seems reasonable.

melinath:
What are the problems besides the /32 generating a random value? That seems like it would be relatively easy to compensate for.

karolgorc:
The "/32" isn't that big of an issue. I already wrote and tested an algorithm that solved this issue but the big problems start when a user adds a new `alias_ip` block in the middle of 2 other ones, or at the beginning. same goes for deletion and update. Then the sorting just doesn't work and gets lost in the midst of wrong ordering in local state, unexplainable sorting from API, and being given unknown values.

Overall a very edge-case heavy implementation and it seems like too much of a hustle to implement this kind of logic for a field that's out of the ordinary when it comes to functionality for Google

melinath:
It sounds like there may be some issues with how the "sort by config order" helper functions work that would be nice to resolve (alongside the /32 issue). I'm not sure I have a good feel for whether the current behavior or the previous permadiff is worse; I don't really want us to back out this behavior just to introduce the old bug.

FWIW it should be possible to do a diff suppress with [CustomizeDiff](https://github.com/hashicorp/terraform-provider-google/issues/18392), since that operates on the entire resource at once. (Though it looks like it might have some trouble with known-after-apply values...)

@shuyama1 what's your take?

karolgorc:
The only CustomDiff that i can think off that wouldn't cause additional edge-cases would be one that ignores any sorting related diff on all `alias_ip_range` blocks inside a single network interface. Not sure how it will behave against the "/32" wildcards though

shuyama1 (commented):
Sorry for the delayed response—I’m a bit behind on reviews.

I agree that a forward fix would be preferable to reverting and reintroducing the previous bug, if possible.  

Adding CustomizeDiff ito ignore reordering, as well as including the logic for netmask comparisons similar to [IpCidrRangeDiffSuppress](https://github.com/GoogleCloudPlatform/magic-modules/blob/1f0d1a4a0e48ce20347c80dc35a954611280ec0a/mmv1/third_party/terraform/services/compute/resource_compute_instance.go.tmpl#L32) might be a potential solution. However, I haven't tested this myself, so there’s a chance it’s much more complex than I think.

@karolgorc Would you mind try implementing CustomizeDiff to see if it resolves the issue? Thanks!

karolgorc:
We are already using the [IpCidrRangeDiffSuppress](https://github.com/GoogleCloudPlatform/magic-modules/blob/1f0d1a4a0e48ce20347c80dc35a954611280ec0a/mmv1/third_party/terraform/services/compute/resource_compute_instance.go.tmpl#L32) function on this field.

```go
"ip_cidr_range": {
  Type:             schema.TypeString,
  Required:         true,
  DiffSuppressFunc: IpCidrRangeDiffSuppress,
  Description:      `The IP CIDR range represented by this alias IP range.`,
},
```

The problem with it is that it works only in the scope of a single IP. Causing additional problems (the 2nd apply to see the ip address thing discussed earlier). 

> @karolgorc Would you mind try implementing CustomizeDiff to see if it resolves the issue? Thanks!

I'll try to build something like that with the function to retain addresses in between requests that's currently implemented

shuyama1 (commented):
> We are already using the [IpCidrRangeDiffSuppress](https://github.com/GoogleCloudPlatform/magic-modules/blob/1f0d1a4a0e48ce20347c80dc35a954611280ec0a/mmv1/third_party/terraform/services/compute/resource_compute_instance.go.tmpl#L32) function on this field.
> 
> ```go
> "ip_cidr_range": {
>   Type:             schema.TypeString,
>   Required:         true,
>   DiffSuppressFunc: IpCidrRangeDiffSuppress,
>   Description:      `The IP CIDR range represented by this alias IP range.`,
> },
> ```
> 
> The problem with it is that it works only in the scope of a single IP. Causing additional problems (the 2nd apply to see the ip address thing discussed earlier).
> 

Understood. That’s why I suggested trying to move the logic to CustomizeDiff, as it operates at the resource level. However, I’m not sure if this will fully resolve the issue.

> > @karolgorc Would you mind try implementing CustomizeDiff to see if it resolves the issue? Thanks!
> 
> I'll try to build something like that with the function to retain addresses in between requests that's currently implemented

Thank you! Please let me know if you have any questions or need any assistance with the implementation.

karolgorc:
Closing this and will repoen with a new solution

shuyama1 (commented):
> After testing i'd say that we are currently blocked by the terraform SDK here. CustomDiff doesn't allow for API operations so we cannot really operate on these values. The only thing we have is `schema.ResourceDiff.GetChanges()`.
> 
> After building a logic to ignore the sorting of these we cannot really single out "some" ranges that are causing problems for us. `schema.ResourceDiff.Clear()` only works for deleting a diff of an entire block
> 
> * we can specify `network_interface.0.alias_ip_range` to be cleared
> * but cannot specify `network_interface.0.alias_ip_range.0` to be cleared
> 
> I've also confirmed the cause of the behaviour in the linked issue. It's not really an issue with the `SortMapsByConfigOrder` helper function but the `IpCidrRangeDiffSuppress` sending wrong data when `/32` is put before any other `alias_ip_range`
> 
> Here is an explanation.
> 
> 1. Let's say that this is our original NIC config
>    image
> 2. And we change it to this
>    image
> 3. This should shift the first 2 alias_ip_ranges and add the last one again as it's the last element of the array
>    image
> 4. But as you can see it doesn't change the first element as per working of IpCidrRangeDiffSuppress to ignore any inputs here that begin with "/". "/32" in our case
> 5. the produced result passed downstream to other functions is [ 192.168.1.10, 192.168.1.10, 192.168.1.20 ]
> 
> IMPORTANT to mention that when `/32` is put as the last element this will not happen as it's not seen as a `CHANGE` and rathen an `ADDITION`
> 
> image image
> 
> problem is that it's a DiffSuppressFunction. It cannot work on external data from outside of itself. I think that moving this function to work the same but as a CustomDiff gives us more wiggle room to do anything. But to perform any actions on these fields we would have to make these `Computed: true`. Customdiff functions like `Clear, SetNew` work only on computed attributes but for that to work we have to remove `Required: true` from `ip_cidr_range` because Computed and Required cannot be set together

Ah, I see where the issue comes from now. Thanks for the deep dive and clear explanation! I'll explore this a bit locally and get back to you soon.

melinath:
@shuyama1 is there a reason this isn't being reviewed?

shuyama1:
We are a bit hesitate to roll back the change as it would reintroduce a previous bug, so we're trying to find a fix forward. I wasn't able to find a solution during my last investigation and haven't had time recently to revisit this. I'll look into it further this/next week. Apologize for the delay.

shuyama1 (commented):
Apologies for the late reply. After further review, I believe a fix-forward approach is better than a rollback. ~I propose we change the field type from list to set (breaking change) and then add a DSF or CustomDiff to suppress the diff. Since changing the field type is a breaking change, this will need to be part of our next major release~. (not the correct approach after investigation. Update: working on a fix-forward) I will take on the fix when the major release window opens.

Converting this PR to a draft then. Appreciate your work on this issue.

PR #12182, opened 2024-10-29
Fix address group ordering for network firewall policy rule

slevenick:
Fixes: https://github.com/hashicorp/terraform-provider-google/issues/18134

```release-note:bug
compute: fixed a diff based on server-side reordering of `match.src_address_groups` and `match.dest_address_groups` in `google_compute_network_firewall_policy_rule`
```


roaks3 (approved):
LGTM

PR #12187, opened 2024-10-29
feat: Add `disks` support to compute_node_template

jjtroberts:
This PR adds Local SSD disk settings support.

Fixes https://github.com/hashicorp/terraform-provider-google/issues/20026

```release-note:enhancement
compute: added `disks` field to `google_compute_node_template` resource
```

### DIFF
--- a/mmv1/products/compute/NodeTemplate.yaml
+++ b/mmv1/products/compute/NodeTemplate.yaml
@@ -175,3 +175,23 @@ properties:
     enum_values:
       - 'ENABLED'
       - 'NONE'
+  - name: 'disks'
+    type: Array
+    description: |
+      List of the type, size and count of disks attached to the
+      node template
+    item_type:
+      type: NestedObject
+      properties:
+        - name: 'diskCount'
+          type: Integer
+          description: |
+            The number of disks exposed to this node template.
+        - name: 'diskType'
+          type: String
+          description: |
+            The type of disks exposed to this node template.
+        - name: 'diskSizeGb'
+          type: Integer
+          description: |
+            The disk pool size in GB.
--- a/mmv1/products/compute/NodeTemplate.yaml
+++ b/mmv1/products/compute/NodeTemplate.yaml
@@ -59,6 +59,10 @@ examples:
     primary_resource_id: 'template'
     vars:
       template_name: 'soletenant-with-accelerators'
+  - name: 'node_template_disks'
+    primary_resource_id: 'template'
+    vars:
+      template_name: 'soletenant-with-disks'
 parameters:
   - name: 'region'
     type: ResourceRef
--- a/mmv1/products/compute/NodeTemplate.yaml
+++ b/mmv1/products/compute/NodeTemplate.yaml
@@ -190,12 +190,12 @@ properties:
         - name: 'diskCount'
           type: Integer
           description: |
-            The number of disks exposed to this node template.
+            Specifies the number of such disks.
         - name: 'diskType'
           type: String
           description: |
-            The type of disks exposed to this node template.
+            Specifies the desired disk type on the node. This disk type must be a local storage type (e.g.: local-ssd). Note that for nodeTemplates, this should be the name of the disk type and not its URL.
         - name: 'diskSizeGb'
           type: Integer
           description: |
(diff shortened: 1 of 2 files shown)


mmv1/products/compute/NodeTemplate.yaml
```diff
@@ -175,3 +179,23 @@ properties:
...
+      type: NestedObject
+      properties:
+        - name: 'diskCount'
+          type: Integer
+          description: |
+            The number of disks exposed to this node template.
+        - name: 'diskType'
+          type: String
+          description: |
+            The type of disks exposed to this node template.
```
zli82016, line 197: Can you copy the description of this field from the [API doc](https://cloud.google.com/compute/docs/reference/rest/v1/nodeTemplates) to here?

mmv1/products/compute/NodeTemplate.yaml
```diff
@@ -175,3 +179,23 @@ properties:
...
       - 'ENABLED'
       - 'NONE'
+  - name: 'disks'
+    type: Array
+    description: |
+      List of the type, size and count of disks attached to the
+      node template
+    item_type:
+      type: NestedObject
+      properties:
```
zli82016, line 189: Just want to double check if these three properties `diskCount`, `diskType` and `diskSizeGb` are required or optional?

mmv1/templates/terraform/examples/node_template_disks.tf.tmpl
```diff
@@ -0,0 +1,16 @@
+data "google_compute_node_types" "central1a" {
+  zone     = "us-central1-a"
+}
+
+resource "google_compute_node_template" "{{$.PrimaryResourceId}}" {
+  name      = "{{index $.Vars "template_name"}}"
+  region    = "us-central1"
+  node_type = "n2-node-80-640"
+
+  disks {
```
zli82016, line 10: Can you add one more `disks` block to test if multiple disks work as expected?

zli82016 (changes requested):

mmv1/templates/terraform/examples/node_template_disks.tf.tmpl
jjtroberts: The API itself is misleading because it does seem to support multiple sets of disk pools, however in actual application (at least for the past two years I've been working with sole-tenant nodes) only one type of disk is available for selection (`local-ssd`) and you have only three choices: None, 16/6TB, 24/9TB.

image

Even the api field descriptions are somewhat inaccurate. Take `diskSizeGb` for example: "integerSpecifies the size of the disk in base-2 GB." However this must only be `375` which is the partition size of local-ssd disks.

The API will only use the last `--disk` argument passed in:

```sh
gcloud compute sole-tenancy node-templates create local-ssd-test \
  --node-type=n2-node-80-640 \
  --region=us-central1 \
  --node-affinity-labels="foo=bar" \
  --disk type=local-ssd,count=16,size=375 \
  --disk type=local-ssd,count=24,size=375
```

```sh
> gcloud compute sole-tenancy node-templates describe local-ssd-test
cpuOvercommitType: NONE
creationTimestamp: '2024-11-02T08:02:56.823-07:00'
disks:
- diskCount: 24
  diskSizeGb: 375
  diskType: local-ssd
id: 'xxxxx'
kind: compute#nodeTemplate
name: local-ssd-test
nodeAffinityLabels:
  foo: bar
nodeType: n2-node-80-640
region: https://www.googleapis.com/compute/v1/projects/xxxxx/regions/us-central1
selfLink: https://www.googleapis.com/compute/v1/projects/xxxxx/regions/us-central1/nodeTemplates/local-ssd-test
serverBinding:
  type: RESTART_NODE_ON_ANY_SERVER
status: READY
```
image

mmv1/products/compute/NodeTemplate.yaml
jjtroberts: Adding a pool of local ssds is completely optional.

mmv1/products/compute/NodeTemplate.yaml
jjtroberts: I've made all three match the API field descriptions.

PR #12208, opened 2024-10-31
add all missing template check cases

NickElliot:
should've been more thorough on the earlier pr. due to the number of cases needing to be covered now (as go templates err toward not using a trailing white space eliminator because it is much more aggressive), this list is looking very silly and I'll make a note to refactor this check to be a simpler regex, but best to have it working for all cases possible now

```release-note:none

```


rileykarson (approved):

wyardley:
Ah, phew. I assume I can get rid of the changes like [this](https://github.com/GoogleCloudPlatform/magic-modules/pull/12216/commits/51d74e0787c9973a55b078735c67438fbead53d9) after this goes in?

tools/template-check/gotemplate/gotemplate.go
```diff
@@ -12,18 +12,25 @@ import (
 // Note: this is allowlisted to combat other issues like `=` instead of `==`, but it is possible we
 // need to add more options to this list in the future, like `private`. The main thing we want to
 // prevent is targeting `beta` in version guards, because it mishandles either `ga` or `private`.
+
+// TODO: refactor to be a regex as the permitted formats are more varied under go templates
```
wyardley, line 16: That was my thought, maybe something like this?
`\{\{-? if (?:ne|eq) \$\.TargetVersionName [\x60\"]ga[\x60\"] \-?\}\}` (assuming regular go regex vs. re2)

(and could also choose to be flexible about spaces, though guessing better not here?)

https://regex101.com/r/jh3Uct/1

NickElliot: Yeah, actually doing the expression wouldnt be hard, just need to adjust a bit of the lower logic (as currently it works by doing a literal string.contain check over the entire array of allowed guards)

PR #12210, opened 2024-10-31
triv: Update resource page in developer docs

gorge511:

```release-note:none
Update outdated links on resource page in developer docs
```


hao-nan-li (approved):
I don't think the TGC build error is related to this change. Also I removed the changelog notes as it's not necessary for a doc change.

gorge511:
Thanks @hao-nan-li. I also think the failed test is unrelated.

Thanks for fixing the release note. I thought that marking it as `release-note:none` is the correct way and then the text is just for this PR.

PR #12220, opened 2024-11-01
transcoder: ignored `start_time` in `JobManifestsExample` test

wyardley:
Ignored `start_time` for `TestAccTranscoderJob_transcoderJobManifestsExample` (assume the other tests aren't failing as often, but could ignore this in some / all of the other ones where `end_time` is already ignored if desired.

Fixes hashicorp/terraform-provider-google#20004

```release-note:none

```

### DIFF
--- a/mmv1/products/transcoder/Job.yaml
+++ b/mmv1/products/transcoder/Job.yaml
@@ -76,6 +76,7 @@ examples:
     ignore_read_extra:
       - 'state'
       - 'end_time'
+      - 'start_time'
 parameters:
   - name: 'location'
     type: String
--- a/mmv1/products/transcoder/Job.yaml
+++ b/mmv1/products/transcoder/Job.yaml
@@ -75,8 +75,8 @@ examples:
       bucket_name: 'transcoder-job'
     ignore_read_extra:
       - 'state'
-      - 'end_time'
       - 'start_time'
+      - 'end_time'
 parameters:
   - name: 'location'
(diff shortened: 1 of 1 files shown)


mmv1/products/transcoder/Job.yaml
```diff
@@ -75,6 +75,7 @@ examples:
     ignore_read_extra:
       - 'state'
       - 'end_time'
+      - 'start_time'
```
wyardley, line 78: should these be alphasorted, or is `state` a special one?

wyardley:
comment (will delete)

mmv1/products/transcoder/Job.yaml
NickElliot: we don't have any particular requirement, I'd prefer `start_time` preceding `end_time` purely for aesthetic reasons though

NickElliot (changes requested):
if you could do that minor reorder, otherwise LGTM!

mmv1/products/transcoder/Job.yaml
wyardley: Okies. Done.

wyardley:
@NickElliot done. q: I think this is the one that's caused issues with tests, apparently, but should the field be added to any other attributes as well, or just see if this fixes it?

NickElliot (approved):
LGTM!

PR #12221, opened 2024-11-01
Added support for additional values KCP_HPA in `google_container_clus…

MalusGreen:
…ter.logging_config

 Adds terraform support for new HPA Observability logs.

```release-note: enhancement
container: added additional value `KCP_HPA` for `logging_config.enable_components` field in `google_container_cluster` resource
```


SarahFrench:
Note: the `mmv1-check-templates / version-guard-check (pull_request)` check isn't a blocker for this PR, and the check was updated in https://github.com/GoogleCloudPlatform/magic-modules/pull/12208.

I'll return to this PR once it's marked as ready for review

SarahFrench:

hao-nan-li (commented):
I'm not sure if the test failure is caused by the change.

Have you tried to run the failed test locally?

MalusGreen:
The cl passed all presubmits before the current sync. I don't believe the current error is caused by the PR. I'll sync and reun the CI.

hao-nan-li:
Are you able to run this test locally and see if it works?

MalusGreen:
> Are you able to run this test locally and see if it works?

This change runs. The problem is our api change has not propagated and is being blocked internally. Will reopen at a later time.

MalusGreen:

roaks3 (changes requested):
The change here is simple enough, but from the test failures, it looks like it still isn't shipped.

Please review go/terraform-launch-stages if you haven't already, but assuming this is an option we can support in the GA `google` provider, is there something else that needs to be changed for the feature to work? The error message mentions the GKE version being too old.

MalusGreen:
> The change here is simple enough, but from the test failures, it looks like it still isn't shipped.
> 
> Please review go/terraform-launch-stages if you haven't already, but assuming this is an option we can support in the GA `google` provider, is there something else that needs to be changed for the feature to work? The error message mentions the GKE version being too old.

The change has been launched to prod but requires, as seen in the error, a higher GKE version than the current default: https://screenshot.googleplex.com/7pkLmTfJjtc36SF

What would be the fastest way to update this default for the VCR tests?

roaks3:
I think you got it! Let me run tests and see if that resolves.

roaks3:
I think the problem is that the field only sets the version on initial creation (I think?). If that's the case, you probably need to specify that version in each of the configs used for that test, ie.
* testAccContainerCluster_basic
* testAccContainerCluster_withLoggingConfigEnabled
* testAccContainerCluster_withLoggingConfigDisabled
* testAccContainerCluster_withLoggingConfigUpdated

roaks3 (approved):

PR #12227, opened 2024-11-01
alloydb: re-enabled VCR for `TestAccAlloydb*` tests

wyardley:
Fixes hashicorp/terraform-provider-google#16231

From [this comment](https://github.com/hashicorp/terraform-provider-google/issues/16231#issuecomment-1986290508), I think these 3 tests can just be re-enabled, unless there's something else specific the owning team has to do?

All 3 of these pass in both recording and replaying mode in my test project, though IIRC, I got a segfault vs. just an error before enabling `servicenetworking` (is there a utility to ensure a specific service is enabled in the default test project). I think you have a way to run this in Teamcity too, which would probably confirm if the test will work there?

```release-note:none

```

### DIFF
--- a/mmv1/products/alloydb/Backup.yaml
+++ b/mmv1/products/alloydb/Backup.yaml
@@ -75,7 +75,6 @@ examples:
     ignore_read_extra:
       - 'reconciling'
       - 'update_time'
-    exclude_test: true
   - name: 'alloydb_backup_full_test'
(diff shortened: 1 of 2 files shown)


wyardley:
I _think_ the unit test failures here are unrelated?

hao-nan-li (approved):
Yeah they are unrelated.

PR #12230, opened 2024-11-01
feat: move emptydir to GA for cloudrun and cloudrunv2

bskaplan:
Move empty_dir support to ga.

```release-note:enhancement
cloudrun: promoted `empty_dir` field in `google_cloud_run_service` to GA
```

```release-note:enhancement
cloudrunv2: promoted `empty_dir` field in `google_cloud_run_v2_service` and `google_cloud_run_v2_job` to GA
```

### DIFF
--- a/mmv1/products/cloudrun/Service.yaml
+++ b/mmv1/products/cloudrun/Service.yaml
@@ -848,7 +848,6 @@ properties:
                       type: NestedObject
                       description: |-
                         Ephemeral storage which can be backed by real disks (HD, SSD), network storage or memory (i.e. tmpfs). For now only in memory (tmpfs) is supported. It is ephemeral in the sense that when the sandbox is taken down, the data is destroyed with it (it does not persist across sandbox runs).
-                      min_version: 'beta'
                       properties:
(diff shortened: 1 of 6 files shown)


hao-nan-li (commented):
For `Error: Inconsistent dependency lock file`, most likely there are some more lines to remove related to Beta/GA settings.

bskaplan:
Still seeing failures on the examples and I don't see any difference between these and the other examples. Is the problem that I'm trying to move an existing example from beta -> GA? Should I leave these examples specifying the beta provider even though it's not needed any more?

hao-nan-li:
> Still seeing failures on the examples and I don't see any difference between these and the other examples. Is the problem that I'm trying to move an existing example from beta -> GA? Should I leave these examples specifying the beta provider even though it's not needed any more?

From the generated diff report, looks like I can't find both failing tests in for `google provider` or the GA provider.

bskaplan:
Found the issue. I was looking in the wrong place before.

hao-nan-li (approved):

PR #12231, opened 2024-11-01
tgc fix: Check that storageClient is not nil before fetching GCS bucket

jalseth:
When running the conversion in offline mode, this client is nil which causes the converter to panic.

```release-note:none
tgc: fixed a panic in storageClient when running tfplan2cai converter in offline mode
```


zli82016 (approved):
LGTM.

PR #12239, opened 2024-11-04
Added controlled-vault feature to backupdr-backupvault

liyun-huang:
```release-note: enhancement
backupdr: added `access_restriction` field to`google_backup_dr_backup_vault` resource (beta)
```


liyun-huang:
TestAccBackupDRBackupPlan_backupDrBackupPlanSimpleExample failed test is not relevant to changed files.

mmv1/products/backupdr/BackupVault.yaml
```diff
@@ -198,3 +198,15 @@ properties:
     description: "Optional. User annotations. See https://google.aip.dev/128#annotations\nStores
       small amounts of arbitrary data. "
     min_version: 'beta'
+  - name: 'accessRestriction'
```
slevenick, line 201: Is this field updatable? If so, please include it in an update test and update the value.

liyun-huang: This field is not updatable - it's only used during create backup vault step.

slevenick: It should be `immutable: true` then

liyun-huang:
@slevenick Hi Sam, please let me know if there are anything I should address for this PR. Thanks!

slevenick (approved):

PR #12240, opened 2024-11-04
Sync main feature branch ephemeral resources

BBBmau:
This is to include the recent go version bump from 1.21 -> 1.23 which is necessary to support ephemerals

```release-note:none

```


SarahFrench:
Looks like a bunch of automation was stopped by evil build killer. I think let's go ahead with merge despite that, as the work on the feature branch is all new code and should not impact any of the new code coming in from main.

@BBBmau When merging the thing I forgot to do previously was merge using this method, which I think requires some repo-wide changes to allow us to do:

>Note: `commit merge` instead of `squash and merge` method is needed for any sync PRs

I recommend talking to the Google team about this before merging this PR, even if the outcome is deciding to squash merge after all

SarahFrench (approved):

SarahFrench:
Oh, I can see that this is actually available as an option without repo-wide changes:

Screenshot 2024-11-05 at 11 19 51

In that case I'll leave merge to you, but make sure to use that merge method

PR #12245, opened 2024-11-05
Add performance config field to Filestore instance

yaelbarak:
Performance configuration support for Filestore.

```release-note:enhancement
filestore: added `performance_config` field to `google_filestore_instance`
```

### DIFF
--- a/mmv1/products/filestore/Instance.yaml
+++ b/mmv1/products/filestore/Instance.yaml
@@ -292,3 +292,39 @@ properties:
     type: String
     description: |
       The reason for enabling deletion protection.
+  - name: 'performanceConfig'
+    type: NestedObject
+    description: |
+      Performance configuration for the instance. If not provided,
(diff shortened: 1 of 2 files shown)


SarahFrench (commented):
Hi, thanks for your PR!

The [test failure reported here](https://github.com/GoogleCloudPlatform/magic-modules/pull/12245#issuecomment-2457343175) fails with this error:

> Error: Error creating Instance: googleapi: Error 400: ConfigurablePerformance is disabled

Do you know if this is a feature that needs to be opted-into at a project level?

yaelbarak:
@SarahFrench The Terraform projects are now allowlisted for the feature guarding experiment. I reran the tests and they're now passing.

SarahFrench (approved):
Thanks for resolving the testing issues!

There isn't any public documentation I can cross-reference with during review, but I'm happy to approve based on the acceptance tests

PR #12246, opened 2024-11-05
spanner: updated docs for database name regex

wyardley:
Fixes hashicorp/terraform-provider-google#17608

This basically just unifies the regex in the doc with the regex that's actually applied.

```release-note:none

```

### DIFF
--- a/mmv1/products/spanner/Database.yaml
+++ b/mmv1/products/spanner/Database.yaml
@@ -103,8 +103,8 @@ properties:
   - name: 'name'
     type: String
     description: |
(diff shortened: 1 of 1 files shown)


wyardley:
@slevenick does it make sense to format the regex (maybe in backticks, which seem to be supported here?)

wyardley:
Updated 👍 

Looks like they show up in pink vs. grey, but I think this will still make it stand out a little more and more readable.

slevenick (approved):

PR #12249, opened 2024-11-05
Add TF support for LocalSsdEncryptionMode flag

halimsam:
Add terraform support for local_ssd_encryption_mode on cluster/nodepool create.

```release-note:enhancement
container: added `local_ssd_encryption_mode` to `node_config` in `google_container_cluster` and `google_container_node_pool`
```


mmv1/third_party/terraform/services/container/node_config.go.tmpl
```diff
@@ -806,6 +806,14 @@ func schemaNodeConfig() *schema.Schema {
 						ForceNew: true,
 						Description: `If enabled boot disks are configured with confidential mode.`,
 				},
+				"local_ssd_encryption_mode": {
+					Type:             schema.TypeString,
+					Optional:         true,
+					Computed:         true,
```
rileykarson, line 812: This means that the field has a default value returned by the server even when unset, and won't be returned as `null`. Is that correct? 

`Computed` means that if the field is removed from the config we will preserve the old value instead of revert to null/empty/falsey. If it will only ever be returned if set explicitly, `Computed` can be removed, which is generally preferable. However, if the API may change the default or return non-null defaults in the future (which is breaking for an API but tends to happen regardless) we should keep it.

mmv1/third_party/terraform/services/container/resource_container_node_pool_test.go.tmpl
```diff
@@ -4157,6 +4157,65 @@ resource "google_container_node_pool" "np" {
...
+	networkName := acctest.BootstrapSharedTestNetwork(t, "gke-cluster")
+	subnetworkName := acctest.BootstrapSubnet(t, "gke-cluster", networkName)
+
+	acctest.VcrTest(t, resource.TestCase{
+		PreCheck:	  func() { acctest.AccTestPreCheck(t) },
+		ProtoV5ProviderFactories: acctest.ProtoV5ProviderFactories(t),
+		CheckDestroy: testAccCheckContainerClusterDestroyProducer(t),
+		Steps: []resource.TestStep{
+			{
+				Config: testAccContainerNodePool_withLocalSsdEncryptionMode(clusterName, np, networkName, subnetworkName, "EPHEMERAL_KEY_ENCRYPTION"),
```
rileykarson, line 4174: Can we add a second config step with an updated value, since it's mutable? https://googlecloudplatform.github.io/magic-modules/develop/test/test/#add-an-update-test

Only this test is fine- GKE cluster shares code so it'd be redundant.

mmv1/third_party/terraform/services/container/node_config.go.tmpl
```diff
@@ -806,6 +806,14 @@ func schemaNodeConfig() *schema.Schema {
...
+					ValidateFunc:     validation.StringInSlice([]string{"LOCAL_SSD_ENCRYPTION_MODE_UNSPECIFIED", "STANDARD_ENCRYPTION", "EPHEMERAL_KEY_ENCRYPTION"}, false),
+					Description:      `LocalSsdEncryptionMode specified the method used for encrypting the local SSDs attached to the node.`,
+					DiffSuppressFunc: tpgresource.EmptyOrDefaultStringSuppress("LOCAL_SSD_ENCRYPTION_MODE_UNSPECIFIED"),
```
rileykarson, line 815: I'd just remove all references to LOCAL_SSD_ENCRYPTION_MODE_UNSPECIFIED unless your API returns it (which I very much doubt- none do that I know of). For Terraform empty/null serves the same purpose.

halimsam: Removed.

mmv1/third_party/terraform/services/container/node_config.go.tmpl
halimsam: This field should only be used (or configured) in boot disk so there shouldn't be any further changes. Will remove `computed`

mmv1/third_party/terraform/services/container/resource_container_node_pool_test.go.tmpl
wyardley: side note: if you want to still test the update path in a test, instead of checking for no deletion in the plancheck, you can just not check whether anything's deleted, or you can do something like this instead:
```
						plancheck.ExpectResourceAction("xxx.yyy", plancheck.ResourceActionDestroyBeforeCreate),
					},
				},
```

rileykarson (dismissed):

rileykarson:
Running tests before merging, but I don't anticipate any surprises.

rileykarson:
Ah, we're waiting for field availability in the client. Dismissing review until that's ready.

mmv1/third_party/terraform/services/container/node_config.go.tmpl
rileykarson: That change seems not to have stuck! I still see it in the ValidateFunc and docs.

halimsam: Apologies, there shouldn't be any references to LOCAL_SSD_ENCRYPTION_MODE_UNSPECIFIED anymore

rileykarson (approved):

PR #12251, opened 2024-11-05
add new resource `google_iam_organizations_policy_binding`

derekchu-google:
addresses hashicorp/terraform-provider-google/issues/20198

```release-note:new-resource
`google_iam_organizations_policy_binding` (beta)
```


mmv1/products/iam3/OrganizationsPolicyBinding.yaml
```diff
@@ -0,0 +1,194 @@
...
+        description: |
+          Required. Immutable. The resource name of the policy to be bound.
+          The binding parent and policy must belong to the same Organization (or Project).
+        immutable: true
+      - name: 'resource'
+        type: String
+        description: |
+          Output only. The globally unique ID of the policy to be bound.
+        immutable: true
+        output: true
```
SarahFrench, line 123: 👋🏻  Hi again!

Your PR generally looks good, I just have one question

[I don't see `resource` as a field that can be set in `target`](https://cloud.google.com/iam/docs/reference/rest/v3beta/folders.locations.policyBindings#PolicyBinding.Target) and the acceptance tests only set target.principal_set, never target.resource. If the field does exist (I assume the public docs are just a bit out of date?) could you please add an acceptance test that uses that field?

derekchu-google: it does exist but it's hidden for future features. i'll remove it.

SarahFrench (approved):

PR #12252, opened 2024-11-05
Resolving test failure: adding the pre-update scope in the updated sc…

sahsagar-google:
…ript as well since it cannot be deleted untill the binding dependency on it is resolved

Fixes https://github.com/hashicorp/terraform-provider-google/issues/18007

```release-note:none
gkehub: fixed failures for TestAccGKEHub2MembershipBinding_gkehubMembershipBindingBasicExample
```


melinath (approved):
Yay, thanks! I would not have known this was the issue!

PR #12255, opened 2024-11-06
Fix plan for google_identity_platform_config always showing a change for phone_number.enabled=false and email.enabled=false

pashanka:
Fixes issue https://github.com/hashicorp/terraform-provider-google/issues/15043

```release-note: bug
identityplatform: Fixed perma-diff in `google_identity_platform_config`
```


mmv1/templates/terraform/custom_flatten/identity_platform_config_phone_number.go.tmpl
rainshen49, line 1: 1. I notice this is different from https://github.com/rainshen49/magic-modules/blame/main/mmv1/templates/terraform/custom_flatten/identity_platform_config_anonymous.go.tmpl#L13, in particular, if the entire value is nil. Is there a motivation for why?
2. Pls figure out indentation situation. Most likely there is a mix of tabs and spaces.

rainshen49:
Do you want to improve the code quality by adding a test? Currently, none of the tests cover this scenario, which is why the bug occurred.

```terraform
anonymous {
        enabled = false
    }
    email {
        enabled = false
    }
    phone_number {
        enabled = false
    }
```

I can walk you through how to create a new test, but it's basically making a new `example` https://github.com/rainshen49/magic-modules/blob/main/mmv1/products/identityplatform/Config.yaml#L47

rainshen49 (changes requested):

mmv1/templates/terraform/custom_flatten/identity_platform_config_phone_number.go.tmpl
pashanka: 1. Yes this is slightly different because, I observed that in case of anonymous field the response comes as anonymous: {}, where as in case of phone_number/email v itself seems to be nil.
2. Fixed indentation.

pashanka:
> Do you want to improve the code quality by adding a test? Currently, none of the tests cover this scenario, which is why the bug occurred.
> 
> ```hcl
> anonymous {
>     enabled = false
> }
> email {
>     enabled = false
> }
> phone_number {
>     enabled = false
> }
> ```
> 
> I can walk you through how to create a new test, but it's basically making a new `example` https://github.com/rainshen49/magic-modules/blob/main/mmv1/products/identityplatform/Config.yaml#L47

Good point, thanks. Added test and identified email also has similar issue. Fixed that as well.

rainshen49 (approved):

SirGitsalot (approved):

PR #12256, opened 2024-11-06
Add IDPF support for the google_compute_instance and google_compute_image.guest_os_features

Asiderr:
closes: https://github.com/hashicorp/terraform-provider-google/issues/19671

This patch adds `IDPF` as an possible value of the `nic_type` in the `google_compute_instance` and adds `IDPF` tag to the list of `guestOsFeatures` in the google_compute_image.

```release-note:enhancement
compute: Added the `IDPF` tag to the list of `guestOsFeatures`
```

```release-note:enhancement
compute: Added `IDPF` as a possible `nic_type` for the `google_compute_instance.network_interface`
```

### DIFF
--- a/mmv1/products/compute/Image.yaml
+++ b/mmv1/products/compute/Image.yaml
@@ -146,6 +146,7 @@ properties:
             - 'VIRTIO_SCSI_MULTIQUEUE'
             - 'WINDOWS'
             - 'GVNIC'
+            - 'IDPF'
             - 'SEV_LIVE_MIGRATABLE'
             - 'SEV_SNP_CAPABLE'
             - 'SUSPEND_RESUME_COMPATIBLE'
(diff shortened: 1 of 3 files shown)


mmv1/third_party/terraform/services/compute/resource_compute_instance_test.go.tmpl
```diff
@@ -1857,6 +1857,13 @@ func TestAccComputeInstance_nictype_update(t *testing.T) {
 						t, "google_compute_instance.foobar", &instance),
 				),
 			},
+			{
+				Config: testAccComputeInstance_nictype(instanceName, "IDPF"),
```
SirGitsalot, line 1861: The test is failing because this config uses an `e2-medium` instance and `IDPF` is only supported on bare metal instances, the smallest of which is `c3-standard-192-metal`. Since this is just an enum value rather than a new field I'm OK with reverting the test addition - spinning up an instance of this size regularly is likely to cause us quota headaches in the future.

SirGitsalot (commented):
Looks good, just need to revert the test change. Thanks!

mmv1/third_party/terraform/services/compute/resource_compute_instance_test.go.tmpl
Asiderr: Fully agree, I reverted the test and I rebased changes on top of the current main.

SirGitsalot (approved):

PR #12259, opened 2024-11-06
privateca: update certificate authority samples with more realistic values

hoexter:
Some of the properties configured here are either wrong or at least not very sensible on root certificates / subordinates.
A similar set of fixes got applied to terraform documentation samples in https://github.com/terraform-google-modules/terraform-docs-samples/pull/631

What is left out here is placing the subordinate CA in it's own pool. I could not figure out what is required here to prepare the test infrastructure to provision another pool dedicated to the sub-ca. Technically those two certificates should not reside in the same pool.

```release-note:none
```

### DIFF
--- a/mmv1/templates/terraform/examples/privateca_certificate_authority_basic.tf.tmpl
+++ b/mmv1/templates/terraform/examples/privateca_certificate_authority_basic.tf.tmpl
@@ -8,40 +8,28 @@ resource "google_privateca_certificate_authority" "{{$.PrimaryResourceId}}" {
   config {
     subject_config {
       subject {
-        organization = "HashiCorp"
+        organization = "ACME"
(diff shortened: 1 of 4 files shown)


melinath (changes requested):
Part of the reason these samples are structured in this way is to ensure that every field is covered in a [create test](https://googlecloudplatform.github.io/magic-modules/develop/test/test/#add-a-create-test) - this is important to make sure the resources work as expected. Given the number of fields removed here, I'm guessing the coverage is no longer complete.

You can use `exclude_docs` on a complete (but impractical) example to have it be used only for test generation, and have the examples being used in docs be less complete.

melinath (approved):
Never mind, the fields are all still covered!

gfxcc:
LGTM

PR #12260, opened 2024-11-06
networkservices: increased `path_matcher` and `host_rule` limits

wyardley:
Increased `path_matcher` and `host_rule limits` from 10 to 50 within `google_network_services_edge_cache_service`

Fixes hashicorp/terraform-provider-google#19780
Fixes hashicorp/terraform-provider-google#20197

```release-note:bug
networkservices: increased `path_matcher` and `host_rule limits` from 10 to 50 within `google_network_services_edge_cache_service`
```

### DIFF
--- a/mmv1/products/networkservices/EdgeCacheService.yaml
+++ b/mmv1/products/networkservices/EdgeCacheService.yaml
@@ -191,7 +191,7 @@ properties:
                 The name of the pathMatcher associated with this hostRule.
               required: true
         min_size: 1
-        max_size: 10
+        max_size: 50
       - name: 'path_matcher'
         type: Array
         description: |
@@ -891,7 +891,7 @@ properties:
               min_size: 1
               max_size: 200
         min_size: 1
-        max_size: 10
+        max_size: 50
   - name: 'logConfig'
     type: NestedObject
     description: |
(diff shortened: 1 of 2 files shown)


wyardley:
I can't test this locally because I'm running into: https://github.com/hashicorp/terraform-provider-google/issues/14903#issuecomment-2414853418 (and even a similar "method not found issue via the CLI").

If wanted, I can update the existing test with dynamic blocks and `for_each` to create 49 or 50 host rules / path matchers, but the plan succeeds, at least, with these changes.

mmv1/products/networkservices/EdgeCacheService.yaml
```diff
@@ -191,7 +191,7 @@ properties:
                 The name of the pathMatcher associated with this hostRule.
               required: true
         min_size: 1
-        max_size: 10
+        max_size: 50
```
wyardley, line 194: These _look_ like they're in the wrong spot in the YAML config, but from the diff and my functional testing, I believe they're in the right spot. Let me know if there's a way to move or indent them so it's more obvious which attributes they line up with.

BBBmau (approved):
This all looks good. Thanks!

PR #12261, opened 2024-11-06
container: removed `master_ipv4_cidr_block` requirement for private clusters

wyardley:
Remove `master_ipv4_cidr_block` from `privateClusterConfigKeys`. It seems safe to assume that on older cluster versions which don't support this, there will be an API level error anyway.

Not sure if this should be a bug or an enhancement; put it as a bug for now.

Tested locally with built provider. I also repurposed an existing test that seems to be completely unused, which tested a different, but similar, configuration.

Fixes hashicorp/terraform-provider-google#20191

```release-note:bug
container: removed `private_cluster_config.master_ipv4_cidr_block` requirement for private clusters (no longer required since v1.29.x)
```

### DIFF
--- a/mmv1/third_party/terraform/services/container/resource_container_cluster_test.go.tmpl
+++ b/mmv1/third_party/terraform/services/container/resource_container_cluster_test.go.tmpl
@@ -1410,6 +1410,30 @@ func TestAccContainerCluster_withPrivateClusterConfigBasic(t *testing.T) {
 	})
 }
 
+func TestAccContainerCluster_withPrivateClusterConfigNoCidrBlock(t *testing.T) {
+	t.Parallel()
+
+	clusterName := fmt.Sprintf("tf-test-cluster-%s", acctest.RandString(t, 10))
+	containerNetName := fmt.Sprintf("tf-test-container-net-%s", acctest.RandString(t, 10))
+
+	acctest.VcrTest(t, resource.TestCase{
+		PreCheck:	  func() { acctest.AccTestPreCheck(t) },
+		ProtoV5ProviderFactories: acctest.ProtoV5ProviderFactories(t),
+		CheckDestroy: testAccCheckContainerClusterDestroyProducer(t),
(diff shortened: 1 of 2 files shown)


mmv1/third_party/terraform/services/container/resource_container_cluster_test.go.tmpl
```diff
@@ -1410,6 +1410,30 @@ func TestAccContainerCluster_withPrivateClusterConfigBasic(t *testing.T) {
...
+	clusterName := fmt.Sprintf("tf-test-cluster-%s", acctest.RandString(t, 10))
+	containerNetName := fmt.Sprintf("tf-test-container-net-%s", acctest.RandString(t, 10))
+
+	acctest.VcrTest(t, resource.TestCase{
+		PreCheck:	  func() { acctest.AccTestPreCheck(t) },
+		ProtoV5ProviderFactories: acctest.ProtoV5ProviderFactories(t),
+		CheckDestroy: testAccCheckContainerClusterDestroyProducer(t),
+		Steps: []resource.TestStep{
+			{
+				Config: testAccContainerCluster_withPrivateClusterConfigNoCidrBlock(containerNetName, clusterName, false),
```
wyardley, line 1425: odd thing is that, so far, this test (which I think should fail) is passing for me locally even without the code change, even though I can reproduce the issue / fix with a very similar config.

For example, this gives me an error on apply (though not on plan):
```tf
resource "google_container_cluster" "test_private" {
  name                = "test-private"
  location            = "us-central1-a"
  initial_node_count  = 1

  private_cluster_config {
    enable_private_nodes = true
  }
  node_config {
    oauth_scopes = [
      "https://www.googleapis.com/auth/cloud-platform",
    ]
  }

  deletion_protection = false
}
```

wyardley:
Ok, this is the wrong fix, also, I can't reproduce the issue with latest git trunk, though I also can't find the fix yet.

PR #12267, opened 2024-11-06
Fix identity type comparison for service perimeters

coder-221:
Fixes hashicorp/terraform-provider-google/issues/17023. The API does not return the `IDENTITY_TYPE_UNSPECIFIED` enum value when it is set so we can treat that as an empty value.

```release-note: bug
accesscontextmanager: fixed comparison of `identity_type` in `ingress_from` and `egress_from` when the `IDENTITY_TYPE_UNSPECIFIED` is set
```


slevenick (approved):
This is a good demonstration on why we shouldn't allow the `_UNSPECIFIED` enum value to be specified in Terraform!

PR #12140, opened 2024-10-25
`ephemeral`: add `ephemeral_google_service_account_token`

BBBmau:
**Release Note Template for Downstream PRs (will be copied)**

See [Write release notes](https://googlecloudplatform.github.io/magic-modules/contribute/release-notes/) for guidance.

```release-note:REPLACEME

```

PR #12268, opened 2024-11-07
move `framework_validators.go` to new `fwvalidators` package

BBBmau:
This PR would make it possible to add validators being used by EphemeralResources.

When attempting to move validators to `framework_provider.go` in `fwprovider` we end up getting an `import cycle` because of `fwprovider` importing `resourcemanager` in order to support ephermeral resources and data sources under plugin-framework.

You can view how `service_account_token` ephemeral would look with this being merged [here](https://github.com/BBBmau/magic-modules/pull/24)

The validators can be seen being used in this resource where they are added in the same reosurce file itself instead of being placed in the validator file. https://github.com/GoogleCloudPlatform/magic-modules/pull/12140

When we merg this we can perform the following:

- merge https://github.com/BBBmau/magic-modules/pull/24 into `service_account_token` resource supprt
-  merge https://github.com/GoogleCloudPlatform/magic-modules/pull/12140 - tests will be passing as i've ran locally with the `fwvalidators` package. This would also allow other resources to add the necessary validators into `fwvalidators` package. 

```release-note:none

```


SarahFrench (changes requested):
Please update how we reference the validators in the fwprovider package now that they've been moved out of there; currently there are build errors:


> google-beta/fwprovider/framework_provider.go:80:6: undefined: CredentialsValidator

mmv1/third_party/terraform/fwvalidators/framework_validators.go
```diff
@@ -0,0 +1,285 @@
+// Copyright (c) HashiCorp, Inc.
+// SPDX-License-Identifier: MPL-2.0
```
SarahFrench, line 2:
```suggestion
```

BBBmau:
pushed the validator tests on top of your changes @SarahFrench just a heads-up!

mmv1/third_party/terraform/fwvalidators/framework_validators.go
```diff
@@ -0,0 +1,288 @@
...
+	return v.Description(ctx)
+}
+
+// ValidateString performs the validation
+func (v ServiceScopeValidator) ValidateString(ctx context.Context, req validator.StringRequest, resp *validator.StringResponse) {
+	if req.ConfigValue.IsNull() || req.ConfigValue.IsUnknown() {
+		return
+	}
+
+	canonicalized := CanonicalizeServiceScope(req.ConfigValue.ValueString())
```
SarahFrench, line 240: [CanonicalizeServiceScope from the tpgresource package](https://github.com/GoogleCloudPlatform/magic-modules/blob/3db026ca076c34e112178ab9c5e5a6a7fb251e8b/mmv1/third_party/terraform/tpgresource/service_scope.go#L5) is a helper to convert shorthand strings into the long version of the scope's name that is API-compatible. In the google_service_account_access_token data source it is [used to change the value stored in state](https://github.com/hashicorp/terraform-provider-google/blob/45df35f12441f4d6749a76f35d26fd89647aa941/google/services/resourcemanager/data_source_google_service_account_access_token.go#L39-L41), not validate the input from users.

`ServiceScopeValidator` currently only flags strings as invalid if they are represented in CanonicalizeServiceScope's map, so I think the validator creates a false sense of security. E.g. "foobar" is found to be valid but "bigquery" isn't.

If we want to enforce that all the inputs are a canonical/long-form scope name then I think we should use something like a regex to check the format of the input strings. We could include the use of CanonicalizeServiceScope to allow the convenience of users supplying short names for some scopes, but I think the output of CanonicalizeServiceScope should be checked using the regex after. [I've asked in the shared chat if there are any gotchas using regex on scopes.
](https://chat.google.com/room/AAAA49vo02s/Cvatj3mIFfM/Cvatj3mIFfM?cls=10)

An alternative, instead of updating the validator, is that we could skip implementing a validator for scopes because scopes have not been validated in the provider previously. What do you think?

mmv1/third_party/terraform/fwvalidators/framework_validators.go
```diff
@@ -0,0 +1,288 @@
...
+
+// Define the possible service account name patterns
+var serviceAccountNamePatterns = []string{
+	`^.+@.+\.iam\.gserviceaccount\.com$`,                     // Standard IAM service account
+	`^.+@developer\.gserviceaccount\.com$`,                   // Legacy developer service account
+	`^.+@appspot\.gserviceaccount\.com$`,                     // App Engine service account
+	`^.+@cloudservices\.gserviceaccount\.com$`,               // Google Cloud services service account
+	`^.+@cloudbuild\.gserviceaccount\.com$`,                  // Cloud Build service account
+	`^service-[0-9]+@.+-compute\.iam\.gserviceaccount\.com$`, // Compute Engine service account
+}
```
SarahFrench, line 138: Please ensure there is a test case covering each of these formats

mmv1/third_party/terraform/fwvalidators/framework_validators.go
SarahFrench: From the discussion I linked it looks like [validating scopes isn't possible](https://chat.google.com/room/AAAA49vo02s/Cvatj3mIFfM/Cvatj3mIFfM?cls=10) in the provider, and users would need to receive API errors for feedback.

I think we cannot make a validator for scopes, but we can use CanonicalizeServiceScope in any ephemeral resources that handle scopes. Due to this, I think we should remove ServiceScopeValidator. What do you think?

mmv1/third_party/terraform/fwvalidators/framework_validators.go
```diff
@@ -0,0 +1,288 @@
...
+			req.Path,
+			"Invalid Service Account Name",
+			"Service account name must match one of the expected patterns for Google service accounts",
+		)
+	}
+}
+
+// Create a custom validator for duration
+type DurationValidator struct {
+}
```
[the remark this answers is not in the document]
BBBmau: This makes sense since in the original data source of this ephemeral this comment is left: `ValidateFunc: verify.ValidateDuration(), // duration <=3600s; TODO: support validateDuration(min,max)`

Latest commit adds this.

mmv1/third_party/terraform/fwvalidators/framework_validators.go
BBBmau: recent commit adds testcases for all patterns here.

mmv1/third_party/terraform/fwvalidators/framework_validators.go
BBBmau: > From the discussion I linked it looks like [validating scopes isn't possible](https://chat.google.com/room/AAAA49vo02s/Cvatj3mIFfM/Cvatj3mIFfM?cls=10) in the provider, and users would need to receive API errors for feedback.
> 
> I think we cannot make a validator for scopes, but we can use CanonicalizeServiceScope in any ephemeral resources that handle scopes. Due to this, I think we should remove ServiceScopeValidator. What do you think?


Yeah it's unfortunate that some scopes don't expand to a uri. In this case we can just remove the ServiceScopeValidator to let the API handle it by returning an error

mmv1/third_party/terraform/fwvalidators/framework_validators.go
```diff
@@ -115,3 +117,109 @@ func (v nonEmptyStringValidator) ValidateString(ctx context.Context, request val
...
+	if err != nil {
+		resp.Diagnostics.AddAttributeError(
+			req.Path,
+			"Invalid Duration Format",
+			"Duration must be a valid duration string (e.g., '3600s', '1h')",
+		)
+		return
+	}
+
+	if v.MinDuration != 0 && duration < v.MinDuration {
```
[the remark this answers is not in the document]
SarahFrench: It might be better to write this logic like:

```go
if duration < v.MinDuration || duration > v.MaxDuration {
    // provide error message saying that the value should be between v.MinDuration and v.MaxDuration
}
```

This would allow provider developers to set values to 0 if they want, and if some bad min/max values were accidentally set the error message would be more useful for debugging (e.g. identifying if the min is 0 and the max is 0 because only zero values are available). This would mean instead of letting the user know a value is too big or too small we'd just tell them it's out of range.

mmv1/third_party/terraform/fwvalidators/framework_validators_test.go
```diff
@@ -0,0 +1,296 @@
...
+				t.Errorf("Expected %d warnings, got %d", tc.ExpectedWarningCount, resp.Diagnostics.WarningsCount())
+			}
+			if resp.Diagnostics.ErrorsCount() > tc.ExpectedErrorCount {
+				t.Errorf("Expected %d errors, got %d", tc.ExpectedErrorCount, resp.Diagnostics.ErrorsCount())
+			}
+		})
+	}
+}
+
+func TestServiceAccountNameValidator(t *testing.T) {
```
SarahFrench, line 71: Looks good, thanks for adding the extra test cases!

mmv1/third_party/terraform/fwvalidators/framework_validators.go
```diff
@@ -115,3 +117,109 @@ func (v nonEmptyStringValidator) ValidateString(ctx context.Context, request val
...
+		resp.Diagnostics.AddAttributeError(
+			req.Path,
+			"Invalid Service Account Name",
+			"Service account name must match one of the expected patterns for Google service accounts",
+		)
+	}
+}
+
+// Create a custom validator for duration
+type DurationValidator struct {
```
SarahFrench, line 181: I'm wondering if this validator can be named more specifically. For example there's a `NonNegativeDurationValidator` in the code base and it's more clear what it does from the name versus `DurationValidator`. Can you think of a more descriptive name that matches what this new validator does?

mmv1/third_party/terraform/fwvalidators/framework_validators.go
```diff
@@ -115,3 +117,109 @@ func (v nonEmptyStringValidator) ValidateString(ctx context.Context, request val
...
+	`^.+@.+\.iam\.gserviceaccount\.com$`,                     // Standard IAM service account
+	`^.+@developer\.gserviceaccount\.com$`,                   // Legacy developer service account
+	`^.+@appspot\.gserviceaccount\.com$`,                     // App Engine service account
+	`^.+@cloudservices\.gserviceaccount\.com$`,               // Google Cloud services service account
+	`^.+@cloudbuild\.gserviceaccount\.com$`,                  // Cloud Build service account
+	`^service-[0-9]+@.+-compute\.iam\.gserviceaccount\.com$`, // Compute Engine service account
+}
+
+// Create a custom validator for service account names
+type ServiceAccountNameValidator struct{}
```
SarahFrench, line 141: nit: isn't this validator validating emails?

mmv1/third_party/terraform/fwvalidators/framework_validators.go
```diff
@@ -115,3 +117,109 @@ func (v nonEmptyStringValidator) ValidateString(ctx context.Context, request val
...
 }
+
+func StringSet(d basetypes.SetValue) []string {
+
+	StringSlice := make([]string, 0)
+	for _, v := range d.Elements() {
+		StringSlice = append(StringSlice, v.(basetypes.StringValue).ValueString())
+	}
+	return StringSlice
+}
```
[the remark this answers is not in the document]
BBBmau: this is used across multiple ephemeral resources, so we'll want to create a utils file for this. I'll remove it and create a separate PR that does this.

mmv1/third_party/terraform/fwvalidators/framework_validators.go
BBBmau: It make sense to include bound in the name something like `BoundedDuration`.

I don't think we need to include `NonNegative` if we're specifying that this Duration is Bounded based on what MinDuration and MaxDuration are set to.

mmv1/third_party/terraform/fwvalidators/framework_validators.go
BBBmau: changed name to `ServiceAccountEmailValidator

mmv1/third_party/terraform/fwvalidators/framework_validators.go
```diff
@@ -52,21 +51,21 @@ func CredentialsValidator() validator.String {
 }
 
 // Non Negative Duration Validator
-type nonnegativedurationValidator struct {
+type nonnegativeBoundedDuration struct {
 }
```
SarahFrench, line 55: Please revert the changes to this validator - it doesn't make assertions about the range/bounds a duration should be in so this name is misleading. My feedback previously was about your validator and showing how this validator's name indicated how it validated the duration.

BBBmau: reverted.

mmv1/third_party/terraform/fwvalidators/framework_validators.go
SarahFrench: Thanks for updating from `DurationValidator` to `BoundedDuration` - that gives more context about what the validator does.

>I don't think we need to include NonNegative if we're specifying that this Duration is Bounded based on what MinDuration and MaxDuration are set to.

Agreed  - `NonNegativeDurationValidator` is validator that's already in the code base that's an example of a more descriptive name. Sorry I wasn't clear

mmv1/third_party/terraform/fwvalidators/framework_validators_test.go
```diff
@@ -0,0 +1,309 @@
...
+			maxDuration:   time.Hour,
+			expectError:   true,
+			errorContains: "Duration Too Long",
+		},
+		"duration exceeds max - minutes": {
+			value:         types.StringValue("120m"),
+			minDuration:   30 * time.Minute,
+			maxDuration:   time.Hour,
+			expectError:   true,
+			errorContains: "Duration Too Long",
```
SarahFrench, line 219:
```suggestion
			errorContains: "Invalid Duration",
```

mmv1/third_party/terraform/fwvalidators/framework_validators_test.go
```diff
@@ -0,0 +1,309 @@
...
+			maxDuration:   time.Hour,
+			expectError:   true,
+			errorContains: "Duration Too Short",
+		},
+		"duration exceeds max - seconds": {
+			value:         types.StringValue("7200s"),
+			minDuration:   30 * time.Minute,
+			maxDuration:   time.Hour,
+			expectError:   true,
+			errorContains: "Duration Too Long",
```
SarahFrench, line 212:
```suggestion
			errorContains: "Invalid Duration",
```

mmv1/third_party/terraform/fwvalidators/framework_validators_test.go
```diff
@@ -0,0 +1,309 @@
...
+			minDuration: 30 * time.Minute,
+			maxDuration: 2 * time.Hour,
+			expectError: false,
+		},
+		"duration below min": {
+			value:         types.StringValue("900s"),
+			minDuration:   30 * time.Minute,
+			maxDuration:   time.Hour,
+			expectError:   true,
+			errorContains: "Duration Too Short",
```
SarahFrench, line 205:
```suggestion
			errorContains: "Invalid Duration",
```

mmv1/third_party/terraform/fwvalidators/framework_validators_test.go
```diff
@@ -0,0 +1,309 @@
...
+			maxDuration:   time.Hour,
+			expectError:   true,
+			errorContains: "Duration Too Long",
+		},
+		"duration exceeds max - hours": {
+			value:         types.StringValue("2h"),
+			minDuration:   30 * time.Minute,
+			maxDuration:   time.Hour,
+			expectError:   true,
+			errorContains: "Duration Too Long",
```
SarahFrench, line 226:
```suggestion
			errorContains: "Invalid Duration",
```

SarahFrench (approved):
LGTM

PR #12273, opened 2024-11-07
support skip_test option in yaml file

iyabchen:
https://github.com/hashicorp/terraform-provider-google/issues/13623

Support a skip_test option in the yaml file, similar to skip_vcr. 
If skip_test var is not an empty string, then the test is generated with a t.Skip("xxx"). 

```release-note:none

```

### DIFF
--- a/mmv1/api/resource/examples.go
+++ b/mmv1/api/resource/examples.go
@@ -149,7 +149,7 @@ type Examples struct {
 	// Or a config with two fine grained resources that have a race condition during create
 	SkipVcr bool `yaml:"skip_vcr,omitempty"`
 
-	// The reason to skip a test. For example, a link to a ticket explaining the issue that needs to be resolved before 
+	// The reason to skip a test. For example, a link to a ticket explaining the issue that needs to be resolved before
 	// unskipping the test. If this is not empty, the test will be skipped.
 	SkipTest string `yaml:"skip_test,omitempty"`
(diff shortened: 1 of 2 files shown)


ScottSuarez:
Reassigning to @melinath to ensure this is the way to go. Not sure if there are additional documentation needed around this.

mmv1/api/resource/examples.go
```diff
@@ -149,6 +149,9 @@ type Examples struct {
 	// Or a config with two fine grained resources that have a race condition during create
 	SkipVcr bool `yaml:"skip_vcr,omitempty"`
 
+	// The reason to skip a test. Only when this is not empty then the test is skipped.
```
melinath, line 152:
```suggestion
	// The reason to skip a test. For example, a link to a ticket explaining the issue that needs to be resolved before 
	// unskipping the test. If this is not empty, the test will be skipped.
```

melinath (approved):
This seems reasonable to me. I don't think we have an obvious place for docs about this to live at the moment but we can add that separately / later.

cc @shuyama1

melinath (changes requested):
looks like gofmt is failing

iyabchen:
> looks like gofmt is failing

Done, seems some trailing whitespace on the comment line.

melinath (approved):

PR #12275, opened 2024-11-07
Added new resource "WasmPlugin"

matheusaleixo-cit:
Adds new resource `google_network_services_wasm_plugin`.

Also updates `google_network_services_lb_route_extension` and `google_network_services_lb_traffic_extension` to document  the support for this in the `extensions.service` field description, as well as related tests.

Fixes: https://github.com/hashicorp/terraform-provider-google/issues/20220

```release-note:new-resource
`google_network_services_wasm_plugin`
```


mmv1/products/networkservices/WasmPlugin.yaml
```diff
@@ -0,0 +1,189 @@
...
+    description: |
+      Optional. A human-readable description of the resource.
+  - name: 'labels'
+    type: KeyValueLabels
+    description: 'Optional. Set of labels associated with the WasmPlugin resource.'
+  - name: 'mainVersionId'
+    type: ResourceRef
+    resource: 'WasmPluginVersion'
+    description: |
+      Optional. The ID of the WasmPluginVersion resource that is the currently serving one. The version referred to must be a child of this WasmPlugin resource and should be listed in the "versions" field.
```
SirGitsalot, line 84: Is there a `google_network_services_wasm_plugin_version` resource on its way?

matheusaleixo-cit: I am currently trying to implement a version resource but I am having some issues since `main_version_id` appears to be required for creation (despite the [API docs](https://cloud.google.com/service-extensions/docs/reference/rest/v1alpha1/projects.locations.wasmPlugins#:~:text=count%22%3A%20%223%22%20%7D.-,mainVersionId,-string) describing it as optional) and needs to refer to a version listed in this plugin resource during request. Trying to list a initial version in the plugin resource and config additional versions as resources inevitably causes permadiffs.
  
Either way, I think this field would be better off as a String since it may not necessarily be used with a future versions resource.

SirGitsalot: I just checked the API proto definition, and `main_version_id` is annotated with both `(google.api.field_behavior) = OPTIONAL` and a `validator.rule` that makes it required (and the latter wins).

The main reason I asked about `google_network_services_wasm_plugin_version` is if that resource can manage versions and so can a writeable `versions` field in `google_network_services_wasm_plugin` then there's no single canonical place to configure versions, which doesn't play well with Terraform.

Thinking about it a little more, I'm not sure that `google_network_services_wasm_plugin_version` would be possible, at least not without a lot of custom code or a sketchy user experience. Since the parent `google_network_services_wasm_plugin` needs to exist before creating a `google_network_services_wasm_plugin_version` child, the configuration sequence would be:

1. Configure a `google_network_services_wasm_plugin` with an empty `main_version_id` (assuming that the `validator.rule` is incorrect and can be removed from the API)
2. Create a `google_network_services_wasm_plugin_version`
3. Update the plugin from step 1 to set `main_version_id` to the version from step 2.

As you've discovered, step 3 is what makes things a PITA. There's a couple of ways to work around the problem:

- The easiest is not to implement `google_network_services_wasm_plugin_version` and make the `versions` field of `google_network_services_wasm_plugin` required instead.
- If `main_version_id` really is optional, remove the `validator.rule` (and wait for the subsequent API rollout). Then, implement `google_network_services_wasm_plugin_version` with an additional virtual boolean field, something like `main_version`. When a `google_network_services_wasm_plugin_version` is configured with `main_version = true` a custom post-create or post-update step would `PATCH` the parent plugin resource to set `main_version` to the version in question.

With the latter, `main_version_id` and `versions` in `google_network_services_wasm_plugin` can be output only or omitted entirely. You'll need custom encoder for `main_version_id` and `versions` for updates so that they don't get zeroed out when updating the other fields in the resource.

I think the separate version resource option is a more idiomatically Terraform approach and a better user experience, but it will be a fair amount of additional work to create and maintain. I'll leave it up to you which you prefer.

mmv1/products/networkservices/product.yaml
```diff
@@ -16,7 +16,7 @@ name: 'NetworkServices'
 display_name: 'Network services'
 versions:
   - name: 'beta'
-    base_url: 'https://networkservices.googleapis.com/v1/'
+    base_url: 'https://networkservices.googleapis.com/v1beta1/'
```
SirGitsalot, line 19: This change is what's causing the new test failures. I think (based on looking at your API service config) that the wasm endpoints are being served on `/v1/` so you should be able to revert.

krz-filip: /v1/ may not be available until Q1 2025. We need to stay with `v1beta1`.

mmv1/products/networkservices/WasmPlugin.yaml
matheusaleixo-cit: `main_version_id` being required seems to be the intended behavior for the API, with versions being managed together with the WasmPlugin), so I will be keeping the versions implementation on the google_network_services_wasm_plugin resource.

mmv1/products/networkservices/product.yaml
SirGitsalot: The [discovery doc](https://networkservices.googleapis.com/$discovery/rest?version=v1) says that the wasm plugin endpoints are available under `/v1/` so it might be worth trying. If it works, it'll save you from having to fix all the tests that are broken when connecting to `/v1beta1/`

matheusaleixo-cit: Running the v1 endpoints for wasmPlugins causes a 403 error: "WasmPlugins v1 api is not yet available. Please use the v1beta1 API." so I guess there is no way around that.

Most of the broken tests are related to NetworkServicesEdgeCache resources, and the docs lists only [v1](https://cloud.google.com/media-cdn/docs/reference/rest/v1/projects.locations.edgeCacheOrigins) and [v1alpha1](https://cloud.google.com/media-cdn/docs/reference/rest/v1alpha1/projects.locations.edgeCacheOrigins) endpoints for these, so maybe we could use v1alpha1 endpoints for the beta provider?

SirGitsalot: It's OK to use a higher-versioned endpoint (like using GA for beta as has been used here) but going the other way is not. This is the first time I've encountered a split where one resource is available in GA but not beta, while another resource is available in beta but not GA. I'll talk to the team and see if there's a precedent for how to handle the situation.

krz-filip: > Most of the broken tests are related to `NetworkServicesEdgeCache*` resources, and the docs lists only [`v1`](https://cloud.google.com/media-cdn/docs/reference/rest/v1/projects.locations.edgeCacheOrigins) and [`v1alpha1`](https://cloud.google.com/media-cdn/docs/reference/rest/v1alpha1/projects.locations.edgeCacheOrigins) endpoints for these, so maybe we could use `v1alpha1` endpoints for the beta provider?
I can't image how NetworkServicesEdgeCache tests may be impacted, as it would require WasmAction resource (a connection between WasmPlugin and EdgeCacheService) which stays in v1alpha and is not a part of the current task (you have access to WasmAction API because your test projects are allowlisted). 

At this moment WasmPlugin is intented to be used with LbTrafficExtension and LbRouteExtensions only.

SirGitsalot: > I can't image how NetworkServicesEdgeCache tests may be impacted

Those tests are unchanged, but Magic Modules only supports a single base URL per version (beta or GA) and this PR changes the base URL for the beta version from `/v1/` to `/v1beta/`.

@matheusaleixo-cit after discussing this with the rest of the team the solution is to move this new resource to a new product (say, `networkservicesplugins`). That should be a quick change:

- Create the new `mmv1/products/networkservicesplugins` directory
- Move `WasmPlugin.yaml` to it
- Copy  `mmv1/products/networkservices/product.yaml` to it, with the beta base url set to `/v1beta`
- Revert the change to `mmv1/products/networkservices/product.yaml`

SirGitsalot (changes requested):
Making GitHub happy here since with my last comment I didn't click the "request changes" button

SirGitsalot (commented):
Looks good, just waiting on the test run to finish successfully now.

@krz-filip PTAL

mmv1/products/networkservicesplugins/WasmPlugin.yaml
krz-filip, line 1: I see that all tests defines the single version configuration. Could you please extend at least one of them to define at least 2 versions? The perfect would updating "update" test to modify WasmPlugin from having v1(main)+v2 versions into v2(main)+v3?

mmv1/third_party/terraform/.teamcity/components/inputs/services_ga.kt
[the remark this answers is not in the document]
SirGitsalot: Looks like the answer is no, as the only product that I can find that exists in beta but not in ga is firebasestorage and it does not have an entry in the ga file.

SirGitsalot (commented):
Kicked off a new test run with the latest changes

mmv1/third_party/terraform/.teamcity/components/inputs/services_ga.kt
```diff
@@ -546,6 +546,11 @@ var ServicesListGa = mapOf(
         "displayName" to "Networkservices",
         "path" to "./google/services/networkservices"
     ),
+    "networkservicesplugins" to mapOf(
```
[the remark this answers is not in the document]
krz-filip: Thanks, I need to discuss it with the product owner (next week). I suppose that he would prefer something like "service_extensions" instead of "network_services_plugins".
AFAIK the main reason of moving it to the new product were failing tests for other network_services resources.
Do you think that there is a chance to make them passing? What should change on networkservices API?

SirGitsalot (commented):
Looks good, just one final question to enable the tests - @krz-filip do you know if any of the examples available in a publicly accessible registry? Some of TPG's other resource tests rely on Docker containers, but we're able to use the service's own example containers in our tests (like Cloud Run has `us-docker.pkg.dev/cloudrun/container/hello`)

mmv1/third_party/terraform/.teamcity/components/inputs/services_ga.kt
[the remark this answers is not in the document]
krz-filip: We prefer `networkservices` product, which means we have to wait to finalize this work until WasmPlugin API is available as `v1`.

krz-filip:
> Looks good, just one final question to enable the tests - @krz-filip do you know if any of the examples available in a publicly accessible registry? Some of TPG's other resource tests rely on Docker containers, but we're able to use the service's own example containers in our tests (like Cloud Run has `us-docker.pkg.dev/cloudrun/container/hello`)

We don't have any public repo :(
If you know what: users, groups, domains, or service accounts are used to run tests I can share/prepare a dedicated repo.

maxi-cit:
Hello folks, I updated this branch with recent changes. Now this should be mergable. Could you run tests please?

mmv1/third_party/terraform/.teamcity/components/inputs/services_ga.kt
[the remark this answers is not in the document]
maxi-cit: Hello there, I am adding this back to the PR since Teamcity checks were failing in lack of it. Also, We are already on Q1 so this should be available soon in anycase.

maxi-cit:
Hello @krz-filip @SirGitsalot, could you run tests please?

matheusaleixo-cit:
I got a confirmation that we should keep the wasm plugins resource in the `networkservices` product config, keeping this on hold until this feature reaches v1 API availability. In that case I will be reverting this new ServiceExtensions product config.

SirGitsalot (commented):
What's the current state of the v1 API? I tested this out locally with a plugin I built from the samples repo and got:

```
Error: Error creating WasmPlugin: googleapi: Error 403: Permission "WasmPlugins v1 api is not yet available. Please use the v1beta1 API." denied on "WasmPlugin"
```

I tried with the `v1beta1` API and it worked (well, mostly worked - there were some issues with the test cleaning up after itself but I may have had dangling resources from previous failed runs).

matheusaleixo-cit:
> What's the current state of the v1 API? I tested this out locally with a plugin I built from the samples repo and got:
> 
> ```
> Error: Error creating WasmPlugin: googleapi: Error 403: Permission "WasmPlugins v1 api is not yet available. Please use the v1beta1 API." denied on "WasmPlugin"
> ```
> 
> I tried with the `v1beta1` API and it worked (well, mostly worked - there were some issues with the test cleaning up after itself but I may have had dangling resources from previous failed runs).

Hi, v1 launch of this feature's API seems to be expected to happen around late June/early July, that's a bit later than I expected when I created this PR.
I that case, should we close this PR and keep a new one in draft while we wait for the v1 launch? Or can we keep this PR open and disable the automatic reminders?

SirGitsalot (commented):
You can drop it into draft mode, which should stop the review reminders (and if it doesn't, I'll add the tag). Thanks!

matheusaleixo-cit:
I'm reopening this PR since WasmPlugins v1 API seems to now be open to the public.

matheusaleixo-cit:
Hello @SirGitsalot , any chance we could give this PR priority now that the API is now available? No changes to the code has been made since last review and this has been in our backlog for quite a while.

mmv1/products/networkservices/WasmPlugin.yaml
```diff
@@ -0,0 +1,194 @@
...
+# limitations under the License.
+
+---
+name: 'WasmPlugin'
+description: |
+  WasmPlugin is a resource representing a service executing a customer-provided Wasm module.
+references:
+  guides:
+    'Configure a route extension': 'https://cloud.google.com/service-extensions/docs/create-plugin'
+  api: 'https://cloud.google.com/service-extensions/docs/reference/rest/v1beta1/projects.locations.wasmPlugins'
```
martijneken: now v1: https://cloud.google.com/service-extensions/docs/reference/rest/v1/projects.locations.wasmPlugins

rileykarson:

mmv1/templates/terraform/custom_expand/wasm_plugin_skip_empty_versions.go.tmpl
```diff
@@ -0,0 +1,72 @@
...
+	if v == nil {
+		return map[string]interface{}{}, nil
+	}
+	m := make(map[string]interface{})
+	for _, raw := range v.(*schema.Set).List() {
+		original := raw.(map[string]interface{})
+		transformed := make(map[string]interface{})
+
+		// Ensure we don't send empty versions
+		if tpgresource.IsEmptyValue(reflect.ValueOf(original["version_name"])) {
```
slevenick, line 11: Why do we not want to send empty versions? Like, can we just require version_name to be specified?

mmv1/third_party/terraform/services/networkservices/resource_network_services_lb_traffic_extension_test.go
```diff
@@ -805,3 +806,374 @@ resource "google_compute_region_backend_service" "callouts_backend_2" {
 }
 `, context)
 }
+
+func TestAccNetworkServicesLbTrafficExtension_crossRegionInternalPluginExtension(t *testing.T) {
+	acctest.SkipIfVcr(t) // Test requires a existing container image that contains the plugin code, published in an Artifact Registry repository.
```
slevenick: Is it possible to add the container image as part of the test? Or how are we supposed to set this up in general?

mmv1/products/networkservices/WasmPlugin.yaml
```diff
@@ -0,0 +1,203 @@
...
+async:
+  actions: ['create', 'delete', 'update']
+  type: 'OpAsync'
+  operation:
+    base_url: '{{op_id}}'
+  result:
+    resource_inside_response: false
+custom_code:
+examples:
+  - name: 'wasm_plugin_basic_docker_repository'
```
slevenick, line 44: Why are these examples skipped for testing? I'm seeing failures when trying to run them

mmv1/templates/terraform/custom_expand/wasm_plugin_skip_empty_versions.go.tmpl
slevenick: Still curious about this

matheusaleixo-cit: Currently `version_name` is already required being the `key_name` for `versions` map, this check is because when removing a existing version from the config, for reasons I can't quite understand the provider sends an empty version with just the `imageDigest` output field in the PATCH request.
This can be reproduced by omitting this `custom_expand` from `versions` and running the `TestAccNetworkServicesWasmPlugin_wasmPluginVersionUpdate` test in the generated provider.

mmv1/third_party/terraform/services/networkservices/resource_network_services_lb_traffic_extension_test.go
matheusaleixo-cit: The container image is set up as described [here](https://cloud.google.com/service-extensions/docs/prepare-plugin-code). The `testAccCheckNetworkServicesWasmPlugin_uploadCompiledCode` testCheckFunc is now responsible in preparing the plugin code for the tests.

(Sorry for the delay,  I wrote a response last week but it seems I accidentally pressed to Start a review button instead of leaving a comment, and my response was only visible to me)

mmv1/third_party/terraform/services/networkservices/resource_network_services_wasm_plugin_test.go
```diff
@@ -0,0 +1,655 @@
...
+		}
+
+		log.Printf("[DEBUG] Finished uploading the compiled file: %#v", res)
+
+		return nil
+	}
+}
+
+// sendRequestRawBodyWithTimeout is derived from sendRequestWithTimeout with direct pass through of request body
+func sendRequestRawBodyWithTimeout(config *transport_tpg.Config, method, project, rawurl, userAgent string, body io.Reader, contentType string, timeout time.Duration, errorRetryPredicates ...transport_tpg.RetryErrorPredicateFunc) (map[string]interface{}, error) {
```
matheusaleixo-cit, line 590: This function is just a straight copy from [apigee_utils.go](https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/third_party/terraform/services/apigee/apigee_utils.go#L54). Should I turn this into a helper function in something shared between resources like in [transport.go](https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/third_party/terraform/transport/transport.go)?

mmv1/products/networkservices/WasmPlugin.yaml
matheusaleixo-cit: I skipped these example tests since they are bound to fail as we are not providing a repository with the required plugin code. These tests could pass if the user already had the required plugin code [manually](https://cloud.google.com/service-extensions/docs/prepare-plugin-code#upload-code) configured in his project.
I don't know if there is a way to run a testCheckFunc for these generated tests in order to upload the plugin code for the generic_repository example, while for the docker_repository example I don't know how to approach this.

mmv1/third_party/terraform/services/networkservices/resource_network_services_wasm_plugin_test.go
matheusaleixo-cit: @SirGitsalot Can I get an update on this particular point?

matheusaleixo-cit:
Hello @slevenick, is the merge for this being blocked by the pending review from @SirGitsalot? Or is the rollout being blocked by something unrelated?

PR #12277, opened 2024-11-07
Add new TPUv2 QueuedResource resource in beta

zeleena:
Add new TPUv2 QueuedResource resource with minimum required fields to google-beta provider.

```release-note:new-resource
`google_tpu_v2_queued_resource` (beta)
```

### DIFF
--- a/mmv1/products/tpuv2/QueuedResource.yaml
+++ b/mmv1/products/tpuv2/QueuedResource.yaml
@@ -1,0 +1,111 @@
+# Copyright 2024 Google Inc.
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+---
+name: 'QueuedResource'
+description: |
+  A Cloud TPU Queued Resource.
+min_version: 'beta'
+references:
+  guides:
+    'Official Documentation': 'https://cloud.google.com/tpu/docs/'
+  api: 'https://cloud.google.com/tpu/docs/reference/rest/v2/projects.locations.queuedResources'
+base_url: 'projects/{{project}}/locations/{{zone}}/queuedResources'
+self_link: 'projects/{{project}}/locations/{{zone}}/queuedResources/{{name}}'
+create_url: 'projects/{{project}}/locations/{{zone}}/queuedResources?queuedResourceId={{name}}'
+update_url: 'projects/{{project}}/locations/{{zone}}/queuedResources/{{name}}'
+update_verb: 'PATCH'
+update_mask: true
+autogen_async: true
+async:
+  actions: ['create', 'delete', 'update']
+  type: 'OpAsync'
+  operation:
+    base_url: '{{op_id}}'
+    path: 'name'
+    wait_ms: 1000
+  result:
+    path: 'response'
+    resource_inside_response: true
+  status:
+    path: 'done'
+    complete: true
+    allowed:
+      - true
+      - false
+  error:
+    path: 'error'
+    message: 'message'
+examples:
+  - name: 'tpu_v2_queued_resource_basic'
+    primary_resource_id: 'qr'
+    min_version: 'beta'
+    vars:
+      qr_name: 'test-qr'
+      tpu_name: 'test-tpu'
+    test_env_vars:
+      project: 'PROJECT_NAME'
+    skip_vcr: true
+parameters:
+  - name: 'zone'
+    type: String
+    description: |
+      The GCP location for the Queued Resource. If it is not provided, the provider zone is used.
+    url_param_only: true
+    immutable: true
+    default_from_api: true
+properties:
+  - name: 'name'
+    type: String
+    description: |
+      The immutable name of the Queued Resource.
+    required: true
+    immutable: true
+    custom_flatten: 'templates/terraform/custom_flatten/name_from_self_link.tmpl'
+  - name: 'tpu'
+    type: NestedObject
+    description: |
+      Defines a TPU resource.
+    properties:
+      - name: 'nodeSpec'
+        type: Array
+        description: |
+          The TPU node(s) being requested.
+        item_type:
+          type: NestedObject
+          properties:
+            - name: 'parent'
+              description: |
+                The parent resource name.
+            - name: 'nodeId'
+              immutable: true
+              description: |
+                Unqualified node identifier used to identify the node in the project once provisioned.
+            - name: 'node'
+              type: NestedObject
+              description: |
+                The node.
+              properties:
+                - name: 'runtimeVersion'
+                  required: true
+                  immutable: true
+                  description: |
+                    Runtime version for the TPU.
+                - name: 'acceleratorType'
+                  immutable: true
+                  default_value: 'v2-8'
+                  description: |
+                    TPU accelerator type for the TPU. If not specified, this defaults to 'v2-8'.
+                - name: 'description'
+                  description: |
+                    Text description of the TPU.
--- a/mmv1/products/tpuv2/QueuedResource.yaml
+++ b/mmv1/products/tpuv2/QueuedResource.yaml
@@ -87,23 +87,22 @@ properties:
             - name: 'parent'
               description: |
                 The parent resource name.
+              required: true
             - name: 'nodeId'
-              immutable: true
               description: |
                 Unqualified node identifier used to identify the node in the project once provisioned.
             - name: 'node'
               type: NestedObject
               description: |
                 The node.
+              required: true
               properties:
                 - name: 'runtimeVersion'
                   required: true
-                  immutable: true
                   description: |
                     Runtime version for the TPU.
                 - name: 'acceleratorType'
-                  immutable: true
-                  default_value: 'v2-8'
+                  default_from_api: true
                   description: |
                     TPU accelerator type for the TPU. If not specified, this defaults to 'v2-8'.
                 - name: 'description'
--- a/mmv1/products/tpuv2/QueuedResource.yaml
+++ b/mmv1/products/tpuv2/QueuedResource.yaml
@@ -28,7 +28,7 @@ update_verb: 'PATCH'
 update_mask: true
 autogen_async: true
 async:
-  actions: ['create', 'delete', 'update']
+  actions: ['create', 'delete']
   type: 'OpAsync'
   operation:
     base_url: '{{op_id}}'
--- a/mmv1/products/tpuv2/QueuedResource.yaml
+++ b/mmv1/products/tpuv2/QueuedResource.yaml
@@ -23,9 +23,7 @@ references:
 base_url: 'projects/{{project}}/locations/{{zone}}/queuedResources'
 self_link: 'projects/{{project}}/locations/{{zone}}/queuedResources/{{name}}'
 create_url: 'projects/{{project}}/locations/{{zone}}/queuedResources?queuedResourceId={{name}}'
-update_url: 'projects/{{project}}/locations/{{zone}}/queuedResources/{{name}}'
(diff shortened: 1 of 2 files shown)


mmv1/templates/terraform/examples/tpu_v2_queued_resource_basic.tf.tmpl
```diff
@@ -0,0 +1,16 @@
...
+
+  name = "{{index $.Vars "qr_name"}}"
+  zone = "us-central1-c"
+
+  tpu {
+    node_spec {
+      parent = "projects/{{index $.TestEnvVars "project"}}/locations/us-central1-c"
+      node_id = "{{index $.Vars "tpu_name"}}"
+      node {
+        runtime_version = "tpu-vm-tf-2.13.0"
```
SarahFrench, line 12: Could you please update this example to include the fields tpu.node_spec.node.description & tpu.node_spec.node.accelerator_type, [mentioned here](https://github.com/GoogleCloudPlatform/magic-modules/pull/12277#issuecomment-2463432967)?

SarahFrench (commented):
Hi, thanks for your PR! Here are some initial review comments

mmv1/templates/terraform/examples/tpu_v2_queued_resource_basic.tf.tmpl
zeleena: Done, updated the basic example to include filled out fields tpu.node_spec.node.description and tpu.node_spec.node.accelerator_type

mmv1/products/tpuv2/QueuedResource.yaml
```diff
@@ -0,0 +1,111 @@
...
+      - name: 'nodeSpec'
+        type: Array
+        description: |
+          The TPU node(s) being requested.
+        item_type:
+          type: NestedObject
+          properties:
+            - name: 'parent'
+              description: |
+                The parent resource name.
```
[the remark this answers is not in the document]
zeleena: Done, updated. I also updated some other fields to more accurately reflect the API docs

mmv1/products/tpuv2/QueuedResource.yaml
```diff
@@ -0,0 +1,111 @@
...
+            - name: 'nodeId'
+              immutable: true
+              description: |
+                Unqualified node identifier used to identify the node in the project once provisioned.
+            - name: 'node'
+              type: NestedObject
+              description: |
+                The node.
```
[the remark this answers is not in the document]
zeleena: Updated

mmv1/products/tpuv2/QueuedResource.yaml
```diff
@@ -0,0 +1,110 @@
...
+  api: 'https://cloud.google.com/tpu/docs/reference/rest/v2/projects.locations.queuedResources'
+base_url: 'projects/{{project}}/locations/{{zone}}/queuedResources'
+self_link: 'projects/{{project}}/locations/{{zone}}/queuedResources/{{name}}'
+create_url: 'projects/{{project}}/locations/{{zone}}/queuedResources?queuedResourceId={{name}}'
+update_url: 'projects/{{project}}/locations/{{zone}}/queuedResources/{{name}}'
+update_verb: 'PATCH'
+update_mask: true
+autogen_async: true
+async:
+  actions: ['create', 'delete', 'update']
```
SarahFrench, line 31: Can this resource be updated at all? The [documentation](https://cloud.google.com/tpu/docs/reference/rest/v2/projects.locations.queuedResources) doesn't show an update methods.

If the overall resource is immutable, we can mark it as immutable by setting [immutable:true](https://googlecloudplatform.github.io/magic-modules/develop/resource/#add-a-resource:~:text=%23-,immutable%3A%20true,-%23%20URL%20for%20the) as a top-level property in this yaml. At the same time it'd be good to remove update from the list of async actions.

mmv1/products/tpuv2/QueuedResource.yaml
```diff
@@ -0,0 +1,110 @@
...
+  - name: 'tpu_v2_queued_resource_basic'
+    primary_resource_id: 'qr'
+    min_version: 'beta'
+    vars:
+      qr_name: 'test-qr'
+      tpu_name: 'test-tpu'
+    test_env_vars:
+      project: 'PROJECT_NAME'
+    skip_vcr: true
+parameters:
```
SarahFrench, line 59: I think you also need to add `project` to the list of parameters. Currently the resource is able to use the default project value that users can set in the Google provider but they're unable to set a project value through the configuration for this resource.

SarahFrench (commented):
Thanks for checking the fields! They all look good now.

I've got two last bits of feedback that need action, but I think after those are addressed we'll be ready to merge 😁

mmv1/products/tpuv2/QueuedResource.yaml
zeleena: Removed 'update'. I don't think the overall resource is immutable because the `state` output field among other nested fields can be changed.

mmv1/products/tpuv2/QueuedResource.yaml
zeleena: I don't think the resources need the project parameter specifically. Looking at other resources, I don't see any other resources with project as a parameter, yet they are able to specify it in their examples and `project` still shows up on the Terraform resource documentation. Compute Engine Instances are one example of [no project parameter or property](https://github.com/GoogleCloudPlatform/magic-modules/blob/39b904188e53f162d10708cf35da3d48cd55731a/mmv1/products/compute/Instance.yaml#L55) yet [Terraform docs say project can still be configured for this resource](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/compute_instance#project-1). Another example is [firebasedatabase Instance which doesn't have the project parameter](https://github.com/GoogleCloudPlatform/magic-modules/blob/39b904188e53f162d10708cf35da3d48cd55731a/mmv1/products/firebasedatabase/Instance.yaml#L70) yet [project can also be configured for this resource](https://registry.terraform.io/providers/hashicorp/google-beta/latest/docs/resources/firebase_database_instance#project-2) and [is present in one of the examples with the resource](https://github.com/GoogleCloudPlatform/magic-modules/blob/0e4f8103665a42efd12b0c474388d44f6ef112c5/mmv1/templates/terraform/examples/firebase_database_instance_basic.tf.tmpl#L3).

I updated the queued resource example with a project field to show in the example that the project can be configured for this resource.

zeleena:
Thanks for your review @SarahFrench, please see my latest replies to your comments and let me know if this PR looks good to you. I was aiming to merge this by Wednesday November 20th in order to meet upcoming deadlines, please let me know if you think this won't be possible or if there's anything I can do to speed up the review process. Thank you in advance!

mmv1/products/tpuv2/QueuedResource.yaml
SarahFrench: Thanks for looking into this - it looks like answer to the problem is that the code generator adds in a project field automatically if the generator detects that the base_url value contains `{{project}}`, departing from what is described in the YAML file. For example [the link to the diff for the google-beta in this diff report](https://github.com/GoogleCloudPlatform/magic-modules/pull/12277#issuecomment-2465592744) shows that the project field is present in the schema of the new resource you are adding despite it not being in the YAML in this PR. As the generator is stopping the problem I described happening I agree you don't need to explicitly include it in the schema.

mmv1/products/tpuv2/QueuedResource.yaml
SarahFrench: When [resources are marked as immutable in the YAML](https://googlecloudplatform.github.io/magic-modules/develop/resource-reference/#immutable) it causes all the fields to be marked as [ForceNew](https://developer.hashicorp.com/terraform/plugin/sdkv2/schemas/schema-behaviors#forcenew). This means if a user changes any user-supplied argument in the resource block for the `google_tpu_v2_queued_resource` resource then the resource will be recreated (delete+create) instead of updated in place. Any changes to output values that come from the API only and aren't user supplied values will still be updated when Terraform performs refresh action.

Does that make sense? I think marking the resource as immutable is necessary due to the lack of an update method in the API but this would not prevent the API changing output values and users being able to see those changed output values in their plans and state.

SarahFrench: When manually testing updating a google_tpu_v2_queued_resource resource as it's currently implemented (immutable:false) I see this:

```
  ~ update in-place

  # google_tpu_v2_queued_resource.qr will be updated in-place
  ~ resource "google_tpu_v2_queued_resource" "qr" {
        id      = "projects/redacted/locations/us-central1-c/queuedResources/tf-test-test-qr"
        name    = "tf-test-test-qr"
        # (2 unchanged attributes hidden)

      ~ tpu {
          ~ node_spec {
                # (2 unchanged attributes hidden)

              ~ node {
                  ~ description      = "Text description of the TPU." -> "UPDATED Text description of the TPU."
                    # (2 unchanged attributes hidden)
                }
            }
        }
    }

  Enter a value: yes

google_tpu_v2_queued_resource.qr: Modifying... [id=projects/redacted/locations/us-central1-c/queuedResources/tf-test-test-qr]
╷
│ Error: Error updating QueuedResource "projects/redacted/locations/us-central1-c/queuedResources/tf-test-test-qr": googleapi: got HTTP response code 404 with body: <!DOCTYPE html>
```

mmv1/products/tpuv2/QueuedResource.yaml
```diff
@@ -0,0 +1,110 @@
...
+        type: Array
+        description: |
+          The TPU node(s) being requested.
+        item_type:
+          type: NestedObject
+          properties:
+            - name: 'parent'
+              description: |
+                The parent resource name.
+              required: true
```
SarahFrench, line 90: We need diff suppression on this field because users can use the project id in the value and the API will return the value containing the project number. Without diff suppression users will see unexpected diffs.

```suggestion
              required: true
              diff_suppress_func: 'tpgresource.ProjectNumberDiffSuppress'
```

mmv1/templates/terraform/examples/tpu_v2_queued_resource_basic.tf.tmpl
```diff
@@ -0,0 +1,19 @@
+resource "google_tpu_v2_queued_resource" "{{$.PrimaryResourceId}}" {
+  provider = google-beta
+
+  name    = "{{index $.Vars "qr_name"}}"
+  zone    = "us-central1-c"
+  project = "{{index $.TestEnvVars "project_id"}}"
```
SarahFrench, line 6: Currently the generated test and documentation from this example leaves this field as an empty string, as there isn't a project_id var. Using `project` fixes this.

```suggestion
  project = "{{index $.TestEnvVars "project"}}"
```

SarahFrench (commented):
In addition to discussions about whether the resource can be updated by users or not [in this thread](https://github.com/GoogleCloudPlatform/magic-modules/pull/12277#discussion_r1842076847), I've found two additional changes that are required. One ensures the acceptance test gets the arguments it needs, and the other suppresses a diff caused by the API changing a field's value away from the value users supply in their config.

SarahFrench:
> Thanks for your review @SarahFrench, please see my latest replies to your comments and let me know if this PR looks good to you. I was aiming to merge this by Wednesday November 20th in order to meet upcoming deadlines, please let me know if you think this won't be possible or if there's anything I can do to speed up the review process. Thank you in advance!

If the 3 points in my review above are addressed today we can merge today 👍🏻

mmv1/templates/terraform/examples/tpu_v2_queued_resource_basic.tf.tmpl
zeleena: Updated to `project`

mmv1/products/tpuv2/QueuedResource.yaml
zeleena: Updated to use this `diff_suppress_func`

mmv1/products/tpuv2/QueuedResource.yaml
zeleena: Updated the resource to be immutable.

That makes sense if it's only the user-supplied fields which cannot be changed and not the output values, and makes sense since there is no update method.

SarahFrench (approved):
Test passes - thanks for making the suggested changes.

The release on December 2nd is skipped, so the next release will be December 9th. This PR will be in that release.

PR #10728, opened 2024-05-16
docs: Correction for google_sql_database_instance database_flags

marchah:
Currently the [Terraform documentation](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/sql_user) for `google_sql_database_instance` `database_flags` is incorrect. It's showing `cloudsql.iam_authentication` instead of `cloudsql_iam_authentication` like it's showed in the official [GCloud  documentation](https://cloud.google.com/sql/docs/mysql/create-edit-iam-instances#terraform) and also [here](https://cloud.google.com/sql/docs/mysql/flags#terraform)

A ticket referencing this issue can be found here https://github.com/hashicorp/terraform-provider-google/issues/16941

**Release Note Template for Downstream PRs (will be copied)**

```release-note:none
```

PR #12282, opened 2024-11-08
doc: fix wrong iam auth flag for sql_user

eliecharra:
This PR was incorrect https://github.com/GoogleCloudPlatform/magic-modules/pull/10728

Flags are slightly inconsistent between MySQL and Postgres.

This fixes the flag for the Postgres example.

image

```release-note:none
cloudsql: fixed example in doc using wrong iam auth flag for PostgresQL in `google_sql_user`
```


wyardley:
I think I ran into this in the past, and was a pain to notice the difference, so thanks for fixing!

shuyama1 (approved):
Thanks for the fix!

PR #12285, opened 2024-11-08
compute: added `numeric_id` to `google_compute_subnetwork`

wyardley:
Added `numeric_id` to `google_compute_subnetwork` resource and data source.

Since Terraform uses `id` internally, follow the example of `google_compute_network`, and make `numeric_id` contain the `id` field from the API.

As mentioned in the issue below, it would be very nice if there were a simpler way to just say "map `id` in the API to `numericId` here", but guessing that's not possible?

Part of hashicorp/terraform-provider-google#20223

```release-note:enhancement
compute: added `numeric_id` to `google_compute_subnetwork` resource and data source
```

### DIFF
--- a/mmv1/products/compute/Subnetwork.yaml
+++ b/mmv1/products/compute/Subnetwork.yaml
@@ -73,6 +73,9 @@ iam_policy:
 custom_code:
   extra_schema_entry: 'templates/terraform/extra_schema_entry/subnetwork.tmpl'
   constants: 'templates/terraform/constants/subnetwork.tmpl'
+  decoder: 'templates/terraform/decoders/compute_subnetwork.go.tmpl'
+  encoder: 'templates/terraform/encoders/compute_subnetwork.go.tmpl'
+  update_encoder: 'templates/terraform/update_encoder/compute_subnetwork.go.tmpl'
   post_update: 'templates/terraform/post_update/compute_subnetwork.go.tmpl'
 custom_diff:
   - 'customdiff.ForceNewIfChange("ip_cidr_range", IsShrinkageIpCidr)'
@@ -203,6 +206,10 @@ properties:
     custom_expand: 'templates/terraform/custom_expand/resourceref_with_validation.go.tmpl'
     resource: 'Network'
     imports: 'selfLink'
+  - name: 'numericId'
+    type: String
+    description: The numeric ID of the resource as a string.
+    output: true
   - name: 'purpose'
     type: String
     description: |
--- a/mmv1/products/compute/Subnetwork.yaml
+++ b/mmv1/products/compute/Subnetwork.yaml
@@ -207,8 +207,8 @@ properties:
     resource: 'Network'
     imports: 'selfLink'
   - name: 'numericId'
-    type: String
-    description: The numeric ID of the resource as a string.
+    type: Integer
+    description: The numeric ID of the resource.
     output: true
   - name: 'purpose'
     type: String
--- a/mmv1/third_party/terraform/services/compute/data_source_google_compute_subnetwork.go.tmpl
+++ b/mmv1/third_party/terraform/services/compute/data_source_google_compute_subnetwork.go.tmpl
@@ -2,6 +2,7 @@ package compute
 
 import (
 	"fmt"
+	"strconv"
 
 	"github.com/hashicorp/terraform-provider-google/google/tpgresource"
 	transport_tpg "github.com/hashicorp/terraform-provider-google/google/transport"
@@ -28,6 +29,10 @@ func DataSourceGoogleComputeSubnetwork() *schema.Resource {
 				Computed: true,
 			},
 			"description": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+			"numeric_id": {
 				Type:     schema.TypeString,
 				Computed: true,
 			},
@@ -98,6 +103,10 @@ func dataSourceGoogleComputeSubnetworkRead(d *schema.ResourceData, meta interfac
 	subnetwork, err := config.NewComputeClient(userAgent).Subnetworks.Get(project, region, name).Do()
 	if err != nil {
 		return transport_tpg.HandleDataSourceNotFoundError(err, d, fmt.Sprintf("Subnetwork Not Found : %s", name), id)
+	}
+
+	if err := d.Set("numeric_id", strconv.Itoa(int(subnetwork.Id))); err != nil {
+		return fmt.Errorf("Error setting numeric_id: %s", err)
 	}
 
 	if err := d.Set("ip_cidr_range", subnetwork.IpCidrRange); err != nil {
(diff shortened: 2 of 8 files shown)


mmv1/products/compute/Subnetwork.yaml
```diff
@@ -73,6 +73,9 @@ iam_policy:
 custom_code:
   extra_schema_entry: 'templates/terraform/extra_schema_entry/subnetwork.tmpl'
   constants: 'templates/terraform/constants/subnetwork.tmpl'
+  decoder: 'templates/terraform/decoders/compute_subnetwork.go.tmpl'
+  encoder: 'templates/terraform/encoders/compute_subnetwork.go.tmpl'
+  update_encoder: 'templates/terraform/update_encoder/compute_subnetwork.go.tmpl'
```
wyardley, line 78: these are all identical to the ones for `compute_network`; I'm assuming there's still a preference to have these be copied / pasted vs. sharing a common file with a more generic name?

mmv1/products/compute/Subnetwork.yaml
```diff
@@ -203,6 +206,10 @@ properties:
     custom_expand: 'templates/terraform/custom_expand/resourceref_with_validation.go.tmpl'
     resource: 'Network'
     imports: 'selfLink'
+  - name: 'numericId'
+    type: String
```
wyardley, line 210: It's an integer in the API, but followed the example from `google_compute_network` of making it a string, so I assume that's preferred?

mmv1/third_party/terraform/services/compute/data_source_google_compute_subnetwork.go.tmpl
```diff
@@ -100,6 +105,10 @@ func dataSourceGoogleComputeSubnetworkRead(d *schema.ResourceData, meta interfac
 		return transport_tpg.HandleDataSourceNotFoundError(err, d, fmt.Sprintf("Subnetwork Not Found : %s", name), id)
 	}
 
+	if err := d.Set("numeric_id", strconv.Itoa(int(subnetwork.Id))); err != nil {
```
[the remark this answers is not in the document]
BBBmau: if we mark numericId as an integer we shouldn't need to do this extra conversion unless there's something that's preventing integer to be used over string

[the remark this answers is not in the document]
wyardley: Google project number, org ID, etc. also seems to be implemented as strings vs. int
https://github.com/GoogleCloudPlatform/magic-modules/blob/817990220aced261b121b0e32dca4ab160e72c55/mmv1/third_party/terraform/services/resourcemanager/resource_google_project.go#L104

wyardley:
@SarahFrench @melinath either of you have any background / perspective on what things should be implemented as strings vs numbers within the provider?

melinath:
in general, it should match the API type. I don't know why the other field you mentioned is a string instead.

wyardley:
So then I guess the question is whether we should follow the other examples within this resource type for consistency, or update it to be a number. I’m happy to do whatever is preferred.

I can test, but is there any type conversion needed if one were interpolating the number value within a string in terraform itself?

melinath:
the only reason I can think of to use a string would be if a user needed to be able to set "0" on an optional + computed field - but I would still hesitate doing that since it would be a hacky workaround, and this isn't an optional + computed field so that wouldn't apply anyway. I'd say just go with matching the API.

wyardley:
@BBBmau Updated to be an integer. Tests seem to pass, and also functionally tested against the built provider.

Might be smart to switch over `google_compute_network` to match at some point?

wyardley:
Created #12339 to update this for the `google_compute_network` resource, while also adding it to the data source for parity.

wyardley:
Is it expected that the "awaiting approval" labels aren't seeming to get applied / removed?

mmv1/third_party/terraform/services/compute/resource_compute_subnetwork_test.go.tmpl
```diff
@@ -94,6 +96,36 @@ func TestAccComputeSubnetwork_basic(t *testing.T) {
...
+
+	acctest.VcrTest(t, resource.TestCase{
+		PreCheck:                 func() { acctest.AccTestPreCheck(t) },
+		ProtoV5ProviderFactories: acctest.ProtoV5ProviderFactories(t),
+		CheckDestroy:             testAccCheckComputeSubnetworkDestroyProducer(t),
+		Steps: []resource.TestStep{
+			{
+				Config: testAccComputeSubnetwork_numericId(cnName, subnetworkName),
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestMatchResourceAttr("google_compute_subnetwork.numeric_id_test", "numeric_id", regexp.MustCompile("^\\d{16,48}$")),
```
wyardley, line 115: There's no builtin check on type here, right? Like to assert that it's an int or that it's a string?

mmv1/third_party/terraform/services/compute/data_source_google_compute_subnetwork.go.tmpl
```diff
@@ -100,6 +104,10 @@ func dataSourceGoogleComputeSubnetworkRead(d *schema.ResourceData, meta interfac
 		return transport_tpg.HandleDataSourceNotFoundError(err, d, fmt.Sprintf("Subnetwork Not Found : %s", name), id)
 	}
 
+	if err := d.Set("numeric_id", subnetwork.Id); err != nil {
```
wyardley, line 107: I didn't seem to need to convert from `int64` to `int` here, even though I did when I was doing string conversion. Let me know if there should be an `int(subnetwork.Id)` here.

wyardley:
@SarahFrench pointed out https://googlecloudplatform.github.io/magic-modules/develop/field-reference/#api_name in the docs, which might be simpler than the custom templates if it works?

I think the test failure is unrelated?

wyardley:
Maybe not needed after #12351?

ScottSuarez:
> Maybe not needed after #12351?

Ack, yeah this can be closed. Although if you want you could add it to the datasource still @wyardley

wyardley:
I will look at it again in a week or two. Opened #12339 a bit back for compute_network as well

wyardley:
> Ack, yeah this can be closed. Although if you want you could add it to the datasource still @wyardley

@ScottSuarez Thanks. Seems like your changes used `foo.foo_id` vs `foo.numeric_id`? If that's the convention, if a breaking change is made later to add the network ID as an integer, maybe it would make sense to make that change? or even to add `google_compute_network.network_id` ahead of time as an integer, and remove `google_compute_network.numeric_id` as a breaking change later, to make network and subnetwork consistent both in type and in naming?

I made a followup PR for adding the data source only (basically cherry-picking part of this PR): #12461

Also made hashicorp/terraform-provider-google#20530 to track some of the inconsistencies of name / type that might be breaking changes.

PR #12287, opened 2024-11-08
Remove gke_hub as part of the DCL endpoints

hao-nan-li:
part of https://github.com/hashicorp/terraform-provider-google/issues/20246

```release-note:none

```


hao-nan-li:
@roaks3 Looks like this change didn't really affect any of the existing test for gkehub, maybe this is not the root cause. Can we still merge it since this line is no longer being used and could potentially cause confusion?

roaks3 (approved):
Sounds good, but might want to remove the `fixes: ` from the description if this doesn't fully resolve that issue.

hao-nan-li:
> Sounds good, but might want to remove the `fixes: ` from the description if this doesn't fully resolve that issue.

Good point, will take that part off.

hao-nan-li:
> > Sounds good, but might want to remove the `fixes: ` from the description if this doesn't fully resolve that issue.
> 
> Good point, will take that part off.

On a second thought, this might be the right fix, since the VCR runs on beta and removing this line doesn't affect anything because everything was already running in beta. I'll check in the nightly runs and see if this fixed the issue.

PR #12291, opened 2024-11-10
Added google_apigee_api_deployment resource

apichick:
This is a PR to add the google_apigee_api_deployment resource

```release-note:new-resource
`google_apigee_api_deployment`
```

### DIFF
--- a/mmv1/products/apigee/ApiDeployment.yaml
+++ b/mmv1/products/apigee/ApiDeployment.yaml
@@ -1,0 +1,198 @@
+# Copyright 2024 Google Inc.
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+---
+name: "ApiDeployment"
+description: |
+  An `ApiDeployment` in Apigee.
+references:
+  guides:
+    "API Proxy Deployments": "https://cloud.google.com/apigee/docs/api-platform/deploy/ui-deploy-overview"
+  api: "https://cloud.google.com/apigee/docs/reference/apis/apigee/rest/v1/organizations.environments.apis.revisions.deployments"
+docs:
+base_url: "{{env_id}}/apis/{{api_name}}/revisions/{{revision}}/deployments"
+self_link: "{{env_id}}/apis/{{api_name}}/revisions/{{revision}}/deployments"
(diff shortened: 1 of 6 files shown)


mmv1/third_party/terraform/website/docs/r/apigee_api_deployment.html.markdown
```diff
@@ -0,0 +1,108 @@
...
+- `update` - Default is 20 minutes.
+- `delete` - Default is 20 minutes.
+
+## Import
+
+
+An API proxy deployment can be imported using any of these accepted formats:
+
+* `organizations/{{org_id}}/environments/{{environment}}/apis/{{api_id}}/revisions/{{revision}}/deployments/{{name}}`
+* `{{org_id}}/{{environment}}/{{sharedflow_id}}/{{revision}}/{{name}}`
```
roaks3, line 92: s/sharedflow_id/api_id/ ?

Similar for various shared flow references below this on the page.

roaks3 (commented):
Overall LGTM, but a couple of high-level questions:
- Does this actually need to be a handwritten resource, or could it be implemented as an mmv1 resource like `mmv1/products/apigee/Environment.yaml`? I didn't notice anything that needed to be handwritten here, and it looked like the sharedflow resource needed `sendRequestRawBodyWithTimeout`, which I don't see used here.
- Do you know why the test needs to be skipped in VCR?

mmv1/third_party/terraform/website/docs/r/apigee_api_deployment.html.markdown
apichick: updated the references

apichick:
Question, I did it not use it using yamls, because the google_apigee_sharedflow_deployment resource which is strcturally similar was not done like that, and I assumed that was the case, because we need to use the same text-fixtures as with google_apigee_api resource. Is it possible to implement resources with yaml and use test-fixtures, where would you place them?

apichick:
I modified the PR so now there resource is done using YAML. However i introduced a change in the google_apigee_api resource and I did that for a reason. Originally I was passing to the google_apigee_api resource the org_id as "{{org_name}}" , for consistency, just as the google_apigee_shareflow resource did. However I realized that forces to declare the dependency of the organization explicitly. So I decided to pass org_id as "organizations/{{org_name}}". I think that should modified in the google_apigee_sharedflow resource too. If that change is not OK because it is backwards incompatible I can drop it, however I think it is better this way.

roaks3 (commented):
Just posting a comment to clear the reminders: this is waiting on the feasibility of supporting both "{{org_name}}" and "organizations/{{org_name}}" formats. Removing the existing format would be considered a breaking change.

osakri:
@roaks3 any updates on this ?
this resource could be very helpful for our infra.

cjhouser:
@roaks3 what did the feasibility study determine?

roaks3:
This PR was closed due to inactivity. Please add future comments regarding this feature on https://github.com/hashicorp/terraform-provider-google/issues/22581 (I'm currently not aware of any timeline for this feature).

PR #12293, opened 2024-11-11
Add status tag for tracking autogen-created resources

slevenick:
Used for tracking purposes, no actual use yet

```release-note:none

```


c2thorn (approved):
are we going to add a check for this in main.go? We could keep all of the `AutogenStatus`'s found as we iterate through the resources and flag for duplicates

slevenick:
> are we going to add a check for this in main.go? We could keep all of the `AutogenStatus`'s found as we iterate through the resources and flag for duplicates

I see the checking as something we can add in the future. Adding this tag is enough for now as we can check them against the resource name when we count everything up (and see any copy/pastes)

PR #12295, opened 2024-11-11
add fwutils package to separate StringSet

BBBmau:
StringSet is used across multiple ephemeral resources. this puts it in a utils file so that it can be used across all.

```release-note:none

```


SarahFrench (changes requested):
I think you need to add changes so that the new package is copied into the generated repo, similar to your PR adding the fwvalidators package

PR #12296, opened 2024-11-11
Add deletion_protection field for cloudExadataInfrastructure resource

tulika-aakriti:
part of : https://github.com/hashicorp/terraform-provider-google/issues/20284

**Release Note Template for Downstream PRs (will be copied)**

See [Write release notes](https://googlecloudplatform.github.io/magic-modules/contribute/release-notes/) for guidance.

```release-note: enhancement
oracledatabase: added `deletion_protection` field to `google_oracle_database_cloud_exadata_infrastructure`
```

PR #12298, opened 2024-11-11
add deletion_protection for autonomousDatabase resource

tulika-aakriti:
part of : https://github.com/hashicorp/terraform-provider-google/issues/20284

```release-note:enhancement
oracledatabase: added `deletion_protection` field to `google_oracle_database_autonomous_database`
```


slevenick:
I don't believe we can add this in a minor version of the provider because it would be changing the default behavior of the resource. @rileykarson is that correct in this case?

rileykarson:
Yeah, we'd strongly prefer to not do so outside of a major. We discuss in https://googlecloudplatform.github.io/magic-modules/best-practices/immutable-fields/ that immutable fields are expected in Terraform. However, if there's undue risk w/ deleting these resources past the standard risks of a database deletion we're glad to continue to discuss an exception here.

tulika-aakriti:
I see that the major release is once every year which might be risk for customer. 
The deletion api will delete all the exisiting data from database and the data in the third party i.e oracleDatabase and there will be no way for a user to restore the data. I see many GCP products like bugQuery and SQL uses this to warn the user regarding the data loss. Moving forward we are planning to add update API and will continue to use deletion_protection as a mitigations for accidental delete.

tulika-aakriti:
We don't support update API as of now, so every update call can lead to accidental deletion.  Can we get exception for 3 of our resources for weekly release. Our service onboarded few weeks ago and we don't have any active customers now.

mmv1/products/oracledatabase/AutonomousDatabase.yaml
```diff
@@ -48,23 +48,39 @@ async:
...
     message: 'message'
 custom_code:
+  pre_delete: 'templates/terraform/pre_delete/oracledatabase_autonomous_database.go.tmpl'
 examples:
   - name: 'oracledatabase_autonomous_database_basic'
     primary_resource_id: 'myADB'
     vars:
       project: 'my-project'
       autonomous_database_id: 'my-instance'
+      deletion_protection: 'true'
```
slevenick, line 58: Have you run this test? I'd imagine it no longer passes because we are effectively preventing the deletion for the teardown of the resource

slevenick (commented):
I'm fine with this being added. We should sync up merging with the other 3 resources so they all go out at the same time

mmv1/products/oracledatabase/AutonomousDatabase.yaml
tulika-aakriti: I am overriding it later to false, in line 62.

slevenick (approved):

slevenick:
Let me know when we are ready to merge this with the other PRs

mmv1/templates/terraform/pre_delete/oracledatabase_autonomous_database.go.tmpl
```diff
@@ -0,0 +1,3 @@
+if d.Get("deletion_protection").(bool) {
+	return fmt.Errorf("cannot destroy instance without setting deletion_protection=false and running `terraform apply`")
```
rileykarson, line 2: drive-by: we probably want to apply a similar change to https://github.com/GoogleCloudPlatform/magic-modules/pull/12296#discussion_r1847173127 here

tulika-aakriti:
> Let me know when we are ready to merge this with the other PRs

All three PR are in final review, We can get it merge once the build run is successfull.

mmv1/templates/terraform/pre_delete/oracledatabase_autonomous_database.go.tmpl
slevenick: Yep, please apply this fix!

tulika-aakriti:
Is there any pending changes required to close this PR?
cc: @rileykarson @slevenick

PR #12303, opened 2024-11-11
Use project name for PSC allowlist

ericgribkoff:
Fixes https://github.com/hashicorp/terraform-provider-google/issues/20264

b/378162820


melinath (approved):
LGTM, looks like the test is passing. thanks!

PR #12305, opened 2024-11-12
fix: derive the project attribute from resource or compute api or fallback to provider

gurusai-voleti:
Fixes https://github.com/hashicorp/terraform-provider-google/issues/17165

```release-note:bug
storage: derive the project attribute from resource or compute api or fallback to provider in `resource_storage_bucket`
```


SirGitsalot (changes requested):
LGTM for the code change, but please update the docs too:

https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/third_party/terraform/website/docs/d/storage_bucket.html.markdown

The "If no value is supplied in the configuration *or through provider defaults* then the data source will use the Compute API..." is not accurate anymore

gurusai-voleti:
> LGTM for the code change, but please update the docs too:
> 
> https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/third_party/terraform/website/docs/d/storage_bucket.html.markdown
> 
> The "If no value is supplied in the configuration _or through provider defaults_ then the data source will use the Compute API..." is not accurate anymore

Updated documentation

SirGitsalot (approved):

PR #12307, opened 2024-11-12
container: fix `TestAccContainerCluster_withFleetConfig` failure

wyardley:
Use a valid (but different) project for the test that migrating the fleet or cluster to a different project fails

Fixes hashicorp/terraform-provider-google/issues/20253

See comments in linked issue

Basically, it just needs some valid project ID... not sure if the right fix here is to use `BootstrapProject` (which exists, but seems to be unused?) to create a project for this purpose, tied to the test project ID, or just hardcode a different project that we know is valid.

I'm not sure if there used to be a project called `random-project` and it got deleted, or if the API behavior just changed. For now, I used `tdx-guest-images`, which I know from other tests is a valid project, and which these accounts wouldn't have permissions on. But let me know if there's a specific project name that would be better to use for this purpose that we know will reliably exist.

```release-note:none

```

### DIFF
--- a/mmv1/third_party/terraform/services/container/resource_container_cluster_test.go.tmpl
+++ b/mmv1/third_party/terraform/services/container/resource_container_cluster_test.go.tmpl
@@ -5232,7 +5232,7 @@ func TestAccContainerCluster_withFleetConfig(t *testing.T) {
 				ImportStateVerifyIgnore: []string{"min_master_version", "deletion_protection"},
 			},
 			{
-				Config:		testAccContainerCluster_withFleetConfig(clusterName, "random-project", networkName, subnetworkName),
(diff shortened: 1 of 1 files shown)


zli82016:
I don't see `tdx-guest-images` is used in other tests in providers.  Is `tdx-guest-images` owned by your team, @wyardley ?

wyardley:
> I don't see tdx-guest-images is used in other tests in providers. Is tdx-guest-images owned by your team, @wyardley

No - see comments in the PR description. I believe / assume it's google-related, and it may host images for confidential compute? I pulled out some docker images that were there in #12148.

Really, this could be anything -- just used this one to prove out the fix. It just has to be a valid GCP project that exists and is unlikely to go away. So if you've got suggestions for something that fits the bill, happy to update it to a different project.

zli82016 (approved):
LGTM.

PR #12315, opened 2024-11-13
Enhanced documentation and property dependency for `quota` in `google_identity_platform_config`

pashanka:
Fixes https://github.com/hashicorp/terraform-provider-google/issues/16172

```release-note:enhancement
identityplatform: marked `quota.0.sign_up_quota_config` subfields conditionally required in `google_identity_platform_config` to move errors from apply time up to plan time, and clarified the rule in documentation
```

Manual testing:
After the change the error message is more clear on the expectations

Sample config:
```
resource "google_project_service" "identitytoolkit" {
  provider = google-beta  
  project = "my-project"
  service = "identitytoolkit.googleapis.com"
}

resource "google_identity_platform_config" "default" {
  provider = google-beta  
  project = "my-project"
  autodelete_anonymous_users = true
  sign_in {
    allow_duplicate_emails = true

    email {
        enabled = false
    }

    phone_number {
      enabled = false
    }

    anonymous {
      enabled = false
    }
  }
  sms_region_config {
    allowlist_only {
      allowed_regions = [
        "US",
        "CA",
      ]
    }
  }
  quota {
    sign_up_quota_config {
      quota = 100
    }
  }
  authorized_domains = [
    "localhost",
    "pavan-test-terraform.firebaseapp.com",
    "pavan-test-terraform.web.app",
  ]

  depends_on = [ google_project_service.identitytoolkit ]
}
```

Error before changes:
```
 Error 400: INVALID_CONFIG : SignUp quota must start between now and 365 days from now.
```

Error after changes:
```
│ Error: Missing required argument
│ 
│   with google_identity_platform_config.default,
│   on main.tf line 36, in resource "google_identity_platform_config" "default":
│   36:       quota = 100
│ 
│ "quota.0.sign_up_quota_config.0.quota": all of
│ `quota.0.sign_up_quota_config.0.quota,quota.0.sign_up_quota_config.0.quota_duration,quota.0.sign_up_quota_config.0.start_time` must be specified
```


mmv1/products/identityplatform/Config.yaml
```diff
@@ -228,20 +228,29 @@ properties:
       - name: 'signUpQuotaConfig'
         type: NestedObject
         description: |
-          Quota for the Signup endpoint, if overwritten. Signup quota is measured in sign ups per project per hour per IP.
+          Quota for the Signup endpoint, if overwritten. Signup quota is measured in sign ups per project per hour per IP. None of the child properties can be skipped.
```
rainshen49, line 231: Be more explicit in case we add new fields in the future: None of quota, startTime, or quotaDuration can be skipped.

rainshen49:
Add `quota_start_time: '2014-10-02T15:01:23Z'` to `test_env_vars` to https://github.com/rainshen49/magic-modules/blob/main/mmv1/products/identityplatform/Config.yaml#L52-L54. This is the value that gets shown in the documentation. Currently, the [documentation](https://registry.terraform.io/providers/hashicorp/google-beta/latest/docs/resources/identity_platform_config) shows empty start_time

```hcl
   sign_up_quota_config {
      quota = 1000
      start_time = ""
      quota_duration = "7200s"
    }
```

Note that tests will use the override value below:
`'quota_start_time': 'time.Now().AddDate(0, 0, 1).Format(time.RFC3339)'`, so the above change is for presentation purposes only.

mmv1/products/identityplatform/Config.yaml
pashanka: Makes sense. Updated

rainshen49 (approved):
Confirmed the diff reflects the documentation changes correctly. 
https://github.com/modular-magician/terraform-provider-google/compare/auto-pr-12315-old..auto-pr-12315#diff-a3655a5dd07dca1febbf680da1c9cceec72e855368c20ab2d187d16dc94123f9

mmv1/products/identityplatform/Config.yaml
```diff
@@ -228,20 +229,29 @@ properties:
...
         description: |
-          Quota for the Signup endpoint, if overwritten. Signup quota is measured in sign ups per project per hour per IP.
+          Quota for the Signup endpoint, if overwritten. Signup quota is measured in sign ups per project per hour per IP. None of quota, startTime, or quotaDuration can be skipped.
         properties:
           - name: 'quota'
             type: Integer
             description: |
-              A sign up APIs quota that customers can override temporarily.
+              A sign up APIs quota that customers can override temporarily. Value can be in between 1 and 1000.
+            required_with:
```
rileykarson, line 238: Using this means it's valid to specify `sign_up_quota_configuration {}` to empty or zero the value, is that what's intended? We could just make all these `required` if that is not valid to write (required-ness is scoped to the parent field for nested fields)

rainshen49: They current TF surface and the API allow an entirely empty (but present) signUpQuotaConfig. Although it probably doesn't do anything different from removing the entire signUpQuotaConfig block, we don't want to break users who are using it this way. Perhaps this is a good candidate for the next major version? @pashanka

rileykarson: Thanks! Seems reasonable to decide out of band but a decision is not merge blocking since we want the current behaviour.

rileykarson (approved):

mmv1/products/identityplatform/Config.yaml
pashanka: Yes, that is correct. Ideally we can make everything as required but that changes from the existing behaviour that empty value works today. Hence, decided to do this to keep it this way for backward compatibility. I can look at fixing this for the next major version.

PR #12316, opened 2024-11-13
Adding warning notes for field loginHint and gcipSettings in IAP Settings

Deepraj1996:
```release-note:none
Terraform docs: Added warning notes in `google_iap_settings` resource
```
```release-note:none
```

### DIFF
--- a/mmv1/products/iap/Settings.yaml
+++ b/mmv1/products/iap/Settings.yaml
@@ -111,6 +111,7 @@ properties:
         properties:
           - name: 'loginHint'
             type: String
+            custom_expand: 'templates/terraform/custom_expand/iap_settings_log_hints.go.tmpl'
             description: |
               Domain hint to send as hd=? parameter in OAuth request flow.
               Enables redirect to primary IDP by skipping Google's login screen.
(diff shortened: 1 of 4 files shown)


roaks3 (commented):
This looks a bit heavy-handed. Do you have an links that explain why you need to do this? The simplest option would generally be to just include these warnings as part of the field description. Also, I think the expanders would only run on apply, so the user would have already made their change by then.

Deepraj1996:
> This looks a bit heavy-handed. Do you have an links that explain why you need to do this? The simplest option would generally be to just include these warnings as part of the field description. Also, I think the expanders would only run on apply, so the user would have already made their change by then.

Deepraj1996:
Sure @roaks3, I will remove the expanders and the note to the description of the field.

roaks3 (commented):
👍

roaks3:
@Deepraj1996 something seems to be wrong with your CLA. Could you click on the details link and try to resolve? We can't move forward without that.

roaks3 (commented):
Mark review

PR #12320, opened 2024-11-13
tpgresource: added support for default location in `tpgresource.ReplaceVars`

wyardley:
This supports inferring `{{location}}` from the provider (similar to what is done with `{{region}}` and `{{zone}}` in
`tpgresource.ReplaceVars()`, vs. just taking a value if it's explicitly set.

Even if all tests pass, it's probably necessary for someone who understands how this is used to weigh in and make sure this isn't likely to have any unintended consequences / side effects.

While `ReplaceVars()` is used all over the place, like the replacements for `{{region}}` and `{{zone}}`, `BuildReplacementFunc()` is using a method (`GetLocation()`, in this case) that was seemingly originally intended for use with GKE clusters, but probably generically will work in some situations.

In this case, for an artifact registry repository resource with a location set at provider level, this fixes some issues with import _if_ the resource is in the same region that's configured at provider level.

Part of hashicorp/terraform-provider-google#19266

```release-note:none

```

### DIFF
--- a/mmv1/third_party/terraform/tpgresource/utils_test.go
+++ b/mmv1/third_party/terraform/tpgresource/utils_test.go
@@ -1079,6 +1079,32 @@ func TestReplaceVars(t *testing.T) {
 			},
 			Expected: "projects/default-project/zones/default-zone/instances",
 		},
+		"location with provider level region": {
+			Template: "projects/{{project}}/locations/{{location}}/repositories",
+			Config: &transport_tpg.Config{
+				Project: "default-project",
+				Region:  "default-region",
+			},
+			Expected: "projects/default-project/locations/default-region/repositories",
+		},
+		"location with provider level region and zone": {
+			Template: "projects/{{project}}/locations/{{location}}/repositories",
+			Config: &transport_tpg.Config{
+				Project: "default-project",
+				Region:  "default-region",
+				Zone:    "default-region-a",
+			},
+			Expected: "projects/default-project/locations/default-region/repositories",
+		},
+		"location with provider level zone": {
+			// May not actually be useful / valid for all use cases.
+			Template: "projects/{{project}}/locations/{{location}}/repositories",
+			Config: &transport_tpg.Config{
+				Project: "default-project",
+				Zone:    "default-region-a",
+			},
+			Expected: "projects/default-project/locations/default-region-a/repositories",
+		},
 		"regional schema values": {
 			Template: "projects/{{project}}/regions/{{region}}/subnetworks/{{name}}",
 			SchemaValues: map[string]interface{}{
@@ -1124,6 +1150,19 @@ func TestReplaceVars(t *testing.T) {
 				"innerzone": "inner",
 			},
 			Expected: "projects/project1/zones/wrapperinnerwrapper/instances/instance1",
+		},
+		"location with schema values": {
+			Template: "projects/{{project}}/locations/{{location}}/repositories/{{repository_id}}",
+			Config: &transport_tpg.Config{
+				Project: "default-project",
+				Region:  "default-region",
+			},
+			SchemaValues: map[string]interface{}{
+				"location":      "other-location",
+				"project":       "project1",
+				"repository_id": "foo",
+			},
+			Expected: "projects/project1/locations/other-location/repositories/foo",
 		},
 		"base path recursive replacement": {
 			Template: "{{CloudRunBasePath}}namespaces/{{project}}/services",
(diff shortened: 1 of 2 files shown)


mmv1/third_party/terraform/tpgresource/utils_test.go
```diff
@@ -1079,6 +1079,32 @@ func TestReplaceVars(t *testing.T) {
...
 			Expected: "projects/default-project/zones/default-zone/instances",
 		},
+		"location with provider level region": {
+			Template: "projects/{{project}}/locations/{{location}}/repositories",
+			Config: &transport_tpg.Config{
+				Project: "default-project",
+				Region:  "default-region",
+			},
+			Expected: "projects/default-project/locations/default-region/repositories",
+		},
```
wyardley, line 1089: q: can these unit tests be run directly from within magic modules (and are they run on any of the CI runs in this project)? I think maybe the provider `build-and-unit-test` steps only test the unit tests within the provider itself?

The only way I could seem to get them to run, whether with `make` or `go test` was first building the provider and then running `go test -v ./google/tpgresource` from the base of the repo... got import issues building them directly within `magic-modules`

wyardley:
I'll take a look at the TGC unit test failures later as well. But would be good to get some input from whoever has some institutional knowledge about this area of the codebase.

mmv1/third_party/terraform/tpgresource/utils_test.go
zli82016: They cannot be run directly from magic modules, unfortunately. 

`build-and-unit-test` is run in the generated provider.

wyardley: Thanks. And I think from the logs, it doesn’t run these tests, maybe, right?

zli82016: The tests should be run in tpgresource folder https://github.com/GoogleCloudPlatform/magic-modules/actions/runs/11824442763/job/32945960561

wyardley: Got it, yeah, I looked there, just don't see anything if I search the logs for `tpgresource`, but could just be missing it.

I ran it locally anyway, was just curious if they really ran in CI somewhere too.

zli82016:
It becomes much more complicated then the original issue and some other tests failed. Do you want to just fix the import issue for `google_artifact_registry_repository` instead of modifying the `tpgresource.ReplaceVars` function?

wyardley:
Unless it's something very trivially fixable, I'm probably going to move this one to draft soon., but if you have ideas (or if there are folks you can pull in who have ideas) about generally speaking whether this is a good idea, and how it should work), would appreciate the comments.

I did take a look at some of the tests, and at least the cloudbuild ones seem to fail for me off of `main` as well, though I don't see an open ticket for them. The bigquery ones seem a bit easier to reproduce... for example, `TestAccBigQueryJob_bigqueryJobCopyExample`. High level, seems like
`location=US` is getting rewritten to `location=us-central` by `ReplaceVars`, maybe because  the`google_bigquery_job` itself doesn't have `location` set, and `US` is the default for bigquery specifically.

wyardley:
@zli82016 I'm definitely going to punt on this for now.

I'm not sure if there are any easy fixes for the `google_artifact_registry_repository` thing that are worth it -- I could switch to custom code for the import function, and use something other than `ReplaceVars()` in it, but it still wouldn't be easy to fix the problem for multi-region repositories anyway, so I'm inclined to PR some of my docs fixes separately (and maybe remove `{{repository_id}}` as a valid import path entirely?). I may leave this open as a draft for a few weeks in case anyone has thoughts / comments about how it "should" work.

zli82016:
> I did take a look at some of the tests, and at least the cloudbuild ones seem to fail for me off of `main` as well, though I don't see an open ticket for them. The bigquery ones seem a bit easier to reproduce... for example, `TestAccBigQueryJob_bigqueryJobCopyExample`. High level, seems like `location=US` is getting rewritten to `location=us-central` by `ReplaceVars`, maybe because the`google_bigquery_job` itself doesn't have `location` set, and `US` is the default for bigquery specifically.

That is the reason for the tests failure in `google_bigquery_job`

Also, I checked the resource `google_cloudbuild_trigger` and there is a [custom code](https://github.com/hashicorp/terraform-provider-google-beta/blob/main/google-beta/services/cloudbuild/resource_cloudbuild_trigger.go#L1995)  for the `location` field in import function.

zli82016:
I talked with out team. As `location` can have different namespaces for different resources, it doesn't make sense to set the default location in the provider level.

Also, it is a breaking change to  remove one specific import format.

wyardley:
There’s a lot of existing precedent for inferring location from provider in the codebase already

Will comment in the other PR, but since the single import format doesn’t actually work in any cases, is it actually breaking to remove it?

PR #12323, opened 2024-11-13
Made traffic_split field use default_from_api

melinath:
Testing behavior of `"{}"`: gpaste/6027469969948672
Testing behavior of `"{}"` if there's already a value set remotely: gpaste/4841573379735552

```release-note:bug
vertexai: fixed issue with google_vertex_ai_endpoint where upgrading to 6.11.0 would delete all traffic splits that were set outside Terraform (which was previously a required step for all meaningful use of this resource).
```


slevenick (approved):
LGTM in general, but did you test specifically unsetting this field by sending the {} value?

In the paste you linked I see sending `trafficSplit: {}` but I don't see that trafficSplit was present with a value before that is sent. I'd be curious if the PATCH actually unsets the field or if it is just ignored because it has an empty value for it

melinath:
yeah, I wasn't able to _set_ the field yesterday - going to test those cases this morning :-)

melinath:
yeah it works: gpaste/4841573379735552

melinath:
aaand I need to approve my own builds now too 🤦

PR #12334, opened 2024-11-15
Fix Failing test TestAccSecurityCenterV2ProjectBigQueryExportConfig_basic + others

vijaykanthm:
This PR aims to fix the TestAccSecurityCenterV2ProjectBigQueryExportConfig_basic + others failing issue, raised in #[20003](https://github.com/hashicorp/terraform-provider-google/issues/20003) 
Bug link: [375653701](https://b.corp.google.com/issues/375653701)

```release-note:none
securitycenterv2: fixed flaky test TestAccSecurityCenterV2ProjectBigQueryExportConfig_basic
```


mmv1/third_party/terraform/services/securitycenter/resource_scc_organization_big_query_export_config_test.go
```diff
@@ -1,25 +1,33 @@
 package securitycenter_test
 
 import (
+	"context"
 	"fmt"
+	"log"
+	"strings"
 	"testing"
 
+	"cloud.google.com/go/bigquery"
```
rileykarson, line 10: Please use https://github.com/googleapis/google-api-go-client instead

mmv1/third_party/terraform/services/securitycenter/resource_scc_organization_big_query_export_config_test.go
```diff
@@ -13,13 +18,21 @@ func TestAccSecurityCenterOrganizationBigQueryExportConfig_basic(t *testing.T) {
 	t.Parallel()
 
 	randomSuffix := acctest.RandString(t, 10)
-	dataset_id := "tf_test_" + randomSuffix
+	datasetID := "tf_test_" + randomSuffix
 	orgID := envvar.GetTestOrgFromEnv(t)
 
+	// Run cleanup before the test starts
```
rileykarson, line 24: Why do we need this? We run sweeper jobs independently. This could cause issues with other tests running at the same time.

mmv1/third_party/terraform/services/securitycenter/resource_scc_organization_big_query_export_config_test.go
```diff
@@ -13,13 +18,21 @@ func TestAccSecurityCenterOrganizationBigQueryExportConfig_basic(t *testing.T) {
...
+	ctx := context.Background()
+	projectID := envvar.GetTestProjectFromEnv()
+	err := cleanupOrganizationBigQueryDatasets(ctx, "tf_test_", projectID)
```
rileykarson, line 27: This seems to be project-scoped, not org scoped?

mmv1/third_party/terraform/services/securitycenter/resource_scc_organization_big_query_export_config_test.go
```diff
@@ -34,7 +47,8 @@ func TestAccSecurityCenterOrganizationBigQueryExportConfig_basic(t *testing.T) {
 		},
 		Steps: []resource.TestStep{
 			{
-				Config: testAccSecurityCenterOrganizationBigQueryExportConfig_basic(context),
+				Config:  testAccSecurityCenterOrganizationBigQueryExportConfig_basic(context),
+				Destroy: true,
```
rileykarson, line 51: What is this line meant to be doing? I suspect that's why we're failing the tests- we're removing all the resources from state and therefore fail to find them.

mmv1/third_party/terraform/services/securitycenter/resource_scc_organization_big_query_export_config_test.go
vijaykanthm: This is to prevent quota issues, clutter, and resource conflicts in test runs.

mmv1/third_party/terraform/services/securitycenter/resource_scc_organization_big_query_export_config_test.go
vijaykanthm: BigQuery datasets are project-scoped, even though the test configures an organization-level, the actual export targets a BigQuery dataset, which is scoped to a specific project (not the organization). Therefore, managing datasets for cleanup requires a project ID.

mmv1/third_party/terraform/services/securitycenter/resource_scc_organization_big_query_export_config_test.go
rileykarson: Let's solve that independently- we'd want the sweepers to work correctly rather than add a separate process. If this runs at the same time as BQ tests are running we'd run into errors as we delete the resources under test.

rileykarson: That means we should be fine to remove this block.

mmv1/third_party/terraform/services/securitycenter/resource_scc_organization_big_query_export_config_test.go
rileykarson: Per above- let's remove this action.

mmv1/third_party/terraform/services/securitycenter/resource_scc_organization_big_query_export_config_test.go
vijaykanthm: Addressed. Thank you for the suggestion, this is causing the issue.

rileykarson (approved):
LGTM, thanks!

rileykarson:
tgc failure looks unrelated.

PR #12338, opened 2024-11-15
Add Code Repository Index resource

JumiDeluxe:

Add new resources for Gemini Code Assist

```release-note:new-resource
`google_gemini_code_repository_index`
```


melinath:
Reassigning this review

melinath:
Updated the assignee in the assignment comment to @SirGitsalot so that future reminders will be directed properly.

mmv1/products/cloudaicompanion/CodeRepositoryIndex.yaml
```diff
@@ -0,0 +1,113 @@
...
+  error:
+    path: error
+    message: message
+  include_project: false
+custom_code:
+  custom_update: templates/terraform/custom_update/code_repository_index.go.tmpl
+parameters:
+  - name: location
+    type: String
+    description: Resource ID segment making up resource `name`. It identifies the resource within its parent collection as described in https://google.aip.dev/122.
```
SirGitsalot, line 62: nit: The description ends up in the Terraform documentation, so it should be use friendly - it can be terse though, so something like:

```
description: The location of the code repository index, for example `us-central1`.
```

mmv1/products/cloudaicompanion/CodeRepositoryIndex.yaml
```diff
@@ -0,0 +1,113 @@
...
+    min_version: 'beta'
+    immutable: true
+    url_param_only: true
+    required: true
+  - name: codeRepositoryIndexId
+    type: String
+    description: Required. Id of the requesting object
```
SirGitsalot, line 69: Ditto on the documentation, something like:

```
description: The ID of the code repository index.
```

SirGitsalot (changes requested):

mmv1/third_party/terraform/acctest/bootstrap_test_utils.go.tmpl
```diff
@@ -1435,6 +1439,89 @@ func SetupProjectsAndGetAccessToken(org, billing, pid, service string, config *t
...
+	//
+	// BootstrapSharedCodeRepositoryIndex returns a persistent code repository index
+	// for a test or set of tests.
+	//
+	// Deletion of code repository index takes a few minutes, and creation of it
+	// currently takes about half an hour.
+	// That is the reason to use the shared code repository indexes for test resources.
+	const SharedCodeRepositoryIndexPrefix = "tf-bootstrap-cri-"
+
+	func BootstrapSharedCodeRepositoryIndex(t *testing.T, codeRepositoryIndexId, location, kmsKey string, labels map[string]string) string {
```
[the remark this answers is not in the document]
JumiDeluxe: Currently not used, it was created for another resource for which I will create a PR after that. I can remove it for now but it will certainly be used.

mmv1/products/cloudaicompanion/CodeRepositoryIndex.yaml
JumiDeluxe: Updated

melinath:
Drive-by comment: It looks like there's inconsistency in terms of calling the service `gemini` vs `cloudaicompanion`. Generally speaking we recommend using the API subdomain (`cloudaicompanion`) because it is not mutable - but if you use gemini instead, you should use it consistently to avoid confusion. So for example, the `products` folder should also be called `gemini`.

You also don't need to have both `gemini` and `cloudaicompanion` listed in the teamcity services - only the one that you end up choosing.

JumiDeluxe:
I've renamed services for consistency by following the resources naming

SirGitsalot (approved):

PR #12339, opened 2024-11-15
compute: added `numeric_id` to `google_compute_network` data source

wyardley:
Add `numeric_id` to `google_compute_network` data source, matching the resource.

Switch `numeric_id` to to be an integer internally within the `google_compute_network` resource for consistency with the API and with `google_compute_subnetwork`.

Followup to #12285
Part of hashicorp/terraform-provider-google#20223

```release-note:enhancement
compute: added `numeric_id` to the `google_compute_network` data source
```

```release-note:enhancement
compute: switched the `numeric_id` attribute of `google_compute_network` to be an integer, matching the API behavior.
```


mmv1/products/compute/Network.yaml
```diff
@@ -124,7 +124,7 @@ properties:
     validation:
       function: 'verify.ValidateGCEName'
   - name: 'numericId'
-    type: String
+    type: Integer
```
rileykarson, line 127: There are some corner cases to do with modules, unfortunately, that mean these corrections are breaking :(

It's probably best to add this as a string in the datasource as well for consistency, and we can track flipping the types in the next major release.

wyardley: Ok, I can update it. Just weird that network and subnetwork won’t be consistent with each other.

Might be good to keep track somewhere of all the fields that are string but should be int, in the board you have setup for breaking changes?

wyardley:
Hi @rileykarson - does this look right now?
See also the new linked issue I created.

rileykarson (commented):
That looks right! Running tests. We could also add `network_id` at-will, which could be a proper int, and deprecate `numeric_id` for it.

Honestly I'd probably advocate for just keeping `numeric_id` deprecated but around for a _while_ as the cost to keep it is low and I'd prefer less churn, although that trades off some clarity for new users.

rileykarson (approved):

PR #12341, opened 2024-11-15
Add tgc_v6 provider

zli82016:

```release-note:none

```


GNUmakefile
```diff
@@ -90,6 +90,7 @@ tgc:
 	cd mmv1;\
 		go run . --version beta --provider tgc --output $(OUTPUT_PATH)/tfplan2cai $(mmv1_compile);\
 		go run . --version beta --provider tgc_cai2hcl --output $(OUTPUT_PATH)/cai2hcl $(mmv1_compile);\
+		go run . --version beta --provider tgc_v6 --output $(OUTPUT_PATH)/v6/pkg $(mmv1_compile);\
```
wyardley, line 93: just a note -- may want to see if chaining them this way is evergreen if the first step fails... see #12314

I think it may work to do `&&` between them instead of `;`, and avoid a situation where the first or second item fails, but the whole make process exits 0.

zli82016: It makes sense to chain them with `&&`.

c2thorn (approved):
LGTM

PR #12346, opened 2024-11-16
artifactregistry: updated repository docs and import format

wyardley:
It turns out that fixing importing regional or multi-region artifact repository resources is challenging. Regional ones could eventually be inferred based on the provider location, if ReplaceVars were updated and / or custom code were used, but multi-region ones will likely not be possible, because (in my understanding), we can't look at the actual resource's defined location when importing, and while we could _maybe_ infer regional location from provider region, the multi-regional path has multiple options, so I don't think there's a way to guess the multi-region location -- for some resources, they'll fall back to global (or a service default, like `US` for Bigquery) if `location` isn't already present or inferrable from the provider, but in this case, doesn't seem safe.

So, this is basically just docs fixes and one other tweak (see my comment about properties vs. parameters).

See description and discussion in #12320, as well as the linked issue, for more on this.

* Update import format to `{{location}}/{{repository_id}}` or the longer version of the same; ~standardize to `repository_id` for the IAM resource~, and add more documentation about locations and what locations are valid.

* Also add a multi-region example test case.

Part of hashicorp/terraform-provider-google#19266

```release-note:none

```

### DIFF
--- a/mmv1/products/artifactregistry/Repository.yaml
+++ b/mmv1/products/artifactregistry/Repository.yaml
@@ -27,7 +27,6 @@ update_verb: 'PATCH'
 update_mask: true
 import_format:
   - 'projects/{{project}}/locations/{{location}}/repositories/{{repository_id}}'
-  - '{{location}}/{{repository_id}}'
 timeouts:
   insert_minutes: 20
   update_minutes: 20
--- a/mmv1/products/artifactregistry/Repository.yaml
+++ b/mmv1/products/artifactregistry/Repository.yaml
@@ -27,7 +27,7 @@ update_verb: 'PATCH'
 update_mask: true
 import_format:
   - 'projects/{{project}}/locations/{{location}}/repositories/{{repository_id}}'
-  - '{{repository_id}}'
+  - '{{location}}/{{repository_id}}'
 timeouts:
   insert_minutes: 20
   update_minutes: 20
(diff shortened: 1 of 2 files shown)


mmv1/products/artifactregistry/Repository.yaml
```diff
@@ -209,15 +215,6 @@ examples:
...
-properties:
-  - name: 'name'
-    type: String
-    description: |-
-      The name of the repository, for example:
-      "repo1"
-    output: true
-    custom_flatten: 'templates/terraform/custom_flatten/name_from_self_link.tmpl'
-    custom_expand: 'templates/terraform/custom_expand/shortname_to_url.go.tmpl'
   - name: 'repository_id'
```
wyardley, line 218: Based on the MMV1 docs, I _think_ `repository_id` and `location` should be parameters, not properties? Previously, everything was under `properties`.

wyardley:
@trodge with the IAM import stuff -- I get now why it has to be `repository` vs `repository_id`. I think the test should now pass with the updated code.

I verified locally that removing that line "fixes" the problem (by disallowing bare imports)

Without the change, it gives a confusing permissions issue when the location isn't given or inferred:
```
% terraform import google_artifact_registry_repository_iam_member.member "test-me roles/artifactregistry.reader user:xxx@gmail.com"
google_artifact_registry_repository_iam_member.member: Importing from ID "test-me  roles/artifactregistry.reader user:xxxx@gmail.com"...
╷
│ Error: Error retrieving IAM policy for artifactregistry repository "projects/xxx/locations/us-central1/repositories/test-me": googleapi: Error 403: The caller does not have permission
```

wyardley:
Not sure if this needs a changelog entry if accepted. Feel free to edit if so.

mmv1/products/artifactregistry/Repository.yaml
```diff
@@ -27,7 +27,6 @@ update_verb: 'PATCH'
 update_mask: true
 import_format:
   - 'projects/{{project}}/locations/{{location}}/repositories/{{repository_id}}'
-  - '{{repository_id}}'
```
wyardley, line 30: note: I originally had `{{location}}/{{repository_id}}` here explicitly, but from what I could see, what's on line 29 seems to already make this work automagically.

trodge (approved):

PR #12353, opened 2024-11-18
Support V6E for type of accelerator_config in google_tpu_v2_vm

zeleena:
```release-note:none
tpuv2: changed `accelerator_config.tpye` field from enum to string in `google_tpu_v2_vm` resource (beta)
```

### DIFF
--- a/mmv1/products/tpuv2/Vm.yaml
+++ b/mmv1/products/tpuv2/Vm.yaml
@@ -264,18 +264,12 @@ properties:
       - accelerator_type
     properties:
       - name: 'type'
-        type: Enum
+        type: String
         description: |
-          Type of TPU.
+          Type of TPU. Please select one of the allowed types: https://cloud.google.com/tpu/docs/reference/rest/v2/AcceleratorConfig#Type
         min_version: 'beta'
         required: true
         immutable: true
-        enum_values:
-          - 'V2'
-          - 'V3'
-          - 'V4'
-          - 'V5P'
-          - 'V6E'
       - name: 'topology'
         type: String
         description: |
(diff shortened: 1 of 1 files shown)


shuyama1 (commented):
The change looks good overall! I just have a quick question: the value doesn’t appear in the API documentation—has support for this value already rolled out?

Would you mind either adding a new test or updating an existing test to include this value? If you’ve already tested it locally and confirmed that it works, that should be fine as well.

https://googlecloudplatform.github.io/magic-modules/develop/test/test/ and https://googlecloudplatform.github.io/magic-modules/develop/test/run-tests/ in case it helps.

zeleena:
Thank you, this value is in Public Preview API (alpha for us) but not GA, and no changes to the API are expected between Public Preview and GA. According to Terraform Launch Stages, it looks like this can be added to `google-beta` until GA API Launch and should be added around now in order to meet deadlines.

V6E is very capacity constrained so adding a new automated test would result in a lot of stockout errors. I did set up a manual test, but tpuv2 uses the GA API in `google-beta` since we do not have a beta API. I verified that the correct request gets sent to the control plane, but actual creation would only succeed once this feature reaches GA, and we want to have it ready by GA.

zeleena:
Like we discussed offline, I changed the type of accelerator_config from Enum to String to be more robust against additions of more types in the future and linked the documentation of types that can be selected. Please let me know what you think.

I tested this change manually using the Basic Example modified to use accelerator config, and changed a field which lead to more capacity and a successful test.

shuyama1 (approved):
LGTM! Thank you!

PR #12355, opened 2024-11-19
fix: Router custom-learned-route-priority undefined behavior

akshat-jindal-nit:
Bug: b/378841773
Issue: custom-learned-route-priority is an optional field .
However, TF would set custom-learned-route-priority to be 0 even if the user has not explicitly set the value to be 0 when there is an update to the resource.

Repro steps:

Create a Router peer resource without custom-learned-route-priority value set.
```
resource "google_compute_router_peer" "peer" {
  name                      = "tf-test-my-router-peer-1"
  router                    = google_compute_router.router.name
  region                    = google_compute_router.router.region
  interface                 = "interface-1"
  peer_asn                  = 65513
  advertise_mode            = "CUSTOM"
}
```
Query gcloud to check the value of custom-learned-route-priority, it will be empty
gcloud compute routers describe {router-name} 

Update the router peer resource (ex: advertised_route_priority = 100) without acustom-learned-route-priority. However, TF would add custom-learned-route-priority = 0 in the update api call.
```
resource "google_compute_router_peer" "peer" {
  name                      = "tf-test-my-router-peer-1"
  router                    = google_compute_router.router.name
  region                    = google_compute_router.router.region
  interface                 = "interface-1"
  peer_asn                  = 65513
  advertised_route_priority = 100
  advertise_mode            = "CUSTOM"
}
```
Query gcloud to check the value of custom-learned-route-priority, it will be 0 even though it is not set by the user.
gcloud compute routers describe {router-name} 

***Release Note Template for Downstream PRs (will be copied)***
```release-note:enhancement
compute: fixed a issue where `custom_learned_route_priority` was accidentally set to 0 during updates in 'google_compute_router_peer'
```
```release-note:enhancement
bug: added support for setting `custom_learned_route_priority` to 0 in 'google_compute_router_peer' by adding the `zero_custom_learned_route_priority` field
```


akshat-jindal-nit:
We see this type of similar issue for  advertised_route_policy , @harshithpatte-g made that fix https://github.com/GoogleCloudPlatform/magic-modules/pull/11613.

harshithpatte-g:
LGTM.

slevenick (changes requested):
Can you add an update test that tests this behavior specifically?

akshat-jindal-nit:
> Can you add an update test that tests this behavior specifically?

I tested it manually by doing a gcloud query for the repro steps. We can't write a test or add into an existing test.
I have tried different permutations ,
update the existing resources without custom_learned_route_priority with the fix.
update the existing resources with custom_learned_route_priority with the fix.
Create a new resource with a fix.
Create a new resource with custom_learned_route_priority with a fix.
Remove custom_learned_route_priority from an existing resource with custom_learned_route_priority.
and so on.

slevenick:
> > Can you add an update test that tests this behavior specifically?
> 
> I tested it manually by doing a gcloud query for the repro steps. We can't write a test or add into an existing test. I have tried different permutations , update the existing resources without custom_learned_route_priority with the fix. update the existing resources with custom_learned_route_priority with the fix. Create a new resource with a fix. Create a new resource with custom_learned_route_priority with a fix. Remove custom_learned_route_priority from an existing resource with custom_learned_route_priority. and so on.

We should be able to write test check steps to validate that a field is set to a certain value after a config has been applied.

That's what we should do here, apply a config with no value and check that it doesn't have one, then apply a config with the 0 value and check that it gets applied, etc.

Look for occurrences of `resource.TestCheckResourceAttr` for examples of this

slevenick (changes requested):
See comment above

akshat-jindal-nit:
> slevenick

I think we can't write the automate test as terraform consider empty or null int field value as 0 .
FYI
I manually check it using these steps https://googlecloudplatform.github.io/magic-modules/develop/test/run-tests/#optional-test-manually.
Here we seeing the field is not set at all when we use the gcloud by fetching out the details of that particular resource.

slevenick:
> > slevenick
> 
> I think we can't write the automate test as terraform consider empty or null int field value as 0 . FYI I manually check it using these steps https://googlecloudplatform.github.io/magic-modules/develop/test/run-tests/#optional-test-manually. Here we seeing the field is not set at all when we use the gcloud by fetching out the details of that particular resource.

We should be able to write a test that sets the field to 0. And then we should be able to verify that Terraform correctly set the value to 0 in the API. Can you add a test that does so?

slevenick:
> @slevenick
> 
> 1. When the customer explicitly sets custom_learned_route_prioirty as 0. TF would consider this as no change and do not make an api call.
> 2. (The important issue) Without explicitly setting custom_learned_route_prioirty as 0, any update to resource like  advertised_route_priority would make a patch call with custom_learned_route_prioirty = 0 (side effect) along with the change (advertised_route_priority = true)
> 
> `So here we go with solution 2. To use ResourceData.getOk() instead of deprecated ResourceData.getOkExists(). Because the data type is an integer, empty value(nil) is always set as 0. ResourceData.getOk() will consider 0 as resource value doesn't exist.`
> 
> Solution to point 1. We don't have it currently as TF SDK doesn't support nil for integer type. The scope of point 1 is greater than just this field. That's why writing automated test is not possible in this case.

Sorry, I think I'm not stating my question well enough.

Is 0 a valid value in the API for this field?

If so, can the user set that value via their Terraform resource?

akshat-jindal-nit:
> Is 0 a valid value in the API for this field?

Yes 0 is the valid value for this particular field `custom_learned_route_priority` image

akshat-jindal-nit:
**The Problem:**

**Before the fix**: The custom_learned_route_priority field was unintentionally being set to 0 whenever any other field was updated. This is bad because it could overwrite a user's intentional configuration and they might not notice.
**After the fix**: The field no longer defaults to 0 on updates, but Terraform can't set it to 0  either . This means users must use gcloud or the Pantheon UI to set it to 0 if needed.

***Why the second fix is better (but not perfect):***
While not ideal, the second fix is preferable because it avoids unexpected behavior and data loss.  Forcing users to explicitly set the field to 0 through other tools ensures they are aware of this specific configuration.

cc: @harshithpatte-g 
@slevenick what's your point of view on this situation.

mmv1/third_party/terraform/services/compute/resource_compute_router_peer.go.tmpl
```diff
@@ -165,7 +165,12 @@ CIDR-formatted string.`,
 This value is applied to all custom learned route ranges for the session. You can choose a value
 from 0 to 65335. If you don't provide a value, Google Cloud assigns a priority of 100 to the ranges.`,
                         },
-
+			"forced_zero_custom_learned_route_priority": {
```
slevenick, line 168: Lets remove "forced" here.

mmv1/third_party/terraform/services/compute/resource_compute_router_bgp_peer_test.go.tmpl
```diff
@@ -248,6 +248,30 @@ func TestAccComputeRouterPeer_Ipv4BasicCreateUpdate(t *testing.T) {
 	})
 }
 
+func TestAccComputeRouterPeer_UpdateRouterCustomLearnedRoutePriority(t *testing.T) {
```
slevenick, line 251: Can you add a step to this test that sets the priority to 0 but sets the boolean to false? We want to make sure that functionality works too.

slevenick:
Looks like we're going to need to add ignore statements on the ImportStateVerify for the failing tests. See `ImportStateVerifyIgnore` in various files

mmv1/third_party/terraform/services/compute/resource_compute_router_bgp_peer_test.go.tmpl
akshat-jindal-nit: i change the debug statement to error and it will be decided by customer which they have to assign it .

mmv1/third_party/terraform/services/compute/resource_compute_router_peer.go.tmpl
akshat-jindal-nit: done

mmv1/third_party/terraform/services/compute/resource_compute_router_bgp_peer_test.go.tmpl
slevenick: Ok great. What happens when the boolean is false/unset and the field is unset? I'm worried this will effectively require the user to set the field.

We should make sure that case works

mmv1/third_party/terraform/services/compute/resource_compute_router_peer.go.tmpl
```diff
@@ -165,7 +165,17 @@ CIDR-formatted string.`,
 This value is applied to all custom learned route ranges for the session. You can choose a value
 from 0 to 65335. If you don't provide a value, Google Cloud assigns a priority of 100 to the ranges.`,
                         },
-
+			"zero_custom_learned_route_priority": {
```
slevenick, line 168: We will need to add this field to the docs at: https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/third_party/terraform/website/docs/r/compute_router_peer.html.markdown

mmv1/third_party/terraform/services/compute/resource_compute_router_bgp_peer_test.go.tmpl
```diff
@@ -1001,6 +1047,42 @@ func testAccComputeRouterPeerWithMd5AuthKeyUpdate(routerName string) string {
 		routerName, routerName)
 }
 
+func testAccComputeRouterPeerCustomLearnedRoutePriority(routerName string, customLearnedRoutePriority int, zeroCustomLearnedRoutePriroity bool) string {
```
slevenick, line 1050:
```suggestion
func testAccComputeRouterPeerCustomLearnedRoutePriority(routerName string, customLearnedRoutePriority int, zeroCustomLearnedRoutePriority bool) string {
```

slevenick (changes requested):

mmv1/third_party/terraform/services/compute/resource_compute_router_peer.go.tmpl
akshat-jindal-nit: done

akshat-jindal-nit:
@slevenick  can you please re-run the test, there was some typo in it, I correct it.

slevenick (approved):

PR #12360, opened 2024-11-19
compute: Implement graceful switch for metadata_startup_script

Asiderr:
This patch checks whether a graceful switch (without ForceNew) is available between `metadata_startup_script` and `metadata.startup-script`. Graceful switch can be executed in two situations:
1. When `metadata_startup_script` is created with the old value of `metadata.startup-script`.
2. When `metadata_startup_script` is deleted and the old value remains in `metadata.startup-script`.

For all other changes in `metadata_startup_script`, `isGracefulMetadataStartupSwitch` sets ForceNew. 

The change is covered by:
`TestAccComputeInstance_metadataStartupScript_update` and `TestAccComputeInstance_metadataStartupScript_gracefulSwitch`

closes: https://github.com/hashicorp/terraform-provider-google/issues/9459

```release-note:enhancement
compute: made `metadata_startup_script` able to be updated via graceful switch in `google_compute_instance`
```

### DIFF
--- a/mmv1/third_party/terraform/services/compute/resource_compute_instance.go.tmpl
+++ b/mmv1/third_party/terraform/services/compute/resource_compute_instance.go.tmpl
@@ -815,7 +815,6 @@ func ResourceComputeInstance() *schema.Resource {
 			"metadata_startup_script": {
 				Type:        schema.TypeString,
 				Optional:    true,
-				ForceNew:    true,
 				Description: `Metadata startup scripts made available within the instance.`,
 			},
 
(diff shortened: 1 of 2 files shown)


c2thorn (commented):
Hi @Asiderr 
Did you test the Forcenew/regular update by hand?

Should we not get a new update test specifically for updating `metadata_startup_script` without Forcenew?

Asiderr:
Hi @c2thorn 
> Did you test the Forcenew/regular update by hand?

Yes, I tested all of the potential configurations and it works as expected.

> Should we not get a new update test specifically for updating metadata_startup_script without Forcenew?

I've added test `TestAccComputeInstance_metadataStartupScript_gracefulSwitch` that covers updating metadata_startup_script without Forcenew

c2thorn (approved):
LGTM thanks @Asiderr for the update test.
Logic looks good for the forcenew function and should be otherwise safe considering the previous logic was always ForceNew

PR #12368, opened 2024-11-20
fix: (storagetransfer) transfer path can be changed to empty string

gurusai-voleti:
Fixes: https://github.com/hashicorp/terraform-provider-google/issues/13673
Fixes: https://github.com/hashicorp/terraform-provider-google/issues/20421

```release-note:bug
storagetransfer: fix to allow empty string for path
```


kautikdk:
Please link the issue in PR description that you are resolving.

gurusai-voleti:
> 
> ## Diff report
> 
> `google` provider: [Diff](https://github.com/modular-magician/terraform-provider-google/compare/auto-pr-12368-old..auto-pr-12368) ( 2 files changed, 122 insertions(+), 4 deletions(-)) `google-beta` provider: [Diff](https://github.com/modular-magician/terraform-provider-google-beta/compare/auto-pr-12368-old..auto-pr-12368) ( 2 files changed, 122 insertions(+), 4 deletions(-))
> 
> ## Breaking Change(s) Detected
> The following breaking change(s) were detected within your pull request.
> 
> * Field `transfer_spec.gcs_data_sink.path` transitioned from optional+computed to optional `google_storage_transfer_job` - [reference](https://googlecloudplatform.github.io/magic-modules/develop/breaking-changes/breaking-changes#field-oc-to-c)
> * Field `transfer_spec.gcs_data_source.path` transitioned from optional+computed to optional `google_storage_transfer_job` - [reference](https://googlecloudplatform.github.io/magic-modules/develop/breaking-changes/breaking-changes#field-oc-to-c)
> 
> If you believe this detection to be incorrect please raise the concern with your reviewer. If you intend to make this change you will need to wait for a [major release](https://www.terraform.io/plugin/sdkv2/best-practices/versioning#example-major-number-increments) window. An `override-breaking-change` label can be added to allow merging.

Path is set as computed field which is user provided i see it setup incorrectly as a computed field, so removed it now and added validation as GCS bucket paths cannot be started with / and added diffsuppressfunc to revert to "" empty string as well

kautikdk:
Can you please briefly explain your solution and how it will allow users to set empty string in path field?

gurusai-voleti:
> Can you please briefly explain your solution and how it will allow users to set empty string in path field?

Path is an optional field, if not provided at first tf apply it is set to empty string,
but in the customer reported issue they used "/" first and later realized to change to "" empty string
which they cannot as in update d.HasChanges will not detect empty string as a change

so added DiffSuppressFunc if the old and new is same it will not show any difference in tf plan, if its different it will show in tf plan and path can be updated to "" empty string and added validation to the path field as gcs object path cannot be started with "/" (preference here https://cloud.google.com/storage-transfer/docs/reference/rest/v1/GcsData)

kautikdk:
This appears to be breaking change as we removing computed from optional+computed so I think this should be scheduled for upcoming major release?

gurusai-voleti:
> This appears to be breaking change as we removing computed from optional+computed so I think this should be scheduled for upcoming major release?

I think it can be scheduled to be in near major release

kautikdk:
If this is planned for major release then I think we should mark this PR draft or we can close it for now until guidelines for next major release is available.

kautikdk:
Hi @gurusai-voleti, Can you please make this PR draft and update internal ticket for the same? because we are not actively looking into this change until next major release which is not going to happen soon.

gurusai-voleti:
> Hi @gurusai-voleti, Can you please make this PR draft and update internal ticket for the same? because we are not actively looking into this change until next major release which is not going to happen soon.

updated on task and converted to draft PR

gurusai-voleti:
closing as this has been done 14377

PR #12371, opened 2024-11-20
Replace use of GetRegionFromSchema and standard field names with use of GetRegion

SarahFrench:
GetRegionFromSchema("region", "zone", d, config) is equivalent to GetRegion(d, config).

In this PR I'm trying to reduce direct usage of GetRegionFromSchema in resources so it's only used by resources who have non-standard location-type fields (for example google_redis_instance's use of [location_id](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/redis_instance#location_id-1)). 

```release-note:none

```

### DIFF
--- a/mmv1/templates/terraform/encoders/location_from_region.go.tmpl
+++ b/mmv1/templates/terraform/encoders/location_from_region.go.tmpl
@@ -12,7 +12,7 @@
 */ -}}
 config := meta.(*transport_tpg.Config)
 if _, ok := d.GetOk("location"); !ok {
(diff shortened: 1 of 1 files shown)


c2thorn (approved):
confirming it is indeed the same thing: https://github.com/GoogleCloudPlatform/magic-modules/blob/8e3d3583c474144f802bee0f7905d7a87ce51fed/mmv1/third_party/terraform/tpgresource/utils.go#L78

PR #12378, opened 2024-11-21
Ensure oneOf condition is honored when expanding the job configs

ahzaz:
Ensure oneOf condition is honored when expanding the job configuration for Hive, Pig, Spark-sql, and Presto. The change ensures that only one of `query_file_uri` or `query_list` is set when expanding job config.

Fixes: https://github.com/hashicorp/terraform-provider-google/issues/13278

```release-note:bug
dataproc: ensured oneOf condition is honored when expanding the job configuration for Hive, Pig, Spark-sql, and Presto in `google_dataproc_job`.
```


hao-nan-li (commented):
Is this change potentially causing any impact on the customers?

ahzaz:
> Is this change potentially causing any impact on the customers?

This should not cause any regression. Earlier config actually did not allow job submission with only query_file_uri set to it should unblock the customers having that issue. This explicitly does not allow setting both the fields when expanding which is expected from schema.

hao-nan-li (approved):

PR #12384, opened 2024-11-21
google_compute_subnetwork resource and data output externalIpv6Prefix

cmseal:
Fixes https://github.com/hashicorp/terraform-provider-google/issues/20424
(just uses `external_ipv6_prefix` vs `ipv6_cidr_range` attribute)

```release-note: enhancement
compute: data source `google_compute_subnetwork` now fetches `externalIpv6Prefix`
```

```release-note: enhancement
compute: resource `google_compute_subnetwork` now outputs `externalIpv6Prefix`
```

### DIFF
--- a/mmv1/products/compute/Subnetwork.yaml
+++ b/mmv1/products/compute/Subnetwork.yaml
@@ -438,6 +438,7 @@ properties:
     description: |
       The range of external IPv6 addresses that are owned by this subnetwork.
     default_from_api: true
+    output: true
   - name: 'allowSubnetCidrRoutesOverlap'
     type: Boolean
     description: |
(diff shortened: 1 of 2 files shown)


cmseal:
Apologies in advance - first time contributing, in an attempt to be helpful and solve my own issue!

shuyama1 (commented):
@cmseal I've commented in the ticket directly. If externalIpv6Prefix shows the correct value needed, can you try using `external_ipv6_prefix` in the config instead? Terraform resources typically map directly to the API and from the log, it looks like that `ipv6CidrRange` wasn't returned. According to the [API doc](https://cloud.google.com/compute/docs/reference/rest/v1/subnetworks) the field is marked for internal use, so it might be intentional that the API doesn't explicitly return it.

PR #12387, opened 2024-11-21
Fix flaky apigee tests

zli82016:

```release-note:none

```


mmv1/third_party/terraform/services/apigee/resource_apigee_environment_update_test.go
```diff
@@ -55,9 +58,16 @@ resource "google_project" "project" {
   deletion_policy = "DELETE"
 }
 
+# Needed for CI tests for permissions to propagate, should not be needed for actual usage
```
ScottSuarez, line 61: Is this true, that it's not needed for actual users? The more I see evidence like this the more I think we should bake in some delay to the creation of eventual consistency resources.

zli82016: The time sleep could be needed for the actual users.

It looks like this is a known issue in API side, b/228371710

PR #12390, opened 2024-11-21
Add view with foreign definitions to google_bigquery_table beta

wj-chen:
Add view with foreign definitions to `google_bigquery_table` beta.

Regarding removing the default value for `use_legacy_sql` - The `foreign_definitions` is mutually exclusive with it in the backend so we can no longer have an unconditional default value. The default of `true` is now coded in at runtime when the new field is absent.

```release-note:enhancement
bigquery: added `foreign_definitions` fields to `google_bigquery_table resource (beta)
```


wj-chen:
@BBBmau Regarding removing the default value for `use_legacy_sql` regarded as a breaking change:

The new field is mutually exclusive with it in the backend so we can no longer have an unconditional default value. The default of `true` is now coded in at runtime when the new field is absent. Let me know if this is ok or you have suggestion on how to set up the relationship between these fields.

mmv1/third_party/terraform/services/bigquery/resource_bigquery_table.go.tmpl
```diff
@@ -2737,21 +2775,87 @@ func expandView(configured interface{}) *bigquery.ViewDefinition {
 	raw := configured.([]interface{})[0].(map[string]interface{})
 	vd := &bigquery.ViewDefinition{Query: raw["query"].(string)}
 
-	if v, ok := raw["use_legacy_sql"]; ok {
-		vd.UseLegacySql = v.(bool)
+	{{- if ne $.TargetVersionName "ga" }}
+	if v, ok := raw["foreign_definitions"]; ok && len(v.([]interface{})) > 0 {
```
BBBmau, line 2779: Having a field that goes from defaulting to true to then defaulting to true **ONLY** if a specific field is not set introduces some unexpected behavior to users. 

Although I understand your reasoning for this, it still wouldn't be possible to merge due to reasoning mentioned above. We'll want to merge this in the next Major Release. 

Any input on your end would be appreciated @c2thorn

c2thorn: What does the API default to? Can we just use the API's default instead of a provider default by setting `use_legacy_sql` to optional + computed. If there is no API default, why do we have a provider default in the first place?

Alternatively, can we have users explicitly set `use_legacy_sql` to false in order to use the new field? Or will the API _only_ accept no value?

We definitely should not have a value change in an expander where it will not show during plan-time _especially_ for a non-Computed value. If we can't do the suggestions above, the best way to set a provider-level conditional default is to make the field Optional+Computed and move the condition logic to a CustomizeDiff

c2thorn: It's relatively new and will need testing, but `d.GetRawConfig().GetAttr("field")` should get the actual config value. you may be able to see if it is equivalent to `cty.UnknownVal(cty.Bool)` to determine whether or not `use_legacy_sql` was set in config or unknown (left to an arbitrary computed value)

c2thorn: Searching `GetRawConfig()` should show some uses of it elsewhere in Magic Modules. You'll probably need to pass in the `d` resourcedata object to the expander or move the logic elsewhere

wj-chen: I updated `view.use_legacy_sql` to be Optional + Computed and cleaned up the related logic in the latest commit. Removing the default value is unavoidable though because the API only accepts no value for `use_legacy_sql` when `foreign_definitions` is set. Because Optional + Computed, existing configs shouldn't be impacted. Is there any way for this to be allowed in a minor release?

wj-chen: @BBBmau @c2thorn This PR is ready for review again.

wj-chen:
@BBBmau @c2thorn Looking for another round of review. Thank you!

mmv1/third_party/terraform/services/bigquery/resource_bigquery_table.go.tmpl
c2thorn: Sorry @wj-chen, I was OOO until yesterday and still catching up. 

So `use_legacy_sql` went from default true to O+C with custom logic to set/unset based on the TF config value.

With the code from https://github.com/GoogleCloudPlatform/magic-modules/pull/12390/files#diff-d9d1047fa09e43341e0a3dfc5af24525b5859851dd6eaf51888d50934952dd8aR2783, what happens to existing configs that use the previous default of true when they upgrade?

If I'm following the new logic from correctly, we will no longer send any value for `vd.UseLegacySql` where before the upgrade we were sending `true`. This would still be a breaking change if so.

Are you able to test this scenario to confirm?

c2thorn: > Because Optional + Computed, existing configs shouldn't be impacted. Is there any way for this to be allowed in a minor release?

If I'm following the logic correctly, existing configs won't be immediately affected because O+C, but the next time the expander runs (like in an update), the config value will be null and we will never send a value to the API which would unset the field on the backend?

c2thorn (commented):
left a review at https://github.com/GoogleCloudPlatform/magic-modules/pull/12390/files#r1878652623

mmv1/third_party/terraform/services/bigquery/resource_bigquery_table.go.tmpl
wj-chen: Just tested and there should be no impact to the existing users:

For users who has a config without `use_legacy_sql`, the current provider version sends `true` in the create request, and table gets created with `use_legacy_sql = true`.

Then after upgrading to using this code, `terraform apply` on the same config without change (i.e. still no `use_legacy_sql` specified) shows no diff. `terraform apply` on the config with change in another field (something not `use_legacy_sql`) shows a diff for only that change, and when run, that update is issued without `use_legacy_sql` in the request, and the API response leaves `use_legacy_sql = true` unchanged for the table.

The biggest impact with this change is that for new configs run with the new provider version, not specifying `use_legacy_sql` means not setting it to `true` unlike the current behavior. For users that want it to be true, they have to set it explicitly now. If this is ok, I will update the release note to call it out.

c2thorn: There's a chance that some users expect the default to be true for new resources and may be surprised that they would now need to set it. From their perspective, it would be a breaking change that they now need to update their configs to get the same resource creation behavior as the previous minor version.

Thanks for confirming that existing resources wouldn't be affected. That takes this from "definitely can't do this" to "probably can't do this" from my perspective. I still think we could get around this last part by implementing a CustomizeDiff to conditionally set the default of true if `foreign_definitions` is not set

c2thorn: Looks like we've done something similar here: https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/templates/terraform/constants/tpu_vm.go.tmpl#L33

c2thorn: I didn't see this earlier, but customizediff logic happens before/during plan which would allow users to see the modified value before applying. Expanders happen right before sending information to the API which would hide what Terraform is doing to the user

wj-chen:
@c2thorn Could you reopen this PR? I didn't see the inactivity warning....

And thanks for the comment, I will look into that once I close out on another PR that will likely cause a sync + conflict resolution on this branch.

mmv1/third_party/terraform/services/bigquery/resource_bigquery_table.go.tmpl
wj-chen: I tried out `CustomizeDiff`, but ran into an issue where the `SetNew` key should be `view.0.use_legacy_sql`, but `SetNew` doesn't support nested key path: https://github.com/hashicorp/terraform-plugin-sdk/blob/da29621efae69dd03dfdc540aa5f63beab1b764c/helper/schema/resource_diff.go#L294 (the last param is `false` so the method returns an invalid key error). I then tried to apply `SetNew` on the entire `view`, but the new error was "Error: SetNew only operates on computed keys - view is not one". I think it's not accurate to mark the entire `view` as `Computed`. Do you know of any other workaround that might help in this case?

c2thorn: spoke with @wj-chen offline and we came to the conclusion that it will be best just to change the default value of `use_legacy_sql` in a breaking change. Marking this PR as `7.0.0` until there is a proper major release feature branch.

wj-chen:
This change is NOT ready for review for 7.0.0. I'm just doing some merging right now.

c2thorn (commented):
noting that I'm about to be OOO for next week, and there is more to review and regain context on for this PR than I could get done this afternoon.

wj-chen:
> noting that I'm about to be OOO for next week, and there is more to review and regain context on for this PR than I could get done this afternoon.

Thank you for the heads up. I'll share a recap here after you are back the week after. Thank you for your support on this PR along the way.

wj-chen:
> > noting that I'm about to be OOO for next week, and there is more to review and regain context on for this PR than I could get done this afternoon.
> 
> Thank you for the heads up. I'll share a recap here after you are back the week after. Thank you for your support on this PR along the way.

@c2thorn Here is a recap of the context for this PR:

`google_bigquery_table` today has a `view` property and within `view` there are `query` and `useLegacySql`. This feature request requires adding another property `foreignDefinitions` inside `view`. `foreignDefinitions` is mutually exclusive with `query` and also with `useLegacySql`, otherwise the API will reject the request. This PR adds support for setting `foreignDefinitions` in this resource with the following changes:

* Add `view.foreignDefinitions`
* Change `view.query` from `required: true` to `optional: true`
* `view.useLegacySql` used to have a default value in TF which was a hard blocker for this feature for failing the API's mutual exclusivity check, so several options for workarounds were discussed in this PR at the time and the final decision was to remove the default value of `useLegacySql` from TF since the API itself doesn't have a default for that field, that breaking change was done in https://github.com/GoogleCloudPlatform/magic-modules/pull/14751 and released in 7.0.0.

Let me know if you have any questions.

PR #12398, opened 2024-11-22
Add diff suppress for computed scratch disks on Z3 instances

karolgorc:
closes https://github.com/hashicorp/terraform-provider-google/issues/19648

may be a breaking change, would love opinions here but CI tests should show if it's breaking.

Z3 instance type creates 12 scratch disks as a base for the instance. They cannot be edited, or deleted and this instance type cannot have more than 12 of these autogenerated local-ssd's so no new can be added. Hence the diff suppression

Not really sure if we want this kind of logic in terraform. The other solution would be to inform the user via docs that he has to add these disks to the configuration to get rid of this diff or use `lifecycle_rule.ignore_changes`. As these disks cannot be changed in any way but will show up as diff regardless

```release-note:enhancement
compute: `google_compute_instance` will no longer show unexpected difference in plan for z3 machine types
```

### DIFF
--- a/mmv1/third_party/terraform/services/compute/resource_compute_instance.go.tmpl
+++ b/mmv1/third_party/terraform/services/compute/resource_compute_instance.go.tmpl
@@ -1007,7 +1007,7 @@ be from 0 to 999,999,999 inclusive.`,
 				Type:        	  schema.TypeList,
 				Optional:    	  true,
 				ForceNew:    	  true,
-				DiffSuppressFunc: scratchDiskDiffSuppress
+				DiffSuppressFunc: scratchDiskDiffSuppress,
 				Description: 	  `The scratch disks attached to the instance.`,
 				Elem: &schema.Resource{
 					Schema: map[string]*schema.Schema{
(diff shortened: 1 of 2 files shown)


c2thorn (commented):
Thanks for the contribution @karolgorc, apologies on the delay here as I've been out of office.

If I'm understanding correctly, Z3 instances will automatically set scratch disks without the user needing to set themselves. Typically, this means that we should make the field Optional + Computed to let the value default to the API's value. 

The issue with doing that is that it would make it impossible to set the list to empty (for all instance types) since Terraform would perceive an empty list as the user wanting to default to the API's value. With how vital the compute instance resource is, I'm hesitant to make that change.

A diff suppress as you've implemented would hide the scratch disks from the resource state which is also not a great solution. Let me bring this up to my team and get back to you on a decision.

karolgorc:
Just to add some context from my testing

> Z3 instances will automatically set scratch disks without the user needing to set themselves

12 blocks of 3GB scratch disks to be exact

> A diff suppress as you've implemented would hide the scratch disks from the resource state which is also not a great solution

Yeah, i agree but keep in mind that these 12 disks cannot be updated in any way by the user. The API will block this behavior. So i'd say that either we hide them or tell the user in the docs that "With more specialized machine types you may have to use `lifecycle_rule.ignore_changes`"

c2thorn:
Closing as I don't think we'd want this behavior. I'd opt for a documentation change as you've suggested for now.

PR #12401, opened 2024-11-22
Remove LocationDescriber interface from google_client_config as it's unneeded

SarahFrench:
Context: I've shared a document with the Google Terraform team about why LocationDescriber was made and why I think it should be replaced.

Here in google_client_config it is unnecessary; there is no region or zone value provided via config to compare to provider-defaults as this resource is only for exposing provider default values. This PR removes use of the LocationDescriber interface and instead passes the values directly into the data source.

Post-muxing fixes the values coming from the provider when there isn't a provider-default will be `""`.
So, when no region or zone values are present:
* the `id` of the data source would be `projects/my-project/regions//zones/` (matching the data source when it was in SDK)
* the `region` will be `""`, not null (matching SDK)
* the `zone` will be `""`, not null (matching SDK)

The data source could be updated to make region and zone null if no provider-default value is present, but that's an explicit change in behaviour and is not in scope for this PR.

```release-note:none

```

### DIFF
--- a/mmv1/third_party/terraform/services/resourcemanager/data_source_google_client_config.go
+++ b/mmv1/third_party/terraform/services/resourcemanager/data_source_google_client_config.go
@@ -8,7 +8,6 @@ import (
 	"github.com/hashicorp/terraform-plugin-framework/datasource/schema"
 	"github.com/hashicorp/terraform-plugin-framework/types"
 	"github.com/hashicorp/terraform-provider-google/google/fwmodels"
-	"github.com/hashicorp/terraform-provider-google/google/fwresource"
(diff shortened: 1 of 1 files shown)


SarahFrench:
TeamCity tests are here: https://hashicorp.teamcity.com/buildConfiguration/TerraformProviders_GoogleCloud_GOOGLE_BETA_MMUPSTREAMTESTS_GOOGLEBETA_PACKAGE_RESOURCEMANAGER/272421?buildTab=overview&focusLine=NaN&hideTestsFromDependencies=false&hideProblemsFromDependencies=false&expandBuildChangesSection=true

SarahFrench:
Thanks for the review! I'll merge tomorrow

PR #12407, opened 2024-11-22
Add details about skipping tests in VCR, and when this is needed

SarahFrench:
Closes https://github.com/hashicorp/terraform-provider-google/issues/20059

```release-note:none

```


docs/content/develop/test/test.md
```diff
@@ -272,6 +272,51 @@ func TestSignatureAlgorithmDiffSuppress(t *testing.T) {
...
+| Bigtable acceptance tests aren't working in VCR mode |  `Requested interaction not found` errors during REPLAYING mode | Currently the provider uses a separate client than the rest of the provider to interact with the Bigtable API. As HTTP traffic to the Bigtable API doesn't go via the shared client it cannot be recorded in RECORDING mode.| Skip the test in VCR ofr Bigtable. |
+| Using multiple provider aliases doesn't work in VCR | You may have two instances of the google provider in the test config but one of them doesn't seem to be using its provider arguments, e.g. using the wrong default project. | See this GitHub issue: https://github.com/hashicorp/terraform-provider-google/issues/20019 . The problem is that, due to how the VCR system works, one provider instance will be configured and the other will be forced to reuse the other instance's configuration, despite them being given different provider arguments. |  Skip the test in VCR is using aliases is unavoidable. |
```
SarahFrench, line 285:
```suggestion
| Using multiple provider aliases doesn't work in VCR | You may have two instances of the google provider in the test config but one of them doesn't seem to be using its provider arguments, e.g. using the wrong default project. | See this GitHub issue: https://github.com/hashicorp/terraform-provider-google/issues/20019 . The problem is that, due to how the VCR system works, one provider instance will be configured and the other will be forced to reuse the first instance's configuration, despite them being given different provider arguments. |  Skip the test in VCR is using aliases is unavoidable. |
```

docs/content/develop/test/test.md
```diff
@@ -272,6 +272,51 @@ func TestSignatureAlgorithmDiffSuppress(t *testing.T) {
 }
 ```
 
+## Skipping tests in VCR mode
+
+As described above, acceptance tests can be run in VCR modes where HTTP requests are either recorded or are replayed. This system can fail in some scenarios, so we will oftern flag tests that are incompatible with VCR so that they're skipped. This means the tests will not run automatically on PRs (in VCR mode) but the tests will still run in our nightly test suite.
```
melinath, line 277:
```suggestion
Acceptance tests are run in VCR replaying mode on PRs (using pre-recorded HTTP requests and responses) to reduce the time it takes to present results to contributors. However, not all resources or tests are possible to run in replaying mode. Incompatible tests should be skipped during VCR replaying mode. They will still run in our nightly test suite.
```

docs/content/develop/test/test.md
```diff
@@ -272,6 +272,51 @@ func TestSignatureAlgorithmDiffSuppress(t *testing.T) {
...
+
+### When should tests be skipped in VCR mode?
```
melinath, line 279: style: [conceptual headings should use noun phrases](https://developers.google.com/style/headings)

```suggestion
### Reasons that tests are skipped in VCR replaying mode
```

docs/content/develop/test/test.md
```diff
@@ -272,6 +272,51 @@ func TestSignatureAlgorithmDiffSuppress(t *testing.T) {
...
+
+| Problem        | Possible symptoms you'll see    | How to fix/Other info  | Skip in VCR? |
```
melinath, line 281:
```suggestion
| Problem        | Possible symptoms you'll see    | How to fix/Other info  | Skip in VCR replaying? |
```

docs/content/develop/test/test.md
```diff
@@ -272,6 +272,51 @@ func TestSignatureAlgorithmDiffSuppress(t *testing.T) {
...
+| -------------- | ------------------------------- | ---------------------- |------------- |
+| Incorrect or insufficient data is present in VCR recordings to replay tests | Tests will fail with `Requested interaction not found` errors during REPLAYING mode | Make sure that you're not introducing randomness into the test, e.g. by unnecessarily using the random provider to set a resource's name.| If you cannot avoid this issue you should skip the test, but try to ensure that it cannot be fixed first.|
+| Bigtable acceptance tests aren't working in VCR mode |  `Requested interaction not found` errors during REPLAYING mode | Currently the provider uses a separate client than the rest of the provider to interact with the Bigtable API. As HTTP traffic to the Bigtable API doesn't go via the shared client it cannot be recorded in RECORDING mode.| Skip the test in VCR ofr Bigtable. |
```
melinath, line 284:
```suggestion
| Bigtable acceptance tests aren't working in VCR mode |  `Requested interaction not found` errors during REPLAYING mode | Currently the provider uses a separate client than the rest of the provider to interact with the Bigtable API. As HTTP traffic to the Bigtable API doesn't go via the shared client it cannot be recorded in RECORDING mode.| Skip the test in VCR for Bigtable. |
```

docs/content/develop/test/test.md
```diff
@@ -272,6 +272,51 @@ func TestSignatureAlgorithmDiffSuppress(t *testing.T) {
...
+| Using multiple provider aliases doesn't work in VCR | You may have two instances of the google provider in the test config but one of them doesn't seem to be using its provider arguments, e.g. using the wrong default project. | See this GitHub issue: https://github.com/hashicorp/terraform-provider-google/issues/20019 . The problem is that, due to how the VCR system works, one provider instance will be configured and the other will be forced to reuse the first instance's configuration, despite them being given different provider arguments. |  Skip the test in VCR is using aliases is unavoidable. |
```
melinath, line 285: style: [avoid e.g.](https://developers.google.com/style/word-list#eg)

```suggestion
| Using multiple provider aliases doesn't work in VCR | You may have two instances of the google provider in the test config but one of them doesn't seem to be using its provider arguments - for example, using the wrong default project. | See this GitHub issue: https://github.com/hashicorp/terraform-provider-google/issues/20019 . The problem is that, due to how the VCR system works, one provider instance will be configured and the other will be forced to reuse the first instance's configuration, despite them being given different provider arguments. |  Skip the test in VCR is using aliases is unavoidable. |
```

docs/content/develop/test/test.md
```diff
@@ -272,6 +272,51 @@ func TestSignatureAlgorithmDiffSuppress(t *testing.T) {
...
+| Using multiple versions of the google/google-beta provider in a single test isn't working in VCR  |  Unexpected test failures during REPLAYING mode in tests with `ExternalProviders` pulling in past versions of the google/google-beta provider | When ExternalProviders is used to pulling in other versions of the provider, any HTTP traffic through the external provider will not be recorded. If the HTTP traffic produces an unexpected result or returns an API error then the test will fail in REPLAYING mode. | Skip the test in VCR when testing the current provider behaviour versus previous released versions. |
+
+Some additional things to bear in mind are that VCR tests in REPLAYING mode will still interact with GCP APIs somewhat. For example:
+
+- When the provider is configured it will use credentials to obtain access tokens from GCP
+- Some acceptance tests use bootstrapping functions that ensure long-lived resources are present in a testing project before the provider is tested.
+
+Due to this you cannot use REPLAYING mode as a way to completely avoid HTTP traffic generally or with GCP APIs.
```
melinath, line 293:
```suggestion
These tests can still run in VCR replaying mode; however, REPLAYING mode can't be used as a way to completely avoid HTTP traffic generally or with GCP APIs.
```

docs/content/develop/test/test.md
```diff
@@ -272,6 +272,51 @@ func TestSignatureAlgorithmDiffSuppress(t *testing.T) {
...
+
+### How can tests be skipped in VCR mode?
```
melinath, line 295: style: [prefer imperative headings](https://developers.google.com/style/headings)

```suggestion
### Skip tests in VCR replaying mode
```

Since this duplicates the parent heading, I'd be inclined to merge this content into the parent heading (that is, above the "reasons" section)

docs/content/develop/test/test.md
```diff
@@ -272,6 +272,51 @@ func TestSignatureAlgorithmDiffSuppress(t *testing.T) {
...
+
+You can make generated and handwritten tests be skipped in VCR mode. If possible, please include a code comment explaining the reason for skipping, or link to a GitHub issue.
```
melinath, line 297:
```suggestion
If you skip a test in VCR mode, include a code comment explaining the reason for skipping (for example, a link to a GitHub issue.)
```

docs/content/develop/test/test.md
```diff
@@ -272,6 +272,51 @@ func TestSignatureAlgorithmDiffSuppress(t *testing.T) {
...
+```
+
+Skipping acceptance tests that are handwritten can be achieved by adding `acctest.SkipIfVcr(t)` at the start of the test:
+
+```go
+func TestAccPubsubTopic_update(t *testing.T) {
+      acctest.SkipIfVcr(t)
+      acctest.VcrTest(t, resource.TestCase{ ... }
+}
+```
```
melinath, line 317: This would be great to wrap in tabs like https://github.com/GoogleCloudPlatform/magic-modules/blob/1a21cf36eea49d191f99ed92d23dd1c10b7833e5/docs/content/develop/test/test.md?plain=1#L46

melinath (changes requested):

SarahFrench:
Thanks for the review, and the link to the developer docs style guide! I'll work through your comments tomorrow and this can be wrapped up after Thanksgiving 😁

SarahFrench:
Here's the new look with tabs and the similar sections being combined:

Screenshot 2024-12-02 at 18 15 39

A current issue is that the table is a bit too wide for the page, despite being displayed on a wide monitor:

Screenshot 2024-12-02 at 18 15 47

docs/content/develop/test/test.md
```diff
@@ -272,6 +272,59 @@ func TestSignatureAlgorithmDiffSuppress(t *testing.T) {
...
+   ```
+
+   If you skip a test in VCR mode, include a code comment explaining the reason for skipping (for example, a link to a GitHub issue.)
+
+   {{< /tab >}}
+{{< /tabs >}}
+
+### Reasons that tests are skipped in VCR replaying mode
+
+| Problem        | Possible symptoms you'll see    | How to fix/Other info  | Skip in VCR replaying? |
```
melinath, line 313: To fix the wideness problem, I wonder if you could reduce the number of columns? For example, combining Problem & Possible symptoms.

Alternatively, you could move this (and maybe https://googlecloudplatform.github.io/magic-modules/develop/test/run-tests/#common-errors too?) to a separate troubleshooting page... but I recognize that's a bigger change that we might want to handle separately later.

cc @nityaravi in case you have any suggestions here

melinath: :100: SGTM

melinath (approved):

SarahFrench:
Thanks!

PR #12411, opened 2024-11-22
Update provider reference docs on quota usage and always use billing_project as quota project if appropriately specified

ScottSuarez:
closes https://github.com/hashicorp/terraform-provider-google/issues/17882

b/360405077#comment8

```release-note:bug
provider: fixed issue where `GOOGLE_CLOUD_QUOTA_PROJECT` env var would override explicit billing_project
```


ScottSuarez:
Adding comment from bug for @SarahFrench 

So current behavior is we don't read quota-project from ADC at all.

First [they attempt to read from an env variable](https://github.com/golang/oauth2/blob/22134a41033e44c2cd074106770ab5b7ca910d15/google/default.go#L199-L206).

Then [they attempt to read](https://github.com/golang/oauth2/blob/22134a41033e44c2cd074106770ab5b7ca910d15/google/default.go#L208-L212
) from a 'well-known-file' aka [adc](https://github.com/golang/oauth2/blob/22134a41033e44c2cd074106770ab5b7ca910d15/google/default.go#L304).

then then fall back to GCE compute runtime credentials. 

The projectId you are seeing never correlates to QuotaProjectID. This is the [credential file's](https://github.com/golang/oauth2/blob/22134a41033e44c2cd074106770ab5b7ca910d15/google/default.go#L269) [projectID](https://github.com/golang/oauth2/blob/22134a41033e44c2cd074106770ab5b7ca910d15/google/google.go#L118) which is still different from the [quota project id](https://github.com/golang/oauth2/blob/22134a41033e44c2cd074106770ab5b7ca910d15/google/google.go#L136). I don't know the significance of the project id here. I suspect its the project that credentials were created from.

It's worth noting that the oauth2 library does pass back the [raw credential json bits](https://github.com/golang/oauth2/blob/22134a41033e44c2cd074106770ab5b7ca910d15/google/default.go#L264) which google-api library exposes an accessor of. Basically they [just index into the raw credential json](https://github.com/googleapis/google-api-go-client/blob/09fa125adbe59d02e1e67470bfeebcb6b9c1596e/internal/creds.go#L257-L260) to see if quota project is there as a fallback.

It's also worth noting that they [reference](https://github.com/googleapis/google-api-go-client/blob/09fa125adbe59d02e1e67470bfeebcb6b9c1596e/internal/creds.go#L248) a env variable for [quota project](https://github.com/googleapis/google-api-go-client/blob/09fa125adbe59d02e1e67470bfeebcb6b9c1596e/internal/creds.go#L27) `GOOGLE_CLOUD_QUOTA_PROJECT`. While we [reference](https://github.com/hashicorp/terraform-provider-google-beta/blob/71c72894f3f01bc2b2762fbfb96fe5773cc98e2f/google-beta/fwtransport/framework_config.go#L457) `GOOGLE_BILLING_PROJECT`.

To align with googleapis library and general naming convention I would recommend we sunset billing project terminology in favor of quota_project. Also to add `GOOGLE_CLOUD_QUOTA_PROJECT` env in addition to `GOOGLE_BILLING_PROJECT` to reference docs.


I think the only thing left I didn't cover is why we don't receive this quota project if google apis library has accessor functions for it. This is because the transport library we use to [create](https://github.com/hashicorp/terraform-provider-google-beta/blob/71c72894f3f01bc2b2762fbfb96fe5773cc98e2f/google-beta/fwtransport/framework_config.go#L1869) [our](https://github.com/googleapis/google-api-go-client/blob/09fa125adbe59d02e1e67470bfeebcb6b9c1596e/transport/dial.go#L24) [client](https://github.com/googleapis/google-api-go-client/blob/09fa125adbe59d02e1e67470bfeebcb6b9c1596e/transport/http/dial.go#L36) [transport](https://github.com/googleapis/google-api-go-client/blob/09fa125adbe59d02e1e67470bfeebcb6b9c1596e/transport/http/dial.go#L57) doesn't have the [rawjson in the credential object](https://github.com/googleapis/google-api-go-client/blob/09fa125adbe59d02e1e67470bfeebcb6b9c1596e/transport/http/dial.go#L182) available because [we only pass token source for credentials](https://github.com/hashicorp/terraform-provider-google-beta/blob/71c72894f3f01bc2b2762fbfb96fe5773cc98e2f/google-beta/fwtransport/framework_config.go#L1869) and their client library doesn't read the[full object](https://github.com/googleapis/google-api-go-client/blob/09fa125adbe59d02e1e67470bfeebcb6b9c1596e/internal/creds.go#L72). [GetQuotaProject needs the rawJSON](https://github.com/googleapis/google-api-go-client/blob/09fa125adbe59d02e1e67470bfeebcb6b9c1596e/transport/http/dial.go#L186) in the [credentials object](https://github.com/googleapis/google-api-go-client/blob/09fa125adbe59d02e1e67470bfeebcb6b9c1596e/internal/creds.go#L257) to parse the quota-project from the credential file.

TLDR
--
google apis client library is kind of funky and the way we interact with it doesn't auto get quota project parsing from the credential file. We should probably handle it ourselves anyways. There is currently undocumented behavior of `GOOGLE_CLOUD_QUOTA_PROJECT` being set causing their [transport layer](https://github.com/googleapis/google-api-go-client/blob/09fa125adbe59d02e1e67470bfeebcb6b9c1596e/internal/creds.go#L248) [to add](https://github.com/googleapis/google-api-go-client/blob/09fa125adbe59d02e1e67470bfeebcb6b9c1596e/transport/http/dial.go#L186) [the](https://github.com/googleapis/google-api-go-client/blob/09fa125adbe59d02e1e67470bfeebcb6b9c1596e/transport/http/dial.go#L105) `X-goog-user-project` header. We should probably document this.

ScottSuarez:
Created two child issues.
https://github.com/hashicorp/terraform-provider-google/issues/20463
https://github.com/hashicorp/terraform-provider-google/issues/20459

SarahFrench:
Just a heads up that some of the files affected by this PR have been deleted as part of the muxing fixes. Instead of parallel implementations of logic that handles credentials etc there are now only the original versions that date from when the provider was 100% SDK-implemented

ScottSuarez:
> Just a heads up that some of the files affected by this PR have been deleted as part of the muxing fixes. Instead of parallel implementations of logic that handles credentials etc there are now only the original versions that date from when the provider was 100% SDK-implemented

all good :)

mmv1/third_party/terraform/transport/config.go.tmpl
```diff
@@ -435,11 +435,19 @@ func (c *Config) LoadAndValidate(ctx context.Context) error {
 	}
 
 	c.TokenSource = tokenSource
-
 	cleanCtx := context.WithValue(ctx, oauth2.HTTPClient, cleanhttp.DefaultClient())
+	clientOptions := []option.ClientOption{option.WithTokenSource(tokenSource)}
+
+	// 'X-Goog-User-Project' gets set by a transport layer within transport.NewHTTPClient
+	// setting explicitly here to override the setting of it by underlying utility function.
```
rileykarson, line 442: I read the linked comment but don't quite understand what we're solving for here- could you reexplain it, sorry?

mmv1/third_party/terraform/transport/config.go.tmpl
```diff
@@ -468,6 +476,8 @@ func (c *Config) LoadAndValidate(ctx context.Context) error {
 
 	// Ensure $userProject is set for all HTTP requests using the client if specified by the provider config
 	// See https://cloud.google.com/apis/docs/system-parameters
+	// We explicitly set this header for logging visibility, even though
```
rileykarson, line 479: Can we also capture why we need to set it for logging visibility in this comment as well?

mmv1/third_party/terraform/website/docs/guides/provider_reference.html.markdown
```diff
@@ -67,11 +67,15 @@ If you are using Terraform on your workstation we recommend that you install
 as a primary authentication method. You can enable ADCs by running the command
 `gcloud auth application-default login`.
 
+<!-- 
+TODO: quota project is not currently read from ADC file b/360405077#comment8
```
rileykarson, line 71: Just to confirm: this also means that the `gcloud` commands below don't work, right? Since that ultimately works through a local credentials file?

So removing this statement in our docs reflects current behaviour.

ScottSuarez: This file is [specifically read](https://github.com/golang/oauth2/blob/22134a41033e44c2cd074106770ab5b7ca910d15/google/default.go#L304):
`$HOME/.config/gcloud/application_default_credentials.json`

We [preserve only](https://github.com/hashicorp/terraform-provider-google/blob/6b679462b2ee101a9a903c8f4370d40d8aebd8aa/google/transport/config.go#L1578-L1586) the [token source](https://github.com/hashicorp/terraform-provider-google/blob/6b679462b2ee101a9a903c8f4370d40d8aebd8aa/google/transport/config.go#L2205).

mmv1/third_party/terraform/transport/config.go.tmpl
ScottSuarez, line 479:
```suggestion
	// We explicitly set this header for logging visibility, even though 
	// option.WithQuotaProject will automatically set it in the client.
	// This duplicate setting ensures the header appears in our request logs
	// since our logging transport sits on top of the Google client library's
	// internal transport that would also sets this header.
```

mmv1/third_party/terraform/transport/config.go.tmpl
ScottSuarez: Basically, the default transport layer will set this header if `GOOGLE_CLOUD_QUOTA_PROJECT` is defined. Since the transport layer is defined below our logging transport their definition takes precedent.  

This is effectively solving for https://github.com/hashicorp/terraform-provider-google/issues/17882 .

mmv1/third_party/terraform/transport/config.go.tmpl
ScottSuarez: done

rileykarson: I'd suggest explaining the issue first, then the solution- 
```suggestion
	// Explicitly set the quota project for visibility in debug logging.
	// option.WithQuotaProject will set it in the HTTP client, but won't appear in provider debug logging
	// because the client is lower level than the logging transport. By also setting it here, we'll ensure it appears
	// in debug logs.
```

(may need to be reflowed)

mmv1/third_party/terraform/transport/config.go.tmpl
rileykarson: Thank you! Maybe something like this?

"The client libraries allow setting the GOOGLE_CLOUD_QUOTA_PROJECT environment variable directly, which unintentionally takes precedence over provider settings. Ensure that provider settings take precedence by applying to the client library's client directly."

mmv1/third_party/terraform/website/docs/guides/provider_reference.html.markdown
rileykarson: Thanks! Does that mean that quota projects _are_ read from the other ADC sources?

ScottSuarez: nope quota project is never read as far as I can tell. We always throw out everything in favor of the token source, which we pass to transport library. The token source doesn't contain any special information

mmv1/third_party/terraform/transport/config.go.tmpl
ScottSuarez: done

mmv1/third_party/terraform/transport/config.go.tmpl
ScottSuarez: added ! Thanks for the rephrase! :)

ScottSuarez (commented):
responding-inline

ScottSuarez:
@rileykarson I'm not sure if there are any other needed changes?

ScottSuarez:
> @ScottSuarez, this PR is waiting for action from you. If no action is taken, this PR will be closed in 28 days.
> 
> Please address any comments or change requests, or [re-request review](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/requesting-a-pull-request-review) from a core reviewer if no action is required.
> 
> 

@melinath -- This state appears incorrect? Unless I'm missing something

melinath:
Agreed, this looks incorrect - @shuyama1 could you take a look?

mmv1/third_party/terraform/transport/config.go.tmpl
```diff
@@ -435,11 +435,20 @@ func (c *Config) LoadAndValidate(ctx context.Context) error {
...
 
 	c.TokenSource = tokenSource
-
 	cleanCtx := context.WithValue(ctx, oauth2.HTTPClient, cleanhttp.DefaultClient())
+	clientOptions := []option.ClientOption{option.WithTokenSource(tokenSource)}
+
+	// The client libraries allow setting the GOOGLE_CLOUD_QUOTA_PROJECT environment variable 
+	// directly, which unintentionally takes precedence over provider settings. Ensure that 
+	// provider settings take precedence by applying to the client library's client directly
+	// b/360405077#comment8 - go/tpg-issue/17882
```
rileykarson, line 444: This will result in a user-impacting change, right? Should there be a changelog entry?

ScottSuarez: Added a changelog entry.

rileykarson (approved):

PR #12416, opened 2024-11-22
Handle operation for ApigeeOrganization delete requests.

masich:
This PR adds `Operation` handling to track the `ApigeeOrganization` deletion progress.

According to the `ApigeeOrganization` public docs for the [delete request](https://cloud.google.com/apigee/docs/reference/apis/apigee/rest/v1/organizations/delete), the result of such a request is [Operation](https://cloud.google.com/apigee/docs/reference/apis/apigee/rest/v1/organizations.operations#Operation). However, the provider does not correctly process the [delete response](https://github.com/hashicorp/terraform-provider-google/blob/v6.12.0/google/services/apigee/resource_apigee_organization.go#L590), immediately logging `Finished deleting Organization`. Deletion operation may take a while, and immediately telling users that deletion is completed needs to be corrected.

```release-note:bug
apigee: made `google_apigee_organization` wait for deletion operation to complete.
```

### DIFF
--- a/mmv1/products/apigee/Organization.yaml
+++ b/mmv1/products/apigee/Organization.yaml
@@ -30,7 +30,7 @@ timeouts:
   delete_minutes: 45
 autogen_async: true
 async:
-  actions: ['create', 'update']
+  actions: ['create', 'delete', 'update']
   type: 'OpAsync'
   operation:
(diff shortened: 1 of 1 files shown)


melinath (approved):
LGTM, thanks!

PR #12418, opened 2024-11-25
Fix client full example to use basename of kms config

rohitjangid:
Add basename function usage in the integrations client full example to reflect correct usage. This would fix the errors due to recent api changes.

```release-note:none
integrations: fixed example in docs to reflect usage of basename instead of full name in `cloud_kms_config` 
```

Fixes: https://github.com/hashicorp/terraform-provider-google/issues/20385
Fixes: https://github.com/hashicorp/terraform-provider-google/issues/20260


rohitjangid:
> see comment

@ScottSuarez I have done the changes. PTAL

ScottSuarez (approved):

PR #12419, opened 2024-11-25
Add Intercept resources to Network Security.

duvni:

All Intercept resources are bundled together in one PR because they're dependent on one another and cannot be tested separately.

```release-note:new-resource
`google_network_security_intercept_deployment`, `google_network_security_intercept_deployment_group`, `google_network_security_intercept_endpoint_group_association`, `google_network_security_intercept_endpoint_group`
```

### DIFF
--- a/mmv1/products/networksecurity/InterceptDeployment.yaml
+++ b/mmv1/products/networksecurity/InterceptDeployment.yaml
@@ -1,0 +1,121 @@
+# Copyright 2024 Google Inc.
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+---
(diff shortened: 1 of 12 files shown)


shuyama1:
I'll be out for a week and won't have time to review this PR before my leave. Reassigning to another reviewer. 

shuyama1:

zli82016 (changes requested):
It is adding four new resources in one PR. Can you split the PR into 4 PRs? Each PR will add a new resource.

duvni:
> It is adding four new resources in one PR and hard to review. Can you split the PR into 4 PRs? Each PR will add one new resource.

As specified in the description, these resources are dependent on each other so they cannot be tested separately, meaning the PRs will have to be stacked, and in case we need to rollback we will have to rollback all of the PRs as rolling back just one has the risk of breaking the others. If you still think this is the way to go let me know.

zli82016:
> As specified in the description, these resources are dependent on each other so they cannot be tested separately, meaning the PRs will have to be stacked, and in case we need to rollback we will have to rollback all of the PRs as rolling back just one has the risk of breaking the others. If you still think this is the way to go let me know.

This PR is really large. It is really easy to miss something and introduce bugs.

There are dependencies among the resources, then we can split the PR based on the dependencies.
PR1 will add `google_network_security_intercept_deployment_group`
PR2 will add `google_network_security_intercept_deployment`.
PR3 will add `google_network_security_intercept_endpoint_group`.
PR4 will add `google_network_security_intercept_endpoint_group_association`.
PR2 and PR3 can be opened in parallel. 

Feel free to let me know if you have any questions.

PR #12420, opened 2024-11-25
Add Mirroring resources to Network Security.

duvni:

All Mirroring resources are bundled together in one PR because they're dependent on one another and cannot be tested separately.

```release-note:new-resource
`google_network_security_mirroring_deployment`, `google_network_security_mirroring_deployment_group`, `google_network_security_mirroring_endpoint_group_association`, `google_network_security_mirroring_endpoint_group`
```

### DIFF
--- a/mmv1/products/networksecurity/MirroringDeployment.yaml
+++ b/mmv1/products/networksecurity/MirroringDeployment.yaml
@@ -15,8 +15,11 @@
 name: 'MirroringDeployment'
 description: MirroringDeployment represents the collectors within a Zone and is associated with a deployment group.
 min_version: 'beta'
+references:
+  guides:
+    'Mirroring deployment overview': 'https://cloud.google.com/network-security-integration/docs/out-of-band/deployments-overview'
+  api: 'https://cloud.google.com/network-security-integration/docs/reference/rest/v1beta1/projects.locations.mirroringDeployments'
 docs:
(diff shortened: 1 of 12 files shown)


mmv1/products/networksecurity/MirroringDeployment.yaml
```diff
@@ -0,0 +1,118 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+---
+name: 'MirroringDeployment'
+description: MirroringDeployment represents the collectors within a Zone and is associated with a deployment group.
+min_version: 'beta'
+docs:
```
trodge, line 18: Could you add docs links for these resources?

trodge (changes requested):

mmv1/products/networksecurity/MirroringDeployment.yaml
duvni: Done.

duvni: Actually docs doesn't seem to support a native url field:
https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/api/resource/docs.go
And I can't find any other resource using docs for these links, so added it under 'references' instead following the convention from other resources.

trodge (approved):

PR #12422, opened 2024-11-25
TF: plan -refresh=false for google_compute_ha_vpn_gateway with gcs backend has resource replacement

akshat-jindal-nit:
Bug: b/376368181
Issue: https://github.com/hashicorp/terraform-provider-google/issues/19978
This particular change is the main culprit of this issue https://screenshot.googleplex.com/79xwUXBEcc5SYAg.
Looks like this issue start occurring from the release v5.40.0 https://screenshot.googleplex.com/367U6wou3dHnzaJ.

```release-note:enhancement
bug: fixed an issue where `terraform plan -refresh=false` with `google_compute_ha_vpn_gateway.gateway_ip_version` would plan a resource replacement if a full refresh had not been run yet. Terraform now assumes that the value is the default value, `IPV4`, until a refresh is completed.
```

### DIFF
--- a/mmv1/templates/terraform/state_migrations/compute_ha_vpn_gateway.go.tmpl
+++ b/mmv1/templates/terraform/state_migrations/compute_ha_vpn_gateway.go.tmpl
@@ -1,0 +1,107 @@
+func resourceComputeHaVpnGatewayResourceV0() *schema.Resource {
+	return &schema.Resource{
+		Schema: map[string]*schema.Schema{
+			"name": {
+				Type:         schema.TypeString,
+				Required:     true,
+				ForceNew:     true,
+				ValidateFunc: verify.ValidateGCEName,
+				Description: `Name of the resource. Provided by the client when the resource is
+created. The name must be 1-63 characters long, and comply with
+RFC1035.  Specifically, the name must be 1-63 characters long and
+match the regular expression '[a-z]([-a-z0-9]*[a-z0-9])?' which means
+the first character must be a lowercase letter, and all following
+characters must be a dash, lowercase letter, or digit, except the last
+character, which cannot be a dash.`,
+			},
+			"network": {
+				Type:             schema.TypeString,
+				Required:         true,
+				ForceNew:         true,
+				DiffSuppressFunc: tpgresource.CompareSelfLinkOrResourceName,
+				Description:      `The network this VPN gateway is accepting traffic for.`,
+			},
+			"description": {
+				Type:        schema.TypeString,
+				Optional:    true,
+				ForceNew:    true,
+				Description: `An optional description of this resource.`,
+			},
+			"gateway_ip_version": {
+				Type:         schema.TypeString,
+				Optional:     true,
+				ForceNew:     true,
+				ValidateFunc: verify.ValidateEnum([]string{"IPV4", "IPV6", ""}),
+				Description:  `The IP family of the gateway IPs for the HA-VPN gateway interfaces. If not specified, IPV4 will be used. Default value: "IPV4" Possible values: ["IPV4", "IPV6"]`,
+				Default:      "IPV4",
+			},
+			"region": {
+				Type:             schema.TypeString,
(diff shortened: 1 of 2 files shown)


harshithpatte-g:
@akshat-jindal-nit can repro this issue with a test? Could we add that test?

akshat-jindal-nit:
> @akshat-jindal-nit can repro this issue with a test? Could we add that test?

@harshithpatte-g this particular change is related to terraform plan -refresh=false related changes. I think we don't write any particulars tests for it.
But this [TestAccDataSourceComputeHaVpnGateway](https://github.com/hashicorp/terraform-provider-google/blob/1f51b9850e752800c36d78c9942f5d27f58fa36d/google/services/compute/resource_compute_router_bgp_peer_test.go#L42C6-L42C44) is running successfully as our changes are intended to impact the default_values only .

rileykarson (commented):
Were you able to [test manually](https://googlecloudplatform.github.io/magic-modules/develop/test/run-tests/#optional-test-manually) to confirm the fix? I'm a bit confused why this works, because we're removing what is traditionally the mitigation for the issue here.

akshat-jindal-nit:
> Were you able to [test manually](https://googlecloudplatform.github.io/magic-modules/develop/test/run-tests/#optional-test-manually) to confirm the fix? I'm a bit confused why this works, because we're removing what is traditionally the mitigation for the issue here.

Yes I test it manually and this particular change is fixing this particular issue.

rileykarson:
Sorry, could you elaborate on how you've reproduced the issue? I continue to see it both before and after this change. Effectively, we assume that a refresh is performed after the upgrade- we'd resolve that with a StateUpgrader to infill the default value where it's null (but really, a full refresh would be better in case the live value is different).

akshat-jindal-nit:
> StateUpgrader

I perform these steps to reproduce the issue [b/37636818](https://b.corp.google.com/issues/376368181#comment2)
1. make a module with the 1st configuration (provider v5.16.0)
2. terraform init
3. terraform apply
4. change google provider version to v5.44.2
5. rm -f .terraform.lock.hcl
6. terraform init
7. terraform plan -refresh=false <--- the issue occurrs here
8. terraform plan <--- doesn't have changes
9. terraform apply <--- doesn't make any changes

By removing the defaulting logic, the fixed version of the function ensures that Terraform accurately reflects the actual state of the gateway_ip_version attribute in Google Cloud. This prevents the false positive change detection and avoids unnecessary resource replacements.

mmv1/templates/terraform/state_migrations/compute_ha_vpn_gateway.go.tmpl
```diff
@@ -0,0 +1,107 @@
...
+			"self_link": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+		},
+		UseJSONNumber: true,
+	}
+}
+
+func ResourceComputeHaVpnGatewayUpgradeV0(_ context.Context, rawState map[string]interface{}, meta interface{}) (map[string]interface{}, error) {
```
rileykarson, line 105: This doesn't make any specific changes to support the change yet- we'd want to read `gateway_ip_version`, and if `nil`, set the default value. That'll preseed the value into state and prevent a diff if a user doesn't do refresh.

I didn't spot any exact prior art, but this is an example of a basic function written here for reference: https://github.com/hashicorp/terraform-provider-google-beta/blob/4e7515540f48e50542ba8a3e70ac7eca316fb6b3/google-beta/services/kms/resource_kms_crypto_key.go#L870-L883

akshat-jindal-nit: thanks for the info will update the function accordingly.

rileykarson (commented):
The code looks good and this works on my machine to solve the original bug- just waiting for generate + test to LGTM and merge.

rileykarson (approved):

PR #12429, opened 2024-11-25
Change markdown links to use ref shortcodes from Hugo

SarahFrench:
Fixes https://github.com/hashicorp/terraform-provider-google/issues/20407

This PR solves the broken link issue above and also replaces other hardcoded links to use ref shortcodes

```release-note:none

```


SarahFrench:
TODO: local testing

PR #12430, opened 2024-11-25
Add enterprise_config field to GKE cluster object

bradhoekstra:
This PR adds support for configuring enterprise_config for GKE clusters (see https://cloud.google.com/kubernetes-engine/docs/how-to/enable-gkee#gcp-cluster).

Bug: b/380901610
Terraform Issue: https://github.com/hashicorp/terraform-provider-google/issues/20355

```release-note:enhancement
container: added `enterprise_config` field to `google_container_cluster` resource
```

### DIFF
--- a/mmv1/third_party/terraform/services/container/resource_container_cluster_test.go.tmpl
+++ b/mmv1/third_party/terraform/services/container/resource_container_cluster_test.go.tmpl
@@ -12834,6 +12834,15 @@ func TestAccContainerCluster_withEnterpriseConfig(t *testing.T) {
 				ImportStateVerify: true,
 				ImportStateVerifyIgnore: []string{"deletion_protection"},
 			},
+			{
+				Config: testAccContainerCluster_removeEnterpriseConfig(pid, clusterName, networkName, subnetworkName),
+			},
+			{
+				ResourceName:      "google_container_cluster.with_enterprise_config",
+				ImportState:       true,
+				ImportStateVerify: true,
+				ImportStateVerifyIgnore: []string{"deletion_protection"},
+			},
 		},
 	})
 }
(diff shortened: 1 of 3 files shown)


mmv1/third_party/terraform/services/container/resource_container_cluster_test.go.tmpl
```diff
@@ -12802,3 +12802,60 @@ resource "google_container_cluster" "primary" {
 }
   `, name, cgroupMode)
 }
+
+func TestAccContainerCluster_withEnterpriseConfig(t *testing.T) {
```
slevenick, line 12806: Can you add a final step to this test that removes the `enterprise_config` block entirely to remove the feature? We often see issues arise from removing a block, so it would be good to test

mmv1/third_party/terraform/website/docs/r/container_cluster.html.markdown
```diff
@@ -1432,6 +1436,12 @@ linux_node_config {
 
 * `enable_alts` - (Required) Whether the alts handshaker should be enabled or not for direct-path. Requires Workload Identity ([workloadPool]((#nested_workload_identity_config)) must be non-empty).
 
+<a name="nested_enterprise_config"></a>The `enterprise_config` block supports:
+
+* `desired_tier` - (Optional) Sets the tier of the cluster. Available options include `STANDARD` and `ENTERPRISE`.
+
+* `cluster_tier` - (Output) Sets the tier of the cluster. Available options include `STANDARD` and `ENTERPRISE`.
```
slevenick, line 1443: The standard way to document an output-only field is to add it to the `Attributes Reference` section below that contains all the computed fields.

Also this description seems to exactly match the desired tier, is that intended?

slevenick (changes requested):

slevenick (approved):
Thanks!

PR #12436, opened 2024-11-26
Correct triple-negative documentation.

StandardMikeBenza:
This documentation was incorrect and poorly worded.  "Whether or not" implies that a `true` value allows deletion and a `false` value prevents deletion, but that's the opposite of what `deletion_protection` does.

The sentence beginning with "Unless" was a bit mind-bending. "Unless" is a negative, "set to false" is a negative, and "will fail" is a negative.

```release-note:note
Corrected documentation for `deletion_protection` in Oracle and GKE clusters.
```


BBBmau:
tests are unrelated and are addressed in a separate issue: https://github.com/hashicorp/terraform-provider-google/issues/19983

BBBmau (approved):
Thanks!

PR #12437, opened 2024-11-26
Changes for b/379856036, b/378888812, b/378726194

nityaravi:

```release-note:REPLACEME

```


melinath (changes requested):
could you split this into separate PRs? One for the "adding documentation" changes, one for moving the reference files, one for moving the tests dir up a directory, one for splitting add-fields from resource.md, and a separate one for renaming resource to add-resource, at a minimum. This is way too much at once for me to review easily.

FYI regarding page weights, they reset within folders, so (for example) having weight of 290 for something in the reference folder is fine but not necessary.

PR #12439, opened 2024-11-27
`compute.disk`: move several parameters into properties since they arent used in URL

BBBmau:
This moves some fields in `resource_compute_disk` that don't follow the parameters description found in the MMv1 docs:
[`Contains a list of fields By convention, these should be the fields that are part URL parameters such as location and name.`](https://googlecloudplatform.github.io/magic-modules/develop/resource-reference/#parameters)

This was found after discovering that `conflicts` does not work if fields are set as a `parameter` rather than `properties`.

```release-note:none

```


c2thorn (approved):
large amount of diffs, but looks like it's just moving the same logic around

PR #12454, opened 2024-11-28
 Fix TF: Failing test(s): TestAccDataprocVirtualCluster_basic

devanshikhatsuriya:
* Updates dataproc virtual gke cluster version 3.1-dataproc-7 to 3.5-dataproc-19
* Fixes https://github.com/hashicorp/terraform-provider-google/issues/19076

```release-note:none
dataproc: fixed test failure due to usage of deprecated version
```

### DIFF
--- a/mmv1/third_party/terraform/services/dataproc/resource_dataproc_cluster_test.go.tmpl
+++ b/mmv1/third_party/terraform/services/dataproc/resource_dataproc_cluster_test.go.tmpl
@@ -112,7 +112,7 @@ func TestAccDataprocVirtualCluster_basic(t *testing.T) {
 	var cluster dataproc.Cluster
 	rnd := acctest.RandString(t, 10)
 	pid := envvar.GetTestProjectFromEnv()
-	version := "3.5-dataproc-19"
+	version := "3.5-dataproc-17"
 	networkName := acctest.BootstrapSharedTestNetwork(t, "gke-cluster")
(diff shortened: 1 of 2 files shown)


devanshikhatsuriya:
> It looks like this value may also be used in the documentation. Would you mind updating that as well?

Are you referring to Dataproc on GKE public docs? They are already updated, 3.1 is marked as deprecated - https://cloud.google.com/dataproc/docs/guides/dpgke/dataproc-gke-versions#spark_engine_35

devanshikhatsuriya:
> I'm referring to https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/third_party/terraform/website/docs/r/dataproc_cluster.html.markdown?plain=1#L217, which shows up on our provider website at https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/dataproc_cluster.

Thank you, I have updated that file too. Ptal

devanshikhatsuriya:
Hi @roaks3, is the test being skipped due to https://github.com/GoogleCloudPlatform/magic-modules/pull/12454. Should we re-enable it, since it is fixed now?

roaks3:
Ah yep, I missed that. Would you mind opening a new PR to un-skip that test and verify it passes?

devanshikhatsuriya:
> Ah yep, I missed that. Would you mind opening a new PR to un-skip that test and verify it passes?

Done, also verified test passed. Ptal - https://github.com/GoogleCloudPlatform/magic-modules/pull/12530

PR #12455, opened 2024-11-28
Increase AlloyDB cluster timeout

sengjea:
When creating an AlloyDB cluster with a backup source, the operation may take longer than 30 minutes to complete and times out. As an example, when creating a cluster with a 14TB AlloyDB backup defined, we saw that it took up to 64 minutes for the operation to complete.

```release-note:enhancement
alloydb: increased default timeout on `google_alloydb_cluster` to 120m from 30m
```


SirGitsalot (commented):
Looks good. I kicked off a test run - I don't see any reason why this would fail, but it's always good to see it pass before approving and merging. Thanks!

SirGitsalot (approved):

PR #12457, opened 2024-11-28
Hcp terraform OIDC poc

SarahFrench:

```release-note:none

```

### DIFF
--- a/mmv1/third_party/terraform/transport/config.go.tmpl
+++ b/mmv1/third_party/terraform/transport/config.go.tmpl
@@ -482,9 +482,34 @@ func (c *Config) LoadAndValidate(ctx context.Context) error {
 
 	c.Context = ctx
 
-	tokenSource, err := c.getTokenSource(c.Scopes, false)
-	if err != nil {
-		return err
+	var tokenSource oauth2.TokenSource
+	var err error
+	if c.ExternalCredentialsHcpTerraform != nil {
+		// Token source is created using external credentials
+		eaConfig := externalaccount.Config{
+			Audience:         c.ExternalCredentialsHcpTerraform.Audience,
+			SubjectTokenType: "urn:ietf:params:oauth:token-type:jwt",
+			// Is ServiceAccountImpersonationURL affected by Universe Domain?
+			ServiceAccountImpersonationURL: fmt.Sprintf("https://iamcredentials.googleapis.com/v1/projects/-/serviceAccounts/%s:generateAccessToken", c.ExternalCredentialsHcpTerraform.ServiceAccountEmail),
+			// ServiceAccountImpersonationLifetimeSeconds ??
+			// QuotaProjectID ??
+			Scopes:               c.Scopes,
+			SubjectTokenSupplier: c.ExternalCredentialsHcpTerraform,
+		}
+		// If UniverseDomain is set, the externalaccount package will use it to set the TokenURL (https://sts.UNIVERSE_DOMAIN/v1/token). Otherwise TokenURL defaults to https://sts.googleapis.com/v1/token
+		if c.UniverseDomain != "" {
+			eaConfig.UniverseDomain = c.UniverseDomain
+		}
+		tokenSource, err = externalaccount.NewTokenSource(ctx, eaConfig)
+		if err != nil {
+			return err
+		}
+	} else {
+		// Token source is created using using credentials/auth_token/ADCs and other related inputs
+		tokenSource, err = c.getTokenSource(c.Scopes, false)
+		if err != nil {
+			return err
+		}
 	}
 
 	c.TokenSource = tokenSource
@@ -498,7 +523,7 @@ func (c *Config) LoadAndValidate(ctx context.Context) error {
 	}
 
 	// Userinfo is fetched before request logging is enabled to reduce additional noise.
-	err = c.logGoogleIdentities()
+	err = c.logGoogleIdentities(ctx)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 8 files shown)


mmv1/third_party/terraform/transport/config.go.tmpl
```diff
@@ -429,9 +482,34 @@ func (c *Config) LoadAndValidate(ctx context.Context) error {
 
 	c.Context = ctx
 
-	tokenSource, err := c.getTokenSource(c.Scopes, false)
-	if err != nil {
-		return err
+	var tokenSource oauth2.TokenSource
```
SarahFrench, line 485: This is ugly and repetitive at the moment - needs refactor when PoC is taken forward

mmv1/third_party/terraform/transport/config.go.tmpl
```diff
@@ -176,6 +178,56 @@ func (f *Formatter) Format(entry *logrus.Entry) ([]byte, error) {
...
+
+var _ externalaccount.SubjectTokenSupplier = ExternalCredentialsHcpTerraform{}
+
+// SubjectToken returns the identity token passed to the provider as an argument from the config.
+// We do not interact with an external system to get a token.
+func (e ExternalCredentialsHcpTerraform) SubjectToken(ctx context.Context, options externalaccount.SupplierOptions) (string, error) {
+	if e.IdentityToken != "" {
+		return e.IdentityToken, nil
+	}
+	return "", errors.New("identity token unavailable in Config when configuring the provider")
```
SarahFrench, line 195: This is an example of an error we 'shouldn't' reach - this function should only be invoked if ExternalCredentialsHcpTerraform is set on the Config struct, i.e. the block is set in the provider block.

All the fields inside external_credentials_hcp_terraform are Required, so if external_credentials_hcp_terraform is present then the identity token should be present.

SarahFrench, line 195:
```suggestion
	return "", errors.New("identity token unavailable in Config when configuring the provider. Please report this issue to the provider developers")
```

mmv1/third_party/terraform/transport/config.go.tmpl
```diff
@@ -593,7 +648,20 @@ func (c *Config) logGoogleIdentities() error {
 
 // Get a TokenSource based on the Google Credentials configured.
 // If initialCredentialsOnly is true, don't follow the impersonation settings and return the initial set of creds.
-func (c *Config) getTokenSource(clientScopes []string, initialCredentialsOnly bool) (oauth2.TokenSource, error) {
+func (c *Config) getTokenSource(ctx context.Context, clientScopes []string, initialCredentialsOnly bool) (oauth2.TokenSource, error) {
+
+	if c.ExternalCredentialsHcpTerraform != nil {
+		// Token source is created using external credentials
+		// Impersonation isn't done here, so ignore use of initialCredentialsOnly bool
```
SarahFrench, line 655: > Impersonation isn't done here

When using external credentials we _are_ impersonating a service account via WIF, but this is different to the impersonation already present in the provider where Terraform might use service account A as it's identity, and A is used to impersonate service account B which is used to actually provision things.

mmv1/third_party/terraform/transport/config.go.tmpl
```diff
@@ -1156,6 +1224,26 @@ type StaticTokenSource struct {
       oauth2.TokenSource
 }
 
+// getExternalAccountConfig returns an externalaccount.Config based on the Config object.
+// The external account config is intended to be used to create a token source.
+func (c *Config) getExternalAccountConfig(clientScopes []string) externalaccount.Config {
+	eaConfig := externalaccount.Config{
```
SarahFrench, line 1230: This is the part of the code with the biggest set of remaining unknowns - what pre-existing provider arguments should impact this externalaccount.Config? Currently we use the `clientScopes` argument, which is created using any values supplied via `scopes` in the provider block, but there are possibly other things to include:

1) Universe Domains

How should the [universe_domain provider argument](https://registry.terraform.io/providers/hashicorp/google/latest/docs/guides/provider_reference#universe_domain-1) be used here? How would we test that? There is current work in progress to add test coverage for use of universe_domain, so we would not be able to test use of `external_credentials_hcp_terraform` when `universe_domain` is also set.

Maybe we should purposefully not make this feature compatible with universe_domain at first, due to inability to test?

2) QuotaProjectID

I believe this would be dependent on user_project_override and billing_project.

3) ServiceAccountImpersonationLifetimeSeconds

Do we want to set this?

mmv1/third_party/terraform/transport/config.go.tmpl
```diff
@@ -1156,6 +1224,26 @@ type StaticTokenSource struct {
...
+		Audience:         c.ExternalCredentialsHcpTerraform.Audience,
+		SubjectTokenType: "urn:ietf:params:oauth:token-type:jwt",
```
SarahFrench, line 1232: This is an example of something that we hardcode that's compatible with HCP Terraform

SarahFrench (commented):
Self-review to help with handover of proof of concept

mmv1/third_party/terraform/transport/config.go.tmpl
SarahFrench, line 1230: See: https://pkg.go.dev/golang.org/x/oauth2@v0.21.0/google/externalaccount#Config

mmv1/third_party/terraform/transport/config.go.tmpl
```diff
@@ -176,6 +178,56 @@ func (f *Formatter) Format(entry *logrus.Entry) ([]byte, error) {
...
+	Audience            string
+	ServiceAccountEmail string
+	IdentityToken    string
+}
+
+var _ externalaccount.SubjectTokenSupplier = ExternalCredentialsHcpTerraform{}
+
+// SubjectToken returns the identity token passed to the provider as an argument from the config.
+// We do not interact with an external system to get a token.
+func (e ExternalCredentialsHcpTerraform) SubjectToken(ctx context.Context, options externalaccount.SupplierOptions) (string, error) {
```
SarahFrench, line 191: See: https://pkg.go.dev/golang.org/x/oauth2@v0.21.0/google/externalaccount#SubjectTokenSupplier

SarahFrench (commented):
Add some links to docs.

Make sure to read [this bit of the externalaccount package's docs](https://pkg.go.dev/golang.org/x/oauth2@v0.21.0/google/externalaccount#pkg-overview:~:text=other%2Dproviders%23create_a_credential_configuration-,To%20use%20a%20custom%20function%20to%20supply%20the%20token%2C%20define%20a,google.com/go/storage%20package%20and%20pass%20in%20option.WithTokenSource(yourTokenSource)),-Note%20that%20this):

>To use a custom function to supply the token, define a struct that implements the [SubjectTokenSupplier](https://pkg.go.dev/golang.org/x/oauth2@v0.21.0/google/externalaccount#SubjectTokenSupplier) interface for OIDC/SAML providers, or one that implements [AwsSecurityCredentialsSupplier](https://pkg.go.dev/golang.org/x/oauth2@v0.21.0/google/externalaccount#AwsSecurityCredentialsSupplier) for AWS providers. This can then be used when building a [Config](https://pkg.go.dev/golang.org/x/oauth2@v0.21.0/google/externalaccount#Config). The [golang.org/x/oauth2.TokenSource](https://pkg.go.dev/golang.org/x/oauth2@v0.21.0#TokenSource) created from the config using [NewTokenSource](https://pkg.go.dev/golang.org/x/oauth2@v0.21.0/google/externalaccount#NewTokenSource) can then be used to access Google Cloud resources. For instance, you can create a new client from the [cloud.google.com/go/storage](https://pkg.go.dev/cloud.google.com/go/storage) package and pass in option.WithTokenSource(yourTokenSource))

SarahFrench:
See also: https://github.com/GoogleCloudPlatform/magic-modules/pull/12714

PR #12458, opened 2024-11-29
Fixed recaptcha_options permadiff causing rules being recreated

matheusaleixo-cit:
Fixes a permadiff related to the `recaptcha_options` field in google_compute_security_policy.
Also code cleanup for the same fix previously applied to `preconfigured_waf_config` field;

Part of: https://github.com/hashicorp/terraform-provider-google/issues/16882, https://github.com/hashicorp/terraform-provider-google/issues/18596

```release-note:bug
compute: fixed permadiff on the `recaptcha_options` field for `google_compute_security_policy` resource
```

### DIFF
--- a/mmv1/third_party/terraform/services/compute/resource_compute_security_policy.go.tmpl
+++ b/mmv1/third_party/terraform/services/compute/resource_compute_security_policy.go.tmpl
@@ -24,6 +24,26 @@ import (
 {{- end }}
 )
 
+func verifyRulePriorityCompareEmptyValues(d *schema.ResourceData, rulePriority int, schemaKey string) bool {
+	if schemaRules, ok := d.GetOk("rule"); ok {
+		for _, itemRaw := range schemaRules.(*schema.Set).List() {
+			if itemRaw == nil {
+				continue
+			}
+			item := itemRaw.(map[string]interface{})
+
+			schemaPriority := item["priority"].(int)
+			if rulePriority == schemaPriority {
+				if tpgresource.IsEmptyValue(reflect.ValueOf(item[schemaKey])) {
+					return true
+				}
+				break
+			}
+		}
+	}
+	return false
+}
+
 // IsEmptyValue does not consider a empty PreconfiguredWafConfig object as empty so we check it's nested values
 func preconfiguredWafConfigIsEmptyValue(config *compute.SecurityPolicyRulePreconfiguredWafConfig) bool {
 	if (tpgresource.IsEmptyValue(reflect.ValueOf(config.Exclusions)) &&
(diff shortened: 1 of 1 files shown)


matheusaleixo-cit:
I don't think this issue can be reproduced in a vcr test, (it's unclear what causes the API to return the empty values, plus we can't intentionally input these values in terraform since the resource is configured to not allow empty values for the recaptcha_options fields, and on rule update the resource uses the patchRule endpoint).

This issue can be tested manually:

- Create a `main.tf` file configuring the security policy resource and some rules and apply the changes;
- Manually update a rule setting the `recaptcha_options.action_token_site_keys` and `recaptcha_options.session_token_site_keys` fields with empty values using the [patch](https://cloud.google.com/compute/docs/reference/rest/v1/securityPolicies/patch) endpoint ([patchRule](https://cloud.google.com/compute/docs/reference/rest/v1/securityPolicies/patchRule) endpoint seems to ignore these fields when setting empty values).
- From here the permadiff can be observed by reapplying the terraform config;
- Configuring a [developer override](https://googlecloudplatform.github.io/magic-modules/develop/test/run-tests/#optional-test-manually) with the changes and running a apply again confirms that the permadiff is fixed;

slevenick (changes requested):
Can you add a test that would have failed with the existing resource?

matheusaleixo-cit:
> Can you add a test that would have failed with the existing resource?

I could not find a way to reproduce the permadiff in a conventional test as there is no way to induce the API to return empty values via terraform config due to how the resource is configured.

I could maybe try to run a separate API request in a PreConfig function in a test step to patch these empty values but I don't know if this is a correct way of approaching this.



> I don't think this issue can be reproduced in a vcr test, (it's unclear what causes the API to return the empty values, plus we can't intentionally input these values in terraform since the resource is configured to not allow empty values for the recaptcha_options fields, and on rule update the resource uses the patchRule endpoint).
> 
> This issue can be tested manually:
> 
> * Create a `main.tf` file configuring the security policy resource and some rules and apply the changes;
> * Manually update a rule setting the `recaptcha_options.action_token_site_keys` and `recaptcha_options.session_token_site_keys` fields with empty values using the [patch](https://cloud.google.com/compute/docs/reference/rest/v1/securityPolicies/patch) endpoint ([patchRule](https://cloud.google.com/compute/docs/reference/rest/v1/securityPolicies/patchRule) endpoint seems to ignore these fields when setting empty values).
> * From here the permadiff can be observed by reapplying the terraform config;
> * Configuring a [developer override](https://googlecloudplatform.github.io/magic-modules/develop/test/run-tests/#optional-test-manually) with the changes and running a apply again confirms that the permadiff is fixed;

slevenick (approved):

PR #12463, opened 2024-12-02
Composer

rlapin-pl:
Remove 2 Composer 1 tests that are failing due to labels with google- prefix.

```release-note:none

```

### DIFF
--- a/mmv1/third_party/terraform/services/composer/resource_composer_environment_test.go.tmpl
+++ b/mmv1/third_party/terraform/services/composer/resource_composer_environment_test.go.tmpl
@@ -77,46 +77,6 @@ func TestAccComposerEnvironment_basic(t *testing.T) {
 	})
 }
 
-// Checks that all updatable fields can be updated in one apply
-// (PATCH for Environments only is per-field)
-func TestAccComposerEnvironment_update(t *testing.T) {
-	// Currently failing
-	acctest.SkipIfVcr(t)
-	t.Parallel()
-
-	envName := fmt.Sprintf("%s-%d", testComposerEnvironmentPrefix, acctest.RandInt(t))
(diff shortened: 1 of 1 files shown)


spapi17 (changes requested):
We can also remove corresponding configurations: testAccComposerEnvironment_update (on line 2588) and testAccComposerEnvironment_webServerCfg (on line 1943), they are not used in any other tests.

rlapin-pl:
> We can also remove corresponding configurations: testAccComposerEnvironment_update (on line 2588) and testAccComposerEnvironment_webServerCfg (on line 1943), they are not used in any other tests.

Done

spapi17 (approved):

spapi17 (approved):

slevenick (changes requested):
These tests are valuable and should not be removed without a good reason.

If the specific label is causing the issue, please rework the test to work as intended but to not set this label *or* add logic to the resource to prevent sending these labels as described in https://github.com/hashicorp/terraform-provider-google/issues/20249#issuecomment-2469020551

rlapin-pl:
> If the specific label is causing the issue, please rework the test to work as intended but to not set this label _or_ add logic to the resource to prevent sending these labels as described in

This label is automatically added "Version 6.0.0 introduced the goog-terraform-provisioned attribution label, which is added automatically to all Terraform-provisioned resources. This is causing problems with the google_composer_environment resource.". 

The tests cannot be reworked, because these labels are passed automatically and are not accepted by the API. We are not planning to change API and I got a green light to delete these tests, because 1) Composer 1 creation is not supported anymore 2) Composer 1 update operations are partially covered with other tests.

slevenick (approved):

PR #12465, opened 2024-12-02
Update Stream.yaml to add cdc method support for MySQL

ravijaingit:
This PR adds support for creating streams for MySQL source which uses gtid based replication cdc method.
Fixes : [hashicorp/terraform-provider-google#20383](https://github.com/hashicorp/terraform-provider-google/issues/20383)

```release-note:enhancement
datastream: added `gtid`, `binaryLogPosition` fields to `datastream_stream` resource
```


ScottSuarez (changes requested):
Can you add test cases using this field and an update test (if the field is updatable).

mmv1/templates/terraform/examples/datastream_stream_mysql_gtid.tf.erb
```diff
@@ -0,0 +1,103 @@
+resource "google_sql_database_instance" "instance" {
```
ScottSuarez, line 1: This test needs to referenced as an example. As is, this test file is not doing anything.

https://github.com/ravijaingit/magic-modules/blob/3697c8acbcd352a17953d607d053068799532cc4/mmv1/products/datastream/Stream.yaml#L57-L189

mmv1/third_party/terraform/services/datastream/resource_datastream_stream_test.go
```diff
@@ -33,7 +33,7 @@ func TestAccDatastreamStream_update(t *testing.T) {
 				ResourceName:            "google_datastream_stream.default",
 				ImportState:             true,
 				ImportStateVerify:       true,
-				ImportStateVerifyIgnore: []string{"create_without_validation", "stream_id", "location", "desired_state", "labels", "terraform_labels"},
+				ImportStateVerifyIgnore: []string{"gtid", "binaryLogPosition", "create_without_validation", "stream_id", "location", "desired_state", "labels", "terraform_labels"},
```
ScottSuarez, line 36: Is there a reason are are adding these to ImportStateVerifyIgnore?

ScottSuarez (changes requested):

mmv1/third_party/terraform/services/datastream/resource_datastream_stream_test.go
```diff
@@ -63,12 +63,22 @@ func TestAccDatastreamStream_update(t *testing.T) {
...
 			},
 			{
 				// Disable prevent_destroy
 				Config: testAccDatastreamStream_datastreamStreamBasicUpdate(context, "RUNNING", false),
 			},
```
ScottSuarez, line 71: We need setting prevent destroy to false to be the last step in this test. Otherwise this resource cannot be cleaned up

mmv1/templates/terraform/examples/datastream_stream_mysql_gtid.tf.erb
```diff
@@ -0,0 +1,103 @@
...
+    name                = "<%= ctx[:vars]['mysql_name'] %>"
```
ScottSuarez, line 2: Also ensure to set prevent destroy to false. Unless this is intended to be a docs only config.

ScottSuarez (changes requested):

mmv1/templates/terraform/examples/datastream_stream_mysql_gtid.tf.erb
ravijaingit: Done.

mmv1/third_party/terraform/services/datastream/resource_datastream_stream_test.go
ravijaingit: Done.

mmv1/third_party/terraform/services/datastream/resource_datastream_stream_test.go
ravijaingit: Tests were failing because of these so ignored them. now removed them after fixing the code.

mmv1/templates/terraform/examples/datastream_stream_mysql_gtid.tf.erb
ravijaingit: done.

ScottSuarez (approved):

PR #12467, opened 2024-12-02
Fix IAM role race condition

damondouglas:
This PR fixes https://github.com/hashicorp/terraform-provider-google/issues/17385 adding a [time_sleep](https://registry.terraform.io/providers/hashicorp/time/latest/docs/resources/sleep) block to fix a race condition when launching a Dataflow flex template Job. Without this fix, it was found that ~8% (on my local machine), the Dataflow Job would fail with various reasons related to accessing the storage bucket. The root cause was the IAM role would not propagate by the time Terraform sent the create Job request.

```release-note:none

```


NickElliot (commented):
Is 5 minutes necessary, or will the race condition be covered with a say 2 minute timeout instead?

damondouglas:
> Is 5 minutes necessary, or will the race condition be covered with a say 2 minute timeout instead?

Thank you for reviewing. Perhaps play it safe for now with the more generous 5 min and reduce the time in the future?

NickElliot (approved):
lgtm

PR #12474, opened 2024-12-03
Add NetworkTier for GlobalForwardingRule.

marcinpilat:
Adds a NetworkTier for GlobalForwardingRule.

Fixes https://github.com/hashicorp/terraform-provider-google/issues/20514

```release-note:enhancement
compute: added `network_tier` field to `google_compute_global_forwarding_rule` resource
```

### DIFF
--- a/mmv1/templates/terraform/examples/global_forwarding_rule_external_managed.tf.tmpl
+++ b/mmv1/templates/terraform/examples/global_forwarding_rule_external_managed.tf.tmpl
@@ -3,6 +3,7 @@ resource "google_compute_global_forwarding_rule" "default" {
   target                = google_compute_target_http_proxy.default.id
   port_range            = "80"
   load_balancing_scheme = "EXTERNAL_MANAGED"
+  network_tier          = "PREMIUM"
 }
 
 resource "google_compute_target_http_proxy" "default" {
--- a/mmv1/products/compute/GlobalForwardingRule.yaml
+++ b/mmv1/products/compute/GlobalForwardingRule.yaml
@@ -485,6 +485,25 @@ properties:
     update_url: 'projects/{{project}}/global/forwardingRules/{{name}}/setTarget'
     update_verb: 'POST'
     diff_suppress_func: 'tpgresource.CompareSelfLinkRelativePaths'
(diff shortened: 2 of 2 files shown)


pawelJas (approved):

mmv1/templates/terraform/examples/global_forwarding_rule_external_managed.tf.tmpl
```diff
@@ -3,6 +3,7 @@ resource "google_compute_global_forwarding_rule" "default" {
   target                = google_compute_target_http_proxy.default.id
   port_range            = "80"
   load_balancing_scheme = "EXTERNAL_MANAGED"
+  network_tier          = "PREMIUM"
```
slevenick, line 6: Can you set this to the non-default value in this test? Just to make sure we can set to something that's not PREMIUM

slevenick (changes requested):

mmv1/templates/terraform/examples/global_forwarding_rule_external_managed.tf.tmpl
marcinpilat: Default is more complex here - `PREMIUM` in general is default for global ForwardingRules unless someone has project defined `STANDARD` network tier.  `STANDARD` is not allowed for global ForwardingRules, so if someone has this per-project setting, then effectively cannot use terraform to setup global FRs.

This also is a reason why we cannot specify here other tier than `PREMIUM`.

I was considering two options:
 * Mimic API and just add `networkTier` to global FR (so users can set `PREMIUM` explicitly), or
 * Add custom code for terraform that would always set `PREMIUM` as networkTier.

slevenick (approved):

mmv1/templates/terraform/examples/global_forwarding_rule_external_managed.tf.tmpl
slevenick: Talked offline, this is only applicable for STANDARD projects that need to specify PREMIUM to be able to create a GFR. We don't have access to these for testing

PR #12481, opened 2024-12-03
Add NetworkConfigs and NetworkConfig.QueueCount to TPU v2 VM

zeleena:
Add NetworkConfigs and NetworkConfig.QueueCount fields. NetworkConfigs (array) and NetworkConfig (singular) conflict. NetworkConfigs should be the field to specify multiple NetworkConfig, so I copied over the NetworkConfig NestedObject properties to its array form, NetworkConfigs.

```release-note:enhancement
tpuv2: added `network_configs` and `network_config.queue_count` fields to `google_tpu_v2_vm` resource
```

### DIFF
--- a/mmv1/products/tpuv2/Vm.yaml
+++ b/mmv1/products/tpuv2/Vm.yaml
@@ -127,6 +127,8 @@ properties:
     min_version: 'beta'
     immutable: true
     default_from_api: true
+    conflicts:
+      - network_configs
     properties:
       - name: 'network'
         type: String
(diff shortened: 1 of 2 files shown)


zeleena:
Hi @roaks3, it's been about two business days and wanted to see if you could please take a look? I was aiming to merge this change prior to the cut for the next release.

roaks3:
@SirGitsalot I'm out tomorrow and already fairly backed up on reviews, and since this isn't an enormous change, I thought it best to re-roll the reviewer. The contributor pinged me and was hoping to get this in before the next release cut.

SirGitsalot (commented):
LGTM, but with a couple of quick questions:

- What's the API serving status of the new field? cl/702456043 was only submitted on Tuesday
- If only a single config is submitted in `network_configs`, will it always be returned in `network_configs`? (that is, is there any point in the API handler where the code says "I only got one element in `network_configs`, so I'm going to treat it as if I was sent `network_config` instead" - I suspect the answer i no, but it would cause problems with Terraform if it does)

zeleena:
Thank you for the quick review! Answers to questions:

- This field should be available in the v2 API by end of next week 12/15 or earlier if the rollout is accelerated since my team is trying to meet a deadline with this feature. With the Terraform release schedule, if this PR is submitted by Tuesday, it would be released approximately 12/16 when the field is available in the v2 API.
- That's correct. If a user sets a single config in the array `network_configs` field, then that info will still be returned in that same `network_configs` field. Single configs shouldn't be moved to the singular `network_config` field.

SirGitsalot (approved):

PR #12485, opened 2024-12-04
Allow for `IPV6_ONLY` stackType configurations

karolgorc:
reopen of https://github.com/GoogleCloudPlatform/magic-modules/pull/12283
related to b/360733056

This got rolled back because it was failing `TestAccComputeInstanceNetworkIntefaceWithSecurityPolicy`. I don't have Cloud Armor on my private cloud environment so can i get a detailed test log to see why it fails and debug it? Preferably with `TF_LOG=DEBUG` while the test is running

- Supports `IPV6_ONLY` stackType for instances and subnetworks
- `ipv6_access_config` is now Computed because when providing an IPV6 subnet the access config will get autofilled from the API

```release-note:enhancement
compute: added support for `IPV6_ONLY` stack_type to `google_compute_subnetwork`, `google_compute_instance`, `google_compute_instance_template` and `google_compute_region_instance_template`.
```

### DIFF
--- a/mmv1/third_party/terraform/services/compute/resource_compute_instance.go.tmpl
+++ b/mmv1/third_party/terraform/services/compute/resource_compute_instance.go.tmpl
@@ -536,7 +536,7 @@ func ResourceComputeInstance() *schema.Resource {
 							Type:         schema.TypeString,
 							Optional:     true,
 							Computed:     true,
-							ValidateFunc: validation.StringInSlice([]string{"IPV4_ONLY", "IPV4_IPV6", ""}, false),
+							ValidateFunc: validation.StringInSlice([]string{"IPV4_ONLY", "IPV4_IPV6", "IPV6_ONLY", ""}, false),
 							Description: `The stack type for this network interface to identify whether the IPv6 feature is enabled or not. If not specified, IPV4_ONLY will be used.`,
 						},
 
@@ -548,7 +548,8 @@ func ResourceComputeInstance() *schema.Resource {
 
 						"ipv6_access_config": {
 							Type:        schema.TypeList,
-                                                        Optional:    true,
+                            Optional:    true,
+							Computed:    true,
 							Description: `An array of IPv6 access configurations for this interface. Currently, only one IPv6 access config, DIRECT_IPV6, is supported. If there is no ipv6AccessConfig specified, then this instance will have no external IPv6 Internet access.`,
 							Elem: &schema.Resource{
 								Schema: map[string]*schema.Schema{
--- a/mmv1/third_party/terraform/services/compute/resource_compute_instance.go.tmpl
+++ b/mmv1/third_party/terraform/services/compute/resource_compute_instance.go.tmpl
@@ -549,7 +549,6 @@ func ResourceComputeInstance() *schema.Resource {
 						"ipv6_access_config": {
 							Type:        schema.TypeList,
                             Optional:    true,
-							Computed:    true,
 							Description: `An array of IPv6 access configurations for this interface. Currently, only one IPv6 access config, DIRECT_IPV6, is supported. If there is no ipv6AccessConfig specified, then this instance will have no external IPv6 Internet access.`,
 							Elem: &schema.Resource{
 								Schema: map[string]*schema.Schema{
--- a/mmv1/products/compute/Subnetwork.yaml
+++ b/mmv1/products/compute/Subnetwork.yaml
@@ -131,7 +131,7 @@ examples:
       subnetwork_name: 'subnet-ipv6-only'
       network_name: 'network-ipv6-only'
   - name: 'subnetwork_ipv6_only_external'
-    primary_resource_id: 'subnetwork-ipv6-only-external'
+    primary_resource_id: 'subnetwork-ipv6-only'
     exclude_docs: true
     vars:
       subnetwork_name: 'subnet-ipv6-only'
--- a/mmv1/products/compute/Subnetwork.yaml
+++ b/mmv1/products/compute/Subnetwork.yaml
@@ -124,6 +124,18 @@ examples:
       network_name: 'network-reserved-secondary-range'
       primary_range_name: 'reserved-primary'
       secondary_range_name: 'reserved-secondary'
+  - name: 'subnetwork_ipv6_only_internal'
+    primary_resource_id: 'subnetwork-ipv6-only'
+    exclude_docs: true
+    vars:
+      subnetwork_name: 'subnet-ipv6-only'
+      network_name: 'network-ipv6-only'
+  - name: 'subnetwork_ipv6_only_external'
+    primary_resource_id: 'subnetwork-ipv6-only-external'
+    exclude_docs: true
+    vars:
+      subnetwork_name: 'subnet-ipv6-only'
+      network_name: 'network-ipv6-only'
 virtual_fields:
   - name: 'send_secondary_ip_range_if_empty'
     description: |
--- a/mmv1/products/compute/Subnetwork.yaml
+++ b/mmv1/products/compute/Subnetwork.yaml
@@ -403,6 +403,7 @@ properties:
     enum_values:
       - 'IPV4_ONLY'
       - 'IPV4_IPV6'
+      - 'IPV6_ONLY'
   - name: 'ipv6AccessType'
     type: Enum
     description: |
--- a/mmv1/templates/terraform/examples/subnetwork_ipv6_only_external.tf.tmpl
+++ b/mmv1/templates/terraform/examples/subnetwork_ipv6_only_external.tf.tmpl
@@ -1,0 +1,12 @@
+resource "google_compute_subnetwork" "subnetwork-ipv6-only" {
+  name          = "{{index $.Vars "subnetwork_name"}}"
+  region        = "us-central1"
+  network       = google_compute_network.custom-test.id
+  stack_type    = "IPV6_ONLY"
+  ipv6_access_type = "EXTERNAL"
+}
+
+resource "google_compute_network" "custom-test" {
+  name                    = "{{index $.Vars "network_name"}}"
+  auto_create_subnetworks = false
+}
--- a/mmv1/templates/terraform/examples/subnetwork_ipv6_only_internal.tf.tmpl
+++ b/mmv1/templates/terraform/examples/subnetwork_ipv6_only_internal.tf.tmpl
@@ -1,0 +1,12 @@
+resource "google_compute_subnetwork" "subnetwork-ipv6-only" {
+  name          = "{{index $.Vars "subnetwork_name"}}"
+  region        = "us-central1"
+  network       = google_compute_network.custom-test.id
+  stack_type    = "IPV6_ONLY"
+  ipv6_access_type = "INTERNAL"
+}
+
+resource "google_compute_network" "custom-test" {
+  name                    = "{{index $.Vars "network_name"}}"
+  auto_create_subnetworks = false
+}
--- a/mmv1/templates/terraform/examples/subnetwork_ipv6_only_internal.tf.tmpl
+++ b/mmv1/templates/terraform/examples/subnetwork_ipv6_only_internal.tf.tmpl
@@ -9,4 +9,5 @@ resource "google_compute_subnetwork" "subnetwork-ipv6-only" {
 resource "google_compute_network" "custom-test" {
   name                    = "{{index $.Vars "network_name"}}"
   auto_create_subnetworks = false
+  enable_ula_internal_ipv6 = true
 }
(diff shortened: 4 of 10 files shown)


zli82016 (changes requested):
The test still failed.

zli82016:
https://paste.googleplex.com/4789365243641856#l=10517

This is the logs for the test TestAccComputeInstanceNetworkIntefaceWithSecurityPolicy. Please let me know if you can open it or not.

karolgorc:
don't have access

zli82016:
[debug-google-beta-549ff5d-277977-TestAccComputeInstanceNetworkIntefaceWithSecurityPolicy__access_config_update_access_config.txt](https://github.com/user-attachments/files/18065340/debug-google-beta-549ff5d-277977-TestAccComputeInstanceNetworkIntefaceWithSecurityPolicy__access_config_update_access_config.txt)


These are the logs.

karolgorc:
I'm currently seeing that our schema is not aligned with the actual API schema here.

We have `security_policy` under `network_interface` where as in the API it looks like this.

image

This is most likely why the `Computed: true` causes problems. Especially with `IPV4_IPV6`

zli82016:
> This is most likely why the `Computed: true` causes problems. Especially with `IPV4_IPV6` . We can only have one sec policy per NIC where in reality it's one sec policy per access config. This is probably a makeshift solution to support the `IPV4_IPV6` and have them set the same values across all the access configs.
> 
> I'm guessing that we've hit some case where the IPV6 sec policy gets left off while updating and it's not covered by this logic.
> 
> This is most likely to get changed and is a breaking change because most of the feature gap between provider and API on NICs is IPV6 related

Right, computed + optional on this field `ipv6_access_config` is the reason for the test failing.

[In this test step](https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/third_party/terraform/services/compute/resource_compute_instance_test.go.tmpl#L4100) the expected error `Error setting security policy to the instance since at least one access config must exist` is from the [Terraform provider code](https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/third_party/terraform/services/compute/compute_instance_network_interface_helpers.go.tmpl#L159) and doesn't happen. The reason is that when `ipv6_access_config` is not specified in the configuration, with computed + optional,  the previous value of this field  in the Terraform state will be used for API request. It means that `ipv6_access_config` exists and this error doesn't  occur.

zli82016:
Is making `ipv6_access_config` computed a requirement to support IPV6_ONLY stackType?

Based on the comment, it looks like `ipv6_access_config` is filled by API if the subnetwork given to the network_interface is a subnetwork of external IPv6 addresses. This case can happen in both IPV6_ONLY and IPV4_IPV6 stackType?

karolgorc:
Will test IPV4_IPV6. But for IPV6_ONLY not having this as computed and providing an IPV6 external subnetwork will cause permadiff for the user, as the access config will be set by the API. The entire block will initialize itself.

zli82016:
> Will test IPV4_IPV6. But for IPV6_ONLY not having this as computed and providing an IPV6 external subnetwork will cause permadiff for the user, as the access config will be set by the API. The entire block will initialize itself.

@karolgorc , have you tested IPV4_IPV6?

zli82016:
Also, can you add a test for `IPV6_ONLY` stackType? Thanks.

ChuAmberZhang:
Hi @karolgorc, thanks for picking up this work! Just curious (I'm not familiar with Terraform at all), why doesn't the stackType field need to be added to mmv1/products/compute/Instance.yaml?

ChuAmberZhang:
> Will test IPV4_IPV6. But for IPV6_ONLY not having this as computed and providing an IPV6 external subnetwork will cause permadiff for the user, as the access config will be set by the API. The entire block will initialize itself.

Also from API implementation, it is the same case with IPV4_IPV6 instances. If no ipv6_access_config is specified, but external IPv6 subnet is specified, the API will auto-fill the ipv6_access_config field for the result instance.

karolgorc:
> @karolgorc , have you tested IPV4_IPV6? 

Will update soon, currently working on adding 12 fields on disks that are not in the provider (initialize_params and kms related)

> Also, can you add a test for IPV6_ONLY stackType? 

yup

>  why doesn't the stackType field need to be added to mmv1/products/compute/Instance.yaml?

As of now Instance is a handwritten resource, not MMv1 because of it's complexity i'm guessing. The difference is described more thoroughly in magic-modules [docs](https://googlecloudplatform.github.io/magic-modules/develop/add-resource/)

> it is the same case with IPV4_IPV6 instances. If no ipv6_access_config is specified, but external IPv6 subnet is specified, the API will auto-fill the ipv6_access_config field for the result instance.

So the behavior persists. I will look into how do we deal with these sec policies inside our Update function and try to normalize this for our implementation but for the future this field should get deprecated and moved into `access_config` and `ipv6_access_config`. Not sure if this would be a breaking change. Should we wait for the major release to do this?

mmv1/third_party/terraform/services/compute/resource_compute_instance.go.tmpl
```diff
@@ -541,7 +541,8 @@ func ResourceComputeInstance() *schema.Resource {
 
 						"ipv6_access_config": {
 							Type:        schema.TypeList,
-                                                        Optional:    true,
+                            Optional:    true,
+							Computed:    true,
```
zli82016, line 545: @karolgorc , @ChuAmberZhang , adding `Computed: true` will suppress the diffs. But it will prevent the removal of `ipv6_access_config` for users, as the previous value will be used. Do you think it will affect a lot of current users with `IPV4_IPV6` and `IP4_ONLY` stackType as `ipv6_access_config` cannot be removed?

zli82016: There is not a way to make the removal of `ipv6_access_config` work with adding `Computed: true` with the current Terraform SDK library.

zli82016: Another way to suppress the diffs is adding a diff suppress function for `IPV4_IPV6` and `IPV6_ONLY` stackType and to not add `Computed: true`. The function will also prevent the removal of `ipv6_access_config`.

zli82016:
> Hi @karolgorc, thanks for picking up this work! Just curious (I'm not familiar with Terraform at all), why doesn't the stackType field need to be added to mmv1/products/compute/Instance.yaml?

`exclude_resource: true` exists in the [file](https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/products/compute/Instance.yaml#L19) to prevent Instance.yaml being used. compute instance is still a handwritten resource. It is hard to migrate it to MMv1 generated resource.

zli82016:
> > it is the same case with IPV4_IPV6 instances. If no ipv6_access_config is specified, but external IPv6 subnet is specified, the API will auto-fill the ipv6_access_config field for the result instance.
> 
> So the behavior persists. I will look into how do we deal with these sec policies inside our Update function and try to normalize this for our implementation but for the future this field should get deprecated and moved into `access_config` and `ipv6_access_config`. Not sure if this would be a breaking change. Should we wait for the major release to do this?

it is not a breaking change to deprecate a field and add a new field in a different place. But we should make sure the new field and the old field both work. The deprecated field can be removed in the next major release 7.0.

mmv1/third_party/terraform/services/compute/resource_compute_instance.go.tmpl
zli82016: Or it is OK for the users to see the diffs now?  Or we can just suppress the diffs for the new `IPV6_ONLY` stacktype?

ChuAmberZhang: > @karolgorc , @ChuAmberZhang , adding `Computed: true` will suppress the diffs. But it will prevent the removal of `ipv6_access_config` for users, as the previous value will be used. Do you think it will affect a lot of current users with `IPV4_IPV6` and `IP4_ONLY` stackType as `ipv6_access_config` cannot be removed? Based on my understanding, `google_compute_subnetwork`, `google_compute_instance`, `google_compute_instance_template` and `google_compute_region_instance_template` are widely used resources.

Hi @zli82016 do you mean the field can't be removed from the request input or output? There are a few cases to consider:

1. This field is not supposed to be populated for `IPV4_ONLY` stack type interfaces and `IPV4_IPV6` stack type interfaces with `INTERNAL `IPv6 access type, and `IPV6_ONLY` stack type interfaces with `INTERNAL` IPv6 access type
2. `IPV4_IPV6` and `IPV6_ONLY` interfaces with `EXTERNAL` IPv6 access type will have this field populated and this field should not be modifiable by the user (or any change attempt will be ignored or error out). When the stack type is updated to `IPV4_ONLY`, the field will be cleared automatically. 
3. When creating new `IPV4_IPV6` or `IPV6_ONLY` interfaces or changing the stack type from `IPV4_ONLY` to `IPV4_IPV6`, users should be able to specify this field in the request.
4. For a single instance, different interfaces may have different stack types and IPv6 access type, and thus may have different `ipv6_access_config` fields.

It's quite complicated, so we need to make sure if we set `Compute: true` the above cases are still honored.

zli82016: With `Compute: true`, even this field is removed from the resource configuration, this field will still be sent to the API with the field's previous value. 

`Computed: true` will be added always for all of the cases.  There is no way to only set this field computed in some cases.

zli82016: Another option is to add the [DiffSuppressFunc](https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/third_party/terraform/services/compute/resource_compute_instance.go.tmpl#L263) to this field to only suppress diffs for the case: IPV4_IPV6 and IPV6_ONLY interfaces with EXTERNAL IPv6 access type . This function will suppress the diffs between the empty value in resource configuration (new value) and the default value stored in Terraform state (old value). 

There are two issues with this option:
1. When this field is removed from the resource configuration, the diffs are suppressed by this function. It is the same issue with `Compute: true`. 
2. I doubt if there is a way to know IPv6 access type is EXTERNAL

karolgorc: This can't be diff suppressed. The info on whether the subnetwork is `External` or `Internal` isn't held by the Instance. It's held by the subnetwork. 

Doing API calls on external resources to suppress a diff seems like a bad design to me.

We can add a condition to the flattening function that will set this field to null unless it's specified by the user in his config. Making this show no diff but it will still get the right values from the API. Something like [this](https://github.com/GoogleCloudPlatform/magic-modules/pull/12552). I will have to test how overriding this works.

> But it will prevent the removal of ipv6_access_config for users, as the previous value will be used.

I don't think so. If the user specifies it he can just update this field. But if not i think it's expected that ipv6_access_config is set automatically when the instance is given a subnetwork with external addresses.

zli82016: > > But it will prevent the removal of ipv6_access_config for users, as the previous value will be used.
> 
> I don't think so. If the user specifies it he can just update this field. But if not i think it's expected that ipv6_access_config is set automatically when the instance is given a subnetwork with external addresses.
> 

With the `Computed: true + Optional: true`, when the user wants to remove `ipv6_access_config` from the configuration, its previous value from the Terraform state will be used to be sent to API. That is how `Computed: true + Optional: true` works.

zli82016: > To sum things up i think that having this as `Computed: true` is a must. Yeah we can just do diff suppress instead of this but this will restrict API's functionality. The user will not be able to query for essential values like the IP on this access config.

Adding `Compute: true` makes it impossible to remove `ipv6_access_config` (setting it to nil from Terraform), even `ipv6_access_config` is removed from the resource configuration. It will affect all of the stack types. I don't think that is what we want.

zli82016: Do you think that the diffs must be suppressed now? Do we consider leaving the diffs right now? I wonder if you have received  a lot of issues related to the diffs for `IPV4_IPV6` stack type?

karolgorc: Not sure if leaving the diff is optimal when shipping this feature. But if you think that removing the `Computed`, adding tests, and maybe a doc update to inform about this is enough. They yeah, for sure.

karolgorc:
Added testing for instance and subnetwork

mmv1/third_party/terraform/services/compute/resource_compute_instance.go.tmpl
zli82016: We can leave the diffs and remove the `Computed` now. You can create a Github issue with this [link](https://github.com/hashicorp/terraform-provider-google/issues/new?assignees=&labels=bug&projects=&template=00_bug.yml) for the diffs

zli82016 (changes requested):

zli82016 (changes requested):
Tests failed.

karolgorc:
These tests were made for computed to be true. Will update them and maybe specify this behaviour in the docs

zli82016 (approved):
LGTM

PR #12487, opened 2024-12-04
Make `google_compute_instance_guest_attributes` return empty results when queried for nonexistent values

karolgorc:
closes https://github.com/hashicorp/terraform-provider-google/issues/20346

- Removed error thrown on Instances.GetGuestAttributes calls
- Added additional Instances.Get call to see if the instance exists at all

```release-note:enhancement
compute: `google_compute_instance_guest_attributes` will return empty arrays and lists when queried values don't exist instead of throwing an error
```

### DIFF
--- a/mmv1/third_party/terraform/services/compute/data_source_google_compute_instance_guest_attributes.go.tmpl
+++ b/mmv1/third_party/terraform/services/compute/data_source_google_compute_instance_guest_attributes.go.tmpl
@@ -92,9 +92,14 @@ func dataSourceGoogleComputeInstanceGuestAttributesRead(d *schema.ResourceData, 
 		return err
 	}
 
+	//Check if instance exists
 	id := fmt.Sprintf("projects/%s/zones/%s/instances/%s", project, zone, name)
-	instanceGuestAttributes := &compute.GuestAttributes{}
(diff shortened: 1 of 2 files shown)


roaks3 (commented):
Could you add a test to confirm the behavior?

I'm also a little unsure about the solution here. It looks like this data source does not represent a list, so I would generally expect it to follow https://github.com/hashicorp/terraform-provider-google/issues/12873 and return an error. For example, I think that would be the expected behavior when `variable_key` is used.

However, this datasource has a `query_path` parameter that makes it a bit weird, and I'm not clear on what behavior we would want. The error the users sees now is obviously not clear, but it seems like technically we would still want an error returned. Could you share more about the use case, and why the user would be purposefully querying for infrastructure that doesn't exist?

roaks3 (changes requested):
Oh I see, I misunderstood the `query_value` field. This does actually look more like a plural datasource, and I don't think a 404 would make sense when the list is empty.

Same comment as before though, we should be able to add a test with an instance that does not have any guest attributes. This would confirm the error with the old behavior, and that it is fixed with this change.

karolgorc:
Added test cases for `empty` and `nonexistent` scenarios

roaks3 (approved):

PR #12488, opened 2024-12-04
Make nodepool concurrent ops scale better

Xylosma:
The purpose of this change is to stagger operation creations, but still keeps operations concurrent, so that operations are not bottlenecked on creation and fail. 
To achieve the above, 
* a cluster-scope lock is held for operation creation
*  cluster-scope lock is released after creation is done
* cluster-scope read lock is held while waiting for operation to complete (as before)

```release-note:enhancement
container: make nodepool concurrent operations scale better
```


ScottSuarez (changes requested):
We should look to upstream this locking behavior to `retryWhileIncompatibleOperation`.

This resource already has custom locking behavior on these calls. If we want to customize that behavior we should customize that function. That would mean passing that operation back and doing the wait within that function rather then introducing a secondary locking mechanism. Otherwise this will be difficult to interpret and new additions will also need to remember to implement this same behavior you have here.

https://github.com/hashicorp/terraform-provider-google-beta/blob/d02bfdd327ae6fbef0aa744c8c4c51c3abd3ba11/google-beta/services/container/resource_container_node_pool.go#L1693

mmv1/third_party/terraform/services/container/resource_container_node_pool.go.tmpl
```diff
@@ -1728,13 +1672,24 @@ func containerNodePoolAwaitRestingState(config *transport_tpg.Config, name, proj
...
+		transport_tpg.MutexStore.Lock(clusterLockKey)
+		op, err := createOpFunc()
+		transport_tpg.MutexStore.Unlock(clusterLockKey)
+		if err != nil {
+			return err
+		}
+		transport_tpg.MutexStore.RLock(clusterLockKey)
+		defer transport_tpg.MutexStore.RUnlock(clusterLockKey)
+		return waitOpFunc(op)
+	}
```
ScottSuarez, line 1692: So I don't believe this to actually be changing anything after going through the behavior loop in my head. `LockedCall(createLockKey, ...)` holds the lock for the entire operation (both creation and waiting). This means the additional locking/unlocking with `waitLockKey` inside the operation serves no purpose, as the outer lock already prevents concurrent execution.

Also, I don't believe concurrent updates are supported in the provider. all operations right now are concurrent

ScottSuarez (changes requested):
I think it would be a good idea to sync over a gvc and talk about this a bit!

mmv1/third_party/terraform/services/container/resource_container_node_pool.go.tmpl
Xylosma: So the goal is that we don't want multiple nodepools within the same cluster to create operations at exactly the same time. Currently, users may trigger creation of multiple (a lot of) nodepools for a single cluster using terraform and they will be created at the exact same time, overwhelming the GKE pipeline. The nodepool lock is fine, because we want sequential creation at cluster level.
Also happy to chat more on GVC.

mmv1/third_party/terraform/services/container/resource_container_node_pool.go.tmpl
```diff
@@ -1688,13 +1672,24 @@ func containerNodePoolAwaitRestingState(config *transport_tpg.Config, name, proj
...
-// retried until the incompatible operation completes, and the newly
-// requested operation can begin.
-func retryWhileIncompatibleOperation(timeout time.Duration, lockKey string, f func() error) error {
+// Retries an operation while the canonical error code is FAILED_PRECONDTION or RESOURCE_EXHAUSTED which indicates
+// there is an incompatible operation already running on the cluster or there are the number of allowed concurrent
+// operations running on the cluster. These errors can be safely retried until the incompatible operation completes,
+// and the newly requested operation can begin.
+// The cluster lock is held during createOpFunc to make opeation creations sequencial, and cluster read lock is held
+// during waitOpFunc to allow concurrency.
+func retryWhileIncompatibleOperation(timeout time.Duration, lockKey string, clusterLockKey string, createOpFunc func() (*container.Operation, error), waitOpFunc func(*container.Operation) error) error {
```
ScottSuarez, line 1681:
```suggestion
// Retries an operation while the canonical error code is FAILED_PRECONDTION or RESOURCE_EXHAUSTED which indicates
// there is an incompatible operation already running on the cluster or there are the number of allowed concurrent
// operations running on the cluster. These errors can be safely retried until the incompatible operation completes,
// and the newly requested operation can begin.
// The npLockKey is held during createOpFunc to make operations sequential within the node pool, and
// clusterLockKey is held during waitOpFunc to allow concurrency on a cluster.
func retryWhileIncompatibleOperation(timeout time.Duration, npLockKey string, clusterLockKey string, createOpFunc func() (*container.Operation, error), waitOpFunc func(*container.Operation) error) error {
```

mmv1/third_party/terraform/services/container/resource_container_node_pool.go.tmpl
ScottSuarez: I believe I understand it a bit better after drawing a diagram. I worry a bit about the readability of this code given that need. I'm okay approving this as is, but could someone from your team also review because the maintenance will fall more broadly there.

```mermaid
sequenceDiagram
    participant Caller
    participant Retry
    participant LockedCall
    participant MutexStore
    participant CreateOp as createOpFunc
    participant WaitOp as waitOpFunc

    Caller->>Retry: retryWhileIncompatibleOperation(timeout, npLockKey, clusterLockKey, ...)
    
    activate Retry
    Note over Retry: Start retry loop with timeout
    
    Retry->>LockedCall: LockedCall(npLockKey, f)
    activate LockedCall
    
    LockedCall->>MutexStore: Lock(npLockKey)
    Note over MutexStore: Write lock on nodepool
    
    LockedCall->>MutexStore: Lock(clusterLockKey)
    Note over MutexStore: Write lock on cluster
    
    LockedCall->>CreateOp: createOpFunc()
    activate CreateOp
    CreateOp-->>LockedCall: Return operation
    deactivate CreateOp
    
    LockedCall->>MutexStore: Unlock(clusterLockKey)
    Note over MutexStore: Release cluster write lock
    
    LockedCall->>MutexStore: RLock(clusterLockKey)
    Note over MutexStore: Read lock on cluster
    
    LockedCall->>WaitOp: waitOpFunc(op)
    activate WaitOp
    WaitOp-->>LockedCall: Return result
    deactivate WaitOp
    
    LockedCall->>MutexStore: RUnlock(clusterLockKey)
    Note over MutexStore: Release cluster read lock
    
    LockedCall->>MutexStore: Unlock(npLockKey)
    Note over MutexStore: Release nodepool lock (via defer)
    
    alt Operation succeeded
        LockedCall-->>Retry: Return nil
        Retry-->>Caller: Return nil
    else IsFailedPreconditionError or IsQuotaError
        LockedCall-->>Retry: Return error
        Note over Retry: Retry operation
        Retry->>LockedCall: LockedCall(npLockKey, f)
    else Other error
        LockedCall-->>Retry: Return error
        Retry-->>Caller: Return error immediately
    end
    
    deactivate LockedCall
    deactivate Retry
```

ScottSuarez (changes requested):

mmv1/third_party/terraform/services/container/node_config.go.tmpl
```diff
@@ -2029,10 +2029,13 @@ func flattenHostMaintenancePolicy(c *container.HostMaintenancePolicy) []map[stri
 // node pool updates in `resource_container_cluster`
 func nodePoolNodeConfigUpdate(d *schema.ResourceData, config *transport_tpg.Config, nodePoolInfo *NodePoolInformation, prefix, name string, timeout time.Duration) error {
 
-	// Nodepool write-lock will be acquired when update function is called.
+	// Cluster write-lock will be acquired when createOpF is called, and read-lock will be acquired when waitOpF is
```
alexmorozov, line 2032: May be worthwhile adding the "why" here. Like, why are we taking the write lock when `createOfF` is called?

alexmorozov (approved):
Looks good overall, thanks for making the change. I'd ask to uplevel the comments in the code that refer to which functions take which locks. That part is more or less evident from the code itself - but it's not clear to an external viewer _why_ that is done. You explain that in a later comment, but I would not expect the reader to be able to find it without guidance.

mmv1/third_party/terraform/services/container/resource_container_node_pool.go.tmpl
Xylosma: Alex from the team has reviewed the PR.

ScottSuarez: Perfect. Are there any other changes you wish to make? If not I can approve. I want to save space for that if so.

mmv1/third_party/terraform/services/container/resource_container_node_pool.go.tmpl
Xylosma: I probably should not have committed the suggestion because it was incorrect.
npLock is held through out this function, just for mutex reason. cluster WRITE lock is held during createOpFunc for sequential creation, and cluster READ lock is held during waitOpFunc to allow concurrency.

ScottSuarez: ah ! sorry about that!

mmv1/third_party/terraform/services/container/resource_container_node_pool.go.tmpl
Xylosma: Now it's ready : )

ScottSuarez (approved):

Xylosma:
Thank you Scott and Alex for the review!

mmv1/third_party/terraform/services/container/resource_container_node_pool.go.tmpl
```diff
@@ -1688,15 +1672,27 @@ func containerNodePoolAwaitRestingState(config *transport_tpg.Config, name, proj
...
+// and clusterLockKey read-lock is held during waitOpFunc to allow concurrency on a cluster.
+func retryWhileIncompatibleOperation(timeout time.Duration, npLockKey string, clusterLockKey string, createOpFunc func() (*container.Operation, error), waitOpFunc func(*container.Operation) error) error {
+	f := func() error {
+		transport_tpg.MutexStore.Lock(clusterLockKey)
+		op, err := createOpFunc()
+		transport_tpg.MutexStore.Unlock(clusterLockKey)
+		if err != nil {
+			return err
+		}
+		transport_tpg.MutexStore.RLock(clusterLockKey)
```
KatrinaHoffert, line 1690: Why are we getting a read lock here? I believe this to be the cause of a regression we've detected. This read lock means that we cannot create new ops while ops are in progress. I believe we only ever wanted the write lock above, to serialize the op creation (temporarily, until concurrency support is improved). I believe this effectively breaks parallelization and thus considerably increases time-to-completion.

ScottSuarez: @Xylosma

Xylosma: Already discussed offline, we will revert this PR first for mitigation. It was the contention between the new acquisition of write lock and pre-existing read lock that caused the regression.

PR #12493, opened 2024-12-04
Allow brand new setup for google_developer_connect_connection

rainshen49:

```release-note:enhancement
developerconnect: added support for setting up a brand new `google_developer_connect_connection`
```

### DIFF
--- a/mmv1/products/developerconnect/Connection.yaml
+++ b/mmv1/products/developerconnect/Connection.yaml
@@ -45,13 +45,24 @@ async:
     message: 'message'
 custom_code:
 examples:
-  - name: 'developer_connect_connection_basic'
+  - name: 'developer_connect_connection_new'
     primary_resource_id: 'my-connection'
-    primary_resource_name: 'fmt.Sprintf("tf-test-connection%s", context["random_suffix"])'
+    primary_resource_name: 'fmt.Sprintf("tf-test-connection-new%s", context["random_suffix"])'
     min_version: 'beta'
     vars:
-      connection_name: 'tf-test-connection'
-  - name: 'developer_connect_connection_github_doc'
+      connection_name: 'tf-test-connection-new'
+    test_env_vars:
+      project: 'PROJECT_NAME'
+  - name: 'developer_connect_connection_existing_credentials'
+    primary_resource_id: 'my-connection'
+    primary_resource_name: 'fmt.Sprintf("tf-test-connection-cred%s", context["random_suffix"])'
+    min_version: 'beta'
+    vars:
+      connection_name: 'tf-test-connection-cred'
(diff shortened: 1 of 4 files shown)


rainshen49:
@jimmygns

mmv1/products/developerconnect/Connection.yaml
```diff
@@ -45,13 +45,24 @@ async:
     message: 'message'
 custom_code:
 examples:
-  - name: 'developer_connect_connection_basic'
+  - name: 'developer_connect_connection_new'
```
BBBmau, line 48: this looks good but we should keep the basic example still, you seem to have replaced it with your new example but didn't delete the basic tfconfig file.

So in the end we should have `[basic, new, developer_connect_connection_existing_credentials]

rainshen49: The old basic example became the `existing_credentials` example. So it's really two tests [new, existing_credentials] + docs

mmv1/products/developerconnect/Connection.yaml
```diff
@@ -99,6 +110,7 @@ properties:
           Represents an OAuth token of the account that authorized the Connection,and
           associated metadata.
         min_version: 'beta'
+        default_from_api: true
```
BBBmau, line 113: My only concern is the inclusion of `default_from_api: true` on these two properties. What is the reason for this?

rainshen49: Sure, this requires some product knowledge, but let me try to explain. `default_from_api: true` means if the user leave it unspecified, Terraforms defers to the server provided value instead of sending an empty value that wipes out the server side value. Consider a typical config to set up a connection

```terraform
resource "google_developer_connect_connection" "default" {
  ...

  github_config {
    github_app = "FIREBASE"
    # Note the unspecified authorizer_credentials field
  }
}

output "next_steps" {
  description = "Follow the action_uri if present to continue setup"
  value = google_developer_connect_connection.default.installation_state
}
```

1. Run `terraform apply`, and the output will provide a URL for the user to visit for an OAuth flow
2. After the OAuth flow, `authorizer_credentials` is populated on the server side, but the Terraform configuration doesn't know about it
3. If the user runs `terraform apply` again on the same config, without `default_from_api`, terraform will detect a diff and accidentally wipe out the `authorizer_credentials`
4. The correct behavior is that terraform will refresh to capture the `authorizer_credentials` populated outside of Terraform, and doesn't complain about a diff.
5. If the user wants to use another `authorizer_credentials`, they can do so and Terraform will send the new value to the server

BBBmau: Thanks for the explanation, this is all I needed!

BBBmau (approved):

PR #12496, opened 2024-12-05
Fix settings.data_cache_config permadiff when set to false

algawivem:
When settings.data_cache_config false a permadiff will result. This is due to the fact that CloudSQL API will not return  data_cache_config when there is no data cache. Update terraform to match CloudSQL API by representing non existent data cache config as set to false .

Mentioned in:
https://github.com/hashicorp/terraform-provider-google/issues/15227
https://github.com/hashicorp/terraform-provider-google/issues/19144

```release-note:enhancement
sql: fixed permadiff when 'settings.data_cache_config' is set to false for 'google_sql_database_instance' resource
```


slevenick (approved):
Thanks!

PR #12499, opened 2024-12-05
Add API hub product, with resource ApiHubInstance

edwinfranciss:
This PR creates support for new GCP product API hub - https://cloud.google.com/apigee/docs/apihub/what-is-api-hub
As part of the product, added support for the resource ApiHubInstance (google_apihub_api_hub_instance) 

```release-note:new-resource
`google_apihub_api_hub_instance`
```


mmv1/products/apihub/product.yaml
```diff
@@ -0,0 +1,22 @@
...
+
+---
+name: Apihub
+display_name: API hub
+scopes:
+- https://www.googleapis.com/auth/cloud-platform
+versions:
+- base_url: https://apihub.googleapis.com/v1/
+  name: ga
+caibaseurl: ""
```
roaks3, line 22: Nit: probably just autogen-ed, but you should be able to remove `caibaseurl`

mmv1/templates/terraform/examples/apihub_api_hub_instance_basic.tf.tmpl
```diff
@@ -0,0 +1,8 @@
+resource "google_apihub_api_hub_instance" "{{$.PrimaryResourceId}}"{
```
roaks3, line 1: You'll need some more tests to check the range of config that can be provided (see https://googlecloudplatform.github.io/magic-modules/test/test/):
- Since this test should be minimal, is `config {}` valid?
- You'll need an "update" test to confirm fields that are mutable can be updated in-place, like description and config
- You'll need to use each field in at least one test (you can do this with the update test, or add an explicit `*_full.tf.tmpl` example with every possible field specified)

roaks3 (changes requested):
Overall LGTM, but will need some more tests

mmv1/products/apihub/ApiHubInstance.yaml
```diff
@@ -0,0 +1,140 @@
...
+id_format: projects/{{project}}/locations/{{location}}/apiHubInstances/{{api_hub_instance_id}}
+import_format:
+- projects/{{project}}/locations/{{location}}/apiHubInstances/{{api_hub_instance_id}}
+examples:
+- name: apihub_api_hub_instance_basic
+  primary_resource_id: apihub-instance
+  vars:
+    instance_id: test-instance
+autogen_async: true
+exclude_delete: true
```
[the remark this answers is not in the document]
edwinfranciss: Yes, currently we do not support delete for this resource.

mmv1/products/apihub/ApiHubInstance.yaml
```diff
@@ -0,0 +1,140 @@
...
+- name: location
+  type: String
+  description: Resource ID segment making up resource `name`. It identifies the resource within its parent collection as described in https://google.aip.dev/122.
+  immutable: true
+  url_param_only: true
+  required: true
+- name: apiHubInstanceId
+  type: String
+  description: |-
+    Optional. Identifier to assign to the Api Hub instance. Must be unique within
```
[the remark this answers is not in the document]
edwinfranciss: Removed the required condition.

mmv1/products/apihub/product.yaml
edwinfranciss: Thanks for pointing it out. Have removed caibaseurl

mmv1/templates/terraform/examples/apihub_api_hub_instance_basic.tf.tmpl
edwinfranciss: - config{} wont be valid. One of disable_search or vertex_location would need to be provided. Default value of disable_search would be false. But in that scenario, vertex_location would need to be provided in the config.
- Currently update is also not supported for this resource
- Currently only one resource/project can be provisioned. And since delete and update is not supported for this resource, was not able to add further test in the same flow. Do you have any suggestions for this scenario.

edwinfranciss: I have added a test with all possible fields, which will be provisioned in a separate project. Hope that is fine?

mmv1/products/apihub/ApiHubInstance.yaml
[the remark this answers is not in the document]
roaks3: So if I'm understanding, only one of these resources can be created per project, it is immutable, and it cannot be deleted, meaning that once an instance is created for a particular project, you can never change it, and you can't touch the Terraform config unless `project` is changed (because TF will attempt a re-creation, which won't work).

Is this the intended behavior? It seems a little confusing to a user to not be able to adjust their config, so perhaps we want to mention that a new project must be used somewhere in the documentation?

mmv1/templates/terraform/examples/apihub_api_hub_instance_basic.tf.tmpl
roaks3: For both tests, could you create a separate project for the instance? Otherwise our test environments will become polluted in a way we seemingly cannot clean up. Ie.:
```
resource "google_project" "project" {
  name       = "{{index $.Vars "project_id"}}"
  project_id = "{{index $.Vars "project_id"}}"
  org_id     = "{{index $.TestEnvVars "org_id"}}"
  deletion_policy = "DELETE"
}
```
And in the examples blocks:
```
    vars:
      project_id: 'apihub-proj'
    test_env_vars:
      org_id: 'ORG_ID'
```

mmv1/products/apihub/ApiHubInstance.yaml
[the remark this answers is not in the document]
edwinfranciss: I have updated the description of the resource, mentioning that the resource cannot be deleted/updated as of now.

mmv1/templates/terraform/examples/apihub_api_hub_instance_basic.tf.tmpl
edwinfranciss: As discussed offline, currently using exclude_test. Have created a separate bug and once one of the pre-requisite bug is competed, will remove this flag.

mmv1/templates/terraform/examples/apihub_api_hub_instance_full.tf.tmpl
```diff
@@ -0,0 +1,15 @@
+resource "google_apihub_api_hub_instance" "{{$.PrimaryResourceId}}"{
+    project = "ah-terraform-provider-1"
```
roaks3, line 2: Nit: You are excluding this test, but it will still show up as an example in the documentation. We may want to tweak this just for clarity to something like `my-project` here and in cmek_key_name.

roaks3 (commented):
One nit, and I'd be curious if you have any more detail regarding the PreConfig option, but otherwise LGTM

mmv1/templates/terraform/examples/apihub_api_hub_instance_full.tf.tmpl
edwinfranciss: Have updated the project used in the example.

edwinfranciss: As discussed offline, there were issues when calling the 1P API using the PreConfig option. The tests will be updated once we have the delete functionality for the resource as well.

roaks3 (approved):

PR #10900, opened 2024-06-06
update protected metadata

bcreddy-gcp:
**Release Note Template for Downstream PRs (will be copied)**

```release-note:enhancement
workbench: Updated the metadata keys managed by the backend.
```

PR #12505, opened 2024-12-05
`gcs-data-bucket` metadata key is now modifiable by the user

bcreddy-gcp:

```release-note: note
workbench : Made `gcs-data-bucket` metadata key modifiable. 
```


roaks3 (commented):
I can't quite tell if this is a breaking change, per https://googlecloudplatform.github.io/magic-modules/breaking-changes/breaking-changes/#field-level-breaking-changes. Do you know what happens if a user has this field specified right now, and then upgrades the provider? It seems like it could produce a new change that could be unexpected. Might be worth a test as well to demonstrate what happens.

bcreddy-gcp:
This key was never allowed to be set nor was set by the backend. We now allow instances to have it set using update. Also newer instances can be created with the said key.

roaks3 (approved):
I was more considering if they had the key currently in their config, even though it is ignored:
```
    metadata = {
      gcs-data-bucket = "some-value"
    }
```

I believe someone could be using this config today, and then with this change, it would suddenly apply the "some-value".

But assuming that's correct, this does seem like fairly unusual usage, and I expect we would be fine including it in a minor version. I also see that there is precedent (we removed `nvidia-driver-gcs-path` in https://github.com/GoogleCloudPlatform/magic-modules/pull/10900).

PR #12510, opened 2024-12-06
Add nullFields logic to `guest_flush` on `resource_google_compute_resource_policy`

karolgorc:
closes https://github.com/hashicorp/terraform-provider-google/issues/20499
closes https://github.com/hashicorp/terraform-provider-google/issues/16936

This field cannot be patched and is set to false automatically when the parent block is added into the configuration causing an error on every change to this block.

```release-note:enhancement
compute: Updating labels on `resource_google_compute_resource_policy` will no longer fail because of a Patch error with `guest_flush`
```

### DIFF
--- a/mmv1/products/compute/ResourcePolicy.yaml
+++ b/mmv1/products/compute/ResourcePolicy.yaml
@@ -258,7 +258,7 @@ properties:
             type: Boolean
             description: |
               Whether to perform a 'guest aware' snapshot.
-            send_empty_value: true
+            send_empty_value: false
             at_least_one_of:
               - 'snapshot_schedule_policy.0.snapshot_properties.0.labels'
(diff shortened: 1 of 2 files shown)


mmv1/products/compute/ResourcePolicy.yaml
```diff
@@ -258,7 +258,7 @@ properties:
             type: Boolean
             description: |
               Whether to perform a 'guest aware' snapshot.
-            send_empty_value: true
```
ScottSuarez, line 261: Can you add a testcase for this ?

ScottSuarez (changes requested):

mmv1/products/compute/ResourcePolicy.yaml
karolgorc: on it

ScottSuarez (approved):

PR #12514, opened 2024-12-06
Remove async operation.(waitms and path)

ScottSuarez:
helps with https://github.com/hashicorp/terraform-provider-google/issues/17098

```release-note:none

```

### DIFF
--- a/mmv1/api/async.go
+++ b/mmv1/api/async.go
@@ -83,11 +83,7 @@ type OpAsync struct {
 type OpAsyncOperation struct {
 	Kind string `yaml:"kind,omitempty"`
 
-	Path string `yaml:"path,omitempty"`
-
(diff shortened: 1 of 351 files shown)


mmv1/api/async.go
```diff
@@ -83,12 +83,8 @@ type OpAsync struct {
 type OpAsyncOperation struct {
 	Kind string `yaml:"kind,omitempty"`
 
-	Path string `yaml:"path,omitempty"`
-
 	BaseUrl string `yaml:"base_url,omitempty"`
 
-	WaitMs int `yaml:"wait_ms,omitempty"`
-
```
ScottSuarez, line 91: I might need to leave it in the serializer otherwise this will break any existing PRs that use this.

PR #12515, opened 2024-12-06
Remove async result.(path) , status.('path', 'allowed') and error.('path', 'message')

ScottSuarez:
helps with https://github.com/hashicorp/terraform-provider-google/issues/17098

```release-note:none

```

### DIFF
--- a/mmv1/products/tpuv2/QueuedResource.yaml
+++ b/mmv1/products/tpuv2/QueuedResource.yaml
@@ -30,8 +30,6 @@ async:
   type: 'OpAsync'
   operation:
     base_url: '{{op_id}}'
-    path: 'name'
-    wait_ms: 1000
   result:
     path: 'response'
     resource_inside_response: true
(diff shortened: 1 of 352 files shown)


mmv1/products/tpuv2/QueuedResource.yaml
```diff
@@ -31,17 +31,11 @@ async:
...
     base_url: '{{op_id}}'
   result:
-    path: 'response'
     resource_inside_response: true
   status:
-    path: 'done'
     complete: true
-    allowed:
       - true
       - false
```
melinath, line 38:
```suggestion
```

melinath (changes requested):

mmv1/products/tpuv2/QueuedResource.yaml
ScottSuarez: ah yeah, just ran into this after running the serializer locally

PR #12522, opened 2024-12-09
Add Intercept Endpoint Group resource to Network Security.

duvni:

```release-note:new-resource
`google_network_security_intercept_endpoint_group` (beta)
```

### DIFF
--- a/mmv1/products/networksecurity/InterceptEndpointGroup.yaml
+++ b/mmv1/products/networksecurity/InterceptEndpointGroup.yaml
@@ -1,0 +1,110 @@
+# Copyright 2024 Google Inc.
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+---
+name: 'InterceptEndpointGroup'
+description: An intercept endpoint group is a global resource in the consumer account representing the producer’s deployment group.
+min_version: 'beta'
+docs:
+id_format: 'projects/{{project}}/locations/{{location}}/interceptEndpointGroups/{{intercept_endpoint_group_id}}'
+base_url: 'projects/{{project}}/locations/{{location}}/interceptEndpointGroups'
+self_link: 'projects/{{project}}/locations/{{location}}/interceptEndpointGroups/{{intercept_endpoint_group_id}}'
+create_url: 'projects/{{project}}/locations/{{location}}/interceptEndpointGroups?interceptEndpointGroupId={{intercept_endpoint_group_id}}'
+update_verb: 'PATCH'
+update_mask: true
+import_format:
+  - 'projects/{{project}}/locations/{{location}}/interceptEndpointGroups/{{intercept_endpoint_group_id}}'
+timeouts:
+  insert_minutes: 20
+  update_minutes: 20
+  delete_minutes: 20
+autogen_async: true
+async:
+  actions: ['create', 'delete', 'update']
+  type: 'OpAsync'
+  operation:
+    base_url: '{{op_id}}'
+    path: 'name'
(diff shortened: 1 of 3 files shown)


mmv1/products/networksecurity/InterceptEndpointGroup.yaml
```diff
@@ -0,0 +1,110 @@
...
+  - name: 'location'
+    type: String
+    description: 'Resource ID segment making up resource `name`. It identifies the resource
+    within its parent collection as described in https://google.aip.dev/122. See documentation
+    for resource type `networksecurity.googleapis.com/InterceptEndpointGroup`. '
+    min_version: 'beta'
+    url_param_only: true
+    required: true
+    immutable: true
+  - name: 'interceptEndpointGroupId'
```
SirGitsalot, line 64: No action required, just noting this here for posterity: this field would normally be called `name` and the output-only `name` field in the body would be omitted. Not requiring a change here because while it's inconsistent with most of the other resources in TPG it's consistent with most of the resources in `networksecurity` (and it's too late to change those).

mmv1/products/networksecurity/InterceptEndpointGroup.yaml
```diff
@@ -0,0 +1,110 @@
...
+    config_path: 'templates/terraform/examples/network_security_intercept_endpoint_group_basic.tf.tmpl'
+    primary_resource_id: 'default'
+    vars:
+      network_name: 'example-network'
+      deployment_group_id: 'example-dg'
+      endpoint_group_id: 'example-eg'
+parameters:
+  - name: 'location'
+    type: String
+    description: 'Resource ID segment making up resource `name`. It identifies the resource
```
SirGitsalot, line 57: This could use a more user friendly description, as this is what ends up in the documentation. It can be short and simple, something like "The location of the Intercept Endpoint Group, for example `us-central1`."

mmv1/products/networksecurity/InterceptEndpointGroup.yaml
```diff
@@ -0,0 +1,110 @@
...
+    within its parent collection as described in https://google.aip.dev/122. See documentation
+    for resource type `networksecurity.googleapis.com/InterceptEndpointGroup`. '
```
SirGitsalot, line 59: All of the descriptions have an extra space before the terminating quote (from the autogen tool maybe?). Please fix.

mmv1/products/networksecurity/InterceptEndpointGroup.yaml
```diff
@@ -0,0 +1,110 @@
...
+    min_version: 'beta'
+    url_param_only: true
+    required: true
+    immutable: true
+  - name: 'interceptEndpointGroupId'
+    type: String
+    description: "Id of the requesting object\nIf auto-generating Id server-side,
```
SirGitsalot, line 66: Ditto on the user friendly description, especially since the id will not be generated server side (and if there's a limitation on the format of the name, please mention that - e.g., Security Endpoint Group has:

>This value should be 1-63 characters long, containing only letters, numbers, hyphens, and underscores, and should not start with a number. E.g. "security_profile_group1".

SirGitsalot (commented):
Mostly LGTM with a couple of nits. Thanks!

mmv1/products/networksecurity/InterceptEndpointGroup.yaml
duvni: Thanks, I also wondered about this, and was trying to track down exactly how this divergence happened. It seems like the current examples (in CCFE) actually do use `id` field instead of `name` for this, so possibly it was changed down the line. But I do find `id` to be slightly better here since `name` should be used for the long uri format (i.e.: `projects/proj/locations/us-east1/resources/id`).

mmv1/products/networksecurity/InterceptEndpointGroup.yaml
duvni: Done.
Also stated that `global` is the only valid option for now.

mmv1/products/networksecurity/InterceptEndpointGroup.yaml
duvni: Done.

mmv1/products/networksecurity/InterceptEndpointGroup.yaml
duvni: Done.

SirGitsalot (approved):

PR #12533, opened 2024-12-10
added iap field to support enable/disable in cloudrun services

Deepraj1996:
```release-note:enhancement
cloudrunv2: added `iap_enabled` field to `google_cloud_run_v2_service` (ga). 
```


bskaplan:
This feature is still in early testing, with access only allowed to a limited set of test projects, so it's not ready to be added to the published Terraform  provider yet.

Deepraj1996:
> This feature is still in early testing, with access only allowed to a limited set of test projects, so it's not ready to be added to the published Terraform provider yet.

Sure, Is it fine I make the changes in Terraform google beta provider?

PR #12535, opened 2024-12-10
Add vulnerability scanning config in Artifact Registry

jimmyqian2:
Support new configurations added to Artifact Registry's Repository object that control vulnerability scanning.

```release-note:enhancement
artifact_registry: added `vulnerability_scanning_config` field to `google_artifact_registry_repository` resource
```


mmv1/products/artifactregistry/Repository.yaml
```diff
@@ -736,3 +743,16 @@ properties:
     description: |-
       If true, the cleanup pipeline is prevented from deleting versions in this
       repository.
+  - name: 'vulnerabilityScanningConfig'
+    type: NestedObject
+    description: |-
+      Configuration for vulnerability scanning of artifacts stored in this repository.
+    allow_empty_object: true
```
melinath, line 750: It looks like this field may need default_from_api because the API returns a default if its unset (which is why most of the tests are failing.) See https://googlecloudplatform.github.io/magic-modules/develop/diffs/ for more info.

mmv1/products/artifactregistry/Repository.yaml
```diff
@@ -207,6 +207,13 @@ examples:
       # Ignore this field as it is INPUT_ONLY. AR will not return this in the
       # response.
       - 'remote_repository_config.0.disable_upstream_validation'
+  - name: 'artifact_registry_repository_vulnerability_scanning'
+    primary_resource_id: 'my-repo'
+    primary_resource_name: 'fmt.Sprintf("tf-test-my-repository%s", context["random_suffix"])'
+    vars:
+      repository_id: 'my-repository'
+      desc: 'example docker repository with vulnerability scanning config'
+      enablement_config: 'INHERITED'
```
melinath, line 216: This should fix this test's failures by ensuring this uses a literal value - or you could hardcode it in the test template (since it's not likely to be reused or modified.)

```suggestion
      enablement_config: 'INHERITED'
    test_vars_overrides:
      enablement_config: 'INHERITED'
```

mmv1/products/artifactregistry/Repository.yaml
jimmyqian2: Thanks for the tip, that fixed the failures.

mmv1/products/artifactregistry/Repository.yaml
jimmyqian2: Moved the field inline in the example.

mmv1/products/artifactregistry/Repository.yaml
```diff
@@ -736,3 +742,31 @@ properties:
     description: |-
       If true, the cleanup pipeline is prevented from deleting versions in this
       repository.
+  - name: 'vulnerabilityScanningConfig'
+    type: NestedObject
+    description: |-
+      Configuration for vulnerability scanning of artifacts stored in this repository.
+    allow_empty_object: true
```
melinath, line 749: This doesn't do anything since this is a default_from_api field.

```suggestion
```

melinath (changes requested):
Overall LGTM - looks like it matches the API implementation and is straightforward. Just one small tweak.

melinath (approved):

PR #12536, opened 2024-12-10
Add envvar GOOGLE_CHRONICLE_INSTANCE_ID for tests

roaks3:

```release-note:none

```


.ci/gcb-generate-diffs-new.yml
```diff
@@ -267,7 +267,7 @@ steps:
     - name: 'gcr.io/graphite-docker-images/go-plus'
       id: gcb-tpg-vcr-test
       entrypoint: '/workspace/.ci/scripts/go-plus/magician/exec.sh'
-      secretEnv: ["GITHUB_TOKEN_DOWNSTREAMS", "GITHUB_TOKEN_MAGIC_MODULES", "GOOGLE_BILLING_ACCOUNT", "GOOGLE_CUST_ID", "GOOGLE_IDENTITY_USER", "GOOGLE_MASTER_BILLING_ACCOUNT", "GOOGLE_ORG", "GOOGLE_ORG_2", "GOOGLE_ORG_DOMAIN", "GOOGLE_PROJECT", "GOOGLE_PROJECT_NUMBER", "GOOGLE_SERVICE_ACCOUNT", "SA_KEY", "GOOGLE_PUBLIC_AVERTISED_PREFIX_DESCRIPTION"]
+      secretEnv: ["GITHUB_TOKEN_DOWNSTREAMS", "GITHUB_TOKEN_MAGIC_MODULES", "GOOGLE_BILLING_ACCOUNT", "GOOGLE_CHRONICLE_INSTANCE_ID", "GOOGLE_CUST_ID", "GOOGLE_IDENTITY_USER", "GOOGLE_MASTER_BILLING_ACCOUNT", "GOOGLE_ORG", "GOOGLE_ORG_2", "GOOGLE_ORG_DOMAIN", "GOOGLE_PROJECT", "GOOGLE_PROJECT_NUMBER", "GOOGLE_SERVICE_ACCOUNT", "SA_KEY", "GOOGLE_PUBLIC_AVERTISED_PREFIX_DESCRIPTION"]
```
shuyama1, line 270: This file is removed in https://github.com/GoogleCloudPlatform/magic-modules/pull/12518

.ci/magician/cmd/check_cassettes.go
```diff
@@ -16,6 +16,7 @@ var ccEnvironmentVariables = [...]string{
 	"GOCACHE",
 	"GOPATH",
 	"GOOGLE_BILLING_ACCOUNT",
+	"GOOGLE_CHRONICLE_INSTANCE_ID",
```
shuyama1, line 19: I wonder if this is backward-compatible. It seems the current change requires all PRs to rebase to include the updates in the cloud build configuration .yml file. Otherwise, the build process might fail due to the missing environment variable `GOOGLE_CHRONICLE_INSTANCE_ID`.

roaks3: I think you could be right, but I don't know enough about how this is set up. Could you elaborate a bit? Is there a set of .ci functionality that runs off of main, and needs to be updated as a "step 1"? (also, is any of this documented for me to get clarification on my own?)

shuyama1: Among the files modified, `.ci/gcb-pr-downstream-generation-and-test.yml` (we usually call it the `generate-diff` job) is the only one that runs off the PR branch's head. The previously used `.ci/gcb-generate-diffs-new.yml` file has already been removed and replaced with this new configuration, we can safely remove it from this PR. This behavior occurs because the Cloud Build trigger is configured to respond to pull request events, invoking the configuration file from the PR branch.

In contrast, other YAML files are for triggers working directly with the main branch and can immediately adapt to code changes once merged.

For the `generate-diff` job, the Magician code should sync at the beginning of execution. Therefore, any updates to the Magician source code take immediate effect for all PRs once the change is merged.

To ensure backward compatibility when introducing a new environment variable, we generally have two options:

1. Split the change in two PRs. 

	- The first PR updates the configuration yml file
	- The second PR updates the Magician source code.
	- Allow a soak period after merging the first PR to ensure most PRs already have the updated configuration checked in before merging the second PR.

2. Make the source code backward compatible 

	- Currently, the Magician checks for existence of all env variables, as seen in this code:  
	``` 
	env := make(map[string]string, len(ttvEnvironmentVariables))
	for _, ev := range ttvEnvironmentVariables {
		val, ok := os.LookupEnv(ev)
		if !ok {
			return fmt.Errorf("did not provide %s environment variable", ev)
		}
		env[ev] = val
	}
	```
	(https://github.com/GoogleCloudPlatform/magic-modules/blob/main/.ci/magician/cmd/test_terraform_vcr.go#L108-L115)
	- To add a new env var without breaking existing functionality, we should make an exception for the new variable, ensuring it does not trigger an error if it is missing. This approach allows the change to take effect immediately for PRs rebased off main after the change is merged.

I hope this helps clarify things.

roaks3: Should be updated now with optional env vars, thanks for the help!

roaks3:
Test failures appear unrelated (same showing in other PRs like https://github.com/GoogleCloudPlatform/magic-modules/pull/12580)

shuyama1 (approved):

PR #12538, opened 2024-12-10
add Ephemeral-write-only subproject into teamcity

BBBmau:

```release-note:REPLACEME

```

### DIFF
--- a/mmv1/third_party/terraform/.teamcity/components/projects/feature_branches/FEATURE-BRANCH-ephemeral-write-only.kt
+++ b/mmv1/third_party/terraform/.teamcity/components/projects/feature_branches/FEATURE-BRANCH-ephemeral-write-only.kt
@@ -1,0 +1,129 @@
+/*
+ * Copyright (c) HashiCorp, Inc.
+ * SPDX-License-Identifier: MPL-2.0
+ */
+
+// This file is maintained in the GoogleCloudPlatform/magic-modules repository and copied into the downstream provider repositories. Any changes to this file in the downstream will be overwritten.
+
+package projects.feature_branches
+
+import ProviderNameBeta
+import ProviderNameGa
+import SharedResourceNameBeta
+import SharedResourceNameGa
+import SharedResourceNameVcr
+import builds.*
+import generated.ServicesListBeta
+import generated.ServicesListGa
+import jetbrains.buildServer.configs.kotlin.Project
+import replaceCharsId
+import vcs_roots.HashiCorpVCSRootBeta
+import vcs_roots.HashiCorpVCSRootGa
+import vcs_roots.ModularMagicianVCSRootBeta
+import vcs_roots.ModularMagicianVCSRootGa
+
+const val featureBranchEphemeralWriteOnly = "FEATURE-BRANCH-ephemeral-write-only"
+const val EphemeralWriteOnlyTfCoreVersion = "1.10.0" // will be changed to 1.11.0 when the new ephemeral values feature is released in release candidates
+
+// featureBranchEphemeralWriteOnlySubProject creates a project just for testing ephemeral write-only attributes.
+// We know that all ephemeral write-only attributes we're adding are part of the Resource Manager service, so we only include those builds.
+// We create builds for testing the resourcemanager service:
+//    - Against the GA hashicorp repo
+//    - Against the GA modular-magician repo
+//    - Against the Beta hashicorp repo
+//    - Against the Beta modular-magician repo
+// These resemble existing projects present in TeamCity, but these all use a more recent version of Terraform including
+// the new ephemeral values feature.
+fun featureBranchEphemeralWriteOnlySubProject(allConfig: AllContextParameters): Project {
+
+    val projectId = replaceCharsId(featureBranchEphemeralWriteOnly)
+
+    val vcrConfig = getVcrAcceptanceTestConfig(allConfig) // Reused below for both MM testing build configs
+    val trigger  = NightlyTriggerConfiguration(
+        branch = "refs/heads/$featureBranchEphemeralWriteOnly" // triggered builds must test the feature branch
+    )
+
+    // All Ephemeral Write-Only attributes are in the following packages
+    var PackagesListWriteOnly = mapOf(
+        "compute" to mapOf(
+            "name" to "compute",
+            "displayName" to "Compute",
+            "path" to "./google/services/compute"
+        ),
+        "secretmanager" to mapOf(
+            "name" to "secretmanager",
+            "displayName" to "Secretmanager",
+            "path" to "./google/services/secretmanager"
+        ),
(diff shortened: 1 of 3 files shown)


mmv1/third_party/terraform/.teamcity/components/get_services.kt
```diff
@@ -0,0 +1,55 @@
+/*
+ * Copyright (c) HashiCorp, Inc.
+ * SPDX-License-Identifier: MPL-2.0
+ */
+import generated.ServicesListGa
+import generated.ServicesListBeta
+
+// This file is maintained in the GoogleCloudPlatform/magic-modules repository and copied into the downstream provider repositories. Any changes to this file in the downstream will be overwritten.
+
+fun getServicesList(Services: Array<String>, version: String): Map<String,Map<String,String>> {
```
SarahFrench, line 10: Could you please add a comment about what the function is used for? Looks like the comment should say that it's used to create a list of a subset of services relevant to testing a feature branch, so only relevant builds are made.

Also, given it's so coupled to feature branch testing, what do you think about moving it into the feature_branches folder?

mmv1/third_party/terraform/.teamcity/components/get_services.kt
```diff
@@ -0,0 +1,55 @@
...
+        }
+    }
+
+    when (version) {
+        "GA" -> return servicesList
+        "Beta" -> {
+            servicesList = servicesList.mapValues { (_, value) ->
+                value + mapOf(
+                    "displayName" to "${value["displayName"]} - Beta",
+                    "path" to (value["path"]?.replace("./google/", "./google-beta/") ?: "")
```
SarahFrench, line 32: Is this transform necessary? I thought that when the version is "Beta" then information in [inputs/services_beta.kt](https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/third_party/terraform/.teamcity/components/inputs/services_beta.kt) would be used, which should be maintained to have beta-specific paths

mmv1/third_party/terraform/.teamcity/components/projects/feature_branches/FEATURE-BRANCH-ephemeral-write-only.kt
```diff
@@ -0,0 +1,82 @@
...
+
+    val trigger  = NightlyTriggerConfiguration(
+        branch = "refs/heads/$featureBranchEphemeralWriteOnly" // triggered builds must test the feature branch
+    )
+    val vcrConfig = getVcrAcceptanceTestConfig(allConfig) // Reused below for both MM testing build configs
+
+    // GA
+    val gaConfig = getGaAcceptanceTestConfig(allConfig)
+    // These are the packages that have resources that will use write-only attributes
+    var ServicesListWriteOnlyGA = getServicesList(arrayOf("compute", "secretmanager", "sql", "bigquerydatatransfer"), "GA")
```
SarahFrench, line 39: Good to see that the new project will be limited to only the necessary services! 🙌🏻 Though the Compute build is one of our longest running builds... so running those tests in the feature branch testing project as well as the nightlies could cause some issues.

That's something to keep an eye on, at least. You could even add services to this list progressively over time, instead of adding them all now, to limit impact. Or you could change the trigger time (see other comment)

mmv1/third_party/terraform/.teamcity/components/projects/feature_branches/FEATURE-BRANCH-ephemeral-write-only.kt
```diff
@@ -0,0 +1,82 @@
...
+import getServicesList
+
+const val featureBranchEphemeralWriteOnly = "FEATURE-BRANCH-ephemeral-write-only"
+const val EphemeralWriteOnlyTfCoreVersion = "1.11.0-alpha20241211"
+
+fun featureBranchEphemeralWriteOnlySubProject(allConfig: AllContextParameters): Project {
+
+    val trigger  = NightlyTriggerConfiguration(
+        branch = "refs/heads/$featureBranchEphemeralWriteOnly" // triggered builds must test the feature branch
+    )
```
SarahFrench, line 33: Please see other comment about conflicting with the nightly tests, particularly for Compute.

To mitigate this, you could update the trigger in the feature branch test project to run on a later schedule versus the nightlies. E.g. default time + 6 hours

SarahFrench (commented):
👋🏻 Hey, here's some comments but I'm happy to approve this PR after they're resolved (either via discussion only or via code changes we identify to make).

mmv1/third_party/terraform/.teamcity/components/get_services.kt
BBBmau: thought I had added a comment explaining the usage, I'll include it. 

I agree it makes sense since it's intended more for feature branches.

mmv1/third_party/terraform/.teamcity/components/projects/feature_branches/FEATURE-BRANCH-ephemeral-write-only.kt
BBBmau: There was also the idea of changing the test prefix within the write-only subproject to only run Tests with `TestAccEphemeral_` though we would miss out on making sure all other tests aren't affected.

I could ensure that all other tests pass by manually triggering them to run the suite entirely, I'm leaning towards this in order to not have to make any TC updates in the future such as including it as part of the `getServices` list.

mmv1/third_party/terraform/.teamcity/components/get_services.kt
BBBmau: the `transformation of path` is unnecessary and was removed in the latest commit. 

The `displayName` is still needed since without explcitly setting the displayName to include beta/MM we get an error due to duplicate name usage.

SarahFrench: That's fine, it was only the transform of `path` that I was interested in. I get the need to alter the names- I've suffered the pain of TeamCity's constraints around name uniqueness despite being at _different levels of nesting_ 😭

mmv1/third_party/terraform/.teamcity/components/projects/feature_branches/FEATURE-BRANCH-ephemeral-write-only.kt
SarahFrench: That sounds good to me, as long as you're aware there's a potential problem and you've got options about how to navigate it then I'm good 👍🏻

SarahFrench (approved):

BBBmau:
although this was approved, it could use a quick PR check since I introduced support for custom test prefix. 

Currently this will only run tests with the regex `TestAcc.*Ephemeral` with the default value being `TestAcc` if unset.

@SarahFrench

PR #12549, opened 2024-12-12
Add support for source_workstation in workstations_workstation resource.

esu:
This adds support for workstation PD cloning using `source_workstation`. It can only be set during workstation resource creation and cannot be updated afterwards: https://cloud.google.com/workstations/docs/reference/rest/v1beta/projects.locations.workstationClusters.workstationConfigs.workstations#Workstation.FIELDS.source_workstation

The primary use case of this field is for recovery, so it doesn't seem possible to write an automated test for it. This is because a workstation must be started manually before it can be cloned, else its PD is not created. If there is a way to write a test that can send API calls outside of Terraform, please let me know.

Fixes https://github.com/hashicorp/terraform-provider-google/issues/19730.

```release-note:enhancement
workstations: add `source_workstation` field to `google_workstations_workstation` resource
```

### DIFF
--- a/mmv1/products/workstations/Workstation.yaml
+++ b/mmv1/products/workstations/Workstation.yaml
@@ -153,3 +153,9 @@ properties:
       - 'STATE_RUNNING'
       - 'STATE_STOPPING'
       - 'STATE_STOPPED'
+  - name: 'source_workstation'
+    type: String
+    description: |
+      Full resource name of the source workstation from which the workstation's persistent
+      directories will be cloned from during creation.
+    immutable: true


ScottSuarez (changes requested):
Can you add tests for this field? Potentially an update and create test. 

You can extend existing ones where appropriate

esu:
It's difficult to write a create test for this field because it requires a manual human interaction on a dependency resource. A create test would have to:
1. Create a workstation
2. Start the workstation so that its PDs are initialized, else the API will reject the next step (and `StartWorkstation` is not possible through Terraform)
3. Create a second workstation specifying `source_workstation` to clone the first workstation's PDs

Is it possible to send GCP API calls without Terraform during a create test? Open to suggestions on how to implement this.
Lastly, if the field is immutable, does it need an update test?

esu:
I am not sure we can do that -- the initial workstation can be created using Terraform, but it's created in a `STOPPED` state in which it is not cloneable, even after the `CREATE` operation is reported as complete by the API.

In your example, creating a secondary workstation using `source_workstation` is "cloning" it, which the Workstations API will reject if the initial workstation is not manually started. Workstations are for human use and cannot be started using Terraform.

ScottSuarez (approved):

ScottSuarez:
tested locally

PR #12550, opened 2024-12-12
`EphemeralWriteOnly`: Initial support for MMv1 Generation in schema & docs

BBBmau:
Goal of this PR is to include `WriteOnly` property as part of generation to support the upcoming Ephemeral WriteOnly Attributes in Terraform 1.11

We add a new field titled `write_only` that includes `WriteOnly` in the schema of any field with `write_only: true`

Some things to know:

The generation will fail when `write_only: true` is set with any of the following:
- `sensitive` (write_only attributes are meant to replace sensitive values in the future)
- `default_from_api`/`output` set properties to be Computed + Optional which we do not support for ephemeral write-only attributes

It's also worth noting that `flatteners` are not needed as write-onlys will not be saved to state.

This PR also introduces `WriteOnlyProperties` that's needed to separate the Ephemeral Write-Only Attributes from a resource into it's own section when generation documentation (similar to how we have arguments and attributes. A new section will be introduced under the name `Ephemeral Attributes Reference`

```release-note:none

```

### DIFF
--- a/mmv1/api/resource.go
+++ b/mmv1/api/resource.go
@@ -547,10 +547,27 @@ func (r Resource) SensitiveProps() []*Type {
 	})
 }
 
+func (r Resource) WriteOnlyProps() []*Type {
+	props := r.AllNestedProperties(r.RootProperties())
+	return google.Select(props, func(p *Type) bool {
+		return p.WriteOnly
+	})
+}
+
 func (r Resource) SensitivePropsToString() string {
 	var props []string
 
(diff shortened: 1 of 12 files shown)


BBBmau:
tests are failing due to some missing upstream commits, they'll be resolved once we do a sync-main merge into the feature branch.

c2thorn (approved):
sorry did not see this until the latest notification. leaving my approval just in case it is still needed.

PR #12551, opened 2024-12-12
Promote port mapping negs to ga

nme01:
Promote the PSC Port Mapping NEG functionality from Beta to GA.

```release-note:enhancement
compute: promoted the PSC Port Mapping NEG functionality from Beta to GA
```

### DIFF
--- a/mmv1/templates/terraform/pre_delete/compute_region_network_endpoint.go.tmpl
+++ b/mmv1/templates/terraform/pre_delete/compute_region_network_endpoint.go.tmpl
@@ -27,7 +27,6 @@ if fqdnProp != "" {
   toDelete["fqdn"] = fqdnProp
 }
 
-{{- if ne $.TargetVersionName "ga" }}
 // Instance
 instanceProp, err := expandNestedComputeRegionNetworkEndpointInstance(d.Get("instance"), d, config)
 if err != nil {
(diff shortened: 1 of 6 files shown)


mmv1/templates/terraform/pre_delete/compute_region_network_endpoint.go.tmpl
```diff
@@ -27,7 +27,6 @@ if fqdnProp != "" {
   toDelete["fqdn"] = fqdnProp
 }
 
-{{- if ne $.TargetVersionName "ga" }}
```
roaks3, line 30: Nit: for this file and possibly the test file, if they no longer need beta-only behavior, you could change them to just be `.go` files. Of course, if you will just be adding more beta fields again soon, then it might be easier just to keep them this way.

mmv1/third_party/terraform/services/compute/resource_compute_region_network_endpoint_test.go.tmpl
```diff
@@ -161,7 +159,7 @@ func TestAccComputeRegionNetworkEndpoint_regionNetworkEndpointPortmapExample(t *
 
 	acctest.VcrTest(t, resource.TestCase{
 		PreCheck:                 func() { acctest.AccTestPreCheck(t) },
-		ProtoV5ProviderFactories: acctest.ProtoV5ProviderBetaFactories(t),
+		ProtoV5ProviderFactories: acctest.ProtoV5ProviderFactories(t),
```
roaks3, line 162: You'll need to remove the references to `provider = google-beta` in these configs to fix the check

roaks3 (changes requested):
Main thing is test needs to be fixed (once that is done, it would be great if you could run the 2 GA tests locally to confirm, since they aren't run in CI)

PR #12552, opened 2024-12-12
Fix autofilling values in `initialize_params.resource_policies` causing instance to recreate

karolgorc:
closes https://github.com/hashicorp/terraform-provider-google/issues/19735

This field will no longer take API values if the field is not set causing an instance recreation.

```release-note:enhancement
compute: Setting resource policies for `google_compute_instance` outside of terraform or using `google_compute_disk_resource_policy_attachment` will no longer affect the `boot_disk.initialize_params.resource_policies` field
```

### DIFF
--- a/mmv1/third_party/terraform/services/compute/resource_compute_instance.go.tmpl
+++ b/mmv1/third_party/terraform/services/compute/resource_compute_instance.go.tmpl
@@ -325,6 +325,7 @@ func ResourceComputeInstance() *schema.Resource {
 										Elem:             &schema.Schema{Type: schema.TypeString},
 										Optional:         true,
 										ForceNew:         true,
+										Computed:         true,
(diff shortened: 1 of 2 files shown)


karolgorc:
I think this will still work with `Computed: true`. All we want to achieve here is for this value not to cause diff if it's not set in the user config.

hao-nan-li:
> I think this will still work with `Computed: true`. All we want to achieve here is for this value not to cause diff if it's not set in the user config.

Yeah I agree with you, but removing `Computed` will cause breaking changes to existing users.

karolgorc:
Tested this with `Computed:true`. It does work, just needed to add a second condition to tighten this up a bit. This should no longer be a breaking change

hao-nan-li (approved):

PR #12555, opened 2024-12-12
Moverefpages

nityaravi:

```release-note:none

```


melinath (changes requested):
it looks like this does end up having merge conflicts with the other PR (because of the changes there) but hopefully they'll be easy to resolve.

nityaravi:
> what would you think about renaming to `reference/resource` and `reference/field` instead, to eliminate the duplicate `-reference`?

sure, i'll do this work in a subsequent PR.

nityaravi:
> it looks like this does end up having merge conflicts with the other PR (because of the changes there) but hopefully they'll be easy to resolve.

Done.

melinath (approved):

PR #12556, opened 2024-12-12
Fix resources with incorrect filenames

roaks3:
These were noticed while generating metadata for handwritten resources. The main purpose for correcting them is for PR notifications to be more accurate.

Of note, changing `usage_export_bucket.html.markdown` to `project_usage_export_bucket.html.markdown` will presumably break links, so it's possible we want to keep the incorrect value (see https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/usage_export_bucket).

```release-note:none

```


melinath (approved):
We should just accept the one-time pain of moving the documentation page; this also fixes the bug of the name showing up incorrectly as `google_usage_export_bucket` in the sidebar.

PR #12562, opened 2024-12-12
Add support for cross bucket replication feature

googlyrahman:
```release-note:enhancement
storagetransfer: add support for `replication_spec` fields to `google_storage_transfer_job` resource
```

### DIFF
--- a/mmv1/third_party/terraform/services/storagetransfer/resource_storage_transfer_job.go.tmpl
+++ b/mmv1/third_party/terraform/services/storagetransfer/resource_storage_transfer_job.go.tmpl
@@ -19,7 +19,7 @@ import (
 )
 
 var (
-	objectConditionsKeys = []string{
+	transferSpecObjectConditionsKeys = []string{
 		"transfer_spec.0.object_conditions.0.min_time_elapsed_since_last_modification",
 		"transfer_spec.0.object_conditions.0.max_time_elapsed_since_last_modification",
 		"transfer_spec.0.object_conditions.0.include_prefixes",
@@ -27,12 +27,26 @@ var (
 		"transfer_spec.0.object_conditions.0.last_modified_since",
 		"transfer_spec.0.object_conditions.0.last_modified_before",
 	}
+	replicationSpecObjectConditionsKeys = []string{
+		"replication_spec.0.object_conditions.0.min_time_elapsed_since_last_modification",
+		"replication_spec.0.object_conditions.0.max_time_elapsed_since_last_modification",
+		"replication_spec.0.object_conditions.0.include_prefixes",
+		"replication_spec.0.object_conditions.0.exclude_prefixes",
+		"replication_spec.0.object_conditions.0.last_modified_since",
+		"replication_spec.0.object_conditions.0.last_modified_before",
+	}
 
-	transferOptionsKeys = []string{
+	transferSpecTransferOptionsKeys = []string{
 		"transfer_spec.0.transfer_options.0.overwrite_objects_already_existing_in_sink",
 		"transfer_spec.0.transfer_options.0.delete_objects_unique_in_sink",
 		"transfer_spec.0.transfer_options.0.delete_objects_from_source_after_transfer",
 		"transfer_spec.0.transfer_options.0.overwrite_when",
+	}
+	replicationSpecTransferOptionsKeys = []string{
+		"replication_spec.0.transfer_options.0.overwrite_objects_already_existing_in_sink",
+		"replication_spec.0.transfer_options.0.delete_objects_unique_in_sink",
+		"replication_spec.0.transfer_options.0.delete_objects_from_source_after_transfer",
+		"replication_spec.0.transfer_options.0.overwrite_when",
 	}
 
 	transferSpecDataSourceKeys = []string{
@@ -47,6 +61,14 @@ var (
 		"transfer_spec.0.gcs_data_sink",
 		"transfer_spec.0.posix_data_sink",
 	}
+
+	replicationSpecDataSourceKeys = []string{
+		"replication_spec.0.gcs_data_source",
+	}
+	replicationSpecDataSinkKeys = []string{
+		"replication_spec.0.gcs_data_sink",
+	}
+
 	awsS3AuthKeys = []string{
 		"transfer_spec.0.aws_s3_data_source.0.aws_access_key",
 		"transfer_spec.0.aws_s3_data_source.0.role_arn",
@@ -98,6 +120,7 @@ func ResourceStorageTransferJob() *schema.Resource {
 				Optional:      true,
 				MaxItems:      1,
 				ConflictsWith: []string{"schedule"},
+				DiffSuppressFunc: diffSuppressEventStream,
 				Elem: &schema.Resource{
 					Schema: map[string]*schema.Schema{
 						"name": {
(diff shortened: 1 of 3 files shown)


mmv1/third_party/terraform/services/storagetransfer/resource_storage_transfer_job.go.tmpl
```diff
@@ -98,6 +120,7 @@ func ResourceStorageTransferJob() *schema.Resource {
 				Optional:      true,
 				MaxItems:      1,
 				ConflictsWith: []string{"schedule"},
+				DiffSuppressFunc: diffSuppressEventStream,
```
kautikdk, line 123: What is the need of DiffSuppressFunc?

mmv1/third_party/terraform/services/storagetransfer/resource_storage_transfer_job.go.tmpl
```diff
@@ -120,14 +143,44 @@ func ResourceStorageTransferJob() *schema.Resource {
...
+						},
+					},
+				},
+				Description: `Replication specification.`,
+			},
 			"transfer_spec": {
 				Type:     schema.TypeList,
-				Required: true,
+				Optional: true,
 				MaxItems: 1,
```
kautikdk, line 179: I think it will be good to add `ConflictsWith` and `ExactlyOneOf` conditions here as well to make it explicit for each fields.

mmv1/third_party/terraform/services/storagetransfer/resource_storage_transfer_job_test.go
```diff
@@ -78,6 +78,63 @@ func TestAccStorageTransferJob_basic(t *testing.T) {
...
+			{
+				Config: testAccStorageTransferReplicationJob_basic(envvar.GetTestProjectFromEnv(), testDataSourceBucketName, testDataSinkName, testUpdatedTransferReplicationJobDescription),
+			},
+			{
+				ResourceName:      "google_storage_transfer_job.transfer_job",
+				ImportState:       true,
+				ImportStateVerify: true,
+			},
+			{
+				Config: testAccStorageTransferReplicationJob_with_transferOptions(envvar.GetTestProjectFromEnv(), testDataSourceBucketName, testDataSinkName, testUpdatedTransferReplicationJobDescription, true, false, testOverwriteWhen[0]),
```
kautikdk, line 112: I wonder if we need separate transfer options update test step as this is common block with `transfer_specs`.

googlyrahman: I don't get you on this completely, i'll discuss this offline

mmv1/third_party/terraform/services/storagetransfer/resource_storage_transfer_job.go.tmpl
googlyrahman: Acknowledge

mmv1/third_party/terraform/services/storagetransfer/resource_storage_transfer_job.go.tmpl
googlyrahman: Good question! I'll give some context:

When using cross bucket replication feature, i.e specifying `replication_spec`, the backend would setup `event_stream` on it's own, and it can't be configurable via user, Hence showing `computed` behvior, However we can't mark it `computed` as the field is configurable while using `transfer_spec`.

So first time when you do `terraform apply`,  the `event_stream` would be stored in file, and in next `terraform plan` it would show a diff `SOMETHING -> ""`

One alternative to this is to not store `eventStream` at all in `Read` while using `replication`.

kautikdk: From the context, Can we use Optional+Computed and ConflictsWith(event_stream and replication_spec)?

googlyrahman: Keeping it `Optional+Computed` would show `event_stream (known after apply)` despite not specifying `event_stream` while working with `transfer_spec`.

But i think, it should be fine - I've updated code to reflect the same!

hao-nan-li:
re-roll as I'll be OOO

mmv1/third_party/terraform/services/storagetransfer/resource_storage_transfer_job.go.tmpl
kautikdk: Also please consider a case when user removes `event_stream` from the config expect it to be removed from the resource. If the attribute is computed then terraform won't consider this change as necessary and won't show any diff. If that is expected then this solution should work otherwise we need to verify if empty list can be used as not-present indicator or not.

googlyrahman: Based on the offline conversation, we're moving forward with `diffSupressFunc` as written earlier.

googlyrahman:
Hey @NickElliot, there's one test failing - doesn't look related to my change - can you please take a look, and can you prioritize the review of this PR?

kautikdk (approved):
LGTM!

mmv1/third_party/terraform/website/docs/r/storage_transfer_job.html.markdown
```diff
@@ -116,7 +116,10 @@ The following arguments are supported:
 
 * `description` - (Required) Unique description to identify the Transfer Job.
 
-* `transfer_spec` - (Required) Transfer specification. Structure [documented below](#nested_transfer_spec).
+* `transfer_spec` - (Optional) Transfer specification. Structure [documented below](#nested_transfer_spec). Only one of `transfer_spec`, or `replication_spec` can be specified.
```
NickElliot, line 119:
```suggestion
* `transfer_spec` - (Optional) Transfer specification. Structure [documented below](#nested_transfer_spec). One of `transfer_spec`, or `replication_spec` must be specified.
```

mmv1/third_party/terraform/website/docs/r/storage_transfer_job.html.markdown
```diff
@@ -116,7 +116,10 @@ The following arguments are supported:
...
+
+* `replication_spec` - (Optional) Replication specification. Structure [documented below](#nested_replication_spec). User should not configure
+`transfer_spec`, `schedule`, `event_stream` with this argument.
```
NickElliot, line 122:
```suggestion
* `replication_spec` - (Optional) Replication specification. Structure [documented below](#nested_replication_spec). User should not configure
`transfer_spec`, `schedule`, `event_stream` with this argument. One of `transfer_spec`, or `replication_spec` must be specified.
```

NickElliot (changes requested):
a couple of needed documentation updates -- will need a bit longer to review the code due to needing to catch up to speed on existing PR history

mmv1/third_party/terraform/website/docs/r/storage_transfer_job.html.markdown
googlyrahman: Done

mmv1/third_party/terraform/website/docs/r/storage_transfer_job.html.markdown
googlyrahman: Done

googlyrahman:
> a couple of needed documentation updates -- will need a bit longer to review the code due to needing to catch up to speed on existing PR history

Would appreciate if you can prioritize it. It's been in review since 2 weeks, I understand that this was assigned to someone else earlier due to which there's delay.

googlyrahman:
Any update on this?

NickElliot (approved):
LGTM! sorry for delayed review with holidays

PR #12567, opened 2024-12-13
Fix diff suppress function ignoring additive_vpc_scope_dns_domain in Autopilot clusters

Michcioperz:
self-appointed successor to #11744

GKE Autopilot allows setting additive VPC scope domain on creation only. However, current implementation of the diff suppress function for Autopilot clusters removes any values in dns_config of Autopilot clusters, all the time.

This change permits setting additive VPC scope domain.

Remaining action items:
- [x] validate that allowing `dns_config.#` doesn't cause issues (it seems necessary for `dns_config.0.additive_vpc_scope_dns_domain` to exist
- [ ] revise test coverage to look for other gaps
- [x] verify that the change in #11744 to ForceNew when domain changes is exactly the behavior we want and copy it over
- [x] write the release note below

```release-note:bug
container: fixed `additive_vpc_scope_dns_domain` being ignored in Autopilot cluster definition
```

### DIFF
--- a/mmv1/third_party/terraform/services/container/resource_container_cluster.go.tmpl
+++ b/mmv1/third_party/terraform/services/container/resource_container_cluster.go.tmpl
@@ -113,6 +113,14 @@ var (
 
 	suppressDiffForAutopilot = schema.SchemaDiffSuppressFunc(func(k, oldValue, newValue string, d *schema.ResourceData) bool {
 		if v, _ := d.Get("enable_autopilot").(bool); v {
+			if k == "dns_config.0.additive_vpc_scope_dns_domain" {
+				return false
+			}
+			if k == "dns_config.#" {
+				if avpcDomain, _ := d.Get("dns_config.0.additive_vpc_scope_dns_domain").(string); avpcDomain != "" || d.HasChange("dns_config.0.additive_vpc_scope_dns_domain") {
+					return false
+				}
+			}
 			return true
 		}
 		return false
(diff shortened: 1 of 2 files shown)


Michcioperz:
@slevenick I think it's looking better now, PTAL.

I know I put an action item to look for more test gaps in the PR description, but I'll file a followup bug internally for this. I think the dns config mutation test I added should be useful enough for now.

slevenick:
Rerolling reviewer as I'm out sick

Michcioperz:
Thanks, hope you get well soon!

ScottSuarez (approved):

mmv1/third_party/terraform/services/container/resource_container_cluster.go.tmpl
```diff
@@ -7074,6 +7082,9 @@ func containerClusterAutopilotCustomizeDiff(_ context.Context, d *schema.Resourc
 			return err
 		}
 	}
+	if d.Get("enable_autopilot").(bool) && d.HasChange("dns_config.0.additive_vpc_scope_dns_domain") {
+		return d.ForceNew("dns_config.0.additive_vpc_scope_dns_domain")
+	}
```
ScottSuarez, line 7087: Can I ask why this check is needed? If I'm reading this right shouldn't the other conditions in the `suppressDiffForAutopilot` sufficiently manage this scenario? I could be wrong or missing something based on your testing.

Michcioperz: Sorry if I overcomplicated it, it was my first Terraform code and I'm not on friendly terms with the documentation. To be clear what I'm trying to accomplish between the diff suppress and here is that 1. the only thing you are allowed to customize in Autopilot clusters is the additive VPC domain 2. after creation of an Autopilot cluster no modifications to DNS config are permitted by container API, any update involving these will fail.

So with that in mind, is what you're saying that I could just use ForceNew in the suppress diff function?

ScottSuarez: Ah! Your context makes sense. Thanks for that. 

I believe you designed it correctly then. I don't believe it's possible to forceNew from a suppress diff function.

ScottSuarez (approved):

PR #12568, opened 2024-12-13
Revert "Etag support for ACM service perimeters"

coder-221:
Reverts GoogleCloudPlatform/magic-modules#12363

Unfortunately, we'd like to revert this change and then try again probably in early January with a new approach for perimeter etags. We have ideas on this but for now we think reverting this change will ensure customers upgrading to the next released version aren't broken.

The broken scenario is when there are perimeters from the same policy in separate google_access_context_manager_service_perimeter resources. Customers with multiple perimeters in a single google_access_context_manager_service_perimeters resource would be fine.

Currently, perimeter etags are generated at the policy level (the parent entity). So, when the first PATCH request is sent for updating perimeter 1, then the etag saved in the Terraform plan for the other perimeters is no longer valid. Those updates will fail.

```release-note:enhancement
accesscontextmanager: added `etag` to `google_access_context_manager_service_perimeter` and `google_access_context_manager_service_perimeters` (reverts)
```

### DIFF
--- a/mmv1/products/accesscontextmanager/ServicePerimeter.yaml
+++ b/mmv1/products/accesscontextmanager/ServicePerimeter.yaml
@@ -65,7 +65,6 @@ async:
 custom_code:
   constants: 'templates/terraform/constants/access_context_manager.go.tmpl'
   encoder: 'templates/terraform/encoders/access_level_never_send_parent.go.tmpl'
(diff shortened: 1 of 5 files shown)


c2thorn (approved):
got approval for the breaking change offline and confirmed this is a standard revert of https://github.com/GoogleCloudPlatform/magic-modules/pull/12363

PR #12569, opened 2024-12-13
Apply maintenance policy updates after upgrades so validation on maintenance policy uses the new versions.

cslink:
Switch the ordering of updates so maintenance policy updates get applied after upgrades. This means that validation on maintenance exclusions will use the new version as opposed to the old version which means you can upgrade to a new version and set an exclusion until end of support for the new version in one `terraform apply` command.

Fixes https://github.com/hashicorp/terraform-provider-google/issues/20556

Release note:

```release-note:enhancement
container: changed `google_container_cluster` to apply maintenance policy updates after upgrades during cluster update
```


c2thorn:
@cslink I'd recommend following the initial steps here to get automatic build runs for any future pull requests: go/terraform-contribution-guide#before-you-begin

c2thorn:
retrying the test

c2thorn (approved):
slightly tweaked the release notes, but PR LGTM. Thanks @cslink

PR #12572, opened 2024-12-14
Fix permadiff when Access Context Manager returns a different order for ingress / egress rule identities

Charlesleonius:
Adds a diff suppress func to ignore ordering for identities in ingress / egress rule resources. The order does not matter and sometimes the API changes it.

Addresses [#19203](https://github.com/hashicorp/terraform-provider-google/issues/19203) and [#20519](https://github.com/hashicorp/terraform-provider-google/issues/20519)

```release-note:bug
accesscontextmanager: fixed permadiff due to reordering on `google_access_context_manager_service_perimeter_dry_run_egress_policy` `egress_from.identities`
```

```release-note:bug
accesscontextmanager: fixed permadiff due to reordering on `google_access_context_manager_service_perimeter_dry_run_ingress_policy` `ingress_from.identities`
```

```release-note:bug
accesscontextmanager: fixed permadiff due to reordering on `google_access_context_manager_service_perimeter_egress_policy` `egress_from.identities`
```
```release-note:bug
accesscontextmanager: fixed permadiff due to reordering on `google_access_context_manager_service_perimeter_ingress_policy` `ingress_from.identities`
```

### DIFF
--- a/mmv1/products/accesscontextmanager/ServicePerimeterDryRunEgressPolicy.yaml
+++ b/mmv1/products/accesscontextmanager/ServicePerimeterDryRunEgressPolicy.yaml
@@ -116,6 +116,8 @@ properties:
           https://cloud.google.com/iam/docs/principal-identifiers#v1 are supported.
         item_type:
           type: String
+        diff_suppress_func: AccessContextManagerServicePerimeterDryRunEgressPolicyEgressFromIdentitiesDiffSuppressFunc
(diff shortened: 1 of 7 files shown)


slevenick (commented):
Is there a way to trigger this permadiff by creating a resource with the rules in a certain order? I'd like to be able to have a test so we can track this going forward

Charlesleonius:
> #20519

Testing the permadiff from the actual API would be difficult. There is way to at least test the comparison logic though so I will do that.

slevenick (approved):

PR #12576, opened 2024-12-16
add ip_filter field to storage_bucket

translucens:

```release-note:enhancement
storage: added `ip_filter` field to `google_compute_foo` resource
```

### DIFF
--- a/mmv1/third_party/terraform/services/storage/resource_storage_bucket.go.tmpl
+++ b/mmv1/third_party/terraform/services/storage/resource_storage_bucket.go.tmpl
@@ -567,6 +567,67 @@ func ResourceStorageBucket() *schema.Resource {
 					},
 				},
 			},
+{{- if ne $.TargetVersionName "ga" }}
+			"ip_filter": {
+				Type:             schema.TypeList,
+				MaxItems:         1,
+				Optional:         true,
+				Description:      `The bucket IP filtering configuration.`,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"mode": {
+							Type: schema.TypeString,
+							Required: true,
+							Description: `The mode of the IP filter. Valid values are 'Enabled' and 'Disabled'.`,
+							ValidateFunc: validation.StringInSlice([]string{"Enabled", "Disabled"}, false),
+						},
+						"public_network_source": {
+							Type:        schema.TypeList,
+							MaxItems:    1,
+							Optional:    true,
+							Description: `The bucket IP filtering configuration.`,
+							Elem:        &schema.Resource{
+								Schema: map[string]*schema.Schema{
+									"allowed_ip_cidr_ranges": {
+										Type:        schema.TypeList,
+										Required:    true,
+										Description: "The list of public IPv4, IPv6 cidr ranges that are allowed to access the bucket.",
+										Elem:        &schema.Schema{
+											Type: schema.TypeString,
+											ValidateFunc: validation.IsCIDR,
+										},
+									},
+								},
+							},
+						},
+						"vpc_network_sources": {
+							Type:        schema.TypeList,
+							Optional:    true,
+							Description: `The bucket IP filtering configuration.`,
+							Elem:        &schema.Resource{
+								Schema: map[string]*schema.Schema{
+									"allowed_ip_cidr_ranges": {
+										Type:     schema.TypeList,
+										Required: true,
+										Description: "The list of public IPv4, IPv6 cidr ranges that are allowed to access the bucket.",
+										Elem:     &schema.Schema{
+											Type: schema.TypeString,
+											ValidateFunc: validation.IsCIDR,
+										},
+									},
+									"network": {
+										Type:        schema.TypeString,
+										Required:    true,
+										Description: "Name of the network. Format: projects/{PROJECT_ID}/global/networks/{NETWORK_NAME}",
+										ValidateFunc: validation.StringMatch(regexp.MustCompile("projects/.+/global/networks/.+"), ""),
+									},
+								},
+							},
+						},
+					},
+				},
+			},
+{{- end }}
 		},
 		UseJSONNumber: true,
 	}
@@ -718,6 +779,12 @@ func resourceStorageBucketCreate(d *schema.ResourceData, meta interface{}) error
 		sb.HierarchicalNamespace = expandBucketHierachicalNamespace(v.([]interface{}))
 	}
 
+{{- if ne $.TargetVersionName "ga" }}
+	if v, ok := d.GetOk("ip_filter"); ok {
+		sb.IpFilter = expandIpFilter(v.([]interface{}))
+	}
+{{- end }}
+
 	var res *storage.Bucket
 
 	err = transport_tpg.Retry(transport_tpg.RetryOptions{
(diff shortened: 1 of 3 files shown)


mmv1/third_party/terraform/services/storage/resource_storage_bucket.go.tmpl
```diff
@@ -745,7 +812,7 @@ func resourceStorageBucketCreate(d *schema.ResourceData, meta interface{}) error
 	// to make sure it exists before moving on
 	err = transport_tpg.Retry(transport_tpg.RetryOptions{
 		RetryFunc: func() (operr error) {
-			_, retryErr := config.NewStorageClient(userAgent).Buckets.Get(res.Name).Do()
+			_, retryErr := config.NewStorageClient(userAgent).Buckets.Get(res.Name).Projection("full").Do()
```
roaks3, line 815: What's the purpose of this change? (Note it would be the only change to the GA provider)

translucens: Default projection does not contains ipFilter and full projection is required only for ipFilter so I added beta version guard.

kautikdk (commented):
We might need `DiffSuppressFunc` function as IpFilter is permanent field and can not be removed once it is configured. Similar to `autoclass` diffsuppressfunc func: https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/third_party/terraform/services/storage/resource_storage_bucket.go.tmpl#L342

Also one suggestion, we can avoid short named variables in expanders and flatteners for better code readability.

mmv1/third_party/terraform/services/storage/resource_storage_bucket.go.tmpl
```diff
@@ -916,7 +996,11 @@ func resourceStorageBucketUpdate(d *schema.ResourceData, meta interface{}) error
 	// to make sure it exists before moving on
 	err = transport_tpg.Retry(transport_tpg.RetryOptions{
 		RetryFunc: func() (operr error) {
+{{- if ne $.TargetVersionName "ga" -}}
+			_, retryErr := config.NewStorageClient(userAgent).Buckets.Get(res.Name).Projection("full").Do()
```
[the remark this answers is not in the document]
translucens: Thank you for your review. Applied your comments.

kautikdk:
@roaks3 It looks like one of the checks is stuck and not moving forward from past 20 hours because of missing approval on cloud build. Can you please re-trigger it or give approval?

mmv1/third_party/terraform/services/storage/resource_storage_bucket.go.tmpl
```diff
@@ -567,6 +567,86 @@ func ResourceStorageBucket() *schema.Resource {
...
+										Elem:     &schema.Schema{
+											Type: schema.TypeString,
+											ValidateFunc: validation.IsCIDR,
+										},
+									},
+									"network": {
+										Type:        schema.TypeString,
+										Required:    true,
+										Description: "Name of the network. Format: projects/{PROJECT_ID}/global/networks/{NETWORK_NAME}",
+										ValidateFunc: validation.StringMatch(regexp.MustCompile("projects/.+/global/networks/.+"), ""),
```
roaks3, line 622: FWIW, networks are passed into a variety of resources across the provider and I don't see them being validated with a regex like this. I'm not sure of the reason, but it could be to support other formats or future changes (like `https://www.googleapis.com/compute/v1/` prefix or similar). If you're sure you want this, I think it's ok to keep, but I would lean toward removing it.

mmv1/third_party/terraform/services/storage/resource_storage_bucket_test.go.tmpl
```diff
@@ -2762,3 +2762,139 @@ resource "google_storage_bucket" "bucket" {
...
+  location = "us-central1"
+	uniform_bucket_level_access = true
+  force_destroy = true
+  ip_filter  {
+    mode = "Enabled"
+		public_network_source {
+			allowed_ip_cidr_ranges = ["0.0.0.0/0"]
+		}
+		vpc_network_sources {
+			network = "projects/%s/global/networks/%s"
```
roaks3, line 2838: Could we avoid testing with the default network? Also, it would be good to pass this as a terraform output like a user would more normally do.

mmv1/third_party/terraform/services/storage/resource_storage_bucket_test.go.tmpl
```diff
@@ -2762,3 +2762,139 @@ resource "google_storage_bucket" "bucket" {
...
+			allowed_ip_cidr_ranges = ["0.0.0.0/0", "::/0"]
+		}
+		vpc_network_sources {
+			network = "projects/%s/global/networks/%s"
+			allowed_ip_cidr_ranges = ["10.201.0.0/16", "10.202.0.0/16"]
+		}
+	}
+}
+`, bucketName, envvar.GetTestProjectFromEnv(), "default", envvar.GetTestProjectFromEnv(), "storage-source-2")
```
roaks3, line 2887: Similar here, with 2 additional points:
- the network name should be passed in to both the definition and the reference
- the network name needs to begin with `tf-test` in order to be cleaned up by our sweepers

roaks3 (commented):
Mostly LGTM, just a few minor comments

mmv1/third_party/terraform/services/storage/resource_storage_bucket_test.go.tmpl
```diff
@@ -2762,3 +2762,156 @@ resource "google_storage_bucket" "bucket" {
...
+				ImportStateVerifyIgnore: []string{"force_destroy"},
+			},
+		},
+	})
+}
+
+func testAccStorageBucket_IPFilter(bucketName string) string {
+	return fmt.Sprintf(`
+resource "google_compute_network" "vpc_gcs_ipfilter1" {
+  name = "tf-test-storage-ipfilter1"
```
roaks3, line 2828: For both tests, you'll also need a random suffix (see other tests for examples), to ensure subsequent/concurrent runs to not affect each other.

roaks3 (approved):
LGTM, we just need the random suffix for test resources, then we can confirm tests pass.

roaks3 (changes requested):
Test failures will need to be resolved

translucens:
It seems the test runner prevented by IP filtering condition.
Could we add a custom role which has `storage.buckets.exemptFromIpFilter` permission and attach the role to test runner's service account?

kautikdk:
I am not very sure about the test error but it might be possible that the machine used to run VCR tests is part of a VPC network and generating network request from that only so public network ip ranges in your test is not taking effect here. I don't know about the internal VPC that is configured or not for running tests. 
I think we can configure IP Filter in disable mode only for acceptance tests. Other fields can be tested in disable mode as well so that should not be an issue.
Regarding the suggestion you provided, we will be basically bypassing IP Filtering through that method so instead we can just configure VPC in tests or can add tests in disable mode only.

roaks3 (commented):
Seems reasonable, clearing review

kautikdk:
Created [this](https://github.com/GoogleCloudPlatform/magic-modules/pull/12945) Draft PR from your PR to check if the solution is working or not. It is working and similar tests can be added in this PR as well. 
As there is no progress in this PR for a while, we will wait for couple more days before closing this one and going ahead with my PR. Let us know if you are still working on this PR.
Also I have changed code to add feature in GA provider and I think we can add this support in GA provider as feature is already in public preview and there is no SLO limitations from server side.
cc: @roaks3

translucens:
I'm not working with for this PR now, so we can continue on #12945.

kautikdk:
Thanks for letting us know.
I will add appropriate credit section in my PR to acknowledge your efforts.

roaks3:
Closing in favor of https://github.com/GoogleCloudPlatform/magic-modules/pull/12945

PR #12580, opened 2024-12-16
DNM: `google_project` plugin framework data source

NickElliot:
Generates normally, can be used with the `data "google_fwprovider_project`"` name

```release-note:none

```

### DIFF
--- a/mmv1/third_party/terraform/fwprovider/framework_provider.go.tmpl
+++ b/mmv1/third_party/terraform/fwprovider/framework_provider.go.tmpl
@@ -295,6 +295,7 @@ func (p *FrameworkProvider) DataSources(_ context.Context) []func() datasource.D
         resourcemanager.NewGoogleClientConfigDataSource,
         resourcemanager.NewGoogleClientOpenIDUserinfoDataSource,
         {{- if ne $.TargetVersionName "ga" }}
+        resourcemanager.NewGoogleProjectFWDataSource,
         firebase.NewGoogleFirebaseAndroidAppConfigDataSource,
         firebase.NewGoogleFirebaseAppleAppConfigDataSource,
         firebase.NewGoogleFirebaseWebAppConfigDataSource,
--- a/mmv1/third_party/terraform/services/resourcemanager/fw_data_source_google_project.go
+++ b/mmv1/third_party/terraform/services/resourcemanager/fw_data_source_google_project.go
@@ -1,0 +1,158 @@
+package resourcemanager
+
+import (
+	"context"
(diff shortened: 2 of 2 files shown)


mmv1/third_party/terraform/services/resourcemanager/fw_data_source_google_project.go
```diff
@@ -0,0 +1,158 @@
...
+
+	"github.com/hashicorp/terraform-provider-google/google/fwmodels"
+	"github.com/hashicorp/terraform-provider-google/google/fwtransport"
+	transport_tpg "github.com/hashicorp/terraform-provider-google/google/transport"
+	"google.golang.org/api/cloudresourcemanager/v1"
+)
+
+// Ensure the implementation satisfies the expected interfaces.
+var (
+	_ datasource.DataSource              = &GoogleProjectFWDataSource{}
```
c2thorn, line 21: Missed this on review of the previous framework datasources - is this actually necessary? My initial thought would be this actually doesn't do anything, and to satisfy interfaces we need to actually name a variable... Have you seen this pattern before?

Very quickly removing this from `data_source_provider_config_plugin_framework.go` and running `make build` doesn't throw a build error but I wonder if it would fail in runtime or something

mmv1/third_party/terraform/fwprovider/framework_provider.go.tmpl
```diff
@@ -295,6 +295,7 @@ func (p *FrameworkProvider) DataSources(_ context.Context) []func() datasource.D
         resourcemanager.NewGoogleClientConfigDataSource,
         resourcemanager.NewGoogleClientOpenIDUserinfoDataSource,
         {{- if ne $.TargetVersionName "ga" }}
+        resourcemanager.NewGoogleProjectFWDataSource,
```
c2thorn, line 298: do we want to make this datasource test-only for now? like in [`a8c3757` (#11599)](https://github.com/GoogleCloudPlatform/magic-modules/pull/11599/commits/a8c375712b0404a2fd409943d82c92cd0aabbe14#diff-7c30a2afcc18eb3bf49388e586c1bcb0f84fcb5768c5a9366d4d34f0dd9dd548R443)

mmv1/third_party/terraform/services/resourcemanager/fw_data_source_google_project.go
NickElliot: I had copied a lot of the base file from the Firebase data sources -- this could've been before the refactor if that removed it?

PR #12590, opened 2024-12-17
fix: (storage) fix failing testcase

gurusai-voleti:
create and update the bucket in same region
Fixes: https://github.com/hashicorp/terraform-provider-google/issues/20577

```release-note:bug
storage: fix a failing testcase, create and update the bucket in same region
```

### DIFF
--- a/mmv1/third_party/terraform/services/storage/resource_storage_bucket_test.go.tmpl
+++ b/mmv1/third_party/terraform/services/storage/resource_storage_bucket_test.go.tmpl
@@ -805,7 +805,7 @@ func TestAccStorageBucket_update(t *testing.T) {
 		CheckDestroy:             testAccStorageBucketDestroyProducer(t),
 		Steps: []resource.TestStep{
 			{
-				Config: testAccStorageBucket_basic(bucketName),
+				Config: testAccStorageBucket(bucketName),
 				Check: resource.ComposeTestCheckFunc(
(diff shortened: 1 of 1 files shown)


mmv1/third_party/terraform/services/storage/resource_storage_bucket_test.go.tmpl
```diff
@@ -824,7 +824,7 @@ func TestAccStorageBucket_update(t *testing.T) {
 				Check: resource.ComposeTestCheckFunc(
 					testAccCheckStorageBucketExists(
 						t, "google_storage_bucket.bucket", bucketName, &recreated),
-					testAccCheckStorageBucketWasRecreated(&recreated, &bucket),
+					testAccCheckStorageBucketWasUpdated(&recreated, &bucket),
```
kautikdk, line 827: We already have a lost of bucket update tests. Even the next test step will check that only so what is the need of this step? 
If we don't want to test ForceNew behavior then we can completely remove this step otherwise if we somehow need to check recreation of bucket, we can add step with different bucket name and that will force recreation of bucket. I am not sure if that will help here reducing flaky test failures.

gurusai-voleti: removed the first step

trodge (approved):

PR #12586, opened 2024-12-16
Add edgenetwork InterconnectAttachment resource.

f1urps:
Create InterconnectAttachment resource in product edgenetwork.

Fixes https://github.com/hashicorp/terraform-provider-google/issues/15665

Tests are disabled for this resource for the same reason as other resources in `edgenetwork`. See: https://github.com/GoogleCloudPlatform/magic-modules/pull/8905

**Release Note Template for Downstream PRs (will be copied)**

```release-note:new-resource
`google_edgenetwork_interconnect_attachment`
```

PR #8905, opened 2023-09-09
Add Network and Subnet resources for edgenetwork

f1urps:
This PR adds two new resources, `Network` and `Subnet`, under a new product `edgenetwork`.

Fixes this issue: https://github.com/hashicorp/terraform-provider-google/issues/15664

**Release Note Template for Downstream PRs (will be copied)**

```release-note:new-resource
edgenetwork : `google_edgenetwork_network`
```

```release-note:new-resource
edgenetwork : `google_edgenetwork_subnet`
```

PR #12594, opened 2024-12-17
Add edgenetwork Router resource.

f1urps:
Create Router resource in product edgenetwork.

Fixes https://github.com/hashicorp/terraform-provider-google/issues/15666

Tests are disabled for this resource for the same reason as other resources in `edgenetwork`. See: https://github.com/GoogleCloudPlatform/magic-modules/pull/8905

This PR depends on another in-review PR: https://github.com/GoogleCloudPlatform/magic-modules/pull/12586

```release-note:new-resource
`google_edgenetwork_router`
```

### DIFF
--- a/mmv1/products/edgenetwork/InterconnectAttachment.yaml
+++ b/mmv1/products/edgenetwork/InterconnectAttachment.yaml
@@ -54,17 +54,6 @@ examples:
     test_env_vars:
       location: 'REGION'
       zone: 'ZONE'
-
-    Settings
-
-Owner avatar
-magic-modules
-Public
-forked from GoogleCloudPlatform/magic-modules
-
-f1urps/magic-modules
-t
-Add file
     exclude_test: true
 parameters:
   - name: 'location'


mmv1/products/edgenetwork/Router.yaml
```diff
@@ -0,0 +1,227 @@
...
+    type: String
+    description: |
+      The canonical name of this resource, with format
+      `projects/{{project}}/locations/{{location}}/zones/{{zone}}/routers/{{router_id}}`
+    output: true
+  - name: 'labels'
+    type: KeyValueLabels
+    description: |
+      Labels associated with this resource.
+    required: false
```
roaks3, line 103: Nit: `required: false` is the default and can be removed in numerous places here

mmv1/templates/terraform/examples/edgenetwork_router_minimal.tf.tmpl
```diff
@@ -0,0 +1,23 @@
+
+resource "google_edgenetwork_router" "{{$.PrimaryResourceId}}" {
+  router_id   = "{{index $.Vars "edgenetwork_router_id"}}"
+  location    = "{{index $.TestEnvVars "location"}}"
+  zone        = "{{index $.TestEnvVars "zone"}}"
+  description = "Example router."
```
roaks3, line 6: The minimal example should not include any optional fields (in part to verify that they are in fact optional).

mmv1/templates/terraform/examples/edgenetwork_router_with_bgp.tf.tmpl
```diff
@@ -0,0 +1,82 @@
+
+resource "google_edgenetwork_router" "{{$.PrimaryResourceId}}" {
```
roaks3, line 2: It looks like the Router resource is not marked as immutable, and a number of resources are therefore presumed to be mutable in-place. If that's the case, this resource will need an [update test](https://googlecloudplatform.github.io/magic-modules/test/test/#add-an-update-test) to verify that those fields can be updated.

roaks3 (commented):
Mainly needs more testing (unless the resource is immutable), but otherwise LGTM

mmv1/templates/terraform/examples/edgenetwork_router_with_bgp.tf.tmpl
f1urps: Can we make an exception and skip this requirement? Due to the dependence on physical hardware, the edgenetwork API can only be tested manually by the internal dev team anyway, so I don't think those sorts of unit tests are useful in this case.

mmv1/templates/terraform/examples/edgenetwork_router_minimal.tf.tmpl
f1urps: Fixed

mmv1/products/edgenetwork/Router.yaml
f1urps: Fixed

mmv1/templates/terraform/examples/edgenetwork_router_with_bgp.tf.tmpl
roaks3: Would it be possible to write the test but have it be skipped, so that your team can run it as a one off prior to submission?

I understand the constraints will be difficult to work around, but conversely, this resource implementation could be broken and we would have no way of knowing until customers use it. If you have some other way to validate the correct mutability properties for each of these fields perhaps we could rely on that, but otherwise I think we want to at least see output from a passing test before submitting.

mmv1/products/edgenetwork/InterconnectAttachment.yaml
```diff
@@ -0,0 +1,134 @@
...
+    type: String
+    description: |
+      The canonical name of this resource, with format
+      `projects/{{project}}/locations/{{location}}/zones/{{zone}}/interconnectAttachments/{{interconnect_attachment_id}}`
+    output: true
+  - name: 'labels'
+    type: KeyValueLabels
+    description: |
+      Labels associated with this resource.
+    required: false
```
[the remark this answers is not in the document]
f1urps: Fixed

mmv1/templates/terraform/examples/edgenetwork_router_with_bgp.tf.tmpl
f1urps: I am unable to write such a test now because I've left the edgenetwork team since opening this PR.

I've marked the Router resource 'immutable' for now. Once this code is merged, others can improve it later if needed.

roaks3 (commented):
@f1urps understood, but we would still need to see working test output (for the other examples), before we can submit. Is there a person who this is passing to? Otherwise I would expect this will become stale unfortunately. Leaving a comment on b/298278262 as well.

PR #12600, opened 2024-12-17
Secure Web Proxy deleteSWGAutoGenRouter wait the operation to finish

Samir-Cit:
Hello folks.
This PR will fix https://github.com/hashicorp/terraform-provider-google/issues/18140 by adding the operation wait when deleting the auto gen routers of the Gateway (SWP)

```release-note:bug
networkservices: fix `google_network_services_gateway` resource so that it correctly waits for the router to be deleted on `terraform destroy`
```


mmv1/templates/terraform/constants/network_services_gateway.go.tmpl
```diff
@@ -122,5 +131,9 @@ func deleteSWGAutoGenRouter(d *schema.ResourceData, config *transport_tpg.Config
 		return err
 	}
 
+	err = tpgcompute.ComputeOperationWaitTime(
```
roaks3, line 134: Should this be `NetworkServicesOperationWaitTime`? Why are we trying to pull in the compute function?

mmv1/templates/terraform/constants/network_services_gateway.go.tmpl
```diff
@@ -122,5 +131,9 @@ func deleteSWGAutoGenRouter(d *schema.ResourceData, config *transport_tpg.Config
...
+		config, res, billingProject, "Deleting autogen router", userAgent,
+		d.Timeout(schema.TimeoutCreate))
```
roaks3, line 136: Nit: `TimeoutDelete` might make more sense?

roaks3 (changes requested):
The concept makes sense, but a few question regarding implementation

mmv1/templates/terraform/constants/network_services_gateway.go.tmpl
Samir-Cit: It's using `ComputeOperationWaitTime` because the `Router` resource is under `Compute`:

```
routerPath := fmt.Sprintf("{{"{{"}}ComputeBasePath{{"}}"}}projects/{{"{{"}}project{{"}}"}}/regions/{{"{{"}}location{{"}}"}}/routers/%s", routerId)
```

mmv1/templates/terraform/constants/network_services_gateway.go.tmpl
Samir-Cit: Yes... I'll change!
Thanks.

mmv1/templates/terraform/constants/network_services_gateway.go.tmpl
roaks3: Ah, missed that, thanks!

mmv1/templates/terraform/constants/network_services_gateway.go.tmpl
```diff
@@ -10,6 +10,15 @@
 	See the License for the specific language governing permissions and
 	limitations under the License.
 */ -}}
+
+import (
```
roaks3, line 14: Do you need to conditionally use different packages here? For example, project just uses the ga wait function: https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/third_party/terraform/services/resourcemanager/resource_google_project.go#L18

Samir-Cit: If I use only the GA, when running on Beta it will show this error:
```
cannot use config (variable of type *"github.com/hashicorp/terraform-provider-google-beta/google-beta/transport".Config) as *"github.com/hashicorp/terraform-provider-google/google/transport".Config value in argument to tpgcompute.ComputeOperationWaitTime
```

Samir-Cit:
Hello @roaks3 
Did you had a chance to look again this PR?

mmv1/templates/terraform/constants/network_services_gateway.go.tmpl
roaks3: Oh I see, the config is the same one used for the network services calls, which is beta in the beta provider

roaks3 (approved):
Just waiting for tests to pass

PR #12601, opened 2024-12-18
CodeRepositoryIndex cascade deletion

JumiDeluxe:

```release-note: enhancement
gemini: added `force_destroy` field to resource `google_code_repository_index`, enabling deletion of the resource even when it has dependent RepositoryGroups
```

### DIFF
--- a/mmv1/products/gemini/CodeRepositoryIndex.yaml
+++ b/mmv1/products/gemini/CodeRepositoryIndex.yaml
@@ -51,6 +51,8 @@ async:
   result:
     resource_inside_response: true
   include_project: false
+custom_code:
+  pre_delete: templates/terraform/pre_delete/code_repository_index_force_delete.go.tmpl
 error_retry_predicates:
   - 'transport_tpg.IsCodeRepositoryIndexUnreadyError'
   - 'transport_tpg.IsRepositoryGroupQueueError'
--- a/mmv1/products/gemini/CodeRepositoryIndex.yaml
+++ b/mmv1/products/gemini/CodeRepositoryIndex.yaml
@@ -59,7 +59,8 @@ error_retry_predicates:
 virtual_fields:
   - name: 'force_destroy'
     description:
-      If set to true, any RepositoryGroups for this CodeRepositoryIndex will also be deleted.
+      If set to true, will allow deletion of the CodeRepositoryIndex even if there are existing
+      RepositoryGroups for the resource. These RepositoryGroups will also be deleted.
     type: Boolean
     default_value: false
 parameters:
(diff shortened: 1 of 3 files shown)


frsv (approved):

NickElliot (commented):
Hi, code LGTM but could you clarify what the "cascade deletion" is doing here so I could make sure the release note wording is what we want to convey for Terraform users? thanks!

JumiDeluxe:
"Cascade deletion" means that we automatically remove children resources (Repository Groups) upon CRIs deletion.

NickElliot (changes requested):
Got it, could you add a test case for this functionality?

JumiDeluxe:
> Got it, could you add a test case for this functionality?

Added to CRI tests, the new test tries to remove CRI without removing RG directly in Terraform.
All children resources should be deleted, so we're expecting a new plan.

NickElliot (commented):
thanks for the test! I wanted to make a suggestion though to fly by:

due to that test working the way it did, I realized we could make users make accidental changes to their states when currently they have a protective guard. Would it be preferable to add a virtual field "`force_delete`" that serves as a flag for sending the `obj["force"] = true` body?

If you have an opinion as the service team member that it shouldn't be an issue, let me know

JumiDeluxe:
Thanks for questions

> Would it be preferable to add a virtual field "`force_delete`" that serves as a flag for sending the `obj["force"] = true` body during DELETE?

Due to how Terraform tries to operate on multiple resources at the same time, and it by default doesn't know about resources dependencies, enforcing that this option is selected always is necessary.
In a case, where user correctly removes all children resources along with `CodeRepositoryIndex` from Terraform files, an error might occur when Terraform tries to delete `CodeRepositoryIndex` before all of `RepositoryGroup` resources owned by it finish deletion.

Of course, a user only needs to apply the changes again, but we don't want users to have to do it just because of how Terraform works.

However, I think we could add `deletion_protection` field, as it seems like it is recommended now: https://googlecloudplatform.github.io/magic-modules/best-practices/deletion-behaviors/

mmv1/products/gemini/CodeRepositoryIndex.yaml
```diff
@@ -51,9 +51,17 @@ async:
...
   include_project: false
+custom_code:
+  pre_delete: templates/terraform/pre_delete/code_repository_index_force_delete.go.tmpl
 error_retry_predicates:
   - 'transport_tpg.IsCodeRepositoryIndexUnreadyError'
   - 'transport_tpg.IsRepositoryGroupQueueError'
+virtual_fields:
+  - name: 'force_destroy'
+    description:
+      If set to true, any RepositoryGroups for this CodeRepositoryIndex will also be deleted.
```
NickElliot, line 62:
```suggestion
      If set to true, will allow deletion of the CodeRepositoryIndex even if there are existing
      RepositoryGroups for the resource. These RepositoryGroups will also be deleted.
```

NickElliot (approved):
LGTM!

PR #12612, opened 2024-12-19
Remove 'diffSuppressSourceRanges' function from Firewall

doron8:
It seems that 'diffSuppressSourceRanges' function purpose is to suppress "0.0.0.0/0" to "" changes in source ranges field.

However:
- 0.0.0.0/0 is not equivalent to "", as "" also include all IPv6 addresses.
- 0.0.0.0/0 also means that source ranges field is specified, which is required in some situations (e.g. there is no other source specified).

The current function also causes an issue in which removing a source range is not applied - added tests to address such situations.

```release-note:bug
Remove 'diffSuppressSourceRanges' function from Firewall
```

### DIFF
--- a/mmv1/templates/terraform/constants/firewall.tmpl
+++ b/mmv1/templates/terraform/constants/firewall.tmpl
@@ -75,38 +75,3 @@ func resourceComputeFirewallSourceFieldsCustomizeDiff(_ context.Context, diff *s
 	return nil
 }
 
-func diffSuppressSourceRanges(k, old, new string, d *schema.ResourceData) bool {
-	if k == "source_ranges.#" {
-		if old == "1" && new == "0" {
-			// Allow diffing on the individual element if we are going from 1 -> 0
-			// this allows for diff suppress on ["0.0.0.0/0"] -> []
-			return true
-		}
-		// For any other source_ranges.# diff, don't suppress
-		return false
-	}
-	kLength := "source_ranges.#"
-	oldLength, newLength := d.GetChange(kLength)
-	oldInt, ok := oldLength.(int)
-
-	if !ok {
-		return false
-	}
-
-	newInt, ok := newLength.(int)
-	if !ok {
-		return false
-	}
-
-	// Diff suppress only should suppress removing the default range
-	// This should probably be newInt == 0, but due to Terraform core internals
-	// (bug?) values found via GetChange may not have the correct new value
-	// in some circumstances
-	if oldInt == 1 && newInt == 1 {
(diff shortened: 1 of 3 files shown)


mmv1/templates/terraform/constants/firewall.tmpl
```diff
@@ -75,38 +75,3 @@ func resourceComputeFirewallSourceFieldsCustomizeDiff(_ context.Context, diff *s
 	return nil
 }
 
-func diffSuppressSourceRanges(k, old, new string, d *schema.ResourceData) bool {
-	if k == "source_ranges.#" {
```
rileykarson, line 79: This was first added as Computed in https://github.com/hashicorp/terraform-provider-google/pull/147 to resolve https://github.com/hashicorp/terraform-provider-google/issues/108, and then later changed to a DiffSuppressFunc for https://github.com/hashicorp/terraform-provider-google/issues/8076 in https://github.com/GoogleCloudPlatform/magic-modules/pull/5309 to support the Firewall resource having evolved and needing more control over when diffs were suppressed / not suppressed.

As-is we'll reintroduce the original https://github.com/hashicorp/terraform-provider-google/issues/108 issue as we no longer have any protection against the `"0.0.0.0/0" => ""` diff. `Computed` is inappropriate, however, for the reasons MM#5309 was made.

I think there are two options:
- Remove the diff suppress func in 7.0.0, as is currently proposed. Users will be expected to specify `["0.0.0.0/0"]`, breaking old configurations. This will ship in late August, exact release schedule TBD (we will announce the timeline internally in a month or so).
- Use a CustomizeDiff, a more powerful function than DiffSuppressFunc, to identify cases where there is a diff from `["0.0.0.0/0"]` -> `[]` and rewrite the new proposed value to `["0.0.0.0/0"]` . This will be a little hairy to get working, and may take some time to develop/review/test- it goes off the "golden path" covered by https://googlecloudplatform.github.io/magic-modules/ so far.

Let me know which you'd prefer!

rileykarson:
Did you get that message when updating the resource, or just creating it? And could you capture `terraform state show` output in an update? IIRC the update case is the one prone to failure here.

(feel free to change project ids/any other identifiers in the state output)

rileykarson (commented):
Marking review to suppress the bot. Waiting on a reply to https://github.com/GoogleCloudPlatform/magic-modules/pull/12612#issuecomment-2638032888 iirc.

I'll be out for a few days, but feel free to message my team's oncall for a reassignment if you need a reviewer before I'm back.

PR #12466, opened 2024-12-02
Migrate google_compute_firewall_policy_association resource from DCL to MMv1

Samir-Cit:
Hello folks.
This PR is to move the Firewall Policy Association (`google_compute_firewall_policy_association`) resource from DCL to MMv1. And this PR includes:
- Creation of the resource on MMv1.
- Creation of the examples and tests.
- Remove references to DCL resource.

_**Note:** This PR doesn't include the deletion of the DCL resource. This change will be done on the [Delarative Resource Client Library](https://github.com/GoogleCloudPlatform/declarative-resource-client-library/tree/main) repository._

**Release Note Template for Downstream PRs (will be copied)**

```release-note:enhancement
compute: `google_compute_firewall_policy_association` now uses MMv1 engine instead of DCL.
```

PR #12614, opened 2024-12-19
Revert "Migrate google_compute_firewall_policy_association resource from DCL to MMv1"

melinath:
Reverts GoogleCloudPlatform/magic-modules#12466. It's missing a few elements needed for safe DCL -> MMv1 migration.

```release-note:enhancement
compute: `google_compute_firewall_policy_association` now uses MMv1 engine instead of DCL. (revert)
```


slevenick:
I'm not sure we actually need to do this. We were worried about the lack of legacy_long_form_project but for this resource in particular it may not be needed. This resource doesn't have a project field and takes a long form reference to a policy

@trodge what do you think?

melinath:
closing per above comment

PR #12417, opened 2024-11-23
Added new resource for managing envirommnent add-ons configuration

sharmajai09:
**Release Note Template for Downstream PRs (will be copied)**

See [Write release notes](https://googlecloudplatform.github.io/magic-modules/contribute/release-notes/) for guidance.

```release-note:REPLACEME

```

PR #12616, opened 2024-12-19
Added new Apigee resource /v1/organizations/<apigee_org>/environments/<apigee_env>/addonsConfig

sharmajai09:
```release-note:new-resource
`google_apigee_environment_addons_config`
```


melinath:
Taking this over as the original reviewer of https://github.com/GoogleCloudPlatform/magic-modules/pull/12417

melinath:
Manual test run: https://hashicorp.teamcity.com/buildConfiguration/TerraformProviders_GoogleCloud_GOOGLE_BETA_MMUPSTREAMTESTS_GOOGLEBETA_PACKAGE_APIGEE/288117

mmv1/products/apigee/EnvironmentAddonsConfig.yaml
```diff
@@ -0,0 +1,66 @@
...
+    'Enable Analytics Add-On': 'https://cloud.google.com/apigee/docs/api-platform/reference/manage-analytics-add-on'
+    'Enable ApiSecurity Add-On': 'https://cloud.google.com/apigee/docs/api-platform/reference/manage-security-add-on'
+  api: 'https://cloud.google.com/apigee/docs/reference/apis/apigee/rest/v1/organizations.environments.addonsConfig/setAddonEnablement'
+docs:
+base_url: '{{env_id}}/addonsConfig'
+self_link: '{{env_id}}/addonsConfig'
+create_url: '{{env_id}}/addonsConfig:setAddonEnablement'
+update_url: '{{env_id}}/addonsConfig:setAddonEnablement'
+update_verb: 'POST'
+exclude_delete: true
```
melinath, line 29: What would you think about sending a POST request to unset AddonEnablement on delete? This would match [the behavior of AddonsConfig](https://github.com/GoogleCloudPlatform/magic-modules/blob/55ec2413bf9eee962665d452b46aa629e3252a49/mmv1/products/apigee/AddonsConfig.yaml#L28-L29).

```suggestion
delete_url: '{{env_id}}/addonsConfig:setAddonEnablement' 
delete_verb: 'POST'
```

mmv1/products/apigee/EnvironmentAddonsConfig.yaml
```diff
@@ -0,0 +1,66 @@
...
+timeouts:
+  insert_minutes: 5
+  update_minutes: 5
```
melinath, line 32: Just want to double-check that this is desired; we usually only lengthen the timeouts, not shorten them.

mmv1/products/apigee/EnvironmentAddonsConfig.yaml
```diff
@@ -0,0 +1,66 @@
...
+async:
+  actions: ['create', 'update']
```
melinath, line 34: (if delete support is added)

```suggestion
  actions: ['create', 'update', 'delete']
```

mmv1/products/apigee/EnvironmentAddonsConfig.yaml
```diff
@@ -0,0 +1,66 @@
...
+  type: 'OpAsync'
+  operation:
+    base_url: '{{op_id}}'
+import_format:
+  - '{{env_id}}/addonsConfig'
+custom_code:
```
melinath, line 40: (if delete support is added) you'll also need a test_check_destroy similar to AddonsConfig https://github.com/GoogleCloudPlatform/magic-modules/blob/55ec2413bf9eee962665d452b46aa629e3252a49/mmv1/products/apigee/AddonsConfig.yaml#L43

mmv1/products/apigee/EnvironmentAddonsConfig.yaml
```diff
@@ -0,0 +1,66 @@
...
+  custom_import: 'templates/terraform/custom_import/apigee_env_addons.go.tmpl'
+  decoder: 'templates/terraform/decoders/apigee_env_addons.go.tmpl'
+examples:
+  - name: 'apigee_env_addons_analytics_test'
+    primary_resource_id: 'apigee_org_addons'
+    test_env_vars:
+      org_id: 'ORG_ID'
+      billing_account: 'BILLING_ACCT'
+    exclude_docs: true
```
melinath, line 49: There needs to be at least one docs example; you could add a doc-only example similar to AddonsConfig: https://github.com/GoogleCloudPlatform/magic-modules/blob/55ec2413bf9eee962665d452b46aa629e3252a49/mmv1/products/apigee/AddonsConfig.yaml#L45-L46

mmv1/products/apigee/EnvironmentAddonsConfig.yaml
```diff
@@ -0,0 +1,66 @@
...
+    skip_vcr: true
```
melinath, line 50: Just want to double-check that this needs to skip VCR tests? More information on the reasons for doing this (or not): https://googlecloudplatform.github.io/magic-modules/test/test/#skip-tests-in-vcr-replaying-mode

AddonsConfig tests don't seem to need this.

mmv1/products/apigee/EnvironmentAddonsConfig.yaml
```diff
@@ -0,0 +1,66 @@
...
+name: 'EnvironmentAddonsConfig'
+api_resource_type_kind: AddonsConfig
+description: Enable/Disable add-ons for an Apigee environment.
+references:
+  guides:
+    'Enable Analytics Add-On': 'https://cloud.google.com/apigee/docs/api-platform/reference/manage-analytics-add-on'
+    'Enable ApiSecurity Add-On': 'https://cloud.google.com/apigee/docs/api-platform/reference/manage-security-add-on'
+  api: 'https://cloud.google.com/apigee/docs/reference/apis/apigee/rest/v1/organizations.environments.addonsConfig/setAddonEnablement'
+docs:
+base_url: '{{env_id}}/addonsConfig'
```
melinath, line 24: The `addonsConfig` part doesn't need to be in the base_url, since this isn't a "real" resource. Technically the [base_url should probably be](https://googlecloudplatform.github.io/magic-modules/reference/resource/#base_url) `organizations/{{org_id}}/environments` but I think given the overrides in place this might actually not get used at all.

```suggestion
base_url: '{{env_id}}'
```

mmv1/products/apigee/EnvironmentAddonsConfig.yaml
```diff
@@ -0,0 +1,66 @@
...
+self_link: '{{env_id}}/addonsConfig'
```
melinath, line 25: This doesn't need addonsConfig, since it's not a "real" resource - leaving it off better matches the [AddonsConfig resource](https://github.com/GoogleCloudPlatform/magic-modules/blob/c53f2b3c7c416174b29f170656da81c320d2e7ed/mmv1/products/apigee/AddonsConfig.yaml#L24) behavior. 

```suggestion
self_link: '{{env_id}}'
```

mmv1/templates/terraform/custom_import/apigee_env_addons.go.tmpl
```diff
@@ -0,0 +1,13 @@
+config := meta.(*transport_tpg.Config)
+
+// current import_formats cannot import fields with forward slashes in their value
+if err := tpgresource.ParseImportId([]string{"(?P<env_id>.+)/addonsConfig"}, d, config); err != nil {
```
melinath, line 4:
```suggestion
if err := tpgresource.ParseImportId([]string{"(?P<env_id>.+)"}, d, config); err != nil {
```

mmv1/templates/terraform/custom_import/apigee_env_addons.go.tmpl
```diff
@@ -0,0 +1,13 @@
...
+	return nil, err
+}
+
+nameParts := strings.Split(d.Get("env_id").(string), "/")
+if len(nameParts) == 5 {
+	id := fmt.Sprintf("organizations/%s/environments/%s", nameParts[1], nameParts[3])
+	d.SetId(id)
+}
```
melinath, line 12: Should be something like this:

```suggestion
id := d.Get("env_id").(string)
nameParts := strings.Split(id, "/")
if len(nameParts) != 4 {  // you could be stricter about this check if you want.
        return nil, fmt.Errorf("env is expected to have shape organizations/{{"{{"}}org_id{{"}}"}}/environments/{{"{{"}}env{{"}}"}}, got %s instead", id)
}
d.SetId(id)
```

Any errors parsing the import id need to be returned - otherwise you might end up with an import that looks like it succeeds but fails to actually set an id (which means the resource won't actually be imported).

melinath (changes requested):

mmv1/products/apigee/EnvironmentAddonsConfig.yaml
sharmajai09: It's not possible with this api because Delete operation test sends null object as input to api. AddonsConfig api allows null object as input but this api doesn't allow. 

```
sharmajai@jai-desktop:~/magic-modules$ curl -H "$AUTH" -H "Content-Type: application/json" -X POST "https://apigee.googleapis.com/v1/organizations/test-apigee-org:setAddons"
{
  "name": "organizations/test-apigee-org/operations/4296f43d-9fa1-4341-bd00-d8221852dafa",
  "metadata": {
    "@type": "type.googleapis.com/google.cloud.apigee.v1.OperationMetadata",
    "operationType": "UPDATE",
    "targetResourceName": "organizations/test-apigee-org",
    "state": "IN_PROGRESS"
  }
}
sharmajai@jai-desktop:~/magic-modules$ curl -H "$AUTH" "https://apigee.googleapis.com/v1/organizations/test-apigee-org/operations/4296f43d-9fa1-4341-bd00-d8221852dafa"
{
  "name": "organizations/test-apigee-org/operations/4296f43d-9fa1-4341-bd00-d8221852dafa",
  "metadata": {
    "@type": "type.googleapis.com/google.cloud.apigee.v1.OperationMetadata",
    "operationType": "UPDATE",
    "targetResourceName": "organizations/test-apigee-org",
    "state": "FINISHED"
  },
  "done": true
}
sharmajai@jai-desktop:~/magic-modules$ curl -H "$AUTH" -H "Content-Type: application/json" -X POST "https://apigee.googleapis.com/v1/organizations/test-apigee-org/environments/test/addonsConfig:setAddonEnablement"
{
  "error": {
    "code": 400,
    "message": "addon enablement request \u003cnil\u003e is not supported",
    "status": "INVALID_ARGUMENT",
    "details": [
      {
        "requestId": "2385760879381416100"
      },
      {
        "@type": "type.googleapis.com/google.rpc.DebugInfo",
        "detail": "[ORIGINAL ERROR] generic::invalid_argument: addon enablement request \u003cnil\u003e is not supported [google.rpc.error_details_ext] { message: \"addon enablement request \u003cnil\u003e is not supported\" details { [type.googleapis.com/google.rpc.RequestInfo] { request_id: \"2385760879381416100\" } } }"
      }
    ]
  }
}
```

mmv1/products/apigee/EnvironmentAddonsConfig.yaml
sharmajai09: 5 minute timeout is ok. This api usually takes < 1 minute.

mmv1/products/apigee/EnvironmentAddonsConfig.yaml
sharmajai09: same as previous comment on supporting delete

mmv1/products/apigee/EnvironmentAddonsConfig.yaml
sharmajai09: same as previous comment on supporting delete

mmv1/products/apigee/EnvironmentAddonsConfig.yaml
sharmajai09: Done

mmv1/products/apigee/EnvironmentAddonsConfig.yaml
sharmajai09: Done

mmv1/products/apigee/EnvironmentAddonsConfig.yaml
sharmajai09: Done

mmv1/templates/terraform/custom_import/apigee_env_addons.go.tmpl
sharmajai09: Done

mmv1/products/apigee/EnvironmentAddonsConfig.yaml
sharmajai09: i had to add skip_vcr check back due to previously known error
```
Error waiting for Deleting Organization: error while retrieving operation: googleapi: Error 403: Permission 'apigee.operations.get' denied on resource 'organizations/tf-test-bawr39q93i/operations/55f073f3-0694-4c76-a387-a39b74c30e49' (or it may not exist)
```

mmv1/products/apigee/EnvironmentAddonsConfig.yaml
melinath: oh, it looks like the `organizations` API includes addonsConfig inline: https://cloud.google.com/apigee/docs/reference/apis/apigee/rest/v1/organizations

But the `environments` API doesn't: https://cloud.google.com/apigee/docs/reference/apis/apigee/rest/v1/organizations.environments

So yes, `/addonsConfig` is necessary here.

mmv1/products/apigee/EnvironmentAddonsConfig.yaml
melinath: unfortunate! Thanks for verifying.

melinath (changes requested):

melinath:

melinath:
reassigning since I'll be OOO next week.

mmv1/templates/terraform/examples/apigee_env_addons_enable_analytics.tf.tmpl
```diff
@@ -0,0 +1,70 @@
+resource "google_project" "project" {
```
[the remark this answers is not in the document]
sharmajai09: done I have kept Org and environment creation steps as this resource is available only when Org is PAYG and environment is COMPREHENSIVE.

mmv1/products/apigee/EnvironmentAddonsConfig.yaml
```diff
@@ -0,0 +1,64 @@
...
+timeouts:
+  insert_minutes: 5
+  update_minutes: 5
+async:
+  actions: ['create', 'update']
+  type: 'OpAsync'
+  operation:
+    base_url: '{{op_id}}'
+import_format:
+  - '{{env_id}}/addonsConfig'
```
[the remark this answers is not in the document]
sharmajai09: I am unable to comment on "This still seems to have addonsConfig - possibly the commit still needs to be pushed, or you backed this out because it was causing test failures (because there are a couple other changes needed to support it)"

So adding comment here: Yes if I remove addonsConfig some test check is failing.

melinath:
Sorry for the delay on this one; we had a miscommunication regarding the vacation plans! I can take it back over now that I'm back.

mmv1/templates/terraform/custom_import/apigee_env_addons.go.tmpl
```diff
@@ -0,0 +1,14 @@
+config := meta.(*transport_tpg.Config)
+
+// current import_formats cannot import fields with forward slashes in their value
+if err := tpgresource.ParseImportId([]string{"(?P<env_id>.+)/addonsConfig"}, d, config); err != nil {
```
melinath, line 4: The regex here matches what users pass in when using `terraform import` & definitely shouldn't contain `addonsConfig`

```suggestion
if err := tpgresource.ParseImportId([]string{"(?P<env_id>.+)"}, d, config); err != nil {
```

mmv1/products/apigee/EnvironmentAddonsConfig.yaml
```diff
@@ -0,0 +1,65 @@
...
+id_format:
+  - '{{env_id}}'
```
melinath, line 40: Sorry, I think I had suggested formatting this as a list but it should actually be just a string. This is causing the yaml parsing error.

```suggestion
id_format: '{{env_id}}'
```

melinath (changes requested):
a couple small changes

mmv1/products/apigee/EnvironmentAddonsConfig.yaml
sharmajai09: done

mmv1/templates/terraform/custom_import/apigee_env_addons.go.tmpl
sharmajai09: done 
somehow when I was running tests locally import worked only when I added /addonsConfig

melinath: I had you make some other changes (like adding the id_format) so that removing this here would work. I think your local testing may have been before those changes?

melinath (changes requested):
one small change but otherwise LGTM. the test failure is expected and is not due to a problem with this resource AFAICT. The update logic is similar enough to the create logic (even using the same API method) that I feel comfortable with not having an update test.

melinath (approved):

PR #12617, opened 2024-12-19
Update REVIEWER_ASSIGNMENT_COMMENT.md

melinath:
b/385135898

I don't think this would make sense to link from the missing test detector, and don't remember why we were considering that specifically - and while it's technically already linked here, adding another link seems reasonable.

```release-note:none

```


ScottSuarez:
I believe the bug, in context of the port mortem, is referring to:
* If a test is missing for a field, 
  * link to common cases where a field is untestable - and a fine-grained resource would be preferred

ScottSuarez (approved):
This change makes sense but doesn't resolve the bug imo

melinath:
Ah yeah that's totally what it was - thanks. I'll merge this and add the details to the ticket.

melinath:
actually the tests are failing; I'll need to fix that first.

melinath:
TGC unit test breakage is unrelated - there are no downstream diffs.

PR #12620, opened 2024-12-19
Add contributor documentation for resource metadata

roaks3:
Adds some light documentation for working with these metadata files (ie. `*_meta.yaml`).

Note that most regular mmv1 resources will not have to do anything additional, and we will be performing regular checks to correct/add files as needed.

```release-note:none

```

### DIFF
--- a/docs/content/reference/metadata.md
+++ b/docs/content/reference/metadata.md
@@ -5,7 +5,7 @@ weight: 35
 
 # MMv1 metadata reference
 
-This page documents all properties for metadata.
+This page documents all properties for metadata. Metadata does not impact the provider itself, but is used by Google internally for coverage metrics.
 
 ## Required
 
--- a/docs/content/reference/metadata.md
+++ b/docs/content/reference/metadata.md
@@ -1,0 +1,32 @@
+---
+title: "MMv1 metadata reference"
+weight: 35
+---
+
+# MMv1 metadata reference
+
+This page documents all properties for metadata.
+
+## Required
+
+### `resource`
+
+The name of the Terraform resource e.g., "google_cloudfunctions2_function".
+
+### `generation_type`
+
+The generation method used to create the Terraform resource e.g., "mmv1", "dcl", "handwritten".
+
+## Optional
+
+### `api_service_name`
+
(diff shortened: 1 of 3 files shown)


docs/content/reference/metadata.md
```diff
@@ -0,0 +1,32 @@
+---
+title: "MMv1 metadata reference"
+weight: 35
+---
+
+# MMv1 metadata reference
+
+This page documents all properties for metadata.
```
ScottSuarez, line 8: Could you add information about when this file is needed? You mention in you PR that most resources would be unaffected. When would they be affected?

ScottSuarez (changes requested):
Do you have more context on this change or a doc I can reference?

docs/content/reference/metadata.md
roaks3: I just added some clarification. The file is autogenerated for MMv1, so it is only really needed for handwritten resources (which is why that guide was updated as well).

roaks3:
Shared doc link with you internally

ScottSuarez:
I have some concerns about the overhead here. commented on the doc.

roaks3:
@ScottSuarez it sounds like maybe you have concerns about the approach, but the metadata described in this change is already implemented. Could we take that discussion offline, and move this through? Otherwise, we are in a place where we've introduced these fields/files without any documentation.

cc @melinath

docs/content/reference/metadata.md
```diff
@@ -0,0 +1,32 @@
+---
+title: "MMv1 metadata reference"
+weight: 35
+---
+
+# MMv1 metadata reference
+
+This page documents all properties for metadata. Metadata does not impact the provider itself, but is used by Google internally for coverage metrics.
```
melinath, line 8: Drive-by suggestion: maybe state this more generically (because that might not be the only thing it gets used for)

```suggestion
This page documents all properties for metadata. Metadata does not impact the provider itself, but enables analysis of the provider that would otherwise be impractical or impossible.
```

ScottSuarez (approved):

nityaravi (approved):

PR #12626, opened 2024-12-20
Fix error drop column with BigQuery flexible column names

shoumoji:
Fix: https://github.com/hashicorp/terraform-provider-google/issues/20750

```release-note:bug
bigquery: fixed DROP COLUMN error with bigquery flexible column names in `google_bigquery_table`
```

### DIFF
--- a/mmv1/third_party/terraform/services/bigquery/resource_bigquery_table.go.tmpl
+++ b/mmv1/third_party/terraform/services/bigquery/resource_bigquery_table.go.tmpl
@@ -2084,7 +2084,11 @@ func resourceBigQueryTableColumnDrop(config *transport_tpg.Config, userAgent str
 	}
 
 	if len(droppedColumns) > 0 {
-		droppedColumnsString := strings.Join(droppedColumns, ", DROP COLUMN ")
+		backquotedDroppedColumns := []string{}
+		for _, column := range droppedColumns {
+			backquotedDroppedColumns = append(backquotedDroppedColumns, fmt.Sprintf("`%s`", column))
+		}
+		droppedColumnsString := strings.Join(backquotedDroppedColumns, ", DROP COLUMN ")
 
 		dropColumnsDDL := fmt.Sprintf("ALTER TABLE `%s.%s.%s` DROP COLUMN %s", tableReference.project, tableReference.datasetID, tableReference.tableID, droppedColumnsString)
 		log.Printf("[INFO] Dropping columns in-place: %s", dropColumnsDDL)
--- a/mmv1/third_party/terraform/services/bigquery/resource_bigquery_table_test.go.tmpl
+++ b/mmv1/third_party/terraform/services/bigquery/resource_bigquery_table_test.go.tmpl
@@ -2206,8 +2206,8 @@ resource "google_bigquery_table" "test" {
     "type": "STRING"
   },
   {
-	"name": "flexible-column-name-dash",
-	"type": "STRING"
+    "name": "flexible-column-name-dash",
+    "type": "STRING"
   }
 ]
 EOH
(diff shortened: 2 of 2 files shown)


mmv1/third_party/terraform/services/bigquery/resource_bigquery_table.go.tmpl
```diff
@@ -2084,7 +2084,11 @@ func resourceBigQueryTableColumnDrop(config *transport_tpg.Config, userAgent str
 	}
 
 	if len(droppedColumns) > 0 {
-		droppedColumnsString := strings.Join(droppedColumns, ", DROP COLUMN ")
+		backquotedDroppedColumns := []string{}
+		for _, column := range droppedColumns {
+			backquotedDroppedColumns = append(backquotedDroppedColumns, fmt.Sprintf("`%s`", column))
+		}
+		droppedColumnsString := strings.Join(backquotedDroppedColumns, ", DROP COLUMN ")
```
shoumoji, line 2091: If `droppedColumns = []string("column1", "column2")` ,

```go
backquotedDroppedColumns = []string("`column1`", "`column2`")
droppedColumnsString = "`column1`, DROP COLUMN `column2`"
dropColumnsDDL = "ALTER TABLE <table_id> DROP Column `column1`, DROP COLUMN `column2`"
```

wj-chen:
Thank you for the fix! Can you also update the test in https://github.com/GoogleCloudPlatform/magic-modules/blob/1a28a91cf77e481d012718eec9bdf7524c663e4f/mmv1/third_party/terraform/services/bigquery/resource_bigquery_table_test.go.tmpl#L114 to cover dropping a column with a reserved word in the name?

shoumoji:
@wj-chen 

The test results revealed that the cause of this error is not a reserved word, but rather the use of [flexible column name](https://cloud.google.com/bigquery/docs/schemas#flexible-column-names) with a dash character.

Therefore, I will add tests and update of the issue and PR.


e.g. 

No Error : the column name `for` ( reserved word **"for"** )
Error       : the column name `foo-bar` ( no reserved word, but with **flexibile column name "-"** )

mmv1/third_party/terraform/services/bigquery/resource_bigquery_table_test.go.tmpl
```diff
@@ -2200,6 +2200,14 @@ resource "google_bigquery_table" "test" {
   {
     "name": "some_int",
     "type": "INTEGER"
+  },
+  {
+	"name": "reserved_word_for",
+	"type": "STRING"
```
shoumoji, line 2206: I added a test case with reserved words just in case, although it did not originally cause an error.

mmv1/third_party/terraform/services/bigquery/resource_bigquery_table_test.go.tmpl
```diff
@@ -2200,6 +2200,14 @@ resource "google_bigquery_table" "test" {
...
+  },
+  {
+	"name": "flexible-column-name-dash",
+	"type": "STRING"
```
shoumoji, line 2210: This case with a **flexible column name** caused an error in the original code.

mmv1/third_party/terraform/services/bigquery/resource_bigquery_table_test.go.tmpl
wj-chen: Can you align the indentation of this new block and the one below with the rest? i.e. using space instead of tab

wj-chen (approved):
One comment about indentation otherwise looks good, thank you

mmv1/third_party/terraform/services/bigquery/resource_bigquery_table_test.go.tmpl
shoumoji:
```suggestion
  "name": "reserved_word_for",
  "type": "STRING"
```

mmv1/third_party/terraform/services/bigquery/resource_bigquery_table_test.go.tmpl
shoumoji:
```suggestion
    "name": "flexible-column-name-dash",
    "type": "STRING"
```

shoumoji:
> One comment about indentation otherwise looks good, thank you

@wj-chen 
Thank you for the review!!
I fixed it.

shuyama1 (approved):
Thank you!

PR #12629, opened 2024-12-20
Allow AUTH_TYPE_UNSPECIFIED in auth config for connection

kanav-google:
Some connectors have auth types which are not among standard ones like Private App Token in Hubspot. These are introduced as part of auth type unspecified and the additional variables.

```release-note:bug
integrationconnectors: allowed `AUTH_TYPE_UNSPECIFIED` option in `google_integration_connectors_connection` resource to support non-standard auth types
```


roaks3 (approved):
Note this is intended to fix b/384477234 and b/380393940

PR #12630, opened 2024-12-20
Add support for parameter manager parameter resource `google_parameter_manager_parameter`

abheda-crest:
Add support for parameter manager parameter resource `google_parameter_manager_parameter`. 

More info about parameter manager: https://cloud.google.com/secret-manager/parameter-manager/docs/overview

```release-note:new-resource
`google_parameter_manager_parameter`
```


mmv1/products/parametermanager/product.yaml
```diff
@@ -0,0 +1,23 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+---
+name: 'ParameterManager'
+display_name: 'Parameter Manager'
+versions:
+  - name: 'ga'
```
slevenick, line 18: It looks like this API is in a Pre-GA state, so we shouldn't add it to the GA provider at this point.

mmv1/products/parametermanager/Parameter.yaml
```diff
@@ -0,0 +1,117 @@
...
+
+---
+name: 'Parameter'
+description: |
+  A Parameter resource is a logical parameter.
+references:
+  guides:
+  api: 'https://cloud.google.com/secret-manager/parameter-manager/docs/reference/rest/v1/projects.locations.parameters'
+docs:
+base_url: 'projects/{{project}}/locations/global/parameters'
```
slevenick, line 22: Will parameters ever be location-based? Or are they strictly limited to global forever?

mmv1/products/parametermanager/product.yaml
gptSanyam: Agreed, let's add this to beta provide only for now.

mmv1/products/parametermanager/Parameter.yaml
gptSanyam: The API is regional and global is considered as a region ( so base URL can be parameterised)
However the URL for global and regional API endpoints vary:
**Global**: parametermanager.googleapis.com
**Regional**: parametermanager.**{location}.rep.**googleapis.com

@slevenick Just because of this we will have to make keep regional and global resources. Can these be combined?

slevenick: Hmmmm, does the regional endpoint work for global? Like `parametermanager.**global.rep.**googleapis.com`?

If so, it should be easy to regionalize it now and allow global as a region.

Otherwise it should be possible, but may require some special casing. A lot of it depends on how you want to surface these resources to customers. Should they be separate resources in Terraform or not?

gptSanyam: Does the regional endpoint work for global? Like parametermanager.**global.rep.**googleapis.com?
 - It will not for global only parametermanager.googleapis.com will work

Regarding separate resources: Given that we don't have any change in logic, I don't have a strong reason why we should have separate resources apart from segregating the endpoints.

mmv1/third_party/terraform/.teamcity/components/inputs/services_ga.kt
```diff
@@ -581,6 +581,11 @@ var ServicesListGa = mapOf(
         "displayName" to "Parallelstore",
         "path" to "./google/services/parallelstore"
     ),
+    "parametermanager" to mapOf(
```
gptSanyam, line 584: DO we need to make change to the ga file - if we are only making a change in the beta provide?

slevenick: We shouldn't need it here

mmv1/products/parametermanager/Parameter.yaml
```diff
@@ -0,0 +1,130 @@
...
+    url_param_only: true
+    required: true
+    immutable: true
+properties:
+  - name: 'name'
+    type: String
+    description: |
+      The resource name of the Parameter. Format:
+      `projects/{{project}}/locations/global/parameters/{{parameter_id}}`
+    min_version: 'beta'
```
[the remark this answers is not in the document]
slevenick: Correct, please remove.

mmv1/products/parametermanager/Parameter.yaml
slevenick: We could do switching of the endpoint based on the value of a location field to combine the resources. We would need to make that decision now though, so we should discuss.

Such an approach would complicate use of custom endpoints for this resource though, so we may need to accommodate that somehow

slevenick (commented):
The resource is small overall, so it seems to be fine. We need to make a decision on how to handle global/regional endpoints either with this resource or by splitting the resource.

We should discuss this offline

mmv1/third_party/terraform/.teamcity/components/inputs/services_ga.kt
abheda-crest: I've removed `parametermanager` product mapping from this file. Just to be clear, previously I added this by referring to other beta-only products like [`apigateway`](https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/products/apigateway/product.yaml), [`firebase`](https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/products/firebase/product.yaml), [`gemini`](https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/products/gemini/product.yaml), etc. All these products are mapped in the [service_ga.kt](https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/third_party/terraform/.teamcity/components/inputs/services_ga.kt) file.

mmv1/products/parametermanager/Parameter.yaml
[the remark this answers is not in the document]
abheda-crest: I've removed `min_version` from the `parameters` and the `properties` of the resource but kept it for the examples. I've kept `min_version` for example because while generating beta provider after removing min_version field resulted in the following error:
Screenshot 2025-01-03 at 2 19 30 PM

mmv1/third_party/terraform/.teamcity/components/inputs/services_ga.kt
abheda-crest: It looks like the one of the Teamcity workflow is failing after I've removed the `parametermanager` mapping from the `services_ga.kt` file: https://github.com/GoogleCloudPlatform/magic-modules/actions/runs/12595156690/job/35103883286?pr=12630

abheda-crest: @slevenick @gptSanyam I've added the `parametermanager` product mapping again as one of the Teamcity workflow was failing. Let me know if I need to do otherwise.

mmv1/products/parametermanager/Parameter.yaml
gptSanyam: As discussed offline we will continue with having separate resources for global and regional Parameters and Parameter Versions.

gptSanyam (approved):

abheda-crest:
@slevenick Could you please review this PR?

abheda-crest:
@BBBmau Could you please review this PR?
**Note:** This PR is equivalent global resource for the PR #12631 that you reviewed and merged yesterday.

shuyama1:
Assigning a new reviewer, as the original reviewer is OOO

shuyama1:

mmv1/products/parametermanager/Parameter.yaml
melinath: It looks like the regional variants may need to be handwritten per https://github.com/hashicorp/terraform-provider-google/issues/12738#issuecomment-1273642960

mmv1/products/parametermanager/Parameter.yaml
```diff
@@ -0,0 +1,121 @@
...
+
+      No more than 64 labels can be assigned to a given resource.
+
+      An object containing a list of "key": value pairs. Example:
+      { "name": "wrench", "mass": "1.3kg", "count": "3" }.
+  - name: 'format'
+    type: Enum
+    description: |
+      The format type of the parameter resource. Default value is UNFORMATTED.
+    default_from_api: true
```
melinath, line 116: Is this default expected to be stable over time? If so, it would be a slightly better UX to use default_value rather than default_from_api. More info at https://googlecloudplatform.github.io/magic-modules/develop/diffs/#default.

mmv1/products/parametermanager/Parameter.yaml
melinath: making a regional variant product folder might be easier.

mmv1/third_party/terraform/services/parametermanager/resource_parameter_manager_parameter_test.go.tmpl
```diff
@@ -0,0 +1,152 @@
...
+  labels = {
+    key1 = "val1"
+    key2 = "val2"
+    key3 = "val3"
+    key4 = "val4"
+    key5 = "val5"
+  }
+}
+`, context)
+}
```
melinath, line 52: I think this fully duplicates the generated `TestAccParameterManagerParameter_parameterWithLabelsExample`: https://github.com/modular-magician/terraform-provider-google-beta/compare/auto-pr-12630-old..auto-pr-12630#diff-ff73922eed881cfc98b75567279d475bd05f6cd8163cb65d6d07ca6cc86309c1R102

```suggestion
```

melinath (changes requested):
A couple small questions/changes but overall LGTM

mmv1/products/parametermanager/Parameter.yaml
melinath: making a regional variant product folder might be easier.

mmv1/products/parametermanager/Parameter.yaml
abheda-crest: > Is this default expected to be stable over time?
What do you think @gptSanyam?

> If so, it would be a slightly better UX to use default_value rather than default_from_api. More info at https://googlecloudplatform.github.io/magic-modules/develop/diffs/#default.
However, AFAIK, default values are being sent in the request body itself. So, I think it would be better that the field is optional + computed. @melinath Let me know your thoughts.

mmv1/third_party/terraform/services/parametermanager/resource_parameter_manager_parameter_test.go.tmpl
abheda-crest: Done.

mmv1/products/parametermanager/Parameter.yaml
abheda-crest: @melinath We have already created a separate regional product folder along with the equivalent regional resource and this has been merged as a part of the PR https://github.com/GoogleCloudPlatform/magic-modules/pull/12631

mmv1/products/parametermanager/Parameter.yaml
melinath: What is the problem with sending default values in the request body? That is also what happens if a user sets a value.

abheda-crest: I think if in future the default from the API side changes than it would require the change to be done in the provider too.

abheda-crest: But, Yeah, I agree with your point that if default from the API is going to stable than we can also use the default values.

melinath: Let's just leave this as default_from_api since the regional variant was already merged with that. For future reference, default_value is preferred where possible for the reasons listed above, but it's not the end of the world to use default_from_api instead.

melinath: Okay, SGTM - the other resource was merged on Wednesday so it should still be safe to change that.

abheda-crest: @melinath I've used `default_value` as `UNFORMATTED` instead of `default_from_api` for both `google_parameter_manager_parameter` and `google_parameter_manager_regional_parameter` resource.

melinath (approved):
LGTM

melinath:
test failures are unrelated

PR #12636, opened 2024-12-22
Upgrade google.golang.org/api to v0.214.0

chayankumar999:

```release-note: note
upgrade google.golang.org/api to v0.214.0
```


chayankumar999:
To me, it seems like the test those are failing are unrelated. The failure of test TestAccComputeRegionPerInstanceConfig_removeInstanceOnDestroy seems a bit strange to me. So I run this test again with the current head of the commit and it's fails there as well but with [different error message](https://screenshot.googleplex.com/AcZ4irFEwAYzAtn). So I am not sure if this is because of the PR or not. 
I double check all the diff for go.mod and go.sum files and verified that all the changes that requires for upgrade the version was added in this PR.

rileykarson (approved):
These don't seem related to the upgrade

chayankumar999:
Thanks @rileykarson for approving the changes. :)

PR #12644, opened 2024-12-23
Promoting namespace to the GA provider

erikmej-code:
In an effort to promote the namespace to the GA provider we do the following:

1. Remove instances of `min_version` from `Namespace.yaml` and include the `uid` field within `properties`.
2. Include a test file, `service_directory_namespace_ga.tf.tmpl`, with the provider omitted.

```release-note:new-resource
`google_service_directory_namespace` (GA)
```

### DIFF
--- a/mmv1/products/servicedirectory/Namespace.yaml
+++ b/mmv1/products/servicedirectory/Namespace.yaml
@@ -46,10 +46,6 @@ examples:
     primary_resource_id: 'example'
     vars:
       namespace_id: 'example-namespace'
-  - name: 'service_directory_namespace_ga'
-    primary_resource_id: 'example-ga'
-    vars:
-      namespace_id: 'example-namespace-ga'
 parameters:
   - name: 'location'
     type: String
--- a/mmv1/products/servicedirectory/Namespace.yaml
+++ b/mmv1/products/servicedirectory/Namespace.yaml
@@ -80,9 +80,3 @@ properties:
       Resource labels associated with this Namespace. No more than 64 user
       labels can be associated with a given resource. Label keys and values can
       be no longer than 63 characters.
-  - name: 'uid'
-    type: String
-    description: |
-      The unique identifier of the namespace.
-    output: true
-    exact_version: 'ga'
--- a/mmv1/products/servicedirectory/Namespace.yaml
+++ b/mmv1/products/servicedirectory/Namespace.yaml
@@ -16,7 +16,6 @@ name: 'Namespace'
 description: |
   A container for `services`. Namespaces allow administrators to group services
   together and define permissions for a collection of services.
-min_version: 'beta'
 references:
   guides:
     'Configuring a namespace': 'https://cloud.google.com/service-directory/docs/configuring-service-directory#configuring_a_namespace'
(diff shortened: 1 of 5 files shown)


mmv1/templates/terraform/examples/service_directory_namespace_ga.tf.tmpl
```diff
@@ -0,0 +1,9 @@
+resource "google_service_directory_namespace" "{{$.PrimaryResourceId}}" {
```
shuyama1, line 1: I believe this is the same example as https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/templates/terraform/examples/service_directory_namespace_basic.tf.tmpl. There's no need to explicitly include an example for ga. It will be automatically generated for both ga and beta providers once the `min_version: 'beta'` flag is removed from the resource.

shuyama1 (commented):
You'll also need to add a GA endpoint in https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/products/servicedirectory/product.yaml

mmv1/templates/terraform/examples/service_directory_namespace_ga.tf.tmpl
erikmej-code: Will Remove.

erikmej-code:
Updated product.yaml file.

mmv1/products/servicedirectory/Namespace.yaml
```diff
@@ -45,17 +44,19 @@ exclude_sweeper: true
 examples:
   - name: 'service_directory_namespace_basic'
     primary_resource_id: 'example'
-    min_version: 'beta'
     vars:
       namespace_id: 'example-namespace'
+  - name: 'service_directory_namespace_ga'
+    primary_resource_id: 'example-ga'
+    vars:
+      namespace_id: 'example-namespace-ga'
```
shuyama1, line 52:
```suggestion
```
We also need to remove this, given the example file is removed

shuyama1 (commented):
We'll need to remove the `provider     = google-beta` in the example file https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/templates/terraform/examples/service_directory_namespace_basic.tf.tmpl#L2, as the example will now be used for GA as well

mmv1/products/servicedirectory/Namespace.yaml
```diff
@@ -77,12 +73,16 @@ properties:
...
-    min_version: 'beta'
     output: true
   - name: 'labels'
     type: KeyValueLabels
     description: |
       Resource labels associated with this Namespace. No more than 64 user
       labels can be associated with a given resource. Label keys and values can
       be no longer than 63 characters.
-    min_version: 'beta'
+  - name: 'uid'
```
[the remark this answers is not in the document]
erikmej-code: Will fix.

erikmej-code:
@shuyama1 these comments should now be resolved. Please review at your earliest convenience. Thanks!

mmv1/templates/terraform/custom_expand/sd_full_url.tmpl
```diff
@@ -4,7 +4,7 @@ func expand{{$.GetPrefix}}{{$.TitlelizeProperty}}(v interface{}, d tpgresource.T
   } else if strings.HasPrefix(v.(string), "https://") {
     return v, nil
   }
-  url, err := tpgresource.ReplaceVars(d, config, "{{"{{"}}ServiceDirectoryBasePath{{"}}"}}" + v.(string))
```
shuyama1, line 7: We generally aim to keep each PR focused on a single change to make issue detection and revert easier. Since this change is separate from the GA promotion, I’ve moved it to a new PR ([#12934](https://github.com/GoogleCloudPlatform/magic-modules/pull/12934)) and added tests to verify the fix. We can remove this change from the current PR and proceed with the other PR for the basepath override. Please let me know if it works for you. Thanks!

shuyama1 (changes requested):

PR #12659, opened 2024-12-27
Add foreign type info and definition to google_bigquery_table beta

wj-chen:
Add foreign type info and definition to `google_bigquery_table` beta.

```release-note:enhancement
bigquery: added `schema_foreign_type_info` field and related schema handling to `google_bigquery_table` resource (beta)
```

### DIFF
--- a/mmv1/products/bigquery/Table.yaml
+++ b/mmv1/products/bigquery/Table.yaml
@@ -270,6 +270,28 @@ properties:
                 - 'TIME'
                 - 'DATETIME'
                 - 'RECORD'
+                - 'FOREIGN'
+            - name: 'foreignTypeDefinition'
+              type: String
+              description: |
+                Definition of the foreign data type.
+                Only valid for top-level schema fields (not nested fields).
+                If the type is FOREIGN, this field is required.
+              min_version: beta
+  - name: 'schemaForeignTypeInfo'
+    type: NestedObject
+    description: |
+      Specifies metadata of the foreign data type definition in field schema.
+    min_version: beta
+    properties:
+      - name: 'typeSystem'
+        type: Enum
+        description: |
+          The foreign type of the table.
+        required: true
+        enum_values:
+          - 'HIVE'
+        min_version: beta
   - name: 'encryptionConfiguration'
     type: NestedObject
     description: Custom encryption configuration
--- a/mmv1/third_party/terraform/services/bigquery/resource_bigquery_table.go.tmpl
+++ b/mmv1/third_party/terraform/services/bigquery/resource_bigquery_table.go.tmpl
@@ -96,6 +96,14 @@ func jsonCompareWithMapKeyOverride(key string, a, b interface{}, compareMapKeyVa
 			unionOfKeys[subKey] = true
 		}
 
+		{{- if ne $.TargetVersionName "ga" }}
+		// Disregard "type" and "fields" if "foreignTypeDefinition" is present since they may have been modified by the server.
+		if _, ok := unionOfKeys["foreignTypeDefinition"]; ok {
+			delete(unionOfKeys, "type")
+			delete(unionOfKeys, "fields")
+		}
+
+		{{- end }}
 		for subKey := range unionOfKeys {
 			eq := compareMapKeyVal(subKey, objectA, objectB)
 			if !eq {
@@ -323,6 +331,15 @@ func resourceBigQueryTableSchemaIsChangeable(old, new interface{}, isExternalTab
 		for key := range objectNew {
 			unionOfKeys[key] = true
 		}
+
+		{{- if ne $.TargetVersionName "ga" }}
+		// Disregard "type" and "fields" if "foreignTypeDefinition" is present since they may have been modified by the server.
+		if _, ok := unionOfKeys["foreignTypeDefinition"]; ok {
+			delete(unionOfKeys, "type")
+			delete(unionOfKeys, "fields")
+		}
+
+		{{- end }}
 		for key := range unionOfKeys {
 			valOld := objectOld[key]
 			valNew := objectNew[key]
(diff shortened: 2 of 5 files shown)


mmv1/products/bigquery/Table.yaml
```diff
@@ -270,6 +270,28 @@ properties:
...
+        min_version: beta
+        properties:
+          - name: 'typeSystem'
+            type: Enum
+            description: |
+              The foreign type of the table.
+            required: true
+            enum_values:
+              - 'HIVE'
+            min_version: beta
```
shuyama1, line 294:
```suggestion
    - name: 'schemaForeignTypeInfo'
      type: NestedObject
      description: |
        Specifies metadata of the foreign data type definition in field schema.
      min_version: beta
      properties:
        - name: 'typeSystem'
          type: Enum
          description: |
            The foreign type of the table.
          required: true
          enum_values:
            - 'HIVE'
          min_version: beta
```
I believe we're not currently using this template file since this resource handwritten. However, let's retain these fields in the file and align the naming for consistency. Additionally, if I understand correctly, this should be placed outside the schema block.

wj-chen: Makes sense, done.

mmv1/third_party/terraform/services/bigquery/resource_bigquery_table.go.tmpl
```diff
@@ -992,6 +1009,25 @@ func ResourceBigQueryTable() *schema.Resource {
 				DiffSuppressFunc: bigQueryTableSchemaDiffSuppress,
 				Description:      `A JSON schema for the table.`,
 			},
+			{{- if ne $.TargetVersionName "ga" }}
+			// SchemaForeignTypeInfo: [Optional] Specifies metadata of the foreign data type definition in field schema.
+			"schema_foreign_type_info": {
```
shuyama1, line 1014: Does this field support updates? If so, could we add an update step in the test?

mmv1/third_party/terraform/services/bigquery/resource_bigquery_table.go.tmpl
```diff
@@ -992,6 +1009,25 @@ func ResourceBigQueryTable() *schema.Resource {
...
+				Type:        schema.TypeList,
+				Optional:    true,
+				MaxItems:    1,
+				Description: "Specifies metadata of the foreign data type definition in field schema.",
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						// TypeSystem: [Optional] Specifies the system which defines the foreign data type.
+						"type_system": {
+							Type:             schema.TypeString,
+							Optional:         true,
```
shuyama1, line 1024: Can we make this field required, given this is currently the only subfield?

shuyama1 (commented):
I wanted to confirm if this feature is already released. Could you share the API doc? I couldn't find it on https://cloud.google.com/bigquery/docs/reference/rest/v2/tables. Also for the bigquery service, we use `https://bigquery.googleapis.com/bigquery/v2/` for both GA and beta providers. Is there a specific reason this feature is limited to beta?

mmv1/third_party/terraform/services/bigquery/resource_bigquery_table.go.tmpl
wj-chen: It doesn't and API will throw an error. The behavior might change between Public Preview and GA or post-GA, so we are also not making it force-new right now.

mmv1/third_party/terraform/services/bigquery/resource_bigquery_table.go.tmpl
wj-chen: Thanks for catching this, it should be required.

wj-chen:
> I wanted to confirm if this feature is already released. Could you share the API doc? I couldn't find it on https://cloud.google.com/bigquery/docs/reference/rest/v2/tables. Also for the bigquery service, we use `https://bigquery.googleapis.com/bigquery/v2/` for both GA and beta providers. Is there a specific reason this feature is limited to beta?

This is scheduled for a Public Preview launch in January, hence beta-provider-only until GA. I will check with the API team to see when the public reference doc is expected to update. In the meantime you can find the corresponding definitions in internal search.

shuyama1 (commented):
> > I wanted to confirm if this feature is already released. Could you share the API doc? I couldn't find it on [cloud.google.com/bigquery/docs/reference/rest/v2/tables](https://cloud.google.com/bigquery/docs/reference/rest/v2/tables). Also for the bigquery service, we use `https://bigquery.googleapis.com/bigquery/v2/` for both GA and beta providers. Is there a specific reason this feature is limited to beta?
> 
> This is scheduled for a Public Preview launch in January, hence beta-provider-only until GA. I will check with the API team to see when the public reference doc is expected to update. In the meantime you can find the corresponding definitions in internal search.

Thanks for confirming! Moving the discussion about the feature launch status to  internal channel.

mmv1/third_party/terraform/services/bigquery/resource_bigquery_table.go.tmpl
```diff
@@ -96,6 +96,14 @@ func jsonCompareWithMapKeyOverride(key string, a, b interface{}, compareMapKeyVa
 			unionOfKeys[subKey] = true
 		}
 
+		{{- if ne $.TargetVersionName "ga" }}
+		// Disregard "type" and "fields" if "foreignTypeDefinition" is present since they may have been modified by the server.
+		if _, ok := unionOfKeys["foreignTypeDefinition"]; ok {
+			delete(unionOfKeys, "type")
+			delete(unionOfKeys, "fields")
```
shuyama1, line 103: Question - Would users be able to change type and fields if foreignTypeDefinition is present?

mmv1/third_party/terraform/services/bigquery/resource_bigquery_table.go.tmpl
shuyama1: Thanks for confirming. I suggest we set it to `forcenew` for now and remove it once the update is supported. Otherwise users won't be able to change this field even if they want to delete and recreate the resource to align with the desired state. However, if the concern is to avoid accidentally deleting the table and the update will be supported soon, we could keep it as is for now

mmv1/third_party/terraform/services/bigquery/resource_bigquery_table.go.tmpl
wj-chen: User-supplied values will be ignored. Service will re-populate type and fields based on the given foreignTypeDefinition value.

wj-chen:
> Thanks for confirming. I suggest we set it to forcenew for now and remove it once the update is supported. Otherwise users won't be able to change this field even if they want to delete and recreate the resource to align with the desired state. However, if the concern is to avoid accidentally deleting the table and the update will be supported soon, we could keep it as is for now

Update support is unplanned so far. Added `ForceNew: true` in the meantime.

shuyama1 (approved):

PR #12664, opened 2024-12-27
Fix nodeConfigDefaults on cluster create.

hoskeri:
The expansion of
`node_pool_defaults.node_config_defaults.insecure_kubelet_readonly_port_enabled` should have a ForceSendFields to allow `FALSE` values to be propagated.

Without this "FALSE" values are not included in the request.

This is not an issue in nodePoolAutoConfig, or during updates. ForceSendFields is set in those cases.

An alternative fix is to use expandNodeKubeletConfig function here as well. The server will reject any values other than the readonly port field.

```release-note:bug
container: fixed propagation of `node_pool_defaults.node_config_defaults.insecure_kubelet_readonly_port_enabled` in node config.
```


c2thorn:
waiting on tests for the scenario

hoskeri:
@c2thorn Added create tests for autopilot and standard clusters, please take a look!

c2thorn (approved):

PR #12667, opened 2024-12-30
artifactregistry: accept all valid durations

Subserial:
Fixes https://github.com/hashicorp/terraform-provider-google/issues/20796

```release-note:bug
artifactregistry: fix `artifact_registry_repository` not accepting durations with 'm', 'h' or 'd'
```


SirGitsalot:
I think you're looking for a custom _expander_ rather than a custom _flattener_ (that is, a function the converts a TF request to an API request) rather than a flattener (which converts an API response to its TF equivalent). The API types that field as a `google.protobuf.Duration` which [must be encoded in seconds](https://protobuf.dev/reference/protobuf/google.protobuf/#duration) in JSON requests.

Subserial:
It's actually more annoying than that: since the field is in a map, the object gets hashed before I read the value, meaning I can't correlate the objects to be compared. I'm moving upwards and writing a diff suppress function on the map itself.

Subserial:
The cleaner solution was to use the `id` field as a hash key. I wonder why this isn't the default behavior for maps.

SirGitsalot (approved):

PR #12689, opened 2025-01-06
Add storage_project_management_hub resource.

kautikdk:
This PR adds support for GCS Storage Intelligence on Project level.

```release-note: new-resource
`google_storage_control_project_intelligence_config`
```


kautikdk:
Hi @trodge, The above VCR tests failures don't look like it's coming from this PR changes. Can you please check if it is common issue across all magic-modules PRs?

kautikdk:
Hi @trodge, The test failures are unrelated so we can start PR review.

trodge:
> While playing around with some customized configs, I found that there is a edge case scenario which we have to handle either from Terraform side or from API side. Nested TypeList fields like `filter.includedCloudStorageBuckets.bucketIdRegexes` and `filter.excludedCloudStorageLocations.locations` can have empty values. These empty values are empty list, `[]`, and they can cause trouble as we can not get these values in the API response if we pass them in the request. Backend service will clear that filter field and won't return empty value in the API response. Now our flatteners will not parse this empty value in the state if it is not coming from the API so ultimately there will be a situation when user config will have empty list defined but state file won't have anything and it will generate permadiff for that field. I am discussing this API behavior internally with the relevant team and will see if it is the expected behavior and won't change in the future. Also if we use `send_empty_value` field then we will end up sending null values of every filter fields which are not specified or have empty value inside them, [reference](https://googlecloudplatform.github.io/magic-modules/reference/field/#send_empty_value). I believe we can handle this scenario with DiffSupressFunc or we can guide users not use empty values inside of the nested attributes rather remove the complete block from the filter block which will eventually do the same thing. Let me know what you think!
> 
> Example scenario,
> 
> ```
> resource "google_storage_control_project_intelligence_config" "project_storage_intelligence" {
>   name = "<intentionally_removed>"
>   edition_config = "STANDARD"
>   filter {
>     included_cloud_storage_buckets{
> 	bucket_id_regexes = []
>     }
>   }
> }
> ```

A DiffSuppressFunc might work for this.

kautikdk:
Hi @trodge, Added some `DiffSuppressFunc` to avoid empty list issue. I am not sure why these unrelated tests are keep failing.

trodge (approved):

PR #12690, opened 2025-01-06
Allow read of PC type from API response to enable non-standard PC imports

swamitagupta:
Remove ignore_read on the 'type' parameter of Private Cloud.

Context:
1. 'type' is an optional and immutable field in the PC resource. If not set by the user, the API response always returns type="STANDARD" post PC creation. 
2. ignore_read was used to retain 'type' as an optional field in terraform and suppress the errors which can arise due to terraform state and API response discrepancies. 
3. This prevented imports of PCs with non-standard types such as "TIME_LIMITED" and "STRETCHED", as the type field would not be read from the API response during the import of the resource and raise warnings to force replacement.

This PR removes ignore_read from the type parameter. The diff suppress on the type field ensures that differences between "" and "STANDARD" types are suppressed to continue supporting the expected behavior as described in point 2. 

The acceptance test has been modified to test the import of the resource with different types including "TIME_LIMITED", "STANDARD" AND "". 

```release-note:enhancement
vmwareengine: allowed import of non-STANDARD private clouds in `google_vmwareengine_private_cloud`
```

### DIFF
--- a/mmv1/products/vmwareengine/PrivateCloud.yaml
+++ b/mmv1/products/vmwareengine/PrivateCloud.yaml
@@ -394,7 +394,6 @@ properties:
     description: |
       Initial type of the private cloud.
     immutable: true
-    ignore_read: true
     diff_suppress_func: 'vmwareenginePrivateCloudStandardTypeDiffSuppressFunc'
     enum_values:
       - 'STANDARD'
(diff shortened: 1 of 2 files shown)


swamitagupta:
Acceptance test passing logs: gpaste/6285415838121984

rileykarson (approved):
LGTM, thanks!

PR #12691, opened 2025-01-06
Create github action to ensure service team labels exist in the correct color

ScottSuarez:
closes https://github.com/hashicorp/terraform-provider-google/issues/20754
```release-note:none

```


tools/issue-labeler/labeler/backfill.go
```diff
@@ -1,95 +1,87 @@
...
 }
 
-func ComputeIssueUpdates(issues []Issue, regexpLabels []RegexpLabel) []IssueUpdate {
+// ComputeIssueUpdates remains the same as it doesn't interact with GitHub API
+func ComputeIssueUpdates(issues []*github.Issue, regexpLabels []RegexpLabel) []IssueUpdate {
 	var issueUpdates []IssueUpdate
 
 	for _, issue := range issues {
-		if len(issue.PullRequest) > 0 {
+		if issue.PullRequestLinks == nil || issue.PullRequestLinks.GetURL() == "" {
```
melinath, line 78: What would you think about using the IsPullRequest helper instead? https://pkg.go.dev/github.com/google/go-github/v68/github#Issue.IsPullRequest

tools/issue-labeler/labeler/github.go
```diff
@@ -0,0 +1,78 @@
...
+		return "", "", fmt.Errorf("unexpected repository format %s", repository)
+	}
+
+	owner = or[0]
+	repo = or[1]
+	return owner, repo, nil
+}
+
+func ensureLabelWithColor(client *github.Client, owner, repo, labelName, color string) error {
+	existingLabel, err := getLabel(client, owner, repo, labelName)
```
melinath, line 37: Rather than get each service label, what would you think about listing all the labels up front using https://docs.github.com/en/rest/issues/labels?apiVersion=2022-11-28#list-labels-for-a-repository?

That would enable a function that takes the list of labels in the file vs the API and computes the labels that need to be edited / created, which could be unit tested, and then the outputs could be used to make the actual API calls.

melinath (changes requested):
I'd be happy to approve the go-github changes in a separate PR while we continue discussing the label implementation, if that works for you?

tools/issue-labeler/labeler/backfill.go
ScottSuarez: sure, I've moved all this to another PR. 
https://github.com/GoogleCloudPlatform/magic-modules/pull/12701/files

tools/issue-labeler/labeler/github.go
ScottSuarez: sounds reasonable

.github/workflows/ensure-service-team-labels.yml
```diff
@@ -0,0 +1,30 @@
+name: ensure-service-team-labels
```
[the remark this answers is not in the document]
ScottSuarez: No not yet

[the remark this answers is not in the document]
ScottSuarez: I'm planning to use a workflow call to re-use file this in tpg

melinath (changes requested):
could you add some basic unit tests? Probably the easiest thing to test would be the core computation of which labels should be updated / added (which would mean factoring it out) but open to other options as well.

melinath (approved):

PR #12696, opened 2025-01-07
Add Terraform support for the Security Gateway Application resource

Berro321:
Adding support for the Security Gateway Application resource from the Beyondcorp API. This change is dependent on 
https://github.com/GoogleCloudPlatform/magic-modules/pull/12695 to submit first as it's the parent resource.

```release-note:new-resource
`google_beyondcorp_application`
```


ScottSuarez:
converting to draft for now.

PR #12700, opened 2025-01-07
add iamPolicyVersion field in MM docs

BBBmau:

adds `iamPolicyVersion` field in MMv1 docs

```release-note:none

```

### DIFF
--- a/docs/content/develop/add-iam-support.md
+++ b/docs/content/develop/add-iam-support.md
@@ -44,6 +44,9 @@ iam_policy:
   # Overrides the HTTP method for setIamPolicy. Default: 'POST'
   # set_iam_policy_verb: 'POST'
 
+  # Overrides the default IamPolicyVersion (version: 1)
+  iam_policy_version: '1'
+
   # Must match the parent resource's `import_format` (or `self_link` if
   # `import_format` is unset), but with the `parent_resource_attribute`
   # value substituted for the final field.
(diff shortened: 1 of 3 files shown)


BBBmau:

docs/content/reference/resource.md
```diff
@@ -292,6 +292,7 @@ attributes – for a full reference, see
   `'REQUEST_BODY'`, `'QUERY_PARAM_NESTED'`. Note: `'QUERY_PARAM_NESTED'` should
   only be used if the query param field contains a `.`
 - `min_version: beta`: Marks IAM support as beta-only.
+- `iam_policy_version`: Version number in the request payload. Default: 1
```
melinath, line 295: The default behavior here is actually more complex. Unset (0) means "use the default IAM policy version set in the provider" rather than having a default set in MMv1. Currently, the providers default to version 3; we could just hardcode that but since it doesn't change frequently I'm concerned we'd forget to update this spot. I also don't want to give the impression that there's an MMv1 default.

Maybe something like this?

```suggestion
- `iam_policy_version`: Version number in the request payload. If set, it overrides the default ([tpgiamresource.IamPolicyVersion](https://github.com/hashicorp/terraform-provider-google-beta/blob/main/google-beta/tpgiamresource/iam.go))
```

docs/content/develop/add-iam-support.md
```diff
@@ -43,6 +43,9 @@ iam_policy:
   # Overrides the HTTP method for setIamPolicy. Default: 'POST'
   # set_iam_policy_verb: 'POST'
 
+  # Overrides the default IamPolicyVersion. Default: 1
```
melinath, line 46: this is pretty rarely used - I only see 1 instance currently - so we probably shouldn't include it in the default template. The detailed reference should be sufficient.

PR #12703, opened 2025-01-07
Add post_read custom_code

coder-221:
This is to unblock a future PR for etags in `access_context_manager` resources. We'll need to access the etag in the GET response that's made from the nested_query code. For consistency, I also added the `post_read` to the normal resource read function as well, but that's not strictly required for our use case.

I'm also not sure what the params `false -` are for in the template, so would appreciate a check to ensure those are correct. Thanks!

```release-note:none

```

### DIFF
--- a/mmv1/templates/terraform/nested_query.go.tmpl
+++ b/mmv1/templates/terraform/nested_query.go.tmpl
@@ -243,6 +243,9 @@ func resource{{ $.ResourceName }}ListForPatch(d *schema.ResourceData, meta inter
     return nil, err
   }
 
+{{- if $.CustomCode.PostRead }}
+  {{ $.CustomTemplate $.CustomCode.PostRead false -}}
+{{- end }}
   var v interface{}
   var ok bool
 {{- range $i, $k := $.NestedQuery.Keys }}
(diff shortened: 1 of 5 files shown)


mmv1/templates/terraform/nested_query.go.tmpl
```diff
@@ -243,6 +243,9 @@ func resource{{ $.ResourceName }}ListForPatch(d *schema.ResourceData, meta inter
     return nil, err
   }
 
+{{- if $.CustomCode.PostRead }}
+  {{ $.CustomTemplate $.CustomCode.PostRead false -}}
```
c2thorn, line 247: the CustomTemplate function is from here: https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/api/resource.go#L1510

the `false` will toggle off the option for appending a new line - a feature we really only cared about for an older migration and should not be relevant either way for you

the `-` tells the go template to delete any whitespace after the statement. Getting an example working with this should show how that ends up looking

c2thorn (changes requested):
Can you provide an example template and use it with `post_read` in some resource's YAML file so we can see that its working?

The template can just add dummy comments so it won't have any functional change to the resource

c2thorn (approved):
Exactly what I wanted to see. Generated output looks good, so looks like `post_read` is good to go. Thanks @coder-221 !

PR #12712, opened 2025-01-08
Add tags field to Filestore Instances for TagsR2401

aniket-gupta:
Part of b/364841739

```release-note:enhancement
filestore: added support for `tags` field to `google_filestore_instance` resource
```


aniket-gupta:
@NickElliot Could you please review this PR

aniket-gupta:
@NickElliot Could you please review this PR

aniket-gupta:
> can you just make the tags a more fixed value instead of using this formatted string+for loop? It looks more complicated to read than it needs to be

tag key and tag value can not be fixed. tag keys and values are get (or create) by acctest.BootstrapSharedTestTagKey and acctest.BootstrapSharedTestTagValue
https://github.com/hashicorp/terraform-provider-google/blob/969fce388d73044bb90a80467629d924be2a5c04/google/acctest/bootstrap_test_utils.go#L1431
https://github.com/hashicorp/terraform-provider-google/blob/969fce388d73044bb90a80467629d924be2a5c04/google/acctest/bootstrap_test_utils.go#L1493

NickElliot (approved):
LGTM

PR #12715, opened 2025-01-09
Colab runtime

bcreddy-gcp:

```release-note:new-resource
`google_colab_runtime`
```

### DIFF
--- a/mmv1/products/colab/Runtime.yaml
+++ b/mmv1/products/colab/Runtime.yaml
@@ -1,0 +1,81 @@
+# Copyright 2025 Google Inc.
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+---
+
+name: 'Runtime'
+description: |
+  'A runtime is a Google-provisioned virtual machine (VM) that can run the code in your notebook (IPYNB file).'
+
+references:
+  guides:
+    'Create a runtime': 'https://cloud.google.com/colab/docs/create-runtime'
+  api: 'https://cloud.google.com/vertex-ai/docs/reference/rest/v1/projects.locations.notebookRuntimes'
+base_url: 'projects/{{project}}/locations/{{location}}/notebookRuntimes'
+self_link: 'projects/{{project}}/locations/{{location}}/notebookRuntimes/{{name}}'
+create_url: 'projects/{{project}}/locations/{{location}}/notebookRuntimes:assign?notebook_runtime_id={{name}}'
+async:
+  type: 'OpAsync'
+  operation:
+    full_url: 'https://{{location}}-aiplatform.googleapis.com/v1/{{op_id}}'
+custom_code:
+  encoder: 'templates/terraform/encoders/colab_runtime.go.tmpl'
+examples:
+  - name: 'colab_runtime_basic'
+    primary_resource_id: 'runtime'
+    primary_resource_name: 'fmt.Sprintf("tf-test-colab-runtime%s", context["random_suffix"])'
+    region_override: 'us-central1'
+    vars:
+      runtime_name: 'colab-runtime'
+  - name: 'colab_runtime_full'
+    primary_resource_id: 'runtime'
+    primary_resource_name: 'fmt.Sprintf("tf-test-colab-runtime%s", context["random_suffix"])'
+    region_override: 'us-central1'
+    vars:
+      runtime_name: 'colab-runtime'
+      key_name: 'my-crypto-key'
+    test_vars_overrides:
+      key_name: 'acctest.BootstrapKMSKeyInLocation(t, "us-central1").CryptoKey.Name'
+parameters:
+  - name: 'location'
+    type: String
+    required: true
+    url_param_only: true
+    description: 'The location for the resource: https://cloud.google.com/colab/docs/locations'
+  - name: 'name'
+    type: String
+    url_param_only: true
+    description: 'The resource name of the Runtime'
+properties:
+  - name: notebookRuntimeTemplateRef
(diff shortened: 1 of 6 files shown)


mmv1/products/colab/Runtime.yaml
```diff
@@ -0,0 +1,91 @@
...
+    description: 'The user email of the NotebookRuntime.'
+  - name: 'displayName'
+    type: String
+    description:
+      Required. The display name of the Runtime.
+    required: true
+  - name: description
+    type: String
+    description: 'The description of the Runtime Template.'
+  - name: 'labels'
```
rileykarson, line 89: @zli82016 bit of a weird case here- the API will reject user labels but will set some system ones. That means we want to mark this output-only for now, and may introduce configurable labels later. This gets weird with default labels / the attribution label which will cause a diff at that point.

To ensure forward compatibility I'm thinking:

1. Add the output-only field as `effective_labels` as TypeMap, i.e. don't enable special labels behaviour yet
1. If/when the API allows user labels, introduce resource-level labels as `labels`, but exclude provider-level labels from the resource. This will allow for mixed resource and server labels without a diff, and won't generate a diff to set provider-level labels, letting us release the functionality in a minor version.
1. In the next possible major version, enable full default label support

Specifically we start with `effective_labels` instead of `labels` as the output field to avoid writing the server labels into `labels`' state, which would require a StateUpgrader to clean up later during step 2. That would also affect configs that read values from the field, in a backwards incompatible way when we do step 2- since we'd null out the old outputs. `effective_labels` will be stable across all 3 steps.

Let me know if you have concerns / feel I missed anything!

bcreddy-gcp: Could we instead remove the labels field and add it once it is supported?

rileykarson: The default provider labels still complicate things- we'd skip step 1, but still do step 2 and 3.

zli82016: In step 1,  `KeyValuePairs` type can be used for the labels field and then add `output: true`. There is a validation for labels that if the field name is  `labels` and its type `KeyValuePairs`, the validation function will throw error. Adding the resource to the [list](https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/api/type.go#L904)  will prevent the error.

madhusuraj (approved):

mmv1/templates/terraform/examples/colab_runtime_basic.tf.tmpl
```diff
@@ -0,0 +1,36 @@
+resource "google_project_iam_member" "colab_enterprise_admin" {
```
rileykarson, line 1: We can clear IAM grants as https://github.com/GoogleCloudPlatform/magic-modules/pull/12772 will grant them (we may want to comment "Needs X permission" or "Needs X role" above or on the `runtime_user` line)

mmv1/products/colab/Runtime.yaml
```diff
@@ -0,0 +1,88 @@
...
+examples:
+  - name: 'colab_runtime_basic'
+    primary_resource_id: 'runtime'
+    primary_resource_name: 'fmt.Sprintf("tf-test-colab-runtime%s", context["random_suffix"])'
+    region_override: 'us-central1'
+    vars:
+      runtime_name: 'colab-runtime'
+    test_env_vars:
+      project_id: 'PROJECT_NAME'
+      service_account: 'SERVICE_ACCT'
```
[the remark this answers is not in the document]
bcreddy-gcp: Removed

mmv1/templates/terraform/examples/colab_runtime_basic.tf.tmpl
```diff
@@ -0,0 +1,36 @@
...
+resource "google_colab_runtime" "{{$.PrimaryResourceId}}" {
+  name = "{{index $.Vars "runtime_name"}}"
+  location = "us-central1" 
+  
+  notebook_runtime_template_ref {
+    notebook_runtime_template = google_colab_runtime_template.my_template.id
+  }
+  
+  display_name = "Runtime basic"
+  runtime_user = "example@example.com"
```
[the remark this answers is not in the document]
bcreddy-gcp: Done

mmv1/templates/terraform/examples/colab_runtime_basic.tf.tmpl
bcreddy-gcp: Done

mmv1/templates/terraform/examples/colab_runtime_full.tf.tmpl
```diff
@@ -0,0 +1,75 @@
...
+  }
+
+  data_persistent_disk_spec {
+    disk_type    = "pd-standard"
+    disk_size_gb = 200
+  }
+
+  network_spec {
+    enable_internet_access = true
+    network = google_compute_network.my_network.id
```
rileykarson, line 31: Alternatively, could we use the default network or a shared network for all tests in the product, or does it need to be dedicated?

mmv1/templates/terraform/examples/colab_runtime_full.tf.tmpl
```diff
@@ -0,0 +1,75 @@
+resource "google_compute_network" "my_network" {
+  name = "{{index $.Vars "network_name"}}"
+  auto_create_subnetworks = false
+}
+
+resource "google_compute_subnetwork" "my_subnetwork" {
```
rileykarson, line 6: Can we chain a `time_sleep` between the subnetwork and runtime template to add time to the deletion? I'd try 30s. https://registry.terraform.io/providers/hashicorp/time/latest/docs/resources/sleep#delay-destroy-usage

bcreddy-gcp: Network is not even a runtime parameter. Getting rid of it.

mmv1/templates/terraform/examples/colab_runtime_full.tf.tmpl
bcreddy-gcp: Network is not even a runtime parameter. Getting rid of it.

mmv1/products/colab/Runtime.yaml
bcreddy-gcp: For now, I am getting rid of the labels field. We can look at it once the API starts supporting it.

rileykarson (approved):
LGTM, thanks!

PR #12722, opened 2025-01-09
Attachment 100g

dkarwowski:
Adds support for BPS_100G InterconnectAttachment enum

```release-note:enhancement
compute: added `BPS_100G` enum value to `bandwidth` field of `google_compute_interconnect_attachment`.
```


shuyama1 (approved):
@dkarwowski Thanks! We might consider switching the field type from enum to string if the possible values are expected to change regularly.

PR #12376, opened 2024-11-21
Remove the tagHoldAdmin for TestAccContainerCluster_resourceManagerTags since we already have bootstrap it

MaChenhao:
Removed the tagHoldAdmin for TestAccContainerCluster_resourceManagerTags since we already have bootstrap it. Also added the tagAdmin IAM permission in order to update the tag.

Fix for https://github.com/hashicorp/terraform-provider-google/issues/20252.

**Release Note Template for Downstream PRs (will be copied)**

```release-note:none
container: Removed the tagHoldAdmin for TestAccContainerCluster_resourceManagerTags since we already have bootstrap it. Also added the tagAdmin IAM permission in order to update the tag.
```

PR #12728, opened 2025-01-10
container: fixed resourceManagerTags tests

wyardley:
Rework of #12376 by @MaChenhao

This fixes some tests that may have started failing after my #12014.

Turned out to be a bigger can of worms than I expected, and is now updated to include autopilot and nodepool variants as well.

Fixes https://github.com/hashicorp/terraform-provider-google/issues/19997
Fixes https://github.com/hashicorp/terraform-provider-google/issues/20252
Closes #12376

```release-note:none

```

### DIFF
--- a/mmv1/third_party/terraform/services/container/resource_container_cluster_test.go.tmpl
+++ b/mmv1/third_party/terraform/services/container/resource_container_cluster_test.go.tmpl
@@ -62,6 +62,7 @@ func TestAccContainerCluster_resourceManagerTags(t *testing.T) {
 	t.Parallel()
 
 	pid := envvar.GetTestProjectFromEnv()
+	project := envvar.GetTestProjectNumberFromEnv()
 
 	randomSuffix := acctest.RandString(t, 10)
 	clusterName := fmt.Sprintf("tf-test-cluster-%s", randomSuffix)
@@ -69,7 +70,8 @@ func TestAccContainerCluster_resourceManagerTags(t *testing.T) {
 	networkName := acctest.BootstrapSharedTestNetwork(t, "gke-cluster")
 	subnetworkName := acctest.BootstrapSubnet(t, "gke-cluster", networkName)
 
-	if acctest.BootstrapPSARoles(t, "service-", "container-engine-robot", []string{"roles/resourcemanager.tagAdmin", "roles/resourcemanager.tagHoldAdmin", "roles/resourcemanager.tagUser"}) {
+	if acctest.BootstrapPSARoles(t, "service-", "container-engine-robot", []string{"roles/resourcemanager.tagAdmin", "roles/resourcemanager.tagHoldAdmin", "roles/resourcemanager.tagUser"}) ||
+		acctest.BootstrapPSARole(t, project, "@cloudservices", "roles/resourcemanager.tagUser") {
 		t.Fatal("Stopping the test because a role was added to the policy.")
 	}
 
@@ -3636,13 +3638,15 @@ func TestAccContainerCluster_withAutopilotResourceManagerTags(t *testing.T) {
 	t.Parallel()
 
 	pid := envvar.GetTestProjectFromEnv()
+	project := envvar.GetTestProjectNumberFromEnv()
 
 	randomSuffix := acctest.RandString(t, 10)
 	clusterName := fmt.Sprintf("tf-test-cluster-%s", randomSuffix)
 	clusterNetName := fmt.Sprintf("tf-test-container-net-%s", randomSuffix)
 	clusterSubnetName := fmt.Sprintf("tf-test-container-subnet-%s", randomSuffix)
 
-	if acctest.BootstrapPSARoles(t, "service-", "container-engine-robot", []string{"roles/resourcemanager.tagAdmin", "roles/resourcemanager.tagHoldAdmin", "roles/resourcemanager.tagUser"}) {
+	if acctest.BootstrapPSARoles(t, "service-", "container-engine-robot", []string{"roles/resourcemanager.tagAdmin", "roles/resourcemanager.tagHoldAdmin", "roles/resourcemanager.tagUser"}) ||
+		acctest.BootstrapPSARole(t, project, "@cloudservices", "roles/resourcemanager.tagUser") {
 		t.Fatal("Stopping the test because a role was added to the policy.")
 	}
 
@@ -3672,7 +3676,7 @@ func TestAccContainerCluster_withAutopilotResourceManagerTags(t *testing.T) {
 				Check: resource.ComposeTestCheckFunc(
 					// Small sleep first, to avoid condition where cluster is ready but underlying GCE
 					// resources apparently aren't.
-					acctest.SleepInSecondsForTest(30),
+					acctest.SleepInSecondsForTest(15),
 					resource.TestCheckResourceAttrSet("google_container_cluster.with_autopilot", "node_pool_auto_config.0.resource_manager_tags.%"),
 				),
 			},
@@ -11776,20 +11780,6 @@ data "google_project" "project" {
   project_id = "%[1]s"
 }
 
-resource "google_project_iam_member" "tag_user" {
-  project = "%[1]s"
-  role    = "roles/resourcemanager.tagUser"
-  member  = "serviceAccount:${data.google_project.project.number}@cloudservices.gserviceaccount.com"
-}
-
-resource "time_sleep" "wait_120_seconds" {
-  create_duration = "120s"
-
-  depends_on = [
-    google_project_iam_member.tag_user,
-  ]
-}
-
 resource "google_tags_tag_key" "key1" {
   parent      = data.google_project.project.id
   short_name  = "foobarbaz-%[2]s"
@@ -11844,8 +11834,6 @@ resource "google_container_cluster" "primary" {
   deletion_protection = false
   network             = "%[4]s"
   subnetwork          = "%[5]s"
-
-  depends_on = [time_sleep.wait_120_seconds]
 }
 `, projectID, randomSuffix, clusterName, networkName, subnetworkName, tagResourceNumber)
 }
@@ -11854,20 +11842,6 @@ func testAccContainerCluster_withAutopilotResourceManagerTags(projectID, cluster
 	return fmt.Sprintf(`
 data "google_project" "project" {
   project_id = "%[1]s"
-}
-
-resource "google_project_iam_member" "tag_user" {
-  project = "%[1]s"
-  role    = "roles/resourcemanager.tagUser"
-  member  = "serviceAccount:${data.google_project.project.number}@cloudservices.gserviceaccount.com"
-}
-
-resource "time_sleep" "wait_120_seconds" {
-  create_duration = "120s"
-
-  depends_on = [
-    google_project_iam_member.tag_user,
-  ]
 }
 
 resource "google_tags_tag_key" "key1" {
@@ -11964,8 +11938,6 @@ resource "google_container_cluster" "with_autopilot" {
   vertical_pod_autoscaling {
     enabled = true
   }
-
-  depends_on = [time_sleep.wait_120_seconds]
 }
 `, projectID, randomSuffix, clusterName, networkName, subnetworkName)
 }
@@ -11974,20 +11946,6 @@ func testAccContainerCluster_withAutopilotResourceManagerTagsUpdate1(projectID, 
 	return fmt.Sprintf(`
 data "google_project" "project" {
   project_id = "%[1]s"
-}
-
-resource "google_project_iam_member" "tag_user" {
-  project = "%[1]s"
-  role    = "roles/resourcemanager.tagUser"
-  member  = "serviceAccount:${data.google_project.project.number}@cloudservices.gserviceaccount.com"
-}
-
-resource "time_sleep" "wait_120_seconds" {
-  create_duration = "120s"
-
-  depends_on = [
-    google_project_iam_member.tag_user,
-  ]
 }
 
 resource "google_tags_tag_key" "key1" {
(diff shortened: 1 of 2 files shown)


wyardley:
@ScottSuarez thanks for running these. Can you see what the error is - if it just needs a re-run after bootstrapping, or if you see a different error?

ScottSuarez:
sure !
```
2025/01/14 18:50:48 [DEBUG] [transport] [server-transport 0xc000850340] loopyWriter exiting with error: transport closed by client 
    resource_container_cluster_test.go:3608: Step 3/6 error: Error running apply: exit status 1
        
        Error: Error waiting for updating GKE cluster node pool auto config resource manager tags: Google Compute Engine: Required 'resourcemanager.tagValueBindings.create' permission for 'tagValues/281481010616245'.
        
          with google_container_cluster.with_autopilot,
          on terraform_plugin_test.tf line 94, in resource "google_container_cluster" "with_autopilot":
          94: resource "google_container_cluster" "with_autopilot" {
        
```

wyardley:
@ScottSuarez I was able to reproduce a similar failure locally on another run. I think the first pass maybe didn't go far enough, so made some more adjustments.

I don't think we can use `acctest.BootstrapPSARole()` on the `serviceAccount:${data.google_project.project.number}@cloudservices.gserviceaccount.com` one (at least without changes to that function, because there is no `iam` in the right hand side of the email, for one thing), but I think we can use it for granting the `container-engine-robot` `roles/resourcemanager.tagUser` permissions. If you've got ideas / suggestions about the other one, just let me know.

I rebased, made some adjustments, and will push up the fix if that seems to help, but you may want to run a couple of times in CI, even if it works. There's an outside chance that it could still be a little flaky, but hopefully this helps.

I ran into some issues when running both tests at the same time, but I assume / hope that won't be an issue with the way these run in the actual test suite (is there a chance that the two tests run in parallel against the same project at any point)?

The autopilot is a pain to test because it takes a while, and is kind of finnicky. With the updated code, I'm getting these failures semi-consistently on the autopilot one, occasionally (on the same step, I am thinking 2nd step, based on the requested tags and since it's 3/6 and there are two runs per step):
```
    resource_container_cluster_test.go:3481: Step 3/6 error: Error running apply: exit status 1
        
        Error: Error waiting for updating GKE cluster node pool auto config resource manager tags: Google Compute Engine: The instance's current status does not support update.
        
          with google_container_cluster.with_autopilot,
          on terraform_plugin_test.tf line 87, in resource "google_container_cluster" "with_autopilot":
          87: resource "google_container_cluster" "with_autopilot" {
```
Under the hood, this is what's happening (heavily snipped, but I think it gives a flavor of what's happening):
```
PUT /v1/projects/xxx/locations/us-central1/clusters/tf-test-cluster-lmdz1yz3cz?alt=json&prettyPrint=false HTTP/1.1
[...]
{
 "update": {
  "desiredNodePoolAutoConfigResourceManagerTags": {
   "tags": {
    "tagKeys/281476493792018": "tagValues/281483533413074",
    "tagKeys/281484657323346": "tagValues/281482241168330"
   }
  }
 }
}
HTTP/2.0 200 OK
[...]
{
 "name": "operation-1736919116983-c37bdb21-ed20-454d-b575-25a6e3e4a896",
 "zone": "us-central1",
 "operationType": "UPDATE_CLUSTER",
 "status": "RUNNING",
 "selfLink": "https://container.googleapis.com/v1/projects/xx/locations/us-central1/operations/operation-1736919116983-c37bdb21-ed20-454d-b575-25a6e3e4a896",
 "targetLink": "https://container.googleapis.com/v1/projects/xx/locations/us-central1/clusters/tf-test-cluster-lmdz1yz3cz",
 "detail": "Updating default-pool, done with 0 out of 1 nodes (0.0%): 1 being processed",
[...]
GET /v1/projects/xxx/locations/us-central1/operations/operation-1736919116983-c37bdb21-ed20-454d-b575-25a6e3e4a896?alt=json&prettyPrint=false HTTP/1.1
HTTP/2.0 200 OK
{
 "name": "operation-1736919116983-c37bdb21-ed20-454d-b575-25a6e3e4a896",
 "zone": "us-central1",
 "operationType": "UPDATE_CLUSTER",
 "status": "DONE",
 "statusMessage": "Google Compute Engine: The instance's current status does not support update.",
 "selfLink": "https://container.googleapis.com/v1/projects/xx/locations/us-central1/operations/operation-1736919116983-c37bdb21-ed20-454d-b575-25a6e3e4a896",
 "targetLink": "https://container.googleapis.com/v1/projects/xx/locations/us-central1/clusters/tf-test-cluster-lmdz1yz3cz",
 "detail": "Google Compute Engine: The instance's current status does not support update.",
[...]
 },
 "clusterConditions": [
  {
   "message": "Google Compute Engine: The instance's current status does not support update.",
   "canonicalCode": "INVALID_ARGUMENT"
  }
 ],
 "error": {
  "code": 3,
  "message": "Google Compute Engine: The instance's current status does not support update."
 }
}
```

If I'm reading this right, the cluster thinks it's in ready state, but the underlying GCE instances aren't actually ready to take the update yet? Not sure if this expected behavior and / or if an issuetracker bug should be filed about the cluster reporting being ready when it's not (I didn't find a public one, nor did I find any search results for this exact error message).

The 120s sleep in the Terraform code won't apply after the first step, because it only applies to create actions.... next thing I'm trying is adding a sleep within step 2 of the test itself, though this will slow down test execution further, even in replaying mode (side note: I see VCR enabled tests with `acctest.SleepInSecondsForTest()` -- maybe there should be a condition skipping the sleep when VCR is enabled there? created an [issue](https://github.com/hashicorp/terraform-provider-google/issues/20916) to ask that)

wyardley:
Another possibility for this failing earlier in CI is, as mentioned above, if the two tests are both recording in parallel, maybe the grant for `[project-number]@cloudservices.gserviceaccount.com` got deleted? The good news is, if this were the case, it should now succeed on the next run if the other test is already passing and thus running in replaying mode.... but that condition would still technically exist. For that to be fixable, I think `acctest.BootstrapPSARole()` would have to be adjusted / extended (e.g., to allow suppressing the `.iam` in the domain), and or a similar mechanism made available for bootstrapping that SA.

mmv1/third_party/terraform/services/container/resource_container_cluster_test.go.tmpl
```diff
@@ -3635,6 +3643,9 @@ func TestAccContainerCluster_withAutopilotResourceManagerTags(t *testing.T) {
 			{
 				Config: testAccContainerCluster_withAutopilotResourceManagerTagsUpdate1(pid, clusterName, clusterNetName, clusterSubnetName, randomSuffix),
 				Check: resource.ComposeTestCheckFunc(
+					// Small sleep first, to avoid condition where cluster is ready but underlying GCE
+					// resources apparently aren't.
+					acctest.SleepInSecondsForTest(30),
```
wyardley, line 3648: Other possibility is that I ran into some sort of shorter term problem while I was testing this and that it's not actually necessary. But this did reliably seem to get rid of the issue I mentioned in the PR comments.

See notes and linked issue about whether this is really needed in the case of VCR, though.

wyardley:
@ScottSuarez so, assuming the error is the same as before, I think the issue may be the issue with the two parallel tests trying to manage the same permissions for `[project-number]@cloudservices.gserviceaccount.com` (and the permission getting deleted by one test when the other one is still running. Does that seem likely to you, and if so, do you or anyone on your team have any thoughts about the best way to approach this? If you give me some direction on whether and how to adjust `acctest.BootstrapPSARole()`, I could take that on (as a separate PR, or as part of this one), but I'm not sure that's the right way to go.

Separately, we could decide whether to keep or get rid of that extra sleep.

ScottSuarez:
```
    resource_container_cluster_test.go:3610: Step 3/6 error: After applying this test step and performing a `terraform refresh`, the plan was not empty.
        stdout
        
        
          + create
        
          # google_project_iam_member.tag_user will be created
          + resource "google_project_iam_member" "tag_user" {
              + etag    = (known after apply)
              + id      = (known after apply)
              + member  = "serviceAccount:xxxx@cloudservices.gserviceaccount.com"
              + project = "ci-test-project-yyyy"
              + role    = "roles/resourcemanager.tagUser"
            }
        
```

ScottSuarez:
> @ScottSuarez so, assuming the error is the same as before, I think the issue may be the issue with the two parallel tests trying to manage the same permissions for `[project-number]@cloudservices.gserviceaccount.com` (and the permission getting deleted by one test when the other one is still running. Does that seem likely to you, and if so, do you or anyone on your team have any thoughts about the best way to approach this? If you give me some direction on whether and how to adjust `acctest.BootstrapPSARole()`, I could take that on (as a separate PR, or as part of this one), but I'm not sure that's the right way to go.
> 
> Separately, we could decide whether to keep or get rid of that extra sleep.

Yeah looks like that might be the case. The terraform refresh seems to suggest that we're modifying a singleton. I would say we should maintain separate values for these two tests if we can. I haven't looked into it in as much detail as you as to if that would be possible

ScottSuarez:
Is there a reason we still maintain a `google_project_iam_member` in the test in question? Couldn't we also bootstrap that permission?

wyardley:
> Is there a reason we still maintain a google_project_iam_member in the test in question? Couldn't we also bootstrap that permission?

@ScottSuarez what I was getting at above is that `.iam` is baked in to the email domain in `BootstrapAllPSARoles()` and related functions currently:
https://github.com/GoogleCloudPlatform/magic-modules/blob/f5c8c97f6c6c79ec00e40fde05adfb7762fdba0a/mmv1/third_party/terraform/acctest/bootstrap_iam_test_utils.go#L41

So from what I can see, I don't believe it's currently possible to use `BootstrapPSARoles()` or `BootstrapPSARole()` to bootstrap permissions on `NNNNNNNNNN@cloudservices.gserviceaccount.com` without code changes in those functions [because there's no `.iam` after the `@`], and I didn't see (from a quick look) any other functions that seem like they'd work. If you or your team would like to give an example of how it's possible to bootstrap that account, or create a PR to support that, I'd be happy to implement it, or if there are clear instructions on how it should be implemented, I could throw up a PR for that separately?

I think with golang not having default function parameters, and the nesting of those functions, it could get a little messy to do, maybe something like if it's nil, use `iam.` and if it's `""` suppress `iam.` entirely?

I can switch the existing set of calls for 3 spearate roles from `BootstrapPSARole()` to `BootstrapPSARoles()`, which at least tidies up the logic in the existing bootstrap logic.

ScottSuarez:
> @ScottSuarez what I was getting at above is that .iam is baked in to the email domain in BootstrapAllPSARoles() and related functions currently:

Ah this makes sense. I think it would be reasonable to make the IAM function more general here and allow modification of the email. Just expose new wrapper functions.

wyardley:
> Ah this makes sense. I think it would be reasonable to make the IAM function more general here and allow modification of the email. Just expose new wrapper functions

Sounds good. Does it make sense to implement that as a separate PR? And, in either event, do you or anyone from your team have any suggestions on the cleanest way to implement it?

ScottSuarez:
We could build the agent name beforehand and just check if the agent name is fully quantified. If so do a passthrough to add the fully quantified name.

https://github.com/hashicorp/terraform-provider-google/blob/1884b7e1073e981bf9e4793b1258506bcd3f56ed/google/acctest/bootstrap_iam_test_utils.go#L43

strings.Contains(s, "@") or similar. 

We could also do some advanced string building.. Similar to what we do for url replacements.
https://github.com/hashicorp/terraform-provider-google/blob/1884b7e1073e981bf9e4793b1258506bcd3f56ed/google/provider/universe/universe_domain_pubsub_test.go#L76

ScottSuarez:
I think if the implementation is simple enough we could add it here, otherwise if you're looking for a a more complicated implementation we should make a new PR.

wyardley:
> If so do a passthrough to add the fully quantified name.

That's a good idea! I will take a look and see if I can figure out how to do that easily.

wyardley:
So the test passing on its own kind of confirms that probably this is what is happening (that role getting revoked). Let me see how hard it will be to extend the bootstrap functions (hopefully in a minimal way).

wyardley:
Ok, I have a pretty simple implementation that is hacky, but works (tested in replaying mode, and testing running both concurrently in recording mode now locally), and that I think is somewhat flexible while not having too many changes.

I pushed it up in its own commit, and cherry-picked into #12784 - my suggestion is that if we can get fast review on this, it will be cleaner for the commit history if it's merged first, and then I can rebase this.

wyardley:
Also @ScottSuarez I tried backing off that sleep to 15s, and got more failures locally. It seems to me like maybe this is a bug. I'm guessing this is something that will not rank as very important, but I created https://issuetracker.google.com/issues/390456348 anyway, this way I can at least link to that in the comment in the code here.

wyardley:
#12785 -- this is a potential solution to the sleep in `REPLAYING` mode, though as currently written, it would affect all ~4 uses of that function.

wyardley:
In the meantime, I'll push up a fix for hashicorp/terraform-provider-google#19997

ScottSuarez (changes requested):
waiting on IAM changes from out of band PR

wyardley:
@ScottSuarez @melinath thanks for the quick fix.... ecc4b3739e81e348f54ce83cdbb078a8838ee2af includes the changes from #12796, and seems to work for me locally. Only open question is whether it's Ok to share that bootstrap function (modeled after the example in there) across the container and node pool test files, or if I should duplicate it and / or extract out to another file. Happy to change the naming of that function if you'd like.

melinath:
I don't have a strong feeling about it. Whatever @ScottSuarez thinks WFM.

ScottSuarez (approved):
Looks good to me! Thanks for your work and persistence here !

Also thanks to Stephen for making the bootstrap function more generalized !

PR #12734, opened 2025-01-13
Promote Cloud Armor Security Policies for Regional Backend Services from Beta to GA

cstanger:
Promote Cloud Armor Security Policies for Regional Backend Services from Beta to GA

```release-note: note
compute:  promoted `security_policy` for `google_compute_region_backend_service`  from Beta to GA
```

### DIFF
--- a/mmv1/third_party/terraform/services/compute/resource_compute_region_backend_service_test.go.tmpl
+++ b/mmv1/third_party/terraform/services/compute/resource_compute_region_backend_service_test.go.tmpl
@@ -1174,7 +1174,6 @@ resource "google_compute_health_check" "health_check" {
 }
 {{- end }}
 
-{{ if ne $.TargetVersionName `ga` -}}
 func TestAccComputeRegionBackendService_withSecurityPolicy(t *testing.T) {
 	t.Parallel()
 
@@ -1223,4 +1222,3 @@ resource "google_compute_region_security_policy" "policy" {
 }
 `, serviceName, polLink, polName)
 }
-{{- end }}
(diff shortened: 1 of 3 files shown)


mmv1/third_party/terraform/services/compute/resource_compute_region_backend_service_test.go.tmpl
```diff
@@ -1174,7 +1174,6 @@ resource "google_compute_health_check" "health_check" {
 }
 {{- end }}
 
-{{ if ne $.TargetVersionName `ga` -}}
 func TestAccComputeRegionBackendService_withSecurityPolicy(t *testing.T) {
```
melinath, line 1177: This test fails in GA with the error:

> The provider hashicorp/google does not support resource type
>        "google_compute_region_security_policy".

It looks like that resource is beta-only. https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/compute_region_security_policy

However, the API for that resource does seem to be available in its GA API - possibly it also needs to be promoted? That would be good to do in a separate PR if that sounds correct.

melinath (changes requested):

mmv1/third_party/terraform/services/compute/resource_compute_region_backend_service_test.go.tmpl
cstanger: thanks for checking - drafted the dependent Pull Request here 
https://github.com/GoogleCloudPlatform/magic-modules/pull/12748

PR #12735, opened 2025-01-13
Finish converting ACM resources to use policy mutex lock

coder-221:
This is a continuation of #12725 to migrate all ACM resources to use a policy level mutex lock. Some resources already have a policy field, so I just used that. Others don't have one so I created a new access_policy_id property and populate it in the encoder. Note, I've signed up for Tuesday office hours to discuss ideas on simplifying this, but I'd prefer to not block this PR on that. Also, I think it's important that both #12725 and this PR go in the same release since migrating some resources to a different lock and not others might cause issues. Thanks!

```release-note:none

```


mmv1/products/accesscontextmanager/AccessLevelCondition.yaml
```diff
@@ -248,3 +249,10 @@ properties:
               description: 'CIDR block IP subnetwork specification. Must be IPv4.'
               item_type:
                 type: String
+  - name: 'accessPolicyId'
+    type: String
+    description: |
+      The name of the Access Policy this resource belongs to.
+    ignore_read: true
```
hao-nan-li, line 256: Why is this field `immutable` and `output true` at the same time? 

Can the user set the policyId from their TF config?

This comment also applies for all the cases below.

hao-nan-li: Thanks for the explanation. 
If `immutable` is set to true, if there's any modification to this field, another new resource will be recreated.  Is that something you'd expect?

hao-nan-li (approved):

PR #12737, opened 2025-01-13
Fix global_forwarding_rule labels

wiktorn:
- enable tests for global_forwarding_rule for `ga` provider
- add test for global_forwarding_rule for PSC target
- do not send labels for global_forwarding_rule for PSC target

Fixes: https://github.com/hashicorp/terraform-provider-google/issues/20873

```release-note:bug
compute: fix failure when creating `google_compute_global_forwarding_rule` with labels targeting PSC endpoint
```

### DIFF
--- a/mmv1/templates/terraform/pre_create/compute_global_forwarding_rule.go.tmpl
+++ b/mmv1/templates/terraform/pre_create/compute_global_forwarding_rule.go.tmpl
@@ -1,8 +1,7 @@
 // Labels cannot be set in a create for PSC forwarding rules, so remove it from the CREATE request.
-
-if targetProp != nil &&
-  (strings.Contains(targetProp.(string), "/serviceAttachments/") ||
-    targetProp.(string) == "all-apis" || targetProp.(string) == "vpc-sc") {
+if strings.Contains(targetProp.(string), "/serviceAttachments/") ||
+    targetProp.(string) == "all-apis" ||
+    targetProp.(string) == "vpc-sc" {
   if _, ok := obj["labels"]; ok {
     delete(obj, "labels")
   }
--- a/mmv1/templates/terraform/pre_create/compute_global_forwarding_rule.go.tmpl
+++ b/mmv1/templates/terraform/pre_create/compute_global_forwarding_rule.go.tmpl
@@ -1,0 +1,9 @@
+// Labels cannot be set in a create for PSC forwarding rules, so remove it from the CREATE request.
+
+if targetProp != nil &&
+  (strings.Contains(targetProp.(string), "/serviceAttachments/") ||
+    targetProp.(string) == "all-apis" || targetProp.(string) == "vpc-sc") {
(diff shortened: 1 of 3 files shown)


BBBmau (changes requested):

mmv1/third_party/terraform/services/compute/resource_compute_global_forwarding_rule_test.go.tmpl
```diff
@@ -473,7 +501,56 @@ resource "google_compute_url_map" "urlmap" {
 }
 `, fr, proxy, backend, hc, urlmap)
 }
-{{- end }}
+
+func testAccComputeGlobalForwardingRule_pscLabels(fr, address string) string {
```
BBBmau, line 505: can we separate this into two separate tests? one for `vpc-sc` and the other for `all-apis`

mmv1/templates/terraform/pre_create/compute_global_forwarding_rule.go.tmpl
```diff
@@ -0,0 +1,9 @@
+// Labels cannot be set in a create for PSC forwarding rules, so remove it from the CREATE request.
+
+if targetProp != nil &&
```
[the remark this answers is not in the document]
wiktorn: Ack on `targetProp != nil`.

Regarding PSC - one way to detect that this is a PSC target is to check that it has `serviceAttachment`, as the docs you link specify.

But there is also special case, [documented here](https://cloud.google.com/vpc/docs/private-service-connect-compatibility#google-apis-global) where instead of normal serviceAttachment you provide `vpc-sc` or `all-apis` value, which is also a Private Service Connect endpoint.

Condition  `strings.Contains(targetProp.(string), "/serviceAttachments/")  && targetProp.(string) == "all-apis"` is never true, so I can't change `||`  to `&&`.

In terms of why it is not possible to set the labels on forwarding rules targeting PSC attachements - I haven't found any reference in public docs, though the same was fixed in #17334 for `google_compute_forwarding_rule`

mmv1/third_party/terraform/services/compute/resource_compute_global_forwarding_rule_test.go.tmpl
wiktorn: Sure, but maybe it's better to have a separate `serviceAttachement` test case?

wiktorn:
@BBBmau updated the PR, leaving one comment open for you to close, if you're happy with my reply.

BBBmau (approved):
thanks for providing the explanation in your previous comment!

Approving and merging after all new tests passing as well as `forwardingRules_labels` test passing: https://hashicorp.teamcity.com/buildConfiguration/TerraformProviders_GoogleCloud_GOOGLE_MMUPSTREAMTESTS_GOOGLE_PACKAGE_COMPUTE/306442?buildTab=tests

image

PR #12740, opened 2025-01-13
Add data source for retrieving all organizations

bestefreund:
Adds a new data source "google_organizations", allowing to retrieve all organizations.

Fixes: https://github.com/hashicorp/terraform-provider-google/issues/20890

```release-note:new-datasource
`google_organizations`
```

### DIFF
--- a/mmv1/third_party/terraform/services/resourcemanager/data_source_google_organizations.go
+++ b/mmv1/third_party/terraform/services/resourcemanager/data_source_google_organizations.go
@@ -1,0 +1,107 @@
+// Copyright (c) HashiCorp, Inc.
+// SPDX-License-Identifier: MPL-2.0
+package resourcemanager
+
+import (
+	"context"
+	"fmt"
+	"path/filepath"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"github.com/hashicorp/terraform-provider-google/google/tpgresource"
+	transport_tpg "github.com/hashicorp/terraform-provider-google/google/transport"
+
+	"google.golang.org/api/cloudresourcemanager/v1"
+)
+
+func DataSourceGoogleOrganizations() *schema.Resource {
+	return &schema.Resource{
+		Read: datasourceGoogleOrganizationsRead,
+		Schema: map[string]*schema.Schema{
+			"filter": {
+				Optional: true,
+				Type:     schema.TypeString,
+			},
+			"organizations": {
+				Computed: true,
+				Type:     schema.TypeList,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"directory_customer_id": {
+							Computed: true,
+							Type:     schema.TypeString,
+						},
+						"display_name": {
+							Computed: true,
+							Type:     schema.TypeString,
+						},
+						"lifecycle_state": {
+							Computed: true,
+							Type:     schema.TypeString,
+						},
+						"name": {
+							Computed: true,
+							Type:     schema.TypeString,
+						},
+						"org_id": {
+							Computed: true,
+							Type:     schema.TypeString,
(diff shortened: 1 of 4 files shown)


mmv1/third_party/terraform/services/resourcemanager/data_source_google_organizations.go
```diff
@@ -0,0 +1,98 @@
...
+	}
+
+	organizations := make([]map[string]interface{}, 0)
+
+	filter := ""
+	if v, ok := d.GetOk("filter"); ok {
+		filter = v.(string)
+	}
+
+	organizationList, err := config.NewResourceManagerClient(userAgent).Organizations.Search(&cloudresourcemanager.SearchOrganizationsRequest{
```
SirGitsalot, line 72: This only retrieves the first page of results - you'll need to loop until `nextPageToken` is empty a la the [projects data source](https://github.com/hashicorp/terraform-provider-google/blob/be8a135f9f02d4836618e9b80eb72b9f0feed4d0/google/services/resourcemanager/data_source_google_projects.go#L78) to retrieve all readable-by-the-caller orgs.

SirGitsalot (changes requested):
One comment, hopefully easy to resolve. Thanks!

mmv1/third_party/terraform/services/resourcemanager/data_source_google_organizations.go
bestefreund: Hi @SirGitsalot,

thanks for the fast reply.

I did [two changes](https://github.com/GoogleCloudPlatform/magic-modules/compare/44df95c92410f6e97bc1ddf897322cff83cb8b36..c6da0f5e69dcfc471d3f114283edb989aad984c8):

1.  Replace generating a checksum with a human readable string for the ID. Instead the provided `filter` will be used and  if no `filter` has been provided a static value is used.
2. For pagination I used the [`Pages`-function from OrganizationsSearchCall](https://pkg.go.dev/google.golang.org/api@v0.214.0/cloudresourcemanager/v1#OrganizationsSearchCall.Pages) which handles pagination.

Please, have a look again.

Kind regards!

mmv1/third_party/terraform/services/resourcemanager/data_source_google_organizations.go
```diff
@@ -0,0 +1,102 @@
...
+	}
+
+	request := config.NewResourceManagerClient(userAgent).Organizations.Search(&cloudresourcemanager.SearchOrganizationsRequest{
+		Filter: filter,
+	})
+
+	err = request.Pages(context.Background(), func(organizationList *cloudresourcemanager.SearchOrganizationsResponse) error {
+		for _, organization := range organizationList.Organizations {
+			organizations = append(organizations, map[string]interface{}{
+				"directory_customer_id": organization.Owner.DirectoryCustomerId,
```
SirGitsalot, line 78: The API docs say that `Owner` is required, so it should never be `nil` here but for safety's sake we should check for `nil` before dereferencing it.

SirGitsalot (changes requested):
Sorry, one more nit. I kicked off the tests for this change to make sure it works with our CI organizations. Thanks again!

mmv1/third_party/terraform/services/resourcemanager/data_source_google_organizations.go
bestefreund: Hi @SirGitsalot,

no worries.
The better the review, the better the result.

I've just added the improvement and I hope I understood you right:
https://github.com/GoogleCloudPlatform/magic-modules/compare/c6da0f5e69dcfc471d3f114283edb989aad984c8..885b464bb967cc41cec87755283e17960c7e3969

Happy weekend!

SirGitsalot (approved):

PR #12746, opened 2025-01-13
Document import syntax for compute_project_metadata_item

chrisst:
This syntax was added in https://github.com/hashicorp/terraform-provider-google/pull/19613 but the documentation wasn't updated at that time.

```release-note:none
```


NickElliot (approved):
thanks for the update!

PR #12748, opened 2025-01-14
Promote Cloud Armor regional Security Policies Resources (Policy, Rules and Edge) from Beta to GA

cstanger:
Promote Cloud Armor regional Security Policies Resources (Policy, Rules and Edge) from Beta to GA

```release-note:note
compute:  promoted `NetworkEdgeSecurityService`, `RegionSecurityPolicy` and `RegionSecurityPolicyRules` from Beta to GA
```

### DIFF
--- a/mmv1/products/compute/NetworkEdgeSecurityService.yaml
+++ b/mmv1/products/compute/NetworkEdgeSecurityService.yaml
@@ -16,7 +16,6 @@ name: 'NetworkEdgeSecurityService'
 kind: 'compute#networkEdgeSecurityService'
 description: |
   Google Cloud Armor network edge security service resource.
-min_version: 'beta'
 references:
   guides:
     'Official Documentation': 'https://cloud.google.com/armor/docs/advanced-network-ddos'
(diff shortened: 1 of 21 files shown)


cstanger:
this is the required promotion for https://github.com/GoogleCloudPlatform/magic-modules/pull/12734

ScottSuarez (changes requested):
For these 3 tests we'll need to switch to using the `acctest.ProtoV5ProviderFactories` in place of the beta one. Test are failing due to this.

https://github.com/modular-magician/terraform-provider-google-beta/blob/eb132ccdea9c49b50e815d5aa8de2a7c525a4ed1/google-beta/services/compute/resource_compute_target_instance_generated_test.go#L161

ScottSuarez (changes requested):
So all resources in a test need to consistently use the same provider version. For these three tests we'll need to also remove google-beta from the other resources, in addition to the `acctest.ProtoV5ProviderFactories` change mentioned above. That is, if the reason we don't test them in GA is because of this feature or these resources alone.

ScottSuarez:
If you need help feel free to reach out over gvc! Happy to hop on a call

PR #12754, opened 2025-01-14
Cleaned up issue labeler

melinath:
This is some minor cleanup to the backfill code & tests. I think it might make sense to merge & see if it solves the failure. https://github.com/GoogleCloudPlatform/magic-modules/pull/12701 accidentally reversed the logic about including vs excluding PRs, which could have exposed the code to issues that were not previously encountered.

Switching to using github.Issue directly in tests allows them to be more explicit & flexible.

```release-note:none

```


shuyama1 (approved):

melinath:
Unfortunately that doesn't seem to have resolved the issue (according to my local testing) but I'll go ahead and merge this so follow-up PRs are more clear.

PR #12761, opened 2025-01-15
Memorystore Cluster_Disabled : Rename Standalone value in Mode field to Cluster_Disabled

arunsatyarth:
- This PR introduces the capability to create a memorystore instance in cluster_disabled mode.
- This is done by passing Mode=CLUSTER_DISABLED
- Previously the mode field was called 'standalone', however a product decision has been made to change it to cluster_disabled.
- This change is safe and not a breaking change because the feature is not launched and protected by an experiment which is disabled in Prod. Hence there are no users using Mode='Standalone' a.t.m and we can safely change it to ClusterDisabled.

```release-note:enhancement
memorystore: Add 'CLUSTER_DISABLED'  to `mode` flag  in  `google_memorystore_instance`
```


mmv1/products/memorystore/Instance.yaml
```diff
@@ -301,74 +301,77 @@ properties:
...
-            type: Enum
-            description:
-              "Output Only. Type of a PSC Connection.
-              \n Possible values:\n CONNECTION_TYPE_DISCOVERY \n CONNECTION_TYPE_PRIMARY \n CONNECTION_TYPE_READER"
-            output: true
-            enum_values:
-              - 'CONNECTION_TYPE_READER'
-              - 'CONNECTION_TYPE_PRIMARY'
-              - 'CONNECTION_TYPE_DISCOVERY'
+      type: NestedObject
```
ScottSuarez, line 304: Could you set up some time with me on GVC. This reads to me like a breaking change. We are changing existing field structure. I can't find this field documented externally on the [rest API](https://cloud.google.com/memorystore/docs/redis/reference/rest/v1beta1/projects.locations.instances).

ScottSuarez (changes requested):

mmv1/products/memorystore/Instance.yaml
arunsatyarth: Actually, this field is not used a.t.m. There is no change in the proto. The previous person who added this field in Terraform had made a mistake in exposing this, but because the feature is not launched yet(its behind mendel), nobody was affected.
Please see this field in the proto, its unchanged since when it was added.
https://source.corp.google.com/piper///depot/google3/google/cloud/memorystore/v1main/memorystore.proto;l=687

This is for a new feature. The docs for it will only be live on launch day.

ScottSuarez: ah got it! We'll need to hold off on merging until the feature is live. Thanks for explaining that.

If possible could we have a handwritten test that checks the presence of one or more of these fields. Modifying an existing test would also be okay.

ScottSuarez:
Checking to see on the status here. I can convert to draft so that I can review when it is ready?

ScottSuarez (approved):

PR #12769, opened 2025-01-15
Add Nightly Test Data Collector

shuyama1:
Part 1 of **Automated Nightly Test Failure Ticket Creation**

This PR adds the functionality to collect nightly test status data and upload the resulting status JSON file to a GCS bucket.

Configuration for the automated job will be added in a separate PR. 
The code for creating corresponding GH tickets based on the test status data will also be introduced in a separate PR.

```release-note:none

```


shuyama1:
Set up a cloud build trigger `test-failure-ticket` along with dedicated SAs, secrets and log bucket. Conducted tests in https://github.com/GoogleCloudPlatform/magic-modules/pull/12780

- [a successful default run with current time](https://pantheon.corp.google.com/cloud-build/builds;region=global/fb42f755-d2e0-40a0-afe2-4cb4ccf3fa46?e=-13802955&inv=1&invt=AbnG-Q&mods=logs_tg_staging&project=graphite-docker-images)
- [a successful run with a date specified](https://pantheon.corp.google.com/cloud-build/builds;region=global/38659455-f0df-4076-8ed5-162c5bc22d13?e=-13802955&inv=1&invt=AbnG-Q&mods=logs_tg_staging&project=graphite-docker-images)

.ci/magician/cmd/collect_nightly_test_status.go
```diff
@@ -0,0 +1,213 @@
...
+			}
+			env[ev] = val
+		}
+
+		tc := teamcity.NewClient(env["TEAMCITY_TOKEN"])
+		gcs := cloudstorage.NewClient()
+
+		date := time.Now()
+		customDate := args[0]
+		//
```
melinath, line 81:
```suggestion
```

.ci/magician/cmd/collect_nightly_test_status.go
```diff
@@ -0,0 +1,213 @@
...
+	},
+}
+
+func listCNTSRequiredEnvironmentVariables() string {
+	var result string
+	for i, ev := range cntsRequiredEnvironmentVariables {
+		result += fmt.Sprintf("\t%2d. %s\n", i+1, ev)
+	}
+	return result
+}
```
melinath, line 101: This makes sense for debugging but we probably don't want to put this in the logs long-term.

```suggestion
```

.ci/magician/cmd/collect_nightly_test_status.go
```diff
@@ -0,0 +1,213 @@
...
+	for _, build := range builds.Builds {
+		serviceTestResults, err := tc.GetTestResults(build)
+		if err != nil {
+			return fmt.Errorf("failed to get test results: %v", err)
+		}
+		if len(serviceTestResults.TestResults) == 0 {
+			continue
+		}
+		// Get service package name
+		serviceName, err := convertServiceName(build.BuildTypeId)
```
melinath, line 139: It looks like this will result in SERVICE_SWEEPER being counted as a package called SWEEPER... it might be nice to special case that.

.ci/magician/cmd/collect_nightly_test_status.go
```diff
@@ -0,0 +1,213 @@
...
+	idx := strings.LastIndex(servicePath, "_")
+
+	if idx != -1 {
+		return strings.ToLower(servicePath[idx+1:]), nil
+	}
+	return "", fmt.Errorf("wrong service path format for %s", servicePath)
+}
+
+// convertErrorMessage returns concise error message
+func convertErrorMessage(rawErrorMessage string) string {
```
melinath, line 193: Could you add some tests for this?

In particular I think that if startMarker is not found it might cause errors.

melinath (approved):

.ci/magician/cmd/collect_nightly_test_status.go
shuyama1: yeah, it makes sense. Currently, sweeper is skipped reply on the previous check `len(serviceTestResults.TestResults) == 0` but after the refactor of sweepers, it may not be working anymore. I agree it's nice to explicit case it.

.ci/magician/cmd/collect_nightly_test_status.go
shuyama1: Good catch! I've modified the logic and added unit tests. Thanks!

.ci/magician/cmd/collect_nightly_test_status.go
shuyama1: removed.

.ci/magician/cmd/collect_nightly_test_status.go
shuyama1: updated

.ci/magician/cmd/collect_nightly_test_status.go
shuyama1: Added an explicit check for sweeper

.ci/magician/cmd/collect_nightly_test_status.go
melinath: I totally misread this as logging the value of the env var, not the name 🤦 please feel free to add it back, sorry about that!

melinath (approved):

.ci/magician/cmd/collect_nightly_test_status.go
shuyama1: lol, I thought you mean we don't need to log the names as well, which, I thought, also makes sense. I'll add it back though

melinath (approved):

PR #12774, opened 2025-01-16
add `google_compute_storage_pool` resource

translucens:
Fixes https://github.com/hashicorp/terraform-provider-google/issues/20434

```release-note:new-resource
`google_compute_storage_pool`
```

### DIFF
--- a/mmv1/products/compute/StoragePool.yaml
+++ b/mmv1/products/compute/StoragePool.yaml
@@ -37,6 +37,8 @@ iam_policy:
   method_name_separator: "/"
   import_format:
     - "projects/{{project}}/zones/{{zone}}/storagePools/{{name}}"
+  iam_conditions_request_type: "QUERY_PARAM"
+  allowed_iam_role: "roles/compute.viewer"
 examples:
   - name: "compute_storage_pool_basic"
     primary_resource_id: "test-storage-pool-basic"
@@ -128,48 +130,58 @@ properties:
         type: String
         description: |
           Timestamp of the last successful resize in RFC3339 text format.
+        output: true
       - name: "diskCount"
         type: String
         description: |
           Number of disks used.
+        output: true
       - name: "poolUsedCapacityBytes"
         type: String
         description: |
           Space used by data stored in disks within the storage pool (in bytes).
           This will reflect the total number of bytes written to the disks in the pool,
           in contrast to the capacity of those disks.
+        output: true
       - name: "poolUserWrittenBytes"
         type: String
         description: |
           Amount of data written into the pool, before it is compacted.
+        output: true
       - name: "totalProvisionedDiskCapacityGb"
         type: String
         description: |
           Sum of all the capacity provisioned in disks in this storage pool.
           A disk's provisioned capacity is the same as its total capacity.
+        output: true
       - name: "maxTotalProvisionedDiskCapacityGb"
         type: String
         description: |
           Maximum allowed aggregate disk size in gigabytes.
+        output: true
       - name: "poolUsedIops"
         type: String
         description: |
           Sum of all the disks' provisioned IOPS, minus some amount
           that is allowed per disk that is not counted towards pool's IOPS capacity.
           For more information, see https://cloud.google.com/compute/docs/disks/storage-pools.
+        output: true
       - name: "totalProvisionedDiskIops"
         type: String
         description: |
           Sum of all the disks' provisioned IOPS.
+        output: true
       - name: "poolUsedThroughput"
         type: String
         description: |
           Sum of all the disks' provisioned throughput in MB/s.
+        output: true
       - name: "totalProvisionedDiskThroughput"
         type: String
         description: |
           Sum of all the disks' provisioned throughput in MB/s,
           minus some amount that is allowed per disk that is not counted towards pool's throughput capacity.
+        output: true
   - name: "storagePoolType"
     type: String
     description: |
@@ -186,45 +198,55 @@ properties:
         type: String
         description: |
           Timestamp of the last successful resize in RFC3339 text format.
+        output: true
       - name: "diskCount"
         type: String
         description: |
           Number of disks used.
+        output: true
       - name: "poolUsedCapacityBytes"
         type: String
         description: |
           Space used by data stored in disks within the storage pool (in bytes).
           This will reflect the total number of bytes written to the disks in the pool, in contrast to the capacity of those disks.
+        output: true
       - name: "poolUserWrittenBytes"
         type: String
         description: |
           Amount of data written into the pool, before it is compacted.
+        output: true
       - name: "totalProvisionedDiskCapacityGb"
         type: String
         description: |
           Sum of all the capacity provisioned in disks in this storage pool.
           A disk's provisioned capacity is the same as its total capacity.
+        output: true
       - name: "maxTotalProvisionedDiskCapacityGb"
         type: String
         description: |
           Maximum allowed aggregate disk size in gigabytes.
+        output: true
       - name: "poolUsedIops"
         type: String
         description: |
           Sum of all the disks' provisioned IOPS, minus some amount that is allowed per disk that is not counted towards pool's IOPS capacity. For more information, see https://cloud.google.com/compute/docs/disks/storage-pools.
+        output: true
       - name: "totalProvisionedDiskIops"
         type: String
         description: |
           Sum of all the disks' provisioned IOPS.
+        output: true
       - name: "poolUsedThroughput"
         type: String
         description: |
           Sum of all the disks' provisioned throughput in MB/s.
+        output: true
       - name: "totalProvisionedDiskThroughput"
         type: String
         description: |
           Sum of all the disks' provisioned throughput in MB/s,
           minus some amount that is allowed per disk that is not counted towards pool's throughput capacity.
+        output: true
   - name: "capacityProvisioningType"
     type: Enum
     description: |
--- a/mmv1/products/compute/StoragePool.yaml
+++ b/mmv1/products/compute/StoragePool.yaml
@@ -257,7 +257,7 @@ properties:
     enum_values:
       - "STANDARD"
       - "ADVANCED"
-    default_value: "ADVANCED"
+    default_from_api: true
   - name: "performanceProvisioningType"
     type: Enum
     description: |
@@ -267,7 +267,7 @@ properties:
     enum_values:
       - "STANDARD"
       - "ADVANCED"
-    default_value: "ADVANCED"
+    default_from_api: true
 virtual_fields:
   - name: "deletion_protection"
     type: Boolean
--- a/mmv1/products/compute/StoragePool.yaml
+++ b/mmv1/products/compute/StoragePool.yaml
@@ -63,7 +63,9 @@ parameters:
     required: false
     default_from_api: true
     custom_expand: "templates/terraform/custom_expand/resourceref_with_validation.go.tmpl"
+    custom_flatten: "templates/terraform/custom_flatten/name_from_self_link.tmpl"
     resource: "Zone"
+    imports: "name"
 properties:
   - name: "kind"
     type: String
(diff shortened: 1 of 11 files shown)


mmv1/third_party/terraform/services/compute/resource_compute_storage_pool_test.go
```diff
@@ -0,0 +1,74 @@
...
+
+func testAccComputeStoragePool_computeStoragePool_update(context map[string]interface{}) string {
+	return acctest.Nprintf(`
+resource "google_compute_storage_pool" "test-storage-pool-full" {
+  name = "tf-test-storage-pool-full%{random_suffix}"
+
+  description = "Hyperdisk Balanced storage pool"
+
+  capacity_provisioning_type   = "STANDARD"
+  pool_provisioned_capacity_gb = "11264"
```
ScottSuarez, line 59: Can we lower the capacity on the tests tests. Looks like we are exceeding quota on out projects

ScottSuarez (changes requested):

mmv1/third_party/terraform/services/compute/resource_compute_storage_pool_test.go
translucens: Hyperdisk storage pools requires minimal initial capacity 10 TiB + 1 TiB increment. The test capacity is smallest size. 

```
Minimum storage pool provisioned capacity	10 TiB
Storage capacity increments	multiples of 1 TiB
```
https://cloud.google.com/compute/docs/disks/storage-pools#hdbsp-limits

ScottSuarez: Okay I put a quota request in. As is these few tests will outstrip current quota

mwarkentin:
Hi, just wondering if the quota request has gone through, eagerly awaiting this functionality. :)

ScottSuarez:
Hi @mwarkentin, let me follow up internally. Seems things have gotten stuck regarding the request. I need to find the correct people to escalate. @translucens, are you on the team in question? Would you be able to assist in granting quota? Feel free to message me out of band.

ScottSuarez:
update, escalated to the right people, waiting to hear back

translucens:
The test requires 71,680 GiB (10 TiB x 7 test cases) of quota `hyperdisk_throughput_storage_pools_advanced_capacity` in us-central1 region.

ScottSuarez:
> The test requires 71,680 GiB (10 TiB x 7 test cases) of quota `hyperdisk_throughput_storage_pools_advanced_capacity` in us-central1 region.

To reduce the amount of quota needed and unblock this PR can you disable auto-generated tests and IAM tests.

For the hand written tests we can make them run linearly with a test suite pattern. So all the current tests are nested under.
https://github.com/ScottSuarez/magic-modules/blob/2346615efcf8e86f961bf6b92987f574682e11fb/mmv1/third_party/terraform/services/tags/resource_tags_test.go#L20-L48

ScottSuarez (commented):
waiting on quota changes

mmv1/products/compute/StoragePool.yaml
ScottSuarez, line 1: Okay so I talked to the owning team. They had concerns about the implementation and asked for deletion protection to be added.

1. You can look at spanner database's implementation of the virtual field for guidance.
https://github.com/ScottSuarez/magic-modules/blob/9d2cc39d0d68ac686996de0bc9a2f039ea62815e/mmv1/products/spanner/Database.yaml#L73-L84

2. You'll need to ignore read on this field in tests (since it's virtual)
https://github.com/ScottSuarez/magic-modules/blob/9d2cc39d0d68ac686996de0bc9a2f039ea62815e/mmv1/products/spanner/Database.yaml#L69-L71

3. And ensure the tests have deletion protection set to false so the infrastructure can be destroyed.

4.Add a pre-delete function that uses this vitural field.
https://github.com/ScottSuarez/magic-modules/blob/9d2cc39d0d68ac686996de0bc9a2f039ea62815e/mmv1/products/spanner/Database.yaml#L59-L60

ScottSuarez (changes requested):
Changes requested by owning team. They said they'll be in touch next week as well. Thanks for your patience here!

mmv1/products/compute/StoragePool.yaml
ScottSuarez: ah perfect ! Thanks !! :) 

You're ahead of me

ScottSuarez:
Tests are failing due to a change outside of this PR. Working to get the upstream fixed.

mwarkentin:
@ScottSuarez looks like `main` is clean again, any chance this could be retried? 

https://github.com/GoogleCloudPlatform/magic-modules/commits/main/

mmv1/templates/terraform/examples/compute_storage_pool_basic.tf.tmpl
```diff
@@ -0,0 +1,13 @@
+resource "google_compute_storage_pool" "{{$.PrimaryResourceId}}" {
+  name = "{{index $.Vars "storage_pool_name"}}"
+
+  pool_provisioned_capacity_gb = "10240"
+
+  pool_provisioned_throughput = 100
+
+  storage_pool_type = "https://www.googleapis.com/compute/v1/projects/${data.google_project.project.project_id}/zones/us-central1-a/storagePoolTypes/hyperdisk-throughput"
```
ScottSuarez, line 8:
```suggestion
  storage_pool_type = "https://www.googleapis.com/compute/beta/projects/${data.google_project.project.project_id}/zones/us-central1-a/storagePoolTypes/hyperdisk-throughput"
```

It appears that v1 is being sent in the api request for `storage_pool_type` but beta is returned.

this field may need to be version dependent? I am not sure if there is president for it. If we can get the storage pool type from a datasource that could be a solution.

ScottSuarez (changes requested):

mmv1/templates/terraform/examples/compute_storage_pool_basic.tf.tmpl
translucens: Adding storagePoolTypes datsource seems reasonable way. I'll add it.

ScottSuarez: can you tag me on the change. I'll prioritize review

cngkaygusuz:
Hey @mwarkentin @translucens, Cengiz here, I'm the person who wrote a vast majority of the storage pool API's and will help yall as the subject matter expert.

Scott probably have told you this before: the `deletion_protection` flag was the point of contention in the past when we looked into this, the advisory we received was that the default value of this flag should be consistent across all the resources (e.g., it can't be true by default for storage pools and false for disks). This time, I personally went after whether if this is the case and indeed this isn't a flag that needs to be consistent all across the board.

I just manually tested the proposed changes and overall it looks good -- `terraform apply` refuses to execute, with a good error message. 

There are a few details I'd like to be sure they're straight before we check this in.

First is the storage pool type field. I think the latest agreement was that this was also going to be a resource reference rather than just a string. This will probably address the oddities I saw, so I'm going to hold off saying anything else until that's in place.

Second, capacity/performance provisioning type defaults. These are coded in the server and the client should delegate it to the source of truth, otherwise this makes change management of this kind of painful for reasons I believe are obvious.

Third, the dependency from disks to storage pools. For a storage pool to be deleted, all the disks in it also needs to be deleted explicitly as well; we intentionally don't do cascading deletions, but I believe such a "feature" is consistent with terraform's planner. Currently, a disk refers to a storage pool simply by its URL, its obviously not modeled as a resource. I'm not sure whats the effort or feasibility of converting that to a storage pool reference but its definitely worth the effort here.

For the test/quota issues, this is indeed something I can help by raising test project quotas but TBH this isn't something I want to do. I think the quota of the project is 20 TiB which should be enough to test updates. To run multiple tests however, I think its best to clean up the resources created by the previous test before the next one runs.

ScottSuarez:
@cngkaygusuz 
> For the test/quota issues, this is indeed something I can help by raising test project quotas but TBH this isn't something I want to do. I think the quota of the project is 20 TiB which should be enough to test updates. To run multiple tests however, I think its best to clean up the resources created by the previous test before the next one runs.

We have limitations on the engine right now where not all of these tests can be run in series. We will likely need some sort of quota increase to be able to handle this.

ScottSuarez:
Looks like there is one failure on the test run 

```
    iam_compute_storage_pool_test.go:114: Step 5/19 error: Error running apply: exit status 1
        
        Error: Error applying IAM policy for compute storagepool "projects/ci-test-project-188019/zones/us-central1-a/storagePools/tf-test-storage-pool-throughputwnb3kxoaic": Error setting IAM policy for compute storagepool "projects/ci-test-project-188019/zones/us-central1-a/storagePools/tf-test-storage-pool-throughputwnb3kxoaic": Post "https://compute.googleapis.com/compute/beta/projects/xxxx/zones/us-central1-a/storagePools/tf-test-storage-pool-throughputwnb3kxoaic/setIamPolicy?alt=json": Requested interaction not found
        
          with google_compute_storage_pool_iam_binding.foo3,
          on terraform_plugin_test.tf line 48, in resource "google_compute_storage_pool_iam_binding" "foo3":
          48: resource "google_compute_storage_pool_iam_binding" "foo3" {
        
```

mmv1/products/compute/StoragePoolType.yaml
```diff
@@ -0,0 +1,155 @@
...
+  Represents a StoragePoolType data source.
+  The type of Hyperdisk Storage Pool that you create determines the type of disks that you can create in the storage pool.
+references:
+  guides:
+    'Types of Hyperdisk Storage Pools': 'https://cloud.google.com/compute/docs/disks/storage-pools#sp-types'
+  api: 'https://cloud.google.com/compute/docs/reference/rest/v1/storagePoolTypes'
+base_url: 'projects/{{project}}/zones/{{zone}}/storagePoolTypes'
+self_link: 'projects/{{project}}/zones/{{zone}}/storagePoolTypes/{{storagePoolType}}'
+immutable: true
+exclude: true
```
ScottSuarez, line 27: this file shouldn't be necessary since we are excluding it from generation.

translucens:
I tested few times with VCR recording mode and there are neither `capacityProvisioningType` nor `performanceProvisioningType` fields in VCR log and audit log.

Storage type URL issue could be solved using `google_compute_storage_pool_types` data source.

VCR log
```json
{"name":"tf-test-storage-pool-basicybr9hzhl6l","poolProvisionedCapacityGb":"10240","poolProvisionedThroughput":"100","storagePoolType":"https://www.googleapis.com/compute/beta/projects/hellomm/zones/us-central1-a/storagePoolTypes/hyperdisk-throughput"}
```

Audit log from Cloud Logging
```
request: {
@type: "type.googleapis.com/compute.storagePools.insert"
name: "tf-test-storage-pool-basic-abcdef"
poolProvisionedCapacityGb: "10240"
poolProvisionedThroughput: "100"
storagePoolType: "https://www.googleapis.com/compute/beta/projects/hellomm/zones/us-central1-a/storagePoolTypes/hyperdisk-throughput"
}
```

Terraform config
```hcl
resource "google_compute_storage_pool" "test-storage-pool-basic" {
  provider = google-beta
  name = "tf-test-storage-pool-basic-abcdef"
  pool_provisioned_capacity_gb = "10240"
  pool_provisioned_throughput = 100
  storage_pool_type = data.google_compute_storage_pool_types.throughput.self_link
  deletion_protection = false
}

data "google_compute_storage_pool_types" "throughput" {
  provider = google-beta

  zone              = "us-central1-a"
  storage_pool_type = "hyperdisk-throughput"
}

data "google_project" "project" {}
```

cngkaygusuz:
@translucens 

Looks like I missed the very last step while building the local binary, it was using the March 6 version 🤦 

Can confirm provisioning types are resolved.

The brief type specification would be the last thing I would insist on. That is the most intuitive form of the types, and as mentioned earlier disks are able to handle it, I imagine it would involve a small formatting function somewhere.

mmv1/products/compute/StoragePoolType.yaml
translucens: I added `ResourceRef` to `StoragePoolType` so I leave this file as is.

ScottSuarez:
@cngkaygusuz could you take another look when you have a moment. Want to ensure your approval here!

dongyingbo:
Hi @ScottSuarez and @cngkaygusuz,
I wonder if there is any chance we are able to accelerate it please?

Thanks!

cngkaygusuz:
This looks good now. Consider this PR approved.

ScottSuarez (approved):

ScottSuarez:
This needs to be added under management of the storage_pool team before it can be merged.

b/409595095

[go/fix-missing-service-labels](https://goto.google.com/fix-missing-service-labels)

Once this is complete then we are finally able to merge.

cngkaygusuz:
@translucens I believe the service label issue should now be fixed, can you run the checks again?

PR #12775, opened 2025-01-16
Add state reason field.

cynicphoenix:
```release-note:enhancement
compute: added `reasons` fields to `Spoke` resource
```

Add StateReason array of enums.


mmv1/products/networkconnectivity/Spoke.yaml
```diff
@@ -319,3 +319,23 @@ properties:
       - 'CREATING'
       - 'ACTIVE'
       - 'DELETING'
+  - name: 'reasons'
+    type: NestedObject
+    description: "Reason why the spoke is inactive."
+    output: true
```
hao-nan-li, line 325: Is there any particular reason to add an output only field here?

PR #12776, opened 2025-01-16
Add google_chronicle_retrohunt resource to chronicle

ankitgoyal0301:

```release-note:new-resource
`google_chronicle_retrohunt`
```

### DIFF
--- a/mmv1/products/chronicle/Retrohunt.yaml
+++ b/mmv1/products/chronicle/Retrohunt.yaml
@@ -1,0 +1,153 @@
+# Copyright 2024 Google Inc.
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+---
+name: Retrohunt
+description: Retrohunt is an execution of a Rule over a time range in the past.
+references:
+  guides:
+    'Google SecOps Guides': 'https://cloud.google.com/chronicle/docs/secops/secops-overview'
+  api: 'https://cloud.google.com/chronicle/docs/reference/rest/v1alpha/projects.locations.instances.rules.retrohunts'
+base_url: projects/{{project}}/locations/{{location}}/instances/{{instance}}/rules/{{rule}}/retrohunts
+immutable: true
+self_link: projects/{{project}}/locations/{{location}}/instances/{{instance}}/rules/{{rule}}/retrohunts/{{retrohunt}}
+create_url: projects/{{project}}/locations/{{location}}/instances/{{instance}}/rules/{{rule}}/retrohunts
+id_format: projects/{{project}}/locations/{{location}}/instances/{{instance}}/rules/{{rule}}/retrohunts/{{retrohunt}}
+import_format:
+- projects/{{project}}/locations/{{location}}/instances/{{instance}}/rules/{{rule}}/retrohunts/{{retrohunt}}
+exclude_delete: true
+
+custom_code:
+  decoder: 'templates/terraform/decoders/chronicle_retrohunt.go.tmpl'
+  post_create: 'templates/terraform/post_create/chronicle_retrohunt_id.go.tmpl'
+examples:
+- name: 'chronicle_retrohunt_basic'
+  primary_resource_id: example
+  min_version: 'beta'
+  exclude_import_test: true
+  vars:
+    start_time: '2025-01-01T00:00:00Z'
+    end_time: '2025-01-01T12:00:00Z'
+  test_env_vars:
+    chronicle_id: 'CHRONICLE_ID'
+  test_vars_overrides:
+    start_time: 'time.Now().Add(time.Hour * (-12)).Format(time.RFC3339)'
+    end_time: 'time.Now().Add(time.Hour * (-1)).Format(time.RFC3339)'
+
+async:
+  actions: [create]
+  type: OpAsync
+  operation:
+    full_url: 'https://{{location}}-chronicle.googleapis.com/v1beta/{{op_id}}'
+    timeouts:
+      insert_minutes: 20
+      update_minutes: 20
+      delete_minutes: 20
+  result:
+    resource_inside_response: true
+exclude_sweeper: true
+parameters:
+- name: location
+  type: String
+  description: The location of the resource. This is the geographical region where the Chronicle instance resides, such as "us" or "europe-west2".
+  immutable: true
+  url_param_only: true
+  required: true
+- name: instance
+  type: String
+  description: The unique identifier for the Chronicle instance, which is the same as the customer ID.
+  immutable: true
+  url_param_only: true
+  required: true
+- name: rule
+  type: String
+  description: The Rule ID of the rule. 
+  immutable: true
+  url_param_only: true
+  required: true
+- name: retrohunt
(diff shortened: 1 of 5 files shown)


ronakgadia (approved):

mmv1/products/chronicle/Retrohunt.yaml
```diff
@@ -0,0 +1,154 @@
...
+      update_minutes: 20
+      delete_minutes: 20
+  result:
+    resource_inside_response: true
+exclude_sweeper: true
+parameters:
+  - name: location
+    type: String
+    description: The location of the resource. This is the geographical region where the Chronicle instance resides, such as "us" or "europe-west2".
+    immutable: true
```
c2thorn, line 64: can we remove immutable on all fields, since the resource itself is already immutable. Just for readability

mmv1/products/chronicle/Retrohunt.yaml
```diff
@@ -0,0 +1,154 @@
...
+    'Google SecOps Guides': 'https://cloud.google.com/chronicle/docs/secops/secops-overview'
+  api: 'https://cloud.google.com/chronicle/docs/reference/rest/v1alpha/projects.locations.instances.rules.retrohunts'
+base_url: projects/{{project}}/locations/{{location}}/instances/{{instance}}/rules/{{rule}}/retrohunts
+immutable: true
+self_link: projects/{{project}}/locations/{{location}}/instances/{{instance}}/rules/{{rule}}/retrohunts/{{retrohunt}}
+create_url: projects/{{project}}/locations/{{location}}/instances/{{instance}}/rules/{{rule}}/retrohunts
+id_format: projects/{{project}}/locations/{{location}}/instances/{{instance}}/rules/{{rule}}/retrohunts/{{retrohunt}}
+import_format:
+  - projects/{{project}}/locations/{{location}}/instances/{{instance}}/rules/{{rule}}/retrohunts/{{retrohunt}}
+exclude_delete: true
```
c2thorn, line 28: since this cannot be deleted on its own, will retrohunts be automatically cleaned up when the parent rule is deleted? Asking mainly for our own test environment

mmv1/products/chronicle/Retrohunt.yaml
```diff
@@ -0,0 +1,154 @@
...
+  - name: rule
+    type: String
+    description: The Rule ID of the rule.
+    immutable: true
+    url_param_only: true
+    required: true
+  - name: retrohunt
+    type: String
+    description: The retrohunt ID of the Retrohunt. A retrohunt is an execution of a Rule over a time range in the past.
+    required: false
```
c2thorn, line 82:
```suggestion
```
more readability, `required: false` is not needed

mmv1/products/chronicle/Retrohunt.yaml
```diff
@@ -0,0 +1,154 @@
...
+
+custom_code:
+  decoder: 'templates/terraform/decoders/chronicle_retrohunt.go.tmpl'
+  post_create: 'templates/terraform/post_create/chronicle_retrohunt_id.go.tmpl'
+examples:
+  - name: 'chronicle_retrohunt_basic'
+    primary_resource_id: example
+    min_version: 'beta'
+    exclude_import_test: true
+    skip_vcr: true
```
c2thorn, line 38: why is this skipped in VCR?

mmv1/third_party/terraform/services/chronicle/chronicle_operation.go
```diff
@@ -0,0 +1,72 @@
...
+
+func (w *ChronicleOperationWaiter) QueryOp() (interface{}, error) {
+	if w == nil {
+		return nil, fmt.Errorf("Cannot query operation, it's unset or nil.")
+	}
+
+	region := tpgresource.GetRegionFromRegionalSelfLink(w.CommonOperationWaiter.Op.Name)
+
+	// Returns the proper get.
+	url := fmt.Sprintf("https://%s-chronicle.googleapis.com/v1alpha/%s", region, w.CommonOperationWaiter.Op.Name)
```
c2thorn, line 27: why are we using v1alpha here instead of v1beta? Alpha APIs are not typically allowed in the public providers

mmv1/third_party/terraform/services/chronicle/chronicle_operation.go
```diff
@@ -0,0 +1,72 @@
+package chronicle
```
c2thorn, line 1: Why do we need a handwritten operation instead of a generated one?

c2thorn (changes requested):
If we are skipping in VCR, we'll need to have the tests pass on a manual run in our environment. I'll trigger that on the next pass after we've gone through some of my questions

mmv1/products/chronicle/Retrohunt.yaml
ankitgoyal0301: Yes, that's correct. So, I have added a virtual field [deletion_policy](https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/products/chronicle/Rule.yaml#L62-L72) to the rules resource specifically for this purpose, which if provided as `deletion_policy: FORCE`, it automatically deletes any child resources too associated to it. In our create test for retrohunt, I have already provided the same [here](https://github.com/GoogleCloudPlatform/magic-modules/pull/12776/files#diff-276d3c334aa459a6fae4896e724d54081104d6ea3cd4b1e27414050ae7434690R5).

mmv1/third_party/terraform/services/chronicle/chronicle_operation.go
ankitgoyal0301: My bad, It is supposed to be v1beta only! Actually the v1beta change got rolled out to prod this thursday only, and I have been testing and working locally with the v1alpha version. Missed updating this part. Will update it! Thanks!

mmv1/third_party/terraform/services/chronicle/chronicle_operation.go
ankitgoyal0301: So the reason we need the handwritten operation is because the generated one isn't taking care of the location parameter in of LROs. This is how we require the location param in the API:

```
https://{{location}}-chronicle.googleapis.com/v1beta/{{op_id}}
```

I think this [YAQs](https://yaqs.corp.google.com/eng/q/7324679282918162432#a2) can give the full context.

mmv1/products/chronicle/Retrohunt.yaml
ankitgoyal0301: I used the `skip_vcr` because the time element will prevent replays from working properly. That's my understanding. I would love to know your thoughts on this.

mmv1/products/chronicle/Retrohunt.yaml
ankitgoyal0301: Done!

mmv1/products/chronicle/Retrohunt.yaml
ankitgoyal0301: Done!

mmv1/third_party/terraform/services/chronicle/chronicle_operation.go
ankitgoyal0301: Done!

mmv1/third_party/terraform/services/chronicle/chronicle_operation.go
c2thorn: I see, that makes sense.

mmv1/products/chronicle/Retrohunt.yaml
c2thorn: makes sense

mmv1/products/chronicle/Retrohunt.yaml
c2thorn: sounds good

c2thorn:
running test run of `TestAccChronicleRetrohunt` tests

c2thorn (approved):
test run in teamcity passed, LGTM

PR #12777, opened 2025-01-16
Fixed cluster resource limits maximum to be marked as required in docs.

mtrqq:
This PR patches parameter [cluster. cluster_autoscaling.resource_limits.maximum](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/container_cluster#maximum-1) to be marked as required and sets a minimum value for it.

Current docs indicate that users are able to configure resource limits like so:

```terraform
resource_limits {
    resource_type = "cpu"
}
```

This is not a valid configuration and it's getting rejected while attempting to create a cluster with the following error:

`Error: googleapi: Error 400: Resource_limit.maximum must be greater than 0.`

```release-note:bug
container: marked `cluster_autoscaling.resource_limits.maximum` as required as requests would fail if it was not set
```

### DIFF
--- a/mmv1/third_party/terraform/services/container/resource_container_cluster.go.tmpl
+++ b/mmv1/third_party/terraform/services/container/resource_container_cluster.go.tmpl
@@ -620,9 +620,10 @@ func ResourceContainerCluster() *schema.Resource {
 										Description: `Minimum amount of the resource in the cluster.`,
 									},
 									"maximum": {
-										Type:        schema.TypeInt,
-										Optional:    true,
-										Description: `Maximum amount of the resource in the cluster.`,
+										Type:         schema.TypeInt,
+										Description:  `Maximum amount of the resource in the cluster.`,
+										Required:     true,
+										ValidateFunc: validation.IntAtLeast(1),
 									},
 								},
 							},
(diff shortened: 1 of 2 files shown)


mtrqq:
IMO - this not a breaking change, as there's 0 changes from the API perspective, it's more about changing customer's expectation from the API which were incorrect beforehand

apeabody:
> IMO - this not a breaking change, as there's 0 changes from the API perspective, it's more about changing customer's expectation from the API which were incorrect beforehand

Agreed - Currently with the Terraform validation would succeed, but upon apply it would be blocked by the API, so it isn't a currently working configuration.

slevenick:
Looks like we have an exception for things like this: https://googlecloudplatform.github.io/magic-modules/breaking-changes/make-a-breaking-change/#in-minor-releases

So it should be fine

slevenick (changes requested):
You'll need to add this change to the docs as well. https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/third_party/terraform/website/docs/r/container_cluster.html.markdown

mtrqq:
> You'll need to add this change to the docs as well. https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/third_party/terraform/website/docs/r/container_cluster.html.markdown

Thanks! Added a patch to docs.

slevenick (approved):

PR #12784, opened 2025-01-16
Support API service agents in `BootstrapPSARole()` utils

wyardley:
Some service agents are in the form
`project-number@cloudservices.gserviceaccount.com`

Support this via a hacky solution where if `@` is present, the interpolation of service accounts will be more limited. Thus either:
`acctest.BootstrapPSARole(t, projectNumber, "@cloudservices", "roles/foo.bar")` (ab)using prefix, or
`acctest.BootstrapPSARole(t, "", "NNNNNNNNNNN@cloudservices", "roles/foo.bar")`

can be used for `NNNNNNNNNNN@cloudservices.gserviceaccount.com`, for example.
https://cloud.google.com/iam/docs/service-account-types#service-agents

```release-note:none

```

### DIFF
--- a/mmv1/third_party/terraform/acctest/bootstrap_iam_test_utils.go
+++ b/mmv1/third_party/terraform/acctest/bootstrap_iam_test_utils.go
@@ -3,6 +3,7 @@ package acctest
 import (
 	"fmt"
 	"log"
+	"strings"
 	"testing"
 
 	"github.com/hashicorp/terraform-provider-google/google/envvar"
@@ -14,6 +15,9 @@ import (
 // policy grants the given service agents the given roles.
 // prefix is usually "service-" and indicates the service agent should have the
 // given prefix before the project number.
+// If an address with an `@` (x@y) is passed, the address will be used
+// verbatim, but still prefixed with `serviceAccount:prefix` and suffixed with
+// `gserviceaccount.com`
 // This is important to bootstrap because using iam policy resources means that
 // deleting them removes permissions for concurrent tests.
 // Return whether the bindings changed.
@@ -38,7 +42,13 @@ func BootstrapAllPSARoles(t *testing.T, prefix string, agentNames, roles []strin
 
 	members := make([]string, len(agentNames))
 	for i, agentName := range agentNames {
-		members[i] = fmt.Sprintf("serviceAccount:%s%d@%s.iam.gserviceaccount.com", prefix, project.ProjectNumber, agentName)
+
+		// Allow partially bypassing the builtin logic of making up the email if an @ is included
+		if strings.Contains(agentName, "@") {
+			members[i] = fmt.Sprintf("serviceAccount:%s%s.gserviceaccount.com", prefix, agentName)
+		} else {
+			members[i] = fmt.Sprintf("serviceAccount:%s%d@%s.iam.gserviceaccount.com", prefix, project.ProjectNumber, agentName)
+		}
 	}
 
 	// Create the bindings we need to add to the policy.


wyardley:
https://github.com/GoogleCloudPlatform/magic-modules/pull/12728 has more context on the reasoning behind this; totally get that it's hacky, and happy to make adjustments if someone has another idea.

cc: @ScottSuarez @melinath 

My thinking is that the history would be cleaner if this can be merged on its own, though we can close this if #12728 comes through first.

Side note: there are some other bits like:
https://github.com/GoogleCloudPlatform/magic-modules/blob/b5bdbf8875474068e54a77780dc9e53e0ddfbc0d/mmv1/third_party/terraform/services/container/resource_container_cluster_test.go.tmpl#L9930-L9942
that might make some sense to move to the bootstrap pattern once this is supported.

ScottSuarez:
I am okay with this pattern @melinath, any resistance or thoughts?

wyardley:
I guess an alternative would be a set of `BootstrapGSA()` wrapper functions similar to the PSA ones?  Could still leave that big function more or less as is, but make the wrapper function interface simpler and the way of calling it less ugly.

melinath:
Overall this seems like a reasonable thing to support in the bootstrap helper functions. I might be inclined to make the helper functions take an object instead of individual arguments at this point; then we could say things like "if no domain is specified, use the default" instead of having to introspect whether there's an `@`-symbol in the address provided.

(There are also some broader things that would be nice to change like: Removing PSA terminology, since folks aren't familiar with it, and maybe reducing the number of helper functions to reduce what users need to know about - but those are likely beyond the scope of this PR.)

wyardley:
Yeah, 100% - this is probably a little clunky, and not ideal interface-wise, and also agree as a user who's not super steeped in all the internal terminology, something more generic would probably be easier to work with.

An object could make some sense (and might make default behavior simpler?), and also just using an array all the time (for both the member(s) and role(s), ideally) vs. having both the array and string variants would be a bit tidier?

I'd be happy to file a separate GH issue if it would help in tracking making some of these improvements.

I'm guessing trying to change the interface too dramatically could get kind of messy, given the number of tests that would probably need adjustment? But I'm totally happy to make small adjustments or changes here if they can be done relatively simply.

mmv1/third_party/terraform/acctest/bootstrap_iam_test_utils.go
```diff
@@ -14,6 +15,9 @@ import (
 // policy grants the given service agents the given roles.
 // prefix is usually "service-" and indicates the service agent should have the
 // given prefix before the project number.
+// If an address with an `@` (x@y) is passed, the address will be used
+// verbatim, but still prefixed with `serviceAccount:prefix` and suffixed with
```
ScottSuarez, line 19: In accordance with Stephen's suggestion. Wdyt about the following pattern? I can understand how it may get difficult to understand weird implicit logic. This change might make things more explicit overall at the consumer level without needing to know about special behavior for certain characters. 
```
// ServiceAccountConfig holds the configuration for service account creation
type ServiceAccountConfig struct {
    // Name of the service account
    Name string
    // Optional domain override. If empty, defaults to project's iam.gserviceaccount.com
    Domain string
    // Optional project number override. If empty, uses the current project number
    ProjectNumber string
}

// BootstrapRoleConfig holds the configuration for bootstrapping IAM roles
type DesiredIAMRoles struct {
    // Prefix to be added to all service account emails
    Prefix string
    // List of service account configurations
    ServiceAccounts []ServiceAccountConfig
    // IAM roles to be assigned
    Roles []string
}

// BootstrapIAMRoles bootstraps IAM roles for service accounts
// Returns whether the bindings changed
func BootstrapIAMRoles(t *testing.T, config DesiredIAMRoles) bool {
```

ScottSuarez:
I left a suggestion. If it's too complex of a change we can merge this as is and open a ticket!

I think the number of changes could be rather small since we usually use wrapper functions to call this.

mmv1/third_party/terraform/acctest/bootstrap_iam_test_utils.go
wyardley: High level, this seems to be a much better design, albeit one I may not want to get too deep into implementing as an afternoon project 😅 

Couple questions:
1) So basically, new function, with the idea that we'd create this method, and then phase out the wrapper functions later? And it would still wrap `BootstrapAllPSARoles()` for now, just providing a cleaner interface, but with the idea of doing more structural fixes later?
2) It seems like there's really two basic common patterns, so like `${prefix}-${projectNumber}@foo.iam.gserviceaccount.com` (is this PSA?), where "prefix" is usually `service-` and then `[project_number]@foo.gserviceaccount.com` (Google Service Agent?). So I guess in this example, you'd have prefix as nil or empty string to do the latter, and set it to `service-` for the former?
In fact, even in my current use case, could possibly hard-code the project number bit into the function, which would simplify my use case further.
3) If we're making naming it "IAM" vs. something specific to service accounts, maybe needs a param for whether to include the `serviceAccount:` prefix too? Not sure if that's too much YAGNI?

ScottSuarez: 1. Exactly, or we can maintain some of the higher level wrappers if we feel they are applicable still. While allowing us to more flexibly use what's underneath.
2. hmm good point on `{prefix}-${projectNumber}@foo.iam.gserviceaccount.com`. We might want to omit project number altogether from the ServiceAccountConfig object. With the existing calling into our function. I believe you are correct on PSA (Project Service Account) being this specific pattern.
3. That sound reasonable to me! I would prefer it to be optional but customizability is always good.

wyardley: Ok, let me see if I can come up with something that uses that pattern, and if it's possible to do this while keeping `BootstrapAllPSARoles()` more or less as it is. It might delay #12728, which I've now rolled a fix for another breaking test into, so up to you if you want to ship that first or wait for this.

Ideally, there would be some tests as well, for the logic of constructing email addresses, at least, and maybe for some of the role related stuff... but that may also have to wait for some bigger project, especially since adding them with the way the functions are structured now might involve either adding some new functions, or figuring out how to stub some of the network calls.

melinath:
Actually, after thinking this over some more, I think we should make an even more basic change... one of the struggles with the current implementation is that from the tests, it's pretty hard to understand what is actually being granted to what. I think we should stop doing templating at all (except for project_number replacement) and just allow a passthrough of the members being bootstrapped.

I have some time & can take a stab at this today.

wyardley:
@melinath Ok, thanks! I thought about this a little more last night, but haven't started doing anything... if you come up with something that's usable, that works great for me, and then I can adjust that other PR with the tests! I actually kind of like _some_ of the magic here, because there really are only 2-3 patterns in the existing (and probably future?) use cases, but I get what you're saying as well.

Hopefully these changes will produce an outcome that will be easier to understand / read for everyone, as well as be more in line with the overall style of the project.

melinath:
yeah, I think there's maybe room for additional helper functions to make it easier to build IAM member names, but that should be separate IMO from the bootstrapping logic.

melinath:
opened https://github.com/GoogleCloudPlatform/magic-modules/pull/12796

wyardley:
Closing since @melinath has opened #12796

PR #12785, opened 2025-01-16
Skip sleep in VCR replaying mode

wyardley:
In `SleepInSecondsForTest()`, skip the sleep when `VCR_MODE` is `REPLAYING`. This makes the assumption that the sleep is always related to a condition that's already mitigated by the fact that we're using prerecorded fixtures, but it could be extended later to override this behavior if necessary.

Fixes #20916

```release-note:none

```


mmv1/third_party/terraform/acctest/resource_test_utils.go
```diff
@@ -90,7 +91,10 @@ func SkipIfVcr(t *testing.T) {
 
 func SleepInSecondsForTest(t int) resource.TestCheckFunc {
 	return func(s *terraform.State) error {
-		time.Sleep(time.Duration(t) * time.Second)
+		// Assume we never want to sleep when we're in replaying mode.
+		if os.Getenv("VCR_MODE") != "REPLAYING" {
```
wyardley, line 95: I tested this both with / without assignment, but gut is that this is probably good enough? it already seems to work fine / not error if the env var is not defined.

slevenick: I think you also want to check if IsVcrEnabled returns true first. It's possible to set VCR_MODE but not VCR_PATH, which is considered as it not being enabled

wyardley: I originally thought about something along those lines. I think we want the behavior to be like this, right?

No VCR: sleep ✅ 😴 
VCR in `RECORDING` mode: sleep ✅ 😴 
VCR in `REPLAYING` mode: no sleep ❌ 😴 

I guess you're thinking about defending against a corner case where `VCR_MODE` is set to `REPLAYING`, but `VCR_PATH` is not set (I'm not sure if this would work, though I can test)? But if we did:
`if IsVcrEnabled() && os.Getenv("VCR_MODE") != "REPLAYING"`?

I think we'd then suppress the sleep in case #1 (no VCR).

wyardley: Wouldn't read that nicely, but I guess could do:
```
if !IsVcrEnabled() || (IsVcrEnabled() && os.Getenv("VCR_MODE") != "REPLAYING)"
```

slevenick: `if IsVcrEnabled() && os.Getenv("VCR_MODE") == "REPLAYING"` is the most straightforward, non inverted form of what we want to check, but that's for skipping the sleep. You could just add this check at the start and return nil, otherwise sleep

The inverse of that would be:
`if !IsVcrEnabled() || os.Getenv("VCR_MODE") != "REPLAYING"` and if that's true then we want to sleep before returning

wyardley: Thanks! That is a good idea and sounds cleaner / easier to read. Let me know if the updated change looks better.
Testing locally, but so far seems to work as you'd expect.

wyardley: I'm assuming that logging anything in either case (sleeping or skipping the sleep) is probably not needed / wanted?

wyardley:
Tested again.

With changes (and a print statement), in `RECORDING` mode:
```
=== RUN   TestAccResourceGoogleProjectDefaultServiceAccountsDisable
=== PAUSE TestAccResourceGoogleProjectDefaultServiceAccountsDisable
=== CONT  TestAccResourceGoogleProjectDefaultServiceAccountsDisable
sleep
--- PASS: TestAccResourceGoogleProjectDefaultServiceAccountsDisable (147.19s)
PASS
ok  	github.com/hashicorp/terraform-provider-google/google/services/resourcemanager	148.332s
```

with 5s sleep bypassed (`REPLAYING` mode):
```
=== RUN   TestAccResourceGoogleProjectDefaultServiceAccountsDisable
=== PAUSE TestAccResourceGoogleProjectDefaultServiceAccountsDisable
=== CONT  TestAccResourceGoogleProjectDefaultServiceAccountsDisable
skipping sleep
--- PASS: TestAccResourceGoogleProjectDefaultServiceAccountsDisable (24.37s)
PASS
ok  	github.com/hashicorp/terraform-provider-google/google/services/resourcemanager	25.501s
```

Without these changes (in `REPLAYING` mode, with `VCR_PATH` set)
```
=== RUN   TestAccResourceGoogleProjectDefaultServiceAccountsDisable
=== PAUSE TestAccResourceGoogleProjectDefaultServiceAccountsDisable
=== CONT  TestAccResourceGoogleProjectDefaultServiceAccountsDisable
--- PASS: TestAccResourceGoogleProjectDefaultServiceAccountsDisable (29.51s)
PASS
ok  	github.com/hashicorp/terraform-provider-google/google/services/resourcemanager	30.659s
```
It does seem like the test will run with VCR_MODE=REPLAYING` set but no `VCR_PATH` set - just operates as if VCR were not enabled. So more tests:

without these changes, and with`VCR_MODE=REPLAYING` set but no `VCR_PATH` set
```
=== RUN   TestAccResourceGoogleProjectDefaultServiceAccountsDisable
=== PAUSE TestAccResourceGoogleProjectDefaultServiceAccountsDisable
=== CONT  TestAccResourceGoogleProjectDefaultServiceAccountsDisable
--- PASS: TestAccResourceGoogleProjectDefaultServiceAccountsDisable (149.44s)
PASS
ok  	github.com/hashicorp/terraform-provider-google/google/services/resourcemanager	150.565s
```

with changes, and with`VCR_MODE=REPLAYING` set but no `VCR_PATH` set
```
=== RUN   TestAccResourceGoogleProjectDefaultServiceAccountsDisable
=== PAUSE TestAccResourceGoogleProjectDefaultServiceAccountsDisable
=== CONT  TestAccResourceGoogleProjectDefaultServiceAccountsDisable
sleep
--- PASS: TestAccResourceGoogleProjectDefaultServiceAccountsDisable (163.71s)
PASS
ok  	github.com/hashicorp/terraform-provider-google/google/services/resourcemanager	164.843s
```

c2thorn:
@slevenick do you mind taking this one over since you have more VCR-specific knowledge?

slevenick (approved):
Thanks! Test failures should be entirely unrelated

PR #12782, opened 2025-01-16
Add IAM policy to SecurityGateway.yaml

Berro321:
```release-note:new-resource
`google_beyondcorp_security_gateway_iam_binding`
```
```release-note:new-resource
`google_beyondcorp_security_gateway_iam_member`
```
```release-note:new-resource
`google_beyondcorp_security_gateway_iam_policy`
```

PR #12788, opened 2025-01-17
Correct SecurityGateway to be global only

Berro321:
Change is for correcting a mistake in the original submission from last week: https://github.com/GoogleCloudPlatform/magic-modules/pull/12695 

```release-note:bug
beyondcorp: corrected `location` to always be global in `google_beyondcorp_security_gateway` 
```

```release-note:deprecation
beyondcorp: deprecated `location` on `google_beyondcorp_security_gateway`. The only valid value is `global`, which is now also the default value. The field will be removed in a future major release.
```

### DIFF
--- a/mmv1/products/beyondcorp/SecurityGateway.yaml
+++ b/mmv1/products/beyondcorp/SecurityGateway.yaml
@@ -14,14 +14,14 @@
 ---
 name: SecurityGateway
 description: Deployment of Security Gateway.
-base_url: projects/{{project}}/locations/global/securityGateways
+base_url: projects/{{project}}/locations/{{location}}/securityGateways
 update_mask: true
-self_link: projects/{{project}}/locations/global/securityGateways/{{security_gateway_id}}
-create_url: projects/{{project}}/locations/global/securityGateways?securityGatewayId={{security_gateway_id}}
+self_link: projects/{{project}}/locations/{{location}}/securityGateways/{{security_gateway_id}}
+create_url: projects/{{project}}/locations/{{location}}/securityGateways?securityGatewayId={{security_gateway_id}}
 update_verb: PATCH
-id_format: projects/{{project}}/locations/global/securityGateways/{{security_gateway_id}}
+id_format: projects/{{project}}/locations/{{location}}/securityGateways/{{security_gateway_id}}
 import_format:
-  - projects/{{project}}/locations/global/securityGateways/{{security_gateway_id}}
+  - projects/{{project}}/locations/{{location}}/securityGateways/{{security_gateway_id}}
 examples:
   - name: beyondcorp_security_gateway_basic
     primary_resource_id: example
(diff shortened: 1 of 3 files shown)


Berro321:
@melinath Is it reasonable to add the `override-breaking-change` label to this PR? This is a correction for a mistake in the previous change, we don't have any customers using this yet.

melinath (changes requested):
I don't think we should roll this back, since it would break folks who start using this resource using v6.16.0. To put it in terms of our [policies](https://googlecloudplatform.github.io/magic-modules/breaking-changes/make-a-breaking-change/), in minor releases we only allow breaking changes that impact _all valid configurations_. However, in this case, setting location to "global" works, and removing the field would break that configuration.

Instead, I would suggest changing the field from required to optional and setting a default_value of `global`.

If the field is expected to continue only supporting global indefinitely, you could also add [client-side validation](https://googlecloudplatform.github.io/magic-modules/reference/field/#validation) to ensure that it is set to "global". Otherwise, you could document that it only allows "global" or link to c.g.c documentation that lists the allowed values.

melinath (approved):
LGTM - breaking change is fine since using values other than global wouldn't work.

By the way, if you want to remove this field, you can do so in the next major release (which will have a contribution window this summer).

mmv1/products/beyondcorp/SecurityGateway.yaml
```diff
@@ -51,7 +51,10 @@ parameters:
     description: Resource ID segment making up resource `name`. It identifies the resource within its parent collection as described in https://google.aip.dev/122.
```
melinath, line 51:
```suggestion
    description: Resource ID segment making up resource `name`. It identifies the resource within its parent collection as described in https://google.aip.dev/122. Must be omitted or set to `global`.
```

Berro321:
Before submitting, I want to test these changes will work with the IAM policy i'm adding in https://github.com/GoogleCloudPlatform/magic-modules/pull/12782. Is it ok to change the import_format for the IAM policy to only have "global"? In our case, only the global location supports IAM.

melinath (changes requested):
I think you will need to change the URL definitions to all use `locations/global` instead of `locations/{{location}}` - that should prevent the location field from being added to the IAM resource, because of how the generation of IAM code works (but it would probably be worth double-checking that it actually works before we merge this.)

If you ever expect to support non-global locations it's probably best to just leave this as is.

If you never expect to support non-global locations, it would probably be good to mark the location field as [deprecated](https://googlecloudplatform.github.io/magic-modules/breaking-changes/make-a-breaking-change/#add-deprecations-and-warnings-to-the-main-branch-of-magic-modules) for extra clarity.

Berro321:
If I change all url defs (IAM policy and all outside of it) then it works and doesn't add the locations field. I'll add the deprecation warning since we don't ever plan to have these as global resources. Since the tests will fail if the location is not 'global', is there a way to limit the VCR tests in the PR to only use the global location? Is that through the [test_env_vars](https://github.com/GoogleCloudPlatform/magic-modules/blob/eae83792920ab7e4acc6f5e65c06413c8ecbd7d9/mmv1/api/resource/examples.go#L78) field in `examples`?

melinath:
The tests in this PR are already using the global location; the tests in the IAM PR should also use the global location once the location field no longer exists.

Berro321:
I'm seeing that the test projects in the IAM PR run out of quota since there is a limit of 1 resource per project. Is there a list of projects that we can increase the quota for?

melinath (approved):

PR #12791, opened 2025-01-17
compute: Add `scheduling.termination_time` field to compute_instance resources

Asiderr:
This patch adds the `scheduling.termination_time` field to the following resources:
* `google_compute_instance`
* `google_compute_instance_from_machine_image` (beta)
* `google_compute_instance_from_template`
* `google_compute_instance_template`
* `google_compute_region_instance_template`

It also adds a helper function `hasTerminationTimeChanged`, which allows to stop the instance while updating the `termination_time`.

```release-note:enhancement
compute: added `scheduling.termination_time` field to `google_compute_instance` resource
```

```release-note:enhancement
compute: added `scheduling.termination_time` field to `google_compute_instance_from_machine_image` resource
```

```release-note:enhancement
compute: added `scheduling.termination_time` field to `google_compute_instance_from_template` resource
```

```release-note:enhancement
compute: added `scheduling.termination_time` field to `google_compute_instance_template` resource
```

```release-note:enhancement
compute: added `scheduling.termination_time` field to `google_compute_region_instance_template` resource
```

### DIFF
--- a/mmv1/third_party/terraform/services/compute/compute_instance_helpers.go.tmpl
+++ b/mmv1/third_party/terraform/services/compute/compute_instance_helpers.go.tmpl
@@ -199,6 +199,9 @@ func expandScheduling(v interface{}) (*compute.Scheduling, error) {
 		scheduling.LocalSsdRecoveryTimeout = transformedLocalSsdRecoveryTimeout
 		scheduling.ForceSendFields = append(scheduling.ForceSendFields, "LocalSsdRecoveryTimeout")
 	}
+	if v, ok := original["termination_time"]; ok {
+		scheduling.TerminationTime = v.(string)
+	}
 	return scheduling, nil
 }
 
@@ -344,6 +347,7 @@ func flattenScheduling(resp *compute.Scheduling) []map[string]interface{} {
 		"provisioning_model":  resp.ProvisioningModel,
 		"instance_termination_action": resp.InstanceTerminationAction,
 		"availability_domain": resp.AvailabilityDomain,
+		"termination_time":    resp.TerminationTime,
 	}
 
 	if resp.AutomaticRestart != nil {
@@ -796,9 +800,12 @@ func schedulingHasChangeRequiringReboot(d *schema.ResourceData) bool {
 {{ if ne $.TargetVersionName `ga` -}}
 	return hasNodeAffinitiesChanged(oScheduling, newScheduling) ||
 		hasMaxRunDurationChanged(oScheduling, newScheduling) ||
-		hasGracefulShutdownChangedWithReboot(d, oScheduling, newScheduling)
+		hasGracefulShutdownChangedWithReboot(d, oScheduling, newScheduling) ||
+		hasTerminationTimeChanged(oScheduling, newScheduling)
 {{- else }}
-	return hasNodeAffinitiesChanged(oScheduling, newScheduling) || hasMaxRunDurationChanged(oScheduling, newScheduling)
+	return (hasNodeAffinitiesChanged(oScheduling, newScheduling) ||
+		hasMaxRunDurationChanged(oScheduling, newScheduling) ||
+		hasTerminationTimeChanged(oScheduling, newScheduling))
 {{- end }}
 }
 
@@ -849,6 +856,24 @@ func schedulingHasChangeWithoutReboot(d *schema.ResourceData) bool {
 		return true
 	}
 {{- end }}
+
+	return false
+}
+
+func hasTerminationTimeChanged(oScheduling, nScheduling map[string]interface{}) bool {
+	oTerminationTime := oScheduling["termination_time"].(string)
+	nTerminationTime := nScheduling["termination_time"].(string)
+
+	if len(oTerminationTime) == 0 && len(nTerminationTime) == 0 {
+		return false
+	}
+	if len(oTerminationTime) == 0 || len(nTerminationTime) == 0 {
+		return true
+	}
+
+	if oTerminationTime != nTerminationTime {
+		return true
+	}
 
 	return false
 }
--- a/mmv1/third_party/terraform/services/compute/resource_compute_instance_from_template_test.go.tmpl
+++ b/mmv1/third_party/terraform/services/compute/resource_compute_instance_from_template_test.go.tmpl
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"regexp"
 	"testing"
+	"time"
 
 	"github.com/hashicorp/terraform-plugin-testing/helper/resource"
 	"github.com/hashicorp/terraform-plugin-testing/terraform"
@@ -401,6 +402,32 @@ func TestAccComputeInstanceFromTemplate_overrideScheduling(t *testing.T) {
 		Steps: []resource.TestStep{
 			{
 				Config: testAccComputeInstanceFromTemplate_overrideScheduling(templateDisk, templateName, instanceName),
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckComputeInstanceExists(t, resourceName, &instance),
+				),
+			},
+		},
+	})
+}
+
+func TestAccComputeInstanceFromTemplate_TerminationTime(t *testing.T) {
+	t.Parallel()
+
+	var instance compute.Instance
+	instanceName := fmt.Sprintf("tf-test-%s", acctest.RandString(t, 10))
+	templateName := fmt.Sprintf("tf-test-%s", acctest.RandString(t, 10))
+	templateDisk := fmt.Sprintf("tf-test-%s", acctest.RandString(t, 10))
+	resourceName := "google_compute_instance_from_template.inst"
+	now := time.Now().UTC()
+	terminationTime := time.Date(now.Year(), now.Month(), now.Day(), 23, 59, 59, 9999, now.Location()).Format(time.RFC3339)
+
+	acctest.VcrTest(t, resource.TestCase{
+		PreCheck:                 func() { acctest.AccTestPreCheck(t) },
+		ProtoV5ProviderFactories: acctest.ProtoV5ProviderFactories(t),
+		CheckDestroy:             testAccCheckComputeInstanceFromTemplateDestroyProducer(t),
+		Steps: []resource.TestStep{
+			{
+				Config: testAccComputeInstanceFromTemplate_terminationTime(templateDisk, templateName, terminationTime, instanceName),
 				Check: resource.ComposeTestCheckFunc(
 					testAccCheckComputeInstanceExists(t, resourceName, &instance),
 				),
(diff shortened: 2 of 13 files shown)


mmv1/third_party/terraform/services/compute/resource_compute_instance_test.go.tmpl
```diff
@@ -1437,6 +1437,31 @@ func TestAccComputeInstance_scheduling(t *testing.T) {
...
+	var instance compute.Instance
+	var instanceName = fmt.Sprintf("tf-test-%s", acctest.RandString(t, 10))
+	now := time.Now().UTC()
+	terminationTime := now.Add(24 * time.Hour).Format(time.RFC3339)
+
+	acctest.VcrTest(t, resource.TestCase{
+		PreCheck:                 func() { acctest.AccTestPreCheck(t) },
+		ProtoV5ProviderFactories: acctest.ProtoV5ProviderFactories(t),
+		CheckDestroy:             testAccCheckComputeInstanceDestroyProducer(t),
+		Steps: []resource.TestStep{
```
melinath, line 1452: could you add an update step here to ensure that changing the termination time works as expected?

melinath (changes requested):
In addition to the comment below, could you add [unit tests](https://googlecloudplatform.github.io/magic-modules/test/test/#add-unit-tests) for schedulingHasChangeRequiringReboot?

melinath (changes requested):
looks like this also needs to be added to google_compute_instance_from_machine_image and google_compute_instance_from_template

mmv1/third_party/terraform/services/compute/resource_compute_instance_test.go.tmpl
Asiderr: I added step where update is checked

Asiderr:
@melinath 
> In addition to the comment below, could you add [unit tests](https://googlecloudplatform.github.io/magic-modules/test/test/#add-unit-tests) for schedulingHasChangeRequiringReboot?

The `schedulingHasChangeRequiringReboot` requires `d *schema.ResourceData` as an input parameter so it would be easier to create unit tests for `hasNodeAffinitiesChanged`, `hasMaxRunDurationChanged` and `hasTerminationTimeChanged` with corresponding test cases. There is no test file for this helper, so should I create a new test file or do we want to move `schedulingHasChangeRequiringReboot` and related functions to the `resource_compute_instance`?

> looks like this also needs to be added to google_compute_instance_from_machine_image and google_compute_instance_from_template

It will be added in the future in a separate PR.

mmv1/third_party/terraform/services/compute/resource_compute_instance_test.go.tmpl
```diff
@@ -1437,6 +1437,40 @@ func TestAccComputeInstance_scheduling(t *testing.T) {
...
+			{
+				Config: testAccComputeInstance_TerminationTime(instanceName, terminationTime),
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckComputeInstanceExists(
+						t, "google_compute_instance.foobar", &instance),
+				),
+			},
+			computeInstanceImportStep("us-central1-a", instanceName, []string{"allow_stopping_for_update"}),
+			{
+				Config: testAccComputeInstance_TerminationTime(instanceName, terminationTimeUpdated),
```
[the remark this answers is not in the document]
Asiderr: I added following step

Asiderr:
@melinath I added all missing fields which were connected to the `termination_time`. So new fields and tests are added to following resources:
* google_compute_instance
* google_compute_instance_from_machine_image (beta)
* google_compute_instance_from_template
* google_compute_instance_template
* google_compute_region_instance_template

melinath:
> @melinath Also I wonder if there is any easy possibility to unit test private functions in helpers? `schedulingHasChangeRequiringReboot` as well as `hasTerminationTimeChanged`, `hasNodeAffinitiesChanged` and `hasMaxRunDurationChanged` are privates due to naming, so I see no easy way to test them.

yes, you can add helper tests in the services/compute/ folder, which will have access to the private helpers in that folder.

melinath:
> regarding unit tests for schedulingHasChangeRequiringReboot: the way we generally get around needing Terraform core structs is by making the function a thin wrapper that gets the necessary data (in this case, the scheduling data) and passes it to the function that's actually tested. A new compute_instance_helpers test file seems reasonable.

It looks like I forgot to circle back to this and say: the wrapper we use is ResourceDataMock: https://github.com/hashicorp/terraform-provider-google-beta/blob/50e64ad11c68fbfe95842bb6e52535e235320a68/google-beta/tpgresource/resource_test_utils.go#L18

Examples of usage: https://github.com/search?q=repo%3Ahashicorp%2Fterraform-provider-google-beta+ResourceDataMock+path%3A%2F%5Egoogle-beta%5C%2Fservices%5C%2F%2F&type=code

melinath (changes requested):
apologies for the delayed review - taking a quick glance, this seems reasonable and I can see that you have added the requested tests for *_from_template and *_from_machine_image, as well as adding the field to instance_template and region_instance_template. If you could add some basic unit tests as previously discussed, that would be great; there's also a merge conflict that needs to be resolved. Other than that I think this should be good to go as long as the tests pass! Thanks!

Asiderr:
> apologies for the delayed review - taking a quick glance, this seems reasonable and I can see that you have added the requested tests for *_from_template and *_from_machine_image, as well as adding the field to instance_template and region_instance_template. If you could add some basic unit tests as previously discussed, that would be great; there's also a merge conflict that needs to be resolved. Other than that I think this should be good to go as long as the tests pass! Thanks!

@melinath I resolved conflicts. I added new test file for helper. The `ResourceDataMock` does not implement `GetChange` method that is used in schedulingHasChangeRequiringReboot so instead of testing `schedulingHasChangeRequiringReboot` I tested function `hasTerminationTimeChanged` that I'm adding to this helper file.

mmv1/third_party/terraform/services/compute/compute_instance_helpers.go.tmpl
```diff
@@ -810,9 +815,12 @@ func schedulingHasChangeRequiringReboot(d *schema.ResourceData) bool {
 {{ if ne $.TargetVersionName `ga` -}}
 	return hasNodeAffinitiesChanged(oScheduling, newScheduling) ||
 		hasMaxRunDurationChanged(oScheduling, newScheduling) ||
-		hasGracefulShutdownChangedWithReboot(d, oScheduling, newScheduling)
+		hasGracefulShutdownChangedWithReboot(d, oScheduling, newScheduling) ||
+		hasTerminationTimeChanged(oScheduling, newScheduling))
```
melinath, line 819:
```suggestion
		hasTerminationTimeChanged(oScheduling, newScheduling)
```

melinath (changes requested):

mmv1/third_party/terraform/services/compute/resource_compute_instance_from_template_test.go.tmpl
```diff
@@ -1482,6 +1509,56 @@ resource "google_compute_instance_from_template" "inst" {
...
+
+  scheduling {
+    instance_termination_action = "STOP"
+    termination_time = "%s"
+  }
+
+  can_ip_forward = true
+}
+
+resource "google_compute_instance_from_template" "inst" {
```
melinath, line 1553: This resource needs to override the termination_time.

(The `testAccComputeInstanceFromTemplate_overrideScheduling` step should have added a similar override for its scheduling fields but apparently the reviewer missed asking them to add it. No need to fix that since this is already a large change.)

mmv1/third_party/terraform/services/compute/resource_compute_region_instance_template_test.go.tmpl
```diff
@@ -1259,6 +1259,37 @@ func TestAccComputeRegionInstanceTemplate_maxRunDuration_onInstanceStopAction(t
 	})
 }
 
+func TestAccComputeRegionInstanceTemplate_instanceTerminationAction_terminationTime(t *testing.T) {
+	t.Parallel()
+
+	var instanceTemplate compute.InstanceTemplate
+	now := time.Now().UTC()
+	terminationTime := now.Add(24 * time.Hour).Format(time.RFC3339)
```
melinath, line 1267: Setting `terminationTime` like this means that the tests become non-deterministic (because running it twice will give two different termination times.) Here & everywhere else - this should be changed to a relatively static time in the future (such as the beginning of the next year).

```suggestion
	terminationTime := time.Date(time.Now().Year(), 12, 31, 0, 0, 0, 0, time.Now().Location()).AddDate(0, 0, 1).Format(time.RFC3339)
```

mmv1/third_party/terraform/services/compute/compute_instance_helpers.go.tmpl
```diff
@@ -199,6 +199,10 @@ func expandScheduling(v interface{}) (*compute.Scheduling, error) {
 		scheduling.LocalSsdRecoveryTimeout = transformedLocalSsdRecoveryTimeout
 		scheduling.ForceSendFields = append(scheduling.ForceSendFields, "LocalSsdRecoveryTimeout")
 	}
+	if v, ok := original["termination_time"]; ok {
+		scheduling.TerminationTime = v.(string)
+		scheduling.ForceSendFields = append(scheduling.ForceSendFields, "TerminationTime")
+	}
```
melinath, line 205: This is causing a bunch of unrelated tests to fail with the error:

```
        Error: Error creating instance: googleapi: Error 400: Invalid value for field 'resource.scheduling.terminationTime': ''. Termination time for given provisioning model is not supported without an instance termination action., invalid
```

This is happening because, on Create, the provider is sending `"terminationTime": ""` in the API request. I suspect this may be caused by something like: `d.Get` at create time returning `""` for string fields (which means it's already "present" in `original`). But I could be wrong.

However, checking whether `v == ""` and only running this code if not would effectively mean that the ForceSendFields is never triggered in the cases where it would actually have an effect.

Do you know whether ForceSendFields is necessary here? If you're not sure, the way to check is to remove it and to add update tests that remove this field from a configuration (without removing the scheduling block overall). If ForceSendFields _is_ necessary, those new tests will fail, and we can try to figure out the best workaround.

melinath (changes requested):

mmv1/third_party/terraform/services/compute/resource_compute_instance_from_template_test.go.tmpl
melinath: This resource (google_compute_instance_from_template.inst) still needs to override `scheduling.termination_time` - this is why the automation is detecting `google_compute_instance_from_template.scheduling.termination_time` as untested: https://github.com/GoogleCloudPlatform/magic-modules/pull/12791#issuecomment-2675165487

mmv1/third_party/terraform/services/compute/compute_instance_helpers.go.tmpl
```diff
@@ -199,6 +199,9 @@ func expandScheduling(v interface{}) (*compute.Scheduling, error) {
 		scheduling.LocalSsdRecoveryTimeout = transformedLocalSsdRecoveryTimeout
 		scheduling.ForceSendFields = append(scheduling.ForceSendFields, "LocalSsdRecoveryTimeout")
 	}
+	if v, ok := original["termination_time"]; ok {
+		scheduling.TerminationTime = v.(string)
```
melinath, line 203: Continuing the discussion from https://github.com/GoogleCloudPlatform/magic-modules/pull/12791#discussion_r1977689485 - it sounds like ForceSendFields doesn't fix the problem of this not getting properly unset? It would definitely match the behavior of other string fields in this expand function, so they might also not behave as expected if this doesn't. (But that's outside the scope of this PR.)

I believe the fix would be to make sure that TerminationTime gets set to `""` (and added to ForceSendFields) if it's not present in `original`.

melinath (changes requested):

mmv1/third_party/terraform/services/compute/resource_compute_instance_from_template_test.go.tmpl
Asiderr: I added overwrite.

mmv1/third_party/terraform/services/compute/compute_instance_helpers.go.tmpl
Asiderr: I added workaround for this issue. Now everything is working as excepted.

Asiderr:
> This resource (google_compute_instance_from_template.inst) still needs to override scheduling.termination_time - this is why the automation is detecting google_compute_instance_from_template.scheduling.termination_time as untested: https://github.com/GoogleCloudPlatform/magic-modules/pull/12791#issuecomment-2675165487
I added overwrite.

mmv1/third_party/terraform/services/compute/compute_instance_helpers.go.tmpl
melinath: What was the workaround?

Asiderr: I added force new to this field. Currently it's not possible to delete this field without destroying the old VM due to the API bug. Once there is no force new and `termination_time` is added to the forcesend, during deletion API is always trying to parse an empty string as a timestamp. One more improvement that could be added is creation of the customdiff that would set force new only when this field is deleted.
BTW other scheduling fields are also affected by this forcesend and could not be deleted once other scheduling fields still remain (I tested this for instance_termination_action).

melinath (approved):
Confirmed via experimentation - API doesn't accept empty values for this field. ForceNew is an acceptable workaround.

Thanks for sticking with this!

PR #12792, opened 2025-01-17
Add field-level metadata generation for mmv1 resources

roaks3:
This is the initial pass at adding field metadata to all mmv1 resources. I was able to check a small set of resources for accuracy, but further accuracy improvements are expected to come after we have the data ingested and can more clearly run checks.

Notable decisions:
* I chose to have the `api_field` default to the value of `field` when it isn't specified, so that it does not need to be included for the vast majority of fields. This massively reduces the size and visual noise in metadata files, which will be helpful for handwritten resources later.
* We do not generate explicit `key`/`value` metadata fields for maps, even though that is how they are described in the API, because it isn't meaningful for coverage. Maps are treated like a single field.

```release-note:none

```

### DIFF
--- a/mmv1/api/resource.go
+++ b/mmv1/api/resource.go
@@ -626,7 +626,7 @@ func (r Resource) RootProperties() []*Type {
 func (r Resource) LeafProperties() []*Type {
 	types := r.AllNestedProperties(google.Concat(r.RootProperties(), r.UserVirtualFields()))
 
-	// Remove types that have children, becasue we only want "leaf" fields
+	// Remove types that have children, because we only want "leaf" fields
 	types = slices.DeleteFunc(types, func(t *Type) bool {
 		nestedProperties := t.NestedProperties()
 		return len(nestedProperties) > 0
--- a/mmv1/api/resource.go
+++ b/mmv1/api/resource.go
@@ -500,6 +500,12 @@ func (r Resource) UserParameters() []*Type {
 	})
 }
 
+func (r Resource) UserVirtualFields() []*Type {
+	return google.Reject(r.VirtualFields, func(p *Type) bool {
+		return p.Exclude
+	})
+}
+
 func (r Resource) ServiceVersion() string {
 	if r.CaiBaseUrl != "" {
 		return extractVersionFromBaseUrl(r.CaiBaseUrl)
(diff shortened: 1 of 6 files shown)


mmv1/api/resource.go
```diff
@@ -615,6 +621,28 @@ func (r Resource) RootProperties() []*Type {
 	return props
 }
 
+// Returns a sorted list of all "leaf" properties, meaning properties that have
+// no children.
+func (r Resource) LeafProperties() []*Type {
+	types := r.AllNestedProperties(google.Concat(r.RootProperties(), r.UserVirtualFields()))
+
+	// Remove types that have children, becasue we only want "leaf" fields
```
melinath, line 629: sp - feel free to ignore

```suggestion
	// Remove types that have children, because we only want "leaf" fields
```

mmv1/api/resource_test.go
```diff
@@ -198,3 +198,105 @@ func TestResourceServiceVersion(t *testing.T) {
 		})
 	}
 }
+
+func TestLeafProperties(t *testing.T) {
```
melinath, line 202: would it be useful to have tests for non-nested leaf properties to cover the base case?

mmv1/templates/terraform/metadata.yaml.tmpl
```diff
@@ -3,3 +3,14 @@ generation_type: 'mmv1'
 api_service_name: '{{ $.ProductMetadata.ServiceName }}'
 api_version: '{{ or $.ProductMetadata.ServiceVersion $.ServiceVersion }}'
 api_resource_type_kind: '{{ or $.ApiResourceTypeKind $.Name }}'
+fields:
+{{- range $p := $.LeafProperties }}
+  -
+    field: '{{ $p.MetadataLineage }}'
```
melinath, line 9: I think this newline isn't necessary / isn't how we generally format our yaml. It adds a lot of whitespace to the yaml file.

```suggestion
  - field: '{{ $p.MetadataLineage }}'
```

melinath (approved):

mmv1/api/resource_test.go
roaks3: Yea that's fair, added a base case

mmv1/templates/terraform/metadata.yaml.tmpl
roaks3: Thanks yea, that's much better

melinath (approved):

PR #12794, opened 2025-01-17
Support apigee environment properties update

Lagu22:
Update google_apigee_environment update operation's http verb from `PATCH` ([modifyEnvironment](https://cloud.google.com/apigee/docs/reference/apis/apigee/rest/v1/organizations.environments/modifyEnvironment)) to `PUT` ([update](https://cloud.google.com/apigee/docs/reference/apis/apigee/rest/v1/organizations.environments/update)). 

Note from the update operation's API doc: `Note: Both PUT and POST methods are supported for updating an existing environment.` 

Partially Fixes: https://github.com/hashicorp/terraform-provider-google/issues/20931

```release-note: bug
apigee: fixed `properties` field update on `google_apigee_environment` resource
```

### DIFF
--- a/mmv1/products/apigee/Environment.yaml
+++ b/mmv1/products/apigee/Environment.yaml
@@ -23,9 +23,6 @@ docs:
 base_url: 'environments'
 self_link: '{{org_id}}/environments/{{name}}'
 create_url: '{{org_id}}/environments'
-update_url: '{{org_id}}/environments/{{name}}'
-update_verb: 'PATCH'
-update_mask: true
 import_format:
   - '{{org_id}}/environments/{{name}}'
   - '{{org_id}}/{{name}}'


mmv1/products/apigee/Environment.yaml
```diff
@@ -23,9 +23,6 @@ docs:
 base_url: 'environments'
 self_link: '{{org_id}}/environments/{{name}}'
 create_url: '{{org_id}}/environments'
-update_url: '{{org_id}}/environments/{{name}}'
-update_verb: 'PATCH'
```
BBBmau, line 27: not sure how this would be much of a benefit. Is the idea that because of the use of `PUT` instead of `PATCH` the reason for changing it despite their being no real difference?

Image showcases what I'm referring to (PUT is mentioned for the endpoint)

image

Lagu22: Since the update mask is no longer used I thought it made sense to use default values when possible, that is the reason for the changes on ln 26 & 27. Update mask is no longer being used as it does not support updating the `properties` field.

BBBmau (approved):

PR #12796, opened 2025-01-17
Implemented simpler IAM bootstrapping logic

melinath:
This should make it easier when looking at tests to see what is being bootstrapped.

I also switched the logic from failing on the first bootstrapping to waiting for the bootstrapping to finish.

I'll follow up with separate PRs to allow easier generation of this from yaml and convert at least some of the existing usage.

```release-note:none

```

### DIFF
--- a/mmv1/third_party/terraform/services/composer/resource_composer_environment_test.go.tmpl
+++ b/mmv1/third_party/terraform/services/composer/resource_composer_environment_test.go.tmpl
@@ -22,14 +22,29 @@ const testComposerNetworkPrefix = "tf-test-composer-net"
 const testComposerBucketPrefix = "tf-test-composer-bucket"
 const testComposerNetworkAttachmentPrefix = "tf-test-composer-nta"
 
-func allComposerServiceAgents() []string {
-	return []string{
-		"cloudcomposer-accounts",
-		"compute-system",
-		"container-engine-robot",
-		"gcp-sa-artifactregistry",
-		"gcp-sa-pubsub",
-	}
+func bootstrapComposerServiceAgents() {
+	acctest.BootstrapIamMembers([]acctest.IamMember{
+		{
+			member: "serviceAccount:service-{project_number}@cloudcomposer-accounts.iam.gserviceaccount.com",
+			role:   "roles/cloudkms.cryptoKeyEncrypterDecrypter",
+		},
+		{
+			member: "serviceAccount:service-{project_number}@compute-system.iam.gserviceaccount.com",
+			role:   "roles/cloudkms.cryptoKeyEncrypterDecrypter",
+		},
+		{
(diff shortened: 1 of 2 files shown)


ScottSuarez (approved):

mmv1/third_party/terraform/services/composer/resource_composer_environment_test.go.tmpl
```diff
@@ -22,27 +22,27 @@ const testComposerNetworkPrefix = "tf-test-composer-net"
...
 		{
-			member: "serviceAccount:service-{project_number}@compute-system.iam.gserviceaccount.com",
-			role:   "roles/cloudkms.cryptoKeyEncrypterDecrypter",
+			Member: "serviceAccount:service-{project_number}@compute-system.iam.gserviceaccount.com",
+			Role:   "roles/cloudkms.cryptoKeyEncrypterDecrypter",
 		},
 		{
-			member: "serviceAccount:service-{project_number}@container-engine-robot.iam.gserviceaccount.com",
-			role:   "roles/cloudkms.cryptoKeyEncrypterDecrypter",
+			Member: "serviceAccount:service-{project_number}@container-engine-robot.iam.gserviceaccount.com",
```
wyardley, line 36: are there ever cases where you want to allow a different TLD? Would it work to end at `foo` or `foo.iam`, or is that too magical?

melinath: I'd prefer not to be at all magical for the bootstrapping... I know it's more verbose & a bit duplicative but it's much more clear about what's happening. We can always consider adding shortcuts for building the members later (and if we did that, we could make those available as provider functions as well).

melinath: I've had too many times (in various situations) where I have to debug something and end up digging through like twelve layers of "helpers" to figure out what is actually happening. 😂

wyardley: Totally makes sense!

And with the various wrappers / helpers also (as discussed in the other tickets), it might be nice to flatten them a bit... right now, there's a helper for if you want one member with multiple roles, another if you want multiple members with one role, and so on....

wyardley:
> I also switched the logic from failing on the first bootstrapping to waiting for the bootstrapping to finish.

This, I think, is great. I had asked / wondered about this a few times... while I get not wanting to wait, this is something that ideally doesn't happen on each run, and having the job fail the first time and succeed (vs. sleeping or waiting) seemed suboptimal.

So I think this will be a huge win.

melinath:
These tests were not impacted by this change - they don't bootstrap IAM resources.

PR #12801, opened 2025-01-18
Updated docs on running single tests

wyardley:
The examples refer to running a single test, but the `-run` command in the example could potentially run lots of tests unexpectedly.

Use a `-run=` example with the correct syntax for passing a literal dollar sign to the regex -- while the docs link to the golang docs, because of shell quoting and because of it getting passed via make, it seems to be necessary to use double dollar signs.

I don't 100% know this is the right solution, but I haven't gotten a direct answer or an example of a better way to do it.

Open to suggestions both as to a better fix

I debated anchoring on the front side as well (`-run=^TestFoo_bar$$`), but seems less necessary. But it is very easy to accidentally kick off a whole bunch of expensive and slow tests if you match too broadly, including if you use a bad regex, forget the dash before `run`, or do any number of other things.

Fixes hashicorp/terraform-provider-google#19934

```release-note:none

```

### DIFF
--- a/docs/content/test/run-tests.md
+++ b/docs/content/test/run-tests.md
@@ -63,7 +63,13 @@ aliases:
     make testacc TEST=./google/services/container TESTARGS='-run=TestAccContainerNodePool_basic$$'
     ```
 
-> **Note:** Acceptance tests create actual infrastructure which can incur costs. Acceptance tests may not clean up after themselves if interrupted, so you may want to check for stray resources and / or billing charges.
+    To run all tests matching, e.g., `TestAccContainerNodePool*`, omit the trailing `$$`:
+
+    ```bash
+    make testacc TEST=./google/services/container TESTARGS='-run=TestAccContainerNodePool'
+    ```
+
+    > **Note:** Acceptance tests create actual infrastructure which can incur costs. Acceptance tests may not clean up after themselves if interrupted, so you may want to check for stray resources and / or billing charges.
 
 1. Optional: Save verbose test output (including API requests and responses) to a file for analysis.
 
@@ -95,7 +101,14 @@ aliases:
     ```bash
     make testacc TEST=./google-beta/services/container TESTARGS='-run=TestAccContainerNodePool'
     ```
-> **Note:** Acceptance tests create actual infrastructure which can incur costs. Acceptance tests may not clean up after themselves if interrupted, so you may want to check for stray resources and / or billing charges.
+
+    To run all tests matching, e.g., `TestAccContainerNodePool*`, omit the trailing `$$`:
+
+    ```bash
+    make testacc TEST=./google/services/container TESTARGS='-run=TestAccContainerNodePool'
+    ```
+
+    > **Note:** Acceptance tests create actual infrastructure which can incur costs. Acceptance tests may not clean up after themselves if interrupted, so you may want to check for stray resources and / or billing charges.
 
 1. Optional: Save verbose test output to a file for analysis.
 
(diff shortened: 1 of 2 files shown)


docs/content/test/run-tests.md
```diff
@@ -60,15 +60,15 @@ aliases:
 1. Run acceptance tests for only modified resources. (Full test runs can take over 9 hours.) See [Go's documentation](https://pkg.go.dev/cmd/go#hdr-Testing_flags) for more information about `-run` and other flags.
 
     ```bash
-    make testacc TEST=./google/services/container TESTARGS='-run=TestAccContainerNodePool'
+    make testacc TEST=./google/services/container TESTARGS='-run=TestAccContainerNodePool_basic$$'
```
shuyama1, line 63: I’d suggest also including an example for running multiple tests. For instance, you could explicitly state that the above example is for running single tests, and then add something like:

To run multiple tests with the `TestAccContainerNodePool` prefix:
```bash
    make testacc TEST=./google/services/container TESTARGS='-run=TestAccContainerNodePool'
```

shuyama1 (commented):
This change makes perfect sense! I’ve built the document locally and verified that `$$` renders correctly. 
Screenshot 2025-01-24 at 11 32 21 AM

docs/content/test/run-tests.md
wyardley: Thanks. That's what I was thinking as well, but because the wording was kind of narrow, and referred to the upstream docs, I wasn't totally sure. I can add something here.

wyardley:
@shuyama1 how does this pass look?

Also, I added some simple docs on how to run a test in VCR recording / replaying mode (see [this comment](https://github.com/hashicorp/terraform-provider-google/issues/20059#issuecomment-2452629162))

docs/content/test/run-tests.md
```diff
@@ -290,6 +303,26 @@ Configure Terraform to use locally-built binaries for `google` and `google-beta`
     TF_LOG=DEBUG TF_LOG_PATH=output.log TF_CLI_CONFIG_FILE="$HOME/tf-dev-override.tfrc" terraform apply
     ```
 
+### Run Tests with VCR Locally
+
+Running tests in "replaying" mode locally can sometimes be useful. In particular, it can allow you to test more quickly, cheaply, and without spinning up real infrastructure, once you've got an initial recording.
+
+It can also be helpful for debugging tests that seem to work locally, but fail in CI in replaying mode.
+   
+Make sure `$VCR_PATH` is set, and points to a local directory which exists, and set `VCR_MODE` appropriately.
```
shuyama1, line 312:
```suggestion
VCR tests record HTTP request/response interactions in cassettes and replay them in future runs without calling the real API.

Running tests in "REPLAYING" mode locally can sometimes be useful. In particular, it can allow you to test more quickly, cheaply, and without spinning up real infrastructure, once you've got an initial recording.

It can also be helpful for debugging tests that seem to work locally, but fail in CI in replaying mode.
   
VCR is controlled via two variables:
- `VCR_MODE`: `REPLAYING` or `RECORDING` mode 
- `VCR_PATH`: Path where recorded cassettes are stored. 

Ensure both variables are configured to properly trigger VCR tests locally.
```

shuyama1: I added a bit more details. Let me know your thoughts!

docs/content/test/run-tests.md
```diff
@@ -290,6 +303,26 @@ Configure Terraform to use locally-built binaries for `google` and `google-beta`
...
+
+If you don't already have an existing cassette that's up to date, first do a run in "recording" mode:
```
shuyama1, line 314:
```suggestion
If you don't already have an existing cassette that's up to date, first do a run in `RECORDING` mode:
```

nit - ensure users can copy and run it directly

docs/content/test/run-tests.md
```diff
@@ -290,6 +303,26 @@ Configure Terraform to use locally-built binaries for `google` and `google-beta`
...
+   
+```bash
+VCR_PATH=$HOME/.vcr/ VCR_MODE=RECORDING  make testacc TEST=./google/services/alloydb TESTARGS='-run=TestAccContainerNodePool_basic$$'
+```
+
+Now run the same test again in "replaying" mode:
```
shuyama1, line 320:
```suggestion
Now run the same test again in `REPLAYING` mode:
```

shuyama1 (commented):
Thanks for helping improving the documentation! Mostly LGTM, only left some small comments

docs/content/test/run-tests.md
wyardley: thanks - I had it that way originally and thought maybe it would look too distracting, but good with it this way.

docs/content/test/run-tests.md
```diff
@@ -95,12 +101,19 @@ aliases:
     ```bash
     make testacc TEST=./google-beta/services/container TESTARGS='-run=TestAccContainerNodePool'
     ```
-> **Note:** Acceptance tests create actual infrastructure which can incur costs. Acceptance tests may not clean up after themselves if interrupted, so you may want to check for stray resources and / or billing charges.
+
+    To run all tests matching, e.g., `TestAccContainerNodePool*`, omit the trailing `$$`:
+
+    ```bash
+    make testacc TEST=./google/services/container TESTARGS='-run=TestAccContainerNodePool'
```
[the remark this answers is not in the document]
wyardley: oops, missed that copy/pasta. fixed now.

shuyama1 (approved):
Thank you!

PR #12802, opened 2025-01-18
Add inputs for filtering service accounts with data source google_service_accounts

bestefreund:
Add inputs `prefix` and `regex`  for filtering service accounts with data source `google_service_accounts`

Fixes https://github.com/hashicorp/terraform-provider-google/issues/20928

```release-note:enhancement
iam: added `prefix` and `regex` fields to `google_service_accounts` data source
```

### DIFF
--- a/mmv1/third_party/terraform/services/resourcemanager/data_source_google_service_accounts.go
+++ b/mmv1/third_party/terraform/services/resourcemanager/data_source_google_service_accounts.go
@@ -5,11 +5,13 @@ package resourcemanager
 import (
 	"context"
 	"fmt"
+	"regexp"
 	"strings"
 
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 	"github.com/hashicorp/terraform-provider-google/google/tpgresource"
 	transport_tpg "github.com/hashicorp/terraform-provider-google/google/transport"
+	"github.com/hashicorp/terraform-provider-google/google/verify"
 	"google.golang.org/api/iam/v1"
 )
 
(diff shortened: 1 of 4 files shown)


mmv1/third_party/terraform/services/resourcemanager/data_source_google_service_accounts.go
```diff
@@ -17,10 +18,18 @@ func DataSourceGoogleServiceAccounts() *schema.Resource {
...
+				Type:     schema.TypeString,
+				Optional: true,
+			},
 			"project": {
 				Type:     schema.TypeString,
 				Optional: true,
 			},
+			"regex": {
+				Type:     schema.TypeString,
+				Optional: true,
```
ScottSuarez, line 31: nit, you could add a validation func here to ensure the REGEX compiles. 

https://github.com/hashicorp/terraform-plugin-sdk/blob/b11308dc8f0ff4a03f750ba012e7bc7d3974e895/helper/schema/schema.go#L342-L349

ScottSuarez (commented):
small suggestion

mmv1/third_party/terraform/services/resourcemanager/data_source_google_service_accounts.go
bestefreund: It's a good suggestion.
For me, to check, if a regex compiles is a repeatable task.

I'd like to add a function to the [`github.com/hashicorp/terraform-provider-google/google/verify`](https://github.com/hashicorp/terraform-provider-google/blob/main/google/verify/validation.go) package and use it here.

ScottSuarez (approved):

PR #11439, opened 2024-08-13
Enable goog-terraform-provisioned provider label

SirGitsalot:
Switch the `goog-terraform-provisioned` label added in [v5.0.0](https://github.com/GoogleCloudPlatform/magic-modules/pull/9808) to be the default behavior. Fixes [#18614](https://github.com/hashicorp/terraform-provider-google/issues/18614).

This PR is starting out in draft form to locate affected tests (of which there **many**). I'll request a review once all the tests broken by this PR are running again.

**Release Note Template for Downstream PRs (will be copied)**

```release-note:breaking-change
changed provider labels to add the `goog-terraform-provisioned: true` label by default.
```

PR #12805, opened 2025-01-19
container: fixed `TestAccContainerNodePool_defaultDriverInstallation`

wyardley:
Remove explicit version on the nodepool; this is unnecessary, and the cluster itself is already using a version based on `google_container_engine_versions`

Fixes hashicorp/terraform-provider-google#20960

```release-note:none
```

### DIFF
--- a/mmv1/third_party/terraform/services/container/resource_container_node_pool_test.go.tmpl
+++ b/mmv1/third_party/terraform/services/container/resource_container_node_pool_test.go.tmpl
@@ -5073,7 +5073,6 @@ resource "google_container_node_pool" "np" {
   location           = "us-central1-a"
   cluster            = google_container_cluster.cluster.name
   initial_node_count = 2
-  version            = "1.30.1-gke.1329003"
 
   node_config {
     service_account = "default"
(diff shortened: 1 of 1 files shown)


shuyama1:
Test `TestAccContainerNodePool_defaultDriverInstallation` is skipped in VCR. Triggered the independent test on [TC](https://hashicorp.teamcity.com/buildConfiguration/TerraformProviders_GoogleCloud_GOOGLE_BETA_MMUPSTREAMTESTS_GOOGLEBETA_PACKAGE_CONTAINER/311042). I'll update the test result after it's finished

wyardley:
> Test TestAccContainerNodePool_defaultDriverInstallation is skipped in VCR. Triggered an independent test on TC.

Seems like #11439 is where a bunch of tests got VCR skipped, but wasn't totally clear from the comments there why (something related to the default labels changes?), and there's no comment next to `acctest.SkipIfVcr()` either. Do you know if it's because of resources / runtime, or because of something non-deterministic that actually breaks VCR tests?

I tried doing a quick run without that line and with VCR, just to see if it could be removed, at least theoretically, but currently getting stockout issues.

shuyama1:
> > Test TestAccContainerNodePool_defaultDriverInstallation is skipped in VCR. Triggered an independent test on TC.
> 
> Seems like #11439 is where a bunch of tests got VCR skipped, but wasn't totally clear from the comments there why (something related to the default labels changes?), and there's no comment next to `acctest.SkipIfVcr()` either. Do you know if it's because of resources / runtime, or because of something non-deterministic that actually breaks VCR tests?
> 
> I tried doing a quick run without that line and with VCR, just to see if it could be removed, at least theoretically, but currently getting stockout issues.

yeah, I don't see a specific reason for skipping this test in VCR based on its configuration. Let's remove `acctest.SkipIfVcr(t)` from the test and see if it works.

shuyama1 (approved):

PR #12807, opened 2025-01-19
resourcemanager: skipped checking `display_name` for test

wyardley:
`display_name` is optional, so unless this test bootstraps a new project and creates service accounts in it, there could be situations where one of the service accounts returned from the default test project doesn't have a display name.

Fixes hashicorp/terraform-provider-google#20258

```release-note:none

```


wyardley:
The main affected test worked for me locally - I _think_ `TestAccEphemeralServiceAccountKey_basic` is unrelated?

That said, there is a downside of this approach, namely that an actual issue with display_name getting pulled could be hidden. I don't know if the juice is worth the squeeze in terms of actually bootstrapping a project and creating a set of service accounts to be able to test in a more predictable way, though I suspect maybe not?

c2thorn:
> That said, there is a downside of this approach, namely that an actual issue with display_name getting pulled could be hidden. I don't know if the juice is worth the squeeze in terms of actually bootstrapping a project and creating a set of service accounts to be able to test in a more predictable way, though I suspect maybe not?

doesn't seem that big of a downside to me, thanks for looking into this

c2thorn (approved):

PR #12808, opened 2025-01-20
fix: Router advertised-route-priority undefined behavior 

akshat-jindal-nit:
b/386726875
Issue: advertised-route-priority is an optional field and by current terraform configs user can't set the value of advertised-route-priority to 0.

```release-note:bug
compute: made `google_compute_router_peer.advertised_route_priority` use server-side default if unset. To set the value to `0` you must also set `zero_advertised_route_priority = true`.
```

```release-note:enhancement
compute: added `zero_advertised_route_priority` field to 'google_compute_router_peer'
```

### DIFF
--- a/mmv1/third_party/terraform/services/compute/resource_compute_router_bgp_peer_test.go.tmpl
+++ b/mmv1/third_party/terraform/services/compute/resource_compute_router_bgp_peer_test.go.tmpl
@@ -3,7 +3,7 @@ package compute_test
 import (
 	"fmt"
 	"testing"
-  "regexp"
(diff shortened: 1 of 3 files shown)


akshat-jindal-nit:
Similar fix is raised for the custom_learned_route_priority https://github.com/GoogleCloudPlatform/magic-modules/pull/12355

melinath:
the contributor-membership-checker failure is because I had already approved the downstream-generation-and-test. :-p

melinath (approved):

PR #12814, opened 2025-01-20
Change Threat Prevention Profile's properties to sets.

duvni:
When creating a Security Profile resource the user can specify overrides for their threat prevention profile.
These overrides are of type Array, and so will show diffs between desired state and actual state if the order of the overrides changes.
However, the overrides are supposed to be agnostic to their order, so this PR fixes this by setting explicitly marking the overrides as Sets.

```release-note:bug
networksecurity: fixed sporadic-diff in `google_network_security_security_profile`
```

### DIFF
--- a/mmv1/products/networksecurity/SecurityProfile.yaml
+++ b/mmv1/products/networksecurity/SecurityProfile.yaml
@@ -111,6 +111,7 @@ properties:
     properties:
       - name: 'severityOverrides'
         type: Array
+        is_set: true
         description: The configuration for overriding threats actions by severity match.
         item_type:
           type: NestedObject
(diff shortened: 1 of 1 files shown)


melinath:

melinath:
reassigning due to OOO, sorry about the confusion!

duvni:
@roaks3, replying again since reviewer has changed and you were re-assigned to this PR.
We can add `override-breaking-change` label to the PR to address the breaking change test.
See my comment above for context.

roaks3 (commented):
Hey, sorry for the delay. I haven't posted an update quite yet because I'm discussing with the team about whether or not we want to consider this a breaking change. Will post back shortly.

roaks3 (approved):
Confirmed this is fine to merge into a minor, primarily because it is both unordered when returned from the API and volatile.

duvni:
> Confirmed this is fine to merge into a minor, primarily because it is both unordered when returned from the API and volatile.

Thanks a lot, appreciate the help with this.

PR #12817, opened 2025-01-21
Add analyticsConfig block to `google_apigee_addons_config`

sam-tombury:
Adds `analyticsConfig` block to `google_apigee_addons_config`  resource. Matches the pattern from other blocks in this resource, and includes all available (output-only) fields.

Fields and descriptions taken from [the REST API documentation](https://cloud.google.com/apigee/docs/reference/apis/apigee/rest/v1/organizations#analyticsconfig).

```release-note:enhancement
apigee: added nested `analyticsConfig` block to `google_apigee_addons_config` resource
```

### DIFF
--- a/mmv1/products/apigee/AddonsConfig.yaml
+++ b/mmv1/products/apigee/AddonsConfig.yaml
@@ -122,3 +122,28 @@ properties:
               Flag that specifies whether the Advanced API Ops add-on is
               enabled.
             output: true
+      - name: 'analyticsConfig'
+        type: NestedObject
+        description: Configuration for the Analytics add-on.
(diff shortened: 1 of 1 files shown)


hao-nan-li (changes requested):
Could you add these added fields to any of the existing tests?

sam-tombury:
Looks like there aren't any tests for this resource, but it turns out there's a [different resource](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/apigee_environment_addons_config) which manages this field on an environment level (not sure what the organization-level field controls). Closing this accordingly

PR #11234, opened 2024-07-23
Move bigqueryanalyticshub beta to v1

obada-ab:
Move bigqueryanalyticshub beta to v1

b/347844459

**Release Note Template for Downstream PRs (will be copied)**

```release-note:none

```

PR #9592, opened 2023-12-06
Add support for Analytics Hub Subscriptions 

Cidan:
This CL adds support for Analytics Hub Subscriptions as a resource, which allows users to subscribe to Analytics Hub Listings.

**Release Note Template for Downstream PRs (will be copied)**

```release-note:new-resource
`google_bigquery_analytics_hub_listing_subscription`
```

PR #12820, opened 2025-01-21
bigquery analyticshub listing subscription

c2thorn:
fixes https://github.com/hashicorp/terraform-provider-google/issues/15144

- adapting https://github.com/GoogleCloudPlatform/magic-modules/pull/9592. Simplified the creation process and removed the custom decoder. tested manually as well.
- The previous PR was blocked due to the v1beta1 API deprecation, but since https://github.com/GoogleCloudPlatform/magic-modules/pull/11234, we no longer use it.
- This resource has a non-typical [creation URL that uses a "subscribe" action on a separate resource](https://cloud.google.com/bigquery/docs/reference/analytics-hub/rest/v1/projects.locations.dataExchanges.listings/subscribe). This PR only implements the `destinationDataset` part of subscribing, as `destinationPubsubSubscription` is not GA yet.

```release-note:new-resource
`google_bigquery_analytics_hub_listing_subscription`
```


c2thorn:
assigning the old reviewer from https://github.com/GoogleCloudPlatform/magic-modules/pull/9592

c2thorn:

c2thorn:
reassigning since @NickElliot is busy with oncall

mmv1/products/bigqueryanalyticshub/Subscription.yaml
```diff
@@ -0,0 +1,150 @@
...
+      The resource name of the subscription. e.g. "projects/myproject/locations/US/subscriptions/123"
+    output: true
+  - name: 'subscription_id'
+    type: String
+    description: |-
+      The subscription id used to reference the subscription.
+    output: true
+    ignore_read: true
+  - name: 'creation_time'
+    type: String
```
wj-chen, line 117: Should this and `last_modify_time` be `type: Time`?

mmv1/products/bigqueryanalyticshub/Subscription.yaml
```diff
@@ -0,0 +1,150 @@
...
+  - name: 'listing_id'
+    type: String
+    description: |-
+      The ID of the listing. Must contain only Unicode letters, numbers (0-9), underscores (_). Should not use characters that require URL-escaping, or characters outside of ASCII, spaces.
+    url_param_only: true
+    required: true
+  - name: 'location'
+    type: String
+    description: |
+      The name of the location this data exchange subscription.
```
wj-chen, line 60: "The name of the location for this subscription."

mmv1/products/bigqueryanalyticshub/Subscription.yaml
```diff
@@ -0,0 +1,150 @@
...
+        type: KeyValuePairs
+        description: |
+          The labels associated with this dataset. You can use these to
+          organize and group your datasets.
+  - name: 'name'
+    type: String
+    description: |-
+      The resource name of the subscription. e.g. "projects/myproject/locations/US/subscriptions/123"
+    output: true
+  - name: 'subscription_id'
```
wj-chen, line 110: Some other property names are in camel case, should we be consistent?

mmv1/products/bigqueryanalyticshub/Subscription.yaml
```diff
@@ -0,0 +1,150 @@
...
+    required: true
+  - name: 'location'
+    type: String
+    description: |
+      The name of the location this data exchange subscription.
+    url_param_only: true
+    required: true
+    custom_flatten: 'templates/terraform/custom_flatten/bigquery_dataset_location.go.tmpl'
+    diff_suppress_func: 'tpgresource.CaseDiffSuppress'
+properties:
```
wj-chen, line 65: Looks like missing two output-only `linkedDatasetMap` and `linkedResources`

mmv1/products/bigqueryanalyticshub/Subscription.yaml
```diff
@@ -0,0 +1,150 @@
...
+---
+name: 'Subscription'
+description: A Bigquery Analytics Hub subscription
+references:
+  guides:
+    'Official Documentation': 'https://cloud.google.com/bigquery/docs/analytics-hub-introduction'
+  api: 'https://cloud.google.com/bigquery/docs/reference/analytics-hub/rest/v1/projects.locations.subscriptions'
+base_url: 'projects/{{project}}/locations/{{location}}/subscriptions'
+self_link: 'projects/{{project}}/locations/{{location}}/subscriptions/{{subscription_id}}'
+create_url: 'projects/{{project}}/locations/{{location}}/dataExchanges/{{data_exchange_id}}/listings/{{listing_id}}:subscribe'
```
wj-chen, line 23: I think it can be created from either https://cloud.google.com/bigquery/docs/reference/analytics-hub/rest/v1/projects.locations.dataExchanges/subscribe or https://cloud.google.com/bigquery/docs/reference/analytics-hub/rest/v1/projects.locations.dataExchanges.listings/subscribe, depending on whether any listing ID is given

c2thorn: I see, now I see why in https://github.com/GoogleCloudPlatform/magic-modules/pull/9592 they went with "listing_subscription" instead of just subscription. I should probably change this PR to follow the same.

@wj-chen do you have any insight into if the listing subscription is the more common usage? Do they both result in the same resource? It's basically not heard of for a Terraform resource to have separate create URLs, so I'm thinking they need to be two different resources

wj-chen: > do you have any insight into if the listing subscription is the more common usage?

I think someone from the BQ Analytics Hub API team that you've been in conversation with can better answer this.

> Do they both result in the same resource?

I believe so, based on the API reference docs - that one can `subscribe` to an data exchange or a listing, and the subscription resource itself will have an oneof of data exchange ID and listing ID - but probably still good to confirm with the API team.

> I'm thinking they need to be two different resources

Is there an option to hand-write the Create for this resource? From customer's perspective it would be less confusing if the Terraform resource aligns closely with the API resource to be a single one, so they don't need to manage two resources that have an almost identical surface, and we don't need to maintain two resources that have almost an identical surface.

c2thorn: @wj-chen I tried implementing the data exchange endpoint and had success implementation-wise, but got permission errors actually testing it out. I didn't get any errors for the listing endpoint and the guides have no instructions on subscribing to data exchanges directly, which makes me think this is a lesser used feature. I'm waiting on feedback from the service team to decide whether or not to include it. We could also include it at a later time if there is customer demand.

mmv1/products/bigqueryanalyticshub/Subscription.yaml
c2thorn: added

mmv1/products/bigqueryanalyticshub/Subscription.yaml
c2thorn: made them consistent

mmv1/products/bigqueryanalyticshub/Subscription.yaml
c2thorn: After talking with the service team, while it is the same resulting resource, the creation URL, request body, and UX are different enough to separate them out into "listing_subscription" and "dcr_subscription"

This PR will only handle the former

hao-nan-li:
> 
> ## Diff report
> 
> `google` provider: [Diff](https://github.com/modular-magician/terraform-provider-google/compare/auto-pr-12820-old..auto-pr-12820) ( 6 files changed, 1240 insertions(+), 2 deletions(-)) `google-beta` provider: [Diff](https://github.com/modular-magician/terraform-provider-google-beta/compare/auto-pr-12820-old..auto-pr-12820) ( 6 files changed, 1240 insertions(+), 2 deletions(-)) `terraform-google-conversion`: [Diff](https://github.com/modular-magician/terraform-google-conversion/compare/auto-pr-12820-old..auto-pr-12820) ( 1 file changed, 169 insertions(+)) Open in Cloud Shell: [Diff](https://github.com/modular-magician/docs-examples/compare/auto-pr-12820-old..auto-pr-12820) ( 4 files changed, 144 insertions(+))
> 
> ## Missing test report
> 
> Resource: `google_bigquery_analytics_hub_listing_subscription` (1 total tests) Please add an acceptance test which includes these fields. The test should include the following:
> 
> ```hcl
> resource "google_bigquery_analytics_hub_listing_subscription" "primary" {
>   linked_dataset_map {
>     resource_name = # value needed
>   }
> }
> ```

Looks like `linked_dataset_map` is an `output` field, but on the downstream `resource_name` is marked as required. May need to check if that cause a diff.

c2thorn:
> Looks like `linked_dataset_map` is an `output` field, but on the downstream `resource_name` is marked as required. May need to check if that cause a diff. Otherwise LGTM

verified that there is no diff detected after that field is populated. So the parent `computed` field overrides the child field's required property.

PR #12826, opened 2025-01-22
Colab schedules

bcreddy-gcp:

```release-note:new-resource
`google_colab_schedule`
```


melinath (changes requested):

mmv1/products/colab/Schedule.yaml
```diff
@@ -0,0 +1,176 @@
...
+
+name: 'Schedule'
+description: |
+  'Colab Enterprise Notebook Execution Schedules.'
+references:
+  guides:
+    'Schedule a notebook run': 'https://cloud.google.com/colab/docs/schedule-notebook-run'
+  api: 'https://cloud.google.com/vertex-ai/docs/reference/rest/v1/projects.locations.schedules'
+base_url: 'projects/{{project}}/locations/{{location}}/schedules'
+self_link: 'projects/{{project}}/locations/{{location}}/schedules/{{name}}'
```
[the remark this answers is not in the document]
bcreddy-gcp: Used the custom_flatten to be consistent to other colab resources.

mmv1/products/colab/Schedule.yaml
```diff
@@ -0,0 +1,209 @@
...
+      start_time: '2014-10-02T15:01:23Z'
+      end_time: '2014-10-10T15:01:23Z'
+    test_env_vars:
+      project_id: 'PROJECT_NAME'
+      location: 'REGION'
+      service_account: 'SERVICE_ACCT'
+    test_vars_overrides:
+      key_name: 'acctest.BootstrapKMSKeyInLocation(t, "us-central1").CryptoKey.Name'
+      start_time: 'time.Now().AddDate(0, 0, 1).Format(time.RFC3339)'
+      end_time: 'time.Now().AddDate(0, 0, 10).Format(time.RFC3339)'
```
melinath, line 76: here & for the handwritten tests - the tests need to use (more) static times to be friendly to VCR recording & replaying, since that will compare the body of the HTTP request.

I assume that the reason you're doing that is that these need to both be times in the future. You could work around the by making it at least be invalidated less frequently - for example, you could make sure it's always from the first day of month after next to the first day of the month after that. (Or the first day of next year to the first day of the year after - even better.) Then the cassette would only be invalidated occasionally.

melinath (changes requested):

mmv1/products/colab/Schedule.yaml
bcreddy-gcp: Changed the start time it to first day of the next year.

mmv1/products/colab/Schedule.yaml
```diff
@@ -0,0 +1,209 @@
...
+      Output only. The state of the schedule.
+    output: true
+  - name: 'createNotebookExecutionJobRequest'
+    immutable: true
+    type: NestedObject
+    description: Request for google_colab_notebook_execution.
+    exactly_one_of:
+      - create_notebook_execution_job_request
+    properties:
+      - name: 'parent'
```
madhusuraj, line 143: Should this be required, unless we have some logic to set the same project/location as the schedule?

madhusuraj (approved):

mmv1/products/colab/Schedule.yaml
```diff
@@ -0,0 +1,209 @@
...
+          - name: 'gcsOutputUri'
+            required: true
+            type: String
+            description: 'The Cloud Storage location to upload the result to. Format:`gs://bucket-name`'
+          - name: 'executionUser'
+            type: String
+            exactly_one_of:
+              - execution_user
+              - service_account
+            description: 'The user email to run the execution as. Only supported by Colab runtimes.'
```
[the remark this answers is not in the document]
bcreddy-gcp: Done

mmv1/products/colab/Schedule.yaml
```diff
@@ -0,0 +1,209 @@
...
+    type: NestedObject
+    description: Request for google_colab_notebook_execution.
+    exactly_one_of:
+      - create_notebook_execution_job_request
+    properties:
+      - name: 'parent'
+        type: String
+        description:
+          Notebook Execution Job Info.
+      - name: 'notebookExecutionJob'
```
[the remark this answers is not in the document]
bcreddy-gcp: Done.

mmv1/products/colab/Schedule.yaml
bcreddy-gcp: Made it required. Can look into QOL improvements at a later point.

mmv1/products/colab/Schedule.yaml
```diff
@@ -0,0 +1,209 @@
...
+    immutable: true
+    type: NestedObject
+    description: Request for google_colab_notebook_execution.
+    exactly_one_of:
+      - create_notebook_execution_job_request
+    properties:
+      - name: 'parent'
+        type: String
+        description:
+          Notebook Execution Job Info.
```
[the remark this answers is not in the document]
bcreddy-gcp: @melinath thoughts?

mmv1/products/colab/Schedule.yaml
```diff
@@ -0,0 +1,209 @@
...
+    default_from_api: true
+parameters:
+  - name: 'location'
+    type: String
+    required: true
+    url_param_only: true
+    description: 'The location for the resource: https://cloud.google.com/colab/docs/locations'
+properties:
+  - name: 'name'
+    type: String
```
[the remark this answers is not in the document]
madhusuraj: Oh sounds good, missed that!

mmv1/products/colab/Schedule.yaml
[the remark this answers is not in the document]
melinath: Yes, if this field is present, it looks like that's the format the API expects. However, having this field feels a little redundant, since the project and location are already present in the URL (and I assume they have to always match.) There are a few ways you could handle this elegantly:

1. Use the `parent` field in the URLs instead of project and location. (I'd recommend against this because the parent isn't a "real" resource that you'd manage with Terraform.)
2. Make the API infer the parent from the request body if unspecified. (This seems best to me but would require changing the API.)
3. Use an [encoder](https://googlecloudplatform.github.io/magic-modules/develop/custom-code/#encoder) to automatically populate this in the request body based on the project & location fields. (This is what I'd recommend for a client-side solution.)

mmv1/templates/terraform/post_create/colab_schedule.go.tmpl
```diff
@@ -0,0 +1,22 @@
+// The response for create request contains the generated name generated name that we need
+// in order to perform a READ. We need to access the object inside of it as
+// a map[string]interface, so let's do that.
+
+longName := res["name"].(string)
+name := tpgresource.GetResourceNameFromSelfLink(longName)
+log.Printf("[DEBUG] Setting resource name to %s", name)
+if err := d.Set("name", name); err != nil {
+	return fmt.Errorf("Error setting name: %s", err)
+}
```
melinath, line 10: Why is this in the post_create code instead of in a flattener for the name field?

mmv1/templates/terraform/post_create/colab_schedule.go.tmpl
```diff
@@ -0,0 +1,22 @@
...
+
+parts := strings.Split(longName, "/")
+parts[1] = project
+updatedLongName := strings.Join(parts, "/")
+d.SetId(updatedLongName)
```
melinath, line 15: This seems to duplicate the standard ID setting logic that's already included - https://github.com/modular-magician/terraform-provider-google-beta/compare/auto-pr-12826-old..auto-pr-12826#diff-ebbc22d7dac2eff1d25f563ec1e7f85691e2e9b1aaf23efe1a2f61918a1507f8R340

```suggestion
```

mmv1/templates/terraform/constants/colab_schedule.go.tmpl
```diff
@@ -0,0 +1,18 @@
+func modifyScheduleState(config *transport_tpg.Config, d *schema.ResourceData, project string, billingProject string, userAgent string, state string) (map[string]interface{}, error) {
+	url, err := tpgresource.ReplaceVars(d, config, "{{"{{"}}ColabBasePath{{"}}"}}projects/{{"{{"}}project{{"}}"}}/locations/{{"{{"}}location{{"}}"}}/schedules/{{"{{"}}name{{"}}"}}:"+state)
+	if err != nil {
+		return nil, err
+	}
+
+	res, err := transport_tpg.SendRequest(transport_tpg.SendRequestOptions{
```
melinath, line 7: Will this wait for the state to change on the API side?

mmv1/third_party/terraform/services/colab/resource_colab_schedule_test.go
```diff
@@ -0,0 +1,539 @@
...
+				Config: testAccColabSchedule_full(context),
+			},
+			{
+				ResourceName:            "google_colab_schedule.schedule",
+				ImportState:             true,
+				ImportStateVerify:       true,
+				ImportStateVerifyIgnore: []string{"location"},
+			},
+			{
+				Config: testAccColabSchedule_update(context),
```
melinath, line 42: As documented at https://googlecloudplatform.github.io/magic-modules/test/test/#add-an-update-test, you can add a check here to ensure that the update step isn't accidentally deleting the resource:

```suggestion
				Config: testAccColabSchedule_update(context),
				ConfigPlanChecks: resource.ConfigPlanChecks{
					PreApply: []plancheck.PlanCheck{
						plancheck.ExpectResourceAction("google_colab_schedule.schedule", plancheck.ResourceActionUpdate),
					},
				},
```

melinath (changes requested):

mmv1/templates/terraform/constants/colab_schedule.go.tmpl
bcreddy-gcp: These methods are not long running. The state change is instantaneous.

mmv1/products/colab/Schedule.yaml
```diff
@@ -0,0 +1,211 @@
...
+      start_time: 'time.Date(time.Now().Year(), 12, 31, 0, 0, 0, 0, time.Now().Location()).AddDate(0, 0, 1).Format(time.RFC3339)'
+      end_time: 'time.Date(time.Now().Year(), 12, 31, 0, 0, 0, 0, time.Now().Location()).AddDate(0, 0, 10).Format(time.RFC3339)'
+    ignore_read_extra:
+      - 'desired_state'
+virtual_fields:
+  - name: 'desired_state'
+    description: |
+      Desired state of the Colab Schedule. Set this field to `ACTIVE` to start/resume the schedule, and `PAUSED` to pause the schedule.
+    type: String
+    default_from_api: true
```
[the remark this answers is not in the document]
bcreddy-gcp: Doesn't the paused example need ignore_read_extra?

mmv1/templates/terraform/post_create/colab_schedule.go.tmpl
bcreddy-gcp: Because name is an output only field that we extract from response. Let me know if I am missing something.

mmv1/templates/terraform/post_create/colab_schedule.go.tmpl
bcreddy-gcp: Without this, I see

  error=
  | exit status 1
  | 
  | Error: Import id "projects/bcreddy-sandbox/locations/us-central1/schedules/" doesn't match any of the accepted formats: [^projects/(?P<project>[^/]+)/locations/(?P<location>[^/]+)/schedules/(?P<name>[^/]+)$ ^(?P<project>[^/]+)/(?P<location>[^/]+)/(?P<name>[^/]+)$ ^(?P<location>[^/]+)/(?P<name>[^/]+)$]
  | 
  | 
   test_step_number=2

mmv1/products/colab/Schedule.yaml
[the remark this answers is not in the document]
melinath: ah, yeah, I guess it would because the field would have a non-default value in config.

mmv1/templates/terraform/post_create/colab_schedule.go.tmpl
melinath: Could you remove this & the code above so that I can look over the failure logs more easily? In a general sense, I can say that error means that the Id isn't being set in state properly, but I can't say for sure why. (But I am fairly confident that the code I'm asking to remove here is unnecessary.)

melinath (changes requested):

melinath:
well that's interesting

melinath:
@bcreddy-gcp Thanks for putting it in this state. The reason this is failing is due to a bug in the generator - I've opened https://github.com/GoogleCloudPlatform/magic-modules/pull/12939 to fix it. If you prefer to move forward with the previous workaround, I'm happy to do that (since there are already a number of post_creates that work around the same problem).

melinath (changes requested):
just a couple small things I noticed while looking into the error

bcreddy-gcp:
Added back the post_create to not wait for the fix @melinath.

mmv1/templates/terraform/post_create/colab_schedule.go.tmpl
melinath: closing the loop here as well - this is required (at least for now) due to https://github.com/GoogleCloudPlatform/magic-modules/pull/12826#issuecomment-2628607697

mmv1/third_party/terraform/services/colab/resource_colab_schedule_test.go
```diff
@@ -0,0 +1,566 @@
...
+        }
+      },
+      "nbformat": 4,
+      "nbformat_minor": 4
+    }
+    EOF
+}
+
+resource "google_colab_schedule" "schedule" {
+  display_name = "Notebook Schedule full"
```
melinath, line 245:
```suggestion
  display_name = "tf-test-schedule%{random_suffix}"
```

mmv1/third_party/terraform/services/colab/resource_colab_schedule_test.go
```diff
@@ -0,0 +1,566 @@
...
```
melinath, line 341:
```suggestion
  display_name = "tf-test-schedule%{random_suffix}"
```

mmv1/third_party/terraform/services/colab/resource_colab_schedule_test.go
```diff
@@ -0,0 +1,566 @@
...
```
melinath, line 439:
```suggestion
  display_name = "tf-test-schedule%{random_suffix}"
```

mmv1/third_party/terraform/services/colab/resource_colab_schedule_test.go
```diff
@@ -0,0 +1,566 @@
...
+        }
+      },
+      "nbformat": 4,
+      "nbformat_minor": 4
+    }
+    EOF
+}
+
+resource "google_colab_schedule" "schedule" {
+  display_name = "Notebook Schedule updated"
```
melinath, line 537:
```suggestion
  display_name = "tf-test-schedule-updated%{random_suffix}"
```

mmv1/third_party/terraform/services/colab/resource_colab_schedule_test.go
```diff
@@ -0,0 +1,566 @@
...
+				ImportStateVerifyIgnore: []string{"location", "desired_state"},
+			},
+		},
+	})
+}
+
+func testAccColabSchedule_full(context map[string]interface{}) string {
+	return acctest.Nprintf(`
+resource "google_colab_runtime_template" "my_runtime_template" {
+  name = ""
```
melinath, line 183:
```suggestion
  name = "tf-test-runtime-template%{random_suffix}"
```

mmv1/third_party/terraform/services/colab/resource_colab_schedule_test.go
```diff
@@ -0,0 +1,566 @@
...
+    google_storage_bucket.output_bucket,
+  ]
+}
+`, context)
+}
+
+func testAccColabSchedule_paused(context map[string]interface{}) string {
+	return acctest.Nprintf(`
+resource "google_colab_runtime_template" "my_runtime_template" {
+  name = ""
```
melinath, line 279:
```suggestion
  name = "tf-test-runtime-template%{random_suffix}"
```

mmv1/third_party/terraform/services/colab/resource_colab_schedule_test.go
```diff
@@ -0,0 +1,566 @@
...
+    google_storage_bucket.output_bucket,
+  ]
+}
+`, context)
+}
+
+func testAccColabSchedule_active(context map[string]interface{}) string {
+	return acctest.Nprintf(`
+resource "google_colab_runtime_template" "my_runtime_template" {
+  name = ""
```
melinath, line 377:
```suggestion
  name = "tf-test-runtime-template%{random_suffix}"
```

mmv1/third_party/terraform/services/colab/resource_colab_schedule_test.go
```diff
@@ -0,0 +1,566 @@
...
+    google_storage_bucket.output_bucket,
+  ]
+}
+`, context)
+}
+
+func testAccColabSchedule_update(context map[string]interface{}) string {
+	return acctest.Nprintf(`
+resource "google_colab_runtime_template" "my_runtime_template" {
+  name = ""
```
melinath, line 475:
```suggestion
  name = "tf-test-runtime-template%{random_suffix}"
```

melinath (changes requested):
a couple small tweaks to the update tests to make the resources sweepable, otherwise LGTM

melinath (approved):

melinath:
note: merging without tests for execution_user because I ended up using that field to run tests locally, so I have confidence it maps properly. However, it would be good to add tests using it as a follow-up.

PR #12829, opened 2025-01-22
Make (R)IGM.PerInstanceConfigs lock on Instance name, not IGM

askubis:
Allow parallelization of per instance config related operations by not locking on the IGM
Fixes b/270736647

```release-note:enhancement
compute: Allow parallelization of google_compute_(region_)per_instance_config by not locking on the parent resource, but including instance name.
```


BBBmau (approved):
approving since this follows the mutex format in docs: https://googlecloudplatform.github.io/magic-modules/reference/resource/#mutex

PR #12836, opened 2025-01-22
Fix Spanner Instance `force_delete` backup bug

Noremac201:
Previously `userAgent` and `billingProjectId` were swapped causing the project to be incorrect when `force_delete` is enabled and backups exist in the Spanner instance.

Fixes https://github.com/hashicorp/terraform-provider-google/issues/21007

```release-note:bug
spanner: fix bug when `force_destroy` enabled on instance that was using wrong project ID to attempt to delete backups
```


rileykarson (approved):
Good catch- LGTM, thanks!

PR #12843, opened 2025-01-23
upgrade upload artifact to v4

ScottSuarez:
https://github.blog/changelog/2024-04-16-deprecation-notice-v3-of-the-artifact-actions/

https://github.com/GoogleCloudPlatform/magic-modules/actions/runs/12935219791/job/36078069848

```release-note:none

```

### DIFF
--- a/.github/workflows/build-downstream.yml
+++ b/.github/workflows/build-downstream.yml
@@ -108,7 +108,7 @@ jobs:
           (current_dir=$(pwd) && cd $OUTPUT_PATH && zip -r "$current_dir/output.zip" .)
 
       - name: Upload built artifacts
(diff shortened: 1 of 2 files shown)


ScottSuarez:
> I think we may need to take some extra care to ensure we're handling the immutable artifacts correctly per [actions/upload-artifact@`main`/docs/MIGRATION.md](https://github.com/actions/upload-artifact/blob/main/docs/MIGRATION.md). FYI I've created [hashicorp/terraform-provider-google#21016](https://github.com/hashicorp/terraform-provider-google/issues/21016) to track this.

Shouldn't be an compatibility issue for use, we don't currently write the same artifacts multiple times within a workflow.

melinath (approved):
ah, if the artifacts have separate namespaces per workflow / run then that should be fine.

PR #12846, opened 2025-01-23
fix format for non-exercised tests comment

shuyama1:

```release-note:none

```


.ci/magician/cmd/templates/vcr/non_exercised_tests.tmpl
```diff
@@ -1,4 +1,5 @@
 {{- if or (gt (len .NotRunBetaTests) 0) (gt (len .NotRunGATests) 0)}}
```
wyardley, line 1: not 100% sure, but might also work to remove the `-` on line 1 and not add the literal space?
```suggestion
{{ if or (gt (len .NotRunBetaTests) 0) (gt (len .NotRunGATests) 0)}}
```

PR #12856, opened 2025-01-24
Fix Firestore field deletion for wildcard fields

rwhogg:
You can update index configs for wildcard fields, but you can't update TTL configs. When we "delete" SFIs, we actually issue a PATCH with empty index and TTL configs. We put both indexConfig and ttlConfig in the update mask for this PATCH accordingly.

For wildcard fields, simply exclude ttlConfig from the update mask when issuing this PATCH to avoid the "unsupported" error.

Fixes https://github.com/hashicorp/terraform-provider-google/issues/20397

```release-note:bug
firestore: fixed deletion of wildcard fields in `google_firestore_field`
```

### DIFF
--- a/mmv1/products/firestore/Field.yaml
+++ b/mmv1/products/firestore/Field.yaml
@@ -86,6 +86,15 @@ examples:
       project_id: 'PROJECT_NAME'
     test_vars_overrides:
       'delete_protection_state': '"DELETE_PROTECTION_DISABLED"'
+  - name: 'firestore_field_wildcard'
+    primary_resource_id: 'wildcard'
+    vars:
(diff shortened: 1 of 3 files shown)


melinath (commented):
Since this is update-specific behavior, would it make sense to add (or modify) an update test instead of adding a new create test? https://googlecloudplatform.github.io/magic-modules/test/test/#add-an-update-test

rwhogg:
@melinath it's not really update-specific behavior. The bug it's fixing is about an inability to delete.

Firestore doesn't _really_ delete SFIs, it just resets them to the default, which is what happens when you destroy a firestore_field in Terraform.

PR #12857, opened 2025-01-24
Mark `content` in storage object content data source as computed

wafrelka:
Fixes https://github.com/hashicorp/terraform-provider-google/issues/15717 in a backward-compatible manner.

The bug was once fixed in #10778, but it is reverted due to backward incompatibility.
This pull request retries to fix the bug without affecting to existing configuration.

```release-note:bug
storage: fixed an issue where plans with a dependency on the `content` field in the `google_storage_bucket_object_content` data source could erroneously fail
```


slevenick:
Odd, so the issue here is that the field *should* be Computed and not Optional, but due to it being marked as Optional in the past we have to keep it that way?

wafrelka:
> Odd, so the issue here is that the field should be Computed and not Optional, but due to it being marked as Optional in the past we have to keep it that way?

Yes.

According to the [document](https://registry.terraform.io/providers/hashicorp/google/latest/docs/data-sources/storage_bucket_object_content), the field is supposed to be Computed and non-Optional.

slevenick (approved):

PR #12858, opened 2025-01-24
Fix error details type on google_database_migration_service_migration_job

kjvellajr:
In the event that there is an error with a MigrationJob in GCP, the Terraform google provider can not properly create a plan because the schema for `google_database_migration_service_migration_job` is incorrect.

`... expected type 'string', got unconvertible type 'map[string]interface {}' ...`

I am not sure if this is the proper way to fix the issue, but based on the details of the error response from Terraform and comparing `magic-modules` source code with `gcloud` output, it seems like this code change will fix the issue. `magic-modules` has a MigrationJob error configured as a `KeyValuePairs` when it is in fact an Object with arbitrary properties.

Terraform execution error:
```
Error: Error reading MigrationJob: error.0.details.1.metadata: '' expected type 'string', got unconvertible type 'map[string]interface {}', value: 'map[errorMessage:The migration job is currently running. Restart is not available on a running Migration job. nativeErrorMessage:Cloud SQL replica is already setup and it's replicating from the source database server. severity:ERROR]'
  with google_database_migration_service_migration_job.dms_task,
  on main.tf line 167, in resource "google_database_migration_service_migration_job" "dms_task":
 167: resource "google_database_migration_service_migration_job" "dms_task" {
```

Data returned from `gcloud database-migration migration-jobs describe dms-task`:
```
...
error:
  code: 9
  details:
  - '@type': type.googleapis.com/google.cloud.clouddms.v1.MigrationJobVerificationError
    errorCode: CANT_RESTART_RUNNING_MIGRATION
    errorDetailMessage: Cloud SQL replica is already setup and it's replicating from
      the source database server.
    errorMessage: "The migration job is currently running. Restart is not available\
      \ on a running Migration job. Details: Cloud SQL replica is already setup and\
      \ it's replicating from the source database server."
  - '@type': type.googleapis.com/google.rpc.ErrorInfo
    domain: datamigration.googleapis.com
    metadata:
      errorMessage: The migration job is currently running. Restart is not available
        on a running Migration job.
      nativeErrorMessage: Cloud SQL replica is already setup and it's replicating
        from the source database server.
      severity: ERROR
    reason: CANT_RESTART_RUNNING_MIGRATION
  - '@type': type.googleapis.com/google.rpc.Help
    links:
    - description: 'Learn more:'
      url: https://cloud.google.com/database-migration/docs/diagnose-issues
  message: The migration job is currently running. Restart is not available on a running
    Migration job.
labels:
  goog-terraform-provisioned: 'true'
...
```

```release-note:bug
fix: error details type on google_database_migration_service_migration_job
```

### DIFF
--- a/mmv1/products/databasemigrationservice/MigrationJob.yaml
+++ b/mmv1/products/databasemigrationservice/MigrationJob.yaml
@@ -155,6 +155,7 @@ properties:
         description: |
           A list of messages that carry the error details.
         output: true
+        custom_flatten: 'templates/terraform/custom_flatten/dms_migration_job_error_details.tmpl'
         item_type:
           type: KeyValuePairs
   - name: 'type'
--- a/mmv1/templates/terraform/custom_flatten/dms_migration_job_error_details.tmpl
+++ b/mmv1/templates/terraform/custom_flatten/dms_migration_job_error_details.tmpl
@@ -1,0 +1,36 @@
+{{/*
+	The license inside this block applies to this file
+	Copyright 2024 Google Inc.
+	Licensed under the Apache License, Version 2.0 (the "License");
+	you may not use this file except in compliance with the License.
+	You may obtain a copy of the License at http://www.apache.org/licenses/LICENSE-2.0
+	Unless required by applicable law or agreed to in writing, software
+	distributed under the License is distributed on an "AS IS" BASIS,
+	WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+	See the License for the specific language governing permissions and
+	limitations under the License.
+*/ -}}
+func flatten{{$.GetPrefix}}{{$.TitlelizeProperty}}(v interface{}, d *schema.ResourceData, config *transport_tpg.Config) interface{} {
+	if v == nil {
+		return nil
+	}
+	detailsArray := v.([]interface{})
+	for i, raw := range detailsArray {
+		m := raw.(map[string]interface{})
+		if len(m) < 1 {
+			// Do not include empty json objects coming back from the api
+			continue
+		}
+		for k, val := range m {
+			if _, ok := val.(string); !ok {
+				b, err := json.Marshal(v)
+				if err != nil {
+					return err
+				}
+				m[k] = string(b)
+			}
+		}
+		detailsArray[i] = m
+	}
+	return detailsArray
+}


mmv1/products/databasemigrationservice/MigrationJob.yaml
```diff
@@ -156,7 +156,9 @@ properties:
           A list of messages that carry the error details.
         output: true
         item_type:
-          type: KeyValuePairs
+          type: NestedObject
+          properties:
+            []
```
melinath, line 161:
```suggestion
          type: KeyValuePairs
          custom_flatten: 'templates/terraform/custom_flatten/dms_migration_job_error_details.tmpl'
```

melinath (changes requested):
I've looked into this a bit and it seems like this field on this resource doesn't have a consistent schema, which means that we need to return nested as a string (containing a JSON representation of the value). However, we also need to maintain backwards-compatibility with existing states / errors that _can_ be represented as key/value pairs. The easiest way to do this is to only marshal values as JSON if they can't already be converted to strings.

You'll need a [custom flattener](https://googlecloudplatform.github.io/magic-modules/develop/custom-code/#custom_flatten) - something like this:

```go
	if v == nil {
		return nil
	}
	m := v.(map[string]interface{})
	for k, val := range m {
		if _, ok := val.(string); !ok {
			b, err := json.Marshal(v)
			if err != nil {
				return err
			}
			m[k] = string(b)
		}
	}
	return string(b)
}
```

kjvellajr:
@melinath thanks for the reply! This sounds great, I have added the suggested changes to the pull request. Please let me know if there are any additional changes would like to see done.

melinath:
Is there any chance you could manually test this change locally & verify it fixes the issue for you? https://googlecloudplatform.github.io/magic-modules/test/run-tests/#optional-test-manually

mmv1/templates/terraform/custom_flatten/dms_migration_job_error_details.tmpl
```diff
@@ -0,0 +1,28 @@
...
+	for k, val := range m {
+		if _, ok := val.(string); !ok {
+			b, err := json.Marshal(v)
+			if err != nil {
+				return err
+			}
+			m[k] = string(b)
+		}
+	}
+	return string(b)
```
melinath, line 27: looks like I had a typo in my suggestion

```suggestion
	return m
```

melinath (changes requested):

mmv1/products/databasemigrationservice/MigrationJob.yaml
```diff
@@ -157,6 +157,7 @@ properties:
         output: true
         item_type:
           type: KeyValuePairs
+          custom_flatten: 'templates/terraform/custom_flatten/dms_migration_job_error_details.tmpl'
```
melinath, line 160: it looks like a custom_flatten at this level isn't honored - there's no diff in the providers - so you'll need to move it up into the main `details` field & modify the code to handle the fact that it's a list of maps (instead of an individual map)

kjvellajr:
> Is there any chance you could manually test this change locally & verify it fixes the issue for you? https://googlecloudplatform.github.io/magic-modules/test/run-tests/#optional-test-manually

I have the recent suggestions on my local. Yes, I'll work on building the provider locally and testing them before pushing.

kjvellajr:
@melinath Please see the latest changes. I have implemented your suggestions of 1) fixing the typo in the return from the suggestion 2) moving the `custom_flatten` to the proper level 3) using a list of maps instead of individual map.

I compiled the latest changes locally, following the docs you provided, and verified that a `terraform plan` fails with the versioned provider and is successful when using the new local compiled provider.

kjvellajr:
@melinath any updates on this PR? Are there any additional changes you would like me to make?

melinath:
sorry for the delay, I've been wanting to run some additional local tests but haven't had the time. I expect to verify this today or tomorrow (and I don't expect any issues, I'm just being paranoid.)

kjvellajr:
No problem, thanks for the update. I appreciate the scrutiny.

melinath:
Do you happen to have advice on how I can trigger an error?

kjvellajr:
In addition to the original error seen in the description, you could also set up a MigrationJob with a source connection profile connected to a proxy VM, then remove the proxy VM (or update your firewall rules to make it so the MigrationJob can no longer talk to the source DB) then the an error will appear on your MigrationJob.

melinath (approved):
yep, all looks good! Thanks!

PR #12863, opened 2025-01-24
Modify labeler to consider issues with more than 3 services impacted as cross-provider

roaks3:
Fixes https://github.com/hashicorp/terraform-provider-google/issues/20818

* I tested locally with some issue bodies and the results looked good
* I chose not to include the backfill job: basically, the backfill job ignores issues with `service/terraform`, so the condition mentioned here would only be triggered after a ticket is created with < 4 services, and then more are added later, which is both unusual, and something we can handle manually.

```release-note:none

```

### DIFF
--- a/tools/issue-labeler/cmd/compute_new_labels.go
+++ b/tools/issue-labeler/cmd/compute_new_labels.go
@@ -45,6 +45,21 @@ func execComputeNewLabels() error {
 	affectedResources := labeler.ExtractAffectedResources(issueBody)
 	labels := labeler.ComputeLabels(affectedResources, regexpLabels)
 
+	// If there are more than 3 service labels, treat this as a cross-provider issue
+	var serviceLabels []string
+	var nonServiceLabels []string
+	for _, l := range labels {
+		if strings.HasPrefix(l, "service/") {
+			serviceLabels = append(serviceLabels, l)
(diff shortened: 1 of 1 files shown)


tools/issue-labeler/cmd/compute_new_labels.go
```diff
@@ -45,6 +45,21 @@ func execComputeNewLabels() error {
...
 	labels := labeler.ComputeLabels(affectedResources, regexpLabels)
 
+	// If there are more than 3 service labels, treat this as a cross-provider issue
+	var serviceLabels []string
+	var nonServiceLabels []string
+	for _, l := range labels {
+		if strings.HasPrefix(l, "service/") {
+			serviceLabels = append(serviceLabels, l)
+		} else {
+			nonServiceLabels = append(serviceLabels, l)
```
c2thorn, line 55: Is this supposed to be 
```suggestion
			nonServiceLabels = append(nonServiceLabels, l)
```
instead? Otherwise we lose track of consecutive non service labels?

roaks3: Yup, thanks for catching this 🤦 . Let me test it a bit more as well.

c2thorn (approved):

PR #12864, opened 2025-01-26
Allow parallelization of per instance config deletions by not locking on the IGM

askubis:
Allow parallelization of per instance config related operations by not locking on the IGM
Fixes b/270736647
follow up to https://github.com/GoogleCloudPlatform/magic-modules/pull/12829

```release-note:enhancement
compute: allow parallelization of `google_compute_per_instance_config` and `google_compute_region_per_instance_config` deletions by not locking on the parent resource, but including instance name.
```


chayankumar999 (approved):

NickElliot (approved):
LGTM!

PR #12865, opened 2025-01-27
Create Subnetwork.yaml with new purpose called PEER_MIGRATION

gvijbha:
We add a new purpose to Subnetwork resource called PEER_MIGRATION. A subnetwork created with purpose PEER_MIGRATION is a user created resource reserved for migrating subnet and its resources from one VPC network to another.

This PR will resolve the following issue:
https://github.com/hashicorp/terraform-provider-google/issues/19936

```release-note:none
compute: added subnet peer migration capability.
```


gvijbha:
Hi, 

Yes, for Terraform client, it is a documentation only change for Create API. thanks for reviewing and approving.

roaks3 (approved):

PR #12892, opened 2025-01-28
Add Gemini Logging Setting resource

PerlMonker303:
**Release Note Template for Downstream PRs (will be copied)**

See [Write release notes](https://googlecloudplatform.github.io/magic-modules/contribute/release-notes/) for guidance.

Add new resources for Gemini Admin Control
```release-note:new-resource
`google_gemini_logging_setting`
```

PR #12893, opened 2025-01-28
Add Gemini Logging Setting Binding resource

PerlMonker303:

Add new resources for Gemini Admin Control
```release-note:new-resource
`google_gemini_logging_setting_binding`
```


PerlMonker303:
Currently failing because it depends on Gemini Logging Setting. Will fix after PR is submitted https://github.com/GoogleCloudPlatform/magic-modules/pull/12892.

mmv1/products/gemini/LoggingSettingBinding.yaml
```diff
@@ -0,0 +1,106 @@
...
+create_url: projects/{{project}}/locations/{{location}}/loggingSettings/{{logging_setting_id}}/settingBindings?settingBindingId={{setting_binding_id}}
+id_format: projects/{{project}}/locations/{{location}}/loggingSettings/{{logging_setting_id}}/settingBindings/{{setting_binding_id}}
+update_verb: PATCH
+update_mask: true
+import_format:
+  - projects/{{project}}/locations/{{location}}/loggingSettings/{{logging_setting_id}}/settingBindings/{{setting_binding_id}}
+mutex: projects/{{project}}/locations/{{location}}/loggingSettings/{{logging_setting_id}}
+examples:
+  - name: gemini_logging_setting_binding_basic
+    min_version: 'beta'
```
SirGitsalot, line 28: Does the example/test need to be beta only? The rest of the resource is not limited to beta.

SirGitsalot (commented):
Mostly LGTM, just want to see the test passing now that the blocking PR is merged. Thanks!

mmv1/products/gemini/LoggingSettingBinding.yaml
PerlMonker303: I set both LoggingSetting and LoggingSettingBinding resources to beta.

mmv1/products/gemini/LoggingSettingBinding.yaml
```diff
@@ -0,0 +1,108 @@
...
+self_link: projects/{{project}}/locations/{{location}}/loggingSettings/{{logging_setting_id}}/settingBindings/{{setting_binding_id}}
+create_url: projects/{{project}}/locations/{{location}}/loggingSettings/{{logging_setting_id}}/settingBindings?settingBindingId={{setting_binding_id}}
+id_format: projects/{{project}}/locations/{{location}}/loggingSettings/{{logging_setting_id}}/settingBindings/{{setting_binding_id}}
+update_verb: PATCH
+update_mask: true
+import_format:
+  - projects/{{project}}/locations/{{location}}/loggingSettings/{{logging_setting_id}}/settingBindings/{{setting_binding_id}}
+mutex: projects/{{project}}/locations/{{location}}/loggingSettings/{{logging_setting_id}}
+examples:
+  - name: gemini_logging_setting_binding_basic
```
SirGitsalot, line 28: You can add `exclude_test: true` to this example.

The reason the tests are failing is that variables in examples that are run as tests get `tf-test-` prepended and a random value tacked on to make them unique, so:

```
    target = "{{index $.Vars "target"}}"
    product = "{{index $.Vars "product"}}"
```

is ending up with something like `product = "tf_test_PRODUCT_UNSPECIFIEDn9eqbteuxc"`. The example is duplicated in the _test.go file, so running the example as a test too isn't adding any value anyway.

SirGitsalot (changes requested):
Left a comment with a one-line fix to get the tests green, otherwise LGTM!

mmv1/products/gemini/LoggingSettingBinding.yaml
PerlMonker303: Done, thanks!

SirGitsalot (changes requested):
The good news is that the bad request is fixed, the bad news is that we're now hitting an IAM error:

```
    "message": "Permission 'cloudaicompanion.settingBindings.loggingSettingsUse' denied on resource '//cloudresourcemanager.googleapis.com/projects/980109375338' (or it may not exist).",
```

I'm assuming that the problem is that the service account that the test is running as doesn't have the permission, but either way the fix is probably similar. I'm not familiar with the API or its requirements, but you may be able to target the project that the test is running in as the initial project a la:

```
data "google_project" "project" {
}

resource "google_gemini_logging_setting_binding" "basic_binding" {
    ...
	target = "projects/${data.google_project.project.number}"
    ...
}
```

For the update step, it's possible to create a new project as part of the test and target that.

mmv1/products/gemini/LoggingSettingBinding.yaml
```diff
@@ -0,0 +1,109 @@
...
+    resource_inside_response: true
+  include_project: false
+autogen_status: U2V0dGluZ0JpbmRpbmdCeVByb2plY3RBbmRMb2NhdGlvbkFuZExvZ2dpbmdzZXR0aW5n
+parameters:
+  - name: location
+    type: String
+    description: Resource ID segment making up resource `name`. It identifies the resource within its parent collection as described in https://google.aip.dev/122.
+    immutable: true
+    url_param_only: true
+    required: true
```
SirGitsalot, line 64: Also, you can remove `required: true` here for the same reason as #12905

SirGitsalot (changes requested):

mmv1/products/gemini/LoggingSettingBinding.yaml
```diff
@@ -0,0 +1,109 @@
...
+properties:
+  - name: labels
+    type: KeyValueLabels
+    description: Optional. Labels as key value pairs.
+  - name: target
+    type: String
+    description: Required. Target of the binding.
+    required: true
+  - name: product
```
SirGitsalot, line 87: The test is failing because `PRODUCT_UNSPECIFIED` is sent to the server, but the server is returning an empty value (which is expected, as that's how proto3 JSON encoding works for zero values). I checked the service proto and `product` is marked as required, but I don't think that's actually true if the zero value is a valid value (that is, setting `product = PRODUCT_UNSPECIFIED` and omitting the `product` field are equivalent, at least when using the JSON API).

You'll know the API better than me, but one option would be to treat the enum in the normal optional TPG enum sense:

- Remove `default_from_api`
- Remove `PRODUCT_UNSPECIFIED` as a possible value
- Remove the word "required" from the description

Then update the tests so they no longer use `PRODUCT_UNSPECIFIED`, either by omitting the `product` field or using one of the other enum values.

SirGitsalot (changes requested):

mmv1/products/gemini/LoggingSettingBinding.yaml
PerlMonker303: Solved by changing the data type from String to Enum. I also removed "required" from the description.

PerlMonker303:
Can be reviewed now.

SirGitsalot (changes requested):
Since the main change in LoggingSetting.yaml is already in the codebase, please revert and make the doc changes in another PR (we generally try to keep PRs to one topic, like adding a single new resource as in this PR).

Otherwise, LGTM - I'll approve and merge once that's done. Thanks!

mmv1/products/gemini/LoggingSetting.yaml
```diff
@@ -14,6 +14,7 @@
 ---
 name: LoggingSetting
 description: The resource for managing Logging settings for Admin Control.
+min_version: 'beta'
```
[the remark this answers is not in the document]
PerlMonker303: Thanks and sorry for the trouble.

PerlMonker303:
Created https://github.com/GoogleCloudPlatform/magic-modules/pull/13015 with description changes as suggested in the comment above.

PerlMonker303:
Can be reviewed.

SirGitsalot (approved):

PR #12901, opened 2025-01-28
omit sweeper generation for subresources

ScottSuarez:

```release-note:none

```

### DIFF
--- a/mmv1/provider/terraform.go
+++ b/mmv1/provider/terraform.go
@@ -24,6 +24,7 @@ import (
 	"os"
 	"path"
 	"path/filepath"
+	"regexp"
 	"slices"
 	"strings"
 	"time"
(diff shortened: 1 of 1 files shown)


melinath (commented):
It looks like this deletes 92 sweepers - just want to double-check that we're confident that these sweepers are not actually needed / useful?

Is this an alternative to implementing dependencies?

melinath:
(I thought the plan was to move forward with implementing support for dependencies so just trying to figure out what's changed, if anything.)

ScottSuarez:
> It looks like this deletes 92 sweepers - just want to double-check that we're confident that these sweepers are not actually needed / useful?
> 
> Is this an alternative to implementing dependencies?

These sweepers are currently broken and do not function. We can look to implement dependency support and add sweepers for them. This would require us to enable lookup of LIST functions for parent resources. 

As is though, I'm looking to clear up noise in the sweeper logs to fix what is actually broken. These urls will 404.

PR #12902, opened 2025-01-28
Add fields to support configuring new Media CDN features via Terraform.

nahurfonseca:
Add fields to support new Media CDN features via Terraform.
 
Feature: Dynamic Compression.
Documentation: http://cloud/media-cdn/docs/dynamic-compression
  - Added compression_mode Enum [DISABLED, AUTOMATIC]

Setting the mode to automatic enables dynamic compression for every eligible response. Further, it instructs Media CDN to automatically choose the best compression algorithm.

Feature: HTTP Method Filtering.
Documentation: https://cloud.google.com/media-cdn/docs/routing#method-filtering
  - Added allowed_methods Array[String]

By default, Media CDN proxies only GET, HEAD, and OPTIONS methods to your origin and filters out the methods that can modify your origin. You can override this default behavior for a specific route rule by specifying the methods that you would like proxied to your origin. Besides GET, HEAD, and OPTIONS, Media CDN supports PUT, POST, DELETE, and PATCH.

Fixes https://github.com/hashicorp/terraform-provider-google/issues/21064

```release-note:enhancement
networkservices: added `compression_mode` and `allowed_methods` fields to `edge_cache_service` resource
```

### DIFF
--- a/mmv1/products/networkservices/EdgeCacheService.yaml
+++ b/mmv1/products/networkservices/EdgeCacheService.yaml
@@ -329,6 +329,24 @@ properties:
                           # TODO: (scottsuarez) conflicts also won't work for array path matchers yet, uncomment here once supported.
                           # conflicts:
                           #   - Routing.PathMatcher.RouteRule.prefixMatch
+                  - name: 'routeMethods'
+                    type: NestedObject
+                    description: |
+                      Allow overriding the set of methods that are allowed for this route.
+                      When not set, Media CDN allows only "GET", "HEAD", and "OPTIONS".
+                    properties:
+                      - name: 'allowedMethods'
+                        type: Array
+                        description: |
+                          The non-empty set of HTTP methods that are allowed for this route.
+
+                          Any combination of "GET", "HEAD", "OPTIONS", "PUT", "POST", "DELETE", and "PATCH".
+                        item_type:
+                          type: String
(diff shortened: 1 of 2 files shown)


mmv1/products/networkservices/EdgeCacheService.yaml
```diff
@@ -329,6 +329,24 @@ properties:
...
+                    properties:
+                      - name: 'allowedMethods'
+                        type: Array
+                        description: |
+                          The non-empty set of HTTP methods that are allowed for this route.
+
+                          Any combination of "GET", "HEAD", "OPTIONS", "PUT", "POST", "DELETE", and "PATCH".
+                        item_type:
+                          type: String
+                        min_size: 1
```
trodge, line 346: Would this mean that leaving this field unspecified is invalid? If so, I think that would be a breaking change for existing users.

nahurfonseca: It is not an error to leave it blank. If you look at the other unit test example, allowed_methods is not specified.
What is an error is to specify it but leave the array empty, like allowed_methods = [].

Now... I may have expressed it wrong, min_size: 1?

But I think it is still correct.
Either the new field is not specified at all, or it has at least one method.

trodge (approved):

PR #12904, opened 2025-01-29
Make colab runtime startable/stoppable

bcreddy-gcp:

```release-note:enhancement
colab: made `google_colab_runtime` startable/stoppable.
```


bcreddy-gcp:
Couple of questions @hao-nan-li.
1) Is this the right way to implement upgrade?
2) Not sure how to test upgrade runtime function. Newly created runtimes are not upgradable and there is no reliable way of creating upgradable runtimes. How do we go about this?

hao-nan-li (commented):
Could you briefly explain what does this change intend to do? 

Is the colab runtime not startable even if the resource was created before?

bcreddy-gcp:
Colab runtime is always created in the "RUNNING" state. This change is to start/stop the runtime.

mmv1/products/colab/Runtime.yaml
```diff
@@ -57,6 +76,7 @@ parameters:
     type: String
     url_param_only: true
     description: 'The resource name of the Runtime'
+    immutable: true
```
hao-nan-li, line 79: Should this field be immutable if it's an url_param? I'd suspect that if the location changes, it refers to a different resource?

bcreddy-gcp: Done

bcreddy-gcp:
Updated the release notes. PTAL.

hao-nan-li (approved):

PR #12910, opened 2025-01-29
Add test for updating persistent disk size and disk type in cloud workstations

joelkattapuram:

```release-note:none
Add test for updating persistent disk size and disk type for cloud workstations configs.
```


hao-nan-li (commented):
I think you can add the labels field into ImportStateVerifyIgnore

PR #12920, opened 2025-01-29
enable merge-group for downstreams build op

ScottSuarez:

```release-note:none

```

### DIFF
--- a/.github/workflows/downstreams.yml
+++ b/.github/workflows/downstreams.yml
@@ -6,9 +6,11 @@ on:
     branches:
       - main
       - 'FEATURE-BRANCH-*'
+  merge_group: 
+
 
 concurrency:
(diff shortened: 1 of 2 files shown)


.github/workflows/downstreams.yml
```diff
@@ -6,9 +6,11 @@ on:
     branches:
       - main
       - 'FEATURE-BRANCH-*'
+  merge_group:
```
melinath, line 9: We should be explicit per https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#merge_group

```suggestion
  merge_group:
    types: [checks_requested]
```

melinath (approved):
LGTM apart from one thing

PR #12922, opened 2025-01-30
Enhance subnetwork.yaml to allow purpose field to be patch-able

gvijbha:
Added necessary code to make 'purpose' field updatable.

This pull request takes care of issue: https://github.com/hashicorp/terraform-provider-google/issues/19937

```release-note:enhancement
compute: added necessary code to make subnetwork's 'purpose' field updatable.
```


mmv1/products/compute/Subnetwork.yaml
```diff
@@ -225,7 +225,10 @@ properties:
       A subnetwork with purpose set to `PRIVATE_NAT` is used as source range for Private NAT gateways.
       Note that `REGIONAL_MANAGED_PROXY` is the preferred setting for all regional Envoy load balancers.
       If unspecified, the purpose defaults to `PRIVATE`.
-    immutable: true
+    update_verb: 'PATCH'
+    update_uri: 'projects/{{project}}/regions/{{region}}/subnetworks/{{name}}'
+    update_id: 'purpose'
+    fingerprint_name: 'fingerprint'
```
melinath, line 231: All you need to do to make the field updatable (if it uses the same API as the resource in general) is remove `immutable`. update_verb etc. are only needed if the resource needs to make an additional call to a separate endpoint to do the update. See https://googlecloudplatform.github.io/magic-modules/reference/field/#update_url for more information.

```suggestion
```

melinath (changes requested):
In addition to the change requested below, please add this field to an update test: https://googlecloudplatform.github.io/magic-modules/test/test/#add-an-update-test

melinath (changes requested):
As previously requested, please add this field to an update test: https://googlecloudplatform.github.io/magic-modules/test/test/#add-an-update-test

mmv1/third_party/terraform/services/compute/data_source_google_compute_subnetworks_test.go
```diff
@@ -11,6 +11,157 @@ import (
 	"github.com/hashicorp/terraform-provider-google/google/envvar"
 )
 
+func TestAccDataSourceGoogleSubnetworks_update(t *testing.T) {
```
melinath, line 14: I'm sorry for not catching this before; this test is currently among the tests for the `google_compute_subnetworks` pural [_data source_](https://developer.hashicorp.com/terraform/language/data-sources) but it should be in the tests for the `google_compute_subnetwork` [_resource_](https://developer.hashicorp.com/terraform/language/resources), which live in https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/third_party/terraform/services/compute/resource_compute_subnetwork_test.go.tmpl.

You could add updating the purpose field to the existing TestAccComputeSubnetwork_update test instead of adding a new test.

mmv1/third_party/terraform/services/compute/data_source_google_compute_subnetworks_test.go
```diff
@@ -11,6 +11,157 @@ import (
...
+		ProtoV5ProviderFactories: acctest.ProtoV5ProviderFactories(t),
+		Steps: []resource.TestStep{
+			{
+				Config: testAccCheckGoogleSubnetworksConfig(context),
+				Check: resource.ComposeTestCheckFunc(
+					// Test schema
+					resource.TestCheckResourceAttrSet("data.google_compute_subnetworks.all", "subnetworks.1.purpose"),
+					// Test Content
+					resource.TestCheckResourceAttr("data.google_compute_subnetworks.two", "subnetworks.0.purpose", "PEER_MIGRATION"),
+				),
```
melinath, line 58: in particular, these lines check values for the data source, but the tests should be checking values for the resource. Normally in a resource test that happens with an ImportState step instead of explicit checks.

mmv1/third_party/terraform/services/compute/data_source_google_compute_subnetworks.go
```diff
@@ -57,6 +57,10 @@ func DataSourceGoogleComputeSubnetworks() *schema.Resource {
 							Type:     schema.TypeBool,
 							Computed: true,
 						},
+						"purpose": {
+							Type:     schema.TypeString,
+							Computed: true,
+						},
```
melinath, line 63: This change seems unrelated to the purpose of the PR; I'm happy to include it in this PR if you prefer, but it's not necessary. Possibly it was something added to resolve the test failures because they were checking values on the data source instead of testing the resource directly?

melinath (changes requested):
I'm sorry for not catching this before; this PR needs update tests for the resource rather than the data source.

mmv1/third_party/terraform/services/compute/resource_compute_subnetwork_test.go.tmpl
```diff
@@ -601,6 +632,22 @@ resource "google_compute_subnetwork" "network-with-private-google-access" {
...
+func testAccComputeSubnetwork_update4(cnName, subnetworkName string) string {
+	return fmt.Sprintf(`
+resource "google_compute_network" "custom-test" {
+  name                    = "%s"
+  auto_create_subnetworks = false
+}
+
+resource "google_compute_subnetwork" "network-with-migration-purpose" {
+  name          = "%s"
+  region        = "us-central1"
```
melinath, line 644: This is missing here, causing the failure: Error: Error creating Subnetwork: googleapi: Error 400: Required field 'resource.ipCidrRange' not specified

```suggestion
  ip_cidr_range = "10.4.0.0/16"
  region        = "us-central1"
```

melinath (changes requested):

melinath (changes requested):
Error:
```
    resource_compute_subnetwork_test.go:114: Resource specified by ResourceName couldn't be found: google_compute_subnetwork.network-with-private-google-access
```

mmv1/third_party/terraform/services/compute/resource_compute_subnetwork_test.go.tmpl
```diff
@@ -99,9 +108,11 @@ func TestAccComputeSubnetwork_update(t *testing.T) {
 	t.Parallel()
 
 	var subnetwork compute.Subnetwork
-
+        var subnetwork1 compute.Subnetwork
```
melinath, line 111:
```suggestion
```

mmv1/third_party/terraform/services/compute/resource_compute_subnetwork_test.go.tmpl
```diff
@@ -601,6 +633,23 @@ resource "google_compute_subnetwork" "network-with-private-google-access" {
...
 }
 
+func testAccComputeSubnetwork_update4(cnName, subnetworkName string) string {
+	return fmt.Sprintf(`
+resource "google_compute_network" "custom-test" {
+  name                    = "%s"
+  auto_create_subnetworks = false
+}
+
+resource "google_compute_subnetwork" "network-with-migration-purpose" {
```
melinath, line 643: This test should update the previously existing network instead of deleting it and creating a new one.

```suggestion
resource "google_compute_subnetwork" "network-with-private-google-access" {
```

mmv1/third_party/terraform/services/compute/resource_compute_subnetwork_test.go.tmpl
```diff
@@ -139,11 +150,24 @@ func TestAccComputeSubnetwork_update(t *testing.T) {
...
 			{
 				ResourceName:      "google_compute_subnetwork.network-with-private-google-access",
 				ImportState:       true,
 				ImportStateVerify: true,
 			},
+			{
+				ResourceName:      "google_compute_subnetwork.network-with-migration-purpose",
+				ImportState:       true,
+				ImportStateVerify: true,
+			},
```
melinath, line 170:
```suggestion
```

mmv1/third_party/terraform/services/compute/resource_compute_subnetwork_test.go.tmpl
```diff
@@ -139,11 +150,24 @@ func TestAccComputeSubnetwork_update(t *testing.T) {
 						t, "google_compute_subnetwork.network-with-private-google-access", &subnetwork),
 				),
 			},
+			{
```
melinath, line 153: This should be added after each config step.

```suggestion
			{
				ResourceName:      "google_compute_subnetwork.network-with-private-google-access",
				ImportState:       true,
				ImportStateVerify: true,
			},
			{
```

mmv1/third_party/terraform/services/compute/resource_compute_subnetwork_test.go.tmpl
```diff
@@ -139,11 +150,24 @@ func TestAccComputeSubnetwork_update(t *testing.T) {
...
+				// update the purpose from PEER_MIGRATION to PRIVATE
+				Config: testAccComputeSubnetwork_update4(cnName, subnetwork2Name),
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckComputeSubnetworkExists(
+						t, "google_compute_subnetwork.network-with-migration-purpose", &subnetwork1),
+				),
```
melinath, line 159: checking whether the subnetwork exists is covered by having an ImportState step after each config step. The  `testAccCheckComputeSubnetworkExists` pattern is only useful for accessing the actual state of the API object in cases where an ImportState check isn't applicable - for example, if the field being checked isn't supported in Terraform, or if you need to reference a value from one step in a following step.

It looks like `testAccCheckComputeSubnetworkExists` was added [over 9 years ago](https://github.com/hashicorp/terraform-provider-google-beta/commit/98f8b0b3c7b93a59fa80d260f5cbf7613fa7f9d2) - I suspect it might pre-date ImportState steps. It would be great to clean up and remove all of these because it's just making this test (and this PR) more confusing than it needs to be. (No need to fix the rest of the usage in this file - just this test would be great.)

Additionally, all steps except the first step should have a ConfigPlanCheck to ensure that the resource under test is actually being updated, not deleted and recreated:

```suggestion
            ConfigPlanChecks: resource.ConfigPlanChecks{
               PreApply: []plancheck.PlanCheck{
                  plancheck.ExpectResourceAction("google_pubsub_topic.foo", plancheck.ResourceActionUpdate),
               },
            },
```

melinath (changes requested):
So, I took a closer look at this. The reason the TestAccComputeSubnetwork_update test is failing is that when it gets to the first ImportState step, the last configuration applied was `testAccComputeSubnetwork_update4`, which didn't include the `google_compute_subnetwork.network-with-private-google-access`. As a result, Terraform isn't able to do an import into that resource.

`testAccComputeSubnetwork_update4` should update the existing resource instead of deleting it and creating a new one.

mmv1/third_party/terraform/services/compute/resource_compute_subnetwork_test.go.tmpl
gvijbha: Thanks for looking in to the failure and for your suggestion. Applied all the suggested changes now.

Regarding replacing SubnetworkExists with ConfigPlan check, i will try this with newly added update step.. once it works, i will replicate the same to other steps as well.

mmv1/third_party/terraform/services/compute/resource_compute_subnetwork_test.go.tmpl
```diff
@@ -8,7 +8,7 @@ import (
 	"github.com/hashicorp/terraform-plugin-testing/helper/resource"
 	"github.com/hashicorp/terraform-plugin-testing/terraform"
 	"github.com/hashicorp/terraform-provider-google/google/acctest"
-	tpgcompute "github.com/hashicorp/terraform-provider-google/google/services/compute"
+	"tpgcompute "github.com/hashicorp/terraform-provider-google/google/services/compute"
```
melinath, line 11: This extra `"` is causing a compilation error

```suggestion
	tpgcompute "github.com/hashicorp/terraform-provider-google/google/services/compute"
```

mmv1/third_party/terraform/services/compute/resource_compute_subnetwork_test.go.tmpl
```diff
@@ -152,6 +160,36 @@ func TestAccComputeSubnetwork_update(t *testing.T) {
 	}
 }
 
+func TestAccComputeSubnetwork_purposeUpdate(t *testing.T) {
```
melinath, line 163: Why is this now a separate test?

melinath (changes requested):
This looks significantly different than I would have expected. It seems like the changes I requested were made but ended up being rolled back. It's a bit difficult to see what happened because there were so many commits that only changed one or two bits of whitespace, and a lot of the changes seem to have failed due to compilation errors before even getting to VCR tests.

Please put this back in the state I had previously requested (unless there's a specific reason it shouldn't be).

Please also be sure to run tests locally before pushing commits to shorten the time it takes to review your PRs: https://googlecloudplatform.github.io/magic-modules/test/run-tests/

mmv1/third_party/terraform/services/compute/resource_compute_subnetwork_test.go.tmpl
gvijbha: looking at how others have added their own updates for separate fields, it made sense to have a separate update for purpose field instead dumping all in one place.

mmv1/third_party/terraform/services/compute/resource_compute_subnetwork_test.go.tmpl
```diff
@@ -152,6 +165,36 @@ func TestAccComputeSubnetwork_update(t *testing.T) {
...
+
+        var subnetwork compute.Subnetwork
+	cnName := fmt.Sprintf("tf-test-%s", acctest.RandString(t, 10))
+	subnetworkName := fmt.Sprintf("tf-test-%s", acctest.RandString(t, 10))
+
+	acctest.VcrTest(t, resource.TestCase{
+		PreCheck:                 func() { acctest.AccTestPreCheck(t) },
+		ProtoV5ProviderFactories: acctest.ProtoV5ProviderFactories(t),
+		CheckDestroy:             testAccCheckComputeSubnetworkDestroyProducer(t),
+		Steps: []resource.TestStep{
```
melinath, line 179: If you're going to have this be a separate test, you need to add a step that sets up the initial subnetwork _before_ the step that updates it. Right now you're just creating the resource, not testing update-in-place.

mmv1/third_party/terraform/services/compute/resource_compute_subnetwork_test.go.tmpl
```diff
@@ -110,34 +119,38 @@ func TestAccComputeSubnetwork_update(t *testing.T) {
...
 				Config: testAccComputeSubnetwork_update1(cnName, "10.2.0.0/24", subnetworkName),
-				Check: resource.ComposeTestCheckFunc(
-					testAccCheckComputeSubnetworkExists(
-						t, "google_compute_subnetwork.network-with-private-google-access", &subnetwork),
-				),
+				ConfigPlanChecks: resource.ConfigPlanChecks{
+                                    PreApply: []plancheck.PlanCheck{
+                                            plancheck.ExpectResourceAction("google_compute_subnetwork.network-with-private-google-access", plancheck.ResourceActionUpdate),
+                                    },
+                                },
```
melinath, line 126: On the first step, the resource is being created, not updated, so this isn't needed.

```suggestion
```

mmv1/third_party/terraform/services/compute/resource_compute_subnetwork_test.go.tmpl
```diff
@@ -60,26 +61,30 @@ func TestAccComputeSubnetwork_basic(t *testing.T) {
 
 	var subnetwork1 compute.Subnetwork
 	var subnetwork2 compute.Subnetwork
+        var subnetwork3 compute.Subnetwork
```
melinath, line 64: If you're using a separate test, all the subnetwork4 / `network-with-migration-purpose` stuff needs to be removed from this function

mmv1/third_party/terraform/services/compute/resource_compute_subnetwork_test.go.tmpl
```diff
@@ -601,6 +652,23 @@ resource "google_compute_subnetwork" "network-with-private-google-access" {
 `, cnName, subnetworkName, cidrRange)
 }
 
+func testAccComputeSubnetwork_update4(cnName, subnetworkName string) string {
```
melinath, line 655: If you're making a separate test, this should have a name closer to the main test function that uses it, like `testAccComputeSubnetwork_purposeUpdate`

mmv1/third_party/terraform/services/compute/resource_compute_subnetwork_test.go.tmpl
```diff
@@ -544,7 +587,15 @@ resource "google_compute_subnetwork" "network-with-private-google-access" {
...
-`, cnName, subnetwork1Name, subnetwork2Name, subnetwork3Name)
+
+resource "google_compute_subnetwork" "network-with-migration-purpose" {
+  name                     = "%s"
+  ip_cidr_range            = "10.4.0.0/16"
+  region                   = "us-central1"
+  network                  = google_compute_network.custom-test.self_link
+  purpose                  = "PEER_MIGRATION"
+}
+`, cnName, subnetwork1Name, subnetwork2Name, subnetwork3Name, subnetwork4Name)
```
melinath, line 598: Same as above - subnetwork4 / `network-with-migration-purpose` should be removed here & elsewhere since you're doing a separate test.

```suggestion
`, cnName, subnetwork1Name, subnetwork2Name, subnetwork3Name)
```

melinath (changes requested):

mmv1/third_party/terraform/services/compute/resource_compute_subnetwork_test.go.tmpl
gvijbha: Added a step that returns a subnet object with purpose set to PEER_MIGRATION. thanks for the pointer.

mmv1/third_party/terraform/services/compute/resource_compute_subnetwork_test.go.tmpl
gvijbha: brought back the resource presence check to the first step and the left the update check for the other steps.

mmv1/third_party/terraform/services/compute/resource_compute_subnetwork_test.go.tmpl
gvijbha: renamed the test as suggested.

mmv1/third_party/terraform/services/compute/resource_compute_subnetwork_test.go.tmpl
gvijbha: removed the migration-purpose related code from the basic test.

mmv1/third_party/terraform/services/compute/resource_compute_subnetwork_test.go.tmpl
```diff
@@ -152,6 +154,45 @@ func TestAccComputeSubnetwork_update(t *testing.T) {
...
+			},
+			{
+				// update the purpose from PEER_MIGRATION to PRIVATE
+				Config: testAccComputeSubnetwork_purposeUpdate1(cnName, subnetworkName),
+				ConfigPlanChecks: resource.ConfigPlanChecks{
+                                    PreApply: []plancheck.PlanCheck{
+                                            plancheck.ExpectResourceAction("google_compute_subnetwork.network-with-migration-purpose", plancheck.ResourceActionUpdate),
+                                    },
+                                },
+			},
```
melinath, line 184: This test is failing with the following error message:

```
    resource_compute_subnetwork_test.go:162: Step 2/3 error: Pre-apply plan check(s) failed:
        'google_compute_subnetwork.network-with-migration-purpose' - expected Update, got action(s): [delete create]
```

The error message means that `google_compute_subnetwork.network-with-migration-purpose` is getting recreated in step 2 (this step) instead of updated. That means an immutable field was changed in the configuration, preventing us from testing the update logic - this is exactly the kind of test issue that the ConfigPlanChecks are there to catch 🎉 

Looking at your configs, I don't see any obvious changes to immutable fields. Digging a bit more, it appears that the overall Subnetwork resource is considered immutable: https://github.com/GoogleCloudPlatform/magic-modules/blob/c721c74c72110cc330eef08932a4011e29beb0ca/mmv1/products/compute/Subnetwork.yaml#L48

This means that only fields that specify a custom [update_url](https://googlecloudplatform.github.io/magic-modules/reference/field/#update_url) can be updated in-place. However, it looks like a number of fields use the normal resource update URL as their "custom URL" - the Subnetwork resource is very old, so this seems like it's probably an oversight / left over from before the resource could be updated in place.

I'd recommend removing the resource-level immutability and adding `immutable: true` to all immutable fields. You'll also need to set the resource-level [update_url](https://googlecloudplatform.github.io/magic-modules/reference/resource/#update_url) and [update_verb](https://googlecloudplatform.github.io/magic-modules/reference/resource/#update_verb) properties. That should fix the issue.

mmv1/third_party/terraform/services/compute/resource_compute_subnetwork_test.go.tmpl
```diff
@@ -152,6 +154,45 @@ func TestAccComputeSubnetwork_update(t *testing.T) {
...
+		ProtoV5ProviderFactories: acctest.ProtoV5ProviderFactories(t),
+		CheckDestroy:             testAccCheckComputeSubnetworkDestroyProducer(t),
+		Steps: []resource.TestStep{
+			{
+				// Create a subnetwork with the purpose set to PEER_MIGRATION
+				Config: testAccComputeSubnetwork_purposeUpdate(cnName, subnetworkName),
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckComputeSubnetworkExists(t, "google_compute_subnetwork.network-with-migration-purpose", &subnetwork),
+				),
+			},
```
melinath, line 175: you should add an `ImportStateVerify` step between steps 1 & 2 to make sure that the API values are set properly in the previous step.

```suggestion
			},
			{
				ResourceName:      "google_compute_subnetwork.network-with-migration-purpose",
				ImportState:       true,
				ImportStateVerify: true,
			},
```

mmv1/third_party/terraform/services/compute/resource_compute_subnetwork_test.go.tmpl
```diff
@@ -111,33 +111,35 @@ func TestAccComputeSubnetwork_update(t *testing.T) {
 			{
 				Config: testAccComputeSubnetwork_update1(cnName, "10.2.0.0/24", subnetworkName),
 				Check: resource.ComposeTestCheckFunc(
-					testAccCheckComputeSubnetworkExists(
-						t, "google_compute_subnetwork.network-with-private-google-access", &subnetwork),
+					testAccCheckComputeSubnetworkExists(t, "google_compute_subnetwork.network-with-private-google-access", &subnetwork),
 				),
 			},
```
melinath, line 116: The config steps should all have ImportStateVerify steps after them.

```suggestion
			},
			{
				ResourceName:      "google_compute_subnetwork.network-with-private-google-access",
				ImportState:       true,
				ImportStateVerify: true,
			},
```

mmv1/third_party/terraform/services/compute/resource_compute_subnetwork_test.go.tmpl
```diff
@@ -111,33 +111,35 @@ func TestAccComputeSubnetwork_update(t *testing.T) {
...
-					testAccCheckComputeSubnetworkExists(
-						t, "google_compute_subnetwork.network-with-private-google-access", &subnetwork),
-				),
+				ConfigPlanChecks: resource.ConfigPlanChecks{
+                                    PreApply: []plancheck.PlanCheck{
+                                            plancheck.ExpectResourceAction("google_compute_subnetwork.network-with-private-google-access", plancheck.ResourceActionUpdate),
+                                    },
+                                },
 			},
 			{
```
melinath, line 126: This step is failing because it modifies an immutable field - this is clearly supposed to be testing update-in-place logic but failed to do so because the ConfigPlanChecks weren't there previously. Please check whether the problem is that it's modifying a field it shouldn't, or whether there's a field it's modifying that is incorrectly (still?) immutable because of the same resource immutability issue.

melinath (changes requested):

mmv1/third_party/terraform/services/compute/resource_compute_subnetwork_test.go.tmpl
melinath: as discussed, leaving this out for now because there's a bug in the resource where one of these steps is not updating in-place (and we have a separate test for the actual change at this point.)

mmv1/third_party/terraform/services/compute/resource_compute_subnetwork_test.go.tmpl
melinath: As discussed, not fixing this in this PR

melinath (approved):
LGTM!

PR #12924, opened 2025-01-30
Clarify when changing a List to a Set is considered breaking

roaks3:
Follow-up to https://github.com/GoogleCloudPlatform/magic-modules/pull/12814

```release-note:none

```

### DIFF
--- a/docs/content/breaking-changes/breaking-changes.md
+++ b/docs/content/breaking-changes/breaking-changes.md
@@ -57,7 +57,9 @@ For more information, see
 * <a name="resource-schema-field-removal-or-rename"></a>Removing or renaming a field 
 * <a name="field-changing-type"></a> Changing field output type
   * Between primitive types, like changing a String to an Integer
-  * Between complex types like changing a List to a Set.
+  * Between complex types like changing a List to a Set. Note that List -> Set
+    can be considered non-breaking if the field is unordered and the order
+    returned from the API changes unpredictably.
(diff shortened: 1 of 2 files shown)


melinath (changes requested):
Exceptions should be documented in https://googlecloudplatform.github.io/magic-modules/breaking-changes/make-a-breaking-change/#in-minor-releases

roaks3:
@melinath I can move it, but just wanted to mention that personally I wouldn't have found that page. Could you confirm we're happy with the discoverability of this note on that page?

melinath:
I think we should be consistent about where we document the exceptions - they should either all be on "Make a breaking change" or all be on "Types of breaking changes".

Currently we have them on the "Make a breaking change" page because the idea is that the change _is_ a breaking change, it's just that we make an exception in certain edge cases. Documenting it on "Types of breaking changes" would be more indicating that the change is not considered breaking in all cases.

For me the thing that makes "Make a breaking change" preferable is that _any_ breaking change could be safe in a minor release, as long as it fixes a bug that impacts all configurations that include a field or resource.

I could potentially be convinced otherwise; if this is something you feel strongly about we should discuss it offline.

roaks3:
Yea, IMO this is just not a place where the note will be found very easily, but I've gone ahead and moved it per your request to move this forward.

roaks3:
K fixed!

melinath (approved):

PR #12926, opened 2025-01-30
Add diff suppression for the renewal_plan field in google_bigquery_capacity_commitment

wj-chen:
Fixes b/392883972. `plan` and `renewal_plan` are both `CommitmentPlan` and should share the same diff suppression logic: https://cloud.google.com/bigquery/docs/reference/reservations/rpc/google.cloud.bigquery.reservation.v1#capacitycommitment.

```release-note:bug
bigquery: added diff suppression for legacy values in `renewal_plan` field in `google_bigquery_capacity_commitment` resource
```


melinath (approved):
LGTM - not actioning the missing test since this is just adding diff suppression.

PR #12930, opened 2025-01-31
Add max staleness to replica table

trodge:

```release-note:none

```


wj-chen:
There might be a bug with max_staleness (reported in b/393456580) that's causing the test failure. Investigating.

wj-chen:
I see that this is to fix the test failure https://github.com/hashicorp/terraform-provider-google/issues/21167. The cause looks to be a backend change, I will update in that issue once I learn more.

PR #12939, opened 2025-01-31
Made resources set fields used in the ID format on post-create

melinath:
Previously this used 'identity' fields, but that's a rarely-used feature that doesn't reflect the actual reason these are needed: filling out the resource ID. There are a number of post_create custom codes that specifically work around this issue.

I didn't remove the existing custom code since there are a bunch of them and some of them are doing more complex things, so we'd need to verify one-by-one that they are safe to back out - but this should at least remove the need to introduce more custom code like this.

One consideration is that we also need to make sure we're setting anything required by Read, which generally uses SelfLink rather than IdFormat - though they are often the same, it might be necessary to check both.

In particular, if self-link is ever set to `{{name}}`, but IdFormat is something else, the previous default behavior for identity fields was to set a value for `{{name}}` - any resources like this will start to fail.

```release-note:none

```

### DIFF
--- a/mmv1/templates/terraform/resource.go.tmpl
+++ b/mmv1/templates/terraform/resource.go.tmpl
@@ -293,6 +293,12 @@ func resource{{ $.ResourceName -}}Create(d *schema.ResourceData, meta interface{
 {{- if or (not $.GetAsync) (not (and ($.GetAsync.IsA "OpAsync") ($.GetAsync.Allow "Create"))) }}
 {{- range $prop := $.GettableProperties }}
 {{-  if $.InIdFormat $prop }}
+{{-  if eq $prop.CustomFlatten "templates/terraform/custom_flatten/id_from_name.tmpl" }}
+    // Setting `name` field so that `id_from_name` flattener will work properly.
+    if err := d.Set("name", flatten{{ if $.NestedQuery -}}Nested{{end}}{{ $.ResourceName -}}Name(res["name"], d, config)); err != nil {
+        return fmt.Errorf(`Error setting computed identity field "name": %s`, err)
+    }
+{{-  end }}
 {{-    if and $prop.Output (not $prop.IgnoreRead) }}
     if err := d.Set("{{ underscore $prop.Name -}}", flatten{{ if $.NestedQuery -}}Nested{{end}}{{ $.ResourceName -}}{{ camelize $prop.Name "upper"  -}}(res["{{ $prop.ApiName -}}"], d, config)); err != nil {
         return fmt.Errorf(`Error setting computed identity field "{{ underscore $prop.Name }}": %s`, err)
--- a/mmv1/templates/terraform/resource.go.tmpl
+++ b/mmv1/templates/terraform/resource.go.tmpl
@@ -293,11 +293,11 @@ func resource{{ $.ResourceName -}}Create(d *schema.ResourceData, meta interface{
 {{- if or (not $.GetAsync) (not (and ($.GetAsync.IsA "OpAsync") ($.GetAsync.Allow "Create"))) }}
 {{- range $prop := $.GettableProperties }}
 {{-  if $.InIdFormat $prop }}
-{{-    if $prop.Output }}
+{{-    if and $prop.Output (not $prop.IgnoreRead) }}
     if err := d.Set("{{ underscore $prop.Name -}}", flatten{{ if $.NestedQuery -}}Nested{{end}}{{ $.ResourceName -}}{{ camelize $prop.Name "upper"  -}}(res["{{ $prop.ApiName -}}"], d, config)); err != nil {
         return fmt.Errorf(`Error setting computed identity field "{{ underscore $prop.Name }}": %s`, err)
     }
-{{-    else if $prop.DefaultFromApi }}
+{{-    else if and $prop.DefaultFromApi (not $prop.IgnoreRead) }}
     // {{ underscore $prop.Name }} is set by API when unset
     if tpgresource.IsEmptyValue(reflect.ValueOf(d.Get("{{ underscore $prop.Name }}"))) {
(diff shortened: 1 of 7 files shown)


melinath:
I added the following to the generation code & it only kicked up one case where this code is triggered and a field was in self link but not in IdFormat:

```go
func (r Resource) InIdFormat(prop Type) bool {
	selfLinkFields := r.ExtractIdentifiers(r.SelfLinkUri())
	fields := r.ExtractIdentifiers(r.GetIdFormat())
	if !slices.Contains(fields, prop.Name) && slices.Contains(selfLinkFields, prop.Name) {
		fmt.Printf("WARNING: %s is in selfLinkFields but not in IdFormat fields. Output: %t DefaultFromApi: %t\n", prop.Name, prop.Output, prop.DefaultFromApi)
	}
	return slices.Contains(fields, prop.Name)
}
```

```
2025/01/31 16:09:14 Generating FolderFeed resource
WARNING: parent is in selfLinkFields but not in IdFormat fields. Output: false DefaultFromApi: false
```

Since the field in question isn't output-only or default_from_api it should be safe.

melinath:
Hm, at least some of these look like they could plausibly be due to this change somehow; I'll convert this to a draft for now while I look into it.

melinath:
- `TestAccCloudQuotasQuotaPreference_cloudquotasQuotaPreferenceBasicExample_update` is [failing in nightlies](https://hashicorp.teamcity.com/test/-8295173404852154014?currentProjectId=TerraformProviders_GoogleCloud_GOOGLE_BETA_NIGHTLYTESTS).
- `TestAccCloudbuildWorkerPool_basic ` is [failing in nightlies](https://hashicorp.teamcity.com/test/-510496357074789472?currentProjectId=TerraformProviders_GoogleCloud_GOOGLE_BETA_NIGHTLYTESTS)
- `TestAccManagedKafkaConnector_managedkafkaConnectorBasicExample` [also](https://hashicorp.teamcity.com/test/-5191058866131656347?currentProjectId=TerraformProviders_GoogleCloud_GOOGLE_BETA_NIGHTLYTESTS)
- `TestAccManagedKafkaConnectCluster_managedkafkaConnectClusterBasicExample` [also](https://hashicorp.teamcity.com/test/8869563437246997016?currentProjectId=TerraformProviders_GoogleCloud_GOOGLE_BETA_NIGHTLYTESTS)

melinath:
@c2thorn @ScottSuarez I'm marking this ready for review; there are more post_creates that can probably be cleaned up but we can do those in separate follow-up PRs... there will be case-by-case decisions needed, which would make it harder to review the overall change.

mmv1/templates/terraform/resource.go.tmpl
```diff
@@ -288,13 +288,29 @@ func resource{{ $.ResourceName -}}Create(d *schema.ResourceData, meta interface{
...
     }
-{{- /* # Set resource properties from create API response (unless it returns an Operation) */}}
-{{- if not (and $.GetAsync ($.GetAsync.IsA "OpAsync")) }}
+{{- /* Set output-only resource properties from create API response (as long as Create doesn't use an async operation) */}}
+{{- /* This is necessary so that the ID is built correctly. */}}
+{{- if or (not $.GetAsync) (not (and ($.GetAsync.IsA "OpAsync") ($.GetAsync.Allow "Create"))) }}
 {{- range $prop := $.GettableProperties }}
-{{-  if and ($.IsInIdentity $prop) $prop.Output }}
+{{-  if $.InIdFormat $prop }}
+{{-  if eq $prop.CustomFlatten "templates/terraform/custom_flatten/id_from_name.tmpl" }}
```
c2thorn, line 296: This works... but if we are directly building logic around a specific custom flatten, maybe the flattener is leaving the realm of custom code? Should `id_from_name` just be a field-level toggle and we move `id_from_name.tmpl` to `mmv1/terraform/templates`?

Then we can add it to https://googlecloudplatform.github.io/magic-modules/reference/field/ and explain when to use it?

c2thorn (commented):
Seeing [several diffs](https://github.com/modular-magician/terraform-provider-google-beta/compare/auto-pr-12939-old..auto-pr-12939#diff-6739ac7cfc6774718bb8f961cd3ce7f3c965a2f86732e6ba4f936a376df8b6d9) in the generated output where resources are losing setting a `name` in the create function. Just double checking that this should be okay since it will be set on read?

mmv1/templates/terraform/resource.go.tmpl
c2thorn: If there's agreement on this, but it's out of scope for this change, I'm okay with that. Especially since if we switch it to a field-level toggle, there should be no diffs generated from this PR

melinath:
> Seeing [several diffs](https://github.com/modular-magician/terraform-provider-google-beta/compare/auto-pr-12939-old..auto-pr-12939#diff-6739ac7cfc6774718bb8f961cd3ce7f3c965a2f86732e6ba4f936a376df8b6d9) in the generated output where resources are losing setting a `name` in the create function. Just double checking that this should be okay since it will be set on read?

Yes, we previously always set `name` after create because it was considered the "default identifier" and therefore necessary to set before read - but most of these resources don't actually use `name` to build their Read URL, which is why they needed separate post_create custom code to set their actual identifiers.

(Another path forward would be to start using the `identifiers` field more broadly to explicitly say what should be set after create, but that seems excessive since we can just pull the necessary fields from the self link.)

(And name will be set on Read.)

mmv1/templates/terraform/resource.go.tmpl
melinath: yes, I think there's probably some improvement to make here (such as making id_from_name a field-level toggle or similar) but I didn't want to get into it as part of this PR.

c2thorn (approved):
spot checked several of the diffs and confirmed that we are just adding automatic (but now duplicate) identifier code or removing unnecessary default code. Should be saving the next contributor that runs into this some time (unless they happen to be copying from a resource that hasn't had its custom code cleaned up yet)

LGTM

ScottSuarez (approved):

melinath:
Created https://github.com/hashicorp/terraform-provider-google/issues/22210 to track converting id_from_name to a field-level flag.

PR #12941, opened 2025-02-01
Add conflict for notification_rate_limit when alert policy is not log based

philipsabri:
Currently, when you try to add `notification_rate_limit` to a non-log alert policy, the GCP API stops you at the apply stage. This change would prevent it at the plan stage instead.

No changes are needed for users, as the API already enforces this restriction. However, if you work in an environment where the apply stage only happens after merging your PR, you would have to create another PR to remove this block. This has happened to me multiple times when copying other policy blocks that included `notification_rate_limit`.

https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/monitoring_alert_policy#notification_rate_limit-1

```release-note: note
Add conflict rules for notification_rate_limit when alert policy is not log based
```

### DIFF
--- a/mmv1/products/monitoring/AlertPolicy.yaml
+++ b/mmv1/products/monitoring/AlertPolicy.yaml
@@ -954,6 +954,11 @@ properties:
             description: |
               Not more than one notification per period.
               A duration in seconds with up to nine fractional digits, terminated by 's'. Example "60.5s".
+        conflicts:
+          - conditions.0.condition_absent
+          - conditions.0.condition_monitoring_query_language
+          - conditions.0.condition_threshold
+          - conditions.0.condition_prometheus_query_language
       - name: 'autoClose'
         type: String
         description: |


philipsabri:
Error:
```
provider_test.go:23: err: resource google_monitoring_alert_policy: ConflictsWith: notification_rate_limit configuration block reference (conditions.0.condition_absent) can only be used with TypeList and MaxItems: 1 configuration blocks
```

Hmm is it not correct how I have configured the conflicts block?

It get generated to this in the provider
```go
ConflictsWith: []string{"conditions.0.condition_absent", "conditions.0.condition_monitoring_query_language", "conditions.0.condition_threshold", "conditions.0.condition_prometheus_query_language"},
```

If I configure it in `magic-module` like this
```yaml
        conflicts:
          - conditions.condition_absent
          - conditions.condition_monitoring_query_language
          - conditions.condition_threshold
          - conditions.condition_prometheus_query_language
```
It wont be generated properly in the provider

trodge:
Is one of the fields in this conflict a list of nested objects?

Currently they are [not supported](https://googlecloudplatform.github.io/magic-modules/reference/field/#conflicts)

PR #12943, opened 2025-02-02
Add source_machine_image_encryption_key to google_compute_instance_from_machine_image resource

Dawid212:
This PR adds support for the source_machine_image_encryption_key field to the google_compute_instance_from_machine_image resource.

```release-note:enhancement
compute: added `source_machine_image_encryption_key` field to `google_compute_instance_from_machine_image` resource  
```


slevenick (changes requested):
Looks like we're getting:

panic: Invalid address to set: []string{"source_machine_image_encryption_key"}

mmv1/third_party/terraform/services/compute/resource_compute_instance_from_machine_image.go.tmpl
```diff
@@ -64,6 +64,39 @@ func computeInstanceFromMachineImageSchema() map[string]*schema.Schema {
...
+				"rsa_encrypted_key": {
+					Type:      schema.TypeString,
+					Optional:  true,
+					Sensitive: true,
+				},
+				"kms_key_name": {
+					Type:     schema.TypeString,
+					Optional: true,
+				},
+				"sha256": {
```
slevenick, line 89: Looks like this shouldn't be settable by the user. It should be marked Computed: true

mmv1/third_party/terraform/services/compute/resource_compute_instance_from_machine_image_test.go.tmpl
```diff
@@ -1031,6 +1156,63 @@ resource "google_compute_instance_from_machine_image" "foobar" {
...
+
+func testAccComputeInstanceFromMachineImage_encryptionKeySha(context map[string]interface{}) string {
+	return fmt.Sprintf(`
+resource "google_compute_instance_from_machine_image" "foobar" {
+		provider = google-beta
+		name = "%s"
+		zone = "us-central1-a"
+		source_machine_image = google_compute_machine_image.foobar.self_link
+		source_machine_image_encryption_key {
+			sha256 = "%s"
```
slevenick, line 1196: According to the docs sha256 is output only: https://cloud.google.com/compute/docs/reference/rest/v1/instances

So how can this possibly work?

slevenick: Still present

Dawid212:
Sorry @slevenick for delay on this PR. It completely slipped my mind.

slevenick (commented):
There's non-determinism in the test due to the rotation time being set on the kms crypto key. Is it possible to remove that from the test to allow for deterministic testing? Alternatively we can skip this test in VCR

slevenick (approved):

PR #2794, opened 2019-12-06
storage bucket iam generation

danawillow:
**Release Note Template for Downstream PRs (will be copied)**

```release-note:enhancement
storage: added support for IAM Conditions to the `google_storage_bucket_iam_*` resources (beta provider only)
```

```release-note:enhancement
storage: updated `id` and `bucket` fields for `google_storage_bucket_iam_*` resources to use `b/{bucket_name}`
```

```release-note:breaking-change
`google_storage_bucket_iam_*` resources now support IAM Conditions (beta provider only). If any conditions had been created out of band before this release, take extra care to ensure they are present in your Terraform config so the provider doesn't try to create new bindings with no conditions. Terraform will show a diff that it is adding the condition to the resource, which is safe to apply. 
```

PR #12944, opened 2025-02-03
fix: (storage) bucket policy delete 412 error

gurusai-voleti:
Fixes: https://github.com/hashicorp/terraform-provider-google/issues/20838

```release-note:bug
storage: fixed a 412 error returned on some `google_storage_bucket_iam_policy` deletions
```


rileykarson (commented):
It seems like we added the etag on delete for GCS bucket in the first place. I can't find context on why yet, but looking into it: https://github.com/GoogleCloudPlatform/magic-modules/pull/2794

edit: original context in https://github.com/hashicorp/terraform-provider-google/pull/1190#issuecomment-377250003

When a blank `bindings` block is sent we get back:

```
{
  "error": {
    "code": 400,
    "message": "A policy to update must be provided.",
    "errors": [
      {
        "message": "A policy to update must be provided.",
        "domain": "global",
        "reason": "required"
      }
    ]
  }
}
```

gurusai-voleti:
> It seems like we added the etag on delete for GCS bucket in the first place. I can't find context on why yet, but looking into it: #2794
> 
> edit: original context in [hashicorp/terraform-provider-google#1190 (comment)](https://github.com/hashicorp/terraform-provider-google/pull/1190#issuecomment-377250003)
> 
> When just an empty `bindings` block is sent we get back:
> 
> ```
> {
>   "error": {
>     "code": 400,
>     "message": "A policy to update must be provided.",
>     "errors": [
>       {
>         "message": "A policy to update must be provided.",
>         "domain": "global",
>         "reason": "required"
>       }
>     ]
>   }
> }
> ```

agree but the request body we are sending will look like below
{
"version": 3
}

the request is accepted and policy got updated and response is 200

kautikdk (commented):
I believe there is a server side validation for rejecting an empty `Buckets:setIamPolicy` request. I tested compute `setIamPolicy` method which allows empty policy and clears if provided. Specifying policy version in the Buckets:setIamPolicy method request should be enough to clear IAM bindings.

rileykarson:
Yep- in my testing any value in addition to the empty `bindings` array passed the validation and cleared the bindings. Sorry for the delay on my end- considering whether we can just remove the etag value here for all IAM policies or not, since that would obviate this change.

gurusai-voleti:
> Yep- in my testing any value in addition to the empty `bindings` array passed the validation and cleared the bindings. Sorry for the delay on my end- considering whether we can just remove the etag value here for all IAM policies or not, since that would obviate this change.

I added below logic to remove etag only when bindings are empty or null in this change and converted auto generated to handwritten resource, below is only extra logic added to existing iam_bucket

if policy.Bindings == nil || len(policy.Bindings) == 0 {
		policy.Etag = ""
}

kautikdk (commented):
Do update here when PR is ready for review. I don't see handwritten documentation file and tests for other cases.

kautikdk (commented):
I believe we also need to handwrite documentation of the data source, google_storage_bucket_iam_policy.

Apart from that, LGTM!

gurusai-voleti:
> I believe we also need to handwrite documentation of the data source, google_storage_bucket_iam_policy.
> 
> Apart from that, LGTM!

@kautikdk added google_storage_bucket_iam_policy doc to handwritten

mmv1/third_party/terraform/services/storage/iam_storage_bucket_test.go
```diff
@@ -0,0 +1,662 @@
...
+		ProtoV5ProviderFactories: acctest.ProtoV5ProviderFactories(t),
+		Steps: []resource.TestStep{
+			{
+				Config: testAccStorageBucketIamPolicy_destroy(),
+			},
+		},
+	})
+}
+
+func testAccStorageBucketIamPolicy_destroy() string {
```
kautikdk, line 31: nit: Individual config generator methods should be placed after all tests.

kautikdk (approved):

mmv1/third_party/terraform/services/storage/iam_storage_bucket_test.go
gurusai-voleti: updated

gurusai-voleti:
@rileykarson  could you please review this PR

pavandagagoogle:
@rileykarson - Could you provide review on the same

rileykarson (approved):
LGTM

Sorry for the delay- I was hoping to find a better solution than dropping to handwritten here, but ended short on cycles to do so.

PR #12945, opened 2025-02-03
Adds IP Filter feature support to google_storage_bucket resource.

kautikdk:
This PR enables users to specify IP Filtering configuration within `google_storage_bucket` resource.  

Original PR: https://github.com/GoogleCloudPlatform/magic-modules/pull/12576
Authors: @translucens, @kautikdk 

```release-note: enhancement
storage: added `ip_filter` to `google_storage_bucket` resource.
```

### DIFF
--- a/mmv1/third_party/terraform/website/docs/r/container_cluster.html.markdown
+++ b/mmv1/third_party/terraform/website/docs/r/container_cluster.html.markdown
@@ -304,6 +304,8 @@ region are guaranteed to support the same version.
     [Google Groups for GKE](https://cloud.google.com/kubernetes-engine/docs/how-to/role-based-access-control#groups-setup-gsuite) feature.
     Structure is [documented below](#nested_authenticator_groups_config).
 
+* `user_managed_keys_config` - (Optional) The custom keys configuration of the cluster Structure is [documented below](#nested_control_plane_endpoints_config).
+
 * `control_plane_endpoints_config` - (Optional) Configuration for all of the cluster's control plane endpoints.
     Structure is [documented below](#nested_control_plane_endpoints_config).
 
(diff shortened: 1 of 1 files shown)


kautikdk:
@roaks3 can you please continue your review on this PR?

mmv1/third_party/terraform/services/storage/resource_storage_bucket.go.tmpl
```diff
@@ -963,7 +1050,7 @@ func resourceStorageBucketRead(d *schema.ResourceData, meta interface{}) error {
 
 	// There seems to be some eventual consistency errors in some cases, so we want to check a few times
 	// to make sure it exists before moving on
-	res, err := config.NewStorageClient(userAgent).Buckets.Get(bucket).Do()
+	res, err := config.NewStorageClient(userAgent).Buckets.Get(bucket).Projection("full").Do()
```
roaks3, line 1053: I can't remember if this came up on the previous PR, but to confirm, this won't change the behavior for existing bucket users?

roaks3 (commented):
One small comment, but as mentioned, this was reviewed before and only blocked on tests (which are now passing)

mmv1/third_party/terraform/services/storage/resource_storage_bucket.go.tmpl
kautikdk: So as per this documentation: https://cloud.google.com/storage/docs/json_api/v1/buckets/get, `projection` is an optional query parameter which can have two values: `full` and `noAcl`. The default is `noAcl` which also excludes ipFilter property from the API response. Now to include IP filter in the response we need to use `full` projection which also adds `owner`, `acl` and `defaultObjectAcl` fields.  These extra fields are not being parsed by flatteners available in `google_storage_bucket` resource but they are available in [google_storage_bucket_acl](google/services/storage/resource_storage_bucket_acl.go). Using `full` projection should not break existing user script as we are not going to parse additional fields coming from the API response. I will still check with Feature team on this if they have plans to add IP filter in `noAcl` projection.

roaks3: Ok yea, seems worth checking if we can just leave it as `noAcl`, but what you described does sound safe. I would consider this non-blocking, but let me know when you hear back.

roaks3 (approved):

mmv1/third_party/terraform/services/storage/resource_storage_bucket.go.tmpl
kautikdk: Update: IP Filter config will be added to the default projection so we don't need to change projection to `full`.

kautikdk: I will update here once this is done from the API side, we will go ahead with the merge of this PR.

PR #12946, opened 2025-02-03
Add missing field to disks and support InstantSnapshot

karolgorc:
closes https://github.com/hashicorp/terraform-provider-google/issues/16426

Added some missing fields to `google_compute_disk`. One of them is `sourceInstantSnapshot` so i've added the resource to create instant snapshots into terraform.

One thing i'm having trouble testing is `sourceStorageObject`. It requires a Bucket with an image stored as `*.tar.gz`. Only way i've found of recreating this in an acceptance test is using `gcloud image export` into the bucket but don't know if we want something like that in our tests. Please LMK!

```release-note:enhancement
compute: added fields `architecture`, `source_instant_snapshot`, `source_storage_object`, `resource_manager_tags`  to `google_compute_disk`.
```

```release-note:new-resource
`google_compute_instant_snapshot`
```


mmv1/products/compute/InstantSnapshot.yaml
```diff
@@ -0,0 +1,104 @@
+---
```
slevenick, line 1: Needs copyright

mmv1/products/compute/InstantSnapshot.yaml
```diff
@@ -0,0 +1,104 @@
...
+    description: Labels to apply to this InstantSnapshot.
+    update_url: 'projects/{{project}}/zones{{zone}}/instantSnapshots/{{name}}/setLabels'
+    update_verb: 'POST'
+  - name: 'labelFingerprint'
+    type: Fingerprint
+    description: |
+      The fingerprint used for optimistic locking of this resource. Used
+      internally during updates.
+    output: true
+    update_url: 'projects/{{project}}/zones/{{zone}}/instantSnapshots/{{name}}/setLabels'
```
slevenick, line 102: No need for update_url or several of these fields on an output only field?

slevenick (commented):
Is labels updatable? The whole resource is marked immutable but it has update_url set

slevenick:
You should be able to add a tar.gz image into our test-fixtures file (for example for the IAP product: https://github.com/GoogleCloudPlatform/magic-modules/tree/main/mmv1/third_party/terraform/services/iap/test-fixtures) and then upload it via a google_storage_object resource

mmv1/products/compute/InstantSnapshot.yaml
karolgorc: I took the code for `labels` and `labelFingerprint` from Disk.yaml. 

Not really sure about the in's and out's of it but LabelFingerprint is updated while any label changes, so it's needed.

This happens when there isn't any update logic in `labelFingerprint`

image

karolgorc:
>Is labels updatable? The whole resource is marked immutable but it has update_url set

Seems that it doesn't matter. It doesn't force replace

image

slevenick:
> > Is labels updatable? The whole resource is marked immutable but it has update_url set
> 
> Seems that it doesn't matter. It doesn't force replace
> 
> image

Interesting! Can you add an update test for labels then?

mmv1/third_party/terraform/services/compute/resource_compute_instant_snapshot_test.go.tmpl
```diff
@@ -0,0 +1,159 @@
...
+	{{- end }}
+)
+
+func TestAccComputeInstantSnapshot_basicFeatures(t *testing.T) {
+	var is compute.InstantSnapshot
+	context := map[string]interface{}{
+		"random_suffix": acctest.RandString(t, 10),
+	}
+
+	resource.Test(t, resource.TestCase{
```
slevenick, line 26: Should be `acctest.VcrTest` instead of resource.Test, this is causing the error in replaying.

slevenick (changes requested):

karolgorc:
The disk had a wrong name in the `instant_snapshot_basic.tf.tmpl` file making it fail in parallel runs. Fixed

mmv1/products/compute/Disk.yaml
```diff
@@ -249,7 +272,17 @@ properties:
...
-
+  - name: 'sourceStorageObject'
+    type: String
+    description: |
+      The full Google Cloud Storage URI where the disk image is stored.
+      This file must be a gzip-compressed tarball whose name ends in .tar.gz or virtual machine disk whose name ends in vmdk.
+      Valid URIs may start with gs:// or https://storage.googleapis.com/.
+      This flag is not optimized for creating multiple disks from a source storage object.
+      To create many disks from a source storage object, use gcloud compute images import instead.
+    api_name: sourceStorageObject
```
slevenick, line 283: Same, remove

mmv1/products/compute/InstantSnapshot.yaml
```diff
@@ -0,0 +1,120 @@
...
+    update_verb: 'POST'
+    key_expander: ''
```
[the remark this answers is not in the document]
karolgorc: Removed. Seems to be working both with or without it. It's used in disk tough

```
- name: 'labelFingerprint'
    type: Fingerprint
    description: |
      The fingerprint used for optimistic locking of this resource.  Used
      internally during updates.
    output: true
    update_url: 'projects/{{project}}/zones/{{zone}}/disks/{{name}}/setLabels'
    update_verb: 'POST'
    key_expander: ''
```

mmv1/products/compute/Disk.yaml
```diff
@@ -249,7 +272,17 @@ properties:
...
+    ignore_read: true
+    diff_suppress_func: 'tpgresource.CompareSelfLinkOrResourceName'
```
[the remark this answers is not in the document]
karolgorc: Removed and tested.

mmv1/products/compute/Disk.yaml
karolgorc: Removed and tested.

mmv1/products/compute/Disk.yaml
```diff
@@ -144,6 +144,29 @@ properties:
...
+    type: String
+    description: |
+      The source instant snapshot used to create this disk. You can provide this as a partial or full URL to the resource.
+      For example, the following are valid values:
+
+      * `https://www.googleapis.com/compute/v1/projects/project/zones/zone/instantSnapshots/instantSnapshot`
+      * `projects/project/zones/zone/instantSnapshots/instantSnapshot`
+      * `zones/zone/instantSnapshots/instantSnapshot`
+
+    api_name: sourceInstantSnapshot
```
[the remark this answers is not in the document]
karolgorc: Removed and tested.

slevenick (approved):

PR #12950, opened 2025-02-03
add sweepers auto-resolver code

ScottSuarez:
Adding missing sweeper regions and introduce auto resolver code. Note that I haven't added unit tests since this code is still evolving. Need to work this for the {{name}} and {{prefixes}}.

```release-note:none

```


tools/sweeper_resolver/main.go
```diff
@@ -0,0 +1,293 @@
...
+		log.SetFlags(log.LstdFlags | log.Lshortfile)
+	}
+
+	logger.Printf("Starting program with directory path: %s", *dirPath)
+
+	if *dirPath == "" {
+		log.Fatal("Please provide a directory path using -dir flag")
+	}
+
+	allTests, errs := reader.ReadAllTests(*dirPath)
```
melinath, line 54: Ideally we'd want this to be integrated in the CI pipeline, which is already reading tests in order to do missing test detection. It would be great to only read them once, then do all the relevant processing of that data (for missing tests, sweeper regions, and sweepable prefix detection), and post any suggested changes as inline comments (or comments on the file in cases where inline doesn't work for some reason).

tools/sweeper_resolver/main.go
```diff
@@ -0,0 +1,293 @@
...
+	}
+}
+
+type MetaData struct {
+	Resource       string `yaml:"resource"`
+	GenerationType string `yaml:"generation_type"`
+	SourceFile     string `yaml:"source_file,omitempty"`
+}
+
+func main() {
```
melinath, line 34: We should run this using cobra to match our other tooling and because it's more readable and abstracts away some of the weird stuff about golang CLI implementation. It should return the basic data about what issues were detected, which would get used by the CI to generate inline comments (ideally using some kind of generic implementation that could handle issue creation for missing test detector and sweepable prefixes as well - maybe also for other arbitrary feedback we might want to create in the future. For example, we could return an object that contains a message plus the line number and file to attach it to. Any multi-line message should most likely be built using a go template.)

melinath (changes requested):

melinath:
This is pretty high level feedback so I'm not getting into details because I don't know how much will change - let me know if you have any questions!

PR #12955, opened 2025-02-04
Update cluster versions in TestAccContainerNodePool_concurrent

bbasata:
At time of this writing, the cluster default version is 1.31.5. The acceptance test fails because the API will not update such a cluster to 1.29.4.

This change updates the static versions and leaves hyper-breadcrumbs for the next traveler on this path.

Fixes https://github.com/hashicorp/terraform-provider-google/issues/21116

```release-note:none

```

### DIFF
--- a/mmv1/third_party/terraform/services/container/resource_container_node_pool_test.go.tmpl
+++ b/mmv1/third_party/terraform/services/container/resource_container_node_pool_test.go.tmpl
@@ -3985,7 +3985,7 @@ resource "google_container_cluster" "cluster" {
   deletion_protection = false
   network             = "%s"
   subnetwork          = "%s"
-  version             = "1.32.0-gke.1448000"
+  min_master_version  = "1.32.0-gke.1448000"
 }
 
 resource "google_container_node_pool" "np1" {
--- a/mmv1/third_party/terraform/services/container/resource_container_node_pool_test.go.tmpl
+++ b/mmv1/third_party/terraform/services/container/resource_container_node_pool_test.go.tmpl
@@ -3960,6 +3960,8 @@ resource "google_container_node_pool" "np1" {
   location           = "us-central1-a"
   cluster            = google_container_cluster.cluster.name
   initial_node_count = 2
+  // 2025-02-03: current default cluster version is 1.31.5-gke.1023000. This will change over time.
+  // Reference:  https://cloud.google.com/kubernetes-engine/docs/release-notes#current_versions
 }
 
 resource "google_container_node_pool" "np2" {
@@ -3967,6 +3969,8 @@ resource "google_container_node_pool" "np2" {
   location           = "us-central1-a"
   cluster            = google_container_cluster.cluster.name
   initial_node_count = 2
+  // 2025-02-03: current default cluster version is 1.31.5-gke.1023000. This will change over time.
+  // Reference:  https://cloud.google.com/kubernetes-engine/docs/release-notes#current_versions
 }
 `, cluster, networkName, subnetworkName, np1, np2)
 }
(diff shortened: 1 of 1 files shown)


trodge (approved):

bbasata:
Oh yes, this needs to be recorded anew. If someone gets to this before I do, that's more than welcome.

melinath:

melinath (commented):
@bbasata the recording is happening automatically - the failures are due to:

```
  
        Error: googleapi: Error 400: Node version "1.32.0-gke.1448000" must not have a greater minor version than master version "1.31.4-gke.1256000".
        Details:
        [
          {
            "requestId": "0x19cad9635fdbce46"
          }
        ]
        , badRequest
        
          with google_container_node_pool.np1,
          on terraform_plugin_test.tf line 11, in resource "google_container_node_pool" "np1":
          11: resource "google_container_node_pool" "np1" {
        
        
        Error: googleapi: Error 400: Node version "1.32.0-gke.1448000" must not have a greater minor version than master version "1.31.4-gke.1256000".
        Details:
        [
          {
            "requestId": "0x2a7c99b7fe91d372"
          }
        ]
        , badRequest
        
          with google_container_node_pool.np2,
          on terraform_plugin_test.tf line 23, in resource "google_container_node_pool" "np2":
          23: resource "google_container_node_pool" "np2" {
```

Possibly relatedly - could this test be rewritten to use the [`google_container_engine_versions` data source](https://registry.terraform.io/providers/hashicorp/google/latest/docs/data-sources/container_engine_versions) instead of hardcoding versions?

bbasata:
> @bbasata the recording is happening automatically - the failures are due to:

Got it, thanks! I think I see what I missed. The node version cannot exceed the cluster (control plane) version. git push. Waiting on checks ...

> could this test be rewritten to use the [google_container_engine_versions data source](https://registry.terraform.io/providers/hashicorp/google/latest/docs/data-sources/container_engine_versions) instead of hardcoding versions?

I'd like to give it a try. If this PR goes green (maybe!), how about we merge, and then follow up with the nicer solution?

melinath:
>  If this PR goes green (maybe!), how about we merge, and then follow up with the nicer solution?

SGTM

mmv1/third_party/terraform/services/container/resource_container_node_pool_test.go.tmpl
```diff
@@ -3953,49 +3953,63 @@ resource "google_container_cluster" "cluster" {
   deletion_protection = false
   network             = "%s"
   subnetwork          = "%s"
+  version             = "1.32.0-gke.1448000"
```
melinath, line 3956: I think these all need to be [min_master_version](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/container_cluster#min_master_version-1)?

```suggestion
  min_master_version  = "1.32.0-gke.1448000"
```

melinath (changes requested):
Error is:

```
        Error: Unsupported argument
        
          on terraform_plugin_test.tf line 9, in resource "google_container_cluster" "cluster":
           9:   version             = "1.32.0-gke.1448000"
        
        An argument named "version" is not expected here.
```

mmv1/third_party/terraform/services/container/resource_container_node_pool_test.go.tmpl
melinath: specifically the ones for google_container_cluster - node pool is correct to use [version](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/container_node_pool#version-1).

bbasata: Makes sense!

KatrinaHoffert:
Thanks for fixing this! I'm not sure why we ever hard coded it.

Bumping the version works for now. A perhaps cleaner long term solution would be to just choose some simpler field. The most important thing is that the operation takes a little bit of time so that there is actual concurrency (since several fields are nearly instant to update). There's many possibilities. The `node_config` field has a few that are probably cheap and easy. I particularly eye [`metadata`](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/container_cluster#metadata-1) as one that seems like it'd be a straightforward one to update. Seems like that'd improve coverage too, as we appear to not have any tests for that field already. Though if that makes it feel too risky, `disk_size_gb` is another option that at least has existing tests.

melinath (approved):
approving this for now to get the test passing; we can consider a longer-term fix separately.

wyardley:
I updated it in https://github.com/GoogleCloudPlatform/magic-modules/pull/13161 with an approach that _hopefully_ accomplishes the goal of the test in a more durable way? I also sized down the node pools and deleted the default pool, which should if nothing else, make it faster and cheaper to run.

I'd come across this PR, but hadn't read through all the comments before (since I'd noticed it was already fixed). I agree with the above, though, that probably updating a different field might make more sense, along with some additional comments on the goals behind the test.

PR #12957, opened 2025-02-04
Make scope optional for google_compute_(region_)resize_request  so that the default can be taken from the provider configuration

askubis:
Makes it consistent with the other compute resources, especially in the instance_group_manager family.

```release-note:bug
compute: fixed google_compute_(region_)resize_request requiring region/zone to be specified in all cases. They can now be pulled from the provider.
```

### DIFF
--- a/mmv1/products/compute/RegionResizeRequest.yaml
+++ b/mmv1/products/compute/RegionResizeRequest.yaml
@@ -59,7 +59,7 @@ parameters:
     description: |
       The reference of the compute region scoping this request.
     url_param_only: true
-    required: true
+    required: false
     resource: 'Region'
     imports: 'name'
   - name: 'instanceGroupManager'
--- a/mmv1/products/compute/RegionResizeRequest.yaml
+++ b/mmv1/products/compute/RegionResizeRequest.yaml
@@ -57,8 +57,10 @@ parameters:
   - name: 'region'
     type: ResourceRef
     description: |
-      The reference of the compute region scoping this request.
-    url_param_only: true
+      The reference of the compute region scoping this request. If it is not provided, the provider region is used.
+    custom_flatten: 'templates/terraform/custom_flatten/name_from_self_link.tmpl'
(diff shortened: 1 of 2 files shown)


mmv1/products/compute/RegionResizeRequest.yaml
```diff
@@ -59,7 +59,7 @@ parameters:
     description: |
       The reference of the compute region scoping this request.
     url_param_only: true
```
rileykarson, line 61:
```suggestion
      The reference of the compute region scoping this request. If it is not provided, the provider region is used.
    custom_flatten: 'templates/terraform/custom_flatten/name_from_self_link.tmpl'
    custom_expand: 'templates/terraform/custom_expand/resourceref_with_validation.go.tmpl'
    default_from_api: true
```

mmv1/products/compute/ResizeRequest.yaml
```diff
@@ -65,7 +65,7 @@ parameters:
     description: |
       The reference of the compute zone scoping this request.
     url_param_only: true
```
rileykarson, line 67:
```suggestion
    description: |
      The reference of the compute zone scoping this request. If it is not provided, the provider zone is used.
    custom_flatten: 'templates/terraform/custom_flatten/name_from_self_link.tmpl'
    custom_expand: 'templates/terraform/custom_expand/resourceref_with_validation.go.tmpl'
    default_from_api: true
```

rileykarson (commented):
Suggestions inline to apply identical settings to https://github.com/GoogleCloudPlatform/magic-modules/blob/f79b8e1529f425d8345efa19fb5d0a017f5c5229/mmv1/products/compute/Address.yaml#L94-L104 / https://github.com/GoogleCloudPlatform/magic-modules/blob/f79b8e1529f425d8345efa19fb5d0a017f5c5229/mmv1/products/compute/RegionDisk.yaml#L87-L95, for consistency.

rileykarson (approved):

PR #12966, opened 2025-02-05
Adds storage_operation to handle LRO related operations

googlyrahman:
```release-note: none

```

### DIFF
--- a/mmv1/third_party/terraform/services/storage/storage_operation.go
+++ b/mmv1/third_party/terraform/services/storage/storage_operation.go
@@ -1,0 +1,80 @@
+// Copyright (c) HashiCorp, Inc.
+// SPDX-License-Identifier: MPL-2.0
+package storage
+
+import (
+	"encoding/json"
+	"errors"
+	"fmt"
+	"time"
+
+	"github.com/hashicorp/terraform-provider-google/google/tpgresource"
+	transport_tpg "github.com/hashicorp/terraform-provider-google/google/transport"
+)
+
+type StorageOperationWaiter struct {
+	Config    *transport_tpg.Config
+	UserAgent string
+	SelfLink  string
+	tpgresource.CommonOperationWaiter
+}
+
+func (w *StorageOperationWaiter) QueryOp() (interface{}, error) {
+	if w == nil {
+		return nil, fmt.Errorf("Cannot query operation, it's unset or nil.")
+	}
+	// Returns the proper get.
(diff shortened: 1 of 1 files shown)


googlyrahman:
I would suggest to compare the code with the operation template, If you're looking for  reason why autogeneration can't be preferred here, you can find that here - https://yaqs.corp.google.com/eng/q/1126624485005328384

With that said, I need prioritization on this PR, I'll communicate the same offline to you.

NickElliot:
reassigning because I am OOO for next few days

mmv1/third_party/terraform/services/storage/storage_operation.go
```diff
@@ -0,0 +1,80 @@
...
+	"time"
+
+	"github.com/hashicorp/terraform-provider-google/google/tpgresource"
+	transport_tpg "github.com/hashicorp/terraform-provider-google/google/transport"
+)
+
+type StorageOperationWaiter struct {
+	Config    *transport_tpg.Config
+	UserAgent string
+	SelfLink  string
```
kautikdk, line 18: I didn't find SelfLink in the documentation of the GCS operations resource, https://cloud.google.com/storage/docs/json_api/v1/operations.
Is this new field or documentation is not updated yet?

googlyrahman: I believe this is pending documentation task, Storage LRO does return `selfLink`,  you can verify the same in [discovery doc](https://source.corp.google.com/piper///depot/google3/third_party/py/googlecloudsdk/generated_clients/apis/storage/v1/storage_v1_messages.py;l=1107)

You might also want to look at the design doc, [go/gcs-lro-api-control-plane](http://go/gcs-lro-api-control-plane)

hao-nan-li (commented):
Thanks for adding this in. Are these functions currently being used? If yes do you expect any diff for existing storage operation tests?

googlyrahman:
@hao-nan-li, These functions are not used anywhere, We're planning to add the support for terraform anywhere-cache which requires LRO code, Hence this PR.

melinath:
@modular-magician reassign-reviewer hao-nan-li

hao-nan-li:
> @hao-nan-li, These functions are not used anywhere, We're planning to add the support for terraform anywhere-cache which requires LRO code, Hence this PR.

I see, in this case I don't think a release-note is needed. I'll handle that.

hao-nan-li (approved):

PR #12967, opened 2025-02-05
EphemeralWriteOnly: add `secretAccessKeyWo`

BBBmau:

```release-note:enhancement
bigquerydatatransfer: added `secret_access_key_wo` write-only field to `google_bigquery_data_transfer_config` resource
```


wj-chen:
Thanks for the PR, could you share more about this change? Is it in response to an issue or bug or feature request?

BBBmau:
@wj-chen this comes from supporting ephemeral write-only attributes, however I am running into a blocker that you could potentially help with. The context:

it looks like the field sensitive_params.secret_access_key doesn't work anymore due to params keys being updated (assuming that the secret_access_key was in the params list in the first place).

Currently looking for some sort of record of this, whether or not it was set in the past. any help would be appreciated!

This comes from doing some testing with the field itself as it's a write-only attribute candidate. Wasn't able to use it normally which led to the discovery.

image

https://cloud.google.com/bigquery/docs/cloud-storage-transfer#bq

wj-chen:
> ~@wj-chen this comes from supporting ephemeral write-only attributes, however I am running into a blocker that you could potentially help with. The context: it looks like the field sensitive_params.secret_access_key doesn't work anymore due to params keys being updated (assuming that the secret_access_key was in the params list in the first place). Currently looking for some sort of record of this, whether or not it was set in the past. any help would be appreciated! This comes from doing some testing with the field itself as it's a write-only attribute candidate. Wasn't able to use it normally which led to the discovery. image https://cloud.google.com/bigquery/docs/cloud-storage-transfer#bq~
> 
> Edit: this has been resolved after coming across: https://cloud.google.com/bigquery/docs/s3-transfer#bq

Glad you were able to find the documentation on the field. Is the support for write-only attributes being rolled out to all google-provider (and other providers)? Do I understand correctly that the recommendation is to use the write-only attribute and deprecate the other one over time? I took only a quick glance at https://developer.hashicorp.com/terraform/plugin/sdkv2/resources/write-only-arguments so maybe it's covered there but  what happens if user wants to change the value if it's not part of the planning phase?

BBBmau:
> > ~@wj-chen this comes from supporting ephemeral write-only attributes, however I am running into a blocker that you could potentially help with. The context: it looks like the field sensitive_params.secret_access_key doesn't work anymore due to params keys being updated (assuming that the secret_access_key was in the params list in the first place). Currently looking for some sort of record of this, whether or not it was set in the past. any help would be appreciated! This comes from doing some testing with the field itself as it's a write-only attribute candidate. Wasn't able to use it normally which led to the discovery. image https://cloud.google.com/bigquery/docs/cloud-storage-transfer#bq~
> > Edit: this has been resolved after coming across: https://cloud.google.com/bigquery/docs/s3-transfer#bq
> 
> Glad you were able to find the documentation on the field. Is the support for write-only attributes being rolled out to all google-provider (and other providers)? Do I understand correctly that the recommendation is to use the write-only attribute and deprecate the other one over time? I took only a quick glance at https://developer.hashicorp.com/terraform/plugin/sdkv2/resources/write-only-arguments so maybe it's covered there but what happens if user wants to change the value if it's not part of the planning phase?

Currently there is no way for Terraform to allow you to update a write-only value. The design perspective is that every ephemeral value is different on every action performed by terraform.

The current work around though that would allow this would be a sidecar field that's meant for triggering an update on an ephemeral value assuming you are in a case where the ephemeral write-only value itself is in a stable state that doesn't change unless manually triggered by a user. Hope this was clear.

On a side note I'll be moving this to `ready for review` as the `secret_access_key` doesn't contain tests due to the nature of the field. I've implemented it in a way that replicates the field when not being used in a `amazon_s3`.

the `tfc_conversion` can be ignored for now as it's being addressed in this PR on downstream tgc repo: https://github.com/GoogleCloudPlatform/terraform-google-conversion/pull/3412

we required the use of `sdk/v2 v2.26.0` while tgc uses `v2.23.0`. The linked PR would bump the needed dependency.

mmv1/templates/terraform/encoders/bigquery_data_transfer.go.tmpl
```diff
@@ -40,6 +40,13 @@ for _, sp := range sensitiveParams {
 	}
 }
 
+for _, sp := range sensitiveWoParams {
+	if auth, _ := d.GetRawConfigAt(cty.GetAttrPath("sensitive_params").IndexInt(0).GetAttr(sp)); !auth.IsNull() {
+		sp = sp[:len(sp)-3] // removes _wo from the end of the param name
```
wj-chen, line 45: Is it a convention that these should always be suffixed "_wo"?  Can we communicate it more in the YAML comment or somewhere so people know the name takes part in this logic?

wj-chen:
> > > ~@wj-chen this comes from supporting ephemeral write-only attributes, however I am running into a blocker that you could potentially help with. The context: it looks like the field sensitive_params.secret_access_key doesn't work anymore due to params keys being updated (assuming that the secret_access_key was in the params list in the first place). Currently looking for some sort of record of this, whether or not it was set in the past. any help would be appreciated! This comes from doing some testing with the field itself as it's a write-only attribute candidate. Wasn't able to use it normally which led to the discovery. image https://cloud.google.com/bigquery/docs/cloud-storage-transfer#bq~
> > > Edit: this has been resolved after coming across: https://cloud.google.com/bigquery/docs/s3-transfer#bq
> > 
> > 
> > Glad you were able to find the documentation on the field. Is the support for write-only attributes being rolled out to all google-provider (and other providers)? Do I understand correctly that the recommendation is to use the write-only attribute and deprecate the other one over time? I took only a quick glance at https://developer.hashicorp.com/terraform/plugin/sdkv2/resources/write-only-arguments so maybe it's covered there but what happens if user wants to change the value if it's not part of the planning phase?
> 
> Currently there is no way for Terraform to allow you to update a write-only value. The design perspective is that every ephemeral value is different on every action performed by terraform.
> 
> The current work around though that would allow this would be a sidecar field that's meant for triggering an update on an ephemeral value assuming you are in a case where the ephemeral write-only value itself is in a stable state that doesn't change unless manually triggered by a user. Hope this was clear.
> 
> On a side note I'll be moving this to `ready for review` as the `secret_access_key` doesn't contain tests due to the nature of the field. I've implemented it in a way that replicates the field when not being used in a `amazon_s3`.
> 
> the `tfc_conversion` can be ignored for now as it's being addressed in this PR on downstream tgc repo: [GoogleCloudPlatform/terraform-google-conversion#3412](https://github.com/GoogleCloudPlatform/terraform-google-conversion/pull/3412)
> 
> we required the use of `sdk/v2 v2.26.0` while tgc uses `v2.23.0`. The linked PR would bump the needed dependency.

Thanks. I'm trying to understand from the user side, if they create a resource with a `secret_access_key_wo`, then regenerate an access key, and want to update their Terraform config with the new key, what will happen when the field is write-only? Sounds like it just doesn't update, unless it's marked as `requires_replace`?

BBBmau:
> Thanks. I'm trying to understand from the user side, if they create a resource with a secret_access_key_wo, then regenerate an access key, and want to update their Terraform config with the new key, what will happen when the field is write-only? Sounds like it just doesn't update, unless it's marked as requires_replace?

The idea is that write-only attributes are different during every action performed by Terraform. Although some values may stay stable it doesn't allow for identifying whether an update of a value occurred or not. Since write-only attributes can accept ephemeral values, write-only attribute configuration values are not expected to be consistent between plan and apply.

So for a typical user they wouldn't need to handle any `require_replace` as it's already expected through Terraform's eyes that the value for this is expected to not remain consistent. @wj-chen

mmv1/templates/terraform/encoders/bigquery_data_transfer.go.tmpl
wj-chen: Rest of the PR LGTM. Wanted to see if you have ideas to make the `sp = sp[:len(sp)-3]` part more future-proof.

wj-chen:
> > Thanks. I'm trying to understand from the user side, if they create a resource with a secret_access_key_wo, then regenerate an access key, and want to update their Terraform config with the new key, what will happen when the field is write-only? Sounds like it just doesn't update, unless it's marked as requires_replace?
> 
> The idea is that write-only attributes are different during every action performed by Terraform. Although some values may stay stable it doesn't allow for identifying whether an update of a value occurred or not. Since write-only attributes can accept ephemeral values, write-only attribute configuration values are not expected to be consistent between plan and apply.
> 
> So for a typical user they wouldn't need to handle any `require_replace` as it's already expected through Terraform's eyes that the value for this is expected to not remain consistent. @wj-chen

I see thanks. This PR looks good to me, but two more questions on the best use case of write-only fields in general:

* Since an AWS secret access key stays unchanged unless the user rotates it, is the use of write-only in this case more about skipping storing the value and removing diffs, and less about the value being ephemeral?
* Does changing a field to/from write-only considered a breaking change?

BBBmau:
> * Since an AWS secret access key stays unchanged unless the user rotates it, is the use of write-only in this case more about skipping storing the value and removing diffs, and less about the value being ephemeral?

WriteOnly is meant to replace sensitive fields since although sensitive prevent sensitive values from being outputted on terraform plans they are still present in the `terraform.tf.state`. WriteOnly is meant to prevent any sort of sensistive data from being stored and only used for when it's needed.

> * Does changing a field to/from write-only considered a breaking change?

It is, however we intend to replace sensitive attributes with WriteOnly in the next major release. Until then we have a `validation.PreferWriteOnlyAttributes` that will inform users to use the new field over the sensitive field.

The PR that adds this validation in the mmv1 generation can be found here: https://github.com/GoogleCloudPlatform/magic-modules/pull/13048

BBBmau:
> 
> I see thanks. This PR looks good to me, but two more questions on the best use case of write-only fields in general:
> 
> * Since an AWS secret access key stays unchanged unless the user rotates it, is the use of write-only in this case more 
about skipping storing the value and removing diffs, and less about the value being ephemeral?

The use-case is more to prevent sensitive information from being stored in state. You can prevent values from beiung shown in the plan with `Sensitive: true` but that doesn't prevent the sensitive info from being shown in `terraform.tf.state`. This is where `WriteOnly` comes in which allows you to use sensitive information when it's necessary while also not needing to store it.

> * Does changing a field to/from write-only considered a breaking change?

It does, however we intend to keep both the WriteOnly and non-Writeonly included with their being Validation added to push users to use the new WriteOnly Attribute. The non-writeOnly attribute would be removed in an upcoming major release. This is an example of the validation being shown to users. PR would support Resource Validation to support this: https://github.com/GoogleCloudPlatform/magic-modules/pull/13048

image

mmv1/templates/terraform/validation/bigquery_data_transfer_config.go.tmpl
```diff
@@ -0,0 +1 @@
+validation.PreferWriteOnlyAttribute(cty.GetAttrPath("sensitive_params").IndexAt(0).GetAttr("secret_access_key"),cty.GetAttrPath("sensitive_params_wo").IndexAt(0).GetAttr("secret_access_key_wo"))
```
wj-chen, line 1: This should be "sensitive_params"

mmv1/templates/terraform/encoders/bigquery_data_transfer.go.tmpl
BBBmau: yeah it's convention to mark write-only attributes with the `_wo` suffix. I can include this as a comment.

mmv1/templates/terraform/validation/bigquery_data_transfer_config.go.tmpl
BBBmau: done

mmv1/templates/terraform/encoders/bigquery_data_transfer.go.tmpl
wj-chen: A comment would help, thanks.

wj-chen (approved):

mmv1/products/bigquerydatatransfer/Config.yaml
```diff
@@ -41,6 +41,7 @@ custom_code:
   post_create: 'templates/terraform/post_create/set_computed_name.tmpl'
   pre_update: 'templates/terraform/pre_update/bigquerydatatransfer_config.tmpl'
   custom_import: 'templates/terraform/custom_import/bigquery_data_transfer_self_link_as_name_set_location.go.tmpl'
+  # validation: 'templates/terraform/validation/bigquery_data_transfer_config.go.tmpl'
```
wj-chen, line 44: Meant to be uncommented?

mmv1/templates/terraform/encoders/bigquery_data_transfer.go.tmpl
```diff
@@ -39,6 +39,14 @@ for _, sp := range sensitiveParams {
 		params[sp] = auth.(string)
 	}
 }
+{{- if ne $.Compiler "terraformgoogleconversion-codegen" }}
+for _, sp := range sensitiveWoParams {
+	if auth, _ := d.GetRawConfigAt(cty.GetAttrPath("sensitive_params").IndexInt(0).GetAttr(sp)); !auth.IsNull() {
+		sp = sp[:len(sp)-3] // _wo is convention for write-only params and are removed here
```
wj-chen, line 45: Thanks, can we add a comment in Config.yaml too? The intention is when people add a new wo field there, they would know something happens in constants/ and encoders/ that requires the new field to be named a certain way and be added to related blocks in those files

mmv1/products/bigquerydatatransfer/Config.yaml
BBBmau: will be commented out once https://github.com/GoogleCloudPlatform/magic-modules/pull/13048 is merged.

mmv1/templates/terraform/encoders/bigquery_data_transfer.go.tmpl
BBBmau: added

wj-chen (approved):

BBBmau:
recent pushes were just rebasing

PR #12972, opened 2025-02-05
Fix various product headings in the Registry

rileykarson:
Saw a wall of BackupDR datasources (filed https://github.com/hashicorp/terraform-provider-google/issues/21261 to mitigate in the future) and did a pass of product names to fix caps / spacing / other minor issues.

We should probably validate "Cloud" prefixes and naming for remaining v2s after b/392147882. I almost removed DM's but then checked and it's inconsistently referred to as "Google Cloud Deployment Manager", "Cloud Deployment Manager", and "Deployment Manager" where I wouldn't have expected "Cloud Deployment Manager" to be in the mix.

To review, I'd recommend checking the handwritten resources and datasources for these products to make sure I didn't orphan any.

```release-note:none
```

### DIFF
--- a/mmv1/products/apihub/product.yaml
+++ b/mmv1/products/apihub/product.yaml
@@ -13,7 +13,7 @@
 
 ---
 name: Apihub
-display_name: API hub
+display_name: "API Hub"
 scopes:
   - https://www.googleapis.com/auth/cloud-platform
 versions:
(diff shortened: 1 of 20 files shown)


c2thorn (approved):
ack'd on validating "cloud" prefixes after b/392147882, made a note in the bug

rileykarson:
That wouldn't necessarily be part of that work fwiw- just voicing thoughts about the next nitpicks to address in the Registry!

PR #12800, opened 2025-01-18
EphemeralWriteOnly: add `secret_data_wo` in `google_secret_version`

BBBmau:
Here's a demo of the payload showcasing the new write-only attribute with logs:
```hcl
2025-02-05T16:38:16.137-0800 [DEBUG] provider.terraform-provider-google_9.9.9_darwin_arm64: ---[ REQUEST ]---------------------------------------
2025-02-05T16:38:16.137-0800 [DEBUG] provider.terraform-provider-google_9.9.9_darwin_arm64: POST /v1/projects/hc-terraform-testing/secrets/secret-version:addVersion?alt=json HTTP/1.1
2025-02-05T16:38:16.137-0800 [DEBUG] provider.terraform-provider-google_9.9.9_darwin_arm64: Host: secretmanager.googleapis.com
2025-02-05T16:38:16.137-0800 [DEBUG] provider.terraform-provider-google_9.9.9_darwin_arm64: User-Agent: Terraform/1.11.0-beta1 (+https://www.terraform.io) Terraform-Plugin-SDK/2.36.0 terraform-provider-google/dev6
2025-02-05T16:38:16.137-0800 [DEBUG] provider.terraform-provider-google_9.9.9_darwin_arm64: Content-Length: 40
2025-02-05T16:38:16.137-0800 [DEBUG] provider.terraform-provider-google_9.9.9_darwin_arm64: Content-Type: application/json
2025-02-05T16:38:16.137-0800 [DEBUG] provider.terraform-provider-google_9.9.9_darwin_arm64: Accept-Encoding: gzip
2025-02-05T16:38:16.137-0800 [DEBUG] provider.terraform-provider-google_9.9.9_darwin_arm64
2025-02-05T16:38:16.137-0800 [DEBUG] provider.terraform-provider-google_9.9.9_darwin_arm64: {
2025-02-05T16:38:16.137-0800 [DEBUG] provider.terraform-provider-google_9.9.9_darwin_arm64:  "payload": {
2025-02-05T16:38:16.137-0800 [DEBUG] provider.terraform-provider-google_9.9.9_darwin_arm64:   "data": "c2VjcmV0LWRhdGE="
2025-02-05T16:38:16.137-0800 [DEBUG] provider.terraform-provider-google_9.9.9_darwin_arm64:  }
2025-02-05T16:38:16.137-0800 [DEBUG] provider.terraform-provider-google_9.9.9_darwin_arm64: }
[...]

PR #12974, opened 2025-02-05
`mmv1`: add `hashicorp/gocty/cty` in resource.go.tmpl

BBBmau:
Was running into issues with the provider generation when a resource required the use of `cty`. Instead of using `hashicorp/go-cty/cty` which is the intended dependency to use in `terraform-sdk-v2` it was using the `zclconf/go-cty/cty` which caused build errors upon generation. This exact build error was ran into when working on the following [PR](https://github.com/GoogleCloudPlatform/magic-modules/pull/12800#issuecomment-2635871281). When using the appropriate dependency it builds and runs just fine.

No diffs on generation for all other resources.

```release-note:none

```


c2thorn (approved):
So we shouldn't expect any diffs here, but it will fix your other PR?

BBBmau:
> So we shouldn't expect any diffs here, but it will fix your other PR?

yes correct! I'll be able to generate the provider within that PR with no extra step now. (extra step was needing to change from `zclconfg` -> `hashicorp`

PR #12975, opened 2025-02-05
make authz extension wire_format O+C

c2thorn:
fixes https://github.com/hashicorp/terraform-provider-google/issues/21129
fixes https://github.com/hashicorp/terraform-provider-google/issues/21120
fixes https://github.com/hashicorp/terraform-provider-google/issues/21180
fixes https://github.com/hashicorp/terraform-provider-google/issues/21171

not sure if the API changed behavior since resource implementation, but the API no longer accepts the provider supplying the default value. Changing the field to be O+C to allow the API to set the default itself.

Should _not_ be a breaking change due to making it O+C while also moving away from a default that is just straight up invalid.

```release-note:bug
networkservices: fixed a bug with `google_network_services_authz_extension.wire_format` sending an invalid default value by removing the Terraform default and letting the API set the default.
```

### DIFF
--- a/mmv1/third_party/terraform/services/networkservices/resource_network_services_authz_extension_test.go
+++ b/mmv1/third_party/terraform/services/networkservices/resource_network_services_authz_extension_test.go
@@ -47,13 +47,13 @@ func TestAccNetworkServicesAuthzExtension_update(t *testing.T) {
 func testAccNetworkServicesAuthzExtension_start(context map[string]interface{}) string {
 	return acctest.Nprintf(`
 resource "google_compute_network" "default" {
-  name                    = "lb-network"
+  name                    = "lb-network%{random_suffix}"
   project                 = "%{project}"
   auto_create_subnetworks = false
 }
 
 resource "google_compute_subnetwork" "default" {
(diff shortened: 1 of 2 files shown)


c2thorn:

mmv1/third_party/terraform/services/networkservices/resource_network_services_authz_extension_test.go
```diff
@@ -47,21 +47,21 @@ func TestAccNetworkServicesAuthzExtension_update(t *testing.T) {
 func testAccNetworkServicesAuthzExtension_start(context map[string]interface{}) string {
 	return acctest.Nprintf(`
 resource "google_compute_network" "default" {
-  name                    = "lb-network"
+  name                    = "lb-network%{random_suffix}"
```
slevenick, line 50: Should we also change these to `tf-test`...?

c2thorn: good idea

slevenick (approved):

PR #12981, opened 2025-02-06
Fixed error message when naming disks

Dawid212:
### Related :

- Resolves: https://github.com/hashicorp/terraform-provider-google/issues/18165

Added validation for disk names in `google_compute_disk`

```release-note:bug
compute: added validation for disk names in `google_compute_disk`
```

### DIFF
--- a/mmv1/products/compute/Disk.yaml
+++ b/mmv1/products/compute/Disk.yaml
@@ -294,6 +294,8 @@ properties:
       character, which cannot be a dash.
     required: true
     immutable: true
+    validation:
+      function: 'verify.ValidateGCEName'
   - name: 'size'
     type: Integer
     description: |


melinath:
Drive-by comment: we need to confirm that this validation function will be correct in all cases before adding it; otherwise, it may be a breaking (and undesired) change.

melinath:

ScottSuarez (approved):
Matches validation requirements [found in the function](https://github.com/ScottSuarez/magic-modules/blob/f7ec8e741b41e69825e825c6e4deb897a3603617/mmv1/third_party/terraform/verify/validation.go#L98) and [rest docs](https://cloud.google.com/compute/docs/reference/rest/v1/disks#:~:text=RFC3339%20text%20format.-,name,-string).

PR #12987, opened 2025-02-06
Backfill autogen status tag

slevenick:

```release-note:none
```

### DIFF
--- a/mmv1/products/apphub/Application.yaml
+++ b/mmv1/products/apphub/Application.yaml
@@ -29,6 +29,7 @@ timeouts:
   update_minutes: 20
   delete_minutes: 20
 autogen_async: true
+autogen_tag: QXBwbGljYXRpb24=
 async:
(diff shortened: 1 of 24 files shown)


mmv1/products/apphub/Application.yaml
```diff
@@ -29,6 +29,7 @@ timeouts:
   update_minutes: 20
   delete_minutes: 20
 autogen_async: true
+autogen_tag: QXBwbGljYXRpb24=
```
melinath, line 32: I think these should all be autogen_status?

```suggestion
autogen_status: QXBwbGljYXRpb24=
```

melinath (changes requested):

mmv1/products/apphub/Application.yaml
slevenick: Hah, was just in 2 files I must have zoned out when typing them

melinath (approved):

PR #12992, opened 2025-02-07
Split resource schema diff returned to magician by added/modified/removed

melinath:
This is a prerequisite to https://github.com/hashicorp/terraform-provider-google/issues/21287 because it will allow distinguishing between added, modified, and removed resources.

I also made the computation of the primary schema diff happen at build time, since it is a static value.

```release-note:none

```


melinath:
Schema diff looks correct in https://pantheon.corp.google.com/cloud-build/builds/2b396317-82c4-45e2-93ba-738baaa91c43;step=17?e=-13802955&inv=1&invt=Abo8vA&mods=logs_tg_staging&project=graphite-docker-images

melinath:
@ScottSuarez this is ready for review.

melinath:
I'm not sure why the `downstreams` jobs are being cancelled.

ScottSuarez (approved):

ScottSuarez:
> I'm not sure why the `downstreams` jobs are being cancelled.

hmm, strange. It's probably related to the concurrency group but I can't see anything obviously wrong here. Any ideas?

https://github.com/GoogleCloudPlatform/magic-modules/blob/main/.github/workflows/downstreams.yml#L14

melinath:
sounds from your PR like this is not a blocker for this PR, so moving forward with merge

PR #12995, opened 2025-02-07
Allow storage_operation to gracefully cancel the LRO if an interrupt is received to prevent the orphaned LRO

googlyrahman:
```release-note:none
Made storage_operation to gracefully cancel the LRO if an interrupt is received to prevent the orphaned LRO
```


shuyama1 (commented):
Is there any possible way that we can test this?

googlyrahman:
@shuyama1, I doubt that we would be able to test this in simple way. The idea is inspired from 
https://github.com/c2thorn/magic-modules/blob/c6f047222b4e420ef1998431a000f18ffa6dd03d/third_party/terraform/resources/resource_compute_instance_group_manager.go.erb#L393

I can't find any tests corresponding to this piece of code for that file.

PR #12998, opened 2025-02-07
log github event metadata to debug canceled workflow

ScottSuarez:
https://github.com/GoogleCloudPlatform/magic-modules/actions/runs/13204479522/job/36864303644?pr=12992

- The called workflow using `workflow_call` [should inherit the full event payload](https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#workflow_call) from the calling workflow, including `github.event.pull_request` context
- However, we're seeing the PR number missing from concurrency group names, suggesting this inheritance isn't working as expected
- Added debug steps to inspect the available context in the called workflow
- Switching to using `github.event.number` For whatever reason `github.event.pull_request.number` is sometimes missing.

```release-note:none

```

### DIFF
--- a/.github/workflows/build-downstream.yml
+++ b/.github/workflows/build-downstream.yml
@@ -9,10 +9,6 @@ on:
         required: true
         type: string
 
-concurrency:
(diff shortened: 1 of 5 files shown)


melinath (approved):
This seems reasonable but I also note that there are still builds being cancelled on this PR, which seems unexpected?

Also, what about the merge_group event - will that propagate correctly?

PR #13006, opened 2025-02-07
Fix integrations test due to missing permission

roaks3:
Fixes https://github.com/hashicorp/terraform-provider-google/issues/21246

Note that this test doesn't run in VCR

```release-note:none

```

### DIFF
--- a/mmv1/products/integrations/Client.yaml
+++ b/mmv1/products/integrations/Client.yaml
@@ -56,6 +56,9 @@ examples:
     primary_resource_id: 'example'
   - name: 'integrations_client_full'
     primary_resource_id: 'example'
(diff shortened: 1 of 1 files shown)


trodge (approved):

melinath (changes requested):
Drive-by comment: it's weird that this is flakey if the problem is a missing bootstrap... I'd expect that to be a 100% failure. Maybe there's something else going on here?

melinath:
Maybe IAM permissions take a minute to propagate to newly-created crypto keys? I wonder if using a bootstrapped crypto key would solve this?

roaks3:
This appears to have been address by the service team.

PR #13021, opened 2025-02-10
Fix: Address issue hashicorp#19621 in discoveryengine module

stefan-malin-db:
Fixed https://github.com/hashicorp/terraform-provider-google/issues/19621
Fixed https://github.com/hashicorp/terraform-provider-google/issues/20259

```release-note:bug
discoveryengine: fixed bug preventing creation of `google_discovery_engine_target_site` resources
```


melinath:
Just want to confirm what issue this relates to? https://github.com/hashicorp/terraform-provider-google/pull/19261 is a PR and doesn't seem related to this change.

melinath (changes requested):
This PR doesn't do anything - please let me know what you're trying to do and what issue you're trying to fix and I may be able to give some pointers.

stefan-malin-db:
> Just want to confirm what issue this relates to? [hashicorp/terraform-provider-google#19261](https://github.com/hashicorp/terraform-provider-google/pull/19261) is a PR and doesn't seem related to this change.

Hello,

@melinath  I need to fix: https://github.com/hashicorp/terraform-provider-google/issues/19621, this PR is intended for this issue.
Thank you!

stefan-malin-db:
> Ah, I mistyped the issue ID when I looked it up - thanks for the link!

no problem @melinath , glad to help :) do we have an update on it or any suggestion how can we fix it?

Thank you

melinath:
I expect to have time to look into this in more detail tomorrow - sorry for the delay!

melinath (changes requested):
Thanks for your patience! This was a weird one so I wanted to be sure to give a complete answer.

The reason this isn't "doing anything" is because of https://github.com/hashicorp/terraform-provider-google/issues/15617. The Operation code is generated based on the first resource configuration it checks, and you've only modified one of the resource yaml files.

Unfortunately, even if you did that, this wouldn't work due to https://github.com/hashicorp/terraform-provider-google/issues/13337. 

And even if that bug didn't exist, there's the issue that the operations code for discoveryengine is _already_ using a locational endpoint based on the DiscoveryEngine base path: https://github.com/hashicorp/terraform-provider-google-beta/blob/7ee2cf3a123bbfedf4f69694b7be28fac0522390/google-beta/services/discoveryengine/discovery_engine_operation.go#L42

It looks like this hasn't been a problem for other DiscoveryEngine resources because they "use operations" but the operations complete immediately, so we don't end up needing to actually make any GET requests to the relevant URLs.

We haven't caught this issue in our nightly tests for TargetSite because those are currently failing due to an unrelated bug.

In conclusion, the actual fix here is to switch to handwritten operations code for this resource.

1. All the yaml files need to have autogen_async: false.
2. Copy the [current operation code](https://github.com/hashicorp/terraform-provider-google/blob/main/google/services/discoveryengine/discovery_engine_operation.go) to [mmv1/third_party/terraform/services/discoveryengine](https://github.com/GoogleCloudPlatform/magic-modules/tree/main/mmv1/third_party/terraform/services/discoveryengine)
3. Add a step _after_ https://github.com/hashicorp/terraform-provider-google/blob/773b763c13635c0514966c09b7190dc8110c748e/google/services/discoveryengine/discovery_engine_operation.go#L42 to replace `{{location}}` with the operation's location (which you will need to extract from w.CommonOperationWaiter.Op.Name)

Attn @jialei-chen FYI

melinath:
> do we have an ETA for above changes?

@stefan-malin-db since this is your PR, the ETA currently depends on you :-) Assuming you are still wanting to work on this?

> Can we whitelist also regions like eu, us, not only global for advanced chunching?

According to the location field documentation, eu and us should already work in addition to global.

melinath:
I've clarified in my comment above that only the yaml files in discoveryengine need to have autogen_async: false - not every yaml file in magic modules.

stefan-malin-db:
> > do we have an ETA for above changes?
> 
> @stefan-malin-db since this is your PR, the ETA currently depends on you :-) Assuming you are still wanting to work on this?
> 
> > Can we whitelist also regions like eu, us, not only global for advanced chunching?
> 
> According to the location field documentation, eu and us should already work in addition to global.

Thank you, @melinath , sure, when I tested terraform resource, eu and us do not reflect changes in ui, only for global, even if they are applied.

stefan-malin-db:
@melinath , new changes have been pushed, please let me know if anything needs to be adapted/changed in order to fix the issue.
Thank you!

mmv1/third_party/terraform/services/discoveryengine/discovery_engine_operation.go
```diff
@@ -0,0 +1,104 @@
...
+// ----------------------------------------------------------------------------
+//
+//     This file is automatically generated by Magic Modules and manual
+//     changes will be clobbered when the file is regenerated.
+//
+//     Please read more about how to change this file in
+//     .github/CONTRIBUTING.md.
+//
+// ----------------------------------------------------------------------------
+
```
melinath, line 17: Since this is being converted to handwritten, the headers can be taken out.

```suggestion
```

mmv1/third_party/terraform/services/discoveryengine/discovery_engine_operation.go
```diff
@@ -0,0 +1,104 @@
...
+	// Extract location from opName.  Example:
+	// "projects/<project>/locations/<location>/operations/<operation-id>"
+	// Requires error handling if not properly formatted.
+	parts := strings.Split(opName, "/")
+	if len(parts) < 5 || parts[2] != "locations" {
+		return nil, fmt.Errorf("unexpected operation name format: %s", opName) // Handle invalid format appropriately
+	}
+	location := parts[3]
+
+	url := fmt.Sprintf("%s/projects/%s/locations/%s/operations/%s", w.Config.DiscoveryEngineBasePath, w.Project, location, parts[5])
```
melinath, line 54: The DiscoveryEngineBasePath will contain the literal string `{{location}}` so that needs to be replaced. I had been thinking it would make sense to do after building the URL but maybe it would be clearer to do it first?

```suggestion
	basePath := strings.Replace(w.Config.DiscoveryEngineBasePath, "{{location}}", location, 1)
	url := fmt.Sprintf("%s%s", basePath, opName)
```

melinath (changes requested):

mmv1/third_party/terraform/services/discoveryengine/discovery_engine_operation.go
stefan-malin-db: updated, thank you for suggestion

stefan-malin-db:
@melinath changes have been pushed.

melinath:
Manually running TestAccDiscoveryEngineTargetSite_discoveryengineTargetsiteBasicExample and TestAccDiscoveryEngineTargetSite_discoveryengineTargetsiteAdvancedExample in beta since I don't trust the cassettes for this resource - the API probably changed its responses. https://hashicorp.teamcity.com/buildConfiguration/TerraformProviders_GoogleCloud_GOOGLE_BETA_MMUPSTREAMTESTS_GOOGLEBETA_PACKAGE_DISCOVERYENGINE/328493

melinath (changes requested):
okay, those both failed with:

```
Error: Error creating TargetSite: googleapi: Error 400: Do not include http or https protocol in the URI pattern.
```

The nightly tests are also failing with this error - it's not new to this PR.

You'll need to fix that so we can see if the underlying operations problem is fixed. On the plus side, that should force cassette invalidation, which means that we will be able to have more confidence in the test results!

So: the lines that need to be updated are:

- https://github.com/GoogleCloudPlatform/magic-modules/blob/6071764f00081d0bdf88801726b7fa631a4f0c91/mmv1/templates/terraform/examples/discoveryengine_targetsite_basic.tf.tmpl#L4
- https://github.com/GoogleCloudPlatform/magic-modules/blob/6071764f00081d0bdf88801726b7fa631a4f0c91/mmv1/templates/terraform/examples/discoveryengine_targetsite_advanced.tf.tmpl#L4

Based on https://github.com/hashicorp/terraform-provider-google/issues/19621 it looks like literally just removing the protocol would probably be sufficient.

stefan-malin-db:
Thank you for update, @melinath ! Updated

melinath:
New failure is:

```
Error: Provider produced inconsistent result after apply
        
        When applying changes to google_discovery_engine_target_site.advanced,
        provider "provider[\"registry.terraform.io/hashicorp/google\"]" produced an
        unexpected new value: Root object was present, but now absent.
        
        issue tracker.
```

Looking into what the fix might be.

melinath:
you're uncovering all sorts of issues with this PR! Thanks!

melinath:
okay this one is actually an issue with the PR

melinath (changes requested):
never mind, this is totally another existing problem with the resource that you've uncovered.

The problem is that the API response for the successful create operation looks like this:

```
{
  "name": "projects/1067888929963/locations/global/collections/default_collection/dataStores/tf-test-data-store-id0jx0ryyj6y/operations/create-target-site-10226669232738830812",
  "done": true,
  "response": {
    "@type": "type.googleapis.com/google.cloud.discoveryengine.v1.TargetSite",
    "name": "projects/1067888929963/locations/global/collections/default_collection/dataStores/tf-test-data-store-id0jx0ryyj6y/siteSearchEngine/targetSites/67c26d0e-0000-21f1-8918-883d24fb7644",
    "type": "INCLUDE",
    "generatedUriPattern": "cloud.google.com/docs/*",
    "updateTime": "2025-02-25T17:31:23.333181897Z",
    "siteVerificationInfo": {
      "siteVerificationState": "UNVERIFIED"
    },
    "indexingStatus": "FAILED",
    "rootDomainUri": "google.com"
  }
}
```

This resource includes [post-create code to set the name of the resource correctly in state](https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/templates/terraform/post_create/set_computed_name.tmpl), which is necessary due to the issue described in https://github.com/GoogleCloudPlatform/magic-modules/pull/12939. However, that code incorrectly sets the resource name to the name of the _operation_.

I think there are two changes needed here:

1. for TargetSite.yaml, async.result.resource_inside_response should be set to `true`.
   - This will make it so that the "response" field from the operation is automatically ["unwrapped" and returned as the result from the operation](https://github.com/modular-magician/terraform-provider-google-beta/blob/a7cafb1f2ac3894ba1bcaccbd12faed1d10d7bbc/google-beta/services/discoveryengine/discovery_engine_operation.go#L64-L77), instead of just the operation itself.
2. https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/templates/terraform/post_create/set_computed_name.tmpl needs to be updated to assume that `res` has already been correctly unwrapped - it can just use res["name"] directly.

(alternatively, just making it so that the post-create code does the unwrapping correctly would be fine too, but IMO using resource_inside_response is a little cleaner)

mmv1/templates/terraform/examples/discoveryengine_targetsite_basic.tf.tmpl
```diff
@@ -1,18 +1,19 @@
-resource "google_discovery_engine_target_site" "basic" {
-  location                    = google_discovery_engine_data_store.basic.location
-  data_store_id               = google_discovery_engine_data_store.basic.data_store_id
-  provided_uri_pattern        = "http://cloud.google.com/docs/*"
+# mmv1/templates/terraform/examples/discoveryengine_targetsite_advanced.tf.tmpl
+resource "google_discovery_engine_target_site" "advanced" {
+  location                    = google_discovery_engine_data_store.advanced.location
+  data_store_id               = google_discovery_engine_data_store.advanced.data_store_id
```
melinath, line 4: Reverting the rename that slipped in here

```suggestion
resource "google_discovery_engine_target_site" "basic" {
  location                    = google_discovery_engine_data_store.basic.location
  data_store_id               = google_discovery_engine_data_store.basic.data_store_id
```

mmv1/templates/terraform/examples/discoveryengine_targetsite_basic.tf.tmpl
```diff
@@ -1,18 +1,19 @@
...
+  provided_uri_pattern        = "cloud.google.com/docs/*"
   type                        = "INCLUDE"
   exact_match                 = false
 }
 
-resource "google_discovery_engine_data_store" "basic" {
+resource "google_discovery_engine_data_store" "advanced" {
```
melinath, line 10:
```suggestion
resource "google_discovery_engine_data_store" "basic" {
```

mmv1/templates/terraform/examples/discoveryengine_targetsite_basic.tf.tmpl
```diff
@@ -1,18 +1,19 @@
...
   location                     = "global"
   data_store_id                = "{{index $.Vars "data_store_id"}}"
-  display_name                 = "tf-test-basic-site-search-datastore"
+  display_name                 = "tf-test-advanced-site-search-datastore"
```
melinath, line 13:
```suggestion
  display_name                 = "tf-test-basic-site-search-datastore"
```

mmv1/templates/terraform/examples/discoveryengine_targetsite_basic.tf.tmpl
```diff
@@ -1,18 +1,19 @@
...
   industry_vertical            = "GENERIC"
   content_config               = "PUBLIC_WEBSITE"
   solution_types               = ["SOLUTION_TYPE_SEARCH"]
-  create_advanced_site_search  = false
+  create_advanced_site_search  = true
```
melinath, line 17:
```suggestion
  create_advanced_site_search  = false
```

melinath (changes requested):

mmv1/templates/terraform/post_create/set_computed_name.tmpl
```diff
@@ -1,17 +1,19 @@
 // `name` is autogenerated from the api so needs to be set post-create
+var name interface{}
```
melinath, line 2:
```suggestion
```

mmv1/templates/terraform/post_create/set_computed_name.tmpl
```diff
@@ -1,17 +1,19 @@
...
-		return fmt.Errorf("Create response didn't contain critical fields. Create may not have succeeded.")
-	}
+    if respBody, ok := res["response"].(map[string]interface{}); ok {
+        name, ok = respBody["name"]
+        if !ok {
+           return fmt.Errorf("Create response didn't contain critical fields. Create may not have succeeded.")
+        }
+    } else {
+       return fmt.Errorf("Create response didn't contain a valid 'response' field.")
+    }
```
melinath, line 12:
```suggestion
	respBody, ok := res["response"]
	if !ok {
		return fmt.Errorf("Create response didn't contain critical fields. Create may not have succeeded.")
	}
```

mmv1/templates/terraform/post_create/set_computed_name.tmpl
```diff
@@ -1,17 +1,19 @@
...
 
-	name, ok = respBody.(map[string]interface{})["name"]
-	if !ok {
-		return fmt.Errorf("Create response didn't contain critical fields. Create may not have succeeded.")
-	}
 }
+
```
melinath, line 15:
```suggestion
```

mmv1/templates/terraform/post_create/set_computed_name.tmpl
```diff
@@ -1,17 +1,19 @@
...
 if err := d.Set("name", name.(string)); err != nil {
-	return fmt.Errorf("Error setting name: %s", err)
+    return fmt.Errorf("Error setting name: %s", err)
```
melinath, line 17:
```suggestion
	return fmt.Errorf("Error setting name: %s", err)
```

mmv1/templates/terraform/post_create/set_computed_name.tmpl
```diff
@@ -1,17 +1,19 @@
...
 }
-d.SetId(name.(string))
+d.SetId(name.(string))
```
melinath, line 19:
```suggestion
d.SetId(name.(string))

```

mmv1/products/discoveryengine/TargetSite.yaml
```diff
@@ -42,7 +42,7 @@ async:
       update_minutes: 60
       delete_minutes: 60
   result:
-    resource_inside_response: false
+    resource_inside_response: true
 custom_code:
   post_create: 'templates/terraform/post_create/set_computed_name.tmpl'
```
melinath, line 47:
```suggestion
```

melinath (changes requested):
okay, so on the one hand we probably shouldn't actually change set_computed_name since it's used on a bunch of other resources as well (though maybe they should all also be using resource_inside_response?)

BUT on the plus side, it looks like now that resource_inside_response is set, this is actually already generating code to correctly extract the name! :-) so removing the post_create altogether should solve the problem.

mmv1/templates/terraform/post_create/set_computed_name.tmpl
```diff
@@ -6,10 +6,6 @@ if !ok {
 		return fmt.Errorf("Create response didn't contain critical fields. Create may not have succeeded.")
 	}
```
melinath, line 8:
```suggestion

	name, ok = respBody.(map[string]interface{})["name"]
	if !ok {
		return fmt.Errorf("Create response didn't contain critical fields. Create may not have succeeded.")
	}
```

melinath (approved):
LGTM, thanks for sticking with it!

stefan-malin-db:
> LGTM, thanks for sticking with it!

Thank you for support :)

PR #13026, opened 2025-02-10
Migrate google_eventarc_channel resource from DCL to MMv1

tommyreddad:
This PR is to move the Eventarc Channel (google_eventarc_channel) resource from DCL to MMv1. This PR includes:
* Creation of the resource on MMv1.
* Creation of the examples and tests.
* Addition of extra test for long-form fields.
* Remove references to DCL resource.
* Adding support for PROJECT_NUMBER env var in generated documentation.

```release-note:note
eventarc: `google_eventarc_channel` now uses MMv1 engine instead of DCL.
```


SirGitsalot (commented):
LGTM, but DCL conversions need special handling - @slevenick @trodge could you take a look?

slevenick (changes requested):
Sorry, took me a while to get to this!

Everything looks good except for one issue:

Looks like the `name` field changes format when the migration happens. It goes from whatever form was specified to the long form. I think this happens because it is being read from the API response which always returns the long form.

We don't need to read this from the API as it's never mutable, so you can either add ignore_read or *maybe* url_param_only depending on how the API works

tommyreddad:
> Sorry, took me a while to get to this!
> 
> Everything looks good except for one issue:
> 
> Looks like the `name` field changes format when the migration happens. It goes from whatever form was specified to the long form. I think this happens because it is being read from the API response which always returns the long form.
> 
> We don't need to read this from the API as it's never mutable, so you can either add ignore_read or _maybe_ url_param_only depending on how the API works

I'm not sure how this can be true. The existing tests use the short-form channel name, but I know for certain that our API does not accept short-form name as input. I'm not sure, but it seems to me like it is being expanded to long-form in DCL: https://github.com/hashicorp/terraform-provider-google/blob/ebad46489e174296791bb2e141cb30c635224e71/google/services/eventarc/resource_eventarc_channel_test.go#L97

I don't think url_param_only is appropriate, since our API does actually require the field to be set. I also tried adding ignore_read, and ran into problems since the generated provider seems to expect a function flattenEventarcChannelName, which is not generated.

slevenick:
> > Sorry, took me a while to get to this!
> > Everything looks good except for one issue:
> > Looks like the `name` field changes format when the migration happens. It goes from whatever form was specified to the long form. I think this happens because it is being read from the API response which always returns the long form.
> > We don't need to read this from the API as it's never mutable, so you can either add ignore_read or _maybe_ url_param_only depending on how the API works
> 
> I'm not sure how this can be true. The existing tests use the short-form channel name, but I know for certain that our API does not accept short-form name as input. I'm not sure, but it seems to me like it is being expanded to long-form in DCL: https://github.com/hashicorp/terraform-provider-google/blob/ebad46489e174296791bb2e141cb30c635224e71/google/services/eventarc/resource_eventarc_channel_test.go#L97
> 
> I don't think url_param_only is appropriate, since our API does actually require the field to be set. I also tried adding ignore_read, and ran into problems since the generated provider seems to expect a function flattenEventarcChannelName, which is not generated.
> 
> The current implementation in this PR makes all tests pass with RELEASE_DIFF=1, so I imagine it should be safe from that perspective.

So what I'm seeing is that the value stored for `name` in Terraform state is changing when I migrate to this version. It may not change the API requests made, but it does change the value that gets stored. It's not a disaster, but would be good to fix in case someone is using the `.name` attribute in their config to reference the value of a particular channel. We don't want to accidentally break their config because the format of this value changes.

Maybe we add a custom flatten that acts to ignore the value coming back from the API?

tommyreddad:
@slevenick I added a custom flattener which implements the logic of ignore_read. Let me know if this works.

tommyreddad:
@slevenick friendly ping

slevenick (approved):
Looks good, thanks!

slevenick (approved):

PR #13027, opened 2025-02-10
Fix instances where no items found in region by continuing to next region

ScottSuarez:
also including a rename of SweepableIdentifierField to identifier_field

```release-note:none

```


mmv1/products/apigee/Organization.yaml
```diff
@@ -43,6 +43,8 @@ async:
 custom_code:
   encoder: 'templates/terraform/encoders/apigee_organization.go.tmpl'
   custom_import: 'templates/terraform/custom_import/apigee_organization.go.tmpl'
+sweeper:
+  identifier_field: organization
```
melinath, line 47: I'm pretty sure this is incorrect since the apigee_organization resource doesn't have an "organization" field - maybe just leave this out for now?

```suggestion
```

mmv1/products/monitoring/MonitoredProject.yaml
```diff
@@ -37,10 +37,10 @@ custom_code:
   custom_import: 'templates/terraform/custom_import/monitoring_monitored_project.go.tmpl'
   test_check_destroy: 'templates/terraform/custom_check_destroy/monitoring_monitored_project.go.tmpl'
 error_retry_predicates:
-
   - 'transport_tpg.IsMonitoringPermissionError'
 schema_version: 1
 state_upgraders: true
+exclude_sweeper: true
```
melinath, line 43: why is this excluded?

melinath (approved):

mmv1/products/monitoring/MonitoredProject.yaml
ScottSuarez: It's done at the organization level and not the project level.

melinath: interesting! Another potential case that would be fixed by moving to more complex list URL handling.

PR #13030, opened 2025-02-10
update maintain a url_substitutions field in place of regions for sweepers

ScottSuarez:

```release-note:none

```


mmv1/templates/terraform/sweeper_file.go.tmpl
```diff
@@ -69,13 +98,23 @@ func testSweep{{ $.ResourceName }}(_ string) error {
...
 		billingId := envvar.GetTestBillingAccountFromEnv(t)
 
+		// Set fallback values for empty region/zone
+		if sub.region == "" {
+			log.Printf("[INFO][SWEEPER_LOG] Empty region provided, falling back to us-central1")
+			sub.region = "us-central1"
+		}
+		if sub.zone == "" {
+			log.Printf("[INFO][SWEEPER_LOG] Empty zone provided, falling back to us-central1-a")
+			sub.zone = "us-central1-a"
```
melinath, line 108: It looks like previously this defaulted to `-` - just wanted to double-check that we want to change this?

mmv1/templates/terraform/sweeper_file.go.tmpl
```diff
@@ -102,7 +141,7 @@ func testSweep{{ $.ResourceName }}(_ string) error {
 		resourceList, ok := res["{{ $.ResourceListKey }}"]
 		if !ok {
 			log.Printf("[INFO][SWEEPER_LOG] Nothing found in response.")
-			continue
+			return nil
```
melinath, line 144: It looks like this got reverted by accident?

```suggestion
			continue
```

melinath (approved):

mmv1/templates/terraform/sweeper_file.go.tmpl
ScottSuarez: ah thanks for this!

mmv1/templates/terraform/sweeper_file.go.tmpl
ScottSuarez: yeah I don't see a reason not to. I'm not sure where '-' is a valid value, but maybe we'll find out after the change gets in 'if any'.

melinath: I don't know for sure, but I'm fine with moving forward with the change & finding out that way.

melinath (approved):

PR #13032, opened 2025-02-11
Surface errors when updating node_config for cluster and node pool.

VeraQin:
Fix issues in https://github.com/hashicorp/terraform-provider-google/issues/19935.

Change-Id: I8ba5d472516f5c358407a9e28476f2ad22ff5134

```release-note:bug
container: fixed node_config update error not surfaced to the operation for cluster and node pool update.
```


roaks3 (commented):
I think this makes sense, but 2 clarifying questions:
* Did you test this locally to confirm the broken scenario is fixed?
* It seems like this method of error handling allows the cluster/node pool to be partially updated. Am I understanding correctly that this is already how the resources operate for other types of updates, and that it is a safe way to change the behavior?

PR #13033, opened 2025-02-11
Encode b64_mac_key in base64url, not in base64

makotonakai:
Fixes https://github.com/hashicorp/terraform-provider-google/issues/20801

```release-note:bug
publicca: encode b64_mac_key in base64url, not in base64
```


hao-nan-li (commented):
Thanks for fixing the customer issue. I'm not sure if I'm following correctly, but why does this validation function help the customer to return the desired attribute `b64url_mac_key`?

makotonakai:
@hao-nan-li 

I added the validation function in order to make sure that the returned string was encoded in base64url.

The public ca is supposed to return a key encoded in base64url, but the key in the API response is actually a base64 string, according to the API documentation. That's why I decided to add a method to re-encode it in base64url.
https://cloud.google.com/certificate-manager/docs/reference/public-ca/rest/v1/projects.locations.externalAccountKeys#ExternalAccountKey

By the way, do you want me to update the field name to `b64url_mac_key`?
I decided to keep as it is for now, because it corresponds to the one in the API response and want to avoid confusion.

brokenjacobs:
I'd recommend adding a new property, otherwise this will break existing terraform state/code

makotonakai:
@brokenjacobs 

Thank you for your suggestion, but I don't fully understand it.
Do you think it's okay to rename the existing field name, or you think it's better to keep the new one and existing one for safety? (Probably the former)

makotonakai:
@hao-nan-li 

I have a question about the field deprecation. The following link only talks about how to include breaking changes before v6.0.0, not after v6.0.0 (or before v7.0.0)
https://googlecloudplatform.github.io/magic-modules/breaking-changes/make-a-breaking-change/

Is it okay to add a deprecation message in the field and add a new one (b64url_mac_key) and push them to the main branch?

hao-nan-li:
> @hao-nan-li
> 
> I have a question about the field deprecation. The following link only talks about how to include breaking changes before v6.0.0, not after v6.0.0 (or before v7.0.0) https://googlecloudplatform.github.io/magic-modules/breaking-changes/make-a-breaking-change/
> 
> Is it okay to add a deprecation message in the field and add a new one (b64url_mac_key) and push them to the main branch?

7.0.0 will occur in August-September 2025. If the change is to add a new field, it's totally fine.

hao-nan-li (approved):

PR #13040, opened 2025-02-11
Terraform config for Firebase App Hosting Backend resource

rainshen49:

```release-note:new-resource
`google_firebase_app_hosting_backend`
```

### DIFF
--- a/mmv1/products/firebaseapphosting/Backend.yaml
+++ b/mmv1/products/firebaseapphosting/Backend.yaml
@@ -1,0 +1,210 @@
+# Copyright 2025 Google Inc.
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+---
+name: Backend
+description: A Backend is the primary resource of App Hosting.
+base_url: projects/{{project}}/locations/{{location}}/backends
+update_mask: true
+self_link: projects/{{project}}/locations/{{location}}/backends/{{backend_id}}
+create_url: projects/{{project}}/locations/{{location}}/backends?backendId={{backend_id}}
+# Always forces deletion because the child resource default `Domain` is always present.
+delete_url: projects/{{project}}/locations/{{location}}/backends/{{backend_id}}?force=true
+update_verb: PATCH
+id_format: projects/{{project}}/locations/{{location}}/backends/{{backend_id}}
+import_format:
+  - "projects/{{project}}/locations/{{location}}/backends/{{backend_id}}"
+  - "{{project}}/{{location}}/{{backend_id}}"
+  - "{{location}}/{{backend_id}}"
+examples:
+  - name: firebase_app_hosting_backend_basic
+    primary_resource_id: example
+    vars:
+      location: 'us-central1'
+      backend_id: 'my-backend'
+      app_id: '1:0000000000:web:674cde32020e16fbce9dbd'
+      display_name: 'My Backend'
+      serving_locality: 'GLOBAL_ACCESS'
+    test_env_vars:
+      project_id: 'PROJECT_NAME'
+    test_vars_overrides:
+      'location': '"us-central1"'
+      'serving_locality': '"GLOBAL_ACCESS"'
+      'app_id': '"1:0000000000:web:674cde32020e16fbce9dbd"'
+  - name: firebase_app_hosting_backend_github
+    primary_resource_id: example
+    vars:
+      location: 'us-central1'
+      backend_id: 'my-backend-gh'
+      app_id: 'firebase:app:id'
+      display_name: 'My Backend'
+      serving_locality: 'GLOBAL_ACCESS'
+    test_env_vars:
+      project_id: 'PROJECT_NAME'
+    test_vars_overrides:
+      'location': '"us-central1"'
+      'serving_locality': '"GLOBAL_ACCESS"'
+    skip_test: true # Can't establish a Github connection in automated tests.
+autogen_async: true
+async:
+  operation:
+    timeouts:
+      insert_minutes: 20
+      update_minutes: 20
+      delete_minutes: 20
+    base_url: "{{op_id}}"
+  actions:
+    - create
+    - delete
+    - update
+  type: OpAsync
+  result:
+    resource_inside_response: true
+  include_project: false
+autogen_status: QmFja2VuZA==
+parameters:
+  - name: location
+    type: String
+    description: The canonical IDs of a Google Cloud location such as "us-east1".
+    immutable: true
+    url_param_only: true
+    required: true
+  - name: backendId
+    type: String
+    description: |-
+      Id of the backend. Also used as the service ID for Cloud Run, and as part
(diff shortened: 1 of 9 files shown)


mmv1/products/firebaseapphosting/Backend.yaml
```diff
@@ -0,0 +1,215 @@
...
+  - "{{project}}/{{location}}/{{backend_id}}"
+  - "{{location}}/{{backend_id}}"
+min_version: beta
+examples:
+  - name: firebase_app_hosting_backend_basic
+    primary_resource_id: example
+    min_version: beta
+    vars:
+      location: "us-central1"
+      backend_id: "my-backend"
```
roaks3, line 36: Besides `backend_id`, these should all be specified directly in the config. The vars here are specifically used for allowing values to differ between docs and tests. Same for the other example.

mmv1/products/firebaseapphosting/Backend.yaml
```diff
@@ -0,0 +1,215 @@
...
+      backend_id: "my-backend-gh"
+      app_id: "firebase:app:id"
+      display_name: "My Backend"
+      serving_locality: "GLOBAL_ACCESS"
+    test_env_vars:
+      project_id: "PROJECT_NAME"
+    test_vars_overrides:
+      "location": '"us-central1"'
+      "serving_locality": '"GLOBAL_ACCESS"'
+    skip_test: true  # Can't establish a Github connection in automated tests.
```
roaks3, line 60: This should be `exclude_test: true` unless you expect to enable this in the near future

mmv1/templates/terraform/examples/firebase_app_hosting_backend_basic.tf.tmpl
```diff
@@ -0,0 +1,76 @@
+# Service account setup only required once per project.
```
roaks3, line 1: For this example, are you intending for users to use this as a reference and potentially copy-and-paste this config as a canonical example to work from? I just couldn't tell, but if it is not suitable for documentation, you may want to rework things so that it does serve that audience (and you may want test-only or docs-only examples).

mmv1/templates/terraform/examples/firebase_app_hosting_backend_basic.tf.tmpl
```diff
@@ -0,0 +1,76 @@
...
+resource "google_service_account" "service_account" {
+  provider = google-beta
+
+  project = "{{index $.TestEnvVars "project_id"}}"
+
+  # Must be firebase-app-hosting-compute
+  account_id                   = "firebase-app-hosting-compute"
```
roaks3, line 8: This should use `vars` so that it gets sweepable prefix/suffix added

mmv1/templates/terraform/examples/firebase_app_hosting_backend_basic.tf.tmpl
```diff
@@ -0,0 +1,76 @@
...
+  display_name                 = "Firebase App Hosting compute service account"
+
+  # Do not throw if already exists
+  create_ignore_already_exists = true
```
roaks3, line 12: What is the purpose of including this? Tests should be deleting this account when they complete

mmv1/templates/terraform/examples/firebase_app_hosting_backend_basic.tf.tmpl
```diff
@@ -0,0 +1,76 @@
...
+resource "google_project_service" "fah" {
+  provider = google-beta
+
+  project = "{{index $.TestEnvVars "project_id"}}"
+  service = "firebaseapphosting.googleapis.com"
+
+  disable_on_destroy = false
+}
+
+resource "google_firebase_app_hosting_backend" "example" {
```
roaks3, line 55: This being the only test I see 2 main issues:
* We never check if optional fields can be omitted. This would normally be done by having one of the test configs be minimal (only required fields), and another with all possible fields specified.
* We never check if mutable fields can be changed (like app_id, service_account, etc.). Unless there is some limitation, we should use an [update test](https://googlecloudplatform.github.io/magic-modules/test/test/#add-an-update-test) for them.

mmv1/templates/terraform/examples/firebase_app_hosting_backend_github.tf.tmpl
```diff
@@ -0,0 +1,70 @@
+resource "google_firebase_app_hosting_backend" "example" {
```
roaks3, line 1: Since these are both going to be shown in the docs as examples, we should be consistent with the ordering ie. dependencies-first or dependencies-last. I think dependencies-last is more common, but either should be fine.

roaks3 (changes requested):
Code LGTM, but had a few test-related comments

mmv1/templates/terraform/examples/firebase_app_hosting_backend_basic.tf.tmpl
rainshen49: Yes, there are some setup that's required only once per project, as opposed to once per backend. It's common for users to just copy-pasta code, so I want them to at least copy everything they need.

mmv1/templates/terraform/examples/firebase_app_hosting_backend_basic.tf.tmpl
rainshen49: This actually must be `firebase-app-hosting-compute` (see comment in the previous line). Otherwise, the experience in other surfaces (such as the Firebase Console UI) will be broken.  This is publicly documented at https://firebase.google.com/support/guides/service-accounts

mmv1/templates/terraform/examples/firebase_app_hosting_backend_basic.tf.tmpl
rainshen49: See comment above (we have to use this name so sweepers can't reach it).

mmv1/templates/terraform/examples/firebase_app_hosting_backend_basic.tf.tmpl
rainshen49: Right, added.

mmv1/templates/terraform/examples/firebase_app_hosting_backend_basic.tf.tmpl
roaks3: Ah sorry, thanks for clarifying!

mmv1/templates/terraform/examples/firebase_app_hosting_backend_full.tf.tmpl
```diff
@@ -0,0 +1,66 @@
+resource "google_firebase_app_hosting_backend" "example" {
+  project          = "{{index $.TestEnvVars "project_id"}}"
+
+  # Choose the region closest to your users
+  location         = "us-central1"
+  backend_id       = "{{index $.Vars "backend_id"}}"
+  app_id           = "1:0000000000:web:674cde32020e16fbce9dbd"
```
roaks3, line 7: You did confirm on last review that these examples looked good for user consumption, but just to double check, this `app_id` is expected to be used by regular users? Sorry to ask again, it just looks very specific to an environment.

roaks3 (approved):
LGTM, just one last nit.

I think the test/check failures are not related, although it might be worth a rebase to see if they resolve. The `terraform-provider-missing-service-labels` check is related to resource ownership for your team, so you'll need to use go/terraform-resource-owners-enrollment to get that set up if you haven't already.

rainshen49:
cl/734321905

roaks3 (commented):
This is just blocked on a rebase of the branch to bring in the latest changes from main, which should at a minimum resolve the failed TeamCity check

roaks3 (approved):
Checked the VCR failures and they look safe to ignore

PR #13042, opened 2025-02-11
Add google_eventarc_message_bus resource

tommyreddad:
Add MessageBus resource for Eventarc.

```release-note:new-resource
`google_eventarc_message_bus`
```


tommyreddad:
@shuyama1 friendly ping

mmv1/products/eventarc/MessageBus.yaml
```diff
@@ -0,0 +1,130 @@
...
+      projects/{project}/locations/{location}/messageBuses/{message_bus}
+    output: true
+  - name: loggingConfig
+    type: NestedObject
+    description: |-
+      The configuration for Platform Telemetry logging for Eventarc Advanced
+      resources.
+    default_from_api: true
+    properties:
+      - name: logSeverity
```
shuyama1, line 75: Can we make this field required given this is the only subfield within `logging_config`. We'd want to prevent user from sending empty blocks, as it may cause issues in some cases, unless sending `logging_config{}` is treated differently than an unset `logging_config`.

mmv1/products/eventarc/MessageBus.yaml
```diff
@@ -0,0 +1,130 @@
...
+        type: Enum
+        description: |-
+          Optional. The minimum severity of logs that will be sent to Stackdriver/Platform
+          Telemetry. Logs at severitiy ≥ this value will be sent, unless it is NONE.
+        default_from_api: true
+        enum_values:
+          - LOG_SEVERITY_UNSPECIFIED
```
shuyama1, line 82: Do we expect user to specify value `LOG_SEVERITY_UNSPECIFIED`? If not, let's remove it from possible value lists

mmv1/products/eventarc/MessageBus.yaml
```diff
@@ -0,0 +1,130 @@
...
+    description: |-
+      Output only. This checksum is computed by the server based on the value of other
+      fields, and might be sent only on update and delete requests to ensure that
+      the client has an up-to-date value before proceeding.
+    output: true
+  - name: createTime
+    type: String
+    description: Output only. The creation time.
+    output: true
+  - name: displayName
```
shuyama1, line 119: Could you include this in the update test to verify its update functionality? Thanks!

mmv1/products/eventarc/MessageBus.yaml
shuyama1: We typically avoid including `_UNSPECIFIED` in `enum_values` list, see https://googlecloudplatform.github.io/magic-modules/reference/field/#enum_values. There're some exceptions in existing resources, but we aim to avoid them in new implementations. Thanks for updating it.

shuyama1 (approved):
Thank you!

PR #13044, opened 2025-02-11
Add location field to google_network_services_grpc_route

AshZhang:
Adds a new `location` field to `google_network_services_grpc_route`.

In the future, this will allow creating a regional GRPCRoute resource, but currently regional GRPCRoutes are internal-only and not yet publicly available, so only the value `'global'` may be set. If omitted, the location defaults to `'global'`.

This PR is essentially the same as https://github.com/GoogleCloudPlatform/magic-modules/pull/12991, but for GRPCRoutes.

```release-note:enhancement
networkservices: added `location` field to `google_network_services_grpc_route` resource
```


AshZhang:
Hi @shuyama1, can you please take a look and review this PR? Please let me know if you have any questions. Thanks!

AshZhang:
I have also created https://cloud-internal-review.git.corp.google.com/c/cloud-graphite-eng/magic-modules-private-overrides/+/50570 to prevent any conflicts in private overrides.

melinath:

mmv1/products/networkservices/GrpcRoute.yaml
```diff
@@ -283,3 +290,13 @@ properties:
                   description: |
                     Specifies the allowed number of retries.
                   min_version: 'beta'
+  - name: 'location'
+    type: String
+    description: |
+      Location (region) of the GRPCRoute resource to be created. Only the value 'global' is currently allowed; defaults to 'global' if omitted.
+    min_version: 'beta'
+    url_param_only: true
+    immutable: true
```
hao-nan-li, line 299: Setting `immutable:true` will make this resource recreated if any change happens to this field. 

Since this is a url param, perhaps this is not necessary here

mmv1/templates/terraform/state_migrations/network_services_grpc_route.go.tmpl
```diff
@@ -0,0 +1,293 @@
+func resourceNetworkServicesGrpcRouteResourceV0() *schema.Resource {
```
hao-nan-li, line 1: Not sure if this addition is relevant to the added field, what is the rational to add this file?

mmv1/products/networkservices/GrpcRoute.yaml
melinath: drive-by comment: URL parameters are almost always immutable.

hao-nan-li: Thanks for the comment, I'm just wondering, if the location change, I'd suspect it points to another resource at another location. 

I think it's also a good option to keep this field immutable.

AshZhang: On the side I did some manual testing for a similar resource (`google_network_services_mesh`), which has a [parameter `name`](https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/products/networkservices/Mesh.yaml#L73) with `url_param_only: true` but no `immutable`. If I create a `google_network_services_mesh` resource, run `terraform apply`, and then change the name and run `terraform apply` again, it will say that the resource will be "updated". The second `terraform apply` will then fail, the error is something like
```
Error updating Mesh (old name): googleapi: Error 404: Resource (the new name) doesn't exist
```
The state file then gets into a weird state where the `google_network_services_mesh` is missing, but the actual underlying Mesh resource is still there, so subsequent `terraform apply` will fail because it will try to create the resource which already exists.

mmv1/templates/terraform/state_migrations/network_services_grpc_route.go.tmpl
AshZhang: This is similar to [the PR adding `location` to `google_network_services_mesh`](https://github.com/GoogleCloudPlatform/magic-modules/pull/12991/files#r1947000446). This is a StateUpgrader for the `location` field for backward compatibility with existing resources / state files that don't have the field. Without it, the provider will substitute an empty value for `{{location}}` in the URL during RUD operations for existing resources since there is no `location` attribute in the state file, causing an invalid resource URL and subsequent errors.

mmv1/products/networkservices/GrpcRoute.yaml
hao-nan-li: Thanks for checking it out

hao-nan-li (approved):
LGTM. @melinath would you like to make another pass on this PR?

melinath:
@hao-nan-li I don't feel the need to unless you want me to, in which case I'd be happy to.

hao-nan-li:
> @hao-nan-li I don't feel the need to unless you want me to, in which case I'd be happy to.

Alright thanks, I think this is good to merge.

PR #13048, opened 2025-02-12
add `raw_resource_config_validation` for `validation.PreferWriteOnlyAttribute` support

BBBmau:
In order include validation for WriteOnly attributes it's necessary to include the following `raw_resource_config_validation` as part of the generation. 

This allows us to include the validation method for `PreferWriteOnlyAttribute` which lets the user know about the new write-only attribute and to transition to the new field instead of using sensitive.

image

it's treated like any other custom_code where we provided the file path which would include the following:
image

This is the yaml config:
image

```release-note:none

```

### DIFF
--- a/mmv1/api/resource/custom_code.go
+++ b/mmv1/api/resource/custom_code.go
@@ -133,4 +133,6 @@ type CustomCode struct {
 	// resource was successfully deleted. Use this if the API responds
 	// with a success HTTP code for deleted resources
 	TestCheckDestroy string `yaml:"test_check_destroy"`
+
+	ValidateRawResourceConfigFuncs string `yaml:"raw_resource_config_validation"`
 }
--- a/mmv1/templates/terraform/resource.go.tmpl
+++ b/mmv1/templates/terraform/resource.go.tmpl
@@ -134,7 +134,11 @@ func Resource{{ $.ResourceName -}}() *schema.Resource {
 {{- end }}
         ),
 {{- end}}
-
+{{- if $.CustomCode.ValidateRawResourceConfigFuncs }}
+		ValidateRawResourceConfigFuncs: []schema.ValidateRawResourceConfigFunc{
+			{{$.CustomTemplate $.CustomCode.ValidateRawResourceConfigFuncs false}},
+		},
+{{- end }}
 {{ if $.DeprecationMessage }}
         DeprecationMessage: "{{ $.DeprecationMessage -}}",
 {{- end}}


tpgtools/resource.go
```diff
@@ -107,6 +107,10 @@ type Resource struct {
 	// after the Creation call for a resource.
 	PostCreateFunction *string
 
+	// ValidationFunction is the name of a function that's called immediately
+	// after the Creation call for a resource.
+	ValidationFunction *string
```
c2thorn, line 112: this would also need to be accessed via template code in https://github.com/GoogleCloudPlatform/magic-modules/blob/main/tpgtools/templates/resource.go.tmpl to affect DCL resource generation

BBBmau: added

tpgtools/templates/resource.go.tmpl
```diff
@@ -76,7 +76,11 @@ func Resource{{$.PathType}}() *schema.Resource {
 {{- end }}
 			Delete: schema.DefaultTimeout({{$.DeleteTimeoutMinutes}} * time.Minute),
 		},
-
+{{- if $.CustomCode.ValidateRawResourceConfigFuncs }}
```
c2thorn, line 79: tpgtools uses a separate resource object entirely than MMv1, so this should mirror the tpgtools/resource.go code you've added like
```suggestion
{{- if $.ValidationFunction }}
```

Then the string should just be output directly, there is no `CustomTemplate` function.

Do you know of a DCL resource needing this functionality already? DCL is on "keeping the lights on" mode and we generally don't need to add new features to tpgtools resources unless it will be immediately important. We have long term goals to migrate off of DCL/tpgtools entirely and are slowly doing so

c2thorn: So it may be easier to just remove the tpgtools changes unless you think it will be necessary for a known resource.

c2thorn: Compute disk is not DCL for what it's worth, here are the list of resources generated by DCL:
https://github.com/hashicorp/terraform-provider-google-beta/blob/main/google-beta/provider/provider_dcl_resources.go#L37 @BBBmau

c2thorn (commented):
Still doubtful the DCL code would be used, and am not confident in the DCL changes would actually work since there are no examples in this PR

See https://github.com/GoogleCloudPlatform/magic-modules/pull/13048#discussion_r1954870976

tpgtools/templates/resource.go.tmpl
```diff
@@ -76,7 +76,11 @@ func Resource{{$.PathType}}() *schema.Resource {
 {{- end }}
 			Delete: schema.DefaultTimeout({{$.DeleteTimeoutMinutes}} * time.Minute),
 		},
-
+{{- if $.ValidationFunction }}
+		ValidateRawResourceConfigFuncs: []schema.ValidateRawResourceConfigFunc{
+			{{$.CustomTemplate $.CustomCode.ValidateRawResourceConfigFuncs false}},
```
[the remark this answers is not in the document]
BBBmau: looked through the list and you're right. No DCL resources require this functionality.

tpgtools/templates/resource.go.tmpl
BBBmau: Thanks I was looking in the wrong place for the DCL resources. Removing it from tpgtools then since no DCL resources require this. https://github.com/GoogleCloudPlatform/magic-modules/pull/13048#discussion_r1954937728

c2thorn (changes requested):
Since we are not using DCL at all for this, can we remove all `tpgtools/*` changes?

c2thorn (approved):
LGTM pending above changes

c2thorn (approved):
LGTM

PR #13053, opened 2025-02-12
Update create_url updateMask for google_chronicle_rule_deployment res…

ankitgoyal0301:
…ource

```release-note:bug
chronicle: fixed creation issues when optional fields were missing for `google_chronicle_rule_deployment` resource
```


ankitgoyal0301:
> Is there more context on this change? What issue might it be fixing and would that be testable?
> 
> It's possible this is just a difference of implementation?

@ScottSuarez So, run_frequency is an optional field for us. Now, during the create call, just assume a case that you don't specify it in your terraform code. Now the updateMask in create_url has run_frequency param (and infact all 4 params, which is the case before this PR because create_url updateMask is hardcoded), the API complains, because it expects a value of the params if they are in updateMask. 

This PR is to fix the issue that if you don't pass "run_frequency" during your terraform create call for a rule_deployment resource, it won't complain and work as expected.

ScottSuarez (approved):

PR #13060, opened 2025-02-12
Adding Apigee Dns Zone resource support

sauragrawal:
Adding support for google_apigee_dns_zone resource

Fixes: https://github.com/hashicorp/terraform-provider-google/issues/21391

```release-note:new-resource
`google_apigee_dns_zone`
```


mmv1/products/apigee/DnsZone.yaml
```diff
@@ -0,0 +1,106 @@
...
+autogen_async: true
+async:
+  actions: ['create', 'delete', 'update']
+  type: 'OpAsync'
+  operation:
+    base_url: '{{op_id}}'
+  result:
+    resource_inside_response: true
+custom_code:
+  custom_import: 'templates/terraform/custom_import/apigee_dns_zone.go.tmpl'
```
melinath, line 44: Since you're defining the import formats the way you are, I don't think you should need a custom import at all (and if you do, it's not written correctly at the moment given your import ids.)

```suggestion
```

mmv1/templates/terraform/examples/apigee_dns_zone_basic_test.tf.tmpl
```diff
@@ -0,0 +1,102 @@
+ resource "google_project" "project" {
```
melinath, line 1: Why does this test need a new project? (Not saying it doesn't, just want to confirm why.)

melinath (changes requested):

mmv1/templates/terraform/examples/apigee_dns_zone_basic_test.tf.tmpl
sauragrawal: My understanding is that every test needs a new project to run.

Also as part of test
want to create Apigee Organization and then create the DNS Zone.

melinath: that's definitely not a restriction for all tests, but it might be true for Apigee organization tests. For example, if it's only possible to have one apigee organization per project?

sauragrawal: Yes, one Apigee Organization per project

mmv1/products/apigee/DnsZone.yaml
sauragrawal: I see other Apigee resources also having similar custom import code

Apigee resource are of format
organizations/<org id>/dnsZones/<dns zone id>

which is different than other normal GCP resources i think, so thats why a custom code might be required

melinath: I'd recommend trying to remove the custom code so we can verify whether it's necessary. Less custom code = easier maintenance.

melinath:
manual test run: https://hashicorp.teamcity.com/buildConfiguration/TerraformProviders_GoogleCloud_GOOGLE_BETA_MMUPSTREAMTESTS_GOOGLEBETA_PACKAGE_APIGEE/332876

mmv1/products/apigee/DnsZone.yaml
sauragrawal: Removed the custom code.

mmv1/templates/terraform/examples/apigee_dns_zone_basic_test.tf.tmpl
```diff
@@ -0,0 +1,102 @@
...
+
+resource "google_project_service_identity" "apigee_sa" {
+  provider = google-beta
+
+  project = google_project.project.project_id
+  service = google_project_service.apigee.service
+  depends_on = [google_project_service.apigee]
+}
+
+resource "google_project_iam_member" "project" {
```
melinath, line 64: This test is currently failing with:

```
Error: Request `Create IAM Members roles/dns.peer serviceAccount:service-908536617266@gcp-sa-apigee.iam.gserviceaccount.com for project "tf-testl8mm620ph2"` returned error: Error applying IAM policy for project "tf-testl8mm620ph2": Error setting IAM policy for project "tf-testl8mm620ph2": googleapi: Error 400: Service account service-908536617266@gcp-sa-apigee.iam.gserviceaccount.com does not exist., badRequest
```

Possibly there's a propagation delay after the `google_project_service_identity` is created? If so, you should add a [time_sleep](https://registry.terraform.io/providers/hashicorp/time/latest/docs/resources/sleep) resource to wait for propagation.

melinath (changes requested):

mmv1/products/apigee/DnsZone.yaml
melinath: looks like import works fine without the custom code!

mmv1/templates/terraform/examples/apigee_dns_zone_basic_test.tf.tmpl
sauragrawal: done

melinath:
manual test run: https://hashicorp.teamcity.com/buildConfiguration/TerraformProviders_GoogleCloud_GOOGLE_BETA_MMUPSTREAMTESTS_GOOGLEBETA_PACKAGE_APIGEE/342006

mmv1/products/apigee/DnsZone.yaml
```diff
@@ -0,0 +1,104 @@
...
+  api: 'https://cloud.google.com/apigee/docs/reference/apis/apigee/rest/v1/organizations.dnsZones/create'
+docs:
+base_url: 'dnsZones'
+self_link: '{{org_id}}/dnsZones/{{dns_zone_id}}'
+create_url: '{{org_id}}/dnsZones?dnsZoneId={{dns_zone_id}}'
+delete_url: '{{org_id}}/dnsZones/{{dns_zone_id}}'
+immutable: true
+import_format:
+  - '{{org_id}}/dnsZones/{{dns_zone_id}}'
+  - '{{org_id}}/{{dns_zone_id}}'
```
melinath, line 30: Adding % here would let org_id contain `/`, and would probably work. But you wouldn't get the same level of information in error messages if users specify it wrong.

```suggestion
  - '{{%org_id}}/dnsZones/{{dns_zone_id}}'
  - '{{%org_id}}/{{dns_zone_id}}'
```

melinath (changes requested):
Sorry for the rework, but it looks like I was wrong (misinterpreted the test failure last time around) and the custom_import _is_ necessary. Specifically, the problem is that orgId contains the `organizations/` part as well. It could be possible to make this behave in a more normal way but then it would potentially diverge from other apigee resources.

The test failure is now:

```
        Error: Import id "organizations/tf-testqi46wps6k0/dnsZones/tf-testqi46wps6k0" doesn't match any of the accepted formats: [^(?P<org_id>[^/]+)/dnsZones/(?P<dns_zone_id>[^/]+)$ ^(?P<org_id>[^/]+)/(?P<dns_zone_id>[^/]+)$]
```

mmv1/products/apigee/DnsZone.yaml
sauragrawal: Added the custom import code

melinath:
manual test run: https://hashicorp.teamcity.com/buildConfiguration/TerraformProviders_GoogleCloud_GOOGLE_BETA_MMUPSTREAMTESTS_GOOGLEBETA_PACKAGE_APIGEE/342019

PR #13063, opened 2025-02-12
Exactly one of breaking rule

trodge:
Adds a breaking change rule for adding exactly one of to an existing field.

```release-note:none

```

### DIFF
--- a/docs/content/breaking-changes/breaking-changes.md
+++ b/docs/content/breaking-changes/breaking-changes.md
@@ -55,7 +55,7 @@ For more information, see
 ## Field-level breaking changes
 
 * <a name="resource-schema-field-removal-or-rename"></a>Removing or renaming a field 
-* <a name="resource-schema-field-addition-of-exactly-one-of"></a>Adding "ExactlyOneOf" to an existing field
+* <a name="resource-schema-field-addition-of-exactly-one-of"></a>Adding an "ExactlyOneOf" constraint that causes one or more previously-optional fields to be required or conflict with each other
 * <a name="field-changing-type"></a> Changing field output type
   * Between primitive types, like changing a String to an Integer
   * Between complex types like changing a List to a Set.
--- a/docs/content/breaking-changes/breaking-changes.md
+++ b/docs/content/breaking-changes/breaking-changes.md
@@ -55,6 +55,7 @@ For more information, see
 ## Field-level breaking changes
 
 * <a name="resource-schema-field-removal-or-rename"></a>Removing or renaming a field 
+* <a name="resource-schema-field-addition-of-exactly-one-of"></a>Adding "ExactlyOneOf" to an existing field
 * <a name="field-changing-type"></a> Changing field output type
   * Between primitive types, like changing a String to an Integer
   * Between complex types like changing a List to a Set.
--- a/docs/content/breaking-changes/breaking-changes.md
+++ b/docs/content/breaking-changes/breaking-changes.md
@@ -54,14 +54,14 @@ For more information, see
 
 ## Field-level breaking changes
 
-* <a name="resource-schema-field-removal-or-rename"></a>Removing or renaming a field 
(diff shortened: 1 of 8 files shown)


docs/content/breaking-changes/breaking-changes.md
```diff
@@ -55,6 +55,7 @@ For more information, see
 ## Field-level breaking changes
 
 * <a name="resource-schema-field-removal-or-rename"></a>Removing or renaming a field 
+* <a name="resource-schema-field-addition-of-exactly-one-of"></a>Adding "ExactlyOneOf" to an existing field
```
melinath, line 58: I'd suggest adding this right after `Making an optional field required or adding a new required field` since it's similarly making something newly required. Also maybe being a little more explicit - something like this?

```suggestion
* <a name="resource-schema-field-addition-of-exactly-one-of"></a>Adding an "ExactlyOneOf" constraint that causes one or more previously-optional fields to be required or conflict with each other
```

tools/diff-processor/breaking_changes/resource_diff_test.go
```diff
@@ -88,3 +107,112 @@ var resourceSchemaRule_RemovingAField_TestCases = []resourceSchemaTestCase{
 		expectedFields: []string{"field-a", "field-b"},
 	},
 }
+
+var resourceSchemaRule_AddingExactlyOneOf_TestCases = []resourceSchemaTestCase{
+	{
+		name: "control",
```
melinath, line 113: nit: control is a slightly uncommon term (and we might want to add more base cases later)

```suggestion
		name: "no changes",
```

tools/diff-processor/breaking_changes/resource_diff_test.go
```diff
@@ -88,3 +107,112 @@ var resourceSchemaRule_RemovingAField_TestCases = []resourceSchemaTestCase{
 		expectedFields: []string{"field-a", "field-b"},
 	},
 }
+
+var resourceSchemaRule_AddingExactlyOneOf_TestCases = []resourceSchemaTestCase{
```
melinath, line 111: We need a few more test cases:

1. [ ] Switching a field from required to ExactlyOneOf with just that field: not breaking
2. [ ] Switching a field from ExactlyOneOf with just that field to required: not breaking
3. [ ] Adding ExactlyOneOf to an existing required field (removing required) and also one or more newly-introduced fields: not breaking
4. [ ] Adding ExactlyOneOf to new fields that are all within a newly-added, optional ancestor: not breaking
5. [ ] Removing some fields from an ExactlyOneOf: not breaking
6. [ ] Removing an entire ExactlyOneOf: not breaking

tools/diff-processor/diff/sets_test.go
```diff
@@ -31,3 +33,83 @@ func TestRemoveZeroPadding(t *testing.T) {
...
+			expected: FieldSet{},
+		},
+		{
+			name:     "set difference with itself is empty set",
+			set1:     FieldSet{"a": {}},
+			set2:     FieldSet{"a": {}},
+			expected: FieldSet{},
+		},
+		{
+			name:     "set difference with subset is subset",
```
melinath, line 95: nit: name is inaccurate

```suggestion
			name:     "set difference with subset is diff",
```

melinath (changes requested):

tools/diff-processor/breaking_changes/resource_diff_test.go
trodge: In the case of removing some fields from an ExactlyOneOf: wouldn't this be breaking?

For example, if we have exactly one of fields a b and c, and a user's config specifies field b, changing the exactly one of to a and c would break the user's config.

melinath: you're correct, good catch! I've updated the list above.

melinath (approved):
LGTM

PR #12190, opened 2024-10-29
`google_artifact_registry_repository`: Fix perma-diff in maven_config

Subserial:
The maven_config field is immutable, but also optional. This PR adds a diff suppress to equate missing, empty and default configs.

Example:
```
resource "google_artifact_registry_repository" "repo" {
  project       = local.project_id
  repository_id = "test-maven-repo"
  format        = "MAVEN"
  location      = "us"
  description   = "Test Maven Repository"
}
```
```
resource "google_artifact_registry_repository" "repo" {
  project       = local.project_id
  repository_id = "test-maven-repo"
  format        = "MAVEN"
  location      = "us"
  description   = "Test Maven Repository"
  maven_config { }
}
```
```
resource "google_artifact_registry_repository" "repo" {
  project       = local.project_id
  repository_id = "test-maven-repo"
  format        = "MAVEN"
  location      = "us"
  description   = "Test Maven Repository"
  maven_config {
    allow_snapshot_overwrites = false
  }
}
```
These should all represent the same configuration, but instead cause perma-diffs when moving from one to any other.

**Release Note Template for Downstream PRs (will be copied)**

```release-note:bug
artifactregistry: fixed perma-diff in `google_artifact_registry_repository`
```
b/376134092

PR #13068, opened 2025-02-13
Add a custom diff function for org_policy rules to fix permadiff

FalconerTC:
This is an attempt to fix the permadiff that can happen when a `google_org_policy_policy` contains multiple rules. Inspired by https://github.com/GoogleCloudPlatform/magic-modules/pull/12190. 

I initially tried to change `Type.UnorderedList` to support nested properties as a solution to this, but was unable to get that approach to work

Fixes: https://github.com/hashicorp/terraform-provider-google/issues/12363

```release-note:bug
orgpolicy: fix permadiff in `google_org_policy_policy`  when multiple rules are present
```


FalconerTC:

rileykarson:
Apologies, have been low on cycles and taking too long to get back to reviews. Picking an alternate reviewer.

FalconerTC:

FalconerTC:
@rileykarson Can you assign another reviewer here?

rileykarson:
Oh, yep, the command needed to be exclusive in the comment at the time 🤦 

melinath (approved):
this seems reasonable. The one issue we run into sometimes with Diff Suppress on complex objects is that when they _don't_ suppress, the diff shown is confusing (which is visible on the linked ticket - it looks like a complex change but actually it's just an ordering problem). The solution is to change the order returned by the API to match what's in the config during "flattening" - more detail about this at https://googlecloudplatform.github.io/magic-modules/develop/diffs/#list-order. However, our existing premade helpers for this require a unique string identifier.

I think it would be possible to do something similar here, though, since you're using [schema.HashResource](https://github.com/hashicorp/terraform-plugin-sdk/blob/b8445ee6b61bd574c02047f3686c74e4d9b4348a/helper/schema/set.go#L33) to compute a unique identifier for the resource - you could use that to match config order as well.

But that's not a blocker since it's not something we currently support. Switching over to that would also be a non-breaking change if we wanted to do it later. I'll open a ticket to track trying it out.

melinath:
Opened https://github.com/hashicorp/terraform-provider-google/issues/21979 and https://github.com/hashicorp/terraform-provider-google/issues/21980 to track potential improvements to our flattener sorting helpers.

Thanks for your work on this PR!

PR #13069, opened 2025-02-13
Test failure ticket creation

shuyama1:

```release-note:none

```


.ci/magician/cmd/create_test_failure_ticket.go
```diff
@@ -0,0 +1,480 @@
...
+	Use:   "create-test-failure-ticket",
+	Short: "Creates GitHub test failure tickets",
+	Long: `This command creates GitHub test failure tickets based on nighlty test status.
+
+	  It then performs the following operations:
+	  1. Calculates test failure rate for last 7 days for all tests.
+	  2. Identifies test that 
+	  		a. failed 100% in last 3 days, or 
+			b. failed 50%+ in last 7 days
+	  3. Retrieves existing active and recently closed(within 24 hours) test failrue tickets
```
melinath, line 98: nit: sp

```suggestion
	  3. Retrieves existing active and recently closed(within 24 hours) test failure tickets
```

.ci/magician/cmd/create_test_failure_ticket.go
```diff
@@ -0,0 +1,480 @@
...
+	  4. Creates new tickets dentified failing tests without corresponding tickets detected in step 3.
```
melinath, line 99:
```suggestion
	  4. Creates new tickets for identified failing tests detected in step 3 that don't already have a corresponding ticket.
```

.ci/magician/cmd/create_test_failure_ticket.go
```diff
@@ -0,0 +1,480 @@
...
+		return "100%", testFailure100
+	}
+
+	failRate := float64(failCount) / float64(n)
+	failRateStr := fmt.Sprintf("%.0f%%", failRate*100)
+
+	switch {
+	case failRate >= 0.5:
+		return failRateStr, testFailure50
+	case failRate >= 0.1:
```
melinath, line 246: I'm a little concerned that this might have false positives - if it's been passing and only the previous night failed, it'll have a 7-day failure rate of 14% and will trigger this case (getting marked as 10% even if it ends up being 100%. But I may be missing something. Maybe we could just skip labeling the 10% for now?

melinath (approved):

.ci/magician/cmd/create_test_failure_ticket.go
shuyama1: We're not currently creating tickets for failures with a lower than 50% rate, so we won't trigger the false positive test case at this time. If we decide to include lower failure rate cases under the current system, I'm thinking using a longer period, like30 days, for < 50% and implementing logic to manage ticket statuses to failure rate labels accordingly.

.ci/magician/cmd/create_test_failure_ticket.go
shuyama1: sorry for the typo

.ci/magician/cmd/create_test_failure_ticket.go
melinath: great, thanks for confirming!

PR #13070, opened 2025-02-13
(feature): Add IPV6 enum for PolicyBasedRoute.yaml and fix incorrect description for default src and dest ranges.

MatthewVu-dev:

```release-note:REPLACEME

```

### DIFF
--- a/mmv1/products/networkconnectivity/PolicyBasedRoute.yaml
+++ b/mmv1/products/networkconnectivity/PolicyBasedRoute.yaml
@@ -87,6 +87,7 @@ properties:
         required: true
         enum_values:
           - 'IPV4'
+          - 'IPV6'
  # probably could have been an enum, but it's a string in the API
       - name: 'ipProtocol'
         type: String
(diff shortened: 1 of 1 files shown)


MatthewVu-dev:
https://googlecloudplatform.github.io/magic-modules/breaking-changes/breaking-changes/#:~:text=Adding%20or%20changing%20a%20default%20value

Based on the link above, I don't think we can change the default value without breaking something. So I opted to only change the description of the protocolVersion. If there's a better way of adding a new default value please let me know.

roaks3 (dismissed):
@MatthewVu-dev correct, we wouldn't want to change an existing default in a minor release. If you plan to change the default, it could be made in the next major release ~Sept.

From what I can tell, this change is safe, since the default only behaves differently if IPV6 is used, which is a new setting.

MatthewVu-dev:
Thanks, let's try merging this PR on the week of March 17-21. That way it will release March 25th according to the new release cut timeline below. 

> "Starting the week of 2025-03-03, we will cut releases on Friday morning and release them the following Tuesday.
> 
> This means that the release that would have been on 2025-03-10 will instead be on 2025-03-11."

Alternatively, @roaks3  would it make sense to merge on the week of March 10-15? This way on Terrafrom IPV6 support will be released March 18th, and GA will subsequently occur 3 days later. This would mitigate any latency for Terraform support.

roaks3:
Yea, earlier seems fine for this. There won't be any effect to users unless they try to use the new field before GA, and the new field will only be communicated through the changelog/release notes. As long as you are comfortable with that level of communication 3 days prior to GA, that's what I would recommend doing.

And if that's the plan, could you possibly make a note to remind me to merge that week? (I will also do my best to remember)

MatthewVu-dev:
Yes lets make it that the plan. We're comfortable on our end. Let's merge this PR in on the week of 3/10/25 - 3/15/25

roaks3:
Release cutoff is tonight, so I believe this one should be merged today

MatthewVu-dev:
@roaks3  let's hold off on this, there's been some delays with a unpredicted freeze which pushing back on GA launch date.

PR #13071, opened 2025-02-13
implement better header for most mmv1 files

ScottSuarez:
partial implementation to not overload reviewer

```release-note:none

```


melinath:
could you elaborate on what the full goal here is?

melinath:
you mentioned it's a partial implementation - what will the rest be, roughly?

ScottSuarez:
> you mentioned it's a partial implementation - what will the rest be, roughly?

All the other headers. This only edits a subset of the headers. The rest are still statically coded

melinath (approved):
This LGTM but it would be great to have some unit tests if possible & also review from someone who is more familiar with the generator.

melinath:
Maybe @c2thorn or @zli82016  could take a look?

ScottSuarez:
linter is failing due to url outstripping the horizontal length allowed by the linter. I think it's still okay for review. I'd have to make changes to the linter settings here.

melinath:
we should just mark the line with nolint (or whatever the go equivalent is) - no need to change the linter settings

ScottSuarez:
> we should just mark the line with nolint (or whatever the go equivalent is) - no need to change the linter settings

adding adding an inline nolint exclusion on over a thousand files is not a good solution imo.

melinath:
> adding adding an inline nolint exclusion on over a thousand files is not a good solution imo.

I don't think we should remove our line length linting (because it is generally a good thing to lint IMO) but if there's some other way around it besides nolint that's fine with me.

ScottSuarez:
> > adding adding an inline nolint exclusion on over a thousand files is not a good solution imo.
> 
> I don't think we should remove our line length linting (because it is generally a good thing to lint IMO) but if there's some other way around it besides nolint that's fine with me.

agreed

mmv1/api/resource.go
```diff
@@ -1830,3 +1833,36 @@ func (r Resource) ShouldGenerateSweepers() bool {
...
+	}
+
+	r.githubUrl = GITHUB_BASE_URL + r.SourceYamlFile
+	return r.githubUrl
+}
+
+func (r Resource) CodeHeader(templatePath string) string {
+	templateUrl := GITHUB_BASE_URL + templatePath
+
+	return fmt.Sprintf(`// ----------------------------------------------------------------------------
```
c2thorn, line 1849: My initial thoughts are that this should be a subtemplate, but code/markdown difference is annoying enough that I'm okay with this

mmv1/provider/template_data.go
```diff
@@ -203,7 +203,14 @@ func (td *TemplateData) GenerateTGCIamResourceFile(filePath string, resource api
 func (td *TemplateData) GenerateFile(filePath, templatePath string, input any, goFormat bool, templates ...string) {
 	templateFileName := filepath.Base(templatePath)
 
-	tmpl, err := template.New(templateFileName).Funcs(google.TemplateFunctions).ParseFiles(templates...)
+	funcMap := template.FuncMap{
+		"TemplatePath": func() string { return templatePath },
+	}
+	for k, v := range google.TemplateFunctions {
```
c2thorn, line 209: Unfortunate we can't get TemplatePath in the original TemplateFunctions map, but I can't think of a better way to pass the path outside of the `input` object

c2thorn (approved):
The new output will be very helpful, thanks @ScottSuarez

mmv1/api/resource.go
```diff
@@ -1830,3 +1833,36 @@ func (r Resource) ShouldGenerateSweepers() bool {
...
+//
+//     ***     AUTO GENERATED CODE    ***    Type: MMv1     ***
+//
+// ----------------------------------------------------------------------------
+//
+//     This code is generated by Magic Modules using the following:
+//
```
zli82016, line 1856: I suspect the trailing whitespace at the end of the line is the linter issue. Can you try to remove it?

ScottSuarez: yeah you're right !

mmv1/provider/template_data.go
ScottSuarez: yeah I thought we could like have a function to return different mapping of templates but that seems excessive

zli82016 (approved):
LGTM

PR #13072, opened 2025-02-13
Update IAM v3 documentation for Principal Access Boundary Policies, and Policy Bindings.

cdutra-google:

```release-note: none
```

### DIFF
--- a/mmv1/products/iam3/PrincipalAccessBoundaryPolicy.yaml
+++ b/mmv1/products/iam3/PrincipalAccessBoundaryPolicy.yaml
@@ -53,7 +53,7 @@ examples:
       display_name: 'PAB policy for Organization'
       pab_id: 'pab-policy-for-org'
   - name: 'iam_organizations_policy_binding'
-    primary_resource_id: 'binding-for-all-org-principals'
+    primary_resource_id: 'my-pab-policy'
     test_env_vars:
       org_id: 'ORG_ID'
     vars:
--- a/mmv1/products/iam3/PrincipalAccessBoundaryPolicy.yaml
+++ b/mmv1/products/iam3/PrincipalAccessBoundaryPolicy.yaml
@@ -13,7 +13,12 @@
 
 ---
 name: 'PrincipalAccessBoundaryPolicy'
-description: An IAM Principal Access Boundary Policy resource
+description: |
+  An IAM Principal Access Boundary Policy resource. This resource has no effect on accesses until is bound to a target through policy bindings.
+  You can see further documentation on policy bindings in:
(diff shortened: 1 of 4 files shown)


mmv1/products/iam3/PrincipalAccessBoundaryPolicy.yaml
```diff
@@ -41,12 +46,20 @@ async:
...
+    primary_resource_id: 'pab-policy-for-org'
     test_env_vars:
       org_id: 'ORG_ID'
     vars:
-      display_name: 'test pab policy'
-      pab_id: 'test-pab-policy'
+      display_name: 'PAB policy for Organization'
+      pab_id: 'pab-policy-for-org'
+  - name: 'iam_organizations_policy_binding'
+    primary_resource_id: 'binding-for-all-org-principals'
```
SirGitsalot, line 56: This is the test that's failing. In this case the test is looking for a PAB policy with the id `binding-for-all-org-principals` but in the [iam_organizations_policy_binding template](https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/templates/terraform/examples/iam_organizations_policy_binding.tf.tmpl) the value is hard coded as `pab_policy`. That said, even if the example were updated it wouldn't be testing anything that isn't already tested in `iam_principal_access_boundary_policy` and the example is still available under OrganizationsPolicyBinding so it's not really adding anything. If you remove this new example block the tests should pass.

SirGitsalot (changes requested):
LGTM except for one suggestion to fix the new test failure.

mmv1/products/iam3/PrincipalAccessBoundaryPolicy.yaml
```diff
@@ -41,12 +46,20 @@ async:
   include_project: true
 examples:
   - name: 'iam_principal_access_boundary_policy'
+    primary_resource_id: 'pab-policy-for-org'
+    test_env_vars:
+      org_id: 'ORG_ID'
+    vars:
+      display_name: 'PAB policy for Organization'
+      pab_id: 'pab-policy-for-org'
+  - name: 'iam_organizations_policy_binding'
```
SirGitsalot, line 55: The test is failing because this block also needs a `external_providers: ["time"]` but really you can just delete the test altogether

SirGitsalot (changes requested):
The unit test failure is due to the build being broken at HEAD, it's being fixed now.

SirGitsalot (commented):
Kicked off the tests - I believe if you [join the Google GitHub org](go/github-docs/accounts#register) the tests will run automatically

SirGitsalot (approved):

PR #13075, opened 2025-02-13
fix backup vault tests

NickElliot:
Fixes https://github.com/hashicorp/terraform-provider-google/issues/21159
Fixes https://github.com/hashicorp/terraform-provider-google/issues/21179

These tests were failing due to not being seeded with the data needed -- the backup vault data source would not generate until a scheduled job is ran (which can not be handled in terraform), and in the time since our CI and beta environments were seeded with the original data, the behavior changed to always use generated random unique identifiers for these data sources rather than aligning with the display name.

This fix has amended the tests to use the unique identifier of the backup vault now present in our GA environment, and include a comment mentioning these tests will not be runnable locally without proper preparation. It might be ugly to be showing this unique identifier, this file is not used for a documentation sample, and functionally it has the same result (i.e. users know the location of a resource in our project). 

Potentially this approach could cause issues down the line if we begin testing GA VCRs in pull request CI, I don't believe we have any plans to do so at this time.

```release-note:none

```

### DIFF
--- a/mmv1/third_party/terraform/services/backupdr/data_source_backup_dr_backup_test.go.tmpl
+++ b/mmv1/third_party/terraform/services/backupdr/data_source_backup_dr_backup_test.go.tmpl
@@ -18,9 +18,9 @@ func TestAccDataSourceGoogleCloudBackupDRBackup_basic(t *testing.T) {
 	project := envvar.GetTestProjectFromEnv()
 	location := "us-central1"
 	backupVaultId := "bv-test"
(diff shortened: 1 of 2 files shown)


melinath (approved):
This will absolutely cause problems when we try to run VCR GA tests; really, this should be configured using env vars. However, we don't have the scaffolding for that at the moment. This works as a quick fix for now.

NickElliot:
I should clarify my "potentially" was in terms of "if we begin doing that" 😅

PR #13078, opened 2025-02-14
make Firewall Policy Association mutable

NickElliot:
Current result of exploration on this issue: using the `replaceExistingAssociation` parameter, the resource is able to be updated following the expected paradigm i.e. the fine grained `firewall_policy_association` is not actually updated, but a new association is added to a different firewall policy, and given a flag allowing it to "steal" the attachment target from the existing association to previous firewall policy. If a `lifecycle.create_before_destroy` flag is set on the firewall policy, and an immutable characteristic is changed, the operations will successfully process for creating a new firewall policy, and updating the association without introducing a gap in policy-association uptime.

However there are problems with this straight forward approach:
* if the `replaceExistingAssociation` addition fails, the resource breaks within Terraform due to some unseen change in the API end. The "existing association" will no longer be reachable, despite still existing. I was unable to determine why this is the case, but all operations on the resource will fail with the following message: "Error 400: Invalid value for field 'name': '{{name}}'. An association with that name does not exist., invalid". While existing infrastructure is undamaged, this will force the resource to be removed from state.
* when a `replaceExistingAssociation` operation succeeds, temporarily the existing association persists, which can cause the older firewall policy to fail to delete. this does not cause a critical failure, as the policy can be deleted on a retry soon afterward
* lastly, these do not mitigate the ability to cause the original critical failure, as without a "create before destroy" being configured, the original delete-delete-create-create steps can still occur and fail

The following config can be used to demo the swapover scenarios, replacing `{{organization}}` with your local test org:
https://paste.googleplex.com/4641467795243008

```release-note:enhancement
compute: update support has been added for the `firewall_policy` field in `google_compute_firewall_policy_association` resource. It is recommended to only perform this operation in combination with a protective lifecycle tag such as "create_before_destroy" or "prevent_destroy" on your previous `firewall_policy` resource in order to prevent situations where a target attachment has no associated policy.
```


mmv1/templates/terraform/post_update/compute_firewall_policy_association_update_sleep.go.tmpl
```diff
@@ -0,0 +1,9 @@
+  // Store the ID now
+  id, err := tpgresource.ReplaceVars(d, config, "locations/global/firewallPolicies/{{"{{"}}firewall_policy{{"}}"}}/associations/{{"{{"}}name{{"}}"}}")
+  if err != nil {
+    return fmt.Errorf("Error constructing id: %s", err)
+  }
+  d.SetId(id)
+
+  //following the swapover briefly both associations exist and this can cause operations to fail
+  time.Sleep(30 * time.Second)
```
c2thorn, line 9: Is this sufficient time in your testing? Should we err on the side of more?

mmv1/products/compute/FirewallPolicyAssociation.yaml
```diff
@@ -52,14 +54,31 @@ examples:
...
     description: |
       The firewall policy of the resource.
+
+      This field can be updated to refer to a different Firewall Policy, which will create a new association from that new
+      firewall policy with the flag to override the existing attachmentTarget's policy association.
+
+      Due to potential risks with this operation it is *highly* recommended to use the `create_before_destroy` life cycle option
+      on your exisiting firewall policy so as to prevent a situation where your attachment target has no associated policy.
+
+      If this update operation fails this FirewallPolicyAssociation will remain in place,
```
c2thorn, line 78: Is there a particular reason why an update operation would fail? Would it fail at this point due to incorrect user input for example?

c2thorn (commented):
couple of questions

mmv1/templates/terraform/post_update/compute_firewall_policy_association_update_sleep.go.tmpl
NickElliot: 30s proved fine over several runs but bumping it to 60 would ensure nothing happens

mmv1/products/compute/FirewallPolicyAssociation.yaml
NickElliot: I think this might be resolved with the ID setting being a part of the update call actually, but I'm not certain. I'll run some manual scenarios, but what originally triggered it for me was simply typoing the new name

mmv1/templates/terraform/post_update/compute_firewall_policy_association_update_sleep.go.tmpl
c2thorn: I think the infrequency of an update should make it fine to bump it to a longer time. I'll leave it up to your discretion

mmv1/products/compute/FirewallPolicyAssociation.yaml
c2thorn: Would the result of your manual scenarios warrant changes in this PR? Like more specific documentation? Or a code change

NickElliot: seeing if the documentation is still necessary

mmv1/products/compute/FirewallPolicyAssociation.yaml
```diff
@@ -52,14 +55,31 @@ examples:
...
       The firewall policy of the resource.
+
+      This field can be updated to refer to a different Firewall Policy, which will create a new association from that new
+      firewall policy with the flag to override the existing attachmentTarget's policy association.
+
+      **Note** Due to potential risks with this operation it is *highly* recommended to use the `create_before_destroy` life cycle option
+      on your exisiting firewall policy so as to prevent a situation where your attachment target has no associated policy.
+
+      If this update operation fails this FirewallPolicyAssociation will remain in place,
+      but the resource will no longer be manageable via Terraform and require manual updates to your resources in gcloud.
```
NickElliot, line 80:
```suggestion
```

NickElliot:
resolved the brick issue -- I'm not totally sure why it happened still, but the state value for `firewall_policy` would become updated by attempting an update and cause all ensuing read attempts to fail, thus preventing any plans. I tried moving around the existing custom expand/removing ignore read/adding a custom flattener, and none of them seemed to make a difference so I have to assume its related to how URL parameters are handled in Terraform core.

mmv1/templates/terraform/custom_update/firewall_policy_association_update.go.tmpl
```diff
@@ -0,0 +1,71 @@
...
+		Method:    "POST",
+		Project:   billingProject,
+		RawURL:    url,
+		UserAgent: userAgent,
+		Body:      obj,
+		Timeout:   d.Timeout(schema.TimeoutUpdate),
+		Headers:   headers,
+	})
+
+	if err != nil {
```
NickElliot, line 52: this is the custom section of. the code

c2thorn (approved):
LGTM pending minor comment

NickElliot:
only diff from previous build is the 3 lines of comments -- will go ahead with merge once checks finish

PR #13080, opened 2025-02-14
Fixes issue #21346 related to google_chronicle_rule_deployment resource

ankitgoyal0301:

```release-note: bug
chronicle: fixed bug setting `enabled` on creation in `google_chronicle_rule_deployment`
```


ankitgoyal0301:
**Some more context on the PR:** 
The problem here was that ruledeployment is singleton and created along with the rules resource (even before we actually make ruledeployment call via terraform). And we don't have a create method for that explicitly.

Currently, for terraform to know about it, we make a PATCH call (underneath what we call the terraform's create call -  first call made by terraform for ruledeployment resource). This makes terraform aware of what the state of resource is. So the issues addressed in the attached github issue were due to this reason.

As a fix, I have added custom create code to it, which basically reads the ruledeployment resource first (before the PATCH call), and then compares the user terraform code with the current state of the resource, and hence only sends PATCH calls for the updates that the user makes. This resolves the issue specified in the issue (enabled is already false, OR run_frequency is already LIVE) because we only make the PATCH call for the differences between the user terraform code and ruledeployment state.

mmv1/templates/terraform/custom_create/chronicle_rule_deployment.go.tmpl
```diff
@@ -0,0 +1,173 @@
...
+    enabledValue, enabledExists:= res["enabled"]
+    if enabledExists {
+        enabled:= flattenChronicleRuleDeploymentEnabled(enabledValue, d, config)
+        if!reflect.DeepEqual(enabledProp, enabled) {
+            obj["enabled"] = enabledProp
+            updateMask = append(updateMask, "enabled")
+        }
+    } else {
+        // Handle the case where "enabled" is missing from the API response
+        if v, ok:= d.GetOkExists("enabled"); ok &&!tpgresource.IsEmptyValue(reflect.ValueOf(v)) {
```
ronakgadia, line 67: && !tpgresource.IsEmptyValue(reflect.ValueOf(v)) - space between && and !value

Here and below

ronakgadia (approved):

mmv1/templates/terraform/custom_create/chronicle_rule_deployment.go.tmpl
ankitgoyal0301: Made the change!

ankitgoyal0301:
The failed unit tests seem to be unrelated to my change to chronicle: https://screenshot.googleplex.com/3BU33jRCz8PBpm6

ankitgoyal0301:
> The failed unit tests seem to be unrelated to my change to chronicle resource: https://screenshot.googleplex.com/3BU33jRCz8PBpm6

@slevenick Tagging this for visibility.

PR #13085, opened 2025-02-14
Fix mmv1 merge function by removing private field from struct

roaks3:
Fixes b/396575323

https://github.com/GoogleCloudPlatform/magic-modules/pull/13071 added a `githubUrl` field that causes an error for our private provider build (mmv1/api/product.go:280):

```
panic: reflect.Value.Interface: cannot return value obtained from unexported field or method
```

Because it is not exportable, it breaks our merge function and generation panics.

It looks like we don't need the caching as best I can tell, so I went ahead and removed it.

```release-note:none

```

### DIFF
--- a/mmv1/api/resource.go
+++ b/mmv1/api/resource.go
@@ -342,7 +342,6 @@ type Resource struct {
 
 	ImportPath     string `yaml:"-"`
 	SourceYamlFile string `yaml:"-"`
-	githubUrl      string
 }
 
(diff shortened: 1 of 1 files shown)


ScottSuarez (approved):

ScottSuarez:
ah ! sorry about that!!

 yeah the caching is very minor optimization and can be omitted

roaks3:
Awesome thanks for confirming!

PR #13088, opened 2025-02-14
Bigtable: fixes updating clusters with autoscaling

andre-sampaio:
Prevents adding `num_nodes` to update cluster requests for clusters that have `autoscaling` enabled, since that results in errors. Also included a test for that scenario.

The github issue has an example of how to trigger this bug.

Fixes https://github.com/hashicorp/terraform-provider-google/issues/21434

```release-note: bug
bigtable: fixed a bug where sometimes updating an instance's cluster list could result in an error if there was an existing cluster with autoscaling enabled
```

### DIFF
--- a/mmv1/third_party/terraform/services/bigtable/resource_bigtable_instance_test.go
+++ b/mmv1/third_party/terraform/services/bigtable/resource_bigtable_instance_test.go
@@ -1007,7 +1007,6 @@ resource "time_offset" "week-in-future" {
 `
 }
 
-
 type cluster struct {
 	clusterId         string
 	zone              string
--- a/mmv1/third_party/terraform/services/bigtable/resource_bigtable_instance_test.go
+++ b/mmv1/third_party/terraform/services/bigtable/resource_bigtable_instance_test.go
@@ -4,8 +4,6 @@ import (
 	"context"
 	"fmt"
 	"regexp"
-	"strconv"
-	"strings"
 	"testing"
 
 	"github.com/hashicorp/terraform-provider-google/google/acctest"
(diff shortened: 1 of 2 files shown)


ron-gal (approved):

melinath:
Manual test run: https://hashicorp.teamcity.com/buildConfiguration/TerraformProviders_GoogleCloud_GOOGLE_BETA_MMUPSTREAMTESTS_GOOGLEBETA_PACKAGE_BIGTABLE/323174

mmv1/third_party/terraform/services/bigtable/resource_bigtable_instance_test.go
```diff
@@ -249,6 +251,80 @@ func TestAccBigtableInstance_createWithAutoscalingAndUpdate(t *testing.T) {
...
+	}
+
+	acctest.VcrTest(t, resource.TestCase{
+		PreCheck:                 func() { acctest.AccTestPreCheck(t) },
+		ProtoV5ProviderFactories: acctest.ProtoV5ProviderFactories(t),
+		CheckDestroy:             testAccCheckBigtableInstanceDestroyProducer(t),
+		Steps: []resource.TestStep{
+			{
+				// Create Autoscaling config with 1 nodes. Default storage_target is set by service based on storage type.
+				Config: getInstanceResource(instanceName, []cluster{*cluster1}),
```
melinath, line 290: Please use a standard config function that uses Nprintf like the others in this file - it's easier to read and it's important for our ability to analyze tests.

melinath (changes requested):
normally we'd want to handle cases like this with improved validation (such as verifying that num nodes isn't set if autoscaling_configs is set) which could be done with a customize diff - but the fact that this is a computed + optional field makes things tricky since it could have a value persisting from when it was last set which then causes the error described. Given that, I think this proposed fix is fine - the test just needs to be written using the standard Nprintf method of building configs. (The manual test run has been queued for a while so I'll have to check on it next week, but I'll need to re-run anyway after you change the test.)

mmv1/third_party/terraform/services/bigtable/resource_bigtable_instance_test.go
andre-sampaio: Changed the test to use printf

andre-sampaio:
> [...] I'll have to check on it next week [...]

np!

brandtnewton (approved):
LGTM. Thanks for the quick fix!

ron-gal (approved):

melinath:
manual test run: https://hashicorp.teamcity.com/buildConfiguration/TerraformProviders_GoogleCloud_GOOGLE_BETA_MMUPSTREAMTESTS_GOOGLEBETA_PACKAGE_BIGTABLE/325349

melinath (approved):
I'm approving this because it seems to fix the issue. There is a small issue with the UX since users could think the field is set (but actually their configuration is being ignored) which will likely surface as a permadiff - but that is a less severe issue than the current bug and can be fixed separately.

PR #13090, opened 2025-02-17
Suppressing diff for database_version for MySQL 8.0

ankushkhanna1998:
Suppressing diff for database_version for MySQL 8.0. With introduction of auto version upgrade for MySQL 8.0.35 and beyond, customers will be auto upgraded to a newer version which will lead to change in database version.

Hence, to avoid breakage of terraform scripts suppressing this diff while maintaining the sanity that the major version does not change, suppression is done only if there is a diff in minor version.

```release-note:enhancement
sql: added `databaseVersionDiffSuppress` to `google_sql_database_instance`. This change will suppress diff for database_version for MySQL 8.0 when the version is updated by auto version upgrade.
```

### DIFF
--- a/mmv1/third_party/terraform/services/spanner/resource_spanner_instance_test.go
+++ b/mmv1/third_party/terraform/services/spanner/resource_spanner_instance_test.go
@@ -602,6 +602,7 @@ resource "google_spanner_instance" "main" {
 func testAccSpannerInstance_spannerInstanceWithAutoscaling(context map[string]interface{}) string {
 	return acctest.Nprintf(`
 resource "google_spanner_instance" "example" {
+  name         = "tf-test-spanner-instance-%{random_suffix}"
   config       = "regional-us-central1"
   display_name = "Test Spanner Instance"
   autoscaling_config {


mmv1/third_party/terraform/services/sql/resource_sql_database_instance.go.tmpl
```diff
@@ -2148,6 +2149,16 @@ func maintenanceVersionDiffSuppress(_, old, new string, _ *schema.ResourceData)
 	}
 }
 
+func databaseVersionDiffSuppress(_, old, new string, _ *schema.ResourceData) bool {
```
isaurabhuttam, line 2152: Please include the test cases covering all the engine to make sure there is no diff is being generated or correct on is not being suppressed.

ankushkhanna1998: Added relevant unit test cases. Please check.

ScottSuarez (changes requested):
Looks like there is a related test failure
TestAccSqlDatabaseInstance_activationPolicy

PR #13094, opened 2025-02-17
Add Optional Log Fields to google_cloud_region_backend_service

Dawid212:
This PR adds optional log fields to to google_cloud_region_backend_service

### Related Issues

- Resolves: https://github.com/hashicorp/terraform-provider-google/issues/19187

### Release Note

```release-note:enhancement
compute: added `log_config.optional_mode` and `log_config.optional_fields` fields to `google_compute_region_backend_service` resource  
```

### DIFF
--- a/mmv1/products/compute/RegionBackendService.yaml
+++ b/mmv1/products/compute/RegionBackendService.yaml
@@ -757,9 +757,7 @@ properties:
     output: true
   - name: 'iap'
     type: NestedObject
-    description: |
-      Settings for enabling Cloud Identity Aware Proxy.
-      If OAuth client is not set, Google-managed OAuth client is used.
+    description: Settings for enabling Cloud Identity Aware Proxy
     default_from_api: true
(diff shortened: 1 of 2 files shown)


roaks3 (approved):
Code LGTM, just need to make sure tests pass

roaks3 (changes requested):
Note that only `TestAccComputeRegionBackendService_withLogConfig` looks relevant to this PR. It seems the update to include optional fields is failing in some way.

roaks3 (changes requested):
Sorry, while the test I highlighted in my last comment is now fixed, the other tests do appear to be a result of this change. I verified by checking our nightly tests, which show these tests passing consistently.

It looks like there is server-side default behavior that you'll need to work out to make sure that Terraform stays happy (I'm guessing `optionalMode` needs `default_from_api: true`, and `optionalFields` might too)

roaks3 (approved):

PR #13111, opened 2025-02-18
bugfix improve provider credentials docs. Fixes hashicorp#21471

dolfandringa:
Adds more docs on the `credentials` argument to the google provider.

## Related Issues

https://github.com/hashicorp/terraform-provider-google/issues/21471

```release-note:none
Added documentation on how to use the `credentials` parameter for the google terraform provider
```

### DIFF
--- a/mmv1/third_party/terraform/website/docs/guides/provider_reference.html.markdown
+++ b/mmv1/third_party/terraform/website/docs/guides/provider_reference.html.markdown
@@ -116,8 +116,8 @@ impersonated service account regardless of the primary identity in use.
 
 ## Authentication Configuration
 
-* `credentials` - (Optional) Either the path to or the contents of a
-[service account key file] in JSON format. You can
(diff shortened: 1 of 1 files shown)


ScottSuarez (approved):

rileykarson:
@ScottSuarez I don't think this change is correct. The `file()` function returns the contents of a path, and is equivalent to supplying the contents directly. We also support supplying a path directly, and open the field directly from the provider. Can you revert? Thanks!

ScottSuarez:
> @ScottSuarez I don't think this change is correct. The `file()` function returns the contents of a path, and is equivalent to supplying the contents directly. We also support supplying a path directly, and open the field directly from the provider. Can you revert and reopen the issue? Thanks!

You got it ! Thanks for double checking this

dolfandringa:
For me supplying a path directly caused an error with the provider trying to interpret json and encountering `/` as the first character. @ScottSuarez and @rileykarson

PR #13116, opened 2025-02-19
manage ticket - remove review label

shuyama1:
This is a temporary workaround to remove `forward/review` label for test failure ticket, as https://github.com/GoogleCloudPlatform/magic-modules/pull/13086 is not working as expected. I'll refactor some of the labeler code for a proper fix later.

```release-note:none

```

### DIFF
--- a/.ci/gcb-test-failure-ticket.yaml
+++ b/.ci/gcb-test-failure-ticket.yaml
@@ -1,24 +1,24 @@
 ---
 steps:
-    # - name: 'gcr.io/graphite-docker-images/go-plus'
-    #   id: collect-nightly-test-status
-    #   entrypoint: '/workspace/.ci/scripts/go-plus/magician/exec.sh'
-    #   secretEnv: ["TEAMCITY_TOKEN"]
-    #   args:
-    #     - 'collect-nightly-test-status'
-    #     - $_CUSTOM_DATE
-    # - name: 'gcr.io/graphite-docker-images/go-plus'
-    #   id: create-test-failure-ticket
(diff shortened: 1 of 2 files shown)


.ci/gcb-test-failure-ticket.yaml
```diff
@@ -14,6 +14,13 @@ steps:
       waitFor: ["collect-nightly-test-status"]
       args:
         - 'create-test-failure-ticket'
+    - name: 'gcr.io/graphite-docker-images/go-plus'
```
melinath, line 17: I think we should wait a bit to ensure that the github labeler had time to run

```suggestion
    - name: 'ubuntu'
      args: ['sleep', '120']
    - name: 'gcr.io/graphite-docker-images/go-plus'
```

melinath (approved):

melinath:
I feel a little bit silly for not suggesting this sooner but another option would be to make create_test_failure_ticket compute & add a service label when creating the issue. Then the labeler would just ignore the issue because it already had a service label.

melinath:
But I'm happy to go with this path since it's already done.

shuyama1:
> I feel a little bit silly for not suggesting this sooner but another option would be to make create_test_failure_ticket compute & add a service label when creating the issue. Then the labeler would just ignore the issue because it already had a service label

ah, good point. yeah, that should work as well. Thanks. Let's check in this first and I'll work on a follow-up fix to either make the labeler list existing labels and skip applying review label or make create_test_failure_ticket to apply service labels.

PR #13119, opened 2025-02-19
fix: (storage) fix multi location rpo test case

gurusai-voleti:
Fixes: https://github.com/hashicorp/terraform-provider-google/issues/21270

```release-note:none
```


mmv1/third_party/terraform/services/storage/resource_storage_bucket_test.go.tmpl
```diff
@@ -379,7 +379,12 @@ func TestAccStorageBucket_dualLocation_rpo(t *testing.T) {
 func TestAccStorageBucket_multiLocation_rpo(t *testing.T) {
```
kautikdk, line 379: Looking into this test deeply, I don't think we even need this test. We can probably add same validations in other basic bucket create tests because `rpo` is having server side `DEFAULT` value for the multi-location buckets. We don't need to have a separate test for checking this explicitly for multi-location buckets.
Probably we can add default validation here: https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/third_party/terraform/services/storage/resource_storage_bucket_test.go.tmpl#L34

gurusai-voleti: can we remove this multilocation_rpo test case completely?

c2thorn: waiting for this before reviewing

c2thorn (commented):
waiting for removal question

mmv1/third_party/terraform/services/storage/resource_storage_bucket_test.go.tmpl
kautikdk: > should this multilocation_rpo test case be removed completely?
Yes, I think so.

kautikdk (approved):

c2thorn (approved):
this change was not user-facing, so I have removed the changelog note FYI.

PR #13050, opened 2025-02-12
Promote Gemini Logging Setting Binding from google-beta to google

PerlMonker303:
**Release Note Template for Downstream PRs (will be copied)**

See [Write release notes](https://googlecloudplatform.github.io/magic-modules/contribute/release-notes/) for guidance.

Promote resources for Gemini Admin Control
```release-note:new-resource
`google_gemini_logging_setting_binding` (ga)
```

PR #13121, opened 2025-02-19
Promote Gemini Data Sharing With Google Setting Binding from google-beta to google

PerlMonker303:

Promote resources for Gemini Admin Control
```release-note:enhancement
gemini: promoted `google_gemini_data_sharing_with_google_setting_binding` resource to ga.
```

### DIFF
--- a/mmv1/products/gemini/DataSharingWithGoogleSettingBinding.yaml
+++ b/mmv1/products/gemini/DataSharingWithGoogleSettingBinding.yaml
@@ -17,7 +17,6 @@ description: The resource for managing DataSharingWithGoogle setting bindings fo
 references:
   guides:
     'Gemini Cloud Assist overview': 'https://cloud.google.com/gemini/docs/cloud-assist/overview'
-min_version: 'beta'
 base_url: projects/{{project}}/locations/{{location}}/dataSharingWithGoogleSettings/{{data_sharing_with_google_setting_id}}/settingBindings
 self_link: projects/{{project}}/locations/{{location}}/dataSharingWithGoogleSettings/{{data_sharing_with_google_setting_id}}/settingBindings/{{setting_binding_id}}
 create_url: projects/{{project}}/locations/{{location}}/dataSharingWithGoogleSettings/{{data_sharing_with_google_setting_id}}/settingBindings?settingBindingId={{setting_binding_id}}
@@ -29,7 +28,6 @@ import_format:
 mutex: projects/{{project}}/locations/{{location}}/dataSharingWithGoogleSettings/{{data_sharing_with_google_setting_id}}
 examples:
   - name: gemini_data_sharing_with_google_setting_binding_basic
-    min_version: 'beta'
     primary_resource_id: example
     exclude_test: true
     vars:
--- a/mmv1/third_party/terraform/services/gemini/resource_gemini_data_sharing_with_google_setting_binding_test.go.tmpl
+++ b/mmv1/third_party/terraform/services/gemini/resource_gemini_data_sharing_with_google_setting_binding_test.go.tmpl
@@ -1,5 +1,4 @@
 package gemini_test
-{{- if ne $.TargetVersionName "ga" }}
 
 import (
     "fmt"
@@ -21,7 +20,7 @@ func TestAccGeminiDataSharingWithGoogleSettingBinding_update(t *testing.T) {
 
 	acctest.VcrTest(t, resource.TestCase{
 		PreCheck:                 func() { acctest.AccTestPreCheck(t) },
-		ProtoV5ProviderFactories: acctest.ProtoV5ProviderBetaFactories(t),
+		ProtoV5ProviderFactories: acctest.ProtoV5ProviderFactories(t),
 		Steps: []resource.TestStep{
 			{
 				Config: testAccGeminiDataSharingWithGoogleSettingBinding_basic(context),
@@ -53,11 +52,9 @@ func TestAccGeminiDataSharingWithGoogleSettingBinding_update(t *testing.T) {
 func testAccGeminiDataSharingWithGoogleSettingBinding_basic(context map[string]interface{}) string {
 	return acctest.Nprintf(`
 data "google_project" "project" {
-    provider = google-beta
 }
 
 resource "google_gemini_data_sharing_with_google_setting" "basic" {
-    provider = google-beta
     data_sharing_with_google_setting_id = "%{data_sharing_with_google_setting_id}"
     location = "global"
     labels = {"my_key" = "my_value"}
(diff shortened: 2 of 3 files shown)


shuyama1:
Running GA tests on [TC](https://hashicorp.teamcity.com/buildConfiguration/TerraformProviders_GoogleCloud_GOOGLE_MMUPSTREAMTESTS_GOOGLE_PACKAGE_GEMINI/326344). I'll update test result after it's done

shuyama1 (commented):
Test failed due to error:
```
resource_gemini_data_sharing_with_google_setting_binding_test.go:23: Step 1/4 error: After applying this test step, the plan was not empty.
        stdout:
          ~ update in-place
          # google_gemini_data_sharing_with_google_setting_binding.basic_binding will be updated in-place
          ~ resource "google_gemini_data_sharing_with_google_setting_binding" "basic_binding" {
                id                                  = "projects/ci-test-project-188019/locations/global/dataSharingWithGoogleSettings/tf-test-ls-anbpja9s28/settingBindings/tf-test-lsb-zpgyv39zxx"
                name                                = "projects/ci-test-project-188019/locations/global/dataSharingWithGoogleSettings/tf-test-ls-anbpja9s28/settingBindings/tf-test-lsb-zpgyv39zxx"
              - product                             = "GEMINI_CLOUD_ASSIST" -> null
                # (9 unchanged attributes hidden)
            }
```
I've checked our nightly tests and it seems the test has failed with the same error in our beta environment. Let's fix that in this PR.

The diff is likely due to the API returns default value for unset field `product`. https://googlecloudplatform.github.io/magic-modules/develop/diffs/#fix-diffs

PerlMonker303:
I have encountered the error before, but it is weird how similar PRs did not have this error and they have identical setups - see https://github.com/GoogleCloudPlatform/magic-modules/pull/13051 and https://github.com/GoogleCloudPlatform/magic-modules/pull/13050.

shuyama1 (changes requested):
> I have encountered the error before, but it is weird how similar PRs did not have this error and they have identical setups - see #13051 and #13050.

I think the issue occurs in `google_gemini_data_sharing_with_google_setting_binding` specifically. Modifying the test config would not fix the issue. This diff happens when the field is not explicitly specified but the API returns value. You'll need to add `default_value: DEFAULT_VALUE` or `default_from_api: true` to the field depends on if it has a fixed default value. See details at https://googlecloudplatform.github.io/magic-modules/develop/diffs/#default

shuyama1 (commented):
> Added a fix after testing locally. This should work.

Thank you for updating the tests. I'd prefer we have the update test set as before - product field unset for creation and change it to a non-default value for update.
Running the previous test in GA, we got the following error:
```
=== RUN   TestAccGeminiDataSharingWithGoogleSettingBinding_update
=== PAUSE TestAccGeminiDataSharingWithGoogleSettingBinding_update
=== CONT  TestAccGeminiDataSharingWithGoogleSettingBinding_update
    resource_gemini_data_sharing_with_google_setting_binding_test.go:23: Step 3/4 error: Error running apply: exit status 1
        Error: Error waiting for Updating DataSharingWithGoogleSettingBinding: Error code 3, message: product "GEMINI_CODE_ASSIST" is not supported by the setting type "dataSharingWithGoogleSettings", allowed product is "GEMINI_CLOUD_ASSIST": invalid argument
          with google_gemini_data_sharing_with_google_setting_binding.basic_binding,
          on terraform_plugin_test.tf line 12, in resource "google_gemini_data_sharing_with_google_setting_binding" "basic_binding":
          12: resource "google_gemini_data_sharing_with_google_setting_binding" "basic_binding" {
--- FAIL: TestAccGeminiDataSharingWithGoogleSettingBinding_update (39.27s)
```
Is the `GEMINI_CLOUD_ASSIST` currently no supported? or is the value only supported in beta?

PerlMonker303:
"GEMINI_CLOUD_ASSIST" should be supported. I have changed the tests now, let me know if it passes.

shuyama1:
> GEMINI_CODE_ASSIST

Sorry, I mean "GEMINI_CODE_ASSIST" (the non-default value). Is that supported, as we set it before in the test and the GA failure indicated that it was not supported

mmv1/third_party/terraform/services/gemini/resource_gemini_data_sharing_with_google_setting_binding_test.go.tmpl
```diff
@@ -53,50 +52,44 @@ func TestAccGeminiDataSharingWithGoogleSettingBinding_update(t *testing.T) {
...
     enable_preview_data_sharing = true
 }
 
 resource "google_gemini_data_sharing_with_google_setting_binding" "basic_binding" {
-    provider = google-beta
     data_sharing_with_google_setting_id = google_gemini_data_sharing_with_google_setting.basic.data_sharing_with_google_setting_id
     setting_binding_id = "%{setting_binding_id}"
     location = "global"
     target = "projects/${data.google_project.project.number}"
+	product = "GEMINI_CLOUD_ASSIST"
```
shuyama1, line 69:
```suggestion
```
Let's remove this. Just want to make sure unset the field does not cause diff now. thanks!

PerlMonker303:
Sorry for the confusion, you are right. GEMINI_CODE_ASSIST is not supported for the product field for this resource. I removed it from possible values. I think things should be good to go now.

mmv1/products/gemini/DataSharingWithGoogleSettingBinding.yaml
```diff
@@ -79,7 +77,7 @@ properties:
       Product type of the setting binding.
     enum_values:
       - 'GEMINI_CLOUD_ASSIST'
-      - 'GEMINI_CODE_ASSIST'
```
shuyama1, line 82: Just need to confirm: Is this value not supported in both GA and beta? Does specifying `GEMINI_CODE_ASSIST` results in API error in all cases? I'm asking this because this change is making the validation of the field more strict, which is a [breaking change](https://googlecloudplatform.github.io/magic-modules/breaking-changes/breaking-changes/#making-validation-more-strict) in Terraform that we'd avoid in minor releases. The only one exception is that if the bug impacts all configurations, then we can fix it in a minor, see https://googlecloudplatform.github.io/magic-modules/breaking-changes/make-a-breaking-change/#in-minor-releases

PerlMonker303: Yes, this value is not supported in any environment (GA/beta). Specifying GEMINI_CODE_ASSIST results in all cases in errors for this resource. We recently added this more strict behavior for the product field.

This is a bug fix that impacts all configurations, therefore according to [the doc you referenced](https://googlecloudplatform.github.io/magic-modules/breaking-changes/make-a-breaking-change/#in-minor-releases) we should be able to fix it in a minor release (here).

shuyama1 (approved):

PR #12672, opened 2024-12-31
Add missing fields to compute resources [KMS and disks]

karolgorc:
closes https://github.com/hashicorp/terraform-provider-google/issues/18050

Hi this is a PR to fix feature gap between the API and the provider on disks and encryption. I'll look for any tickets that this fixes and update them later.

Tested this for basic breaking changes (created an instance on upstream provider and did a terraform apply on the custom provider) - no changes in configuration shown

On some of these fields i wasn't able to add a `ConflictsWith` param. LMK if the API errors are sufficient in these cases or do we want this to throw an error on terraform's side

Also for the `guest_os_features` field i've made it so that it is computed but it will show null until the user sets it in his config. This field changes based on the source used for the disk and is ForceNew, so i think that this it's a safer behavior to have it null until the user specifficaly sets it in his config.

Fields added:

- **google_compute_instance**
  - boot_disk.architecture
  - boot_disk.guest_os_features
  - boot_disk.disk_encryption_key_rsa
  - boot_disk.disk_encryption_service_account
  - boot_disk.initialize_params.source_image_encryption_key
    - boot_disk.initialize_params.source_image_encryption_key.raw_key
    - boot_disk.initialize_params.source_image_encryption_key.rsa_encrypted_key
    - boot_disk.initialize_params.source_image_encryption_key.kms_key_self_link
    - boot_disk.initialize_params.source_image_encryption_key.kms_key_service_account
    - boot_disk.initialize_params.source_image_encryption_key.sha256
  - boot_disk.initialize_params.snapshot
  - boot_disk.initialize_params.source_snapshot_encryption_key
    - boot_disk.initialize_params.source_snapshot_encryption_key.raw_key
    - boot_disk.initialize_params.source_snapshot_encryption_key.rsa_encrypted_key
    - boot_disk.initialize_params.source_snapshot_encryption_key.kms_key_self_link
    - boot_disk.initialize_params.source_snapshot_encryption_key.kms_key_service_account
    - boot_disk.initialize_params.source_snapshot_encryption_key.sha256
  - attached_disk.disk_encryption_key_rsa
  - attached_disk.disk_encryption_service_account
  - instance_encryption_key
    - instance_encryption_key.kms_key_self_link
    - instance_encryption_key.kms_key_service_account
    - instance_encryption_key.sha256

- **google_compute_image**
  - image_encryption_key.raw_key
  - image_encryption_key.rsa_encrypted_key

- **google_compute_snapshot**
[...]

PR #13125, opened 2025-02-19
Revert "Add missing fields to compute resources [KMS and disks]"

NickElliot:
Reverts GoogleCloudPlatform/magic-modules#12672

usages of `boot_disk` are segfaulting due to a bad check here: https://github.com/karolgorc/magic-modules/blob/bff8190d005193b98bab2cb07587a1d6c8cdea4f/mmv1/third_party/terraform/services/compute/resource_compute_instance.go.tmpl#L3637

additionally, the preceding line is a bad check as `guest_os_features` is not an implemented member of the `initialize_params` schema


NickElliot:
`mmv1/third_party/tgc/tfdata/fake_resource_data_test.go` was moved to be generated within MMV1 to resolve TGC issues that manifested as a result of updating the compute resources, the lines that required change as a result of the original change have been altered within this revert

NickElliot:
@karolgorc FYI, if you want to get ahead on revising the original. I encountered the segfault running the `TestAccComputeInstance_regionBootDisk` test locally, let me know if you need help getting a local env for it online.

zli82016:
It looks like the release note is missing.

mmv1/third_party/tgc/tfdata/fake_resource_data_test.go
```diff
@@ -242,9 +242,7 @@ func TestFakeResourceData_getOkTypeObject(t *testing.T) {
 	assert.Equal(t, map[string]interface{}{
 		"device_name":                     "test-device_name",
 		"disk_encryption_key_raw":         "",
-		"disk_encryption_key_rsa":         "",
```
zli82016, line 245: This file needs go format after removing the fields.

NickElliot:
added

PR #8432, opened 2023-07-24
Add internal IPV6 support for GoogleComputeAddress and ComputeInstance resources

simonebruzzechesse:
This PR includes:
- new field ipVersion for resource google_compute_address for reserving private ipv6 addresses (possible values IPV4 and IPV6)
- new acceptance test for ipv6 on google_compute_address resource
- new field ipv6Address in networkInterface for google_compute_instance resource to assign static ipv6 prefix address (new field support both ip address and reference to existing google_compute_address ipv6 resource).
- update to ipv6 address, although being implemented, seems not to work due to issue in the patch API for network interface (tested with API and got same behaviour), I will open a ticket on the API to check for that

If this PR is for Terraform, I acknowledge that I have:

- [ ] Searched through the [issue tracker](https://github.com/hashicorp/terraform-provider-google/issues) for an open issue that this either resolves or contributes to, commented on it to claim it, and written "fixes {url}" or "part of {url}" in this PR description. If there were no relevant open issues, I opened one and commented that I would like to work on it (not necessary for very small changes).
- [ ] Ensured that all new fields I added that can be set by a user appear in at least one [example](https://github.com/GoogleCloudPlatform/magic-modules/tree/main/mmv1/templates/terraform/examples) (for generated resources) or [third_party test](https://github.com/GoogleCloudPlatform/magic-modules/tree/main/mmv1/third_party/terraform/tests) (for handwritten resources or update tests).
- [X] [Generated Terraform providers](https://github.com/GoogleCloudPlatform/magic-modules/blob/main/docs/content/docs/getting-started/generate-providers.md), and ran [`make test` and `make lint`](https://googlecloudplatform.github.io/magic-modules/docs/getting-started/run-provider-tests/#run-unit-tests) in the generated providers to ensure it passes unit and linter tests.
- [X] [Ran](https://github.com/GoogleCloudPlatform/magic-modules/blob/main/docs/content/docs/getting-started/run-provider-tests.md) relevant acceptance tests using my own Google Cloud project and credentials (If the acceptance tests do not yet pass or you are unable to run them, please let your reviewer know).
- [X] Read [Write release notes](https://googlecloudplatform.github.io/magic-modules/contribute/release-notes/) before writing my release note below.

**Release Note Template for Downstream PRs (will be copied)**

```release-note:enhancement
compute: added `ipVersion` field to `google_compute_address` resource
compute: added `ipv6Address` field to `google_compute_instance` resource
```

PR #13133, opened 2025-02-20
fix: compute instance internal ipv6 prefix length is not set in creation and read in syncing

jingjia-peng:
Fixes [hashicorp/terraform-provider-google#21520](https://github.com/hashicorp/terraform-provider-google/issues/21520).

This ensures that Terraform correctly reads the `compute_instance.network_interface.internal_ipv6_prefix_length` from GCP and stores it in `terraform.state`. Previously, the absence of this value in `terraform.state` caused `compute_instance` to be considered modified, even when no changes were made to the configuration.

```release-note:bug
compute: fix `compute_instance.network_interface.internal_ipv6_prefix_length` not being set or read in Terraform state  
```


roaks3 (commented):
Thanks for the fix. It looks like at least part of this was added with https://github.com/GoogleCloudPlatform/magic-modules/pull/8432, and for some reason this field was left out of the flattener/expander. It seems like there may have even been cases where the value of this field wasn't being submitted to the server as expected.

Will run tests and approve once they pass.

roaks3 (approved):

PR #13011, opened 2025-02-08
EphemeralWriteOnly: add `password_wo` field to `google_sql_user`

BBBmau:
**Release Note Template for Downstream PRs (will be copied)**

See [Write release notes](https://googlecloudplatform.github.io/magic-modules/contribute/release-notes/) for guidance.

```release-note:enhancement
sql: added `password_wo` and `password_wo_version` fields to `google_sql_user` resource
```

PR #13135, opened 2025-02-20
feature: ephemeral write-only support

BBBmau:
This PR adds support for Write-Only attributes with the release of TF 1.11

It includes the following attributes:
- https://github.com/GoogleCloudPlatform/magic-modules/pull/13011
- https://github.com/GoogleCloudPlatform/magic-modules/pull/12967
- https://github.com/GoogleCloudPlatform/magic-modules/pull/12800

Some PRs that were necessary to support Write-Only attributes:
- https://github.com/GoogleCloudPlatform/magic-modules/pull/13048
- https://github.com/GoogleCloudPlatform/magic-modules/pull/12974
- https://github.com/GoogleCloudPlatform/magic-modules/pull/12550

```release-note:enhancement
sql: added `password_wo` and `password_wo_version` fields to `google_sql_user` resource
```

```release-note:enhancement
bigquerydatatransfer: added `secret_access_key_wo` write-only field to `google_bigquery_data_transfer_config` resource
```

```release-note:enhancement
secretmanager: added `secret_version_wo` write-only field to `google_secret_version` resource
```


mmv1/third_party/terraform/services/sql/resource_sql_user_test.go
```diff
@@ -55,6 +55,41 @@ func TestAccSqlUser_mysql(t *testing.T) {
 	})
 }
 
+func TestAccSqlUser_password_wo(t *testing.T) {
+	// Multiple fine-grained resources
+	acctest.SkipIfVcr(t)
```
rileykarson, line 60:
```suggestion
```

This is not the case for the test as written

mmv1/products/bigquerydatatransfer/Config.yaml
```diff
@@ -226,6 +233,19 @@ properties:
...
           The Secret Access Key of the AWS account transferring data from.
-
-        required: true
         sensitive: true
+        at_least_one_of:
+          - 'sensitive_params.0.secretAccessKey'
+          - 'sensitive_params.0.secretAccessKeyWo'
+        conflicts:
+          - 'sensitive_params.0.secretAccessKeyWo'
+      - name: 'secretAccessKeyWo' # Wo is convention for write-only properties
```
rileykarson, line 242: Can we add a test that exercises this field?

BBBmau:
TeamCity Builds testing the resources with Write-Only attributes:

[`Bigquerydatatransfer`](https://hashicorp.teamcity.com/buildConfiguration/TerraformProviders_GoogleCloud_EPHEMERALWRITEONLYGA__MM_GOOGLE_PACKAGE_BIGQUERYDATATRANSFER?branch=refs%2Fheads%2Fauto-pr-13135&buildTypeTab=overview#all-projects)
[`Sql`](https://hashicorp.teamcity.com/buildConfiguration/TerraformProviders_GoogleCloud_EPHEMERALWRITEONLYGA__MM_GOOGLE_PACKAGE_SQL?branch=refs%2Fheads%2Fauto-pr-13135&buildTypeTab=overview#all-projects)
[`secretManager`](https://hashicorp.teamcity.com/buildConfiguration/TerraformProviders_GoogleCloud_EPHEMERALWRITEONLYGA__MM_GOOGLE_PACKAGE_SECRETMANAGER?branch=refs%2Fheads%2Fauto-pr-13135&buildTypeTab=overview#all-projects)

rileykarson (commented):
Going through failing tests for now, I'll make a full run through the code tomorrow. The BQ change and a docs change didn't go through earlier dedicated review, and not sure if a full review was the expectation on the feature branch merge PR.

Breakdown by test:

- [x] TestAccDataSourceGoogleGkeHubFeature_basic seems unrelated. I'm not entirely sure what's going wrong, but it's VCR mechanics and doesn't seem to have been touched here.
- [x] TestAccSpannerInstance_spannerInstanceWithAutoscaling has a non-deterministic source of randomisation. I'll fix that. **edit**: https://github.com/GoogleCloudPlatform/magic-modules/pull/13141
- [x] TestAccAccessContextManager is likely failing due to messiness/conflicts in the environment.
- [x] TestAccCloudbuildWorkerPool_basic is [failing in nightlies](https://github.com/hashicorp/terraform-provider-google/issues/21106)
- [ ] TestAccEphemeralServiceAccountKey_basic seems like it may be related
- [x] TestAccSecretManagerSecretVersion_secretVersionBasicWriteOnlyExample can't pass yet in VCR but did in TC
- [x] TestAccSecretManagerSecretVersion_secretVersionWithBase64StringSecretDataWriteOnlyExample can't pass yet in VCR but did in TC
- [x] TestAccSqlUser_password_wo was added but won't pass yet in VCR but did in TC

@BBBmau can you link successful runs for the WO tests, and investigate and provide context on or fix the `TestAccEphemeralServiceAccountKey_basic` failure? **edit** ah, you just beat me with most of them. Thanks!

See comment inline on `bigquerydatatransfer`, also, a test/test(s) should be added.

rileykarson:
> [Bigquerydatatransfer](https://hashicorp.teamcity.com/buildConfiguration/TerraformProviders_GoogleCloud_EPHEMERALWRITEONLYGA__MM_GOOGLE_PACKAGE_BIGQUERYDATATRANSFER?branch=refs%2Fheads%2Fauto-pr-13135&buildTypeTab=overview#all-projects)

No tests ran in that one, see comment above!

mmv1/products/bigquerydatatransfer/Config.yaml
BBBmau: when first  attempting to add a test in for this in the separate PR i noticed that we don't include a test for the regular `secret_access_key` field. 

I had attempted to add one myself anyways when working on the `secret_access_key` PR early on but found that the reason for not including a test is due to it being a params for AWS. (requiring us to mark the `data_source` as `amazon_s3`

We would have to include valid params for `data_source=amazon_s3` such a s3 uri, destination_table_name, file_format, and more in order to create a good test for both `secret_access_key` and `secret_access_key_wo`.

fwiw I'm able to successful run `terraform apply` when applying these values, though this was just for my own testing. They are not valid values for `amazon_s3`. This is to show that the field `secret_access_key_wo` is being used when creating `google_bigquery_data_transfer_config`.

```hcl
  10   │ resource "google_bigquery_data_transfer_config" "query_config" {
  11   │   depends_on = [google_project_iam_member.permissions]
  12   │
  13   │   display_name           = "my-query"
  14   │   location               = "asia-northeast1"
  15   │   schedule               = "first sunday of quarter 00:00"
  16   │   destination_dataset_id = google_bigquery_dataset.my_dataset.dataset_id
  17   │  data_source_id = "amazon_s3"
  18   │
  19   │   sensitive_params {
  20   │   secret_access_key_wo = "aB3dEfgH1IjK2LmN4OpQrStUvWxYz+/5678ABCD="
  21   │   }
  22   │
  23   │   params = {
  24   │     file_format = "CSV"
  25   │     data_path = "s3://test"
  26   │     access_key_id = "ASDJFOAFOJAIFOAJDOFI"
  27   │     destination_table_name_template = "my_table"
  28   │     write_disposition               = "WRITE_APPEND"
  29   │   }
  30   │ }
```

Where I was attempting this can be found here: https://github.com/GoogleCloudPlatform/magic-modules/pull/12967#issuecomment-2640808926

I did uncover a diff within secret_access_key, it's been addressed in the latest commit.

rileykarson: Thanks!

mmv1/third_party/terraform/website/docs/guides/using_write_only_attributes.html.markdown
```diff
@@ -0,0 +1,156 @@
...
+---
+
+# Write-only attributes in the Google Cloud provider
+
+Write-only attributes allow users to access and use data in their configurations without that data being stored in Terraform state.
+
+Write-only attributes are available in Terraform v1.11 and later. For more information, see the [official HashiCorp documentation for Write-only Attributes](https://developer.hashicorp.com/terraform/plugin/sdkv2/resources/write-only-arguments).
+
+To mark the launch of write-only attributes, the Google Cloud provider has added the following write-only attributes:
+- [`google_compute_disk: disk_encryption_key.raw_key_wo`](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/compute_disk.html#raw_key_wo)
```
rileykarson, line 14: Remove disk entries here

mmv1/third_party/terraform/website/docs/guides/using_write_only_attributes.html.markdown
```diff
@@ -0,0 +1,156 @@
...
+  How to use write-only attributes in the Google Cloud provider
+---
+
+# Write-only attributes in the Google Cloud provider
+
+Write-only attributes allow users to access and use data in their configurations without that data being stored in Terraform state.
+
+Write-only attributes are available in Terraform v1.11 and later. For more information, see the [official HashiCorp documentation for Write-only Attributes](https://developer.hashicorp.com/terraform/plugin/sdkv2/resources/write-only-arguments).
+
+To mark the launch of write-only attributes, the Google Cloud provider has added the following write-only attributes:
```
rileykarson, line 13: Can this text be made evergreen? This reads as temporary- like an announcement blog post- but will be used by users as evergreen content.

mmv1/third_party/terraform/website/docs/guides/using_write_only_attributes.html.markdown
```diff
@@ -0,0 +1,156 @@
...
+
+These were chosen due to them being marked as sensitive already in the provider. Although sensitive attributes do not appear in `terraform plan`, they are still stored in the Terraform state. Write-only attributes allow users to access and use data in their configurations without that data being stored in Terraform state.
+
+## Use the Google Cloud provider's new write-only attributes
+
+The following sections show how to use the new write-only attributes in the Google Cloud provider.
+
+### Applying a write-only attribute
+
+The following example shows how to apply a write-only attribute. All write-only attributes are marked with the `wo` suffix and can not be used with the `sensitive` attribute.
```
rileykarson, line 29: This should probably be more clear- what sensitive attribute?

mmv1/third_party/terraform/website/docs/guides/using_write_only_attributes.html.markdown
```diff
@@ -0,0 +1,156 @@
...
+
+resource "google_secret_manager_secret_version" "secret-version-basic" {
+  secret = google_secret_manager_secret.secret-basic.id
+  secret_data_wo = "new-secret-data"
+  secret_data_wo_version = 2
+  enabled = true
+}
+```
+
+A `terraform apply` of this configuration will allow you to update the write-only attribute as it will destroy and recreate the resource.
```
rileykarson, line 135: I would recommend using a different resource as your example, one that can be updated. This teaches a first-time reader that updating a WO attribute destroys their resource when it only does if the corresponding sensitive value was immutable!

mmv1/templates/terraform/resource.html.markdown.tmpl
```diff
@@ -58,6 +58,11 @@ To get more information about {{$.Name}}, see:
 values will be stored in the raw state as plain text: {{ $.SensitivePropsToString }}.
 [Read more about sensitive data in state](https://www.terraform.io/language/state/sensitive-data).
 {{ end }}
+{{- if $.WriteOnlyProps }}
+~> **Warning:** All arguments including the following potentially write-only
+values will be stored in the raw state as plain text: {{ $.WriteOnlyPropsToString }}.
```
rileykarson, line 63: Isn't the whole point that they aren't stored in state? What am I missing here?

mmv1/products/bigquerydatatransfer/Config.yaml
```diff
@@ -210,6 +211,12 @@ properties:
       **NOTE** : If you are attempting to update a parameter that cannot be updated (due to api limitations) [please force recreation of the resource](https://www.terraform.io/cli/state/taint#forcing-re-creation-of-resources).
     required: true
     custom_flatten: 'templates/terraform/custom_flatten/json_to_string_map.go.tmpl'
+
+  - name: 'sensitiveParamsWoVersion'
```
rileykarson, line 215: Shouldn't this be a nested field parallel to the WO fied? Is there an explanation for why it isn't and can't be?

mmv1/api/resource/docs.go
```diff
@@ -32,5 +32,7 @@ type Docs struct {
 
 	OptionalProperties string `yaml:"optional_properties"`
 
+	WriteOnlyProperties string `yaml:"write_only_properties"`
```
rileykarson, line 35: Please give yourself a followup to add guidance on _adding_ WO fields to the contributor guide.

mmv1/products/bigquerydatatransfer/Config.yaml
rileykarson, line 242: Will WO fields always be 1:1 with sensitive fields? It seems error prone to require separate MM definitions if so, particularly where the version sidecar also needs to be configured by hand.

Before adding new WO fields I think we should:
* Figure out a plan for what support looks like long-term. Is this an opt-in option for sensitive fields, a default option for sensitive fields, can resources be added with WO variants and **not** sensitive variants or should sensitive always be required / be required until our min version is 1.11, etc.
* Modify the definition and template so that following the plan is obvious. Look at `KeyValueLabels` as an example, where it is structurally impossible to get that (complicated!) set of fields wrong because one type in MM adds multiple fields to Terraform.

BBBmau:
@rileykarson when looking into the `TestAccEphemeralServiceAccountKey_basic` tests look to be passing locally

```hcl
󰀵 mau  …/terraform-provider-google   main $!?⇣   v1.23.4   11:31   envchain GCLOUD make testacc TEST=./google/services/resourcemanager TESTARGS='-run=TestAccEphemeralServiceAccountKey_basic'     
TF_ACC=1 TF_SCHEMA_PANIC_ON_ERROR=1 go test ./google/services/resourcemanager -v -run=TestAccEphemeralServiceAccountKey_basic -timeout 240m -ldflags="-X=github.com/hashicorp/terraform-provider-google/version.ProviderVersion=acc"
=== RUN   TestAccEphemeralServiceAccountKey_basic
=== PAUSE TestAccEphemeralServiceAccountKey_basic
=== CONT  TestAccEphemeralServiceAccountKey_basic
2025/02/21 11:32:07 [INFO] Authenticating using configured Google JSON 'credentials'...
2025/02/21 11:32:07 [INFO]   -- Scopes: [https://www.googleapis.com/auth/cloud-platform https://www.googleapis.com/auth/userinfo.email]
2025/02/21 11:32:07 [INFO] Authenticating using configured Google JSON 'credentials'...
2025/02/21 11:32:07 [INFO]   -- Scopes: [https://www.googleapis.com/auth/cloud-platform https://www.googleapis.com/auth/userinfo.email]
2025/02/21 11:32:07 [DEBUG] Waiting for state to become: [success]
2025/02/21 11:32:07 [INFO] Terraform is using this identity: mauricio-alvarezleon@hc-terraform-testing.iam.gserviceaccount.com
2025/02/21 11:32:07 [DEBUG] Verifying projects/hc-terraform-testing/serviceAccounts/tf-bootstrap-sa-key-basic@hc-terraform-testing.iam.gserviceaccount.com as bootstrapped service account.
2025/02/21 11:32:07 [INFO] Instantiating Google Cloud IAM client for path https://iam.googleapis.com/
2025/02/21 11:32:07 [DEBUG] Retry Transport: starting RoundTrip retry loop
2025/02/21 11:32:07 [DEBUG] Retry Transport: request attempt 0
2025/02/21 11:32:08 [DEBUG] Retry Transport: Stopping retries, last request was successful
2025/02/21 11:32:08 [DEBUG] Retry Transport: Returning after 1 attempts
2025/02/21 11:32:08 [DEBUG] Setting service account permissions.
2025/02/21 11:32:08 [INFO] Instantiating Google Cloud IAM client for path https://iam.googleapis.com/
2025/02/21 11:32:08 [DEBUG] Retry Transport: starting RoundTrip retry loop
2025/02/21 11:32:08 [DEBUG] Retry Transport: request attempt 0
2025/02/21 11:32:08 [DEBUG] Retry Transport: Stopping retries, last request was successful
2025/02/21 11:32:08 [DEBUG] Retry Transport: Returning after 1 attempts
--- PASS: TestAccEphemeralServiceAccountKey_basic (29.34s)
PASS
ok      github.com/hashicorp/terraform-provider-google/google/services/resourcemanager  30.877s
```

When looking at the error itself in TC the `precondition` failure comes from the `google_service_account_key` resource itself and not from the ephemeral resource.

[ephemeral `google_service_account_key` TC Fail](https://hashicorp.teamcity.com/buildConfiguration/TerraformProviders_GoogleCloud_GOOGLE_MMUPSTREAMTESTS_GOOGLE_PACKAGE_RESOURCEMANAGER/326832?buildTab=tests&expandedTest=build%3A%28id%3A326832%29%2Cid%3A2000000000)

The first step where the error comes from is using a config that only has resource `google_service_account_key` for initial setup:
```hcl
func testAccEphemeralServiceAccountKey_setup(serviceAccount string) string {
	return fmt.Sprintf(`
resource "google_service_account_key" "key" {
  service_account_id = "%s"
  public_key_type = "TYPE_X509_PEM_FILE"
}
`, serviceAccount)
}
```

I looked into whether we see a similar failure with just the resource itself and found no issues: It also uses the same configuration:
```hcl
func testAccServiceAccountKey(account, name string) string {
	return fmt.Sprintf(`
resource "google_service_account" "acceptance" {
  account_id   = "%s"
  display_name = "%s"
}

resource "google_service_account_key" "acceptance" {
  service_account_id = google_service_account.acceptance.name
  public_key_type    = "TYPE_X509_PEM_FILE"
}
`, account, name)
}
```
[resource `google_service_account_key` TC Pass](https://hashicorp.teamcity.com/buildConfiguration/TerraformProviders_GoogleCloud_GOOGLE_MMUPSTREAMTESTS_GOOGLE_PACKAGE_RESOURCEMANAGER/326840?buildTab=tests&expandedTest=build%3A%28id%3A326840%29%2Cid%3A2000000000)

* note: the process of setting up a serviceAccount for tests is the same across all ephemeral resources 

so far it's not exactly clear what could be causing the ephemeral test to fail when:
1) write-only attributes do not interact with `resourcemanager` product
2) the ephemeral test fails due to the resource google_service_account_key
3) the tests on the resource google_service_account_key itself show passing
4) attempts at running locally show passing on both ephemeral and non-ephemeral resources.
5) [nightly-tests](https://hashicorp.teamcity.com/test/-9103025145046913377?currentProjectId=TerraformProviders_GoogleCloud_GOOGLE_NIGHTLYTESTS) show consistent passing tests on the ephemeral test itself

mmv1/third_party/terraform/website/docs/r/sql_user.html.markdown
```diff
@@ -155,6 +157,15 @@ The read only `password_policy.status` subblock supports:
 
 * `password_expiration_time` - (read only) Password expiration duration with one week grace period.
 
+## Ephemeral Attributes Reference
+
+The following write-only attributes are supported:
+
+* `password_wo` -
```
rileykarson, line 164: `password` has considerably more documentation. Should more be added here, to make it more similar?

mmv1/third_party/terraform/website/docs/r/sql_user.html.markdown
```diff
@@ -130,6 +130,8 @@ The following arguments are supported:
 
     Possible values are: `ABANDON`.
 
+* `password_wo_version` - (Optional) The version of the password_wo.
```
rileykarson, line 133: The WO field docs are all fairly terse. Should we link out to guides explaining the version convention and how to update WO values?

mmv1/third_party/terraform/website/docs/guides/using_write_only_attributes.html.markdown
BBBmau: removed

mmv1/third_party/terraform/website/docs/guides/using_write_only_attributes.html.markdown
BBBmau: recent commit made it more clear the different between write-only and non writeonly

mmv1/third_party/terraform/website/docs/guides/using_write_only_attributes.html.markdown
BBBmau: used `password_wo` to showcase updating the write-only where the resource isn't immutable.

mmv1/templates/terraform/resource.html.markdown.tmpl
BBBmau: this is wrong and has been addressed in the latest commit.

mmv1/products/bigquerydatatransfer/Config.yaml
BBBmau: this is due to how sensitiveParams works

sensitive_params is meant to take in fields for `params`. The api can only take valid params in regards to the `data_source` being used. although it may have been an option to include it in params and remove it before sending the request I felt that it made more sense to leave the values within params to be what's sent and have fields that are not sent outside of the params object.

mmv1/third_party/terraform/website/docs/r/sql_user.html.markdown
BBBmau: description contains link to updating-write-only section now

mmv1/third_party/terraform/website/docs/r/sql_user.html.markdown
BBBmau: updated to match `password` doc

mmv1/products/bigquerydatatransfer/Config.yaml
BBBmau: >Figure out a plan for what support looks like long-term. Is this an opt-in option for sensitive fields, a default option for sensitive fields, can resources be added with WO variants and not sensitive variants or should sensitive always be required / be required until our min version is 1.11, etc.

Long term we would want `WriteOnly` to replace `Sensitive` entirely but until the min version is 1.11 we'll want to include both sensitive and writeOnly field support. There may be a case where a field that wasn't previously sensitive could be a WriteOnly field candidate though there hasn't been an example of this.

>Modify the definition and template so that following the plan is obvious. Look at KeyValueLabels as an example, where it is structurally impossible to get that (complicated!) set of fields wrong because one type in MM adds multiple fields to Terraform.

it would've been nice to modify the sensitive template to include the writeOnly field in order to prevent future errors. (i.e setting `sensitive: true` would create both the normal field and _wo field + the version sidecar field) though we run into cases where if the field is nested it may need to be added outside of the nestedObject itself. (what comes to mind is `sensitiveParams`)

mmv1/third_party/terraform/website/docs/r/sql_user.html.markdown
```diff
@@ -130,6 +130,8 @@ The following arguments are supported:
 
     Possible values are: `ABANDON`.
 
+* `password_wo_version` - (Optional) The version of the password_wo. For more info see [updating write-only attributes](https://registry.terraform.io/providers/hashicorp/google/6.15.0/docs/guides/using_write_only_attributes#updating-write-only-attributes).
```
rileykarson, line 133: We should avoid version pins, can you use a relative link?

mmv1/products/bigquerydatatransfer/Config.yaml
```diff
@@ -210,6 +211,12 @@ properties:
...
+    type: Integer
+    immutable: true
+    description: |
+      The version of the sensitive params - used to trigger updates of the write-only params. For more info see [updating write-only attributes](https://registry.terraform.io/providers/hashicorp/google/6.15.0/docs/guides/using_write_only_attributes#updating-write-only-attributes)
```
rileykarson, line 219: use relative link

mmv1/products/secretmanager/SecretVersion.yaml
```diff
@@ -130,13 +145,29 @@ properties:
...
         api_name: data
-        required: true
+        conflicts:
+          - 'secretDataWo'
         immutable: true
         sensitive: true
-        custom_expand: 'templates/terraform/custom_expand/secret_version_secret_data.go.tmpl'
+      - name: 'secretDataWo'
+        type: String
+        description: The secret data. Must be no larger than 64KiB. For more info see [updating write-only attributes](https://registry.terraform.io/providers/hashicorp/google/6.15.0/docs/guides/using_write_only_attributes#updating-write-only-attributes)
```
rileykarson, line 161: note: use relative link

mmv1/products/secretmanager/SecretVersion.yaml
```diff
@@ -130,13 +145,29 @@ properties:
...
+        required_with:
+          - 'SecretDataWoVersion'
+        conflicts:
+          - 'payload.0.secretData'
+        write_only: true
+      - name: 'SecretDataWoVersion'
+        type: Integer
+        default_value: 0
+        ignore_read: true
+        description: Triggers update of secret data write-only. For more info see [updating write-only attributes](https://registry.terraform.io/providers/hashicorp/google/6.15.0/docs/guides/using_write_only_attributes#updating-write-only-attributes)
```
rileykarson, line 172: note: use relative link

mmv1/templates/terraform/decoders/bigquery_data_transfer.go.tmpl
```diff
@@ -14,11 +14,7 @@ if paramMap, ok := res["params"]; ok {
 	params := paramMap.(map[string]interface{})
 	for _, sp := range sensitiveParams {
 		if _, apiOk := params[sp]; apiOk {
-			if _, exists := d.GetOkExists("sensitive_params.0." + sp); exists {
-				delete(params, sp)
-			} else {
-				params[sp] = d.Get("params." + sp)
```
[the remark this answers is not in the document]
BBBmau: you're correct, the regression comes from if a user decides to explicitly set a `secret_access_key` into params directly instead of `sensitive_params`. If a user were to do this with the current change it would delete from state (regression).

Latest commit reverts this and adds a config check of `_wo` attribute with `RawConfigAt` https://github.com/GoogleCloudPlatform/magic-modules/pull/13135/commits/d5b47a446b6bc372d99a3751702153f31d8f3a9a

mmv1/products/bigquerydatatransfer/Config.yaml
BBBmau: sounds good, i made the change that moves it into sensitive_params while also adding in a check that requires both _wo_version and _wo to be set. https://github.com/GoogleCloudPlatform/magic-modules/pull/13135/commits/3eb3cab85be0b553c41f81a69e7a444c5b6e8ae9

RequiredWith doesn't appear to work.

mmv1/templates/terraform/decoders/bigquery_data_transfer.go.tmpl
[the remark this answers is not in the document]
BBBmau: >There might be something wrong with how RawConfigAt works. It only works on creation. Was getting null as the result. Enforcing the check with version fixes the diff just fine but feels janky. I'll get back to you on whether this is a bug or if this is meant to be like this, just need to confirm. The new commit: https://github.com/GoogleCloudPlatform/magic-modules/commit/113b096acf08e18ac733087f7fe7f50d84e4ccb8

After reaching out, the Read in providers is intended to return values as null for Write-Only. The `non-null` value on Read is being carried over from Create hence why on creation the check from the initial commit mentioned in this thread worked just fine. A check  with `secret_key_wo_version` is more suitable and will prevent the diff.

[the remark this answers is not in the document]
rileykarson: Thanks!

mmv1/templates/terraform/nested_property_documentation.html.markdown.tmpl
```diff
@@ -1,10 +1,10 @@
 {{ "" }}
-{{- if $.FlattenObject }}
+{{- if and $.FlattenObject (not $.WriteOnlyProperties) }}
   {{- range $np := $.NestedProperties }}
 {{- trimTemplate "nested_property_documentation.html.markdown.tmpl" $np -}}
   {{- end -}}
-{{- else if $.NestedProperties }}
-<a name="nested_{{$.LineageAsSnakeCase}}"></a>The `{{ underscore $.Name }}` block {{ if $.Output }}contains{{ else }}supports{{ end }}:
```
rileykarson, line 7: The change from `$.LineageAsSnakeCase` to `underscore $.Name` breaks many block links where those values are different. What was the intent of the change?

If we are changing this line we may want to break it into a dedicated change- that'll be easier to review, and will cut down tremendously on the changed files here.

mmv1/templates/terraform/custom_flatten/secret_version_access.go.tmpl
```diff
@@ -12,6 +12,9 @@
 */ -}}
 func flatten{{$.GetPrefix}}{{$.TitlelizeProperty}}(v interface{}, d *schema.ResourceData, config *transport_tpg.Config) interface{} {
 	transformed := make(map[string]interface{})
+	if d.Get("secret_data").(string) == "" {
```
rileykarson, line 15: Are we able to do an existence check on the wo field instead of a non-existence check on the sensitive field? (I am almost certain either another reviewer or myself have asked as much- the explanation has not stuck for me, so asking again to have it reexplained again!)

also nit: Can you comment here to explain why we're not checking the value? Similar to L16/L19 (left/right side line) below.

mmv1/products/secretmanager/SecretVersion.yaml
```diff
@@ -130,13 +145,29 @@ properties:
...
+      - name: 'secretDataWo'
+        type: String
+        description: The secret data. Must be no larger than 64KiB. For more info see [updating write-only attributes](/docs/providers/google/guides/using_write_only_attributes.html#updating-write-only-attributes)
+        api_name: data
+        required_with:
+          - 'SecretDataWoVersion'
+        conflicts:
+          - 'payload.0.secretData'
+        write_only: true
+      - name: 'SecretDataWoVersion'
```
rileykarson, line 168: These should all be `url_param_only: true`, right? This one is `ignore_read: true`

mmv1/products/bigquerydatatransfer/Config.yaml
```diff
@@ -222,10 +223,28 @@ properties:
       to a different credential configuration in the config will require an apply to update state.
     url_param_only: true
     properties:
+      - name: 'secretAccessKeyWoVersion'
```
rileykarson, line 226: `url_param_only: true` probably, based on other discussion

mmv1/products/bigquerydatatransfer/Config.yaml
```diff
@@ -222,10 +223,28 @@ properties:
...
+        type: Integer
+        immutable: true
```
rileykarson, line 228: This is wrong, right? Otherwise the WO variant is immutable while the sensitive one isn't.

mmv1/third_party/terraform/services/sql/resource_sql_user_test.go
```diff
@@ -55,6 +55,39 @@ func TestAccSqlUser_mysql(t *testing.T) {
...
+		CheckDestroy:             testAccSqlUserDestroyProducer(t),
+		Steps: []resource.TestStep{
+			{
+				Config: testGoogleSqlUser_password_wo(instance, "password"),
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckGoogleSqlUserExists(t, "google_sql_user.user1"),
+				),
+			},
+			{
+				// Update password
```
rileykarson, line 74: here & other update checks, use the update plancheck: https://googlecloudplatform.github.io/magic-modules/test/test/#add-an-update-test

mmv1/templates/terraform/nested_property_documentation.html.markdown.tmpl
BBBmau: There must've been a change since opening the PR to now because I couldn't find any reason to do this. 

The original PR that adds the [initial changes to the MMv1 generation can be found here.](https://github.com/GoogleCloudPlatform/magic-modules/pull/12550/files#diff-0ee44b6dfb99752bf6dae3214ef04057cc0443041b6671cc6a1e3eb3355b9403) So it looks like it was changed but after that PR was merged. I've updated this here: https://github.com/GoogleCloudPlatform/magic-modules/pull/13135/commits/e7535e2d604a492bb33f8bade7013e445fbab02e

mmv1/templates/terraform/custom_flatten/secret_version_access.go.tmpl
BBBmau: we aren't able to do this since on `ReadResource` the RawConfigAt of a write-only attribute will always return Null. So the check is for the sensitive field itself. We can change this to check for `secret_data_wo_version` instead since that must be set with write-only attribute.

mmv1/products/bigquerydatatransfer/Config.yaml
BBBmau: yeah this is wrong and addressed here: https://github.com/GoogleCloudPlatform/magic-modules/pull/13135/commits/31f7b4ee0e49430c1d8f26bf8f4c1ec3bd23b91c

mmv1/products/bigquerydatatransfer/Config.yaml
BBBmau: https://github.com/GoogleCloudPlatform/magic-modules/pull/13135/commits/31f7b4ee0e49430c1d8f26bf8f4c1ec3bd23b91c

mmv1/products/secretmanager/SecretVersion.yaml
BBBmau: https://github.com/GoogleCloudPlatform/magic-modules/pull/13135/commits/6c9d5245c5f2e95090876fa316d90557c7fd994c

mmv1/templates/terraform/custom_flatten/secret_version_access.go.tmpl
BBBmau: https://github.com/GoogleCloudPlatform/magic-modules/pull/13135/commits/eb7578b6f56aeadd75fc89a468aa0aed189711a5

mmv1/third_party/terraform/services/sql/resource_sql_user_test.go
BBBmau: https://github.com/GoogleCloudPlatform/magic-modules/pull/13135/commits/eab0a5b48adcd42cd1ffd99b0c7dd85746b50610

rileykarson (approved):
LGTM, thanks!

PR #13146, opened 2025-02-21
Add information about failing member into the error message

wiktorn:
When working with dynamically generated list of IAM resources, the error message is not that useful:
```
Error: invalid value for member (IAM members must have one of the values outlined here: https://cloud.google.com/billing/docs/reference/rest/v1/Policy#Binding)

  with module.projects-iam["service1"].google_project_iam_member.shared_vpc_host_iam["rw"],
  on ../project/shared-vpc.tf line 131, in resource "google_project_iam_member" "shared_vpc_host_iam":
 131:   member     = each.value
```
It makes hard to actually identify offending value. Returning failing member value helps to narrow down, which value was passed incorrectly.

```release-note:note
iam: added member value to the error message when member validation fails for google_project_iam_*
```


wiktorn:
All 3 failures looks like unrelated to the change

melinath (approved):

melinath:
agreed - failing tests are unrelated. This seems like a straightforward improvement.

melinath:
Note: This changes all IAM resources, but doesn't need to modify all tests because most tests don't cover the error behavior. VCR tests all reran so there shouldn't be any new failures due to this.

PR #13147, opened 2025-02-21
Instance feature gap for disk related fields

karolgorc:
Reopen https://github.com/GoogleCloudPlatform/magic-modules/pull/12672

fixed the error with `guest_os_features` causing this to be breaking

```release-note:enhancement
compute: added several `boot_disk`, `attached_disk`, and `instance_encryption_key` fields for improved encryption key support in `google_compute_instance` and template resources
```

```release-note:enhancement
compute: added support for `image_encryption_key.raw_key` and ` image_encryption_key.rsa_encrypted_key` to `google_compute_image` resource
```

```release-note:enhancement
compute: added support for `snapshot_encryption_key.rsa_encrypted_key` to `google_compute_snapshot` resource
```


roaks3 (changes requested):
Given this was previously reverted and contains a large number of changes at once, I'd request that this be broken up so that we can be more confident in the review. Perhaps along resource lines, or by grouping similar fields together (like all of the *key fields)?

karolgorc:
>  I'd request that this be broken up so that we can be more confident in the review

I'll link these back to this PR and then close it. Need to check if order of implementing these can be breaking

karolgorc:
Found some dependencies and outlined them in the smaller PR's. Closing

PR #13150, opened 2025-02-21
Fix perma-diff in `google_container_cluster` when `cluster_dns_scope` is unspecified

Michcioperz:
Terraform defaults this field to DNS_SCOPE_UNSPECIFIED, but the API
corrects this to an empty value.

Following https://googlecloudplatform.github.io/magic-modules/develop/diffs/#default_if_empty
to tackle this issue.

To make writing DNS tests faster, I refactored the helper function `testAccContainerCluster_autopilot_withDNSConfig` I added previously into `testAccContainerCluster_withAdvancedDNSConfig`.

Fixes https://github.com/hashicorp/terraform-provider-google/issues/20285

```release-note:bug
container: fixed perma-diff in `google_container_cluster` when `cluster_dns_scope` is unspecified
```

### DIFF
--- a/mmv1/third_party/terraform/services/container/resource_container_cluster.go.tmpl
+++ b/mmv1/third_party/terraform/services/container/resource_container_cluster.go.tmpl
@@ -2206,8 +2206,7 @@ func ResourceContainerCluster() *schema.Resource {
 						},
 						"cluster_dns_scope": {
 							Type:         schema.TypeString,
-							Default:      "DNS_SCOPE_UNSPECIFIED",
-							ValidateFunc: validation.StringInSlice([]string{"DNS_SCOPE_UNSPECIFIED", "CLUSTER_SCOPE", "VPC_SCOPE"}, false),
+							ValidateFunc: validation.StringInSlice([]string{"CLUSTER_SCOPE", "VPC_SCOPE"}, false),
 							Description:  `The scope of access to cluster DNS records.`,
 							Optional:     true,
 						},
--- a/mmv1/third_party/terraform/services/container/resource_container_cluster.go.tmpl
+++ b/mmv1/third_party/terraform/services/container/resource_container_cluster.go.tmpl
@@ -6783,7 +6783,7 @@ func flattenDnsConfig(c *container.DNSConfig) []map[string]interface{} {
 		{
 			"additive_vpc_scope_dns_domain": 	c.AdditiveVpcScopeDnsDomain,
 			"cluster_dns":        				c.ClusterDns,
-			"cluster_dns_scope":  				flattenClusterDnsScope(c.ClusterDnsScope),
+			"cluster_dns_scope":  				c.ClusterDnsScope,
 			"cluster_dns_domain": 				c.ClusterDnsDomain,
 		},
 	}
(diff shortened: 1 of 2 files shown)


melinath (changes requested):
We usually don't include UNSPECIFIED values in enums because they're the equivalent of the empty string.

Changing default values or making validation stricter is a breaking change; however, given that users are experiencing a bug (this permadiff) I think it would fall under the [fixing a bug exemption](https://googlecloudplatform.github.io/magic-modules/breaking-changes/make-a-breaking-change/#in-minor-releases) and be allowed in a minor release.

In particular, I would recommend:

1. Remove the default (which would be the same as changing to `Default: ""`).
2. Remove `DNS_SCOPE_UNSPECIFIED` from the ValidateFunc for the field so that it is no longer an allowed value.

What would you think of that? Any objection?

Michcioperz:
> Remove DNS_SCOPE_UNSPECIFIED from the ValidateFunc for the field so that it is no longer an allowed value.

@melinath Ah, so the idea here would be that, since setting this value is always causing permadiff right now, it's okay if we disallow it and force people change it to empty? Is that the rationale more or less?

melinath:
(I've checked with other folks on the team and confirmed that this falls under the bug exemption.)

mmv1/third_party/terraform/services/container/resource_container_cluster.go.tmpl
```diff
@@ -6784,12 +6783,19 @@ func flattenDnsConfig(c *container.DNSConfig) []map[string]interface{} {
...
 	}
 }
 
+func flattenClusterDnsScope(s string) string {
+	if s == "" {
+		return "DNS_SCOPE_UNSPECIFIED"
+	}
+	return s
+}
+
```
melinath, line 6798: No longer needed

```suggestion
```

mmv1/third_party/terraform/services/container/resource_container_cluster_test.go.tmpl
```diff
@@ -6054,6 +6054,31 @@ func TestAccContainerCluster_autopilot_withAdditiveVPC(t *testing.T) {
...
+	acctest.VcrTest(t, resource.TestCase{
+		PreCheck:                 func() { acctest.AccTestPreCheck(t) },
+		ProtoV5ProviderFactories: acctest.ProtoV5ProviderFactories(t),
+		CheckDestroy:             testAccCheckContainerClusterDestroyProducer(t),
+		Steps: []resource.TestStep{
+			{
+				Config: testAccContainerCluster_withAdvancedDNSConfig(clusterName, false, true, true, false, ""),
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr("google_container_cluster.primary", "dns_config.0.cluster_dns_scope", ""),
+				),
```
melinath, line 6070: We generally don't do this kind of check - it (more or less) duplicates work done by the following ImportState step.

```suggestion
```

melinath (changes requested):

mmv1/third_party/terraform/services/container/resource_container_cluster.go.tmpl
Michcioperz: Ah, I thought I had reverted it. Thanks

melinath (approved):

PR #13151, opened 2025-02-21
Suppressing diff for database_version for MySQL 8.0

ankushkhanna1998:
Suppressing diff for database_version for MySQL 8.0. With introduction of auto version upgrade for MySQL 8.0.35 and beyond, customers will be auto upgraded to a newer version which will lead to change in database version.

Hence, to avoid breakage of terraform scripts suppressing this diff while maintaining the sanity that the major version does not change, suppression is done only if there is a diff in minor version.

```release-note:enhancement
sql: added `databaseVersionDiffSuppress` to `google_sql_database_instance`. This change will suppress diff for database_version for MySQL 8.0 when the version is updated by auto version upgrade.
```

### DIFF
--- a/mmv1/third_party/terraform/services/sql/resource_sql_database_instance.go.tmpl
+++ b/mmv1/third_party/terraform/services/sql/resource_sql_database_instance.go.tmpl
@@ -7,6 +7,7 @@ import (
 	"log"
 	"reflect"
 	"slices"
+	"strconv"
 	"strings"
 	"time"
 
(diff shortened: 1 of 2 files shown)


ankushkhanna1998:
Please help merge https://github.com/GoogleCloudPlatform/magic-modules/pull/13152 before this PR, to fix the flakiness of the test.

mmv1/third_party/terraform/services/sql/resource_sql_database_instance.go.tmpl
```diff
@@ -2148,6 +2149,16 @@ func maintenanceVersionDiffSuppress(_, old, new string, _ *schema.ResourceData)
 	}
 }
 
+func databaseVersionDiffSuppress(_, old, new string, _ *schema.ResourceData) bool {
+	// Ignore the minor version in database version diff for MySQL 8.0.
+	if strings.Contains(old, "MYSQL_8_0") && strings.Contains(new, "MYSQL_8_0") {
```
melinath, line 2154: Drive-by comment: This likely needs more complex logic to actually check the patch version. As written, it will prevent manual upgrades from (for example) MYSQL_8_0_18 to MYSQL_8_0_26.

melinath:

melinath:
Reassigning to myself since I've already taken a look at it

mmv1/third_party/terraform/services/sql/resource_sql_database_instance.go.tmpl
melinath, line 2154: This needs more complex logic to actually check the patch version. As written, it will prevent manual upgrades from (for example) MYSQL_8_0_18 to MYSQL_8_0_26.

mmv1/third_party/terraform/services/sql/resource_sql_database_instance_internal_test.go
```diff
@@ -36,3 +36,40 @@ func TestMaintenanceVersionDiffSuppress(t *testing.T) {
 		})
 	}
 }
+
+func TestDatabaseVersionDiffSuppress(t *testing.T) {
```
melinath, line 40: This needs additional tests for at least the following cases (shouldn't suppress unless stated otherwise):

- Upgrading from 8_0 non-auto-upgrade version to 8_4
- Upgrading from 8_0 non-auto-upgrade version to 8_0 auto-upgrade version
- Upgrading from 5.x to 8.4
- Upgrading from 8.0 auto-upgrade version to a different 8.0 auto-upgrade version (should suppress)
- Downgrading from 8.0 auto-upgrade version to a different 8.0 auto-upgrade version (should suppress)
- Downgrading from 8.4 to 8.0 auto-upgrade version
- Downgrading from 8.4 to 8.0 non-auto-upgrade version
- Downgrading from 8.4 to 5.x
- Downgrading from 8.0 auto-upgrade version to 8.0 non-auto-upgrade version

melinath (changes requested):

mmv1/third_party/terraform/services/sql/resource_sql_database_instance.go.tmpl
ankushkhanna1998: Done

mmv1/third_party/terraform/services/sql/resource_sql_database_instance_internal_test.go
melinath: Thanks!

PR #13156, opened 2025-02-21
Fix metadata for recent resource changes 20250221

roaks3:
2 recently added Gemini resources that follow an existing pattern

```release-note:none

```


rileykarson (approved):

rileykarson:
The default we assume is SettingsBindings?

roaks3:
Well, in the service config these all happen to be variants of SettingsBindings.

PR #13161, opened 2025-02-21
container: Fixed container nodepool tests

wyardley:
Update some container versions to either remove implicit references to old minimum versions, or set newer / available versions.
I don't think v 1.32 is quite the default version in stable yet, but I removed the sysctl TODOs, and they seem to work for that test (`TestAccContainerNodePool_withKubeletConfig`)?

I assume there's a reason `TestAccContainerNodePool_concurrent` has the versions hard-coded? I didn't 100% understand the comments around some of those, so I took the more conservative approach of bumping it, but if it would work / be preferred, I could configure it to use the latest stable version using a data resource instead, as many other tests do?

Fixes hashicorp/terraform-provider-google#21488
Fixes hashicorp/terraform-provider-google#21484
Fixes hashicorp/terraform-provider-google#21444

Ran the following locally, in both recording and replaying modes:
✅ `TestAccContainerNodePool_concurrent`
✅ `TestAccContainerNodePool_withKubeletConfig`
`TestAccContainerNodePool_fastSocket`
✅ `TestAccContainerNodePool_enableQueuedProvisioning`
✅ `TestAccContainerNodePool_secondaryBootDisks`

```release-note:none
```


mmv1/third_party/terraform/services/container/resource_container_node_pool_test.go.tmpl
```diff
@@ -3981,7 +3977,7 @@ resource "google_container_cluster" "cluster" {
   deletion_protection = false
   network             = "%s"
   subnetwork          = "%s"
-  min_master_version  = "1.32.0-gke.1448000"
+  min_master_version  = "1.32.1-gke.1357001"
```
wyardley, line 3980: as mentioned in PR description, just bumping this seems safest, but also obviously fragile. So if it would work to use a data resource and the stable channel (like [this](https://github.com/GoogleCloudPlatform/magic-modules/blob/ae1e116040cd10d4a34c376845c6c5525350adbc/mmv1/third_party/terraform/services/container/resource_container_cluster_test.go.tmpl#L875), or to omit the version entirely, let me know.

mmv1/third_party/terraform/services/container/resource_container_node_pool_test.go.tmpl
```diff
@@ -4013,15 +4009,15 @@ resource "google_container_cluster" "cluster" {
...
 }
 
 resource "google_container_node_pool" "np1" {
   name               = "%s"
   location           = "us-central1-a"
   cluster            = google_container_cluster.cluster.name
   initial_node_count = 2
-  version            = "1.32.0-gke.1448000"
+  version            = "1.32.1-gke.1357001"
```
wyardley, line 4020: I don't know if they're testing something specific with an explicit version here? normally, setting the node pool version to the same version as the control plane will Just Work that way anyway without being configured? I took a look at some of the PRs mentioned in the comments here, but still couldn't figure out what is going on.

zli82016:
The failed terraform-google-conversion test is unrelated. The fix is already merged into the main branch.

wyardley:
> The failed terraform-google-conversion test is unrelated. The fix is already merged into the main branch.

Thanks, let me know if you want me to rebase.

I'm still trying to run `TestAccContainerNodePool_fastSocket` locally - may be running into issues with quota and / or availability for the accelerator and / or the n1 instance types.

LMK if you have any thoughts about the hard-coded versions for the other test. Updating this should fix it, it's just kind of kicking the can down the road, though.

mmv1/third_party/terraform/services/container/resource_container_node_pool_test.go.tmpl
zli82016: I don't think they are testing something specific.

mmv1/third_party/terraform/services/container/resource_container_node_pool_test.go.tmpl
zli82016: I think the test will be more stable to use the data source. Do you minding trying to change it to the data source? Thanks.

wyardley: Yes, I can do that. I still can't figure out why the first step (`testAccContainerNodePool_concurrentCreate`) _doesn't_ have the version set on node pools, and then `testAccContainerNodePool_concurrentUpdate` does, though? Since it's supposed to "update" something, maybe the original idea is that, by having the _minimum_ version for the control plane be _higher_, that the update test should then be updating it to a _different_ (but close) version?

This would mean that just using the data source might not get it to work quite as easily... and yet, my earlier "fix" also wouldn't accomplish the intent of the test? But maybe we can do something like this, to create with "latest" version, and then update with "stable"? or is the change in the # of default nodes on one of the two nodes sufficient for the update test?

```diff
diff --git a/mmv1/third_party/terraform/services/container/resource_container_node_pool_test.go.tmpl b/mmv1/third_party/terraform/services/container/resource_container_node_pool_test.go.tmpl
index 399218c92..f859eba35 100644
--- a/mmv1/third_party/terraform/services/container/resource_container_node_pool_test.go.tmpl
+++ b/mmv1/third_party/terraform/services/container/resource_container_node_pool_test.go.tmpl
@@ -3970,6 +3970,10 @@ resource "google_container_node_pool" "np" {
 
 func testAccContainerNodePool_concurrentCreate(cluster, np1, np2, networkName, subnetworkName string) string {
 	return fmt.Sprintf(`
+data "google_container_engine_versions" "central1a" {
+  location = "us-central1-a"
+}
+
 resource "google_container_cluster" "cluster" {
   name                = "%s"
   location            = "us-central1-a"
@@ -3977,7 +3981,7 @@ resource "google_container_cluster" "cluster" {
   deletion_protection = false
   network             = "%s"
   subnetwork          = "%s"
-  min_master_version  = "1.32.1-gke.1357001"
+  min_master_version  = data.google_container_engine_versions.uscentral1a.release_channel_latest_version["STABLE"]
 }
 
 resource "google_container_node_pool" "np1" {
@@ -3985,8 +3989,6 @@ resource "google_container_node_pool" "np1" {
   location           = "us-central1-a"
   cluster            = google_container_cluster.cluster.name
   initial_node_count = 2
-  // 2025-02-03: current default cluster version is 1.31.5-gke.1023000. This will change over time.
-  // Reference:  https://cloud.google.com/kubernetes-engine/docs/release-notes#current_versions
 }
 
 resource "google_container_node_pool" "np2" {
@@ -3994,14 +3996,16 @@ resource "google_container_node_pool" "np2" {
   location           = "us-central1-a"
   cluster            = google_container_cluster.cluster.name
   initial_node_count = 2
-  // 2025-02-03: current default cluster version is 1.31.5-gke.1023000. This will change over time.
-  // Reference:  https://cloud.google.com/kubernetes-engine/docs/release-notes#current_versions
 }
 `, cluster, networkName, subnetworkName, np1, np2)
 }
 
 func testAccContainerNodePool_concurrentUpdate(cluster, np1, np2, networkName, subnetworkName string) string {
 	return fmt.Sprintf(`
+data "google_container_engine_versions" "central1a" {
+  location = "us-central1-a"
+}
+
 resource "google_container_cluster" "cluster" {
   name                = "%s"
   location            = "us-central1-a"
@@ -4009,7 +4013,7 @@ resource "google_container_cluster" "cluster" {
   deletion_protection = false
   network             = "%s"
   subnetwork          = "%s"
-  min_master_version  = "1.32.1-gke.1357001"
+  min_master_version  = data.google_container_engine_versions.uscentral1a.release_channel_latest_version["STABLE"]
 }
 
 resource "google_container_node_pool" "np1" {
@@ -4017,11 +4021,13 @@ resource "google_container_node_pool" "np1" {
   location           = "us-central1-a"
   cluster            = google_container_cluster.cluster.name
   initial_node_count = 2
-  version            = "1.32.1-gke.1357001"
-  // The node version must remain within one minor version of the cluster ("master") version, and it must not exceed the cluster ("master") version
-  // Cross-ref: https://github.com/hashicorp/terraform-provider-google/issues/21116
-  // Cross-ref: https://github.com/GoogleCloudPlatform/magic-modules/pull/11115
-  // Reference:  https://cloud.google.com/kubernetes-engine/docs/release-notes#current_versions
+  # Force an update by going to a different, but still compatible version.
+  version            = data.google_container_engine_versions.uscentral1a.release_channel_default_version["STABLE"]
+  # The node version must remain within one minor version of the cluster
+  # ("master") version, and it must not exceed the cluster ("master") version
+  # Cross-ref: https://github.com/hashicorp/terraform-provider-google/issues/21116
+  # Cross-ref: https://github.com/GoogleCloudPlatform/magic-modules/pull/11115
+  # Reference:  https://cloud.google.com/kubernetes-engine/docs/release-notes#current_versions
 }
 
 resource "google_container_node_pool" "np2" {
@@ -4030,10 +4036,11 @@ resource "google_container_node_pool" "np2" {
   cluster            = google_container_cluster.cluster.name
   initial_node_count = 2
   version            = "1.32.1-gke.1357001"
-  // The node version must remain within one minor version of the cluster ("master") version, and it must not exceed the cluster ("master") version
-  // Cross-ref: https://github.com/hashicorp/terraform-provider-google/issues/21116
-  // Cross-ref: https://github.com/GoogleCloudPlatform/magic-modules/pull/11115
-  // Reference:  https://cloud.google.com/kubernetes-engine/docs/release-notes#current_versions
+  # The node version must remain within one minor version of the cluster
+  # ("master") version, and it must not exceed the cluster ("master") version
+  # Cross-ref: https://github.com/hashicorp/terraform-provider-google/issues/21116
+  # Cross-ref: https://github.com/GoogleCloudPlatform/magic-modules/pull/11115
+  # Reference:  https://cloud.google.com/kubernetes-engine/docs/release-notes#current_versions
 }
 `, cluster, networkName, subnetworkName, np1, np2)
 }
```

zli82016: > Yes, I can do that. I still can't figure out why the first step (`testAccContainerNodePool_concurrentCreate`) _doesn't_ have the version set on node pools, and then `testAccContainerNodePool_concurrentUpdate` does, though? Since it's supposed to "update" something, maybe the original idea is that, by having the _minimum_ version for the control plane be _higher_, that the update test should then be updating it to a _different_ (but close) version?
> 
I agree that it is just for updating test. 

> This would mean that just using the data source might not get it to work quite as easily... and yet, my earlier "fix" also wouldn't accomplish the intent of the test? But maybe we can do something like this, to create with "latest" version, and then update with "stable"? or is the change in the # of default nodes on one of the two nodes sufficient for the update test?

If it is not easy to use the data source, the original hardcoded way is also acceptable.

wyardley: Alternative approach would be to do something like `data.google_container_engine_versions.central1a.valid_master_versions[0]` combined with `data.google_container_engine_versions.central1a.valid_node_versions[2]`?

zli82016 (approved):
LGTM

PR #13165, opened 2025-02-22
Apply service label during test ticket creation

shuyama1:
This PR updates the test failure tickets creator to add service labels and reverts the changes to compute_new_labels made in https://github.com/GoogleCloudPlatform/magic-modules/pull/13086 since it was working as expected, while keep the labeler backfill changes made in that PR unchanged as it should be working.

```release-note:none

```

### DIFF
--- a/.ci/magician/cmd/create_test_failure_ticket.go
+++ b/.ci/magician/cmd/create_test_failure_ticket.go
@@ -30,6 +30,8 @@ import (
 	"github.com/google/go-github/v68/github"
 	"github.com/spf13/cobra"
 
+	"github.com/GoogleCloudPlatform/magic-modules/tools/issue-labeler/labeler"
+
 	_ "embed"
 )
 
@@ -177,9 +179,6 @@ func execCreateTestFailureTicket(now time.Time, gh *github.Client, gcs Cloudstor
 			if err != nil {
 				return fmt.Errorf("error creating test failure ticket: %w", err)
 			}
-			fmt.Println("Yes, should create ticket")
-		} else {
-			fmt.Println("No, should not create ticket")
 		}
 	}
 	return nil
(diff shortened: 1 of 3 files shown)


shuyama1:
I've tested it over the weekend, and can confirm that service labels are added during creation job. [Test failure tickets](https://github.com/hashicorp/terraform-provider-google/issues?q=sort%3Aupdated-desc%20is%3Aissue%20is%3Aopen%20label%3Atest-failure)

shuyama1:
> So my understanding is:
> 
> * Auto-created test tickets automatically forward
> * Backfill jobs automatically forward test tickets (we run these very occasionally)
> * One-off job runs or manually created test tickets need to be reviewed
> 
> Is that right? Just want to make sure that's the behavior we're looking for, and it could be worth comments in the code (or some place else?) to mention the different treatment with forwarding.

Correct! Thanks for the review.

- auto-created test tickets have service label applied during creation (implemented in this PR), so they should be automatically forwarded
- labeler backfill job [skips apply `forward/review` label for test tickets](https://github.com/GoogleCloudPlatform/magic-modules/blob/main/tools/issue-labeler/labeler/backfill.go#L107), so they get automatically forwarded. (and I think we run backfill job [daily](https://github.com/hashicorp/terraform-provider-google/actions/workflows/backfill-issue-labels.yml) nowadays)
- we also have a [manage ticket job](https://github.com/GoogleCloudPlatform/magic-modules/blob/main/.ci/magician/cmd/manage_test_failure_ticket.go#L91) that removes review label to get test ticket forwarded. It runs daily, so it should handle one-off job runs and manually created test tickets

So technically, all test tickets will be automatically forwarded within 24 hours.

I've added some comments in the code for now. I'm working on documentation on the test ticket automation and will make sure to include these different forwarding situations.

roaks3:
Makes sense, thanks for explaining!

PR #13169, opened 2025-02-24
Update Gemini Logging Setting Binding Resource - Change product possible values

PerlMonker303:

```release-note:note
gemini: removed unsupported value `GEMINI_CLOUD_ASSIST` for field `product` in `google_gemini_logging_setting_binding` resource
```

### DIFF
--- a/mmv1/products/gemini/LoggingSettingBinding.yaml
+++ b/mmv1/products/gemini/LoggingSettingBinding.yaml
@@ -88,7 +88,6 @@ properties:
     description: |-
       Product type of the setting binding.
     enum_values:
-      - 'GEMINI_CLOUD_ASSIST'
       - 'GEMINI_CODE_ASSIST'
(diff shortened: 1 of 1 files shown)


shuyama1 (commented):
Based on the changlog, I assume we only want to exclude it from the GA resource? It doesn't seem we have a test coverage on this value, so I can't confirm. 

If so, I think we lack the capability to generate distinct enum lists for GA and beta resources currently. I recommend updating the documentation to clarify that `GEMINI_CLOUD_ASSIST` is only supported in beta resource.

PerlMonker303:
This change should be applied in both google-beta and google (GA) releases, I updated the changelog correspondingly.

shuyama1 (approved):

shuyama1:
Sorry, hit the wrong button. Meant to merge it.

PR #13176, opened 2025-02-25
Adds StorageControl service folder for GCS JSON v2 resource support.

kautikdk:
```release-note: none

```


kautikdk:
Hi @shuyama1,
I am not sure why the beta build is failing. I checked locally and it works for ga provider but not for beta because of `provider_mmv1_resources.go` file being modified. It generates one additional module import for beta provider which is not available. 
Can you please look into it what is happening?

PR #13177, opened 2025-02-25
Add examples for AlloyDB Major Version Upgrade

GauravJain21:

```release-note:none
alloydb: add examples for AlloyDB Major Version Upgrade for `google_alloydb_cluster` resource.
```


slevenick (changes requested):
Looks like the tests aren't being generated with all the right values.

What's the goal here, to show before and after for updating the Postgres version? Is that a difficult upgrade path?

GauravJain21:
> Looks like the tests aren't being generated with all the right values.
> 
> What's the goal here, to show before and after for updating the Postgres version? Is that a difficult upgrade path?

Fixed the issue. It's a trivial CUJ but it's mandatory to launch this feature in GA.

slevenick (approved):

PR #13188, opened 2025-02-26
Remove immutable annotation from node type to support vertical scaling for Valkey.

singh-shreya321:

```release-note:enhancement
memorystore: enabled update support for `node_type` field in `google_memorystore_instance` resource
```

### DIFF
--- a/mmv1/products/memorystore/Instance.yaml
+++ b/mmv1/products/memorystore/Instance.yaml
@@ -211,9 +211,8 @@ properties:
   - name: 'nodeType'
     type: String
     description:
-      "Optional. Immutable. Machine type for individual nodes of the instance.
+      "Optional. Machine type for individual nodes of the instance.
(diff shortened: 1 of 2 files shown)


NickElliot (changes requested):
can you add a test for update support with this? thanks!

NickElliot (changes requested):
> google/services/memorystore/resource_memorystore_instance_test.go:208:6: unknown field node_type in struct literal of type InstanceParams

is the experienced failure -- can you run builds locally to catch these? let me know if you need help with that, thanks!

singh-shreya321:
@NickElliot updated the test and tested it locally. PTAL.

NickElliot (commented):
what changed in your latest commit? can you please avoid force pushing a squashed commit like this in the future, it makes it very hard to keep track of review history. I want to be able to see each iteration as its own new commit.

singh-shreya321:
Apologies. Noted, I will take care of this for my future commits.
In the most recent commit, I just updated the comment to "create cluster with node type" from ""create cluster with deletion protection enabled". Same for "update cluster with node type", in the test file.

NickElliot (approved):
LGTM

PR #8265, opened 2023-07-04
Added windows_node_config support for GKE Clusters

pranavparuchuri:
Adding support for windows_node_config in gke cluster node_pools. This is required to specify windows 2022 as os version for the nodes. The corresponding issue is https://github.com/hashicorp/terraform-provider-google/issues/14979

If this PR is for Terraform, I acknowledge that I have:

- [ X] Searched through the [issue tracker](https://github.com/hashicorp/terraform-provider-google/issues) for an open issue that this either resolves or contributes to, commented on it to claim it, and written "fixes {url}" or "part of {url}" in this PR description. If there were no relevant open issues, I opened one and commented that I would like to work on it (not necessary for very small changes).
- [X ] Ensured that all new fields I added that can be set by a user appear in at least one [example](https://github.com/GoogleCloudPlatform/magic-modules/tree/main/mmv1/templates/terraform/examples) (for generated resources) or [third_party test](https://github.com/GoogleCloudPlatform/magic-modules/tree/main/mmv1/third_party/terraform/tests) (for handwritten resources or update tests).
- [X ] [Generated Terraform providers](https://github.com/GoogleCloudPlatform/magic-modules/blob/main/docs/content/docs/getting-started/generate-providers.md), and ran [`make test` and `make lint`](https://googlecloudplatform.github.io/magic-modules/docs/getting-started/run-provider-tests/#run-unit-tests) in the generated providers to ensure it passes unit and linter tests.
- [X ] [Ran](https://github.com/GoogleCloudPlatform/magic-modules/blob/main/docs/content/docs/getting-started/run-provider-tests.md) relevant acceptance tests using my own Google Cloud project and credentials (If the acceptance tests do not yet pass or you are unable to run them, please let your reviewer know).
- [ ] Read [Write release notes](https://googlecloudplatform.github.io/magic-modules/contribute/release-notes/) before writing my release note below.

**Release Note Template for Downstream PRs (will be copied)**

```release-note:enhancement
GKE: added `node_config.windows_node_config` field to `google_container_node_pool` resource
```

PR #13197, opened 2025-02-26
Add support for WindowsNodeConfig to GKE NodePool

quoctruong:
Add support for windows_node_config to GKE NodePool. This is based on https://github.com/GoogleCloudPlatform/magic-modules/pull/8265. The corresponding issue is https://github.com/hashicorp/terraform-provider-google/issues/14979

```release-note:enhancement
container: added `node_config.windows_node_config` field to `google_container_node_pool` resource.
```

### DIFF
--- a/mmv1/third_party/terraform/services/container/node_config.go.tmpl
+++ b/mmv1/third_party/terraform/services/container/node_config.go.tmpl
@@ -710,6 +710,7 @@ func schemaNodeConfig() *schema.Schema {
 				"windows_node_config": {
 					Type:        schema.TypeList,
 					Optional:    true,
+					Computed:    true,
 					MaxItems:    1,
 					Description: `Parameters that can be configured on Windows nodes.`,
 					Elem: &schema.Resource{
--- a/mmv1/third_party/terraform/services/container/node_config.go.tmpl
+++ b/mmv1/third_party/terraform/services/container/node_config.go.tmpl
@@ -707,6 +707,24 @@ func schemaNodeConfig() *schema.Schema {
 						},
 					},
 				},
+				"windows_node_config": {
+					Type:        schema.TypeList,
+					Optional:    true,
+					MaxItems:    1,
+					Description: `Parameters that can be configured on Windows nodes.`,
+					Elem: &schema.Resource{
+						Schema: map[string]*schema.Schema{
+							"osversion": {
+								Type:         schema.TypeString,
+								Optional:     true,
+								ForceNew:     true,
+								Default:      "OS_VERSION_UNSPECIFIED",
+								Description:  `The OS Version of the windows nodepool.Values are OS_VERSION_UNSPECIFIED,OS_VERSION_LTSC2019 and OS_VERSION_LTSC2022`,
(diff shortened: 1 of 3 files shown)


mmv1/third_party/terraform/services/container/node_config.go.tmpl
```diff
@@ -707,6 +707,24 @@ func schemaNodeConfig() *schema.Schema {
 						},
 					},
 				},
+				"windows_node_config": {
+					Type:        schema.TypeList,
+					Optional:    true,
```
SirGitsalot, line 712: I *think* this needs a `Computed: true`. Looking at the logs for the failing non-Windows tests, the API is returning an empty `windowsNodeConfig` in its responses. That wasn't a problem before since the provider ignores any API field it doesn't recognize, but with this change the field is getting returned to Terraform. That means Terraform is sending a config with no `windows_node_config` to the provider, but getting back `windows_node_config {}`. Annoyingly, this behavior is different from  `linuxNodeConfig` which doesn't seem to be returned in API responses is not set in requests.

For instance, here's an API response for one of the non-Windows tests:

```
HTTP/2.0 200 OK
Date: Wed, 05 Mar 2025 00:53:04 GMT
Server: ESF
Vary: Origin
Vary: X-Origin
Vary: Referer
X-Frame-Options: SAMEORIGIN
X-Xss-Protection: 0

{
 "name": "tf-test-cluster-f8lokszref",
 "initialNodeCount": 3,
 "nodeConfig": {
  "machineType": "e2-medium",
  "diskSizeGb": 100,
  "oauthScopes": [
   "https://www.googleapis.com/auth/monitoring",
   "https://www.googleapis.com/auth/devstorage.read_only",
   "https://www.googleapis.com/auth/logging.write",
   "https://www.googleapis.com/auth/compute"
  ],
  "metadata": {
   "disable-legacy-endpoints": "true"
  },
  "imageType": "COS_CONTAINERD",
  "serviceAccount": "default",
  "diskType": "pd-balanced",
  "shieldedInstanceConfig": {
   "enableIntegrityMonitoring": true
  },
  "kubeletConfig": {
   "insecureKubeletReadonlyPortEnabled": true
  },
  "resourceLabels": {
   "goog-gke-node-pool-provisioning-model": "on-demand"
  },
  "loggingConfig": {},
  "windowsNodeConfig": {},
  "effectiveCgroupMode": "EFFECTIVE_CGROUP_MODE_V2"
 },
 ...
```

SirGitsalot (changes requested):

mmv1/third_party/terraform/services/container/node_config.go.tmpl
quoctruong: Thank you. Added.

quoctruong:
@SirGitsalot Thanks for the review. PTAL

SirGitsalot (commented):
LGTM, but I'm going to hold off on approving until Monday to run the `computed: true` solution by some folks who know more than me (I mean, it worked! but if it's subtly wrong it won't be fixable as removing it would be a breaking change once submitted)

quoctruong:
@SirGitsalot Have you had a chance to run `computed: true` for this?

SirGitsalot (approved):

quoctruong:
Thank you @SirGitsalot. Do you know when it will make it to the release?

SirGitsalot:
Looks like it will land in next week's 6.27.0 release.

PR #13205, opened 2025-02-26
Add Backend Service Custom Metrics.

efimki:

```release-note:REPLACEME

```

### DIFF
--- a/mmv1/products/compute/BackendService.yaml
+++ b/mmv1/products/compute/BackendService.yaml
@@ -147,8 +147,8 @@ properties:
             Specifies the balancing mode for this backend.
 
             For global HTTP(S) or TCP/SSL load balancing, the default is
-            UTILIZATION. Valid values are UTILIZATION, RATE (for HTTP(S))
-            and CONNECTION (for TCP/SSL).
+            UTILIZATION. Valid values are UTILIZATION, RATE (for HTTP(S)),
+            CUSTOM_METRICS (for HTTP(s)) and CONNECTION (for TCP/SSL).
 
             See the [Backend Services Overview](https://cloud.google.com/load-balancing/docs/backend-service#balancing-mode)
             for an explanation of load balancing modes.
@@ -157,6 +157,7 @@ properties:
             - 'UTILIZATION'
             - 'RATE'
             - 'CONNECTION'
+            - 'CUSTOM_METRICS'
         - name: 'capacityScaler'
           type: Double
           description: |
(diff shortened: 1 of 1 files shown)


pawelJas (approved):

mmv1/products/compute/BackendService.yaml
```diff
@@ -261,6 +270,39 @@ properties:
             Used when balancingMode is UTILIZATION. This ratio defines the
             CPU utilization target for the group. Valid range is [0.0, 1.0].
           default_from_api: true
+        - name: 'customMetrics'
+          type: Array
+          description: |
+            The set of custom metrics that are used for <code>CUSTOM_METRICS</code> BalancingMode.
+          is_set: false
```
slevenick, line 277: This should be the default, so it's not needed

mmv1/products/compute/BackendService.yaml
```diff
@@ -960,6 +1010,31 @@ properties:
               description: |
                 An optional, arbitrary JSON object with configuration data, understood
                 by a locally installed custom policy implementation.
+  - name: 'customMetrics'
```
slevenick, line 1013: Does this not need maxUtilization?

mmv1/templates/terraform/examples/region_backend_service_ilb_custom_metrics.tf.tmpl
```diff
@@ -0,0 +1,52 @@
...
+    group = google_compute_network_endpoint_group.default.id
+    balancing_mode = "CUSTOM_METRICS"
+    custom_metrics {
+      name    = "orca.cpu_utilization"
+      max_utilization = 0.9
+      dry_run = true
+    }
+    custom_metrics {
+      name    = "orca.named_metrics.foo"
+      max_utilization = 0.9
```
slevenick, line 39: Can you not set this value in one of these tests?

slevenick (changes requested):
Please add an update test for updating some of these fields. Things like:
1. Adding a custom metric when several already exist
2. Removing a custom metric from several
3. Updating dry_run from true -> false. We typically see issues with this type of update, so it would be good to check

mmv1/products/compute/BackendService.yaml
efimki: Done.

mmv1/products/compute/BackendService.yaml
efimki: No, only customMetrics that are set on the Backend (as opposed to BackendService) need maxUtilization.

mmv1/templates/terraform/examples/region_backend_service_ilb_custom_metrics.tf.tmpl
efimki: Done

efimki:
> Please add an update test for updating some of these fields. Things like:
> 
> 1. Adding a custom metric when several already exist
> 2. Removing a custom metric from several
> 3. Updating dry_run from true -> false. We typically see issues with this type of update, so it would be good to check

Sounds good, I will try to add those. Is https://googlecloudplatform.github.io/magic-modules/test/test/#add-an-update-test the right instruction to follow?

efimki (commented):
Thank you for quick review!

efimki:
> Please add an update test for updating some of these fields. Things like:
> 
> 1. Adding a custom metric when several already exist
> 2. Removing a custom metric from several
> 3. Updating dry_run from true -> false. We typically see issues with this type of update, so it would be good to check

I've added `TestAccComputeBackendService_backendServiceCustomMetrics_update`, PTAL.

slevenick (approved):

PR #13207, opened 2025-02-26
Add a diffSupressFunc() for template_gcs_path.

harrisonlimh:
Fixes https://github.com/hashicorp/terraform-provider-google/issues/21567

Add a diffSupressFunc() for template_gcs_path.

```release-note:bug
dataflow: fixed a permadiff on `template_gcs_path` in `google_dataflow_job` resource
```

### DIFF
--- a/mmv1/third_party/terraform/services/dataflow/resource_dataflow_job.go.tmpl
+++ b/mmv1/third_party/terraform/services/dataflow/resource_dataflow_job.go.tmpl
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"log"
+	"regexp"
 	"strings"
 	"time"
 
@@ -299,7 +300,7 @@ func resourceDataflowJobTypeCustomizeDiff(_ context.Context, d *schema.ResourceD
 				if field == "template_gcs_path" {
 					old, new := d.GetChange("template_gcs_path")
 
-					if d.Get("region") != nil && isRegionSuffixedPathMatch(old.(string), new.(string), d.Get("region").(string)) {
+					if d.Get("region") != nil && isRegionSuffixedPathMatch(old.(string), new.(string))) {
 						continue
 					}
 				}
--- a/mmv1/third_party/terraform/services/dataflow/resource_dataflow_job.go.tmpl
+++ b/mmv1/third_party/terraform/services/dataflow/resource_dataflow_job.go.tmpl
@@ -300,7 +300,7 @@ func resourceDataflowJobTypeCustomizeDiff(_ context.Context, d *schema.ResourceD
 				if field == "template_gcs_path" {
 					old, new := d.GetChange("template_gcs_path")
 
-					if d.Get("region") != nil && isRegionSuffixedPathMatch(old.(string), new.(string))) {
+					if d.Get("region") != nil && isRegionSuffixedPathMatch(old.(string), new.(string)) {
 						continue
 					}
 				}
--- a/mmv1/third_party/terraform/services/dataflow/resource_dataflow_job.go.tmpl
+++ b/mmv1/third_party/terraform/services/dataflow/resource_dataflow_job.go.tmpl
@@ -72,9 +72,16 @@ func ResourceDataflowJobTemplateGcsPathDiffSuppress(k, old, new string, d *schem
 }
 
 
-func isRegionSuffixedPathMatch(old, new, region string) bool {
-	old = strings.Replace(old, "-"+region, "", 1)
-	return old == new
+func isRegionSuffixedPathMatch(old, new string) bool {
+    re := regexp.MustCompile(`gs://([a-z0-9\-]+)-[a-z0-9]+-[a-z0-9]+(/.*)?`)
+    matches := re.FindStringSubmatch(old)
+    
+    if len(matches) == 3 && matches[2] != "" {
+        modifiedOld := "gs://" + matches[1] + matches[2]
+        return modifiedOld == new
+    }
+
+    return false
 }
 
 func ResourceDataflowJob() *schema.Resource {
--- a/mmv1/third_party/terraform/services/dataflow/resource_dataflow_job.go.tmpl
+++ b/mmv1/third_party/terraform/services/dataflow/resource_dataflow_job.go.tmpl
@@ -313,16 +313,6 @@ func resourceDataflowJobTypeCustomizeDiff(_ context.Context, d *schema.ResourceD
 	return nil
 }
 
-// zoneToRegion converts a zone to a region.
-// For example, "us-central1-a" -> "us-central1".
-func zoneToRegion(zone string) string {
-	parts := strings.Split(zone, "-")
-	if len(parts) != 3 {
-		return ""
-	}
-	return fmt.Sprintf("%s-%s", parts[0], parts[1])
-}
-
 // return true if a job is in a terminal state, OR if a job is in a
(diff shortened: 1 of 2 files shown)


mmv1/third_party/terraform/services/dataflow/resource_dataflow_job.go.tmpl
```diff
@@ -87,6 +87,7 @@ func ResourceDataflowJob() *schema.Resource {
 			"template_gcs_path": {
 				Type:        schema.TypeString,
 				Required:    true,
+				DiffSuppressFunc: tpgresource.AlwaysDiffSuppress,
```
shuyama1, line 90: We'd want to only suppress this specific diff - bucket name adds region suffix, instead of all diffs, as we still need to detect changes for this field.

I don't think we currently have a diff suppress function that can handle this case specifically. You'll need to write a new function yourself, likely named `resourceDataflowJobTemplateGcsPathDiffSuppress`.

See [`resourceDataflowJobLabelDiffSuppress`](https://github.com/harrisonlimh/magic-modules/blob/317edc826144c910eeb9b7653d3882d39a2d8fcc/mmv1/third_party/terraform/services/dataflow/resource_dataflow_job.go.tmpl#L38) function as an example. 

`old` is the value read from the API, like "gs://dataflow-templates-us-central1/latest/Word_Count" 
`new` is what in the config, like "gs://dataflow-templates/latest/Word_Count"
`region` value could likely be retrieved by `d.Get("region")`

Please let me know if you have any other questions, thank you.

harrisonlimh: Thank you! Done.

shuyama1:
Kicking off skipped VCR tests on [TC](https://hashicorp.teamcity.com/buildConfiguration/TerraformProviders_GoogleCloud_GOOGLE_BETA_MMUPSTREAMTESTS_GOOGLEBETA_PACKAGE_DATAFLOW/330584). I'm also testing some cases locally. Will update the result here soon.

shuyama1 (commented):
DSF looks good to me - thanks for adding the unit tests. Could you change `testDataflowJobTemplateWordCountUrl` to `"gs://dataflow-templates/latest/Word_Count"` (global bucket) and leave `testDataflowJobTemplateTextToPubsub` pointing to the regional bucket (as it is today). This will ensure we test both scenarios? Thank you!

harrisonlimh:
> DSF looks good to me - thanks for adding the unit tests. Could you change `testDataflowJobTemplateWordCountUrl` to `"gs://dataflow-templates/latest/Word_Count"` (global bucket) and leave `testDataflowJobTemplateTextToPubsub` pointing to the regional bucket (as it is today). This will ensure we test both scenarios? Thank you!

Done! PTAL!

shuyama1:
Running skipped tests on [TC](https://hashicorp.teamcity.com/buildConfiguration/TerraformProviders_GoogleCloud_GOOGLE_BETA_MMUPSTREAMTESTS_GOOGLEBETA_PACKAGE_DATAFLOW/333870)

mmv1/third_party/terraform/services/dataflow/resource_dataflow_job.go.tmpl
```diff
@@ -51,6 +51,37 @@ func resourceDataflowJobLabelDiffSuppress(k, old, new string, d *schema.Resource
...
 
+// ResourceDataflowJobTemplateGcsPathDiffSuppress suppresses diffs on the template_gcs_path field
+// when the only diff is the region name in the bucket.
+// `old` is the value set by service and `new` is what is stored in the config. When using a public
+// template bucket, it will be automatically turned to use a regional bucket path. In this case,
+// we want to supress the diff.
+// Example:
+//   - old: gs://template-bucket-us-central1/path/to/file.txt
+//   - new: gs://template-bucket/path/to/file.txt
+func ResourceDataflowJobTemplateGcsPathDiffSuppress(k, old, new string, d *schema.ResourceData) bool {
```
shuyama1, line 62: `resourceDataflowJobTypeCustomizeDiff` forces the resource recreation (makes every field `forcenew`) if any non-virtual fields (except `terraform_labels` field) changed for batch jobs, we need to extend it with this diff suppress logic. 

A working example is available in https://github.com/GoogleCloudPlatform/magic-modules/pull/13351, but the code can be optimized. Hope this helps

harrisonlimh: Done.

shuyama1:
Running skipped tests on [TC](https://hashicorp.teamcity.com/buildConfiguration/TerraformProviders_GoogleCloud_GOOGLE_BETA_MMUPSTREAMTESTS_GOOGLEBETA_PACKAGE_DATAFLOW/338146)

mmv1/third_party/terraform/services/dataflow/resource_dataflow_job.go.tmpl
```diff
@@ -51,6 +52,39 @@ func resourceDataflowJobLabelDiffSuppress(k, old, new string, d *schema.Resource
...
+	if old == new {
+		return true
+	}
+
+	if d.Get("region") == nil {
+		return false
+	}
+
```
shuyama1, line 71:
```suggestion
```
as we don't need to check region anytime

mmv1/third_party/terraform/services/dataflow/resource_dataflow_job.go.tmpl
```diff
@@ -262,13 +297,20 @@ func resourceDataflowJobTypeCustomizeDiff(_ context.Context, d *schema.ResourceD
 			}
 
 			if field != "terraform_labels" && d.HasChange(field) {
+				if field == "template_gcs_path" {
+					old, new := d.GetChange("template_gcs_path")
+
+					if d.Get("region") != nil && isRegionSuffixedPathMatch(old.(string), new.(string)) {
```
shuyama1, line 303:
```suggestion
					isRegionSuffixedPathMatch(old.(string), new.(string)) {
```
same here.

mmv1/third_party/terraform/services/dataflow/resource_dataflow_job.go.tmpl
```diff
@@ -262,13 +293,20 @@ func resourceDataflowJobTypeCustomizeDiff(_ context.Context, d *schema.ResourceD
 			}
 
 			if field != "terraform_labels" && d.HasChange(field) {
+				if field == "template_gcs_path" {
+					old, new := d.GetChange("template_gcs_path")
+
+					isRegionSuffixedPathMatch(old.(string), new.(string)) {
```
shuyama1, line 299:
```suggestion
					if isRegionSuffixedPathMatch(old.(string), new.(string)) {
```

shuyama1:
Triggered tests on [TC](https://hashicorp.teamcity.com/buildConfiguration/TerraformProviders_GoogleCloud_GOOGLE_BETA_MMUPSTREAMTESTS_GOOGLEBETA_PACKAGE_DATAFLOW/338277)

eyalzek:
@shuyama1 @harrisonlimh can we get a clarification as to why this permadiff started happening? Was it a change in the upstream Dataflow API?
We're seeing this behavior with GKE Config Connector - it creates a new job on every reconciliation interval when using a non-regional GCS template path, so we're trying to understand the root cause.

Thanks!

PR #13216, opened 2025-02-27
Skip changelog generation when changelog type is none

wyardley:
Could use some help in terms of doing some functional testing on this and / or suggesting a way I could add some test coverage. But basically, the idea here is to skip generating a changelog entry when the changelog type is `none`

Mostly, this should prevent adding an extra changelog entry (whether or not any other changes are generated), avoiding creating extra noise in the downstream repos to create a bunch of empty files.

Separately, I could make a PR to clean up the empty changelog files in the tpg / tpg-beta repos if there's interest.

Fixes hashicorp/terraform-provider-google#20134

```release-note:none

```

I made a first pass at some unit tests locally (this file currently has 0 coverage), but a bit stuck on the best way to mock the runner and whether or not to use `mock_runner_test.go`.

Looking at downstream, I do see that there are a handful of cases where there's a note even though release notes is `none`, e.g., (in `9889.txt`):
````
```release-note:none
Fixes: b/304402329, b/304432955, b/304432958
```
````
It would probably be a little harder, but would be possible to retain those if we did want to send them downstream, though I think the changelog is only used for release notes anyway?

### DIFF
--- a/.ci/magician/cmd/generate_downstream.go
+++ b/.ci/magician/cmd/generate_downstream.go
@@ -1,3 +1,18 @@
+/*
+* Copyright 2025 Google LLC. All Rights Reserved.
+*
+* Licensed under the Apache License, Version 2.0 (the "License");
+* you may not use this file except in compliance with the License.
+* You may obtain a copy of the License at
+*
+*     http://www.apache.org/licenses/LICENSE-2.0
+*
+* Unless required by applicable law or agreed to in writing, software
+* distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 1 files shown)


trodge (approved):

wyardley:
Thanks @trodge. were you able to find a way to test? from what I can see, this generated nothing downstream, and https://github.com/hashicorp/terraform-provider-google/pull/22229, which came after mine, did seem to generate one, so hopefully it's working(ish)?

Is it worth me making a PR to remove the empty changelog entries as well (directly against tpg / tpg-beta repos)?

I noticed some PRs have multiple release notes, but I assume it wouldn't be expected to have a "none" type mixed with other / real types?

PR #13227, opened 2025-02-27
Added documentation about how to handle "root object was present, but now absent"

melinath:
ref yaqs/5773424506777370624

Screenshot 2025-02-27 at 1 59 37 PM

Screenshot 2025-02-27 at 1 59 50 PM

```release-note:none

```

### DIFF
--- a/docs/content/develop/diffs.md
+++ b/docs/content/develop/diffs.md
@@ -300,3 +300,13 @@ func flattenResourceNameFieldName(v interface{}, d *schema.ResourceData, config 
 {{< /tabs >}}
 
 For other Array fields, convert the field to a Set – this is a [breaking change]({{< ref "/breaking-changes/breaking-changes" >}}) and can only happen in a major release.
+
+## API is eventually consistent {#eventually-consistent}
+
+Eventually consistent APIs are not compliant with [AIP-121's strong consistency requirement](https://google.aip.dev/121#strong-consistency), but they do happen sometimes, and can lead to diffs.
+
+The fix is to add [post-create custom code]({{< ref "/develop/custom-code/#pre_post_injection" >}}) which adds a sleep after creation completes. There are three premade templates that can be used for this:
+
+- [`templates/terraform/post_create/sleep.go.tmpl`](https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/templates/terraform/post_create/sleep.go.tmpl)
+- [`templates/terraform/post_create/sleep_2_min.go.tmpl`](https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/templates/terraform/post_create/sleep_2_min.go.tmpl)
+- [`templates/terraform/post_create/sleep_5_min.go.tmpl`](https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/templates/terraform/post_create/sleep_5_min.go.tmpl)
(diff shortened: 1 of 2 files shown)


docs/content/develop/diffs.md
```diff
@@ -300,3 +300,15 @@ func flattenResourceNameFieldName(v interface{}, d *schema.ResourceData, config
...
 
 For other Array fields, convert the field to a Set – this is a [breaking change]({{< ref "/breaking-changes/breaking-changes" >}}) and can only happen in a major release.
+
+## API is eventually consistent or returns an Operation {#eventually-consistent}
+
+If a resource's creation or update takes a long time before the results will be returned from a Read call, it can lead to diffs.
+
+ If the method returns an Operation, the best thing to do is wait for the operation to complete. In MMv1 resources, this can be done by setting [`autogen_async: true`](https://googlecloudplatform.github.io/magic-modules/reference/resource/#autogen_async) and configuring the resource's [`async`](https://googlecloudplatform.github.io/magic-modules/reference/resource/#async) settings.
+
+Eventually consistent APIs that do not return an  Operation are not compliant with [AIP-121's strong consistency requirement](https://google.aip.dev/121#strong-consistency), but they do exist. The fix in this case is to add [post-create custom code]({{< ref "/develop/custom-code/#pre_post_injection" >}}) which adds a sleep after creation completes. There are three premade templates that can be used for this:
```
c2thorn, line 310:
```suggestion
If the method returns an Operation, the best thing to do is wait for the operation to complete. In MMv1 resources, this can be done by setting [`autogen_async: true`](https://googlecloudplatform.github.io/magic-modules/reference/resource/#autogen_async) and configuring the resource's [`async`](https://googlecloudplatform.github.io/magic-modules/reference/resource/#async) settings.

Eventually consistent APIs that do not return an Operation are not compliant with [AIP-121's strong consistency requirement](https://google.aip.dev/121#strong-consistency), but they do exist. The fix in this case is to add [post-create custom code]({{< ref "/develop/custom-code/#pre_post_injection" >}}) which adds a sleep after creation completes. There are three premade templates that can be used for this:
```

docs/content/test/run-tests.md
```diff
@@ -127,12 +127,25 @@ aliases:
...
+### Blocks of type "FIELD_NAME" are not expected here
+
+The field called `FIELD_NAME` does not exist; this is either because it has not been implemented or because the test is running for the `google` provider and the field is only implemented in the `google-beta` provider. See [Add resource tests]({{< ref "/test/test" >}}) for information on using version guards to exclude beta-only fields from GA tests, or [Promote from beta to GA]({{< ref "/develop/promote-to-ga" >}}) for information on how to promote fields that were accidentally made beta-only.
+
+### Provider produced inconsistent result after apply ... Root object was present, but now absent.
+
+This indicates that after an apply to create or update a resource, the resource was not present in Terraform state. This generally means one of a few things:
+
+- [API is eventually consistent or returns an Operation]({{< ref "/develop/diffs#eventually-consistent" >}})
+- The URL for Reads was built incorrectly. The exact fix will depend on why this is happening.
```
c2thorn, line 147: "Check the URL in the debug log if this is the case" ?
Or is checking the debug log implied for most of these?

c2thorn (approved):

docs/content/develop/diffs.md
c2thorn: looks like the spaces don't even appear in the preview, but might as well?

docs/content/test/run-tests.md
melinath: I think that totally makes sense to call out, thanks for the suggestion!

PR #13228, opened 2025-02-27
Use the location of the folder when configuring folder logging bucket

SirGitsalot:
This should resolve https://github.com/hashicorp/terraform-provider-google/issues/17108 for real this time. The previous "fix" got it passing in the GitHub action integration tests, but it continued failing in the nightly test runs. The problem is that `google_logging_organization_settings` is a singleton, and some tests need it configured one way and others a different way. The changes here prevent the tests that depend on it from running concurrently, and each test configures it as needed rather than relying on a previously run test to do so.

```release-note: none

```


SirGitsalot:

roaks3 (approved):

PR #13229, opened 2025-02-27
Added documentation about how to use bootstrapping

melinath:
Fixed https://github.com/hashicorp/terraform-provider-google/issues/12540

Screenshot 2025-02-27 at 3 19 05 PM
Screenshot 2025-02-27 at 3 19 13 PM
Screenshot 2025-02-27 at 3 19 22 PM

```release-note:none

```

### DIFF
--- a/docs/content/test/test.md
+++ b/docs/content/test/test.md
@@ -38,9 +38,44 @@ For more information about testing, see the [official Terraform documentation](h
    git checkout main && git clean -f . && git checkout -- . && git pull
    ```
 
+## Add unit tests
+
+A unit test verifies functionality that is not related to interactions with the API, such as
+[diff suppress functions]({{<ref "/reference/field#diff_suppress_func" >}}),
+[validation functions]({{<ref "/reference/field#validation" >}}),
+CustomizeDiff functions, and so on.
+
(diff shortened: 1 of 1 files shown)


docs/content/test/test.md
```diff
@@ -236,40 +271,133 @@ An update test is a test that creates the target resource and then makes updates
...
+}
+```
+{{< /tab >}}
+{{< /tabs >}}
+
+### Networks
+
+Bootstrapping networks can be useful for two reasons:
+
+1. Resources like service_networking_connection use a consumer network and create a complementing tenant network which we don't control. These tenant networks never get cleaned up and they can accumulate to the point where a limit is reached for the organization. By reusing a consumer network across test runs, we can reduce the number of tenant networks that are needed. (Googlers: See b/146351146 for more context.)
```
shuyama1, line 361:
```suggestion
1. Resources like `google_service_networking_connection` use a consumer network and create a complementing tenant network which we don't control. These tenant networks never get cleaned up and they can accumulate to the point where a limit is reached for the organization. By reusing a consumer network across test runs, we can reduce the number of tenant networks that are needed. (Googlers: See b/146351146 for more context.)
```
Not sure if we want to use the Terraform resource name here or just point it as a general reference

shuyama1 (approved):

docs/content/test/test.md
melinath: I think we should list explicit resource names since people reading this document won't necessarily know if a particular API resource has this behavior. If we find additional similar resources in the future we can add them here as well.

shuyama1 (approved):

PR #13233, opened 2025-02-28
fix: correct output value path

apeabody:

```release-note: none
```

### DIFF
--- a/mmv1/third_party/terraform/website/docs/d/container_attached_versions.html.markdown
+++ b/mmv1/third_party/terraform/website/docs/d/container_attached_versions.html.markdown
@@ -18,7 +18,7 @@ data "google_container_attached_versions" "uswest" {
 
 
 output "first_available_version" {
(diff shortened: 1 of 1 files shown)


NickElliot (commented):
all our tests are using the `.versions`  right now so I would assume this example is correct, is there some other docs you could link or reasoning for this change?

apeabody:
> all our tests are using the `.versions` right now so I would assume this example is correct, is there some other docs you could link or reasoning for this change?

Hi @NickElliot - The example currently does not work as-is because the `google_container_attached_versions` data resource is called (`uswest`)[https://github.com/GoogleCloudPlatform/magic-modules/pull/13233/files#diff-5e1ac297f96242f5f8f5bbfdd1965fc1e158a1b13d510486c7d116c4af53d7d4R14].  If desired I can instead change `google_container_attached_versions` to `versions`, just as long as they are aligned.

NickElliot (approved):
my bad, I misread everything! lgtm!

PR #13234, opened 2025-02-28
Made nested query resources replace the whole resource instead of merging new over old

melinath:
This allows fields to be unset using the same logic that any other field on any other resource would use.

Related to https://github.com/hashicorp/terraform-provider-google/issues/14848

b/394945456#comment14

```release-note:bug
compute: fixed bug in `google_compute_router_nat` where `max_ports_per_vm` couldn't be unset once set.
```

Note: Technically this also allows `nat_ip_allocate_option` to be unset, but that field must be set for type = PUBLIC and _can't_ be set for type = PRIVATE, and modifying `type` forces recreation, so there doesn't seem to be a case where it can actually be unset on update.

This doesn't impact `google_compute_router_nat` because the only optional field has send_empty_value set.

### DIFF
--- a/mmv1/third_party/terraform/services/compute/resource_compute_router_nat_test.go.tmpl
+++ b/mmv1/third_party/terraform/services/compute/resource_compute_router_nat_test.go.tmpl
@@ -1229,40 +1229,6 @@ resource "google_compute_router_nat" "foobar" {
 `, routerName, routerName, routerName, routerName, routerName)
 }
 
-func testAccComputeRouterNatUnsetOptionalFields(routerName string) string {
-	return fmt.Sprintf(`
-resource "google_compute_router" "foobar" {
-  name    = "%s"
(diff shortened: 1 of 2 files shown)


mmv1/third_party/terraform/services/compute/resource_compute_router_nat_test.go.tmpl
```diff
@@ -108,37 +124,11 @@ func TestAccComputeRouterNat_update(t *testing.T) {
...
-			{
-				ResourceName:      "google_compute_router_nat.foobar",
-				ImportState:       true,
-				ImportStateVerify: true,
-			},
-		},
-	})
-}
-
-func TestAccComputeRouterNat_removeLogConfig(t *testing.T) {
```
melinath, line 121: this is now covered by the main update test.

melinath:

roaks3 (approved):
Nice!

PR #13236, opened 2025-02-28
Update disk type schema field to support db-c4a-highmem-4 tier

shuang2018g:
New CloudSQL performance tier does not support PD_SSD, specifying PD_SSD as default disk type can cause creation failure unnecessarily. Update the field to computed without default allows CloudSQL to automatically choose the right disk type when not specified. 

```release-note:bug
sql: the Terraform-based default value for `settings.disk_type` in `google_sql_database_instance` has been replaced with server-assigned default, allowing for compatibility with machine types that require `HyperDisk_Balanced`
```


mmv1/third_party/terraform/services/sql/resource_sql_database_instance.go.tmpl
```diff
@@ -424,10 +424,10 @@ is set to true. Defaults to ZONAL.`,
 						"disk_type": {
 							Type:             schema.TypeString,
 							Optional:         true,
-							Default:          "PD_SSD",
+							Computed:         true,
 							ForceNew:         true,
 							DiffSuppressFunc: caseDiffDashSuppress,
-							Description:      `The type of data disk: PD_SSD or PD_HDD. Defaults to PD_SSD.`,
+							Description:      `The type of supported data disk is tier dependent and can be PD_SSD or PD_HDD or HyperDisk_Balanced `,
```
NickElliot, line 430:
```suggestion
							Description:      `The type of data disk. The default type is tier dependent and can be PD_SSD, PD_HDD, or HyperDisk_Balanced.`,
```

NickElliot (changes requested):
clarifying the description, let me know if this wording isnt accurate!

mmv1/third_party/terraform/services/sql/resource_sql_database_instance.go.tmpl
shuang2018g: The default type can be PD_SSD or HyperDisk_Balanced, but not PD_HDD

skysarthak:
@NickElliot can we make the following edit on L1675 in `mmv1/third_party/terraform/services/sql/resource_sql_database_instance_test.go`:
```
-      enterprisePlusTier := "db-perf-optimized-C4A-4"
+      enterprisePlusTier := "db-c4a-highmem-4"
```
I created a [PR](https://github.com/shuang2018g/magic-modules/pull/1) to update this value but Shuang is OOTO. Reason for this request is that the public tier name for C4A was changed, you can refer the new name in the documentation [here](https://cloud.google.com/sql/docs/mysql/machine-series-overview#c4a).

NickElliot (commented):
A question about default behaviors on the resource:
while this wouldnt be a breaking change, is there scenarios where a resource made today with Terraform (that would've ended up with the present `PD_SSD` default) will end up with a different default `disk_type` upon creation of an identical configuration?

skysarthak:
> A question about default behaviors on the resource: while this wouldnt be a breaking change, is there scenarios where a resource made today with Terraform (that would've ended up with the present `PD_SSD` default) will end up with a different default `disk_type` upon creation of an identical configuration?

No, Hyperdisk will only be set as the default disk type for newly launched C4A (still in preview) machine series in CloudSQL. So customer would have to explicitly configure the resource to use C4A series public tier for the disk type to default to Hyperdisk.

NickElliot (approved):
LGTM, change is non breaking due to the `default` being replaced with `computed`. All current configs will produce the `PD_SSD` output by default still, change allows for compatibility with new machine types.

PR #13237, opened 2025-03-01
artifactregistry: fix cleanup policies updates

Subserial:
Fixes hashicorp/terraform-provider-google#21348

[The original fix](https://github.com/GoogleCloudPlatform/magic-modules/pull/12667) panicked on type assertion failures and incorrectly hashed objects. Turns out, for objects in Maps (stored as sets) the diff suppress func doesn't matter as the values hash differently. This change moves the diff suppress functionality to the hashing step.

```release-note:bug
artifactregistry: fixed failing update when changing cleanup policies in `google_artifact_registry_repository`
```

### DIFF
--- a/mmv1/products/artifactregistry/Repository.yaml
+++ b/mmv1/products/artifactregistry/Repository.yaml
@@ -423,11 +423,13 @@ properties:
               description: |-
                 Match versions older than a duration.
               custom_expand: 'templates/terraform/custom_expand/duration_to_seconds.go.tmpl'
+              diff_suppress_func: 'durationDiffSuppress'
             - name: 'newerThan'
               type: String
               description: |-
                 Match versions newer than a duration.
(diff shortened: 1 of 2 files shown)


melinath (changes requested):
Please add tests that would have caught this. Definitely unit tests: https://googlecloudplatform.github.io/magic-modules/test/test/

It would probably also be good to extend [TestAccArtifactRegistryRepository_update](https://github.com/GoogleCloudPlatform/magic-modules/blob/32fa743f4372065cc1c73fdcd2f3224379714ca7/mmv1/third_party/terraform/services/artifactregistry/resource_artifact_registry_repository_test.go.tmpl#L11) to cover at least one failure instance.

joshdickersongoogle:
Per [melinath](https://github.com/melinath) request please add unit tests.

PR #13120, opened 2025-02-19
Adding magic module config for osconfigv2

yerniyazN:
**Release Note Template for Downstream PRs (will be copied)**

See [Write release notes](https://googlecloudplatform.github.io/magic-modules/contribute/release-notes/) for guidance.

```release-note:new-resource
`google_os_config_v2_policy_orchestrator`
```

PR #13244, opened 2025-03-03
Add osconfigv2 permissions to test project and organization

yerniyazN:
Needed for google_os_config_v2_policy_orchestrator and google_os_config_v2_policy_orchestrator_for_organization tests, which need a permissioned service account.
Added for #13120

```release-note:none

```


melinath:

yerniyazN:
@NickElliot how can I address canceled contributor-membership-checker and queued downstream-generation-and-test ?

NickElliot:
reassigning to reviewer for main associated change

melinath:
@modular-magician reassign-reviewer @slevenick

melinath (changes requested):
The downstream generation isn't running because you're modifying the .ci folder; this is expected.

However, permissions like this should generally be bootstrapped in the relevant test instead of provisioned with Terraform: https://googlecloudplatform.github.io/magic-modules/test/test/#iam-resources

(Unless I'm missing something.)

yerniyazN:
> The downstream generation isn't running because you're modifying the .ci folder; this is expected.
> 
> However, permissions like this should generally be bootstrapped in the relevant instead of provisioned with Terraform: https://googlecloudplatform.github.io/magic-modules/test/test/#iam-resources
> 
> (Unless I'm missing something.)

Yes, IAM bindings for project can be bootstraped. 
But what about org level service accounts? 
"serviceAccount:service-org-${org_id}@gcp-sa-osconfig.iam.gserviceaccount.com" doesn't replace org_id with actual organization

melinath:
ah - I scrolled to the bottom & missed that there were org-level IAM changes. Those should also be bootstrapped to avoid conflicting with other org-level IAM, yes. But the boostrapping logic doesn't currently support it. Would you be willing to add that support in this PR? It would require modifying this file: https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/third_party/terraform/acctest/bootstrap_iam_test_utils.go

I would suggest allowing `{organization_id}` in addition to `{project_number}`; the function could then split into member lists for project vs org and apply each only if needed. You can use `GetTestOrgFromEnv` to get the current org id.

melinath:

melinath (changes requested):
marking reviewed

PR #13261, opened 2025-03-05
feat: (storage) remove validation of max 2 items for data_locations in custom_placement_config field

gurusai-voleti:

```release-note:bug
storage: lowered the minimum required items for `custom_placement_config.data_locations` from 2 to 1, and removed the Terraform-enforced maximum item limit for the field in `google_storage_bucket`
```


mmv1/third_party/terraform/services/storage/resource_storage_bucket_test.go.tmpl
```diff
@@ -710,6 +710,24 @@ func TestAccStorageBucket_storageClass(t *testing.T) {
 	})
 }
 
+func TestAccStorageBucket_invalidStorageClass(t *testing.T) {
```
kautikdk, line 713: What exactly is this test validating and how is it related to the change in this PR?

mmv1/third_party/terraform/services/storage/resource_storage_bucket_test.go.tmpl
```diff
@@ -710,6 +710,24 @@ func TestAccStorageBucket_storageClass(t *testing.T) {
...
+	bucketName := fmt.Sprintf("tf-test-acc-bucket-%d", acctest.RandInt(t))
+
+	acctest.VcrTest(t, resource.TestCase{
+		PreCheck:                 func() { acctest.AccTestPreCheck(t) },
+		ProtoV5ProviderFactories: acctest.ProtoV5ProviderFactories(t),
+		CheckDestroy:             testAccStorageBucketDestroyProducer(t),
+		Steps: []resource.TestStep{
+			{
+				Config:      testAccStorageBucket_storageClass(bucketName, "HOTLINE", "US"),
+				ExpectError: regexp.MustCompile("The storage class you specified is not valid"),
```
kautikdk, line 725: Why are we putting error validation in the acceptance test when `storage_class` attribute doesn't have any special behavior? i.e. DiffSupressFunc,StateFunc,etc.

mmv1/third_party/terraform/services/storage/resource_storage_bucket_test.go.tmpl
gurusai-voleti: the change is to allow atleast one location in placement-config to allow creation of RAPID buckets, this test case is  validating the API response if any invalid storage classes are sent

mmv1/third_party/terraform/services/storage/resource_storage_bucket_test.go.tmpl
NickElliot: We don't need a test to validate API error messages in Terraform, negative test cases are used for when the validation or other special behavior is contained within Terraform as Kautik mentioned

NickElliot (changes requested):
that test case is unneeded, removing the max items limit should be fine as is as long as the api does accept it

mmv1/third_party/terraform/services/storage/resource_storage_bucket_test.go.tmpl
gurusai-voleti: removed the test case

kautikdk (approved):
LGTM for the change!
I would leave the validation constraints on your decision. Not sure if there should be max limit or not as I'm not aware of the requirement.

gurusai-voleti:
> LGTM for the change! I would leave the validation constraints on your decision. Not sure if there should be max limit or not as I'm not aware of the requirement.

no need of max limit as its validated by backend API

NickElliot (approved):
all lgtm! sorry for last delay, was flagging an issue with the breaking change detector to our team

PR #13266, opened 2025-03-05
Allow reassign-reviewer anywhere in the comment

melinath:

```release-note:none

```

### DIFF
--- a/.github/workflows/reassign-reviewer.yml
+++ b/.github/workflows/reassign-reviewer.yml
@@ -22,7 +22,7 @@ jobs:
         uses: actions-ecosystem/action-regex-match@d50fd2e7a37d0e617aea3d7ada663bd56862b9cc # v2.0.2
         with:
           text: ${{ github.event.comment.body }}
-          regex: '@modular-magician reassign-reviewer ?@?([a-zA-Z0-9-]+)?$'
+          regex: '@modular-magician reassign-reviewer ?@?([a-zA-Z0-9-]+)?'
       - name: Checkout Repository
         if: steps.read-comment.outputs.match != ''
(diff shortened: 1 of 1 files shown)


rileykarson (approved):

melinath:
> This will still require it end the comment due to the `$` right? Honestly I would have assumed the regex would capture any specific line of the comment as written before.

good point, removed the $ as well

.github/workflows/reassign-reviewer.yml
```diff
@@ -22,7 +22,7 @@ jobs:
         uses: actions-ecosystem/action-regex-match@d50fd2e7a37d0e617aea3d7ada663bd56862b9cc # v2.0.2
         with:
           text: ${{ github.event.comment.body }}
-          regex: '^@modular-magician reassign-reviewer ?@?([a-zA-Z0-9-]+)?$'
+          regex: '@modular-magician reassign-reviewer ?@?([a-zA-Z0-9-]+)?'
```
rileykarson, line 25: Is there a way we could enforce that we only match individual lines? How are new lines represented in `github.event.comment.body`?

melinath: If you don't have any objection to requiring standalone lines, I do like that better. :-)

taking a quick scan of https://api.github.com/repos/GoogleCloudPlatform/magic-modules/issues/comments, looks like \r\n or \n - guess it depends on the user OS. Apparently `\r` is also a possible value on some OSes. I've adjusted the regex accordingly.

rileykarson (approved):

.github/workflows/reassign-reviewer.yml
rileykarson: Yeah standalone line is what I'd expect, sole entry in a comment isn't.

PR #13278, opened 2025-03-06
Fix the issue that user can't use regional disk in global_instance_template

yaoyaoyaoGit:
Allow users to specify url link to zonal or regional disk resources in instance template.
Fixes https://github.com/hashicorp/terraform-provider-google/issues/10186

```release-note:bug
compute: fixed the issue that user can't use regional disk in `global_instance_template`.
```


shuyama1 (commented):
LGTM. only some nitpicks. Thank you!

shuyama1 (approved):
the failing test is unrelated to this PR.

PR #13279, opened 2025-03-06
Add support for continuous query to google_bigquery_job in beta

wj-chen:
Add support for continuous query to `google_bigquery_job` in beta

```release-note:enhancement
bigquery: added support for continuous query to `google_bigquery_job` (beta)
```

### DIFF
--- a/mmv1/products/bigquery/Job.yaml
+++ b/mmv1/products/bigquery/Job.yaml
@@ -57,6 +57,15 @@ examples:
     ignore_read_extra:
       - 'etag'
       - 'status.0.state'
+  - name: 'bigquery_job_query_continuous'
+    exclude_docs: true
+    primary_resource_id: 'job'
+    vars:
+      job_id: 'job_query_continuous'
+    ignore_read_extra:
+      - 'etag'
+      - 'status.0.state'
+    min_version: beta
   - name: 'bigquery_job_query_table_reference'
     primary_resource_id: 'job'
     vars:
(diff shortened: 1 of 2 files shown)


wj-chen:
@melinath could you rerun the checks? Not sure why one was stuck.

melinath:
hm, weird! I don't see any reason it should've been cancelled. /gcbrun

mmv1/products/bigquery/Job.yaml
```diff
@@ -392,6 +401,11 @@ properties:
                 enum_values:
                   - 'LAST'
                   - 'FIRST_SELECT'
+          - name: 'continuous'
```
melinath, line 404: Could you add modification of this field to an update test as well? (Ideally including setting it from true to false (or to unset))

melinath (changes requested):

mmv1/products/bigquery/Job.yaml
wj-chen: This resource doesn't support Update: https://github.com/hashicorp/terraform-provider-google/blob/63cdd6a13df80017b27cde4cd7488cf365cc19ee/google/services/bigquery/resource_bigquery_job.go#L1220

Public API: https://cloud.google.com/bigquery/docs/reference/rest/v2/jobs

melinath: ah, yep! Thanks!

melinath (approved):

PR #13283, opened 2025-03-07
feat: (storagebatchoperations)  added new resource storage batch operations

gurusai-voleti:

```release-note:new-resource
`google_storage_batch_operations_job`
```


gurusai-voleti:
@trodge  could you help enable SBO API on test project please to run the tests?

gurusai-voleti:
@trodge all the test cases related to the change are successful, please review

the failing test cases are unrelated to this change

mmv1/third_party/terraform/services/storagebatchoperations/resource_storage_batch_operations_job_test.go
```diff
@@ -0,0 +1,496 @@
...
+import (
+	"fmt"
+	"regexp"
+	"testing"
+
+	"github.com/hashicorp/terraform-plugin-testing/helper/resource"
+	"github.com/hashicorp/terraform-provider-google/google/acctest"
+)
+
+func TestAccStorageBatchOperationsJobs_storageBatchOperationsError(t *testing.T) {
```
kautikdk, line 18: What's the need of this test? We usually don't add error tests in the acceptance testing until we are having a very customized implementation.

mmv1/third_party/terraform/services/storagebatchoperations/resource_storage_batch_operations_job_test.go
```diff
@@ -0,0 +1,496 @@
...
+	return fmt.Sprintf(`
+resource "google_storage_bucket" "bucket" {
+  name     = "%s"
+  location = "us-central1"
+  uniform_bucket_level_access = true
+  force_destroy = true
+}
+resource "google_storage_batch_operations_job" "job" {
+	job_id     = "%s"
+	location = "global"
```
kautikdk, line 204: Spacing can be reformatted in the test configs.

mmv1/third_party/terraform/services/storagebatchoperations/resource_storage_batch_operations_job_test.go
```diff
@@ -0,0 +1,496 @@
...
+				ResourceName:            "google_storage_batch_operations_job.job",
+				ImportState:             true,
+				ImportStateVerify:       true,
+				ImportStateVerifyIgnore: []string{"job_id", "location", "delete_protection"},
+			},
+		},
+	})
+}
+
+func TestAccStorageBatchOperationsJobs_jobWithPrefixDeleteLiveObject(t *testing.T) {
```
kautikdk, line 84: I believe, the above test is testing the same structured config. Do we need this to be tested separately with different test-case?

mmv1/products/storagebatchoperations/Job.yaml
```diff
@@ -0,0 +1,275 @@
...
+        type: Boolean
+        description: 'enable flag to permanently delete object and all object versions if versioning is enabled on bucket'
+        immutable: true
+        required: true
+    exactly_one_of:
+      - 'delete_object'
+      - 'put_metadata'
+      - 'rewrite_object'
+      - 'put_object_hold'
+    custom_expand: 'templates/terraform/custom_expand/storage_batch_operations_job_delete_object.go.tmpl'
```
kautikdk, line 161: Why do we need custom expand and custom flatten for this field?

mmv1/third_party/terraform/.teamcity/components/inputs/services_beta.kt
```diff
@@ -766,6 +766,11 @@ var ServicesListBeta = mapOf(
         "displayName" to "Storage",
         "path" to "./google-beta/services/storage"
     ),
+    "storagebatchoperations" to mapOf(
+        "name" to "storagebatchoperations",
+        "displayName" to "StorageBatchJobOperations",
```
kautikdk, line 771: Should display name be "StorageBatchOperations"? Ditto for other file.

kautikdk (commented):
Instead of using custom delete and custom update, can we utilize pre delete and pre update?

mmv1/third_party/terraform/services/storagebatchoperations/resource_storage_batch_operations_job_test.go
gurusai-voleti: there is change in flag permanent_object_deletion_enabled true and false

mmv1/products/storagebatchoperations/Job.yaml
gurusai-voleti: here if we set the permanent_object_deletion_enabled as false the auto generated code is not working as expected so written an custom flatten and expansion

gurusai-voleti:
> Instead of using custom delete and custom update, can we utilize pre delete and pre update?

entire delete and update are overridden so I think better to use custom instead of pre

mmv1/third_party/terraform/services/storagebatchoperations/resource_storage_batch_operations_job_test.go
gurusai-voleti: removed the test case

mmv1/third_party/terraform/services/storagebatchoperations/resource_storage_batch_operations_job_test.go
gurusai-voleti: updated

mmv1/third_party/terraform/services/storagebatchoperations/resource_storage_batch_operations_job_test.go
kautikdk: But that may not require separate test rather can be added in the previous test as new step. This is to maintain readability and we don't end up having a test suite that are hard to understand for future changes.

kautikdk:
> > Instead of using custom delete and custom update, can we utilize pre delete and pre update?
> 
> entire delete and update are overridden so I think better to use custom instead of pre

What is the need to override entire update and delete? Is there any API behavior which is forcing us write full custom code?

mmv1/products/storagebatchoperations/Job.yaml
```diff
@@ -0,0 +1,275 @@
...
+---
+name: 'Job'
+kind: 'storagebatchoperations#jobs'
+id_format: 'projects/{{project}}/locations/{{location}}/jobs/{{job_id}}'
+base_url: 'projects/{{project}}/locations/{{location}}/jobs'
+self_link: 'projects/{{project}}/locations/{{location}}/jobs/{{job_id}}'
+create_url: 'projects/{{project}}/locations/{{location}}/jobs?jobId={{job_id}}'
+import_format:
+  - 'projects/{{project}}/locations/{{location}}/jobs/{{job_id}}'
+timeouts:
```
kautikdk, line 22: Do we have any SLO documentation for this timeout values?

mmv1/products/storagebatchoperations/Job.yaml
```diff
@@ -0,0 +1,275 @@
...
+      The timestamp at which this storage batch operation was most recently updated.
+    output: true
+  - name: 'jobId'
+    type: String
+    description: 'The ID of the job.'
+    url_param_only: true
+    immutable: true
+    validation:
+      regex: '^[a-z0-9]([-a-z0-9]*[a-z0-9])?.$'
+  - name: 'location'
```
kautikdk, line 74: I think, we are not allowing users to specify this field rather it is global for every operation.

mmv1/products/storagebatchoperations/Job.yaml
```diff
@@ -0,0 +1,275 @@
...
+    description: |
+      The timestamp at which this storage batch operation was most recently updated.
+    output: true
+  - name: 'jobId'
+    type: String
+    description: 'The ID of the job.'
+    url_param_only: true
+    immutable: true
+    validation:
+      regex: '^[a-z0-9]([-a-z0-9]*[a-z0-9])?.$'
```
kautikdk, line 73: Do we need this client side validation?

mmv1/products/storagebatchoperations/Job.yaml
```diff
@@ -0,0 +1,275 @@
...
+    description: |
+      The time that the job was completed.
+    output: true
+  - name: 'state'
+    type: String
+    description: |
+      State of the job.
+    output: true
+properties:
+  - name: 'bucketList'
```
kautikdk, line 96: Is this required field?

mmv1/products/storagebatchoperations/Job.yaml
```diff
@@ -0,0 +1,275 @@
...
+        description: contains object_prefix and bucket
+        immutable: true
+        item_type:
+          type: NestedObject
+          properties:
+            - name: 'bucket'
+              type: string
+              required: true
+              immutable: true
+              description: 'bucket on which operations performed'
```
kautikdk, line 116: "Bucket name for the objects to be transformed."

mmv1/products/storagebatchoperations/Job.yaml
```diff
@@ -0,0 +1,275 @@
...
+    description: List of buckets and their objects to be transformed. Currently, only one bucket configuration is supported. If multiple buckets are specified, an error will be returned
+    immutable: true
+    min_size: 1
+    max_size: 1
+    properties:
+      - name: 'buckets'
+        type: Array
+        min_size: 1
+        max_size: 1
+        description: contains object_prefix and bucket
```
kautikdk, line 107: You can follow API field description for better readability: https://cloud.google.com/storage/docs/storagebatchoperations/reference/rest/v1/projects.locations.jobs#BucketList

mmv1/products/storagebatchoperations/Job.yaml
```diff
@@ -0,0 +1,275 @@
...
+              type: string
+              required: true
+              immutable: true
+              description: 'bucket on which operations performed'
+            - name: 'prefixList'
+              type: NestedObject
+              immutable: true
+              min_size: 1
+              max_size: 1
+              description: 'bucket on which operations performed'
```
kautikdk, line 122: Description is not aligning with the field.

mmv1/products/storagebatchoperations/Job.yaml
```diff
@@ -0,0 +1,275 @@
...
+                  item_type:
+                    type: String
+            - name: 'manifest'
+              type: NestedObject
+              immutable: true
+              required: false
+              exactly_one_of:
+                - 'bucket_list.0.buckets.0.prefix_list'
+                - 'bucket_list.0.buckets.0.manifest'
+              description: 'contain the manifest source file that is a CSV file in a Google Cloud Storage bucket'
```
kautikdk, line 140: Ditto as suggested above for the description.

mmv1/products/storagebatchoperations/Job.yaml
```diff
@@ -0,0 +1,275 @@
...
+                  immutable: true
+                  description: 'list of objects to perform operations in csv format on storage bucket'
+  - name: 'deleteObject'
+    type: NestedObject
+    description: 'allows batch operations to delete objects in bucket'
+    immutable: true
+    properties:
+      - name: 'permanentObjectDeletionEnabled'
+        type: Boolean
+        description: 'enable flag to permanently delete object and all object versions if versioning is enabled on bucket'
```
kautikdk, line 153: nit: "...bucket."

kautikdk:
> > > > Instead of using custom delete and custom update, can we utilize pre delete and pre update?
> > > 
> > > 
> > > entire delete and update are overridden so I think better to use custom instead of pre
> > 
> > 
> > What is the need to override entire update and delete? Is there any API behavior which is forcing us write full custom code?
> 
> all the fields are immutable and delete_protection field is only update able so customised the code for update

So can we achieve similar behavior with `pre_update` and `pre_delete`?

kautikdk (commented):
One general feedback would to use period/fullstop to indicate completion of field description for all fields.

mmv1/products/storagebatchoperations/Job.yaml
gurusai-voleti: we don't have any, added default times for now

mmv1/products/storagebatchoperations/Job.yaml
gurusai-voleti: yes will add default value as global, and remove required

mmv1/products/storagebatchoperations/Job.yaml
gurusai-voleti: min_size is set to 1  and max_size it will suffice the required, will add required: true also

mmv1/products/storagebatchoperations/Job.yaml
gurusai-voleti: updated description

mmv1/products/storagebatchoperations/Job.yaml
gurusai-voleti: updated

mmv1/products/storagebatchoperations/Job.yaml
gurusai-voleti: updated description

mmv1/products/storagebatchoperations/Job.yaml
gurusai-voleti: updated description

gurusai-voleti:
> > > > > Instead of using custom delete and custom update, can we utilize pre delete and pre update?
> > > > 
> > > > 
> > > > entire delete and update are overridden so I think better to use custom instead of pre
> > > 
> > > 
> > > What is the need to override entire update and delete? Is there any API behavior which is forcing us write full custom code?
> > 
> > 
> > all the fields are immutable and delete_protection field is only update able so customised the code for update
> 
> So can we achieve similar behavior with `pre_update` and `pre_delete`?

we can but along with it we get auto generated code for update which we dont need

mmv1/products/storagebatchoperations/Job.yaml
kautikdk: Ahh, got it!

kautikdk:
> > > > > > Instead of using custom delete and custom update, can we utilize pre delete and pre update?
> > > > > 
> > > > > 
> > > > > entire delete and update are overridden so I think better to use custom instead of pre
> > > > 
> > > > 
> > > > What is the need to override entire update and delete? Is there any API behavior which is forcing us write full custom code?
> > > 
> > > 
> > > all the fields are immutable and delete_protection field is only update able so customised the code for update
> > 
> > 
> > So can we achieve similar behavior with `pre_update` and `pre_delete`?
> 
> we can but along with it we get auto generated code for update which we dont need

So we get unnecessary code for update because most of the fields are immutable but what is the issue `delete`? We prefer to keep code autogenerated as long as we can because of the easy maintenance benefit. Adding two methods with custom code will require us to individually maintain these two files for any future changes rather we can try to avoid it if possible.

mmv1/products/storagebatchoperations/Job.yaml
kautikdk: I don't understand the issue. What is the issue with expander? and what API behavior is causing us to write custom flattener? Can you please explain the issue in a little bit more detail because the autogeneration code should not be value specific and it there is a bug in autogen tool then we can create a ticket right away.

gurusai-voleti:
> > > > > > > Instead of using custom delete and custom update, can we utilize pre delete and pre update?
> > > > > > 
> > > > > > 
> > > > > > entire delete and update are overridden so I think better to use custom instead of pre
> > > > > 
> > > > > 
> > > > > What is the need to override entire update and delete? Is there any API behavior which is forcing us write full custom code?
> > > > 
> > > > 
> > > > all the fields are immutable and delete_protection field is only update able so customised the code for update
> > > 
> > > 
> > > So can we achieve similar behavior with `pre_update` and `pre_delete`?
> > 
> > 
> > we can but along with it we get auto generated code for update which we dont need
> 
> So we get unnecessary code for update because most of the fields are immutable but what is the issue `delete`? We prefer to keep code autogenerated as long as we can because of the easy maintenance benefit. Adding two methods with custom code will require us to individually maintain these two files for any future changes rather we can try to avoid it if possible.

we have delete_protection field in delete which required to make it custom

mmv1/products/storagebatchoperations/Job.yaml
gurusai-voleti: when the flag is set to false, in autogenerated code there is some logic which is  ignoring its value and empty map being sent

https://github.com/GoogleCloudPlatform/magic-modules/blob/a507efe4277d57b06383d0d2d93e88f2d0fbd856/mmv1/templates/terraform/custom_expand/storage_batch_operations_job_delete_object.go.tmpl#L25 

I debugged and removed some  logic over here when I created it in both flattener and expander

mmv1/third_party/terraform/services/storagebatchoperations/resource_storage_batch_operations_job_test.go
gurusai-voleti: as all the fields are immutable, a seperate job should be created for each of the case so seperated the test

gurusai-voleti: updated test case into one test as suggested

mmv1/templates/terraform/custom_delete/storage_batch_operations_job.go.tmpl
```diff
@@ -0,0 +1,3 @@
+if d.Get("delete_protection").(bool) {
```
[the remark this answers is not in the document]
gurusai-voleti: updated

mmv1/products/storagebatchoperations/Job.yaml
trodge: Not sending empty objects is the default magic modules behavior. There is a send empty flag when different behavior is needed

gurusai-voleti: added send_empty_value and removed custom expansion, working as expected

pavandagagoogle:
@trodge - Can you review? This is a feature which is already GA

mmv1/products/storagebatchoperations/Job.yaml
```diff
@@ -0,0 +1,289 @@
...
+custom_code:
+  custom_update: 'templates/terraform/custom_update/storage_batch_operations_job.go.tmpl'
+  pre_delete: 'templates/terraform/pre_delete/storage_batch_operations_job.go.tmpl'
+virtual_fields:
+  - name: 'delete_protection'
+    description: |
+      If set to `true`, the storage batch operation job will not be deleted and new job will be created.
+    type: Boolean
+    default_value: true
+parameters:
```
kautikdk, line 57: nit: I believe fields with `url_param_only` should be specified in this section, other fields should be added in properties section. Ultimately both are same but this is the kind of pattern we follow. @trodge can you please confirm this?

mmv1/third_party/terraform/services/storagebatchoperations/resource_storage_batch_operations_job_test.go
```diff
@@ -0,0 +1,379 @@
...
+		ProtoV5ProviderFactories: acctest.ProtoV5ProviderFactories(t),
+		Steps: []resource.TestStep{
+			{
+				Config: testAccStorageBatchOperationsCreateJobWithPrefix(bucketName, jobID),
+			},
+			{
+				ResourceName:            "google_storage_batch_operations_job.job",
+				ImportState:             true,
+				ImportStateVerify:       true,
+				ImportStateVerifyIgnore: []string{"job_id", "location", "delete_protection"},
```
kautikdk, line 28: nit: we don't need `location` here as we removed this field.

kautikdk (approved):
LGTM, Left some comments but shouldn't be blocker for the merge.

mmv1/third_party/terraform/services/storagebatchoperations/resource_storage_batch_operations_job_test.go
gurusai-voleti: updated

kautikdk (approved):

trodge (approved):

PR #13286, opened 2025-03-07
Fix integrations tests to bootstrap clients and kms

rohitjangid:

Fix integrations flaky tests

```release-note: none
integrations: Fixes flaky tests by bootstrapping kms and gmek client
```

Fixes: https://github.com/hashicorp/terraform-provider-google/issues/21245
Fixes: https://github.com/hashicorp/terraform-provider-google/issues/21246
Fixes: https://github.com/hashicorp/terraform-provider-google/issues/21511

### DIFF
--- a/mmv1/third_party/terraform/acctest/bootstrap_test_utils.go.tmpl
+++ b/mmv1/third_party/terraform/acctest/bootstrap_test_utils.go.tmpl
@@ -33,6 +33,7 @@ import (
 	cloudresourcemanager "google.golang.org/api/cloudresourcemanager/v1"
 	iam "google.golang.org/api/iam/v1"
 	"google.golang.org/api/iamcredentials/v1"
+	"google.golang.org/api/integrations/v1"
 	"google.golang.org/api/servicenetworking/v1"
 	"google.golang.org/api/serviceusage/v1"
 	sqladmin "google.golang.org/api/sqladmin/v1beta4"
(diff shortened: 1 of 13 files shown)


mmv1/third_party/terraform/acctest/bootstrap_test_utils.go.tmpl
```diff
@@ -2194,3 +2195,28 @@ func BootstrapSharedTestTagValue(t *testing.T, testId string, tagKey string) str
...
+	config := BootstrapConfig(t)
+	if config == nil {
+		return BootstrapClient{}
+	}
+
+	projectID := envvar.GetTestProjectFromEnv()
+	parent := fmt.Sprintf("projects/%s/locations/%s", projectID, locationID)
+
+	integrationsClient := config.NewIntegrationsClient(config.UserAgent)
+	request := &integrations.GoogleCloudIntegrationsV1alphaProvisionClientRequest{}
```
slevenick, line 2213: Adding a dependency on an alpha client scares me. Can we do this without using a client at all? The function directly above this has an example of using a direct HTTP call

slevenick (commented):
Looks like the tests are failing due to permission errors on a service account. Do we need specific permissions to be able to bootstrap these keys?

mmv1/third_party/terraform/acctest/bootstrap_test_utils.go.tmpl
rohitjangid: our v1alpha is GA so it should be fine. However I have replaced it wht http call considering the golang client is reflecting older version.

rohitjangid:
@slevenick Can you help me understand why other tests are failing? They should not have been impacted by my change. Also I am not sure what is failing in VCR test.

slevenick (approved):

slevenick:
Failing tests are unrelated

PR #13289, opened 2025-03-07
Fix google_service_networking_peered_dns_domain docs

rileykarson:
Fix bad preamble/header, and add API ref

```release-note:none
```


melinath (approved):
The file also shouldn't have a google_ prefix but it's probably fine.

rileykarson:
Yeah- not worth fixing on its own imo, particularly since it breaks links, maybe if we batched in a major version or could do redirects.

PR #13297, opened 2025-03-08
fix issue-19051 max retention period for Storage bucket

mjogdand:
Fixes https://github.com/hashicorp/terraform-provider-google/issues/19051

```release-note: bug
storage: fixed retention_period in `google_storage_bucket`
```

### DIFF
--- a/mmv1/third_party/terraform/services/storage/resource_storage_bucket.go.tmpl
+++ b/mmv1/third_party/terraform/services/storage/resource_storage_bucket.go.tmpl
@@ -6,7 +6,6 @@ import (
 	"errors"
 	"fmt"
 	"log"
-	"math"
 	"regexp"
 	"runtime"
 	"strconv"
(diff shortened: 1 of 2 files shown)


mjogdand:
@roaks3 Can you please advice on the change?

mjogdand:
@roaks3 - Sorry to trouble you but can you please review and approve this PR?

roaks3 (commented):
@mjogdand unfortunately, we won't be able to accept this change as-is, see https://github.com/hashicorp/terraform-provider-google/issues/19051#issuecomment-2725537669. You could go ahead and convert the field to a string, but we won't be able to merge a change like that until ~August (the next major release).

mjogdand:
> @mjogdand unfortunately, we won't be able to accept this change as-is, see [hashicorp/terraform-provider-google#19051 (comment)](https://github.com/hashicorp/terraform-provider-google/issues/19051#issuecomment-2725537669). You could go ahead and convert the field to a string, but we won't be able to merge a change like that until ~August (the next major release).

Hello @roaks3 - Thank you the review, regarding updating the field type to string, I have a one query.
consider we have updated field type to string and we set retention_period to "3155760000" the can you please guide me how we will compare the strings.

in general to compare strings we use 
1. len function 
2. strings.compare() method -- it compares two strings lexicographically.
3. parse string to int -- if number is greater than 2147483647 then it will throw error while parsing to int.

Kindly request you to guide me on this what approach we can take here to compare the strings.

PR #13299, opened 2025-03-09
Increase range of CloudSQL Query Insights string length & plans per minute to match Enterprise Plus limits

jbuck:
After following [the docs](https://cloud.google.com/sql/docs/mysql/using-query-insights#enable-enterprise-plus-query-insights) to increase the CloudSQL Query Insights string length and plan per minute limits we found that the Terraform provider's validation function hasn't been increased to match the new-ish Enterprise Plus limits. This PR modifies the validation function to account for the Enterprise Plus limits.

I see in the Google docs that this is marked as a preview feature. I'm not sure if that means these increased limits should be a beta-only change, or if it's okay to have it in both. I also wonder if changing the validation function based on the edition of the instance is a better idea?

Anyways, let me know what you think and I'm happy to update this PR to match

```release-note:enhancement
sql: Increased `settings.insights_config.query_string_length` and `settings.insights_config.query_string_length` limits for Enterprise Plus edition `sql_database_instance` resource.
```


SirGitsalot (approved):
The failing test is unrelated, it's been failing in nightly for a couple of weeks with the same error.

PR #13300, opened 2025-03-09
Add secret_manager_stored_password field to Datastream connection profiles

yardenvan:

```release-note:enhancement
datastream: added `secret_manager_stored_password` field to `google_datastream_connection_profile` resource
```


roaks3 (commented):
Code seems fine, but would it be feasible to add this to a test somewhere, to confirm it works as expected?

roaks3 (changes requested):
So as written, the new examples here wouldn't be tested because they don't have an associated `examples` block in the yaml.

However, I would suggest not added 4 new tests for this one field, for test performance reasons, unless you have to. Are there any existing tests where this could be added instead?

yardenvan:
I don't think that any of our exisiting tests can be used here because they actually test against our backend and this will require creating the relevant secret manager resource. Since we don't want to do that, I left only one test which uses the "create_without_validation" flag to bypass that (we only want to test the validity of terraform here).
Also added the test to the examples block.

mmv1/products/datastream/ConnectionProfile.yaml
```diff
@@ -169,9 +173,12 @@ properties:
         type: String
         description: |
           Password for the Oracle connection.
-        required: true
         sensitive: true
         custom_flatten: 'templates/terraform/custom_flatten/datastream_connection_profile_oracle_profile_password.go.tmpl'
+      - name: 'secret_manager_stored_password'
```
roaks3, line 178: Whoops, missed this earlier. The names should be camel cased ie. `secretManagerStoredPassword`

roaks3 (changes requested):

mmv1/products/datastream/ConnectionProfile.yaml
yardenvan: Fixed it, thanks!

PR #13314, opened 2025-03-11
[AlloyDB] External Deny Periods Support

vmiglani:
Description:
Supporting External Deny Periods in Terraform.

Issue - https://b.corp.google.com/issues/338101611

```release-note:enhancement
alloydb: added `maintenanceUpdatePolicy.denyMaintenancePeriods ` field to `google_alloydb_cluster` resource
```


mmv1/products/alloydb/Cluster.yaml
```diff
@@ -529,6 +529,7 @@ properties:
     type: NestedObject
     description: |
       MaintenanceUpdatePolicy defines the policy for system updates.
+    default_from_api: true
```
SirGitsalot, line 532: Is this `default_from_api` necessary? The new test always has this field set so it's not necessary there, and I would think that existing cluster tests would be failing if it was needed.

SirGitsalot (commented):
Is `denyMaintenancePeriods` fully rolled out for the GA API? I see that the earlier test failures were "Maintenance deny periods are not enabled", and it's not documented for GA (but [is for beta](https://cloud.google.com/alloydb/docs/reference/rest/v1beta/projects.locations.clusters#denymaintenanceperiod))

mmv1/products/alloydb/Cluster.yaml
vmiglani: Also, yes, the feature has been fully rolled out and is GA. It seems there is a delay in client documentation where something was published for beta first before it gets published to GA.

However, please find, gCloud reference - https://cloud.google.com/sdk/gcloud/reference/alloydb/clusters/create#--deny-maintenance-period-start-date

SirGitsalot (changes requested):

PR #13315, opened 2025-03-11
Feat/function cloud nat dyna port

Maarc-D:
This function will help setup correct value for cloud nat dynamic port that must be pow of 2, used here : https://registry.terraform.io/providers/public-cloud-wl/tools/latest/docs/functions/GetClosestInferiorGCPCloudNatDynamicPortFunction

https://github.com/public-cloud-wl/terraform-provider-tools/blob/main/internal/provider/get_closest_inferior_gcp_cloud_nat_dynamic_port_function.go

And I think it is a good idea to backport it here ;) 

```release-note:none
function: added `cloud_nat_dynamic_port_from_number` function
```


Maarc-D:
@SirGitsalot can I let you fix the last issue for unit test ? the setup here is a bit mindfuck seems I did not understand something :/

not as easy as a classic provider ;)

PR #13317, opened 2025-03-11
Add FutureReservation resource to Compute

JaylonmcShan03:
This PR adds support for the FutureReservation resource for Compute Engine. Future Reservations allow users to reserve Compute Engine capacity for a future time window

```release-note:new-resource
`google_compute_future_reservation`
```


roaks3:
* The `terraform-provider-missing-service-labels` check is fine to ignore, we can handle that on our end.
* For the 400 error, I'm not sure for this specific resource. Unfortunately, it is fairly difficult to tell what fields are supported, especially if they require another field to be set a specific way. Normally, in a case like yours, you'd need some trial-and-error to determine what is available from Terraform (ideally in a local dev environment). Worst case, if that can't be done or becomes too cumbersome, you could omit the offending fields to be added in future contributions (the downside being that we could be missing coverage of these fields in the meantime).

roaks3:
I'll trigger the tests, but note that the plan check is used for update tests to ensure that the resource is in fact updated in-place (and not replaced by TF). Normally, if you can't use the plan check, it means that you are updating fields with ForceNew (which shouldn't be done if you are trying to test update).

roaks3:
@JaylonmcShan03 it looks like you solved this with the update_encoder. I did dig a bit deeper because I was surprised you needed this, and it looks like the generator assumes that when PATCH is used immutable fields should not be included in the body (which does generally make sense). So I think basically you hit a strange edge case, and the encoder is the correct way to correct the behavior.

How is this going otherwise? I'm seeing tests passing now, does that mean it's complete and ready for a review?

JaylonmcShan03:
@roaks3 Yes, it is now ready for review :)

mmv1/templates/terraform/examples/shared_future_reservation.tf.tmpl
```diff
@@ -0,0 +1,53 @@
...
+}
+
+resource "google_project" "guest_project" {
+  project_id      = "tf-test-2%{random_suffix}"
+  name            = "tf-test-2%{random_suffix}"
+  org_id          = "{{index $.TestEnvVars "org_id"}}"
+  deletion_policy = "DELETE"
+}
+
+resource "google_organization_policy" "shared_future_reservation_org_policy" {
```
roaks3, line 22: I suspect that we'll need to work out how we want to set up this policy for any test environments that run this test. The main thing is that our PR tests and multiple nightly tests fall under the same org, so modifying it in the tests can lead to unintended side-effects. Perhaps this is isolated enough though because of the test-specific project id being used.

mmv1/third_party/terraform/services/compute/resource_compute_future_reservation_test.go.tmpl
```diff
@@ -0,0 +1,99 @@
...
+  provider = google-beta
+  name     = "tf-fr%{random_suffix}"
+  name_prefix = "fr-%{random_suffix}"
+  project  = "%{project}"
+  auto_delete_auto_created_reservations = true
+  description      = "test future reservation"
+  time_window {
+    start_time = "2025-08-01T00:00:00Z"
+    end_time   = "2025-08-02T00:00:00Z"
+  }
```
roaks3, line 64: It looks like there are a lot of fields not marked as immutable that could be added here. For example, `planning_status` is defined as a field that can be specified by the user, and `reservation_name` is defined to be mutable.

roaks3 (changes requested):
Some minor feedback in general, and then we'll need to work out the shared test

mmv1/products/compute/FutureReservation.yaml
```diff
@@ -0,0 +1,563 @@
...
+    description: |
+      Planning state before being submitted for evaluation
+    enum_values:
+      - 'DRAFT'
+      - 'PLANNING_STATUS_UNSPECIFIED'
+      - 'SUBMITTED'
+    default_from_api: true
+  - name: 'autoDeleteAutoCreatedReservations'
+    type: Boolean
+    ignore_read: true
```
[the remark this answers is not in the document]
JaylonmcShan03: Yes! autoDeleteAutoCreatedReservations fails during the` ImportStateVerify` and` post-apply `plan check due to a persistent diff without `ignore_read`. 

The API response omits it / defaults to false even though the field is explicitly set to true in the config!

mmv1/products/compute/FutureReservation.yaml
```diff
@@ -0,0 +1,563 @@
...
+      - name: 'startTime'
+        type: String
+        description: |
+          Start time of the future reservation in RFC3339 format.
+        required: true
+      - name: 'endTime'
+        type: String
+        description: |
+          End time of the future reservation in RFC3339 format.
+        required: true
```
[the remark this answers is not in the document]
JaylonmcShan03: Ah, that makes total sense — thanks for pointing it out! I agree that we should allow either endTime or duration, but keep startTime as required.

mmv1/templates/terraform/examples/shared_future_reservation.tf.tmpl
JaylonmcShan03: Thanks for pointing this out! I based this setup on the existing google_compute_reservation test, which also updates the constraints/compute.sharedReservationsOwnerProjects org policy using a test-scoped project approach similar to my configuration.

Since the config is leveraging a randomized tf-test%{random_suffix} project and the policy only whitelists that one project, I believe the impact radius will be very limited. The test also sets deletion_policy = "DELETE" to clean up properly.

Additionally, this test is skipped during PR runs due to skip_vcr: true, so it won’t affect the shared org in CI. It will only run in our nightly suite, just like the reservation resource tests.

Please let me know, if there is something I am not accounting for!

roaks3: Awesome, makes sense, thanks for explaining!

roaks3 (approved):
Sorry for the delay on this, the notifications caused it to get lost in my queue. LGTM

roaks3:
A little unfortunate here (depending how you look at it). It looks like we added a check that caught a new failure immediately before we were going to merge.

```
2025/05/16 18:14:23 Failed to find project environment variable defined in YAML file when validating the file templates/terraform/examples/future_reservation_basic.tf.tmpl. Please define this in test_env_vars
exit status 1
make: *** [GNUmakefile:72: mmv1] Error 1
```

PR #13318, opened 2025-03-11
Update Connection.yaml to have default from api for lock and log configs

arshitg-google:

```release-note:REPLACEME

```


NickElliot (commented):
these are both required fields, while default from api is equivalent to "optional + computed" which is to say these are conflicting field attributes. what is the reasoning for this attempted change?

PR #13325, opened 2025-03-11
Reconcile Dataproc Metastore Autoscaling Configs w/ API status

Noremac201:
Fixes https://github.com/hashicorp/terraform-provider-google/issues/21812

```release-note:enhancement
metastore: promoted `scaling_config` field to GA
```

### DIFF
--- a/mmv1/products/metastore/Service.yaml
+++ b/mmv1/products/metastore/Service.yaml
@@ -135,22 +135,18 @@ examples:
       metastore_service_name: 'backup'
   - name: 'dataproc_metastore_service_autoscaling_max_scaling_factor'
     primary_resource_id: 'test_resource'
-    min_version: 'beta'
     vars:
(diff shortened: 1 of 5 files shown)


mmv1/templates/terraform/examples/.tmpl
```diff
@@ -0,0 +1,22 @@
+resource "google_dataproc_metastore_service" "{{$.PrimaryResourceId}}" {
```
roaks3, line 1: This looks like a stray file, or possibly a problem with the filename?

roaks3 (changes requested):
Overall LGTM except for the mentioned example file

roaks3 (approved):

roaks3:
I got a strange 404 on 2 of these tests when running locally, but the other 2 passed. Given the nature of the change, I'm going to assume this is ok to merge.

PR #13331, opened 2025-03-11
feat(bigquery): added security_mode option for google_bigquery_routine

sachinpro:
Resolves https://github.com/hashicorp/terraform-provider-google/issues/19221.

```release-note:enhancement
bigquery: added `security_mode` option for `google_bigquery_routine` resource
```

### DIFF
--- a/mmv1/products/bigquery/Routine.yaml
+++ b/mmv1/products/bigquery/Routine.yaml
@@ -236,11 +236,11 @@ properties:
   - name: 'securityMode'
     type: Enum
     description: Optional. The security mode of the routine, if defined. If not defined, the security mode is automatically determined from the routine's configuration.
-    ignore_read: true
     enum_values:
       - 'SECURITY_MODE_UNSPECIFIED'
       - 'DEFINER'
       - 'INVOKER'
+    diff_suppress_func: 'tpgresource.EmptyOrDefaultStringSuppress("SECURITY_MODE_UNSPECIFIED")'
   - name: 'sparkOptions'
     type: NestedObject
     description: |
--- a/mmv1/products/bigquery/Routine.yaml
+++ b/mmv1/products/bigquery/Routine.yaml
@@ -233,6 +233,14 @@ properties:
     description: If set to DATA_MASKING, the function is validated and made available as a masking function. For more information, see https://cloud.google.com/bigquery/docs/user-defined-functions#custom-mask
     enum_values:
       - 'DATA_MASKING'
+  - name: 'securityMode'
+    type: Enum
+    description: Optional. The security mode of the routine, if defined. If not defined, the security mode is automatically determined from the routine's configuration.
+    ignore_read: true
+    enum_values:
+      - 'SECURITY_MODE_UNSPECIFIED'
+      - 'DEFINER'
+      - 'INVOKER'
   - name: 'sparkOptions'
     type: NestedObject
     description: |
--- a/mmv1/products/bigquery/Routine.yaml
+++ b/mmv1/products/bigquery/Routine.yaml
@@ -237,10 +237,8 @@ properties:
     type: Enum
     description: Optional. The security mode of the routine, if defined. If not defined, the security mode is automatically determined from the routine's configuration.
     enum_values:
-      - 'SECURITY_MODE_UNSPECIFIED'
       - 'DEFINER'
       - 'INVOKER'
-    diff_suppress_func: 'tpgresource.EmptyOrDefaultStringSuppress("SECURITY_MODE_UNSPECIFIED")'
   - name: 'sparkOptions'
     type: NestedObject
     description: |
--- a/mmv1/templates/terraform/examples/bigquery_routine_basic.tf.tmpl
+++ b/mmv1/templates/terraform/examples/bigquery_routine_basic.tf.tmpl
@@ -7,6 +7,7 @@ resource "google_bigquery_routine" "sproc" {
   routine_id     = "{{index $.Vars "routine_id"}}"
   routine_type = "PROCEDURE"
   language = "SQL"
+  security_mode = "INVOKER"
   definition_body = "CREATE FUNCTION Add(x FLOAT64, y FLOAT64) RETURNS FLOAT64 AS (x + y);"
 }
 
--- a/mmv1/third_party/terraform/services/bigquery/resource_bigquery_routine_test.go
+++ b/mmv1/third_party/terraform/services/bigquery/resource_bigquery_routine_test.go
@@ -50,6 +50,7 @@ resource "google_bigquery_routine" "sproc" {
   routine_id     = "%s"
   routine_type = "SCALAR_FUNCTION"
   language = "SQL"
+  security_mode = "INVOKER"
   definition_body = "1"
 }
 `, dataset, routine)
@@ -66,6 +67,7 @@ resource "google_bigquery_routine" "sproc" {
   routine_id     = "%s"
   routine_type = "SCALAR_FUNCTION"
   language = "JAVASCRIPT"
+  security_mode = "DEFINER"
   definition_body = "CREATE FUNCTION multiplyInputs return x*y;"
   arguments {
     name = "x"


mmv1/products/bigquery/Routine.yaml
```diff
@@ -233,6 +233,14 @@ properties:
     description: If set to DATA_MASKING, the function is validated and made available as a masking function. For more information, see https://cloud.google.com/bigquery/docs/user-defined-functions#custom-mask
     enum_values:
       - 'DATA_MASKING'
+  - name: 'securityMode'
+    type: Enum
+    description: Optional. The security mode of the routine, if defined. If not defined, the security mode is automatically determined from the routine's configuration.
+    ignore_read: true
```
sachinpro, line 239: API wasn't returning this parameter so I had to use `ignore_read`.

```
POST /bigquery/v2/projects/<redacted>/datasets/tf_test_dataset_id8s1c4q7e6r/routines?alt=json HTTP/1.1
Host: bigquery.googleapis.com
User-Agent: Terraform/1.4.6 (+https://www.terraform.io) Terraform-Plugin-SDK/2.36.0 terraform-provider-google/acc
Content-Length: 326
Accept-Encoding: gzip

{
 "definitionBody": "CREATE FUNCTION Add(x FLOAT64, y FLOAT64) RETURNS FLOAT64 AS (x + y);",
 "language": "SQL",
 "routineReference": {
  "datasetId": "tf_test_dataset_id8s1c4q7e6r",
  "projectId": "<redacted>",
  "routineId": "tf_test_routine_id8s1c4q7e6r"
 },
 "routineType": "PROCEDURE",
 "securityMode": "SECURITY_MODE_UNSPECIFIED"
}

2025/03/11 22:18:09 [DEBUG] Google API Response Details:
HTTP/2.0 200 OK
Alt-Svc: h3=":443"; ma=2592000,h3-29=":443"; ma=2592000
Date: Tue, 11 Mar 2025 21:18:09 GMT
Server: ESF
Vary: Origin
Vary: X-Origin
Vary: Referer
X-Frame-Options: SAMEORIGIN
X-Xss-Protection: 0

{
  "etag": "0L2vYHpmABM6CxkbWG1ShA==",
  "routineReference": {
    "projectId": "<redacted>",
    "datasetId": "tf_test_dataset_id8s1c4q7e6r",
    "routineId": "tf_test_routine_id8s1c4q7e6r"
  },
  "routineType": "PROCEDURE",
  "creationTime": "1741727889311",
  "lastModifiedTime": "1741727889311",
  "language": "SQL",
  "definitionBody": "CREATE FUNCTION Add(x FLOAT64, y FLOAT64) RETURNS FLOAT64 AS (x + y);"
}
```

mmv1/templates/terraform/examples/bigquery_routine_basic.tf.tmpl
```diff
@@ -7,6 +7,7 @@ resource "google_bigquery_routine" "sproc" {
   routine_id     = "{{index $.Vars "routine_id"}}"
   routine_type = "PROCEDURE"
   language = "SQL"
+  security_mode = "SECURITY_MODE_UNSPECIFIED"
```
c2thorn, line 10: is `SECURITY_MODE_UNSPECIFIED` even a necessary value to include? For most enums it is the same as just removing the field entirely.
also, I believe this means we need to test a different value than `SECURITY_MODE_UNSPECIFIED` ?

mmv1/products/bigquery/Routine.yaml
c2thorn: okay, unfortunate but should be fine

c2thorn (changes requested):

mmv1/templates/terraform/examples/bigquery_routine_basic.tf.tmpl
sachinpro: True that it was functionally the same as not using the field.

I am testing with a non-default value `INVOKER` now. Although the test still doesn't validate the behaviour specific to [INVOKER](https://pkg.go.dev/google.golang.org/api/bigquery/v2#Routine.SecurityMode:~:text=defines%20it.%0A%09//-,%22INVOKER%22,-%2D%20The%20routine%20is) field.

c2thorn: > I am testing with a non-default value `INVOKER` now. Although the test still doesn't validate the behaviour specific to [INVOKER](https://pkg.go.dev/google.golang.org/api/bigquery/v2#Routine.SecurityMode:~:text=defines%20it.%0A%09//-,%22INVOKER%22,-%2D%20The%20routine%20is) field.

What do you mean by this? Is the example now a bad one to be recommending to users, since it will appear in the Terraform docs? Or is it still fine. 
If it is a "bad" configuration but acceptable for testing, we should have a separate test-only example

c2thorn: @sachinpro

mmv1/products/bigquery/Routine.yaml
wj-chen: Is it the same case even for a mode that's not the default `SECURITY_MODE_UNSPECIFIED`? Can we check with someone from the Routine's team to see if it's expected?

sachinpro: I just checked. It is indeed the issue only with "SECURITY_MODE_UNSPECIFIED". I have contacted the TL. Can we mention this issue as a 'note' and remove `ignore_read`?

c2thorn: You mean a note in the description?
Am I understanding correctly that the default `SECURITY_MODE_UNSPECIFIED` is the same as not returned at all by the API? Maybe we can remove ignore_read and add a diff suppress:
```
diff_suppress_func: 'tpgresource.EmptyOrDefaultStringSuppress("SECURITY_MODE_UNSPECIFIED")'
```
?

c2thorn (commented):
waiting for API owner response

c2thorn (commented):
resetting review counter

mmv1/templates/terraform/examples/bigquery_routine_basic.tf.tmpl
sachinpro: @c2thorn This is not a bad example per se, it is just not the default one. I am not sure what default behaviour is supported by Bigquery Routine based on the only [documentation](https://pkg.go.dev/google.golang.org/api/bigquery/v2#Routine.SecurityMode) I found.

In general, for any database, "INVOKER" is the recommended option which should be set. A good explanation for this is [here](https://pkg.go.dev/google.golang.org/api/bigquery/v2#Routine.SecurityMode).

mmv1/products/bigquery/Routine.yaml
sachinpro: Yes, you are correct.

I have now removed `ignore_read` and using `diff_suppress_func`. Thanks for suggesting the fix.

mmv1/products/bigquery/Routine.yaml
```diff
@@ -233,6 +233,14 @@ properties:
     description: If set to DATA_MASKING, the function is validated and made available as a masking function. For more information, see https://cloud.google.com/bigquery/docs/user-defined-functions#custom-mask
     enum_values:
       - 'DATA_MASKING'
+  - name: 'securityMode'
+    type: Enum
+    description: Optional. The security mode of the routine, if defined. If not defined, the security mode is automatically determined from the routine's configuration.
+    enum_values:
+      - 'SECURITY_MODE_UNSPECIFIED'
```
c2thorn, line 240: Will a user ever need to set `SECURITY_MODE_UNSPECIFIED`? Reading your previous comment, if it is the same as just removing the field, we can remove `SECURITY_MODE_UNSPECIFIED` and the diff suppress altogether.

c2thorn (changes requested):
Is this field updatable between Definer and invoker? If so we should test it with an [update test](https://googlecloudplatform.github.io/magic-modules/test/test/#add-an-update-test) either by creating a new one or adding values to an existing one. Here is the update test file: https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/third_party/terraform/services/bigquery/resource_bigquery_routine_test.go
If it is not updatable, we should add `immutable: true`

mmv1/products/bigquery/Routine.yaml
sachinpro: Yes.
I only added it for feature completeness based on the [documentation](https://pkg.go.dev/google.golang.org/api/bigquery/v2#Routine.SecurityMode). The server will technically support `SECURITY_MODE_UNSPECIFIED`, but for the user, it is same as not setting it.

We _can_ remove it as you said.

sachinpro: I have removed it.

sachinpro:
> Is this field updatable between Definer and invoker? If so we should test it with an [update test](https://googlecloudplatform.github.io/magic-modules/test/test/#add-an-update-test) either by creating a new one or adding values to an existing one. Here is the update test file: https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/third_party/terraform/services/bigquery/resource_bigquery_routine_test.go If it is not updatable, we should add `immutable: true`

It was updatable. I have modified the existing test.

wj-chen (approved):

c2thorn (approved):

PR #13347, opened 2025-03-13
`docs`: add `write_only` field in `field_references`

BBBmau:

```release-note:none

```

### DIFF
--- a/docs/content/reference/field.md
+++ b/docs/content/reference/field.md
@@ -98,6 +98,22 @@ Example:
 sensitive: true
 ```
 
+### `write_only`
+If true, the field is considered "write-only", which means that its value will
+be obscured in Terraform output as we all as not be stored in state. This field is meant to replace `sensitive` as it doesn't store the value in state.
+See [Ephemerality in Resources - Use Write-only arguments](https://developer.hashicorp.com/terraform/language/resources/ephemeral/write-only)
+for more information.
+
+This field cannnot be used in conjuction with `immutable` or `sensitive`.
+
+**Note**: Due to write-only not being read from the API, it is not possible to update the field directly unless a sidecar field is used. (e.g. `password` as a write-only field and `password_wo_version` as an immutable field meant for updating).
+
+Example:
+
+```yaml
+write_only: true
+```
+
 ### `ignore_read`
 If true, the provider sets the field's value in the resource state based only
 on the user's configuration. If false or unset, the provider sets the field's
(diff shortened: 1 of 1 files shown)


docs/content/reference/field.md
```diff
@@ -98,6 +98,22 @@ Example:
 sensitive: true
 ```
 
+### `write_only`
+If true, the field is considered "write-only", which means that its value will
+be obscured in Terraform output as we all as not be stored in state. This field is meant to replace `sensitive` as it doesn't store the value in state.
+See [Ephemerality in Resources - Use Write-only arguments](https://developer.hashicorp.com/terraform/language/resources/ephemeral/write-only)
+for more information.
+
+This field cannnot be used in conjuction with `immutable` or `sensitive`.
```
rileykarson, line 107:
```suggestion
This field cannot be used in conjuction with `immutable` or `sensitive`.
```

rileykarson: Isn't `immutable` fine? The resource will recreate if you increment the version. Or does the _version_ need to be the one that's immutable?

docs/content/reference/field.md
```diff
@@ -98,6 +98,22 @@ Example:
 sensitive: true
 ```
 
+### `write_only`
+If true, the field is considered "write-only", which means that its value will
+be obscured in Terraform output as we all as not be stored in state. This field is meant to replace `sensitive` as it doesn't store the value in state.
```
rileykarson, line 103:
```suggestion
be obscured in Terraform output as well as not be stored in state. This field is meant to replace `sensitive` as it doesn't store the value in state.
```

docs/content/reference/field.md
```diff
@@ -98,6 +98,22 @@ Example:
...
+See [Ephemerality in Resources - Use Write-only arguments](https://developer.hashicorp.com/terraform/language/resources/ephemeral/write-only)
+for more information.
+
```
rileykarson, line 106:
```suggestion

Write-only fields are only supported in Terraform v1.11+. Because the provider supports earlier Terraform versions, write only fields must be paired with (mutually exclusive) `sensitive` fields covering the same functionality for compatibility with those older versions.
```

docs/content/reference/field.md
BBBmau: the version field (the non-write-only field) would need to be immutable.

rileykarson (approved):

PR #13349, opened 2025-03-13
add pod_autoscaling field in google_container_cluster

translucens:
Fixes https://github.com/hashicorp/terraform-provider-google/issues/21727

```release-note:enhancement
container: added `pod_autoscaling` field to `google_container_cluster` resource
```

### DIFF
--- a/mmv1/third_party/terraform/services/container/resource_container_cluster.go.tmpl
+++ b/mmv1/third_party/terraform/services/container/resource_container_cluster.go.tmpl
@@ -2370,6 +2370,26 @@ func ResourceContainerCluster() *schema.Resource {
 					},
 				},
 			},
+{{ if ne $.TargetVersionName `ga` -}}
+			"pod_autoscaling": {
+				Type:        schema.TypeList,
+				Optional:    true,
+				Computed:    true,
+				Description: `The config for pod autoscaling.`,
+				MaxItems:    1,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"hpa_profile": {
+							Type:        schema.TypeString,
+							Optional:    true,
+							Computed:    true,
+							Description: `Horizontal Pod Autoscaling profile. See [HPAProfile](https://cloud.google.com/kubernetes-engine/docs/reference/rest/v1beta1/projects.locations.clusters#hpaprofile) for allowed values.`,
+							DiffSuppressFunc: tpgresource.EmptyOrDefaultStringSuppress("HPA_PROFILE_UNSPECIFIED"),
+						},
+					},
+				},
+			},
+{{- end }}
 		},
 	}
 }
@@ -2544,6 +2564,9 @@ func resourceContainerClusterCreate(d *schema.ResourceData, meta interface{}) er
 {{- end }}
 		CostManagementConfig: expandCostManagementConfig(d.Get("cost_management_config")),
 		EnableK8sBetaApis:   expandEnableK8sBetaApis(d.Get("enable_k8s_beta_apis"), nil),
+{{- if ne $.TargetVersionName "ga" }}
+		PodAutoscaling: expandPodAutoscaling(d.Get("pod_autoscaling")),
+{{- end }}
 	}
 
 	v:= d.Get("enable_shielded_nodes")
@@ -3258,6 +3281,12 @@ func resourceContainerClusterRead(d *schema.ResourceData, meta interface{}) erro
 	if err := d.Set("enterprise_config", flattenEnterpriseConfig(cluster.EnterpriseConfig)); err != nil {
 		return err
 	}
+
+{{ if ne $.TargetVersionName `ga` -}}
+	if err := d.Set("pod_autoscaling", flattenPodAutoscaling(cluster.PodAutoscaling)); err != nil {
+		return err
+	}
+{{- end }}
 
 	return nil
 }
@@ -4670,6 +4699,22 @@ func resourceContainerClusterUpdate(d *schema.ResourceData, meta interface{}) er
 		log.Printf("[INFO] GKE cluster %s's WorkloadALTSConfig has been updated", d.Id())
 	}
 {{- end }}
+{{- if ne $.TargetVersionName "ga" }}
+	if d.HasChange("pod_autoscaling") {
+		req := &container.UpdateClusterRequest{
+			Update: &container.ClusterUpdate{
+				DesiredPodAutoscaling: expandPodAutoscaling(d.Get("pod_autoscaling")),
+			},
+		}
+
+		updateF := updateFunc(req, "updating GKE cluster PodAutoscaling")
+		if err := transport_tpg.LockedCall(lockKey, updateF); err != nil {
+			return err
+		}
+
+		log.Printf("[INFO] GKE cluster %s's PodAutoscaling has been updated", d.Id())
+	}
+{{- end }}
 	return resourceContainerClusterRead(d, meta)
 }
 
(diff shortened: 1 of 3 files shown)


mmv1/third_party/terraform/services/container/resource_container_cluster.go.tmpl
```diff
@@ -2370,6 +2370,26 @@ func ResourceContainerCluster() *schema.Resource {
...
+{{ if ne $.TargetVersionName `ga` -}}
+			"pod_autoscaling": {
+				Type:        schema.TypeList,
+				Optional:    true,
+				Computed:    true,
+				Description: `The config for pod autoscaling.`,
+				MaxItems:    1,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"hpa_profile": {
```
melinath, line 2382: Would it make sense to add validation to make sure this only contains one of the valid enum values? (Generally that's good to do unless the list of valid values is expected to change in the future.)

https://pkg.go.dev/github.com/hashicorp/terraform-plugin-sdk/helper/validation#StringInSlice would be the func to use.

mmv1/third_party/terraform/services/container/resource_container_cluster.go.tmpl
```diff
@@ -2370,6 +2370,26 @@ func ResourceContainerCluster() *schema.Resource {
...
+							Type:        schema.TypeString,
+							Optional:    true,
+							Computed:    true,
```
melinath, line 2385: just want to double-check that the API returns a complex default? Optional + Computed has some drawbacks, documented at https://googlecloudplatform.github.io/magic-modules/develop/diffs/.

mmv1/third_party/terraform/services/container/resource_container_cluster.go.tmpl
```diff
@@ -2370,6 +2370,26 @@ func ResourceContainerCluster() *schema.Resource {
...
+							Description: `Horizontal Pod Autoscaling profile. See [HPAProfile](https://cloud.google.com/kubernetes-engine/docs/reference/rest/v1beta1/projects.locations.clusters#hpaprofile) for allowed values.`,
+							DiffSuppressFunc: tpgresource.EmptyOrDefaultStringSuppress("HPA_PROFILE_UNSPECIFIED"),
```
melinath, line 2387: Generally we don't expect users to specify UNSPECIFIED as a value; if the enum is expected to change in the future, you could still validate that they aren't setting UNSPECIFIED if it would cause diffs. That would be preferable to a diff suppress func, unless UNSPECIFIED has a special meaning and users need to be able to send it (which is generally not the case.)

mmv1/third_party/terraform/services/container/resource_container_cluster_test.go.tmpl
```diff
@@ -6297,6 +6297,83 @@ func testAccCheckContainerClusterDestroyProducer(t *testing.T) func(s *terraform
...
+			{
+				ResourceName:      "google_container_cluster.with_pod_autoscaling_config",
+				ImportState:       true,
+				ImportStateVerify: true,
+				ImportStateVerifyIgnore: []string{"deletion_protection"},
+				Check: resource.TestCheckResourceAttr(
+					"google_container_cluster.with_pod_autoscaling_config", "pod_autoscaling.hpa_profile", "HPA_PROFILE_UNSPECIFIED"),
+			},
+			{
+				Config: testAccContainerCluster_withPodAutoscalingConfig(pid, networkName, subnetworkName, clusterName, false),
```
melinath, line 6325: Update tests should include ConfigPlanChecks on all the Config steps except the first one, as described at https://googlecloudplatform.github.io/magic-modules/test/test/#add-an-update-test

mmv1/third_party/terraform/services/container/resource_container_cluster_test.go.tmpl
```diff
@@ -6297,6 +6297,83 @@ func testAccCheckContainerClusterDestroyProducer(t *testing.T) func(s *terraform
...
+			},
+			{
+				ResourceName:      "google_container_cluster.with_pod_autoscaling_config",
+				ImportState:       true,
+				ImportStateVerify: true,
+				ImportStateVerifyIgnore: []string{"deletion_protection"},
+				Check: resource.TestCheckResourceAttr(
+					"google_container_cluster.with_pod_autoscaling_config", "pod_autoscaling.hpa_profile", "NONE"),
+			},
+						{
```
melinath, line 6335: nit: whitespace

mmv1/third_party/terraform/services/container/resource_container_cluster_test.go.tmpl
```diff
@@ -13278,3 +13355,63 @@ resource "google_container_cluster" "with_enterprise_config" {
...
+  data "google_project" "project" {
+    provider = google-beta
+    project_id = "%s"
+  }
+  resource "google_container_cluster" "with_pod_autoscaling_config_autopilot" {
+    provider = google-beta
+    name               = "%s"
+    location           = "us-central1"
+    enable_autopilot = true
+    min_master_version = "1.32.1-gke.1729000"
```
melinath, line 13412: use the [google_container_engine_versions](https://registry.terraform.io/providers/hashicorp/google/latest/docs/data-sources/container_engine_versions) datasource instead of hardcoding a version; otherwise this test will break in the future when this version is removed.

mmv1/third_party/terraform/services/container/resource_container_cluster_test.go.tmpl
```diff
@@ -13278,3 +13355,63 @@ resource "google_container_cluster" "with_enterprise_config" {
...
+data "google_project" "project" {
+  provider   = google-beta
+  project_id = "%s"
+}
+resource "google_compute_network" "network" {
+  provider                 = google-beta
+  name                     = "%s"
+  auto_create_subnetworks  = false
+}
+resource "google_compute_subnetwork" "subnet" {
```
melinath, line 13381: Use a bootstrapped network & subnetwork here instead of creating new ones - search for BootstrapSharedTestNetwork and BootstrapSubnet in this file for examples.

mmv1/third_party/terraform/services/container/resource_container_cluster_test.go.tmpl
```diff
@@ -13278,3 +13355,63 @@ resource "google_container_cluster" "with_enterprise_config" {
...
+      hpa_profile = "NONE"
+    }
+  `
+  }
+
+  return fmt.Sprintf(`
+data "google_project" "project" {
+  provider   = google-beta
+  project_id = "%s"
+}
```
melinath, line 13375: unused
```suggestion
```

mmv1/third_party/terraform/services/container/resource_container_cluster_test.go.tmpl
```diff
@@ -13278,3 +13355,63 @@ resource "google_container_cluster" "with_enterprise_config" {
...
+`, projectID, networkName, subnetworkName, name, podAutoscalingBlock)
+}
+
+func testAccContainerCluster_withPodAutoscalingConfigAutopilot(projectID, name string) string {
+  return fmt.Sprintf(`
+  data "google_project" "project" {
+    provider = google-beta
+    project_id = "%s"
+  }
```
melinath, line 13406:
```suggestion
```

mmv1/third_party/terraform/services/container/resource_container_cluster.go.tmpl
```diff
@@ -6983,6 +7043,19 @@ func flattenWorkloadAltsConfig(c *container.WorkloadALTSConfig) []map[string]int
 }
 {{- end }}
 
+{{ if ne $.TargetVersionName `ga` -}}
+func flattenPodAutoscaling(c *container.PodAutoscaling) []map[string]interface{} {
```
melinath, line 7047: If the API is returning UNSPECIFIED when an empty string is sent (and this behavior is consistent - that is, doesn't change based on other factors) you can intercept here & convert it to an empty string rather than using a diff suppress; generally speaking, it's better to make that kind of change in a flattener if possible.

melinath (changes requested):

translucens:
The same field has been added #13359

PR #13358, opened 2025-03-14
Cancel the LRO if anything fails while waiting on LRO for anywhere_cache_resource

googlyrahman:
```release-note:enhancement
storage: made `google_storage_anywhere_cache` cancel the long-running operation after create and update requests timeout
```


c2thorn (commented):
Hi @googlyrahman, is this related to a buganizer issue or Github issue for context?
Also, is there a way to test this functionality?

mmv1/templates/terraform/custom_create/storage_anywhere_cache.go.tmpl
```diff
@@ -0,0 +1,116 @@
...
+var opRes map[string]interface{}
+err = StorageOperationWaitTimeWithResponse(
+    config, res, &opRes, "Creating AnywhereCache", userAgent,
+    d.Timeout(schema.TimeoutCreate))
+
+if err != nil {
+    // The resource didn't actually create
+    d.SetId("")
+
+    // If any error comes due to LRO, cancel the LRO
```
c2thorn, line 72: What's the behavior of the provider code before this change? Will it get stuck or result in a tainted resource?

googlyrahman: I'll explain:

The resource create, and update returns LRO which can take upto 48h, However under normal circumstances it should be created under 4h of time. We maintain a timeout of 4h at client side. Before this change, if it timeout happens for more than 4h, there would be orphaned LRO, and we would not cancel the LRO.

Hence this change enables cancellation of the LRO on any error while polling the LRO.

googlyrahman:
> Hi @googlyrahman, is this related to a buganizer issue or Github issue for context? Also, is there a way to test this functionality?

Context: The cache creation LRO (Long Running Operation) can take up to 48 hours to complete. However, based on discussions with the Terraform and backend teams, we understand that under normal circumstances, the cache should be created within 2-3 hours. Therefore, we have set a time limit of 4 hours. Additionally, if a bucket contains any caches (even those with pending LROs), it cannot be deleted.

Given this context, if a timeout occurs (typically due to a service outage), we would receive an error as expected.  In that case, currently user is expected to cancel the LRO manually, and then hit "terraform destory" which might be a bit of overhead to our customers.

To address this, the change ensures that if any error occurs, the LRO cancellation request is made. With a wait time of approximately 1 hour and 30 minutes on the bucket during the destroy process, the bucket can be deleted without any errors.

Let me know if you've any more questions!

googlyrahman:
> > Hi @googlyrahman, is this related to a buganizer issue or Github issue for context? Also, is there a way to test this functionality?
> 
> Context: The cache creation LRO (Long Running Operation) can take up to 48 hours to complete. However, based on discussions with the Terraform and backend teams, we understand that under normal circumstances, the cache should be created within 2-3 hours. Therefore, we have set a time limit of 4 hours. Additionally, if a bucket contains any caches (even those with pending LROs), it cannot be deleted.
> 
> Given this context, if a timeout occurs (typically due to a service outage), we would receive an error as expected. In that case, currently user is expected to cancel the LRO manually, and then hit "terraform destory" which might be a bit of overhead to our customers.
> 
> To address this, the change ensures that if any error occurs, the LRO cancellation request is made. With a wait time of approximately 1 hour and 30 minutes on the bucket during the destroy process, the bucket can be deleted without any errors.
> 
> Let me know if you've any more questions!


We already've few tests related to create and update of the resources which passes successfully, I just introduced a change where a LRO cancellation request is made if any failure happens, I'm unsure how to stimulate the same in tests, do you've any examples for the same?

googlyrahman:
cc: @melinath

c2thorn:
@googlyrahman Are you able to simulate a failure to verify the cancel code works as expected for create and destroy?

googlyrahman:
> @googlyrahman Are you able to simulate a failure to verify the cancel code works as expected for create and destroy?

Yes, I'm able to simulate the behavior, and it is working as expected. I adjusted the timeout to 2m, and since i know thatLRO completion is gonna take more than 2m, i would get a timeout error, and it should cancel the LRO in such situation, Here is the logs for the same (including config) - https://paste.googleplex.com/6413406712889344

Another scenario is the interrupt, Here is the logs for scenario in case if interrupt is received - https://paste.googleplex.com/6331367468826624

Let me know if anything else is required!

c2thorn (approved):

PR #13362, opened 2025-03-14
Fix cross-project listing subscriptions

c2thorn:
fixes https://github.com/hashicorp/terraform-provider-google/issues/21402

google_bigquery_analytics_hub_listing_subscription is unfortunately unique in that the create_url and self_link differ completely. What was missed in the original implementation was a cross-project setup that would cause one project needed for the create_url, but a separate one for the self_link.

This fix rectifies the issue by inserting pre_read and pre_delete code to overwrite the template URL to use the correct destination dataset project instead of the `project` field.

```release-note:bug
bigqueryanalyticshub: fixed a bug in `google_bigquery_analytics_hub_listing_subscription` where a subscription using a different project than the dataset would not work
```

### DIFF
--- a/mmv1/products/bigqueryanalyticshub/ListingSubscription.yaml
+++ b/mmv1/products/bigqueryanalyticshub/ListingSubscription.yaml
@@ -27,6 +27,8 @@ import_format:
   - 'projects/{{project}}/locations/{{location}}/subscriptions/{{subscription_id}}'
 custom_code:
   post_create: templates/terraform/post_create/analytics_hub_subscription.go.tmpl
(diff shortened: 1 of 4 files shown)


sachinpro:
Is this WIP? `TestAccBigqueryAnalyticsHubListingSubscription_differentProject` is failing.

c2thorn:
> Is this WIP? `TestAccBigqueryAnalyticsHubListingSubscription_differentProject` is failing.

It is failing on the import step which I still need to fix. Working on that now

hao-nan-li (approved):

c2thorn (commented):
All existing TestAccBigqueryAnalyticsHub tests passed in a manual Teamcity build

PR #13367, opened 2025-03-17
Add `locations` field to mirroring deployment group.

duvni:
Add the new `locations` output field to Mirroring Deployment Group resource.
Used to expose the locations where the deployment group is currently deployed.

```release-note:enhancement
networksecurity: added `locations` field to `google_network_security_mirroring_deployment_group` resource
```

### DIFF
--- a/mmv1/products/networksecurity/MirroringDeploymentGroup.yaml
+++ b/mmv1/products/networksecurity/MirroringDeploymentGroup.yaml
@@ -138,3 +138,26 @@ properties:
     description: |-
       User-provided description of the deployment group.
       Used as additional context for the deployment group.
+  - name: locations
+    type: Array
+    is_set: true
+    description: |-
+      The list of locations where the deployment group is present.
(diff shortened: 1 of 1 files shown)


mmv1/products/networksecurity/MirroringDeploymentGroup.yaml
```diff
@@ -138,3 +138,26 @@ properties:
     description: |-
       User-provided description of the deployment group.
       Used as additional context for the deployment group.
+  - name: locations
+    type: Array
+    is_set: true
```
BBBmau, line 143: is there a reason that it must be a set?

duvni: Yes, locations are unique and the back-end may return them in different order every time.
So to avoid diffs, I changed it to be a set.

BBBmau (approved):

PR #13369, opened 2025-03-17
Add `locations` field to mirroring endpoint group association.

duvni:
Add the new `locations` output field to Mirroring Endpoint Group Association resource.
Used to expose the locations where the producer is currently deployed.

```release-note:enhancement
networksecurity: added `locations` field to `google_network_security_mirroring_endpoint_group_association` resource
```

### DIFF
--- a/mmv1/products/networksecurity/MirroringEndpointGroupAssociation.yaml
+++ b/mmv1/products/networksecurity/MirroringEndpointGroupAssociation.yaml
@@ -119,6 +119,8 @@ properties:
       The list of locations where the association is present. This information
       is retrieved from the linked endpoint group, and not configured as part
       of the association itself.
+    deprecation_message: |-
+      `locationsDetails` is deprecated and will be removed in a future major release. Use `locations` instead.
     output: true
     item_type:
       type: NestedObject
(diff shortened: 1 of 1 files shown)


mmv1/products/networksecurity/MirroringEndpointGroupAssociation.yaml
```diff
@@ -158,3 +160,26 @@ properties:
       operation (e.g. adding a new location to the target deployment group).
       See https://google.aip.dev/128.
     output: true
+  - name: locations
+    type: Array
+    is_set: true
```
BBBmau, line 165: is there a reason why it needs to be a set?

mmv1/products/networksecurity/MirroringEndpointGroupAssociation.yaml
```diff
@@ -119,6 +119,8 @@ properties:
       The list of locations where the association is present. This information
       is retrieved from the linked endpoint group, and not configured as part
       of the association itself.
+    deprecation_message: |-
```
BBBmau, line 122: we'll want to include a conflicts field to prevent both from being set.

BBBmau (approved):

PR #13372, opened 2025-03-17
Add `connectedDeploymentGroups` and `associations` fields to mirroring endpoint group.

duvni:
Add the new `connectedDeploymentGroups` and `associations` output fields to Mirroring Endpoint Group resource.
These fields expose to the user more data on the connected deployment groups and the associated networks.

```release-note:enhancement
networksecurity: added `connectedDeploymentGroups` and `associations` fields to `google_network_security_mirroring_endpoint_group` resource
```

### DIFF
--- a/mmv1/products/networksecurity/MirroringEndpointGroup.yaml
+++ b/mmv1/products/networksecurity/MirroringEndpointGroup.yaml
@@ -164,11 +164,6 @@ properties:
             OUT_OF_SYNC
             DELETE_FAILED
           output: true
-  - name: description
-    type: String
-    description: |-
-      User-provided description of the endpoint group.
-      Used as additional context for the endpoint group.
   - name: connectedDeploymentGroups
     type: Array
     is_set: true
--- a/mmv1/products/networksecurity/MirroringEndpointGroup.yaml
+++ b/mmv1/products/networksecurity/MirroringEndpointGroup.yaml
@@ -128,3 +128,82 @@ properties:
     description: |-
       User-provided description of the endpoint group.
       Used as additional context for the endpoint group.
+  - name: associations
+    type: Array
+    is_set: true
+    description: |-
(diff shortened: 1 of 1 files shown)


mmv1/products/networksecurity/MirroringEndpointGroup.yaml
```diff
@@ -128,3 +128,77 @@ properties:
     description: |-
       User-provided description of the endpoint group.
       Used as additional context for the endpoint group.
+  - name: associations
+    type: Array
+    is_set: true
```
BBBmau, line 133: is there a reason why it needs to be a set?

mmv1/products/networksecurity/MirroringEndpointGroup.yaml
```diff
@@ -128,3 +128,77 @@ properties:
...
+            ACTIVE
+            CREATING
+            DELETING
+            CLOSED
+            OUT_OF_SYNC
+            DELETE_FAILED
+          output: true
+  - name: connectedDeploymentGroups
+    type: Array
+    is_set: true
```
BBBmau, line 169: same here

mmv1/products/networksecurity/MirroringEndpointGroup.yaml
duvni: The associations are unique (no duplicates) and the order they're returned from the back-end might change.
From previous issues I've experienced, if the order changes it creates a bogus diff, so I mad it a set to avoid that.

mmv1/products/networksecurity/MirroringEndpointGroup.yaml
duvni: Same as the other comment, the connected deployment groups are unique and order is not guranteed.

BBBmau (approved):

PR #13373, opened 2025-03-17
Add `connectedDeploymentGroup` and `associations` fields to intercept endpoint group.

duvni:
Add the new `connectedDeploymentGroup` and `associations` output fields to Intercept Endpoint Group resource.
These fields expose to the user more data on the connected deployment group and the associated networks.

```release-note:enhancement
networksecurity: added `connectedDeploymentGroup` and `associations` fields to `google_network_security_intercept_endpoint_group` resource
```

### DIFF
--- a/mmv1/products/networksecurity/InterceptEndpointGroup.yaml
+++ b/mmv1/products/networksecurity/InterceptEndpointGroup.yaml
@@ -183,6 +183,7 @@ properties:
     description: |-
       The endpoint group's view of a connected deployment group.
     min_version: 'beta'
+    output: true
     properties:
       - name: name
         type: String
--- a/mmv1/products/networksecurity/InterceptEndpointGroup.yaml
+++ b/mmv1/products/networksecurity/InterceptEndpointGroup.yaml
@@ -138,3 +138,82 @@ properties:
       User-provided description of the endpoint group.
       Used as additional context for the endpoint group.
     min_version: 'beta'
+  - name: associations
+    type: Array
+    is_set: true
+    description: |-
+      List of associations to this endpoint group.
+    min_version: 'beta'
+    output: true
+    item_type:
+      type: NestedObject
+      properties:
+        - name: name
+          type: String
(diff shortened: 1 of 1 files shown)


mmv1/products/networksecurity/InterceptEndpointGroup.yaml
```diff
@@ -138,3 +138,82 @@ properties:
...
+            STATE_UNSPECIFIED
+            ACTIVE
+            CREATING
+            DELETING
+            CLOSED
+            OUT_OF_SYNC
+            DELETE_FAILED
+          min_version: 'beta'
+          output: true
+  - name: connectedDeploymentGroup
```
c2thorn, line 181: should this parent field be output-only?

c2thorn (changes requested):
This is fairly larger than the last one, have you verified the fields show correctly in a manual test or running an existing one?

mmv1/products/networksecurity/InterceptEndpointGroup.yaml
duvni: Yes, users use the `intercept_deployment_group` field to set the name of the connected deployment group on creation. This field gives output-only data on the state of the connected deployment group and the zones where it is currently deployed.
While this could have been combined, it'll be a big breaking change to do it now so we just added a new output-only field to the API to accommodate for this.

duvni:
> This is fairly larger than the last one, have you verified the fields show correctly in a manual test or running an existing one?

Yes, I verified the output manually, but I ran the update test as well now and found that as you've noticed the output annotation was missing from the connected_deployment_group parent field and it caused the test to fail.
I fixed it and all related intercept resource tests are passing now.
Sorry about combining the two new output fields in one PR, I wanted to reduce the review toll as there are 5 other related PRs for other fields in other similar resources, so I figured that combining these two instead of chaining can be maybe help a bit.

c2thorn (approved):

PR #13213, opened 2025-02-27
Add bac to backend service

porky256:
```release-note:enhancement
compute: added `tls_settings` field to `google_compute_backend_service` resource
```

PR #13374, opened 2025-03-17
Fix compute.BackendService bug

porky256:
```release-note:bug
compute: fixed broken test in `google_compute_backend_service`.
```


porky256:
Fixes https://github.com/hashicorp/terraform-provider-google/issues/21884

porky256:
Turned out there is some blockers on GCP side that I was not aware of, and we can't introduce BackendService.tlsSettings yet. I'll revert this: https://github.com/GoogleCloudPlatform/magic-modules/pull/13213.

PR #13378, opened 2025-03-17
Revert "Add bac to backend service"

porky256:
```release-note:breaking-change
compute: remove tlsSettings from `google_compute_backend_service` (revert)
```
Reverts GoogleCloudPlatform/magic-modules#13213
Causing https://github.com/hashicorp/terraform-provider-google/issues/21884, and can't be fixed at the moment because of the blockers on the GCP side.


porky256:
I'm not sure about release note type since no clients is using this, but marking as breaking_change to follow guide.

porky256:
No customers is using this feature yet, so it's not a breaking change.

zli82016 (approved):
LGTM

`authentication_config` field depends on `google_network_security_backend_authentication_config` resource, which is only available in beta

zli82016:
I tested it locally, and the field does't work at all with the error
```
Invalid value for field 'resource.tlsSettings':
Tls Settings are not supported for global EXTERNAL backend services
```

porky256:
Yep, that is the reason why I want to roll this changes back. Once error you see will be fixed, I’ll open a new pr with this changes. Thanks for reviewing this!

PR #13384, opened 2025-03-18
add iam bootstraping support for organizations

yerniyazN:

```release-note:none

```

### DIFF
--- a/mmv1/third_party/terraform/acctest/bootstrap_iam_test_utils.go
+++ b/mmv1/third_party/terraform/acctest/bootstrap_iam_test_utils.go
@@ -52,7 +52,7 @@ func BootstrapIamMembers(t *testing.T, members []IamMember) {
 		for _, pm := range projectMembers {
 			replacedMember := strings.ReplaceAll(pm.Member, "{project_number}", strconv.FormatInt(project.ProjectNumber, 10))
 			projectBindings = append(projectBindings, &cloudresourcemanager.Binding{
-				Role:    member.Role,
+				Role:    pm.Role,
 				Members: []string{replacedMember},
 			})
 		}
@@ -70,7 +70,7 @@ func BootstrapIamMembers(t *testing.T, members []IamMember) {
 		for _, om := range orgMembers {
 			replacedMember := strings.ReplaceAll(om.Member, "{organization_id}", orgId)
 			orgBindings = append(orgBindings, &cloudresourcemanager.Binding{
-				Role:    member.Role,
+				Role:    om.Role,
 				Members: []string{replacedMember},
 			})
 		}
--- a/mmv1/third_party/terraform/acctest/bootstrap_iam_test_utils.go
+++ b/mmv1/third_party/terraform/acctest/bootstrap_iam_test_utils.go
@@ -34,40 +34,78 @@ func BootstrapIamMembers(t *testing.T, members []IamMember) {
 		t.Fatalf("Error getting project with id %q: %s", project.ProjectId, err)
 	}
 
-	// Create the bindings we need to add to the policy.
-	var newBindings []*cloudresourcemanager.Binding
+	// Get the organization ID from environment if any
+	orgIdFromEnv := envvar.GetTestOrgFromEnv(t)
+
+	// Separate bindings into project-level vs. org-level
+	var projectBindings []*cloudresourcemanager.Binding
+	var orgBindings []*cloudresourcemanager.Binding
+
 	for _, member := range members {
-		newBindings = append(newBindings, &cloudresourcemanager.Binding{
-			Role:    member.Role,
-			Members: []string{strings.ReplaceAll(member.Member, "{project_number}", strconv.FormatInt(project.ProjectNumber, 10))},
-		})
+	        // Replace {project_number} and {organization_id} if present
+		replacedMember := strings.ReplaceAll(member.Member, "{project_number}", strconv.FormatInt(project.ProjectNumber, 10))
+		replacedMember = strings.ReplaceAll(replacedMember, "{organization_id}", orgIdFromEnv)
+
+		// If the original member string indicates organization usage, store it as org binding
+		if strings.Contains(member.Member, "{organization_id}") {
+			orgBindings = append(orgBindings, &cloudresourcemanager.Binding{
+				Role:    member.Role,
+				Members: []string{replacedMember},
+			})
+		} else {
+			// Otherwise, treat it as a project binding
+			projectBindings = append(projectBindings, &cloudresourcemanager.Binding{
+				Role:    member.Role,
+				Members: []string{replacedMember},
+			})
+		}
 	}
+
+	// Apply project-level bindings if any
+	if len(projectBindings) > 0 {
+		applyProjectIamBindings(t, client, project.ProjectId, projectBindings)
+	}
+
+	// Apply org-level bindings if any
+	if len(orgBindings) > 0 {
+		if orgIdFromEnv == "" {
+			t.Fatal("Error: Org-level IAM was requested, but no organization ID was set in the environment.")
+		}
+		orgName := "organizations/" + orgIdFromEnv
+		applyOrgIamBindings(t, client, orgName, orgBindings)
+	}
+}
+
+func applyProjectIamBindings(t *testing.T,
+	client *cloudresourcemanager.Service,
+	projectId string,
+	newBindings []*cloudresourcemanager.Binding) {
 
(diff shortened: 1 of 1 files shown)


yerniyazN:
@melinath Hi!
Could you please take a look?
It is follow up of https://github.com/GoogleCloudPlatform/magic-modules/pull/13244#issuecomment-2711084918

melinath:
Reassigning to myself since I have context on this

mmv1/third_party/terraform/acctest/bootstrap_iam_test_utils.go
```diff
@@ -34,40 +34,78 @@ func BootstrapIamMembers(t *testing.T, members []IamMember) {
...
+	// Apply project-level bindings if any
+	if len(projectBindings) > 0 {
+		applyProjectIamBindings(t, client, project.ProjectId, projectBindings)
 	}
 
+	// Apply org-level bindings if any
+	if len(orgBindings) > 0 {
+		if orgIdFromEnv == "" {
+			t.Fatal("Error: Org-level IAM was requested, but no organization ID was set in the environment.")
+		}
```
melinath, line 73: Hm, so, I really like the idea of deferring errors unless the specific requirement is present.

`GetTestOrgFromEnv` and `GetTestProjectFromEnv` both already skip the test if there's no org/project available in the environment, so you don't need another check to make sure they're set. However, contributors may not have all env vars set locally when running tests, and we shouldn't skip their test run unless the env vars are actually required for the tests they're running.

So, `GetTestOrgFromEnv` should only run if there are orgBindings and `GetTestProjectFromEnv` (and fetching the project number!) should only run if there are projectBindings.

Could you move the env var fetching & replacement logic for org and project to only happen if there are org / project bindings present, respectively?

mmv1/third_party/terraform/acctest/bootstrap_iam_test_utils.go
```diff
@@ -34,40 +34,78 @@ func BootstrapIamMembers(t *testing.T, members []IamMember) {
 		t.Fatalf("Error getting project with id %q: %s", project.ProjectId, err)
 	}
 
-	// Create the bindings we need to add to the policy.
-	var newBindings []*cloudresourcemanager.Binding
+	// Get the organization ID from environment if any
+	orgIdFromEnv := envvar.GetTestOrgFromEnv(t)
```
melinath, line 38: nit: can simplify this since there's no other org Id to disambiguate from.

```suggestion
	orgId := envvar.GetTestOrgFromEnv(t)
```

melinath (changes requested):
awesome, thanks! A couple things suggested below.

mmv1/third_party/terraform/acctest/bootstrap_iam_test_utils.go
yerniyazN: Sure. Separated IAM members into two groups: projectMembers and orgMembers. And accessing env variables only if we use it

mmv1/third_party/terraform/acctest/bootstrap_iam_test_utils.go
```diff
@@ -28,46 +28,87 @@ func BootstrapIamMembers(t *testing.T, members []IamMember) {
...
+		project, err := client.Projects.Get(envvar.GetTestProjectFromEnv()).Do()
+		if err != nil {
+			t.Fatalf("Error getting project with id %q: %s", project.ProjectId, err)
+		}
+
+		var projectBindings []*cloudresourcemanager.Binding
+		for _, pm := range projectMembers {
+			replacedMember := strings.ReplaceAll(pm.Member, "{project_number}", strconv.FormatInt(project.ProjectNumber, 10))
+			projectBindings = append(projectBindings, &cloudresourcemanager.Binding{
+				Role:    member.Role,
```
melinath, line 55:
```suggestion
				Role:    pm.Role,
```

mmv1/third_party/terraform/acctest/bootstrap_iam_test_utils.go
```diff
@@ -28,46 +28,87 @@ func BootstrapIamMembers(t *testing.T, members []IamMember) {
...
+		orgId := envvar.GetTestOrgFromEnv(t)
+		if orgId == "" {
+			t.Fatal("Error: Org-level IAM was requested, but no organization ID was set in the environment.")
+		}
+
+		var orgBindings []*cloudresourcemanager.Binding
+		for _, om := range orgMembers {
+			replacedMember := strings.ReplaceAll(om.Member, "{organization_id}", orgId)
+			orgBindings = append(orgBindings, &cloudresourcemanager.Binding{
+				Role:    member.Role,
```
melinath, line 73:
```suggestion
				Role:    om.Role,
```

melinath (changes requested):

melinath (changes requested):
I think I may have led you down a false path 🤦 (though I might be missing something.)

Service accounts are project-level, so replacing the organization ID in the member doesn't actually make any sense. The thing that is org-level is the IAM grant. So the only replacement that needs to happen is project number. 🎉  But there needs to be a way to mark something as org IAM.

I would suggest adding another field to the IamMember struct:

- Add to MMv1 yaml: https://github.com/GoogleCloudPlatform/magic-modules/blob/e17bd4d0b0b648af8e4bf6e7cbbbb86f55017594/mmv1/api/resource/examples.go#L31
- Add to bootstrap_iam_test_utils: https://github.com/GoogleCloudPlatform/magic-modules/blob/e17bd4d0b0b648af8e4bf6e7cbbbb86f55017594/mmv1/third_party/terraform/acctest/bootstrap_iam_test_utils.go#L16
- Generate the new field as appropriate if set: https://github.com/GoogleCloudPlatform/magic-modules/blob/e17bd4d0b0b648af8e4bf6e7cbbbb86f55017594/mmv1/templates/terraform/examples/base_configs/test_file.go.tmpl#L52

I'd suggest maybe a string field (Parent?) that can take a special value ("organization"?) to trigger going into the org-level list instead of the project-level list.

melinath:
It would probably be good to temporarily set the new field on a random resource just to make sure the code gets generated properly.

yerniyazN:
> I think I may have led you down a false path 🤦 (though I might be missing something.)
> 
> Service accounts are project-level, so replacing the organization ID in the member doesn't actually make any sense. The thing that is org-level is the IAM grant. So the only replacement that needs to happen is project number. 🎉 But there needs to be a way to mark something as org IAM.
> 
> I would suggest adding another field to the IamMember struct:
> 
> * Add to MMv1 yaml parser: https://github.com/GoogleCloudPlatform/magic-modules/blob/e17bd4d0b0b648af8e4bf6e7cbbbb86f55017594/mmv1/api/resource/examples.go#L31
> * Add to bootstrap_iam_test_utils: https://github.com/GoogleCloudPlatform/magic-modules/blob/e17bd4d0b0b648af8e4bf6e7cbbbb86f55017594/mmv1/third_party/terraform/acctest/bootstrap_iam_test_utils.go#L16
> * Generate the new field as appropriate if set: https://github.com/GoogleCloudPlatform/magic-modules/blob/e17bd4d0b0b648af8e4bf6e7cbbbb86f55017594/mmv1/templates/terraform/examples/base_configs/test_file.go.tmpl#L52
> 
> I'd suggest maybe a string field (Parent?) that can take a special value ("organization"?) to trigger going into the org-level list instead of the project-level list.

Hi, I didn't fully get it. 
If service account associated with the org, and contains org_id in the name, why not to have replacement for org? 

Yes, all service accounts are project-level, but it is the different project than the one we are replacing {project_number} with, and the one customer using. It's service team who owns those projects where service accounts are created as a resource. But the project_number on the name of service account are different.

I would say we are not replacing parent of the resource, but just a placeholder with customers project_number of org_id

melinath:
@yerniyazN do you know if org-level service agents _must_ have org-level IAM, or could we grant them project-level IAM?

melinath:
Looks like for https://github.com/GoogleCloudPlatform/magic-modules/pull/13244 at least, org-level is required. From https://cloud.google.com/compute/vm-manager/docs/os-policies/orchestrator-prerequisites#permissions:

> For the policy orchestrator to work properly, you must grant all required service agent roles to the respective service accounts, at the parent resource of the orchestrator

It seems likely that this is common, so I do think tying the replacement with the parent resource type makes sense, at least for now.

melinath (approved):

melinath:
test failures are unrelated

PR #13386, opened 2025-03-18
Add storage_control_organization_intelligence_config resource.

kautikdk:
This PR adds support for GCS Storage Intelligence on Organization level.

```release-note:new-resource
`google_storage_control_organization_intelligence_config`
```


kautikdk:
Hi @trodge, Created this follow up PR to add Storage Intelligence support for Organization hierarchy level. It would be good if you can review this since you have context on the feature and have reviewed previous PR.
Previous PR: https://github.com/GoogleCloudPlatform/magic-modules/pull/12689

Also, I will create PR soon for GCP folders.

kautikdk:
We are planning to release all Storage Intelligence [resources](https://github.com/hashicorp/terraform-provider-google/issues/20763) together. Part 1 PR:https://github.com/GoogleCloudPlatform/magic-modules/pull/12689 is already merged and other two PRs are in review. This is a request if we can complete review and merge remaining two PRs by today itself so that we can have s support of full feature in next Tuesday release. It's completely fine if we have some concern which can block PR but if we don't then it would be good to move forward. Thanks!

SirGitsalot (commented):
Mostly LGTM, but I'm not familiar with the API - as an org-level singleton, will this have side effects that might break other tests? It's running in our usual test org, which is where a bunch of other tests creating and deleting buckets and objects will be running concurrently (the integration test passed, but it only ran the tests in this PR)

kautikdk:
> Mostly LGTM, but I'm not familiar with the API - as an org-level singleton, will this have side effects that might break other tests? It's running in our usual test org, which is where a bunch of other tests creating and deleting buckets and objects will be running concurrently (the integration test passed, but it only ran the tests in this PR)

So the Storage Intelligence is a Resource Management tool and it's not similar to org policy which restricts the functionality usage. It shouldn't/won't break any existing tests since it's a new service all together and just configures Intelligence plan and doesn't do any kind of enforcements on resources.

SirGitsalot (approved):

PR #13391, opened 2025-03-18
Bigtable: add NodeScalingFactor support for Clusters

shantstepanian:

```release-note:enhancement
bigtable: added `cluster.node_scaling_factor` field to `google_bigtable_instance` resource
```


mmv1/third_party/terraform/services/bigtable/resource_bigtable_instance.go
```diff
@@ -754,7 +779,7 @@ func resourceBigtableInstanceClusterReorderTypeListFunc(diff tpgresource.Terrafo
 		return err
 	}
 
-	// Clusters can't have their zone, storage_type or kms_key_name updated,
+	// Clusters can't have their zone, storage_type kms_key_name, or node_scaling_factor updated,
```
ron-gal, line 782: missing , after `storage_type`

mmv1/third_party/terraform/services/bigtable/resource_bigtable_instance.go
```diff
@@ -793,6 +818,14 @@ func resourceBigtableInstanceClusterReorderTypeListFunc(diff tpgresource.Terrafo
 				return fmt.Errorf("Error setting cluster diff: %s", err)
 			}
 		}
+
+		oNsf, nNsf := diff.GetChange(fmt.Sprintf("cluster.%d.node_scaling_factor", i))
+		if oNsf != nNsf {
+			err := diff.ForceNew(fmt.Sprintf("cluster.%d.node_scaling_factor", i))
```
ron-gal, line 824: Force new will re-create the instance. I'm not saying it's forbidden, I'm saying let's make sure this is what we want, as it will cause data loss on change

ron-gal (changes requested):
Tests?

translucens:
The client library's PR googleapis/google-cloud-go#11540 contains `Adding node scaling factor` and it could be simplify the implementation of Terraform provider.

mmv1/third_party/terraform/services/bigtable/resource_bigtable_instance.go
shantstepanian: I've updated this to just return an error if we see a difference. Is that okay?

shantstepanian:
Test passes locally:

```
TF_ACC=1 TF_SCHEMA_PANIC_ON_ERROR=1 go test ./google/services/bigtable -v -run=TestAccBigtableInstance_createWithNodeScalingFactorThenCannotUpdate -timeout 240m -ldflags="-X=github.com/hashicorp/terraform-provider-google/version.ProviderVersion=acc"
=== RUN   TestAccBigtableInstance_createWithNodeScalingFactorThenCannotUpdate
=== PAUSE TestAccBigtableInstance_createWithNodeScalingFactorThenCannotUpdate
=== CONT  TestAccBigtableInstance_createWithNodeScalingFactorThenCannotUpdate
--- PASS: TestAccBigtableInstance_createWithNodeScalingFactorThenCannotUpdate (36.23s)
PASS
ok      github.com/hashicorp/terraform-provider-google/google/services/bigtable 36.351s
```

mmv1/third_party/terraform/services/bigtable/resource_bigtable_instance_internal_test.go
```diff
@@ -167,24 +167,30 @@ func TestUnitBigtable_flattenBigtableCluster(t *testing.T) {
 						"storage_target": 60,
 					},
 				},
+				// rongal@ - I don't know what this test does and if we should
+				// expect node scaling factor to be specified. It works if we do specify it
+				// (see next step). How do you suggest I write this test?
+				"node_scaling_factor": "",
```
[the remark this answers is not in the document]
ron-gal: It seems to test the functionality added in https://github.com/GoogleCloudPlatform/magic-modules/pull/13391/files#diff-5c1f11f8027fe52d5d44b39e37826e1c80725d255d110569c13daa9c09e9a0d9R546, so let's keep it with whatever values you want to test (empty to "" sounds like a good test case)

mmv1/third_party/terraform/services/bigtable/resource_bigtable_instance_test.go
```diff
@@ -546,6 +546,79 @@ func TestAccBigtableInstance_forceDestroyBackups(t *testing.T) {
...
+}
+
+func testAccBigtableInstance_nodeScalingFactor_allowDestroy(instanceName string, numNodes int, nodeScalingFactor string) string {
+	return testAccBigtableInstance_nodeScalingFactor(instanceName, numNodes, nodeScalingFactor, true)
+}
+
+func testAccBigtableInstance_nodeScalingFactor(instanceName string, numNodes int, nodeScalingFactor string, allow_destroy bool) string {
+	deletion_protection := ""
+	if allow_destroy {
+		deletion_protection = "deletion_protection = false"
```
ron-gal, line 605: you can have deletion protection false by default, I don't think you care about the true->false transition... Just hardcoded it to false (same as in `testAccBigtableInstance_multipleClustersSameID`)

mmv1/third_party/terraform/services/bigtable/resource_bigtable_instance.go
```diff
@@ -754,14 +779,21 @@ func resourceBigtableInstanceClusterReorderTypeListFunc(diff tpgresource.Terrafo
...
 	// ForceNew if it's changed. This will show a diff with the old state on
 	// the left side and the unmodified new state on the right and the ForceNew
 	// attributed to the _old state index_ even if the diff appears to have moved.
 	// This depends on the clusters having been reordered already by the prior
 	// SetNew call.
 	// We've implemented it here because it doesn't return an error in the
 	// client and silently fails.
+	//
+	// On the other hand, clusters can't change node_scaling_factor, but we
+	// _don't_want to ForceNew in that case. The server side will reject such updates,
```
ron-gal, line 792: _don't_want looks like a broken formatting... is it?

mmv1/third_party/terraform/go.mod
```diff
@@ -3,7 +3,7 @@ module github.com/hashicorp/terraform-provider-google
 go 1.23.0
 
 require (
-	cloud.google.com/go/bigtable v1.33.0
+	cloud.google.com/go/bigtable v1.37.0
```
ron-gal, line 6: These belong in a separate PR

shantstepanian: Done - https://github.com/GoogleCloudPlatform/magic-modules/pull/13737

mmv1/third_party/terraform/services/bigtable/resource_bigtable_instance.go
shantstepanian: fixed

mmv1/third_party/terraform/services/bigtable/resource_bigtable_instance_test.go
shantstepanian: Done, similar to the other test

shantstepanian:
> Tests?

Fixed now

mmv1/third_party/terraform/go.mod
shantstepanian: Done - the version upgrade is merged in, and the files no longer in this PR

mmv1/third_party/terraform/services/bigtable/resource_bigtable_instance_test.go
shantstepanian: I decided to put the deletion protection back in, as I really want to make sure that any attempt to change the node scaling factor will not cause a deletion

mmv1/third_party/terraform/services/bigtable/resource_bigtable_instance_internal_test.go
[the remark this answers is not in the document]
shantstepanian: I've left the test as is. I've added an integration test that specifies no value to ensure that the default is correctly applied

ron-gal (approved):

ron-gal (approved):

mmv1/third_party/terraform/services/bigtable/resource_bigtable_instance.go
```diff
@@ -793,6 +825,12 @@ func resourceBigtableInstanceClusterReorderTypeListFunc(diff tpgresource.Terrafo
 				return fmt.Errorf("Error setting cluster diff: %s", err)
 			}
 		}
+
+		// Do NOT ForceNew for node scaling factor. Return the error on diffs.
+		oNsf, nNsf := diff.GetChange(fmt.Sprintf("cluster.%d.node_scaling_factor", i))
+		if oNsf != nNsf {
+			return fmt.Errorf("Immutable field 'node_scaling_factor' cannot be updated.")
+		}
```
zli82016, line 833: Instead of returning errors, using `ForeceNew` is a standard way to handle immutable field.

mmv1/third_party/terraform/services/bigtable/resource_bigtable_instance_test.go
```diff
@@ -546,6 +546,127 @@ func TestAccBigtableInstance_forceDestroyBackups(t *testing.T) {
...
+		Steps: []resource.TestStep{
+			{
+				// Create config with nothing specified for node scaling factor.
+				// Ensure that we get 1X back.
+				Config: testAccBigtableInstance_nodeScalingFactor_disallowDestroy(instanceName, 2, ""),
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr("google_bigtable_instance.instance", "cluster.0.num_nodes", "2"),
+					resource.TestCheckResourceAttr("google_bigtable_instance.instance", "cluster.0.node_scaling_factor", "NodeScalingFactor1X"),
+				),
+			},
```
zli82016, line 569: You can move step 3 here, as the resource will be destroyed after all of the test steps finish.

mmv1/third_party/terraform/services/bigtable/resource_bigtable_instance_test.go
```diff
@@ -546,6 +546,127 @@ func TestAccBigtableInstance_forceDestroyBackups(t *testing.T) {
...
+				ImportStateVerifyIgnore: []string{"deletion_protection", "instance_type"}, // we don't read instance type back
+			},
+			{
+				// Update deletion protection to allow cleanup after the test. Node scaling factor should remain as is.
+				Config: testAccBigtableInstance_nodeScalingFactor_allowDestroy(instanceName, 2, ""),
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr("google_bigtable_instance.instance", "cluster.0.num_nodes", "2"),
+					resource.TestCheckResourceAttr("google_bigtable_instance.instance", "cluster.0.node_scaling_factor", "NodeScalingFactor1X"),
+				),
+			},
```
zli82016, line 583: The step3 can be used for step 1.

mmv1/third_party/terraform/services/bigtable/resource_bigtable_instance_test.go
```diff
@@ -546,6 +546,127 @@ func TestAccBigtableInstance_forceDestroyBackups(t *testing.T) {
...
+
+	acctest.VcrTest(t, resource.TestCase{
+		PreCheck:                 func() { acctest.AccTestPreCheck(t) },
+		ProtoV5ProviderFactories: acctest.ProtoV5ProviderFactories(t),
+		CheckDestroy:             testAccCheckBigtableInstanceDestroyProducer(t),
+		Steps: []resource.TestStep{
+			{
+				// Create config with nothing specified for node scaling factor.
+				// Ensure that we get 1X back.
+				Config: testAccBigtableInstance_nodeScalingFactor_disallowDestroy(instanceName, 2, ""),
```
zli82016, line 564:
```suggestion
				Config: testAccBigtableInstance_nodeScalingFactor_allowDestroy(instanceName, 2, ""),
```

mmv1/third_party/terraform/services/bigtable/resource_bigtable_instance_test.go
```diff
@@ -546,6 +546,127 @@ func TestAccBigtableInstance_forceDestroyBackups(t *testing.T) {
...
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr("google_bigtable_instance.instance", "cluster.0.num_nodes", "2"),
+					resource.TestCheckResourceAttr("google_bigtable_instance.instance", "cluster.0.node_scaling_factor", "NodeScalingFactor1X"),
+				),
+			},
+		},
+	})
+}
+
+func TestAccBigtableInstance_createWithNodeScalingFactorThenCannotUpdate(t *testing.T) {
```
zli82016, line 588: This test is not needed after making the new field immutable with `ForceNew`

mmv1/third_party/terraform/services/bigtable/resource_bigtable_instance.go
shantstepanian: We explicitly don't want to ForceNew in this case, as we don't want folks to accidentally wipe out their database by changing a field. I haven't used Terraform personally, so I don't know what the actual user experience is in this case, e.g. would they ever get a prompt to warn them about recreating the instance, or would it silently proceed and make this change?

I know the other fields are implemented with ForceNew, but I know the SME on one of those fields isn't fond of that ForceNew behavior either.

Ron initially suggested that we omit this code entirely if the server side will cause a failure. The grpc API will fail if called in this case. However, we have a quirk with our Go client where an update mask prevents the field from getting updated at all. In this case, Terraform will try the update, not change the field, and then the verification will fail (it will query the state and give the original attribute).

Given all this, I'd like to still validate that the attribute doesn't change in Terraform. Is this acceptable? Happy to discuss on chat or GVC if necessary

zli82016: > We explicitly don't want to ForceNew in this case, as we don't want folks to accidentally wipe out their database by changing a field. I haven't used Terraform personally, so I don't know what the actual user experience is in this case, e.g. would they ever get a prompt to warn them about recreating the instance, or would it silently proceed and make this change?
> 
The terraform will not silently make the change.
The terraform plan will show that the resource will be replaced and the changed immutable field will be shown with `# forces replacement`. 

```
  # google_compute_instance.test-vm-1 must be replaced
-/+ resource "google_compute_instance" "test-vm-1" {
         ......
          ~ source                     = "test-vm-10111701" -> "test" # forces replacement
        ......
```

> I know the other fields are implemented with ForceNew, but I know the SME on one of those fields isn't fond of that ForceNew behavior either.
> 
Do you know what fields don't have `ForceNew: true`?

> Ron initially suggested that we omit this code entirely if the server side will cause a failure. The grpc API will fail if called in this case. However, we have a quirk with our Go client where an update mask prevents the field from getting updated at all. In this case, Terraform will try the update, not change the field, and then the verification will fail (it will query the state and give the original attribute).
> 
> Given all this, I'd like to still validate that the attribute doesn't change in Terraform. Is this acceptable? Happy to discuss on chat or GVC if necessary
I will discuss with our team and get back to you soon.

mmv1/third_party/terraform/services/bigtable/resource_bigtable_instance_test.go
shantstepanian: Done

mmv1/third_party/terraform/services/bigtable/resource_bigtable_instance_test.go
shantstepanian: Wasn't sure how to read the two comments here, but the gist I got was that if I remove the deletion protection when creating the instance, then I can skip the last step. Let me know if this now looks how you expect

mmv1/third_party/terraform/services/bigtable/resource_bigtable_instance_test.go
shantstepanian: (see above comment) Wasn't sure how to read the two comments here, but the gist I got was that if I remove the deletion protection when creating the instance, then I can skip the last step. Let me know if this now looks how you expect

mmv1/third_party/terraform/services/bigtable/resource_bigtable_instance.go
shantstepanian: I will check on if other fields aren't covered by "ForceNew", and also check w/ our wider team on the stance of using ForceNew in general

shantstepanian: The only field not covered by ForceNew is ClusterId. Though that's somewhat of a primary key field anyway. If I try that in Terraform, TF lets the update go through, but the update fails in the server (which is expected)

I'll share some notes offline on our team's stance on this

mmv1/third_party/terraform/services/bigtable/resource_bigtable_instance_test.go
zli82016: That is what i mean. Sorry for the confusion.

ron-gal (approved):

ron-gal (approved):

mmv1/third_party/terraform/services/bigtable/resource_bigtable_instance.go
shantstepanian: Per offline discussion, we have implemented this via ForceNew

mmv1/third_party/terraform/services/bigtable/resource_bigtable_instance_test.go
shantstepanian: I've kept the test to show a provided value being used, and then changed it such that the update works via ForceNew instead of getting rejected

ron-gal (approved):

mmv1/third_party/terraform/services/bigtable/resource_bigtable_instance.go
```diff
@@ -152,6 +152,13 @@ func ResourceBigtableInstance() *schema.Resource {
 							Computed:    true,
 							Description: `The state of the cluster`,
 						},
+						"node_scaling_factor": {
+							Type:         schema.TypeString,
```
zli82016, line 156: Instead of using diff.ForceNew function, `ForceNew:    true,` can be used for this field.

mmv1/third_party/terraform/services/bigtable/resource_bigtable_instance.go
```diff
@@ -793,6 +818,14 @@ func resourceBigtableInstanceClusterReorderTypeListFunc(diff tpgresource.Terrafo
...
 			}
 		}
+
+		oNsf, nNsf := diff.GetChange(fmt.Sprintf("cluster.%d.node_scaling_factor", i))
+		if oNsf != nNsf {
+			err := diff.ForceNew(fmt.Sprintf("cluster.%d.node_scaling_factor", i))
+			if err != nil {
+				return fmt.Errorf("Error setting node scaling factor: %s", err)
+			}
+		}
```
zli82016, line 828: Instead of using diff.ForceNew function, `ForceNew:    true,` can be added for the field.

zli82016 (changes requested):

mmv1/third_party/terraform/services/bigtable/resource_bigtable_instance.go
shantstepanian: Done

mmv1/third_party/terraform/services/bigtable/resource_bigtable_instance.go
shantstepanian: Done

zli82016:
Tests passed in [Teamcity](https://hashicorp.teamcity.com/buildConfiguration/TerraformProviders_GoogleCloud_GOOGLE_BETA_MMUPSTREAMTESTS_GOOGLEBETA_PACKAGE_BIGTABLE/365662?buildTab=tests)

PR #13394, opened 2025-03-19
Add storage_control_folder_intelligence_config resource.

kautikdk:
This PR adds support for GCS Management Hub on GCP Folder level.

```release-note:new-resource
`google_storage_control_folder_intelligence_config`
```


kautikdk:
Hi @trodge, Created this follow up PR to add Storage Intelligence support for Folder hierarchy level. It would be good if you can review this since you have context on the feature and have reviewed previous PR.
Previous PR: https://github.com/GoogleCloudPlatform/magic-modules/pull/12689

mmv1/products/storagecontrol/FolderIntelligenceConfig.yaml
```diff
@@ -0,0 +1,179 @@
...
+docs:
+  warning: |
+    Storage Intelligence Config is a singleton resource which cannot be created or deleted. A single instance of Storage Intelligence Config exist for each GCP Folder. Terraform does not create or destroy this resource.
+    Terraform resource creation for this resource is simply an update operation on existing resource with specified properties. Terraform deletion won't have any effect on this resource rather it will only remove it from the state file.
+
+# URL for the resource's standard Get method. https://google.aip.dev/131
+self_link: 'folders/{{name}}/locations/global/intelligenceConfig'
+
+custom_code:
+  custom_create: templates/terraform/custom_create/storage_control_folder_intelligence.go.tmpl
```
ScottSuarez, line 33: why is custom create needed? pre_create seems sufficient to just alter the url. There is the question of, if we need this to be all on v2 if this is the right way to go about things. Because we will be creating on v2 and reading on v1 with the current implmentation. Open to call to discuss.

ScottSuarez (changes requested):

mmv1/products/storagecontrol/FolderIntelligenceConfig.yaml
kautikdk: Hi @ScottSuarez, 
URL modification code was written a quite some time ago before we decided to create a new service `storagecontrol` for GCS JSON v2. I forgot to remove it, My bad! Regarding custom_create, The reason is the update_mask and PATCH method in create operation. Storage Intelligence can not be created so we mapped create operation to PATCH method but the concern is related to permanent and optional fields. `edition_config` is a permanent field and back-filled by server side so user can change but it is not the field that we can clear but the `filter` block is optional and not back filled by server side. Now just to simulate the create operation, we can put full update_mask of every fields which are optional but the no need to send empty value for the fields which are computed or has server side value. 
Why we want put `update_mask` for optional fields?
The main reason is we are clearly stating that the create operation is actually update operation under the hood so the first create operation diff should represent whatever will be the end state and should not generate diff after apply. That necessitates requirement of update_mask for optional fields and if we don't do it then some diff can be generated after apply operation. That will be just similar to import operation that we already support. 

I have removed that URL modification line but open to discuss any concerns here.

kautikdk: Also empty value for `edition_config` will result in server side error so we can not put update_mask for that field until it is specified by the user in create operation.

kautikdk:
Referring comment from other PR:https://github.com/GoogleCloudPlatform/magic-modules/pull/13386#issuecomment-2739721130

mmv1/products/storagecontrol/FolderIntelligenceConfig.yaml
```diff
@@ -0,0 +1,179 @@
...
+update_mask: true
+
+exclude_delete: true
+
+import_format:
+  - 'folders/{{name}}/locations/global/intelligenceConfig'
+
+# If true, code for handling long-running operations is generated along with
+# the resource. If false, that code is not generated.
+autogen_async: false
```
ScottSuarez, line 48: Can we add an autogen test that is excluded from tests so that there is an example in documentation (if we decided to support this).

kautikdk:
> > Storage Intelligence Config is a singleton resource which cannot be created or deleted. A single instance of Storage Intelligence Config exist for each GCP Folder. Terraform does not create or destroy this resource.
> 
> I'm concerned about this resource not having the ability to be cleared. This is a one way set of resource and doesn't appear to be working within the terraform model. Deletion should clear any set config ideally. Given that it does not, I am not sure this api make sense to support at the terraform level.

Yes, This resource is different than other GCS resources(in terms of singleton behavior). We have guidelines for singleton resources and there is a mention of resources having no delete endpoint and for them, we have [exclude_delete: true](https://googlecloudplatform.github.io/magic-modules/reference/resource/#create_url) which basically skips the deletion rather removes it from the Terraform only. I believe that's the standard Terraform pattern that we are following across TPG though I am don't know about any other similar resources outside of GCS. 

Storage Intelligence is a very important feature for large scale GCS resource management and it should be supported through Terraform as well. We have ask from feature team to provide Terraform support with high priority. I will share reference documentations internally.

mmv1/products/storagecontrol/FolderIntelligenceConfig.yaml
kautikdk: Added!

kautikdk: Excluded test generation since there is no creation or deletion operation.

melinath:
@kautikdk is correct that singletons don't need to unset their configuration; this is how they generally work in the provider. Best practices for singleton resources are documented at https://googlecloudplatform.github.io/magic-modules/best-practices/common-resource-patterns/.

mmv1/products/storagecontrol/FolderIntelligenceConfig.yaml
melinath: +1 to @ScottSuarez's point. This should be possible to do with `pre_create` custom code rather than replacing the whole create method.

melinath: I'm a little confused what the concern is with O+C fields; in general, sending an empty value for them (triggering server-side default behavior) should be correct. Although it would be possible to persist the values that already exist on the server side that might be more confusing than beneficial.

kautikdk: Ahh yes, checked how exactly pre-create injects code and realized it adds code snippet just before making an API call. Changed custom_create to pre_create and kept previous update_mask logic as it is while we are still discussing this behavior.

kautikdk: > I'm a little confused what the concern is with O+C fields; in general, sending an empty value for them (triggering server-side default behavior) should be correct. Although it would be possible to persist the values that already exist on the server side that might be more confusing than beneficial.

So just to give an example in the Storage Intelligence resource, `edition_config` is a property of the resource and all the resources are having this field backfilled by the server. Users can configure this field, has server side value initially, and can not be removed so it should be Optional+Computed. Now since we can not remove it, we can not send empty value and update_mask together as it results in a InvalidValue error so we have to put update_mask only when user specifies it otherwise we will keep server side value in the create operation.

melinath: So to make sure I understand right, `edition_config` behaves as follows:

- The initial configuration is "empty"
- The first time that a user calls this API, they can choose to supply a value; if they don't, the API will choose a value on the server side
- Either way, once the field has a value, it can _never be changed_.

Is that correct?

kautikdk: Umm, Not quite like that.

API behavior is,
-  The initial value of `edition_config` field is `INHERIT`.
-  User can change the field value but can never remove `edition_config` property from the resource.

Terraform mapping,
-  Marked as O+C. That will allow users to keep server side value if not specified in the config and they can change it if they want.

Regarding the empty value reference, I believe my wordings made some confusion. There is an API side validation which checks if the update_mask is provided for `edition_config`. if it is provided and field is not present in the API request body then it will throw an InvalidValue error.
In the create operation, we will put update_mask of `edition_config` only when it is specified by user in the resource config. Just like whatever is being done here for every field: https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/templates/terraform/update_mask.go.tmpl

ScottSuarez (approved):

PR #13395, opened 2025-03-19
Fix TestAccFirebaseAppHostingBackend

rainshen49:
Partially fix:

https://github.com/hashicorp/terraform-provider-google/issues/21886
https://github.com/hashicorp/terraform-provider-google/issues/21891

The reason is that multiple tests are creating & destroying the same service account, causing collision.

```release-note:none
```


rileykarson:
Thanks!

PR #13398, opened 2025-03-19
Feature gap: Add `groupPlacementPolicy.tpuTopology`

Cezarus27:

```release-note:enhancement
compute: added `group_placement_policy.0.tpu_topology` field to `google_compute_resource_policy` resource.
```

### DIFF
--- a/mmv1/products/compute/ResourcePolicy.yaml
+++ b/mmv1/products/compute/ResourcePolicy.yaml
@@ -341,9 +341,6 @@ properties:
         description: |
           Specifies the shape of the TPU slice.
         immutable: true
-        required_with:
-          - 'group_placement_policy.0.vmCount'
-          - 'group_placement_policy.0.collocation'
         min_version: 'beta'
   - name: 'instanceSchedulePolicy'
     type: NestedObject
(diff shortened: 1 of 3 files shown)


mmv1/products/compute/ResourcePolicy.yaml
```diff
@@ -316,6 +316,15 @@ properties:
           - group_placement_policy.0.max_distance
         immutable: true
         min_version: 'beta'
+      - name: 'tpuTopology'
+        type: String
+        description: |
+          Specifies the shape of the TPU slice.
+        immutable: true
+        required_with:
```
roaks3, line 324: Can you help me understand why we need `required_with` here? It seems like that would break users who already specify only those fields.

Cezarus27: API requires those fields to be set, when `tpu_topology` is set anyway. Beside that it won't break anything if users have `vm_count` and `collocation` already set in their configuration.

roaks3 (changes requested):

mmv1/products/compute/ResourcePolicy.yaml
Cezarus27: Okay, I agree. Change pushed.

Cezarus27:
@roaks3 kind request to review this PR.

roaks3 (approved):

PR #13401, opened 2025-03-19
Adding automated backup config to Redis 

NA2047:

```release-note:enhancement
redis: added `automated_backup_config` field to `google_redis_cluster`
```


mmv1/products/redis/Cluster.yaml
```diff
@@ -250,6 +250,56 @@ properties:
     description: |
       System assigned, unique identifier for the cluster.
     output: true
+  - name: 'automatedBackupConfig'
+    type: NestedObject
```
rileykarson, line 254: `default_from_api: true`

mmv1/products/redis/Cluster.yaml
```diff
@@ -250,6 +250,56 @@ properties:
...
+    description: "The automated backup config for a instance."
+    properties:
+      - name: 'automatedBackupMode'
```
rileykarson, line 257: Seems like `DISABLED` is a default value, judging by those tests!

mmv1/products/redis/Cluster.yaml
```diff
@@ -250,6 +250,56 @@ properties:
...
+        type: Enum
+        description: |
+          "Optional. The automated backup mode. If the mode is disabled, the other fields will be ignored."
+          1. `AUTOMATED_BACKUP_MODE_UNSPECIFIED`: Default value. Automated backup config is not specified.
+          2. `DISABLED`: Automated backup config disabled.
+          3. `ENABLED`: Automated backup config enabled.
+        enum_values:
+          - 'AUTOMATED_BACKUP_MODE_UNSPECIFIED'
```
rileykarson, line 265: We can remove this from here + the description and just make `DISABLED` the default.

mmv1/products/redis/Cluster.yaml
```diff
@@ -250,6 +250,56 @@ properties:
...
+        properties:
+          - name: 'startTime'
+            type: NestedObject
+            description: |
+              Required. The start time of every automated backup in UTC.
+              It must be set to the start of an hour. This field is required.
+            required: true
+            properties:
+              - name: 'hours'
+                type: Interger
```
rileykarson, line 280: I am surprised we don't return an error in some way for this! Should be `Integer` here + below

fyi @c2thorn

mmv1/products/redis/Cluster.yaml
```diff
@@ -250,6 +250,55 @@ properties:
...
+    default_from_api: true
+    description: "The automated backup config for a instance."
+    properties:
+      - name: 'automatedBackupMode'
+        type: Enum
+        description: |
+          "Optional. The automated backup mode. If the mode is disabled, the other fields will be ignored."
+        default_value: "DISABLED"
+        enum_values:
+          - 'AUTOMATED_BACKUP_MODE_UNSPECIFIED'
```
rileykarson, line 264:
```suggestion
```

mmv1/products/redis/Cluster.yaml
```diff
@@ -250,6 +250,55 @@ properties:
...
     output: true
+  - name: 'automatedBackupConfig'
+    type: NestedObject
+    default_from_api: true
+    description: "The automated backup config for a instance."
+    properties:
+      - name: 'automatedBackupMode'
+        type: Enum
+        description: |
+          "Optional. The automated backup mode. If the mode is disabled, the other fields will be ignored."
```
rileykarson, line 261:
```suggestion
          "Optional. The automated backup mode. If the mode is DISABLED, the other fields will be ignored."
```

mmv1/products/redis/Cluster.yaml
```diff
@@ -250,6 +250,55 @@ properties:
...
+              - name: 'seconds'
+                type: Integer
+                description: |
+                  Seconds of a minute. Must be greater than or equal to 0 and typically must be less than or equal to 59.
+                  An API may allow the value 60 if it allows leap-seconds.
+              - name: 'nanos'
+                type: Integer
+                description: |
+                  Fractions of seconds, in nanoseconds. Must be greater than or equal to 0 and less than or equal to 999,999,999.
+      - name: 'retention'
```
rileykarson, line 296: Can you include a test configuration that omits `retention`?

mmv1/third_party/terraform/services/redis/resource_redis_cluster_test.go.tmpl
```diff
@@ -270,6 +270,70 @@ func TestAccRedisCluster_createUpdateDeletionProtection(t *testing.T) {
...
+func testAccRedisCluster_automatedBackupConfig(context map[string]interface{}) string {
+	return acctest.Nprintf(`
+
+resource "google_redis_cluster" "cluster_abc" {
+  name                           = "tf-test-redis-abc%{random_suffix}"
+  shard_count                    = 1
+  region                         = "us-central1"
+  deletion_protection_enabled    = false
+  automated_backup_config {
+   automated_backup_mode         = "ENABLED"
```
rileykarson, line 321: This should draw the value from the context rather than hardcode it, right?

NA2047: Normally yes, deletion_protection_enabled is set by default to true by the provider, so I need explicitly set it to false so it can be destroyed by the tests

mmv1/products/redis/Cluster.yaml
NA2047: @rileykarson You would like me to remove it form the existing test or make an additional test without `retention` ?

rileykarson: Based on the test failure, we should set the default value of `"3024000s"` (and then add a non-default `retention` value back to the current test)

mmv1/templates/terraform/custom_flatten/redis_cluster_automated_backup_config.go.tmpl
```diff
@@ -0,0 +1,65 @@
+func flatten{{$.GetPrefix}}{{$.TitlelizeProperty}}(v interface{}, d *schema.ResourceData, config *transport_tpg.Config) interface{} {
+	if v == nil {
+		return nil
+	}
+	original := v.(map[string]interface{})
+	if len(original) == 0 {
+		return nil
+	}
+	// if automated_backup_config is not defined 
+    if len(d.Get("automated_backup_config").([]interface{})) < 1 {
```
rileykarson, line 10: Sorry- can you walk me through why we need to check state again? I think we should be checking for DISABLED in the API response here.

mmv1/third_party/terraform/services/redis/resource_redis_cluster_test.go.tmpl
```diff
@@ -263,14 +262,78 @@ func TestAccRedisCluster_createUpdateDeletionProtection(t *testing.T) {
...
+		CheckDestroy:             testAccCheckRedisClusterDestroyProducer(t),
+		Steps: []resource.TestStep{
+			{
+				Config: testAccRedisCluster_automatedBackupConfig(context),
+			},
+			{
+				ResourceName:            "google_redis_cluster.cluster_abc",
+				ImportState:             true,
+				ImportStateVerify:       true,
+				ImportStateVerifyIgnore: []string{"automated_backup_config", "automated_backup_config.0", "automated_backup_config.0.fixed_frequency_schedule", "automated_backup_config.0.fixed_frequency_schedule.0.start_time", "utomated_backup_config.0.fixed_frequency_schedule.0.start_time.0.hours", "automated_backup_config.0.retention"},
```
rileykarson, line 290: ISVI here indicates an issue with import- there's no good reason we can't match both states exactly here. (The fix to not check state in Read should fix this)

NA2047: The API docs are still being worked on right now. but they the implmetation is the same as the ones I have provided :)

NA2047: Totally agree, should be good now :)

mmv1/products/redis/Cluster.yaml
```diff
@@ -250,6 +251,36 @@ properties:
...
     output: true
+  - name: 'automatedBackupConfig'
+    type: NestedObject
+    description: "The automated backup config for a instance."
+    custom_flatten: 'templates/terraform/custom_flatten/redis_cluster_automated_backup_config.go.tmpl'
+    custom_expand: 'templates/terraform/custom_expand/redis_cluster_automated_backup_config.go.tmpl'
+    properties:
+      - name: 'fixedFrequencySchedule'
+        type: NestedObject
+        description: "Optional. Trigger automated backups at a fixed frequency."
```
rileykarson, line 262:
```suggestion
        required: true
        description: "Trigger automated backups at a fixed frequency."
```

> "Error creating Cluster: googleapi: Error 400: invalid argument: invalid automated backup config: start time must be specified"

when I omit this locally. We can remove this if other options are added later.

NA2047: Fixed, this should be required thanks for the catch

rileykarson (approved):

PR #13404, opened 2025-03-19
Make multiple secondary cluster operations non-concurrent

199201shubhamsahu:
fixed b/402516733
Make multiple secondary cluster operations non-concurrent

```release-note:enhancement
alloydb: added a mutex to `google_alloydb_cluster` to prevent conflicts among multiple cluster operations
```


BBBmau (commented):
looks good,  I'd like to know the reason for this change and what caused it, were we running into issues with tests? I'll hit approve after a brief explanation.

zli82016:
> looks good, I'd like to know the reason for this change and what caused it, were we running into issues with tests? I'll hit approve after a brief explanation.

Based on b/402516733, concurrent creation/deletion/promotion of multiple secondaries can potentially cause failures and leave the secondaries or primary in an invalid state.

BBBmau (approved):

PR #13405, opened 2025-03-20
Add tags field to Datafusion Instances for TagsR2401

rajeshgupthar:
Part of b/337048265

```datafusion: added `tags` field to `datafusion_instance` to allow setting tags for instances at creation time```

### DIFF
--- a/mmv1/products/datafusion/Instance.yaml
+++ b/mmv1/products/datafusion/Instance.yaml
@@ -375,3 +375,12 @@ properties:
           enum_values:
             - 'ENABLED'
             - 'DISABLED'
+  - name: 'tags'
+    type: KeyValuePairs
+    description: |
+      A map of resource manager tags.
+      Resource manager tag keys and values have the same definition as resource manager tags.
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Review by c2thorn — CHANGES_REQUESTED** [review]
@rajeshgupthar the test seems to fail - there is a permission issue on the tag key. Are you able to view the log

File: mmv1/products/datafusion/Instance.yaml [hunk 1]
```diff
@@ -375,3 +375,12 @@ properties:
...
             - 'DISABLED'
+  - name: 'tags'
+    type: KeyValuePairs
+    description: |
+      A map of resource manager tags.
+      Resource manager tag keys and values have the same definition as resource manager tags.
+      Keys must be in the format tagKeys/{tag_key_id}, and values are in the format tagValues/{tag_value_id}.
+      The field is ignored (both PUT & PATCH) when empty.
+    immutable: true
+    ignore_read: true
```
c2thorn, line 386: why is this `ignore_read`?

PR #13414, opened 2025-03-20
Adding magic module config for PolicyOrchestratorForOrganization [osconfigv2]

yerniyazN:

```release-note:new-resource
`google_os_config_v2_policy_orchestrator_for_organization`
```


ScottSuarez:
we need to be careful with these tests as it seems they will make policy changes and infrastructure changes at an organization level across our projects. Is it possible it make it so no actual changes are made? What permissions need to be granted to the SA?

ScottSuarez (changes requested):
tests are failing due to SA permissions

yerniyazN:
> tests are failing due to SA permissions

Looks like it is failing to get and set IAM bindings for org via bootstraping. Seems user running VCR test doesn't have right permissions. Can we grant it, same way as it was done for the test project ?

mmv1/products/osconfigv2/PolicyOrchestratorForOrganization.yaml
```diff
@@ -0,0 +1,1266 @@
...
+      - member: "serviceAccount:service-org-{organization_id}@gcp-sa-osconfig-rollout.iam.gserviceaccount.com"
+        role: "roles/osconfig.rolloutServiceAgent"
+      - member: "serviceAccount:service-org-{organization_id}@gcp-sa-progrollout.iam.gserviceaccount.com"
+        role: "roles/progressiverollout.serviceAgent"
+    vars:
+      policy_orchestrator_id: po-org
+      orchestrated_resource_id: test-orchestrated-resource-org
+      os_policy_id: test-os-policy-org
+    test_env_vars:
+      org_id: ORG_ID
```
melinath, line 44:
```suggestion
      org_id: ORG_TARGET
```

mmv1/products/osconfigv2/PolicyOrchestratorForOrganization.yaml
```diff
@@ -0,0 +1,1266 @@
...
+update_verb: PATCH
+id_format: organizations/{{organization_id}}/locations/global/policyOrchestrators/{{policy_orchestrator_id}}
+import_format:
+  - organizations/{{organization_id}}/locations/global/policyOrchestrators/{{policy_orchestrator_id}}
+min_version: beta
+examples:
+  - name: osconfigv2_policy_orchestrator_for_organization_basic
+    primary_resource_id: policy_orchestrator_for_organization
+    bootstrap_iam:
+      - member: "serviceAccount:service-org-{organization_id}@gcp-sa-osconfig.iam.gserviceaccount.com"
```
melinath, line 33: The bootstrapping logic will need to use `GetTestOrgTargetFromEnv` to target the correct org.

mmv1/products/osconfigv2/PolicyOrchestratorForOrganization.yaml
```diff
@@ -0,0 +1,1266 @@
...
+        role: "roles/osconfig.rolloutServiceAgent"
+      - member: "serviceAccount:service-org-{organization_id}@gcp-sa-progrollout.iam.gserviceaccount.com"
+        role: "roles/progressiverollout.serviceAgent"
+    vars:
+      policy_orchestrator_id: po-org
+      orchestrated_resource_id: test-orchestrated-resource-org
+      os_policy_id: test-os-policy-org
+    test_env_vars:
+      org_id: ORG_ID
+      project_number: PROJECT_NUMBER
```
melinath, line 45: You'll need to provision a new project here instead of using the project number from the test environment, since you'll be in a different org. (Unless that doesn't matter.)

melinath (changes requested):
leaving a few comments related to the org-level bootstrapping!

mmv1/products/osconfigv2/PolicyOrchestratorForOrganization.yaml
yerniyazN: I see you already PR for it 
https://github.com/GoogleCloudPlatform/magic-modules/pull/13467/files#diff-21a9db155a318c8ab73cb3f9024f96077de7e2db7a328d653d095de43922cedbR64

mmv1/products/osconfigv2/PolicyOrchestratorForOrganization.yaml
yerniyazN: Project selectors are optional, so I just removed it

ScottSuarez:
internal error. I wonder if there is a issue because they are both writing to organization policies simultaneously. We might need these tests to run synchronously

yerniyazN:
> internal error. I wonder if there is a issue because they are both writing to organization policies simultaneously. We might need these tests to run synchronously

Yes, you are right.
Verified logs. It is saying: 'SetIamPolicy;There were concurrent policy changes'

Removed parallel running of the test.

ScottSuarez (approved):

PR #13421, opened 2025-03-20
Add a test check function to get cai assets metadata

zli82016:
This change implements a check function to get and print out the CAI asset name and Terraform configuration of the created main resource in the test, and adds the check function to a couple of tests for `google_project`. These values will be extracted from nightly tests to get raw config and CAI asset.

The printed values are in https://paste.googleplex.com/5502359491575808. 

```release-note:none

```

### DIFF
--- a/mmv1/third_party/terraform/acctest/tgc_utils.go
+++ b/mmv1/third_party/terraform/acctest/tgc_utils.go
@@ -13,18 +13,24 @@ import (
 
 // Hardcode the Terraform resource name -> API service name mapping temporarily.
 // TODO: [tgc] read the mapping from the resource metadata files.
-var CaiProductBackendNames = map[string]string{
-	"google_compute_instance": "compute",
-	"google_project":          "cloudresourcemanager",
+var ApiServiceNames = map[string]string{
+	"google_compute_instance": "compute.googleapis.com",
+	"google_project":          "cloudresourcemanager.googleapis.com",
 }
 
 // Gets the test metadata for tgc:
 //   - test config
 //   - cai asset name
 //     For example: //compute.googleapis.com/projects/ci-test-188019/zones/us-central1-a/instances/tf-test-mi3fqaucf8
-func GetTestMetadataForTgc(service, resourceType, resourceName, config string) resource.TestCheckFunc {
+func GetTestMetadataForTgc(service, address, config string) resource.TestCheckFunc {
 	return func(s *terraform.State) error {
-		address := fmt.Sprintf("%s.%s", resourceType, resourceName)
+		splits := strings.Split(address, ".")
+		if splits == nil || len(splits) < 2 {
+			return fmt.Errorf("The resource address %s is invalid.", address)
+		}
+		resourceType := splits[0]
+		resourceName := splits[1]
+
 		rState := s.RootModule().Resources[address]
 		if rState == nil || rState.Primary == nil {
 			return fmt.Errorf("The resource state is unavailable. Please check if the address %s.%s is correct.", resourceType, resourceName)
@@ -32,7 +38,7 @@ func GetTestMetadataForTgc(service, resourceType, resourceName, config string) r
 
 		// Convert the resource ID into CAI asset name
 		// and then print out the CAI asset name in the logs
-		if productName, ok := CaiProductBackendNames[resourceType]; !ok {
+		if apiServiceName, ok := ApiServiceNames[resourceType]; !ok {
 			return fmt.Errorf("The Cai product backend name for resource %s doesn't exist.", resourceType)
 		} else {
 			var rName string
@@ -43,10 +49,12 @@ func GetTestMetadataForTgc(service, resourceType, resourceName, config string) r
 				rName = rState.Primary.ID
 			}
 
-			caiAssetName := fmt.Sprintf("//%s.googleapis.com/%s", productName, rName)
+			caiAssetName := fmt.Sprintf("//%s/%s", apiServiceName, rName)
 			log.Printf("[DEBUG]TGC CAI asset name: %s", caiAssetName)
 		}
 
+		// The acceptance tests names will be also used for the tgc tests.
+		// "service" is logged and will be used to put the tgc tests into specific service packages.
 		log.Printf("[DEBUG]TGC Terraform service: %s", service)
 		log.Printf("[DEBUG]TGC Terraform resource: %s", resourceType)
 
--- a/mmv1/third_party/terraform/acctest/tgc_utils.go
+++ b/mmv1/third_party/terraform/acctest/tgc_utils.go
@@ -11,15 +11,17 @@ import (
 	"github.com/hashicorp/terraform-provider-google/google/tpgresource"
 )
 
+// Hardcode the Terraform resource name -> API service name mapping temporarily.
+// TODO: [tgc] read the mapping from the resource metadata files.
 var CaiProductBackendNames = map[string]string{
-	"compute":         "compute",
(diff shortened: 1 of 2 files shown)


mmv1/third_party/terraform/acctest/tgc_utils.go
```diff
@@ -0,0 +1,65 @@
...
+
+var CaiProductBackendNames = map[string]string{
+	"compute":         "compute",
+	"resourcemanager": "cloudresourcemanager",
+}
+
+// Gets the test metadata for tgc:
+//   - test config
+//   - cai asset name
+//     For example: //compute.googleapis.com/projects/terraform-dev-zhenhuali/zones/us-central1-a/instances/tf-test-mi3fqaucf8
```
melinath, line 22: Maybe use one of the public test projects here.

```suggestion
//     For example: //compute.googleapis.com/projects/ci-test-188019/zones/us-central1-a/instances/tf-test-mi3fqaucf8
```

mmv1/third_party/terraform/acctest/tgc_utils.go
```diff
@@ -0,0 +1,65 @@
...
+		}
+
+		log.Printf("[DEBUG]TGC Terraform service: %s", service)
+		log.Printf("[DEBUG]TGC Terraform resource: %s", resourceType)
+
+		re := regexp.MustCompile(`\"(tf[-_]?test[-_]?.*?)([a-z0-9]+)\"`)
+		config = re.ReplaceAllString(config, `"${1}tgc"`)
+
+		// Replace resource name with the resource's real name
+		// For example, replace `"google_compute_instance" "foobar"` with `"google_compute_instance" "tf-test-mi3fqaucf8"`
```
melinath, line 55: Why do this replacement instead of recording the ID of the resource under test explicitly in a separate line? Maybe because the exported config will use the ID in this way? That would be good to document here in a comment if so.

mmv1/third_party/terraform/acctest/tgc_utils.go
```diff
@@ -0,0 +1,65 @@
...
+		n := tpgresource.GetResourceNameFromSelfLink(rState.Primary.ID)
+		old := fmt.Sprintf(`"%s" "%s"`, resourceType, resourceName)
+		new := fmt.Sprintf(`"%s" "%s"`, resourceType, n)
+		config = strings.Replace(config, old, new, 1)
+
+		log.Printf("[DEBUG]TGC raw_config starts %s", config)
+		log.Printf("[DEBUG]End of TGC raw_config")
```
melinath, line 62: It seems like it might be good to have this all in one log line so that nothing can get between them

mmv1/third_party/terraform/acctest/tgc_utils.go
```diff
@@ -0,0 +1,65 @@
...
+
+	"github.com/hashicorp/terraform-plugin-testing/helper/resource"
+	"github.com/hashicorp/terraform-plugin-testing/terraform"
+	"github.com/hashicorp/terraform-provider-google/google/tpgresource"
+)
+
+var CaiProductBackendNames = map[string]string{
+	"compute":         "compute",
+	"resourcemanager": "cloudresourcemanager",
+}
```
melinath, line 17: I would recommend basing this on just the TF resource name instead of product -> service name. We can easily generate a comprehensive TF resource name -> API service name mapping, either generated from MMv1 or by reading the `resource` and `api_service_name` fields from metadata files like https://github.com/hashicorp/terraform-provider-google-beta/blob/main/google-beta/services/accessapproval/resource_organization_access_approval_settings_generated_meta.yaml

melinath (changes requested):

mmv1/third_party/terraform/acctest/tgc_utils.go
```diff
@@ -0,0 +1,67 @@
...
+	"github.com/hashicorp/terraform-plugin-testing/terraform"
+	"github.com/hashicorp/terraform-provider-google/google/tpgresource"
+)
+
+// Hardcode the Terraform resource name -> API service name mapping temporarily.
+// TODO: [tgc] read the mapping from the resource metadata files.
+var CaiProductBackendNames = map[string]string{
+	"google_compute_instance": "compute",
+	"google_project":          "cloudresourcemanager",
+}
```
melinath, line 19:
```suggestion
var CaiProductBackendNames = map[string]string{
	"google_compute_instance": "compute.googleapis.com",
	"google_project":          "cloudresourcemanager.googleapis.com",
}
```

melinath (changes requested):
we should use the full service domain here because that's how we store the information in the metadata files

mmv1/third_party/terraform/acctest/tgc_utils.go
```diff
@@ -0,0 +1,67 @@
...
+var ApiServiceNames = map[string]string{
+	"google_compute_instance": "compute.googleapis.com",
+	"google_project":          "cloudresourcemanager.googleapis.com",
+}
+
+// Gets the test metadata for tgc:
+//   - test config
+//   - cai asset name
+//     For example: //compute.googleapis.com/projects/ci-test-188019/zones/us-central1-a/instances/tf-test-mi3fqaucf8
+func GetTestMetadataForTgc(service, resourceType, resourceName, config string) resource.TestCheckFunc {
```
melinath, line 25: the service is no longer necessary here

```suggestion
func GetTestMetadataForTgc(resourceType, resourceName, config string) resource.TestCheckFunc {
```

mmv1/third_party/terraform/acctest/tgc_utils.go
```diff
@@ -0,0 +1,67 @@
...
+	return func(s *terraform.State) error {
+		address := fmt.Sprintf("%s.%s", resourceType, resourceName)
```
melinath, line 27: what would you think about passing in the address instead, and then breaking it into the resource type and resource name? That might make the signature more similar to other checks (which usually take an address).

(I have a slight inclination to pass the full address but don't care strongly about it.)

melinath (changes requested):

mmv1/third_party/terraform/acctest/tgc_utils.go
zli82016: The acceptance test name is also used for the tgc tests. `service` is logged and will be used to put the tgc tests  into specific service packages.

mmv1/third_party/terraform/acctest/tgc_utils.go
zli82016: Passing the address works here to be consistent with other checks.

mmv1/third_party/terraform/acctest/tgc_utils.go
zli82016: Added the comment.

mmv1/third_party/terraform/acctest/tgc_utils.go
melinath: Ah, that makes sense!

If this should be the service package, it could be inferred based on the service package containing this resource's definition instead of passed in as part of the call...

Unless the plan is to automatically add the calls to this function to all VCR tests (and the mechanism adding the calls will have access to that information)?

melinath (approved):
LGTM

mmv1/third_party/terraform/acctest/tgc_utils.go
melinath: I think this would likely be better handle with automation but we can always decide to change it later if it causes problems.

zli82016:
cc @hao-nan-li

PR #13424, opened 2025-03-21
fix: (storage) added content_base64 field in google_storage_bucket_object_content datasource

gurusai-voleti:
Fixes: https://github.com/hashicorp/terraform-provider-google/issues/10835

```release-note:enhancement
storage: added `content_base64` field in `google_storage_bucket_object_content` datasource
```


zli82016 (changes requested):
Thanks for adding the field. Can you add a test for the use case in https://github.com/hashicorp/terraform-provider-google/issues/10835#issue-1093838248? Thanks.

gurusai-voleti:
@zli82016 added test case, could you please review the PR again

mmv1/third_party/terraform/services/storage/data_source_storage_bucket_object_content_test.go
```diff
@@ -22,12 +27,71 @@ func TestAccDataSourceStorageBucketObjectContent_Basic(t *testing.T) {
...
+					resource.TestCheckResourceAttrSet("data.google_storage_bucket_object_content.default", "content_base64"),
+					resource.TestCheckResourceAttr("data.google_storage_bucket_object_content.default", "content_base64", base64.StdEncoding.EncodeToString([]byte(content))),
 				),
 			},
 		},
 	})
 }
 
+func TestAccDataSourceStorageBucketObjectContent_FileContentBase64(t *testing.T) {
+	acctest.SkipIfVcr(t)
```
kautikdk, line 39: What is the need of skipping VCR test?

kautikdk:
> Thanks for adding the field. Can you add a test for the use case in [hashicorp/terraform-provider-google#10835 (comment)](https://github.com/hashicorp/terraform-provider-google/issues/10835#issue-1093838248)? Thanks.

I wonder if we need that use-case in our test since this is not a bug fix rather a feature request. We are adding one output field only and validating value of that field in the existing test case. I think that should be enough from our side for testing purpose and I am not sure if we need to take whole scenario of original ticket. @zli82016 correct me if I am wrong.

mmv1/third_party/terraform/services/storage/data_source_storage_bucket_object_content_test.go
gurusai-voleti: it is throwing an error in VCR test (replaying mode), as we need only zip file verification disabled the VCR test (replaying mode)

2025/03/24 11:18:58 [DEBUG] [transport] [server-transport 0xc0038a64e0] loopyWriter exiting with error: transport closed by client 
    data_source_storage_bucket_object_content_test.go:53: Step 1/1 error: After applying this test step, the plan was not empty.

https://github.com/GoogleCloudPlatform/magic-modules/pull/13424#issuecomment-2747778931

zli82016: Thanks for adding the test.

`acctest.SkipIfVcr(t)` will skip the test during vcr testing, one step of the checks. 

The test will still be run during the nightly test.

PR #13431, opened 2025-03-21
Fix expandNodeKubeletConfig unset field check.

hoskeri:
Use the schema.ResourceData.HasChange function on the Kubelet readonly port field to detect if the value of the readonly port field was actually set in this request.

GetOk would return a value if it was not present in the config.

Fixes #21227

```release-note:bug
container: fixed a bug in `google_container_cluster` where empty `node_pool_defaults.node_config_defaults` would break in Autopilot clusters. 
```

### DIFF
--- a/mmv1/third_party/terraform/services/container/node_config.go.tmpl
+++ b/mmv1/third_party/terraform/services/container/node_config.go.tmpl
@@ -917,7 +917,7 @@ func schemaNodePoolAutoConfigLinuxNodeConfig() *schema.Schema {
 	}
 }
 
-func expandNodeConfigDefaults(configured interface{}) *container.NodeConfigDefaults {
+func expandNodeConfigDefaults(configured interface{}, d *schema.ResourceData) *container.NodeConfigDefaults {
 	configs := configured.([]interface{})
 	if len(configs) == 0 || configs[0] == nil {
 		return nil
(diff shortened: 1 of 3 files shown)


hoskeri:
@c2thorn please review.

c2thorn:
build may have failed due to a GH transient error, I will retry

mmv1/third_party/terraform/services/container/node_config.go.tmpl
```diff
@@ -926,12 +926,17 @@ func expandNodeConfigDefaults(configured interface{}) *container.NodeConfigDefau
 
 	nodeConfigDefaults := &container.NodeConfigDefaults{}
 	nodeConfigDefaults.ContainerdConfig = expandContainerdConfig(config["containerd_config"])
-	if v, ok := config["insecure_kubelet_readonly_port_enabled"]; ok {
-		nodeConfigDefaults.NodeKubeletConfig = &container.NodeKubeletConfig{
-			InsecureKubeletReadonlyPortEnabled: expandInsecureKubeletReadonlyPortEnabled(v),
-			ForceSendFields:                    []string{"InsecureKubeletReadonlyPortEnabled"},
+
+	// Set a value of insecure_kubelet_readonly_port_enabled only if set in the upstream config.
+ 	if d.HasChange("node_pool_defaults.node_config_defaults.node_kubelet_config.insecure_kubelet_readonly_port_enabled") {
```
c2thorn, line 931: `d.HasChange` Will only check if the value has changed since the last state - meaning if the user had set `insecure_kubelet_readonly_port_enabled` and then updated a separate value, this check would fail and not correctly send `insecure_kubelet_readonly_port_enabled`. 
Also, `node_pool_defaults.node_config_defaults.node_kubelet_config.insecure_kubelet_readonly_port_enabled` is not the Terraform path, `node_kubelet_config` is only in the API path, so this check was always `false` since the Terraform path did not exist. I'm not sure why the test `TestAccContainerCluster_withInsecureKubeletReadonlyPortEnabledDefaultsUpdates` did not run in the build, but I am fairly certain it is failing as-is.

Instead, we need to read the user config directly. I've tested the following manually, but we'll need to confirm with all of the other tests:

```suggestion
	readOnlyPortConfig, _ := d.GetRawConfigAt(cty.GetAttrPath("node_pool_defaults").IndexInt(0).
		GetAttr("node_config_defaults").IndexInt(0).GetAttr("insecure_kubelet_readonly_port_enabled"))

	if readOnlyPortConfig.IsKnown() && !readOnlyPortConfig.IsNull() {
```

you'll also need to add the import
```
	"github.com/hashicorp/go-cty/cty"
```

c2thorn (changes requested):

PR #13435, opened 2025-03-24
Fix failing Managed Kafka Connect resource test failures

johndee7310:
Fix failures in Managed Kafka ConnectCluster and Connector resource tests

This will resolve https://github.com/hashicorp/terraform-provider-google/issues/21912 and https://github.com/hashicorp/terraform-provider-google/issues/21892

```release-note:none

```

### DIFF
--- a/mmv1/products/managedkafka/ConnectCluster.yaml
+++ b/mmv1/products/managedkafka/ConnectCluster.yaml
@@ -49,6 +49,7 @@ examples:
     test_env_vars:
       org_id: 'ORG_ID'
       billing_account: 'BILLING_ACCT'
+    skip_vcr: true
     external_providers: ["time"]
 parameters:
   - name: 'location'
--- a/mmv1/products/managedkafka/ConnectCluster.yaml
+++ b/mmv1/products/managedkafka/ConnectCluster.yaml
@@ -46,6 +46,9 @@ examples:
       connect_cluster_id: 'my-connect-cluster'
       cluster_id: 'my-cluster'
       secondary_subnet_id: 'my-secondary-subnetwork'
+    test_env_vars:
+      org_id: 'ORG_ID'
+      billing_account: 'BILLING_ACCT'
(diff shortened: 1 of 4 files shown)


shuyama1 (commented):
Thanks for working on fixing the tests.

quick question: what do we need to use the default network instead of creating a separate one for this test? We have resource tests of all services running in parallel for our nightly tests, so we'd want to avoid modification to the default network to prevent race conditions

johndee7310:
@shuyama1 I actually tried a version of my code that created new networks and had the subnets defined on that network, but the tests kept failing (ex: see last failing test before switching to the default network https://screenshot.googleplex.com/BTvrfqPjJzTqw6d). I didn't see why the tests kept failing with `The subnetwork resource is already being used` when trying to delete the resources.

Have you seen this before? Would you know I had something wrong before?

shuyama1 (commented):
> @shuyama1 I actually tried a version of my code that created new networks and had the subnets defined on that network, but the tests kept failing (ex: see last failing test before switching to the default network [screenshot.googleplex.com/BTvrfqPjJzTqw6d](https://screenshot.googleplex.com/BTvrfqPjJzTqw6d)). I didn't see why the tests kept failing with `The subnetwork resource is already being used` when trying to delete the resources.
> 
> Have you seen this before? Would you know if I had something wrong before?

Looking into the previous failing issue

shuyama1 (commented):
I think you also need to active the compute service as well, since it's not automatically enabled for newly created projects. Example: https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/templates/terraform/examples/apigee_developer_basic_test.tf.tmpl#L28

johndee7310:
Noted, I'll push an update now!

Izzy-CodersPlanet (approved):
Looks good! ✅

shuyama1 (approved):

PR #13439, opened 2025-03-24
Add tlsSettings field to the compute.BackendService resource

porky256:
```release-note:enhancement
compute: added `tls_settings` field to `google_compute_backend_service` resource
```


mmv1/products/compute/BackendService.yaml
```diff
@@ -127,6 +127,13 @@ examples:
       default_neg_name: 'network-endpoint'
       health_check_name: 'health-check'
       network_name: 'network'
+  - name: 'backend_service_tls_settings'
+    primary_resource_id: 'default'
+    min_version: 'beta'
```
rileykarson, line 132: Should the feature be marked as beta as well, or can we mark the example as GA?

porky256: field itself is GA, however resource used in it (backendAuthenticationConfig) is beta. But I agree, field itself should be marked beta then. Thanks for catching!

rileykarson (approved):

PR #13441, opened 2025-03-24
feat(bigtable) Add support for deletion protection for Logical Views

ron-gal:

```release-note:enhancement
bigtable: added `deletion_protection` field to `google_bigtable_logical_view` resource
```


mmv1/products/bigtable/LogicalView.yaml
```diff
@@ -71,3 +76,7 @@ properties:
     description:
       'The logical view''s select query.'
     required: true
+  - name: 'deletionProtection'
```
c2thorn, line 79: Typically deletion protection fields are either fields from the API body itself, or [a virtual field](https://googlecloudplatform.github.io/magic-modules/reference/resource/#virtual_fields) in MMv1 combined with [custom code](https://googlecloudplatform.github.io/magic-modules/develop/custom-code/) that prevents the delete functionality in Terraform. 
I've looked through the [documentation for logical view](https://cloud.google.com/bigtable/docs/reference/admin/rest/v2/projects.instances.logicalViews#LogicalView), and I don't see an API field. Am I looking in the right place? Or were you wanting this to be a Terraform-only field?

c2thorn (approved):
Noting that API documentation is not updated yet. This is an API field, which does not return a value if defaulting to false. Which means we shouldn't need `default_from_api: true`.

PR #13449, opened 2025-03-24
Scope down RNEG lifecycle rule recommendation

rileykarson:
`lifecycle.create_before_destroy` reorders the plan and can be passed through to other resources with dependencies on the initial resource. We want users to use it only when strictly required.

```release-note:none
```

### DIFF
--- a/mmv1/products/compute/RegionNetworkEndpointGroup.yaml
+++ b/mmv1/products/compute/RegionNetworkEndpointGroup.yaml
@@ -19,9 +19,11 @@ description: |
   A regional NEG that can support Serverless Products, proxying traffic to
   external backends and providing traffic to the PSC port mapping endpoints.
 
-  Recreating a region network endpoint group that's in use by another resource will give a
-  `resourceInUseByAnotherResource` error. Use `lifecycle.create_before_destroy`
-  to avoid this type of error.
+  When in use by a resource that can be updated, recreating a RegionNetworkEndpointGroup
+  will give a `resourceInUseByAnotherResource` error because Terraform will attempt to
+  delete the   RegionNetworkEndpointGroup first, but an in-use RegionNetworkEndpointGroup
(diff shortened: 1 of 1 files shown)


mmv1/products/compute/RegionNetworkEndpointGroup.yaml
```diff
@@ -19,9 +19,11 @@ description: |
   A regional NEG that can support Serverless Products, proxying traffic to
   external backends and providing traffic to the PSC port mapping endpoints.
 
-  Recreating a region network endpoint group that's in use by another resource will give a
-  `resourceInUseByAnotherResource` error. Use `lifecycle.create_before_destroy`
-  to avoid this type of error.
+  When in use by a resource that can be updated, recreating a RegionNetworkEndpointGroup
+  will give a `resourceInUseByAnotherResource` error because Terraform will attempt to
+  delete the   RegionNetworkEndpointGroup first, but an in-use RegionNetworkEndpointGroup
```
rileykarson, line 24:
```suggestion
  delete the  RegionNetworkEndpointGroup first, but an in-use RegionNetworkEndpointGroup
```

mmv1/products/compute/RegionNetworkEndpointGroup.yaml
```diff
@@ -19,9 +19,11 @@ description: |
...
+  can't be deleted in the API. Use `lifecycle.create_before_destroy` to reorder the plan
+  and create the new resource first, allowing the deletion to go through successfully.
```
rileykarson, line 26:
```suggestion
  and create the new resource first, allowing the deletion to go through successfully.
  This is only recommended when strictly necessary, as the `create_before_destroy`
  directive can be passed onto further dependencies, creating unexpected plans.
```

SirGitsalot (approved):

PR #13455, opened 2025-03-25
Deprecate GCR

joshdickersongoogle:
Deprecate GCR.

Fixes https://github.com/hashicorp/terraform-provider-google/issues/19661

```release-note:deprecation
google_container_registry: Deprecated product. Use google_artifact_registry_repository instead.
```


slevenick (changes requested):
We'll need to add this message to the markdown docs as well: https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/third_party/terraform/website/docs/r/container_registry.html.markdown

slevenick (approved):
Thanks!

PR #13457, opened 2025-03-25
Add tags field to Metastore Federation for TagsR2401

rajeshgupthar:
Add tags field to federation resource to allow setting tags on federation resources at creation time.
Part of b/367307391

```release-note:enhancement
metastore: added `tags` field to `metastore_federation` to allow setting tags for federations at creation time
```

### DIFF
--- a/mmv1/third_party/terraform/services/networksecurity/resource_network_security_gateway_security_policy_rule_test.go
+++ b/mmv1/third_party/terraform/services/networksecurity/resource_network_security_gateway_security_policy_rule_test.go
@@ -49,6 +49,8 @@ func TestAccNetworkSecurityGatewaySecurityPolicyRule_update(t *testing.T) {
 }
 
 func TestAccNetworkSecurityGatewaySecurityPolicyRule_multiple(t *testing.T) {
+	// Skip for now to avoid leaking resources until the fix for b/400293188 rolls out
+	t.Skip()
 	t.Parallel()
 
 	context := map[string]interface{}{


mmv1/products/metastore/Federation.yaml
```diff
@@ -144,3 +144,12 @@ properties:
             - 'METASTORE_TYPE_UNSPECIFIED'
             - 'DATAPROC_METASTORE'
             - 'BIGQUERY'
+  - name: 'tags'
+    type: KeyValuePairs
+    description: |
+      A map of resource manager tags.
+      Resource manager tag keys and values have the same definition as resource manager tags.
+      Keys must be in the format tagKeys/{tag_key_id}, and values are in the format tagValues/{tag_value_id}.
+      The field is ignored (both PUT & PATCH) when empty.
```
melinath, line 153: what does it mean that the field is ignored when empty? Especially since this is an immutable field? Possibly this line should just be removed.

```suggestion
```

mmv1/third_party/terraform/services/dataprocmetastore/resource_dataproc_metastore_federation_test.go
```diff
@@ -0,0 +1,73 @@
...
+	}
+	tagKey := acctest.BootstrapSharedTestTagKey(t, "metastore-federations-tagkey")
+	tagValue := acctest.BootstrapSharedTestTagValue(t, "metastore-federations-tagvalue", tagKey)
+
+	acctest.VcrTest(t, resource.TestCase{
+		PreCheck:                 func() { acctest.AccTestPreCheck(t) },
+		ProtoV5ProviderFactories: acctest.ProtoV5ProviderFactories(t),
+		Steps: []resource.TestStep{
+			{
+				Config: testAccMetastoreFederationTags(context, map[string]string{org + "/" + tagKey: tagValue}),
```
melinath, line 27: Our test reader isn't able to parse this config, which is why the tag field is showing up as missing tests. I'd recommend simplifying the formatting - specifically, to have a single string that you format the tag values into, instead of splitting the config into multiple chunks that get concatenated.

I see that this is a common pattern that you've copied; that is unfortunate, since it means detection for all those tests is also broken. Still, we should fix it here :-)

melinath (changes requested):

mmv1/third_party/terraform/services/dataprocmetastore/resource_dataproc_metastore_federation.go
```diff
@@ -0,0 +1,672 @@
+// Copyright (c) HashiCorp, Inc.
```
melinath, line 1: this file should be deleted; it's already being generated based on the yaml.

mmv1/third_party/terraform/services/dataprocmetastore/resource_dataproc_metastore_federation_test.go
SamanthaMathews: But this acceptance test was passing on running
 make testacc TEST=./google/services/dataprocmetastore TESTARGS='-run=TestAccMetastoreFederation_tags'

melinath: yes, the tests are able to pass, but our static test analyzer can't parse this to check what fields are covered.

melinath (changes requested):
please fix the merge conflicts & address the feedback in https://github.com/GoogleCloudPlatform/magic-modules/pull/13457#discussion_r2012739269

mmv1/third_party/terraform/services/dataprocmetastore/resource_dataproc_metastore_federation_test.go
```diff
@@ -0,0 +1,73 @@
...
+		},
+	})
+}
+
+func testAccMetastoreFederationTags(context map[string]interface{}, tags map[string]string) string {
+
+	r := acctest.Nprintf(`
+
+       resource "google_dataproc_metastore_service" "default" {
+         service_id = "metastore-srv-sep16"
```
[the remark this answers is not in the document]
melinath: "sweepable" meaning that we can automatically clean up dangling resources if the test ever panics partway through.

rajeshgupthar:
https://github.com/GoogleCloudPlatform/magic-modules/pull/13625 created a new PR for the same Change.

PR #13458, opened 2025-03-25
Add 1 minute delay after project creation before running KMS KAJ test

bohrjoce:
I think this might be flaky because when CSS can't find a project, it returns unenrolled. So this passes when the CSS has registered the ephemeral project, but fails when the project is not registered yet.

Fixes: https://github.com/hashicorp/terraform-provider-google/issues/21508

```release-note:none
```


c2thorn (approved):
seems reasonable to me

PR #13459, opened 2025-03-25
Set google_compute_region_url_map > path_matcher > default_service to required.

Dawid212:
This PR sets google_compute_region_url_map > path_matcher > default_service to required. 

### Related Issues

- Resolves: https://github.com/hashicorp/terraform-provider-google/issues/21069

```release-note:bug
compute: fixed `google_compute_url_map.path_matcher.default_service` to be required
```

### DIFF
--- a/mmv1/products/compute/RegionUrlMap.yaml
+++ b/mmv1/products/compute/RegionUrlMap.yaml
@@ -244,6 +244,7 @@ properties:
           # exactly_one_of:
           #   - path_matchers.0.default_service
           #   - path_matchers.0.default_url_redirect
+          required: true
           resource: 'RegionBackendService'
           imports: 'selfLink'
         - name: 'description'
--- a/mmv1/products/compute/UrlMap.yaml
+++ b/mmv1/products/compute/UrlMap.yaml
@@ -303,6 +303,7 @@ properties:
           #   - path_matchers.0.default_service
           #   - path_matchers.0.default_url_redirect
           #   - path_matchers.0.default_route_action.0.weighted_backend_services
+          required: true
           custom_expand: 'templates/terraform/custom_expand/reference_to_backend.tmpl'
           resource: 'BackendService'
           imports: 'selfLink'


zli82016:
It is a breaking change to make a field required. I am going to confirm with the service team if this field should be always required.

zli82016:
@Dawid212 , I don't think this field `path_matcher.default_service` should be set be required. This field is not always be required. Can you share the reason that you make this changes? Thanks.

mmv1/products/compute/RegionUrlMap.yaml
```diff
@@ -244,6 +244,7 @@ properties:
           # exactly_one_of:
           #   - path_matchers.0.default_service
           #   - path_matchers.0.default_url_redirect
+          required: true
```
zli82016, line 247: Instead of making this field `required`, we can add a note to the `description` of this field that one of `default_service` and `default_url_redirect` should be provided.

mmv1/products/compute/UrlMap.yaml
```diff
@@ -303,6 +303,7 @@ properties:
           #   - path_matchers.0.default_service
           #   - path_matchers.0.default_url_redirect
           #   - path_matchers.0.default_route_action.0.weighted_backend_services
+          required: true
```
zli82016, line 306: Instead of making this field `required`, we can add a note to the `description` of this field that one of `default_service` and `default_url_redirect` should be provided.

zli82016 (changes requested):

glouasse:
> @Dawid212 , I don't think this field `path_matcher.default_service` should be set to `required`. This field is not always be required. Can you share the reason that you make this changes? Thanks.

Hello, if I may help : you're correct. I opened the issue related to this pull request. The problem is that actually ANY `path_matcher` needs a `default_service`(in the documentation, it's supposed to be optional). This leads to many `default_service` for a load balancer (the required one on `google_compute_region_url_map` and any backend in `path_matcher`) and the result is a unusable load balancer. The only required `default_service` should be on `google_compute_region_url_map` and optional for each `path_matcher`.

GeoffreyLou:
> @Dawid212 , I don't think this field `path_matcher.default_service` should be set to `required`. This field is not always be required. Can you share the reason that you make this changes? Thanks.

Hello, if I may help : you're correct. I opened the issue related to this pull request. The problem is that actually **ANY** `path_matcher` needs a `default_service`(in the documentation, it's supposed to be **optional**). This leads to many `default_service` for a load balancer (the required one on `google_compute_region_url_map` and any backend in `path_matcher`) and the result is a **unusable load balancer** -> the default url will forward to the last `default_service` set.

The only required `default_service` should be on `google_compute_region_url_map` and optional for each `path_matcher`.

zli82016:
> > @Dawid212 , I don't think this field `path_matcher.default_service` should be set to `required`. This field is not always be required. Can you share the reason that you make this changes? Thanks.
> 
> Hello, if I may help : you're correct. I opened the issue related to this pull request. The problem is that actually **ANY** `path_matcher` needs a `default_service`(in the documentation, it's supposed to be **optional**). This leads to many `default_service` for a load balancer (the required one on `google_compute_region_url_map` and any backend in `path_matcher`) and the result is a **unusable load balancer** -> the default url will forward to the last `default_service` set.
> 

@GeoffreyLou, thanks for the information. Based on my understanding, one of  `path_matchers.0.default_service` and `path_matchers.0.default_url_redirect` must be provided.  But the nested `exactly_one_of` validation won't work for array path matchers yet (github.com/hashicorp/terraform-plugin-sdk/issues/470).

What we can do is to add a note in the `description` of the fields `path_matchers.0.default_service` and `path_matchers.0.default_url_redirect`.

PR #13464, opened 2025-03-26
feat: (storage) added md5hexhash field in storage object

gurusai-voleti:
Fixes: https://github.com/hashicorp/terraform-provider-google/issues/3888

```release-note:enhancement
storage: added md5hexhash field in storage object
```

### DIFF
--- a/mmv1/third_party/terraform/services/storage/resource_storage_bucket_object.go
+++ b/mmv1/third_party/terraform/services/storage/resource_storage_bucket_object.go
@@ -17,6 +17,7 @@ import (
 
 	"crypto/sha256"
 	"encoding/base64"
+	"encoding/hex"
 	"net/http"
 
 	"google.golang.org/api/googleapi"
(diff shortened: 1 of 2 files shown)


kautikdk (commented):
Can you please check if there is a way to achieve similar results by Terraform functions or specific providers like mentioned in this comment: https://github.com/hashicorp/terraform-provider-google/issues/3888#issuecomment-1520539496 ?
If there is, I don't think we have to introduce a new field just for a specific use case.

gurusai-voleti:
> Can you please check if there is a way to achieve similar results by Terraform functions or specific providers like mentioned in this comment: [hashicorp/terraform-provider-google#3888 (comment)](https://github.com/hashicorp/terraform-provider-google/issues/3888#issuecomment-1520539496) ? If there is, I don't think we have to introduce a new field just for a specific use case.

I see there is no builtin function for this

trodge (approved):

PR #13468, opened 2025-03-26
Detect if PR was already merged

melinath:

```release-note:none

```

### DIFF
--- a/.ci/magician/github/set.go
+++ b/.ci/magician/github/set.go
@@ -157,9 +157,10 @@ func (gh *Client) MergePullRequest(owner, repo, prNumber, commitSha string) erro
 				return fmt.Errorf("failed to check if PR was already merged: %w", err)
 			}
 			if pr.Merged {
+				fmt.Printf("Pull request %s was already merged\n", prNumber)
 				return nil
 			}
-			fmt.Printf("Pull request %s was not already merged\n", prNumber)
+			fmt.Printf("Pull request %s wasn't already merged\n", prNumber)
 		}
 		return fmt.Errorf("failed to merge pull request: %w", err)
 	}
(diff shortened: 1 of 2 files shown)


melinath:

.ci/magician/github/set.go
```diff
@@ -149,7 +149,19 @@ func (gh *Client) MergePullRequest(owner, repo, prNumber, commitSha string) erro
 			fmt.Printf("Pull request %s is already being merged\n", prNumber)
 			return nil
 		}
-		return fmt.Errorf("failed to merge pull request: %s", err)
+		// Check if the PR is already merged (returns 405 Pull Request is not mergeable)
+		if strings.Contains(err.Error(), "Pull Request is not mergeable") {
+			fmt.Printf("Pull request %s is not mergeable; checking if it was already merged\n", prNumber)
```
ScottSuarez, line 154: maybe we should wait 10 seconds or before we check this.

ScottSuarez (approved):

.ci/magician/github/set.go
```diff
@@ -149,7 +149,19 @@ func (gh *Client) MergePullRequest(owner, repo, prNumber, commitSha string) erro
...
+			pr, err := gh.GetPullRequest(prNumber)
+			if err != nil {
+				return fmt.Errorf("failed to check if PR was already merged: %w", err)
+			}
+			if pr.Merged {
+				return nil
```
ScottSuarez, line 160:
```suggestion
				fmt.Printf("Pull request %s is already merged\n", prNumber)
				return nil
```

.ci/magician/github/set.go
melinath: what would that help with?

melinath: is this a blocker?

ScottSuarez: hmm, not a blocker. I thought it might help if there is an eventual consistency issue on the API

melinath: I think I'd like to leave it as-is for now, in that case. If it turns out to be eventually consistent we can add a sleep later.

ScottSuarez: Makes sense !

ScottSuarez (approved):

PR #13470, opened 2025-03-27
Colab: Use location instead of region for Colab resources in base_url

bcreddy-gcp:

```release-note:bug
colab: fixed an issue where `google_colab_*` resources incorrectly required a provider-level region matching the resource location
```


bcreddy-gcp:
https://github.com/hashicorp/terraform-provider-google/issues/21893#issuecomment-2755766332
cc: @rileykarson @SirGitsalot

mmv1/products/colab/product.yaml
```diff
@@ -16,10 +16,10 @@ name: 'Colab'
 display_name: 'Colab Enterprise'
 versions:
   - name: 'ga'
-    base_url: 'https://{{region}}-aiplatform.googleapis.com/v1/'
+    base_url: 'https://{{location}}-aiplatform.googleapis.com/v1/'
```
BBBmau, line 19: i'm not sure what this is fixing since location represents the same value as region. Can you elaborate a bit more?

SirGitsalot: None of the colab resources have a `region` field, they all use `location`. The two generally have the same semantic meaning, but by declaring the `base_url` to use `region` it means that `region` has to be defined somewhere. Since there's no 'region` field in the resources, it has to be defined at the provider level or you won't be able to create colab resources.

I don't think this is a breaking change, but since @rileykarson has already been tagged on this PR I'll defer to him.

rileykarson: What happens if the provider region and resource location mismatch? Is there an error? I think this is just a bugfix unless the locational endpoint works for resources in another location.

> i'm not sure what this is fixing since location represents the same value as region

* `region` represents the GCE regions and can generally be assumed to be reusable across domains, which is why we can support a provider-level default.
* `location` is a domain-specific (at the service granularity) set of locations that are _generally_ the GCE regions but can also include multiregions, `global`, zones, etc. Locations can't safely be reused between services.

So in this case while the region value technically worked it's not safe to have reused it given this service is not necessarily constrained to the GCE regions.

bcreddy-gcp: It will be an INVALID_ARGUMENT error.

rileykarson: Thanks! This should definitely be safe then, since any mismatch will have failed. I did a quick pass and all the resources in the product have a required `location` field and no region field, so no concerns from me!

rileykarson (approved):

PR #13471, opened 2025-03-27
Add deletion_protection field to Metastore Federation

rajeshgupthar:
Part of b/366209903

```release-note:enhancement
metastore: added `deletion_protection` field to `google_dataproc_metastore_federation`
```

### DIFF
--- a/mmv1/products/metastore/Federation.yaml
+++ b/mmv1/products/metastore/Federation.yaml
@@ -44,6 +44,7 @@ iam_policy:
     - 'projects/{{project}}/locations/{{location}}/federations/{{federation_id}}'
     - '{{federation_id}}'
 custom_code:
+  pre_delete: 'templates/terraform/pre_delete/metastore_federation.go.tmpl'
 examples:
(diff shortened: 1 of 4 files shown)


mmv1/products/metastore/Federation.yaml
```diff
@@ -44,19 +44,30 @@ iam_policy:
...
       federation_id: 'metastore-fed'
       service_id: 'metastore-service'
+virtual_fields:
+  - name: 'deletion_protection'
+    description: |
+      Whether Terraform will be prevented from destroying the federation. Defaults to false.
+      When the field is set to true in Terraform state, a `terraform apply`
+      or `terraform destroy` that would delete the federation will fail.
+    type: Boolean
+    default_value: false
```
slevenick, line 70: With a default of false we won't prevent the destruction of most resources. It will only be opt-in. That is different from most of the existing implementations of deletion_protection, are you sure that's what you want here?

slevenick (approved):

PR #13093, opened 2025-02-17
Feature gap: Add missed `enable_flow_logs` & `state` to subnetwork

Cezarus27:
This patch introduces new fields to the Terraform provider for the `subnetwork` resource:  
- `enable_flow_logs` - a field that allows enabling flow logs. If the `log_config` field is not set in the configuration, the default `log_config` value returned by the API will be stored in `tfstate`.  
- `state` - a read-only field. During testing, it was observed that the API does not return this field in response to a GET request. Nevertheless, the field is stored in `tfstate` as an empty string.  

**Release Note Template for Downstream PRs (will be copied)**

See [Write release notes](https://googlecloudplatform.github.io/magic-modules/contribute/release-notes/) for guidance.

```release-note:enhancement
compute: added `enable_flow_logs` and `state` fields to `google_compute_subnetwork` resource
```

PR #13485, opened 2025-03-27
Re-deprecate enable_flow_logs

c2thorn:
closes https://github.com/hashicorp/terraform-provider-google/issues/21950

This PR re-deprecates `enable_flow_logs` and removes its functionality by reverting the expander logic related to it.

Testing the config mentioned in this comment: https://github.com/hashicorp/terraform-provider-google/issues/21950#issuecomment-2748876288
gpaste/5824658669305856 shows `enableFlowLogs` and `logConfig.enabled` to be correctly set to true.

Unfortunately, https://github.com/GoogleCloudPlatform/magic-modules/pull/13093 also set `log_config` to `Optional` + `Computed`, which removed its ability to be removed. To fix this, I removed `default_from_api` but also modified the diffsuppress for `log_config` to check whether or not the `log_config` exists in the user's config. If it does not exist but the deprecated `enable_flow_logs` does, we should suppress the diff so users aren't hit with a breaking change to disable their flow logs.
To test this, I applied the following config in 6.27:
```
resource "google_compute_subnetwork" "this" {
  provider      = google-beta
  name          = "<name>"
  ip_cidr_range = "10.2.0.0/16"
  region        = "us-central1"
  network       = google_compute_network.custom-test.name

  enable_flow_logs = true
}
```

Here is the difference when upgrading to versions with the fix w/ and w/o diffsuppress:

Without diffsuppress + enable_flow_logs=true + no log_config:
```
  # google_compute_subnetwork.this will be updated in-place
  ~ resource "google_compute_subnetwork" "this" {
      - log_config {
          - aggregation_interval = "INTERVAL_5_SEC" -> null
          - flow_sampling        = 0.5 -> null
          - metadata             = "EXCLUDE_ALL_METADATA" -> null
          - metadata_fields      = [] -> null
        }
    }
```

With diffsuppress
```
```

Given the circumstances of the O+C in `log_config` only being present in 6.26-6.27, and the diff suppress, I believe we move forward with removing O+C. There's a possibility that someone created a subnetwork in 6.26 or 6.27, enabled flow logs WITHOUT enable_flow_logs or log_config (maybe outside of Terraform), and upgrading to 6.28 will show a diff to remove flow logs since log_config is not in their configuration and no longer O+C. I believe this should be a very small set and we'd be saving more user pain overall.

```release-note:deprecation
compute: deprecated `enable_flow_logs` in favor of `log_config`.  If `log_config` is present, flow logs are enabled and `enable_flow_logs` can be safely removed.
```
```release-note:bug
compute: fixed a regression in `google_compute_subnetwork` where setting `log_config` would not enable flow logs without `enable_flow_logs` also being set to true. To enable or disable flow logs, please use `log_config`. `enable_flow_logs` is now deprecated and will be removed in the next major release.
```


c2thorn:
Found an issue with the upgrade path 6.25 -> 6.26 -> patch, will look into it

NickElliot (approved):
LGTM -- walked through test scenarios and tried to come up with any missed holes.

This change as implemented will be a semver break due to removing computed status from `log_config`, but this field has only been available in a computed capacity for a few weeks so it is unlikely users implemented states with other attributes dependent on that field value, and even if they did, the present state of the resource is highly disruptive to far more existing users attempting to update.

PR #13487, opened 2025-03-28
Set a default value for envs in Cloud Run Job

oliver-hohn:
What?
- Make a similar change to https://github.com/GoogleCloudPlatform/magic-modules/pull/13363 but for the `google_cloud_run_v2_job` resource.
  - Fixes https://github.com/hashicorp/terraform-provider-google/issues/7467
  - Fixes https://github.com/hashicorp/terraform-provider-google/issues/10634

Why?
- Currently, when a new env variable is added to the `env` set, the `env` set is re-ordered which results in a large diff, see https://github.com/hashicorp/terraform-provider-google/issues/7467 for an example. The cause for this seems to be that in the Terraform state the env `value` is `""` when using a `value_source` (i.e. a reference to a `google_secret_manager_secret`), but in the plan the value is `nil` (as it has not been set), which then triggers all of the envs to have changed.

```release-note:bug
cloudrunv2: fixed the diffs for unchanged `template.template.containers.env` in `google_cloud_run_v2_job` resource
```

### DIFF
--- a/mmv1/products/cloudrunv2/Job.yaml
+++ b/mmv1/products/cloudrunv2/Job.yaml
@@ -349,6 +349,10 @@ properties:
                         required: true
                       - name: 'value'
                         type: String
+                        # env is a set.
+                        # The env.value has value "" in Terraform state, but it has value nil in Terraform plan,
+                        # which causes the diffs for unchanged env. default_value: "" is to suppress the diffs.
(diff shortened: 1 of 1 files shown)


ScottSuarez (approved):

ScottSuarez:
waiting for integration tests

PR #13489, opened 2025-03-28
compute: Documented empty string for `load_balancing_scheme`

wyardley:
While it's not listed in the possible enum values, empty string is also used in certain situations as a valid value for a
compute_forwarding_rule's `load_balancing_scheme`. Since the valid values are auto-generated, add a note about the empty string as well.

This can be confusing, since using `null` or unsetting the value will result in the default (`"EXTERNAL"`).

Fixes https://github.com/hashicorp/terraform-provider-google/issues/22098

```release-note:none

```

### DIFF
--- a/mmv1/products/compute/ForwardingRule.yaml
+++ b/mmv1/products/compute/ForwardingRule.yaml
@@ -368,6 +368,10 @@ properties:
     description: |
       Specifies the forwarding rule type.
 
+      Note that an empty string value (`""`) is also supported for some use
+      cases, for example PSC (private service connection) regional forwarding
+      rules.
+
       For more information about forwarding rules, refer to
       [Forwarding rule concepts](https://cloud.google.com/load-balancing/docs/forwarding-rule-concepts).
     default_value: "EXTERNAL"


BBBmau (changes requested):
we should add a test for this in the event that this behavior changes in the future. Can you go ahead and add one?

wyardley:
> we should add a test for this in the event that this behavior changes in the future. Can you go ahead and add one?

I think the existing examples should cover this probably, since IIRC, tests are generated based on them.

For example, [this one](https://github.com/GoogleCloudPlatform/magic-modules/blob/678b9623bb39cc8ee03e8ec6dc75e873f3f65396/mmv1/templates/terraform/examples/private_service_connect_google_apis.tf.tmpl#L28-L35)

Some additional context / discussion in the linked issue.

wyardley:
@BBBmau does that make sense to you? I think the docs you're talking about should already be there, but let me know if you think something is missing.

BBBmau (approved):
thanks! This makes sense, apologies for the late review.

PR #13492, opened 2025-03-31
Strip redundant information from descriptions in autogen resources

slevenick:
Remove boilerplate from descriptions. This gets autogenerated in here during OpenAPI creation, and we add it again during MMv1 generation, so strip it before we make the MMv1 YAML

```release-note:none

```

### DIFF
--- a/mmv1/openapi_generate/parser.go
+++ b/mmv1/openapi_generate/parser.go
@@ -304,7 +304,7 @@ func parseOpenApi(resourcePath, resourceName string, root *openapi3.T) []any {
 		if strings.TrimSpace(description) == "" {
 			description = "No description"
 		}
-		paramObj.Description = trimSpacesFromDescription(description)
+		paramObj.Description = trimDescription(description)
 
 		if param.Value.Name == "requestId" || param.Value.Name == "validateOnly" || paramObj.Name == "" {
 			continue
@@ -410,7 +410,7 @@ func writeObject(name string, obj *openapi3.SchemaRef, objType openapi3.Types, u
 		description = "No description"
 	}
 
-	field.Description = trimSpacesFromDescription(description)
+	field.Description = trimDescription(description)
 
 	if urlParam {
 		field.UrlParamOnly = true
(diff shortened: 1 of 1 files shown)


mmv1/openapi_generate/parser.go
```diff
@@ -451,7 +451,12 @@ func buildProperties(props openapi3.Schemas, required []string) []*api.Type {
 
 // Trims whitespace from the ends of lines in a description to force multiline
 // formatting for strings with newlines present
-func trimSpacesFromDescription(description string) string {
+// Also trim "Output only." and "Required." from descriptions as this gets duplicated
+func trimDescription(description string) string {
+	description, _ = strings.CutPrefix(description, "Optional. ")
+	description, _ = strings.CutPrefix(description, "Output only. ")
+	description, _ = strings.CutPrefix(description, "Required. ")
+	description, _ = strings.CutPrefix(description, "Immutable. ")
```
melinath, line 459: Immutable actually doesn't currently get added to descriptions by MMv1 (though arguably it should be IMO - but that might be better in a separate PR)

mmv1/openapi_generate/parser.go
```diff
@@ -451,7 +451,12 @@ func buildProperties(props openapi3.Schemas, required []string) []*api.Type {
 
 // Trims whitespace from the ends of lines in a description to force multiline
 // formatting for strings with newlines present
-func trimSpacesFromDescription(description string) string {
+// Also trim "Output only." and "Required." from descriptions as this gets duplicated
+func trimDescription(description string) string {
+	description, _ = strings.CutPrefix(description, "Optional. ")
+	description, _ = strings.CutPrefix(description, "Output only. ")
```
melinath, line 457: Possibly also `[Output Only]` and `[Output only]`?

mmv1/openapi_generate/parser.go
```diff
@@ -451,7 +451,12 @@ func buildProperties(props openapi3.Schemas, required []string) []*api.Type {
 
 // Trims whitespace from the ends of lines in a description to force multiline
 // formatting for strings with newlines present
-func trimSpacesFromDescription(description string) string {
+// Also trim "Output only." and "Required." from descriptions as this gets duplicated
+func trimDescription(description string) string {
```
melinath, line 455: This should potentially do a loop while a prefix remains to address cases like https://github.com/GoogleCloudPlatform/magic-modules/blob/8ce3b475fd363aea309851730f35bd82d97044a9/mmv1/products/developerconnect/GitRepositoryLink.yaml#L96

melinath (changes requested):

mmv1/openapi_generate/parser.go
slevenick: Have we seen that? That doesn't look like the standard form that gets generated

slevenick: Ah, developer connect. It looks like they manually wrote [Output only] in their comment that gets copied downstream. I'd lean towards not including that as it's a one-off manual addition

melinath: okay sgtm

mmv1/openapi_generate/parser.go
melinath: ignoring this as a one-time issue

melinath (approved):
LGTM apart from Immutable

PR #13495, opened 2025-03-31
Strip immutable from descriptions in autogen resources

melinath:
This is a follow-up to https://github.com/GoogleCloudPlatform/magic-modules/pull/13492 - after thinking about it some more, we should just have this in place - regardless of what we're generating from MMv1 today - so that we don't have to go in and fix them later if/when we start generating that string from MMv1.

```release-note:none

```


melinath:
Opened https://github.com/hashicorp/terraform-provider-google/issues/22180 to track generating the word Immutable.

slevenick (approved):

PR #13496, opened 2025-03-31
HTTP transport functions for fwprovider

NickElliot:
`SendRequest()` is largely the same as in `transport_tpg`, but is updated for "diags" error handling

`getProjectFromFrameworkSchema()` is updated to a more generalized `getProviderDefaultFromFrameworkSchema()`, the original function as implemented by Sarah is largely accommodating of this change, "project" as a string being something it needed supplied as an arg rather than being inbuilt to the function

`ReplaceVarsRecursive()` and related functions have been rewritten to work with the `[crud]Request` types from plugin framework, relying on interface (type) checking to access the correct values when needed

```release-note:none

```


mmv1/third_party/terraform/fwprovider/framework_provider.go.tmpl
```diff
@@ -319,6 +319,7 @@ func (p *FrameworkProvider) DataSources(_ context.Context) []func() datasource.D
         resourcemanager.NewGoogleClientConfigDataSource,
         resourcemanager.NewGoogleClientOpenIDUserinfoDataSource,
         {{- if ne $.TargetVersionName "ga" }}
+        resourcemanager.NewGoogleProjectFWDataSource,
```
c2thorn, line 322: let's move the datasource to test-only for now, like in https://github.com/GoogleCloudPlatform/magic-modules/pull/12741/files#diff-7c30a2afcc18eb3bf49388e586c1bcb0f84fcb5768c5a9366d4d34f0dd9dd548

mmv1/third_party/terraform/services/resourcemanager/fw_data_source_google_project.go
```diff
@@ -0,0 +1,158 @@
...
+	ProjectId types.String `tfsdk:"project_id"`
+	Name      types.String `tfsdk:"name"`
+	OrgId     types.String `tfsdk:"org_id"`
+	FolderId  types.String `tfsdk:"folder_id"`
+	Number    types.String `tfsdk:"number"`
+}
+
+// Metadata returns the data source type name.
+func (d *GoogleProjectFWDataSource) Metadata(_ context.Context, req datasource.MetadataRequest, resp *datasource.MetadataResponse) {
+	resp.TypeName = req.ProviderTypeName + "_fwprovider_project"
```
c2thorn, line 47: shall we shorten to something like `_fw_project` or `project_fw`? We should probably standardize to something for now. Sarah had https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/third_party/terraform/fwprovider/data_source_provider_config_plugin_framework.go#L58, but that feels different enough

c2thorn (changes requested):
seems good on initial pass. Want to add some simple datasource style tests for it?

c2thorn (approved):

PR #13503, opened 2025-04-01
Add Finding resource template

sarthakvashistha:
Related to http://b/394718493.
Issue No: https://b.corp.google.com/issues/394718493

Documentation: https://cloud.google.com/security-command-center/docs/how-to-api-create-manage-findings

Template for the REST resource - https://cloud.google.com/security-command-center/docs/reference/rest/v2/organizations.sources.locations.findings

```release-note:REPLACEME
`google_scc_v2_organization_source_finding`
```

### DIFF
--- a/.ci/magician/cmd/interfaces.go
+++ b/.ci/magician/cmd/interfaces.go
@@ -33,6 +33,7 @@ type GithubClient interface {
 	PostComment(prNumber, comment string) error
 	UpdateComment(prNumber, comment string, id int) error
 	RequestPullRequestReviewers(prNumber string, reviewers []string) error
+	RemovePullRequestReviewers(prNumber string, reviewers []string) error
 	AddLabels(prNumber string, labels []string) error
 	RemoveLabel(prNumber, label string) error
 	CreateWorkflowDispatchEvent(workflowFileName string, inputs map[string]any) error
--- a/.ci/magician/cmd/mock_github_test.go
+++ b/.ci/magician/cmd/mock_github_test.go
@@ -74,6 +74,11 @@ func (m *mockGithub) RequestPullRequestReviewers(prNumber string, reviewers []st
 	return nil
 }
 
+func (m *mockGithub) RemovePullRequestReviewers(prNumber string, reviewers []string) error {
+	m.calledMethods["RemovePullRequestReviewers"] = append(m.calledMethods["RemovePullRequestReviewers"], []any{prNumber, reviewers})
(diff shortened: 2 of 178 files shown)


mmv1/products/securitycenterv2/Findings.yaml
```diff
@@ -0,0 +1,81 @@
+# Copyright 2024 Google Inc.
```
melinath, line 1:
```suggestion
# Copyright 2025 Google Inc.
```

melinath (changes requested):
The unit tests are currently failing with:

```
provider_test.go:25: err: resource google_scc_v2_findings: All fields are ForceNew or Computed w/out Optional, Update is superfluous
```

Either at least one field needs to be updatable, or the whole resource should be marked immutable (`immutable: true` at the resource level instead of on individual fields.)

mmv1/products/securitycenterv2/Findings.yaml
```diff
@@ -0,0 +1,81 @@
...
+    required: true
+    immutable: true
+  - name: 'location_id'
+    type: String
+    description: |
+      When data residency controls are enabled, this attribute specifies the location in which the resource is located and applicable.
+      If not provided, Use global as default.
+    url_param_only: true
+    required: true
+    immutable: false
```
melinath, line 65: location is almost never immutable. Just want to confirm that moving a finding between locations is a valid operation on update? The docs linked only give an example of [updating a finding's state](https://cloud.google.com/security-command-center/docs/how-to-api-create-manage-findings#updating_a_findings_state) - and state doesn't even seem to be a field at the moment.

melinath: sorry, mistyped - location is almost never _not_ immutable.

sarthakvashistha: so I checked this https://pkg.go.dev/cloud.google.com/go/securitycenter/apiv2/securitycenterpb and verified that- location is not updatable as it depends on the ssc location settings.. and the only functions available for update are the state changes, so I've updated it now, checking internally as well with the test run fixes https://github.com/GoogleCloudPlatform/magic-modules/pull/13503/commits/3486100c9057cf190686af8e8dcf546fc3f83c47

svgoogl:
@melinath I've addressed all review comments, will open and link to a new PR from the official username as the current one may have compliance and sync issues going forward.

PR #13509, opened 2025-04-01
Add initial variant patterns

roaks3:
This adds the new `api_variant_patterns` field to mmv1 and metadata yaml files. Its use is described in go/terraform-resource-variant-coverage, which is basically to capture the distinction of resource "variants".

This should capture all variants that we have clear data for, however when we start matching these values to the API, we should be able to see if there are any remaining gaps.

```release-note:none

```


melinath (approved):
LGTM unless this is no longer what we want the values to look like

roaks3:
Yea, I think let's stick with this and see if we need to do anything differently regarding http_rules or otherwise later

PR #13510, opened 2025-04-01
Pub/Sub - Change Allowed Persistence Regions to Set in 'google_pubsub_topic'

abbrowne126:

```release-note:breaking-change
Fixes https://github.com/hashicorp/terraform-provider-google/issues/21650

This resource field has a permadiff as it stands now. Changing its type to set should resolve this.

NOTE: This is a breaking change, but should be allowed in a minor release [as per the docs](https://googlecloudplatform.github.io/magic-modules/breaking-changes/make-a-breaking-change/#in-minor-releases).
```

### DIFF
--- a/mmv1/products/pubsub/Topic.yaml
+++ b/mmv1/products/pubsub/Topic.yaml
@@ -130,6 +130,7 @@ properties:
     default_from_api: true
     properties:
       - name: 'allowedPersistenceRegions'
+        is_set: true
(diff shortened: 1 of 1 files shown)


hao-nan-li (commented):
Looks like this PR contains a breaking change that can only be included in major releases. The next major release is scheduled on August or September 2025.

PR #13513, opened 2025-04-02
Support custom subject alternative name for CloudSQL instances

pandirigoog:

```release-note:enhancement
sql: added `custom_subject_alternative_names` field to `instances` resource
```

### DIFF
--- a/mmv1/third_party/terraform/services/sql/resource_sql_database_instance.go.tmpl
+++ b/mmv1/third_party/terraform/services/sql/resource_sql_database_instance.go.tmpl
@@ -82,7 +82,7 @@ var (
 		"settings.0.ip_configuration.0.ssl_mode",
 		"settings.0.ip_configuration.0.server_ca_mode",
 		"settings.0.ip_configuration.0.server_ca_pool",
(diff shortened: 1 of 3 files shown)


slevenick (commented):
Looks like it's failing with:

Received the following error from Certificate Authority Service (CAS): Permission 'privateca.certificates.create' denied on '<CAS_CA_POOL>'
Are you able to run this yourself locally? It could be an environment issue

pandirigoog:
> Looks like it's failing with:
> 
> Received the following error from Certificate Authority Service (CAS): Permission 'privateca.certificates.create' denied on '<CAS_CA_POOL>' Are you able to run this yourself locally? It could be an environment issue

Yes, I was able to run the test locally with my own project that has the CAS ca pool set-up. Any instructions on which project will be used for running the test ? Or how do we ensure that the set-up exists for these projects?

slevenick (changes requested):
We chatted offline. There's an expectation that all permissions are granted as part of the test running process for most tests, so we'll need to see what permissions are missing and grant them as part of this test config

slevenick (approved):

PR #13516, opened 2025-04-02
Adding terraform support for GKEBackup Backup and Restore channel resources.

aru31:
This PR adds support for newly created GKEBackup channel resources in beta version.

### DIFF
--- a/mmv1/products/gkebackup/BackupChannel.yaml
+++ b/mmv1/products/gkebackup/BackupChannel.yaml
@@ -1,0 +1,116 @@
+# Copyright 2025 Google Inc.
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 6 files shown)


ScottSuarez (commented):
looks like the tests are failing because the api endpoint doesn't exist

ScottSuarez:
suspect this resource is pre-release?

aru31:
> suspect this resource is pre-release?

The newly added tests are failing in this PR because the new fields are hidden behind API visibility label. I will wait till the visibility labels are removed and pushed to production and then re-trigger the tests.

ScottSuarez:
converted to draft for now. Feel free to mark as ready for review whenever you want another pass.

ScottSuarez (changes requested):
i am okay with these changes but I notice your PM says that these tests cannot run in parallel. Could the tests in question be only handwritten and adopt a t.Run test suite pattern?

The auto-gen examples should be excluded from test generation with the appropriate flags

aru31:
[Commit](https://github.com/GoogleCloudPlatform/magic-modules/pull/13516/commits/d6d06339dd2b6532fe5ec74b72adaf9c078e6703) solved the issue of parallel tests. There shouldn't be parallel tests issue anymore.

ScottSuarez (approved):

PR #13520, opened 2025-04-02
Handle custom decoders as part of setting fields in post_create

melinath:
https://github.com/GoogleCloudPlatform/magic-modules/pull/12939 started triggering generation of post-create code, but it missed adding support for decoders (which are already used by the [equivalent code for pulling computed fields from operation responses](https://github.com/GoogleCloudPlatform/magic-modules/blob/d0f8dc4dd63e11d1b7cdabe28458f8c66f65140d/mmv1/templates/terraform/resource.go.tmpl#L346-L354).) I also added setting of fields for resources that use PollAsync on create.

Decoder support is necessary to properly handle some resources, like bigqueryanalyticshub_listing_subscription. In this specific case, we do end up just moving some code around from one custom code template to another; however, this may reduce the need for post_create custom code on other resources as well.

However, we can't run decoders for PollAsync resources (because the decoder returns `nil` until the resource exists).

(I think we should split this logic into a separate function & use it for post-create for OpAsync resources as well - they currently rely on "identity" fields - but I'm leaving that for a separate PR.)

I had to mark monitored project as using async for create to prevent it trying to decode the operation that it gets on Create.

Fixed https://github.com/hashicorp/terraform-provider-google/issues/22239

```release-note:none

```


melinath:
This one can't be removed (at least, not now) because the Create method is on the listing and returns the actual object nested inside `subscription`, and decoders aren't available for the create response.

melinath:
Actually, we do use decoders in Create for resources that need to extract data from an operation result, so it should be fine to do so for resources that need to extract from a normal response too.

melinath:
This isn't working quite right - for example, a few Chronicle resources are no longer having their fields set on post_create after this change.

ScottSuarez (approved):

c2thorn (commented):
sorry for the trouble with Analyticshub listing subscription, it's a particularly unusual API that if I were to do again I think I would handwrite. 

Looking through the generated diffs, it looks like the new generated code + decoder does what the old post-create does, but I'd like to test `TestAccBigqueryAnalyticsHubListingSubscription_differentProject` in TC one more time to make sure

c2thorn (approved):
passed

melinath:
> sorry for the trouble with Analyticshub listing subscription, it's a particularly unusual API that if I were to do again I think I would handwrite.

honestly, the resource itself was easy to fix; the problem was figuring out how to not break everything else 😂

PR #13522, opened 2025-04-02
Added support for NAT64 when configuring Router NAT

matheusaleixo-cit:
Adds `sourceSubnetworkIpRangesToNat64` and `nat64Subnetwork` fields to `compute` RouterNat, allowing the support for public Router NAT services with NAT64 enabled.

Also adds `dns64Config` field to `dns` Policy, allowing the configuration of DNS64 server policies, required when using NAT64.

Fixes: [#22116](https://github.com/hashicorp/terraform-provider-google/issues/22116) 

```release-note:enhancement
compute: added `source_subnetwork_ip_ranges_to_nat64` and `nat64_subnetwork` fields  to `google_compute_router_nat` resource
```
```release-note:enhancement
dns: added `dns64_config` field  to `google_dns_policy` resource
```


mmv1/products/dns/Policy.yaml
```diff
@@ -103,6 +103,18 @@ properties:
     update_url: 'projects/{{project}}/policies/{{name}}'
     update_verb: 'PATCH'
     default_value: "Managed by Terraform"
+  - name: 'dns64Config'
+    type: NestedObject
```
rileykarson, line 107:
```suggestion
    type: NestedObject
    default_from_api: true
```

mmv1/third_party/terraform/services/compute/resource_compute_router_nat_test.go.tmpl
```diff
@@ -804,6 +804,40 @@ func testAccCheckComputeRouterNatDelete(t *testing.T, n string) resource.TestChe
...
+func TestAccComputeRouterNat_withNat64Configuration(t *testing.T) {
+	t.Parallel()
+
+	context := map[string]interface{}{
+		"random_suffix": acctest.RandString(t, 10),
+	}
+
+	acctest.VcrTest(t, resource.TestCase{
+		PreCheck:                 func() { acctest.AccTestPreCheck(t) },
+		ProtoV5ProviderFactories: acctest.ProtoV5ProviderBetaFactories(t),
```
rileykarson, line 817:
```suggestion
		ProtoV5ProviderFactories: acctest.ProtoV5ProviderFactories(t),
```

& remove `google-beta` from test configs- the implicit `google` provider will be the beta provider inside these tests. We only use the specific beta factory for example tests so the docs use the beta provider!

mmv1/products/dns/Policy.yaml
```diff
@@ -103,6 +103,19 @@ properties:
...
+    default_from_api: true
+    description: Configurations related to DNS64 for this Policy.
+    properties:
```
rileykarson, line 110: Should these subfields be required or would it be valid to write `dns64_config {}`? The configurations below seem like we could start w/ required.

mmv1/products/compute/RouterNat.yaml
```diff
@@ -258,6 +258,38 @@ properties:
...
+    enum_values:
+      - 'ALL_IPV6_SUBNETWORKS'
+      - 'LIST_OF_IPV6_SUBNETWORKS'
+    min_version: 'beta'
+  - name: 'nat64Subnetwork'
+    type: Array
+    description: |
+      One or more subnetwork NAT configurations whose traffic should be translated by NAT64 Gateway.
+      Only used if `source_subnetwork_ip_ranges_to_nat64` is set to `LIST_OF_IPV6_SUBNETWORKS`
+    api_name: nat64Subnetworks
```
rileykarson, line 278: Note: This is consistent with `subnetwork` above, although we'd generally avoid depluralising nowadays.

mmv1/products/dns/Policy.yaml
matheusaleixo-cit: I did some testing and using `dns64_config {}` causes a permadiff, so I guess we should set it as required.
Setting `dns64_config { scope {} }` seems to be functional when creating a new resource and when updating existing  dns64_config values, so maybe we can keep `allQueries` as optional?

PR #13527, opened 2025-04-03
Provide support for deleting pending caches present on bucket, This w…

googlyrahman:
```release-note: enhancement
storage: Add support for deleting pending caches present on bucket with force_destory=true in storage_bucket resource
```

### DIFF
--- a/mmv1/third_party/terraform/services/storage/resource_storage_bucket.go.tmpl
+++ b/mmv1/third_party/terraform/services/storage/resource_storage_bucket.go.tmpl
@@ -101,7 +101,7 @@ func ResourceStorageBucket() *schema.Resource {
 				Type:        schema.TypeBool,
 				Optional:    true,
 				Default:     false,
-				Description: `When deleting a bucket, this boolean option will delete all contained objects, or anywhereCaches (if any). If you try to delete a bucket that contains objects or anywhereCaches, Terraform will fail that run.`,
+				Description: `When deleting a bucket, this boolean option will delete all contained objects, or anywhereCaches (if any). If you try to delete a bucket that contains objects or anywhereCaches, Terraform will fail that run, deleting anywhereCaches may take 80 minutes to complete.`,
 			},
 
 			"labels": {
@@ -644,7 +644,7 @@ func deleteAnywhereCacheIfAny(config *transport_tpg.Config, bucket string) error
 		if !ok {
 			continue // If state is not a string, skip this item
 		}
-		if state != "running" && state != "paused" {
+		if !strings.EqualFold(state, "running") && !strings.EqualFold(state, "paused")  {
 			continue
 		}
 
@@ -679,7 +679,7 @@ func deleteAnywhereCacheIfAny(config *transport_tpg.Config, bucket string) error
 		return nil
 	}
 
-	return fmt.Errorf("error while deleting the cache!")
+	return fmt.Errorf("Error while deleting the cache: caches still exists post 80mins of their disable time")
 }
 
 func resourceDataplexLabelDiffSuppress(k, old, new string, d *schema.ResourceData) bool {
@@ -1129,6 +1129,13 @@ func resourceStorageBucketDelete(d *schema.ResourceData, meta interface{}) error
 		// is it scheduled to be plumbed to individual providers.
 		wp := workerpool.New(runtime.NumCPU() - 1)
 
+		wp.Submit(func() {
+			err = deleteAnywhereCacheIfAny(config, bucket)
+			if err != nil {
+				deleteObjectError = fmt.Errorf("error deleting the caches on the bucket %s : %w", bucket, err)
+			}
+		})
+
 		for _, object := range res.Items {
 			log.Printf("[DEBUG] Found %s", object.Name)
 			object := object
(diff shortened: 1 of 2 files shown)


mmv1/third_party/terraform/services/storage/resource_storage_bucket.go.tmpl
```diff
@@ -590,6 +590,98 @@ func labelKeyValidator(val interface{}, key string) (warns []string, errs []erro
...
+		if !ok {
+			return fmt.Errorf("unexpected type for resource list item: %T", item)
+		}
+
+		// Check the state of the object
+		state, ok := obj["state"].(string)
+		if !ok {
+			continue // If state is not a string, skip this item
+		}
+		if state != "running" && state != "paused" {
```
SirGitsalot, line 647: Are these strings lower case when returned from the API? The [docs](https://cloud.google.com/storage/docs/json_api/v1/AnywhereCache) say "Can be CREATING, RUNNING, DISABLED, or PAUSED."

In either case, it's probably safer to do something like:

```
if !strings.EqualFold("state", "running") && !strings.EqualFold("state", "paused") {
```

What's the expected behavior if state == "creating"?

mmv1/third_party/terraform/services/storage/resource_storage_bucket.go.tmpl
```diff
@@ -590,6 +590,98 @@ func labelKeyValidator(val interface{}, key string) (warns []string, errs []erro
...
+	if err != nil {
+		return err
+	}
+
+	// Check if the cache list is now empty
+	if len(cacheList) == 0 {
+		return nil
+	}
+
+	return fmt.Errorf("error while deleting the cache!")
```
SirGitsalot, line 682: This could use a more user-friendly error like `"error while deleting the cache: cache is not empty"` (or something)

mmv1/third_party/terraform/services/storage/resource_storage_bucket.go.tmpl
```diff
@@ -1046,6 +1144,12 @@ func resourceStorageBucketDelete(d *schema.ResourceData, meta interface{}) error
 			})
 		}
 
+		// While we wait our objects to be deleted in background, delete the cache if any!
```
SirGitsalot, line 1147: Since this can end up waiting 80 minutes it should run in parallel with the object deletions, and ideally get kicked off first to minimize overall wait time, so switch this to something like:

```
wp.Submit(func() {
		err = deleteAnywhereCacheIfAny(config, bucket)
		if err != nil {
			deleteObjectError = fmt.Errorf("error deleting the caches on the bucket %s : %w", bucket, err)
		}
})
```

And move it above the `for` loop.

mmv1/third_party/terraform/services/storage/resource_storage_bucket.go.tmpl
```diff
@@ -101,7 +101,7 @@ func ResourceStorageBucket() *schema.Resource {
 				Type:        schema.TypeBool,
 				Optional:    true,
 				Default:     false,
-				Description: `When deleting a bucket, this boolean option will delete all contained objects. If you try to delete a bucket that contains objects, Terraform will fail that run.`,
+				Description: `When deleting a bucket, this boolean option will delete all contained objects, or anywhereCaches (if any). If you try to delete a bucket that contains objects or anywhereCaches, Terraform will fail that run.`,
```
SirGitsalot, line 104:
```suggestion
				Description: `When deleting a bucket, this boolean option will delete all contained objects, or anywhereCaches (if any). If you try to delete a bucket that contains objects or anywhereCaches, Terraform will fail that run. Deleting anywhereCaches may take 80 minutes to complete.`,
```

(just to avoid surprising people!)

SirGitsalot (changes requested):

mmv1/third_party/terraform/services/storage/resource_storage_bucket.go.tmpl
googlyrahman: The API does returns them in lower case, However i like the idea of keeping it generic to avoid any issues due to backend changes.

The API would throw an error when trying to disable a cache with "creating" state.

mmv1/third_party/terraform/services/storage/resource_storage_bucket.go.tmpl
googlyrahman: Updated.

mmv1/third_party/terraform/services/storage/resource_storage_bucket.go.tmpl
googlyrahman: Though this also happens in parallel, However it's good to keep it first.

Updated as suggested by you.

mmv1/third_party/terraform/services/storage/resource_storage_bucket.go.tmpl
googlyrahman: Thanks! Updated.

googlyrahman:
Hi @SirGitsalot, Let me know if anything is pending here!

SirGitsalot (approved):
Ran this against the upstream tests in TeamCity to be safe, it passed there too. LGTM

PR #13529, opened 2025-04-03
Pub/Sub - Change Allowed Persistence Regions to Set in 'google_pubsub_topic'

abbrowne126:
Fixes https://github.com/hashicorp/terraform-provider-google/issues/21650

```release-note: bug
pubsub: Fixes https://github.com/hashicorp/terraform-provider-google/issues/21650 by changing `allowed_persistence_regions` on `google_pubsub_topic` to set type.
```


roaks3:
Per https://googlecloudplatform.github.io/magic-modules/breaking-changes/make-a-breaking-change/#in-minor-releases, this seems like it could qualify as a change for a minor release:

> Changing a List to a Set where the field is unordered and the order returned from the API changes unpredictably.

Could you confirm if this issue occurs whenever the field is used, and that the order of elements is not predictable (ie. it can keep changing over time). Effectively, if these things are true, there is virtually no way someone would be able to use the field as a List.

abbrowne126:
> Per https://googlecloudplatform.github.io/magic-modules/breaking-changes/make-a-breaking-change/#in-minor-releases, this seems like it could qualify as a change for a minor release:
> 
> > Changing a List to a Set where the field is unordered and the order returned from the API changes unpredictably.
> 
> Could you confirm if this issue occurs whenever the field is used, and that the order of elements is not predictable (ie. it can keep changing over time). Effectively, if these things are true, there is virtually no way someone would be able to use the field as a List.

I opened a PR as a minor release and was told to put it in the major release https://github.com/GoogleCloudPlatform/magic-modules/pull/13510

I can confirm this - there is a permadiff on this field, and additionally the field should be a set based on its behavior

abbrowne126:
Sounds good, thank you Ryan. Just one question: I am guessing I shouldn't list it as a breaking change [in the release notes](https://screenshot.googleplex.com/BoGq7vdT8T2L3fi) if it goes into a minor release?

abbrowne126:
> @abbrowne126 yea, if we do a minor release you don't need to classify it as a breaking change, and it can go to main.
> 
> To help land on the best solution, can you speak to the extent that this could be in use as an output of the resource? Like are these values often needed for subsequent config? If they are, we may be safer with custom logic that preserves the order supplied by the user (as opposed to a set that would potentially break those users).

This would generally not be used further. It's a retention configuration for messages sent to the topic only, and no such option (regionalization sort of knob) exists elsewhere in Pub/Sub. 

As a team we discussed this and think this is best represented as a set.

roaks3 (changes requested):
Ok, this should be fine to include in a minor release.

Only remaining change would be to base the change against `main` instead of the 6.0.0 branch.

abbrowne126:
Thanks - just changed that as well as changed the type to bugfix.

Really appreciate your initiative on finding the right type for this

mmv1/products/pubsub/Topic.yaml
```diff
@@ -111,6 +111,7 @@ properties:
     properties:
       - !ruby/object:Api::Type::Array
         name: 'allowedPersistenceRegions'
+        is_set: True
```
roaks3, line 114: Nit: could this be changed to `true`, if only to stay consistent?

roaks3 (approved):
No problem! Just one minor nit, but otherwise LGTM

mmv1/products/pubsub/Topic.yaml
abbrowne126: Done

PR #13533, opened 2025-04-03
document sweeper field usage

ScottSuarez:
Documenting sweeper field usage. Still need to document how to run sweepers but i want to simplify the execution pattern for them first. It's kinda weird right now.

```release-note:none

```

### DIFF
--- a/docs/content/reference/resource.md
+++ b/docs/content/reference/resource.md
@@ -325,6 +325,139 @@ Example:
 mutex: 'alloydb/instance/{{name}}'
 ```
 
+## Sweeper customization
+
+Sweepers are a testing infrastructure mechanism that automatically clean up resources created during tests. They run before tests start and can be run manually to clean up dangling resources. Sweepers help prevent test failures due to resource quota limits and reduce cloud infrastructure costs by removing test resources that were not properly cleaned up.
+
+Sweeper generation is enabled by default, except in the following conditions which require customization here:
+- Resources with custom deletion code
+- Resources with parent-child relationships (unless the parent relationship is configured)
+- Resources with complex URL parameters that aren't simple region/project parameters
+
+Simply defining the sweeper block in a resource will override these exclusions and enable sweeper generation for that resource, even with minimal configuration.
+
+### `exclude_sweeper`
+
+If set to `true`, no sweeper will be generated for this resource. This is useful for resources that cannot or should not be automatically cleaned up.
+
+Default: `false`
+
+Example:
+
+```yaml
+exclude_sweeper: true
+```
+
+### `sweeper`
+
+Configures how test resources are swept (cleaned up) after tests. The sweeper system helps ensure resources created during tests are properly removed, even when tests fail unexpectedly. All fields within the `sweeper` block are optional, with reasonable defaults provided when not specified.
+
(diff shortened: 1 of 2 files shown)


docs/content/reference/resource.md
```diff
@@ -325,6 +325,139 @@ Example:
...
+  
+  - `parent_field_regex`: A regex pattern with a capture group to extract a specific portion of the parent field value. This is useful when you need more control over extracting parts of complex resource identifiers. The pattern must contain at least one capture group (in parentheses), and the first capture group's match will be used as the extracted value.
+
+- `query_string`: Allows appending additional query parameters to the resource's delete URL when performing delete operations. Format should include the starting character, e.g., "?force=true" or "&verbose=true". If not specified, no additional query parameters are added.
+
+- `ensure_value`: Specifies a field that must be set to a specific value before deletion. Used for resources that have fields like 'deletionProtectionEnabled' that must be explicitly disabled before the resource can be deleted. All fields within the `ensure_value` block are required except `include_full_resource`:
+  
+  - `field`: The API field name that needs to be updated before deletion. Can include dot notation for nested fields (e.g., "settings.deletionProtectionEnabled").
```
melinath, line 383: style: don't use `e.g.` https://developers.google.com/style/abbreviations#dont-use

here & elsewhere

```suggestion
  - `field`: The API field name that needs to be updated before deletion. Can include dot notation for nested fields (for example, "settings.deletionProtectionEnabled").
```

docs/content/reference/resource.md
```diff
@@ -325,6 +325,139 @@ Example:
...
+
+- `parent`: Configures sweeping for resources that depend on parent resources (like a nodepool that belongs to a cluster).
+
+  Required fields:
+  - `resource_type`: The type of the parent resource (e.g., "google_container_cluster")
+  - `child_field`: The field in your resource that references the parent (e.g., "cluster")
+  - At least one of `parent_field` or `template` is required
+
+  Options for getting parent reference:
+  - `parent_field`: The field from parent to use (typically "name" or "id")
```
melinath, line 371: It would be worth clarifying here & for child_field whether this should be Terraform field name or API field name. Based on the examples it looks like they're likely API field name.

docs/content/reference/resource.md
```diff
@@ -325,6 +325,139 @@ Example:
...
+
+```yaml
+exclude_sweeper: true
+```
+
+### `sweeper`
+
+Configures how test resources are swept (cleaned up) after tests. The sweeper system helps ensure resources created during tests are properly removed, even when tests fail unexpectedly. All fields within the `sweeper` block are optional, with reasonable defaults provided when not specified.
+
+- `identifier_field`: Specifies which field in the resource object should be used to identify resources for deletion. If not specified, defaults to "name" if present in the resource, otherwise falls back to "id".
```
melinath, line 355:
```suggestion
- `identifier_field`: Specifies which field in the API resource object should be used to identify resources for deletion. If not specified, defaults to "name" if present in the resource, otherwise falls back to "id".
```

melinath (approved):
LGTM, a few minor suggestions below

docs/content/reference/resource.md
ScottSuarez: will resolve with parent rework

PR #13537, opened 2025-04-03
Made chronicle retrohunt use id_from_name flattener instead of custom decoder + post_create

melinath:
It looks like this was mistakenly implemented as a decoder + post_create when actually they just wanted id_from_name behavior.

```release-note:none

```


melinath:

trodge (approved):

melinath:
@ankitgoyal0301 caught that this change isn't backwards-compatible; the retrohunt field doesn't get set properly after this change. We didn't see an issue in the tests because the tests for this resource skip VCR. I believe the issue is that this resource uses OpAsync, which uses "identity" fields instead of the Id Format based logic. I'll work on converting OpAsync creates to the new behavior - probably initially in separate PRs.

melinath:
Closing this PR in favor of https://github.com/GoogleCloudPlatform/magic-modules/pull/13693, which makes the necessary async changes. @ankitgoyal0301 added you as a reviewer there

PR #13543, opened 2025-04-04
Add delete branches command

trodge:
fixes b/384994863

```release-note:none

```

### DIFF
--- a/.ci/magician/cmd/delete_branches.go
+++ b/.ci/magician/cmd/delete_branches.go
@@ -1,0 +1,141 @@
+/*
+* Copyright 2024 Google LLC. All Rights Reserved.
+*
+* Licensed under the Apache License, Version 2.0 (the "License");
+* you may not use this file except in compliance with the License.
+* You may obtain a copy of the License at
+*
+*     http://www.apache.org/licenses/LICENSE-2.0
+*
+* Unless required by applicable law or agreed to in writing, software
+* distributed under the License is distributed on an "AS IS" BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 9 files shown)


trodge:

.ci/magician/cmd/delete_branches.go
```diff
@@ -0,0 +1,141 @@
...
+		return "", err
+	}
+
+	controller.SetPath(repo)
+
+	if err := runner.PushDir(repo.Path); err != nil {
+		return "", err
+	}
+
+	message, err := runner.Run("git", []string{
```
rileykarson, line 94: There's no path to using GitHub to do this assocation, is there? Parsing Git seems finicky.

.ci/magician/cmd/delete_branches.go
```diff
@@ -0,0 +1,141 @@
...
+	}, nil)
+	if err != nil {
+		return "", fmt.Errorf("error getting commit message: %s", err)
+	}
+
+	messageParts := strings.Split(message, " ")
+
+	prNumber := messageParts[len(messageParts)-1]
+
+	return strings.Trim(prNumber, "()#\n"), nil
```
rileykarson, line 108: Can we return this as an integer or convert it to one to confirm we didn't match an extra word or token before returning?

rileykarson (commented):
As a general note, non-blocking, here and for other PRs I'd recommend both:

- Describing the intent of the change and a brief 1-2 sentence summary of the choice(s) or considerations you made (for a relatively straightforward case and as much if it takes if it isn't!)
- Linking to an issue, bug, or doc as appropriate to tie in the work to the larger feature or context. (If there isn't one, the summary above is generally fine)

rileykarson (approved):

PR #13549, opened 2025-04-04
feat(LbTrafficExtension): Add LBTrafficExtension metadata support

Yavis:
The GCP LbTrafficExtension supports to give an object of metadata on setup.
This behaviour shall be added with this change.

You can setup an loadbalancer traffic extension with a `traffic.yml` like this:
```yaml
name: queueit-traffic-extension
forwardingRules:
  - https://www.googleapis.com/compute/v1/projects/[PROJECT]/regions/global/forwardingRules/[GCP_FORWARD_RULE]
loadBalancingScheme: EXTERNAL_MANAGED
extensionChains:
  - name: "chain1"
    matchCondition:
      celExpression: 'request.host = 'example.com'
    extensions:
      - name: 'cloud-connector'
        authority: [AUTHORITY]
        service: https://www.googleapis.com/compute/v1/projects/[PROJECT]/regions/global/backendServices/[BACKEND_SERVICE]
        failOpen: true
        timeout: 1s
        supportedEvents:
          - REQUEST_HEADERS
          - REQUEST_BODY
          - RESPONSE_HEADERS
        metadata:
          "customerId": "test"
          "key": "value"
```

Which will be added to metadata in the traffic service extension. The metadata object need to be a JSON object to be valid. Afterwards the command:

```bash
$ gcloud service-extensions lb-traffic-extensions import test-traffic-extension --source=traffic.yaml --location=global
```

will successfully create an extension including the metadata.

```release-note:enhancement
networkservices: added `metadata` field to `google_networkservices_lbtrafficextension` resource
```

### DIFF
--- a/mmv1/products/networkservices/LbTrafficExtension.yaml
+++ b/mmv1/products/networkservices/LbTrafficExtension.yaml
@@ -86,6 +86,9 @@ properties:
   - name: 'labels'
     type: KeyValueLabels
     description: 'Set of labels associated with the LbTrafficExtension resource.'
+  - name: 'metadata'
+    type: Object
+    description: 'Set of metadata associated with the LbTrafficExtension resource.'
   - name: 'forwardingRules'
     type: Array
     description: |
--- a/mmv1/products/networkservices/LbTrafficExtension.yaml
+++ b/mmv1/products/networkservices/LbTrafficExtension.yaml
@@ -86,9 +86,6 @@ properties:
   - name: 'labels'
     type: KeyValueLabels
     description: 'Set of labels associated with the LbTrafficExtension resource.'
-  - name: 'metadata'
-    type: Object
-    description: 'Set of metadata associated with the LbTrafficExtension resource.'
   - name: 'forwardingRules'
     type: Array
     description: |
@@ -105,6 +102,7 @@ properties:
       Match conditions for each extension chain are evaluated in sequence for a given request.
       The first extension chain that has a condition that matches the request is executed.
       Any subsequent extension chains do not execute. Limited to 5 extension chains per resource.
+      Further information can be found at https://cloud.google.com/service-extensions/docs/reference/rest/v1/ExtensionChain
     required: true
     item_type:
       type: NestedObject
@@ -134,6 +132,7 @@ properties:
             A set of extensions to execute for the matching request.
             At least one extension is required. Up to 3 extensions can be defined for each extension chain for
             LbTrafficExtension resource. LbRouteExtension chains are limited to 1 extension per extension chain.
+            Further documentation to be found at https://cloud.google.com/service-extensions/docs/reference/rest/v1/ExtensionChain#Extension
           required: true
           item_type:
             type: NestedObject
@@ -188,6 +187,13 @@ properties:
                 item_type:
                   type: String
                 min_size: 1
+              - name: 'metadata'
+                type: KeyValuePairs
+                description: |
+                  Metadata associated with the extension. This field is used to pass metadata to the extension service.
+                  You can set up key value pairs for metadata as you like and need.
+                  f.e. {"key": "value", "key2": "value2"}.
+
   - name: 'loadBalancingScheme'
     type: Enum
     description: |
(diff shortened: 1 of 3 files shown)


mmv1/products/networkservices/LbTrafficExtension.yaml
```diff
@@ -86,6 +86,9 @@ properties:
   - name: 'labels'
     type: KeyValueLabels
     description: 'Set of labels associated with the LbTrafficExtension resource.'
+  - name: 'metadata'
+    type: Object
```
roaks3, line 90: I think this needs to be `KeyValuePairs`. See https://googlecloudplatform.github.io/magic-modules/develop/add-fields/.

mmv1/third_party/terraform/services/networkservices/resource_network_services_lb_traffic_extension_test.go
```diff
@@ -26,7 +26,7 @@ func TestAccNetworkServicesLbTrafficExtension_update(t *testing.T) {
 				ResourceName:            "google_network_services_lb_traffic_extension.default",
 				ImportState:             true,
 				ImportStateVerify:       true,
-				ImportStateVerifyIgnore: []string{"location", "name", "labels", "terraform_labels"},
+				ImportStateVerifyIgnore: []string{"location", "name", "labels", "metadata", "terraform_labels"},
```
roaks3, line 29: Was there a specific reason for ignoring the new field in these tests?

mmv1/products/networkservices/LbTrafficExtension.yaml
roaks3: Whoops, if this actually needs to be a json blob, you'll need it to be a `String` with some custom fields to include json parsing.

Yavis:
Thanks for the review @roaks3 Actually you're right I'm not completely sure why I added the field onto ignore in the tests. I will fix that and test it as well as checking the API for response tomorrow (I'm from Germany).

Yavis:
Found out, the api supports a non json format if set at the correct place, will have to rewrite a littlebit more, I converted the PR to draft state for the meantime, will keep you updated most likely tomorrow

mmv1/products/networkservices/LbTrafficExtension.yaml
Yavis: Reworked it a littlebit, as metadata can be set as KeyValuePair when set on extension level. So I did.

Yavis:
> This will need to be tested, either by removing the lines that ignore the new field, or running some tests locally.
> 
> There are also likely some other questions to answer: is this field returned from the API, and are there ever defaults that are set server-side?

There are no defaults on server side and as far as I could find out in 
* https://cloud.google.com/service-extensions/docs/configure-traffic-extensions#gcloud
* https://cloud.google.com/service-extensions/docs/reference/rest/v1/ExtensionChain#Extension

mmv1/third_party/terraform/services/networkservices/resource_network_services_lb_traffic_extension_test.go
```diff
@@ -247,6 +247,9 @@ resource "google_network_services_lb_traffic_extension" "default" {
 
       supported_events = ["REQUEST_HEADERS"]
       forward_headers  = ["custom-header"]
+      metadata = {
+        "exampleId": "test"
```
roaks3, line 251: I think here and below you need `=`

roaks3 (commented):
One minor comment, then we can run the tests to verify

mmv1/third_party/terraform/services/networkservices/resource_network_services_lb_traffic_extension_test.go
Yavis: Overlooked that and changed, thanks @roaks3 !

Yavis:
@roaks3 Seems like everything ran well, while I van't get the reports on the tests it at least says they passed. 😆 
Thanks for your support and let me know if there is anything more to do on this on my side. I believe otherwise the only thing left is to get an approve on the PR isn't it?

roaks3 (approved):
Yep, this LGTM

PR #13553, opened 2025-04-04
compute: add certificate_wo write-only field to google_compute_ssl_certificate and google_compute_region_ssl_certificate

BBBmau:
`private_key_wo` already exists on both resources from #16183. This PR adds the matching write-only path for `certificate`.

The `private_key` description is also updated so it is no longer labeled write-only. That text is shared with both `private_key` and the generated `private_key_wo` field.

```release-note:enhancement
compute: added `certificate_wo` write-only field and `certificate_wo_version` field in `google_compute_ssl_certificate`
```
```release-note:enhancement
compute: added `certificate_wo` write-only field and `certificate_wo_version` field in `google_compute_region_ssl_certificate`
```


BBBmau:
tests are unrelated

melinath:
This is skipped in VCR. Manual test runs:

- GA: https://hashicorp.teamcity.com/buildConfiguration/TerraformProviders_GoogleCloud_GOOGLE_MMUPSTREAMTESTS_GOOGLE_PACKAGE_COMPUTE/757620
- Beta: https://hashicorp.teamcity.com/buildConfiguration/TerraformProviders_GoogleCloud_GOOGLE_BETA_MMUPSTREAMTESTS_GOOGLEBETA_PACKAGE_COMPUTE/757621

melinath (changes requested):
looks like this causes a permadiff in the relevant tests.

melinath (commented):
Could you do a manual test run or make these not skipped in VCR? The comment says they use "id.UniqueId" but I don't see that actually being used in the samples; maybe it's outdated?

PR #13559, opened 2025-04-07
Allow setting SSL type on connection profiles

owaismawasy:
Cloud DMS API & UI started to support setting the SSL type on source connection profiles. This change allows setting the SSL type in the Terraform templates.

```release-note:enhancement
databasemigrationservice: added `ssl.type` as an input field to `google_database_migration_service_connection_profile` resource
```


hao-nan-li (commented):
Is it possible to test one of the added enum values here? 
For example type = "required" in any existing test?

PR #13562, opened 2025-04-07
Update BackupPlanAssociation support added

niharika-98:

```release-note:enhancement
"feat: Add support for update BPA".
```


roaks3 (commented):
Couple of comments, but it also looks like there are a number of resources/datasources being promoted from beta to GA in this PR. I would consider splitting those 2 changes, mainly for more isolated rollback should something go wrong.

PR #12965, opened 2025-02-05
Added IAP Enable/Disable flag in cloudrunv2 service

Deepraj1996:
```release-note:enhancement
cloudrunv2: added `iap_enabled` field to `google_cloud_run_v2_service` (beta).
```

PR #13575, opened 2025-04-07
Added IAP Enable/Disable flag in cloudrunv2 service.

bryan0515:

```release-note:enhancement
cloudrunv2: added `iapEnabled` field to `mmv1/products/cloudrunv2/Service.yaml` resource
```

This PR is cloned from https://github.com/GoogleCloudPlatform/magic-modules/pull/12965


mmv1/templates/terraform/examples/cloudrunv2_service_iap.tf.tmpl
```diff
@@ -0,0 +1,16 @@
+resource "google_cloud_run_v2_service" "{{$.PrimaryResourceId}}" {
+  provider = google-beta
+  name     = "{{index $.Vars "cloud_run_service_name"}}"
+  location = "us-central1"
+  deletion_protection = false
+  ingress = "INGRESS_TRAFFIC_ALL"
+  launch_stage = "BETA"
```
SirGitsalot, line 7: The fix is to remove the `launch_stage` line.

I think this is a bug involving the `launch_stage` field: it's marked `default_from_api` but that's not how it works - it's close, but not the same. `default_from_api` means "if the user doesn't provide a value, the API will provide one" but `launch_stage` may *override* a user provided value (as you see in this test).

SirGitsalot (changes requested):

mmv1/templates/terraform/examples/cloudrunv2_service_iap.tf.tmpl
bryan0515: The feature is launch stage protected, so I think simply remove the line also breaks - https://github.com/GoogleCloudPlatform/magic-modules/pull/13575#issuecomment-2784745710

I saw other tests ignore the launch stage field. Would that be the solution to this issue?

SirGitsalot: Hmmm in that case it sounds like a bug in the API. But yeah, ignoring the change will work around it.

mmv1/third_party/terraform/services/cloudrunv2/resource_cloud_run_v2_service_test.go.tmpl
```diff
@@ -1557,3 +1557,79 @@ resource "google_project_iam_member" "logs_writer" {
...
+
+func testAccCloudRunV2Service_cloudrunv2ServiceIapUpdate(context map[string]interface{}) string {
+	return acctest.Nprintf(`
+resource "google_cloud_run_v2_service" "default" {
+  provider = google-beta
+  name     = "tf-test-cloudrun-iap-service%{random_suffix}"
+  location = "us-central1"
+  deletion_protection = false
+  ingress = "INGRESS_TRAFFIC_ALL"
+  launch_stage = "BETA"
```
yanweiguo, line 1624: This will be changed to GA by Cloud Run since iap_enabled is false.

mmv1/templates/terraform/examples/cloudrunv2_service_iap.tf.tmpl
yanweiguo: In the API side, this feature is still beta so in the example we need `launch_stage = "BETA"`. However in the feature when API promotes this feature to GA, it will break the test. Ideally it should be ignored.

mmv1/third_party/terraform/services/cloudrunv2/resource_cloud_run_v2_service_test.go.tmpl
bryan0515: Done.

mmv1/templates/terraform/examples/cloudrunv2_service_iap.tf.tmpl
bryan0515: Done.

SirGitsalot (approved):

PR #13577, opened 2025-04-07
Adding Memorystore import sources 

NA2047:
```release-note:enhancement
memorystore: added `managed_backup_source` and ` gcs_source` fields to `google_memorystore_instance`
```

### DIFF
--- a/mmv1/products/memorystore/Instance.yaml
+++ b/mmv1/products/memorystore/Instance.yaml
@@ -706,3 +706,36 @@ properties:
           description:
             "Output only. Ports of the exposed endpoint."
           output: true
+  - name: 'gcsSource'
+    type: NestedObject
+    description: "GCS source for the instance."
+    immutable: true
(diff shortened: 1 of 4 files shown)


mmv1/products/memorystore/Instance.yaml
```diff
@@ -706,3 +706,40 @@ properties:
...
+  - name: 'gcsSource'
+    type: NestedObject
+    description: "GCS source for the instance."
+    immutable: true
+    ignore_read: true
+    conflicts:
+      - managedBackupSource
+    properties:
+      - name: 'uris'
+        type: Array
```
rileykarson, line 718:
```suggestion
        type: Array
        is_set: true
```

We won't get diffs because these are ignore_read, but this will mean reordering doesn't trigger recreations.

rileykarson:
Sorry- one last thing I remembered 😅

PR #13579, opened 2025-04-08
resource_sql_database_instance.go normalizes failover_dr_replica_name name for diff

lychgoogle:
Fixes https://github.com/hashicorp/terraform-provider-google/issues/21724

* Implementation: Just wrote an DiffSupressFunc that compares normalized values of failover_dr_replica_name instead of raw strings.  For shortened version of failover_dr_replica name (See [doc](https://registry.terraform.io/providers/hashicorp/google-beta/latest/docs/resources/sql_database_instance#failover_dr_replica_name-1)), we just add project name on the left of the instance name to make it normalized.

* Tests in existing `TestAccSqlDatabaseInstance_MysqlSwitchoverSuccess()` and `TestAccSqlDatabaseInstance_PostgresSwitchoverSuccess()`, we already set failover_dr_replica_name twice. We just modify one of setting action to use shortened version of failover_dr_replica_name, so testing setting of failover_dr_replica_name of both standard and shortened format.

```release-note:bug
sql: fixed the issue of shortened version of failover_dr_replica_name causes unnecessary diff 
```

### DIFF
--- a/mmv1/third_party/terraform/services/compute/resource_compute_instance_from_machine_image_test.go.tmpl
+++ b/mmv1/third_party/terraform/services/compute/resource_compute_instance_from_machine_image_test.go.tmpl
@@ -84,6 +84,8 @@ func TestAccComputeInstanceFromMachineImage_maxRunDuration(t *testing.T) {
 }
 
 func TestAccComputeInstanceFromMachineImage_terminationTime(t *testing.T) {
+	// Uses time.Now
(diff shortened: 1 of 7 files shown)


lychgoogle:
For reviewers. http://b/406065779#comment9 has some internal discussion.

SirGitsalot (approved):

lychgoogle:
The fix is merged. Let's wait for the release.

PR #13581, opened 2025-04-08
Add VPC network support to Security Gateway terraform

Berro321:
```release-note:enhancement
beyondcorp: Added `upstreams` fields to `google_beyondcorp_application` resource
```


mmv1/products/beyondcorp/Application.yaml
```diff
@@ -109,6 +115,33 @@ properties:
...
+            - name: regions
+              type: Array
+              description: Required. List of regions where the application sends traffic to.
+              required: true
+              item_type:
+                type: String
+        - name: network
+          type: NestedObject
+          exactly_one_of:
+            - network
```
slevenick, line 137: exactly_one_of doesn't work inside arrays. You can remove this

mmv1/products/beyondcorp/Application.yaml
```diff
@@ -109,6 +115,33 @@ properties:
...
+          description: Network to forward traffic to.
+          properties:
+            - name: name
```
slevenick, line 140: This should probably be required unless there's a reason to specify a network without a name.

slevenick (approved):

PR #13583, opened 2025-04-08
Deprecate `run_as_service_account` field in client

rohitjangid:

Deprecate `run_as_service_account` field in `google_integrations_client`. Add iam binding for integration p4sa.

```release-note:deprecation
integrations: deprecated `run_as_service_account` field in `google_integrations_client` resource
```

Fixes https://github.com/hashicorp/terraform-provider-google/issues/21887

### DIFF
--- a/mmv1/products/integrations/Client.yaml
+++ b/mmv1/products/integrations/Client.yaml
@@ -46,12 +46,19 @@ examples:
     vars:
       key_ring_name: 'my-keyring'
       crypto_key_name: 'my-crypto-key'
-      service_account_id: 'service-acc'
     test_vars_overrides:
       key_ring_name: '"tftest-shared-keyring-1"'
       crypto_key_name: '"tftest-shared-key-1"'
       kms_key: 'acctest.BootstrapKMSKeyInLocation(t, "us-east1")'
+    bootstrap_iam:
+      - member: "serviceAccount:service-{project_number}@gcp-sa-integrations.iam.gserviceaccount.com"
+        role: "roles/cloudkmskacls.serviceAgent"
     skip_vcr: true
+  - name: 'integrations_client_service_account'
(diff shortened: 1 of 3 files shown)


mmv1/products/integrations/Client.yaml
```diff
@@ -113,3 +120,4 @@ properties:
       User input run-as service account, if empty, will bring up a new default service account.
     immutable: true
     ignore_read: true
+    deprecation_message: '`run_as_service_account` is deprecated and will be removed in a future major release.'
```
c2thorn, line 123: Is there an alternative users should use instead?

c2thorn (changes requested):
can you also create a new github issue to remove the field in a future major release?

c2thorn:
reassigning to @NickElliot to follow up as I'll be out for a few days.

mmv1/products/integrations/Client.yaml
rohitjangid: This field is a no-op and is not used anymore by the service. There are no alternative to this field.

rohitjangid:
> can you also create a new github issue to remove the field in a future major release?

Created https://github.com/hashicorp/terraform-provider-google/issues/22288

rohitjangid:
> reassigning to @NickElliot to follow up as I'll be out for a few days.

Hi @NickElliot 
Can you take a look at the PR and review?

NickElliot (commented):
LGTM but could you clarify the timeline on the field becoming non operational?

rohitjangid:
> LGTM but could you clarify the timeline on the field becoming non operational?

The field is already non-operational

NickElliot (approved):
got it, given the test still passes I think its alright to include for coverage's sake!

PR #13595, opened 2025-04-08
Add Operation to healthcare pipeline job creation and update

zli82016:
fixes https://github.com/hashicorp/terraform-provider-google/issues/20943
part of https://github.com/hashicorp/terraform-provider-google/issues/21294

Add Operation to healthcare pipeline job creation and update, as the response of CREATE and PATCH is an operation.

```release-note:enhancement
healthcare: made `google_healthcare_pipeline_job` wait for creation and update operation to complete
```

### DIFF
--- a/mmv1/products/healthcare/PipelineJob.yaml
+++ b/mmv1/products/healthcare/PipelineJob.yaml
@@ -29,6 +29,14 @@ update_verb: PATCH
 update_mask: true
 id_format: '{{dataset}}/pipelineJobs/{{name}}'
 import_format: ['{{%dataset}}/pipelineJobs/{{name}}', '{{name}}', '{{dataset}}/pipelineJobs?pipelineJobId={{name}}']
+autogen_async: true
+async:
+  actions: ['create', 'update']
+  type: 'OpAsync'
+  operation:
+    base_url: '{{op_id}}'
+  result:
+    resource_inside_response: true
 examples:
   - name: 'healthcare_pipeline_job_reconciliation'
     primary_resource_id: 'example-pipeline'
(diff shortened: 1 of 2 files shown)


zli82016:
The failure of test TestAccHealthcarePipelineJob_healthcarePipelineJobBackfillExample is unrelated.

mmv1/products/healthcare/PipelineJob.yaml
```diff
@@ -29,6 +29,14 @@ update_verb: PATCH
 update_mask: true
 id_format: '{{dataset}}/pipelineJobs/{{name}}'
 import_format: ['{{%dataset}}/pipelineJobs/{{name}}', '{{name}}', '{{dataset}}/pipelineJobs?pipelineJobId={{name}}']
+autogen_async: true
+async:
+  actions: ['create', 'update']
```
zli82016, line 34: Only the response of CREATE and PATCH APIs is an operation.

slevenick:
While you're here can you fix the other failing test? It looks like it just needs a new name

zli82016:
Oh, never mind, I think I know the reason. Let me give it one more try.

slevenick (approved):
If you can fix the issue with the failing test that would be great, otherwise it's already failing

mmv1/templates/terraform/examples/healthcare_pipeline_job_backfill.tf.tmpl
```diff
@@ -3,11 +3,72 @@ resource "google_healthcare_pipeline_job" "{{$.PrimaryResourceId}}" {
   location = "us-central1"
   dataset = google_healthcare_dataset.dataset.id
   backfill_pipeline_job {
-    mapping_pipeline_job = "${google_healthcare_dataset.dataset.id}/pipelinejobs/{{index $.Vars "mapping_pipeline_name"}}"
```
zli82016, line 6: `pipelinejobs` should be `pipelineJobs`. 

I use id here to replace the whole string.

After this fix, a new error comes up. I will stop here.

zli82016: Oh, it is better to just change it to `pipelineJobs`

PR #13619, opened 2025-04-10
Factored setting of post-create id format fields into a separate function

melinath:
This PR doesn't change any functionality - just moves it into a separate function. This is a follow-up to https://github.com/GoogleCloudPlatform/magic-modules/pull/13520 and prepares for changes that will be required for https://github.com/GoogleCloudPlatform/magic-modules/pull/13537 (which uses OpAsync - we will need to first convert OpAsync cases to use this function as well).

I've factored this change into a standalone PR so that it'll be easier to review.

```release-note:none

```


melinath:
VCR failures look unrelated.

ScottSuarez (approved):

PR #13622, opened 2025-04-11
Fix perma-diff in google_identity_platform_config for default false values

pashanka:
Fix [#20406](https://github.com/hashicorp/terraform-provider-google/issues/20406)

Customer reported that google_identity_platform_config always shows a diff if the values are set to false for blockingFunctions.forwardInboundCredentials. Tried reproducing the issue for all the properties in the google_identity_platform_config resource and the plan always shows a diff, even though the value set is the same as the actual value of the config.

```

      ~ blocking_functions {
          + forward_inbound_credentials {
              + access_token  = false
              + id_token      = false
              + refresh_token = false
            }

            # (1 unchanged block hidden)
        }

      + multi_tenant {
          + allow_tenants = false
        }

        # (5 unchanged blocks hidden)
``` 
Above is the diff that is always present if the value set to these properties are false.

This PR fixes this perma-diff by implementing custom_flatten to prevent this from happening. Also added test to cover all such parameters to make sure that Config resource is covered.

Similar fix in the past - https://github.com/GoogleCloudPlatform/magic-modules/pull/12255

```release-note:bug
identityplatform: Fixed perma-diff in `google_identity_platform_config`
```

### DIFF
--- a/mmv1/products/identityplatform/Config.yaml
+++ b/mmv1/products/identityplatform/Config.yaml
@@ -208,7 +208,7 @@ properties:
         type: NestedObject
         description: |
           The user credentials to include in the JWT payload that is sent to the registered Blocking Functions.
-        custom_flatten: 'templates/terraform/custom_flatten/identity_platform_config_forward_inbound_credentials.go.tmpl'  
+        custom_flatten: 'templates/terraform/custom_flatten/identity_platform_config_forward_inbound_credentials.go.tmpl'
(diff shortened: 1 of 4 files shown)


hao-nan-li (commented):
Could you explain a bit what does this fix do?

pashanka:
> Could you explain a bit what does this fix do?

Added to the PR description.

PR #13626, opened 2025-04-11
Added IPv6 support for Internal Ranges in Terraform.

dorianverna:
This PR aims to modify the Internal Range Terraform support as follows:

Add new comments to Internal Range Terraform documentation to reflect IPv6 support.
Add a template file showing how the feature is used.
Added acceptance test for the newly added feature.

```release-note:enhancement
networkconnectivity: added IPv6 support to `google_network_connectivity_internal_range` resource
```

+CC: @sebkalis


SirGitsalot (changes requested):
LGTM, but it looks like this PR crossed paths with #13586 so that needs to be resolved first

SirGitsalot (approved):

PR #13634, opened 2025-04-11
Removed logging metric set_computed_name post_create

melinath:
This is unused code, since name is a required field. Part of https://github.com/hashicorp/terraform-provider-google/issues/22214

```release-note:none

```


rileykarson (commented):
Can you break down where/how this is duplicated? I don't see replacement logic in the diff.

It seems more likely that this was not needed at all, at first glance. The metric name seems to be user-provided and not server generated.

melinath:
It looks like you're correct and this is just unnecessary code, not unused code. I've updated the PR description for clarity.

melinath:
Merged in main to resolve the unrelated MM unit test failure.

PR #13639, opened 2025-04-11
Removed osconfig guestpolicies set_computed_name post_create

melinath:
As far as I can tell this code is just completely unnecessary. The `name` field has the same format as self_link (which gets used to determine the id format). For example, from a recent [nightly test](https://hashicorp.teamcity.com/buildConfiguration/TerraformProviders_GoogleCloud_GOOGLE_BETA_NIGHTLYTESTS_GOOGLEBETA_PACKAGE_OSCONFIG/353069?buildTab=artifacts) you can see that the name is in the format `Finished creating GuestPolicies "projects/653407317329/guestPolicies/tf-test-guest-policy6shqkt7xf4"`. Additionally, `guest_policy_id` is a required field and [always has been](https://github.com/GoogleCloudPlatform/magic-modules/commit/34fa2676afcd68b5adfc34e2a4299000dc8fca7d)

That said, if there's a concern, I could set id_format to `{{name}}` (which it used to be set to) to reduce the difference between the old / new code.

```release-note:none

```

### DIFF
--- a/mmv1/products/osconfig/GuestPolicies.yaml
+++ b/mmv1/products/osconfig/GuestPolicies.yaml
@@ -38,8 +38,6 @@ timeouts:
   delete_minutes: 20
 identity:
   - guestPolicyId
-custom_code:
-  post_create: 'templates/terraform/post_create/set_computed_name.tmpl'
 examples:
   - name: 'os_config_guest_policies_basic'
     primary_resource_id: 'guest_policies'


melinath:
@modular-magician request-reviewer rileykarson

melinath:
@modular-magician reassign-reviewer @rileykarson

melinath:
This resource actually seems to be unique among set_computed_name resources in that it doesn't have an id_format of `{{name}}` currently.

rileykarson (approved):
Hmm, yeah, seems to do nothing.

PR #13651, opened 2025-04-11
Made firewall policy with rules ignore decoding operations on create

melinath:
Technically, using standard operation handling would be better - but it turns out ~Compute has a special operation signature that's not compatible with generation.~ FirewallPolicyWithRules doesn't have a project in its signature - but the operation code for Compute expects a project to be present.

```release-note:none

```


melinath:
Manual test run just to be safe: https://hashicorp.teamcity.com/buildConfiguration/TerraformProviders_GoogleCloud_GOOGLE_BETA_MMUPSTREAMTESTS_GOOGLEBETA_PACKAGE_COMPUTE/353413

zli82016 (approved):
LGTM supposed that the test will pass.

PR #13656, opened 2025-04-12
container: allow updating accelerators

DrFaust92:

```release-note:enhancement
container: allow update in place for `guest_accelerator`
```

### DIFF
--- a/mmv1/third_party/terraform/services/container/node_config.go.tmpl
+++ b/mmv1/third_party/terraform/services/container/node_config.go.tmpl
@@ -149,20 +149,17 @@ func schemaNodeConfig() *schema.Schema {
 					Type:     schema.TypeList,
 					Optional: true,
 					Computed: true,
-					ForceNew: true,
 					Description: `List of the type and count of accelerator cards attached to the instance.`,
 					Elem: &schema.Resource{
 						Schema: map[string]*schema.Schema{
 							"count": {
 								Type:     schema.TypeInt,
 								Required: true,
-								ForceNew: true,
 								Description: `The number of the accelerator cards exposed to an instance.`,
 							},
 							"type": {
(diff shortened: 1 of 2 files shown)


mmv1/third_party/terraform/services/container/node_config.go.tmpl
```diff
@@ -148,21 +148,18 @@ func schemaNodeConfig() *schema.Schema {
 				"guest_accelerator": {
 					Type:     schema.TypeList,
 					Optional: true,
-					Computed: true,
```
DrFaust92, line 151: this will recreate when guest_accelerator is removed as there no method to remove via API (I tried thoroughly) but will allow updating if changes to nested args are made

NickElliot (commented):
removing computed status from the entire block is definitely a breaking change as it no longer is providing output, is it possible to keep `guest_accelerator` as O+C?

DrFaust92:
[NickElliot](https://github.com/NickElliot) no, as it suppresses diff

NickElliot (changes requested):
even if users are unable to destroy by removing a `guest_accelerator{}` block, that is fine as that is current resource behavior, even if it would force users to manually recreate rather than having it come through as part of the plan.

that attribute of `guest_accelerator` could potentially change in the next major release, but it can't happen with the rest of this in-place update support.

DrFaust92:
> even if users are unable to destroy by removing a `guest_accelerator{}` block, that is fine as that is current resource behavior, even if it would force users to manually recreate rather than having it come through as part of the plan.
> 
> that attribute of `guest_accelerator` could potentially change in the next major release, but it can't happen with the rest of this in-place update support.

Hi @NickElliot not sure what you are suggesting? leave bugged state as is and ignore removal? just support update?

PR #13658, opened 2025-04-13
container: allow updating windows_node_config in place

DrFaust92:

```release-note:enhancement
container: allow updaing `windows_node_config` in place
```

tests and update logix exists, but the forcenew is extraneous. tested locally with flag and before to validate behavour change and that node pool is not recreated. other changes is just fmt

### DIFF
--- a/mmv1/third_party/terraform/services/container/node_config.go.tmpl
+++ b/mmv1/third_party/terraform/services/container/node_config.go.tmpl
@@ -630,41 +630,41 @@ func schemaNodeConfig() *schema.Schema {
 								Description: `Controls the maximum number of processes allowed to run in a pod.`,
 							},
 							"container_log_max_size": {
-                                                                Type:        schema.TypeString,
-                                                                Optional:    true,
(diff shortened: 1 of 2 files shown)


mmv1/third_party/terraform/services/container/node_config.go.tmpl
```diff
@@ -718,7 +718,6 @@ func schemaNodeConfig() *schema.Schema {
 							"osversion": {
 								Type:         schema.TypeString,
 								Optional:     true,
-								ForceNew:     true,
```
DrFaust92, line 721: actual change, everything else is fmt

roaks3 (commented):
Do you think we might be able to add a plan check to `TestAccContainerNodePool_withWindowsNodeConfig`, as specified [here](https://googlecloudplatform.github.io/magic-modules/test/test/#add-an-update-test)? That would be nice to confirm the update in place, since the test was passing before.

Otherwise the change LGTM

DrFaust92:
[roaks3](https://github.com/roaks3) neat, I didnt know of this check. added and passing locally

PR #13665, opened 2025-04-14
location attribute incorrectly marked as optional for cross-cloud interconnect

abhijeetkjha-google:
```release-note:enhancement
compute: marked `location` field as required in `google_compute_interconnect` resource
```

On the resource 'google_compute_interconnect', the attribute 'location' is incorrectly marked as optional, even for cross-cloud interconnect. It needs to be required.

https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/compute_interconnect
location - (Optional) URL of the InterconnectLocation object that represents where this connection is to be provisioned. Specifies the location inside Google's Networks, should not be passed in case of cross-cloud interconnect.

It should be marked as a required field, the customer has to provide a location for all interconnects regardless of whether it's cross cloud or not.

### DIFF
--- a/mmv1/products/compute/Interconnect.yaml
+++ b/mmv1/products/compute/Interconnect.yaml
@@ -81,10 +81,11 @@ properties:
     type: ResourceRef
     description: |
       URL of the InterconnectLocation object that represents where this connection is to be provisioned.
-      Specifies the location inside Google's Networks, should not be passed in case of cross-cloud interconnect.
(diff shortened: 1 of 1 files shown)


abhijeetkjha-google:
Hi @BBBmau , Could you please take a look at this PR when you have a moment? Thanks!

BBBmau (approved):
thanks for catching this! Confirming the issue from testing this the example configuration but without the field set

image

PR #13666, opened 2025-04-14
Add StateReason to NCC Spoke

amanMahendroo:

```release-note:enhancement
networkconnectivity: added `state_reason` field to `spoke` resource
```


shuyama1:
rerun missing doc detector
/gcbrun

ScottSuarez:
Could you add a test for this field or add it to an existing test? If not, why not?

amanMahendroo:
@ScottSuarez This is an output only field which is populated by a different API that isn't supported on Terraform and as such can't be added in a create or update test. The `StateReason` array would otherwise remain empty unless we use this API. I can't think of a reasonable test for this field since its only purpose is to be read by user in a Get request (other than asserting that it's an empty array upon resource creation, but I don't know of a way to do that in a test).

mmv1/products/networkconnectivity/Spoke.yaml
```diff
@@ -329,3 +329,9 @@ properties:
       - 'CREATING'
       - 'ACTIVE'
       - 'DELETING'
+  - name: 'stateReason'
```
ScottSuarez, line 332: The public API documentation suggestions that stateReason does not exist.

[Perhaps reason is what you are looking for?](https://cloud.google.com/network-connectivity/docs/reference/networkconnectivity/rest/v1beta/projects.locations.spokes#:~:text=reasons%5B%5D,the%20INACTIVE%20state.)

ScottSuarez (changes requested):

ScottSuarez:
> @ScottSuarez This is an output only field which is populated by a different API that isn't supported on Terraform and as such can't be added in a create or update test. The `StateReason` array would otherwise remain empty unless we use this API. I can't think of a reasonable test for this field since its only purpose is to be read by user in a Get request (other than asserting that it's an empty array upon resource creation, but I don't know of a way to do that in a test).

I see, I think I am okay with this, but have issue with how we are using vs what is documented publicly on the docs. Given that we have no tests I would still like some form of validation that this field works. Local testing or instructions on how to verify functionality.

amanMahendroo:
PTAL at https://paste.googleplex.com/4819292224487424 for instructions on how to verify functionality. I'm unable to test my changes due to unrelated errors in other modules, specifically:

```
too many arguments in call to sweeper.AddTestSweepers
	have (string, func(_ string) error)
	want (*sweeper.Sweeper)
```

(if there is a known fix to this, it would be very helpful!)

amanMahendroo:
There seems to be a failing VCR test, but clicking it shows that the test has passed. Not sure how to fix this

ScottSuarez:
The test crashed with a panic because the API is returning a map[string]string for the new Reasons field.
```
panic: reasons.0: '' expected type 'string', got unconvertible type 'map[string]interface {}', value: 'map[code:PENDING_REVIEW message:Spoke is Pending Review]'

goroutine 34637 [running]:
```

For your local development, make sure both the downstream (terraform-provider-google) and the upstream (magic-modules) are both up to date before generating (git pull upstream main && git clean -fdx .)

`git clean -fdx .` will throw away uncommitted changes (just an fyi)

amanMahendroo:
Updated the field. When testing locally using the config provided above, the output is returned as an empty array as expected.

ScottSuarez (changes requested):
I provisioned the config in the comment above and was unable to see the reasons field populated. I need to a way to verify functionality be approving. 

As seen above, an invalid schema can cause the provider to crash and break users if incorrect.

amanMahendroo:
For that, I would have to call Get on the Spoke resource after updating it. However, data blocks for NCC Spoke resource don't seem to be supported yet. Is the VCR test not sufficient to verify schema correctness?

ScottSuarez (commented):
with current VCR implementation I cannot see the POST and GET requests.. I would feel more comfortable if there was a way to to directly look locally. If either of us can provision a machine that has these fields populated that would be perfect !

amanMahendroo:
Hi @ScottSuarez . I have updated the above paste (https://paste.googleplex.com/4819292224487424) to reflect the changes necessary to see the updated response. I was able to see the "reasons" field updated as well. All screenshots and steps are mentioned in the paste itself

ScottSuarez (approved):
Perfect ! Thanks for helping me be thorough about testing. I hope you can understand my position. Really appreciate it!!

LGTM

PR #13672, opened 2025-04-14
pubsub_lite_reservation demo fwprovider conversion

NickElliot:

```release-note:none

```


mmv1/third_party/terraform/services/pubsublite/fw_resource_pubsub_lite_reservation.go
```diff
@@ -0,0 +1,383 @@
...
+	}, &resp.Diagnostics)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	tflog.Trace(ctx, "create fwprovider google_pubsub_lite resource")
+
+	// Put data in model
+	data.Id = types.StringValue(fmt.Sprintf("projects/%s/locations/%s/reservations/%s", data.Project.ValueString(), data.Region.ValueString(), data.Name.ValueString()))
+	data.ThroughputCapacity = types.Int64Value(res["throughputCapacity"].(int64))
```
c2thorn, line 167: This is the only field that really needs to be read, so we don't need to call Read or any shared function right?

mmv1/third_party/terraform/services/pubsublite/fw_resource_pubsub_lite_reservation_test.go
```diff
@@ -0,0 +1,39 @@
...
+
+	context := map[string]interface{}{
+		"random_suffix": acctest.RandString(t, 10),
+	}
+	acctest.VcrTest(t, resource.TestCase{
+		PreCheck:                 func() { acctest.AccTestPreCheck(t) },
+		ProtoV5ProviderFactories: acctest.ProtoV5ProviderFactories(t),
+		Steps: []resource.TestStep{
+			{
+				Config: testAccResourceFWPubsubLiteReservation_basic(context),
```
c2thorn, line 21: update step?

mmv1/third_party/terraform/services/pubsublite/fw_resource_pubsub_lite_reservation.go
NickElliot: yeah, I could add the read call for if we want it because of it being what the template will do, but its unneeded for this

c2thorn (approved):
i'm not seeing `TestAccResourceFWPubsubLiteReservation_basic` in the passing list of tests, but it didn't show in the failing ones?

NickElliot:
if you go into the debug log for all tests it appears to have passed in recording? I'm not sure why it isnt there on this end

PR #13677, opened 2025-04-15
Fix for 

NA2047:
Fix for Issue   [Issue 21775]( https://github.com/hashicorp/terraform-provider-google/issues/21775?reload=1)

```release-note:bug
google_compute_network_firewall_policy_rule:  Adding diffsuppress  for Issue 21775
```

### DIFF
--- a/mmv1/templates/terraform/constants/network_firewall_policy_rule.go.tmpl
+++ b/mmv1/templates/terraform/constants/network_firewall_policy_rule.go.tmpl
@@ -1,0 +1,15 @@
+func validateSecurityProfileGroupPrefix(v interface{}, k string) (warnings []string, errors []error) {
+	value := v.(string)
+	lenValue := len(value)
+	if lenValue == 0 {
+		errors = append(errors, fmt.Errorf("the empty string is not a valid security_profile_group value"))
+		return nil, errors
+	} else if lenValue >= 2 {
+		secondString := value[1]
+		if secondString != '/' {
+			errors = append(errors, fmt.Errorf("%q (%q) security_profile_group value is required to start with two forward slashes, not just one ", k, value))
+			return nil, errors
+		}
+	}
+	return nil, nil
+}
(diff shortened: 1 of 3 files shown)


mmv1/third_party/terraform/services/compute/resource_compute_network_firewall_policy_rule_test.go.tmpl
```diff
@@ -491,7 +520,7 @@ resource "google_compute_network_firewall_policy_rule" "fw_policy_rule1" {
   priority               = 9000
   enable_logging         = true
   action                 = "apply_security_profile_group"
-  security_profile_group = "//networksecurity.googleapis.com/${google_network_security_security_profile_group.security_profile_group_updated.id}"
+  security_profile_group = "${security_profile_group_prefix}networksecurity.googleapis.com/${google_network_security_security_profile_group.security_profile_group_updated.id}"
```
SirGitsalot, line 523:
```suggestion
  security_profile_group = "%{security_profile_group_prefix}networksecurity.googleapis.com/${google_network_security_security_profile_group.security_profile_group_updated.id}"
```

mmv1/templates/terraform/constants/network_firewall_policy_rule.go.tmpl
```diff
@@ -0,0 +1,8 @@
+// Match a string that starts with one / slash followed .* t then suppress it as API takes // or / as prefixes
+func SecurityProfileGroupPrefixSlashes(_, old, new string, _ *schema.ResourceData) bool {
+	re := regexp.MustCompile(`^/[^/].*`)
+	if re.MatchString(new) {
+		return true
```
SirGitsalot, line 5: From the bug report and the comment it looks like the issue is that the API accepts values that start with a single or double slash, but will always return a double slash. If that's the case, I think the logic here is wrong, as it will always suppress the diff if the new value begins with a single slash (which prevents changing the value from one single-leading slash value to a different single slash value).

A common approach in diffSuppressFuncs is to normalize the old and values and then compare; in this case you could make them both single slash values and compare, something like:

```
if strings.HasPrefix(old, "//") {
    old = old[1:]
}
if strings.HasPrefix(new, "//") {
    new = new[1:]
}
return old == new
```

SirGitsalot (changes requested):

SirGitsalot (approved):

PR #13687, opened 2025-04-15
Fix issue labeler backfill

shuyama1:
Currently the `Issue Labeler Backfill` GHA is failing with error `getting github issues: listing issues: GET https://api.github.com/repos/hashicorp/terraform-provider-google/issues?direction=desc&page=100&per_page=100&since=1973-01-01T00%3A00%3A00Z&sort=updated&state=all: 422 Pagination with the page parameter is not supported for large datasets, please use cursor based pagination (after/before) []`

```release-note:none

```

### DIFF
--- a/tools/issue-labeler/labeler/backfill.go
+++ b/tools/issue-labeler/labeler/backfill.go
@@ -3,6 +3,8 @@ package labeler
 import (
 	"context"
 	"fmt"
+	"net/http"
+	"regexp"
 	"sort"
 	"time"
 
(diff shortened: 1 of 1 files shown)


roaks3 (approved):
This looks good to me, but did you already have a look at cl/747885807? The main difference I noticed was special handling for the parameters being ordered differently (I'm not sure if that's possible).

shuyama1:
> This looks good to me, but did you already have a look at cl/747885807? The main difference I noticed was special handling for the parameters being ordered differently (I'm not sure if that's possible).

Oh, I didn't notice cl/747885807 when making this change. It looks like that CL also switched to using the link header to fix the same issue. I'll update my change to follow what was done there for consistency.

roaks3 (approved):

PR #13693, opened 2025-04-16
Switched async post-create setting of fields to check id format instead of 'identity'

melinath:
This brings it more in line with the PostCreateSetComputedFields method (the template lines are copied from there, modified to use opRes instead) but shows the diffs inline in Create for easier comparison. Part of https://github.com/hashicorp/terraform-provider-google/issues/22214.

Special notes per resource:

- Chronicle retrohunt: Brought over changes from https://github.com/GoogleCloudPlatform/magic-modules/pull/13537 which are necessary to ensure that `name` still gets set as required.
- Firebase AndroidApp: `name` doesn't look necessary to me
- Firebase AppleApp: `name` doesn't look necessary to me
- Firebase WebApp: `name` doesn't look necessary to me
- Spanner Instance: default_from_api field only needs to be set if unset
- Spanner InstanceConfig: default_from_api field only needs to be set if unset
- Workflows Workflow: default_from_api field only needs to be set if unset

```release-note:none

```


melinath:
Manual test runs for Chronicle Retrohunt, which skips VCR: https://hashicorp.teamcity.com/buildConfiguration/TerraformProviders_GoogleCloud_GOOGLE_BETA_MMUPSTREAMTESTS_GOOGLEBETA_PACKAGE_CHRONICLE/356171

Same for Spanner Instance (which also includes Spanner InstanceConfig even though that will run in VCR too): https://hashicorp.teamcity.com/buildConfiguration/TerraformProviders_GoogleCloud_GOOGLE_BETA_MMUPSTREAMTESTS_GOOGLEBETA_PACKAGE_SPANNER/356172

melinath:
Spanner tests passed. Chronicle Retrohunt failed - made some changes that I think should fix it. Unfortunately I can't test this resource locally.

melinath:
Chronicle retrohunt manual test run: https://hashicorp.teamcity.com/buildConfiguration/TerraformProviders_GoogleCloud_GOOGLE_BETA_MMUPSTREAMTESTS_GOOGLEBETA_PACKAGE_CHRONICLE/356175

melinath:
Manual test run: https://hashicorp.teamcity.com/buildConfiguration/TerraformProviders_GoogleCloud_GOOGLE_BETA_MMUPSTREAMTESTS_GOOGLEBETA_PACKAGE_CHRONICLE/356177

melinath:
okay, yeah, this isn't fixable because the resource data isn't in the operation response - it's in operation["metadata"]. The previous logic is actually parsing the _operation_ name rather than the retrohunt name (I guess they're always identical, at least right now?) We could fix this by allowing async resources to customize where in the operation they look for the response (or the API could start returning the data in response.) I'll open a ticket for potentially allowing that configuration knob, but it would need at least a few more resources to require it before we'd implement it.

melinath:
Opened https://github.com/hashicorp/terraform-provider-google/issues/22392 to track that issue.

melinath:
correction: The previous decoder was trying to pull the resource name from the resource data in the response (which isn't set); the previous post_create was properly extracting it from the metadata field.

melinath:
Note: this fixes a bug in Chronicle retrohunt where the ID was never being set correctly after the initial Create - only after the subsequent Read.

melinath:
Okay, that got Retrohunt passing 🎉

c2thorn:
> Note: this fixes a bug in Chronicle retrohunt where the ID was never being set correctly after the initial Create - only after the subsequent Read.

this was never practically user-facing since Create calls Read right?

melinath:
> > Note: this fixes a bug in Chronicle retrohunt where the ID was never being set correctly after the initial Create - only after the subsequent Read.
> 
> this was never practically user-facing since Create calls Read right?

Yes, this was probably never user-facing.

c2thorn (approved):

ankitgoyal0301 (approved):
LGTM!

The above problem of name and retrohunt ID not being set seems to be fixed now.

https://paste.googleplex.com/6241351785775104
https://paste.googleplex.com/6728694040887296

Thanks @melinath for taking care of this.

PR #13695, opened 2025-04-16
Removed unnecessary check for name in spanner database post_create

melinath:
spanner database. post_create relied specifically on accessing opRes, but name is actually a settable & required field on that resource so it's not necessary to pull it from the operation as far as I can tell.

This code will stop working as a consequence of changes for https://github.com/hashicorp/terraform-provider-google/issues/22214, because this resource will no longer be detected as requiring opRes to be set (because there are no computed fields that need to be set after create.)

This part of the post_create template was added in [this commit](https://github.com/GoogleCloudPlatform/magic-modules/pull/6141/commits/3b78b94cf46073a2491a6b30add63b7cc35e9880), after [this round of review](https://github.com/GoogleCloudPlatform/magic-modules/pull/6141#pullrequestreview-1014203786). I don't see anything indicating that this was added in response to a specific problem, and `name` should always be available anyway (because it's a required field) so I think this may have been a mistake.

```release-note:none

```

### DIFF
--- a/mmv1/templates/terraform/post_create/spanner_database.go.tmpl
+++ b/mmv1/templates/terraform/post_create/spanner_database.go.tmpl
@@ -3,12 +3,6 @@
 // `terraform apply` twice to get their desired outcome, the provider does not set
 // `extraStatements` in the call to the `create` endpoint and all DDL (other than
 //  <CREATE DATABASE>) is run post-create, by calling the `updateDdl` endpoint 
-
-_, ok := opRes["name"]
-if !ok {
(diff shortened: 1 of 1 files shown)


melinath:
@modular-magician reassign-reviewer @c2thorn

melinath:
Manual test run since this skips VCR: https://hashicorp.teamcity.com/buildConfiguration/TerraformProviders_GoogleCloud_GOOGLE_BETA_MMUPSTREAMTESTS_GOOGLEBETA_PACKAGE_SPANNER/356173

c2thorn (approved):

PR #13696, opened 2025-04-16
Add field to support configuring new Media CDN feature (flexible shielding) via Terraform

rthorn:
Add field to support new Media CDN feature (flexible shielding) via Terraform.

Added flex_shielding NestedObject[flex_shielding_regions Array[Enum [AFRICA_SOUTH1, ME_CENTRAL1]]]

Setting this field configures Media CDN to always fetch content from origin from the configured region, instead of Media CDN's default global origin shielding, which is still used if flexible shielding is not configured. If flexible shielding is configured, exactly one region must be set (but it is an Array in order to potentially support multiple regions in the future).

Fixes https://github.com/hashicorp/terraform-provider-google/issues/22376

```release-note:enhancement
networkservices: added `flex_shielding` field to `edge_cache_origin` resource
```


ScottSuarez (commented):
looks like this field, after being set, is not being returned in the API response. 

Is this field under some whitelist or is this expected behavior?

rthorn:
The change to remove the visibility restriction on this API field was submitted last week but likely has not rolled out yet.

ScottSuarez (commented):
waiting for api rollout

rthorn:
API rollout finished, tests now pass.

mmv1/products/networkservices/EdgeCacheOrigin.yaml
```diff
@@ -332,3 +332,26 @@ properties:
...
           type: String
         max_size: 5
+  - name: 'flexShielding'
+    type: NestedObject
+    description: |
+      The FlexShieldingOptions to be used for all routes to this origin.
+
+      If not set, defaults to a global caching layer in front of the origin.
+    properties:
+      - name: flexShieldingRegions
```
ScottSuarez, line 342: Can this field be updated? If so can we add an update test?

rthorn: Sorry, updated as in can we add new regions in the future? Yes but I'm not sure how to test that. Or updated as in users can update the value? I can set it to a different value in the EdgeCacheOrigin_update_1 test if that's sufficient

ScottSuarez: Sorry, updated as in can we add new regions in the future?
- Are you expecting the region list to be highly volatile? If so I would recommend using string for this with a external link to valid values in the description. (markdown is supported). This is recommended as maintenance can become a burden. 

Or updated as in users can update the value? I can set it to a different value in the EdgeCacheOrigin_update_1 test if that's sufficient.
- Yep that would work !

ScottSuarez (changes requested):

ScottSuarez:
I'm comfortable with this change as is. I can submit if you are ready.

rthorn:
Good with me, thank you!

PR #13701, opened 2025-04-16
Allow serial-port-logging-enable in google_workbench_instance metadata

google-mike:
Issue: https://github.com/hashicorp/terraform-provider-google/issues/22381

```release-note:bug
google_workbench_instance: `metadata` can now modify serial-port-logging-enable.
```

### DIFF
--- a/mmv1/templates/terraform/constants/workbench_instance.go.tmpl
+++ b/mmv1/templates/terraform/constants/workbench_instance.go.tmpl
@@ -68,7 +68,6 @@ var WorkbenchInstanceProvidedMetadata = []string{
     "report-system-status",
     "resource-url",
     "restriction",
-    "serial-port-logging-enable",
     "service-account-mode",
     "shutdown-script",
     "title",
--- a/mmv1/third_party/terraform/services/workbench/resource_workbench_instance_test.go.tmpl
+++ b/mmv1/third_party/terraform/services/workbench/resource_workbench_instance_test.go.tmpl
@@ -377,6 +377,7 @@ resource "google_workbench_instance" "instance" {
 
   gce_setup {
     metadata = {
+      "serial-port-logging-enable" = "true",
       terraform = "true"
       "resource-url" = "new-fake-value",
     }
@@ -401,6 +402,7 @@ resource "google_workbench_instance" "instance" {
       terraform = "true",
       "idle-timeout-seconds" = "10800",
       "image-url" = "fake-value",
+      "serial-port-logging-enable" = "true",
     }
   }
 


shuyama1 (commented):
Can you modify an existing test to verify this change? Also would you mind following go/terraform-contribution-guide#before-you-begin to get your account registered? Thanks!

google-mike:
Hello, sorry had to figure out some CLA stuff.  Should be all set now, I added the serial-port-enable metadata to the metadata tests.

google-mike:
Hello, I'm not sure why the tests are failing--I can't find any other references to serial-port-access, but it seems like the test failures still indicate there's a bug somewhere with the workbench metadata.  Any chance someone can assist me with this?

shuyama1 (commented):
Tests failed with error:
```
resource_workbench_instance_test.go:20: Step 3/4 error: After applying this test step, the plan was not empty.
        stdout:
        
        
          ~ update in-place
        
          # google_workbench_instance.instance will be updated in-place
          ~ resource "google_workbench_instance" "instance" {
                id                          = "projects/ci-test-project-188019/locations/us-central1-a/instances/tf-test-workbench-instancebhhdxu809e"
                name                        = "tf-test-workbench-instancebhhdxu809e"
                # (16 unchanged attributes hidden)
        
              ~ gce_setup {
                  ~ metadata             = {
                      - "serial-port-logging-enable" = "true" -> null
                        # (19 unchanged elements hidden)
                    }
                    tags                 = [
                        "deeplearning-vm",
                        "notebook-instance",
                    ]
        
                    # (6 unchanged blocks hidden)
                }
            }
        
        Plan: 0 to add, 1 to change, 0 to destroy
```
This is likely due to that the API returns the value while it's not explicitly specified and therefore it causes a permadiff. https://googlecloudplatform.github.io/magic-modules/develop/diffs/#fix-diffs

bcreddy-gcp:
TLDR:
`serial-port-logging-enable` can be set while creation. If nothing is specified, it is said to true and is returned by default. It cannot be modified. It is in diff suppress for 2 reasons. 
- It is returned on creation even when not specified.
- It cannot be modified.
Removing it from diffsuppress will create a perma-diff. Thoughts on how to solve this @shuyama1

shuyama1 (commented):
Okay, I think I understand the issue now. This PR aims to resolve the problem that it's unable to set `serial-port-logging-enable = FALSE` using Terraform. 

I've looked into this resource and suspect it might be a broader issue - The current implementation of the [diff suppress function for metadata](https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/templates/terraform/constants/workbench_instance.go.tmpl#L86) seems to prevent explicitly sending any of the flags listed in [WorkbenchInstanceProvidedMetadata](https://github.com/GoogleCloudPlatform/magic-modules/blob/665dd6d6ea5d673db3d7ccab59ccbb554b2af7c6/mmv1/templates/terraform/constants/workbench_instance.go.tmpl#L26C5-L26C38) during create or update API calls. What's worse, it doesn't report any diffs for these values to users, even when the values in the configuration do not align with the actual deployed state, because all diffs for these specific flags are suppressed.

@bcreddy-gcp Do you happen to know which values in the list of `WorkbenchInstanceProvidedMetadata` can be explicitly set at creation, which can be updated, and which are server-generated (output-only)?

bcreddy-gcp:
Let me compile a list of metadata keys with all the possible behaviors and we can look at how to address each of the behaviors.

bcreddy-gcp:
https://github.com/GoogleCloudPlatform/magic-modules/pull/14243 should solve what this PR is trying to do.

PR #13708, opened 2025-04-17
Add support to enable dns64 field in google_dns_policy resource

shellyvilenko:

```release-note:enhancement
dns: added 'enable_dns64_all_queries'  field to 'google_dns_policy' resource for all versions.
```


ScottSuarez:

mmv1/templates/terraform/pre_create/dns_policy_enable_dns64_all_queries.go.tmpl
```diff
@@ -0,0 +1,7 @@
+if val, ok := d.Get("enableDns64AllQueries").(bool); ok {
```
ScottSuarez, line 1:
```suggestion
if val, ok := d.Get("enable_dns64_all_queries").(bool); ok {
```

mmv1/products/dns/Policy.yaml
```diff
@@ -31,6 +31,7 @@ timeouts:
   delete_minutes: 20
 custom_code:
   pre_delete: 'templates/terraform/pre_delete/detach_network.tmpl'
+  pre_create: 'templates/terraform/pre_create/dns_policy_enable_dns64_all_queries.go.tmpl'
```
ScottSuarez, line 34: will also need pre_update (if supported) and post_read (to set the value from api read)

mmv1/templates/terraform/post_read/dns_policy_enable_dns64_all_queries.go.tmpl
```diff
@@ -0,0 +1,3 @@
+if err := d.Set("enable_dns64_all_queries", res["dns64Config"]["scope"]["allQueries"]); err != nil {
```
ScottSuarez, line 1: res["dns64Config"]["scope"]["allQueries"]

I think we might be indexing into this object through an unsafe pattern.

mmv1/templates/terraform/pre_create/dns_policy_enable_dns64_all_queries.go.tmpl
```diff
@@ -0,0 +1,7 @@
+if val, ok := d.Get("enable_dns64_all_queries").(bool); ok {
```
ScottSuarez, line 1:
```suggestion
if val, ok := d.Get("enable_dns64_all_queries").(bool); ok && val {
```

mmv1/products/dns/Policy.yaml
```diff
@@ -50,6 +53,12 @@ examples:
       org_id: 'ORG_ID'
       billing_account: 'BILLING_ACCT'
     exclude_docs: true
+virtual_fields:
+  - name: 'enableDns64AllQueries'
```
ScottSuarez, line 57:
```suggestion
  - name: 'enable_dns64_all_queries'
```

since this api doesn't have a field of this name we can use the terraform name here

ScottSuarez (changes requested):
so i tested this locally and I think we should be using d.GetOk here or just not sending if not true.

In the case that `enable_dns64_all_queries` is unset d.Get is returning false here for enable_dns64_all_queries causing the field to be sent over API.

mmv1/templates/terraform/pre_create/dns_policy_enable_dns64_all_queries.go.tmpl
```diff
@@ -0,0 +1,7 @@
...
+	dns64Config := make(map[string]interface{})
+	scope := make(map[string]interface{})
+	scope["allQueries"] = val
+	dns64Config["scope"] = scope
+	obj["dns64Config"] = dns64Config
+}
```
ScottSuarez, line 7:
```suggestion
if val, ok := d.GetOk("enable_dns64_all_queries"); ok {
	castedVal, _ := val.(bool)
	dns64Config := make(map[string]interface{})
	scope := make(map[string]interface{})
	scope["allQueries"] = castedVal
	dns64Config["scope"] = scope
	obj["dns64Config"] = dns64Config
}
```

ScottSuarez (changes requested):

shellyvilenko:
> so i tested this locally and I think we should be using d.GetOk here or just not sending if not true.
> 
> In the case that `enable_dns64_all_queries` is unset d.Get is returning false here for enable_dns64_all_queries causing the field to be sent over API.

Synced offline and decided to go with this approach as it can cause issues later on if we decide the deprecate the enable_dns64_all_queries and generate a new field that will support extensibility.

Anyways, our server default for this is false - so this should not make any difference now

PR #13711, opened 2025-04-17
Fix TestAccDNSRecordSet_routingPolicy

nbodepudi:
Fix https://github.com/hashicorp/terraform-provider-google/issues/21770 (b/401955357)

```release-note:REPLACEME
Fix broken TestAccDNSRecordSet_routingPolicy test
```

### DIFF
--- a/mmv1/third_party/terraform/services/dns/resource_dns_record_set_test.go.tmpl
+++ b/mmv1/third_party/terraform/services/dns/resource_dns_record_set_test.go.tmpl
@@ -320,7 +320,7 @@ func TestAccDNSRecordSet_routingPolicy(t *testing.T) {
 				ImportStateVerify: true,
 			},
 			{
-				Config: testAccDnsRecordSet_routingPolicyRegionalL7PrimaryBackup(networkName, proxySubnetName, httpHealthCheckName, backendName, urlMapName, httpProxyName, forwardingRuleName, zoneName, 300),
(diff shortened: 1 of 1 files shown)


mmv1/third_party/terraform/services/dns/resource_dns_record_set_test.go.tmpl
```diff
@@ -320,7 +320,7 @@ func TestAccDNSRecordSet_routingPolicy(t *testing.T) {
 				ImportStateVerify: true,
 			},
 			{
-				Config: testAccDnsRecordSet_routingPolicyRegionalL7PrimaryBackup(networkName, proxySubnetName, httpHealthCheckName, backendName, urlMapName, httpProxyName, forwardingRuleName, zoneName, 300),
+				Config: testAccDnsRecordSet_routingPolicyRegionalL7PrimaryBackup(networkName, "regional-l7-"+proxySubnetName, httpHealthCheckName, backendName, urlMapName, httpProxyName, "regional-l7-"+forwardingRuleName, zoneName, 300),
```
melinath, line 323: here & below: the value passed in for resource names needs to start with `tf-test` so that we can identify them as needing to be cleaned up if they leak. The variables already do this, but adding a prefix breaks it. Could you add this as a suffix instead?

melinath (changes requested):
It looks like the test is now passing, but I'm not sure I understand why. Could you explain briefly?

mmv1/third_party/terraform/services/dns/resource_dns_record_set_test.go.tmpl
nbodepudi: Cool will update it to suffix. I think something has changed in between with how the proxy subnets are handled which broke things with name conflict, so I update the tests to use unique names and seems working. Previously we used same resource for all the tests. 
Screenshot 2025-04-17 at 1 10 03 PM

PR #9348, opened 2023-10-24
Update node pool tests to use bootstrapped network

shuyama1:
**Release Note Template for Downstream PRs (will be copied)**

```release-note:none

```

PR #9265, opened 2023-10-14
bootstrap network used for vpc-native GKE clusters

shuyama1:
**Release Note Template for Downstream PRs (will be copied)**

```release-note:none
```

PR #13713, opened 2025-04-17
use bootstrapped network for gke cluster tests

shuyama1:
Switched to use the bootstrapped network instead of default network for [vpc-native (default mode) GKE clusters](https://cloud.google.com/kubernetes-engine/docs/concepts/alias-ips) to to avoid traffic to the default network. This follows the direction as https://github.com/GoogleCloudPlatform/magic-modules/pull/9265, https://github.com/GoogleCloudPlatform/magic-modules/pull/9348 and https://github.com/GoogleCloudPlatform/magic-modules/pull/9664

```release-note:none

```


zli82016 (approved):
LGTM

PR #13716, opened 2025-04-18
`appengine` add `oauth2_client_secret` write-only field to `resource_google_app_engine_application`

BBBmau:
```hcl
󰀵 mau  …/terraform-provider-google   main $!⇣   v1.24.2   14:00   envchain GCLOUD make test TEST=./google/services/appengine TESTARGS='--run TestAccAppEngineApplication_IAP_wo_basic'
sh -c "'/Users/mau/Dev/terraform-provider-google/scripts/gofmtcheck.sh'"
go vet
go test --run TestAccAppEngineApplication_IAP_wo_basic -timeout=30s ./google/services/appengine
ok      github.com/hashicorp/terraform-provider-google/google/services/appengine        1.486s
```

```release-note:enhancement
appengine: added `oauth2_client_secret` write-only field and `oauth2_client_secret_wo_version` field in `google_app_engine_application` resource
```

### DIFF
--- a/mmv1/third_party/terraform/services/appengine/resource_app_engine_application.go
+++ b/mmv1/third_party/terraform/services/appengine/resource_app_engine_application.go
@@ -6,6 +6,7 @@ import (
 	"log"
 	"time"
 
+	"github.com/hashicorp/go-cty/cty"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/customdiff"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/validation"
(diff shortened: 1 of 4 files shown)


mmv1/third_party/terraform/services/appengine/resource_app_engine_application.go
```diff
@@ -138,11 +138,11 @@ func ResourceAppEngineApplication() *schema.Resource {
 				Description: `The GCR domain used for storing managed Docker images for this app.`,
 			},
 			"iap": {
-				Type:        schema.TypeList,
-				Optional:    true,
-				Computed:    true,
-				MaxItems:    1,
-				Description: `Settings for enabling Cloud Identity Aware Proxy`,
+				Type:             schema.TypeList,
+				Optional:         true,
```
melinath, line 142: Removing Computed is a breaking change - is that necessary? https://googlecloudplatform.github.io/magic-modules/breaking-changes/breaking-changes/

BBBmau: unfortunately yes, the way write-only is designed prevents `Computed: true` from being set as you wouldn't expect to receive a value back from the API when the field has a `write_only: true` set on it.

The diffSuppress introduced follows a similar implementation of a recent write-only that was introduced:
- https://github.com/GoogleCloudPlatform/magic-modules/pull/16837
- https://github.com/GoogleCloudPlatform/magic-modules/pull/18840

basically to allow write-only a diffSuppress would be added where we suppress diffs only when the `iap` block is unset

i did some manual validation and did not experience any breaking changes with the diffSuppress put in place:, the following actions were performed in order:

## 8.4.0
1. when iap is empty -> iap is empty list in state  
2. when iap is populated -> iap has one value in the list
3. when iap is removed from config -> iap still has one value in list

## after upgrading to this branch
4. unsetting iap produces no infra changes -> introduces the new write-only values in state  
5. setting iap produces no infra changes
6. Replacing `oauth2_client_secret` with write_only field (same value)  -> produces a state change for `wo_version` + leaves `secret_sha256` value the exact same as what it was before using write-only
7. updating both `oauth2_client_secret_wo` and `oauth2_client_secret_wo_version` with new values -> performs an update to the `secret_sha256` that is stored in state 
- (the updated value to secret_sha256 is not shown in plan, though this isn't an issue since the value doesn't show on 8.4.0 when a new secret_sha256 is being introduced anyways)
8. removing the iap block from config -> no infra changes

knowing all this we should be good to merge as this wouldn't produce any infra changes that are unexpected.

PR #13724, opened 2025-04-20
Modifies AnywhereCache Deletion logic in `google_storage_bucket` resource

kautikdk:
This PR Modifies AnywhereCache deletion logic to respect `StorageBasePath` value rather than using static API endpoint for fetching and deleting AnywhereCaches.

This will allow Bucket deletion in other environments as well which can be configured by user with `storage_custom_endpoint` field in the provider block. Currently it's throwing an error while deleting bucket when using different storage endpoint.

```release-note:none

```

### DIFF
--- a/mmv1/third_party/terraform/services/storage/resource_storage_bucket.go.tmpl
+++ b/mmv1/third_party/terraform/services/storage/resource_storage_bucket.go.tmpl
@@ -590,9 +590,12 @@ func labelKeyValidator(val interface{}, key string) (warns []string, errs []erro
 	return
 }
 
-func getAnywhereCacheListResult(config *transport_tpg.Config, bucket string) ([]interface{}, error) {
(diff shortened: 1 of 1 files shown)


googlyrahman:
The idea of making direct calls, rather than using transport was taken from existing code present in  [resource_storage_bucket_sweeper](https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/third_party/terraform/services/storage/resource_storage_bucket_sweeper.go#L189)

Had an offline conversation with @kautikdk on this, This do not have any customer dependency, and it's required for pre-prod development. 

LGTM

googlyrahman (approved):

kautikdk:
> Just wondering, is this fix somewhat related to [hashicorp/terraform-provider-google#22359](https://github.com/hashicorp/terraform-provider-google/issues/22359)?

This appears to be a different issue and @googlyrahman is looking into this.

hao-nan-li (approved):
Thanks for letting me know.

PR #13725, opened 2025-04-21
Add a copy file header to state the source file for handwritten files

iyabchen:
Only adds copy file header for .go and .markdown files, let me know if more files needs that. 

https://github.com/hashicorp/terraform-provider-google/issues/21862

```release-note:none

```


iyabchen:
Most test results seem to aligned with the vcr-nighly: https://pantheon.corp.google.com/cloud-build/builds;region=global/076639e1-e0fa-4870-9c07-4aa843310efd;step=0?e=13803378&mods=monitoring_api_prod&project=graphite-docker-images
except `TestAccDataSourceGoogleGkeHubFeature_basic` return "no cassette found on disk for TestAccDataSourceGoogleGkeHubFeature_basic", but the recording was green before rerunning replay.

mmv1/provider/terraform.go
```diff
@@ -507,10 +509,62 @@ func (t Terraform) CompileFileList(outputFolder string, files map[string]string,
...
+// ----------------------------------------------------------------------------
+//
+//     This code is generated by Magic Modules using the following:
+//
+//     Source file: %s
+//
+//     DO NOT EDIT this file directly. Any changes made to this file will be
+//     overwritten during the next generation cycle.
+//
+// ----------------------------------------------------------------------------` + "\n"
```
shuyama1, line 549: Would it be possible to move this to a md template instead of hardcoding it?

iyabchen: Done.

mmv1/provider/terraform.go
```diff
@@ -507,10 +509,62 @@ func (t Terraform) CompileFileList(outputFolder string, files map[string]string,
...
+	}
+
+	srcStr := string(sourceByte)
+	if strings.Contains(srcStr, "***     AUTO GENERATED CODE    ***    Type: MMv1     ***") {
+		return
+	}
+
+	headerFormat := `// ----------------------------------------------------------------------------
+//
+//     ***     AUTO GENERATED CODE    ***    Type: MMv1     ***
```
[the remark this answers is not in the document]
iyabchen: Done.

mmv1/provider/terraform.go
```diff
@@ -507,10 +510,74 @@ func (t Terraform) CompileFileList(outputFolder string, files map[string]string,
...
+	if err != nil {
+		log.Fatalf("Cannot read file %s to add copy file header: %s", targetFile, err)
+	}
+
+	srcStr := string(sourceByte)
+	if strings.Contains(srcStr, "***     AUTO GENERATED CODE    ***    Type: Handwritten     ***") {
+		return
+	}
+
+	templateFile := "templates/terraform/handwritten_file.go.tmpl"
```
shuyama1, line 537: Okay, may bad for missing the syntax difference between Go and Markdown files. Now, thinking about it more, managing two templates could create extra maintenance work. Your initial idea of using inline code seems a better solution, and it's also consistent with how the generated files are handled.  https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/api/resource.go#L1895
I think we should go back to that. Apologies for it!

shuyama1 (commented):
Test failures are unrelated. Sorry for the noise. Please remove the two templates. Thanks!

mmv1/provider/terraform.go
iyabchen: Done

shuyama1 (approved):
LGTM! Thank you @iyabchen. Just to be extra cautious, I'll merge this tomorrow to avoid any provider changes this afternoon, as we are cutting release tonight.

shuyama1:
Will merge after the patch release today.

shuyama1:
final check before merging
/gcbrun

PR #13727, opened 2025-04-21
Add directory services config field to Filestore instance in beta

michaeltur3:
Directory Services configuration support for Filestore.

```release-note: enhancement
filestore: added `directory_services` field to `google_filestore_instance` (beta)
```


melinath (changes requested):
Please also test updating the field as described in https://googlecloudplatform.github.io/magic-modules/test/test/#add-an-update-test

michaeltur3:
We do not support updating this field at this moment.
Added `immutable: true` to it.

melinath (changes requested):
immutability is not automatically inherited; it's possible in GCP to have a top-level field that's immutable but subfields can be updated in-place. Is that the case here, or should all of these fields be marked immutable?

melinath:
@michaeltur3 pointed out that our contribution docs say that immutability _is_ inherited, so I need to double-check this. If that's correct then this will be ready to merge.

melinath:
Doesn't seem like the docs are correct - please add immutable: true to all the fields.

michaeltur3:
> Doesn't seem like the docs are correct - please add immutable: true to all the subfields.

Done

melinath (approved):

PR #13728, opened 2025-04-21
Update Misleading Documentation in CloudSQL Data Cache Behaviour

nizmitz:
This PR update the documentation of `google_sql_database_instance` where `data_cache` argument doesn't meet with the current api behavior where data_cache_config will defaults to `true` instead of `false`  

Fixes: https://github.com/hashicorp/terraform-provider-google/issues/22427

```release-note: note
cloudsql: update documentation of `google_sql_database_instance`
```


melinath (commented):
This fix is not technically correct (though I can see why it's confusing). I'm not sure offhand what the best solution will be; I'll need to think about it.

PR #13730, opened 2025-04-21
Remove extra permission required for doing a destory in resource_storage_bucket

googlyrahman:
Fixes https://github.com/hashicorp/terraform-provider-google/issues/22359

```release-note: bug
storage: removed extra permission (storage.anywhereCaches.list) required for doing a destroy in `resource_storage_bucket`
```

### DIFF
--- a/mmv1/third_party/terraform/services/storage/resource_storage_bucket.go.tmpl
+++ b/mmv1/third_party/terraform/services/storage/resource_storage_bucket.go.tmpl
@@ -1086,7 +1086,9 @@ func resourceStorageBucketDelete(d *schema.ResourceData, meta interface{}) error
 
 		cacheList, cacheListErr := getAnywhereCacheListResult(config, bucket)
 		if cacheListErr != nil {
-			return cacheListErr
+			// If we get any error, try deleting the bucket anyway in case it's empty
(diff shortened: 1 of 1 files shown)


mmv1/third_party/terraform/services/storage/resource_storage_bucket.go.tmpl
```diff
@@ -1086,7 +1086,9 @@ func resourceStorageBucketDelete(d *schema.ResourceData, meta interface{}) error
 
 		cacheList, cacheListErr := getAnywhereCacheListResult(config, bucket)
 		if cacheListErr != nil {
-			return cacheListErr
+			// If we get any error, try deleting the bucket anyway in case it's empty
+			// This would help our customers to avoid requiring extra storage.anywhereCaches.list permission.
+			break
```
ScottSuarez, line 1091: I not sure I agree with this implementation. The user should not be able to delete with-out force_destory in this case as we do not know the state of the buckets content. See line 1112

ScottSuarez: reasonable. thanks!

ScottSuarez (approved):

PR #13731, opened 2025-04-21
Expand Account Connector terraform to add new providers

ihnarayanan:
fixes hashicorp/terraform-provider-google#22439

```release-note:none
```

### DIFF
--- a/mmv1/products/developerconnect/AccountConnector.yaml
+++ b/mmv1/products/developerconnect/AccountConnector.yaml
@@ -44,6 +44,14 @@ examples:
     primary_resource_id: 'my-account-connector'
     vars:
       account_connector_name: tf-test-ac
+  - name: developer_connect_account_connector_new_relic
+    primary_resource_id: 'my-account-connector'
+    vars:
+      account_connector_name: tf-test-ac
+  - name: developer_connect_account_connector_datastax
+    primary_resource_id: 'my-account-connector'
+    vars:
+      account_connector_name: tf-test-ac
 autogen_async: true
 async:
   operation:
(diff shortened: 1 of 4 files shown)


c2thorn (commented):
this seems to be a documentation-only change because the `systemProviderId` field already supports the new values due to the lack of provider-level validation on the field.
Typically we do not leave a release note for documentation-only changes. Do you think this is worth calling out anyway? @ihnarayanan

ihnarayanan:
> this seems to be a documentation-only change because the `systemProviderId` field already supports the new values due to the lack of provider-level validation on the field. Typically we do not leave a release note for documentation-only changes. Do you think this is worth calling out anyway? @ihnarayanan

I don't think it's necessary to call out in a release note, so I'll remove it from the PR. Lmk if that works @c2thorn

ihnarayanan:
> > this seems to be a documentation-only change because the `systemProviderId` field already supports the new values due to the lack of provider-level validation on the field. Typically we do not leave a release note for documentation-only changes. Do you think this is worth calling out anyway? @ihnarayanan
> 
> I don't think it's necessary to call out in a release note, so I'll remove it from the PR. Lmk if that works @c2thorn

Nvm, it looks like some release note is required for the changelog-checker. I changed the type from `enhancement` to `note` as outlined [here](https://googlecloudplatform.github.io/magic-modules/code-review/release-notes/#write-release-notes).

c2thorn:
> > this seems to be a documentation-only change because the `systemProviderId` field already supports the new values due to the lack of provider-level validation on the field. Typically we do not leave a release note for documentation-only changes. Do you think this is worth calling out anyway? @ihnarayanan
> 
> I don't think it's necessary to call out in a release note, so I'll remove it from the PR. Lmk if that works @c2thorn

thanks for the confirmation, a release note of type `none` is suitable. I've updated it now

c2thorn (approved):

PR #13733, opened 2025-04-21
Use the correct form of `sha256.Sum256`.

mcsaucy:
`sha256.New().Sum(foo)` appends the SHA-256 hash for nil to foo. See https://go.dev/play/p/vSW0U3Hq4qk

Adapted from hashicorp/terraform-provider-google#22428

```release-note:bug
compute: fixed improper SHA256 calculations on `google_compute_ssl_certificate` when suppressing diff
```

### DIFF
--- a/mmv1/templates/terraform/constants/compute_certificate.go.tmpl
+++ b/mmv1/templates/terraform/constants/compute_certificate.go.tmpl
@@ -1,5 +1,6 @@
 // sha256DiffSuppress
 // if old is the hex-encoded sha256 sum of new, treat them as equal
 func sha256DiffSuppress(_, old, new string, _ *schema.ResourceData) bool {
-	return hex.EncodeToString(sha256.New().Sum([]byte(old))) == new
+	h := sha256.Sum256([]byte(old))
+	return hex.EncodeToString(h[:]) == new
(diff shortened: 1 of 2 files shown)


slevenick (commented):
Is this testable in a reasonable way? It would be good to keep a test for this piece of code to make sure it never drifts

mcsaucy:
Are you thinking testing the underlying behavior in the rendered templates? I'm having trouble seeing how this all works past the templates, so I'm having trouble thinking of tests. Is there any prior art I can copy from?

slevenick (commented):
What I'd like to test is some failure case where we were getting the wrong SHA and suppressing a diff based on that. Or writing the wrong value to state based on the incorrect flatten function.

One way to do this is to create a resource with a deterministic hash and then assert the SHA is what we think it should be. For example, using a Check function: https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/third_party/terraform/provider/provider_billing_project_test.go#L98

Because this value is coming from the API though I don't know if it's possible to deterministically produce

Another way to figure out what to test is to look at failure cases for the existing resource. Why did you notice this wasn't returning the right value?

PR #13736, opened 2025-04-22
Support setting up date/time in rollout_restriction with datetime strings.

ChrisGe4:
Fixes https://github.com/hashicorp/terraform-provider-google/issues/22458. 

This resource is newly added.  We found out it's not easy to set up the resource because users need to manually create the date/time proto message. So we replaced the messages with string fields.

```release-note:enhancement
PRODUCT: added `start` and `end` string fields to `one_time_windows`  and changed the type of `start_time` and `end_time` fields to string in `weekly_windows` in  `google_clouddeploy_deploy_policy` resource.
```


mmv1/products/clouddeploy/DeployPolicy.yaml
```diff
@@ -175,71 +175,19 @@ properties:
...
-                            description: "Month of a year. Must be from 1 to 12, or 0 to specify a year without a month and day."
-                          - name: day
-                            type: Integer
-                            description: "Day of a month. Must be from 1 to 31 and valid for the year and month, or 0 to specify a year by itself or a year and month where the day isn't significant."
-                      - name: "endDate"
-                        type: NestedObject
-                        description: "Required. End date."
+                      - name: "end"
+                        type: String
+                        description: "Required. End date and time,  in the datetime format, e.g. 2025-01-02 09:00:00."
```
BBBmau, line 189: no need to include required in the description the generator includes it in the docs automatically when a field has `required: true`

mmv1/products/clouddeploy/DeployPolicy.yaml
```diff
@@ -175,71 +175,19 @@ properties:
...
                   description: "Optional. One-time windows within which actions are restricted."
+                  custom_expand: "templates/terraform/custom_expand/clouddeploy_policy_onetime_windows.go.tmpl"
+                  custom_flatten: "templates/terraform/custom_flatten/clouddeploy_policy_onetime_windows.go.tmpl"
                   item_type:
                     type: NestedObject
                     properties:
-                      - name: "startDate"
-                        type: NestedObject
-                        description: "Required. Start date."
+                      - name: "start"
```
BBBmau, line 183: what's the reasoning behind wanting to change it to just a string? This would be a breaking change due to removing the original nestedObject to instead just be string inputs. Is the current method not supporting a specific use case?

ChrisGe4: The intention is to make it easier for user to set the date time.  To avoid the breaking change, if I keep the existing fields(defined in the API) and add the two new string fields, what is the best way to flatten the rules array to pass the acceptance test? I tried to write custom flatten in various ways but always get `ImportStateVerify attributes not equivalent. Difference is shown below. The - symbol indicates attributes missing after import.` error.

ChrisGe4:
After discussed with the Terraform team, we are not going to make this change.

PR #13681, opened 2025-04-15
upgrade google.golang.org/api to v0.229.0

halimsam:
**Release Note Template for Downstream PRs (will be copied)**

See [Write release notes](https://googlecloudplatform.github.io/magic-modules/contribute/release-notes/) for guidance.

```release-note:none
upgrade google.golang.org/api to v0.229.0
```

Ran `go get google.golang.org/api@latest` and `go mod tidy` to get the latest Go API version.

PR #11883, opened 2024-10-01
Upgrade dcl to 1.74.0

c2thorn:
also made `make upgrade-dcl` work again after the go rewrite

**Release Note Template for Downstream PRs (will be copied)**

```release-note:enhancement
containeraws: added `kubelet_config` field group to the `google_container_aws_node_pool` resource
```
```release-note:enhancement
assuredworkloads: added `HEALTHCARE_AND_LIFE_SCIENCES_CONTROLS` and `HEALTHCARE_AND_LIFE_SCIENCES_CONTROLS_WITH_US_SUPPORT` enum values to `compliance_regime` in the `google_assuredworkload_workload` resource
```

PR #13737, opened 2025-04-22
Bigtable: upgrade cloud.google.com/go/bigtable to v1.37.0

shantstepanian:
```release-note:none
```

Transitive dependencies updated via `go get -u` and `go mod tidy`


ron-gal (approved):

hao-nan-li (commented):
I don't think a changelog is necessary for this PR. Is there any particular reason to upgrade the version?

shantstepanian:
> I don't think a changelog is necessary for this PR. Is there any particular reason to upgrade the version?

Hi @hao-nan-li , the reason for upgrading the Bigtable client version is #13391 (that code relies on the client version upgrade). I was asked to split the required version upgrade for that into this CL, presumably since this is a simpler change and can safely go on its own

Re: the changelog, I'm flexible w/ the message to write. I was modeling my commit message after others like #13681  and  #11883 . Is there a specific way you want me to rewrite the title or commit message? (e.g. remove the 'bigtable:'?)

hao-nan-li:
> > I don't think a changelog is necessary for this PR. Is there any particular reason to upgrade the version?
> 
> Hi @hao-nan-li , the reason for upgrading the Bigtable client version is #13391 (that code relies on the client version upgrade). I was asked to split the required version upgrade for that into this CL, presumably since this is a simpler change and can safely go on its own
> 
> Re: the changelog, I'm flexible w/ the message to write. I was modeling my commit message after others like #13681 and #11883 . Is there a specific way you want me to rewrite the title or commit message? (e.g. remove the 'bigtable:'?)

Sounds good, thanks for the explanation. I can edit the changelog so no worries.

hao-nan-li (approved):

PR #13743, opened 2025-04-23
`resourceIdentity`: initial beta pkg setup - sdk/v2 + plugin-testing

BBBmau:
initial setup of necessary dependencies for resourceIdentity support

```release-note:none

```


BBBmau:
it looks like it might've just done a fresh cassette?

mmv1/third_party/terraform/go.mod
```diff
@@ -38,6 +38,14 @@ require (
 	google.golang.org/protobuf v1.36.6
 )
 
+replace github.com/hashicorp/terraform-plugin-sdk/v2 => github.com/hashicorp/terraform-plugin-sdk/v2 v2.37.0-beta.1
+
+replace github.com/hashicorp/terraform-plugin-go => github.com/hashicorp/terraform-plugin-go v0.27.0-alpha.1
+
+replace github.com/hashicorp/terraform-plugin-mux => github.com/hashicorp/terraform-plugin-mux v0.19.0-alpha.1
+
+replace github.com/hashicorp/terraform-plugin-testing => github.com/hashicorp/terraform-plugin-testing v1.13.0-beta.1
```
melinath, line 47: I'm concerned about the idea of forcing these to point to alpha versions - unless this is just for testing?

melinath (changes requested):
these look like real failures to me. Please let me know if there's something more specific you're looking for from me!

mmv1/third_party/terraform/go.mod
BBBmau: yeah this is just for initial testing, this would be changed once there's a non-alpha / non-beta offering

BBBmau: i've removed the use of `terraform-plugin-testing -> 1.13.0-beta1` due to experiencing this [bug](https://github.com/hashicorp/terraform-plugin-testing/issues/327) when uprading to `1.6.0 >`

melinath:
Converted to draft for now since I think you're still iterating? Apologies if I misunderstood.

BBBmau:
> Converted to draft for now since I think you're still iterating? Apologies if I misunderstood.

this is fine, the latest commit was more just to get some logs that come from using <`1.5.0` plugin-testing to see what we can do about the testing issues coming from anything past `>1.6.0`

Once something is more solidified around `plugin-testing` ill move back to in review though i imagine it'll involve reverting the previous commit to stick with using `1.5.1` plugin-testing

melinath (commented):
It looks like this is still in process - relying on alpha versions of packages and lots of failing tests. I'm not sure exactly what I'm supposed to be reviewing?

BBBmau:
> It looks like this is still in process - relying on alpha versions of packages and lots of failing tests. I'm not sure exactly what I'm supposed to be reviewing?

The failing tests exist due to them already failing currently on main, taking a look at the nightly-tests we can see examples for the following test [- TestAccCloudbuildWorkerPool_basic](https://hashicorp.teamcity.com/buildConfiguration/TerraformProviders_GoogleCloud_GOOGLE_BETA_NIGHTLYTESTS_GOOGLEBETA_PACKAGE_CLOUDBUILD/365992)

but we see some tests failing that were previously failing and were just recently resolved after the `FEATURE-BRANCH` was created. an example is [TestAccComputeResourcePolicy_resourcePolicyPlacementPolicyGpuTopologyExample](https://hashicorp.teamcity.com/test/-3900219190016506658?currentProjectId=TerraformProviders_GoogleCloud_GOOGLE_BETA_NIGHTLYTESTS)

The only real test that is concerning to me is [TestAccAlloydbDatabaseInstanceDatasourceConfig](https://hashicorp.teamcity.com/test/-8645780363068218824?currentProjectId=TerraformProviders_GoogleCloud_GOOGLE_BETA_NIGHTLYTESTS) since we shouldn't be seeing any failure introduced from this PR as it's only a dependency bump of terraform packages. Though [I came across a failure](https://hashicorp.teamcity.com/test/-8645780363068218824?currentProjectId=TerraformProviders_GoogleCloud_GOOGLE_NIGHTLYTESTS&logView=flowAware) that matches the failure being outputted in the PR, I believe it's also part of the list of tests that are fixed after this PR was made.

I'll perform a sync to see how that addresses the tests on here, when it comes to the alpha packages they only exist as an initial setup but would be updated to the GA version upon it's release. However in order to start with the reviews of individual resourceIdentity support PRs having this merged is necessary. @melinath

melinath (approved):
oh, this is a PR against a feature branch. Sorry, I missed that part. go for it.

PR #13750, opened 2025-04-23
http rules principals coming unavailability

maxi-cit:
According to recent communication from a member from GCP networking. This 'principal' field will be under some reworking process. Dev ask for a way to temporarily remove/make it unavailable until they are done with the rework.

Since this might incur in a breaking-change, we just want to add a warning message so users are discouraged to use 'principal' field which will throw a not valid field error if they try to. The idea is to remove the warning once field is available again.

```release-note:note
networksecurity: added warning to prevent use of `http_rules.source.principals` and `http_rules.not_source.principals` from `google_network_security_authz_policy`.
```


melinath (commented):

PR #13757, opened 2025-04-23
Remove bad ExpectError from data. google_kms_crypto_key_latest_versio…

rileykarson:
Fixes https://github.com/hashicorp/terraform-provider-google/issues/20110

I think the contributor misunderstood when to use ExpectError? We're referencing a bootstrapped key that has a version in it.

```release-note:none
```


rileykarson:
VCR had a destroyed key as latest, but nightlies have an enabled one. Added a new VCR key.

/gcbrun

ScottSuarez (approved):

PR #13766, opened 2025-04-24
Add View resource for contactcenterinsights

scawful:
Add `View` resource for `contactcenterinsights`
https://cloud.google.com/contact-center/insights/docs/reference/rest/v1/projects.locations.views

```release-note:new-resource
`google_contact_center_insights_view`
```


c2thorn:
enabling contactcenterinsights.googleapis.com in our test projects

c2thorn (changes requested):
```
    resource_contact_center_insights_view_generated_test.go:40: Step 1/2 error: Error running apply: exit status 1
        
        Error: Error creating View: googleapi: Error 400: Invalid query parameters. Empty query parameter names are not allowed.
        
          with google_contact_center_insights_view.basic_view,
          on terraform_plugin_test.tf line 2, in resource "google_contact_center_insights_view" "basic_view":
           2: resource "google_contact_center_insights_view" "basic_view" {
```

c2thorn (changes requested):
seeing a `LOCATION_POLICY_VIOLATED` error. Does this pass locally with your own credentials and is just specific to our environment?

scawful:
The test run passes locally for me, I'm not sure why the region name is randomly generated in the debug log.

`ok  	github.com/hashicorp/terraform-provider-google/google/services/contactcenterinsights	0.230s`

mmv1/products/contactcenterinsights/View.yaml
```diff
@@ -0,0 +1,70 @@
...
+update_mask: true
+id_format: projects/{{project}}/locations/{{location}}/views/{{view}}
+import_format:
+  - projects/{{project}}/locations/{{location}}/views/{{view}}
+examples:
+  - name: 'contact_center_insights_view_basic'
+    primary_resource_id: 'basic_view'
+    vars:
+      resource_name: 'basic_view'
+      location_name: 'us-central1'
```
c2thorn, line 33: The use of a variable for `location_name` automatically appends a random suffix to the variable value. There doesn't seem to be a need for location to be a variable in these examples, so you can hard code them to be `us-central1` in the yaml.

c2thorn (changes requested):

scawful:
Looking at the debug log there seems to be a problem with properly deleting the resource https://cloud.google.com/contact-center/insights/docs/reference/rest/v1/projects.locations.views/delete 

I added a delete_url to the yaml but that didn't seem to help.

c2thorn (commented):
Refreshing the review timer as it is waiting on valid test runs
@scawful - do you feel like you have an idea of what needs to be fixed for each iteration? It seems like there were multiple different fixes you've gone through. Let me know if you have specific questions or get stuck

scawful:
@c2thorn Thanks for refreshing that and sorry if this PR has been noisy. I think I understand mostly but it still seems I've configured the resource name or ID incorrectly where it's not able to properly delete the resources. Could you take a look and see why that would be the case?

`Deleting View "projects/ci-test-project-188019/locations/us-central1/views/"`

The logs show that it omitted the ID of the view from the request and there is a small collection of view resources from previous test runs. Would this be a case where I should use test sweepers? https://googlecloudplatform.github.io/magic-modules/reference/resource/#sweeper

mmv1/products/contactcenterinsights/View.yaml
```diff
@@ -0,0 +1,71 @@
...
+  api: 'https://cloud.google.com/contact-center/insights/docs/reference/rest/v1/projects.locations.views'
+base_url: projects/{{project}}/locations/{{location}}/views
+self_link: projects/{{project}}/locations/{{location}}/views/{{view}}
+create_url: projects/{{project}}/locations/{{location}}/views
+delete_url: projects/{{project}}/locations/{{location}}/views/{{view}}
+update_verb: PATCH
+update_mask: true
+id_format: projects/{{project}}/locations/{{location}}/views/{{view}}
+import_format:
+  - projects/{{project}}/locations/{{location}}/views/{{view}}
```
c2thorn:
```suggestion
self_link: projects/{{project}}/locations/{{location}}/views/{{name}}
create_url: projects/{{project}}/locations/{{location}}/views
delete_url: projects/{{project}}/locations/{{location}}/views/{{name}}
update_verb: PATCH
update_mask: true
id_format: projects/{{project}}/locations/{{location}}/views/{{name}}
import_format:
  - projects/{{project}}/locations/{{location}}/views/{{name}}
```
name is acting as {{view}} from with the use of the custom flattener. Whatever is in {{ }} needs to be a field that exists in this YAML schema

mmv1/third_party/terraform/services/contactcenterinsights/resource_contact_center_insights_view_test.go
```diff
@@ -0,0 +1,108 @@
...
+		create_time = "2024-01-01T00:00:00Z"
+		update_time = "2024-01-01T00:00:00Z"
+		value    = "medium=\"PHONE_CALL\""
+	}
+	`, context)
+}
+
+func testAccContactCenterInsightsView_full(context map[string]interface{}) string {
+	return acctest.Nprintf(`
+resource "google_contact_center_insights_view" "full_view" {
```
c2thorn, line 54: I think going from testAccInsightsView -> testAccContactCenterInsightsView_full is destroying the "default" resource and creating a new "full_view" resource. Just making sure that this was your intention

mmv1/third_party/terraform/services/contactcenterinsights/resource_contact_center_insights_view_test.go
```diff
@@ -0,0 +1,108 @@
...
+import (
+	"testing"
+
+	"github.com/hashicorp/terraform-plugin-testing/helper/resource"
+	"github.com/hashicorp/terraform-plugin-testing/plancheck"
+	"github.com/hashicorp/terraform-provider-google/google/acctest"
+	"github.com/hashicorp/terraform-provider-google/google/envvar"
+)
+
+func TestInsightsView(t *testing.T) {
```
c2thorn: Let's rename this to 
```suggestion
func TestAccContactCenterInsightsView_update(t *testing.T) {
```
for consistency with the generated examples, and to actually have it running in CI. I think it needs to be prefixed with `TestAcc`

c2thorn (changes requested):
Seeing the following interaction in the debug logs:
```
POST /v1/projects/<proj>/locations/us-central1/views?alt=json HTTP/1.1

{
 "displayName": "view-display-name",
 "name": "tf-test-insights-view-basic",
 "value": "medium=\"CHAT\""
}

2025/06/05 17:34:11 [DEBUG] Google API Response Details:
HTTP/2.0 200 OK

{
  "name": "projects/<num>/locations/us-central1/views/8903367393471769374",
  "displayName": "view-display-name",
  "createTime": "2025-06-05T17:34:11.533326Z",
  "updateTime": "2025-06-05T17:34:11.533326Z",
  "value": "medium=\"CHAT\""
}
```
we're sending the name as `tf-test-insights-view-basic` but getting it back as the long name. Are we supposed to send the name there?

mmv1/third_party/terraform/services/contactcenterinsights/resource_contact_center_insights_view_test.go
scawful: Thanks for pointing this out

mmv1/third_party/terraform/services/contactcenterinsights/resource_contact_center_insights_view_test.go
scawful: I decided to remove that test to focus on the one resource.

scawful:
> we're sending the name as `tf-test-insights-view-basic` but getting it back as the long name. Are we supposed to send the name there?

As it turns out, and I should've doubled checked this but the view_id at the end of the resource name is actually randomly generated by the system. Is there any specific guidance on handling this? I've removed name fields and marked it as output rather than required but not sure if that's the correct way to handle this with the  "examples" section of the resource yaml which has a notion of resource_id and resource_name which don't necessarily apply when the ID is randomly generated. @c2thorn

c2thorn:
> As it turns out, and I should've doubled checked this but the view_id at the end of the resource name is actually randomly generated by the system. Is there any specific guidance on handling this?

For the resource schema, I think you've done it correctly. We used to require custom post-create code to apply the generated ID field after the resource is created, but I think we do that automatically now recently as part of https://github.com/hashicorp/terraform-provider-google/issues/22214 and https://github.com/GoogleCloudPlatform/magic-modules/pull/14146

> I've removed name fields and marked it as output rather than required but not sure if that's the correct way to handle this with the "examples" section of the resource yaml which has a notion of resource_id and resource_name which don't necessarily apply when the ID is randomly generated. @c2thorn

primary_resource_id is just for the Terraform config id of the resource, which you have correctly set.
You no longer need a resource_name as it will be randomly generated.

mmv1/products/contactcenterinsights/View.yaml
```diff
@@ -0,0 +1,71 @@
...
+    url_param_only: true
+    required: true
+properties:
+  - name: name
+    type: String
+    description: |-
+      Immutable. The resource name of the view.
+      Format:
+      projects/{project}/locations/{location}/views/{view}
+    immutable: true
```
c2thorn:
```suggestion
```

mmv1/products/contactcenterinsights/View.yaml
```diff
@@ -0,0 +1,71 @@
...
+    url_param_only: true
+    required: true
+properties:
+  - name: name
+    type: String
+    description: |-
+      Immutable. The resource name of the view.
+      Format:
+      projects/{project}/locations/{location}/views/{view}
```
c2thorn: It's not the long form so it isn't in this format?

mmv1/products/contactcenterinsights/product.yaml
```diff
@@ -0,0 +1,21 @@
+# Copyright 2024 Google Inc.
```
c2thorn:
```suggestion
# Copyright 2025 Google Inc.
```

mmv1/products/contactcenterinsights/View.yaml
```diff
@@ -0,0 +1,71 @@
+# Copyright 2024 Google Inc.
```
c2thorn:
```suggestion
# Copyright 2025 Google Inc.
```

mmv1/products/contactcenterinsights/View.yaml
scawful: I see, I think I got confused trying to do a 1:1 translation of our API docs to terraform, but didn't fully understand how Terraform handles resource names/IDs yet.

scawful:
Looks like the Insights tests are passing now @c2thorn and the failing tests are unrelated, maybe just need to sync the repo. How does it look to you?

c2thorn (approved):
LGTM, the failing tests should be unrelated.

PR #13770, opened 2025-04-24
Fix handling of 405 merge in progress

ScottSuarez:

```release-note:none

```

### DIFF
--- a/.ci/magician/github/get.go
+++ b/.ci/magician/github/get.go
@@ -116,7 +116,7 @@ func (gh *Client) GetCommitMessage(owner, repo, sha string) (string, error) {
 		} `json:"commit"`
 	}
 
-	err := utils.RequestCall(url, "GET", gh.token, &commit, nil)
+	err := utils.RequestCallSilent(url, "GET", gh.token, &commit, nil)
 	if err != nil {
 		return "", err
 	}
--- a/.ci/magician/github/get.go
+++ b/.ci/magician/github/get.go
@@ -116,7 +116,7 @@ func (gh *Client) GetCommitMessage(owner, repo, sha string) (string, error) {
 		} `json:"commit"`
 	}
 
-	err := utils.RequestCallSilent(url, "GET", gh.token, &commit, nil)
+	err := utils.RequestCall(url, "GET", gh.token, &commit, nil)
 	if err != nil {
 		return "", err
 	}
--- a/.ci/magician/github/set.go
+++ b/.ci/magician/github/set.go
@@ -19,6 +19,7 @@ import (
 	"fmt"
 	utils "magician/utility"
 	"strings"
+	"time"
 )
 
 func (gh *Client) PostBuildStatus(prNumber, title, state, targetURL, commitSha string) error {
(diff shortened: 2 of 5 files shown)


.ci/magician/github/get.go
```diff
@@ -116,7 +116,7 @@ func (gh *Client) GetCommitMessage(owner, repo, sha string) (string, error) {
 		} `json:"commit"`
 	}
 
-	err := utils.RequestCall(url, "GET", gh.token, &commit, nil)
+	err := utils.RequestCallSilent(url, "GET", gh.token, &commit, nil)
```
melinath, line 119: Why does this need to be silent? We generally want to log all of these requests for debugging purposes.

.ci/magician/github/set.go
```diff
@@ -165,7 +166,10 @@ func (gh *Client) MergePullRequest(owner, repo, prNumber, commitSha string) erro
 		// Check if the error is "Merge already in progress" (405)
 		if strings.Contains(err.Error(), "Merge already in progress") {
 			fmt.Printf("Pull request %s is already being merged\n", prNumber)
-			return nil
+			// This status does not indicate that the Pull Request was merged
+			// Try again after 10s
+			time.Sleep(20000)
+			return gh.MergePullRequest(owner, repo, prNumber, commitSha)
```
melinath, line 172:
```suggestion
			// This status does not indicate that the Pull Request was merged
			// Try again after 20s
			time.Sleep(20 * time.Second)
			return gh.MergePullRequest(owner, repo, prNumber, commitSha)
```

melinath (changes requested):

.ci/magician/github/get.go
ScottSuarez: Thomas remove the printing of the result for POST requests because this API outputs a lot. I am pretty sure we are listing a lot of commits so I can understand.

I added the printing of the POST back but thomas was saying that this caused the test output to be too long. I don't think we should be testing this in a presubmit anyways but this is just to unblock adding back the POST result to console

.ci/magician/utility/utils.go
```diff
@@ -119,7 +125,17 @@ func processResponse(resp *http.Response, respBodyBytes []byte, result any) erro
...
-	resp, respBodyBytes, err := makeHTTPRequest(url, method, credentials, body)
+	resp, respBodyBytes, err := makeHTTPRequest(url, method, credentials, body, true)
+	if err != nil {
+		return err
+	}
+
+	return processResponse(resp, respBodyBytes, result)
+}
+
+// RequestCall makes a single HTTP request without retries
```
melinath, line 136:
```suggestion
// RequestCallSilent makes a single HTTP request without retries and does not log the request and response
```

melinath (approved):
I would somewhat prefer that we fix the unit test to not call the API rather than remove logging, and I have a slight concern that this extra method will live longer than we want it to - but I don't currently feel strongly enough to block merging this.

ScottSuarez:
> I would somewhat prefer that we fix the unit test to not call the API rather than remove logging, and I have a slight concern that this extra method will live longer than we want it to - but I don't currently feel strongly enough to block merging this.

Okay, I changed the testcase instead. Note that this will mean the delete branch operation might be a bit noisy in cloud build.

for-examples:
https://api.github.com/repos/hashicorp/terraform-provider-google-beta/commits/8c6e61bb62d52c950008340deafc1e2a2041898a

melinath (approved):
ack re: noise.

PR #13784, opened 2025-04-25
Make capacity immutable in google_network_security_address_group

rosmo:
Make `capacity` immutable for `google_network_security_address_group` since it cannot be updated in-place (in-place update says `googleapi: Error 400: capacity can't be changed: invalid argument`).

```release-note:bug
networksecurity: marked google_network_security_address_group.capacity as immutable because it can't be updated in place.
```

### DIFF
--- a/mmv1/products/networksecurity/AddressGroup.yaml
+++ b/mmv1/products/networksecurity/AddressGroup.yaml
@@ -135,6 +135,7 @@ properties:
     description: |
       Capacity of the Address Group.
     required: true
+    immutable: true
   - name: 'purpose'
     type: Array
     description: |


rosmo:
@SirGitsalot I know you like big commits, but could we merge this little one?

SirGitsalot (commented):
Adding `immutable` isn't [a breaking change](https://googlecloudplatform.github.io/magic-modules/breaking-changes/breaking-changes/#field-level-breaking-changes) but it does have to be handled carefully because for Terraform `immutable` doesn't mean "can't be changed" (yeah, I know...) it means "delete then recreate". For some resources that's not acceptable, as it can mean data loss (not for this resource, it looks like) or a period of unavailability. It's the latter I'm concerned about:

- What are the consequences if the address group disappears for a while? I don't know if it's the case here, but in some security contexts that could mean a window of vulnerability
- What are the consequences if the address group is deleted *but is never recreated*?  Say, if the new value for `capacity` is invalid? in that case, the group will be deleted, but fail on the recreate.

rosmo:
The first concern is definitely a concern. I need to give that a test, but making the `capacity` immutable is *more correct* in my opinion than not having it immutable (what would be the alternative?).

SirGitsalot:
Adding @duvni as a reviewer from Network Security team and who is also familiar with TPG for their take.

duvni:
Sorry, missed the notification on this review.
This LGTM, I'm going to run this by @mihhalj from the relevant dev team as well before approving.

mihhalj:
LGTM

PR #13792, opened 2025-04-25
Add knowledge_connector_settings to dialogflow_cx_flow/page

ArtoriaRen:
Add `knowledge_connector_settings` to `dialogflow_cx_flow` and `dialogflow_cx_page` resources.

```release-note:enhancement
dialogflowcx: added `knowledge_connector_settings` field to `google_dialogflow_cx_flow` and `google_dialogflow_cx_page` resources
```

### DIFF
--- a/mmv1/third_party/terraform/services/dialogflowcx/resource_dialogflowcx_flow_test.go
+++ b/mmv1/third_party/terraform/services/dialogflowcx/resource_dialogflowcx_flow_test.go
@@ -375,25 +375,8 @@ func testAccDialogflowCXFlow_full(context map[string]interface{}) string {
           }
         }
       }
-      data_store_connections {
-        data_store_type = "UNSTRUCTURED"
-        data_store = "projects/${data.google_project.project.number}/locations/${google_dialogflow_cx_agent.agent_entity.location}/collections/default_collection/dataStores/${google_discovery_engine_data_store.my_datastore.data_store_id}"
-        document_processing_mode = "DOCUMENTS"
-      }
       target_flow = google_dialogflow_cx_agent.agent_entity.start_flow
     }
-  }
-
-  resource "google_discovery_engine_data_store" "my_datastore" {
-    location          = "global"
-    data_store_id     = "datastore-flow-update%{random_suffix}"
-    display_name      = "datastore-flow-update"
-    industry_vertical = "GENERIC"
-    content_config    = "NO_CONTENT"
-    solution_types    = ["SOLUTION_TYPE_CHAT"]
-  }
-
-  data "google_project" "project" {
   }
 `, context)
 }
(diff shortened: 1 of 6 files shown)


mmv1/third_party/terraform/services/dialogflowcx/resource_dialogflowcx_flow_test.go
```diff
@@ -364,6 +364,40 @@ func testAccDialogflowCXFlow_full(context map[string]interface{}) string {
...
+        data_store_type = "UNSTRUCTURED"
+        data_store = "projects/${data.google_project.project.number}/locations/${google_dialogflow_cx_agent.agent_entity.location}/collections/default_collection/dataStores/datastore-flow-update"
+        document_processing_mode = "DOCUMENTS"
+      }
+      target_page = google_dialogflow_cx_page.my_page.id
+    }
+  }
+
+  resource "google_dialogflow_cx_page" "my_page" {
+    parent       = google_dialogflow_cx_agent.agent_entity.my_flow
```
NickElliot, line 388:
```suggestion
    parent       = google_dialogflow_cx_agent.my_flow.agent_entity
```

NickElliot (changes requested):
>   Error: Unsupported attribute
        
          on terraform_plugin_test.tf line 318, in resource "google_dialogflow_cx_page" "my_page":
         318:     parent       = google_dialogflow_cx_agent.agent_entity.my_flow


The reference in the test is structured in the wrong order, can you double check and rerun the test locally to verify?

NickElliot (commented):
mostly looks good to me but I wanted to followup on a pretty wide reaching thing. most of these fields are optional, but I wanted to verify whether or not some should be flagged as `default_from_api`. are fields like `speech_settings.endpointer_sensitivity` truly optional i.e. there is an explicit meaning in excluding them/unsetting them? If they have a default value other than 0, they should probably be adjusted.

ArtoriaRen:
> Most of these fields are optional, but I wanted to verify whether or not some should be flagged as `default_from_api`. are fields like `speech_settings.endpointer_sensitivity` truly optional i.e. there is an explicit meaning in excluding them/unsetting them? If they have a default value other than 0, they should probably be adjusted.

I tried setting the speechSetting like this in a manual test:
```
        speech_settings {
          no_speech_timeout             = "3.500s"
          use_timeout_based_endpointing = true
        }
```
Note that `endpointer_sensitivity` is not set here.  Then I run `terraform apply` followed by `terraform plan`. The  `terraform plan` returned `No changes`. I also downloaded the agent locally and confirmed that  `endpointer_sensitivity`  was not set. So I think this field is optional.

But I agree that if some fields have a custom default value, we should flag it as `default_from_api`. I'll keep this in mind and create another PR to update such fields if I notice any as I use `dialogflowcx` resources in the next few months .

ArtoriaRen:
Oh I just tried removing the `speech_settings`. Now `terraform plan` shows `- endpointer_sensitivity        = 0 -> null`, which confirms the default value is 0.

NickElliot (approved):
in that case LGTM

PR #13793, opened 2025-04-25
Add startup probe support to Cloud Run v2 job resource

yanweiguo:

```release-note:enhancement
cloudrunv2: added `startup_probe` field to `google_cloud_run_v2_service` resource.
```


NickElliot (approved):
LGTM!

mmv1/products/cloudrunv2/Job.yaml
```diff
@@ -430,6 +430,100 @@ properties:
...
+                  type: NestedObject
+                  description: |-
+                    Startup probe of application within the container. All other probes are disabled if a startup probe is provided, until it succeeds. Container will not be added to service endpoints if the probe fails. More info: https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle#container-probes
+                  default_from_api: true
+                  properties:
+                    - name: 'initialDelaySeconds'
+                      type: Integer
+                      description: |-
+                        Number of seconds after the container has started before the probe is initiated. Defaults to 0 seconds. Minimum value is 0. Maximum value is 240. More info: https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle#container-probes
+                      default_value: 0
```
NickElliot, line 443: example line

NickElliot (changes requested):
sorry for asking changes after a delay

do these need to be a Terraform based default? given the block itself is `default_from_api`, it seems to me there is not an actual meaning in un-setting these fields, and the subfields with integer values should also be `default_from_api`

yanweiguo:
> sorry for asking changes after a delay
> 
> do these need to be a Terraform based default? given the block itself is `default_from_api`, it seems to me there is not an actual meaning in un-setting these fields, and the subfields with integer values should also be `default_from_api`

Good question. I copied the blocker from https://github.com/GoogleCloudPlatform/magic-modules/blob/3ae602271284d50261c8a7a1295dcfbe2b0f2249/mmv1/products/cloudrunv2/Service.yaml#L689

For Service, the whole `startupProbe` will have a default value if not specified. If some field in it is set, other fields would get a default value different than the ones when the whole `startupProbe` is not specified.

For Job, it is different. The fields in `startupProbe` will only have a default value when some fields in it are set in client side. In this case, I think removing `default_from_api` for `startupProbe` field is the preferred way?

NickElliot (commented):
could you go ahead and try to test the resource locally to see how it behaves with default_from_api shuffled around to see what best reflects the API behavior?

yanweiguo:
Closed in favor of https://github.com/GoogleCloudPlatform/magic-modules/pull/13838

PR #13806, opened 2025-04-28
Add a license check for new files

iyabchen:
https://github.com/hashicorp/terraform-provider-google/issues/20127
- Check license type (using a go-licenses library)
- Check copyright year (using strings.Contains)

Tested in https://github.com/GoogleCloudPlatform/magic-modules/pull/13805.

```release-note:none

```


shuyama1 (commented):
Hey @melinath, just want to confirm with you if this change addresses what's asked in https://github.com/hashicorp/terraform-provider-google/issues/20127 

This license text validation is currently implemented to check files in the mmv1 folder. We'd like to confirm if this is the desired scope, or should it be broader?

Also, the ticket description mentions "We should automate validation of its presence and correctness". Could you please clarify what specific aspects of `correctness` this should be validating?

melinath:
Correctness:

- Is the license text being added the correct license text?
- Does it have the current year (if it's new)

Which files: Basically, any files that correspond to other files that currently have it. MMv1 yaml, definitely. tpgtools we can ignore for now because it's not getting new stuff added. I'm not sure if there are other files.

iyabchen:
The current code is checking only the mmv1 folder for .tmpl, .go, .yaml files, and only for new files. If checking for just modified files, then many files will need a touch for the copyright year.

tools/license-check/cmd/root.go
```diff
@@ -0,0 +1,105 @@
...
+	return nil
+}
+
+func checkCopyright(filePath string, year int) error {
+	b, err := os.ReadFile(filePath)
+	if err != nil {
+		return err
+	}
+	expectStr := fmt.Sprintf("Copyright %d Google", year)
+	if !strings.Contains(strings.ToLower(string(b)), strings.ToLower(expectStr)) {
```
shuyama1: Would it be easy if we just have an explicit error message for incorrect year? I assume this is a common error people make when copying and pasting from old files.

shuyama1 (commented):
Sorry for the delay. Missed this in my review list somehow. Mostly LGTM! Thank you!

tools/license-check/cmd/root.go
iyabchen: Done. Changed to use regular expression to match and find substring.

iyabchen:
Rebase is done, the pr size is still large probably it considers the testdata files.

PR #13834, opened 2025-05-01
Corrected immutability docs

melinath:
Corrected these docs to match the actual behavior per IsForceNew, and split that logic apart so that it's easier to understand the choices we're making at each stage (instead of having a single massive && || construct.)

```release-note:none

```


melinath:

trodge (approved):

PR #13836, opened 2025-05-01
Removed post-create setting of identity fields that aren't computed or in id_format

melinath:
This is a follow-up to https://github.com/GoogleCloudPlatform/magic-modules/pull/13619 (and part of https://github.com/hashicorp/terraform-provider-google/issues/22214). It removes the logic to set "identity" fields on post-create _unless_ they're also in id_format and computed.

If a field is not in id_format, we do not consider it necessary to set on post-create even if it's computed.

If a field is not computed, it doesn't need to be set on post-create because it will already be set (or not set) on the config - the value will not change based on the API response. (Generally speaking, fields that are part of id_format and not computed should be required.)

The failure scenario if there were an issue with this code is that the read after create would fail.

```release-note:none

```

### DIFF
--- a/mmv1/templates/terraform/resource.go.tmpl
+++ b/mmv1/templates/terraform/resource.go.tmpl
@@ -309,7 +309,7 @@ func resource{{ $.ResourceName -}}Create(d *schema.ResourceData, meta interface{
 
 {{if and $.GetAsync ($.GetAsync.Allow "Create") -}}
 {{  if ($.GetAsync.IsA "OpAsync") -}}
-{{    if and $.GetAsync.Result.ResourceInsideResponse (or $.GetIdentity $.HasComputedIdFormatFields) -}}
+{{    if and $.GetAsync.Result.ResourceInsideResponse $.HasComputedIdFormatFields -}}
     // Use the resource in the operation response to populate
     // identity fields and d.Id() before read
     var opRes map[string]interface{}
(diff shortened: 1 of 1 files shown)


melinath:
- TestAccDataSourceGoogleGkeHubFeature_basic is not related because this PR doesn't modify datasources.
- TestAccCloudbuildWorkerPool_basic is not related because this PR doesn't modify that resource.
- TestAccHealthcarePipelineJob_healthcarePipelineJobBackfillExample is unrelated - failing in the same way in nightlies. https://github.com/hashicorp/terraform-provider-google/issues/21294 https://hashicorp.teamcity.com/buildConfiguration/TerraformProviders_GoogleCloud_GOOGLE_BETA_NIGHTLYTESTS_GOOGLEBETA_PACKAGE_HEALTHCARE/378664?buildTab=tests&status=failed&expandedTest=build%3A%28id%3A378664%29%2Cid%3A2000000034

c2thorn (approved):

melinath:
It turns out that there actually are a few fields that are optional and in id_format, but they are all either _only_ in id_format (and don't have a change in behavior from this PR) or are actually required by the API. There's one that should be optional + computed, but we use d.Set in the encoder so it's not a problem in this context (and the behavior wasn't changed by this PR.)

CloudTasks/Queue: name - should be required
Spanner/BackupSchedule: name - generates an ID client-side in the encoder. Should be optional + computed.
VertexAI/FeatureGroup: name - name is required by API
Notebooks/Location: name - deprecated, not functional
Apigee/EnvKeystore: name - required by API
Compute/GlobalNetworkEndpoint: ipAddress - idformat only
Compute/GlobalNetworkEndpoint: fqdn - idformat only
Compute/ManagedSslCertificate: name - should be required
Compute/NetworkEndpoint: instance - idformat only
Compute/NetworkEndpoint: port - idformat only
Compute/NodeGroup: name - required by API
Compute/NodeTemplate: name - required by API
Compute/RegionNetworkEndpoint: ipAddress - idformat only
Compute/RegionNetworkEndpoint: fqdn - idformat only

PR #13127, opened 2025-02-19
remove or fix various sweepers

ScottSuarez:
**Release Note Template for Downstream PRs (will be copied)**

See [Write release notes](https://googlecloudplatform.github.io/magic-modules/contribute/release-notes/) for guidance.

```release-note:none

```

PR #13842, opened 2025-05-02
feat: `google_iam_workforce_pool_provider_key` resource creation

ramonvermeulen:
Closes https://github.com/hashicorp/terraform-provider-google/issues/22082

```release-note:new-resource
`google_iam_workforce_pool_provider_key`
```


ramonvermeulen:
First time doing a "whole resource" from scratch.

Acceptance tests pass locally, only thing I noticed is that the API spec has the values `KEY_USE_UNSPECIFIED` and `KEY_SPEC_UNSPECIFIED` for [KeyUse](https://cloud.google.com/iam/docs/reference/rest/v1/locations.workforcePools.providers.keys#KeyUse) and [KeySpec](https://cloud.google.com/iam/docs/reference/rest/v1/KeyData#KeySpec). However, when you provide them in a POST to the API it actually gives an error of "invalid value", so therefore I didn't add them to the `WorkforcePoolProviderKey.yaml` file.

mmv1/products/iamworkforcepool/WorkforcePoolProviderKey.yaml
```diff
@@ -0,0 +1,166 @@
...
+      Immutable. Public half of the asymmetric key.
+    required: true
+    properties:
+      - name: 'format'
+        type: Enum
+        description: |
+          Output only. The format of the key.
+        output: true
+        enum_values:
+          - 'KEY_FORMAT_UNSPECIFIED'
```
BBBmau, line 103: typically don't include this as possible enum values to use. it's also brought up in the [documentation](https://cloud.google.com/iam/docs/reference/rest/v1/KeyData#keyformat) of the field as well:

image

mmv1/products/iamworkforcepool/WorkforcePoolProviderKey.yaml
```diff
@@ -0,0 +1,166 @@
...
+          - 'RSA_X509_PEM'
+      - name: 'notBeforeTime'
+        type: String
+        description: |
+          Output only. Earliest timestamp when this key is valid. Attempts to use this key before this time will fail.
```
BBBmau, line 108: no need to include `output only` in description as it does this automatically when provider is generated. Be sure to address this on all fields. The same applies to `required: true`

mmv1/products/iamworkforcepool/WorkforcePoolProviderKey.yaml
```diff
@@ -0,0 +1,166 @@
...
+        required: true
+        enum_values:
+          - 'RSA_2048'
+          - 'RSA_3072'
+          - 'RSA_4096'
+  - name: 'state'
+    type: Enum
+    description: |
+      Output only. The state of the key.
+      * STATE_UNSPECIFIED: State unspecified.
```
BBBmau, line 143: there's also no need to specify the enum values in the description. Generation does this also, see an example in [PubSub.yaml](https://github.com/GoogleCloudPlatform/magic-modules/blob/a100501b1a8d2a0c27e0fcdf8512490548c34ad3/mmv1/products/pubsub/Topic.yaml#L166-L173) with the generated docs being found [here](https://registry.terraform.io/providers/hashicorp/google/6.34.0/docs/resources/pubsub_topic#encoding-1)

BBBmau (changes requested):

ramonvermeulen:
@BBBmau 

Thanks for your review! Just applied your feedback.

mmv1/products/iamworkforcepool/WorkforcePoolProviderKey.yaml
```diff
@@ -0,0 +1,159 @@
...
+  - name: 'iam_workforce_pool_provider_saml_key_basic'
+    primary_resource_id: "example"
+    vars:
+      workforce_pool_id: 'example-pool'
+      provider_id: 'example-prvdr'
+      key_id: 'example-key'
+    test_env_vars:
+      org_id: 'ORG_ID'
+parameters:
+properties:
```
BBBmau, line 54: you typically would include the fields that are used in the URL as children of `parameters`.

So fields such as workforcePoolId, location, etc. should be under parameters (refer to [docs that mentions what parameters is meant for](https://googlecloudplatform.github.io/magic-modules/reference/resource/#parameters))

ramonvermeulen: Thanks, good catch! Will update the PR.

mmv1/products/iamworkforcepool/WorkforcePoolProviderKey.yaml
```diff
@@ -0,0 +1,159 @@
...
+self_link: 'locations/{{location}}/workforcePools/{{workforce_pool_id}}/providers/{{provider_id}}/keys/{{key_id}}'
+create_url: 'locations/{{location}}/workforcePools/{{workforce_pool_id}}/providers/{{provider_id}}/keys?workforcePoolProviderKeyId={{key_id}}'
+immutable: true
+import_format:
+  - 'locations/{{location}}/workforcePools/{{workforce_pool_id}}/providers/{{provider_id}}/keys/{{key_id}}'
+timeouts:
+  insert_minutes: 20
+  delete_minutes: 20
+autogen_async: true
+exclude_sweeper: true
```
BBBmau, line 34: this all looks good but could you include a comment here explaining the reasoning for excluding this resource from the sweeper?

ramonvermeulen: I accidentally copied this over from the `WorkForcePoolProvider.yaml` when creating the resource. 

I am not sure why that one actually has the sweeper disabled, seems to be disabled with this PR:
https://github.com/GoogleCloudPlatform/magic-modules/pull/13127

I'll remove the line for this yaml.

BBBmau (approved):
this all looks good, thanks for working on this!

PR #13845, opened 2025-05-03
[WIP] Ansible code generation

thekad:
I am trying to bring back ansible code generation to this repo. I have looked at the state the old ruby compiler was before it was rewritten in Go and I made a few changes:

* Modified the current compiler to add another (ansible) provider
* Adjusted the existing template data file to account for this change
* Added a new `-no-tests` flag to the compiler (not really needed, just helped me during my short hacking session)
* Added some helper methods for Ansible in template rendering code
* Added a few templates just to test out this new provider

```release-note:none
```


thekad:
Hi :wave: I am a Red Hat engineer and am trying to open up the conversation with this pull request, before I continue any further I'd like to get input from current developers and see if this is going in the right direction or not.
At this point my plan would be to continue porting the old ruby templates to Go templates to generate quasi-identical python code, but I am unsure if I am missing obvious stuff.
As far as I can tell, the old ruby compiler had override hooks in place and I saw a bunch of overrides for TF (and still do with the current Go compiler) but I couldn't find any special overrides for ansible from back then.
Do you see this going in the right direction?

TIA

thekad:
Rebased

rileykarson:
@modular-magician reassign-reviewer rileykarson

PR #3196, opened 2020-03-02
Initial retry transport

emilymye:
At this stage, this only adds a retry transport but doesn't change the retry loops outside of the transport (i.e. those wrapping httpClient.Do(req) in `sendRequest()` or golangClient.Request().Do()). 

I think removal of retry-loop work needs to be done more carefully, in case we end up changing the behavior, but we should hopefully move to having sendRequest/resource.Retry only send API/resource specific errorRetryPredicates. 

**Release Note Template for Downstream PRs (will be copied)**

```release-note:enhancement
provider: Added provider-wide request retries for common temporary GCP error codes and network errors
```

PR #13850, opened 2025-05-05
Removed goto label

melinath:
This was part of the initial retry transport implementation https://github.com/GoogleCloudPlatform/magic-modules/pull/3196 but as far as I can tell was never used

https://go.dev/ref/spec#Goto_statements

```release-note:none

```


melinath:
Never mind, this is used as a break label: https://go.dev/ref/spec#Break_statements

PR #13851, opened 2025-05-05
feat: implementation of spanner `instance_type` to be able to provision a `FREE_INSTANCE` via terraform

ramonvermeulen:
Solves https://github.com/hashicorp/terraform-provider-google/issues/22596

feat: implementation of spanner `instance_type` to be able to provision a `FREE_INSTANCE` via terraform
* had to implement a custom version of `exactly_one_of` because when provisioning a spanner instance with instance_type = `FREE_INSTANCE` neither one of `num_nodes` or `processing_units` or `autoscaling_config` is allowed to be configured.

**Added / direct affected tests pass locally:**

image

Couldn't run all tests locally with `-run=TestAccSpannerInstance_` because I ran into some org / quota policies on my sandbox project.

**Another thing to keep in mind, that makes testing a bit harder:**

image

E.g.: tests always have to run on a "new" GCP project, not sure if that is solved within the CI?

```release-note:enhancement
spanner: added field `instance_type` to the `google_spanner_instance` resource
```

### DIFF
--- a/mmv1/products/spanner/Instance.yaml
+++ b/mmv1/products/spanner/Instance.yaml
@@ -112,6 +112,11 @@ properties:
       autoscaling_config must be present in terraform except when instance_type = FREE_INSTANCE.
     api_name: nodeCount
     default_from_api: true
+    at_least_one_of:
+      - 'num_nodes'
+      - 'processing_units'
+      - 'autoscaling_config'
+      - 'instance_type'
     conflicts:
       - 'processing_units'
       - 'autoscaling_config'
@@ -121,6 +126,11 @@ properties:
       The number of processing units allocated to this instance. Exactly one of either num_nodes,
       processing_units or autoscaling_config must be present in terraform except when instance_type = FREE_INSTANCE.
     default_from_api: true
+    at_least_one_of:
+      - 'num_nodes'
+      - 'processing_units'
+      - 'autoscaling_config'
+      - 'instance_type'
     conflicts:
       - 'num_nodes'
       - 'autoscaling_config'
@@ -146,6 +156,11 @@ properties:
       When autoscaling is enabled, num_nodes and processing_units are treated as,
       OUTPUT_ONLY fields and reflect the current compute capacity allocated to
       the instance.
+    at_least_one_of:
+      - 'num_nodes'
+      - 'processing_units'
+      - 'autoscaling_config'
+      - 'instance_type'
     conflicts:
       - 'num_nodes'
       - 'processing_units'
@@ -267,6 +282,11 @@ properties:
       usage restrictions, quotas and billing. Currently this is used to distinguish FREE_INSTANCE vs PROVISIONED instances.
       When configured as FREE_INSTANCE, the field `edition` should not be configured.
     default_from_api: true
+    at_least_one_of:
+      - 'num_nodes'
+      - 'processing_units'
+      - 'autoscaling_config'
+      - 'instance_type'
     enum_values:
       - 'PROVISIONED'
       - 'FREE_INSTANCE'
--- a/mmv1/products/spanner/Instance.yaml
+++ b/mmv1/products/spanner/Instance.yaml
@@ -112,8 +112,7 @@ properties:
       must be present in terraform.
     api_name: nodeCount
     default_from_api: true
-    exactly_one_of:
-      - 'num_nodes'
+    conflicts:
       - 'processing_units'
       - 'autoscaling_config'
   - name: 'processingUnits'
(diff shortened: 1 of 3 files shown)


mmv1/templates/terraform/encoders/spanner_instance.go.tmpl
```diff
@@ -1,3 +1,22 @@
+if obj["instanceType"] == nil || obj["instanceType"] != nil && obj["instanceType"] == "PROVISIONED" {
```
slevenick, line 1:
```suggestion
if obj["instanceType"] == nil ||  obj["instanceType"] == "PROVISIONED" {
```

I believe this is equivalent?

mmv1/templates/terraform/encoders/spanner_instance.go.tmpl
```diff
@@ -1,3 +1,22 @@
...
+	}
+	if obj["nodeCount"] != nil {
+		count++
+	}
+	if obj["processingUnits"] != nil {
+		count++
+	}
+
+	if count != 1 {
+		return nil, fmt.Errorf("Exactly one of `autoscaling_config`, `num_nodes`, or `processing_units` must be specified")
```
slevenick, line 16: I don't think we want to throw this type of error here. exactly_one_of should present a plan-time error, but this is a runtime error.

We may need to remove the exactly_one_of, or replace it with at_least_one_of and include instance_type

mmv1/templates/terraform/encoders/spanner_instance.go.tmpl
```diff
@@ -12,5 +31,11 @@ if obj["name"] == nil {
 } else {
 	newObj["instanceId"] = obj["name"]
 }
+if obj["instanceType"] != nil && obj["instanceType"] == "FREE_INSTANCE" {
+	// when provisioning a FREE_INSTANCE, the following fields cannot be specified
```
slevenick, line 35: We should probably throw an error here rather than ignoring the user's value for these fields. If they set them it will cause a permadiff

slevenick (changes requested):

mmv1/templates/terraform/encoders/spanner_instance.go.tmpl
ramonvermeulen: So the problem I ran into is that `exactly_one_of` that was already configured for `num_nodes`, `processing_units` and `autoscaling_config`. However, this now conflicts with `instance_type` == `FREE_INSTANCE`, e.g. with FREE_INSTANCE none of these fields are allowed to be configured, while with PROVISIONED instance exactly one of these fields should be configured. So my thought was to do a custom implementation in the encoder to solve this issue.

What would be your suggestion to solve this problem without causing a runtime error.

slevenick: I would probably move to at_least_one_of on all 4 of the fields. It won't be the most restrictive that we could be, but it will accomplish most of the goals at plan time. Most users will continue not setting `instance_type`, and instead have to specify one of the other fields

mmv1/products/spanner/Instance.yaml
```diff
@@ -261,6 +258,16 @@ properties:
       - 'STANDARD'
       - 'ENTERPRISE'
       - 'ENTERPRISE_PLUS'
+  - name: 'instanceType'
+    type: Enum
+    description: |
```
slevenick, line 263: Can you add the restrictions on not setting `processing_units` and the other values here? Moving that to the description will increase visibility if we remove the validation

mmv1/templates/terraform/encoders/spanner_instance.go.tmpl
ramonvermeulen: Just applied this, but thinking about it now this doesn't solve the problem (test also failed). Because `at_least_one_of` requires that either `num_nodes`, `processing_units` or `autoscaling_config` is set. For a `FREE_INSTANCE` none of these fields can be set.

slevenick: I would add `instance_type` to the `at_least_one_of`. So a user can set instance_type = FREE instead of setting one of the other values.

Most people won't specify instance_type = PROVISIONED as examples won't specify, so the at_least_one_of will mostly work as intended

ramonvermeulen: Hmm I see, only weird scenario is then indeed when a user configures only `instance_type=PROVISIONED` (although not in any of the examples) on the api endpoint (during apply) you will probably still get an error even though the plan succeeds. But that is in this case the trade-off I guess.

slevenick (approved):
Thanks!

PR #13852, opened 2025-05-05
Tweaked resource version guard docs to be clearer

melinath:

```release-note:none

```

### DIFF
--- a/docs/content/develop/add-resource.md
+++ b/docs/content/develop/add-resource.md
@@ -163,9 +163,8 @@ For more information about types of resources and the generation process overall
    - Replace all occurrences of `github.com/hashicorp/terraform-provider-google-beta/google-beta` with `github.com/hashicorp/terraform-provider-google/google`
    - Remove the `Example` suffix from all test function names.
    - Remove the comments at the top of the file.
-   - If beta-only fields are being tested, do the following:
-     - Change the file suffix to `.go.tmpl`
-     - Wrap each beta-only test in a separate version guard: `{{- if ne $.TargetVersionName "ga" -}}...{{- else }}...{{- end }}`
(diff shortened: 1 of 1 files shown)


melinath:

docs/content/develop/add-resource.md
```diff
@@ -163,9 +163,8 @@ For more information about types of resources and the generation process overall
    - Replace all occurrences of `github.com/hashicorp/terraform-provider-google-beta/google-beta` with `github.com/hashicorp/terraform-provider-google/google`
    - Remove the `Example` suffix from all test function names.
    - Remove the comments at the top of the file.
-   - If beta-only fields are being tested, do the following:
-     - Change the file suffix to `.go.tmpl`
-     - Wrap each beta-only test in a separate version guard: `{{- if ne $.TargetVersionName "ga" -}}...{{- else }}...{{- end }}`
+   - If any of the added Go code (including any imports) is beta-only, change the file suffix to `.go.tmpl` and wrap the beta-only code in a version guard: `{{- if ne $.TargetVersionName "ga" -}}...{{- else }}...{{- end }}`.
+     - If the whole resource is beta-only, wrap everything except package declarations. Otherwise, wrap each logically-related block of code in a version guard (field, test, etc) rather than grouping adjacent version-guarded sections - it's easier to read and easier to modify as things move out of beta.
```
c2thorn, line 167: a bit unclear to me what this suggestion is. is this suggesting to version wrap entire functions instead of pieces of it? or to not move code around into single giant version guards?

melinath: The thing I'm trying to get across is that if for example the resource overall is GA and there are multiple fields that are beta only, you should have individual version guards for each field (and for each field's expander/flattener) rather than a version guard that holds multiple fields / functions.

The way this is phrased on the field docs is: `Add a new guard rather than adding the field to an existing guard; it is easier to read.`

PR #13860, opened 2025-05-06
Added immutability support to Internal Ranges in Terraform.

dorianverna:
This PR aims to modify the Internal Range Terraform support as follows:

Add a new field available to the Internal Range resource called immutability.
Add a template file showing how the feature is used.
Added acceptance test for the newly added feature.

```release-note:enhancement
networkconnectivity: added `immutability` field to `google_network_connectivity_internal_range` resource
```

+CC: @sebkalis


mmv1/products/networkconnectivity/InternalRange.yaml
```diff
@@ -183,3 +183,8 @@ properties:
           may not exist yet.
           For example /projects/{project}/regions/{region}/subnetworks/{subnet}
         required: true
+  - name: 'immutable'
+    type: Boolean
+    description: |
+      Immutable ranges cannot have their fields modified, except for labels and description.
+    immutable: true
```
roaks3, line 190: Just to clarify: the meaning of this config is that the field cannot be changed without recreating the entire resource. Is that true, even if the user is trying to change from `immutable: false` to `immutable: true`?

mmv1/third_party/terraform/services/networkconnectivity/resource_network_connectivity_internal_range_test.go
```diff
@@ -337,3 +337,118 @@ resource "google_compute_network" "default" {
...
+}
+
+resource "google_compute_network" "default" {
+  name                    = "tf-test-internal-ranges%{random_suffix}"
+  auto_create_subnetworks = false
+}
+`, context)
+}
+
+func TestAccNetworkConnectivityInternalRange_networkConnectivityInternalRangesImmutableExample_update(t *testing.T) {
```
roaks3, line 399: I think you can completely remove this test and the template below it. Update tests are used to test that a field can be changed in-place, but since this field is immutable that can't happen.

roaks3 (approved):
Mainly need clarification on the behavior, but tentatively approving because it otherwise seems fine.

The other thing I noticed was that this field didn't seem to be present in the API docs. If this field is not yet released, then we will need to be sure that the timing works out (primarily that the provider release with this change goes out when the field is actually available). See go/terraform-releases#schedule.

mmv1/products/networkconnectivity/InternalRange.yaml
dorianverna: Yes, that is right, the user can only change the labels and description of the internal range if immutability is set to true. Also, this immutable field cannot be modified (just as you pointed out).

dorianverna:
> Mainly need clarification on the behavior, but tentatively approving because it otherwise seems fine.
> 
> The other thing I noticed was that this field didn't seem to be present in the API docs. If this field is not yet released, then we will need to be sure that the timing works out (primarily that the provider release with this change goes out when the field is actually available). See go/terraform-releases#schedule.

The change is present in the discovery docs, but it may not be present in the public documentation yet. We actually aim to have API, gcloud and Terraform functionalities in place before making it visible. Let me know if that is fine.

mmv1/third_party/terraform/services/networkconnectivity/resource_network_connectivity_internal_range_test.go
dorianverna: Done.

mmv1/templates/terraform/examples/network_connectivity_internal_ranges_immutability.tf.tmpl
```diff
@@ -0,0 +1,14 @@
+resource "google_network_connectivity_internal_range" "{{$.PrimaryResourceId}}" {
```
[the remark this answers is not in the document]
dorianverna: Done

roaks3 (approved):

PR #13863, opened 2025-05-06
Image error

hao-nan-li:
Fixes https://github.com/hashicorp/terraform-provider-google/issues/22566

```release-note:none

```


mmv1/third_party/terraform/envvar/envvar_utils.go
```diff
@@ -146,6 +150,10 @@ func GetTestUniverseDomainFromEnv(t *testing.T) string {
 	return transport_tpg.MultiEnvSearch(UniverseDomainEnvVars)
 }
 
+func GetProjectPrefixFromEnv() string {
```
rileykarson, line 153: Can you document context / why this needs to be used?

mmv1/third_party/terraform/provider/universe/universe_domain_compute_test.go
```diff
@@ -85,6 +105,23 @@ resource "google_compute_instance_template" "instance_template" {
...
+  universe_domain = "%s"
+}
+
+resource "google_compute_disk" "primary" {
+  name  = "async-test-disk"
+  type  = "pd-ssd"
+  zone  = "%s"
+
+  physical_block_size_bytes = 4096
+  image = "projects/%s:debian-cloud/global/images/debian-12"
```
rileykarson, line 120: Can we inject the full project name here as `prefixedProject` instead of `prefix`? For alternate universe that's `example.com:debian-cloud` instead of `example.com`, and that way we can supply `debian-cloud` for GCP. That'll let us remove the skip as the test will pass in GCP and other universes.

mmv1/third_party/terraform/envvar/envvar_utils.go
hao-nan-li: Will do

mmv1/third_party/terraform/provider/universe/universe_domain_compute_test.go
hao-nan-li: Yes, you are right about injecting the full project name here instead of the `prefix`. There's just a little problem for the image name `debian-12` which is not the actual name in different universe. (But looks like this should work in GCP)

rileykarson: As discussed offline, can we use `data.google_compute_images` to identify an image appropriate to the environment?

mmv1/third_party/terraform/provider/universe/universe_domain_compute_test.go
```diff
@@ -33,6 +33,33 @@ func TestAccUniverseDomainDisk(t *testing.T) {
 	})
 }
 
+func TestAccUniverseDomainDiskImage(t *testing.T) {
+	// Skip this test in all env since this can only run in specific test project.
+	// t.Skip()
```
rileykarson, line 38:
```suggestion
```

rileykarson (commented):
Hmm- the new test isn't running. Do we need to register something somewhere to have it get picked up?

hao-nan-li:
> Do we need to register something somewhere to have its package get picked up?

Looks like this test actually passed in https://storage.mtls.cloud.google.com/ci-vcr-logs/beta/refs/heads/auto-pr-13863/artifacts/7b3e92a9-bb66-47fb-a694-79ad3ff89ed0/build-log/replaying_test.log. For some reason it didn't show up on the UI

mmv1/third_party/terraform/provider/provider.go.tmpl
```diff
@@ -416,6 +416,13 @@ func ProviderConfigure(ctx context.Context, d *schema.ResourceData, p *schema.Pr
 		return nil, diag.FromErr(err)
 	}
 
+	// Bypass checkDestroyer failure with universe domain set
```
hao-nan-li, line 419: @c2thorn FYI, this is added to bypass check destroy error

hao-nan-li:
the unit test failure looks unrelated to this change, perhaps it's something needed to be fixed...

rileykarson:
FYI @ScottSuarez as this may be related to https://github.com/GoogleCloudPlatform/magic-modules/pull/13929

In the meantime can you rebase this PR on top of `main`? That could be a possible trigger.

mmv1/third_party/terraform/provider/provider.go.tmpl
rileykarson: Why does this work exactly? I don't think we're generally expected to call `d.Set` inside the provider as it's not a full resource, so I'm a bit surprised it even works.

c2thorn: If this works, this is the exact same as removing the check later in https://github.com/hashicorp/terraform-provider-google/blob/main/google/provider/provider.go#L1266. There's no point in having both.

I was hoping to either set the universe domain in a test-only setup function, or skip the check altogether by adding something to 
```
else if config.UniverseDomain != "" && config.UniverseDomain != "googleapis.com"
```
that verifies if it is a test run.

rileykarson: What is the issue we're seeing exactly? Does https://github.com/hao-nan-li/magic-modules/blob/2b5b82e44b188362545a5ca43a854c7397079d75/mmv1/third_party/terraform/provider/universe/universe_domain_compute_test.go#L145 not have the domain set?

hao-nan-li: Here's the flow: 
1) https://github.com/hashicorp/terraform-provider-google/blob/main/google/provider/universe/universe_domain_compute_test.go#L100
2) https://github.com/hashicorp/terraform-provider-google/blob/main/google/acctest/provider_test_utils.go#L45
3) The configure function then calls https://github.com/hashicorp/terraform-provider-google/blob/ff829fe23a4235e6566fd07be74c52e32e376b69/google/provider/provider.go#L921

Finally an error occurs because universe_domain is not set in d (schema.ResourceData). And the code right below will return an error, then the check destroyer function fails.

Adding this block of code works around this problem

rileykarson: Discussed offline; we can likely set the value on https://github.com/hashicorp/terraform-provider-google/blob/8ef7e5fbcf4f8d5a2a141dfff29bffc52d119e65/google/acctest/provider_test_utils.go#L44 to later successfully call configure.

More broadly, we should figure out how the two initialization paths (through the provider factory and there) are different, and ideally align them.

hao-nan-li: Applied fix in provider_test_utils.

Will take a look at the second suggestion.

mmv1/third_party/terraform/envvar/envvar_utils.go
```diff
@@ -146,6 +150,11 @@ func GetTestUniverseDomainFromEnv(t *testing.T) string {
 	return transport_tpg.MultiEnvSearch(UniverseDomainEnvVars)
 }
 
+// Project Prefix of different universes
+func GetProjectPrefixFromEnv() string {
```
rileykarson, line 154: In a followup PR could you change the env var to include "Universe" in the namespace? i.e. `GOOGLE_UNIVERSE_PROJECT_PREFIX` / `GetUniverseProjectPrefixFromEnv`

hao-nan-li: Will do

PR #13868, opened 2025-05-06
Print original resource address for tgc testing

zli82016:
`raw_config` will be used to check if the config changes or not compared to the previous day's config. Only fetch cai assets for the resources with changed configs. 

This PR will keep the original resource name, e.g. `foobar`, instead of the resource id, e.g. `tf-test-random-string`. The resource id has a random string, which causes the config always changing.

```release-note:none

```


zli82016:
Failed tests are unrelated.

hao-nan-li (approved):

PR #13871, opened 2025-05-07
fix: do not enforce recreate for `node_scaling_factor` on bigtable instances when using default value

ramonvermeulen:
Solves https://github.com/hashicorp/terraform-provider-google/issues/22617

```release-note:bug
bigtable: fixed an issue where setting the default value for `node_scaling_factor` in `google_bigtable_instance` would cause the resource to be recreated
```

### DIFF
--- a/mmv1/third_party/terraform/services/bigtable/resource_bigtable_instance.go
+++ b/mmv1/third_party/terraform/services/bigtable/resource_bigtable_instance.go
@@ -155,7 +155,7 @@ func ResourceBigtableInstance() *schema.Resource {
 						"node_scaling_factor": {
 							Type:         schema.TypeString,
 							Optional:     true,
-							ForceNew:     false,
+							ForceNew:     true,
 							Default:      "NodeScalingFactor1X",
 							ValidateFunc: validation.StringInSlice([]string{"NodeScalingFactor1X", "NodeScalingFactor2X"}, false),
 							Description:  `The node scaling factor of this cluster. One of "NodeScalingFactor1X" or "NodeScalingFactor2X". Defaults to "NodeScalingFactor1X".`,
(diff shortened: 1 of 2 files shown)


ramonvermeulen:
Added boilerplate for a test that should succeed, and logging in `CustomizeDiff` to further investigate issue when I have time.

rjmco:
@ramonvermeulen thanks for this. I added some details from our setup on https://github.com/hashicorp/terraform-provider-google/issues/22617#issuecomment-2860352895 , I am not sure it will be helpful with the tests. I hope it does.

ramonvermeulen:
Struggling to write tests, please see discussion for more information: https://github.com/hashicorp/terraform-provider-google/issues/22617#issuecomment-2861926438

The problem is that when I try to create a test case, the `nodeScalingFactor` in the API will be set to and returned as `NodeScalingFactor1X` anyways. So I really need to have an old instance (before the API change) to test it on, or somehow manipulate the response from the API via an interceptor if something like that is possible.

Would love to hear your ideas and learn if there are better ways for scenarios like this (e.g. API adds field with default, but doesn't return field/default for "old" instances).

shuyama1:
@hao-nan-li Passing this review to you, as you are already working with the service team on a resolution for the issue.

mmv1/third_party/terraform/services/bigtable/resource_bigtable_instance.go
```diff
@@ -695,6 +696,21 @@ func resourceBigtableInstanceUniqueClusterID(_ context.Context, diff *schema.Res
 	return nil
 }
 
+func resourceBigtableInstanceIgnoreNodeScalingFactorNullFromRecreate(_ context.Context, diff *schema.ResourceDiff, meta interface{}) error {
+	oldCount, _ := diff.GetChange("cluster.#")
+
+	for i := 0; i < oldCount.(int); i++ {
+		oldNodeScalingFactor, newNodeScalingFactor := diff.GetChange(fmt.Sprintf("cluster.%d.node_scaling_factor", i))
+		if oldNodeScalingFactor == "" && newNodeScalingFactor == "NodeScalingFactor1X" {
+			if err := diff.SetNew(fmt.Sprintf("cluster.%d.node_scaling_factor", i), oldNodeScalingFactor); err != nil {
```
zli82016, line 705: SetNew only works for computed and root field.

ramonvermeulen: Hmm I see, what would be the suggestion here to use instead? The goal is to prevent the plan to see a "difference" between node_scaling_factor being returned as `null` or as `NodeScalingFactor1X` (which newer resources have by default since the field was added to the API).

ramonvermeulen:
Issue is fixed with: https://github.com/GoogleCloudPlatform/magic-modules/pull/13902

PR #13878, opened 2025-05-07
Cloud DMS PrivateConnection support for create_without_validation

oferhandel-google:
Cloud DMS PrivateConnection support for create_without_validation for skipping overlapping ranges validation

```release-note:enhancement
databasemigrationservice: added `create_without_validation` field to `google_database_migration_service_private_connection` resource
```


oferhandel-google:
@BBBmau any chance to review this soon?

NickElliot (commented):
could you verify the behavior of this field on import? i.e. how does it behave when a private connection resource is imported to terraform, when that resource is not declaring this field

oferhandel-google:
> could you verify the behavior of this field on import? i.e. how does it behave when a private connection resource is imported to terraform, when that resource is not declaring this field

this field does not persist on the resource and therefore I believe it doesn't have any special meaning for import. Its very similar to the datastream_private_connection resource

NickElliot (changes requested):
test is failing, would say we just handle it like this

NickElliot (approved):
LGTM

PR #13879, opened 2025-05-07
Prevent storagecontrol target project from being swept

shuyama1:
Update the bootstrapped project prefix from `tf-test` to `tf-boot` to prevent the project from being deleted by the sweeper. Though the bootstrap function will [restore](https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/third_party/terraform/acctest/bootstrap_test_utils.go.tmpl#L995) the project during each run, [there might be a delay before a restored project can be linked to a billing account](https://cloud.google.com/resource-manager/docs/creating-managing-projects#restoring_a_project). This caused tests to fail when setting the billing account.

fixes https://github.com/hashicorp/terraform-provider-google/issues/22537, fixes https://github.com/hashicorp/terraform-provider-google/issues/22536

```release-note:none

```


rileykarson (approved):
Thanks! I used `tf-test-` on autopilot 🤦

PR #13889, opened 2025-05-08
Switched to only one bootstrapped code repository index to reduce quota issues

melinath:
I don't see a specific reason that this test needs to be using a separate bootstrapped instance, so I think we can save our limited quota by only using one.

```release-note:none

```


melinath:

SirGitsalot (approved):

melinath:
deleted the second one in VCR test project; manual test run: https://hashicorp.teamcity.com/buildConfiguration/TerraformProviders_GoogleCloud_GOOGLE_BETA_MMUPSTREAMTESTS_GOOGLEBETA_PACKAGE_GEMINI/373457

melinath:
The related tests seem to be passing in the manual test as well - moving forward with merge.

PR #13917, opened 2025-05-12
Deprecated bool_to_object custom expander

hao-nan-li:
fixes https://github.com/hashicorp/terraform-provider-google/issues/16626

```release-note:none
```


mmv1/products/billingbudget/Budget.yaml
```diff
@@ -412,7 +412,7 @@ properties:
           - 'amount.0.specified_amount'
           - 'amount.0.last_period_amount'
         custom_flatten: 'templates/terraform/custom_flatten/object_to_bool.go.tmpl'
-        custom_expand: 'templates/terraform/custom_expand/bool_to_object.go.tmpl'
+        custom_expand: 'templates/terraform/custom_expand/deprecated_bool_to_object.go.tmpl'
```
c2thorn, line 415: Thanks for picking this up. Just to really drive home that we should NOT use this going forward, would you mind adding a comment above all 4 current YAML uses, explaining that this is reserved for forward compatibility but should NOT copied from or replicated

then maybe another one at the top of the template itself?
The template one being a template comment like 
```
{{/* THIS TEMPLATE IS DEPRECATED, DO NOT USE FOR NEW FIELDS */}}
```

hao-nan-li: Thanks for pointing it out, just added the message.

mmv1/products/billingbudget/Budget.yaml
```diff
@@ -412,7 +412,8 @@ properties:
           - 'amount.0.specified_amount'
           - 'amount.0.last_period_amount'
         custom_flatten: 'templates/terraform/custom_flatten/object_to_bool.go.tmpl'
-        custom_expand: 'templates/terraform/custom_expand/bool_to_object.go.tmpl'
+        # {{/* THIS TEMPLATE IS DEPRECATED, DO NOT USE FOR NEW FIELDS */}}
```
c2thorn, line 415:
```suggestion
        # THIS TEMPLATE IS DEPRECATED, DO NOT USE FOR NEW FIELDS
```

c2thorn (approved):

PR #13918, opened 2025-05-12
Add `google_beyondcorp_security_gateway_application` resource.

Berro321:
```release-note:new-resource
`google_beyondcorp_security_gateway_application`
```


mmv1/third_party/terraform/services/beyondcorp/resource_beyondcorp_security_gateway_application_test.go
```diff
@@ -0,0 +1,85 @@
...
+  security_gateway_id = "default-sg%{random_suffix}"
+  display_name = "My Security Gateway resource"
+  hubs { region = "us-central1" }
+}
+
+resource "google_beyondcorp_security_gateway_application" "example" {
+  security_gateway_id = google_beyondcorp_security_gateway.default.security_gateway_id
+  display_name = "Updated Name"
+  application_id = "google-sga%{random_suffix}"
+  endpoint_matchers {
```
SirGitsalot, line 80: Please add some updates for `endpoint_matchers` and `upstreams` to expand the test.

mmv1/products/beyondcorp/SecurityGatewayApplication.yaml
```diff
@@ -0,0 +1,151 @@
...
+  - name: createTime
+    type: String
+    description: Output only. Timestamp when the resource was created.
+    output: true
+  - name: displayName
+    type: String
+    description: |-
+      Optional. An arbitrary user-provided name for the Application resource.
+      Cannot exceed 64 characters.
+  - name: endpointMatchers
```
SirGitsalot, line 90: Does the API do any kind of normalization or reordering of `endpointMatchers` or `upstreams`? I ask because hostnames are the kind of string that is often normalized, e.g., `EXAMPLE.COM` -> `example.com` and arrays often get reordered. If either is the case, it will cause a [permadiff](https://googlecloudplatform.github.io/magic-modules/develop/diffs/)

SirGitsalot (changes requested):

mmv1/products/beyondcorp/SecurityGatewayApplication.yaml
Berro321: They do not. They keep it as the customer configured them originally.

Berro321:
@SirGitsalot Is the multiple-resources error because of the IAM resources also contained in this change? If so, can we still submit this PR or is there something else needed from me?

SirGitsalot:
Yeah, it's the IAM resources but that's fine - I've added the necessary label to override the check.

mmv1/products/beyondcorp/SecurityGatewayApplication.yaml
```diff
@@ -0,0 +1,151 @@
...
+        - name: hostname
+          type: String
+          description: Required. Hostname of the application.
+          required: true
+        - name: ports
+          type: Array
+          description: Optional. Ports of the application.
+          item_type:
+            type: Integer
+  - name: upstreams
```
SirGitsalot, line 118: One last question: I just checked the service proto and `upstreams` has visibility restrictions on it - is the field usable by non-allowlisted external customers? And if not, is the resource usable without the `upstreams` field?

Berro321: it's visibility restricted for now but will be available to all customers within the next few weeks. The resource is still usable w/out the upstream field.

Berro321:
friendly ping for review

mmv1/products/beyondcorp/SecurityGatewayApplication.yaml
SirGitsalot: In that case let's remove `upstreams` from this PR and add in back in a future PR after the visibility restriction is removed and it's been rolled out for all customers.

SirGitsalot (changes requested):

NickElliot:
Hi going to drive-by for a moment here: because this resource isn't merged yet we're modifying the deprecation message on  `beyondcorp_application` https://github.com/GoogleCloudPlatform/magic-modules/pull/13991 to avoid suggesting using this (unreleased) resource for the release this week.

if you could restore the current deprecation message along with this PR, that would be great thanks!

SirGitsalot (approved):

PR #13921, opened 2025-05-13
`resourceIdentity`: add identity logic in `ParseImportId`

BBBmau:
This supports the use of the identity block within the `import` block.

The following example uses this same logic to import a service_account successfully:
image

```release-note:none

```


BBBmau:
moving into draft until the sync PR has been merged. a lot of the test failures come from failures that were present prior to the feature-branch being created.

mmv1/third_party/tgc/tfdata/fake_resource_data.go
```diff
@@ -92,6 +98,10 @@ func (d *FakeResourceData) Get(name string) interface{} {
 	return val
 }
 
+func (d *FakeResourceData) GetRawConfig() cty.Value {
+	return d.GetRawConfig()
```
zli82016, line 102: Does the function GetRawConfig call itself? Or I miss something.

mmv1/third_party/terraform/tpgresource/import.go
```diff
@@ -64,18 +65,46 @@ func ParseImportId(idRegexes []string, d TerraformResourceData, config *transpor
...
 
 			// The first id format is applied first and contains all the fields.
-			err := setDefaultValues(idRegexes[0], d, config)
+			err := setDefaultValues(idRegexes[0], nil, d, config)
 			if err != nil {
 				return err
 			}
 
 			return nil
+		} else if d.Id() == "" {
```
zli82016, line 74: We can move `if d.Id() == "" ` before line 30 to first check if d.Id()  is empty?

zli82016:
> This supports the use of the identity block within the `import` block.
> 
> The following example uses this same logic to import a service_account successfully: image
> 

`google_service_account` has id. The import still uses the old way?

BBBmau:
> > This supports the use of the identity block within the `import` block.
> > The following example uses this same logic to import a service_account successfully: image
> 
> `google_service_account` has [id](https://github.com/hashicorp/terraform-provider-google/blob/50b92ef5dec8df11b9eede61c8fc1c806d8fd0a4/google/services/resourcemanager/resource_google_service_account.go#L154). The import still uses the old way?

This is an example of how the importer works but the actual implementation for `google_service_account` will be done in a separate PR that will be opened once this PR gets merged

mmv1/third_party/tgc/tfdata/fake_resource_data.go
BBBmau: i've moved it to the list of mappers that are required but not needed or used in tgc

mmv1/third_party/terraform/tpgresource/resource_test_utils.go
```diff
@@ -35,6 +37,10 @@ func (d *ResourceDataMock) Get(key string) interface{} {
 	return v
 }
 
+func (d *ResourceDataMock) GetRawConfig() cty.Value {
+	return d.GetRawConfig()
```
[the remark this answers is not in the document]
BBBmau: I've updated the function however the reason we must implement this is for the ResourceDataMock to follow the interface of `ResourceData`. They aren't used however, the latest commit i've added but left a comment on why it's included and how we don't use or need it.

mmv1/third_party/terraform/tpgresource/import.go
BBBmau: this made sense to do, i've applied it in the latest commit

mmv1/third_party/tgc_next/pkg/tfplan2cai/models/fake_resource_data.go
```diff
@@ -107,6 +113,10 @@ func (d *FakeResourceData) GetOk(name string) (interface{}, bool) {
 	return r.Value, exists
 }
 
+func (d *FakeResourceData) GetRawConfig() cty.Value {
+	return d.GetRawConfig()
```
zli82016, line 117:
```suggestion
	return cty.NullVal(cty.String) 
```

zli82016 (approved):

PR #13928, opened 2025-05-13
transition functions to go-github

ScottSuarez:
I've transitioned all functions in .ci/magician to use go-github except for the merge function. We have special retry logic based on very specific errors. This will break if we bring it over to the new system.

I also changed all interfaces to go-github natives for consistency and eliminated one unneeded version of go-github
closes https://github.com/hashicorp/terraform-provider-google/issues/17191

```release-note:none

```


ScottSuarez:
@melinath I don't quite know how to test this. I can try to check every function individually but it's large change with no integration testing possible.

melinath:
yeah, this is hard to test. Part of that is that it's a lot of changes at once. Would it be possible to move over smaller pieces at a time? Then you could specifically test each piece on the PR that changes it, and there would also be less risk per PR of breaking things by accident.

PR #13929, opened 2025-05-13
validate template syntax needed

ScottSuarez:
closes https://github.com/hashicorp/terraform-provider-google/issues/21407
converted 147 templates to go code, condensed mmv1 unit tests into one github workflow. Added unit test to confirm no new templates that can be go code are being added.

```release-note:none

```


mmv1/validate_third_party_test.go
```diff
@@ -0,0 +1,87 @@
...
+			return err
+		}
+
+		// Skip directories
+		if info.IsDir() {
+			return nil
+		}
+
+		// Only check .go.tmpl files
+		if filepath.Ext(path) != ".tmpl" || !strings.HasSuffix(path, ".go.tmpl") {
```
c2thorn, line 46: is there a case that `filepath.Ext(path) != ".tmpl"` covers that isn't covered in `!strings.HasSuffix(path, ".go.tmpl")`?

c2thorn (approved):
LGTM, thanks for adding this!

.github/workflows/unit-test-mmv1.yml
```diff
@@ -1,13 +1,33 @@
...
 permissions: read-all
 
 on:
   pull_request:
     paths:
-      - 'mmv1/products/**'
+      - 'mmv1/**'
 
 jobs:
+  version-guard-check:
```
c2thorn, line 11: was this move just to consolidate similar checks?

ScottSuarez: correct!

mmv1/validate_third_party_test.go
ScottSuarez: just one

        The following 1 .go.tmpl files in third_party directory don't contain any template syntax and no longer need to be templates:
    validate_third_party_test.go:79:   - third_party/terraform/terraform-registry-manifest.json.tmpl

ScottSuarez: I've changed it to include all tmpl files instead of just .go.tmpl. Lmk what you think

PR #13938, opened 2025-05-13
Check example vars

slevenick:
Fixes: https://github.com/hashicorp/terraform-provider-google/issues/15679#event-17597385755

Check Vars and TestEnvVars to make sure that all variables used in examples have values specified in YAML configs. This does add significant file IO as we have to read every example file prior to generating tests.

```release-note:none

```


zli82016:
> Check Vars and TestEnvVars to make sure that all variables used in examples have values specified in YAML configs. This does add significant file IO as we have to read every example file prior to generating tests.
> 

Does it cause the provider generation time longer?

zli82016:
The examples files are read automatically and then parsed during the template execution with the function `ParseFiles` in  [ExecuteTemplate](https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/api/resource/examples.go#L318)
```
tmpl, err := template.New(templateFileName).Funcs(google.TemplateFunctions).ParseFiles(templates...)
```

To avoid reading the files twice, one for parsing, and the other one for validation, we can read the files once. 

We can first read the file content manually for each file, and then parse the content using `tmpl.Parse()`.

The content of the example file can be used for variables validation. 

Does this option make sense?

slevenick:
> The examples files are read automatically and then parsed during the template execution with the function `ParseFiles` in [ExecuteTemplate](https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/api/resource/examples.go#L318)
> 
> ```
> tmpl, err := template.New(templateFileName).Funcs(google.TemplateFunctions).ParseFiles(templates...)
> ```
> 
> To avoid reading the files twice, one for parsing, and the other one for validation, we can read the files once.
> 
> We can first read the file content manually for each file, and then parse the content using `tmpl.Parse()`.
> 
> The content of the example file can be used for variables validation.
> 
> Does this option make sense?

Yeah, that option makes sense. I'm a little confused at the structure of ExecuteTemplate though.

ExecuteTemplate within examples.go seems to be the central place that Go templates are executed through for the entire provider generation (resources... types... etc go through here). It's not limited to example generation, so it doesn't make sense to modify it for example-specific behavior.

I can split off ExecuteTemplate to be specific to *Examples, but I'm not sure if that's worth it.

zli82016:
> ExecuteTemplate within examples.go seems to be the central place that Go templates are executed through for the entire provider generation (resources... types... etc go through here). It's not limited to example generation, so it doesn't make sense to modify it for example-specific behavior.
> 
> I can split off ExecuteTemplate to be specific to *Examples, but I'm not sure if that's worth it.

Sorry for the confusion.
[ExecuteTemplate](https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/api/resource/examples.go#L305C6-L305C21) in examples.go is only for *Examples. We can modify it for examples execution.

The other generation (resources, tests and docs) uses the template in
https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/provider/template_data.go

zli82016 (approved):
LGTM. Thanks for the fix.

Rebasing the main branch is needed to address the unit test failure.

ScottSuarez:
> Fixes: [hashicorp/terraform-provider-google#15679 (comment)](https://github.com/hashicorp/terraform-provider-google/issues/15679#event-17597385755)
> 
> Check Vars and TestEnvVars to make sure that all variables used in examples have values specified in YAML configs. This does add significant file IO as we have to read every example file prior to generating tests.
> 
> **Release Note Template for Downstream PRs (will be copied)**
> 
> See [Write release notes](https://googlecloudplatform.github.io/magic-modules/contribute/release-notes/) for guidance.
> 
> ```
> 
> ```

Why don't we add this as a unit test instead. Similar to https://github.com/GoogleCloudPlatform/magic-modules/pull/13929/files#diff-70a2cc3f932572186c59a6a51bb24b1406f8764e1b4891d62971011f412cf610

ScottSuarez:
> For both cases could they be returned by the compiler as warnings, and upgraded to errors in CI? Unit tests are unlikely to be run by contributors locally, while errors at runtime can block development. Warnings w/ warn2err in CI would theoretically both notify folks in local runs & enforce compliance.

My nit is adding compilation runtime for what I would consider unit test behavior. I agree that the user would catch more often before proposing a pr if apart of build, but I am not sure it is worth an impact to overall build time. Pending the actual impact to build time it is likely a non issue. It is just the more of the things like this we do, deaths by a thousand cuts.

rileykarson:
Fair consideration! I think bazel may allow us to parallelise more (especially if we integrate more deeply over time, i.e. breaking service packages into dedicated build rules) and will help keep latency down. If it doesn't or we do see runtime increases we'd want to restructure, yeah.

PR #13939, opened 2025-05-13
Add github membership data to tools

trodge:
This will allow membership data to be shared with the good build approver.

```release-note:none

```

### DIFF
--- a/tools/github-membership/go.mod
+++ b/tools/github-membership/go.mod
@@ -1,0 +1,8 @@
+module github.com/trodge/magic-modules/membership-tools/tools/github-membership
+
+go 1.25
+
+require (
+	github.com/google/go-cmp v0.7.0
+	golang.org/x/exp v0.0.0-20250506013437-ce4c2cf36ca6
+)
--- a/tools/github-membership/membership.go
+++ b/tools/github-membership/membership.go
@@ -1,0 +1,119 @@
+/*
+* Copyright 2023 Google LLC. All Rights Reserved.
+*
+* Licensed under the Apache License, Version 2.0 (the "License");
+* you may not use this file except in compliance with the License.
+* You may obtain a copy of the License at
+*
+*     http://www.apache.org/licenses/LICENSE-2.0
+*
+* Unless required by applicable law or agreed to in writing, software
+* distributed under the License is distributed on an "AS IS" BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+ */
+package membership
+
+import (
+	"fmt"
+	"math/rand"
+	"slices"
+	"time"
+
+	"golang.org/x/exp/maps"
+)
+
+type UserType int64
+
+const (
+	CommunityUserType UserType = iota
+	GooglerUserType
+	CoreContributorUserType
+)
+
+func (ut UserType) String() string {
+	switch ut {
+	case GooglerUserType:
+		return "Googler"
+	case CoreContributorUserType:
+		return "Core Contributor"
+	default:
+		return "Community Contributor"
+	}
+}
+
+func (gh *Client) GetUserType(user string) UserType {
+	if IsCoreContributor(user) {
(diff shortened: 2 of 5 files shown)


trodge:

zli82016:

tools/github-membership/membership.go
```diff
@@ -0,0 +1,119 @@
...
+	case GooglerUserType:
+		return "Googler"
+	case CoreContributorUserType:
+		return "Core Contributor"
+	default:
+		return "Community Contributor"
+	}
+}
+
+func (gh *Client) GetUserType(user string) UserType {
```
ScottSuarez: undefined: Clientcompiler[UndeclaredName](https://pkg.go.dev/golang.org/x/tools/internal/typesinternal#UndeclaredName)

Should this define a client interface that has this function?

tools/github-membership/go.mod
```diff
@@ -0,0 +1,8 @@
+module github.com/trodge/magic-modules/membership-tools/tools/github-membership
+
+go 1.25
```
ScottSuarez:
```suggestion
go 1.23
```
consistency with other packages

tools/github-membership/membership_test.go
ScottSuarez, line 1: add unit test executions to https://github.com/ScottSuarez/magic-modules/blob/7f69b69d2cb6d375afb81a5082ab8930ae571f33/.github/workflows/unit-test-tools.yml#L12

ScottSuarez (changes requested):
Do we want to migrate membership checker to use this library so aren't duplicating code?

trodge:
> Do we want to migrate membership checker to use this library so aren't duplicating code?

Yes, but I want to copy it first so that importing is easier.

trodge:
It's now occurring to me that this approach won't work because I'll need to import the `Client` methods used here, and they can't be imported because they're in the .ci directory. I think a better solution is going to be to move the entire magician into tools so that it can be imported from anywhere, including the good build approver.

ScottSuarez:
> It's now occurring to me that this approach won't work because I'll need to import the `Client` methods used here, and they can't be imported because they're in the .ci directory. I think a better solution is going to be to move the entire magician into tools so that it can be imported from anywhere, including the good build approver.

I'm not sure I understand the purpose of moving the code? We can just import it where it lives today, no?

trodge:
> > It's now occurring to me that this approach won't work because I'll need to import the `Client` methods used here, and they can't be imported because they're in the .ci directory. I think a better solution is going to be to move the entire magician into tools so that it can be imported from anywhere, including the good build approver.
> 
> I'm not sure I understand the purpose of moving the code? We can just import it where it lives today, no?

Packages from paths containing a leading . cannot be imported. https://pkg.go.dev/golang.org/x/mod/module#CheckImportPath

ScottSuarez:
> > > It's now occurring to me that this approach won't work because I'll need to import the `Client` methods used here, and they can't be imported because they're in the .ci directory. I think a better solution is going to be to move the entire magician into tools so that it can be imported from anywhere, including the good build approver.
> > 
> > 
> > I'm not sure I understand the purpose of moving the code? We can just import it where it lives today, no?
> 
> Packages from paths containing a leading . cannot be imported. [pkg.go.dev/golang.org/x/mod/module#CheckImportPath](https://pkg.go.dev/golang.org/x/mod/module#CheckImportPath)

I'm okay with this philosophically but would like Stephen to sign off as well just to make sure I'm not missing anything !

melinath:
I'm initially hesitant to make this large of a change when the stated purpose is pretty narrow. It feels like we ought to be able to only move the membership code + client layer code, rather than needing to move everything.

There's also a security question: if we're moving _any_ code from .ci to tools, we should take a moment to consider whether we should automatically consider changes to tools unsafe, similar to .ci.

That being said - checking membership is now significantly easier than it was previously so perhaps we don't need to share code with the GBA at all?

melinath:
As discussed - switching to a yaml file for membership / vacation data, will pull that into the GBA as needed. We'll implement the membership check in the GBA separately instead of sharing code.

PR #13941, opened 2025-05-13
Make VCR correctly handle compound tests

trodge:
fixes https://github.com/hashicorp/terraform-provider-google/issues/19930

```release-note:none

```


trodge:

zli82016:

slevenick (approved):

trodge:
Had to rework this after testing.

slevenick:
Error message for the failing test is giving me a 404, is that related to this change?

trodge:
> Error message for the failing test is giving me a 404, is that related to this change?

Ah, good catch. It looks like the build logs for compound tests use the compound test name while the debug logs use the subtest name. There is probably a more elegant solution to this problem but it would require changing the Result struct around.

PR #13955, opened 2025-05-14
expand tgc metadata coverage to 100% of testcases

ScottSuarez:

```release-note:none

```

### DIFF
--- a/mmv1/third_party/terraform/acctest/tgc_utils.go
+++ b/mmv1/third_party/terraform/acctest/tgc_utils.go
@@ -1,3 +1,5 @@
+// Copyright (c) HashiCorp, Inc.
+// SPDX-License-Identifier: MPL-2.0
 package acctest
 
 import (
@@ -61,4 +63,92 @@ func GetTestMetadataForTgc(service, address, rawConfig string) resource.TestChec
 		log.Printf("[DEBUG]TGC raw_config starts %sEnd of TGC raw_config", rawConfig)
 		return nil
 	}
+}
+
+// parseResources extracts all resources from a Terraform configuration string
+func parseResources(config string) []string {
+	// This regex matches resource blocks in Terraform configurations
+	resourceRegex := regexp.MustCompile(`resource\s+"([^"]+)"\s+"([^"]+)"`)
+	matches := resourceRegex.FindAllStringSubmatch(config, -1)
+
+	var resources []string
+	for _, match := range matches {
+		if len(match) >= 3 {
+			// Combine resource type and name to form the address
+			resources = append(resources, fmt.Sprintf("%s.%s", match[1], match[2]))
+		}
+	}
+
+	return resources
+}
+
+// getServicePackage determines the service package for a resource type
+func getServicePackage(resourceType string) string {
+	var ServicePackages = map[string]string{
+		"google_compute_":   "compute",
+		"google_storage_":   "storage",
+		"google_sql_":       "sql",
+		"google_container_": "container",
+		"google_bigquery_":  "bigquery",
+		"google_project":    "resourcemanager",
+		"google_cloud_run_": "cloudrun",
+	}
+
+	// Check for exact matches first
+	if service, ok := ServicePackages[resourceType]; ok {
+		return service
+	}
+
+	// Check for prefix matches
+	for prefix, service := range ServicePackages {
+		if strings.HasPrefix(resourceType, prefix) {
+			return service
+		}
+	}
+
+	// Default to "unknown" if no match found
+	return "unknown"
+}
+
+// extendWithTGCData adds TGC metadata check functions to each TestStep
+func extendWithTGCData(c resource.TestCase) resource.TestCase {
+	var updatedSteps []resource.TestStep
+
+	for _, step := range c.Steps {
+		if step.Config != "" {
+			// Parse resources from the config
+			resources := parseResources(step.Config)
+
+			// Create TGC metadata checks for each resource
+			var tgcChecks []resource.TestCheckFunc
+			for _, res := range resources {
+				parts := strings.Split(res, ".")
+				if len(parts) >= 2 {
+					resourceType := parts[0]
+					service := getServicePackage(resourceType)
+					tgcChecks = append(tgcChecks, GetTestMetadataForTgc(service, res, step.Config))
+				}
+			}
+
+			// If there are TGC checks to add
+			if len(tgcChecks) > 0 {
+				// If there's an existing check function, wrap it with ours
+				if step.Check != nil {
+					existingCheck := step.Check
+					step.Check = resource.ComposeTestCheckFunc(
+						existingCheck,
+						resource.ComposeTestCheckFunc(tgcChecks...),
+					)
+				} else {
+					// Otherwise, just use our TGC checks
+					step.Check = resource.ComposeTestCheckFunc(tgcChecks...)
+				}
+			}
+		}
+
+		updatedSteps = append(updatedSteps, step)
+	}
+
+	c.Steps = updatedSteps
+	return c
 }
--- a/mmv1/third_party/terraform/acctest/vcr_utils.go
+++ b/mmv1/third_party/terraform/acctest/vcr_utils.go
@@ -148,6 +148,8 @@ func VcrTest(t *testing.T, c resource.TestCase) {
 		c = initializeReleaseDiffTest(c, t.Name())
 	}
 
+	c = extendWithTGCData(c)
+
 	// terraform_labels is a computed field to which "goog-terraform-provisioned": "true" is always
 	// added by the provider. ImportStateVerify "checks for strict equality and does not respect
 	// DiffSuppressFunc or CustomizeDiff" so any test using ImportStateVerify must ignore
--- a/mmv1/third_party/terraform/services/compute/resource_compute_instance_test.go.tmpl
+++ b/mmv1/third_party/terraform/services/compute/resource_compute_instance_test.go.tmpl
@@ -159,7 +159,7 @@ func computeInstanceImportStep(zone, instanceName string, additionalImportIgnore
 	// metadata is only read into state if set in the config
 	// importing doesn't know whether metadata.startup_script vs metadata_startup_script is set in the config,
 	// it always takes metadata.startup-script
-	ignores := []string{"metadata.%", "metadata.startup-script", "metadata_startup_script", "boot_disk.0.initialize_params.0.resource_manager_tags.%",  "params.0.resource_manager_tags.%"}
+	ignores := []string{"metadata.%", "metadata.startup-script", "metadata_startup_script", "boot_disk.0.initialize_params.0.resource_manager_tags.%", "params.0.resource_manager_tags.%"}
 
 	return resource.TestStep{
 		ResourceName:            "google_compute_instance.foobar",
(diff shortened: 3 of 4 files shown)


mmv1/third_party/terraform/acctest/tgc_utils.go
```diff
@@ -62,3 +63,91 @@ func GetTestMetadataForTgc(service, address, rawConfig string) resource.TestChec
...
+			// Parse resources from the config
+			resources := parseResources(step.Config)
+
+			// Create TGC metadata checks for each resource
+			var tgcChecks []resource.TestCheckFunc
+			for _, res := range resources {
+				parts := strings.Split(res, ".")
+				if len(parts) >= 2 {
+					resourceType := parts[0]
+					service := getServicePackage(resourceType)
```
zli82016, line 127: For the unknown service, I think we don't need to add the check function.  The `unknown` asset name will be filtered out before calling the cai assets API eventually.

mmv1/third_party/terraform/acctest/tgc_utils.go
```diff
@@ -62,3 +63,91 @@ func GetTestMetadataForTgc(service, address, rawConfig string) resource.TestChec
...
+
+	// Default to "unknown" if no match found
+	return "unknown"
+}
+
+// extendWithTGCData adds TGC metadata check functions to each TestStep
+func extendWithTGCData(c resource.TestCase) resource.TestCase {
+	var updatedSteps []resource.TestStep
+
+	for _, step := range c.Steps {
```
zli82016, line 116: `[batchGetAssetsHistory](https://cloud.google.com/asset-inventory/docs/reference/rest/v1/TopLevel/batchGetAssetsHistory)` is called to get the cai asset history and the latest data in the history is used for testing. Based on that, only the raw config of the last step is needed.

mmv1/third_party/terraform/acctest/tgc_utils.go
ScottSuarez: I would like to add the metadata for all tests so I can parse them for separate tasks. Maybe we can sync up here on how to be expose this metadata for both our usecases.

mmv1/third_party/terraform/acctest/tgc_utils.go
```diff
@@ -62,3 +63,91 @@ func GetTestMetadataForTgc(service, address, rawConfig string) resource.TestChec
...
+		if step.Config != "" {
+			// Parse resources from the config
+			resources := parseResources(step.Config)
+
+			// Create TGC metadata checks for each resource
+			var tgcChecks []resource.TestCheckFunc
+			for _, res := range resources {
```
zli82016, line 123: I wonder if there is a way to know the main resource address(google_compute_instance.foobar) or main resource type(google_compute_instance). We would like to only test the main resource.

mmv1/third_party/terraform/acctest/tgc_utils.go
zli82016: If `unknown` is needed for your use cases, we can keep it. I can filter out them before calling the API.

mmv1/third_party/terraform/acctest/tgc_utils.go
```diff
@@ -38,7 +38,8 @@ func GetTestMetadataForTgc(service, address, rawConfig string) resource.TestChec
 		// Convert the resource ID into CAI asset name
 		// and then print out the CAI asset name in the logs
 		if apiServiceName, ok := ApiServiceNames[resourceType]; !ok {
-			return fmt.Errorf("The Cai product backend name for resource %s doesn't exist.", resourceType)
+			unknownType := "unkown"
```
zli82016, line 41:
```suggestion
			unknownType := "unknown"
```

mmv1/third_party/terraform/acctest/tgc_utils.go
ScottSuarez: Is there a reason to only test the main resource. If we have full coverage then we will have more test data for other resources.

mmv1/third_party/terraform/acctest/tgc_utils.go
ScottSuarez: got it, I'll revise

mmv1/third_party/terraform/acctest/tgc_utils.go
zli82016: The test data for other resources will be duplicate and not necessary?

Also, the time to run the tests will become much longer.

ScottSuarez: I changed it to emit primary resource and only one per last config. as discussed

mmv1/third_party/terraform/acctest/tgc_utils.go
```diff
@@ -1,64 +1,307 @@
...
+	"runtime"
 	"strings"
 
 	"github.com/hashicorp/terraform-plugin-testing/helper/resource"
 	"github.com/hashicorp/terraform-plugin-testing/terraform"
 )
 
+type ResourceMetadata struct {
+	CaiAssetName    string         `json:"cai_asset_name"`
+	ResourceType    string         `json:"resource_type"`
```
zli82016, line 18: Do we want to keep `ResourceType` with `ResourceAddress`?

mmv1/third_party/terraform/acctest/tgc_utils.go
```diff
@@ -1,64 +1,307 @@
...
+}
+
+// CollectAllTgcMetadata collects metadata for all resources in a test step
+func CollectAllTgcMetadata(config string, resourceMetadata map[string]ResourceMetadata) resource.TestCheckFunc {
 	return func(s *terraform.State) error {
-		splits := strings.Split(address, ".")
-		if splits == nil || len(splits) < 2 {
-			return fmt.Errorf("The resource address %s is invalid.", address)
+		// Get the name of the test that invoked this function
+		testName := getCallerInfo()
```
zli82016, line 57: t.Name can be used to get the test name. Or I missing something.

ScottSuarez: t is not in scope of the vcr functions

mmv1/third_party/terraform/acctest/tgc_utils.go
ScottSuarez: Personally I'm inclined to keep it. I realize the information can be retrieved through the data, but it's nice to have for ease of access and not harmful? What do you think?

mmv1/third_party/terraform/acctest/tgc_utils.go
```diff
@@ -1,64 +1,307 @@
...
+			if nextStep.ImportStateIdPrefix != "" {
+				metadata.Id = fmt.Sprintf("%s<AUTO_ID>", nextStep.ImportStateIdPrefix)
+				return metadata
 			}
-			caiAssetName := fmt.Sprintf("//%s/%s", apiServiceName, rName)
-			log.Printf("[DEBUG]TGC CAI asset names start\n%s\nEnd of TGC CAI asset names", caiAssetName)
+
+			// If ImportStateIdFunc is set, get function info
+			if nextStep.ImportStateIdFunc != nil {
+				metadata.Id = "<DYNAMIC_IMPORT_ID>"
```
zli82016, line 189: I don't see `<DYNAMIC_IMPORT_ID>` is replaced anywhere.

mmv1/third_party/terraform/acctest/tgc_utils.go
```diff
@@ -1,64 +1,307 @@
...
-		if splits == nil || len(splits) < 2 {
-			return fmt.Errorf("The resource address %s is invalid.", address)
+		// Get the name of the test that invoked this function
+		testName := getCallerInfo()
+
+		// Create the consolidated TGC payload
+		tgcPayload := TgcMetadataPayload{
+			TestName:         testName,
+			RawConfig:        config,
+			ResourceMetadata: resourceMetadata,
```
zli82016, line 63: `tgcPayload` can be initialized and then passed to the function `CollectAllTgcMetadata`?

mmv1/third_party/terraform/acctest/tgc_utils.go
```diff
@@ -1,64 +1,307 @@
...
+					resourceType := parts[0]
+
+					// Determine import metadata if the next step is an import step
+					importMeta := determineImportMetadata(c.Steps, i, res)
+
+					// Create metadata for this resource
+					resourceMetadata[res] = ResourceMetadata{
+						ResourceType:    resourceType,
+						ResourceAddress: res,
+						ImportMetadata:  importMeta,
```
zli82016, line 251: Is `Service` missing?

mmv1/third_party/terraform/acctest/tgc_utils.go
zli82016: `t` is the parameter of the function [func VcrTest(t *testing.T, c resource.TestCase) {](https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/third_party/terraform/acctest/vcr_utils.go#L142), where `extendWithTGCData(c)` is called.

mmv1/third_party/terraform/acctest/tgc_utils.go
zli82016: Then we can keep it.

mmv1/third_party/terraform/acctest/tgc_utils.go
ScottSuarez: This is true, I'll make the change

mmv1/third_party/terraform/acctest/tgc_utils.go
ScottSuarez: Service was removed from the test and put on the resource directly

mmv1/third_party/terraform/acctest/tgc_utils.go
ScottSuarez: DYNAMIC_IMPORT_ID is stating that a function is used to supply the import id. There are about 3 or 4 instances of this. Unfortunately I dont see a way to supply the was import function at this time. Dynamic is intended.

mmv1/third_party/terraform/acctest/tgc_utils.go
ScottSuarez: changed to account for this

mmv1/third_party/terraform/acctest/tgc_utils.go
ScottSuarez: we could use reflection on the testing object to find where which service the test was defined in but there are two things here. The service the resource is defined in and the service the test is defined in.

mmv1/third_party/terraform/acctest/tgc_utils.go
ScottSuarez: hmm actually we might be able to send terraform state to the function directly

ScottSuarez: But I would prefer to do this in another pr

mmv1/third_party/terraform/acctest/tgc_utils.go
ScottSuarez: ah sorry you are right. I added it

zli82016 (approved):
LGTM.  Great work. Thanks.

PR #13964, opened 2025-05-15
EdgeCacheService: Fixed `defaultTtl/maxTtl` Config validation failed error when switching `cache_mode`

Asing1001:
To align with [EdgeCacheService - CDNpolicy](https://cloud.google.com/media-cdn/docs/reference/rest/v1/projects.locations.edgeCacheServices#cdnpolicy) behavior, the `defaultTtl/maxTtl` must respect the client config instead of using the `default_from_api` which is actually the `last state` from api. By doing this, the omitted field will not be sent to the api.

Fixes: 
- https://github.com/hashicorp/terraform-provider-google/issues/16200
- https://github.com/hashicorp/terraform-provider-google/issues/14903
- https://github.com/hashicorp/terraform-provider-google/issues/20661

```release-note: bug
networkservices: fixed validation error when modifying the `cache_mode` field in `edge_cache_service`
```

### DIFF
--- a/mmv1/templates/terraform/encoders/network_services_edge_cache_service.go.tmpl
+++ b/mmv1/templates/terraform/encoders/network_services_edge_cache_service.go.tmpl
@@ -1,0 +1,53 @@
+// This encoder ensures TTL fields are handled correctly based on cache mode
+routing, ok := obj["routing"].(map[string]interface{})
+if !ok {
+	return obj, nil
+}
+
+pathMatchers, ok := routing["pathMatchers"].([]interface{})
+if !ok || len(pathMatchers) == 0 {
+	return obj, nil
+}
+
+for _, pm := range pathMatchers {
+	pathMatcher, ok := pm.(map[string]interface{})
+	if !ok {
+		continue
+	}
+
+	routeRules, ok := pathMatcher["routeRules"].([]interface{})
+	if !ok {
+		continue
+	}
+
+	for _, rr := range routeRules {
+		routeRule, ok := rr.(map[string]interface{})
+		if !ok {
+			continue
+		}
+
+		routeAction, ok := routeRule["routeAction"].(map[string]interface{})
+		if !ok {
+			continue
+		}
+
+		cdnPolicy, ok := routeAction["cdnPolicy"].(map[string]interface{})
+		if !ok {
+			continue
+		}
+
+		// Handle TTL fields based on cache mode
+		if cacheMode, ok := cdnPolicy["cacheMode"].(string); ok {
+			switch cacheMode {
+			case "USE_ORIGIN_HEADERS", "BYPASS_CACHE":
+				delete(cdnPolicy, "clientTtl")
+				delete(cdnPolicy, "defaultTtl")
+				delete(cdnPolicy, "maxTtl")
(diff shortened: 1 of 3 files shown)


SirGitsalot (changes requested):
Removing `default_from_api` is a [breaking change](https://googlecloudplatform.github.io/magic-modules/breaking-changes/breaking-changes/#field-level-breaking-changes) and would have to wait until the 7.0 release.

From reading the attached bugs, I agree that [customer encoders](https://github.com/hashicorp/terraform-provider-google/issues/14903#issuecomment-1806151482), at least for https://github.com/hashicorp/terraform-provider-google/issues/14903 and https://github.com/hashicorp/terraform-provider-google/issues/20661 (and for https://github.com/hashicorp/terraform-provider-google/issues/16200 on creation but maybe not updates; see below). We have [some basic documentation](https://googlecloudplatform.github.io/magic-modules/develop/custom-code/) for writing the encoder(s). The encoder should check the value of `cache_mode` and do the right thing vis-a-vis `max_ttl`: set it to the requested value, omit it, set it to `null`, etc.

Terraform Provider Google uses GCP REST APIs, so requests and responses are sent JSON-encoded. `max_ttl` is of type `google.protobuf.Duration` which is normally a proto message type, but the [JSON encoding is a string](https://protobuf.dev/reference/protobuf/google.protobuf/#duration). Normally omitting a string from a JSON encoded proto (in proto3 anyway) is the same as including an empty string, I assume that is not the case for durations (however it works, it'll need to be accounted for in the custom encoder).

Asing1001:
It's fine to use `custom encoder`, but the drawback is: It is hard-coded in terraform client to reflect backend constraint, so every time the logic change in backend, it will need to change to reflect. e.g. introduce a new `cacheMode` which require omitted `defaultTtl`

mmv1/templates/terraform/encoders/network_services_edge_cache_service.go.tmpl
```diff
@@ -0,0 +1,53 @@
...
+		if cacheMode, ok := cdnPolicy["cacheMode"].(string); ok {
+			switch cacheMode {
+			case "USE_ORIGIN_HEADERS", "BYPASS_CACHE":
+				delete(cdnPolicy, "clientTtl")
+				delete(cdnPolicy, "defaultTtl")
+				delete(cdnPolicy, "maxTtl")
+			case "FORCE_CACHE_ALL":
+				delete(cdnPolicy, "maxTtl")
+			}
+		}
```
Asing1001, line 49: This block need to reflect the backend proto constraint, which is I prefer directly use `default_from_api=false` implementation.

SirGitsalot: That will work. IMO adding `default_from_api` should be a last resort due to the downsides that you're bumping up against. This PR will fix the immediate problems that users are facing today, and then you can [make the breaking change](https://googlecloudplatform.github.io/magic-modules/breaking-changes/make-a-breaking-change/#in-the-700-major-release) at your leisure (well, not *too* leisurely: the 7.0 release is scheduled for the end of August)

SirGitsalot (approved):

PR #13966, opened 2025-05-15
Revert "Removed cloud identity group set_computed_name post_create"

melinath:
Reverts GoogleCloudPlatform/magic-modules#13627.

It looks like this does cause real failures (due to introducing a bug).

```release-note:none
```


slevenick (approved):

melinath:
This brings us back to the previous error :-)

PR #13980, opened 2025-05-15
Create stub product.go files for products

rileykarson:
Fixes https://github.com/hashicorp/terraform-provider-google/issues/15633

There are a few symptoms discussed in the bug, I think this addresses them all.

* MMv1 already seems to be generating import statements for the versions supported by a product, probably after the Go rewrite
* There is now always a Go file present for a product that is valid in a version
* Import statements that are generated are always used

This adds the first resourceless product level Generate call- Operation was once per product but ultimately just picked the first valid resource, assuming all resources shared the same settings. That required adding Compiler/SetCompiler, a runtime generator input that got propagated on Resource (and now Product).

```release-note:none
```


rileykarson:
You can see the impact of the changes on some edge cases in that diff / the original checks! https://github.com/GoogleCloudPlatform/magic-modules/compare/ba82d8a428103e3b2984fff79c39eea672f0a468..088bb69d9fab1992e6c43617e66307481a6b3556 has the MM changes.

mmv1/third_party/terraform/provider/provider_mmv1_resources.go.tmpl
```diff
@@ -508,3 +508,14 @@ var handwrittenIAMResources = map[string]*schema.Resource{
 	"google_service_account_iam_policy":            tpgiamresource.ResourceIamPolicy(resourcemanager.IamServiceAccountSchema, resourcemanager.NewServiceAccountIamUpdater, resourcemanager.ServiceAccountIdParseFunc),
 	// ####### END non-generated IAM resources ###########
 }
+
+// UseGeneratedProducts uses every generated product to avoid "imported and not used" errors.
```
c2thorn, line 512: Nice way around this!

mmv1/api/product.go
```diff
@@ -68,6 +69,9 @@ type Product struct {
 	LegacyName string `yaml:"legacy_name,omitempty"`
 
 	ClientName string `yaml:"client_name,omitempty"`
+
+	// The compiler to generate the downstream files, for example "terraformgoogleconversion-codegen".
+	Compiler string `yaml:"-"`
```
c2thorn, line 74: I was thinking it may be worthwhile to have a more universal way to check which MMv1 compiler/provider a template is being called from. But I guess all inputs into templates so far are either a resource which has a reference to the product object, or your new template that takes in the product itself.

c2thorn (approved):

mmv1/api/product.go
rileykarson: Yeah- I just shimmed this for now, but I was thinking we may want to introduce a GeneratorSettings/GeneratorRunSettings struct that we [embed](https://gobyexample.com/struct-embedding) in the api types & fill in during shared Generate code. Gets rid of the opportunity for drift/error.

rileykarson:
Failing tests are unrelated (this just kicked every package and we're carrying a few around)

PR #13990, opened 2025-05-16
Suggest alternative measure for pr size

ScottSuarez:

```release-note:none

```


ScottSuarez:
@melinath lmk what you think here. I think this measure would better account for changes we care about. IMO we should probably also ignore new resource yaml additions but this is not accounted for here.

melinath (approved):
This seems reasonable. I think we do want to include yaml files, at least for now, since they correspond to non-test code changes in the downstreams.

PR #13993, opened 2025-05-16
Allow `pod_ipv4_cidr_block` to accept netmask for `container_node_pool`

jackweinbender:
This PR brings the `pod_ipv4_cidr_block` attribute in the `container_node_pool` resource into alignment with attributes of the same name and description from `container_cluster`.

Notably, at present the documentation is incorrect—validation of this field fails if you try to include a `/netmask` value to this field.

There's been an open issue about this for some time: https://github.com/hashicorp/terraform-provider-google/issues/11024 and this PR simply implements the suggested fix.

```release-note:bug
container: fixed nodepool secondary range validation to allow the use of netmasks.
```


roaks3 (commented):
Seems reasonable and probably correct, but we do need to be careful because if this new diff suppress is more restrictive in any way, then users could experience a failure out of the blue.

Running tests now as a first pass at this, but I'll probably also need to do a scan through each implementation to double check. If you happen to know of precedence for changing safely from `ValidateIpCidrRange` -> `CidrOrSizeDiffSuppress`, that would help as well.

roaks3:
I believe this change is safe to make. We are removing the CIDR validation entirely to support this, which means users won't benefit from that client-side validation when using a full CIDR, but it also shouldn't break anyone because we are making it less restrictive. The new diff suppress seems reasonably targeted _only_ toward the use case of handling a specified size that is expanded by the server, which we definitely want fixed.

roaks3 (approved):

jackweinbender:
@roaks3 Sorry I kinda dropped the ball here. Thanks for picking it up!

PR #13995, opened 2025-05-16
Allow updates to the mtu property in the Network resource.

AshwinGururagh:
Currently using terraform to update the mtu property of the Network results in forcing the replacement of the Network resource.

This change marks the mtu property as mutable, and instead uses the PATCH API on the Network resource to update the mtu property of the network.

Fixes https://github.com/hashicorp/terraform-provider-google/issues/22543

```release-note: bug
compute: fixed VPC Network's MTU to be updatable without forcing a replacement of the Network resource.
```


roaks3 (approved):
Code and tests LGTM

PR #13901, opened 2025-05-09
Fixed TestAccComputeHealthCheck_tcpAndSsl_shouldFail

melinath:
This was broken by https://github.com/GoogleCloudPlatform/magic-modules/pull/13608 and is a result of https://github.com/hashicorp/terraform-provider-google/issues/19079 plus https://github.com/hashicorp/terraform-provider-google/issues/15427

Fixed https://github.com/hashicorp/terraform-provider-google/issues/22651
Fixed https://github.com/hashicorp/terraform-provider-google/issues/22659

**Release Note Template for Downstream PRs (will be copied)**

See [Write release notes](https://googlecloudplatform.github.io/magic-modules/contribute/release-notes/) for guidance.

```release-note:none

```

PR #14009, opened 2025-05-18
remove TestAccComputeHealthCheck_tcpAndSsl_shouldFail

shuyama1:
The test started failing in beta after a conflicting field was introduced in https://github.com/GoogleCloudPlatform/magic-modules/pull/13608. While https://github.com/GoogleCloudPlatform/magic-modules/pull/13901 addressed these beta failures by updating the expected error message to include the new field, it caused the tests to fail in GA because the field is beta-only

Instead of fixing the tests in GA, I propose removing them for the following reasons:
1. Limited value of the tests. The tests are narrowly focused on verifying that the provider errors out when two specific conflicting fields are set. It's basically testing the conflict_with behavior for only this particular pair of fields, while we don't apply the same validation to other conflicting field.
2. High Maintenance Cost. These tests require updates whenever a new conflicting field is added

Fixes https://github.com/hashicorp/terraform-provider-google/issues/22812

```release-note:none

```

### DIFF
--- a/mmv1/third_party/terraform/services/compute/resource_compute_health_check_test.go.tmpl
+++ b/mmv1/third_party/terraform/services/compute/resource_compute_health_check_test.go.tmpl
@@ -192,26 +192,6 @@ func TestAccComputeHealthCheck_typeTransition(t *testing.T) {
 	})
 }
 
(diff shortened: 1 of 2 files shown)


melinath (approved):
SGTM

PR #13570, opened 2025-04-07
remove service_folder_identity from ga provider and add note to mlengine_model

ScottSuarez:
https://hashicorp.teamcity.com/test/4428545972839984415?currentProjectId=TerraformProviders_GoogleCloud_GOOGLE_NIGHTLYTESTS

**Release Note Template for Downstream PRs (will be copied)**

See [Write release notes](https://googlecloudplatform.github.io/magic-modules/contribute/release-notes/) for guidance.

```release-note:none

```

PR #14019, opened 2025-05-19
Removing folder policy orchestrator tests from ga provider

darthmelonder:
fixes https://github.com/hashicorp/terraform-provider-google/issues/22518

```release-note:none
```

### DIFF
--- a/mmv1/third_party/terraform/services/osconfigv2/resource_os_config_v2_policy_orchestrator_for_folder_test.go.tmpl
+++ b/mmv1/third_party/terraform/services/osconfigv2/resource_os_config_v2_policy_orchestrator_for_folder_test.go.tmpl
@@ -1,3 +1,4 @@
+{{- if ne $.TargetVersionName "ga" -}}
 package osconfigv2_test
 
 import (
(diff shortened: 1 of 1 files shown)


darthmelonder:
@NickElliot , this is to address the failing tests for osconfig v2 policy orchestrator resource on folder level.

More context:

a. For project level and org level, we have setup project and organization for our tests. The service identities (P4SAs) are setup already there, and tests run successfully.
b. For folder level, we create folder, setup P4SAs, add iAM permissions, setup IAM policies, run policy orchestrators tests, clean up folder.

Since folder level tests require P4SAs, it depends on `resource_folder_service_identity`. However, this was removed from GA in https://github.com/GoogleCloudPlatform/magic-modules/pull/13570. And now our folder level tests are failing. The only way I could think was to remove the tests for GA (we will still have tests running for beta provider though). Let me know if you have any other ideas.

NickElliot (approved):
yeah, if these cant be ran in GA and an override to breaking change remove them from GA already happened, makes sense to disable the tests.

PR #14022, opened 2025-05-19
fix: (storage) delete non current version of objects 

gurusai-voleti:
Fixes: 

```release-note:bug
storage: delete non current objects
```

### DIFF
--- a/mmv1/third_party/terraform/services/storage/resource_storage_bucket.go.tmpl
+++ b/mmv1/third_party/terraform/services/storage/resource_storage_bucket.go.tmpl
@@ -1093,6 +1093,7 @@ func resourceStorageBucketDelete(d *schema.ResourceData, meta interface{}) error
 	bucket := d.Get("name").(string)
 
 	var listError, deleteObjectError error
+	var emptyBucket bool
 	for deleteObjectError == nil {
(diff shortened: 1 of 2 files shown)


kautikdk (commented):
Is your solution related to the deletion of non current version of object? I believe the original issue is about `force_destroy` is not working for buckets with versioning enabled.

slevenick (commented):
What's the next step here? As far as I can tell we're waiting on review from @kautikdk

gurusai-voleti:
> What's the next step here? As far as I can tell we're waiting on review from @kautikdk

I am verifying whether the current logic is deleting all the objects or not even version objects, and it seems no need to update the logic, will confirm if existing logic working fine  we need some more info from the bug reporter in which case customer faced the error

gurusai-voleti:
@slevenick  @kautikdk created a bucket with more than 1000 version objects and few live objects and tested the current code on main branch (latest tf version), no changes are required for delete functionality as current logic is working fine, until we get some info on the customer issue, we will keep this PR as draft

kautikdk:
@gurusai-voleti so that's what I was thinking. The PR code might not be completely related to the issue that was linked. If you think there is an separate bug that this PR is fixing, you can create one GitHub issue with storage tag and link that issue to this PR.

PR #14031, opened 2025-05-19
Convert review functions to go-gh

ScottSuarez:
closes https://github.com/hashicorp/terraform-provider-google/issues/17191

```release-note:none

```

### DIFF
--- a/.ci/magician/github/set.go
+++ b/.ci/magician/github/set.go
@@ -80,6 +80,10 @@ func (c *Client) UpdateComment(prNumber, comment string, id int) error {
 
 // RequestPullRequestReviewers adds reviewers to a pull request
 func (c *Client) RequestPullRequestReviewers(prNumber string, reviewers []string) error {
+	if len(reviewers) == 0 {
+		return nil
+	}
+
 	num, err := strconv.Atoi(prNumber)
 	if err != nil {
 		return err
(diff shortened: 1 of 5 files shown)


.ci/magician/github/set.go
```diff
@@ -78,39 +78,46 @@ func (c *Client) UpdateComment(prNumber, comment string, id int) error {
...
-		"reviewers":      reviewers,
-		"team_reviewers": {},
+	// Create the reviewers request
+	reviewersRequest := gh.ReviewersRequest{
+		Reviewers:     reviewers,
+		TeamReviewers: nil,
 	}
 
-	err := utils.RequestCallWithRetry(url, "POST", gh.token, nil, body)
+	_, _, err = c.gh.PullRequests.RequestReviewers(c.ctx, defaultOwner, defaultRepo, num, reviewersRequest)
```
melinath, line 94: It looks like this line is causing a failure because it doesn't match the schema GitHub expects:

```
422 Invalid request.

No subschema in "anyOf" matched.
"reviewers" wasn't supplied.
"team_reviewers" wasn't supplied.
```


I think previously we were getting around this by accident because we were sending an empty object for team_reviewers - now we probably need to make this function a no-op if the list of reviewers is empty. (And we don't need to set TeamReviewers at all.)

.ci/magician/github/set.go
```diff
@@ -78,39 +78,46 @@ func (c *Client) UpdateComment(prNumber, comment string, id int) error {
...
-	body := map[string][]string{
-		"reviewers":      reviewers,
-		"team_reviewers": {},
+	reviewersRequest := gh.ReviewersRequest{
+		Reviewers:     reviewers,
+		TeamReviewers: nil,
 	}
 
-	err := utils.RequestCall(url, "DELETE", gh.token, nil, body)
+	_, err = c.gh.PullRequests.RemoveReviewers(c.ctx, defaultOwner, defaultRepo, num, reviewersRequest)
```
melinath, line 115: I bet we need the same logic here.

melinath (approved):
LGTM apart from the failure requesting service team reviewers.

PR #14033, opened 2025-05-19
Disable siteverification domain test

SirGitsalot:
This is probably what I meant to do when I first wrote the test, since one of the prerequisites is effectively "register a new, real domain that has properly configured DNS that has propagated globally". The storage bucket test is still there during all test runs, and the API hasn't changed in like 15 years so we should be safe.

Fixes https://github.com/hashicorp/terraform-provider-google/issues/19533

```release-note: none

```


SirGitsalot:

hao-nan-li (approved):

PR #14034, opened 2025-05-19
Add Managed Kafka Acl resource and tests.

amfisher-404:
Description: Added a new Managed Kafka Acl terraform resource to the Google Terraform provider. Includes a basic create and update test.

Issue: https://b.corp.google.com/issues/374342184

```release-note:new-resource
`google_managed_kafka_acl`
```


mmv1/products/managedkafka/Acl.yaml
```diff
@@ -0,0 +1,149 @@
...
+    output: true
+  - name: 'aclEntries'
+    type: Array
+    is_set: true
+    description: "The acl entries that apply to the resource pattern. The maximum number of allowed
+      entries is 100."
+    item_type:
+      type: NestedObject
+      properties:
+        - name: 'principal'
```
slevenick, line 105: required: true?

mmv1/products/managedkafka/Acl.yaml
```diff
@@ -0,0 +1,149 @@
...
+          type: String
+          description: 'The principal. Specified as Google Cloud account, with the Kafka
+            StandardAuthorizer prefix User:". For example: "User:test-kafka-client@test-project.iam.gserviceaccount.com".
+            Can be the wildcard "User:*" to refer to all users.'
+        - name: 'permissionType'
```
slevenick, line 110: Required?

mmv1/products/managedkafka/Acl.yaml
```diff
@@ -0,0 +1,149 @@
...
+          type: String
+          default_value: "ALLOW"
+          description: 'The permission type. Accepted values are (case insensitive): ALLOW, DENY.'
+        - name: 'operation'
```
slevenick, line 114: Required?

mmv1/products/managedkafka/Acl.yaml
```diff
@@ -0,0 +1,149 @@
...
+          type: String
+          description: |
+            The operation type. Allowed values are (case insensitive): ALL, READ,
+            WRITE, CREATE, DELETE, ALTER, DESCRIBE, CLUSTER_ACTION, DESCRIBE_CONFIGS,
+            ALTER_CONFIGS, and IDEMPOTENT_WRITE. See https://kafka.apache.org/documentation/#operations_resources_and_protocols
+            for valid combinations of resource_type and operation for different Kafka API requests.
+        - name: 'host'
```
slevenick, line 121: Required?

mmv1/products/managedkafka/Acl.yaml
```diff
@@ -0,0 +1,149 @@
...
+          type: String
+          default_value: "*"
+          description: 'The host. Must be set to "*" for Managed Service for Apache Kafka.'
+  - name: 'etag'
```
slevenick, line 125: I think this should be output_only unless there's a reason to specify etag within Terraform

mmv1/products/managedkafka/Acl.yaml
amfisher-404: To clarify: For an Array of NestedObject where the array must not be empty, should the `required` annotation go on the aclEntries array field, or within the item_type NestedObject part? (I'm thinking the former.)

slevenick: The former.

I would guess most of the nested fields should also have `required` on them as well. I'd guess if someone specifies a specific `acl_entry` it must have a principal, so principal should be required. Similar for any other fields that must be specified for an individual entry.

amfisher-404: Thanks!

> I would guess most of the nested fields should also have required on them as well.

Agreed. I added `required` for the fields without default values in prior commit ddb99a224d9120140e3478fc1fe4bc811a5f0f5b.

mmv1/products/managedkafka/Acl.yaml
amfisher-404: The issue is, the etag does need to be provided in the ACL for future updates to succeed.

Test failure with etag as output_only:

```
resource_managed_kafka_acl_test.go:19: Step 3/4 error: Error running apply: exit status 1
        
        Error: Error updating Acl "projects/ci-test-project-188019/locations/us-central1/clusters/tf-test-my-cluster8wqckoe84a/acls/topic/tf-test-my-acl8wqckoe84a": googleapi: Error 400: The request was invalid: etag must be specified on UpdateAcl. Use GetAcl to get the current etag
        Details:
        [
          {
            "fieldViolations": [
              {
                "field": "etag"
              }
            ]
          },
          {
            "requestId": "4b3e5414714372b5"
          }
        ]
        
          with google_managed_kafka_acl.example,
          on terraform_plugin_test.tf line 18, in resource "google_managed_kafka_acl" "example":
          18: resource "google_managed_kafka_acl" "example" {
```

This is why I had gone with `default_from_api`, so the API-provided value is persisted in future requests. However, IMO it doesn't make sense for tests to exercise the user providing this field. Is there some better alternative you'd propose?

mmv1/products/managedkafka/Acl.yaml
```diff
@@ -0,0 +1,152 @@
+# Copyright 2024 Google Inc.
```
slevenick, line 1:
```suggestion
# Copyright 2025 Google Inc.
```

mmv1/products/managedkafka/Acl.yaml
slevenick: Does the API not accept requests without the etag set? In general etag should be an optional request parameter: https://google.aip.dev/154

If it's required on purpose you may be able to mark the etag field as output: true and change the type to `type: Fingerprint`. I believe that adds the field to every request

amfisher-404: We've implemented the etag on the resource level (not the request level) as specified in AIP-154. The etag is only required on Update requests, to meet the need for consistency/concurrency control: https://screenshot.googleplex.com/8NobwFxUKjof6y3. As such, there are no REQUIRED field annotations on `etag`; the API will return INVALID_ARGUMENT if the etag in UpdateAcl is missing.

Thanks for the suggestion. Marking etag as output:true with type:Fingerprint achieves what we're after:
* Users can't provide the field themselves: https://paste.googleplex.com/6450133052686336
* Its value is persisted from the API and sent on subsequent UpdateAcl requests: https://paste.googleplex.com/6546873382600704

slevenick (approved):
Thanks!

PR #14035, opened 2025-05-20
tgc: Support IAM Deny Policies

jalseth:
```release-note:enhancement
tgc: Support IAM Deny Policies
```

### DIFF
--- a/mmv1/third_party/tgc/tests/data/example_iam_deny_policy.json
+++ b/mmv1/third_party/tgc/tests/data/example_iam_deny_policy.json
@@ -1,0 +1,41 @@
+[
+  {
+    "name": "//iam.googleapis.com/policies/cloudresourcemanager.googleapis.com%2Fprojects%2F{{.Provider.project}}/denypolicies/my-deny-policy",
+    "asset_type": "iam.googleapis.com/DenyPolicy",
+    "ancestry_path": "{{.Ancestry}}/project/{{.Provider.project}}",
+    "resource": {
+      "version": "v2",
+      "discovery_document_uri": "https://www.googleapis.com/discovery/v1/apis/iam/v2/rest",
+      "discovery_name": "DenyPolicy",
+      "parent": "//cloudresourcemanager.googleapis.com/projects/{{.Provider.project}}",
+      "data": {
+        "displayName": "A deny rule",
+        "rules": [
+            {
+                "description": "First rule",
+                "denyRule": {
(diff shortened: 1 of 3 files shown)


mmv1/third_party/tgc/tests/data/example_iam_deny_policy.json
```diff
@@ -0,0 +1,41 @@
+[
+  {
+    "name": "//iam.googleapis.com/policies/cloudresourcemanager.googleapis.com%2Fprojects%2F{{.Provider.project}}/denypolicies/my-deny-policy",
+    "asset_type": "iam.googleapis.com/DenyPolicy",
```
zli82016, line 4: The asset type should be [iam.googleapis.com/PolicyV2](https://cloud.google.com/iam/docs/reference/rest/v2/policies). 

Maybe the generated asset type in `iam2.ResourceConverterIAM2DenyPolicy()` is wrong.

jalseth: You're right, the CAI name is PolicyV2, it's just more commonly referred to as a DenyPolicy because V2 IAM policies can only be configured as deny policies. I also realized the converter is not setting the `name` field like it should.

What is the path forward? Does this require a handwritten converter?

zli82016: What should the `name` field be?

jalseth: Looking at a real CAI export, `name` should be `policies/cloudresourcemanager.googleapis.com%2Fprojects%2F{{project_number}}/denypolicies/{{tf_name_field}}` but in a TF plan the project number might not be known so using project ID instead is fine. There is also a `kind` which should be `DenyPolicy` but I don't care about that field for my use case since all V2 policies are of DenyPolicy kind.

zli82016: When generating the cai asset type, the resource name is used.
https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/templates/tgc/resource_converter.go.tmpl#L50


We can add another field `CaiName` or `TgcName` to https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/api/resource.go#L39. 


If CaiName is provided, it is used to generate the cai asset type. Otherwise, the resource name is used.

PR #14046, opened 2025-05-20
Create a terraform resource for the Fleet RBACRolebindingActuation Feature for supporting custom roles in scope RBACRoleBindings

tonybayvas:

```release-note:REPLACEME

```


shumiao (approved):

c2thorn:

mmv1/templates/terraform/pre_delete/gkehub_existing_feature.go.tmpl
```diff
@@ -0,0 +1,5 @@
+// Make sure that mandatory/default features continue to exist.
```
ScottSuarez, line 1: Why is pre-delete necessary ?  If this is pre-existing/mandatory wouldn't we not be creating in the first place?

tonybayvas: That also makes sense, but the downside is that if more mandatory/default features are added to the if statement, that requires adding new logic for each feature's behavior since the specs are different. Since the pre_create already does an overwriting update if called after a destroy, would it be okay to keep it as is?

ScottSuarez: I am not sure this makes sense in the terraform world. If I am a client I would expect terraform destroy to unset any thing my resource created. If a delete and do not recreate then that is state that has persisted.

tonybayvas: I've updated the pre_delete logic to clear the state and verified locally that it works, please take a look

mmv1/templates/terraform/pre_delete/gkehub_existing_feature.go.tmpl
```diff
@@ -0,0 +1,41 @@
...
+          if err := d.Set("spec", spec); err != nil {
+            return err
+          }
+          return resourceGKEHub2FeatureUpdate(d, meta)
+        }
+      }
+    }
+    return nil
+  }
+}
```
ScottSuarez, line 41:
```suggestion
// Special handling for the mandatory 'rbacrolebindingactuation' feature.
// Instead of deleting it, we reset it to a default state by sending a PATCH request.
if d.Get("name").(string) == "rbacrolebindingactuation" {
	log.Printf("[DEBUG] Mandatory feature 'rbacrolebindingactuation' detected. Resetting instead of deleting.")

	patchUrl, err := tpgresource.ReplaceVarsForId(d, config, "{{"{{"}}GKEHub2BasePath{{"}}"}}projects/{{"{{"}}project{{"}}"}}/locations/{{"{{"}}location{{"}}"}}/features/{{"{{"}}name{{"}}"}}")
	if err != nil {
		return err
	}

	// Construct the request body to clear the desired field.
	obj := map[string]interface{}{
		"spec": map[string]interface{}{
			"rbacrolebindingactuation": map[string]interface{}{
				"allowedCustomRoles": []string{},
			},
		},
	}

	// A specific updateMask is required for a PATCH request.
	updateMask := "spec.rbacrolebindingactuation.allowedCustomRoles"
	url, err := transport_tpg.AddQueryParams(patchUrl, map[string]string{"updateMask": updateMask})
	if err != nil {
		return err
	}

	log.Printf("[DEBUG] Sending PATCH to reset Feature %q: %#v", d.Id(), obj)

	// Send the raw PATCH request.
	res, err := transport_tpg.SendRequest(transport_tpg.SendRequestOptions{
		Config:    config,
		Method:    "PATCH",
		Project:   billingProject,
		RawURL:    url,
		UserAgent: userAgent,
		Body:      obj,
		Timeout:   d.Timeout(schema.TimeoutDelete), // Use the delete timeout for this reset operation.
		Headers:   headers,
	})
	if err != nil {
		return fmt.Errorf("error resetting Feature %q: %s", d.Id(), err)
	}

	// Wait for the long-running operation to complete.
	err = GKEHub2OperationWaitTime(
		config, res, tpgresource.GetResourceNameFromSelfLink(project), "Resetting Feature", userAgent,
		d.Timeout(schema.TimeoutDelete))

	if err != nil {
		return fmt.Errorf("error waiting to reset Feature %q: %s", d.Id(), err)
	}

	log.Printf("[DEBUG] Finished resetting Feature %q", d.Id())
	return nil
}
```

I would prefer we just unset this specific field instead of calling update function.

ScottSuarez (changes requested):

mmv1/templates/terraform/pre_delete/gkehub_existing_feature.go.tmpl
tonybayvas: Thanks for the detailed fixes, I have applied and tested it

ScottSuarez (approved):

PR #14049, opened 2025-05-21
Fix issue on hubs in google_beyondcorp_security_gateway

shuyama1:

```release-note:bug
beyondcorp: fixed the issue where `hubs.internet_gateway.assigned_ips` was not populated correctly in the `google_beyondcorp_security_gateway` resource
```

### DIFF
--- a/mmv1/products/beyondcorp/SecurityGateway.yaml
+++ b/mmv1/products/beyondcorp/SecurityGateway.yaml
@@ -114,7 +114,6 @@ properties:
         - name: internetGateway
           type: NestedObject
           description: Internet Gateway configuration.
-          default_from_api: true
           properties:
             - name: assignedIps
               type: Array
--- a/mmv1/products/beyondcorp/SecurityGateway.yaml
+++ b/mmv1/products/beyondcorp/SecurityGateway.yaml
@@ -54,6 +54,8 @@ async:
     resource_inside_response: true
   include_project: false
 autogen_status: U2VjdXJpdHlHYXRld2F5
+custom_code:
+  constants: 'templates/terraform/constants/beyondcorp_security_gateway.go.tmpl'
 parameters:
   - name: location
     type: String
@@ -104,6 +106,7 @@ properties:
       as a key.
     key_name: region
     key_description: The region to deploy the hub in.
+    set_hash_func: 'beyondcorpSecurityGatewayHubsHash'
     value_type:
       name: Hub
       type: NestedObject
--- a/mmv1/products/beyondcorp/SecurityGateway.yaml
+++ b/mmv1/products/beyondcorp/SecurityGateway.yaml
@@ -111,6 +111,7 @@ properties:
         - name: internetGateway
           type: NestedObject
           description: Internet Gateway configuration.
+          default_from_api: true
           properties:
             - name: assignedIps
               type: Array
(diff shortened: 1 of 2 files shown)


shuyama1:

mmv1/products/beyondcorp/SecurityGateway.yaml
```diff
@@ -104,15 +106,17 @@ properties:
       as a key.
     key_name: region
     key_description: The region to deploy the hub in.
+    set_hash_func: 'beyondcorpSecurityGatewayHubsHash'
```
melinath, line 109: It looks like this just returns the key name. If I remember correctly, I think this will suppress diffs as long as the set_hash_func's return value doesn't change - which means that changing the value for internet_gateway would be ignored unless the region also changed?

shuyama1: The `internet_gateway` block currently contains only the output-only fields `assigned_ips` which is not configurable by users at this point, but service team plan to add configurable fields to the block in the future. 
So yes, for now, we may want to suppress the following diff unless the region is changed, and will need to update the hash function once those non-computed fields are implemented.

```
# google_beyondcorp_security_gateway.example will be updated in-place
          ~ resource "google_beyondcorp_security_gateway" "example" {
                id                  = "projects/ci-test-project-188019/locations/global/securityGateways/defaulttc0ljpg4em"
                name                = "projects/ci-test-project-188019/locations/global/securityGateways/defaulttc0ljpg4em"
                # (8 unchanged attributes hidden)
        
              - hubs {
                  - region = "us-central1" -> null
        
                  - internet_gateway {
                      - assigned_ips = [
                          - "34.68.0.163",
                          - "34.173.76.190",
                        ] -> null
                    }
                }
              + hubs {
                  + region = "us-central1"
        
                  + internet_gateway (known after apply)
                }
            }
```

I'm not certain this is the best path forward. Please let me know your thoughts on this approach, and if there's alternative solutions. Thanks!

mmv1/products/beyondcorp/SecurityGateway.yaml
```diff
@@ -104,15 +106,17 @@ properties:
...
     value_type:
       name: Hub
       type: NestedObject
       properties:
-        - name: internet_gateway
+        - name: internetGateway
           type: NestedObject
           description: Internet Gateway configuration.
+          default_from_api: true
```
shuyama1, line 117: I think we can remove `default_from_api` here(?). I initially added it to fix the diff, but it didn't' work

shuyama1, line 117:
```suggestion
```

shuyama1: nope, looks like it's still needed

melinath (approved):
LGTM. I'm a little confused why the data source test is passing at all, but the API is definitely returning a value for internetGateway.  default_from_api seems reasonable given there will be settable fields in the future. 🤷

melinath:
oh right, the test is passing today (prior to this change) because both the data source and resource are failing to read the `internetGateway` field from the API.

shuyama1:
> oh right, the test is passing today (prior to this change) because both the data source and resource are failing to read the `internetGateway` field from the API.

Yes, that's correct, the fields were not read correctly. I started this PR believing it was a straightforward fix for the field names, but then found out it was more complicated than I thought

PR #14052, opened 2025-05-21
allow sending propagatedConnectionLimit zero value for ServiceAttachment

steved:
When creating or modifying `google_compute_service_attachment`, the `propagated_connection_limit` cannot be set to `0` and the default is used instead. This adds a new `send_propagated_connection_limit_if_zero` flag that, when enabled, will send the zero value.

https://github.com/hashicorp/terraform-provider-google/issues/22939

```release-note:enhancement
compute: added `send_propagated_connection_limit_if_zero` to `google_compute_service_attachment` to resolve an issue where `propagated_connection_limit` were not working for 0 value previously. Now setting `send_propagated_connection_limit_if_zero = true` will send `propagated_connection_limit = 0` when it's unset or set to `0`.
```

### DIFF
--- a/mmv1/products/compute/ServiceAttachment.yaml
+++ b/mmv1/products/compute/ServiceAttachment.yaml
@@ -38,6 +38,7 @@ async:
 custom_code:
   constants: 'templates/terraform/constants/compute_service_attachment.go.tmpl'
   update_encoder: 'templates/terraform/update_encoder/compute_service_attachment.go.tmpl'
+  encoder: 'templates/terraform/encoders/compute_service_attachment.go.tmpl'
 sweeper:
   url_substitutions:
     - region: "us-west2"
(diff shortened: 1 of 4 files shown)


steved:
> Sorry for the delayed review, I was out last week. Would it be possible to add a new test, or modify an existing one, to explicitly specify `propagated_connection_limit = 0`? Thank you!
> 
> https://googlecloudplatform.github.io/magic-modules/test/test/#add-resource-tests

No worries, gave it a shot in d4fdc5e3c6ba93665a833c782016f55ee1810cf6. Unfortunately, seems I ran into an edge case where we want the default from the API (`default_from_api`) but only want to send the 0 value if it's set rather than null / unset (somewhat discussed in https://github.com/hashicorp/terraform-provider-google/issues/19847). I set the `default_value = 250` but I'm not sure if that's the correct solution.

shuyama1:
The previous build failed, likely due to a transient server error. Triggering a rerun.

/gcbrun

steved:
> I think this is a breaking change that we should prevent in minor release. We can introduce a virtual field `send_propagated_connection_limit_if_zero` (default to false) to control if we want to explicitly send 0 value for this field. Similar case like #11410 and #10717
> 
> Please let me know if you have any questions! Thank you!

Thanks! Gave it a shot, let me know what you think.

steved:
Just cleaned up the documentation a bit in 77e6d9035ed5612ed1305da2bd4fb8300739ddca, no major changes.

shuyama1 (approved):
LGTM! I've modified the changelog a bit. Please let me know if that makes sense. Thank you!

PR #14057, opened 2025-05-21
Suppress the diff for false value of force_attach field

zli82016:
fixes https://github.com/hashicorp/terraform-provider-google/issues/22950

The immutable `force_attach` is added to the `attached_disk` with the default value false.

This PR suppresses the permadiff between false and empty value for this field.

As this field is immutable, the resource recreation is expected when adding `force_attach = true` to the config.

```release-note:bug
compute: fixed forced instance recreation when adding a `attached_disk` with `force_attach` being `false` to `google_compute_instance` 
```


melinath (approved):
tested locally, prevents the destructive diff.

PR #14059, opened 2025-05-21
Extend breaking change detector to work on new required and new optional with default

ScottSuarez:
closes b/300515447

```release-note:none

```


tools/diff-processor/breaking_changes/breaking_changes.go
```diff
@@ -43,9 +43,10 @@ func ComputeBreakingChanges(schemaDiff diff.SchemaDiff) []BreakingChange {
 			}
 		}
 
+		isPreExistingResource := !schemaDiff.IsNewResource(resource)
 		for field, fieldDiff := range resourceDiff.Fields {
 			for _, rule := range FieldDiffRules {
-				for _, message := range rule.Messages(resource, field, fieldDiff) {
+				for _, message := range rule.Messages(resource, field, fieldDiff, isPreExistingResource) {
```
melinath, line 49: Rather than passing a bool isPreExistingResource, what would you think of passing the entire resourceDiff object? That would be a simple change that would give us more flexibility for potentially introducing more complex logic in the future.

tools/diff-processor/breaking_changes/field_diff.go
```diff
@@ -214,3 +217,43 @@ func FieldRemovingDiffSuppressMessages(resource, field string, fieldDiff diff.Fi
...
+	}
+	return nil
+}
+
+var FieldNewOptionalFieldWithDefault = FieldDiffRule{
+	Identifier: "no-new-optional-default",
+	Messages:   FieldNewOptionalFieldWithDefaultMessages,
+}
+
+func FieldNewOptionalFieldWithDefaultMessages(resource, field string, fieldDiff diff.FieldDiff, isPreExistingResource bool) []string {
```
melinath, line 246: I'm not convinced this is inherently a breaking change. However, we've had a couple of patch releases recently due to optional fields with defaults that (crucially) are ForceNew:

- https://github.com/hashicorp/terraform-provider-google-beta/pull/9907/files#diff-37939071c02493a83384db446ab712613747c834878940780f604092145ac297R157-R161
- https://github.com/hashicorp/terraform-provider-google/pull/22801/files#diff-e34b264d37ad2b66100c88995746ce1ea3e6cd81ef8d5b68ae938a0e04a32316R554-R559

Note that optional + default + forcenew can require a patch release even if it's a nested field. In fact, both of these changes were to nested fields.

I think even optional + default + forcenew is not _necessarily_ a breaking change (depending on API behavior / diff suppression) but the consequences when it is breaking are severe enough that we should always flag it for additional attention.

ScottSuarez: > optional + default + forcenew is all client side, regardless of API it would be breaking, no?

melinath: I don't think so, not necessarily. It's only breaking if it triggers a diff. If a field has a consistent API-side default value, we currently recommend setting that default value on the field to avoid a diff. The problem in both of these cases was that the API returned an empty value if the field was unset, which triggered a diff and therefore recreation. (We can also work around the API behavior on the client side with a flattener or diff suppression.)

ScottSuarez: I see, I think we should ignore or clarify this scenario and still flag it. Allow reviewer to override

melinath: yes, I think flagging all optional + default + forcenew fields and requiring an override seems reasonable.

melinath: optional + default without forcenew is pretty common and less harmful if there's an issue, so I think we shouldn't flag that.

ScottSuarez: agreed

tools/diff-processor/breaking_changes/breaking_changes.go
ScottSuarez: changed to pass interface, would rather not pass entire object, difficult to unit test

ScottSuarez:
checks are failing due to noop virtual fields (will be removed before checkin)

docs/content/breaking-changes/breaking-changes.md
```diff
@@ -60,7 +60,8 @@ For more information, see
   * Between complex types like changing a List to a Set.
   * Changing the field type between primitive and complex data
     types is not possible. For this scenario, field renames are preferred.
-* <a name="field-optional-to-required"></a> Making an optional field required or adding a new required field
+* <a name="field-optional-to-required"></a> Making an optional field required
+* <a name="no-new-required"></a> Adding a required field to a pre-existing resource or object
```
melinath, line 64: "object" isn't clear - maybe something like:

```suggestion
* <a name="no-new-required"></a> Adding a required field to a pre-existing resource at any level of nesting, unless it is being added at the same time as an optional ancestor
```

docs/content/breaking-changes/breaking-changes.md
```diff
@@ -81,13 +82,17 @@ For more information, see
     if the change will destroy and recreate the resource due to changing an immutable value.
     Default changes in the provider are comparable in impact to default changes in an API,
     and modifying examples and modules may achieve the intended effect with a smaller blast radius.
+* <a name="no-new-optional-default"></a> Adding an optional field with a default value and force new to a pre-existing resource or object
```
melinath, line 85:
```suggestion
* <a name="no-new-optional-default"></a> Adding an optional field with a default value and force new to a pre-existing resource at any level of nesting, unless it is being added at the same time as an optional ancestor
```

docs/content/breaking-changes/breaking-changes.md
```diff
@@ -81,13 +82,17 @@ For more information, see
...
+  * This can be allowed if there is a confirmed API-level default that matches the schema default
+  * Please work with your reviewer and ensure this scenario is debuged carefully
```
melinath, line 87:
```suggestion
  * Please work with your reviewer and ensure this scenario is debugged carefully to avoid a destructive permadiff
```

tools/diff-processor/diff/diff.go
```diff
@@ -313,3 +326,42 @@ func setKey(set FieldSet) string {
...
+		rcd = rd.ResourceConfig
+	}
+	if rcd.Old == nil && rcd.New != nil {
+		return true
+	}
+	return false
+}
+
+// IsFieldInNewNestedStructure determines if a field is part of a completely new nested structure
+func (sd SchemaDiff) IsFieldInNewNestedStructure(resourceName, fieldPath string) bool {
```
melinath, line 342: Rather than checking whether the parent is new, we need to check whether there is at least one new ancestor (possibly a grandparent) that was optional. (I _think_ a new O+C with a nested required field is non-breaking; optional without computed definitely is. But if all the ancestors are required, then this will be a breaking change.)

Having a helper function for this makes a lot of sense since it's a complex operation that would be easy to get wrong - but I don't know that it makes sense as a method on the SchemaDiff object, since it should operate on a specific resource diff + field name.

tools/diff-processor/breaking_changes/field_diff.go
```diff
@@ -13,13 +13,15 @@ import (
 // regarding field attribute changes
 type FieldDiffRule struct {
 	Identifier string
-	Messages   func(resource, field string, fieldDiff diff.FieldDiff) []string
+	Messages   func(resource, field string, fieldDiff diff.FieldDiff, schemaDiff diff.SchemaDiffInterface) []string
```
melinath, line 16: to clarify - I didn't mean passing the entire schema diff here, just the resource diff. I don't think there would ever be a case where a field on one resource needs information about a field on a different resource.

melinath (changes requested):

tools/diff-processor/diff/diff.go
ScottSuarez: discussed offline

tools/diff-processor/breaking_changes/field_diff.go
ScottSuarez: simplified scope to use minimal interface for needed action

tools/diff-processor/breaking_changes/mock_schema_diff_test.go
```diff
@@ -0,0 +1,36 @@
+package breaking_changes
+
+// MockSchemaDiff implements the diff.SchemaDiff interface for testing
+type MockSchemaDiff struct {
```
melinath, line 4: I would prefer to not mock per go/choose-test-double?polyglot=go#prefer-real-implementations, but I'm not blocking on it.

melinath (approved):

tools/diff-processor/breaking_changes/mock_schema_diff_test.go
ScottSuarez: ack, this is checked upstream in another integration test. I would like to keep these simple and relevant to the logic they need to check.

PR #14063, opened 2025-05-22
Enable mmv1 unit tests to run on main and all paths

ScottSuarez:

```release-note:none

```


melinath (approved):

melinath (commented):
the lint checker is now passing with warnings, but the warnings will show up on the files tab (which is cool but maybe we should resolve those, and make the lint checker more strict in the future?) Maybe in separate PRs

PR #14067, opened 2025-05-22
Adding delegatingServiceAccount to output

palramanathan:
```release-note:enhancement
beyondcorp: added `delegatingServiceAccount` fields to `SecurityGateway` resource
```

### DIFF
--- a/mmv1/products/beyondcorp/SecurityGateway.yaml
+++ b/mmv1/products/beyondcorp/SecurityGateway.yaml
@@ -135,3 +135,8 @@ properties:
     type: String
     description: Identifier. Name of the resource.
     output: true
+  - name: delegatingServiceAccount
+    type: String
+    description: |-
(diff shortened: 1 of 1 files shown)


mmv1/products/beyondcorp/SecurityGateway.yaml
```diff
@@ -135,3 +135,8 @@ properties:
     type: String
     description: Identifier. Name of the resource.
     output: true
+  - name: delegatingServiceAccount
+    type: String
+    description: |-
+      Output only. Service account used for operations that involve resources in consumer projects.
```
BBBmau, line 141: no need to include `Output only.` in description since having the `output_only: true` as part of the field automatically adds it as part of the generation in the docs

palramanathan: Done

PR #14070, opened 2025-05-22
Private CA - Differentiate unset and default values for is_ca/max_issuer_path_length in Certificate Templates

rohanchawla23:
Previously, unset and default values for the `is_ca` and `max_issuer_path_length` fields were treated as exactly the same for Certificate Templates, though the API service does distinguish between the two. This led to an issue where it was not possible to use a Certificate Template to create a CA with a `max_issuer_path_length` of 0. To address this, two new fields have been added to the Certificate Template interface to distinguish between the unset and default values for these two fields. The name_constraints field has also been added to the Certificate Template resource, to enable compatibility with a new expander. This expander contains the logic for using the two new fields to distinguish between the unset and default values. Note that the pre-existing expander for X509 Configs could not be used, as the behavior there would not be backwards compatible with the current behavior of CertificateTemplate when the `is_ca` field is false.

I acknowledge that I have:

- [x] Searched through the [issue tracker](https://github.com/hashicorp/terraform-provider-google/issues) for an open issue that this either resolves or contributes to, commented on it to claim it, and written "fixes {url}" or "part of {url}" in this PR description. If there were no relevant open issues, I opened one and commented that I would like to work on it (not necessary for very small changes).
- [x] Ensured that all new fields I added that can be set by a user appear in at least one [example](https://github.com/GoogleCloudPlatform/magic-modules/tree/main/mmv1/templates/terraform/examples) (for generated resources) or [third_party test](https://github.com/GoogleCloudPlatform/magic-modules/tree/main/mmv1/third_party/terraform/tests) (for handwritten resources or update tests).
- [x] [Generated Terraform providers](https://github.com/GoogleCloudPlatform/magic-modules/blob/main/docs/content/docs/getting-started/generate-providers.md), and ran [make test and make lint](https://googlecloudplatform.github.io/magic-modules/docs/getting-started/run-provider-tests/#run-unit-tests) in the generated providers to ensure it passes unit and linter tests.
- [x] [Ran](https://github.com/GoogleCloudPlatform/magic-modules/blob/main/docs/content/docs/getting-started/run-provider-tests.md) relevant acceptance tests using my own Google Cloud project and credentials (If the acceptance tests do not yet pass or you are unable to run them, please let your reviewer know).
- [x] Read the [Release Notes Guide](https://googlecloudplatform.github.io/magic-modules/contribute/release-notes/) before writing my release note below.

```release-note:enhancement
privateca: added support for setting default values for basic constraints for `google_privateca_certificate_template` via the `null_ca` and `zero_max_issuer_path_length` fields. also added `name_constraints` field for `google_privateca_certificate_template`.
```

### DIFF
--- a/mmv1/products/privateca/CertificateTemplate.yaml
+++ b/mmv1/products/privateca/CertificateTemplate.yaml
@@ -69,6 +69,8 @@ properties:
   - name: 'predefinedValues'
     type: NestedObject
     description: Optional. A set of X.509 values that will be applied to all issued certificates that use this template. If the certificate request includes conflicting values for the same properties, they will be overwritten by the values defined here. If the issuing CaPool's IssuancePolicy defines conflicting baseline_values for the same properties, the certificate issuance request will fail.
+    custom_flatten: 'templates/terraform/custom_flatten/privateca_certificate_template_509_config.go.tmpl'
+    custom_expand: 'templates/terraform/custom_expand/privateca_certificate_template_509_config.go.tmpl'
     properties:
       - name: 'keyUsage'
         type: NestedObject
@@ -188,6 +190,7 @@ properties:
               If both `is_ca` and `null_ca` are unset, the "CA" in Basic Constraints extension will be set to false.
               Note that the behavior when `is_ca = false` for this resource is different from the behavior in the Certificate Authority, Certificate and CaPool resources.
             url_param_only: true
+            send_empty_value: true
           - name: 'maxIssuerPathLength'
             type: Integer
             description: |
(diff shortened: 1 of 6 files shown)


mmv1/products/privateca/CertificateTemplate.yaml
```diff
@@ -178,11 +185,29 @@ properties:
         properties:
           - name: 'isCa'
             type: Boolean
-            description: Optional. Refers to the "CA" X.509 extension, which is a boolean value. When this value is missing, the extension will be omitted from the CA certificate.
+            description: |
+              Optional. Refers to the "CA" X.509 extension, which is a boolean value. When this value is true, the "CA" in Basic Constraints extension will be set to true.
+            send_empty_value: true
+          - name: 'nullCa'
```
SirGitsalot, line 191: I'm *pretty* sure this PR is good to go but want to make sure my understanding is correct: the current behavior is that a basic constraints extension is always created; setting`isCa` to `true` creates a BC extension with `ca: true`, and setting it to `false` or omitting it entirely creates a BC extension with `ca: false` (and as a result, matching the behavior of other cert resources that use `nonCa` would cause existing `google_privateca_certificate_template` configs with `isCa = false` to stop creating BC extensions rather than creating them with `ca: false`)

rohanchawla23: Yes, that's correct!

SirGitsalot (approved):

PR #14071, opened 2025-05-22
fix(tgc): Correctly handle ancestry for resources with the "parent" f…

jalseth:
…ield

Previously, ancestrymanager only tried to use the project, which breaks resouces that use the "parent" field.

```release-note:bug
fix(tgc): Correctly handle ancestry for resources with the "parent" field
```


BBBmau (approved):
this makes sense to do as each ID is being set prior to the `switch cai.Type` it also matches the behavior of all the other cases in the switch + tests are passing.

@zli82016 I'm approving on my end but would like a second opinion as you're more familiar with the TGC repo before merging

mmv1/third_party/tgc/ancestrymanager/ancestrymanager_test.go
```diff
@@ -500,6 +511,36 @@ func TestGetAncestors(t *testing.T) {
...
+			asset: &resources.Asset{
+				Type: "orgpolicy.googleapis.com/CustomConstraint",
+			},
+			want:       []string{"organizations/12345"},
+			wantParent: "//cloudresourcemanager.googleapis.com/organizations/12345",
+		},
+		{
+			name: "Folder-level CuOP set with parent field",
+			data: tfdata.NewFakeResourceData(
+				"google_org_policy_custom_constraint",
```
zli82016, line 532: `google_org_policy_custom_constraint` is not an good resource to test `Folder-level` CuOP set with parent field, as its parent can only be `organization`. Are there any other resources to replace it?

jalseth: `google_folder` is a good candidate, and there's already test cases for that. In fact, this change allows us to remove the switch cases for the folder and custom IAM role resources. PTAL.

zli82016 (approved):

PR #14076, opened 2025-05-23
`mmv1/identity`: support `identities` schema generation in `tpgtools/templates/resource.go.tmpl`

BBBmau:
> [!IMPORTANT]
> This PR is opened on `mmv1/identity-generation`, serving as the first PR being opened to support ResourceIdentity in MMv1 with more PRs to follow. The `mmv1/identity-generation` branch is meant to be merged upon the merging of `FEATURE-BRANCH-resource-identity`

Only adds support for resources being generated from `tpgtools/templates/resource.go.tmpl`, below is the generation result on `ResourceApiKeysKey` as well as how the identity matches the regex import

image

image

- [X] IdentitySchema Generated
- [ ] IdentitySchemaVersion

```release-note:none

```

### DIFF
--- a/tpgtools/id.go
+++ b/tpgtools/id.go
@@ -48,17 +48,17 @@ func PatternToRegex(s string, allowForwardSlash bool) string {
 }
 
 // Finds the correct resource id based on the schema and any overrides. Returns whether a custom ID override was used.
-func findResourceID(schema *openapi.Schema, overrides Overrides, location string) (string, bool, error) {
+func findResourceID(schema *openapi.Schema, overrides Overrides, location string) (identities []Property, id string, customID bool, err error) {
 	id, ok := schema.Extension["x-dcl-id"].(string)
 	if !ok {
-		return "", false, fmt.Errorf("Malformed or missing x-dcl-id: %v", schema.Extension["x-dcl-id"])
+		return nil, "", false, fmt.Errorf("Malformed or missing x-dcl-id: %v", schema.Extension["x-dcl-id"])
 	}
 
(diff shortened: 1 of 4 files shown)


BBBmau:
Closing this since upon further investigation, if we want to support identity schema in DCL resources we would need to include them in the [DCL repo](https://github.com/GoogleCloudPlatform/declarative-resource-client-library?tab=readme-ov-file) which is where the resource information is obtained for generation.

c2thorn (commented):
Is this something that needs to be supported by DCL at all? Do you plan on adding a test with existing DCL resources? Are we going to backfill?

c2thorn:
We don't allow any new resources in DCL and we are slowly migrating the existing ones

BBBmau:
> Is this something that needs to be supported by DCL at all? Do you plan on adding a test with existing DCL resources? Are we going to backfill?

We could consider not supporting it since they're transitioning to MMv1, though what's the timeline for that?

 Tests for the DCL resources would be added in a separate PR that i've opened here:
- https://github.com/GoogleCloudPlatform/magic-modules/pull/14150

Since there are no plans for new DCL resources we wouldn't have to worry about backfilling new DCL resources.

c2thorn:
> We could consider not supporting it since they're transitioning to MMv1, though what's the timeline for that?

No clear timeline, it missed the prioritization line last planning, but should be considered for the next one. I'd be really disappointed if we haven't migrated mostly off by next year. Depends on what your goals for coverage for the identities schema is

c2thorn (approved):

PR #13561, opened 2025-04-07
Feature gap: Add `backends.preference`, `maxStreamDuration` and `cdnPolicy.requestCoalescing` to BackendService

Cezarus27:
**Release Note Template for Downstream PRs (will be copied)**

See [Write release notes](https://googlecloudplatform.github.io/magic-modules/contribute/release-notes/) for guidance.

```release-note:enhancement
compute: added `log_config.optional_mode`, `log_config.optional_fields`, `backend.preference`, `max_stream_duration` and `cdn_policy.request_coalescing` fields to `google_compute_backend_service` resource
```

PR #14083, opened 2025-05-23
compute: Add example for `log_config` `optional_mode` and `optional_fields` in `google_compute_backend_service`.

efimki:

```release-note:none
```

Note that these fields were added in #13561 but included example of optional fields like "serverInstance", "serverGkeDetails.cluster", "serverGkeDetails.pod.podNamespace" are not allowed and would fail the test.


efimki:
The failing tests seem unrelated. How can I trigger the retest?

efimki:
Thank you!

PR #14085, opened 2025-05-23
Remove the hard coded timeout while doing bucket deletion.

googlyrahman:
Remove the hardcoded timeout used during bucket deletion, and instead rely on the resource default `destroy_timeout` in the google_storage_bucket resource. Currently, the actual deletion of the cache occurs around 62–63 minutes, but we've been using a hardcoded 80-minute timeout as a best-effort cleanup buffer.

```release-note:enhancement
storage: removed the hardcoded 80m timeout used during `google_storage_bucket` deletion when removing an anywhere cache, polling instead. This should speed up deletion in these cases.
```

### DIFF
--- a/mmv1/third_party/terraform/services/storage/resource_storage_bucket.go.tmpl
+++ b/mmv1/third_party/terraform/services/storage/resource_storage_bucket.go.tmpl
@@ -710,71 +710,63 @@ func getAnywhereCacheListResult(d *schema.ResourceData, config *transport_tpg.Co
 }
 
 func deleteAnywhereCacheIfAny(d *schema.ResourceData, config *transport_tpg.Config) error {
-	// Get the initial list of Anywhere Caches
-	cacheList, err := getAnywhereCacheListResult(d, config)
-	if err != nil {
-		return err
-	}
-
-	// If no cache exists initially, return early
-	if len(cacheList) == 0 {
-		return nil
-	}
-
-	// Iterate over each object in the resource list
-	for _, item := range cacheList {
-		// Ensure the item is a map
-		obj, ok := item.(map[string]interface{})
-		if !ok {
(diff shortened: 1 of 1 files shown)


kautikdk:
>Not exactly, 50 API requests are like peanuts.

That's very subjective though. 

>We use a similar method when dealing with LROs, where we poll every minute for up to four hours, IIRC.

If there is no backing customer issue or request, I believe we should discuss it internally first? Let me know if it's already done and agreed.

kautikdk (approved):
Since the approach is already been discussed with stockholders, LGTM from me. @rileykarson can review if they find anything concerning.

rileykarson (commented):
Code changes look good- just double checking before LGTM+merge, do you have the exact error returned if the anywhere cache deletion is still in progress? https://github.com/googlyrahman/magic-modules/blob/8425183101d24c9db7969d8de4ecfd6ce022d420/mmv1/third_party/terraform/services/storage/resource_storage_bucket.go.tmpl#L1263-L1265 will retry provided it's a 429.

I kicked off a run of `TestAccStorageAnywhereCache_update ` in https://hashicorp.teamcity.com/buildConfiguration/TerraformProviders_GoogleCloud_GOOGLE_BETA_MMUPSTREAMTESTS_GOOGLEBETA_PACKAGE_STORAGE/386801 (restricted access, unfortunately) to see if that triggers the error and retries it correctly.

googlyrahman:
@rileykarson Addressed comments.

googlyrahman:
@rileykarson any update here?

rileykarson (approved):
LGTM, thanks!

Sorry for the delay- I wanted to run through the flow locally w/ an externally added anywhereCache to confirm the behaviour, since our tests don't cover that.

PR #14086, opened 2025-05-23
feat: added `google_compute_snapshot_settings` resource

ramonvermeulen:
Closes https://github.com/hashicorp/terraform-provider-google/issues/22799

```release-note:new-resource
`google_compute_snapshot_settings`
```

### DIFF
--- a/mmv1/third_party/terraform/services/compute/resource_compute_snapshot_settings_test.go
+++ b/mmv1/third_party/terraform/services/compute/resource_compute_snapshot_settings_test.go
@@ -22,6 +22,9 @@ func TestAccComputeSnapshotSettings_snapshotSettings_update(t *testing.T) {
 	acctest.VcrTest(t, resource.TestCase{
 		PreCheck:                 func() { acctest.AccTestPreCheck(t) },
 		ProtoV5ProviderFactories: acctest.ProtoV5ProviderFactories(t),
+		ExternalProviders: map[string]resource.ExternalProvider{
+			"time": {},
+		},
 		Steps: []resource.TestStep{
 			{
 				Config: testAccComputeSnapshotSettings_snapshotSettings_basic(context),


ramonvermeulen:
Acceptance test locally:

```
❯ make testacc TEST=./google/services/compute TESTARGS='-run=TestAccComputeSnapshotSettings_snapshotSettingsSpecificLocationsExample'
TF_ACC=1 TF_SCHEMA_PANIC_ON_ERROR=1 go test ./google/services/compute -v -run=TestAccComputeSnapshotSettings_snapshotSettingsSpecificLocationsExample -timeout 240m -ldflags="-X=github.com/hashicorp/terraform-provider-google/version.ProviderVersion=acc"
=== RUN   TestAccComputeSnapshotSettings_snapshotSettingsSpecificLocationsExample
=== PAUSE TestAccComputeSnapshotSettings_snapshotSettingsSpecificLocationsExample
=== CONT  TestAccComputeSnapshotSettings_snapshotSettingsSpecificLocationsExample
--- PASS: TestAccComputeSnapshotSettings_snapshotSettingsSpecificLocationsExample (5.11s)
PASS
ok      github.com/hashicorp/terraform-provider-google/google/services/compute  5.841s
```

mmv1/templates/terraform/examples/snapshot_settings_specific_locations.tf.tmpl
```diff
@@ -0,0 +1,10 @@
+resource "google_compute_snapshot_settings" "{{$.PrimaryResourceId}}" {
+    project   = "{{index $.TestEnvVars "project"}}"
```
melinath, line 2: here & in the other tests, the test project will be set by default & doesn't need to be set explicitly.

mmv1/third_party/terraform/services/compute/resource_compute_snapshot_settings_test.go
```diff
@@ -0,0 +1,74 @@
...
+	"github.com/hashicorp/terraform-plugin-testing/plancheck"
+	"testing"
+
+	"github.com/hashicorp/terraform-plugin-testing/helper/resource"
+
+	"github.com/hashicorp/terraform-provider-google/google/acctest"
+	"github.com/hashicorp/terraform-provider-google/google/envvar"
+)
+
+func TestAccComputeSnapshotSettings_snapshotSettings_basic(t *testing.T) {
```
melinath, line 13: nit: usually we call this `_update` to indicate it's an update test.

```suggestion
func TestAccComputeSnapshotSettings_snapshotSettings_update(t *testing.T) {
```

melinath (changes requested):
Two things:

1. It looks like this resource returns an operation rather than being synchronous. That means that you'll need to add an `async` block to the resource yaml: https://googlecloudplatform.github.io/magic-modules/reference/resource/#async
2. This is a singleton resource, so you'll need to worry about whether the tests conflict with each other or will have impact on other tests. It might make sense to exclude the test for the example and have the update test create a testing subproject.

mmv1/templates/terraform/examples/snapshot_settings_specific_locations.tf.tmpl
melinath: Project will need to be set if you decide you need a non-default project, though!

ramonvermeulen:
> Two things:
> 
> 1. It looks like this resource returns an operation rather than being synchronous. That means that you'll need to add an `async` block to the resource yaml: https://googlecloudplatform.github.io/magic-modules/reference/resource/#async
> 2. This is a singleton resource, so you'll need to worry about whether the tests conflict with each other or will have impact on other tests. It might make sense to exclude the test for the example and have the update test create a testing subproject.

Thanks for your review!

1. Good to know, I will take a look at this and see if I can implement this in the right way. Just for my understanding, since I see this quite often in the debug logs. Any resource where the API endpoint on a POST / PUT / PATCH etc. returns any kind of "Operation" should have this `async` block configured? So that after the initial request it polls the operation until it is `done`? (e.g. because some operations are not instant, provisioning a cluster or whatsoever). And what are the implications if you do not configure it? I guess that will lead to a sort of "fire and forget" situation where it doesn't get validated if the resource actually succeed in updating/creating.
2. My bad, probably figured this out myself if I ran all tests instead of just the one that I added. I will disable the example tests and let the update test run in a dedicated subproject.

ramonvermeulen:
@melinath 

Updated the PR with https://github.com/GoogleCloudPlatform/magic-modules/pull/14086/commits/445f006d569934a1d93165ad2a0f1901fd0d56f7

I only ran into the project quota on my personal GCP org, so I cannot run the acceptance tests locally.

mmv1/third_party/terraform/services/compute/resource_compute_snapshot_settings_test.go
```diff
@@ -0,0 +1,91 @@
...
+
+func testAccComputeSnapshotSettings_snapshotSettings_basic(context map[string]interface{}) string {
+	return acctest.Nprintf(`
+resource "google_project" "project" {
+  project_id      = "tf-test%{random_suffix}"
+  name            = "tf-test%{random_suffix}"
+  org_id          = "%{org_id}"
+  billing_account = "%{billing_account}"
+  deletion_policy = "DELETE"
+}
```
melinath, line 59: here & below - you'll need to enable the compute service for the new project:

```suggestion
}

resource "google_project_service" "compute" {
  project = google_project.project.project_id
  service = "compute.googleapis.com"
}

resource "time_sleep" "wait_120_seconds" {
  create_duration = "120s"
  depends_on = [google_project_service.compute]
}

```

mmv1/third_party/terraform/services/compute/resource_compute_snapshot_settings_test.go
```diff
@@ -0,0 +1,91 @@
...
+
+resource "google_compute_snapshot_settings" "tf_test_snapshot_settings" {
+    project   = google_project.project.project_id
+    storage_location {
+        policy    = "SPECIFIC_LOCATIONS"
+        locations {
+            name     = "us-central1"
+            location = "us-central1"
+        }
+    }
```
melinath, line 69:
```suggestion
    }
    depends_on = [time_sleep.wait_120_seconds]
```

PR #14088, opened 2025-05-23
Support Cloud SQL read pool

psung:
Adds support for:
* instance_type=READ_POOL_INSTANCE
* new node_count field

```release-note:enhancement
sql: added `node_count` field to `sql_database_instance` resource
sql: added new value `READ_POOL_INSTANCE` to the `instance_type` field of `sql_database_instance` resource
```

### DIFF
--- a/mmv1/third_party/terraform/services/sql/resource_sql_database_instance.go.tmpl
+++ b/mmv1/third_party/terraform/services/sql/resource_sql_database_instance.go.tmpl
@@ -315,6 +315,7 @@ settings.backup_configuration.enabled is set to true.
 For MySQL instances, ensure that settings.backup_configuration.binary_log_enabled is set to true.
 For Postgres instances, ensure that settings.backup_configuration.point_in_time_recovery_enabled
 is set to true. Defaults to ZONAL.`,
+							DiffSuppressFunc: availabilityTypeDiffSuppress,
 						},
 						"backup_configuration": {
 							Type:     schema.TypeList,
@@ -874,7 +875,15 @@ is set to true. Defaults to ZONAL.`,
 				Type:        schema.TypeString,
 				Computed:    true,
 				Optional:    true,
-				Description: `The type of the instance. The valid values are:- 'SQL_INSTANCE_TYPE_UNSPECIFIED', 'CLOUD_SQL_INSTANCE', 'ON_PREMISES_INSTANCE' and 'READ_REPLICA_INSTANCE'.`,
+				Description: `The type of the instance. The valid values are:- 'SQL_INSTANCE_TYPE_UNSPECIFIED', 'CLOUD_SQL_INSTANCE', 'ON_PREMISES_INSTANCE', 'READ_REPLICA_INSTANCE', and 'READ_POOL_INSTANCE'.`,
+			},
+
+			"node_count": {
+				Type:             schema.TypeInt,
+				Computed:         true,
+				Optional:         true,
+				Description:      `For a read pool instance, the number of nodes in the read pool.`,
+				DiffSuppressFunc: nodeCountDiffSuppress,
 			},
 
 			"replica_configuration": {
@@ -1251,6 +1260,10 @@ func resourceSqlDatabaseInstanceCreate(d *schema.ResourceData, meta interface{})
 
 	if _, ok := d.GetOk("instance_type"); ok {
 		instance.InstanceType = d.Get("instance_type").(string)
+	}
+
+	if _, ok := d.GetOk("node_count"); ok {
+		instance.NodeCount = int64(d.Get("node_count").(int))
 	}
 
 	instance.RootPassword = d.Get("root_password").(string)
@@ -1840,6 +1853,11 @@ func resourceSqlDatabaseInstanceRead(d *schema.ResourceData, meta interface{}) e
 	if err := d.Set("instance_type", instance.InstanceType); err != nil {
 		return fmt.Errorf("Error setting instance_type: %s", err)
 	}
+	if instance.NodeCount != 0 {
+		if err := d.Set("node_count", instance.NodeCount); err != nil {
+			return fmt.Errorf("Error setting node_count: %s", err)
+		}
+	}
 	if err := d.Set("settings", flattenSettings(instance.Settings, d)); err != nil {
 		log.Printf("[WARN] Failed to set SQL Database Instance Settings")
 	}
@@ -2191,6 +2209,10 @@ func resourceSqlDatabaseInstanceUpdate(d *schema.ResourceData, meta interface{})
 		instance.InstanceType = d.Get("instance_type").(string)
 	}
 
+	if _, ok := d.GetOk("node_count"); ok {
+		instance.NodeCount = int64(d.Get("node_count").(int))
+	}
+
 	// Database Version is required for all calls with Google ML integration enabled or it will be rejected by the API.
 	if d.Get("settings.0.enable_google_ml_integration").(bool) {
 		instance.DatabaseVersion = databaseVersion
@@ -2277,6 +2299,24 @@ func databaseVersionDiffSuppress(_, oldVersion, newVersion string, _ *schema.Res
 		}
 	}
 
+	return false
+}
+
+func availabilityTypeDiffSuppress(_, old, new string, d *schema.ResourceData) bool {
+	instance_type := d.Get("instance_type").(string)
+	if instance_type == "READ_POOL_INSTANCE" {
+		log.Printf("[DEBUG] Instance is a read pool. Suppressing diff for availability_type (%s => %s)", old, new)
+		return true
+	}
+	return false
+}
+
+func nodeCountDiffSuppress(_, old, new string, d *schema.ResourceData) bool {
+	instance_type := d.Get("instance_type").(string)
+	if instance_type != "READ_POOL_INSTANCE" {
+		log.Printf("[DEBUG] Instance is not a read pool. Suppressing diff for node_count (%s => %s)", old, new)
+		return true
+	}
 	return false
 }
 
--- a/mmv1/third_party/terraform/services/sql/resource_sql_database_instance_test.go.tmpl
+++ b/mmv1/third_party/terraform/services/sql/resource_sql_database_instance_test.go.tmpl
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"regexp"
 	"strconv"
+	"strings"
 	"testing"
 	"time"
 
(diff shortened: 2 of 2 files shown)


mmv1/third_party/terraform/services/sql/resource_sql_database_instance.go.tmpl
```diff
@@ -2280,6 +2302,24 @@ func databaseVersionDiffSuppress(_, oldVersion, newVersion string, _ *schema.Res
 	return false
 }
 
+func availabilityTypeDiffSuppress(_, old, new string, d *schema.ResourceData) bool {
```
melinath, line 2305: could you give a brief description about why the diff suppress funcs are necessary?

mmv1/third_party/terraform/services/sql/resource_sql_database_instance.go.tmpl
```diff
@@ -874,7 +875,15 @@ is set to true. Defaults to ZONAL.`,
 				Type:        schema.TypeString,
 				Computed:    true,
 				Optional:    true,
-				Description: `The type of the instance. The valid values are:- 'SQL_INSTANCE_TYPE_UNSPECIFIED', 'CLOUD_SQL_INSTANCE', 'ON_PREMISES_INSTANCE' and 'READ_REPLICA_INSTANCE'.`,
+				Description: `The type of the instance. The valid values are:- 'SQL_INSTANCE_TYPE_UNSPECIFIED', 'CLOUD_SQL_INSTANCE', 'ON_PREMISES_INSTANCE', 'READ_REPLICA_INSTANCE', and 'READ_POOL_INSTANCE'.`,
```
melinath, line 878: This description field doesn't automatically modify the docs page - you'll need to update https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/third_party/terraform/website/docs/r/sql_database_instance.html.markdown as well.

Additionally, it may make sense to change the docs to point to the canonical list of enum values in the API docs instead of hardcoding them here.

mmv1/third_party/terraform/services/sql/resource_sql_database_instance.go.tmpl
psung: Done.

mmv1/third_party/terraform/services/sql/resource_sql_database_instance.go.tmpl
psung: Done, I linked both to the canonical API docs.

psung:
Thanks for the review! Note that the acceptance tests are also waiting for the rollout of some backend changes before they will pass.

psung:
@melinath, tests are passing now, can you please take a look? Thanks!

mmv1/third_party/terraform/services/sql/resource_sql_database_instance.go.tmpl
```diff
@@ -2280,6 +2302,29 @@ func databaseVersionDiffSuppress(_, oldVersion, newVersion string, _ *schema.Res
 	return false
 }
 
+func availabilityTypeDiffSuppress(_, old, new string, d *schema.ResourceData) bool {
+	// For Read Pool instances, suppress the diff because
+	// availability_type is not set by the client but managed by Cloud
+	// SQL (it can change whenever the node_count is modified).
+	instance_type := d.Get("instance_type").(string)
+	if instance_type == "READ_POOL_INSTANCE" {
+		log.Printf("[DEBUG] Instance is a read pool. Suppressing diff for availability_type (%s => %s)", old, new)
```
melinath: I'm not sure these diff suppress funcs will do quite what you want. Diff suppress funcs will make it so that diffs on this specific field will not trigger an update. However, if an update is triggered for _any other reason_, the diff will still show up for users as part of the plan, which can be confusing. It also _doesn't_ prevent the values from being detected as changed or from being sent to the API.

Generally speaking, fields like this that are partially managed by the user and partially by the server don't work super well in Terraform. There are a few different ways you could handle availability type:

- Have an "effective_availability_type" field that is output-only, and make availability_type write-only (by ignoring the values returned from the API). I think this is _probably_ the best option. yaqs/6543863299476291584
- Add validation (via a customize diff func) that says you can't set a value availability_type that's different than what the server returned, if it's a READ_POOL_INSTANCE
- Maybe something similar but "fancier", where you try to modify the behavior of availability type based on whether it's a read_pool_instance - however, that is likely to be more trouble than it's worth IMO, and would need to be careful to not be a [breaking change](https://googlecloudplatform.github.io/magic-modules/breaking-changes/breaking-changes/)

mmv1/third_party/terraform/services/sql/resource_sql_database_instance.go.tmpl
```diff
@@ -2280,6 +2302,29 @@ func databaseVersionDiffSuppress(_, oldVersion, newVersion string, _ *schema.Res
...
+	}
+	return false
+}
+
+func nodeCountDiffSuppress(_, old, new string, d *schema.ResourceData) bool {
+	// Node count will only be supplied for read pool instances, it
+	// should be ignored for non-pool instances.
+	instance_type := d.Get("instance_type").(string)
+	if instance_type != "READ_POOL_INSTANCE" {
+		log.Printf("[DEBUG] Instance is not a read pool. Suppressing diff for node_count (%s => %s)", old, new)
```
melinath: For node count, I can't tell from the comment if it's partially server managed or if you're just not allowed to set this field unless the instance type is READ_POOL_INSTANCE. If you're just not allowed to set it, a diff suppress won't prevent that - you'd need to use a customizeDiff function. (You can also choose to rely on the API to return error messages for fields with complex validation.)

psung: I removed the validation here as the API will validate it anyway.

mmv1/third_party/terraform/services/sql/resource_sql_database_instance.go.tmpl
psung: Per our discussion,
* I added an effective_availability_type field that always reports what the API said, and,
* For the availability_type field, for read pools I made it always report whatever was previously in the TF resource, so that it never attempts any update on that field. (This is fine, since if we do end up trying to update the resource due to some other field, the API's behavior is to ignore any change to availability_type anyway.)

psung:
> Hi there, I'm the Modular magician. I've detected the following information about your changes: [...]
> 
> ## Missing test report
> 
> Resource: `google_sql_database_instance` (187 total tests) Please add an acceptance test which includes these fields. The test should include the following:
> 
> ```hcl
> resource "google_sql_database_instance" "primary" {
>   node_count = # value needed
> }
> ```

TestAccSqlDatabaseInstance_PostgresReadPoolScaleOutSuccess does set node_count. If it's complaining that the **primary** specifically never has node_count set, that is in fact the case, it's not supported on primary instances.

> ## Missing doc report (experimental)
> The following resources have fields missing in documents.
> 
> * `google_sql_database_instance`
>   
>   * Expected Document Path: `/website/docs/r/sql_database_instance.html.markdown`
>   * Fields: `[settings.effective_availability_type]`

`mmv1/third_party/terraform/website/docs/r/sql_database_instance.html.markdown` has new docs added for effective_availability_type. Not sure if it's complaining because I need to follow some specific convention for this.

mmv1/third_party/terraform/services/sql/resource_sql_database_instance.go.tmpl
```diff
@@ -1840,10 +1860,14 @@ func resourceSqlDatabaseInstanceRead(d *schema.ResourceData, meta interface{}) e
 	if err := d.Set("instance_type", instance.InstanceType); err != nil {
 		return fmt.Errorf("Error setting instance_type: %s", err)
 	}
-	if err := d.Set("settings", flattenSettings(instance.Settings, d)); err != nil {
+	if instance.NodeCount != 0 {
```
melinath: hmm, why not set NodeCount to 0?

mmv1/third_party/terraform/website/docs/r/sql_database_instance.html.markdown
```diff
@@ -327,11 +327,17 @@ The `settings` block supports:
...
-  instance, high availability (`REGIONAL`) or single zone (`ZONAL`).' For all instances, ensure that
+  instance, high availability (`REGIONAL`) or single zone (`ZONAL`). For all instances, ensure that
   `settings.backup_configuration.enabled` is set to `true`.
   For MySQL instances, ensure that `settings.backup_configuration.binary_log_enabled` is set to `true`.
   For Postgres and SQL Server instances, ensure that `settings.backup_configuration.point_in_time_recovery_enabled`
   is set to `true`. Defaults to `ZONAL`.
+  For read pool instances, this field is read-only. The availability type is changed by specifying
+  the number of nodes (`node_count`).
+
+* `effective_availability_type` - (Computed) The availability type of the Cloud SQL instance, high availability
```
melinath: effective_availability_type is an output-only field; those are expected to go in the ["Attributes reference"](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/sql_database_instance#attributes-reference) section, which is why the doc report thinks it's missing. This would be good to fix for consistency. (It would be listed as `settings.0.effective_availability_type`, I think?)

mmv1/third_party/terraform/services/sql/resource_sql_database_instance_test.go.tmpl
```diff
@@ -4594,6 +4779,81 @@ resource "google_sql_database_instance" "original-replica" {
...
+  database_version    = "%s"
+  instance_type       = "CLOUD_SQL_INSTANCE"
+  deletion_protection = false
+
+  settings {
+    tier              = "db-perf-optimized-N-2"
+    edition           = "ENTERPRISE_PLUS"
+    backup_configuration {
+      enabled                        = true
+%s
```
melinath, line 4916: It looks to me like our test parsing is choking on this (primaryTxnLogs) - the test parser assumes that all %-formatting is done in a "value" position, not replacing a block/line like this.

I don't have a workaround (apart from splitting this into multiple functions) & it's not a trivial fix in the test parser so let's just leave this as-is.

mmv1/third_party/terraform/services/sql/resource_sql_database_instance_test.go.tmpl
```diff
@@ -2895,6 +2896,190 @@ func TestAccSqlDatabaseInstance_PostgresSwitchoverSuccess(t *testing.T) {
...
+				ImportState:             true,
+				ImportStateVerify:       true,
+				ImportStateVerifyIgnore: []string{"deletion_protection"},
+			},
+			{
+				ResourceName:            "google_sql_database_instance.original-read-pool",
+				ImportState:             true,
+				ImportStateVerify:       true,
+				// node_count is no longer populated after disable pool
+				ImportStateVerifyIgnore: []string{"deletion_protection", "node_count"},
```
melinath: I see that node_count is set to optional + computed, which means it will persist values returned by the API... but if the API is no longer populating the field, I'd expect you wouldn't need to ignore it during ImportStateVerify? Is this caused by not setting it to `0` in Read?

mmv1/third_party/terraform/services/sql/resource_sql_database_instance.go.tmpl
psung: Ah, right. Done.

mmv1/third_party/terraform/website/docs/r/sql_database_instance.html.markdown
psung: Done. I think this means I documented `node_count` in the wrong section as well, so I've moved it.

mmv1/third_party/terraform/services/sql/resource_sql_database_instance_test.go.tmpl
psung: Ack.

mmv1/third_party/terraform/services/sql/resource_sql_database_instance_test.go.tmpl
psung: Thanks, I didn't make the connection between the treatment of the field here and the logic above. With the change above, it passes tests now without having node_count ignored.

melinath (approved):
LGTM

PR #14091, opened 2025-05-23
Add `Tool` resource to product Dialgoflow CX

ArtoriaRen:
```release-note:new-resource
`google_dialogflow_cx_tool`
```

### DIFF
--- a/mmv1/products/dialogflowcx/Agent.yaml
+++ b/mmv1/products/dialogflowcx/Agent.yaml
@@ -54,12 +54,12 @@ examples:
       - 'enable_stackdriver_logging'
       - 'advanced_settings.0.logging_settings'
 virtual_fields:
-- name: 'delete_chat_engine_on_destroy'
-  description: |
-    If set to `true`, Terraform will delete the chat engine linked to the agent when destroying the agent.
-    Otherwise, Terraform will not delete the chat engine.
-  type: Boolean
-  default_value: false
+  - name: 'delete_chat_engine_on_destroy'
+    description: |
+      If set to `true`, Terraform will delete the chat engine linked to the agent when destroying the agent.
+      Otherwise, Terraform will not delete the chat engine.
+    type: Boolean
+    default_value: false
 parameters:
 properties:
   - name: 'name'
--- a/mmv1/products/dialogflowcx/Agent.yaml
+++ b/mmv1/products/dialogflowcx/Agent.yaml
@@ -22,7 +22,6 @@ references:
 docs:
 id_format: 'projects/{{project}}/locations/{{location}}/agents/{{name}}'
 base_url: 'projects/{{project}}/locations/{{location}}/agents'
-delete_url: 'projects/{{project}}/locations/{{location}}/agents/{{name}}?deleteContents={{delete_contents_on_destroy}}'
 update_verb: 'PATCH'
 update_mask: true
 import_format:
@@ -34,9 +33,11 @@ timeouts:
 custom_code:
   # An engine resource https://cloud.google.com/generative-ai-app-builder/docs/reference/rest/v1/projects.locations.collections.engines
   # will be automatically created when we specify dataStoreConnections in Flow, Page, or Tool resources associated with the Agent.
-  # We need to delete the engine before deleting the Agent. Otherwise, terraform wouldn't be able to delete the DataStore resource
-  # created in acceptance tests and would fail the tests.
-  # pre_delete: 'templates/terraform/pre_delete/dialogflowcx_agent.go.tmpl'
+  # We need to extract the engine ID from the Agent, delete the Agent, and then delete the engine. Otherwise, Terraform wouldn't be
+  # able to delete the DataStore resource created in acceptance tests.
+  # TODO: remove pre_delete and post_delete after the API is updated to clean up the engine resource.
+  pre_delete: 'templates/terraform/pre_delete/dialogflowcx_agent.go.tmpl'
+  post_delete: 'templates/terraform/post_delete/dialogflowcx_agent.go.tmpl'
 exclude_sweeper: true
 examples:
   - name: 'dialogflowcx_agent_full'
--- a/mmv1/products/dialogflowcx/Agent.yaml
+++ b/mmv1/products/dialogflowcx/Agent.yaml
@@ -56,8 +56,20 @@ examples:
 virtual_fields:
   - name: 'delete_chat_engine_on_destroy'
     description: |
-      If set to `true`, Terraform will delete the chat engine linked to the agent when destroying the agent.
-      Otherwise, Terraform will not delete the chat engine.
+      If set to `true`, Terraform will delete the chat engine associated with the agent when the agent is destroyed.
+      Otherwise, the chat engine will persist.
+
+      This virtual field addresses a critical dependency chain: `agent` -> `engine` -> `data store`. The chat engine is automatically
+      provisioned when a data store is linked to the agent, meaning Terraform doesn't have direct control over its lifecycle as a managed
+      resource. This creates a problem when both the agent and data store are managed by Terraform and need to be destroyed. Without
+      delete_chat_engine_on_destroy set to true, the data store's deletion would fail because the unmanaged chat engine would still be
+      using it. This setting ensures that the entire dependency chain can be properly torn down.
+      See `mmv1/templates/terraform/examples/dialogflowcx_tool_data_store.tf.tmpl` as an example.
+
+      Data store can be linked to an agent through the `knowledgeConnectorSettings` field of a [flow](https://cloud.google.com/dialogflow/cx/docs/reference/rest/v3/projects.locations.agents.flows#resource:-flow)
+      or a [page](https://cloud.google.com/dialogflow/cx/docs/reference/rest/v3/projects.locations.agents.flows.pages#resource:-page)
+      or the `dataStoreSpec` field of a [tool](https://cloud.google.com/dialogflow/cx/docs/reference/rest/v3/projects.locations.agents.tools#resource:-tool).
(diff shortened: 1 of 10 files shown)


mmv1/templates/terraform/pre_delete/dialogflowcx_agent.go.tmpl
```diff
@@ -0,0 +1,53 @@
...
+_, err = transport_tpg.SendRequest(transport_tpg.SendRequestOptions{
+	Config:    config,
+	Method:    "GET",
+	Project:   project,
+	RawURL:    engineUrl,
+	UserAgent: userAgent,
+})
+if err != nil {
+	log.Printf("[DEBUG] engine %s doesn't exist. No need to delete", engineID)
+	return nil
```
ScottSuarez, line 37: We shouldn't return nil here, correct? If the engine doesn't exist we still want to delete the agent later on

mmv1/templates/terraform/pre_delete/dialogflowcx_agent.go.tmpl
```diff
@@ -0,0 +1,53 @@
...
+}
+
+// delete the engine
+log.Printf("[DEBUG] Deleting engine %v", engineID)
+_, err = transport_tpg.SendRequest(transport_tpg.SendRequestOptions{
```
ScottSuarez, line 42: We will likely need to wait for this operation to complete

tpgservicusage.ServiceUsageOperationWaitTimeWithResponse or similar.

mmv1/templates/terraform/pre_delete/dialogflowcx_agent.go.tmpl
```diff
@@ -0,0 +1,53 @@
+s := d.Get("gen_app_builder_settings")
+log.Printf("[DEBUG] gen_app_builder_settings: %v", s)
+settings, ok := s.([]interface{})
+if !ok {
+	return fmt.Errorf("Error converting gen_app_builder_settings %s to  []interface{}", s)
+}
+if len(settings) == 0 {
+    // no engine to delete
+    return nil
```
ScottSuarez, line 9: If engine does not exist we still need to delete the agent

mmv1/products/dialogflowcx/Tool.yaml
```diff
@@ -0,0 +1,326 @@
...
+                * DOCUMENTS: Documents are processed as documents.
+                * CHUNKS: Documents are converted to chunks.
+              enum_values:
+                - 'DOCUMENTS'
+                - 'CHUNKS'
+      - name: 'fallbackPrompt'
+        type: NestedObject
+        description: |
+          Fallback prompt configurations to use.
+        required: true
```
ScottSuarez, line 297: I am confused. this is required but is empty. Why require the customer to define this block? would we add a more detailed description for the dev

mmv1/products/dialogflowcx/Tool.yaml
```diff
@@ -0,0 +1,326 @@
...
+              enum_values:
+                - 'PUBLIC_WEB'
+                - 'UNSTRUCTURED'
+                - 'STRUCTURED'
+            - name: 'dataStore'
+              type: String
+              description: |
+                The full name of the referenced data store. Formats: projects/{project}/locations/{location}/collections/{collection}/dataStores/{dataStore} projects/{project}/locations/{location}/dataStores/{dataStore}
+            - name: 'documentProcessingMode'
+              type: Enum
```
ScottSuarez, line 285: here and other places.. consider using string over Enum if more values are to be added in the future. Supporting additional enum types can be a maintenance burden.  

You could like to the valid type values in the field description.

ScottSuarez (changes requested):

mmv1/products/dialogflowcx/Tool.yaml
ArtoriaRen: That's a good point. This is consistent with the API doc though https://cloud.google.com/dialogflow/cx/docs/reference/rest/v3/projects.locations.agents.tools#datastoretool. I'll check with the Dialogflow team.

mmv1/products/dialogflowcx/Tool.yaml
ArtoriaRen: Do we need to be consistent with the REST API, which uses  Enum https://cloud.google.com/dialogflow/cx/docs/reference/rest/v3/DataStoreConnection#documentprocessingmode?

mmv1/products/dialogflowcx/Agent.yaml
```diff
@@ -42,6 +49,14 @@ examples:
...
       - 'enable_stackdriver_logging'
       - 'advanced_settings.0.logging_settings'
+virtual_fields:
+  - name: 'delete_contents_on_destroy'
+    description: |
+      If set to `true`, delete the resources linked to the agent (e.g., chat engine) when destroying the agent.
+      Otherwise, Terraform acceptance test `dialogflowcx_tool_data_store.tf.tmpl` would fail at deleting the data store resource because
+      the chat engine depending on the data store still exists.
+    type: Boolean
+    default_value: false
```
ScottSuarez: Is this still used?

mmv1/products/dialogflowcx/Agent.yaml
```diff
@@ -31,6 +31,13 @@ timeouts:
   update_minutes: 40
   delete_minutes: 20
 custom_code:
+  # An engine resource https://cloud.google.com/generative-ai-app-builder/docs/reference/rest/v1/projects.locations.collections.engines
+  # will be automatically created when we specify dataStoreConnections in Flow, Page, or Tool resources associated with the Agent.
+  # We need to extract the engine ID from the Agent, delete the Agent, and then delete the engine. Otherwise, Terraform wouldn't be
+  # able to delete the DataStore resource created in acceptance tests.
+  # TODO: remove pre_delete and post_delete after the API is updated to clean up the engine resource.
+  pre_delete: 'templates/terraform/pre_delete/dialogflowcx_agent.go.tmpl'
+  post_delete: 'templates/terraform/post_delete/dialogflowcx_agent.go.tmpl'
```
ScottSuarez, line 44: Why have both pre and post delete. Don't we only need pre-delete?

ArtoriaRen:
> looks like a test is failing due to not cleaning up the agents engine

Yes, I was trying the `delete_contents_on_destroy` virtual field approach, but it didn't work.

mmv1/products/dialogflowcx/Agent.yaml
ArtoriaRen: When I delete the engine in pre-delete, deleting the agent would fail with `404 Not Found`. I attached the log file for these two deletes [log_pre_delete_only.log](https://github.com/user-attachments/files/20560825/log_pre_delete_only.log).

On the other hand, deleting the agent first won't result in any errors. As engine ID is stored in a field of the agent, we have to obtain the engine ID before deleting the agent, so I put this part of the logic in pre-delete and the actual engine delete logic in post-delete.

mmv1/products/dialogflowcx/Agent.yaml
ArtoriaRen: Ah good catch! Thanks for noticing it. I pushed a new commit to drop this.

mmv1/products/dialogflowcx/Agent.yaml
ScottSuarez: I am confused, I thought deleting the engine before the the agent was the usecase and why we needed this code. We can delete the agent without deleting the engine?

> When I delete the engine in pre-delete, deleting the agent would fail with 404 Not Found. I attached the log file for these two deletes [log_pre_delete_only.log](https://github.com/user-attachments/files/20560825/log_pre_delete_only.log).

Do we know why this happens?

ArtoriaRen: > I am confused, I thought deleting the engine before the the agent was the usecase and why we needed this code. We can delete the agent without deleting the engine?

Right, we can delete agent without deleting engine. But then we wouldn't be able to delete the [data store resource](https://github.com/GoogleCloudPlatform/magic-modules/pull/14091/files#diff-5526156aededeb7d665573aa28608d321601ed802374570fbf7a40a527e8861eR31).  It's all because there is a chain of dependency: agent -> engine -> data store. 

> Do we know why this happens?

It's because agent depends on engine, which in turn depends on data store.

mmv1/templates/terraform/pre_delete/dialogflowcx_agent.go.tmpl
ArtoriaRen: You're right! thanks for catching this. Actually, the agent depends on the engine, so I should delete the engine in `post_delete` instead of `pre_delete`.

mmv1/templates/terraform/pre_delete/dialogflowcx_agent.go.tmpl
ArtoriaRen: Do we still need this now that we delete engine in `post_delete`?

mmv1/templates/terraform/pre_delete/dialogflowcx_agent.go.tmpl
ArtoriaRen: Ack

mmv1/products/dialogflowcx/Tool.yaml
ArtoriaRen: As we discussed in chat, this field is required for now.

mmv1/products/dialogflowcx/Tool.yaml
ArtoriaRen: Ack. I 've replaced Enum with String for all the Enum fields.

zli82016:
I am going to merge the PR as the reviewer is OOO.

PR #14093, opened 2025-05-24
fix: string based ordering for `google_compute_region_security_policy` causes recreate after apply

ramonvermeulen:
Closes https://github.com/hashicorp/terraform-provider-google/issues/22995

```release-note:bug
compute: fixed an issue where rules ordering in `google_compute_region_security_policy` caused a diff after apply
```

### DIFF
--- a/mmv1/templates/terraform/constants/region_security_policy.go.tmpl
+++ b/mmv1/templates/terraform/constants/region_security_policy.go.tmpl
@@ -1,0 +1,31 @@
+{{- if ne $.Compiler "terraformgoogleconversion-codegen" }}
+func resourceComputeRegionSecurityPolicySpecRulesDiffSuppress(k, o, n string, d *schema.ResourceData) bool {
+    oldCount, newCount := d.GetChange("rules.#")
+    var count int
+    // There could be duplicates - worth continuing even if the counts are unequal.
+    if oldCount.(int) < newCount.(int) {
+        count = newCount.(int)
+    } else {
+        count = oldCount.(int)
+    }
+
+    old := make([]interface{}, 0, count)
+    new := make([]interface{}, 0, count)
+    for i := 0; i < count; i++ {
+        o, n := d.GetChange(fmt.Sprintf("rules.%d", i))
+
+        if o != nil {
+            old = append(old, o)
+        }
+        if n != nil {
+            new = append(new, n)
+        }
+    }
+
+    oldSet := schema.NewSet(schema.HashResource(ResourceComputeRegionSecurityPolicy().Schema["rules"].Elem.(*schema.Resource)), old)
+    newSet := schema.NewSet(schema.HashResource(ResourceComputeRegionSecurityPolicy().Schema["rules"].Elem.(*schema.Resource)), new)
+
+    return oldSet.Equal(newSet)
+}
+
+{{- end }}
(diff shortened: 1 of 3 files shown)


ramonvermeulen:
@rileykarson 

I suspect a similar problem here as in https://github.com/GoogleCloudPlatform/magic-modules/pull/14086#pullrequestreview-2876655567, can you please confirm this from the VCR test logs? In that case I will change the tests so that it will use a dedicated project.

mmv1/templates/terraform/constants/region_security_policy.go.tmpl
```diff
@@ -0,0 +1,31 @@
+{{- if ne $.Compiler "terraformgoogleconversion-codegen" }}
+func resourceComputeRegionSecurityPolicySpecRulesDiffSuppress(k, o, n string, d *schema.ResourceData) bool {
+    oldCount, newCount := d.GetChange("rules.#")
+    var count int
+    // There could be duplicates - worth continuing even if the counts are unequal.
```
rileykarson, line 5: Will the API reject duplicates? If so, should we use a CustomizeDiff to find duplicates and reject them if they're present? (non blocking for this change)

mmv1/third_party/terraform/services/compute/resource_compute_region_security_policy_test.go.tmpl
```diff
@@ -722,6 +722,121 @@ func testAccComputeRegionSecurityPolicy_withMultipleEnforceOnKeyConfigs_update(c
...
+		"random_suffix": acctest.RandString(t, 10),
+	}
+
+	acctest.VcrTest(t, resource.TestCase{
+		PreCheck:                 func() { acctest.AccTestPreCheck(t) },
+		ProtoV5ProviderFactories: acctest.ProtoV5ProviderFactories(t),
+		CheckDestroy:             testAccCheckComputeRegionSecurityPolicyDestroyProducer(t),
+		Steps: []resource.TestStep{
+			{
+				Config: testAccComputeRegionSecurityPolicy_ruleOrderingWithMultipleRules_create(context),
```
rileykarson, line 738: Just making sure I understand correctly- this test would have failed before the change and passes now. Previous tests all passed because they generally contained two rules, with priority `100`/`1000` and then `2147483647`. Those got returned by the API in that order.

Since this has `3000` followed by `2147483647`, the API returns `2147483647` followed by `3000` and we currently get a diff (but the test change it).

ramonvermeulen: Yes exactly, took a TDD approach for this bug so on https://github.com/GoogleCloudPlatform/magic-modules/pull/14093/commits/a66d24e7681c3cda9e045ae5f59a55a3fbc784d0 it failed and on https://github.com/GoogleCloudPlatform/magic-modules/pull/14093/commits/2ebc839634a17e74822ac12f4b459a2bbd74e063 it succeed. See https://github.com/hashicorp/terraform-provider-google/issues/22995 for a little more context, but indeed the problem was that the API returns the rules in a different order than in which you configure them (based on priority).

mmv1/templates/terraform/constants/region_security_policy.go.tmpl
ramonvermeulen: I took this example over from https://github.com/GoogleCloudPlatform/magic-modules/pull/13068 which had a similar issue, honestly I am not sure if the API rejects duplicates. This is something I have to test, otherwise it might be "more efficient" to already return `false` if the counts are not equal.

rileykarson (approved):
LGTM, thanks!

PR #14096, opened 2025-05-25
Fix default disk type in docs

boomanaiden154:
This patch updates the default disk type in GKE container clusters. It is not pd-standard.

```release-note:REPLACEME

```


SirGitsalot (approved):
Weirdly the API docs still say `pd-standard` is the default but I checked our test logs and `pd-balanced` is indeed returned when `diskType` is unspecified (which matches [this doc](https://cloud.google.com/compute/docs/disks/performance#type_comparison) but we're using the GKE API not the GCE API)

PR #14103, opened 2025-05-26
feat: (storage) add custom_headers for folder resource

gurusai-voleti:

```release-note:enhancement
storage: added `custom_headers` for resource `google_storage_folder`
```


mmv1/products/storage/Folder.yaml
```diff
@@ -72,6 +74,10 @@ parameters:
     validation:
       regex: '/$'
 properties:
+  - name: 'customHeaders'
```
ScottSuarez, line 77: This should be labeled as a virtual_field with name of type custom_headers (virtual fields don't map to api fields)

see https://googlecloudplatform.github.io/magic-modules/reference/resource/#virtual_fields

mmv1/templates/terraform/custom_delete/storage_folder_delete.go.tmpl
```diff
@@ -55,7 +55,14 @@ for deleteObjectError == nil {
...
 err = retry.Retry(1*time.Minute, func() *retry.RetryError {
-    err := config.NewStorageClient(userAgent).Folders.Delete(bucket, name).Do()
+    deleteCall := config.NewStorageClient(userAgent).Folders.Delete(bucket, name)
+    headers := make(http.Header)
+    if v_headers, ok := d.GetOk("custom_headers"); ok {
+    for k, v := range v_headers.(map[string]interface{}) {
+            headers.Add(k, v.(string))
+        }
+    }
+    err = deleteCall.Do()
```
ScottSuarez, line 65: Since this function is implemented using the client library the headers object is not used. We will need to pass headers to the client library.

ScottSuarez (changes requested):

ScottSuarez:
converting to draft for now in accordance with the bellow comment

mmv1/products/storage/Folder.yaml
gurusai-voleti: this is suggested in one of the yaqs so implemented url_param_only, if this is not correct will change

ScottSuarez (commented):
Is this Pull Request ready for review? I saw in the internal bug there was some discussion around implementation so I just want to be sure.

PR #14107, opened 2025-05-27
Support Cross-Site Interconnect. 

NandiniAgrawal15:

```release-note: new-resource 
`google_compute_interconnect`
```


mmv1/third_party/terraform/services/compute/resource_compute_interconnect_macsec_test.go.tmpl
```diff
@@ -0,0 +1,185 @@
...
+					resource.TestCheckResourceAttr("google_compute_interconnect.example-interconnect-cross-site-network", "description", "example description updated")),
+			},
+		},
+	})
+}
+
+func testAccComputeInterconnect_crossSiteInterconnectBasic(context map[string]interface{}) string {
+	return acctest.Nprintf(`
+data "google_project" "project" {
+provider    = google-beta}
```
melinath: This is causing the test to fail to even run. (Of course once it runs it will have the same problem as the other test).

```suggestion
  provider    = google-beta
}
```

mmv1/products/compute/Interconnect.yaml
```diff
@@ -410,5 +420,17 @@ properties:
       ports. If not present then the Interconnect connection is provisioned on non-MACsec capable
       ports and MACsec isn't supported and enabling MACsec fails).
     output: true
+    item_type:
+      type: Enum
+      description: interconnects.list of available features for this Interconnect connection.
+      enum_values:
+        - 'MACSEC'
+        - 'CROSS_SITE_NETWORK'
```
melinath: this code has no effect because this is an output-only field.

```suggestion
```

melinath (changes requested):
It looks like the tests for this change are failing due to a permadiff bug.

The diff on interconnect_type is correct; it would normally be hidden due to [InterconnectTypeDiffSuppress](https://github.com/hashicorp/terraform-provider-google-beta/blob/acdc0bd63d5f5a6b981e179ca9080b6b865a554d/google-beta/services/compute/resource_compute_interconnect.go#L37). Same is true for the diff on macsec.

They are being displayed because there's a diff on requested_features:

```
              ~ requested_features       = [ # forces replacement
                  + "CROSS_SITE_NETWORK",
                ]
```

this means that the API is not returning a value for requested_features, as you can see in the test's debug log. Why is that the case?

melinath (changes requested):
see previous change request

PR #14048, opened 2025-05-20
Add workload identity pool managed identity.

stevenyang72:
**Release Note Template for Downstream PRs (will be copied)**

See [Write release notes](https://googlecloudplatform.github.io/magic-modules/contribute/release-notes/) for guidance.

```release-note:new-resource
`google_iam_workload_identity_pool_managed_identity` (beta)
```

PR #14110, opened 2025-05-27
Remove id validation on mwlid namespace and managed identity.

stevenyang72:

```release-note:none

```

Remove ID validation on WorkloadIdentityPool Namespace and Managed Identity, clean up associated test files.

Context: https://github.com/GoogleCloudPlatform/magic-modules/pull/14048#discussion_r2103484559


melinath (commented):
Validation functions are generally useful because they allow errors to happen at plan time instead of apply time. We do sometimes remove them if there's evidence that the validation won't be stable over time - is that the case here? Otherwise I don't think there's a need to remove this validation.

Is there some additional context for this change? Like a bug that the validation is causing?

cc @zli82016 as the commenter of https://github.com/GoogleCloudPlatform/magic-modules/pull/14048#discussion_r2103448135

zli82016:
> Validation functions are generally useful because they allow errors to happen at plan time instead of apply time. We do sometimes remove them if there's evidence that the validation won't be stable over time - is that the case here? Otherwise I don't think there's a need to remove this validation.
> 
> Is there some additional context for this change? Like a bug that the validation is causing?
> 
> cc @zli82016 as the commenter of [#14048 (comment)](https://github.com/GoogleCloudPlatform/magic-modules/pull/14048#discussion_r2103448135)

There isn't a bug related to the validation itself. My main concern is simply that I remember we prefer to avoid client-side validation, as it could cause issue if the validation changes on the service side.

melinath:
Ack, thanks @zli82016!

@stevenyang72 is this field likely to have looser validation in the future? If so we should just remove the validation; if it's expected to be static we should probably leave the validation in place.

stevenyang72:
> oh you already said:
> 
> > I'm not sure if there'll be changes on the server-side validation, but IMO the change would probably rarely happen because that will likely be a breaking change.
> 
> I think I'd lean towards leaving the validation there in that case.

Got it. Thanks for the input! Then I'll open another PR to bring back the ID validation on Managed Identity.

melinath:
@stevenyang72 I checked with other folks on the team just to be sure and it sounds like we generally allow validation of ID segments like this. If you'd like to open a PR to add that validation back you're welcome to.

PR #14112, opened 2025-05-27
tgc-revival: HclWriteBlocks function modification

zli82016:
1. Add `provider = google-beta` to all of the converted HCL, as we use the schema from TPGB. Also, the google-beta provider is used for testing in tgc.
2. Generate empty object if the type of the field is `TypeSet` to handle the case that the field is required. For example, [scopes](https://github.com/GoogleCloudPlatform/terraform-google-conversion/blob/main/pkg/tfplan2cai/converters/services/compute/compute_instance.go) is required and has the type `scopes`.

```release-note:none

```

### DIFF
--- a/mmv1/third_party/tgc_next/pkg/cai2hcl/models/hcl_block.go
+++ b/mmv1/third_party/tgc_next/pkg/cai2hcl/models/hcl_block.go
@@ -3,7 +3,6 @@ package models
 import (
 	"fmt"
 
-	"github.com/hashicorp/hcl/hcl/printer"
 	"github.com/hashicorp/hcl/v2/hclwrite"
 	"github.com/zclconf/go-cty/cty"
 )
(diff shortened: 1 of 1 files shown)


mmv1/third_party/tgc_next/pkg/cai2hcl/models/hcl_block.go
```diff
@@ -20,12 +19,15 @@ func HclWriteBlocks(blocks []*TerraformResourceBlock) ([]byte, error) {
...
+		resourceBody := hclBlock.Body()
+		resourceBody.SetAttributeRaw("provider", hclwrite.TokensForIdentifier("google-beta"))
+
 		if err := hclWriteBlock(resourceBlock.Value, hclBlock.Body()); err != nil {
 			return nil, err
 		}
 	}
 
-	return printer.Format(f.Bytes())
+	return hclwrite.Format(f.Bytes()), nil
```
zli82016, line 30: Example https://paste.googleplex.com/6602186907451392

melinath (approved):

PR #14113, opened 2025-05-27
Terraform support for Dataplex publishing

XianJiChen:
```release-note:enhancement
Dataplex: added `catalog_publishing_enabled` field to `google_dataplex_datascan` resource
```


BBBmau (dismissed):
We get some permission issues on the project used for CI:

```
    resource_dataplex_datascan_test.go:38: Step 1/4 error: Error running apply: exit status 1
        
        Error: Error creating Datascan: googleapi: Error 403: Permission denied for user to access source BigQuery table [project:dataplex-back-end-dev-project, dataset:austin_bikeshare, table:bikeshare_stations]. You should have `bigquery.tables.get` permission for the specified table to create a DataScan for it. Please grant `bigquery.tables.get` permission for the specified table via appropriate IAM roles e.g. BigQuery Data Viewer.
        
          with google_dataplex_datascan.full_quality,
          on terraform_plugin_test.tf line 2, in resource "google_dataplex_datascan" "full_quality":
           2: resource "google_dataplex_datascan" "full_quality" {
```

Can you take a look by chance? @melinath

melinath:
I don't think this is a problem I can help with - the test should either provision the data and set the correct permissions or use something publicly visible.

XianJiChen:
@melinath Thanks for the clarification! Are these the account we should grant permission to? 
hashicorp-test-runner@ci-test-project-nightly-beta.iam.gserviceaccount.com
hashicorp-test-runner@ci-test-project-nightly.iam.gserviceaccount.com

BBBmau (commented):
You'll want to refer to the [bootstrap IAM resources](https://googlecloudplatform.github.io/magic-modules/test/test/#iam-resources) to handle the permission issues showing up in generation examples

XianJiChen:
@BBBmau Looks like both the creation and update tests have passed. Please take a look when you get a chance and let me know if there's any issues. Thanks!

mmv1/products/dataplex/Datascan.yaml
```diff
@@ -84,6 +84,14 @@ examples:
       datascan_name: 'dataquality-full'
     test_env_vars:
       project_name: 'PROJECT_NAME'
+    exclude_test: true
```
BBBmau, line 87: could i get an explanation on why this test is skipped?

XianJiChen: Sure! We are trying to add the field `catalog_publishing_enabled` to the existing resource `google_dataplex_datascan`. However, having this field set to true will require `bigquery.dataOwner` permission on the resource table that is used to create the datascan. Since we don't have the permission on the bigquery-public-dataset and should not have, we're creating the `google_bigquery_dataset` and `google_bigquery_table` from scratch so that the test project will automatically have the permission. But this will make the documentation quite messy, showing all the steps that are not necessary ,and could be misleading the users to create everything from scratch. So after consulting with the internal terraform team member, we decided to maintain two templates, one for generating the documentation and one for generating the tests, which is similar to these [examples](https://github.com/GoogleCloudPlatform/magic-modules/blob/65d00044411b0a52709b0cd4da996af30d2eac05/mmv1/products/alloydb/User.yaml#L39-L64). Hope this clarifies your concern.

trodge (approved):

PR #14117, opened 2025-05-28
fix: (storage) added source_md5hash field in bucket object

gurusai-voleti:
Fixes: https://github.com/hashicorp/terraform-provider-google/issues/18618

```release-note:enhancement
storage: added `source_md5hash` field in `google_storage_bucket_object`
```


mmv1/third_party/terraform/services/storage/resource_storage_bucket_object.go
```diff
@@ -132,9 +132,16 @@ func ResourceStorageBucketObject() *schema.Resource {
...
+				Type:        schema.TypeString,
+				Optional:    true,
+				Description: `User-provided md5hash, Base 64 MD5 hash of the object data.`,
+			},
+
 			// Detect changes to local file or changes made outside of Terraform to the file stored on the server.
 			"detect_md5hash": {
-				Type: schema.TypeString,
+				Type:       schema.TypeString,
+				Deprecated: "`detect_md5hash` field will be deprecated in future release. Start using `source_md5hash` instead",
```
kautikdk, line 144: >"...will be deprecated..."

It should be "is deprecated and will be removed in future release".

mmv1/third_party/terraform/services/storage/resource_storage_bucket_object_test.go
```diff
@@ -87,12 +88,9 @@ func TestAccStorageObject_recreate(t *testing.T) {
...
 			{
 				PreConfig: func() {
-					err := os.Rename(updatedName, testFile.Name())
-					if err != nil {
-						t.Errorf("Failed to rename %s to %s", updatedName, testFile.Name())
-					}
+					writeFile(testFile.Name(), []byte("datum"))
 				},
-				Config: testGoogleStorageBucketsObjectBasic(bucketName, testFile.Name()),
+				Config: testGoogleStorageBucketsObjectFileMd5(bucketName, testFile.Name()),
```
kautikdk, line 93: What's the need of changing this test? Also add one specific test to check behavior of the `source_md5hash` field with steps of value updates. That will ensure workings of the field itself.

mmv1/third_party/terraform/website/docs/r/storage_bucket_object.html.markdown
```diff
@@ -87,6 +87,8 @@ One of the following is required:
 
 * `kms_key_name` - (Optional) The resource name of the Cloud KMS key that will be used to [encrypt](https://cloud.google.com/storage/docs/encryption/using-customer-managed-keys) the object.
 
+* `source_md5hash` - (Optional) User-provided md5hash, Base 64 MD5 hash of the object data.
```
kautikdk, line 90: @roaks3 should we specify in the public documentation that the field is only to generate diff for the content change? Reason is we have couple of similar output only [md5hash fields](https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/third_party/terraform/services/storage/resource_storage_bucket_object.go#L113) so just to clarify things.

kautikdk (commented):
Please add reference of the previous PR in description where the solution was discussed.

mmv1/third_party/terraform/services/storage/resource_storage_bucket_object_test.go
gurusai-voleti: in this test we are passing source_md5hash explicitly saying change of object data and ensure object update happen on storage bucket

mmv1/third_party/terraform/website/docs/r/storage_bucket_object.html.markdown
```diff
@@ -87,6 +87,8 @@ One of the following is required:
 
 * `kms_key_name` - (Optional) The resource name of the Cloud KMS key that will be used to [encrypt](https://cloud.google.com/storage/docs/encryption/using-customer-managed-keys) the object.
 
+* `source_md5hash` - (Optional) User-provided md5hash, Base 64 MD5 hash of the object data. Note: Use only static md5 value, donot calculate or use dynamic values
```
kautikdk: Please follow this section guidelines for adding notes: https://googlecloudplatform.github.io/magic-modules/document/handwritten-docs-style-guide/#callouts

Nit: "Use only static value, do not calculate or use dynamic values."

mmv1/third_party/terraform/services/storage/resource_storage_bucket_object.go
```diff
@@ -477,8 +487,8 @@ func resourceStorageBucketObjectRead(d *schema.ResourceData, meta interface{}) e
 	if err := d.Set("md5hexhash", md5HexHash); err != nil {
 		return fmt.Errorf("Error setting md5hexhash: %s", err)
 	}
-	if err := d.Set("detect_md5hash", res.Md5Hash); err != nil {
-		return fmt.Errorf("Error setting detect_md5hash: %s", err)
+	if err := d.Set("source_md5hash", d.Get("source_md5hash")); err != nil {
```
kautikdk, line 497: Why are removing `detect_md5hash` changes? There can be users who are dependent on this field, it can break their scripts so keep that changes as well.

mmv1/third_party/terraform/services/storage/resource_storage_bucket_object.go
```diff
@@ -147,6 +154,9 @@ func ResourceStorageBucketObject() *schema.Resource {
 				// 3. Don't suppress the diff iff they don't match
 				DiffSuppressFunc: func(k, old, new string, d *schema.ResourceData) bool {
 					localMd5Hash := ""
+					if v, ok := d.GetOkExists("source_md5hash"); ok && v != "" {
```
kautikdk: @BBBmau @roaks3 there is one concern related to this logic. I believe d.GetOkExists or d.GetOk only returns true for the user provided values, Right? What if we want to check for "(Known After Apply)" or Computed values? Is there any easier way to do so? If yes, we can probably make this new field useful for dynamic content as well.

kautikdk:
Hi @roaks3 @BBBmau, can you please start PR review as this is quite complicated issue and related to Terraform core limitations?

mmv1/third_party/terraform/services/storage/resource_storage_bucket_object.go
BBBmau: d.GetOk obtains and checks whether the field is set in state through `resourceData` so this would cover non-user provided values also. `GetRawConfig` is specifically for obtaining the values from terraform configuration.

BBBmau (changes requested):
Seems like the reason for the failing test and why the m5hash isn't being updated is due to it being missing in the update check for HasChanges:

https://github.com/GoogleCloudPlatform/magic-modules/pull/14117/files#diff-e198b25b4e5d21b679dd1a9da36c9c939af82f891a6066fa14301d3b87ee9935L395-L397

mmv1/third_party/terraform/services/storage/resource_storage_bucket_object.go
gurusai-voleti: added

mmv1/third_party/terraform/website/docs/r/storage_bucket_object.html.markdown
gurusai-voleti: updated

mmv1/third_party/terraform/services/storage/resource_storage_bucket_object.go
kautikdk: @gurusai-voleti can you please check if we can utilize both d.GetOk and d.GetRawConfig to fetch `source_md5hash` value in DiffSuppressFunc of the `detect_md5hash` field to suppress diff for the dynamic values?

BBBmau: are you saying that `source_md5hash` can produce computed values? I thought the field was meant as a user-provided field so no computed values should be expected. It makes sense the current implementation since it will obtain the `source_md5hash` from state and will thus suppress the diff on `detect_md5hash`

gurusai-voleti: source_md5hash is a user provided but can be dynamic value like in below example, here the value of source_md5hash will be unknown when executing tf plan and value will be finalized when tf apply executed so handling this unknown value is needed as well @kautikdk  please correct/add if I am wrong 

resource "google_storage_bucket" "bucket" {
  name     = "${var.prefix}-my-bucket3"
  project  = "storage-sdk-vendor"
  location = "EU"
  uniform_bucket_level_access = true
}

data "http" "example" {
  url = "https://httpstat.us/200"
}

resource "local_file" "test" {
  content  =  jsonencode(data.http.example.response_headers)
  filename = "test-file-a1"
}

resource "google_storage_bucket_object" "bo2" {
  source         = local_file.test.filename
  source_md5hash = local_file.test.content_md5
  bucket         = google_storage_bucket.bucket.name
  name           = "test-file-a"
}

mmv1/third_party/terraform/services/storage/resource_storage_bucket_object.go
```diff
@@ -147,6 +155,13 @@ func ResourceStorageBucketObject() *schema.Resource {
 				// 3. Don't suppress the diff iff they don't match
 				DiffSuppressFunc: func(k, old, new string, d *schema.ResourceData) bool {
 					localMd5Hash := ""
+					if d.GetRawConfig().GetAttr("source_md5hash") == cty.UnknownVal(cty.String) {
+						log.Printf("source md5hash is not present")
```
kautikdk: Are we suppressing diff because `source_md5hash` is not present? I believe we are suppressing if the value is unknown, Right?

mmv1/third_party/terraform/services/storage/resource_storage_bucket_object_test.go
```diff
@@ -509,6 +509,54 @@ func TestResourceStorageBucketObjectUpdate_ContentChange(t *testing.T) {
...
+				Config: testGoogleStorageBucketsObjectBasic(bucketName, testFile.Name()),
+				Check:  testAccCheckGoogleStorageObject(t, bucketName, objectName, dataMd5),
+			},
+			{
+				PreConfig: func() {
+					if err := ioutil.WriteFile(testFile.Name(), updatedata, 0644); err != nil {
+						t.Errorf("error writing file: %v", err)
+					}
+				},
+				Config: testGoogleStorageBucketsObjectFileMd5(bucketName, testFile.Name(), updatedDataMd5),
```
kautikdk, line 556: Please add one field update step as well.

mmv1/third_party/terraform/website/docs/r/storage_bucket_object.html.markdown
```diff
@@ -87,6 +87,8 @@ One of the following is required:
 
 * `kms_key_name` - (Optional) The resource name of the Cloud KMS key that will be used to [encrypt](https://cloud.google.com/storage/docs/encryption/using-customer-managed-keys) the object.
 
+* `source_md5hash` - (Optional) User-provided md5hash, Base 64 MD5 hash of the object data. Note: Use only static value, do not calculate or use dynamic values.
```
kautikdk: If the solution is working for computed values, should we remove the note?

mmv1/third_party/terraform/services/storage/resource_storage_bucket_object_test.go
```diff
@@ -509,6 +509,54 @@ func TestResourceStorageBucketObjectUpdate_ContentChange(t *testing.T) {
 	})
 }
 
+func TestAccStorageObject_sourceMd5Hash(t *testing.T) {
```
kautikdk, line 512: This test can not validate the fix against "Inconsistent final plan" error scenario.
Please consider adding test coverage that can reproduce the issue and validate fix like suggested here: https://github.com/GoogleCloudPlatform/magic-modules/pull/12032#issuecomment-2662770949

mmv1/third_party/terraform/services/storage/resource_storage_bucket_object.go
gurusai-voleti: yes if source_md5hash is unknown/dynamic we are suppressing detect_md5hash func

mmv1/third_party/terraform/website/docs/r/storage_bucket_object.html.markdown
gurusai-voleti: yes removed already

mmv1/third_party/terraform/services/storage/resource_storage_bucket_object_test.go
gurusai-voleti: updated

mmv1/third_party/terraform/services/storage/resource_storage_bucket_object_test.go
gurusai-voleti: added the sample suggested

mmv1/third_party/terraform/services/storage/resource_storage_bucket_object_test.go
```diff
@@ -847,3 +938,116 @@ func getNewTmpTestFile(t *testing.T, prefix string) *os.File {
...
+}
+
+resource "local_file" "test" {
+  content  = jsonencode(google_storage_bucket_object.changing.content)
+  filename = "%s"
+}
+
+resource "google_storage_bucket_object" "bo" {
+  source = local_file.test.filename
+  //source_md5hash = local_file.test.content_md5
```
BBBmau: we should remove code that's commented

mmv1/third_party/terraform/website/docs/r/storage_bucket_object.html.markdown
BBBmau: perhaps we can include in documentation as well as in the schema that we only want one or the other to be set, I can't imagine there's a case where both would be set. That can make it clear to the user that they'd want to utilize this over `md5hash`

mmv1/third_party/terraform/services/storage/resource_storage_bucket_object_test.go
gurusai-voleti: removed

mmv1/third_party/terraform/services/storage/resource_storage_bucket_object_test.go
gurusai-voleti: added test case

BBBmau:
> 
> ## Diff report
> 
> `google` provider: [Diff](https://github.com/modular-magician/terraform-provider-google/compare/5087f248a5c6776734b6fee4e85d73b595f53e2a..01e3faa6a6957096bfec2f6b6a60b299664c4b5a) ( 3 files changed, 233 insertions(+), 2 deletions(-)) `google-beta` provider: [Diff](https://github.com/modular-magician/terraform-provider-google-beta/compare/3189d2cf972364cb4b5a8191ee7a44cfa486b806..10198b3b3ada4980bed2fd5ba56ad8665901587d) ( 3 files changed, 233 insertions(+), 2 deletions(-))
> 
> ## Missing test report
> 
> Resource: `google_storage_bucket_object` (243 total tests) Please add an acceptance test which includes these fields. The test should include the following:
> 
> ```hcl
> resource "google_storage_bucket_object" "primary" {
>   detect_md5hash = # value needed
> }
> ```

~~seems like we've never included `detect_md55hash` in our tests ever, can we do so in this PR?~~

actually nevermind, this doesn't make sense to do now since this field is being deprecated and will be removed soon.

BBBmau (approved):

mmv1/third_party/terraform/services/storage/resource_storage_bucket_object_test.go
```diff
@@ -509,6 +511,95 @@ func TestResourceStorageBucketObjectUpdate_ContentChange(t *testing.T) {
...
+		},
+	})
+}
+
+func TestAccStorageObject_knownAfterApply(t *testing.T) {
+	t.Parallel()
+
+	bucketName := acctest.TestBucketName(t)
+	destinationDirPath := t.TempDir()
+	destinationFilePath := filepath.Join(destinationDirPath, "local_file")
```
roaks3: I think you may want a randomized string at the end of this, so that subsequent runs of the test don't try to reuse the same file. I'm not sure the `tf-test` prefix matters here, but you may want to use it just in case (most other files used in tests appear to use it).

mmv1/third_party/terraform/website/docs/r/storage_bucket_object.html.markdown
roaks3: I would suggest guidance similar to https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/lambda_function#source_code_hash-1. Just saying it is user-provided is probably a little too vague for the average user.

mmv1/third_party/terraform/services/storage/resource_storage_bucket_object_test.go
gurusai-voleti: updated test file  name

mmv1/third_party/terraform/website/docs/r/storage_bucket_object.html.markdown
gurusai-voleti: Updated description

roaks3 (approved):

PR #14119, opened 2025-05-28
Add resource_manager_tags support to Network api

Veronichka:
Added resource manager tags support to Network.
Part of https://github.com/hashicorp/terraform-provider-google/issues/23225

```release-note:enhancement
compute: add `params.resourceManagerTags` field to the `google_compute_network`
```


shuyama1:

Assign a new reviewer, as I'll be OOO most of the week.

mmv1/third_party/terraform/acctest/bootstrap_test_utils.go.tmpl
```diff
@@ -2148,29 +2150,34 @@ func bootstrapSharedTestTagKey(t *testing.T, testId, parent string, obj map[stri
 		t.Fatalf("Error getting shared tag key %q: %s", sharedTagKey, err)
 	}
 
-	return sharedTagKey
+	return map[string]interface{}{
```
melinath: given these are both strings, would it make sense to instead use:

```suggestion
	return map[string]string{
```

or maybe even a struct?

mmv1/third_party/terraform/services/compute/resource_compute_network_test.go.tmpl
```diff
@@ -838,3 +872,17 @@ resource "google_compute_network" "acc_network_firewall_policy_enforcement_order
...
 }
+
+func testAccComputeNetwork_resourceManagerTags(context map[string]interface{}) string {
+	return acctest.Nprintf(`
+resource "google_compute_network" "acc_network_with_resource_manager_tags" {
+  name                            = "%{network_name}"
+  auto_create_subnetworks         = false
+  params {
+	resource_manager_tags = {
+	  "%{tag_key_id}" = "%{tag_value_id}"
```
melinath, line 883: It looks like you're making significant changes to the tag bootstrapping logic. What format will these values have? Is it significantly different than the other resources that have already added tag support?

mmv1/third_party/terraform/acctest/bootstrap_test_utils.go.tmpl
Veronichka: Good point, updated to use `map[string]string`

mmv1/third_party/terraform/services/compute/resource_compute_network_test.go.tmpl
Veronichka: Difference between other resources and compute resources is that compute resources don't support "namespaced" tags. So when for other resources it's possible to pass the short name, the compute resources support just tags in format "tagKeys/12345", "tagValues/543132". I haven't changed in the PR how the other resources are supporting tags and what is sent in their request, I just added additional method which allows me to use "name" for compute resources - tag key id and tag value id from the tags get response (e.g. response for tag key https://cloud.google.com/resource-manager/reference/rest/v3/tagKeys#TagKey).

melinath:
failing tests look unrelated.

PR #14121, opened 2025-05-28
Update Gemini GeminiGcpEnablementSettingBinding resource -- change allowed products

mwiatrowski:
```release-note:enhancement
gemini: updated allowed product values for field `product` in `google_gemini_gemini_gcp_enablement_setting_binding` resource
```


mwiatrowski:
@NickElliot That enum value still exists, but it's no longer accepted for that resource (creation/update will return an error). What should I do in that case? Is it better to list all possible values, or only the valid ones?

NickElliot (commented):
when you say cant be created or updated, do you mean resources presently using the value cant be used at all as long as they have the other enum, or it just isnt eligible to be changed to? if its the latter, then the semi-unsupported value can stay, since users who have existing configs can keep their existing resources that have that value

PR #14124, opened 2025-05-28
Documented validation best practices

melinath:
Added documentation for validation best practices.

localhost_1313_magic-modules_best-practices_validation_

```release-note:none

```

### DIFF
--- a/docs/content/best-practices/validation.md
+++ b/docs/content/best-practices/validation.md
@@ -19,7 +19,7 @@ If a resource URL looks like:
 projects/{project}/folders/{folder}/resource/{resource_id}
 ```
 
-Adding validation for the last part of the path (`resource_id`) may be safe if there are specific restrictions that are not going to change. However, if the API was ever less restrictive (or becomes less restrictive later), resources created with other tools and then imported into Terraform may be impossible to actually manage with Terraform (without deleting & recreating them) because the ID which was valid in the API violates the more restrictive validation in the provider.
+Adding validation for the last part of the path (`resource_id`) may be safe if there are specific restrictions that aren't going to change, such as following an external RFC or other spec/standard. However, if the API was ever less restrictive (or becomes less restrictive later), resources created with other tools and then imported into Terraform may be impossible to actually manage with Terraform (without deleting & recreating them) because the ID which was valid in the API violates the more restrictive validation in the provider.
 
 ## Enum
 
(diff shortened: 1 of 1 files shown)


docs/content/best-practices/validation.md
```diff
@@ -0,0 +1,36 @@
...
+
+Adding validation for the last part of the path (`resource_id`) is likely safe in most cases, especially for GCE resources.
+
+## Enum
+
+Enums are generally okay if they seem to cover the entire scope of possible values for a clearly defined domain. Otherwise, it is better to use a string field and add a link to the API documentation as a reference for the possible values.
+
+## Inter-field relationships
+
+`conflicts_with`, `required_with`, and similar are safe types of client-side validation, because they are intrinsically linked to fields in the provider. If a new field is added to the API that invalidates a rule, users will need to update the provider to get access to that field anyway, so there isn't a future-compatibility concern.
```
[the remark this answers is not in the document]
rileykarson: We've seen a good number of cases where these rules are changed. I can't recall any specific ones offhand, but probably every month or two a RequiredWith or ConflictsWith gets outdated.

docs/content/best-practices/validation.md
```diff
@@ -0,0 +1,34 @@
...
+
+## URL segments
+
+If a resource URL looks like:
+
+```
+projects/{project}/folders/{folder}/resource/{resource_id}
+```
+
+Adding validation for the last part of the path (`resource_id`) may be safe if there are specific restrictions that are not going to change. However, if the API was ever less restrictive (or becomes less restrictive later), resources created with other tools may not be possible to import into Terraform safely.
```
rileykarson, line 22: Can you elaborate on what would be _unsafe_ specifically? I think that users would not be successful because we would reject their config and/or fail on import if we had apply time assumptions that were wrong.

melinath: Done.

docs/content/best-practices/validation.md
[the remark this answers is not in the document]
rileykarson: Probably not common enough to discourage relative to the utility of the rule. I would just change this to state that if future changes are anticipated (or even just likely enough), we should use the more relaxed schema.

> without any other changes to the API

APIs change constantly. I assume you mean without a co-required schema change, i.e. a new field that triggers the validationn change. Yes, probably at a quarterly frequency. Also consider that we may have misses where we add the co-required field and miss the related relaxation of the schema rules.

docs/content/best-practices/validation.md
```diff
@@ -0,0 +1,34 @@
...
+
+## URL segments
+
+If a resource URL looks like:
+
+```
+projects/{project}/folders/{folder}/resource/{resource_id}
+```
+
+Adding validation for the last part of the path (`resource_id`) may be safe if there are specific restrictions that are not going to change. However, if the API was ever less restrictive (or becomes less restrictive later), resources created with other tools and then imported into Terraform may be impossible to actually manage with Terraform (without deleting & recreating them) because the ID which was valid in the API violates the more restrictive validation in the provider.
```
rileykarson, line 22: maybe "specific restrictions that are not going to change like following an external RFC or other spec/standard", like GCE does.

melinath: done

docs/content/best-practices/validation.md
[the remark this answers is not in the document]
melinath: rewrote

zli82016 (approved):

rileykarson (approved):

PR #14126, opened 2025-05-28
Enforce id validation for WorkloadIdentityPool Managed Identity

stevenyang72:

```release-note: enhancement
iambeta: enforced `workload_identity_pool_managed_identity_id` field validation per the documented specifications
```

Add back ID validation on WorkloadIdentityPool Managed Identity. Context: https://github.com/GoogleCloudPlatform/magic-modules/pull/14110#issuecomment-2917187493


roaks3 (approved):
Seems reasonable consider the context, noting that new validation could _technically_ impact some existing configs, but it seems unlikely that there would be values failing this validation while succeeding previously

PR #14127, opened 2025-05-28
Allow folder ancestor list only contains folder itself inside

lujunTompson:
Context: when offline mode is enabled, folder ancestor list only contains folder itself inside. But this should be accepted and doesn't return error. 

```release-note:note
fixed the `assertParent` function to support the situation that ancestors only contain folder resource itself
```

### DIFF
--- a/mmv1/third_party/tgc/ancestrymanager/ancestryutil.go
+++ b/mmv1/third_party/tgc/ancestrymanager/ancestryutil.go
@@ -24,8 +24,8 @@ func assetParent(cai *resources.Asset, ancestors []string) (string, error) {
 				return fmt.Sprintf("//cloudresourcemanager.googleapis.com/%s", ancestors[1]), nil
 			}
 		}
-		if len(ancestors) == 1 && strings.HasPrefix(ancestors[0], "organizations/") {
-			// organizations/unknown
+		if len(ancestors) == 1 {
+			// organizations/unknown or folder itself
 			return fmt.Sprintf("//cloudresourcemanager.googleapis.com/%s", ancestors[0]), nil
 		}
 	case "cloudresourcemanager.googleapis.com/Organization":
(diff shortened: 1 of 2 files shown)


BBBmau:

roaks3 (approved):
@zli82016 would you mind weighing in? This change looks correct to me, but could use a second set of eyes.

zli82016:
@lujunTompson, what [ancestry](https://github.com/GoogleCloudPlatform/terraform-google-conversion/blob/main/cmd/tfplan2cai/convert.go#L104) do you provide with the offline mode?

How to reproduce the issue `Context: when offline mode is enabled, folder ancestor list only contains folder itself inside.`? 

Thanks.

lujunTompson:
> @lujunTompson, what [ancestry](https://github.com/GoogleCloudPlatform/terraform-google-conversion/blob/main/cmd/tfplan2cai/convert.go#L104) do you provide with the offline mode?
> 
> How to reproduce the issue `Context: when offline mode is enabled, folder ancestor list only contains folder itself inside.`?
> 
> Thanks.

@zli82016 , in our code, the resource is "folders/xxxx" and the ancestry is ["folders/xxxx"]

zli82016:
> @zli82016 , in our code, the resource is "folders/xxxx" and the ancestry is ["folders/xxxx"]

The ancestry for a folder should have one more entry, either `organizations/xxxx` or another folder `folder/folder2`,  besides itself.

lujunTompson:
> > @zli82016 , in our code, the resource is "folders/xxxx" and the ancestry is ["folders/xxxx"]
> 
> The ancestry for a folder should have one more entry, either `organizations/xxxx` or another folder `folder/folder2`, besides itself.

Our intention is to not error out and proceed with partial ancestry info when offline mode is enabled. In our case, we are ok with partial ancestry info that's why I propose this code change. Wonder why one more entry is necessary for folder resource. It should not error out with partial info because offline mode we don't fetch the ancestry info.

zli82016:
A folder must have a parent. I think that is the reason one more entry is necessary for folder resource.

lujunTompson:
Ok, sounds good. I will talk with my team members about this soon and see what's the next step. Until then I will keep this pull request open.

PR #14128, opened 2025-05-28
Remove custom expand function for 'enabled' field in SecretVersion/Re…

jalseth:
…gionalSecretVersion

```release-note:REPLACEME

```


jalseth:
This is actually needed, but should not be a part of the expand func. Will send a follow up PR.

PR #13847, opened 2025-05-04
feat: implementation of write_only support for sensitive_labels

ramonvermeulen:
Closes https://github.com/hashicorp/terraform-provider-google/issues/21855

First time implementing a `write_only` attribute, took a look at `bigquery_data_transfer` for inspiration.

**Release Note Template for Downstream PRs (will be copied)**

See [Write release notes](https://googlecloudplatform.github.io/magic-modules/contribute/release-notes/) for guidance.

```release-note:enhancement
monitoring: added `write-only` support for `sensitive_labels` via `sensitive_labels_wo` for `monitoring_notification_channel` resource
```

PR #14138, opened 2025-05-29
compute: added `shared_secret_wo` and `shared_secret_wo_version` fields to `google_compute_vpn_tunnel` + changes to `write_only` implementation in generator/templating

ramonvermeulen:
Closes https://github.com/hashicorp/terraform-provider-google/issues/23058

```release-note:enhancement
compute: added `shared_secret_wo` and `shared_secret_wo_version` fields to `google_compute_vpn_tunnel`
```


mmv1/products/compute/VpnTunnel.yaml
```diff
@@ -164,9 +164,35 @@ properties:
     description: |
       Shared secret used to set the secure session between the Cloud VPN
       gateway and the peer VPN gateway.
-    required: true
     ignore_read: true
     sensitive: true
+    exactly_one_of:
+      - 'shared_secret'
+      - 'shared_secret_wo'
+    conflicts:
```
melinath: conflicts isn't necessary since exactlyoneof is present.

melinath (changes requested):
So, you've stumbled on an area that is not well-lit and has a lot of rough edges. If you would like to make significant contributions to the generator logic, that is probably something we would accept, but we may need to come to consensus on our end on some points first. For example, we may want to automatically generate the *Version field similar to how we handle adding sibling fields for [labels fields](https://github.com/GoogleCloudPlatform/magic-modules/blob/db8c9e5caa0b1122802446aa90f7d2849790fe6d/mmv1/api/resource.go#L713).

Any contribution making significant generator changes will need to start out as _only_ changing generator logic and cleaning up various parts of the handwritten code that was necessary for the initial PR implementing a few write-only fields: https://github.com/GoogleCloudPlatform/magic-modules/pull/13135 Probably split into multiple PRs to clean up various portions. Once that refactor was done, we could then think about adding new fields.

Alternatively, you could likely follow the patterns of the original PR to add support with no template changes. In particular, you would need to add a custom_expand on the parent field and add a `raw_resource_config_validation` custom code template.

If you're open to / interested in working on the generator changes, I can put together a more formal list of refactors we'd want to see - lmk!

ramonvermeulen:
> So, you've stumbled on an area that is not well-lit and has a lot of rough edges. If you would like to make significant contributions to the generator logic, that is probably something we would accept, but we may need to come to consensus on our end on some points first. For example, we may want to automatically generate the *Version field similar to how we handle adding sibling fields for [labels fields](https://github.com/GoogleCloudPlatform/magic-modules/blob/db8c9e5caa0b1122802446aa90f7d2849790fe6d/mmv1/api/resource.go#L713).
> 
> Any contribution making significant generator changes will need to start out as _only_ changing generator logic and cleaning up various parts of the handwritten code that was necessary for the initial PR implementing a few write-only fields: #13135 Probably split into multiple PRs to clean up various portions. Once that refactor was done, we could then think about adding new fields.
> 
> Alternatively, you could likely follow the patterns of the original PR to add support with no template changes. In particular, you would need to add a custom_expand on the parent field and add a `raw_resource_config_validation` custom code template.
> 
> If you're open to / interested in working on the generator changes, I can put together a more formal list of refactors we'd want to see - lmk!

I see, and that also makes a lot of sense to me. There are most likely a lot of edge cases (that I definitely cannot grasp on my own), and changes to the generator code can (and most likely will) have significant downstream impact. Therefore I mainly made these changes to open up the discussion itself, and demonstrate how in my head the ideal yaml configuration would look like for write only arguments (combination of `api_name: 'sharedSecret'` and `write_only: true`), but I am interested in your (and other maintainers) their thoughs on this. For **this specific implementation** my changes to the generator seem to work, but I suspect it will break a lot of other implementations ;)

I am definitely open to contribute to the generator code, with the goal of making the addition of write-only arguments more standardized, straight-forward, and better documented where possible. Would love to hear your (and possibly other maintainers) there thoughts in this idea. Maybe we could open an issue in the provider, describing the desired config and list of refactors we'd want to see?

Some other thoughts:

- Good point about generating the `*Version` field similar to how adding sibling fields for [labels fields](https://github.com/GoogleCloudPlatform/magic-modules/blob/db8c9e5caa0b1122802446aa90f7d2849790fe6d/mmv1/api/resource.go#L713) is handled. This seems like a good idea to me, you need to add the `_wo_version` field anyways for every write-only implementation, so I think it will be nice if this can be standardized and generated for you. 
- If this is indeed an aspect within magic-modules that is desired to get improved, I think the refactor is easier to do now than in for example 1 or 2 years. Write-only arguments are still fairly new, but the amount of write-only implementations will probably grow over the years, making the impact of a refactor on this subject bigger and harder.
- For now moving forward with the same pattern as on the original PR #13135 seems like a good idea to me, I will take a look and see if I am able to implement write-only attributes for this PR in a similar manner. Would you like to see that on this specific PR, or rather keep this one open for reference for the refactor? And open another PR to implement the `shared_secret_wo` for `google_compute_vpn_tunnel` in the patterns of the original PR?

melinath:
> If this is indeed an aspect within magic-modules that is desired to get improved, I think the refactor is easier to do now than in for example 1 or 2 years. Write-only arguments are still fairly new, but the amount of write-only implementations will probably grow over the years, making the impact of a refactor on this subject bigger and harder.

Big +1 to this.

> For now moving forward with the same pattern as on the original PR https://github.com/GoogleCloudPlatform/magic-modules/pull/13135 seems like a good idea to me, I will take a look and see if I am able to implement write-only attributes for this PR in a similar manner. Would you like to see that on this specific PR, or rather keep this one open for reference for the refactor? And open another PR to implement the shared_secret_wo for google_compute_vpn_tunnel in the patterns of the original PR?

Yeah, if you'd like to move forward with this specific change before modifying the generator logic, doing so on this PR seems reasonable.

melinath:
@ramonvermeulen After talking things over with @BBBmau, we'd actually like to ask you to make the generator changes _first_, then the resource changes after. We're working on putting together a plan for what generator changes will be required - should have that ready in a few days.

Apologies for the back-and-forth!

ramonvermeulen:
> @ramonvermeulen After talking things over with @BBBmau, we'd actually like to ask you to make the generator changes _first_, then the resource changes after. We're working on putting together a plan for what generator changes will be required - should have that ready in a few days.
> 
> Apologies for the back-and-forth!

No problem, haven't worked on this in the meantime.

Sounds like a good idea! Might not be worth the effort to first do a more custom implementation on this PR, then make the generator changes, and then have to refactor the "more custom implementation" from this PR right away.

BBBmau:
@ramonvermeulen thanks for the patience! We've opened github issues in regards to the refactoring work you've expressed interest in 
- https://github.com/hashicorp/terraform-provider-google/issues/23210#issue-3131689461

You'll want to open the PRs onto a refactoring branch (`refactor-write-only`). Once we have all that's needed for the refactor merged we can then proceed with merging `refactor-write-only` into main.

melinath:
@BBBmau wouldn't it be easier to merge each part of the fix directly into main like normal? The incremental changes should be safe?

BBBmau:
> @BBBmau wouldn't it be easier to merge each part of the fix directly into main like normal? The incremental changes should be safe?

we can do normal merges to main also since like you said these changes are safe to do.

ramonvermeulen:
@melinath @BBBmau 

Thanks for following up on this! 🙌

Just a small FYI, from this Thursday 12th till Monday 16th I am out-of-office, so I will probably follow up on most of the work after the 16th. I am fine with either approaches, of doing all the subtasks in one big PR (e.g. `refactor-write-only` branch), or each in an individual PR (e.g. `refactor-write-only-<issue-number>` branches).

Are you sure the incremental changes should be safe to merge separately? Just high-over thinking about the refactor, for example I can imagine https://github.com/hashicorp/terraform-provider-google/issues/23213 (eliminating the need of `*_wo_version`) conflicting with "existing" wo implementations, resulting in for example a double wo_version (e.g. original one + the one being generated with new generation code). Unless it will really be based upon a field configuration such as `generate_wo_version: true` next to `write_only: true`.

**EDIT:**

Another thought on the generated `*_wo_version`, I guess the `required_with` logic should also be automatically generated, e.g. where you used to do:

```yaml
        required_with:
          - 'password_wo'
```

Also, do we still want the user to set this on a write-only field or generate such logic as well?:

```yaml
        conflicts:
          - 'password'
```

I think it might be better to keep that part as it is, because it offers slightly more flexibility. Maybe we also need to add a subtask to add/update the documentation for users on how to add write-only fields?

Desired end config from a user perspective would look like?:

```yaml
  - name: 'password'
    type: String
    description: |
      foo
    ignore_read: true
    sensitive: true
    # cannot configure both at the same time
    exactly_one_of:
      - 'password'
      - 'password_wo'
    conflicts:
      - 'password_wo'
  - name: 'passwordWo'
    type: String
    description: |
      bar
    # so that the same underlaying field will be used during resource creation
    api_name: 'password'
    write_only: true
    # cannot configure both at the same time (not sure if this configuration is required on both ends)
    exactly_one_of:
      - 'password'
      - 'password_wo'
    conflicts:
      - 'password'
    # not sure if we want to do this "explicit" or "implicit", in case we want this "explicit":
    generate_wo_version: true
```

melinath:
we should avoid doing a single large branch. That said, each incremental change will need to be _complete_. For example, generating expander behavior should be in a separate PR from eliminating manual *_wo_version fields, but removing *_wo_version fields should _include_ replacing the existing *_wo_version fields. (It might be possible to split some of the tickets into multiple PR stages for easier review.)

melinath:
regarding the UX, see https://github.com/hashicorp/terraform-provider-google/issues/23213#issuecomment-2959923206

ramonvermeulen:
> we should avoid doing a single large branch. That said, each incremental change will need to be _complete_. For example, generating expander behavior should be in a separate PR from eliminating manual *_wo_version fields, but removing *_wo_version fields should _include_ replacing the existing *_wo_version fields. (It might be possible to split some of the tickets into multiple PR stages for easier review.)

@melinath @BBBmau 
Check, for generating the `_wo` and `_wo_version` properties, would it make more sense to do this from within the templating, or even before the templating within the data structure? 

~For example by modifying the following method:
https://github.com/GoogleCloudPlatform/magic-modules/blob/80c697e199a05c12ff68fb882fcf66d6fe3f4f0a/tpgtools/property.go#L517~

~I think this method loops over all the "parsed" properties, and applies recursion at several places, for instance for `Object` type properties. Thinking about it, it might make more sense to already solve the problem from within the data structure itself (which is the input for the templating).~

Another option is to do it slightly later, maybe in a similar manner as:
https://github.com/GoogleCloudPlatform/magic-modules/blob/80c697e199a05c12ff68fb882fcf66d6fe3f4f0a/mmv1/main.go#L240

Would love to hear your opinion(s) about this approach.

melinath:
Adding the new fields should happen at yaml parsing time, so that they look like they've "always been there" and can be handled like any other fields. It should be similar to how we handle label-related fields. I'll put more details on the ticket.

ramonvermeulen:
> Adding the new fields should happen at yaml parsing time, so that they look like they've "always been there" and can be handled like any other fields. It should be similar to how we handle label-related fields. I'll put more details on the ticket.

Check, this makes sense, so similar as:
https://github.com/GoogleCloudPlatform/magic-modules/blob/80c697e199a05c12ff68fb882fcf66d6fe3f4f0a/mmv1/main.go#L240
https://github.com/GoogleCloudPlatform/magic-modules/blob/80c697e199a05c12ff68fb882fcf66d6fe3f4f0a/mmv1/api/resource.go#L713

And also apply recursion for `NestedObject` type properties to support "nested" write-only properties.

ramonvermeulen:
@melinath tagging for visibility, since this is the first PR using the new write-only system.

With this PR I found out that I still had to do some changes to the generated expander logic to make this work (https://github.com/hashicorp/terraform-provider-google/issues/23211). At first glance I thought it wasn't necessary because for the existing _wo arguments it generated without issues, but apparently it is required for this one. 

For example, if it stays just `{{ $prop.ApiName }}Prop := ` ... you will get conflicting variable names, and `d.Get` cannot be used to retrieve the write-only value, instead [d.GetRawConfig(...)](https://developer.hashicorp.com/terraform/plugin/sdkv2/resources/write-only-arguments#retrieving-write-only-values) should be used. I could also move this to a seperate PR if that is preferred.

**Tests:**
```
❮ make testacc TEST=./google/services/compute TESTARGS='-run=TestAccComputeVpnTunnel_ -parallel 3'
TF_ACC=1 TF_SCHEMA_PANIC_ON_ERROR=1 go test ./google/services/compute -v -run=TestAccComputeVpnTunnel_ -parallel 3 -timeout 240m -ldflags="-X=github.com/hashicorp/terraform-provider-google/version.ProviderVersion=acc"
=== RUN   TestAccComputeVpnTunnel_vpnTunnelBasicExample
=== PAUSE TestAccComputeVpnTunnel_vpnTunnelBasicExample
=== RUN   TestAccComputeVpnTunnel_regionFromGateway
=== PAUSE TestAccComputeVpnTunnel_regionFromGateway
=== RUN   TestAccComputeVpnTunnel_router
=== PAUSE TestAccComputeVpnTunnel_router
=== RUN   TestAccComputeVpnTunnel_routerWithSharedSecretWo_update
=== PAUSE TestAccComputeVpnTunnel_routerWithSharedSecretWo_update
=== RUN   TestAccComputeVpnTunnel_defaultTrafficSelectors
=== PAUSE TestAccComputeVpnTunnel_defaultTrafficSelectors
=== CONT  TestAccComputeVpnTunnel_vpnTunnelBasicExample
=== CONT  TestAccComputeVpnTunnel_routerWithSharedSecretWo_update
=== CONT  TestAccComputeVpnTunnel_defaultTrafficSelectors
--- PASS: TestAccComputeVpnTunnel_vpnTunnelBasicExample (224.41s)
=== CONT  TestAccComputeVpnTunnel_router
--- PASS: TestAccComputeVpnTunnel_defaultTrafficSelectors (252.69s)
=== CONT  TestAccComputeVpnTunnel_regionFromGateway
--- PASS: TestAccComputeVpnTunnel_routerWithSharedSecretWo_update (318.62s)
--- PASS: TestAccComputeVpnTunnel_router (156.78s)
--- PASS: TestAccComputeVpnTunnel_regionFromGateway (228.20s)
PASS
ok      github.com/hashicorp/terraform-provider-google/google/services/compute  481.826s
```

<hr>

P.S. I also played a bit around on [another PR](https://github.com/GoogleCloudPlatform/magic-modules/pull/13847) for write-only arguments and found out that currently `exactly_one_of` will cause some problems when you have multiple write-only fields. Just a small heads-up that the write-only generation probably still requires a couple iterations before it tackles all cases.

melinath:
> Just a small heads-up that the write-only generation probably still requires a couple iterations before it tackles all cases.

Yep, that's expected! Thanks for flagging.

mmv1/templates/terraform/resource.go.tmpl
```diff
@@ -170,6 +170,34 @@ func resource{{ $.ResourceName }}{{ camelize $prop.Name "upper" }}SetStyleDiff(_
 }
 {{- end}}
 
+
+func resource{{ $.ResourceName -}}GetRawConfigAttributeAsString(d *schema.ResourceData, key string) string {
```
melinath: Rather than generating this function per-resource, could you make it a helper function? Nothing here seems to rely on knowledge of a specific resource. (Overall I'd like us to have more helper functions for accessing ResourceData values so that we can document & reproduce specific behaviors across handwritten and generated resources instead of relying on the generator to do the right thing and just winging it for handwritten.)

The helper would definitely go in `tpgresource` - maybe in https://github.com/hashicorp/terraform-provider-google-beta/blob/main/google-beta/tpgresource/utils.go?

ramonvermeulen: Sure, also makes more sense than doing everything within templates. I just wasn't fully sure if I can call `tpgresource` directly from the template, but looking at the generated code it is always imported so that should be fine. Thanks for the suggestion, is indeed a cleaner approach.

ramonvermeulen:
> > With this PR I found out that I still had to do some changes to the generated expander logic to make this work ([hashicorp/terraform-provider-google#23211](https://github.com/hashicorp/terraform-provider-google/issues/23211)). At first glance I thought it wasn't necessary because for the existing _wo arguments it generated without issues, but apparently it is required for this one.
> 
> Yeah, so it would be good to know why the existing fields don't seem to have a problem with this!
> 
> Taking a look, it seems like:
> 
> * `google_bigquery_data_transfer_config`: sensitiveParams is (incorrectly IMO) marked as url_param_only, which means it "doesn't get sent". Then the resource has a custom encoder that handles setting the sensitiveParams value in the API request, including handling the write-only field. This would be nice to fix if possible but isn't a blocker.
> * `google_monitoring_uptime_check_config`: the field is nested, so the generated code is inside an expander, which uses [this template](https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/templates/terraform/expand_property_method.go.tmpl) (as opposed to the field you're working on here, which is top-level). The expander template uses `transformed{{$prop.TitlelizeProperty}}` as the variable name, instead of relying on `$prop.ApiName` - I would be inclined to match that if possible rather than create a new helper function.
> * `google_secret_manager_secret_version`: Nested field. But also, it uses a custom expander for the parent field to handle the expansion, so it doesn't rely on the templates to expand the write_only field.
> 
> So yeah, I'd recommend changing the top-level template to use `$prop.TitlelizeProperty` for variable names. That should happen in a separate PR because it could potentially change every resource, so it should be isolated.

Thanks a lot for your investigation on this! I will put all template related things into a separate PR, but I will ensure that this PR includes the exact same changes so that I can test it easily. Will tag you when the PR is ready. Eventually went with a new `$prop.CamelizeProperty` instead of `$prop.TitlelizeProperty` so that most prop variable names stay the same. When I used `$prop.TitlelizeProperty` a ton of tests started failing.

melinath (commented):
I'm removing this from my queue for now since there's a couple other in-flight PRs related to this - when it's time to get back to this one, just leave a comment or re-request review from me.

ramonvermeulen:
closed by https://github.com/GoogleCloudPlatform/magic-modules/pull/14933

PR #14146, opened 2025-05-29
Check for fields that are needed for the read URL explicitly

melinath:
Follow-up to https://github.com/GoogleCloudPlatform/magic-modules/pull/13836. It turns out that active directory peering has divergent id_format and read url. Even though that's the only case right now, I think we should just handle both to be safe.

```release-note:none

```

### DIFF
--- a/mmv1/api/resource.go
+++ b/mmv1/api/resource.go
@@ -1218,8 +1218,15 @@ func (r Resource) GetIdFormat() string {
 
 // Returns true if the Type is in the ID format and false otherwise.
 func (r Resource) InPostCreateComputed(prop Type) bool {
-	fields := r.ExtractIdentifiers(r.GetIdFormat())
-	return slices.Contains(fields, google.Underscore(prop.Name))
+	fields := map[string]struct{}{}
+	for _, f := range r.ExtractIdentifiers(r.GetIdFormat()) {
+		fields[f] = struct{}{}
+	}
+	for _, f := range r.ExtractIdentifiers(r.SelfLinkUri()) {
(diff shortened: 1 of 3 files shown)


melinath:
there are a couple other cases to add here - cancelling the test run.

melinath:
So, I thought there was a case where a field marked as URL param only could end up not being set correctly, such as https://github.com/modular-magician/terraform-provider-google-beta/blob/dc199f02bfae6811c8f0100d4c2b0480d2eb2e21/google-beta/services/apigateway/resource_api_gateway_api_config.go.

However, looking at the history of that file, we were _never_ setting that field correctly as part of the ID. https://github.com/hashicorp/terraform-provider-google-beta/commits/main/google-beta/services/apigateway/resource_api_gateway_api_config.go We don't actually have a way to set it at all because it's a URL param.

So, running tests again.

mmv1/api/resource.go
```diff
@@ -1217,20 +1217,30 @@ func (r Resource) GetIdFormat() string {
 }
 
 // Returns true if the Type is in the ID format and false otherwise.
-func (r Resource) InIdFormat(prop Type) bool {
-	fields := r.ExtractIdentifiers(r.GetIdFormat())
-	return slices.Contains(fields, google.Underscore(prop.Name))
+func (r Resource) InPostCreateComputed(prop Type) bool {
```
ScottSuarez, line 1220: I am not sure I understand the name of this new function. InIdFormat made sense because it was checking presence of field in ids. This also seems to do the same thing?

ScottSuarez (approved):

PR #14152, opened 2025-05-30
Removed Composer 1 tests

rlapin-pl:

```release-note: none
cleanup

```


roaks3 (commented):
This conceptually seems fine, but 2 questions:
1. Is there a rationale I can take a look at, perhaps a bug or something? I know the composer v1 is older, but is it unavailable/deprecated?
2. The GitHub diff is a bit unfortunate and hard for me to parse. I assume you removed code and then rearranged things? Were any other changes made that need review? If possible, separating commits could help here (one to remove the code, one to re-order as you see fit)

rlapin-pl:
1) You can find the list of github issues in the descreption. One of the issue is https://github.com/hashicorp/terraform-provider-google/issues/23030.  We deprecate creation of C1 environments and that is why we decided to fix the issue with missing service accounts to remove these tests.
2) I've removed Composer 1 tests and I see some autoformatting in the imports.

Also there is a followup pullrequest where I added service accounts in other tests.https://github.com/GoogleCloudPlatform/magic-modules/pull/14153

PR #14153, opened 2025-05-30
Fix service account issue

rlapin-pl:

```release-note:none

```


rlapin-pl:
[NickElliot](https://github.com/NickElliot) could you please review my changes.

NickElliot (commented):
updates look good, but I wanted to verify the reduced number of test cases, were all these cases not able to be updated for using service accounts + environment, or was there a meaningful amount of redundancy?

rlapin-pl:
Composer 1 tests were removed. They also were failing but as we are going to deprecate Composer 1 creation we decided to delete these tests instead of fixing them.

NickElliot (changes requested):
got it! can you also update the composer environment docs to clarify that `node_config.service_account` is a required field for all newly created environments?

rlapin-pl:
It is still an optional field in the Environment schema and currently creation is only blocked by our API. It seems that we can have some cases when we can have this field as empty.

I would update the documentation once we decide to mark the field as required in the schema.

NickElliot:
that aligns with what Im saying, I want the description to say "Required for newly created environments" as a purely textual addition, as you say it can still be empty for existing resources

rlapin-pl:
Updated documentation https://github.com/GoogleCloudPlatform/magic-modules/pull/14153/commits/e20022afc451b51be4ffbac9115f17ebdcd777e8

PR #14154, opened 2025-05-30
Add regex to handle image link in universes

hao-nan-li:

```release-note:none

```

### DIFF
--- a/mmv1/templates/terraform/constants/disk.tmpl
+++ b/mmv1/templates/terraform/constants/disk.tmpl
@@ -43,16 +43,18 @@ func IsDiskShrinkage(_ context.Context, old, new, _ interface{}) bool {
 	return new.(int) < old.(int)
 }
 
-func matchImageLink(old string) ([]string) {
+func matchImageLink(old string) (string, string, bool) {
 	// 'old' is read from the API.
-	// It has the format 'https://www.googleapis.com/compute/v1/projects/(%s)/global/images/(%s)'
-	// Except if the image is created in different universes
+	// In GCP It has the format 'https://www.googleapis.com/compute/v1/projects/(%s)/global/images/(%s)'
 	matches := resolveImageLink.FindStringSubmatch(old)
 	if matches == nil {
-		// Need to do an extra check if the image is created is in different universes
+		// In alternate universes, it has the format https://compute.%s/compute/[a-z0-9]+/projects/(%s)/global/images/(%s)
 		matches = resolveImageUniverseLink.FindStringSubmatch(old)
+		if matches == nil {
+			return "", "", false
+		}
 	}
-	return matches
+	return matches[1], matches[2], true
 }
 
 // We cannot suppress the diff for the case when family name is not part of the image name since we can't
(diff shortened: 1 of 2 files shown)


mmv1/third_party/terraform/services/compute/image.go
```diff
@@ -28,6 +29,7 @@ var (
 	resolveImageFamily                 = regexp.MustCompile(fmt.Sprintf("^(%s)$", resolveImageFamilyRegex))
 	resolveImageImage                  = regexp.MustCompile(fmt.Sprintf("^(%s)$", resolveImageImageRegex))
 	resolveImageLink                   = regexp.MustCompile(fmt.Sprintf("^https://www.googleapis.com/compute/[a-z0-9]+/projects/(%s)/global/images/(%s)", verify.ProjectRegex, resolveImageImageRegex))
```
rileykarson, line 31: Can you take steps to remove the modification of this global value inside of ResolveImage and use the same function you just introduced there too? Effectively, we want to encapsulate "identify and/or break up an image self link" into a function now that it is sufficiently complicated.

mmv1/templates/terraform/constants/disk.tmpl
```diff
@@ -43,6 +43,18 @@ func IsDiskShrinkage(_ context.Context, old, new, _ interface{}) bool {
 	return new.(int) < old.(int)
 }
 
+func matchImageLink(old string) ([]string) {
+	// 'old' is read from the API.
+	// It has the format 'https://www.googleapis.com/compute/v1/projects/(%s)/global/images/(%s)'
```
rileykarson, line 48:
```suggestion
	// In GCP it has the format 'https://www.googleapis.com/compute/v1/projects/(%s)/global/images/(%s)'
```

mmv1/templates/terraform/constants/disk.tmpl
```diff
@@ -43,6 +43,18 @@ func IsDiskShrinkage(_ context.Context, old, new, _ interface{}) bool {
 	return new.(int) < old.(int)
 }
 
+func matchImageLink(old string) ([]string) {
```
rileykarson, line 46: Since we're in a dedicated function now, should we just return the values (and possibly a success/failure boolean) directly rather than just return the array? That'll be clearer to callers.

mmv1/templates/terraform/constants/disk.tmpl
```diff
@@ -43,6 +43,18 @@ func IsDiskShrinkage(_ context.Context, old, new, _ interface{}) bool {
...
+	// 'old' is read from the API.
+	// It has the format 'https://www.googleapis.com/compute/v1/projects/(%s)/global/images/(%s)'
+	// Except if the image is created in different universes
```
rileykarson, line 49:
```suggestion
```

mmv1/templates/terraform/constants/disk.tmpl
```diff
@@ -43,6 +43,18 @@ func IsDiskShrinkage(_ context.Context, old, new, _ interface{}) bool {
...
+	matches := resolveImageLink.FindStringSubmatch(old)
+	if matches == nil {
```
rileykarson, line 51:
```suggestion
	// In alternate universes, it has the format https://compute.%s/compute/[a-z0-9]+/projects/(%s)/global/images/(%s)
	if matches == nil {
```

mmv1/third_party/terraform/services/compute/image.go
hao-nan-li: For this commit, I took the override out and added a case for `resolveImageUniverseLink`

mmv1/templates/terraform/constants/disk.tmpl
hao-nan-li: Applied the comment to this line. But I prefer to keep the first if check since we only do resolveImageUniverseLink if resolveImageLink didn't match

hao-nan-li:
> LGTM- can you confirm this resolves the issue you were able to repro?

Attached execution log on the buganizer ticket.

PR #14156, opened 2025-05-30
tgc-revival: implement testing logic

zli82016:
Implement the main testing logic.
1. Read the nightly tests from the GCS bucket
2. Each test has raw_config and export_asset as the input
  * Step 1: Use cai2hcl to convert export assets into a Terraform configuration (export config).
	Compare all of the fields in raw_config are in export_config.
  * Step 2: Run a terraform plan using export_config.
	Use tfplan2cai to convert the generated plan into CAI assets (roundtrip_assets).
	Convert roundtrip_assets back into a Terraform configuration (roundtrip_config) using cai2hcl.
	Compare roundtrip_config with export_config to ensure they are identical.

```release-note:none

```

### DIFF
--- a/mmv1/third_party/tgc_next/test/assert_test_files.go
+++ b/mmv1/third_party/tgc_next/test/assert_test_files.go
@@ -17,13 +17,10 @@ import (
 	"github.com/GoogleCloudPlatform/terraform-google-conversion/v6/pkg/tfplan2cai"
 	tfplan2caiconverters "github.com/GoogleCloudPlatform/terraform-google-conversion/v6/pkg/tfplan2cai/converters"
 
+	"go.uber.org/zap"
 	"go.uber.org/zap/zaptest"
 
 	"github.com/google/go-cmp/cmp"
-
-	"github.com/hashicorp/hcl/v2"
-	"github.com/hashicorp/hcl/v2/hclparse"
-	"github.com/hashicorp/hcl/v2/hclsyntax"
 )
 
 var (
@@ -31,14 +28,15 @@ var (
 	tmpDir     = os.TempDir()
 )
 
-func AssertTestFile(t *testing.T, ignoredFields []string) {
-	var err error
-	cacheMutex.Lock()
-	TestsMetadata, err = ReadTestsDataFromGcs()
+func TestBidirectionalConversion(t *testing.T, ignoredFields []string) {
+	resourceTestData, primaryResource, err := prepareTestData(t.Name())
 	if err != nil {
-		log.Fatal(err)
+		t.Fatalf("Error preparing the input data: %#v", err)
 	}
-	cacheMutex.Unlock()
+
+	if resourceTestData == nil {
+		t.Skipf("The test data is unavailable.")
+	}
 
 	// Create a temporary directory for running terraform.
 	tfDir, err := os.MkdirTemp(tmpDir, "terraform")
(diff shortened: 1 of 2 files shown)


mmv1/third_party/tgc_next/test/assert_test_files.go
```diff
@@ -0,0 +1,399 @@
...
+	resourceMetadata := testMetadata.ResourceMetadata
+	if len(resourceMetadata) == 0 {
+		log.Printf("Data of test is unavailable: %s", t.Name())
+		return
+	}
+
+	rawTfFile := fmt.Sprintf("%s.tf", t.Name())
+	err = os.WriteFile(rawTfFile, []byte(testMetadata.RawConfig), 0644)
+	if err != nil {
+		log.Fatalf("error writing to file %s: %#v", rawTfFile, err)
```
melinath: should these use `t.Fatalf` instead so that they cause the test to fail gracefully instead of causing os.Exit?

mmv1/third_party/tgc_next/test/assert_test_files.go
```diff
@@ -0,0 +1,399 @@
...
+			testSingleResource(t, metadata, rawConfigMap[address], tfDir, ignoredFields)
+		}
+	}
+}
+
+// Tests a single resource
+func testSingleResource(t *testing.T, resourceMetadata *ResourceMetadata, rawConfig map[string]interface{}, tfDir string, ignoredFields []string) {
+	resourceType := resourceMetadata.ResourceType
+	if _, ok := tfplan2caiconverters.ConverterMap[resourceType]; !ok {
+		log.Printf("Test %s fails as tfplan2cai conversion for %s is not supported.", t.Name(), resourceType)
```
melinath: Here & elsewhere - if this is considered a failure mode it should either use t.Fatalf or this function should return an error that triggers t.Fatalf in the caller.

```suggestion
		t.Fatalf("Test %s fails as tfplan2cai conversion for %s is not supported.", t.Name(), resourceType)
```

mmv1/third_party/tgc_next/test/assert_test_files.go
```diff
@@ -0,0 +1,399 @@
...
+		testSingleResource(t, primaryMetadata, rawConfigMap[address], tfDir, ignoredFields)
+	} else {
+		for address, metadata := range resourceMetadata {
+			testSingleResource(t, metadata, rawConfigMap[address], tfDir, ignoredFields)
+		}
+	}
+}
+
+// Tests a single resource
+func testSingleResource(t *testing.T, resourceMetadata *ResourceMetadata, rawConfig map[string]interface{}, tfDir string, ignoredFields []string) {
```
melinath: I'm inclined to suggest removing `t *testing.T` from the signature and having the caller handle errors but I don't feel strongly about it.

```suggestion
func testSingleResource(testName string, resourceMetadata *ResourceMetadata, rawConfig map[string]interface{}, tfDir string, ignoredFields []string) error {
```

mmv1/third_party/tgc_next/test/assert_test_files.go
```diff
@@ -0,0 +1,399 @@
...
+	"github.com/hashicorp/hcl/v2/hclparse"
+	"github.com/hashicorp/hcl/v2/hclsyntax"
+)
+
+var (
+	cacheMutex = sync.Mutex{}
+	tmpDir     = os.TempDir()
+)
+
+func AssertTestFile(t *testing.T, ignoredFields []string) {
```
melinath: This name is a little bit confusing - it's not clear what's being asserted. Maybe something like `AssertRoundtripConversionWorks` or `AssertConversionResultsCorrect`? I'm not really satisfied with either of those either, though.

Alternatively, you could structure the functions differently:

- Function for downloading & parsing the files (doesn't automatically run tests)
- Function for testing a single resource's cai2hcl & roundtrip behavior (since roundtrip depends on cai2hcl's output splitting them might not make sense)

This would essentially mean that callers are responsible for determining whether to run for one resource in the config or all of them; they would also be able to choose how to handle data loading and parsing errors vs how to handle actual problems in the comparison vs expected output. You'd need to also think about what data structure would be best for the loading & parsing function to return.

I'm pretty ambivalent about the structure in this moment, but the entrypoint(s) definitely needs to have a more clear name.

mmv1/third_party/tgc_next/test/assert_test_files.go
zli82016: That is right.

mmv1/third_party/tgc_next/test/assert_test_files.go
```diff
@@ -0,0 +1,281 @@
...
+
+	return missingKeys
+}
+
+// Converts a tfplan to CAI asset, and then converts the CAI asset into HCL
+func getRoundtripConfig(t *testing.T, testName string, tfDir string, ancestryCache map[string]string, logger *zap.Logger) ([]byte, error) {
+	fileName := fmt.Sprintf("%s_export", testName)
+
+	// Run terraform init and terraform apply to generate tfplan.json files
+	terraformWorkflow(t, tfDir, fileName)
```
zli82016, line 245: I am trying to remove `t *testing.T` from the signature of the functions and return error. 

The function `terraformWorkflow` needs the same modification. I will do that in a separate PR later and then remove  `t *testing.T` from the signature of the functions.

mmv1/third_party/tgc_next/test/assert_test_files.go
zli82016: It is a good practice to remove `t *testing.T` from the signature and return errors.

mmv1/third_party/tgc_next/test/assert_test_files.go
zli82016: What about the name `TestBidirectionalConversion`?

I put the downloading and parsing the files to a function `prepareTestData`

melinath: I think `TestBidirectionalConversion` would be run as a test since it starts with `Test`, which could cause problems. What about just `BidirectionalConversion` since this is inside the `test` package? So it would be `test.BidirectionalConversion`. Alternatively, `CompareBidirectionalConversion` or `BidirectionalConversionDiff` could work?

zli82016: `test.BidirectionalConversion` sounds good!

melinath (approved):

PR #14164, opened 2025-05-30
feat: added 'data_source_id' for update requests through  bigquery_data_transfer_config

sachinpro:
b/421406404

The custom org policy in this bug was not able to verify what value was in the “dataSourceId” field (which is a requirement for that custom org policy) during update requests and hence was denying the access.

```release-note:enhancement
bigquery: added `data_source_id` to update requests through `google_bigquery_data_transfer_config`
```


c2thorn (approved):
I can see the `dataSourceId` in the debug log now.

PR #14166, opened 2025-05-31
`mmv1/identity`: add `CustomImportSchema` field in yanl

BBBmau:
There are cases where the ExtractIdentifiers doesn't return fields that match the regex:

image

Having a `CustomImportSchema` will resolve this case as it's only in two resources

```release-note:none

```


BBBmau:
> When will you add the custom identity schema to the resources you've identified need it?

they'd be added after this PR is merged, though if it's easier I can keep it all in a single PR the resources that require the `CustomImportSchema` is low.

c2thorn:
> > When will you add the custom identity schema to the resources you've identified need it?
> 
> they'd be added after this PR is merged, though if it's easier I can keep it all in a single PR the resources that require the `CustomImportSchema` is low.

up to you, I think it could make sense to add here

c2thorn (approved):

PR #14167, opened 2025-05-31
Add suspended field to dataplex data scan rules

uedar:
Fixes https://github.com/hashicorp/terraform-provider-google/issues/23123
```release-note:enhancement
dataplex: added `suspended` field to `google_dataplex_datascan` resource
```

### DIFF
--- a/mmv1/products/dataplex/Datascan.yaml
+++ b/mmv1/products/dataplex/Datascan.yaml
@@ -369,6 +369,7 @@ properties:
               type: Boolean
               description: |
                 Whether the Rule is active or suspended. Default = false.
+              default_value: false
             - name: 'description'
               type: String
               description: |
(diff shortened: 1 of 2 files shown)


mmv1/products/dataplex/Datascan.yaml
```diff
@@ -341,6 +341,10 @@ properties:
                 The maximum length is 63 characters.
                 Must start with a letter.
                 Must end with a number or a letter.
+            - name: 'suspended'
```
hao-nan-li, line 368: If this field has a default value, could you add it in? Also can you add this field to an existing test?

uedar: Sorry for the late reply...
I've added a default value to the field and updated the existing test.
https://github.com/GoogleCloudPlatform/magic-modules/pull/14167/commits/1484a04acf59776e0194febb0c57fa984ff49fe3
https://github.com/GoogleCloudPlatform/magic-modules/pull/14167/commits/ceac2443213d26882ea8a02a32a06b6aa6e14a41

PR #14168, opened 2025-06-01
Add a custom role field for the Scope RBACRolebindings

tonybayvas:

```release-note:enhancement
gkehub: added `custom_role` field to `google_gke_hub_scope_rbac_role_binding` resource
```

### DIFF
--- a/mmv1/third_party/terraform/services/gkehub2/resource_gke_hub_scope_rbac_role_binding_test.go
+++ b/mmv1/third_party/terraform/services/gkehub2/resource_gke_hub_scope_rbac_role_binding_test.go
@@ -53,7 +53,7 @@ resource "google_gke_hub_scope" "scoperbacrolebinding" {
 
 resource "google_gke_hub_scope_rbac_role_binding" "scoperbacrolebinding" {
   scope_rbac_role_binding_id = "tf-test-scope-rbac-role-binding%{random_suffix}"
-  scope_id = "tf-test-scope%{random_suffix}"
+  scope_id = google_gke_hub_scope.scoperbacrolebinding.scope_id
   user = "test-email@gmail.com"
   role {
     predefined_role = "ADMIN"
@@ -73,7 +73,7 @@ resource "google_gke_hub_scope" "scoperbacrolebinding" {
 
 resource "google_gke_hub_scope_rbac_role_binding" "scoperbacrolebinding" {
   scope_rbac_role_binding_id = "tf-test-scope-rbac-role-binding%{random_suffix}"
-  scope_id = "tf-test-scope%{random_suffix}"
+  scope_id = google_gke_hub_scope.scoperbacrolebinding.scope_id
   group = "test-email2@gmail.com"
   role {
     predefined_role = "VIEW"


tonybayvas:
Hi @roaks3, this PR is ready for review. Could you please take a look? Thank you

mmv1/templates/terraform/examples/gkehub_scope_rbac_custom_role_binding_basic.tf.tmpl
```diff
@@ -0,0 +1,26 @@
...
+  spec {
+    rbacrolebindingactuation {
+      allowed_custom_roles = ["my-custom-role"]
+    }
+  }
+}
+
+resource "google_gke_hub_scope_rbac_role_binding" "{{$.PrimaryResourceId}}" {
+  scope_rbac_role_binding_id = "tf-test-scope-rbac-role-binding%{random_suffix}"
+  scope_id = google_gke_hub_scope.scope.scope_id
```
corentone, line 17: this dependency may allow you to skip the depends_on?

or maybe only if its an output?

mmv1/products/gkehub2/ScopeRBACRoleBinding.yaml
```diff
@@ -51,6 +51,11 @@ examples:
     primary_resource_name: 'fmt.Sprintf(\"tf-test-scope%s\", context[\"random_suffix\"]), fmt.Sprintf(\"tf-test-rbac-role-binding%s\", context[\"random_suffix\"])'
     test_env_vars:
       project: 'PROJECT_NAME'
+  - name: 'gkehub_scope_rbac_custom_role_binding_basic'
```
roaks3, line 54: I think you may want `exclude_test: true` here so that it doesn't generate a test (this is an assumption based of some of the existing tests you've written that seem to require a dedicated project, presumably because the `rbacrolebindingactuation` is global)

mmv1/products/gkehub2/ScopeRBACRoleBinding.yaml
```diff
@@ -143,6 +148,16 @@ properties:
           - 'ADMIN'
           - 'EDIT'
           - 'VIEW'
+        exactly_one_of:
+          - 'predefinedRole'
```
roaks3: I believe these need to be defined like `role.0.predefined_role`

mmv1/templates/terraform/examples/gkehub_scope_rbac_custom_role_binding_basic.tf.tmpl
roaks3: Yea, I think you can remove `google_gke_hub_scope.scope` from `depends_on`

mmv1/third_party/terraform/services/gkehub2/resource_gke_hub_scope_rbac_role_binding_test.go
```diff
@@ -84,3 +84,127 @@ resource "google_gke_hub_scope_rbac_role_binding" "scoperbacrolebinding" {
 }
 `, context)
 }
+
+func TestAccGKEHub2ScopeRBACRoleBinding_gkehubScopeRbacCustomRoleBindingBasicExample_update(t *testing.T) {
```
roaks3, line 88: Per some of the existing comments I'm seeing in the code, note that you may need to skip VCR for this test (since multiple services need to be enabled, VCR might have trouble with the batching)

mmv1/third_party/terraform/services/gkehub2/resource_gke_hub_scope_rbac_role_binding_test.go
```diff
@@ -84,3 +84,127 @@ resource "google_gke_hub_scope_rbac_role_binding" "scoperbacrolebinding" {
...
+
+resource "google_gke_hub_feature" "rbacrolebindingactuation" {
+  name = "rbacrolebindingactuation"
+  location = "global"
+  spec {
+    rbacrolebindingactuation {
+      allowed_custom_roles = ["my-custom-role", "my-custom-role-2"]
+    }
+  }
+  depends_on = [google_project_service.anthos, google_project_service.gkehub]
```
roaks3, line 174: Nit: it seems like the `google_gke_hub_scope ` would also need to depend on `google_project_service.gkehub`?

roaks3 (changes requested):
Some relatively minor comments, but mostly looks good

mmv1/templates/terraform/examples/gkehub_scope_rbac_custom_role_binding_basic.tf.tmpl
tonybayvas: Thanks, I see the scope_id is a required field, removed the depends_on from both the predefined and custom role examples

mmv1/products/gkehub2/ScopeRBACRoleBinding.yaml
tonybayvas: Done, added it for both examples.

mmv1/products/gkehub2/ScopeRBACRoleBinding.yaml
tonybayvas: Done

mmv1/third_party/terraform/services/gkehub2/resource_gke_hub_scope_rbac_role_binding_test.go
tonybayvas: I added acctest.SkipIfVcr(t), please let me know if that's what you meant

mmv1/third_party/terraform/services/gkehub2/resource_gke_hub_scope_rbac_role_binding_test.go
tonybayvas: Added the APIs, but it does work without these dependencies

tonybayvas:
Hi @roaks3, this PR is ready for another round of review

mmv1/third_party/terraform/services/gkehub2/resource_gke_hub_scope_rbac_role_binding_test.go
```diff
@@ -1,25 +1,35 @@
...
 
 	"github.com/hashicorp/terraform-provider-google/google/acctest"
 	"github.com/hashicorp/terraform-provider-google/google/envvar"
+	"github.com/hashicorp/terraform-provider-google/google/tpgresource"
+	transport_tpg "github.com/hashicorp/terraform-provider-google/google/transport"
 )
 
 func TestAccGKEHub2ScopeRBACRoleBinding_gkehubScopeRbacRoleBindingBasicExample_update(t *testing.T) {
+	// VCR fails to handle batched project services
+	acctest.SkipIfVcr(t)
```
roaks3: I'm confused about a lot of the test changes here, and maybe I confused things by jumping ahead. Were these tests broken? If they were working, I would leave them. For the new tests, do they need the dedicated project, or was that added from copying other tests?

mmv1/products/gkehub2/ScopeRBACRoleBinding.yaml
```diff
@@ -51,6 +51,13 @@ examples:
     primary_resource_name: 'fmt.Sprintf(\"tf-test-scope%s\", context[\"random_suffix\"]), fmt.Sprintf(\"tf-test-rbac-role-binding%s\", context[\"random_suffix\"])'
     test_env_vars:
       project: 'PROJECT_NAME'
+    exclude_test: true
```
roaks3: Hmm, so I guess I'm not clear why we want to also remove this test. Was it failing in some way? I was only suggesting that for the other test, it seems like the test would have failed, because it was not creating a dedicated project.

mmv1/products/gkehub2/ScopeRBACRoleBinding.yaml
```diff
@@ -143,6 +150,16 @@ properties:
           - 'ADMIN'
           - 'EDIT'
           - 'VIEW'
+        exactly_one_of:
```
roaks3, line 152: Ah ok, so this is the source of the breaking change. As written, users were previously able to use a config where `predefined_role` was not specified, and now that same config would not pass validation. Do you know what would have happened previously if it was not specified?

tonybayvas: The role was also a required field before my changes and even if an empty role was passed on the terraform side, the call would still fail on the API side `Error: Error creating ScopeRBACRoleBinding: googleapi: Error 400: MissingFieldError for field role: Missing role`. The exactly_one_of field being added does not change the behavior of this field since it was the only available field before this PR.

Tested not setting a role block:
Error: Insufficient role blocks
│ 
│   on main.tf line 27, in resource "google_gke_hub_scope_rbac_role_binding" "scope_rbac_role_binding":
│   27: resource "google_gke_hub_scope_rbac_role_binding" "scope_rbac_role_binding" {
│ 
│ At least 1 "role" blocks are required.

mmv1/third_party/terraform/services/gkehub2/resource_gke_hub_scope_rbac_role_binding_test.go
tonybayvas: Everything is tied back to making the fleet scopes depend on the APIs. If the scopes depend on the APIs, that means I have to include the setup for the project which enables those APIs and as a result skip the VCR tests. I also don't want to have the scopes depend on the APIs for one test and not the other. The new tests need the dedicated project since they depend on a feature. I'll try adding the generated test back for the predefined role case and avoid setting any depends_on for the scopes

mmv1/products/gkehub2/ScopeRBACRoleBinding.yaml
tonybayvas: I'll remove it, see my other comment on the testing file.

tonybayvas:
Hi @roaks3, I've updated some files, please take a look. I would like to get this submitted today to be part of the rollout, thanks

mmv1/third_party/terraform/services/gkehub2/resource_gke_hub_scope_rbac_role_binding_test.go
```diff
@@ -13,13 +12,16 @@ func TestAccGKEHub2ScopeRBACRoleBinding_gkehubScopeRbacRoleBindingBasicExample_u
 	t.Parallel()
 
 	context := map[string]interface{}{
-		"project":       envvar.GetTestProjectFromEnv(),
-		"random_suffix": acctest.RandString(t, 10),
+		"project":         envvar.GetTestProjectFromEnv(),
+		"random_suffix":   acctest.RandString(t, 10),
+		"org_id":          envvar.GetTestOrgFromEnv(t),
```
roaks3, line 17: It seems like org_id and billing_account are not used for these tests, so I think they should be removed (unless I've missed where they're used)

mmv1/third_party/terraform/services/gkehub2/resource_gke_hub_scope_rbac_role_binding_test.go
```diff
@@ -59,7 +61,6 @@ resource "google_gke_hub_scope_rbac_role_binding" "scoperbacrolebinding" {
   labels = {
       key = "value" 
   }
-  depends_on = [google_gke_hub_scope.scoperbacrolebinding]
```
roaks3, line 62: With this removed, you'll need to establish the implicit dependency by updating the scope_id above (ie. to `google_gke_hub_scope.scoperbacrolebinding.scope_id`)

mmv1/third_party/terraform/services/gkehub2/resource_gke_hub_scope_rbac_role_binding_test.go
```diff
@@ -80,7 +81,133 @@ resource "google_gke_hub_scope_rbac_role_binding" "scoperbacrolebinding" {
   labels = {
       key = "updated_value" 
   }
-  depends_on = [google_gke_hub_scope.scoperbacrolebinding]
```
roaks3, line 83: Same here, you'll need the implicit dependency

roaks3 (commented):
LGTM after a few small changes, and sounds good for the change not being breaking

mmv1/third_party/terraform/services/gkehub2/resource_gke_hub_scope_rbac_role_binding_test.go
tonybayvas: It's used in the gkeHubRRBActuationProjectSetupForGA function

mmv1/third_party/terraform/services/gkehub2/resource_gke_hub_scope_rbac_role_binding_test.go
tonybayvas: Done, thanks

roaks3 (approved):
This is technically a breaking change, but practically `predefined_role` was previously required and the resource could not be used without it.

PR #14169, opened 2025-06-02
Add fine-grained resource GenerativeSettings to DialogflowCX

ArtoriaRen:

```release-note:new-resource
`google_dialogflow_cx_generative_settings`
```

### DIFF
--- a/mmv1/products/dialogflowcx/GenerativeSettings.yaml
+++ b/mmv1/products/dialogflowcx/GenerativeSettings.yaml
@@ -21,7 +21,7 @@ references:
   api: 'https://cloud.google.com/dialogflow/cx/docs/reference/rest/v3/projects.locations.agents/getGenerativeSettings'
 docs:
 base_url: '{{parent}}/generativeSettings'
-self_link: '{{parent}}/generativeSettings?languageCode={{lang_code}}'
+self_link: '{{parent}}/generativeSettings?languageCode={{language_code}}'
 create_url: '{{parent}}/generativeSettings'
 update_url: '{{parent}}/generativeSettings'
 create_verb: 'PATCH'
@@ -52,13 +52,6 @@ parameters:
     description: |
       The agent to create a flow for.
       Format: projects/<Project ID>/locations/<Location ID>/agents/<Agent ID>.
-    url_param_only: true
-    immutable: true
-  - name: 'langCode'
-    type: String
-    default_value: 'en'
-    description: |
-      Language code of the generative settings.
     url_param_only: true
     immutable: true
 properties:
--- a/mmv1/products/dialogflowcx/GenerativeSettings.yaml
+++ b/mmv1/products/dialogflowcx/GenerativeSettings.yaml
@@ -54,7 +54,7 @@ parameters:
       Format: projects/<Project ID>/locations/<Location ID>/agents/<Agent ID>.
     url_param_only: true
     immutable: true
-  - name: 'lang_code'
+  - name: 'langCode'
     type: String
     default_value: 'en'
     description: |
--- a/mmv1/products/dialogflowcx/GenerativeSettings.yaml
+++ b/mmv1/products/dialogflowcx/GenerativeSettings.yaml
@@ -154,6 +154,7 @@ properties:
     type: String
     description: |
       Language for this settings.
+    required: true
   - name: 'llmModelSettings'
     type: NestedObject
     description: |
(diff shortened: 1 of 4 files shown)


mmv1/third_party/terraform/services/dialogflowcx/resource_dialogflowcx_generative_settings_test.go
```diff
@@ -0,0 +1,125 @@
...
+				ImportStateVerifyIgnore: []string{"fallback_settings.0.prompt_templates"},
+			},
+		},
+	})
+}
+
+func testAccDialogflowCXGenerativeSettings_full(context map[string]interface{}) string {
+	return acctest.Nprintf(`
+  resource "google_dialogflow_cx_agent" "agent" {
+      display_name               = "tf-test-%{random_suffix}update"
```
NickElliot:
```suggestion
    display_name               = "tf-test-%{random_suffix}update"
```

mmv1/products/dialogflowcx/GenerativeSettings.yaml
```diff
@@ -0,0 +1,177 @@
...
+      agent_name: 'dialogflowcx-agent'
+parameters:
+  - name: 'parent'
+    type: String
+    description: |
+      The agent to create a flow for.
+      Format: projects/<Project ID>/locations/<Location ID>/agents/<Agent ID>.
+    url_param_only: true
+    immutable: true
+  # not sure if this is the right way to specify a query parameter and it's value.
```
NickElliot:
```suggestion
```

mmv1/products/dialogflowcx/GenerativeSettings.yaml
```diff
@@ -0,0 +1,177 @@
...
+  - name: 'lang_code'
```
NickElliot:
```suggestion
  - name: 'langCode'
```

NickElliot (commented):
what is the difference between `language_code` and `lang_code`

ArtoriaRen:
> what is the difference between `language_code` and `lang_code`

`language_code` is a field of `GenerativeSettings` https://cloud.google.com/dialogflow/cx/docs/reference/rest/v3/GenerativeSettings, whereas`lang_code` is the query parameter of the `getGenerativeSettings` method.

NickElliot (commented):
are langCode and languageCode always the same value? my question is more or less wondering if the "url_param_only" langCode could be replaced by having languageCode as a field be required and also the URL parameter

ArtoriaRen:
> are langCode and languageCode always the same value? my question is more or less wondering if the "url_param_only" 
langCode could be replaced by having languageCode as a field be required and also the URL parameter

The Get request https://cloud.google.com/dialogflow/cx/docs/reference/rest/v3/projects.locations.agents/getGenerativeSettings#http-request doesn't include a `GenerativeSettings`  object, so I'm not sure where we can store `languageCode` in this case.

langCode and languageCode won't be always the same value. Users can specify an invalid langCode in the Get request, in which case they will receive an error response.

ArtoriaRen:
> is a langCode invalid if it doesnt match the languageCode?

Yes. A dialogflow agent can support multiple languages, and  each of them is associated with a `GenerativeSettings`. If langCode isn't one of the supported languages, the response code will be 400. Example https://screenshot.googleplex.com/6Nu5dpDY5rebwqv.

> is languageCode an exported attribute i.e. will the API return a value for the getGenerativeSettings with the field even if a user does not explicitly supply that?

No. Example https://screenshot.googleplex.com/AkF6L5xv2LtdgDT

> Is there an actual scenario where a user will have a different value for both of these fields that doesn't produce an error?

As long as langCode is one of the supported languages of the agent, the get request won't produce an error.  For example,    an agent can support two languages `ja` and `en`, so we have two GenerativeSettings, one with `languageCode=ja` and the other with `languageCode=en`.  A get request on `https://{endpoint}/v3/{name=projects/*/locations/*/agents/*/generativeSettings?languageCode=en` would succeed.  Same for the `?languageCode=ja` case. A Get request on any other languageCode would return 400.

NickElliot (changes requested):
based on what you said and reviewing the documentation, it seems like `langCode` is the same value as `languageCode` within a given instance of the fine grained resource, being functionally the Unique Identifier for the fine grained resource.

Could you remove the `langCode` url parameter and update the `self_link` of the resource to be `?languageCode={{language_code}}', and set `languageCode` as required?

ArtoriaRen:
> based on what you said and reviewing the documentation, it seems like `langCode` is the same value as `languageCode` within a given instance of the fine grained resource, being functionally the Unique Identifier for the fine grained resource.
> 
> Could you remove the `langCode` url parameter and update the `self_link` of the resource to be `?languageCode={{language_code}}', and set `languageCode` as required?

Thanks for the suggestion! I've dropped the the `langCode` url parameter. Could you please take another look? Thank you!

NickElliot (approved):
LGTM!

PR #14172, opened 2025-06-02
Add support for `managed_server_ca` to `google_redis_cluster` resource

liuce-:

```release-note:enhancement
redis: added `managed_server_ca` to `google_redis_cluster` resource

```

Fixes https://github.com/hashicorp/terraform-provider-google/issues/17958

### DIFF
--- a/mmv1/products/redis/Cluster.yaml
+++ b/mmv1/products/redis/Cluster.yaml
@@ -20,6 +20,8 @@ references:
     'Official Documentation': 'https://cloud.google.com/memorystore/docs/cluster/'
   api: 'https://cloud.google.com/memorystore/docs/cluster/reference/rest/v1/projects.locations.clusters'
 docs:
+  attributes: |
+    * `managed_server_ca` - Cluster's Certificate Authority. This field will only be populated if Redis Cluster's transit_encryption_mode is TRANSIT_ENCRYPTION_MODE_SERVER_AUTHENTICATION
   note: |
     For [Multiple VPC Networking](https://cloud.google.com/memorystore/docs/cluster/about-multiple-vpc-networking) if you want to use
     [User-registered PSC Connections](https://cloud.google.com/memorystore/docs/cluster/about-multiple-vpc-networking#psc_connection_types),
@@ -117,6 +119,8 @@ async:
     resource_inside_response: false
 custom_code:
   encoder: 'templates/terraform/encoders/redis_cluster.go.tmpl'
+  decoder: 'templates/terraform/decoders/redis_cluster.go.tmpl'
+  extra_schema_entry: templates/terraform/extra_schema_entry/redis_cluster.tmpl
 sweeper:
   ensure_value:
     field: deletionProtectionEnabled
@@ -208,6 +212,18 @@ examples:
     test_vars_overrides:
       'deletion_protection_enabled': 'false'
       'kms_key_name': 'acctest.BootstrapKMSKeyInLocation(t, "us-central1").CryptoKey.Name'
+    oics_vars_overrides:
+      'deletion_protection_enabled': 'false'
+  - name: "redis_cluster_tls_enabled"
+    primary_resource_id: 'cluster-tls'
+    vars:
+      cluster_name: 'tls-cluster'
+      policy_name: 'my-policy'
+      subnet_name: 'my-subnet'
+      network_name: 'my-network'
+      deletion_protection_enabled: 'true'
+    test_vars_overrides:
+      'deletion_protection_enabled': 'false'
     oics_vars_overrides:
       'deletion_protection_enabled': 'false'
 parameters:
--- a/mmv1/templates/terraform/decoders/redis_cluster.go.tmpl
+++ b/mmv1/templates/terraform/decoders/redis_cluster.go.tmpl
@@ -1,0 +1,91 @@
+config := meta.(*transport_tpg.Config)
+
+userAgent, err :=  tpgresource.GenerateUserAgentString(d, config.UserAgent)
+if err != nil {
+  return nil, err
+}
+
+// Only clusters with TRANSIT_ENCRYPTION_MODE_SERVER_AUTHENTICATION mode have certificate authority set
+if v, ok := res["transitEncryptionMode"].(string); !ok || v!="TRANSIT_ENCRYPTION_MODE_SERVER_AUTHENTICATION" {
+    return res, nil
+}
+
+url, err := tpgresource.ReplaceVars(d, config, "{{"{{"}}RedisBasePath{{"}}"}}projects/{{"{{"}}project{{"}}"}}/locations/{{"{{"}}region{{"}}"}}/clusters/{{"{{"}}name{{"}}"}}/certificateAuthority")
+if err != nil {
+    return nil, err
+}
+
+billingProject := ""
+
+project, err := tpgresource.GetProject(d, config)
+if err != nil {
+    return nil, fmt.Errorf("Error fetching project for Cluster: %s", err)
+}
+
+billingProject = project
+
+// err == nil indicates that the billing_project value was found
+if bp, err := tpgresource.GetBillingProject(d, config); err == nil {
+    billingProject = bp
+}
+
+certificateAuthority, err := transport_tpg.SendRequest(transport_tpg.SendRequestOptions{
+    Config: config,
+    Method: "GET",
+    Project: billingProject,
+    RawURL: url,
+    UserAgent: userAgent,
+})
+if err != nil {
+    return nil, fmt.Errorf("Error reading certificateAuthority: %s", err)
+}
+
+// TODO: DO NOT SUBMIT without removing the comment
+log.Printf("[calvinliu] getting responses from GetCA %+v", certificateAuthority)
+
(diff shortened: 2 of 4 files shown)


liuce-:
Hi @c2thorn , just want to let you know that this PR is ready for review. Feel free to take a look and let me know if you have any questions & concerns, Thank you!

mmv1/templates/terraform/extra_schema_entry/redis_cluster.tmpl
```diff
@@ -0,0 +1,30 @@
+"managed_server_ca": {
+    Type:           schema.TypeList,
+    Computed:       true,
+    Optional:       true,
```
c2thorn: Since its only read in the decoder, these attributes appear to be read-only. The documentation line implies this as well.
If it is read-only we should remove the `Optional: true`. If they are meant to be configurable, I believe it needs to be exposed in the Encoder with expanders, and we need to test configuring it.

mmv1/templates/terraform/examples/redis_cluster_tls_enabled.tf.tmpl
```diff
@@ -0,0 +1,57 @@
+resource "google_redis_cluster" "{{$.PrimaryResourceId}}" {
```
c2thorn, line 1: This example appears to set things up, but doesn't actually use the `managed_server_ca` anywhere. How would a user typically use the output of this field?
If it can't be captured in the Terraform config itself, we should make a handwritten test in https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/third_party/terraform/services/redis/resource_redis_cluster_test.go with a [TestCheckResourceAttr](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-sdk/helper/resource#TestCheckResourceAttr) that verifies the attribute is set correctly

mmv1/products/redis/Cluster.yaml
```diff
@@ -20,6 +20,8 @@ references:
     'Official Documentation': 'https://cloud.google.com/memorystore/docs/cluster/'
   api: 'https://cloud.google.com/memorystore/docs/cluster/reference/rest/v1/projects.locations.clusters'
 docs:
+  attributes: |
+    * `managed_server_ca` - Cluster's Certificate Authority. This field will only be populated if Redis Cluster's transit_encryption_mode is TRANSIT_ENCRYPTION_MODE_SERVER_AUTHENTICATION
```
c2thorn: I think we should at least have bullets for the two nested fields ca_certs and certificates. May be weird formatting with this docs.attributes function, but I'd be fine with best effort for now

mmv1/templates/terraform/decoders/redis_cluster.go.tmpl
```diff
@@ -0,0 +1,84 @@
+config := meta.(*transport_tpg.Config)
```
c2thorn, line 5: A comment explaining that this custom code is necessary because `managed_server_ca` must be retrieved from the distinct https://cloud.google.com/memorystore/docs/cluster/reference/rest/v1/projects.locations.clusters/getCertificateAuthority endpoint would be helpful IMO

mmv1/templates/terraform/examples/redis_cluster_tls_enabled.tf.tmpl
c2thorn: or if we just care that it is set at all, we can use https://pkg.go.dev/github.com/hashicorp/terraform-plugin-sdk/helper/resource#TestCheckResourceAttrSet

mmv1/products/redis/Cluster.yaml
rileykarson: Could we use `ignore_read` on an `output` field or something to add the full field to the schema? There should be _some_ overrides that gives us a target that we will in w/ custom code rather than an extra schema entry.

Or, since we're using a `decoder` anyways we could just add the whole field as if it was returned in Get, and rely on the decoder merging it into the response.

c2thorn: yeah could be worth trying. Nested fields existing inside the extra schema entry is not ideal. Moving it back into the YAML and relying on the decoder would get documentation and we could split the expanders out of the big custom decoder block.

liuce-: I didn't do too much coding in the Terraform provider before so I would like to understand the proposal here: 

Are you suggesting that I should move the nested structure to the `properties` section, so magic module would generate the custom flatten code and the doc section would be easier to read? 

My understanding is my current PR would achieve the exact same thing in the generated Google Provider Code. I am not sure how to do that with the new proposal. The generated code with this proposal would try to access the field in `managed_server_ca` in GetCluster responses, which doesn't exist. We would need a place to inject the code to call the getCertificateAuthority API and flatten the responses.

rileykarson: Specifically I'm talking through this path, which I think I prefer:

> Or, since we're using a decoder anyways we could just add the whole field as if it was returned in Get, and rely on the decoder merging it into the response.

Yes! You'd add the field to `properties` as if it was returned by GetCluster, and then inside the `decoder` you'd call `getCertificateAuthority` and add it to `res`, the original GetCluster response. This is similar to how we retrieve `managementCluster` in https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/templates/terraform/decoders/private_cloud.go.tmpl and attach it to `res`, simulating it getting returned in Get.

mmv1/templates/terraform/decoders/redis_cluster.go.tmpl
liuce-: done. please check.

mmv1/templates/terraform/extra_schema_entry/redis_cluster.tmpl
liuce-: yes you are right. this is a read-only field. Removing the optional field here.

mmv1/products/redis/Cluster.yaml
liuce-: ohhh that's a smart way to do that!!! Let me try it out and see if it would solve the issue.

mmv1/templates/terraform/decoders/redis_cluster.go.tmpl
c2thorn: looks good, thanks

mmv1/templates/terraform/extra_schema_entry/redis_cluster.tmpl
```diff
@@ -0,0 +1,29 @@
...
+    Type:           schema.TypeList,
+    Computed:       true,
+    MaxItems:       1,
+    Description:    "server ca information",
+    Elem: &schema.Resource{
+        Schema: map[string]*schema.Schema{
+            "ca_certs": {
+                Type:           schema.TypeList,
+                Computed:       true,
+                Optional:       true,
```
c2thorn:
```suggestion
```

mmv1/templates/terraform/extra_schema_entry/redis_cluster.tmpl
```diff
@@ -0,0 +1,29 @@
...
+                Description:    "The PEM encoded CA certificate chains for redis managed server authentication",
+                Elem: &schema.Resource{
+                    Schema: map[string]*schema.Schema{
+                        "certificates":{
+                            Type:        schema.TypeList,
+                            Computed:    true,
+                            Optional:    true,
```
c2thorn:
```suggestion
```

liuce-: Done.

mmv1/templates/terraform/extra_schema_entry/redis_cluster.tmpl
liuce-: Done.

mmv1/products/redis/Cluster.yaml
liuce-: Done. please check

mmv1/third_party/terraform/services/redis/resource_redis_cluster_test.go
```diff
@@ -1250,3 +1250,94 @@ func createRedisClusterResourceConfig(params *ClusterParams, isSecondaryCluster
...
+	}
+
+	acctest.VcrTest(t, resource.TestCase{
+		PreCheck:                 func() { acctest.AccTestPreCheck(t) },
+		ProtoV5ProviderFactories: acctest.ProtoV5ProviderFactories(t),
+		// CheckDestroy:             testAccCheckRedisClusterDestroyProducer(t),
+		Steps: []resource.TestStep{
+			{
+				Config: testAccRedisCluster_redisClusterTlsEnabled(context),
+				Check: resource.ComposeTestCheckFunc(
```
c2thorn, line 1269: Looks like this check fails
`google_redis_cluster.cluster-tls: Attribute 'managedServerCa.caCerts.0.certificates.0' expected to be set`

liuce-: Thanks for pointing it out! The test is passing now. please take a look.

PR #14175, opened 2025-06-02
Improve Node Pool Reconciliation time by listing and caching node pools instead of getting them separately

martin-guillen:
Fixes https://github.com/hashicorp/terraform-provider-google/issues/23128

```release-note:none
```

### DIFF
--- a/mmv1/third_party/terraform/services/container/resource_container_node_pool.go.tmpl
+++ b/mmv1/third_party/terraform/services/container/resource_container_node_pool.go.tmpl
@@ -5,7 +5,7 @@ import (
 	"log"
 	"regexp"
 	"strings"
-{{ if ne $.TargetVersionName `ga` }}
+{{- if ne $.TargetVersionName `ga` }}
 	"sync"
 {{- end }}
 	"time"
@@ -951,7 +951,7 @@ func resourceContainerNodePoolExists(d *schema.ResourceData, meta interface{}) (
 	}
 
 	name := getNodePoolName(d.Id())
-{{ if eq $.TargetVersionName `ga` }}
+{{- if eq $.TargetVersionName `ga` }}
 	clusterNodePoolsGetCall := config.NewContainerClient(userAgent).Projects.Locations.Clusters.NodePools.Get(nodePoolInfo.fullyQualifiedName(name))
 	if config.UserProjectOverride {
 		clusterNodePoolsGetCall.Header().Add("X-Goog-User-Project", nodePoolInfo.project)


martin-guillen:
I looked into the failing tests and both seem to be failing due to reasons unrelated to my changes, for the resource manager tags test the subnetwork wasn't ready and for the sole tenant config there was this error:
` Error: Error waiting to create NodeGroup: Error waiting for Creating NodeGroup: The zone 'projects/ci-test-project-188019/zones/us-central1-a' does not have enough resources available to fulfill the request.  Try a different zone, or try again later.`

This is my first time contributing to terraform so I'm not sure if the tests can be reran without changes?

mmv1/third_party/terraform/services/container/resource_container_node_pool.go.tmpl
```diff
@@ -26,6 +27,66 @@ import (
...
+
+func (nodePoolCache *nodePoolCache) invalidate() {
+  nodePoolCache.mutex.Lock()
+  defer nodePoolCache.mutex.Unlock()
+  nodePoolCache.valid = false
+}
+
+var npCache = &nodePoolCache{
+  nodePools: make(map[string]*container.NodePool),
+  ttl:       10 * time.Minute,
```
BBBmau: i'm failing to understand how this would be helpful aside from the case where you are managing a cluster containing a large list of node pools.

 From my understanding this would set it up so that a large list of node pools can be stored and accessed more efficiently, however it is only accessing one after cache. 
 
This seems to make sense when handling a large amount of node pools (in the 100+). However what is the reasoning for the specific 10 minute TTL? I may need assistance on this from the service team.

Edit: Forgot to be mentioned that I've confirmed that both tests are flaky after reviewing the [nightly builds](https://hashicorp.teamcity.com/buildConfiguration/TerraformProviders_GoogleCloud_GOOGLE_BETA_NIGHTLYTESTS_GOOGLEBETA_PACKAGE_CONTAINER/389611?buildTab=tests&name=TestAccContainerNodePool_)

martin-guillen: This is mainly meant to help the cases where there is a large number of node pools, as that is what benefits most from listing the node pools vs getting them separately.

The TTL was just meant to ensure freshness if the node pools have not been listed in a while, however it might not be needed because of the invalidation. It also doesn't have to be 10 minutes if you think there could be a better amount of time

martin-guillen: @BBBmau it's been 2 days since your last response, could you take another look when you have the chance?

martin-guillen:
Had a teammate take a look and made changes for the following suggestions:

- Shorten TTL to 30 seconds
- Store the updated time with the node pool in the cache
- Use the fully qualififed name instead
- Only use the new code in beta
- Fix indenting (looks like I may have to fix it again, IntelliJ is not being very friendly)
- Take a lock when checking if a refresh is needed
- Remove the valid and lastUpdated fields and use the node pool's updated time instead, instead of invalidating the cache remove invalidated node pools, if a node pool is not in the cache then refresh it

mmv1/third_party/terraform/services/container/resource_container_node_pool.go.tmpl
```diff
@@ -26,6 +29,78 @@ import (
...
+func (nodePoolCache *nodePoolCache) refreshIfNeeded(d *schema.ResourceData, config *transport_tpg.Config, userAgent string, nodePoolInfo *NodePoolInformation, name string) error {
+	if !nodePoolCache.needsRefresh(nodePoolInfo.fullyQualifiedName(name)) {
+		return nil
+	}
+
+	nodePoolCache.mutex.Lock()
+	defer nodePoolCache.mutex.Unlock()
+
+	parent := fmt.Sprintf("projects/%s/locations/%s/clusters/%s", nodePoolInfo.project, nodePoolInfo.location, nodePoolInfo.cluster)
+	clusterNodePoolsListCall := config.NewContainerClient(userAgent).Projects.Locations.Clusters.NodePools.List(parent)
```
rileykarson, line 63: @martin-guillen very out of the blue question, but could there be more NPs on a cluster than there are results returned in a single List call? Should we call .Pages (if available, or do pagination manually) here to ensure we get every NP on a cluster? If we fail an NP hit we don't break through, so this would currently return an incorrect error for the customer saying a node pool that does exist does not.

(If I'm missing something already handling pages please flag it, I'm admittedly skimming a bit while looking into something else!)

martin-guillen: Was wondering the same thing when I added caching for the Instance Group Managers, however node pool list calls don't have pagination https://pkg.go.dev/google.golang.org/api/container/v1#ProjectsLocationsClustersNodePoolsListCall
where as instance group manager list calls do
https://pkg.go.dev/google.golang.org/api/compute/v1#InstanceGroupManagersListCall

rileykarson: That's the homework I didn't do. Thank you! And I see you called Pages on MIG, awesome. Sorry for the interrupt!

martin-guillen: No problem!

PR #14179, opened 2025-06-02
colab: fixed perma-diff in `google_colab_runtime_template` caused by the API returning a non-null default value.

bcreddy-gcp:

```release-note:bug
colab: fixed perma-diff in `google_colab_runtime_template` caused by the API returning a non-null default value.
```


bcreddy-gcp:
Adding @c2thorn since they were looking at the issue. 

Is `ignore_changes lifecycle` required?  Recreating a runtime template should be ok since it is not a resource that has any "data", so there shouldn't be any data loss. That said, if the resources depending on template are recreated, that might lead to a data loss. Lmk the suggested route for minimal disruption to the customer.

cc: @haofan

c2thorn:
> Adding @c2thorn since they were looking at the issue.
> 
> Is `ignore_changes lifecycle` required? Recreating a runtime template should be ok since it is not a resource that has any "data", so there shouldn't be any data loss. That said, if the resources depending on template are recreated, that might lead to a data loss. Lmk the suggested route for minimal disruption to the customer.
> 
> cc: @haofan

`ignore_changes` is a Terraform-wide attribute that is not controlled by the provider. Users can use it right now and take it off after they upgrade or simply leave it on.

c2thorn:
I'll run colab tests in our CI just to double check things

c2thorn (approved):

PR #14180, opened 2025-06-03
Add lustre instance performance tiers

stephanecharite:
```release-note:breaking-change
lustre: added `per_unit_storage_throughput` as a required field to `google_lustre_instance` resource in response to a change in the API surface
```


stephanecharite:
Test failures are due to quota issues. The project that runs this presubmit only has enough capacity for 2 Lustre instances and not 3. I suspect when https://github.com/GoogleCloudPlatform/magic-modules/pull/13810 was made this wasn't accounted for. I am addressing on the Google side but for now I suspect I can just rerun presubmits for this PR and VPC should pass since previous successful runs are cached.

Will also send quota increase request for nightly ga and beta projects as well as I suspect they will have the same issue post submission

zli82016:
@stephanecharite, besides adding the required field `per_unit_storage_throughput`, this PR is adding some other required fields, such as `capacity_gib`, `network`, `filesystem `? Is it expected? In b/421976343, only `per_unit_storage_throughput` is mentioned.

stephanecharite:
@zli82016 Looking at the diff generated by the PR the fields `capacity_gib`, `network`, and `filesystem` have always been required. It seemed the order got rearranged when I used autogen to generate this PR. If need be I can reorganize the output if it helps.

mmv1/third_party/terraform/services/lustre/resource_lustre_instance_test.go
```diff
@@ -45,7 +45,7 @@ func TestAccLustreInstance_update(t *testing.T) {
 				ResourceName:            "google_lustre_instance.instance",
 				ImportState:             true,
 				ImportStateVerify:       true,
-				ImportStateVerifyIgnore: []string{"instance_id", "labels", "gke_support_enabled", "location", "terraform_labels"},
+				ImportStateVerifyIgnore: []string{"instance_id", "labels", "gke_support_enabled", "per_unit_storage_throughput", "location", "terraform_labels"},
```
zli82016: What is the reason to ignore `per_unit_storage_throughput` here?

stephanecharite: No reason at all, this was an accident. I removed this change

mmv1/third_party/terraform/services/lustre/resource_lustre_instance_test.go
```diff
@@ -28,7 +28,7 @@ func TestAccLustreInstance_update(t *testing.T) {
...
```
zli82016: Should `per_unit_storage_throughput` be removed as well?

stephanecharite: My apologies, I didn't realized I didn't remove it here as well. I updated the code

zli82016 (dismissed):

mmv1/third_party/terraform/services/lustre/resource_lustre_instance_test.go
```diff
@@ -82,17 +83,18 @@ data "google_compute_network" "lustre-network" {
...
-  network             = data.google_compute_network.lustre-network.id
-	description         = "test-description"
-	labels              = {
+  instance_id                 = "tf-test-my-instance%{random_suffix}"
+  location                    = "us-central1-a"
+  filesystem                  = "testfs"
+  capacity_gib                = 18000
+  network                     = data.google_compute_network.lustre-network.id
+  description                 = "test-description"
+  per_unit_storage_throughput = 1000
```
zli82016, line 92: The field can be updated to a different value to test the update.

zli82016: If the update feature is not ready, you can add `immutable: true` to this new field.

stephanecharite: done

mmv1/products/lustre/Instance.yaml
```diff
@@ -73,10 +74,42 @@ parameters:
     url_param_only: true
     required: true
 properties:
+  - name: capacityGib
+    type: String
+    description: |-
+      The storage capacity of the instance in gibibytes (GiB). Allowed values
+      are from `18000` to `954000`, in increments of 9000.
+    required: true
+    immutable: true
```
zli82016: Is the field `capacityGib` updatable? `immutable: true` is just added to it.

zli82016: What happens to update the field `capacityGib` now without `immutable: true`? Does it always return an error?

mmv1/third_party/terraform/services/lustre/resource_lustre_instance_test.go
stephanecharite: Not supported at the moment

mmv1/products/lustre/Instance.yaml
stephanecharite: It will be updatable in 2 weeks, personally I think we shouldn't have to include immutable here. The user will get an error if they try to udpate it prior to GA.

zli82016: Can you remove `immutable: true` for `capacityGib`, as `capacityGib` is not a new field? Thanks.

stephanecharite: done

zli82016:
cc @SirGitsalot

SirGitsalot (approved):
Adding some context here for posterity as a lot of the discussion of this breaking change took place in other venues:

- `per_unit_storage_throughput` has been supported in the API for a while as an optional field
- Until this PR the field has not been supported in Terraform Provider Google
- An API change that makes the field required is in the process of being rolled out

The combination of those three facts means that without the change the resource will become unusable, so this is one of the rare instances where we are allowing a breaking change in a minor release. The team has indicated that they have notified affected customers.

PR #14185, opened 2025-06-03
tgc-revival: cai2hcl  resource converter template

zli82016:
The generated compute address cai2hcl converter is  in the [diff](pkg/cai2hcl/converters/services/compute/compute_address.go).

```release-note:none

```

### DIFF
--- a/mmv1/api/resource.go
+++ b/mmv1/api/resource.go
@@ -1980,7 +1980,7 @@ func (r Resource) MarkdownHeader(templatePath string) string {
 // ====================
 // TGC
 // ====================
-// Converts most properties during cai2hcl, exept for computed properties
+// Filters out computed properties during cai2hcl
 func (r Resource) ReadPropertiesForTgc() []*Type {
 	return google.Reject(r.AllUserProperties(), func(v *Type) bool {
 		return v.Output
--- a/mmv1/api/resource.go
+++ b/mmv1/api/resource.go
@@ -1976,3 +1976,13 @@ func (r Resource) CodeHeader(templatePath string) string {
 func (r Resource) MarkdownHeader(templatePath string) string {
 	return strings.Replace(r.CodeHeader(templatePath), "//", "#", -1)
 }
+
+// ====================
+// TGC
+// ====================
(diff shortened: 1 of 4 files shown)


mmv1/api/resource.go
```diff
@@ -1976,3 +1976,13 @@ func (r Resource) CodeHeader(templatePath string) string {
 func (r Resource) MarkdownHeader(templatePath string) string {
 	return strings.Replace(r.CodeHeader(templatePath), "//", "#", -1)
 }
+
+// ====================
+// TGC
+// ====================
+// Converts most properties during cai2hcl, exept for computed properties
```
c2thorn:
```suggestion
// Filters out computed properties during cai2hcl
```
Doesn't seem to do any converting on its own, just filtering?

mmv1/templates/tgc_next/cai2hcl/resource_converter.go.tmpl
```diff
@@ -0,0 +1,130 @@
...
+
+	res := asset.Resource.Data
+	config := utils.NewConfig()
+	d := &schema.ResourceData{}
+
+	hclData := make(map[string]interface{})
+
+{{ range $prop := $.ReadPropertiesForTgc }}
+{{   if $prop.FlattenObject -}}
+// Terraform must set the top level schema field, but since this object contains collapsed properties
```
c2thorn: This code is shared with the resource.go template code right? If they are the same, and need to be kept identical, should we move the shared parts to a subtemplate?

c2thorn (commented):
generated output looks good!

mmv1/api/resource.go
zli82016: That is right.

mmv1/templates/tgc_next/cai2hcl/resource_converter.go.tmpl
zli82016: It shouldn't use `d.Set()` to set values in TGC. Instead, it should set values to `hclData`

I am going to add a TODO comment how to convert FlattenObject

c2thorn (approved):

PR #14187, opened 2025-06-03
Introduce confidential_instance_type into confidential_nodes config

G-jackyli:
This adds support for the "confidential_instance_type" enum into "confidential_nodes" at both cluster and node pool level.

```release-note:enhancement
container: added `confidential_nodes.confidential_instance_type` field to `google_container_cluster` resource
```

```release-note:enhancement
container: added `node_config.confidential_nodes.confidential_instance_type` field to `google_container_node_pool` resource
```


G-jackyli:
This is my first PR in Terraform, please let me know if I have to do anything different. 

With regard to the `Missing test report`, I do have the test including those fields under `resource "google_container_cluster" "confidential_nodes"` instead of `resource "google_container_cluster" "primary"`. This is consistent with the rest of file and existing code. It's a false alarm but please let me know if I had to update it as well. 

For `Missing doc report (experimental)`, I do have doc updated but the checked seems to only checking the full path so it's a false alarm as well.

PR #14190, opened 2025-06-04
Introduce confidential_instance_type into confidential_nodes config

G-jackyli:
This change adds support for the "confidential_instance_type" enum into "confidential_nodes" at both cluster and node pool level.

```release-note:enhancement
container: added `confidential_nodes.confidential_instance_type` field to `google_container_cluster` resource
```

```release-note:enhancement
container: added `node_config.confidential_nodes.confidential_instance_type` field to `google_container_node_pool` resource
```

### DIFF
--- a/mmv1/third_party/terraform/services/container/resource_container_cluster_migratev1.go.tmpl
+++ b/mmv1/third_party/terraform/services/container/resource_container_cluster_migratev1.go.tmpl
@@ -1092,7 +1092,7 @@ func resourceContainerClusterResourceV1() *schema.Resource {
                                                         ForceNew:    true,
                                                         DiffSuppressFunc: suppressDiffForConfidentialNodes,
 							Description: `Defines the type of technology used by the confidential node.`,
-                                                        ValidateFunc: validation.StringInSlice([]string{"CONFIDENTIAL_INSTANCE_TYPE_UNSPECIFIED", "SEV", "SEV_SNP", "TDX"}, false),
+                                                        ValidateFunc: validation.StringInSlice([]string{"SEV", "SEV_SNP", "TDX"}, false),
                                                 },
 					},
 				},
--- a/mmv1/third_party/terraform/services/container/node_config.go.tmpl
+++ b/mmv1/third_party/terraform/services/container/node_config.go.tmpl
@@ -844,7 +844,7 @@ func schemaNodeConfig() *schema.Schema {
 								ForceNew:    true,
 								DiffSuppressFunc: suppressDiffForConfidentialNodes,
 								Description: `Defines the type of technology used by the confidential node.`,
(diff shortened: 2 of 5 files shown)


G-jackyli:
This is my first PR in Terraform, please let me know if I have to do anything different. 

With regard to the `Missing test report`, I do have the test including those fields under `resource "google_container_cluster" "confidential_nodes"` instead of `resource "google_container_cluster" "primary"`. This is consistent with the rest of file and existing code. It's a false alarm but please let me know if I had to update it as well. 

For `Missing doc report (experimental)`, I do have doc updated but the checked seems to only checking the full path so it's a false alarm as well.

mmv1/third_party/terraform/services/container/resource_container_cluster_migratev1.go.tmpl
```diff
@@ -1086,6 +1086,14 @@ func resourceContainerClusterResourceV1() *schema.Resource {
 							ForceNew:    true,
 							Description: `Whether Confidential Nodes feature is enabled for all nodes in this cluster.`,
 						},
+						"confidential_instance_type": {
+                                                        Type:        schema.TypeString,
+                                                        Optional:    true,
+                                                        ForceNew:    true,
+                                                        DiffSuppressFunc: suppressDiffForConfidentialNodes,
+							Description: `Defines the type of technology used by the confidential node.`,
+                                                        ValidateFunc: validation.StringInSlice([]string{"CONFIDENTIAL_INSTANCE_TYPE_UNSPECIFIED", "SEV", "SEV_SNP", "TDX"}, false),
```
SirGitsalot: Don't include `CONFIDENTIAL_INSTANCE_TYPE_UNSPECIFIED` - [enum values shouldn't include the zero value](https://googlecloudplatform.github.io/magic-modules/reference/field/#enum_values) To unset the field (or set it to the zero value, depending on your perspective) users will omit the field from their configs entirely.

Not sure what's going on with the test and doc reports as the docs are obviously there and I checked the debug log and the tests are indeed running and changing the field's value, so don't worry about it.

SirGitsalot (changes requested):

mmv1/third_party/terraform/services/container/resource_container_cluster_migratev1.go.tmpl
G-jackyli: Got it.  I was following the code samples in the same file and it seems the "unspecified" value is used in many places (e.g. [1]). I think you are right though, user should just omit the field if they want to unset the field. 

Question: should I remove the `CONFIDENTIAL_INSTANCE_TYPE_UNSPECIFIED` in the `node_config.go.tmpl` file and `resource_container_cluster.go.tmpl` file as well? I assume yes but want to double confirm before I go ahead and update the tests.

[1] https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/third_party/terraform/services/container/resource_container_cluster_migratev1.go.tmpl#L1649

SirGitsalot (approved):

PR #12571, opened 2024-12-13
Fix service_attachment.target_service validation

Samir-Cit:
Hello folks.
This PR is to add back the validation of the resource that was removed on this other PR: https://github.com/GoogleCloudPlatform/magic-modules/pull/11107

[Buganizer](https://buganizer.corp.google.com/issues/352822740)

The resource now accepts ForwardingRules and Gateways so this validation, that used to have, needed to change to accept both kind of resources.

Note: This PR includes a small refactoring on the test

**Release Note Template for Downstream PRs (will be copied)**

See [Write release notes](https://googlecloudplatform.github.io/magic-modules/contribute/release-notes/) for guidance.

```release-note:enhancement
compute: add back the validation for `target_service` field on the `google_compute_service_attachment` resource to validade a `ForwardingRule` or `Gateway` URL.
```

PR #14195, opened 2025-06-04
Fix `target_server` validation for service attachment

DanielRieske:
Originally addressed in https://github.com/GoogleCloudPlatform/magic-modules/pull/12571 where validation was added back for the `target_service` argument.

Unfortunately if an invalid format of a target service is provided, it will results in an out of bound exception due to the order of error handling/validating.

```
Stack trace from the terraform-provider-google_v6.35.0_x5 plugin:

panic: runtime error: index out of range [-1]

goroutine 186 [running]:
github.com/hashicorp/terraform-provider-google/google/services/compute.expandComputeServiceAttachmentTargetService({0x4090440, 0xc001677eb0}, {0x52fbf00, 0xc001f46e80}, 0xc00060b408)
	github.com/hashicorp/terraform-provider-google/google/services/compute/resource_compute_service_attachment.go:894 +0x1b3
```

```release-note:bug
compute: fixed validation for `target_service`  in `google_compute_service_attachment`
```


melinath (approved):
Seems straightforward. Thanks for the fix!

PR #14204, opened 2025-06-05
Add new fields in QAS resource

vbhadoriaB:
```release-note:enhancement
cloudquotas: added `inherited`  and `inherited_from` fields to `google_cloud_quotas_quota_adjuster_settings` resource
```

### DIFF
--- a/mmv1/products/cloudquotas/QuotaAdjusterSettings.yaml
+++ b/mmv1/products/cloudquotas/QuotaAdjusterSettings.yaml
@@ -66,14 +66,12 @@ properties:
   - name: 'effectiveContainer'
     type: String
     description: |
-      Deprecated: The resource container that determines if the quota adjuster is set for this project.
-      Expect this field to be empty currently.
+      The resource container that determines if the quota adjuster is set for this project.
     output: true
   - name: 'effectiveEnablement'
     type: Enum
     description: |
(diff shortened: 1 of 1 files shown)


melinath (changes requested):
It looks like there are two separate things going on in this PR. `effectiveEnablement` and `effectiveContainer` are being marked as deprecated, and two new fields are being added. Please only do one of those. Additionally, see https://googlecloudplatform.github.io/magic-modules/breaking-changes/make-a-breaking-change/ for information on how to do deprecations and removals.

mmv1/products/cloudquotas/QuotaAdjusterSettings.yaml
```diff
@@ -67,15 +67,25 @@ properties:
...
     enum_values:
       - 'DEFAULT'
       - 'ENABLED'
       - 'DISABLED'
+  - name: 'inherited'
+    type: Boolean
+    description: |
+      Indicates whether the setting is inherited or explicitly specified.
+    output: true
+  - name: 'inherited_from'
```
melinath: This needs to match the field's JSON name (rather than proto name) - otherwise it won't be set properly on the resource.

```suggestion
  - name: 'inheritedFrom'
```

melinath (approved):

PR #14206, opened 2025-06-05
Adding KMS CMEK support for Dialogflow

zhihaos:
fulfill request https://github.com/hashicorp/terraform-provider-google/issues/22914

```release-note:new-resource
`google_dialogflow_encryption_spec`
```


zhihaos:
@SirGitsalot , could you please take a look? Is there anything I missed?

zhihaos:
The error code is autogenerated and I think it is from `autogen_async: true`, which is used for returning LRO. I am unable to find why there is one extra parameters in the call. @SirGitsalot could you please take a look? Thanks!

mmv1/products/dialogflow/EncryptionSpec.yaml
```diff
@@ -0,0 +1,64 @@
...
+          Only the key in the same location as this resource is allowed to be used for encryption.
+          Format: projects/{project}/locations/{location}/keyRings/{keyRing}/cryptoKeys/{key}
+
+examples:
+  - name: "dialogflow_encryption_spec_basic"
+    primary_resource_id: "example-encryption-spec"
+    vars:
+      kms_key_name: 'kms-key'
+    test_vars_overrides:
+      kms_key_name: 'acctest.BootstrapKMSKey(t).CryptoKey.Name'
```
SirGitsalot: Try:

```
kms_key_name: 'acctest.BootstrapKMSKeyInLocation(t, "us-central1").CryptoKey.Name'
```

The test is failing with `Please refer to https://cloud.google.com/dialogflow/cx/docs/concept/region#avail and https://cloud.google.com/dialogflow/es/docs/how/region#regions to find the correct endpoint to access resources located in 'us-central1'."` The request for the encryption spec is using `us-central`, but the KMS key's location is `global` which makes me think it's running afoul of [this limitation](https://cloud.google.com/dialogflow/cx/docs/concept/cmek#limits)

SirGitsalot (changes requested):

mmv1/products/dialogflow/EncryptionSpec.yaml
zhihaos: Thanks for the review. I updated it and running the tests now.

zhihaos:
> Is `encryptionSpec` available in the v3 API? If it is, moving these files to the dialogflowcx directory might get them working.
> 
> As you saw, changing the KMS key to `us-central1` didn't fix the problem (I think it's still necessary, but the tests don't seem to be getting that far). It looks to me like the immediate problem is that the test is using `https://dialogflow.googleapis.com` when it needs to be using `https://us-central1-dialogflow.googleapis.com`.
> 
> Magic Modules defines the API endpoint in `product.yaml`, and for [dialogflow](https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/products/dialogflow/product.yaml) that's `https://dialogflow.googleapis.com/v2/`, but for [dialogflowcx](https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/products/dialogflowcx/product.yaml) it's `https://{{location}}-dialogflow.googleapis.com/v3/`

Thanks for the finding. Yes, you are right, from my test, the endpoint pointed to the `{{location}}-dialogflow`. I will move them to the cx.

zhihaos:
@SirGitsalot Sorry for the bothering. I moved the resource to dialogflow cx now. But the checks failed with an error that I don't understand. Could you please take a look? 🙏

zhihaos:
> @SirGitsalot Sorry for the bothering. I moved the resource to dialogflow cx now. But the checks failed with an error that I don't understand. Could you please take a look? 🙏

The error is gone after I commented out following:
```
# autogen_async: true
# async:
#   actions: ['create']
#   operation:
#     base_url: '{{op_id}}'
```

zhihaos:
> > @SirGitsalot Sorry for the bothering. I moved the resource to dialogflow cx now. But the checks failed with an error that I don't understand. Could you please take a look? 🙏
> 
> The error is gone after I commented out following:
> 
> ```
> # autogen_async: true
> # async:
> #   actions: ['create']
> #   operation:
> #     base_url: '{{op_id}}'
> ```

after inspecting the file `go/src/github.com/hashicorp/terraform-provider-google/google/services/dialogflowcx/dialogflow_cx_operation.go`

It is marked as:
```
//	***     AUTO GENERATED CODE    ***    Type: Handwritten     ***
```

and the original file is found at https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/third_party/terraform/services/dialogflowcx/dialogflow_cx_operation.go

zhihaos:
> > > @SirGitsalot Sorry for the bothering. I moved the resource to dialogflow cx now. But the checks failed with an error that I don't understand. Could you please take a look? 🙏
> > 
> > 
> > The error is gone after I commented out following:
> > ```
> > # autogen_async: true
> > # async:
> > #   actions: ['create']
> > #   operation:
> > #     base_url: '{{op_id}}'
> > ```
> 
> after inspecting the file `go/src/github.com/hashicorp/terraform-provider-google/google/services/dialogflowcx/dialogflow_cx_operation.go`
> 
> It is marked as:
> 
> ```
> //	***     AUTO GENERATED CODE    ***    Type: Handwritten     ***
> ```
> 
> and the original file is found at https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/third_party/terraform/services/dialogflowcx/dialogflow_cx_operation.go

<del>@SirGitsalot Could you please take a look? What should I do in this case? 🙏 </del>

I removed autogen_async, and it passed the test locally.

zhihaos:
The current issues is that the `project` is in the `base_url`, when calling `OperationWaitTime`, the argument will include the `project`, however the handwritten operation function does not include the `project` as the parameter of `OperationWaitTime` function. Need to figure out any workaround. how to exclude the `project` as the argument?

zhihaos:
> The current issues is that the `project` is in the `base_url`, when calling `OperationWaitTime`, the argument will include the `project`, however the handwritten operation function does not include the `project` as the parameter of `OperationWaitTime` function. Need to figure out any workaround. how to exclude the `project` as the argument?

I am trying to overload the `OperationWaitTime` function, or adding default parameter value. However, both solutions are not supported in Go. Continue investigating.

zhihaos:
> > The current issues is that the `project` is in the `base_url`, when calling `OperationWaitTime`, the argument will include the `project`, however the [handwritten operation function](https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/third_party/terraform/services/dialogflowcx/dialogflow_cx_operation.go) does not include the `project` as the parameter of `OperationWaitTime` function. Need to figure out any workaround. how to exclude the `project` as the argument?
> 
> I am trying to overload the `OperationWaitTime` function, or adding default parameter value. However, both solutions are not supported in Go. Continue investigating.

tried to add `include_project: false`, the project argument is still added because it uses `if hasProject() or hasIncludeProject` [code](https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/templates/terraform/resource.go.tmpl#L392). As long as there is `{{project}}` in the yaml file, the project argument will be added.

zhihaos:
The failing test may be due to the gcp outage, will re trigger the test once it is recovered.

zhihaos:
It seems that `encryptionSpec` was not publicly accessible, will look into how to modify the endpoint of the v2 dialogflow.

zhihaos:
@SirGitsalot , I am a little confused about the following error:
```
Error: Inconsistent dependency lock file
        
The following dependency selections recorded in the lock file are
inconsistent with the current configuration:
  - provider registry.terraform.io/hashicorp/google-beta: required by this configuration but no version is selected

To make the initial dependency selections that will initialize the dependency
lock file, run:
  terraform init
```
Could you please have a look?

mmv1/products/dialogflow/EncryptionSpec.yaml
```diff
@@ -0,0 +1,69 @@
...
+  guides:
+    'Official ES Documentation': 'https://cloud.google.com/dialogflow/es/docs'
+    'Official CX Documentation': 'https://cloud.google.com/dialogflow/cx/docs'
+  api: 'https://cloud.google.com/dialogflow/es/docs/reference/rest/v2/projects.locations.encryptionSpec'
+
+base_url: 'projects/{{project}}/locations/{{location}}/encryptionSpec'
+
+create_url: 'projects/{{project}}/locations/{{location}}/encryptionSpec:initialize'
+
+immutable: true
```
SirGitsalot, line 33: If the resource is only in beta you'll need `min_version: "beta"` [at the root level too](https://googlecloudplatform.github.io/magic-modules/reference/resource/#min_version-beta) (I don't know if that will fix the current error message, but it's needed)

zhihaos: Thank you for the review. The new resource is not only in beta. One of the resources in the example is only in beta. I changed all of them to beta in the example. It is working now.

zhihaos:
The error is that the resource does not support import which is correct and I have `exclude_import: true`. Need to figure out where the error comes from.

zhihaos:
> The error is that the resource does not support import which is correct and I have `exclude_import: true`. Need to figure out where the error comes from.

I set `exclude_import_test: true`, all tests passed. @SirGitsalot Could you please review? Thanks. 🙏

mmv1/products/dialogflow/EncryptionSpec.yaml
```diff
@@ -0,0 +1,78 @@
...
+        description: |
+          The name of customer-managed encryption key that is used to secure a resource and its sub-resources.
+          If empty, the resource is secured by the default Google encryption key.
+          Only the key in the same location as this resource is allowed to be used for encryption.
+          Format: projects/{project}/locations/{location}/keyRings/{keyRing}/cryptoKeys/{key}
+
+examples:
+  - name: "dialogflow_encryption_spec_basic"
+    primary_resource_id: "my-encryption-spec"
+    min_version: "beta"
```
SirGitsalot, line 68: Are the `min_version: "beta"` declarations still needed?

SirGitsalot (commented):
I'm running the dialogflow tests on our upstream CI test system right now since there's specific IAM requirement which can be touchy (as you've seen!). It should pass since it's passing on Cloud Build though. Once that goes through LGTM.

mmv1/products/dialogflow/EncryptionSpec.yaml
zhihaos: Done. Thanks for the catch.

zhihaos: I have to add this line or the test will fail with
```
Error: Inconsistent dependency lock file
        
The following dependency selections recorded in the lock file are
inconsistent with the current configuration:
  - provider registry.terraform.io/hashicorp/google-beta: required by this configuration but no version is selected
        
To update the locked dependency selections to match a changed configuration,
run:
terraform init -upgrade
```

SirGitsalot: Ah I see - `google_project_service_identity` is beta only. LGTM!

SirGitsalot (approved):

PR #14207, opened 2025-06-05
Update TestAccGeminiGeminiGcpEnablementSettingBinding_update value

rileykarson:
Error:

```
Error code 3, message: product "GEMINI_IN_BIGQUERY" is not supported by the setting type "geminiGcpEnablementSettings", allowed product is "GEMINI_CLOUD_ASSIST": invalid argument
```

```release-note:bug
gemini: removed overly restrictive `product` validation on `google_gemini_gemini_gcp_enablement_setting_binding`, `google_gemini_data_sharing_with_google_setting_binding`. New values like `GOOGLE_CLOUD_ASSIST` will now be accepted.
```

### DIFF
--- a/mmv1/products/gemini/DataSharingWithGoogleSettingBinding.yaml
+++ b/mmv1/products/gemini/DataSharingWithGoogleSettingBinding.yaml
@@ -77,7 +77,7 @@ properties:
   - name: product
     type: String
     description: |-
-      Product type of the setting binding. Values include GEMINI_IN_BIGQUERY, GEMINI_CLOUD_ASSIST, etc. See [product reference](https://cloud.google.com/gemini/docs/api/reference/rest/v1/projects.locations.dataSharingWithGoogleSettings.settingBindings) fpr a complete list.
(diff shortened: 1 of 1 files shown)


mmv1/products/gemini/GeminiGcpEnablementSettingBinding.yaml
```diff
@@ -82,11 +82,9 @@ properties:
     description: Target of the binding.
     required: true
   - name: product
-    type: Enum
+    type: String
     description: |-
-      Product type of the setting binding.
-    enum_values:
-      - 'GEMINI_IN_BIGQUERY'
+      Product type of the setting binding. Values include GEMINI_IN_BIGQUERY, GEMINI_CLOUD_ASSIST, etc. See [product reference](https://cloud.google.com/gemini/docs/api/reference/rest/v1/projects.locations.dataSharingWithGoogleSettings.settingBindings) for a complete list.
```
rileykarson, line 87: The reference I'm linking here is technically for the wrong resource type- and I think the actual enum type link on this page doesn't work- but this resource kind is missing docs entirely. It appears to be a shared type given the values in range, so it seems better than nothing.

roaks3 (approved):

rileykarson:
(Not sure if that will entirely fix the GHAs, I maybe made a few too many commits at once lol)

PR #14211, opened 2025-06-05
add pagination function for plural data-sources

BBBmau:
Fixes https://github.com/hashicorp/terraform-provider-google/issues/19157

This simplifies the implementation of plural data-sources by creating a generic function that can be used regardless of the data-source.

The PR represents two uses of the function, with one simplifying implementation within an already existing List function and the second showcasing the function implementation within a Read function 

```release-note:none

```

### DIFF
--- a/mmv1/third_party/terraform/services/alloydb/data_source_alloydb_locations.go
+++ b/mmv1/third_party/terraform/services/alloydb/data_source_alloydb_locations.go
@@ -86,64 +86,39 @@ func dataSourceAlloydbLocationsRead(d *schema.ResourceData, meta interface{}) er
 	if err != nil {
 		return fmt.Errorf("Error setting api endpoint")
 	}
-	res, err := transport_tpg.SendRequest(transport_tpg.SendRequestOptions{
-		Config:    config,
-		Method:    "GET",
-		Project:   billingProject,
-		RawURL:    url,
-		UserAgent: userAgent,
-	})
+
+	locations, err := transport_tpg.PluralDataSourceGet(d, config, billingProject, userAgent, url, nil, "locations")
 	if err != nil {
-		return transport_tpg.HandleDataSourceNotFoundError(err, d, fmt.Sprintf("Locations %q", d.Id()), url)
+		return err
 	}
-	var locations []map[string]interface{}
-	for {
-		fetchedLocations := res["locations"].([]interface{})
-		for _, loc := range fetchedLocations {
-			locationDetails := make(map[string]interface{})
-			l := loc.(map[string]interface{})
-			if l["name"] != nil {
-				locationDetails["name"] = l["name"].(string)
-			}
-			if l["locationId"] != nil {
-				locationDetails["location_id"] = l["locationId"].(string)
-			}
-			if l["displayName"] != nil {
-				locationDetails["display_id"] = l["displayName"].(string)
-			}
-			if l["labels"] != nil {
-				labels := make(map[string]string)
-				for k, v := range l["labels"].(map[string]interface{}) {
-					labels[k] = v.(string)
-				}
-				locationDetails["labels"] = labels
-			}
-			if l["metadata"] != nil {
-				metadata := make(map[string]string)
-				for k, v := range l["metadata"].(map[interface{}]interface{}) {
-					metadata[k.(string)] = v.(string)
-				}
-				locationDetails["metadata"] = metadata
-			}
-			locations = append(locations, locationDetails)
+
+	for _, loc := range locations {
+		locationDetails := make(map[string]interface{})
+		l := loc.(map[string]interface{})
+		if l["name"] != nil {
+			locationDetails["name"] = l["name"].(string)
 		}
-		if res["nextPageToken"] == nil || res["nextPageToken"].(string) == "" {
-			break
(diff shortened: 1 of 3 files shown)


mmv1/third_party/terraform/transport/transport.go
```diff
@@ -189,3 +189,130 @@ func IsApiNotEnabledError(err error) bool {
 	}
 	return false
 }
+
+func PluralDataSourceGetList(d *schema.ResourceData, config *Config, billingProject *string, userAgent string, u string, listFlattener func(config *Config, res interface{}) ([]interface{}, error), params map[string]string, resourecToList string) ([]interface{}, error) {
```
rileykarson: Would you mind structuring this as an options type similar to SendRequest? That gives a good opportunity to add documentation for the parameters, and makes understanding the function from the callsite easier.

rileykarson: In particular it's not obvious when you would/wouldn't include a listFlattener IMO (I _think_ it's for brownfield reasons?)

mmv1/third_party/terraform/services/alloydb/data_source_alloydb_locations.go
```diff
@@ -86,64 +86,39 @@ func dataSourceAlloydbLocationsRead(d *schema.ResourceData, meta interface{}) er
...
 	}
-	res, err := transport_tpg.SendRequest(transport_tpg.SendRequestOptions{
-		Config:    config,
-		Method:    "GET",
-		Project:   billingProject,
-		RawURL:    url,
-		UserAgent: userAgent,
-	})
+
+	listedLocations, err := transport_tpg.PluralDataSourceGetList(d, config, &billingProject, userAgent, url, nil, nil, "locations")
```
rileykarson: Could we use the function outside of plural datasources? A name like `GetPaginatedList`/`GetListPages` might be more clear.

(Or, I guess `GetPaginatedListSlice` or `GetPaginatedListMap` would be the actual names)

mmv1/third_party/terraform/transport/transport.go
```diff
@@ -189,3 +189,130 @@ func IsApiNotEnabledError(err error) bool {
...
+		}
+		if billingProject != nil {
+			opts.Project = *billingProject
+		}
+		res, err := SendRequest(opts)
+		if err != nil {
+			return nil, HandleNotFoundError(err, d, fmt.Sprintf("%s %q", resourecToList, d.Id()))
+		}
+
+		if res == nil {
```
rileykarson: A plural datasource can return zero items and still be valid, I think? I'd expect us to drop the id if the _parent_ is gone i.e. a crypto key with no keys in it still exists, it just has no keys. (for most resources that will be the project or the project+location pair)

rileykarson: I would further consider making this the responsibility of the caller, where this function's responsibility is just to walk the paginated list, leaning on my comment about `GetPaginatedList` a bit more!

mmv1/third_party/terraform/transport/transport.go
```diff
@@ -189,3 +189,130 @@ func IsApiNotEnabledError(err error) bool {
...
+			log.Printf("[DEBUG] Removing %s because it no longer exists.", resourecToList)
+			d.SetId("")
+			return nil, nil
+		}
+
+		var newItems []interface{}
+		if listFlattener != nil {
```
rileykarson: I think `flattenKMSKeyRingsList` is actually a _decoder_, similar to how NestedQuery can apply encoders/decoders to list items during the nested query. Not a new issue to your change, but may be worth correcting!

The distinguishing factor is the level of granularity. Encoders/decoders are resource level, and expanders/flatteners are field level. This abstraction get a bit messy with big nested objects, where multiple fields are within the scope of the parent field.

rileykarson: (we could also consider making this the responsibility of the caller, potentially)

mmv1/third_party/terraform/services/alloydb/data_source_alloydb_locations.go
BBBmau: the naming can be more clear i agree and yeah technically if you need to paginate through a list for a specific case you can do so as long as you provide the appropriate options

mmv1/third_party/terraform/transport/transport.go
BBBmau: yeah having it be an options type will be better.

In regards to the `listFlattener` you're correct, it's due to brown field reasons as in if theres an existing flattener being used in the exist plural data source we can continue to use it.

mmv1/third_party/terraform/transport/transport.go
BBBmau: yeah we can change this to where it traverses the list, regardless of whether it contains anything or not.

> I'd expect us to drop the id if the parent is gone i.e. a crypto key with no keys in it still exists, it just has no keys. (for most resources that will be the project or the project+location pair)

this makes sense to check whether the actual parent (resource we need that contains the list) is present.

we completely remove the id of the plural data sources of crypto keys if the `crypto key` itself that contains the list is nonexistent.

PR #14214, opened 2025-06-06
Fix runner to surface generic errors

roaks3:
Note that this is not currently breaking or blocking anything I'm aware of, but when working through the EAP VCR setup, the runner appeared to be swallowing gsutil errors. I made this change directly in the CL, which seemed to work, so it seems like something we want upstream. My theory is that we can't assume all errors coming from CLI tools will be `ExitError` or `PathError`.

```release-note:none

```

### DIFF
--- a/.ci/magician/exec/runner.go
+++ b/.ci/magician/exec/runner.go
@@ -131,6 +131,9 @@ func (ar *Runner) Run(name string, args []string, env map[string]string) (string
 		return "", fmt.Errorf("path error running %s: %v", name, typedErr)
 
 	}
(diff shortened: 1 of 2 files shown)


trodge (dismissed):

trodge:
Could you make a test change (a comment in any handwritten resource file is fine) somewhere that will show up in the provider so we can trigger a run and confirm this won't break anything else in the ci that uses `Runner`?

roaks3:
Will do, thanks for that suggestion!

trodge (approved):

PR #14219, opened 2025-06-09
fix for scaling down to zero replica in Valkey

somesh-m:
```release-note:enhancement
memorystore: added the ability to set the `replica_count`field  in `google_memorystore_instance` to zero.
```

Changes:

- Set send_empty_value as true in instance.yaml for memorystore to support sending 0 as the value for replica count.
- Modified acceptance test to update replica count to 0

Fixes: https://github.com/hashicorp/terraform-provider-google/issues/22783

### DIFF
--- a/mmv1/products/memorystore/Instance.yaml
+++ b/mmv1/products/memorystore/Instance.yaml
@@ -259,6 +259,7 @@ properties:
       "Optional. Number of replica nodes per shard. If omitted the default
       is 0 replicas. "
     default_from_api: true
+    send_empty_value: true
   - name: 'authorizationMode'
     type: String
     description:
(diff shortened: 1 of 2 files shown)


SirGitsalot (commented):
Unfortunately, `send_empty_value` is [incompatible](https://googlecloudplatform.github.io/magic-modules/reference/field/#send_empty_value) with `default_from_api `. The easy workaround would be to remove `default_from_api`, but that's a [breaking change](https://googlecloudplatform.github.io/magic-modules/breaking-changes/breaking-changes/#field-level-breaking-changes). This change will need to go into [TPG v7](https://googlecloudplatform.github.io/magic-modules/breaking-changes/make-a-breaking-change/), which is only a couple of months away.

SirGitsalot:
In the time since I last reviewed the [guidance on `send_empty_value` with `default_from_api` has changed](https://github.com/GoogleCloudPlatform/magic-modules/pull/14029). Running the tests now...

SirGitsalot (approved):

PR #14221, opened 2025-06-09
Add Async logic and small fixes

c2thorn:
this PR won't build, but changes for data fusion instance can be seen here https://github.com/c2thorn/terraform-provider-google-beta/commit/dd525cddcb99f2e79c333f4fe7dac7505b1299b2

```release-note:none

```

### DIFF
--- a/mmv1/templates/terraform/resource_fw.go.tmpl
+++ b/mmv1/templates/terraform/resource_fw.go.tmpl
@@ -94,6 +94,9 @@ type {{$.ResourceName}}FWModel struct {
 	{{ if $.HasProject -}}
 	Project            types.String `tfsdk:"project"`
     {{- end }}
+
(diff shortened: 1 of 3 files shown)


mmv1/templates/terraform/resource_fw.go.tmpl
```diff
@@ -94,6 +94,9 @@ type {{$.ResourceName}}FWModel struct {
 	{{ if $.HasProject -}}
 	Project            types.String `tfsdk:"project"`
     {{- end }}
+
+	Id       types.String `tfsdk:"id"`{{/* TODO should this be gated behind a greenfield/brownfield flag? */}}
```
NickElliot, line 98: I think we can assume this always is generated in, as users may expect that ID to be an exported attribute for every single resource much like SDKv2 resources

c2thorn: That's probably what's going to happen, but I just want to get acknowledgement from the team at some point that this is what we are deciding to do

PR #14225, opened 2025-06-10
Adding support for log_linked_dataset_query_user_email attribute for BigQuery Analytics Hub listings.

arnavdham:
This Pull Request introduces support for the log_linked_dataset_query_user_email attribute within the google_bigquery_analytics_hub_listing resource.

This new optional boolean attribute allows users to control whether the email addresses of users querying linked datasets from an Analytics Hub listing are logged to Cloud Logging for audit purposes. Previously, this logging behavior was not directly configurable via Terraform. [API reference](https://cloud.google.com/bigquery/docs/reference/analytics-hub/rest/v1/projects.locations.dataExchanges.listings#resource:-listing)

@c2thorn @shashambhavi

```release-note:enhancement
bigqueryanalyticshub: added `log_linked_dataset_query_user_email` field to `google_bigquery_analytics_hub_listing` resource
```


mmv1/products/bigqueryanalyticshub/Listing.yaml
```diff
@@ -194,3 +194,9 @@ properties:
         type: Boolean
         description:
           If true, restrict export of query result derived from restricted linked dataset table.
+  - name: 'logLinkedDatasetQueryUserEmail'
+    type: Boolean
+    description:
+      If true, subscriber email logging is enabled and all queries on the linked dataset will log the email address of the querying user.
+    default_value: false
```
c2thorn:
```suggestion
```
Does this default value need to be here? Terraform usually defaults to false for booleans

mmv1/templates/terraform/examples/bigquery_analyticshub_listing_log_linked_dataset_query_user.tf.impl
```diff
@@ -0,0 +1,26 @@
+resource "google_bigquery_analytics_hub_data_exchange" "{{$.PrimaryResourceId}}_log_email" {
```
c2thorn: This should be added to the "examples" block in Listing.yaml, which will add it to our documentation and auto-generate tests.
You can see the examples already in the resource YAML, and instructions here: https://googlecloudplatform.github.io/magic-modules/test/test/#add-a-create-test

After you do this, you can probably delete the handwritten TestAccBigqueryAnalyticsHubListing_logLinkedDatasetQueryUserEmailCreate test, since it seems to do the exact same thing as this?

arnavdham: I have added the examples block in listing.yaml. I don't understand how to run the tests after rewriting code if I remove the TestAccBigqueryAnalyticsHubListing_logLinkedDatasetQueryUserEmailCreate test. According to [this](https://googlecloudplatform.github.io/magic-modules/test/run-tests/),The whole test takes about 9 hours to run and I don't have the handwritten test to pinpoint and run.

mmv1/products/bigqueryanalyticshub/Listing.yaml
arnavdham: fixed

mmv1/templates/terraform/examples/bigquery_analyticshub_listing_log_linked_dataset_query_user.tf.impl
arnavdham: I have tested using manual tests and it works fine. Also fixed minor issues and removed hard-coded parts.

PR #14227, opened 2025-06-10
feat: generate expander behavior for write-only fields

ramonvermeulen:
Closes https://github.com/hashicorp/terraform-provider-google/issues/23211

**Release Note Template for Downstream PRs (will be copied)**

See [Write release notes](https://googlecloudplatform.github.io/magic-modules/contribute/release-notes/) for guidance.

```release-note:enhancement
provider: generate expander behavior for write-only fields
```

PR #14230, opened 2025-06-10
 provider: eliminated the need to manually add `*_wo` and `*_wo_version` for write-only properties

ramonvermeulen:
Closes https://github.com/hashicorp/terraform-provider-google/issues/23213
Closes https://github.com/hashicorp/terraform-provider-google/issues/23212
Closes https://github.com/hashicorp/terraform-provider-google/issues/23214

~Goes together with https://github.com/GoogleCloudPlatform/magic-modules/pull/14227 ensuring there is no overlap for expander fields between the original and the `Wo` field.~

Closed ^, if any of it is still required I rather do it in this single PR to make testing easier.

```release-note:enhancement
provider: eliminated the need to manually add `*_wo` and `*_wo_version` for write-only properties
```


ramonvermeulen:
@BBBmau @melinath 

FYI: Just a quick update since returning from vacation and resuming work on this refactor.

I am making some progress toward generating the desired code, though currently encountering some issues with slice properties being generated as empty slices (e.g., `RequiredWith`, `ConflictsWith` and `ExactlyOneOf`). Wrapping up for now, but I plan to set up a local debugging environment later this week to investigate further (would be nice to set breakpoints).

If anyone is interested in reviewing or investigating this in the meantime, I've added some comments and log statements to help clarify the issue.

**EDIT:**
Slice properties seem to be generated now as well, issue was that the values need to be snake_case terraform resource access paths, and it needs to work for nested write-only fields as well. Will later on look into alternatives to:
- the `camelToSnake` (maybe there is something in the API)
- the `requiredWith = parent.TerraformLineage() + ".0." + writeOnlyField.TerraformLineage()` examples feels a bit "ducktapy"

At least for the current resources the generated code looks close to what it should be:

<details>

**resource_bigquery_data_transfer_config.go**
image

**resource_secret_manager_secret_version.go**
image


</details>

melinath:
@modular-magician reassign-reviewer @BBBmau

ramonvermeulen:
I know it isn't 100% done yet, but would like to get a first review in on what I have so far.

The main question I have at the moment is if there is a way to get the full field path for a `*Type` instance, e.g. `http_check.0.auth_info.0.password` via any of the existing methods. For the issue I am currently facing with the `buildPath` method something like this is required to set properties such as Conflicts, RequiredWith, ExactlyOneOf in a correct manner. Otherwise I might have to build something custom for this, keeping track of the nesting depth constructing a key during the applied recursion for nested properties.

Also I would love to hear your opinion on `mark_write_only_version_mutable`, wondering if there are better ways of not introducing a breaking change. If we decide to take this approach, we must ensure removing this field doesn't get forgotten when moving to the next major (7.0.0).

Another thing I noticed is that this might introduce a second breaking change (`TypeString` will change to `TypeInt`):
https://github.com/GoogleCloudPlatform/magic-modules/blob/7de4ff38ee8db85e6059ec6270531b7b4385e4ba/mmv1/products/monitoring/UptimeCheckConfig.yaml#L264

melinath:
Apologies for the delayed review; I've been caught up in a project and haven't had a chance to take the time to review this. I will be OOO next week but hope to go back through this the week after.

melinath:
(and the other related PR)

ramonvermeulen:
> Apologies for the delayed review; I've been caught up in a project and haven't had a chance to take the time to review this. I will be OOO next week but hope to go back through this the week after.

No worries!

FYI I will also be OOO the first two weeks of July, so I might be a bit slow in replies.

rileykarson:
Both reviewers that have been on this change are unavailable for different periods. @melinath can you pick up review or hand off when you're back?

melinath:
yes, I can pick up the review now that I'm back.

melinath:

mmv1/api/resource.go
```diff
@@ -713,14 +713,88 @@ func (r Resource) GetIdentity() []*Type {
 	})
 }
 
-func (r *Resource) AddLabelsRelatedFields(props []*Type, parent *Type) []*Type {
+func buildFieldPath(parent *Type, fieldName string) string {
+	// TODO doesn't seem to work for deeper nested fields, such as "http_check.0.auth_info.0.password" in monitoring uptime check
+	// parent.TerraformLineage() just returns "auth_info" for the above example resulting in "auth_info.0.password" instead of "http_check.0.auth_info.0.password"
+	// tried to look through other available methods, but couldn't find a better one yet -> looking for input, otherwise will need to implement a custom one
```
melinath: I think the problem here is that that parent metadata is currently set as part of SetDefault, which is called _after_ the `AddExtraFields` function. I'm not sure whether we can safely move SetDefault earlier, but possibly we could restructure how we handle specifically setting ParentMetadata so it can happen earlier? @zli82016 any chance you have specific recommendations / context to add about this?

mmv1/api/type.go
```diff
@@ -173,6 +173,11 @@ type Type struct {
 
 	WriteOnly bool `yaml:"write_only,omitempty"` // Adds `WriteOnly: true` to the schema
 
+	// TODO: remove this field right before the next major release which is 7.0.0
+	// temporary solution to support bigquerydatatransfer being already mutable (not introducing a breaking change)
+	// https://github.com/hashicorp/terraform-provider-google/issues/23214
+	MarkWriteOnlyVersionMutable bool `yaml:"mark_write_only_version_mutable,omitempty"` // Adds `MarkWriteOnlyVersionMutable: true` to the schema
```
melinath: Possibly I'm missing or have forgotten something specific to this case, but generally speaking it's not a breaking change to make a formerly-immutable field mutable. Left a comment on https://github.com/hashicorp/terraform-provider-google/issues/23214 to this effect as well.

mmv1/api/resource.go
```diff
@@ -713,14 +713,88 @@ func (r Resource) GetIdentity() []*Type {
...
+		return parent.TerraformLineage() + ".0." + fieldName
+	}
+	return fieldName
+}
+
+func buildWriteOnlyField(name string, parent *Type, originalField *Type) *Type {
+	description := fmt.Sprintf("%s Note: This property is write-only and will not be read from the API. For more info see [updating write-only attributes](/docs/providers/google/guides/using_write_only_attributes.html#updating-write-only-attributes)", google.Underscore(originalField.Description))
+	conflictsWith := buildFieldPath(parent, originalField.TerraformLineage())
+	exactlyOneOfOriginalField := buildFieldPath(parent, originalField.TerraformLineage())
+	exactlyOneOfWriteOnlyField := buildFieldPath(parent, google.Underscore(name))
```
melinath: This is probably fine for now because of the limited usage, but we'll want to eventually be more comprehensive, because the original field could also have a conflicts / exactlyOneOf / other relationship with other fields as well. I think the logic we'll need is:

- If (and only if) there is a conflicts, exactlyOneOf, or atLeastOneOf set that includes the original field, we should extend it to include the write only field. I _think_ that should work properly even though it would only be set on the original field and write only field, not the other fields in the set.
- I think we probably need to return an error / halt generation if the original field has requiredWith set, since there's no way to split that between the two fields
- If there are _none_ of conflicts, exactlyOneOf, or atLeastOneOf set on the original field, then we should add a conflicts constraint if the field is optional and an exactlyOneOf if it's required (and unset required). (This is already partially implemented but happens in all cases, not checking for potential other sets first.)

But that doesn't need to happen in this PR. @BBBmau maybe we should make a ticket to track this? What do you think?

mmv1/api/resource.go
```diff
@@ -713,14 +713,88 @@ func (r Resource) GetIdentity() []*Type {
...
+	}
+
+	options := []func(*Type){
+		propertyWithType("String"),
+		propertyWithRequired(false),
+		propertyWithDescription(description),
+		propertyWithWriteOnly(true),
+		propertyWithApiName(apiName),
+		propertyWithIgnoreRead(true),
+		propertyWithConflicts([]string{conflictsWith}),
```
melinath: We should only add conflicts if the original field isn't required.

mmv1/api/resource.go
```diff
@@ -713,14 +713,88 @@ func (r Resource) GetIdentity() []*Type {
...
+	}
+
+	if originalField.Required {
+		options = append(options, propertyWithExactlyOneOf([]string{exactlyOneOfOriginalField, exactlyOneOfWriteOnlyField}))
+	}
+
+	return NewProperty(name, originalField.ApiName, options)
+}
+
+func buildWriteOnlyVersionField(name string, parent *Type, writeOnlyField *Type, immutable bool) *Type {
```
melinath: We need to make sure that the write only field isn't sent to the API or set based on the API response. The old YAML accomplished this by setting `url_param_only: true`, but that's not... strictly correct IMO. Really, we should be setting `ClientSide: true` to indicate this is a [client-side field](https://googlecloudplatform.github.io/magic-modules/develop/client-side-fields/), but we [don't currently support client-side fields at arbitrary levels of nesting](https://github.com/hashicorp/terraform-provider-google/issues/19246) so I'm not sure whether that would be sufficient or have additional edge cases.

The labels fields work around this by explicitly [hardcoding the generated effectiveLabels field as provider-only](https://github.com/GoogleCloudPlatform/magic-modules/blob/88d2f81608fee70e7e88c59bbac4434ef3c9294c/mmv1/api/type.go#L1203-L1206) and having specific custom logic in a bunch of other places too.

I'd be inclined to say: let's mark this field as `ClientSide: true` and solve any problems that come up from that perspective. For example, [`IgnoreReadPropertiesToString`](https://github.com/GoogleCloudPlatform/magic-modules/blob/88d2f81608fee70e7e88c59bbac4434ef3c9294c/mmv1/api/resource.go#L1173) should include a check for ClientSide: true (or just call [ProviderOnly](https://github.com/GoogleCloudPlatform/magic-modules/blob/88d2f81608fee70e7e88c59bbac4434ef3c9294c/mmv1/api/type.go#L1202)).

cc @zli82016 @c2thorn in case you have any insight / context / concerns / alternative thoughts

mmv1/products/monitoring/UptimeCheckConfig.yaml
```diff
@@ -246,27 +246,10 @@ properties:
...
-            type: String
-            description: The password to authenticate.
-            exactly_one_of:
-              - 'passwordWo'
-              - 'password'
-            required_with:
-              - 'http_check.0.auth_info.0.password_wo_version'
-            write_only: true
-          - name: 'passwordWoVersion'
-            type: String
```
melinath, line 264: I think you mentioned elsewhere but changing this from string to integer is breaking, so this might need to be customizable for now. (Adding a default value is also a breaking change.)

I'm a little tempted to say that we just hard-code in an exception for this field and then remove it in the major release branch rather than making it fully customizable.

cc @NickElliot & @c2thorn on this since it would require a breaking change in the 7.0.0 release.

mmv1/products/bigquerydatatransfer/Config.yaml
```diff
@@ -223,30 +223,14 @@ properties:
...
-        required_with:
-          - 'sensitive_params.0.secretAccessKeyWo'
-        description: |
-          The version of the sensitive params - used to trigger updates of the write-only params. For more info see [updating write-only attributes](/docs/providers/google/guides/using_write_only_attributes.html#updating-write-only-attributes)
       - name: 'secretAccessKey'
         type: String
         description: |
           The Secret Access Key of the AWS account transferring data from.
         sensitive: true
-        at_least_one_of:
```
melinath, line 237: changing from `at_least_one_of` to `exactly_one_of` is technically a breaking change, since previously users would be able to set both and now they would be able to only set one. If that would have always caused an error, then it's fine to change, otherwise we'd need to stick with at_least_one_of (and switch in a major release).

ramonvermeulen:
> Thanks for your patience! Overall this looks really great - definitely the right direction! But there are some issues we'll need to work out.

Thanks a lot for the extensive review! I just got back from OOO and it's quite busy. Will probably take some days to get back on all the points of feedback.

melinath (commented):
removing from my queue for now

mmv1/api/resource.go
ramonvermeulen: Check, because if `SetDefault` would be called before `AddExtraFields` this method will always return the full lineage path instead of only the lineage derived to it's parent?

mmv1/api/resource.go
ramonvermeulen: Hmm I think I get your point here! The 2nd one is easy to implement, and I think I will do that right away in this PR. For the first and 3rd point I think quite some testing of edge cases is required, with combinations of making the field required or optional and setting the conflicts, exactlyOneOf and atLeastOneOf on the original field.

mmv1/api/resource.go
ramonvermeulen: Even if the field is required the `_wo` still conficts with the original field right? Or because we use exactlyOneOf in that case it is redundant to set on a required field? Will make this change by adding an else clause to the `if originalField.Required` below.

mmv1/api/resource.go
ramonvermeulen: Small nit, but want to make sure we understand eachother correct, in your first sentence you say "We need to make sure that the **_write only field_** isn't ..." I guess you meant to say "We need to make sure that the **_write only version field_** isn't ...".

But I agree with the approach, and will then mark the write only version field as `ClientSide: true`!

mmv1/api/type.go
ramonvermeulen: I applied this logic because of the subissue: https://github.com/hashicorp/terraform-provider-google/issues/23214, that states `removing the field being immutable is a breaking change`. I thought this could offer a temporary solution and can then later be revoked in a major release. However, if it is not a breaking change it is redundant and I will remove it. 

Maybe @BBBmau can confirm this?

mmv1/products/bigquerydatatransfer/Config.yaml
ramonvermeulen: I get your point here, however, since it also has `conflicts` set (e.g. you cannot configure the `_wo` and the original field at the same time) it would've previously caused an error when setting both at the same time. Which I guess is similar functionality to what `exactly_one_of` would do for this case (or I might be overlooking something).

mmv1/products/monitoring/UptimeCheckConfig.yaml
ramonvermeulen: Yes it is a bit of a hard one, since these fields are now "generated" instead of created via the yaml. Maybe somehow I could add a special check that when it is this specific resource (UptimeCheck) it handles it as a string, and for the other cases as an integer? With a comment that the condition should be removed in the next major release?

mmv1/api/resource.go
melinath: I think the problem is that if SetDefault were called before AddExtraFields, SetDefault wouldn't be called for the added fields (unless we called it while adding them.)

mmv1/api/resource.go
melinath: Yeah, makes sense. For now, how about throwing an error if the initial field has a conflicts/exactlyOneOf/atLeastOneOf set, so that we make sure we notice if/when we need to implement that part of the logic (and don't have to deal with it here)

mmv1/api/resource.go
melinath: because conflicts is redundant with exactlyOneOf, yeah :-)

mmv1/api/resource.go
melinath: yes - I did mean write only _version_ field - thanks!

mmv1/products/monitoring/UptimeCheckConfig.yaml
melinath: yeah, that's what I'm leaning towards.

mmv1/products/bigquerydatatransfer/Config.yaml
melinath: you're right - conflicts means that they wouldn't be able to actually set both! And `at_least_one_of` means setting one is required... so it should be equivalent. Weird!

melinath (commented):
responded above to comments; removing from queue for now.

mmv1/api/resource.go
zli82016: What do you think about this option:  add the TerraformLineage of parent as one more parameter for the function `AddExtraFields`? In the function buildFieldPath, we can get the final TerraformLineage of the parent. 


Or we can have a new function to handle specifically setting ParentMetadata. The new function could be called both before and after `AddExtraFields`

mmv1/api/resource.go
zli82016: Sounds good to me to add a flag. I wonder should the flag be called `virtual_filed: true` or `ClientSide: true`

melinath: The flag is already called `ClientSide: true`. (As a side note, personally I think we should standardize on ClientSide over "virtual" wherever possible, since it's more descriptive & a more common term... but that's not something we need to decide right now.)

mmv1/api/resource.go
melinath: I don't think we can trust any of the Lineage functions before ParentMetadata is set...

So I'd lean towards a new function for specifically setting ParentMetadata. I think we'd only need to call it before AddExtraFields, since none of the fields we add are Arrays, Maps, or NestedObjects.

mmv1/products/monitoring/UptimeCheckConfig.yaml
melinath: Discussed with @c2thorn & @NickElliot and an easier path forward here might be to standardize the "version" field on strings instead of integers, since converting an int to a string _should_ be backwards-compatible (since Terraform will automatically convert it to a string.) @BBBmau would there be any issue with doing that from your perspective?

mmv1/api/resource.go
ramonvermeulen: Just a small FYI because I've been a bit radio silent, next Friday I have an innovation day, meaning I have a lot of time to look into this and go through all feedback. Will also use that time to look into what `SetDefault` actually does to have a better understanding what is required for the `SetParentMetadata`.

zli82016: Sorry, I should add more details for the first option.

We don't use the function `TerraformLineage`. Instead, we manually build the lineage by passing `parentLineage` as an parameter to AddExtraFields. `parentLineage` is empty string as the initial value and then it will append the property name by calling the function `AddExtraFields` recursively.
```
func (r *Resource) AddExtraFields(props []*Type, parent *Type, parentLineage string) []*Type {
	for _, p := range props {
		if p.WriteOnly && !strings.HasSuffix(p.Name, "Wo") {
			props = r.addWriteOnlyFields(props, parent, p, lineage)
		}
		if p.IsA("KeyValueLabels") {
			props = r.addLabelsFields(props, parent, p)
		} else if p.IsA("KeyValueAnnotations") {
			props = r.addAnnotationsFields(props, parent, p)
		} else if p.IsA("NestedObject") && len(p.AllProperties()) > 0 {
                         lineage := fmt.Sprintf("%s%s", parentLineage, p.Name)
			p.Properties = r.AddExtraFields(p.AllProperties(), p, lineage)
		}
	}
      
}
```

melinath: oh, I see! that makes sense, thanks for the details! I would still lean towards a separate function for setting up the lineage data so that we only have one way that lineage is determined - otherwise I think we're setting ourselves up for confusion later.

zli82016: That makes sense.

mmv1/api/type.go
BBBmau: ah yes i must've misunderstood something when first writing the issue. This would be a breaking change if we were moving from mutable to immutable but this isn't the case.

mmv1/api/resource.go
BBBmau: I'll create the issue, these are good points, i imagine that with the 3rd point

> If there are none of conflicts, exactlyOneOf, or atLeastOneOf set on the original field, then we should add a conflicts constraint if the field is optional and an exactlyOneOf if it's required (and unset required). (This is already partially implemented but happens in all cases, not checking for potential other sets first.)

involves using the constraints where both original + wo are the only fields used. e.g value is optional therefore it has conflicts with [original, wo] ; value is required therefore it has exactlyOneOf with [original, wo]

Edit: issue created https://github.com/hashicorp/terraform-provider-google/issues/23698

mmv1/products/monitoring/UptimeCheckConfig.yaml
ramonvermeulen: Check, applied this in https://github.com/GoogleCloudPlatform/magic-modules/pull/14230/commits/c779d023fe7f5051a2fbf18104224f4e291fc95b.

Does this also mean I do not need to make an exception clause for the following two:

mmv1/products/secretmanager/SecretVersion.yaml
mmv1/products/bigquerydatatransfer/Config.yaml

Because these both "used" to be `Integer`? If this is still required I will do a check somewhere in the `AddExtraFields` logic, and make specifically these two fields Integer with a TODO comment to remove the checks just before the next major release.

mmv1/api/type.go
ramonvermeulen: Check, removed this temporary immutability configuration in https://github.com/GoogleCloudPlatform/magic-modules/pull/14230/commits/228ba85950bb1d1c6918025ad12d3fd46cfd58d4, also makes the overall code simpler :)

mmv1/api/resource.go
ramonvermeulen: Check! Applied in https://github.com/GoogleCloudPlatform/magic-modules/pull/14230/commits/635a47979d6099c48061f17dc1c0206c4edb058c.

mmv1/api/resource.go
ramonvermeulen: @melinath @zli82016 
What do you think of the following approach that I implemented in https://github.com/GoogleCloudPlatform/magic-modules/pull/14230/commits/858a3840c858d56bf9b01b3cbb7bc39c7d43ff36 and https://github.com/GoogleCloudPlatform/magic-modules/pull/14230/commits/16e70591e5e7232cf312e61035fe905871a39397?:

1) I swapped the order of execution, calling `SetDefault` before `AddExtraFields`
2) To ensure every "generated" property also has the defaults set, I call `SetDefault` on every property that is created by the `AddExtraFields` method.

Could potentially also make a wrapper, instead of `append()` it does `props := setDefaultAndAppend(props, extraPropX, extraPropY)` so that new contributors do not forget to call the `SetDefault` for extra fields?

mmv1/api/resource.go
ramonvermeulen: Applied in https://github.com/GoogleCloudPlatform/magic-modules/pull/14230/commits/e959127d97aafb98b58bf72a5ecab2bf31c0a05b.

mmv1/api/resource.go
ramonvermeulen: Tried to apply this in:

https://github.com/GoogleCloudPlatform/magic-modules/pull/14230/commits/5f8c41eff789be4da3e90e9e375eca78452fe1fb
And
https://github.com/GoogleCloudPlatform/magic-modules/pull/14230/commits/b58cb0180f61144ed21d69c732325b6e38cc7098

If this has other side-effects or doesn't work as expected I could also move to your point of just throwing an error if either one of these are configured on the original field.

mmv1/products/monitoring/UptimeCheckConfig.yaml
melinath: The theory here is that changing from an Integer to a String _may not_ be a breaking change, because the String field might accept integer inputs. However, we should definitely double-check that before releasing.

ramonvermeulen: Yes, makes sense. Any Integer should be able to safely be casted to a string, however, not the other way around.

Just wondering if this works without breaking change, in a sense that when you already have a terraform configuration like:

```hcl
blabla_wo_version = 3
```

It keeps working and applies the casting to the string without any kind of type error.

mmv1/api/resource.go
melinath: If this were working the way I'd been imagining, you would be able to call field.TerraformLineage() directly - buildFieldPath would be unnecessary.

The problem is that SetDefault works recursively starting from the root of the tree and working its way down. Thinking through this again, even a separate function for setting up the lineage data would struggle with this - basically, the new fields can't have their lineage set until they're already in the tree - but we're creating new fields _outside_ of the tree and then adding them, so of course the lineage data isn't available for setting properties on those fields.

So - the thing I was imagining definitely not completely feasible with the current implementation (and may just not be possible).

That being said, SetDefault needs to be called on the _parent_ (if there is one) after the fields are added, not on the fields themselves (because it's recursive). That would probably (?) be sufficient for now.

ramonvermeulen: Hmm, I get that point. However, we need the lineage at a point even before the field itself is generated. Therefore I thought it could make sense to swap the `SetDefault` and `AddExtraFields` so that in the AddExtraFields, at least all the "non-generated" fields have the lineage data and all other defaults available. And then use TerraformLineage() on the parent of the original field. Then after field generation I call SetDefault on themselves because I don't know what the downstream impact is if these generated fields don't have their defaults set.

As far I can see in the generated code, at least for the 3 cases of Wo fields that we have, it seems to work. However, I also see some (maybe related?) side effects, have to investigate what is going on there:

image

EDIT:
Above is indeed caused by swapping the `SetDefault` and `AddExtraFields` order. If I call `SetDefault` before AND after `AddExtraFields` it works fine, but that's quite inefficient so will look further for a solution.

mmv1/products/monitoring/UptimeCheckConfig.yaml
melinath: yeah that's the question

mmv1/api/resource.go
melinath: yes, calling SetDefault first makes sense, and using buildFieldPath makes sense. But instead of calling SetDefault on the new fields, you need to call it on the parent field they've been added to - that will call it recursively on each of the added fields and set them up properly.

ramonvermeulen: Check, will play a bit around with this and also look into potential other solutions.

mmv1/products/monitoring/UptimeCheckConfig.yaml
ramonvermeulen: I will later build the provider and have a pre-existing wo resource, then do some manual testing to check if that works as expected 👍

mmv1/api/resource.go
ramonvermeulen: I tried quite a lot of things, e.g. checking if parent != nil and then calling SetDefault, but couldn't get it to work. Eventually I got two solutions to work properly:

1) Just call SetDefault on every field again: https://github.com/GoogleCloudPlatform/magic-modules/pull/14230/commits/6d38e482331b093bc44c599ff953cfc02f04aabb (don't like this so much, because you're esentially doing things twice)
2) Revert to old order, e.g. first AddExtraFields and then SetDefault. And build up the lineage yourself in the AddExtraFields, like in https://github.com/GoogleCloudPlatform/magic-modules/pull/14230/commits/10b19ec88eff38e3b369ec962e3dbbc7956f1f59

For now I call it a day, but I'm open for any other (potential better) solutions. I think the last solution is quite okay, however, slightly less readable, but maybe a bit of renaming/small refactors can make it better.

ramonvermeulen:
@melinath @BBBmau 

Tried to apply all of the feedback and also updated the connected PR (https://github.com/GoogleCloudPlatform/magic-modules/pull/14227) for expander behaviour.

Would love another review! Maybe it is also good to let the downstream generation run so that we can take a look and discuss the generated code as well. Eventually went for recursively creating the lineage within the method itself via `buildCurrentPropLineage`, at least for the existing write-only fields this seems to work well and SetDefault can just stay as it is and also run for all newly generated fields.

Will still do the test case for breaking changes when I have time, as discussed in:
https://github.com/GoogleCloudPlatform/magic-modules/pull/14230/files#r2231552501.

**EDIT:**

Good news about the compatibility test, maybe someone else can follow the same steps to confirm it twice, but the int to string conversion seems to work without compatibility issues! I described all my testing steps in the following details block to make it reproducable:

<details>
<summary><b>OPEN ME</b></summary>

Here a brief walkthrough of the exact steps executed to do this test:

### 1. Terraform version

```bash
❯ terraform version
Terraform v1.12.2
on darwin_arm64
+ provider registry.terraform.io/hashicorp/google v6.45.0
```

### 2. Configuration
Masked out project id and numbers throughout the examples.

```hcl
locals {
  project_id = "my-gcp-project"
}

provider "google" {
  project = local.project_id
}

resource "google_secret_manager_secret" "default" {
  secret_id = "test-secret-wo-compatibility"
  replication {
    auto {}
  }
}

resource "google_secret_manager_secret_version" "default" {
  secret                 = google_secret_manager_secret.default.id
  secret_data_wo_version = 1
  secret_data_wo         = "I am write-only ;)"
}
```

### 3. Applied configuration

```
❯ terraform apply -auto-approve

  + create

  # google_secret_manager_secret.default will be created
  + resource "google_secret_manager_secret" "default" {
      + create_time           = (known after apply)
      + deletion_protection   = false
      + effective_annotations = (known after apply)
      + effective_labels      = {
          + "goog-terraform-provisioned" = "true"
        }
      + expire_time           = (known after apply)
      + id                    = (known after apply)
      + name                  = (known after apply)
      + project               = "my-gcp-project"
      + secret_id             = "test-secret-wo-compatibility"
      + terraform_labels      = {
          + "goog-terraform-provisioned" = "true"
        }

      + replication {
          + auto {
            }
        }
    }

  # google_secret_manager_secret_version.default will be created
  + resource "google_secret_manager_secret_version" "default" {
      + create_time            = (known after apply)
      + deletion_policy        = "DELETE"
      + destroy_time           = (known after apply)
      + enabled                = true
      + id                     = (known after apply)
      + is_secret_data_base64  = false
      + name                   = (known after apply)
      + secret                 = (known after apply)
      + secret_data_wo         = (write-only attribute)
      + secret_data_wo_version = 1
      + version                = (known after apply)
    }

Plan: 2 to add, 0 to change, 0 to destroy.
google_secret_manager_secret.default: Creating...
google_secret_manager_secret.default: Creation complete after 2s [id=projects/my-gcp-project/secrets/test-secret-wo-compatibility]
google_secret_manager_secret_version.default: Creating...
google_secret_manager_secret_version.default: Creation complete after 3s [id=projects/XXXXXXXXX/secrets/test-secret-wo-compatibility/versions/1]

Apply complete! Resources: 2 added, 0 changed, 0 destroyed.
```

And the terraform state after apply looks like the following:

```
{
  "version": 4,
  "terraform_version": "1.12.2",
  "serial": 31,
  "lineage": "4812b715-adbd-0d16-f5f0-8be8ba3fe232",
  "outputs": {},
  "resources": [
    {
      "mode": "managed",
      "type": "google_secret_manager_secret",
      "name": "default",
      "provider": "provider[\"registry.terraform.io/hashicorp/google\"]",
      "instances": [
        {
          "schema_version": 0,
          "attributes": {
            "annotations": null,
            "create_time": "2025-07-30T06:18:08.589955Z",
            "deletion_protection": false,
            "effective_annotations": {},
            "effective_labels": {
              "goog-terraform-provisioned": "true"
            },
            "expire_time": "",
            "id": "projects/my-gcp-project/secrets/test-secret-wo-compatibility",
            "labels": null,
            "name": "projects/XXXXXXXXXXX/secrets/test-secret-wo-compatibility",
            "project": "my-gcp-project",
            "replication": [
              {
                "auto": [
                  {
                    "customer_managed_encryption": []
                  }
                ],
                "user_managed": []
              }
            ],
            "rotation": [],
            "secret_id": "test-secret-wo-compatibility",
            "tags": null,
            "terraform_labels": {
              "goog-terraform-provisioned": "true"
            },
            "timeouts": null,
            "topics": [],
            "ttl": null,
            "version_aliases": null,
            "version_destroy_ttl": ""
          },
          "sensitive_attributes": [],
          "identity_schema_version": 0,
          "private": "eyJlMmJmYjczMC1lY2FhLTExZTYtOGY4OC0zNDM2M2JjN2M0YzAiOnsiY3JlYXRlIjoxMjAwMDAwMDAwMDAwLCJkZWxldGUiOjEyMDAwMDAwMDAwMDAsInVwZGF0ZSI6MTIwMDAwMDAwMDAwMH19"
        }
      ]
    },
    {
      "mode": "managed",
      "type": "google_secret_manager_secret_version",
      "name": "default",
      "provider": "provider[\"registry.terraform.io/hashicorp/google\"]",
      "instances": [
        {
          "schema_version": 0,
          "attributes": {
            "create_time": "2025-07-30T06:18:12.092229Z",
            "deletion_policy": "DELETE",
            "destroy_time": "",
            "enabled": true,
            "id": "projects/XXXXXXXXXXX/secrets/test-secret-wo-compatibility/versions/1",
            "is_secret_data_base64": false,
            "name": "projects/XXXXXXXXXXX/secrets/test-secret-wo-compatibility/versions/1",
            "secret": "projects/my-gcp-projectv/secrets/test-secret-wo-compatibility",
            "secret_data": null,
            "secret_data_wo": null,
            "secret_data_wo_version": 1,
            "timeouts": null,
            "version": "1"
          },
          "sensitive_attributes": [
            [
              {
                "type": "get_attr",
                "value": "secret_data"
              }
            ]
          ],
          "identity_schema_version": 0,
          "private": "eyJlMmJmYjczMC1lY2FhLTExZTYtOGY4OC0zNDM2M2JjN2M0YzAiOnsiY3JlYXRlIjoxMjAwMDAwMDAwMDAwLCJkZWxldGUiOjEyMDAwMDAwMDAwMDAsInVwZGF0ZSI6MTIwMDAwMDAwMDAwMH19",
          "dependencies": [
            "google_secret_manager_secret.default"
          ]
        }
      ]
    }
  ],
  "check_results": null
}

```

### 4. Build the provider based upon the branch of this PR

```bash
make provider VERSION=ga OUTPUT_PATH="$GOPATH/src/github.com/hashicorp/terraform-provider-google"
cd "$GOPATH/src/github.com/hashicorp/terraform-provider-google"
make build
```

And create the tf-dev-override.tfrc in the same directory as the terraform config:

```hcl
provider_installation {
  dev_overrides {
      "hashicorp/google" = "/Users/ramon/go/bin/" # your bin location
      "hashicorp/google-beta" = "/Users/ramon/go/bin/"  # your bin location
  }
  direct {}
}
```

### 5. Execute terraform plan using `TF_CLI_CONFIG_FILE` environment variable

```
➜ TF_CLI_CONFIG_FILE="$(pwd)/tf-dev-override.tfrc" terraform plan

╷
│ 
│ 
╵
google_secret_manager_secret.default: Refreshing state... [id=projects/my-gcp-project/secrets/test-secret-wo-compatibility]
google_secret_manager_secret_version.default: Refreshing state... [id=projects/XXXXXXXXXXX/secrets/test-secret-wo-compatibility/versions/1]

```

### 6. Execute terraform apply using `TF_CLI_CONFIG_FILE` environment variable

```
 
➜ TF_CLI_CONFIG_FILE="$(pwd)/tf-dev-override.tfrc" terraform apply 

╷
│ 
│ 
╵
google_secret_manager_secret.default: Refreshing state... [id=projects/my-gcp-project/secrets/test-secret-wo-compatibility]
google_secret_manager_secret_version.default: Refreshing state... [id=projects/XXXXXXXXXXX/secrets/test-secret-wo-compatibility/versions/1]

```

### 7. Verify terraform state file afterwards, and indeed changed from int to string

```
{
  "version": 4,
  "terraform_version": "1.12.2",
  "serial": 32,
  "lineage": "4812b715-adbd-0d16-f5f0-8be8ba3fe232",
  "outputs": {},
  "resources": [
    {
      "mode": "managed",
      "type": "google_secret_manager_secret",
      "name": "default",
      "provider": "provider[\"registry.terraform.io/hashicorp/google\"]",
      "instances": [
        {
          "schema_version": 0,
          "attributes": {
            "annotations": {},
            "create_time": "2025-07-30T06:18:08.589955Z",
            "deletion_protection": false,
            "effective_annotations": {},
            "effective_labels": {
              "goog-terraform-provisioned": "true"
            },
            "expire_time": "",
            "id": "projects/my-gcp-project/secrets/test-secret-wo-compatibility",
            "labels": {},
            "name": "projects/XXXXXXXXXXX/secrets/test-secret-wo-compatibility",
            "project": "my-gcp-project",
            "replication": [
              {
                "auto": [
                  {
                    "customer_managed_encryption": []
                  }
                ],
                "user_managed": []
              }
            ],
            "rotation": [],
            "secret_id": "test-secret-wo-compatibility",
            "tags": null,
            "terraform_labels": {
              "goog-terraform-provisioned": "true"
            },
            "timeouts": null,
            "topics": [],
            "ttl": null,
            "version_aliases": {},
            "version_destroy_ttl": ""
          },
          "sensitive_attributes": [],
          "identity_schema_version": 0,
          "private": "eyJlMmJmYjczMC1lY2FhLTExZTYtOGY4OC0zNDM2M2JjN2M0YzAiOnsiY3JlYXRlIjoxMjAwMDAwMDAwMDAwLCJkZWxldGUiOjEyMDAwMDAwMDAwMDAsInVwZGF0ZSI6MTIwMDAwMDAwMDAwMH19"
        }
      ]
    },
    {
      "mode": "managed",
      "type": "google_secret_manager_secret_version",
      "name": "default",
      "provider": "provider[\"registry.terraform.io/hashicorp/google\"]",
      "instances": [
        {
          "schema_version": 0,
          "attributes": {
            "create_time": "2025-07-30T06:18:12.092229Z",
            "deletion_policy": "DELETE",
            "destroy_time": "",
            "enabled": true,
            "id": "projects/XXXXXXXXXXX/secrets/test-secret-wo-compatibility/versions/1",
            "is_secret_data_base64": false,
            "name": "projects/XXXXXXXXXXX/secrets/test-secret-wo-compatibility/versions/1",
            "secret": "projects/my-gcp-project/secrets/test-secret-wo-compatibility",
            "secret_data": null,
            "secret_data_wo": null,
            "secret_data_wo_version": "1",
            "timeouts": null,
            "version": "1"
          },
          "sensitive_attributes": [
            [
              {
                "type": "get_attr",
                "value": "secret_data"
              }
            ]
          ],
          "identity_schema_version": 0,
          "private": "eyJlMmJmYjczMC1lY2FhLTExZTYtOGY4OC0zNDM2M2JjN2M0YzAiOnsiY3JlYXRlIjoxMjAwMDAwMDAwMDAwLCJkZWxldGUiOjEyMDAwMDAwMDAwMDAsInVwZGF0ZSI6MTIwMDAwMDAwMDAwMH19",
          "dependencies": [
            "google_secret_manager_secret.default"
          ]
        }
      ]
    }
  ],
  "check_results": null
}
```

As you can see the `            "secret_data_wo_version": "1",` changed from int to str.
</details>

I see the tests just ran, and the tests related to `MonitoringUptimeCheck` seem to fail, I will later check what is causing this and see if I can determine the problem and update the test cases if necessary.

mmv1/products/monitoring/UptimeCheckConfig.yaml
ramonvermeulen: @melinath @BBBmau 

Just tested this, see https://github.com/GoogleCloudPlatform/magic-modules/pull/14230#issuecomment-3133724171.

Seems to work fine! Would maybe be nice if someone else can do the same test to ensure this 100%, but for me it converts fine from int to str and doesn't give any compatibility/diff issues.

ramonvermeulen:
> 🔴 Tests failed during RECORDING mode: `TestAccMonitoringUptimeCheckConfig_changeNonUpdatableFields` [[Error message](https://storage.cloud.google.com/ci-vcr-logs/beta/refs/heads/auto-pr-14230/artifacts/2f37b9df-91dd-4e21-8785-ec1ea63fbd07/build-log/recording_build/TestAccMonitoringUptimeCheckConfig_changeNonUpdatableFields_recording_test.log)] [[Debug log](https://storage.cloud.google.com/ci-vcr-logs/beta/refs/heads/auto-pr-14230/artifacts/2f37b9df-91dd-4e21-8785-ec1ea63fbd07/recording/TestAccMonitoringUptimeCheckConfig_changeNonUpdatableFields.log)] `TestAccMonitoringUptimeCheckConfig_noProjectId` [[Error message](https://storage.cloud.google.com/ci-vcr-logs/beta/refs/heads/auto-pr-14230/artifacts/2f37b9df-91dd-4e21-8785-ec1ea63fbd07/build-log/recording_build/TestAccMonitoringUptimeCheckConfig_noProjectId_recording_test.log)] [[Debug log](https://storage.cloud.google.com/ci-vcr-logs/beta/refs/heads/auto-pr-14230/artifacts/2f37b9df-91dd-4e21-8785-ec1ea63fbd07/recording/TestAccMonitoringUptimeCheckConfig_noProjectId.log)] `TestAccMonitoringUptimeCheckConfig_update` [[Error message](https://storage.cloud.google.com/ci-vcr-logs/beta/refs/heads/auto-pr-14230/artifacts/2f37b9df-91dd-4e21-8785-ec1ea63fbd07/build-log/recording_build/TestAccMonitoringUptimeCheckConfig_update_recording_test.log)] [[Debug log](https://storage.cloud.google.com/ci-vcr-logs/beta/refs/heads/auto-pr-14230/artifacts/2f37b9df-91dd-4e21-8785-ec1ea63fbd07/recording/TestAccMonitoringUptimeCheckConfig_update.log)] `TestAccMonitoringUptimeCheckConfig_update_wo` [[Error message](https://storage.cloud.google.com/ci-vcr-logs/beta/refs/heads/auto-pr-14230/artifacts/2f37b9df-91dd-4e21-8785-ec1ea63fbd07/build-log/recording_build/TestAccMonitoringUptimeCheckConfig_update_wo_recording_test.log)] [[Debug log](https://storage.cloud.google.com/ci-vcr-logs/beta/refs/heads/auto-pr-14230/artifacts/2f37b9df-91dd-4e21-8785-ec1ea63fbd07/recording/TestAccMonitoringUptimeCheckConfig_update_wo.log)] `TestAccMonitoringUptimeCheckConfig_uptimeCheckConfigHttpPasswordWoExample` [[Error message](https://storage.cloud.google.com/ci-vcr-logs/beta/refs/heads/auto-pr-14230/artifacts/2f37b9df-91dd-4e21-8785-ec1ea63fbd07/build-log/recording_build/TestAccMonitoringUptimeCheckConfig_uptimeCheckConfigHttpPasswordWoExample_recording_test.log)] [[Debug log](https://storage.cloud.google.com/ci-vcr-logs/beta/refs/heads/auto-pr-14230/artifacts/2f37b9df-91dd-4e21-8785-ec1ea63fbd07/recording/TestAccMonitoringUptimeCheckConfig_uptimeCheckConfigHttpPasswordWoExample.log)]
> 
> 🔴 Errors occurred during RECORDING mode. Please fix them to complete your PR.
> 
> View the [build log](https://storage.cloud.google.com/ci-vcr-logs/beta/refs/heads/auto-pr-14230/artifacts/2f37b9df-91dd-4e21-8785-ec1ea63fbd07/build-log/recording_test.log) or the [debug log](https://console.cloud.google.com/storage/browser/ci-vcr-logs/beta/refs/heads/auto-pr-14230/artifacts/2f37b9df-91dd-4e21-8785-ec1ea63fbd07/recording) for each test

These failing tests seems to be related to how the custom flatten is now generated for the `_wo_version` field, this should be the same as on the left side:

image

So just returning how it is configured, since it is a client-side only field.

Added a fix in https://github.com/GoogleCloudPlatform/magic-modules/pull/14230/commits/c147a5c5c164db841e4a8893165a7f51c755de2e.

All tests locally are succeeding now, except one but that seems to be a permission issue:

<details>
<summary><b>Logs</b></summary>

```
❯ make testacc TEST=./google/services/monitoring TESTARGS='-run=TestAccMonitoringUptimeCheckConfig_'

TF_ACC=1 TF_SCHEMA_PANIC_ON_ERROR=1 go test ./google/services/monitoring -v -run=TestAccMonitoringUptimeCheckConfig_ -timeout 240m -ldflags="-X=github.com/hashicorp/terraform-provider-google/version.ProviderVersion=acc"
=== RUN   TestAccMonitoringUptimeCheckConfig_uptimeCheckConfigHttpExample
=== PAUSE TestAccMonitoringUptimeCheckConfig_uptimeCheckConfigHttpExample
=== RUN   TestAccMonitoringUptimeCheckConfig_uptimeCheckConfigHttpPasswordWoExample
=== PAUSE TestAccMonitoringUptimeCheckConfig_uptimeCheckConfigHttpPasswordWoExample
=== RUN   TestAccMonitoringUptimeCheckConfig_uptimeCheckConfigStatusCodeExample
=== PAUSE TestAccMonitoringUptimeCheckConfig_uptimeCheckConfigStatusCodeExample
=== RUN   TestAccMonitoringUptimeCheckConfig_uptimeCheckConfigHttpsExample
=== PAUSE TestAccMonitoringUptimeCheckConfig_uptimeCheckConfigHttpsExample
=== RUN   TestAccMonitoringUptimeCheckConfig_uptimeCheckTcpExample
=== PAUSE TestAccMonitoringUptimeCheckConfig_uptimeCheckTcpExample
=== RUN   TestAccMonitoringUptimeCheckConfig_uptimeCheckConfigSyntheticMonitorExample
=== PAUSE TestAccMonitoringUptimeCheckConfig_uptimeCheckConfigSyntheticMonitorExample
=== RUN   TestAccMonitoringUptimeCheckConfig_update
=== PAUSE TestAccMonitoringUptimeCheckConfig_update
=== RUN   TestAccMonitoringUptimeCheckConfig_update_wo
=== PAUSE TestAccMonitoringUptimeCheckConfig_update_wo
=== RUN   TestAccMonitoringUptimeCheckConfig_noProjectId
=== PAUSE TestAccMonitoringUptimeCheckConfig_noProjectId
=== RUN   TestAccMonitoringUptimeCheckConfig_changeNonUpdatableFields
=== PAUSE TestAccMonitoringUptimeCheckConfig_changeNonUpdatableFields
=== RUN   TestAccMonitoringUptimeCheckConfig_jsonPathUpdate
=== PAUSE TestAccMonitoringUptimeCheckConfig_jsonPathUpdate
=== CONT  TestAccMonitoringUptimeCheckConfig_uptimeCheckConfigHttpExample
=== CONT  TestAccMonitoringUptimeCheckConfig_update
=== CONT  TestAccMonitoringUptimeCheckConfig_changeNonUpdatableFields
=== CONT  TestAccMonitoringUptimeCheckConfig_uptimeCheckConfigHttpsExample
=== CONT  TestAccMonitoringUptimeCheckConfig_uptimeCheckConfigSyntheticMonitorExample
=== CONT  TestAccMonitoringUptimeCheckConfig_uptimeCheckTcpExample
=== CONT  TestAccMonitoringUptimeCheckConfig_noProjectId
=== CONT  TestAccMonitoringUptimeCheckConfig_jsonPathUpdate
=== CONT  TestAccMonitoringUptimeCheckConfig_uptimeCheckConfigStatusCodeExample
=== CONT  TestAccMonitoringUptimeCheckConfig_update_wo
=== CONT  TestAccMonitoringUptimeCheckConfig_uptimeCheckConfigHttpPasswordWoExample
--- PASS: TestAccMonitoringUptimeCheckConfig_uptimeCheckConfigStatusCodeExample (17.00s)
--- PASS: TestAccMonitoringUptimeCheckConfig_uptimeCheckConfigHttpsExample (18.71s)
--- PASS: TestAccMonitoringUptimeCheckConfig_uptimeCheckConfigHttpPasswordWoExample (19.64s)
--- PASS: TestAccMonitoringUptimeCheckConfig_uptimeCheckConfigHttpExample (20.38s)
--- PASS: TestAccMonitoringUptimeCheckConfig_jsonPathUpdate (25.19s)
--- PASS: TestAccMonitoringUptimeCheckConfig_noProjectId (26.01s)
--- PASS: TestAccMonitoringUptimeCheckConfig_uptimeCheckTcpExample (27.30s)
--- PASS: TestAccMonitoringUptimeCheckConfig_update (28.24s)
--- PASS: TestAccMonitoringUptimeCheckConfig_update_wo (29.50s)
=== NAME  TestAccMonitoringUptimeCheckConfig_uptimeCheckConfigSyntheticMonitorExample
    resource_monitoring_uptime_check_config_generated_test.go:372: Step 1/2 error: Error running apply: exit status 1

        Error: Error waiting to create function: Error waiting for Creating function: Error code 3, message: Build failed with status: FAILURE. Could not build the function due to a missing permission on the build service account. If you didn't revoke that permission explicitly, this could be caused by a change in the organization policies. Please refer to the following documentation for more details and resolution: https://cloud.google.com/functions/docs/troubleshooting#build-service-account
         You can also view the logs at https://console.cloud.google.com/cloud-build/builds;region=us-central1/ed61e683-c1bd-46a7-9f1c-d0525c195e8c?project=XXXXXXXX.


          with google_cloudfunctions2_function.function,
          on terraform_plugin_test.tf line 14, in resource "google_cloudfunctions2_function" "function":
          14: resource "google_cloudfunctions2_function" "function" {

--- FAIL: TestAccMonitoringUptimeCheckConfig_uptimeCheckConfigSyntheticMonitorExample (34.39s)
--- PASS: TestAccMonitoringUptimeCheckConfig_changeNonUpdatableFields (37.70s)
FAIL
FAIL    github.com/hashicorp/terraform-provider-google/google/services/monitoring       38.502s
FAIL
make: *** [testacc] Error 1
```

</details>

melinath:
Thanks! I'll take another pass on this tomorrow.

mmv1/api/resource.go
```diff
@@ -730,14 +730,90 @@ func (r Resource) GetIdentity() []*Type {
 	})
 }
 
-func (r *Resource) AddLabelsRelatedFields(props []*Type, parent *Type) []*Type {
+func buildWriteOnlyField(name string, originalField *Type, originalFieldLineage string) *Type {
+	description := fmt.Sprintf("%s Note: This property is write-only and will not be read from the API. For more info see [updating write-only attributes](/docs/providers/google/guides/using_write_only_attributes.html#updating-write-only-attributes)", google.Underscore(originalField.Description))
```
melinath: This isn't necessary - messes up the formatting. ([Example](https://github.com/modular-magician/terraform-provider-google-beta/compare/0f5853760f309f26df789249b5cada3888009b31..43f7bd63bcc6d2784dc4aa3186092a7842579a1a#diff-2bd7e175492f07dbf376c98182c9474c67ceaa61534923bf2ff5bf9fab1e54c1R103))

```suggestion
	description := fmt.Sprintf("%s Note: This property is write-only and will not be read from the API. For more info see [updating write-only attributes](/docs/providers/google/guides/using_write_only_attributes.html#updating-write-only-attributes)", originalField.Description)
```

mmv1/api/resource.go
```diff
@@ -730,14 +730,90 @@ func (r Resource) GetIdentity() []*Type {
...
+	return NewProperty(name, originalField.ApiName, options)
+}
+
+func buildWriteOnlyVersionField(name string, originalField *Type, writeOnlyField *Type, originalFieldLineage string) *Type {
+	description := fmt.Sprintf("Triggers update of %s write-only. For more info see [updating write-only attributes](/docs/providers/google/guides/using_write_only_attributes.html#updating-write-only-attributes)", google.Underscore(writeOnlyField.Name))
+	requiredWith := strings.ReplaceAll(originalFieldLineage, google.Underscore(originalField.Name), google.Underscore(writeOnlyField.Name))
+
+	options := []func(*Type){
+		propertyWithType("String"),
+		propertyWithImmutable(true),
```
melinath: I think this should only be immutable if the originalField is immutable. @BBBmau can you confirm?

mmv1/templates/terraform/flatten_property_method.go.tmpl
```diff
@@ -18,7 +18,7 @@
     {{- $.CustomTemplate $.CustomFlatten false -}}
 {{- else -}}
 func flatten{{$.GetPrefix}}{{$.TitlelizeProperty}}(v interface{}, d *schema.ResourceData, config *transport_tpg.Config) interface{} {
-  {{- if $.IgnoreRead }}
+  {{- if or $.IgnoreRead (hasSuffix $.Name "WoVersion") }}
```
melinath: I think we can rely on `ClientSide` here instead of the field name.

```suggestion
  {{- if or $.IgnoreRead $.ClientSide }}
```

mmv1/products/monitoring/UptimeCheckConfig.yaml
melinath: I've also tested moving from a version where a WoVersion field is an int to a version where it's a string - worked fine, not even a diff.

Actually, it looks like there is one case that changes slightly: if no value is set, there's a diff from 0 to null, which does trigger replacement if that field is ForceNew (but I think that should only be the case for secret_manager_secret_version, which I think is probably acceptable.)

melinath: Actually, it looks like there is one case that changes slightly: if no value is set, there's a diff from `0` to `null`, which does trigger replacement if that field is ForceNew (but I think that should only be the case for secret_manager_secret_version, which I think is probably acceptable.)

mmv1/api/resource.go
melinath: yeah, it would be _nice_ to clean this up but given it wasn't easy to do, this workaround is fine 👍  We can always come back to it later.

mmv1/api/resource.go
```diff
@@ -730,14 +730,90 @@ func (r Resource) GetIdentity() []*Type {
...
+	description := fmt.Sprintf("%s Note: This property is write-only and will not be read from the API. For more info see [updating write-only attributes](/docs/providers/google/guides/using_write_only_attributes.html#updating-write-only-attributes)", google.Underscore(originalField.Description))
+	fieldPathOriginalField := originalFieldLineage
+	fieldPathCurrentField := strings.ReplaceAll(originalFieldLineage, google.Underscore(originalField.Name), google.Underscore(name))
+
+	apiName := originalField.ApiName
+	if apiName == "" {
+		apiName = originalField.Name
+	}
+
+	options := []func(*Type){
```
melinath, line 748: It looks like we need to specify RequiredWith for the write only field, too - without it, I'm able to specify the write only field without the version field, which results in unexpected behavior (permadiff where the original field has a value set and Terraform keeps trying to unset it)

I guess this works opposite of how I expected it!

mmv1/api/resource.go
```diff
@@ -730,14 +730,90 @@ func (r Resource) GetIdentity() []*Type {
...
+		propertyWithType("String"),
+		propertyWithRequired(false),
+		propertyWithDescription(description),
+		propertyWithWriteOnly(true),
+		propertyWithApiName(apiName),
+		propertyWithIgnoreRead(true),
+	}
+
+	if originalField.Required {
+		exactlyOneOf := append(originalField.ExactlyOneOf, fieldPathOriginalField, fieldPathCurrentField)
```
melinath, line 759: If the original field already has an exactlyoneof, it should already be included there, so we only need to add the current field.

```suggestion
		exactlyOneOf := append(originalField.ExactlyOneOf, fieldPathCurrentField)
```

mmv1/api/resource.go
```diff
@@ -730,14 +730,90 @@ func (r Resource) GetIdentity() []*Type {
...
+		options = append(options, propertyWithExactlyOneOf(exactlyOneOf))
+	} else {
+		conflicts := append(originalField.Conflicts, fieldPathOriginalField)
+		options = append(options, propertyWithConflicts(conflicts))
+	}
+
+	if len(originalField.AtLeastOneOf) > 0 {
+		atLeastOneOf := append(originalField.AtLeastOneOf, fieldPathOriginalField, fieldPathCurrentField)
```
melinath: same here

```suggestion
		atLeastOneOf := append(originalField.AtLeastOneOf, fieldPathCurrentField)
```

melinath (changes requested):
This is looking really great! I think the test failure is unrelated to your changes, but we can find out on the next test run.

ramonvermeulen:
> This is looking really great! I think the test failure is unrelated to your changes, but we can find out on the next test run.

Thanks again for reviewing! Will go over the comments this Sunday or Monday.

melinath (changes requested):
removing from my queue

mmv1/api/resource.go
ramonvermeulen: Are you completely sure about this one? Becuase this means that in the yaml configuration you need to set an `exactly_one_of` on a generated field which seems a bit unlogical to me (e.g. `exactly_one_of: ["foo.bar.0.password_wo"]` on the password field spec).

By doing it the current way, when `required: true` is set, it will ensure that exactly_one_of is set in such a way that it ensures that the original field OR the wo field OR any other field set on `exactly_one_of` in the original field are required.

I'm happy to change this, just want to double check this before I do.

mmv1/api/resource.go
ramonvermeulen: ^ See comment https://github.com/GoogleCloudPlatform/magic-modules/pull/14230/files#r2250639132

mmv1/api/resource.go
ramonvermeulen: I see, so it works in one direction. E.g. when you set the version you are REQUIRED to also set the `_wo`. However, if you only set the `_wo` you are NOT REQUIRED to set the version.

mmv1/products/monitoring/UptimeCheckConfig.yaml
ramonvermeulen: Nice, thanks for testing this as well! :)

mmv1/api/resource.go
ramonvermeulen: Applied this in https://github.com/GoogleCloudPlatform/magic-modules/pull/14230/commits/28e36ec354b7ceac38ec6397fa881235f581a3b9, please let me know if it needs to be reverted.

ramonvermeulen:
> removing from my queue

Thanks, applied all your feedback in https://github.com/GoogleCloudPlatform/magic-modules/pull/14230/commits/d8288aa5e1b7fed2b07c5d794347fc0c1c1f37d0 and https://github.com/GoogleCloudPlatform/magic-modules/pull/14230/commits/28e36ec354b7ceac38ec6397fa881235f581a3b9, I left two comments open because I wan't 100% sure if we should change this.

mmv1/api/resource.go
melinath: hmm, no, I was getting confused because of the `append` - you're right, if it's just `required` with no existing `exactly_one_of` constraint, we need to make sure both fields are present. Thanks!

I'm not sure whether it will cause problems to extend an existing exactly_one_of like this or not... but also, in that case `required` might not be set at all, so this branch wouldn't even activate.

Let's just leave this as-is for now.

mmv1/api/resource.go
melinath: This one should be changed - since you're checking whether `at_least_one_of` is set, we can trust that it will already contain the original field.

ramonvermeulen: I agree, thought it was a similar case as the other one. But it is indeed not, thanks for pointing out, will update PR in a bit.

ramonvermeulen:
> responded to the questions :-)

Thanks!

Once this is merged I want to change these two towards the new `wo` generation, which I think is directly a good check to find out if the set-up works well for newly added write-only fields:

https://github.com/GoogleCloudPlatform/magic-modules/pull/14138
https://github.com/GoogleCloudPlatform/magic-modules/pull/13847

And after that I would like to work on https://github.com/hashicorp/terraform-provider-google/issues/22614 to make it easier to implement ephemeral resources based upon a data source.

BBBmau:
> * Field `secret_data_wo_version` default value changed from `0` to `<nil>` on `google_secret_manager_secret_version` - [reference](https://googlecloudplatform.github.io/magic-modules/breaking-changes/breaking-changes#field-changing-default-value)

This is fine since we only ever want to set the version of a write-only when we are using the write-only field (setting it to "1" when setting the field for the first time)

ramonvermeulen:
> > * Field `secret_data_wo_version` default value changed from `0` to `<nil>` on `google_secret_manager_secret_version` - [reference](https://googlecloudplatform.github.io/magic-modules/breaking-changes/breaking-changes#field-changing-default-value)
> 
> This is fine since we only ever want to set the version of a write-only when we are using the write-only field (setting it to "1" when setting the field for the first time)

Check, I indeed removed the default value for `_wo_version` because it is `required_with` the `_wo` field anyways (both ways), requiring to set it explicitly.

melinath (approved):
The generated code seems functionally equivalent / backwards-compatible to me with the caveats we've noted above. Specifically:

- Switching field type from int to string seems to be backwards-compatible.
- Changing ConflictsWith + AtLeastOneOf to ExactlyOneOf should be identical in terms of functionality.
- Changing ForceNew true -> false is backwards-compatible
- We are adding RequiredWith in a couple places, which is technically breaking, but it's necessary to ensure that the write only version is set properly.

melinath:
The failing test seems unrelated to this change to me.

melinath:
rerunning to see if tests will pass - /gcbrun

melinath:
@BBBmau could you take a look at this PR and see if it looks good to you?

melinath:
opened https://github.com/hashicorp/terraform-provider-google/issues/23899 to track an issue I noticed while taking a look at this. Not a blocker for this PR.

ramonvermeulen:
> rerunning to see if tests will pass - /gcbrun

Thanks! I see again a fail during REPLAYING mode, is this related to the PR?

BBBmau:
> @BBBmau could you take a look at this PR and see if it looks good to you?

I did a review of the PR earlier this week and all looks good to me! Thanks for assisting in the reviews of the generation code on this. @melinath

melinath:
Since the failures on the build seem persistent, and aren't happening in our nightly tests, I'll take a closer look at that prior to merge.

melinath:
I'm not able to reproduce the failure locally - trying an ad hoc run in TeamCity. https://hashicorp.teamcity.com/buildConfiguration/TerraformProviders_GoogleCloud_GOOGLE_BETA_MMUPSTREAMTESTS_GOOGLEBETA_PACKAGE_SECRETMANAGER/434592

melinath:
The secret manager tests all passed in the TeamCity run. Looking at where the panic is happening - the problem is that `flattenSecretManagerSecretVersionPayload` sometimes returns an `error`, but that's not a valid return value - the calling code expects a `[]interface{}`. Basically if there's an API error reading the secret value, this will always panic. Probably the logic should be in a custom decoder instead (because that can return an error meaningfully).

This behavior isn't the result of this PR or changed by it, so I'm going to go ahead with the merge.

melinath:
TGC test failures also look unrelated to this PR

ramonvermeulen:
> The secret manager tests all passed in the TeamCity run. Looking at where the panic is happening - the problem is that `flattenSecretManagerSecretVersionPayload` sometimes returns an `error`, but that's not a valid return value - the calling code expects a `[]interface{}`. Basically if there's an API error reading the secret value, this will always panic. Probably the logic should be in a custom decoder instead (because that can return an error meaningfully).
> 
> This behavior isn't the result of this PR or changed by it, so I'm going to go ahead with the merge.

Thanks for your effort on doing the investigation on this! 

With the PR merged now I will put some effort in open PRs for write-only functionality after the weekend.

melinath:
Opened a few issues to track things I noticed while reviewing this PR:

- https://github.com/hashicorp/terraform-provider-google/issues/23923
- https://github.com/hashicorp/terraform-provider-google/issues/23924
- https://github.com/hashicorp/terraform-provider-google/issues/23925

PR #14231, opened 2025-06-10
disk sweeper pagination

shuyama1:

```release-note:none

```

### DIFF
--- a/mmv1/third_party/terraform/services/compute/resource_compute_disk_sweeper.go
+++ b/mmv1/third_party/terraform/services/compute/resource_compute_disk_sweeper.go
@@ -33,64 +33,73 @@ func testSweepDisk(region string) error {
 	zones := []string{"us-central1-a", "us-central1-b", "us-central1-c", "us-central1-f", "us-east1-b", "us-east1-c", "us-east1-d", "us-west1-a", "us-west1-b", "us-west1-c"}
 	for _, zone := range zones {
 		servicesUrl := "https://compute.googleapis.com/compute/v1/projects/" + config.Project + "/zones/" + zone + "/disks"
-		res, err := transport_tpg.SendRequest(transport_tpg.SendRequestOptions{
-			Config:    config,
-			Method:    "GET",
-			Project:   config.Project,
-			RawURL:    servicesUrl,
-			UserAgent: config.UserAgent,
-		})
-		if err != nil {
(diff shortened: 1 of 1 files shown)


mmv1/third_party/terraform/services/compute/resource_compute_disk_sweeper.go
```diff
@@ -33,64 +33,73 @@ func testSweepDisk(region string) error {
...
-			return nil
-		}
-
-		resourceList, ok := res["items"]
-		if !ok {
-			log.Printf("[INFO][SWEEPER_LOG] Nothing found in response.")
-			return nil
-		}
+		listUrl := servicesUrl
+		for {
```
shuyama1: The diff is bit hard to read, but L36-37 and L99-102 are the only changes

rileykarson: GH actually has a button to hide whitespace changes (or you can append `?w=1` directly to the URL), which reduces this to +- 13!

mmv1/third_party/terraform/services/compute/resource_compute_disk_sweeper.go
```diff
@@ -33,64 +33,73 @@ func testSweepDisk(region string) error {
...
-			log.Printf("[INFO][SWEEPER_LOG] Error in response from request %s: %s", servicesUrl, err)
-			return nil
-		}
-
-		resourceList, ok := res["items"]
-		if !ok {
-			log.Printf("[INFO][SWEEPER_LOG] Nothing found in response.")
-			return nil
-		}
+		listUrl := servicesUrl
```
rileykarson:
```suggestion
		// Page zero's URL is the raw list URL. Successive pages will return the token for the next page.
		pageUrl := servicesUrl
```

rileykarson (approved):
LGTM! One small nit on naming suggested, but non-blocking.

mmv1/third_party/terraform/services/compute/resource_compute_disk_sweeper.go
rileykarson: (note the variable name change, it doesn't get highlighted for some reason)

shuyama1: renamed the variable

PR #14234, opened 2025-06-10
remove ignoredPorjectServices in google_project_service

zli82016:
b/422413011

Remove the `ignoredPorjectServices`  validation 
1. `dataproc-control.googleapis.com` is enabled automatically when `dataproc.googleapis.com` is enabled, and need to be disabled separately.
2. `source.googleapis.com` has been deprecated
3. `stackdriverprovisioning.googleapis.com` can be enabled and disabled separately.

```release-note:enhancement
resourcemanager: allowed `dataproc-control.googleapis.com` and `stackdriverprovisioning.googleapis.com` services in `google_project_service` resource 
```

### DIFF
--- a/mmv1/third_party/terraform/services/resourcemanager/resource_google_project_service.go.tmpl
+++ b/mmv1/third_party/terraform/services/resourcemanager/resource_google_project_service.go.tmpl
@@ -21,11 +21,6 @@ import (
     "google.golang.org/api/serviceusage/v1"
 )
 
-// These services can only be enabled as a side-effect of enabling other services,
-// so don't bother storing them in the config or using them for diffing.
-var ignoredProjectServices = []string{"dataproc-control.googleapis.com", "source.googleapis.com", "stackdriverprovisioning.googleapis.com"}
-var ignoredProjectServicesSet = tpgresource.GolangSetFromStringSlice(ignoredProjectServices)
-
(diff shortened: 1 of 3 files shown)


mmv1/third_party/terraform/services/resourcemanager/resource_google_project_service.go.tmpl
```diff
@@ -65,9 +60,11 @@ var renamedServicesByOldAndNewServiceNames = tpgresource.MergeStringMaps(Renamed
 const maxServiceUsageBatchSize = 20
 
 func validateProjectServiceService(val interface{}, key string) (warns []string, errs []error) {
-	bannedServicesFunc := verify.StringNotInSlice(append(ignoredProjectServices, bannedProjectServices...), false)
-	warns, errs = bannedServicesFunc(val, key)
-	if len(errs) > 0 {
+	var errors []error
+	bannedServicesFunc := verify.StringNotInSlice(bannedProjectServices, false)
+	warns, errors = bannedServicesFunc(val, key)
```
rileykarson: I think we can continue to just return `errs` directly here? This will change the error message in a way that may be confusing to end users, since we didn't previously express this as "banned services" iirc.

zli82016: Changed it back.

PR #14236, opened 2025-06-11
[#23232] Add region url_parameter to google_vertex_ai_index_endpoint_…

LucaPrete:
Adds `region` url_parameter to `google_vertex_ai_index_endpoint_deployed_index`.

Reference [hashicorp/terraform-provider-google/issues/23232](https://github.com/hashicorp/terraform-provider-google/issues/23232)

```release-note:bug
vertexai: added missing region url_param in `google_vertex_ai_index_endpoint_deployed_index`
```

### DIFF
--- a/mmv1/products/vertexai/IndexEndpointDeployedIndex.yaml
+++ b/mmv1/products/vertexai/IndexEndpointDeployedIndex.yaml
@@ -108,6 +108,11 @@ parameters:
     immutable: true
     resource: 'IndexEndpoint'
     imports: 'name'
+  - name: 'region'
+    type: String
+    description: The region of the index endpoint deployment. eg us-central1
+    url_param_only: true
+    immutable: true
 properties:
   - name: 'name'
     type: String


mmv1/products/vertexai/IndexEndpointDeployedIndex.yaml
```diff
@@ -108,6 +108,11 @@ parameters:
     immutable: true
     resource: 'IndexEndpoint'
     imports: 'name'
+  - name: 'region'
```
hao-nan-li, line 111: Is this field required?

LucaPrete: Yes. I'll add it.

LucaPrete: Sorry @hao-nan-li since now there are still lots of modules around using this without region (but using instead the provider default), I'd keep it NOT mandatory. So the code should be good as you see here.

hao-nan-li (approved):

PR #14239, opened 2025-06-11
Fix TTL zero value handling in compute backend service

ruchikawa:
Add send_empty_value: true to TTL fields (defaultTtl, maxTtl, clientTtl) to ensure zero values are included in API requests rather than being treated as empty and omitted.

This resolves the issue where setting client_ttl=0 would cause the API to apply its default value of 3600, leading to validation errors when combined with lower max_ttl values.

Also adds test case to verify TTL zero value behavior works correctly.

Fixes: https://github.com/hashicorp/terraform-provider-google/issues/21632

```release-note:bug
compute: fixed `client_ttl`, `default_ttl`, and `max_ttl` fields in `google_compute_backend_service` and `google_compute_region_backend_service` not being sent to API when set to 0, which caused the API to apply default values and potentially trigger validation errors```

### DIFF
--- a/mmv1/products/compute/BackendService.yaml
+++ b/mmv1/products/compute/BackendService.yaml
@@ -683,16 +683,19 @@ properties:
           Specifies the default TTL for cached content served by this origin for responses
           that do not have an existing valid TTL (max-age or s-max-age).
         default_from_api: true
+        send_empty_value: true
       - name: 'maxTtl'
         type: Integer
         description: |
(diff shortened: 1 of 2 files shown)


ruchikawa:
Thank you for the review. I overlooked the concern you raised. I also confirmed that while the tests ran, they're now failing with errors. Since I can't access the detailed error logs, I'll investigate this again in my local environment. 
Please give me some time to confirm.

PR #13944, opened 2025-05-14
Add Terraform Support for WireGroups.

NandiniAgrawal15:
```release-note: new-resource 
`google_compute_wire_group`
```
The WireGroup resource represents a group of redundant wires between interconnects in two different metros.
---

# google_compute_wire_group

The WireGroup resource represents a group of redundant wires between interconnects in two different metros. Each WireGroup belongs to a CrossSiteNetwork. A wire group defines endpoints and the wires which exist between them.

~> **Warning:** This resource is in beta, and should be used with the terraform-provider-google-beta provider.
See [Provider Versions](https://terraform.io/docs/providers/google/guides/provider_versions.html) for more details on beta resources.

To get more information about WireGroup, see:

* [API documentation](https://cloud.google.com/compute/docs/reference/rest/beta/wireGroups)
* How-to Guides
    * [Create a WireGroup](https://cloud.google.com/network-connectivity/docs/interconnect/how-to/cross-site/modify-network#add-wire-group)

## Example Usage - Compute Wire Group Basic

```hcl
data "google_project" "project" {
provider = google-beta
}

resource "google_compute_cross_site_network" "example-cross-site-network" {
  name        = "tf-test-cross-site-network"
  description = "Example cross site network"
  provider    = google-beta
}

resource "google_compute_wire_group" "example-test-wire-group" {
  name               = "test-wire-group"
  description        = "Example Wire Group"
  cross_site_network = "test-cross-site-network"
  provider           = google-beta
  depends_on = [
    google_compute_cross_site_network.example-cross-site-network
  ]
  wire_properties {
    bandwidth_unmetered = 10
  }
  wire_group_properties {
    type = "WIRE"
  }
}
```

## Argument Reference

The following arguments are supported:

* `name` -
  (Required)
  Name of the resource. Provided by the client when the resource is created. The name must be
  1-63 characters long, and comply with RFC1035. Specifically, the name must be 1-63 characters
  long and match the regular expression `[a-z]([-a-z0-9]*[a-z0-9])?` which means the first
  character must be a lowercase letter, and all following characters must be a dash,
  lowercase letter, or digit, except the last character, which cannot be a dash.

* `cross_site_network` -
  (Required)
  Required cross site network to which wire group belongs.

- - -

* `description` -
  (Optional)
  An optional description of this resource. Provide this property when you create the resource.

* `endpoints` -
  (Optional)
  Endpoints grouped by location, each mapping to interconnect configurations.

* `admin_enabled` -
  (Optional)
[...]

PR #14241, opened 2025-06-11
Change the type of the endpoint

NandiniAgrawal15:
```release-note:bug
compute: fixed endpoint type  in `google_compute_wire_group`
```

This doesn't constitute a breaking change, since without this, users were unable to add an acceptable endpoint value themselves.

### DIFF
--- a/mmv1/products/compute/WireGroup.yaml
+++ b/mmv1/products/compute/WireGroup.yaml
@@ -87,23 +87,33 @@ properties:
     validation:
       regex: '^[a-z]([-a-z0-9]*[a-z0-9])?$'
   - name: endpoints
-    type: KeyValuePairs
+    type: Map
     description: |
       Endpoints grouped by location, each mapping to interconnect configurations.
-    properties:
-      - name: interconnects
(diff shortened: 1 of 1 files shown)


NandiniAgrawal15:
The previous change is yet to be released [https://github.com/GoogleCloudPlatform/magic-modules/pull/13944]. Is there some way we can exempt this from being labelled as breaking change ? If so, any directions would be helpful. Thanks !

hao-nan-li (commented):
Is this change trying to fix a problem that has not been released yet? In that case I don't think this is a breaking-change.

Could you also tag this PR to the previously merged one? I just need to verify a bit and see if it's safe to merge this in.

hao-nan-li:
@NickElliot FYI since you had context with https://github.com/GoogleCloudPlatform/magic-modules/pull/13944

NandiniAgrawal15:
The Resource `WireGroup` was added via https://github.com/GoogleCloudPlatform/magic-modules/pull/13944. This change is yet to be released.The field `Endpoint` should have a different type.

PR #14058, opened 2025-05-21
Don't suppress diff for Workbench Instance serial port logging metadata.

madhusuraj:
**Release Note Template for Downstream PRs (will be copied)**

See [Write release notes](https://googlecloudplatform.github.io/magic-modules/contribute/release-notes/) for guidance.

```release-note:bug
workbench: fixed `metadata` serial-port-logging-enable diff on `google_workbench_instance`.
```

PR #14243, opened 2025-06-11
Allow setting `serial-port-logging-enable` in google_workbench_instance metadata

bcreddy-gcp:

```release-note:bug
google_workbench_instance: serial-port-logging-enable` metadata key is now settable.
```

### DIFF
--- a/mmv1/templates/terraform/constants/workbench_instance.go.tmpl
+++ b/mmv1/templates/terraform/constants/workbench_instance.go.tmpl
@@ -287,14 +287,18 @@ func workbenchMetadataCustomizeDiff(_ context.Context, diff *schema.ResourceDiff
 			oldValue, oldOk := oldMetadata[key]
 			newValue, newOk := newMetadata[key]
 
-			// If the key exists in both old and new metadata and their values differ, or
-			// if the key exists in one but not the other, then a meaningful change has occurred.
+			// Condition to force new:
+			// 1. The key exists in both old and new metadata AND their values differ.
+			// 2. The key exists in new but not in old (meaning it was added).
+			//
+			// The key exists in old but not in new (meaning it was removed) is ignored.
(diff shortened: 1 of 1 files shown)


bcreddy-gcp:
Fixing the issue from
https://github.com/GoogleCloudPlatform/magic-modules/pull/13701#pullrequestreview-2820006519
https://github.com/GoogleCloudPlatform/magic-modules/pull/14058

cc: @shuyama1

SirGitsalot (changes requested):
LGTM, but with the diff suppress changes it'd be great to see an update test, maybe by adding `serial-port-logging-enable = "false"` [here or so](https://github.com/GoogleCloudPlatform/magic-modules/blob/645ff8c36870e134106a99ccfaf10ae38d9577aa/mmv1/third_party/terraform/services/workbench/resource_workbench_instance_test.go#L381)

bcreddy-gcp:
This metadata key is only settable but not updatable. This change makes the key now settable. That said when users update the key, apply goes through but the plan always leaves a diff. I don't see an easy way that prevents changing the key. Any ideas?

bcreddy-gcp:
I don't think that's an option here. We don't want the resource to be recreated anytime there is a metadata change. Is there a way to trigger this only when this key is changed?

SirGitsalot:
Sorry, I said "diff suppress" but meant "custom diff" (they're similar in basic purpose but the former is a field-level annotation and the latter is resource level and more powerful). It's the latter I'm suggesting - you might be able to check that the user is attempting to change that one entry and `ForceNew` (and/or `SetNewComputed`, since this is a `default_from_api` field).

Example: https://github.com/GoogleCloudPlatform/magic-modules/blob/f29644c592cdfdaece4725b7b7cda29012d13d2a/mmv1/templates/terraform/constants/secret_manager_secret.go.tmpl#L2

SirGitsalot (approved):

PR #14248, opened 2025-06-12
Added cipher_suite to VpnTunnel

Dawid212:
This PR adds cipher_suite with its all fields (phase1, phase1.encryption, phase1.integrity, phase1.prf, phase1.dh, phase2. encryption, phase2.integrity, phase2.pfs)

```release-note:enhancement
compute: added `cipher_suite` block with phase1 and phase2 encryption configurations to `google_compute_vpn_tunnel` resource.
```


ScottSuarez (approved):

Gorlami96:
@Dawid212 we are from the Google and we handle the Cloud VPN resource. This implementation has bug because cipher order should not be changed by terrraform after the user has specified it. In this implementation, since the fields are Sets, Terraform treats them as unordered maps and orders them randomly according to the set hash function. This breaks the contract with the user as we have mentioned that cipher ordering will not change.

We were curious as to what exactly prompted you to make this change?

PR #14249, opened 2025-06-12
Add Network Firewall Policy "Policy Type" field

jhamhader:
The policy type field is a new field, set on policy creation and is then immutable.
Used to define which resources the policy can be associated with (depending on the VPC type), and imposes limitations and validations on the rules in the policy.

The default policy type is VPC_POLICY. RDMA_ROCE_POLICY can be used with RoCE networks.

https://github.com/GoogleCloudPlatform/magic-modules/pull/14249

```release-note:enhancement
compute: added `policyType` to `google_compute_network_firewall_policy`, `google_compute_network_firewall_policy_with_rules`, `google_compute_region_network_firewall_policy`, `google_compute_region_network_firewall_policy_with_rules`.
```

### DIFF
--- a/mmv1/products/compute/NetworkFirewallPolicy.yaml
+++ b/mmv1/products/compute/NetworkFirewallPolicy.yaml
@@ -57,6 +57,16 @@ properties:
   - name: 'description'
     type: String
     description: An optional description of this resource. Provide this property when you create the resource.
+  - name: 'policyType'
+    type: Enum
+    immutable: true
+    description: |
+      Policy type is used to determine which resources (networks) the policy can be associated with.
+      A policy can be associated with a network only if the network has the matching policyType in its network profile.
+      Different policy types may support some of the Firewall Rules features.
+    min_version: 'beta'
+    enum_values:
+      - 'VPC_POLICY'
   - name: 'fingerprint'
     type: Fingerprint
     description: Fingerprint of the resource. This field is used internally during updates of this resource.
--- a/mmv1/products/compute/NetworkFirewallPolicy.yaml
+++ b/mmv1/products/compute/NetworkFirewallPolicy.yaml
@@ -65,6 +65,7 @@ properties:
       A policy can be associated with a network only if the network has the matching policyType in its network profile.
       Different policy types may support some of the Firewall Rules features.
     min_version: 'beta'
+    default_from_api: true
     enum_values:
       - 'VPC_POLICY'
   - name: 'fingerprint'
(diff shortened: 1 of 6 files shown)


mmv1/products/compute/NetworkFirewallPolicy.yaml
```diff
@@ -57,6 +57,16 @@ properties:
   - name: 'description'
     type: String
     description: An optional description of this resource. Provide this property when you create the resource.
+  - name: 'policyType'
```
mihhalj, line 60: I think we have to add also default_from_api: true attribute for this new field - otherwise if we backfill it to the existing resources terraform would report changes) - not sure how' default_from_api' works with 'immutable: true'

jhamhader: Done.

mmv1/products/compute/NetworkFirewallPolicyWithRules.yaml
```diff
@@ -76,6 +76,16 @@ properties:
   - name: 'description'
     type: String
     description: An optional description of this resource.
+  - name: 'policyType'
```
roaks3, line 79: We can see if the tests catch it, but per the comment above, should any of these other fields be `default_from_api`? If they don't have `default_from_api` and the server provides a default value, I think a diff will result, which is especially bad because the fields are immutable.

mmv1/products/compute/RegionNetworkFirewallPolicy.yaml
```diff
@@ -38,6 +38,10 @@ examples:
     primary_resource_id: 'policy'
     vars:
       policy_name: 'tf-test-policy'
+  - name: 'region_network_firewall_policy_roce'
+    primary_resource_id: 'policy'
+    vars:
+      policy_name: 'tf-test-policy'
```
roaks3: The `tf-test` prefix is actually added for you as long as your name includes a `-`. So for example, this could be changed to `rnf-policy`.

roaks3: Oh also, this may need to be `min_version: beta`

roaks3 (commented):
Largely LGTM, just want to make sure about the server-provided defaults

mmv1/products/compute/NetworkFirewallPolicyWithRules.yaml
jhamhader: Yes, updated, thank you.

mmv1/products/compute/RegionNetworkFirewallPolicy.yaml
jhamhader: Done, thanks.

roaks3 (approved):

PR #14250, opened 2025-06-12
Add create_ignore_already_exists to google_cloud_identity_group_membership

bjfish25:
Adding create_ignore_already_exists field to google_cloud_identity_group_membership resource

Part of https://github.com/hashicorp/terraform-provider-google/issues/19790

```release-note: enhancement
cloudidentity: added `create_ignore_already_exists` field to `google_cloud_identity_group_membership` resource
```

### DIFF
--- a/mmv1/templates/terraform/pre_create/cloud_identity_group_membership.go.tmpl
+++ b/mmv1/templates/terraform/pre_create/cloud_identity_group_membership.go.tmpl
@@ -22,7 +22,15 @@ if d.Get("create_ignore_already_exists").(bool) {
         Headers:   headers,
     })
     // Do normal create if membership does not exist
-    if err == nil {
+    
+    if err != nil {
+        gerr, ok := err.(*googleapi.Error)
+        notFound := ok && gerr.Code == 404
+        // If group membership is not found, we can proceed with the create operation.
+        if !notFound {
+            return fmt.Errorf("Error checking if GroupMembership exists: %s", err)
+        }
+    } else { 
         // Set computed resource properties from create API response so that they're available on the subsequent Read call.
         err = resourceCloudIdentityGroupMembershipPostCreateSetComputedFields(d, meta, res)
         if err != nil {


mmv1/products/cloudidentity/GroupMembership.yaml
```diff
@@ -37,6 +39,8 @@ timeouts:
   update_minutes: 20
   delete_minutes: 20
 custom_code:
+  extra_schema_entry: templates/terraform/extra_schema_entry/cloud_identity_group_membership.tmpl
```
slevenick: Instead of extra_schema_entry I believe you can use a virtual_field. See https://googlecloudplatform.github.io/magic-modules/develop/client-side-fields/

mmv1/products/cloudidentity/GroupMembership.yaml
```diff
@@ -37,6 +39,8 @@ timeouts:
...
+  custom_create: templates/terraform/custom_create/cloud_identity_group_membership.go.tmpl
```
slevenick: Instead of an entire custom_create I think this could be accomplished with a pre_create that does the following:

Make the GET request to check if the resource already exists
If no, continue with normal create
If yes, do any necessary setting of computed fields, id, name etc. 

This would make it easier in the future to add new fields to this resource, as the generated create functionality will continue to work as new fields are introduced

bjfish25: Using pre_create duplicates code, but the normal creation flow is more standardized. Should be updated

mmv1/templates/terraform/pre_create/cloud_identity_group_membership.go.tmpl
```diff
@@ -0,0 +1,61 @@
...
+    res, err := transport_tpg.SendRequest(transport_tpg.SendRequestOptions{
+        Config:    config,
+        Method:    "GET",
+        Project:   billingProject,
+        RawURL:    getUrl,
+        UserAgent: userAgent,
+        Headers:   headers,
+    })
+    // Do normal create if membership does not exist
+    if err == nil {
```
slevenick: We should check that err is actually a 404 not found error rather than anything else

bjfish25: I'm not sure what you mean. If `err == nil`, then group membership exists and set detail.

I have updated it to go forward with create if `gerr.Code == 404`, and return a failure otherwise if that is what you mean

slevenick: Yeah, checking that the specific error is a not found error. We could see other errors here (500 internal error, 429 rate limiting, etc.) and we don't want to always fall back to attempting to create the resource

bjfish25: Alright, that should be handled then with the latest commit. Such that if the error is not notFound, we return an error: https://github.com/GoogleCloudPlatform/magic-modules/pull/14250/files#diff-8d325e085eab083e981091f7bb4927b81ea3a321b5c920b669e14348befcf966R30

slevenick (approved):

PR #14259, opened 2025-06-12
Remove static check for connection creation

anjalisoni3655:
This will help in resolving one of the internal bug where static check of exactly_one_of is not needed with authType UNSPECIFIED.

release-note:bug
`google_integration_connectors_connection`

Fixes: https://github.com/hashicorp/terraform-provider-google/issues/23426


anjalisoni3655:
> can you resolve the conflicts on this PR? Also can you explain a bit more what this `UNSPECIFIED` type is coming from? To me this `exactly_one_of` is to prevent conflicts from occuring in auth_config, It's not clear where this unspecified type comes from that you're referring to.

Hello, apologies for the typo. So, the auth type can also be "AUTH_TYPE_UNSPECIFIED" For this case, the user won't be providing an auth object like user_password, oauth2ClientCredentials etc. 
But due to the above exactly_one_of check, the auth type "AUTH_TYPE_UNSPECIFIED" doesn't work. 
e.g;   auth_config{ auth_type = "AUTH_TYPE_UNSPECIFIED"} , the below error is thrown. So, to fix the issue, I am currently removing this check 
image

rileykarson:

NickElliot (commented):
hi it looks like this PR has no changes now

PR #14263, opened 2025-06-12
Fix the bug that sql instance can not enable mcp

xuebaoZ:

```release-note:bug
sql: fixed an error in updating `connection_pool_config` in `google_sql_database_instance`
```

### DIFF
--- a/mmv1/third_party/terraform/services/sql/resource_sql_database_instance.go.tmpl
+++ b/mmv1/third_party/terraform/services/sql/resource_sql_database_instance.go.tmpl
@@ -424,6 +424,7 @@ is set to true. Defaults to ZONAL.`,
 						"disk_size": {
 							Type:     schema.TypeInt,
 							Optional: true,
+							// Default is likely 10gb, but it is undocumented and may change.
 							Computed:    true,
 							Description: `The size of data disk, in GB. Size of a running instance cannot be reduced but can be increased. The minimum value is 10GB for PD_SSD, PD_HDD and 20GB for HYPERDISK_BALANCED.`,
 						},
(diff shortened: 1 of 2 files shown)


mmv1/third_party/terraform/services/sql/resource_sql_database_instance.go.tmpl
```diff
@@ -433,7 +434,7 @@ is set to true. Defaults to ZONAL.`,
 							Computed:         true,
 							ForceNew:         true,
 							DiffSuppressFunc: caseDiffDashSuppress,
-							Description:      `The type of supported data disk is tier dependent and can be PD_SSD or PD_HDD or HYPERDISK_BALANCED.`,
+							Description:      `The type of supported data disk is tier dependent and can be PD_SSD or PD_HDD or HyperDisk_Balanced `,
```
c2thorn: Where is it documented that it is lower cased? I can only find https://github.com/googleapis/google-api-go-client/blob/main/sqladmin/v1beta4/sqladmin-gen.go#L4642

mmv1/third_party/terraform/services/sql/resource_sql_database_instance.go.tmpl
```diff
@@ -2170,6 +2171,7 @@ func resourceSqlDatabaseInstanceUpdate(d *schema.ResourceData, meta interface{})
 	s := d.Get("settings")
 	instance = &sqladmin.DatabaseInstance{
 		Settings: expandSqlDatabaseInstanceSettings(desiredSetting.([]interface{}), databaseVersion),
+		DatabaseVersion: databaseVersion,
```
c2thorn: what is MCP and why does this enable it?

mmv1/third_party/terraform/services/sql/resource_sql_database_instance.go.tmpl
xuebaoZ: This is modified by mistake, I have updated it.

mmv1/third_party/terraform/services/sql/resource_sql_database_instance.go.tmpl
xuebaoZ: MCP is a feature for instance. When we enable MCP for an existed instance, it will fail because the code does not include the 'databaseVersion' in the payload which need to be checked. I have modified the code. Please check. Thanks.

c2thorn (approved):

PR #14272, opened 2025-06-13
updated required property

luckyswaminathan:
In reference to https://github.com/hashicorp/terraform-provider-google/issues/23080, where it was mentioned that for both google_compute_node_group
and google_compute_node_template, the name schema is in reality required by the API and there is an error if it doesn't exist, but on our end it is marked as optional. I updated both their .yaml configuration files to mark them as required.

```release-note:REPLACEME

```


roaks3:
Rerolling since shuya is out

roaks3:

PR #14282, opened 2025-06-16
Fix TestAccDataSourceGoogleGkeHubFeature_basic failing in VCR

roaks3:
Fixes https://github.com/hashicorp/terraform-provider-google/issues/22091

This test runs and passes in VCR, but nothing is recorded. I notcied it is the only test using `Providers:    acctest.TestAccProviders`, so I've updated that in hopes that it fixes the issue. I'm wondering if there is something happening with beta vs ga when we use the old config.

```release-note:none

```


roaks3:
Seems to resolve the issue.

zli82016 (approved):

PR #14284, opened 2025-06-17
tgc-revival: add IncludeInTGCNext field

zli82016:
1. Setting `IncludeInTGCNext` to true will add the resource to tgc new package. `include_in_tgc_next_DO_NOT_USE` is for our team to support resources, not ready for using by contributors.
2. If a field is not CAI asset, then add `is_missing_in_cai: true` to the field and ignore the field in the test.

```release-note:none

```

### DIFF
--- a/mmv1/provider/terraform_tgc_next.go
+++ b/mmv1/provider/terraform_tgc_next.go
@@ -37,7 +37,7 @@ import (
 type TerraformGoogleConversionNext struct {
 	ResourceCount int
 
-	ResourcesForVersion []map[string]string
+	ResourcesForVersion []ResourceIdentifier
 
 	TargetVersionName string
 
@@ -46,6 +46,12 @@ type TerraformGoogleConversionNext struct {
 	Product *api.Product
 
 	StartTime time.Time
+}
+
+type ResourceIdentifier struct {
+	ServiceName   string
+	TerraformName string
+	ResourceName  string
 }
 
 func NewTerraformGoogleConversionNext(product *api.Product, versionName string, startTime time.Time) TerraformGoogleConversionNext {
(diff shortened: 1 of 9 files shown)


mmv1/third_party/tgc_next/test/assert_test_files.go
```diff
@@ -66,11 +67,12 @@ func BidirectionalConversion(t *testing.T, ignoredFields []string) {
 }
 
 // Tests a single resource
-func testSingleResource(t *testing.T, testName string, testData ResourceTestData, tfDir string, ignoredFields []string, logger *zap.Logger) error {
+func testSingleResource(t *testing.T, testName string, testData ResourceTestData, tfDir string, ignoredFields []string, logger *zap.Logger, primaryResource bool) error {
 	resourceType := testData.ResourceType
+	var tfplan2caiNotSupported, cai2hclNotSupported bool
```
trodge: I'm finding the double negatives here confusing.

Could we make these variables called `tfplan2caiSupported` and `cai2hclSupported`?

That way we could also do `_, tfplan2caiSupported := tfplan2caiconverters.ConverterMap[resourceType]`

mmv1/third_party/tgc_next/test/assert_test_files.go
```diff
@@ -81,11 +83,24 @@ func testSingleResource(t *testing.T, testName string, testData ResourceTestData
...
+		if primaryResource {
+			return fmt.Errorf("conversion of the primary resource %s is not supported in tgc", testData.ResourceAddress)
+		} else {
+			log.Printf("Test for %s is skipped as the resource is not supported in tgc.", resourceType)
 			return nil
 		}
 	}
 
+	if tfplan2caiNotSupported || cai2hclNotSupported {
+		return fmt.Errorf("resource %s is not supported in both tfplan2cai and cai2hcl in tgc", resourceType)
```
trodge: This could be ambiguous, not sure if we mean "resource %s is not supported in either direction" or "resource %s is supported in one direction but not both"

mmv1/provider/terraform_tgc_next.go
```diff
@@ -34,6 +35,10 @@ import (
 // TerraformGoogleConversionNext is for both tfplan2cai and cai2hcl conversions
 // and copying other files, such as transport.go
 type TerraformGoogleConversionNext struct {
+	ResourceCount int
+
+	ResourcesForVersion []map[string]string
```
trodge: Does this need to be a map rather than a struct?

zli82016: Changed it to a struct

mmv1/third_party/tgc_next/test/assert_test_files.go
zli82016: right, changed

mmv1/third_party/tgc_next/test/assert_test_files.go
zli82016: Good point. Changed.

trodge (approved):

PR #14289, opened 2025-06-17
Adding effective_reserved_ip_range and fix for null value in redis datasource 

NA2047:
Fixes https://github.com/hashicorp/terraform-provider-google/issues/14154

```release-note:enhancement 
redis: added output-only `effective_reserved_ip_range` field in `google_redis_instance` resource to display the value of `reservedIpRange` from API
```

```release-note:bug 
redis: fixed the bug that the `reserved_ip_range` field is null in `google_redis_instance` data source
```

### DIFF
--- a/mmv1/products/redis/Instance.yaml
+++ b/mmv1/products/redis/Instance.yaml
@@ -445,6 +445,16 @@ properties:
     # but will be a subset of the range.
     ignore_read: true
     default_from_api: true
+  - name: 'effectiveReservedIpRange'
+    type: String
+    description: |
+      The CIDR range of internal addresses that are reserved for this
+      instance. If not provided, the service will choose an unused /29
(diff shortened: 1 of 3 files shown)


zli82016:
Thanks for the fix.

Can you add a check to the [test](https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/third_party/terraform/services/redis/data_source_redis_instance_test.go#L20) that the field `reserved_ip_range` is set in the data source?

Example: https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/third_party/terraform/services/accessapproval/data_source_access_approval_folder_service_account_test.go#L31

zli82016 (changes requested):
Add a check in the test

NA2047:
> Add a check in the test

Will do @zli82016  :)

zli82016 (approved):

PR #14290, opened 2025-06-17
Support adc impersonation in different universes

hao-nan-li:

```release-note:enhancement
provider: added support for adc impersonation in different universes
```

### DIFF
--- a/mmv1/third_party/terraform/transport/config.go.tmpl
+++ b/mmv1/third_party/terraform/transport/config.go.tmpl
@@ -1358,11 +1358,34 @@ func (c *Config) GetCredentials(clientScopes []string, initialCredentialsOnly bo
 	var creds *googleoauth.Credentials
 	var err error
 	if c.ImpersonateServiceAccount != "" && !initialCredentialsOnly {
-		opts := option.ImpersonateCredentials(c.ImpersonateServiceAccount, c.ImpersonateServiceAccountDelegates...)
-		creds, err = transport.Creds(context.TODO(), opts, option.WithScopes(clientScopes...))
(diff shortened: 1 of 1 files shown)


mmv1/third_party/terraform/transport/config.go.tmpl
```diff
@@ -1358,11 +1358,34 @@ func (c *Config) GetCredentials(clientScopes []string, initialCredentialsOnly bo
 	var creds *googleoauth.Credentials
 	var err error
 	if c.ImpersonateServiceAccount != "" && !initialCredentialsOnly {
-		opts := option.ImpersonateCredentials(c.ImpersonateServiceAccount, c.ImpersonateServiceAccountDelegates...)
-		creds, err = transport.Creds(context.TODO(), opts, option.WithScopes(clientScopes...))
+		jsonCreds, err := credentials.DetectDefault(&credentials.DetectOptions{
```
c2thorn: can we rename this to adcCreds or defaultCreds? There is no longer a direct JSON file so I think it will be less confusing.

c2thorn (changes requested):

mmv1/third_party/terraform/transport/config.go.tmpl
c2thorn: I guess there IS a JSON file, but we aren't supplying it like we did with CredentialsJSON option

hao-nan-li: Yeah, I think I removed this line because the variable content was undefined, since the content was not something required previously. I can rename it to `defaultCreds`

c2thorn (approved):

PR #14295, opened 2025-06-17
remove dependsOn field as it is not supported for workerpools and fix failing test.

paridhishah18:

```release-note:breaking-change
Field `template.containers.depends_on` within `resource google_cloud_run_v2_worker_pool` removed
```

### DIFF
--- a/mmv1/products/cloudrunv2/WorkerPool.yaml
+++ b/mmv1/products/cloudrunv2/WorkerPool.yaml
@@ -477,12 +477,6 @@ properties:
               type: String
               description: |-
                 Container's working directory. If not specified, the container runtime's default will be used, which might be configured in the container image.
-            - name: 'dependsOn'
-              type: Array
-              description: |-
-                Containers which should be started before this container. If specified the container will wait to start until all containers with the listed names are healthy.
-              item_type:
-                type: String
       - name: 'volumes'
         type: Array
         description: |-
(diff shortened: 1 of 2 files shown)


paridhishah18:
Additional context:

We need to remove this field and this field doesn't work for Worker pool resource. This fields requires some additional fields to be set and those field are not supported in WorkerPool resource. WorkerPools API is in "BETA".

paridhishah18:
> Removing a field is a breaking-change. You mentioned this field does not work... What does it mean exactly? Are all use cases expected to fail if this field is used?

Yes all cases would fail if this field is used. Recently we added api validation for worker pools resource which enforces setting `template.containers.startUpProbe` field when `template.containers.depends_on` is set. `template.containers.startUpProbe` is not supported in Worker pools resource. 

```Error: Error creating WorkerPool: googleapi: Error 400: spec.configuration.revision_template.spec.containers[0].depends_on: Dependent container 'hello-2' must have startup probe specified
        Details:
        [
          {
            "fieldViolations": [
              {
                "description": "Dependent container 'hello-2' must have startup probe specified",
                "field": "spec.configuration.revision_template.spec.containers[0].depends_on"
              }
            ]
          }
        ]
        
          with google_cloud_run_v2_worker_pool.default
```
          
error in the current test if depends_on field is used.

kminsu-google:
lgtm

hao-nan-li:
I think this PR belongs to a major release. The contribution window should be open now.

hao-nan-li:
Please refer to https://googlecloudplatform.github.io/magic-modules/breaking-changes/make-a-breaking-change/#make-a-breaking-change to make a breaking change

hao-nan-li (approved):

PR #14300, opened 2025-06-18
fix: (storagetransfer) transfer path can be changed to empty string

gurusai-voleti:
Fixes: https://github.com/hashicorp/terraform-provider-google/issues/13673

```release-note:bug
storagetransfer: fix to allow empty string for path
```

### DIFF
--- a/mmv1/third_party/terraform/services/storagetransfer/resource_storage_transfer_job.go.tmpl
+++ b/mmv1/third_party/terraform/services/storagetransfer/resource_storage_transfer_job.go.tmpl
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"log"
 	"reflect"
+	"regexp"
 	"strings"
 	"time"
 
@@ -698,9 +699,28 @@ func gcsDataSchema() *schema.Resource {
 				Computed:    true,
 				Type:        schema.TypeString,
 				Description: `Google Cloud Storage path in bucket to transfer`,
+				ValidateFunc: validatePath,
+				DiffSuppressFunc: func(k, old, new string, d *schema.ResourceData) bool {
+					old, new = strings.TrimSpace(old), strings.TrimSpace(new)
+					if strings.ToLower(old) == strings.ToLower(new) {
+						return true
+					}
+					return false
+				},
 			},
 		},
 	}
+}
+
+func validatePath(v interface{}, k string) (ws []string, errors []error) {
+	value := v.(string)
+	value = strings.TrimSpace(value)
+	// checks if path not started with "/"
+	regex, err := regexp.Compile("^/+")
+	if err == nil && len(value) > 0 && regex.Match([]byte(value)) {
+		errors = append(errors, fmt.Errorf("%q cannot start with /", k))
+	}
+	return
 }
 
 func awsS3DataSchema() *schema.Resource {
--- a/mmv1/third_party/terraform/services/storagetransfer/resource_storage_transfer_job_test.go
+++ b/mmv1/third_party/terraform/services/storagetransfer/resource_storage_transfer_job_test.go
@@ -4,6 +4,7 @@ package storagetransfer_test
 
 import (
 	"fmt"
+	"regexp"
 	"testing"
 
 	"github.com/hashicorp/terraform-plugin-testing/helper/resource"
@@ -568,6 +569,28 @@ func TestAccStorageTransferJob_hdfsSource(t *testing.T) {
 				ResourceName:      "google_storage_transfer_job.transfer_job",
 				ImportState:       true,
 				ImportStateVerify: true,
+			},
+		},
+	})
+}
+
+func TestAccStorageTransferJob_transferPathError(t *testing.T) {
+	t.Parallel()
+
+	testDataSourceBucketName := acctest.RandString(t, 10)
+	testDataSinkName := acctest.RandString(t, 10)
+	testTransferJobDescription := acctest.RandString(t, 10)
+	testTransferJobName := fmt.Sprintf("tf-test-transfer-job-%s", acctest.RandString(t, 10))
+	errorMessage := string('"') + "transfer_spec.0.gcs_data_sink.0.path" + string('"') + " cannot start with /"
+
+	acctest.VcrTest(t, resource.TestCase{
+		PreCheck:                 func() { acctest.AccTestPreCheck(t) },
+		ProtoV5ProviderFactories: acctest.ProtoV5ProviderFactories(t),
+		CheckDestroy:             testAccStorageTransferJobDestroyProducer(t),
+		Steps: []resource.TestStep{
+			{
+				Config:      testAccStorageTransferJob_transferJobPathError(envvar.GetTestProjectFromEnv(), testDataSourceBucketName, testDataSinkName, testTransferJobDescription, testTransferJobName),
+				ExpectError: regexp.MustCompile(errorMessage),
 			},
 		},
 	})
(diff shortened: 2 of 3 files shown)


kautikdk:
Hi @gurusai-voleti, what is the solution that you are proposing here?

ScottSuarez:
This appears to be a breaking change, see https://github.com/GoogleCloudPlatform/magic-modules/pull/14300#issuecomment-2984189491

gurusai-voleti:
> Hi @gurusai-voleti, what is the solution that you are proposing here?

firstly the field is missing validation, added validation to the path field as gcs object path cannot be started with "/" (reference here https://cloud.google.com/storage-transfer/docs/reference/rest/v1/GcsData) and field is an user provided and cannot be marked as computed so changed path to optional field only

as the customer reported issue they used "/" first and later realized to change to "" empty string
which they cannot as in update d.HasChanges will not detect empty string as a change and an empty string cannot be considered for computed field as per terraform sdk (error occurred with computed set to true **_A computed value with the empty string as the new value and a non-empty old value was found. Interpreting the empty string as "unset" to align with legacy behavior._**) so tf plan/apply will show no changes required, so removed computed for the field and as it is user provided no need of computed as well

as d.HasChanges and computed cannot consider empty string as a change or input, added DiffSuppressFunc if the old and new is same it will not show any difference in tf plan, if its different it will show difference in tf plan and path can be updated to "" empty string

gurusai-voleti:
> This appears to be a breaking change, see [#14300 (comment)](https://github.com/GoogleCloudPlatform/magic-modules/pull/14300#issuecomment-2984189491)

yes it's a breaking change, so created an PR on top 7.0 feature branch, Anything else needs to be taken care?

mmv1/third_party/terraform/services/storagetransfer/resource_storage_transfer_job_test.go
```diff
@@ -573,6 +574,28 @@ func TestAccStorageTransferJob_hdfsSource(t *testing.T) {
...
+	errorMessage := string('"') + "transfer_spec.0.gcs_data_sink.0.path" + string('"') + " cannot start with /"
+
+	acctest.VcrTest(t, resource.TestCase{
+		PreCheck:                 func() { acctest.AccTestPreCheck(t) },
+		ProtoV5ProviderFactories: acctest.ProtoV5ProviderFactories(t),
+		CheckDestroy:             testAccStorageTransferJobDestroyProducer(t),
+		Steps: []resource.TestStep{
+			{
+				Config:      testAccStorageTransferJob_transferJobPathError(envvar.GetTestProjectFromEnv(), testDataSourceBucketName, testDataSinkName, testTransferJobDescription, testTransferJobName),
+				ExpectError: regexp.MustCompile(errorMessage),
```
kautikdk: I don't think we need error validation in the acceptance tests. We should add unit test instead to validate the change which is not related to the API integratiion.

Reference: https://googlecloudplatform.github.io/magic-modules/test/test/#add-resource-tests

mmv1/third_party/terraform/services/storagetransfer/resource_storage_transfer_job_test.go
```diff
@@ -2399,3 +2422,159 @@ resource "google_storage_transfer_job" "transfer_job" {
...
+  
+    depends_on = [
+      google_storage_bucket_iam_member.data_source,
+      google_storage_bucket_iam_member.data_sink,
+    ]
+  }
+  `, project, dataSourceBucketName, project, dataSinkBucketName, project, testTransferJobName, transferJobDescription, project)
+}
+
+func testAccStorageTransferJob_transferJobGcsPath(project string, dataSourceBucketName string, dataSinkBucketName string, transferJobDescription string, testTransferJobName string, gcsPath string) string {
```
kautikdk, line 2429: Where is this being used?

mmv1/third_party/terraform/services/storagetransfer/resource_storage_transfer_job.go.tmpl
```diff
@@ -695,14 +696,32 @@ func gcsDataSchema() *schema.Resource {
 			},
 			"path": {
 				Optional:    true,
-				Computed:    true,
 				Type:        schema.TypeString,
 				Description: `Google Cloud Storage path in bucket to transfer`,
+				ValidateFunc: validatePath,
+				DiffSuppressFunc: func(k, old, new string, d *schema.ResourceData) bool {
```
kautikdk: If we are removing Computed behavior then do we need this DiffSupressFunc?

kautikdk:
If the field is currently Computed, do we know why it was marked as such? i.e Was there any change from the API side which mandated this field to get server side value as well if not provided by user?

mmv1/third_party/terraform/website/docs/guides/version_7_upgrade.html.markdown
```diff
@@ -102,8 +102,8 @@ Description of the change and how users should adjust their configuration (if ne
 
 ## Resources
 
-## Resource: `google_product_resource`
+## Resource: `google_storage_transfer_job`
 
-### Resource-level change example header
+### `path` does not allow strings starting with /
```
kautikdk, line 107: We should add full field name here including parent block names otherwise it will mislead users. Only specifying `path` will be interpreted as top level field within the `google_storage_transfer_job` resource.

Reference: https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/third_party/terraform/website/docs/guides/version_6_upgrade.html.markdown#advanced_datapath_observability_configrelay_mode-is-now-removed

gurusai-voleti:
> If the field is currently Computed, do we know why it was marked as such? i.e Was there any change from the API side which mandated this field to get server side value as well if not provided by user?

even if its computed or not, as its optional field  if API response has value, it can be set otherwise not

ScottSuarez:
> > This appears to be a breaking change, see [#14300 (comment)](https://github.com/GoogleCloudPlatform/magic-modules/pull/14300#issuecomment-2984189491)
> 
> yes it's a breaking change, so created an PR on top 7.0 feature branch, Anything else needs to be taken care?

oops, I didn't realize we opened for changes to the major release branch. Thanks for that !!

ScottSuarez:
Looks like merging main into this branch has added unintended changes here.

gurusai-voleti:
> Looks like merging main into this branch has added unintended changes here.

git rebase went wrong

ScottSuarez:
> > If the field is currently Computed, do we know why it was marked as such? i.e Was there any change from the API side which mandated this field to get server side value as well if not provided by user?
> 
> even if its computed or not, as its optional field if API response has value, it can be set otherwise not

Do we have understanding of when/in what scenario server sets a value for this field independent of user input? While yes we can switch to optional here, I am looking to see how disruptive this might be for the user.

If the user needs to deploy then define this field post deployment because the server returned some value this is not an optimal experience. In such a case we should likely explicitly set the value to empty string if valid to avoid two step apply. As long as such a change isn't very divergent from normal user flow.

mmv1/third_party/terraform/services/storagetransfer/resource_storage_transfer_job.go.tmpl
ScottSuarez: May still be relevant if API alters user input

gurusai-voleti:
this branch has been corrupted created a new PR here https://github.com/GoogleCloudPlatform/magic-modules/pull/14377, please check the new PR @kautikdk  @ScottSuarez please add yourself as a reviewer for new PR

mmv1/third_party/terraform/services/storagetransfer/resource_storage_transfer_job.go.tmpl
gurusai-voleti: no API doesn't alter user input, removed the diffsuppressfunc

mmv1/third_party/terraform/services/storagetransfer/resource_storage_transfer_job_test.go
gurusai-voleti: updated test case

mmv1/third_party/terraform/services/storagetransfer/resource_storage_transfer_job_test.go
gurusai-voleti: removed the validation test case

mmv1/third_party/terraform/website/docs/guides/version_7_upgrade.html.markdown
gurusai-voleti: updated description

gurusai-voleti:
> why it was marke



> > > If the field is currently Computed, do we know why it was marked as such? i.e Was there any change from the API side which mandated this field to get server side value as well if not provided by user?
> > 
> > 
> > even if its computed or not, as its optional field if API response has value, it can be set otherwise not
> 
> Do we have understanding of when/in what scenario server sets a value for this field independent of user input? While yes we can switch to optional here, I am looking to see how disruptive this might be for the user.
> 
> If the user needs to deploy then define this field post deployment because the server returned some value this is not an optimal experience. In such a case we should likely explicitly set the value to empty string if valid to avoid two step apply. As long as such a change isn't very divergent from normal user flow.

if the user doesn't provide any value, API doesn't set any value, it just reflects what user provided

PR #14311, opened 2025-06-18
Update Access Context Manager Test for user types

coder-221:

```release-note:none
```

I think this test is failing because the members in the access level are incorrect. The IAM principal prefix should match the type of the principal. The principal type in this case should be group not user since the principals are groups.

### DIFF
--- a/mmv1/third_party/terraform/services/accesscontextmanager/resource_access_context_manager_access_level_test.go.tmpl
+++ b/mmv1/third_party/terraform/services/accesscontextmanager/resource_access_context_manager_access_level_test.go.tmpl
@@ -218,7 +218,7 @@ resource "google_access_context_manager_access_level" "test-access" {
     combining_function = "AND"
     conditions {
       ip_subnetworks = ["192.0.4.0/24"]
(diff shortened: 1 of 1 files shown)


coder-221:
Removing the draft label. I think the failing test is unrelated and I see the test I'm updating is passing again.

zli82016:
The failed test `TestAccAccessContextManager__gcp_user_access_binding` is unrelated.

@coder-221, can you add the Github issue you are trying to fix to the `description` section of this PR? Thanks.

coder-221:
Thanks @zli82016 for confirming the failing test is unrelated. I don't think there's a public GitHub issue tracking this issue. Do you have a link to the CI failures and maybe we can add that if it would help?

zli82016:
> Thanks @zli82016 for confirming the failing test is unrelated. I don't think there's a public GitHub issue tracking this issue. Do you have a link to the CI failures and maybe we can add that if it would help?

@coder-221, thanks for the information. What issue/error are you trying to fix in this PR?

zli82016 (approved):

PR #14315, opened 2025-06-19
Adds new IP Filter access fields

kautikdk:
This PR adds two new boolean fields to the `google_storage_bucket.ip_filter` block,
1. allow_cross_org_vpcs
2. allow_all_service_agent_access

```release-note:enhancement
storage: added `allow_cross_org_vpcs` and `allow_all_service_agent_access` fields to `google_storage_bucket` resource.
```

### DIFF
--- a/mmv1/third_party/terraform/services/storage/resource_storage_bucket.go.tmpl
+++ b/mmv1/third_party/terraform/services/storage/resource_storage_bucket.go.tmpl
@@ -635,7 +635,7 @@ func ResourceStorageBucket() *schema.Resource {
 							Type:        schema.TypeBool,
 							Optional:    true,
 							Description: `Whether to allow cross-org VPCs in the bucket's IP filter configuration.`,
-              RequiredWith: []string{"ip_filter.0.vpc_network_sources"},
+							RequiredWith: []string{"ip_filter.0.vpc_network_sources"},
 						},
 						"allow_all_service_agent_access" : {
 							Type:        schema.TypeBool,
--- a/mmv1/third_party/terraform/services/storage/resource_storage_bucket.go.tmpl
+++ b/mmv1/third_party/terraform/services/storage/resource_storage_bucket.go.tmpl
@@ -631,6 +631,17 @@ func ResourceStorageBucket() *schema.Resource {
 								},
 							},
 						},
+						"allow_cross_org_vpcs" : {
+							Type:        schema.TypeBool,
+							Optional:    true,
+							Description: `Whether to allow cross-org VPCs in the bucket's IP filter configuration.`,
+              RequiredWith: []string{"ip_filter.0.vpc_network_sources"},
+						},
+						"allow_all_service_agent_access" : {
+							Type:        schema.TypeBool,
+							Optional:    true,
+							Description: `Whether to allow all service agents to access the bucket regardless of the IP filter configuration.`,
+						},
 					},
 				},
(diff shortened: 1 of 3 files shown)


kautikdk:
Hi @rileykarson, It appears that the recent API change in `ip_filter` block marked `allow_all_service_agent_access` required. This change caused this failure: https://github.com/hashicorp/terraform-provider-google/issues/23370. Discussing offline for further clarifications.

Reference: https://cloud.google.com/storage/docs/ip-filtering-overview, The feature is in Preview stage.

kautikdk:
Hi @rileykarson, I think we can start review of this PR.

mmv1/third_party/terraform/services/storage/resource_storage_bucket.go.tmpl
```diff
@@ -631,6 +631,17 @@ func ResourceStorageBucket() *schema.Resource {
 								},
 							},
 						},
+						"allow_cross_org_vpcs" : {
+							Type:        schema.TypeBool,
+							Optional:    true,
+							Description: `Whether to allow cross-org VPCs in the bucket's IP filter configuration.`,
+							RequiredWith: []string{"ip_filter.0.vpc_network_sources"},
```
rileykarson, line 638: I think we need to remove this- it needs to be set on both fields at once and it's backwards incompatible on an existing field. If this is valid and `ip_filter.0.vpc_network_sources` can be set independently, can you point to a test that confirms as much? I see at least one test modified, which could be due to backwards incompatibility.

kautikdk: Yes, So the field `allow_cross_org_vpcs` only takes effect with `ip_filter.0.vpc_network_sources`. If the VPC network source is not specified then the backend will just ignore the field regardless of it's value. With that behavior, it will be misleading to allow users specifying this field without VPC network source as we have to mitigate the unwanted diff which will will be the false positive for the behavior correctness. To handle this, I put `RequiredWith` validation with only `allow_cross_org_vpcs` field so that we can ensure plan time validation and it aligns with field usage as well.

Regarding test modification, I think I mistakenly added `allow_cross_org_vpcs` field in third test step. Removed that to showcase use-case of the new field and behavior validation.
One more thing related to the breaking change, So there is one breaking change from API side that the feature team is doing deliberately. `allow_all_service_agent_access` field is required when `mode` is set to Enabled. The change is coming from the API side so it impacts all the clients not only Terraform. From the Terraform side, we have to add this field support so that if any user is blocked on this issue, they can mitigate it by configuring the new field.

rileykarson: A user could set this field to `false` and the `RequiredWith` would then require them to set `ip_filter.0.vpc_network_sources` which would also be invalid, right? But admittedly a bit rarer than the case you're guarding against.

I think we'd want a custom validation (through a CustomizeDiff) or rely on an API error/diff to correct invalid specifications here. That could be added later and it'd be fine, so to move forward we'd just need to remove `RequiredWith`.

`allow_all_service_agent_access`: All reads as reasonable to me, thanks for the context!

rileykarson: Offline discussion; `false` also theoretically requires that value set. False and null are distinct.

PR #14320, opened 2025-06-20
feat: (storage) add flag to force empty content type

gurusai-voleti:
Fixes: https://github.com/hashicorp/terraform-provider-google/issues/17058

```release-note:enhancement
storage: added `force_empty_content_type` flag to force empty content type in resource `google_storage_bucket_object`
```


mmv1/third_party/terraform/services/storage/resource_storage_bucket_object.go
```diff
@@ -406,10 +417,11 @@ func resourceStorageBucketObjectUpdate(d *schema.ResourceData, meta interface{})
 	bucket := d.Get("bucket").(string)
 	name := d.Get("name").(string)
 
-	if d.HasChange("content") || d.HasChange("source_md5hash") || d.HasChange("detect_md5hash") {
+	if d.HasChange("content") || d.HasChange("source_md5hash") || d.HasChange("detect_md5hash") || d.HasChange("force_empty_content_type") {
```
kautikdk: Do we need to trigger a POST call when `force_empty_content_type` is changed? Why the field can not be just updated with PATCH call?

mmv1/third_party/terraform/services/storage/resource_storage_bucket_object.go
```diff
@@ -406,10 +417,11 @@ func resourceStorageBucketObjectUpdate(d *schema.ResourceData, meta interface{})
...
 		// The KMS key name are not able to be set on create :
 		// or you get error: Error uploading object test-maarc: googleapi: Error 400: Malformed Cloud KMS crypto key: projects/myproject/locations/myregion/keyRings/mykeyring/cryptoKeys/mykeyname/cryptoKeyVersions/1, invalid
 		d.Set("kms_key_name", nil)
+		d.Set("content_type", nil)
```
kautikdk: Why explicit nil for content_type?

mmv1/third_party/terraform/services/storage/resource_storage_bucket_object.go
```diff
@@ -379,7 +386,11 @@ func resourceStorageBucketObjectCreate(d *schema.ResourceData, meta interface{})
 
 	insertCall := objectsService.Insert(bucket, object)
 	insertCall.Name(name)
-	insertCall.Media(media)
+	if v, ok := d.GetOk("force_empty_content_type"); ok && v.(bool) {
```
kautikdk, line 391: What if users specify both `content_type` and `force_empty_content_type`? Should we put validation for not allowing users to specify both at the same time?

gurusai-voleti: seems correct will add conflictwith, if user provides both force_empty_content_type is not working, the content_type provided by user is preferred

mmv1/third_party/terraform/services/storage/resource_storage_bucket_object.go
gurusai-voleti: 1) create a object and initially force_empty_content_type is false
2) object content type is set with text/plain or octect stream
3) update force_empty_content_type= true
4)  now content type should be force set to empty from the client side

similarly in vice versa case from true to false, either user should provide the value (as we are thinking both cannot exist at same time) or content type should be calculated from client side

to minimize I am updating the field to ForceNew: true

mmv1/third_party/terraform/services/storage/resource_storage_bucket_object.go
gurusai-voleti: removing this

mmv1/third_party/terraform/services/storage/resource_storage_bucket_object.go
kautikdk: How the scenarios are relevant to the question? What I am asking is that the object metadata including content-type can be updated using PATCH call then why we have to override the object using POST call.
Anyways since `content_type` is immutable, we can do similar for the new field as well. We should create a bug ticket for removing ForceNew behavior from Object metadata fields if that is applicable.

kautikdk (approved):
Left one optional comment. Apart from that LGTM!

mmv1/third_party/terraform/services/storage/resource_storage_bucket_object.go
gurusai-voleti: we can update using a PATCH call, 
If the flag changed from false to true we can force send the contentType to empty string,

If the flag changed from true to false
content_type needs to be calculated but for updateCall we don't have such provision like in insert, either some action need to be done which allows backend to calculate the content_type

gurusai-voleti: I see content_type cannot be mutable, if the client doesn't provide any it will always be calculated in backend and same is set in state file until anything changes either the object or content_type (user provided)

PR #12859, opened 2025-01-24
Add google_dialogflow_conversation_profile for (https://github.com/hashicorp/terraform-provider-google/issues/12466)

Spheny1:
Add support for Dialogflow ConversationProfiles.

**Release Note Template for Downstream PRs (will be copied)**

```release-note:new-resource
`google_dialogflow_convesation_profile`
```

PR #14324, opened 2025-06-20
Add google_dialogflow_conversation_profile for (https://github.com/hashicorp/terraform-provider-google/issues/12466) #12859

Spheny1:
Add support for Dialogflow ConversationProfiles.

```release-note:new-resource
`google_dialogflow_convesation_profile`
```


mmv1/third_party/terraform/services/dialogflow/resource_dialogflow_conversation_profile_test.go
```diff
@@ -0,0 +1,755 @@
...
+						type = "FAQ"
+					}
+					query_config {
+						context_filter_settings {
+							drop_handoff_messages       = false
+							drop_ivr_messages           = false
+							drop_virtual_agent_messages = false
+						}
+						knowledge_base_query_source {
+							knowledge_bases = ["projects/${google_project.agent_project.id}/locations/global/knowledgeBases/MTIxNTkzNjM1NzY3NjY2NjA2MDk"]
```
c2thorn: Is this a hardcoded source? Is it available to all, and what is it exactly

mmv1/third_party/terraform/services/dialogflow/resource_dialogflow_conversation_profile_test.go
```diff
@@ -0,0 +1,755 @@
...
+			}
+		}
+		logging_config {
+			enable_stackdriver_logging = true
+		}
+		language_code = "en-US"
+	}
+	`, context)
+}
+func testAccDialogflowConversationProfile_dialogflowAgentKnowledgebase2(context map[string]interface{}) string {
```
c2thorn: this is a very large config and can be hard to spot the difference between this and the first. Can you comment what's intended to change between the configs?

mmv1/third_party/terraform/services/dialogflow/resource_dialogflow_conversation_profile_test.go
```diff
@@ -0,0 +1,755 @@
...
+						type = "SMART_REPLY"
+					}
+					query_config {
+						context_filter_settings {
+							drop_handoff_messages       = false
+							drop_ivr_messages           = false
+							drop_virtual_agent_messages = false
+						}
+						document_query_source {
+							documents = ["projects/ccai-shared-external/locations/global/knowledgeBases/smart_messaging_kb/documents/NzU1MDYzOTkxNzU0MjQwODE5Mg"]
```
c2thorn: please provide more info on these, and where they came from.

mmv1/third_party/terraform/services/dialogflow/resource_dialogflow_conversation_profile_test.go
```diff
@@ -0,0 +1,755 @@
...
+				ResourceName:            "google_dialogflow_conversation_profile.profile",
+				ImportState:             true,
+				ImportStateVerify:       true,
+				ImportStateVerifyIgnore: []string{"location"},
+			},
+		},
+	})
+}
+
+// TODO make sure to flip agent back
```
c2thorn: Is this addressed?

c2thorn (changes requested):
some initial questions before running the build

Spheny1:
@c2thorn Im not sure why the "Build" Step is failing and, can we pass the large pr step? Conversation profiles are pretty large resources and getting all of the tests to verify each field is acting as it should takes quite a few lines.

mmv1/third_party/terraform/services/dialogflow/resource_dialogflow_conversation_profile_test.go
Spheny1: Yes this is a hardcoded source. Knowledge Bases are not currently in the terraform google provider is there a better way to handle this?

c2thorn: where did you get this value from?

Spheny1: This is created manually

Spheny1: @c2thorn Is there a change required here due to manually created knowledge base?

mmv1/templates/terraform/examples/dialogflow_conversation_profile_basic.tf.tmpl
```diff
@@ -0,0 +1,19 @@
+resource "google_dialogflow_agent" "basic_agent" {
+  display_name = "example_agent"
+  default_language_code = "en-us"
+  time_zone = "America/New_York"
+}
+resource "google_dialogflow_conversation_profile" "{{$.PrimaryResourceId}}" {
+  display_name = "{{index $.Vars "profile_name"}}"
+  location = "global"
+  default_language_code = "en-US"
```
c2thorn: should this be `language_code` instead?

c2thorn (changes requested):

mmv1/third_party/terraform/services/dialogflow/resource_dialogflow_conversation_profile_test.go
c2thorn: These tests will need to run in an idempotent way in multiple of our test environments. Hardcoded resources need to be either universally available or we'll need to use bootstrap functions to ensure they exist everytime the test is run.

Do you have information on these specific values were created? Like do you own them specifically?

mmv1/third_party/terraform/services/dialogflow/resource_dialogflow_conversation_profile_test.go
```diff
@@ -0,0 +1,753 @@
...
+
+	resource "google_dialogflow_agent" "agent" {
+		project = "${google_project.agent_project.id}"
+		display_name = "tf-test-agent-%{random_suffix}"
+		default_language_code = "en-us"
+		time_zone = "America/New_York"
+		depends_on = [google_project_iam_member.agent_create]
+	}
+
+	resouce "google_pubsub_topic" "topic" {
```
ArtoriaRen: typo `resouce`. Please fix. Thanks

mmv1/third_party/terraform/services/dialogflow/resource_dialogflow_conversation_profile_test.go
```diff
@@ -0,0 +1,753 @@
...
+	}
+
+	resource "google_project_iam_member" "agent_create" {
+		project = "${google_project.agent_project.id}"
+		role    = "roles/dialogflow.admin"
+		member  = "serviceAccount:${google_service_account.dialogflow_service_account.email}"
+	}
+
+	resource "google_dialogflow_agent" "agent" {
+		project = "${google_project.agent_project.id}"
```
ArtoriaRen: Found this error from the [debug log](https://ff87e700822b2708a99fdf4667b24f52292e81fcac874a67197e0b4-apidata.googleusercontent.com/download/storage/v1/b/ci-vcr-logs/o/beta%2Frefs%2Fheads%2Fauto-pr-14324%2Fartifacts%2F9586085e-deca-4c66-a9de-fb273ebb4ff3%2Frecording%2FTestAccDialogflowConversationProfile_update_document.log?jk=AXbWWmmqn4_hVJF9A5K3aGtMUSpzvXxA5N09NKKpXvFQLvbiy0IywHdijpOi3d1ailu7eqZuCfjq8M3Zsgtdmde8V1OE8G_72HQ-OXqaUjqIPPnFooPqqr9eBbEAJWPEjLIjK01VXRnTF2BaROn8gs1c1PjmcAhdcWzJNO7jaDwUwNUJSsYxw32rVAeoacPu7YmgdZBXsUPcGY08KK6rAhqRm6SCriooREwR34d0GyM63lPiC4dEuHjaIpCW5G0tldjW39HVCLtDAhnlVLmlKvGG8oNvDfrEHmo2LeziDKrFySBWNjiLem7ayWPYDsGqONqA1myeNddP15wAp9hrWWLuxR8upG_I3EDiu2c4tQijERgUFMVj2n6YXE2Hd-qiqnKxmug1tVfyveHsdxO2hnNZ9qX063pxhlDZIpni2PgVexavubQGNZsnk6dAdYMBOK98nm68cv_V8GR0JqObrfmkD0DboHO5efwoy10WZ5mLgJxsXvj0XKpRPm4S8yyu0jAJXkVm4JEHvPhuvPJbxA_Qp03scEUGVRe5q7-2ObcH4w5ZHLmhf12b7bPs3tRwK5FCLhLPs8xwh8ud0_ddEzct8WHuBeW659jDza-cZGg9X-DuS_Yzij2sIz_-weX8OKujAoTSUa5EWauu8HW5nDanjutuYeyFSNE-qu-J7jBAcem_uwOrax-2touOer9IqLldn8ycNP3_hkypSOuV56QypwrAsBpAp96FRnFOPVi-C8h17wrOCp6uXHZRtdj4WoFTc8Xn8eGFLNm9L4lFaHeznQC4tfwJN3VLlfdfa-legOjMqyTzIpmvAv3KkrKuIW0nJsapOns8cXm1aBmxdk7XBb9A0LXbG_e-mIysrPbg6rNed8lmGpZpPABPO0yS3Gfs-P4EmrYI8qSW5_iUBp_dIdpkBhDN8iFeviV-s7FYX80Nls_8rB6WPruh3Ieeg9FRAdZR9kICss0XfFO6jnL0oIWULUcZ7PbmTlMD7F6--5NEEmhB_n070P18uRj6iw1AoEWxIXRJcyBKLeccTMwsudXsHv2EhWkqrv7JZ_RoXNfrlRFDQH-wv7KRhXc6DsqSM1Dv3aMB34Qp-M5chPLpqp9QQdq9o6exYjMlE0e-bCrE-jQsFlse4Cv2LTCqSH7Zn7j-a8c1d_MUDTcVlS9LH3cjzrKrXn-O4myTO0EpECsNOUx9xuFa2UyFpja5onF5PWGqxQ0veCNENJvPko2yoEsE1RPQOps8xBu6_GmMwxHHQwXWn0msxiqVUf46dIMQGjOqI4cuwm52CLXT4KKSgM_pyjRNLD049Ktm5aUh7inQy6cmVC6PSBg22eLN3557WgXE&isca=1): `The requested URL <code>/v2/projects/projects/tf-test-dialogflow-o9bbm2mnzy/agent?alt=json</code> was not found on this server.`  Problem is that `projects` showed up twice.


`projects/${google_project.agent_project.project_id}/... ` should work according to https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/third_party/terraform/services/dialogflow/resource_dialogflow_intent_test.go#L207

mmv1/third_party/terraform/services/dialogflow/resource_dialogflow_conversation_profile_test.go
```diff
@@ -0,0 +1,753 @@
...
+						type = "FAQ"
+					}
+					query_config {
+						context_filter_settings {
+							drop_handoff_messages       = false
+							drop_ivr_messages           = false
+							drop_virtual_agent_messages = false
+						}
+						knowledge_base_query_source {
+							knowledge_bases = ["projects/${google_project.agent_project.id}/locations/global/knowledgeBases/MTQ5NzI0MjQ0OTQwMTI5NTY2NzM"]
```
ArtoriaRen: Saw error `Error: Error creating ConversationProfile: googleapi: Error 400: Invalid resource name: 'projects/projects/tf-test-dialogflow-0kvg4kbd4x/locations/global/knowledgeBases/MTQ5NzI0MjQ0OTQwMTI5NTY2NzM'. Valid resource patterns are: 'projects/*/knowledgeBases/*'...` in the [debug log](https://fff48102d239757c03943bf88a28595e38a0b30e1e3958d6fd6abd5-apidata.googleusercontent.com/download/storage/v1/b/ci-vcr-logs/o/beta%2Frefs%2Fheads%2Fauto-pr-14324%2Fartifacts%2F9586085e-deca-4c66-a9de-fb273ebb4ff3%2Fbuild-log%2Frecording_build%2FTestAccDialogflowConversationProfile_update_knowledgebase_recording_test.log?jk=AXbWWmnmeAxLB2BXAgmoaeoZKCn83at25LowvyXX0-v9xTJFCBofXoPP1rmZzvlTg4n2LymBDR3Ms_AtM5ybvapRX887iVfhV5q9inDrT4C7fwRDi5oFE8E7APUgot9KD_GO0ZDnZ1xyw3m8m0U_u6WvSGicG9V0Ri5QNtQyjNy-9Kl-bMUAJDIV0lnVeJnT-bxkWg0EOlvevFTBblvphA3psTDM0P95ZO2blqxJzvpZPuSskiz9dZAaB1vgP98ECdUY83NYUOIDX5CDmDSqseaTxCEXoYM1A9DawKHQ88r6TQkM75X3Pf3XsRh3Vm6BDDyj62lxJom3NRBwo9qC2MVTzFr7GY2znDydGbXA7mdWW5ROeedDY--FmScDUnZDFjIRCbg5A7RBViOSFDJY-al46E8Nk5J_X60NSYoML0zblICXRZdg23-03w-GVdjiXtNCZVdOp8xzIp73Orfqh_BeYCZVMmT-UHlhNavlCKIWqe9uzKd1uT4YnqRv7Ba2kLwmNh-RS2nNT8BIrISL5g-IS5mZz7fzsKx0e5ZhhtP4bKYoF3IYGNTf1TJWJHK0eBTBS9Kgo2WPWk2fkXjv8MheJmfMdYh_PU55O-632AVUudCJGDovotd26rfyQ-qUwVFaQuWe1ZoOjDUZ4leFymRLJrC6Yp08DyN03TxY-w19fWdkKvlk-GyQ99uqYkDrStYMT40W49PEXnTmxy5UCDT-D9X7-vW1dT-LsMEnZwz-OS7XR9zL-7DwatL2qgfCIsh7retY9mRzrHx8Gz36hETT_EJ2SoPGbS_bn88HrT4WQYYY-uv9tIouaPGoYuue0_QVF-nKKhVqJDxdS2oQgkqX5_X5X1bq3vK6jJX58izZC3v3851Ta3GPv4PwdP7ioonijohWlVIwiCztnVNNhA1nWBYohfoyaYY_PRjtXDCb4VFgN4E1oEFy5q-QQcu1rID5PVlXnUKl9hwwJYdvCtsRu6-ThFF0TYBLspZYqJP01Ror4snz89QEYqqLzbVAwYLQ6Nhl0qPpu31R4S9TmhmHHlfkG0y3ia1mP9MDOdY6OF-_h_1r4dL9wRwTm6kc_AlRyelYgY3wWz2JRUl14RV2rYwwF4LHwBnNNz6Prw3zf1dXnHtycctxI6s69ZmjTcl2CYpcvIvcz0XvRqva65v1Tl3pPbgC6lZrHRo6L4M6NTBTcsMD05B_hGCSi84pR7FIxzz0NFvUSai2LrYgUbtC79oT_rv-GMiXFYzcIGmQ3KnO1IBB77iU6tgNXXrGbsLY6YpKUemcOfz72rUFrGi7W75r1Ijxdiacypm5ISdIYXA5rylHBFTbr-gWZJ82_gFRegChdQS9wDWf0QOFu-JIk_yYlTvWmXOqQJa-iwLA-h_JZosAv4toog&isca=1). `projects` appeared twice in the URL. 

`projects/${google_project.agent_project.project_id}/...`  should work  according to https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/third_party/terraform/services/dialogflow/resource_dialogflow_intent_test.go#L207

mmv1/templates/terraform/examples/dialogflow_conversation_profile_basic.tf.tmpl
ArtoriaRen: > should this be `language_code` instead?

Yeah, it seems so.


@Spheny1 Could you please run the test locally by following https://googlecloudplatform.github.io/magic-modules/test/run-tests/? This is a large resource and the `create` and `update` tests can hit  many errors. Including as many fields as possible in the `create` tests and testing the `create` test locally have worked well for me. The `TF_LOG=DEBUG` flag is also super helpful. You can find how to use it in the doc I linked. Thanks!

ArtoriaRen:
@Spheny1  Can you please add the missing fields mentioned in https://github.com/GoogleCloudPlatform/magic-modules/pull/14324#issuecomment-3053552348  to your tests? Thanks!

mmv1/third_party/terraform/services/dialogflow/resource_dialogflow_conversation_profile_test.go
c2thorn: Do you know if the value of the knowledge bases are actually verified by the API during the create request? If not, then we can maybe use placeholder values instead of a real value.

If the API does verify, then we may need to drop this field from tests.

Spheny1: It seems the API requires a valid knowledge base (With document query source as well) in order for the conversation profile to be created.

ArtoriaRen: Hi @Spheny1, the `knowledgeBaseQuerySource` and `documentQuerySource` fields are related to the [Agent Assist Article Suggestion](https://cloud.google.com/agent-assist/docs/article-suggestion) and [Agent Assist FAQ Assist](https://cloud.google.com/agent-assist/docs/faq) features, which are superseded by [Generative knowledge assist](https://cloud.google.com/agent-assist/docs/generative-knowledge-assist). So Google doesn't plan to add support for  `Agent Assist Article Suggestion` or `Agent Assist FAQ Assist` in Terraform. Can you please drop the `knowledgeBaseQuerySource` and `documentQuerySource` fields in your PR? Thanks.

ArtoriaRen:
Hi @Spheny1, could you please take a look at these two comments: 
1.  https://github.com/GoogleCloudPlatform/magic-modules/pull/14324#discussion_r2195773939 
2. https://github.com/GoogleCloudPlatform/magic-modules/pull/14324#issuecomment-3099290314

Spheny1:
@ArtoriaRen Im running into troubles getting the tests to complete without using my already standing project. Agent's need their own project due to the limitation of one Dialogflow ES agent per project and I am following a similar pattern to what already exists in entity or fulfillment resources but when creating the resources I am getting HTML 404 errors instead of any other meaningful errors. This seems to happen after a post request for the Pubsub topic. Pasting logs in this comment.
[log.txt](https://github.com/user-attachments/files/21373079/log.txt)

ArtoriaRen:
> @ArtoriaRen Im running into troubles getting the tests to complete without using my already standing project. Agent's need their own project due to the limitation of one Dialogflow ES agent per project and I am following a similar pattern to what already exists in entity or fulfillment resources but when creating the resources I am getting HTML 404 errors instead of any other meaningful errors. This seems to happen after a post request for the Pubsub topic. Pasting logs in this comment. [log.txt](https://github.com/user-attachments/files/21373079/log.txt)

Hi @Spheny1, in the log file you uploaded, I saw `projects` appeared twice in the request URL. 
```
PUT /v1/projects/projects/tf-test-dialogflow-r69l0o1dbq/topics/tf-test-topic-r69l0o1dbq?alt=json HTTP/1.1
Host: pubsub.googleapis.com
User-Agent: Terraform/1.8.5 (+https://www.terraform.io) Terraform-Plugin-SDK/2.36.0 terraform-provider-google/acc
Content-Length: 49
Accept-Encoding: gzip

{
 "labels": {
  "goog-terraform-provisioned": "true"
 }
}

2025/07/21 11:17:17 [DEBUG] Google API Response Details:
HTTP/2.0 404 Not Found
```
I suspect this is because you are using `project = "${google_project.agent_project.id}"` when creating the pubsub topic. Can you please try `project = "${google_project.agent_project.project_id}"` instead? There are many other occurences. Could you please replace all of them? Thanks.

mmv1/templates/terraform/custom_flatten/conversation_profile_stack_driver.go.tmpl
```diff
@@ -0,0 +1,22 @@
...
+	Unless required by applicable law or agreed to in writing, software
+	distributed under the License is distributed on an "AS IS" BASIS,
+	WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+	See the License for the specific language governing permissions and
+	limitations under the License.
+*/ -}}
+func flatten{{$.GetPrefix}}{{$.TitlelizeProperty}}(v interface{}, d *schema.ResourceData, config *transport_tpg.Config) interface{} {
+	transformed := make(map[string]interface{})
+	if v == nil {
+		transformed["enable_stackdriver_logging"] = false
```
c2thorn, line 16: This field being set to false from null is causing the test to fail.
These custom flatteners were added to the initial commit, what was the behavior for these nested fields without the custom flattener that made you add them?

mmv1/products/dialogflow/ConversationProfile.yaml
```diff
@@ -0,0 +1,668 @@
...
+      - type: Boolean
+        name: 'enableWordInfo'
+        description: |
+          If true, Dialogflow returns SpeechWordInfo in StreamingRecognitionResult with information about the recognized speech words.
+      - type: Boolean
+        name: 'useTimeoutBasedEndpointing'
+        description: |
+          Use timeout based endpointing, interpreting endpointer sensitivy as seconds of timeout value.
+  - type: String
+    name: 'languageCode'
```
c2thorn, line 601: this field showed up with a diff at the end of the test of
```
language_code     = "en-US" -> null
```
which indicates there is a default. Let's add `default_from_api: true` to this field.

c2thorn (changes requested):

mmv1/templates/terraform/custom_flatten/conversation_profile_stack_driver.go.tmpl
Spheny1: Without the custom flatten setting them to false manually would cause it to not appear on the object returned by the API.

c2thorn: Thank you for the information. I think this is better resolved as a diff suppression, as it is discrepancy between the provider and the API over essentially the same state.

I know we have the common diff suppress [tpgresource.EmptyOrUnsetBlockDiffSuppress](https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/third_party/terraform/tpgresource/common_diff_suppress.go#L36), but I am not sure if that would work since the values are specifically false, not empty. You can try that or create new diff suppress functions to check for the values equal to false. It should be somewhat similar to the EmptyOrUnsetBlockDiffSuppress.

c2thorn: Hmm, looks like other dialogflow resources worked around this by just ignoring read:
https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/products/dialogflowcx/Page.yaml#L1486

Probably fine to do the same here, as long as we leave a comment.

Spheny1:
Question regarding the endpoint. I realize the reason it was working locally and not in CICD (or at least I think) is due to my local using the v2beta1 endpoint instead of just v2 is there a way to change the enum values available, validations on fields, or custom flattening methods based on the version of the rest api being used?

c2thorn:
> Question regarding the endpoint. I realize the reason it was working locally and not in CICD (or at least I think) is due to my local using the v2beta1 endpoint instead of just v2 is there a way to change the enum values available, validations on fields, or custom flattening methods based on the version of the rest api being used?

[min_version](https://googlecloudplatform.github.io/magic-modules/reference/field/#min_version-beta) is how to set entire fields to be beta-only. 

For custom code, instead of switching entire templates based on the version, we use template-level if statements to guard code behind different versions like in: https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/templates/terraform/decoders/containeranalysis_attestation_field_name.go.tmpl#L13

For enum values, I would rather change the enum to a string to remove all validation so we don't have to worry about updating the versioned values down the line if the API changes.

Similar for validation, I would rather loosen the validation than try to manage 2 separate versions of it. Or, use custom code to validate and control the versioned logic in the code. The way to do this in YAML-only is cumbersome and would mean defining two entire versions of the field with different `exact_version` attributes which is not recommended.

c2thorn (changes requested):
gave some recommendations for versioning in https://github.com/GoogleCloudPlatform/magic-modules/pull/14324#issuecomment-3165297111

But also would like to call out my previous review with 
https://github.com/GoogleCloudPlatform/magic-modules/pull/14324#discussion_r2240341168

and https://github.com/GoogleCloudPlatform/magic-modules/pull/14324#discussion_r2245968658

Spheny1:
I want to keep iterating on this to get the tests to pass however whenever clicking on the build log I am receiving a 403 error Forbidden.

c2thorn:
looks like two fields: automated_agent_config.agent and logging_config.enable_stackdriver_logging
```
          # google_dialogflow_conversation_profile.basic_profile will be updated in-place
          ~ resource "google_dialogflow_conversation_profile" "basic_profile" {
...
              ~ automated_agent_config {
                  ~ agent       = "projects/<project-id>/locations/global/agent/environments/draft" -> "projects/<project-id>/agent/environments/draft"
                }
        
              - logging_config {
                  - enable_stackdriver_logging = false -> null
                }
        
                # (1 unchanged block hidden)
            }
```

c2thorn (commented):
I still believe we should make `enable_stackdriver_logging` ignore_read as suggested in https://github.com/GoogleCloudPlatform/magic-modules/pull/14324#discussion_r2245968658 which should allow us to get rid of the custom flatten for it.

mmv1/templates/terraform/examples/dialogflow_conversation_profile_basic.tf.tmpl
```diff
@@ -0,0 +1,18 @@
+resource "google_dialogflow_agent" "basic_agent" {
+  display_name = "example_agent"
+  default_language_code = "en-us"
+  time_zone = "America/New_York"
+}
+resource "google_dialogflow_conversation_profile" "{{$.PrimaryResourceId}}" {
+  display_name = "{{index $.Vars "profile_name"}}"
+  location = "global"
+  automated_agent_config {
+    agent = "projects/${google_dialogflow_agent.basic_agent.id}/agent/environments/draft"
```
c2thorn, line 10:
```suggestion
    agent = "projects/${google_dialogflow_agent.basic_agent.id}/locations/global/agent/environments/draft"
```

c2thorn:
@Spheny1 Thanks for your efforts in this tricky resource. We are close, I believe these last two changes will be all that's needed. If you are able to push these fixes we should be good to merge.

The cutoff for the 6.49 release will be tomorrow night (Pacific Time). This will be the last release before 7.0.0, so it would be worthwhile getting it in before then. I'd push these changes myself, but I think the deletion of the custom_flatten is not feasible to do from the Github PR UI, and I don't have permission to push changes directly against your fork.

If you read this message, would you give me permission to adopt the changes in this PR in my own separate pull request if needed to get this out by 6.49?

Spheny1:
> @Spheny1 Thanks for your efforts in this tricky resource. We are close, I believe these last two changes will be all that's needed. If you are able to push these fixes we should be good to merge.
> 
> The cutoff for the 6.49 release will be tomorrow night (Pacific Time). This will be the last release before 7.0.0, so it would be worthwhile getting it in before then. I'd push these changes myself, but I think the deletion of the custom_flatten is not feasible to do from the Github PR UI, and I don't have permission to push changes directly against your fork.
> 
> If you read this message, would you give me permission to adopt the changes in this PR in my own separate pull request if needed to get this out by 6.49?

Absolutely I am traveling at the moment so I could not guarantee getting this updated fixed and tested locally before tomorrow night.

c2thorn:
> Absolutely I am traveling at the moment so I could not guarantee getting this updated fixed and tested locally before tomorrow night.

got it, I'm starting a separate pull request with https://github.com/GoogleCloudPlatform/magic-modules/pull/14849.

Appreciate your efforts on getting this essentially to the finish line. I'll do what I can go get it in the release cut, but it should be close.

PR #14330, opened 2025-06-20
tgc-revival: support compute tgc autoscaler

zli82016:

```release-note:none

```

### DIFF
--- a/mmv1/products/compute/Autoscaler.yaml
+++ b/mmv1/products/compute/Autoscaler.yaml
@@ -42,6 +42,7 @@ async:
   result:
     resource_inside_response: false
 collection_url_key: 'items'
+include_in_tgc_next_DO_NOT_USE: true
 custom_code:
 sweeper:
   url_substitutions:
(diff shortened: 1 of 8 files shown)


mmv1/products/compute/Autoscaler.yaml
```diff
@@ -83,6 +84,8 @@ parameters:
     ignore_read: true
     default_from_api: true
     custom_expand: 'templates/terraform/custom_expand/resourceref_with_validation.go.tmpl'
+    custom_tgc_flatten: 'templates/tgc_next/custom_flatten/compute_auto_scaler.go.tmpl'
```
trodge, line 87: Since the `custom_expand` is a template used by many resources, can we use a similar approach in TGC to not need a separate template for each resource?

trodge (approved):
As discussed offline, we'll merge this for now and refactor later to a more generic method of handling `zone`, `region` and `location` fields when we have more resources covered.

PR #14336, opened 2025-06-23
Supporting update BP

niharika-98:

```release-note:enhancement
BACKUPDR: added support to update Backupplan resource
```

### DIFF
--- a/mmv1/products/backupdr/BackupPlan.yaml
+++ b/mmv1/products/backupdr/BackupPlan.yaml
@@ -15,7 +15,8 @@ name: 'BackupPlan'
 base_url: projects/{{project}}/locations/{{location}}/backupPlans
 create_url: projects/{{project}}/locations/{{location}}/backupPlans/?backup_plan_id={{backup_plan_id}}
 self_link: projects/{{project}}/locations/{{location}}/backupPlans/{{backup_plan_id}}
-immutable: true
+update_verb: 'PATCH'
+update_mask: true
 description: A backup plan defines when and how to back up a resource, including the backup's schedule, retention, and location.
 import_format:
   - 'projects/{{project}}/locations/{{location}}/backupPlans/{{backup_plan_id}}'
@@ -26,6 +27,7 @@ references:
 autogen_async: true
 timeouts:
   insert_minutes: 60
+  update_minutes: 60
   delete_minutes: 60
 examples:
   - name: 'backup_dr_backup_plan_simple'
--- a/mmv1/third_party/terraform/services/backupdr/resource_backup_dr_backup_plan_test.go
+++ b/mmv1/third_party/terraform/services/backupdr/resource_backup_dr_backup_plan_test.go
@@ -1,0 +1,212 @@
+package backupdr_test
+
+import (
+	"github.com/hashicorp/terraform-plugin-testing/helper/resource"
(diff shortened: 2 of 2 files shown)


mmv1/third_party/terraform/services/backupdr/resource_backup_dr_backup_plan_test.go
```diff
@@ -0,0 +1,212 @@
...
+	acctest.SkipIfVcr(t)
+
+	t.Parallel()
+
+	timeNow := time.Now().UTC()
+	referenceTime := time.Date(timeNow.Year(), timeNow.Month(), timeNow.Day(), 0, 0, 0, 0, time.UTC)
+
+	context := map[string]interface{}{
+		"project":        envvar.GetTestProjectFromEnv(),
+		"effective_time": referenceTime.Add(24 * time.Hour).Format(time.RFC3339),
```
melinath: This doesn't seem to be used - removing it would allow you to not skip VCR, which would be preferable.

As a side note, if the issue is that this is a future time, a workaround is to set a specific time much farther in the future - for example, +1 year on April 1st. Then you can still use VCR - it'll just need to be re-recorded once per year (which will happen automatically).

mmv1/third_party/terraform/services/backupdr/resource_backup_dr_backup_plan_test.go
```diff
@@ -0,0 +1,212 @@
...
+      days_of_month   = [15]
+      time_zone       = "UTC"
+
+      backup_window {
+        start_hour_of_day = 2  # Backup starts at 2:00 AM UTC
+        end_hour_of_day   = 8  # Optional, backup window ends at 3:00 AM
+      }
+    }
+  }
+}
```
melinath: Please make sure that all updatable fields get new values.

mmv1/products/backupdr/BackupPlan.yaml
```diff
@@ -15,7 +15,8 @@ name: 'BackupPlan'
 base_url: projects/{{project}}/locations/{{location}}/backupPlans
 create_url: projects/{{project}}/locations/{{location}}/backupPlans/?backup_plan_id={{backup_plan_id}}
 self_link: projects/{{project}}/locations/{{location}}/backupPlans/{{backup_plan_id}}
-immutable: true
```
melinath, line 18: Note: Since this resource was previously immutable at the resource level, you'll need to go through and add `immutable: true` to all immutable fields on the resource - at a minimum, that should probably include location and backup plan id.

mmv1/third_party/terraform/services/backupdr/resource_backup_dr_backup_plan_test.go
niharika-98: Addressed comment

mmv1/products/backupdr/BackupPlan.yaml
niharika-98: Added immutable field

mmv1/third_party/terraform/services/backupdr/resource_backup_dr_backup_plan_test.go
niharika-98: Removed effective time as it is not used

PR #14339, opened 2025-06-23
Add SecureSourceManager UpdateRepository

ericpang777:

```release-note:enhancement
securesourcemanager: supported in-place update for `description` in `google_secure_source_manager_repository` resource
```

### DIFF
--- a/mmv1/products/securesourcemanager/Repository.yaml
+++ b/mmv1/products/securesourcemanager/Repository.yaml
@@ -98,7 +98,6 @@ properties:
     description: |
       The name of the instance in which the repository is hosted.
     required: true
-    immutable: true
     diff_suppress_func: 'tpgresource.ProjectNumberDiffSuppress'
   - name: 'uid'
     type: String
@@ -141,6 +140,7 @@ properties:
     description: |
       Initial configurations for the repository.
     ignore_read: true
+    immutable: true
     properties:
       - name: 'defaultBranch'
         type: String
--- a/mmv1/products/securesourcemanager/Repository.yaml
+++ b/mmv1/products/securesourcemanager/Repository.yaml
@@ -98,6 +98,7 @@ properties:
     description: |
       The name of the instance in which the repository is hosted.
     required: true
+    immutable: true
     diff_suppress_func: 'tpgresource.ProjectNumberDiffSuppress'
   - name: 'uid'
     type: String
--- a/mmv1/products/securesourcemanager/Repository.yaml
+++ b/mmv1/products/securesourcemanager/Repository.yaml
@@ -73,12 +73,14 @@ examples:
 parameters:
   - name: 'location'
     type: String
+    immutable: true
     description: |
       The location for the Repository.
     url_param_only: true
     required: true
   - name: 'repository_id'
     type: String
+    immutable: true
     description: |
       The ID for the Repository.
     url_param_only: true
@@ -145,10 +147,12 @@ properties:
     properties:
       - name: 'defaultBranch'
         type: String
+        immutable: true
         description: |
           Default branch name of the repository.
       - name: 'gitignores'
         type: Array
+        immutable: true
         description: |
           List of gitignore template names user can choose from.
           Valid values can be viewed at https://cloud.google.com/secure-source-manager/docs/reference/rest/v1/projects.locations.repositories#initialconfig.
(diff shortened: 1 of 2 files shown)


ericpang777:
Will check back next week, needs some time

mmv1/third_party/terraform/services/securesourcemanager/resource_secure_source_manager_repository_update_test.go
```diff
@@ -0,0 +1,100 @@
...
+      prevent_destroy = "%{prevent_destroy}"
+    }
+}
+
+resource "google_secure_source_manager_repository" "default" {
+    location = "us-central1"
+    repository_id = "tf-test-my-repository%{random_suffix}"
+    instance = google_secure_source_manager_instance.instance.name
+
+	description = "new description"
```
shuyama1: nit: we use spaces for indentation in resource configs.

ericpang777: Done

ericpang777:
> > Will check back next week, needs some time
> 
> Sure, take your time. I've added a small comment regarding spacing and also updated the changelog a bit. Please let me know if you need help debugging. Thanks!

Oh no worries, I am just waiting for the API visibility to lift for this, I think that is why the test is failing :)

shuyama1 (commented):
This comment is to disable the review-reminder, as the feature is pending API visibility.  Please comment or mark the PR as ready for review when the feature is available for testing, and I will re-run the tests.

ericpang777:
@shuyama1  Should be good now

shuyama1 (commented):
Thank you for the update! Could you please verify if all fields within `parameters` and `properties` support in-place updates? Any fields that are immutable will need the `immutable` flag at the field level.

Also, it'd be great to expand the test coverage to include more field update scenarios.

ericpang777:
Description is the only updatable field. The rest of the fields are output only or input only (`initialConfig` is input only, and I have it marked with `ignore_read`), do I still need to add it to those?

I think the only field left is `instance`. I will test it out what happens if I put a `immutable` tag on it

shuyama1 (commented):
> Description is the only updatable field. The rest of the fields are output only or input only (`initialConfig` is input only, and I have it marked with `ignore_read`), do I still need to add it to those?
> 
> I think the only field left is `instance`. I will test it out what happens if I put a `immutable` tag on it

Thanks for confirming! Yeah, for all non-output fields that are immutable, including input-only fields, you'll need to add the `immutable` flag. 

This flag adds `ForceNew` to the field in the generated Terraform resource. As a result, if the immutable field's value is changed in the configuration, Terraform will be forced to delete the existing resource and recreate a new one.

Input-only (or `ignore_read` flag) is just to read the value directly from the configuration into the Terraform state because the API doesn't return it. We still need to mark the input-only immutable fields as `immutable` to ensure that any values changes would trigger Terraform to correctly recreate the resource.

References in case it helps:
- https://googlecloudplatform.github.io/magic-modules/best-practices/immutable-fields/#immutable-fields
- https://googlecloudplatform.github.io/magic-modules/reference/field/#immutable
- https://developer.hashicorp.com/terraform/plugin/sdkv2/schemas/schema-behaviors#forcenew
- https://googlecloudplatform.github.io/magic-modules/develop/diffs/#ignore_read

ericpang777:
Thanks for the links! I've made both an immutable field now.

One question, would it be better to not add ForceNew behaviour to the `instance` field if I am wary of users changing the field and then Terraform attempts to replace it? Since we want to avoid accidental deletions if possible.

I am going to add deletion protection to this resource though, so maybe if I add that, it should cover most realistic cases? I guess https://googlecloudplatform.github.io/magic-modules/best-practices/immutable-fields/#immutable-fields basically describes the same scenario.

mmv1/products/securesourcemanager/Repository.yaml
```diff
@@ -139,6 +141,7 @@ properties:
     description: |
       Initial configurations for the repository.
     ignore_read: true
+    immutable: true
```
shuyama1, line 146: I assume all its subfields are immutable? If so, you'll need to add `immutable: true` to the subfields as well.

ericpang777: Done, thanks for the catch!

ericpang777: Added immutable to location and repo id too

shuyama1 (approved):
Thank you!

PR #14346, opened 2025-06-24
Add Labels field to StoragePool

Dawid212:
Add Labels field to StoragePool

```release-note:enhancement
compute: added `labels` field to `google_compute_storage_pool` resource
```


hao-nan-li (commented):
Could you add this field to an existing test?

mmv1/products/compute/StoragePool.yaml
```diff
@@ -277,6 +277,13 @@ properties:
       - "STANDARD"
       - "ADVANCED"
     default_from_api: true
+  - name: "labels"
+    type: KeyValueLabels
+    description: |
+      Labels to apply to this storage pool. These can be later modified by the setLabels method.
+    update_url: "projects/{{project}}/zones/{{zone}}/storagePools/{{name}}/setLabels"
+    update_verb: "POST"
+    default_from_api: true
```
hao-nan-li: Why is this field default from api?

mmv1/products/compute/StoragePool.yaml
```diff
@@ -277,6 +277,13 @@ properties:
       - "STANDARD"
       - "ADVANCED"
     default_from_api: true
+  - name: "labels"
+    type: KeyValueLabels
+    description: |
+      Labels to apply to this storage pool. These can be later modified by the setLabels method.
+    update_url: "projects/{{project}}/zones/{{zone}}/storagePools/{{name}}/setLabels"
```
hao-nan-li: What's the reason to add an update_url for this particular field? Is the update on this field different from updating the resource?

mmv1/products/compute/StoragePool.yaml
Dawid212: You are right. It doesn't need to be there.

mmv1/products/compute/StoragePool.yaml
Dawid212: You are right. It doesn't need to be there.

mmv1/products/compute/StoragePool.yaml
```diff
@@ -277,6 +277,11 @@ properties:
       - "STANDARD"
       - "ADVANCED"
     default_from_api: true
+  - name: "labels"
+    type: KeyValueLabels
+    description: |
+      Labels to apply to this storage pool. These can be later modified by the setLabels method.
+    update_verb: "POST"
```
hao-nan-li: I don't think this should be here if the update_url could be removed.

Dawid212: you are right. removed it

hao-nan-li (approved):

PR #14359, opened 2025-06-25
Make reassign-reviewer regex less strict

trodge:
The existing regex misses comments that contain the command string with anything else on the same line.

The new regex will match comments such as

```
On vacation next week, @modular-magician reassign review @trodge.

Could you take over this review?
```
```
@modular-magician reassign-review @trodge # (extra space)
```
or
```
@modular-magician reassign-review @hypothetical_enterprise_managed_user
```

```release-note:none

```

### DIFF
--- a/.github/workflows/reassign-reviewer.yml
+++ b/.github/workflows/reassign-reviewer.yml
@@ -22,7 +22,7 @@ jobs:
         uses: actions-ecosystem/action-regex-match@d50fd2e7a37d0e617aea3d7ada663bd56862b9cc # v2.0.2
         with:
           text: ${{ github.event.comment.body }}
-          regex: '(?:^|\n|\r)@modular-magician reassign-reviewer ?@?([a-zA-Z0-9-]+)?(?:$|\n|\r)'
+          regex: '.*@modular-magician reassign[- ]reviewer ?@?([a-zA-Z0-9-_]+)?.*'
       - name: Checkout Repository
         if: steps.read-comment.outputs.match != ''
         uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.2


trodge:

.github/workflows/reassign-reviewer.yml
```diff
@@ -22,7 +22,7 @@ jobs:
         uses: actions-ecosystem/action-regex-match@d50fd2e7a37d0e617aea3d7ada663bd56862b9cc # v2.0.2
         with:
           text: ${{ github.event.comment.body }}
-          regex: '(?:^|\n|\r)@modular-magician reassign-reviewer ?@?([a-zA-Z0-9-]+)?(?:$|\n|\r)'
+          regex: '.*@modular-magician reassign[- ]reviewer ?@?([a-zA-Z0-9-_]+)?.*'
```
ScottSuarez:
```suggestion
          regex: '.*@modular-magician reassign[- ]+reviewer *@?([a-zA-Z0-9-_]+).*'
```

trodge: I think this would make choosing a reviewer mandatory. The current behavior is to pick a random reviewer when none is specified (i.e. capture group is empty string).

trodge:
I've made it so that `@` is required when specifying a reviewer and not when reassigning to a random available reviewer.

I've also re-enabled caching for building the magician on the theory that the magician shouldn't have the same caching issues as the provider.

.github/workflows/reassign-reviewer.yml
```diff
@@ -22,7 +22,7 @@ jobs:
         uses: actions-ecosystem/action-regex-match@d50fd2e7a37d0e617aea3d7ada663bd56862b9cc # v2.0.2
         with:
           text: ${{ github.event.comment.body }}
-          regex: '.*@modular-magician reassign[- ]reviewer ?@?([a-zA-Z0-9-_]+)?.*'
+          regex: '.*@modular-magician reassign[- ]+review[^@]*@?([a-zA-Z0-9-_]+)?.*'
```
ScottSuarez: I think [^@] might be a little two permissive, we wouldn't want to include new lines in this capture.

trodge: I've made it exclude newlines and carriage returns. It's possible to make it allow omitting the @ but I think requiring @ is preferable anyway.

ScottSuarez (approved):

PR #14360, opened 2025-06-25
chore(ci): update frontmatter check to only cover website/docs folder

iyabchen:
Skip non website/docs filepath to avoid false positive alert for frontmatter format. 

```release-note:none

```


iyabchen:
Note that, the subcategory check seems to be duplicate of docscheck.sh in tpg/tpgb repo.  @shuyama1 should we just remove the subcategory check on our side?

iyabchen:
docscheck.sh purely use grep command to detect subcategory. If to combine the frontmatter format chat, then it needs move some go code there and apply the library. Not sure whether it worths to do so.

.ci/magician/cmd/generate_comment.go
```diff
@@ -685,6 +685,9 @@ func init() {
 func checkDocumentFrontmatter(repo source.Repo) []string {
 	var errs []string
 	for _, f := range repo.ChangedFiles {
+		if !strings.HasPrefix(f, "website/docs/r/") && !strings.HasPrefix(f, "website/docs/d/") {
```
shuyama1, line 688: Given we're removing the subcategory check, I think we could have the frontmatter check across all doc files in website/docs

shuyama1:
> docscheck.sh purely use grep command to detect subcategory. If to combine the frontmatter format chat, then it needs move some go code there and apply the library. Not sure whether it worths to do so.

Makes sense, let's keep the frontmatter chech in MM then. Thanks!

PR #14364, opened 2025-06-25
Add google_network_management_connectivity_test_run data source

rosmo:
Adds `google_network_management_connectivity_test_run` data source that allows to trigger/re-trigger connectivity checks. Useful when used in conjunction with Terraform's postcondition validations, for example:

```
data "google_network_management_connectivity_test_run" "test_run" {
  name     = google_network_management_connectivity_test.test.name

  lifecycle {
    postcondition {
      condition     = self.reachability_details[0].result == "REACHABLE"
      error_message = "Connectivity test reports connectivity broke."
    }
  }
 depends_on = [
    google_compute_firewall.rule
 ]
}

resource "google_network_management_connectivity_test" "test" {
# ...
}

resource "google_compute_firewall" "rule" {
# ...
}
```

```release-note:new-datasource
`google_network_management_connectivity_test_run`
```

### DIFF
--- a/mmv1/third_party/terraform/website/docs/d/network_management_connectivity_test_run.html.markdown
+++ b/mmv1/third_party/terraform/website/docs/d/network_management_connectivity_test_run.html.markdown
@@ -8,6 +8,9 @@ description: |-
 
 # google_network_management_connectivity_test_run
 
+
+!> This datasource triggers side effects on the target resource. It will take a long time to refresh (i.e. `terraform plan` will take much longer than usual) and may modify the state of the parent resource or other copies of the resource copying the same parent.
+
 A connectivity test is a static analysis of your resource configurations
 that enables you to evaluate connectivity to and from Google Cloud
 resources in your Virtual Private Cloud (VPC) network. This data source allows


mmv1/third_party/terraform/services/networkmanagement/data_source_network_management_connectivity_test_run_test.go
```diff
@@ -0,0 +1,113 @@
...
+						"reachability_details.0.result", regexp.MustCompile("REACHABLE")),
+				),
+			},
+		},
+	})
+}
+
+func testAccNetworkManagementConnectivityTestRun_instanceToInstance(context map[string]interface{}) string {
+	connTestCfg := acctest.Nprintf(`
+data "google_network_management_connectivity_test_run" "conn-test" {
```
rileykarson, line 37: Writing this as a datasource will mean that we create a new run on every `terraform apply` / `terraform plan`, and that could include as part of an automated drift detection system. This is user defined and we have no controls other than API quotas- so that could mean a system doing that every 10 minutes, creating and abandoning a run per `google_network_management_connectivity_test_run` instance!

On the other hand, a resource could create a single run during creation and then report on the run's results each `terraform apply`/`terraform plan`. A new run would be created with a new resource block.

A datasource seems reasonable given the domain (and GCP quotas / costs if any would provide controls on customers' request pressure) but I wanted to explicitly flag it and confirm!

rosmo: It's a fair point that you raise, and I would say just three things:
1) Generally most customer when they start a Terraform run (whether from cli or TFE), it takes a minute to start. Of course you could do this 1000x in parallel (both within the same TF code and just parallel runs). But it just doesn't seem likely and you could say the same for every data source.
2) If you roll back one commit on this PR, there is a resource implementation of this. On one hand, it's way easier for me (and you) since it's all autogenerated. But this is one of those things that really falls between the cracks. It's not really a data source (but it is in the sense that it will return data), but it's not also a resource (it doesn't really create anything new, there is no backlog of runs like you would see for example in Cloud Build),
3) Last but not the least, when you operate a "public" API, you have to be like a strict parent. The kids (the provider) should behave, but the parent (API endpoint) should have the last word :)

Ideally the best solution to me, that most reflects the nature of this, would be an ephemeral resource. But I think that's a bit off still and I'm not particularly looking to break new ground here (unless you feel strongly about it).

rileykarson: Ephemeral resources will behave like datasources anyways, from the perspective that they'll be evaluated on every plan (and the evaluation of this involves creating a test run + retrieving results). We're gonna wrap 'em all sometime in the near future so all datasources will be available as ephemeral resources: https://github.com/hashicorp/terraform-provider-google/issues/22614

Datasource is fine, just wanted to make sure I brought it up!

rosmo: Thanks!

rileykarson: Oh, this is just a "rerun" anyways. I thought it created a "run" resource, I misread it!

mmv1/third_party/terraform/website/docs/d/network_management_connectivity_test_run.html.markdown
```diff
@@ -0,0 +1,202 @@
...
+subcategory: "Network Management"
+description: |-
+  A connectivity test is a static analysis of your resource configurations
+  that enables you to evaluate connectivity to and from Google Cloud
+  resources in your Virtual Private Cloud (VPC) network.
+---
+
+# google_network_management_connectivity_test_run
+
+A connectivity test is a static analysis of your resource configurations
```
rileykarson, line 14:
```suggestion

!> This datasource triggers side effects on the target resource. It will take a long time to refresh (i.e. `terraform plan` will take much longer than usual) and may modify the state of the parent resource or other copies of the resource copying the same parent.

A connectivity test is a static analysis of your resource configurations
```

rileykarson (approved):
LGTM, thanks!

PR #14365, opened 2025-06-25
firestore_database: Remove the deprecation message for deletion_policy

rwhogg:
This is still deprecated, but this deprecation message caused more confusion than benefit. If you want your databases to be deleted upon running `terraform destroy`, you have no real choice other than to keep using this field (until we merge it with `delete_protection_state` in a future major release). The message should probably have never been included, because it wasn't truly actionable at the time for users with requirements to delete these databases.

```release-note:enhancement
firestore: revoked deprecation of `deletion_policy` field in `google_firestore_database` resource
```

### DIFF
--- a/mmv1/products/firestore/Database.yaml
+++ b/mmv1/products/firestore/Database.yaml
@@ -145,7 +145,9 @@ virtual_fields:
       See also `delete_protection`.
     type: String
     default_value: "ABANDON"
-    deprecation_message: '`deletion_policy` is deprecated and will be removed in a future major release. Use `delete_protection_state` instead.'
+    # `deletion_policy` is deprecated and will be removed in a future major release.
+    # Once that release happens, you should use `delete_protection_state` instead.
+    # For now though, setting this field is necessary if you wish for your Firestore databases to be deleted upon `terraform destroy`.
 parameters:
 properties:
   - name: 'name'


mmv1/products/firestore/Database.yaml
```diff
@@ -145,7 +145,9 @@ virtual_fields:
       See also `delete_protection`.
     type: String
     default_value: "ABANDON"
-    deprecation_message: '`deletion_policy` is deprecated and will be removed in a future major release. Use `delete_protection_state` instead.'
```
zli82016, line 148: deprecation_message: '`deletion_policy` is deprecated and will be removed in a future major release. Once that major release happens, use `delete_protection_state` instead. For now though, setting this field is necessary to delete the databases upon `terraform destroy`.'

zli82016: I think we can keep the deprecation_message and modify it a little bit.

bw-yuhogg: This change was specifically made in response to a support case with a customer objecting to the deprecation message's inclusion, so I'd personally rather move forward with getting rid of it.

zli82016: To remove the field in the next major release, the field needs to be deprecated first according to our team's [best practice](https://googlecloudplatform.github.io/magic-modules/breaking-changes/make-a-breaking-change/#add-deprecations-and-warnings-to-the-main-branch-of-magic-modules).

bw-yuhogg: They also say that the deprecation message had to be actionable, and there's not really anything users can do about it until we smoosh `deletion_policy` and `delete_protection_state` together.

zli82016: OK. It is fine to remove it after discussed with our team.

zli82016 (approved):

PR #14376, opened 2025-06-26
Fix JWKS JSON permadiff for workload identity pool provider

fujin:
## Summary
Fixes JWKS JSON permadiff issues for workload identity pool providers by implementing state_func normalization.

## Description
This PR addresses the permadiff issue where GCP normalizes JWKS JSON formatting (whitespace, key ordering) causing Terraform to detect false diffs on every plan/apply cycle.

### Solution
Uses \`state_func\` with \`structure.NormalizeJsonString()\` to normalize JSON values when stored in state. This approach:

- Follows the established pattern used in BigQuery and other Google provider resources
- Normalizes JSON on storage, eliminating the need for custom diff suppression
- Is cleaner and more maintainable than custom diff logic

### Implementation
- **Added**: \`state_func: 'func(v interface{}) string { s, _ := structure.NormalizeJsonString(v); return s }'\` to the jwks_json field
- **Pattern**: Mirrors the approach used in [BigQuery table schema normalization](https://github.com/hashicorp/terraform-provider-google-beta/blob/73b34d8a4ecfcdfde6e3a6db5aa2f1a55176326e/google-beta/services/bigquery/resource_bigquery_table.go#L580-L584)

### Benefits
- ✅ Eliminates permadiff for JWKS JSON formatting differences
- ✅ Handles whitespace variations and key ordering 
- ✅ Uses standard Terraform provider patterns
- ✅ No custom code maintenance burden
- ✅ Works consistently with other JSON fields in the provider

## Release Note Template
```release-note:bug
iambeta: fixed permadiff issue with `jwks_json` field in `google_iam_workload_identity_pool_provider` resource when GCP normalizes JSON formatting
```

### DIFF
--- a/mmv1/products/iambeta/WorkloadIdentityPoolProvider.yaml
+++ b/mmv1/products/iambeta/WorkloadIdentityPoolProvider.yaml
@@ -310,6 +310,7 @@ properties:
           }
           ```
         required: false
+        state_func: 'func(v interface{}) string { s, _ := structure.NormalizeJsonString(v); return s }'
   - name: 'saml'
     type: NestedObject
     description:


mmv1/products/iambeta/WorkloadIdentityPoolProvider.yaml
```diff
@@ -310,6 +310,7 @@ properties:
           }
           ```
         required: false
+        diff_suppress_func: 'jwksJsonDiffSuppressFunc'
```
ScottSuarez:
```suggestion
        state_func: 'func(v interface{}) string { s, _ := structure.NormalizeJsonString(v); return s }'
```

While this solution works I would recommend we just normalize the value we store in state similar to 
https://github.com/hashicorp/terraform-provider-google-beta/blob/73b34d8a4ecfcdfde6e3a6db5aa2f1a55176326e/google-beta/services/bigquery/resource_bigquery_table.go#L580-L584

fujin: oh very nice. is that all I need? I'll add your suggestion.

ScottSuarez: This should be all we need correct

fujin: In [7ef98a0](https://github.com/GoogleCloudPlatform/magic-modules/pull/14376/commits/7ef98a0d67c07f628fd453c00c7e18ffd3c6e509).

fujin:
woops, removed the changelog one sec

mmv1/third_party/terraform/services/iambeta/resource_iam_workload_identity_pool_provider_state_func_test.go
```diff
@@ -0,0 +1,127 @@
+package iambeta
+
+import (
+	"testing"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/structure"
+)
+
+func TestWorkloadIdentityPoolProviderJwksJsonStateFunc(t *testing.T) {
```
ScottSuarez: Having unit tests seems a bit overkill. We're just testing a native library function without terraform interaction. I think we can revert testing. If we insist on some form of testing I would prefer to have an integration test where we call the API

fujin: understood, will remove.

fujin: 893f06fa8

PR #14377, opened 2025-06-26
fix: (storagetransfer) path validation for GCS path source and sink

gurusai-voleti:
**Release Note Template for Downstream PRs (will be copied)**

See [Write release notes](https://googlecloudplatform.github.io/magic-modules/contribute/release-notes/) for guidance.
Fixes: https://github.com/hashicorp/terraform-provider-google/issues/13673
```release-note:bug
storagetransfer: path validation for GCS path source and sink
```

PR #14379, opened 2025-06-26
Resolved gcs_data_source path variable not updated when reset to empty in google_storage_transfer_job

malhotrasagar2212:
Resolved gcs_data_source path variable not updated when reset to empty in google_storage_transfer_job.

Removed Computed:true and added a DiffSuppressFunc for it.

```release-note:enhancement
removed computed:true and added a DiffSuppressFunc in gcsDataSchema() to resolve gcs_data_source path variable not updated when reset to empty in google_storage_transfer_job


```


kautikdk (commented):
This appears to be the duplicate PR of https://github.com/GoogleCloudPlatform/magic-modules/pull/14377 which is already submitted. We can probably close this PR!

PR #14381, opened 2025-06-26
Update VpcFlowLogs to support subnet and vpc on Project Level

maayanbeltzer:

```release-note: enhancement
networkmanagement: added `subnet` and `network` fields to the `google_network_management_vpc_flow_logs_config` resource (beta)
```

```release-note: enhancement
networkmanagement: added output-only field `target_resource_state` to the `google_network_management_vpc_flow_logs_config` resource
```


mmv1/third_party/terraform/services/networkmanagement/resource_network_management_vpc_flow_logs_config_test.go
```diff
@@ -182,6 +182,169 @@ resource "google_network_management_vpc_flow_logs_config" "example" {
 	return fmt.Sprintf("%s\n\n%s\n\n", vpcFlowLogsCfg, testAccNetworkManagementVpcFlowLogsConfig_baseResources(context))
 }
 
+func TestAccNetworkManagementVpcFlowLogsConfig_network(t *testing.T) {
```
shuyama1, line 186: I'd expect this test to specifically testing the update functionality of the network field, like updating from using one network to another. 

Same for the `subnet` field.

mmv1/products/networkmanagement/VpcFlowLogsConfig.yaml
```diff
@@ -153,6 +180,16 @@ properties:
     type: String
     description: |
       Traffic will be logged from the VPN Tunnel. Format: projects/{project_id}/regions/{region}/vpnTunnels/{name}
+  - name: 'subnet'
```
shuyama1, line 151: Question: Is it possible to specify both `network` and `subnet` together? If so, could you please add a test to include this case, otherwise, we may need to mark them as conflict_with.

mmv1/products/networkmanagement/VpcFlowLogsConfig.yaml
```diff
@@ -79,6 +79,32 @@ examples:
       esp_forwarding_rule_name: 'full-test-fresp'
       route_name: 'full-test-route'
       vpc_flow_logs_config_id: 'full-test-id'
+  - name: 'network_management_vpc_flow_logs_config_network_basic'
```
shuyama1, line 63: Thanks for the thorough test cases! To keep the example for the [web documentation](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/network_management_vpc_flow_logs_config) concise, let's pare this down to just the basic and update tests. If these aren't intended for the web docs, the update tests alone will be sufficient.

mmv1/third_party/terraform/services/networkmanagement/resource_network_management_vpc_flow_logs_config_test.go
```diff
@@ -182,6 +182,169 @@ resource "google_network_management_vpc_flow_logs_config" "example" {
...
+}
+
+resource "google_compute_network" "network" {
+  name = "tf-test-flow-logs-network-%{random_suffix}"
+}
+
+resource "google_network_management_vpc_flow_logs_config" "network-test" {
+  vpc_flow_logs_config_id = "tf-test-network-id-%{random_suffix}"
+  location                = "global"
+  network                 = "projects/${data.google_project.project.number}/global/networks/${google_compute_network.network.name}"
```
shuyama1: The field description specifies the format `projects/{project_id}/global/networks/{name}`. Is there a specific reason for using the project number here instead of the id? I think we can simplify this by referencing the network directly `network = google_compute_network.network.id`

Same for the `subnet` field.

shuyama1 (commented):
Thanks for the work. I've left a few comments and questions.

mmv1/third_party/terraform/services/networkmanagement/resource_network_management_vpc_flow_logs_config_test.go
maayanbeltzer: The existing tests already validate the general update mechanism by testing changes to fields like state and aggregation_interval. The network and subnet fields follow this same standard update path, much like the existing vpn_tunnel field and ica field. Since there is no special handling required for them, I believe additional, specific update tests would be redundant. This will require creating another test and another subnet and make the test more complicated for no good reason.

mmv1/products/networkmanagement/VpcFlowLogsConfig.yaml
maayanbeltzer: It is not possible to have both network and subnet fields (only one of them). I added conflict_with to each of the resources (including vpn and ica), thanks!

mmv1/third_party/terraform/services/networkmanagement/resource_network_management_vpc_flow_logs_config_test.go
maayanbeltzer: The API returns the project number.

mmv1/products/networkmanagement/VpcFlowLogsConfig.yaml
maayanbeltzer: You're right, that's much cleaner for the docs. I've removed the _full examples you pointed out, and also trimmed the ones for vpn and ica to be consistent

mmv1/products/networkmanagement/VpcFlowLogsConfig.yaml
maayanbeltzer: I've been trying to implement your suggestion to mark the network, subnet, vpn_tunnel, and interconnect_attachment fields as mutually exclusive, but I've run into some trouble finding the correct syntax for the YAML file.

I first attempted to use conflicts_with and then ConflictsWith, but both resulted in build failures because the property was not found. I then tried using exactly_one_of at the resource level, which also caused a build failure.

[Failed run with ConflictsWith](https://github.com/GoogleCloudPlatform/magic-modules/actions/runs/15993162868/job/45110436528?pr=14381)
[Failed run with exactly_one_of](https://github.com/GoogleCloudPlatform/magic-modules/actions/runs/15993675546/job/45112059659?pr=14381)
For context, the underlying API for this resource uses a oneof field called target_resource, which can be one of the four fields we're discussing.

Given this, what would be the best approach? I can either add a new acceptance test that confirms the provider returns an error when multiple fields are set, or I can leave it as-is and rely on the API-level validation. If there's a declarative way to handle this in MMv1 that I've missed, please let me know.

shuyama1: I believe [`Conflicts`](https://github.com/GoogleCloudPlatform/magic-modules/blob/2a22759ec0725321ddcb6ae69a41d133ef0db933/mmv1/api/type.go#L126C2-L126C11) is what you're looking for. `exactly_one_of` would require one of the fields to be set.

However, since `vpn_tunnel` and `interconnect_attachment` are already implemented and lack of this validation on the Terraform side, now adding it would be a breaking change. I'd suggest we state that at most one of the fields can be specified in the field descriptions and rely on the API-level validation

https://googlecloudplatform.github.io/magic-modules/breaking-changes/breaking-changes/#field-level-breaking-changes

mmv1/products/networkmanagement/VpcFlowLogsConfig.yaml
shuyama1: Thanks!

mmv1/third_party/terraform/services/networkmanagement/resource_network_management_vpc_flow_logs_config_test.go
shuyama1: This update test should specifically validates the network field's update functionality within the Terraform provider. A simple test case could be udpating the configuration from a specified network to its absence. We can simply update the `testAccNetworkManagementVpcFlowLogsConfig_networkUpdate` function to be something like the following 
```
func testAccNetworkManagementVpcFlowLogsConfig_networkUpdate(context map[string]interface{}) string {
	return acctest.Nprintf(`
data "google_project" "project" {
}
resource "google_compute_network" "network" {
  name = "tf-test-flow-logs-network-%{random_suffix}"
}
resource "google_network_management_vpc_flow_logs_config" "network-test" {
  vpc_flow_logs_config_id = "tf-test-network-id-%{random_suffix}"
  location                = "global"
  state = "ENABLED"
}
`, context)
}
```

Please let me know if that makes sense!

shuyama1: This update test should specifically validates the network field's update functionality within the Terraform provider. A simple test case could be udpating the configuration from a specified network to its absence. We can simply update the `testAccNetworkManagementVpcFlowLogsConfig_networkUpdate` function to be something like the following 
```
func testAccNetworkManagementVpcFlowLogsConfig_networkUpdate(context map[string]interface{}) string {
	return acctest.Nprintf(`
data "google_project" "project" {
}
resource "google_compute_network" "network" {
  name = "tf-test-flow-logs-network-%{random_suffix}"
}
resource "google_network_management_vpc_flow_logs_config" "network-test" {
  vpc_flow_logs_config_id = "tf-test-network-id-%{random_suffix}"
  location                = "global"
  state = "ENABLED"
}
`, context)
}
```

Please let me know if that makes sense!

mmv1/products/networkmanagement/VpcFlowLogsConfig.yaml
shuyama1: I believe [`Conflicts`](https://github.com/GoogleCloudPlatform/magic-modules/blob/2a22759ec0725321ddcb6ae69a41d133ef0db933/mmv1/api/type.go#L126C2-L126C11) is what you're looking for. `exactly_one_of` would require one of the fields to be set.

However, since `vpn_tunnel` and `interconnect_attachment` are already implemented and lack of this validation on the Terraform side, now adding it would be a breaking change. I'd suggest we state that at most one of the fields can be specified in the field descriptions and rely on the API-level validation

https://googlecloudplatform.github.io/magic-modules/breaking-changes/breaking-changes/#field-level-breaking-changes

shuyama1 (changes requested):

mmv1/products/networkmanagement/VpcFlowLogsConfig.yaml
maayanbeltzer: Done

mmv1/third_party/terraform/services/networkmanagement/resource_network_management_vpc_flow_logs_config_test.go
maayanbeltzer: It won't work the config must have a target resource. I added another network and subnet to the test.

mmv1/third_party/terraform/services/networkmanagement/resource_network_management_vpc_flow_logs_config_test.go
```diff
@@ -182,6 +182,176 @@ resource "google_network_management_vpc_flow_logs_config" "example" {
 	return fmt.Sprintf("%s\n\n%s\n\n", vpcFlowLogsCfg, testAccNetworkManagementVpcFlowLogsConfig_baseResources(context))
 }
 
+func TestAccNetworkManagementVpcFlowLogsConfig_network(t *testing.T) {
```
shuyama1, line 186: We'll need to make these tests as beta-only, given `network` and `subnet` are beta-only fields. 
Details at https://googlecloudplatform.github.io/magic-modules/test/test/#add-an-update-test step 5:

If beta-only fields are being tested, do the following:

- Change the file suffix to `.go.tmpl`
- Wrap each beta-only test in a separate version guard: `{{- if ne $.TargetVersionName "ga" -}}...{{- else }}...{{- end }}`
- In each beta-only test, ensure that the TestCase sets `ProtoV5ProviderFactories: acctest.ProtoV5ProviderBetaFactories(t)`
- In each beta-only test, ensure that all Terraform resources in all configs have `provider = google-beta` set

mmv1/third_party/terraform/services/networkmanagement/resource_network_management_vpc_flow_logs_config_test.go
shuyama1: okay, but this doesn't seem align with the suggested format in the field description `Format: projects/{project_id}/global/networks/{name}` and `projects/{project_id}/regions/{region}/subnetworks/{name}`
So, the API accepts both the project ID and number, but always outputs the project number?

maayanbeltzer: Until recently, our API accepted both project number and project ID, always returning the project number in the target resource (subnet, vpc, ica, vpn). We've updated it recently, and now the API will return the project ID in *some* cases, the default is still to return the project number. 

I understand the confusion, but this change is specific to the API's output, not the user input format described in the field.

mmv1/third_party/terraform/services/networkmanagement/resource_network_management_vpc_flow_logs_config_test.go
maayanbeltzer: Done

mmv1/third_party/terraform/services/networkmanagement/resource_network_management_vpc_flow_logs_config_test.go.tmpl
```diff
@@ -182,6 +182,189 @@ resource "google_network_management_vpc_flow_logs_config" "example" {
...
+  flow_sampling           = 0.05
+  metadata                = "INCLUDE_ALL_METADATA"
+  description             = "Updated description for network test"
+}
+`, context)
+}
+{{- else }}
+func TestAccNetworkManagementVpcFlowLogsConfig_network(t *testing.T) {
+	t.Skip("Test skipped for GA provider: network field is beta-only")
+}
```
shuyama1:
```suggestion
```
We could just exclude the entire generation of this test in GA

mmv1/third_party/terraform/services/networkmanagement/resource_network_management_vpc_flow_logs_config_test.go.tmpl
```diff
@@ -182,6 +182,189 @@ resource "google_network_management_vpc_flow_logs_config" "example" {
...
+
+resource "google_network_management_vpc_flow_logs_config" "subnet-test" {
+  vpc_flow_logs_config_id = "tf-test-subnet-id-%{random_suffix}"
+  location                = "global"
+  subnet                  = "projects/${data.google_project.project.number}/regions/${google_compute_subnetwork.subnet_update.region}/subnetworks/${google_compute_subnetwork.subnet_update.name}"
+  state                   = "ENABLED"
+}
+`, context)
+}
```
shuyama1: This test also needs to be wrapped in Beta, since `subnet` is beta-only as well

mmv1/third_party/terraform/services/networkmanagement/resource_network_management_vpc_flow_logs_config_test.go.tmpl
maayanbeltzer: Done

mmv1/third_party/terraform/services/networkmanagement/resource_network_management_vpc_flow_logs_config_test.go.tmpl
maayanbeltzer: Hi @shuyama1. Thanks for the review, and my apologies for the delay. I've addressed all your comments, and the tests appear to be passing now

maayanbeltzer: Sorry, it seems like some of the tests are still failing. I'll TAL

maayanbeltzer: There is one test that is failing - according to duckie it is possible that this is just flaky, is there anything I can do to fix it?

shuyama1 (approved):

PR #14388, opened 2025-06-26
Add support for GKE anonymous authentication config

dave-garred:

```release-note:enhancement
container: added support for `anonymous_authentication_config`
```

### DIFF
--- a/mmv1/third_party/terraform/services/container/resource_container_cluster.go.tmpl
+++ b/mmv1/third_party/terraform/services/container/resource_container_cluster.go.tmpl
@@ -2483,6 +2483,26 @@ func ResourceContainerCluster() *schema.Resource {
 					},
 				},
 			},
+			"anonymous_authentication_config": {
+				Type:        schema.TypeList,
+				Optional:    true,
+				MaxItems:    1,
+				Computed:    true,
+				Description: `AnonymousAuthenticationConfig allows users to restrict or enable anonymous access to the cluster.`,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"mode": {
+							Type:         schema.TypeString,
+							Required:     true,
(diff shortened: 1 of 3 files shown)


mmv1/third_party/terraform/services/container/resource_container_cluster.go.tmpl
```diff
@@ -2483,6 +2483,26 @@ func ResourceContainerCluster() *schema.Resource {
 					},
 				},
 			},
+			"anonymous_authentication_config": {
+				Type:        schema.TypeList,
+				Optional:    true,
+				MaxItems:    1,
+				Computed:    true,
```
roaks3, line 2490: Do you know if this needs to be computed? Normally that indicates the field could be set outside of the user's config (most common case is server-side defaults). If nothing like that is happening here, you might be able to remove this line.

roaks3 (commented):
Code LGTM, just one minor comment. Tests pass and I believe the `TestAccContainerNodePool_withNetworkConfig` failure is unrelated.

mmv1/third_party/terraform/services/container/resource_container_cluster.go.tmpl
dave-garred: That appears to be correct, and I've removed it. Thanks for that catch.

dave-garred: I think I misinterpreted your previous comment, this does need to be computed as we will set a default value on any new clusters.

roaks3: Got it, doesn't hurt to try it out when in doubt :)

roaks3 (approved):

PR #14392, opened 2025-06-27
Remove validation from authConfigs in integration connector's connection resource.

anjalisoni3655:
This will help in resolving one of the internal bug where static check of exactly_one_of is not needed with authType UNSPECIFIED.

release-note:bug
`google_integration_connectors_connection`

Fixes: https://github.com/hashicorp/terraform-provider-google/issues/23426


roaks3 (approved):
Makes sense, just need to wait for tests

PR #14393, opened 2025-06-27
Main

Harshita1603:
Fixes https://b.corp.google.com/issues/394160015
Self Service Update(i.e. maintenance_version field) can only be done to a higher version. Consider a case where the instance gets created on the most latest version, then Self Service Update cannot be done and it will fail. There is no consumer API via which we can first downgrade the instance' version, and then use Self Service Update to further update it.

```release-note:none
Fixes failing tests
```

### DIFF
--- a/mmv1/third_party/terraform/services/redis/resource_redis_instance_test.go
+++ b/mmv1/third_party/terraform/services/redis/resource_redis_instance_test.go
@@ -374,26 +374,6 @@ resource "google_redis_instance" "cache" {
 `, context)
 }
 
-func testAccRedisInstance_selfServiceUpdate20240411_00_00(context map[string]interface{}) string {
-	return acctest.Nprintf(`
-resource "google_redis_instance" "cache" {
-  name           = "tf-test-memory-cache%{random_suffix}"
-  memory_size_gb = 1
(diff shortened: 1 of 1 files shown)


mmv1/third_party/terraform/services/redis/resource_redis_instance_test.go
```diff
@@ -255,40 +255,6 @@ func TestAccRedisInstance_redisInstanceAuthEnabled(t *testing.T) {
...
-				Config: testAccRedisInstance_selfServiceUpdate20240411_00_00(context),
-			},
-			{
-				ResourceName:            "google_redis_instance.cache",
-				ImportState:             true,
-				ImportStateVerify:       true,
-				ImportStateVerifyIgnore: []string{"region"},
-			},
-			{
-				Config: testAccRedisInstance_selfServiceUpdate20240503_00_00(context),
```
ScottSuarez, line 280: If these dependent config functions are no longer needed can we delete them as well

Harshita1603: Thanks for pointing it out, Done.

PR #14402, opened 2025-06-29
Fix: Make "match" field in Route Policy's terms as required

harshithpatte-g:
Update match field in Route Policy as required.
Fixes https://github.com/hashicorp/terraform-provider-google/issues/23061

```release-note: bug
compute: updated match field in Route Policy as required
```

### DIFF
--- a/mmv1/products/compute/RouterRoutePolicy.yaml
+++ b/mmv1/products/compute/RouterRoutePolicy.yaml
@@ -110,7 +110,8 @@ properties:
         - name: 'match'
           type: NestedObject
           description: |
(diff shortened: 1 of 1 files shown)


roaks3 (commented):
@harshithpatte-g can you confirm this resource was not possible to create without the `match` being specified (ie. there was always a 400 error)? Otherwise this could be a breaking change for some users.

harshithpatte-g:
@roaks3 I confirm that it always resulted in an error.

> @harshithpatte-g can you confirm this resource was not possible to create without the `match` being specified (ie. there was always a 400 error)? Otherwise this could be a breaking change for some users.

roaks3 (approved):

PR #14403, opened 2025-06-30
Promote autokey resources to GA

nb-goog:

```release-note: enhancement
Promoting Autokey resources to GA from beta
```


mmv1/products/kms/AutokeyConfig.yaml
```diff
@@ -59,6 +58,7 @@ exclude_sweeper: true
 examples:
   - name: 'kms_autokey_config_all'
     primary_resource_id: 'example-autokeyconfig'
+    # test depends upon google_project_service_identity service which is still in beta, so we need to keep test limited to beta
```
ScottSuarez, line 61: Hello, is there a way to add testing in GA provider. I am worried about supporting a resource that is entirely untested in a specific provider version

nb-goog: I also had same worry, so tried to search for all the entities having dependency on google_project_service_identity and all of them are having their tests in beta only.

I am not sure how to have and run those tests in GA without service-identity, but i will still give it a try

PR #14416, opened 2025-06-30
Add Endpoint for Vertex Model Garden OSS Gen AI model deployment

lizzij:
This PR adds the necessary configuration files and examples to enable Terraform users to deploy a Model Garden LLM to an endpoint for inference with Vertex Model Garden services.

Specifically, this PR:
- Adds the EndpointWithModelGardenDeployment.yaml file to define the resource.
- Includes a basic Terraform example.
- Adds a create and update test case for the resource.

Add endpoint with Model Garden model deployed resource.
- Vertex AI Model Garden service
- Based on [ModelGardenService.deploy](https://cloud.google.com/vertex-ai/docs/reference/rest/v1beta1/projects.locations/deploy)

Addresses [hashicorp/terraform-provider-google#23217](https://github.com/hashicorp/terraform-provider-google/issues/23217) and [hashicorp/terraform-provider-google#22478](https://github.com/hashicorp/terraform-provider-google/issues/22478)

```release-note:new-resource
`google_vertex_ai_endpoint_with_model_garden_deployment`
```


mmv1/products/vertexai/EndpointWithModelGardenDeployment.yaml
```diff
@@ -0,0 +1,972 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+---
+name: EndpointWithModelGardenDeployment
+api_resource_type_kind: Endpoint
+description: Create an Endpoint and deploy a Model Garden model to it.
+references:
+  guides: null
```
slevenick: Null? I think this can be omitted if we don't have one

mmv1/products/vertexai/EndpointWithModelGardenDeployment.yaml
```diff
@@ -0,0 +1,972 @@
+# Copyright 2024 Google Inc.
```
slevenick:
```suggestion
# Copyright 2025 Google Inc.
```

mmv1/products/vertexai/EndpointWithModelGardenDeployment.yaml
```diff
@@ -0,0 +1,972 @@
...
+examples:
+- name: 'vertex_ai_deploy'
+  primary_resource_id: 'deploy'
+  primary_resource_name: 'fmt.Sprintf("tf_test_example_deploy%s", context["random_suffix"])'
+  vars:
+    project: 'vertex-ai'
+    publisher_model_name: 'publisher_model_name'
+  ignore_read_extra:
+  - 'project'
+  exclude_test: true
```
slevenick: Why exclude test?

slevenick (commented):
Is this resource updatable? If so, can you add a test for that?

mmv1/templates/terraform/examples/vertex_ai_deploy.tf.tmpl
```diff
@@ -0,0 +1,9 @@
+resource "google_vertex_ai_endpoint_with_model_garden_deployment" "{{$.PrimaryResourceId}}" {
+  publisher_model_name = "publishers/google/models/paligemma@paligemma-224-float32"
+  location             = "us-central1"
+  model_config {
+    accept_eula =  true
+  }
+}
+
+data "google_project" "project" {}
```
slevenick, line 9: Do we need this project here? It doesn't look like it's used

mmv1/templates/terraform/custom_delete/resource_vertexai_endpoint_with_model_garden_deployment.go.tmpl
```diff
@@ -0,0 +1,38 @@
+log.Printf("[DEBUG] Beginning custom_delete for Vertex AI Endpoint with Model Garden Deployment")
+
+// Log resource ID
+log.Printf("[DEBUG] Resource ID: %s", d.Id())
+
+// Log fields using d.Get
```
slevenick: Why do we need any of these things to be logged?

mmv1/templates/terraform/custom_delete/resource_vertexai_endpoint_with_model_garden_deployment.go.tmpl
```diff
@@ -0,0 +1,38 @@
...
+	log.Printf("[DEBUG] deployedModels.0.id not found in state attributes")
+}
+
+// Surpress userAgent not used error
+log.Printf("[DEBUG] userAgent for deletion: %s", userAgent)
+
+// Continue with actual deletion logic here (undeploy model, delete endpoint, etc.)
+// ...
+log.Printf("[DEBUG] Finished deleting endpoint.")
+return nil
```
slevenick: It looks like this entire custom_delete is just made to log field values? Does this need to actually do anything?

If it's the case that these things can't be deleted, we have built-in support for those via exclude_delete

mmv1/templates/terraform/custom_delete/resource_vertexai_endpoint_with_model_garden_deployment.go.tmpl
```diff
@@ -0,0 +1,38 @@
...
+if idVal, ok := d.State().Attributes["deployed_models.0.id"]; ok {
+	log.Printf("[DEBUG] deployedModels[0].id = %s", idVal)
+} else {
+	log.Printf("[DEBUG] deployedModels.0.id not found in state attributes")
+}
+
+// Surpress userAgent not used error
+log.Printf("[DEBUG] userAgent for deletion: %s", userAgent)
+
+// Continue with actual deletion logic here (undeploy model, delete endpoint, etc.)
```
slevenick: Should this exist?

mmv1/templates/terraform/post_create/resource_vertexai_endpoint_with_model_garden_deployment.go.tmpl
```diff
@@ -0,0 +1,59 @@
...
+
+// Log res structure
+log.Printf("[DEBUG] Top-level keys in res:")
+for k := range res {
+	log.Printf("[DEBUG]   - %s", k)
+}
+
+// Declare and populate opRes
+var opRes map[string]interface{}
+err = VertexAIOperationWaitTimeWithResponse(
```
slevenick, line 20: Are we not correctly following the operation in the normal create flow? Why do we need to handle it in a post_create?

mmv1/templates/terraform/post_create/resource_vertexai_endpoint_with_model_garden_deployment.go.tmpl
```diff
@@ -0,0 +1,59 @@
...
+endpointFull, ok := opRes["endpoint"].(string)
+if !ok || endpointFull == "" {
+	log.Printf("[ERROR] 'endpoint' not found or empty in opRes. Full opRes: %#v", opRes)
+	return fmt.Errorf("Create response didn't contain 'endpoint'. Create may not have succeeded.")
+}
+log.Printf("[DEBUG] Extracted full endpoint from opRes: %s", endpointFull)
+
+// Check format and extract endpoint name without strict project name match
+parts := strings.Split(endpointFull, "/")
+if len(parts) != 6 || parts[0] != "projects" || parts[2] != "locations" || parts[4] != "endpoints" {
```
slevenick, line 45: Instead of doing parsing within post_create (which would cause problems if this resource became updatable in the future) you can prevent diffs on the endpoint field via a diff suppress function. There are many examples of these if you search for `diffSuppressFunc` in the code base

mmv1/third_party/terraform/services/vertexai/resource_vertex_ai_deploy_test.go
```diff
@@ -0,0 +1,56 @@
...
+		},
+	})
+}
+
+func testAccVertexAIEndpointWithModelGardenDeployment_basic(context map[string]interface{}) string {
+	return acctest.Nprintf(`
+resource "google_vertex_ai_endpoint_with_model_garden_deployment" "test" {
+  publisher_model_name = "publishers/google/models/paligemma@paligemma-224-float32"
+  location             = "us-central1"
+  project              = "%{project}"
```
slevenick: project should be inherited from provider config, this can be removed

lizzij:
> Is this resource updatable? If so, can you add a test for that?

No the resource is not updatable.

mmv1/products/vertexai/EndpointWithModelGardenDeployment.yaml
lizzij: Removed.

mmv1/products/vertexai/EndpointWithModelGardenDeployment.yaml
lizzij: Updated.

mmv1/products/vertexai/EndpointWithModelGardenDeployment.yaml
lizzij: Removed.

mmv1/third_party/terraform/services/vertexai/resource_vertex_ai_deploy_test.go
francis-ohara: Removed.

mmv1/products/vertexai/EndpointWithModelGardenDeployment.yaml
francis-ohara: I wanted to confirm if adding `exclude_test` to the example may be necessary since we are writing handwritten tests for the resource and may not need to generate an automated acceptance test for it.

mmv1/templates/terraform/post_create/resource_vertexai_endpoint_with_model_garden_deployment.go.tmpl
lizzij: The created Endpoint resource is represented as a resource name in the LRO response. The post_create read the LRO response and read the Endpoint resource. This is the recommended approach from Terraform office hour.

mmv1/products/vertexai/EndpointWithModelGardenDeployment.yaml
slevenick: Unless there's a particular reason the test is invalid we should prefer to test here and in handwritten tests.

This example is used in documentation, so whenever possible we would like to make sure those examples actually work

mmv1/products/vertexai/EndpointWithModelGardenDeployment.yaml
```diff
@@ -0,0 +1,985 @@
...
+    base_url: "{{op_id}}"
+  result:
+    resource_inside_response: true
+custom_code:
+  post_create: "templates/terraform/post_create/resource_vertexai_endpoint_with_model_garden_deployment.go.tmpl"
+  custom_delete: "templates/terraform/custom_delete/resource_vertexai_endpoint_with_model_garden_deployment.go.tmpl"
+examples:
+  - name: "vertex_ai_deploy"
+    primary_resource_id: "deploy"
+    primary_resource_name: 'fmt.Sprintf("tf_test_example_deploy%s", context["random_suffix"])'
```
slevenick: This is only needed when generating IAM resources, which I don't think will make sense for this resource

mmv1/products/vertexai/EndpointWithModelGardenDeployment.yaml
```diff
@@ -0,0 +1,985 @@
...
+---
+name: EndpointWithModelGardenDeployment
+api_resource_type_kind: Endpoint
+description: Create an Endpoint and deploy a Model Garden model to it.
+references:
+  guides:
+    "Use models in Model Garden": "https://cloud.google.com/vertex-ai/generative-ai/docs/model-garden/use-models"
+    "Overview of Model Garden": "https://cloud.google.com/vertex-ai/generative-ai/docs/model-garden/explore-models"
+    "Overview of self-deployed models": "https://cloud.google.com/vertex-ai/generative-ai/docs/model-garden/self-deployed-models"
+  api: "https://cloud.google.com/vertex-ai/docs/reference/rest/v1beta1/projects.locations/deploy"
```
slevenick: This should point at the GA docs rather than v1beta1

mmv1/products/vertexai/EndpointWithModelGardenDeployment.yaml
```diff
@@ -0,0 +1,985 @@
...
+    immutable: true
+    url_param_only: true
+    required: true
+  - name: endpoint
+    type: String
+    description:
+      Resource ID segment making up resource `endpoint`. It identifies the
+      resource within its parent collection as described in https://google.aip.dev/122.
+    url_param_only: true
+    output: true
```
slevenick, line 77: If endpoint is output-only and server generated it shouldn't be in `parameters`. It should move to `properties`

mmv1/products/vertexai/EndpointWithModelGardenDeployment.yaml
```diff
@@ -0,0 +1,985 @@
...
+properties:
+  - name: publisherModelName
```
slevenick, line 78: `required: true` probably

mmv1/third_party/terraform/services/vertexai/resource_vertex_ai_deploy_test.go
```diff
@@ -0,0 +1,53 @@
...
+		},
+	})
+}
+
+func testAccVertexAIEndpointWithModelGardenDeployment_basic(context map[string]interface{}) string {
+	return acctest.Nprintf(`
+resource "google_vertex_ai_endpoint_with_model_garden_deployment" "test" {
+  publisher_model_name = "publishers/google/models/paligemma@paligemma-224-float32"
+  location             = "us-central1"
+  model_config {
```
slevenick, line 34: Can you add some additional fields here to test them? Some sub-fields of model_config & deploy_config would be great. We don't need to test *every* field as there are so many in this resource, but more coverage would be good

mmv1/templates/terraform/custom_delete/resource_vertexai_endpoint_with_model_garden_deployment.go.tmpl
slevenick: Please address this

mmv1/products/vertexai/EndpointWithModelGardenDeployment.yaml
lizzij: This is optional since user could either specify this field or Hugging Face model ID

mmv1/products/vertexai/EndpointWithModelGardenDeployment.yaml
lizzij: Update to GA doc.

mmv1/products/vertexai/EndpointWithModelGardenDeployment.yaml
lizzij: Test is covered by handwritten test and seems like the common practice for other Vertex AI resources.

mmv1/third_party/terraform/services/vertexai/resource_vertex_ai_deploy_test.go
lizzij: Added `TestAccVertexAIEndpointWithModelGardenDeployment_withConfigs` that includes both `model_config` and `deploy_config`.

mmv1/products/vertexai/EndpointWithModelGardenDeployment.yaml
lizzij: Removed.

mmv1/products/vertexai/EndpointWithModelGardenDeployment.yaml
lizzij: Moved to `properties`.

mmv1/templates/terraform/post_create/resource_vertexai_endpoint_with_model_garden_deployment.go.tmpl
lizzij: Endpoint field itself is nested in the LRO response field. It seems not possible to read it without post_create.

mmv1/templates/terraform/custom_delete/resource_vertexai_endpoint_with_model_garden_deployment.go.tmpl
francis-ohara: `custom_delete` has been removed and resource has been updated to currently not support deletion.

mmv1/templates/terraform/examples/vertex_ai_deploy.tf.tmpl
lizzij: It seems to read the env variable set up in https://googlecloudplatform.github.io/magic-modules/test/run-tests/#before-you-begin step for testing.

I am following similar Vertex AI examples like `mmv1/templates/terraform/examples/vertex_ai_index_endpoint_with_psc.tf.tmpl` and `mmv1/templates/terraform/examples/vertex_ai_index_endpoint.tf.tmpl`

mmv1/templates/terraform/custom_delete/resource_vertexai_endpoint_with_model_garden_deployment.go.tmpl
francis-ohara: The `custom_delete` template has been removed as resource currently does not support deletion.

mmv1/templates/terraform/custom_delete/resource_vertexai_endpoint_with_model_garden_deployment.go.tmpl
francis-ohara: The `custom_delete` template has been removed.

mmv1/products/vertexai/EndpointWithModelGardenDeployment.yaml
slevenick: Then can we add `exactly_one_of` to require a user to set either this or the Hugging Face model ID?

mmv1/products/vertexai/EndpointWithModelGardenDeployment.yaml
```diff
@@ -0,0 +1,984 @@
...
+    ignore_read: true
+  - name: huggingFaceModelId
+    type: String
+    description: |-
+      The Hugging Face model to deploy.
+      Format: Hugging Face model ID like `google/gemma-2-2b-it`.
+  - name: modelConfig
+    type: NestedObject
+    description: The model config to use for the deployment.
+    ignore_read: true
```
slevenick: Does this not get returned by the API? I'm worried that if we need this here, we'll need it on other parts of this gigantic schema.

What's going on with this resource that it needs ignore_read in so many places?

lizzij: As discussed offline, the `ignore_read` is because the created resource is an `Endpoint` whereas the fields with `ignore_read` are representing input. The model itself cannot be read and cannot be updated. `DeployedModel` is a different entity from Model Garden models. The former is a pointer to the model deployment, whereas the latter is a shared public representation of an OSS GenAI model.

mmv1/products/vertexai/EndpointWithModelGardenDeployment.yaml
```diff
@@ -0,0 +1,991 @@
...
+    "Overview of self-deployed models": "https://cloud.google.com/vertex-ai/generative-ai/docs/model-garden/self-deployed-models"
+  api: "https://cloud.google.com/vertex-ai/docs/reference/rest/v1/projects.locations/deploy"
+docs:
+id_format: "projects/{{project}}/locations/{{location}}/endpoints/{{endpoint}}"
+base_url: "projects/{{project}}/locations/{{location}}:deploy"
+self_link: "projects/{{project}}/locations/{{location}}/endpoints/{{endpoint}}"
+create_url: "projects/{{project}}/locations/{{location}}:deploy"
+immutable: true
+exclude_delete: true  # the resource does not yet support deletion
+import_format:
```
slevenick: We should remove import_format and import entirely. I think you can add ignore_import: true to do this

mmv1/third_party/terraform/services/vertexai/resource_vertex_ai_deploy_test.go
```diff
@@ -0,0 +1,105 @@
...
+	acctest.VcrTest(t, resource.TestCase{
+		PreCheck:                 func() { acctest.AccTestPreCheck(t) },
+		ProtoV5ProviderFactories: acctest.ProtoV5ProviderFactories(t),
+		CheckDestroy:             nil,
+		Steps: []resource.TestStep{
+			{
+				Config: testAccVertexAIEndpointWithModelGardenDeployment_withConfigs(context),
+			},
+			{
+				ResourceName:      "google_vertex_ai_endpoint_with_model_garden_deployment.test_with_configs",
```
slevenick: You can remove this step entirely as it is based on Import

mmv1/third_party/terraform/services/vertexai/resource_vertex_ai_deploy_test.go
```diff
@@ -0,0 +1,105 @@
...
+	acctest.VcrTest(t, resource.TestCase{
+		PreCheck:                 func() { acctest.AccTestPreCheck(t) },
+		ProtoV5ProviderFactories: acctest.ProtoV5ProviderFactories(t),
+		CheckDestroy:             nil, // the resource does not yet support deletion
+		Steps: []resource.TestStep{
+			{
+				Config: testAccVertexAIEndpointWithModelGardenDeployment_basic(context),
+			},
+			{
+				ResourceName:      "google_vertex_ai_endpoint_with_model_garden_deployment.test",
```
slevenick: Remove step entirely

mmv1/products/vertexai/EndpointWithModelGardenDeployment.yaml
francis-ohara: Removed.

mmv1/third_party/terraform/services/vertexai/resource_vertex_ai_deploy_test.go
francis-ohara: Removed.

mmv1/third_party/terraform/services/vertexai/resource_vertex_ai_deploy_test.go
francis-ohara: Removed.

mmv1/products/vertexai/EndpointWithModelGardenDeployment.yaml
lizzij: Added `exactly_one_of`.

slevenick (approved):

PR #14420, opened 2025-06-30
tgc-revival: support google_compute_backend_bucket

zli82016:
https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/products/compute/BackendBucket.yaml

The field `cdn_policy` is C + O
The raw_config in one test doesn’t have `cdn_policy`.
In the exported cai asset, `cdnPolicy` doesn't have `signedUrlCacheMaxAgeSec`
After the cai2hcl from the exported cai asset, the converted exported_config has `cdn_policy`.
After tfplan2cai from exported_config, in the converted cai asset (roundtrip_asset), "signedUrlCacheMaxAgeSec": 0 as it has `send_empty_value: true`. 
`signedUrlCacheMaxAgeSec` is removed from the roundtrip_asset during the test with `tgc_test_ignore_in_asset` in yaml file. Then the roundtrip_asset will be converted to roundtrip_config.

```release-note:none

```

### DIFF
--- a/mmv1/third_party/tgc_next/test/assert_test_files.go
+++ b/mmv1/third_party/tgc_next/test/assert_test_files.go
@@ -353,19 +353,19 @@ func deleteFieldsFromAsset(assets []caiasset.Asset, ignoredResourceDataFields []
 		if len(parts) <= 1 {
 			continue
 		}
-		if parts[0] == "resource" {
-			if _, ok := ignoredFieldsMap["resource"]; !ok {
-				ignoredFieldsMap["resource"] = make([]Field, 0)
+		if parts[0] == "RESOURCE" {
+			if _, ok := ignoredFieldsMap["RESOURCE"]; !ok {
+				ignoredFieldsMap["RESOURCE"] = make([]Field, 0)
 			}
 			f := Field{Path: parts[1:]}
-			ignoredFieldsMap["resource"] = append(ignoredFieldsMap["resource"], f)
+			ignoredFieldsMap["RESOURCE"] = append(ignoredFieldsMap["RESOURCE"], f)
 		}
 	}
 
 	for _, asset := range assets {
 		if asset.Resource != nil && asset.Resource.Data != nil {
 			data := asset.Resource.Data
-			for _, ignoredField := range ignoredFieldsMap["resource"] {
+			for _, ignoredField := range ignoredFieldsMap["RESOURCE"] {
 				path := ignoredField.Path
 				deleteMapFieldByPath(data, path)
 			}
--- a/mmv1/third_party/tgc_next/test/assert_test_files.go
+++ b/mmv1/third_party/tgc_next/test/assert_test_files.go
@@ -344,7 +344,7 @@ type Field struct {
 	Path []string
 }
 
-// Deletes fileds from the resource data of CAI assets
+// Deletes fields from the resource data of CAI assets
 func deleteFieldsFromAssets(assets []caiasset.Asset, ignoredResourceDataFields []string) []caiasset.Asset {
 	// The key is the content type, such as "resource"
 	ignoredFieldsMap := make(map[string][]Field, 0)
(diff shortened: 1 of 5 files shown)


zli82016:
cc @hao-nan-li

mmv1/third_party/tgc_next/test/assert_test_files.go
```diff
@@ -328,6 +330,45 @@ func getRoundtripConfig(t *testing.T, testName string, tfDir string, ancestryCac
...
+					}
+				}
+				if i == len(ignoredFieldPathes)-1 {
+					delete(data, ignoredFieldPathes[i])
+				}
+			}
+		}
+	}
+	return assets
+}
```
trodge: This implementation seems a bit confusing to me. Perhaps we could move L353-L365 into a helper method.

We could also add a `DeletePath` method to the `AssetResource` type and call it here.

It might also be worth refactoring so that the paths in the ignored fields map are a type that holds a slice of steps.

mmv1/third_party/tgc_next/test/assert_test_files.go
```diff
@@ -328,6 +330,45 @@ func getRoundtripConfig(t *testing.T, testName string, tfDir string, ancestryCac
...
+
+	for _, asset := range assets {
+		if asset.Resource != nil && asset.Resource.Data != nil {
+			data := asset.Resource.Data
+			for _, ignoredFieldPathes := range ignoredFieldsMap["resource"] {
+				i := 0
+				for i < len(ignoredFieldPathes)-1 {
+					k := ignoredFieldPathes[i]
+					if v, ok := data[k]; ok {
+						data = v.(map[string]interface{})
```
trodge: This looks like it might panic if the path is incorrect.

trodge (changes requested):

mmv1/third_party/tgc_next/test/assert_test_files.go
zli82016: Good catch

mmv1/third_party/tgc_next/test/assert_test_files.go
zli82016: Moved L353-L365 into a helper method and added a type Field.

`DeletePath` is just for tests, so I will hold it now. If necessary, we can add it later.

mmv1/third_party/tgc_next/test/assert_test_files.go
```diff
@@ -328,6 +330,80 @@ func getRoundtripConfig(t *testing.T, testName string, tfDir string, ancestryCac
...
+//			"user": "admin",
+//		},
+//	}
+//
+// Path of "host" in "data" is ["database", "host"]
+type Field struct {
+	Path []string
+}
+
+// Deletes fileds from the resource data of CAI assets
```
trodge:
```suggestion
// Deletes fields from the resource data of CAI assets
```

mmv1/third_party/tgc_next/test/assert_test_files.go
```diff
@@ -328,6 +330,80 @@ func getRoundtripConfig(t *testing.T, testName string, tfDir string, ancestryCac
...
+func deleteFieldsFromAsset(assets []caiasset.Asset, ignoredResourceDataFields []string) []caiasset.Asset {
```
trodge:
```suggestion
func deleteFieldsFromAssets(assets []caiasset.Asset, ignoredResourceDataFields []string) []caiasset.Asset {
```

(assuming that this deletes from all assets)

mmv1/third_party/tgc_next/test/assert_test_files.go
trodge: Even if it is just for tests, having it in the Asset types makes more sense to me. Otherwise we're duplicating the structure of the asset here in the form of a map that could be a struct.

trodge (approved):

mmv1/third_party/tgc_next/test/assert_test_files.go
zli82016: Which map could be a struct here?

zli82016: I see. Do you mean ignoredFieldsMap?

PR #14426, opened 2025-07-01
Make openapi-generate output more stable

IngmarStein:
Currently, `openapi-generate`'s output is non-deterministic because properties are unordered (they are stored in a slice which is produced by iterating over a map). This causes unnecessary diffs when regenerating the yaml files.

This commit sorts the property keys lexicographically to provide a stable output.

```release-note:none
openapi-generate: produce stable output
```


zli82016:
@slevenick, can you help to review this PR? I don't know so much about openapi? Thanks.

PR #14427, opened 2025-07-01
feat:(storagetransfer) added federated identity config for azure storage transfer

gurusai-voleti:

```release-note:enhancement
storagetransfer: added `federated_identity_config` to resource `google_storage_transfer_job`
```


mmv1/third_party/terraform/services/storagetransfer/resource_storage_transfer_job.go.tmpl
```diff
@@ -1129,6 +1145,30 @@ func resourceStorageTransferJobStateImporter(d *schema.ResourceData, meta interf
...
+}
+
+func flattenAzureFederatedIdentifyConfig(d *schema.ResourceData) []map[string]interface{} {
+	if ((d.Get("transfer_spec.0.azure_blob_storage_data_source.0.federated_identity_config.0.client_id") == "") || (d.Get("transfer_spec.0.azure_blob_storage_data_source.0.federated_identity_config.0.tenant_id") == "") ){
+		return []map[string]interface{}{}
+	}
+
+	data := map[string]interface{}{
+		"client_id": d.Get("transfer_spec.0.azure_blob_storage_data_source.0.federated_identity_config.0.client_id"),
+		"tenant_id": d.Get("transfer_spec.0.azure_blob_storage_data_source.0.federated_identity_config.0.tenant_id"),
```
kautikdk, line 1172: So we don't get this credentials back from the API response and we have to set them from the state file as it is. Is that correct?

gurusai-voleti: yes we need to set from the user terraform config only, as it will not return in API response

gurusai-voleti:
> While the `federated_identity_config` field changes looks good to me, I was wondering why the `credentials_secret` is being promoted to GA in the PR and also `azure_credentials` is changed to optional from required? I will also wait for manual test runs before approving this PR.

yes azure has three kinds of creds can be provided to authenticate sources, any one of them is required in API requests so it creds made optional and using ExactlyOneOf to ensure one is present

kautikdk (approved):
Overall LGTM,
I would leave `credentials_secret` flag promotion decision to @SirGitsalot.

kautikdk:
We may need to update this file as well: [mmv1/third_party/terraform/services/storagetransfer/resource_storage_transfer_job_meta.yaml.tmpl](https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/third_party/terraform/services/storagetransfer/resource_storage_transfer_job_meta.yaml.tmpl)

gurusai-voleti:
> Is it possible to add tests? Everything looks fine, but it would be nice to see it actually running. It doesn't look like there are any existing tests related to Azure auth; I assume there may be some issue with testing it.

these tests have sensitive data, so we did manual tests and provided logs, please check the corresponding buganizer ticket, i have tagged you on buganizer and provided the test logs link there

kautikdk:
> Is it possible to add tests? Everything looks fine, but it would be nice to see it actually running. It doesn't look like there are any existing tests related to Azure auth; I assume there may be some issue with testing it.

We don't have any Storage Transfer Service tests for the third party integration as we currently don't have required infra to add tests. So we are going ahed with the manual testing till we setup Azure and AWS accounts for the testing. As @gurusai-voleti mentioned, they have provided manual testing logs internally to validate the working of the new fields. I think that's the only way to move forward as of now.

PR #14428, opened 2025-07-01
sql: fix nil pointer while importing `google_sql_database` and remove invalid import_format

ramonvermeulen:
Closes https://github.com/hashicorp/terraform-provider-google/issues/22728

```release-note:bug
sql: fixed a provider crash when importing `google_sql_database` resource
```

```release-note:note
sql: removed the invalid import format `{{name}}` for `google_sql_database` resource
```

### DIFF
--- a/mmv1/products/sql/Database.yaml
+++ b/mmv1/products/sql/Database.yaml
@@ -26,7 +26,6 @@ import_format:
   - '{{project}}/{{instance}}/{{name}}'
   - 'instances/{{instance}}/databases/{{name}}'
   - '{{instance}}/{{name}}'
-  - '{{name}}'
 timeouts:
   insert_minutes: 20
   update_minutes: 20
--- a/mmv1/templates/terraform/pre_read/sql_database_activation_policy.tmpl
+++ b/mmv1/templates/terraform/pre_read/sql_database_activation_policy.tmpl
@@ -1,8 +1,8 @@
 instance := d.Get("instance").(string)
 databaseInstance, err := config.NewSqlAdminClient(userAgent).Instances.Get(project, instance).Do()
-	if err != nil {
-		return err
-	}
-	if databaseInstance.Settings.ActivationPolicy != "ALWAYS" {
-		return nil
-	}
+if err != nil {
+	return err
+}
+if databaseInstance.Settings != nil && databaseInstance.Settings.ActivationPolicy != "ALWAYS" {
+	return nil
+}


ramonvermeulen:
Please read: https://github.com/hashicorp/terraform-provider-google/issues/22728#issuecomment-3022920222

Also maybe the `{{ name }}` entry should be removed from the import formats?:

https://github.com/GoogleCloudPlatform/magic-modules/blob/9ddf9fe53ec97a7bebae0793c783f378aca63b5c/mmv1/products/sql/Database.yaml#L29

If so I will update that onto this PR as well.

EDIT:

Already pushed ^ in https://github.com/GoogleCloudPlatform/magic-modules/pull/14428/commits/5ce489c03749e30938e6619fb77d3d66ce54aff6 since I think it makes sense to remove this `import_format`.

mmv1/products/sql/Database.yaml
```diff
@@ -26,7 +26,6 @@ import_format:
   - '{{project}}/{{instance}}/{{name}}'
   - 'instances/{{instance}}/databases/{{name}}'
   - '{{instance}}/{{name}}'
-  - '{{name}}'
```
shuyama1, line 29: You're right that {{name}} seems an incorrect import format, however, removing an import format is a breaking change that we need to avoid in a minor release. I'd suggest we hold off on this change until the next major release.

https://googlecloudplatform.github.io/magic-modules/breaking-changes/breaking-changes/#resource-level-breaking-changes

ramonvermeulen: Check! Would you want me to put that change in a separate PR? I don't know if there is a certain labeling strategy or whatsoever so that there is a list of PRs that are waiting for the next major?

On the other hand I was also thinking, does it really matter in this case that the change is "breaking"? I mean my guess would be that no one will be impacted since the import format is unusable anyways (gives an error if you try it via {{ name }} only), but for me it is hard to determine what the "actual" impact will be so I am also fine with waiting for the next major.

shuyama1: Apologies for the delay in my reply, this didn't appear in my review list somehow.

The contribution window for the 7.0.0 major release is actually open now https://googlecloudplatform.github.io/magic-modules/breaking-changes/make-a-breaking-change/#in-the-700-major-release. 

However, I think you're right. {{name}} is never a valid import format for `google_sql_database`. As per the [guidelines](https://googlecloudplatform.github.io/magic-modules/breaking-changes/make-a-breaking-change/#in-minor-releases): "If a breaking change fixes a bug that impacts all configurations that include a field or resource, it is generally allowed in a minor release", we'd be fine to make this change in minor releases,

shuyama1 (approved):

PR #14446, opened 2025-07-02
Adding initial Playbook support for Dialogflow CX

zhihaos:
fulfill request https://github.com/hashicorp/terraform-provider-google/issues/23328

```release-note:new-resource
`google_dialogflow_cx_playbook` (initial basic support, full features to follow in a later release)
```


zhihaos:
After checking the test log, the requests were succeeded and responses were expected. Unsure what the error means. 

```
sdk.helper_resource: Unexpected error: test_name=TestAccDialogflowCXPlaybook_dialogflowcxPlaybookBasicExample test_terraform_path=/bin/terraform test_working_directory=/tmp/plugintest3453659892 test_step_number=1
  error=
  | After applying this test step, the plan was not empty.
  | stdout:
  | 
  | 
  |   ~ update in-place
  | 
  | 
  |   # google_dialogflow_cx_playbook.my-playbook will be updated in-place
  |   ~ resource "google_dialogflow_cx_playbook" "my-playbook" {
  |         id                   = "projects/ci-test-project-188019/locations/global/agents/8eadb94d-5494-404f-b4a4-a0ad8cb02d0e/playbooks/6d90aeb4-ee02-4e9e-8456-04d8a6b65a1a"
  |         name                 = "6d90aeb4-ee02-4e9e-8456-04d8a6b65a1a"
  |         # (8 unchanged attributes hidden)
  | 
  |       ~ instruction {
  |             # (1 unchanged attribute hidden)
  | 
  |           ~ steps {
  |               ~ steps = jsonencode(
  |                   ~ [
  |                       - {
  |                           - steps = [
  |                               - {
  |                                   - text = "step 1 1"
  |                                 },
  |                               - {
  |                                   - steps = [
  |                                       - {
  |                                           - text = "step 1 2 1"
  |                                         },
  |                                       - {
  |                                           - text = "step 1 2 2"
  |                                         },
  |                                     ]
  |                                   - text  = "step 1 2"
  |                                 },
  |                               - {
  |                                   - text = "step 1 3"
  |                                 },
  |                             ]
  |                           - text  = ""
  |                         },
  |                     ] -> {
  |                       + steps = [
  |                           + {
  |                               + text = "step 1 1"
  |                             },
  |                           + {
  |                               + steps = [
  |                                   + {
  |                                       + text = "step 1 2 1"
  |                                     },
  |                                   + {
  |                                       + text = "step 1 2 2"
  |                                     },
  |                                 ]
  |                               + text  = "step 1 2"
  |                             },
  |                           + {
  |                               + text = "step 1 3"
  |                             },
  |                         ]
  |                     }
  |                 )
  |                 # (1 unchanged attribute hidden)
  |             }
  | 
  |             # (2 unchanged blocks hidden)
  |         }
  |     }
  | 
  | Plan: 0 to add, 1 to change, 0 to destroy.
```

<del>@rileykarson could you please give any pointers here? 🙏</del>
NVM, I found one issue in the code, testing the fix right now.

zhihaos:
The current issue is that the following value:
```
    "steps": [
     "{\"text\":\"step 1 1\"}",
     "{\"steps\":[{\"text\":\"step 1 2 1\"},{\"text\":\"step 1 2 2\"}],\"text\":\"step 1 2\"}",
     "{\"text\":\"step 1 3\"}"
    ],
```
is supposed to be array of json code like:
```
"steps": [
  {
    "test": "step 1 1"
  },
  ...
]
```
and it seems that it is parsed as string with quotations around it, which raises the following error:
`Invalid value at 'playbook.instruction.steps[0].steps[0]' (type.googleapis.com/google.cloud.dialogflow.cx.v3.Playbook.Step), \"{\"text\":\"step 1 1\"}\"`

zhihaos:
> I see this is in draft- let me know when this is ready to review, I think iterative passes would be more trouble than they're worth given the size of the change!

@rileykarson thanks for the reply. Yes, I am adding test coverage to reach 100%. Will let you know when it is ready.

melinath:
@zhihaos follow-up from office hours. I've looked into the issue with ignore_read on the nested fields and found a couple things:

1. the field path is being generated wrong: https://github.com/modular-magician/terraform-provider-google-beta/blob/53298b5366f07841687f1ef3b52af0b3d52a7a46/google-beta/services/dialogflowcx/resource_dialogflow_cx_playbook.go#L3467-L3469 This starts with `handlers.0.handlers.0.event_handlers` but handlers doesn't contain a handlers field - it should just be `handlers.0.event_handlers`. This is a bug in the generator due to the `handlers` field being an [array](https://github.com/GoogleCloudPlatform/magic-modules/blob/ab923c8c57704663a42c375ab0ccb4df2202b1a0/mmv1/api/type.go#L327).
2. (Bigger issue) Because `handlers` is an array field, the flattener isn't going to work properly even without that bug. Some context: for historical reasons, Terraform models "nested objects" as lists with a single item. The `0` in the path used in the flattener indicates the index of the nested object in that array. However, `handlers` is _actually_ an array, which means that the `0` would need to be replaced with the actual index of the handler to work properly. But that index isn't passed down into the flattener that would need it.

Basically, ignore_read is supported on nested objects, but _not_ inside arrays of nested objects. Supporting it properly would require some consideration and possibly a significant refactoring of how flatteners are generated.

That doesn't block you. The workaround is unfortunately a bit hacky, though. You need to add a [custom_flatten](https://googlecloudplatform.github.io/magic-modules/develop/custom-code/#custom_flatten) for `handlers` - copy in the [existing generated code and all called funcs](https://github.com/modular-magician/terraform-provider-google-beta/blob/53298b5366f07841687f1ef3b52af0b3d52a7a46/google-beta/services/dialogflowcx/resource_dialogflow_cx_playbook.go#L2621-L2639). Then you can modify the `handlers` flattener to, for each item in the array, set the right value for the nested ignore_read field (including the correct handler index in the path passed to d.Get).

Opened https://github.com/hashicorp/terraform-provider-google/issues/23630 to track this issue.

melinath:
I think for the ImportStateVerify issues you need to use the entire path as it appears in the error:

```yaml
ignore_read_extra:
- "handlers.2.event_handler.0.fulfillment.0.advanced_settings.0.audio_export_gcs_destination.#"
- "handlers.2.event_handler.0.fulfillment.0.advanced_settings.0.audio_export_gcs_destination.0.%"
- "handlers.2.event_handler.0.fulfillment.0.advanced_settings.0.audio_export_gcs_destination.0.uri"
- "handlers.2.event_handler.0.fulfillment.0.advanced_settings.0.logging_settings.#"
- "handlers.2.event_handler.0.fulfillment.0.advanced_settings.0.logging_settings.0.%"
- "handlers.2.event_handler.0.fulfillment.0.advanced_settings.0.logging_settings.0.disable_logging"
- "handlers.2.event_handler.0.fulfillment.0.advanced_settings.0.logging_settings.0.enable_consent_based_redaction"
- "handlers.2.event_handler.0.fulfillment.0.advanced_settings.0.logging_settings.0.enable_interaction_logging"
- "handlers.2.event_handler.0.fulfillment.0.advanced_settings.0.logging_settings.0.enable_stackdriver_logging"
- "handlers.6.lifecycle_handler.0.fulfillment.0.advanced_settings.0.audio_export_gcs_destination.#"
- "handlers.6.lifecycle_handler.0.fulfillment.0.advanced_settings.0.audio_export_gcs_destination.0.%"
- "handlers.6.lifecycle_handler.0.fulfillment.0.advanced_settings.0.audio_export_gcs_destination.0.uri"
- "handlers.6.lifecycle_handler.0.fulfillment.0.advanced_settings.0.logging_settings.#"
- "handlers.6.lifecycle_handler.0.fulfillment.0.advanced_settings.0.logging_settings.0.%"
- "handlers.6.lifecycle_handler.0.fulfillment.0.advanced_settings.0.logging_settings.0.disable_logging"
- "handlers.6.lifecycle_handler.0.fulfillment.0.advanced_settings.0.logging_settings.0.enable_consent_based_redaction": "true","
- "handlers.6.lifecycle_handler.0.fulfillment.0.advanced_settings.0.logging_settings.0.enable_interaction_logging"
- "handlers.6.lifecycle_handler.0.fulfillment.0.advanced_settings.0.logging_settings.0.enable_stackdriver_logging"
```

melinath:
@zhihaos I was talking about this with @rileykarson earlier and he suggested using a [custom decoder](https://googlecloudplatform.github.io/magic-modules/develop/custom-code/#decoder) rather than a custom flattener. I didn't suggest this initially because you can't call d.Set for nested fields inside a decoder, but he pointed out that you can modify the API response instead to inject the values using d.Get (and then have the flattener treat it as "returned by the API"). The advantage of this approach is that you don't have to override all the descendant flatteners.

Apologies for not thinking of this last week. Moving forward with the custom flattener probably isn't a blocker if you'd like to keep with that for now (unless @rileykarson disagrees as reviewer) but the decoder would probably be more maintainable long-term.

zhihaos:
> @zhihaos I was talking about this with @rileykarson earlier and he suggested using a [custom decoder](https://googlecloudplatform.github.io/magic-modules/develop/custom-code/#decoder) rather than a custom flattener. I didn't suggest this initially because you can't call d.Set for nested fields inside a decoder, but he pointed out that you can modify the API response instead to inject the values using d.Get (and then have the flattener treat it as "returned by the API"). The advantage of this approach is that you don't have to override all the descendant flatteners.
> 
> Apologies for not thinking of this last week. Moving forward with the custom flattener probably isn't a blocker if you'd like to keep with that for now (unless @rileykarson disagrees as reviewer) but the decoder would probably be more maintainable long-term.

@melinath Thanks for the information! Do you have an example of using `custom decoder` to modify the response? I do find it hard to maintain for rest of the testing as there are more fields inside an array that need `ignore_read`.

melinath:
I don't think there's an exact example that matches what you'll want to do, but this one might be similar: https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/templates/terraform/decoders/memorystore_instance.go.tmpl

zhihaos:
@rileykarson the PR is ready for review. Thank you! And sorry for this extremely large PR.

josuehuamanzaz:
My apologies for jumping into this conversation. I was just curious to know if the Playbook feature will eventually be added to Terraform. We need it for an upcoming project. Sorry for the interruption, and thank you in advance.

josueelias9:
My apologies for jumping into this conversation. I was just curious to know if the Playbook feature will eventually be added to Terraform. We need it for an upcoming project. Sorry for the interruption, and thank you in advance.

zhihaos:
> My apologies for jumping into this conversation. I was just curious to know if the Playbook feature will eventually be added to Terraform. We need it for an upcoming project. Sorry for the interruption, and thank you in advance.

@josueelias9 yes, it is currently waiting for review.

mmv1/templates/terraform/decoders/dialogflowcx_playbook.go.tmpl
```diff
@@ -0,0 +1,366 @@
+///////////////////////////////////////////////////////////////////////////////////////////////////
+// Modify response for the following fields that are ignore_read (i.e. removed in the response): //
+///////////////////////////////////////////////////////////////////////////////////////////////////
+displayName, ok := res["displayName"].(string)
+if !ok { return nil, fmt.Errorf("'displayName' is not a string") }
+
+if displayName == "Playbook Example with Fulfillment" {
+	log.Printf("[DEBUG] Received response for Playbook %q: %#v", d.Id(), res)
+
+	//   - 'handlers.2.event_handler.0.fulfillment.0.advanced_settings.0.audio_export_gcs_destination'
```
rileykarson: It's unusual that we need to reference specific indices in a list like this. Does this mean that the service expects handlers to be sent in a specific sequence/order and validation will block sending a different ordering? I didn't see documentation suggesting as much at https://cloud.google.com/dialogflow/cx/docs/reference/rest/v3/projects.locations.agents.playbooks#Playbook.Handler or in the concepts pages in c.g.c (but may not have looked in the right spot)

mmv1/templates/terraform/decoders/dialogflowcx_playbook.go.tmpl
```diff
@@ -0,0 +1,366 @@
+///////////////////////////////////////////////////////////////////////////////////////////////////
+// Modify response for the following fields that are ignore_read (i.e. removed in the response): //
+///////////////////////////////////////////////////////////////////////////////////////////////////
+displayName, ok := res["displayName"].(string)
+if !ok { return nil, fmt.Errorf("'displayName' is not a string") }
+
+if displayName == "Playbook Example with Fulfillment" {
```
rileykarson: Oh wait- this looks like we're keyed off the specific name used in the test. Can you elaborate on why we need to do this? How would this work for a customer with a different name?

zhihaos: Yes, this is used for distinguish the examples as there are two examples, and the decoder is only for modifying the API response for one example, so the other example won't be affected by the decoder.

The decoder is used for the full version example because there are more fields that are covered in the full version example, and some of the fields are in the list and require `ignore_read: true`, so I modify its API response in the decoder.

mmv1/templates/terraform/decoders/dialogflowcx_playbook.go.tmpl
zhihaos: No, the service does not expect any ordering from the handlers. It is just that there are many handlers in this full version example (for 100% test coverage) and some handlers contains the fields that is `ignore_read: true` (i.e. not exist in the response), so I modified the API response for those fields (using the index) to make sure the test wont fail due to mismatch.

mmv1/templates/terraform/decoders/dialogflowcx_playbook.go.tmpl
rileykarson: We'll need to make the logic more generic to ensure support arbitrary customer configurations as well! Is there a meaningful spot we could split the PR to do the base resource now and handlers later? That could potentially speed up the initial resource and break out that review/iteration.

zhihaos: SG, removed.

rileykarson:
Updated changelog to the following for clarity for customers, lmk if you'd prefer otherwise!

> `google_dialogflow_cx_playbook` (initial basic support, full features to follow in a later release)

rileykarson (approved):

PR #14447, opened 2025-07-02
Update service account creation to prevent failures due to eventual consistency

chasevedder:
This changes service account creation to no longer require a final GetServiceAccount request to succeed. Instead, we will wait until we get 3 consecutive successful responses from GetServiceAccount before considering the create successful.

closes https://github.com/hashicorp/terraform-provider-google/issues/22546
```release-note:bug
iam: Update service account creation to prevent failures due to eventual consistency
```

### DIFF
--- a/mmv1/third_party/terraform/services/resourcemanager/resource_google_service_account.go
+++ b/mmv1/third_party/terraform/services/resourcemanager/resource_google_service_account.go
@@ -124,7 +124,7 @@ func resourceGoogleServiceAccountCreate(d *schema.ResourceData, meta interface{}
 
 	d.SetId(fmt.Sprintf("projects/%s/serviceAccounts/%s@%s.iam.gserviceaccount.com", project, aid, project))
 
-	iamClient := config.NewIamClient(userAgent) 
+	iamClient := config.NewIamClient(userAgent)
 	sa, err = iamClient.Projects.ServiceAccounts.Create("projects/"+project, r).Do()
 	if err != nil {
 		gerr, ok := err.(*googleapi.Error)
--- a/mmv1/third_party/terraform/services/resourcemanager/resource_google_service_account.go
+++ b/mmv1/third_party/terraform/services/resourcemanager/resource_google_service_account.go
@@ -124,14 +124,15 @@ func resourceGoogleServiceAccountCreate(d *schema.ResourceData, meta interface{}
 
 	d.SetId(fmt.Sprintf("projects/%s/serviceAccounts/%s@%s.iam.gserviceaccount.com", project, aid, project))
 
-	sa, err = config.NewIamClient(userAgent).Projects.ServiceAccounts.Create("projects/"+project, r).Do()
+	iamClient := config.NewIamClient(userAgent) 
+	sa, err = iamClient.Projects.ServiceAccounts.Create("projects/"+project, r).Do()
 	if err != nil {
 		gerr, ok := err.(*googleapi.Error)
 		alreadyExists := ok && gerr.Code == 409 && d.Get("create_ignore_already_exists").(bool)
 		if alreadyExists {
 			err = transport_tpg.Retry(transport_tpg.RetryOptions{
 				RetryFunc: func() (operr error) {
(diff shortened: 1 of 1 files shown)


mmv1/third_party/terraform/services/resourcemanager/resource_google_service_account.go
```diff
@@ -122,54 +122,71 @@ func resourceGoogleServiceAccountCreate(d *schema.ResourceData, meta interface{}
...
 	if err != nil {
 		gerr, ok := err.(*googleapi.Error)
 		alreadyExists := ok && gerr.Code == 409 && d.Get("create_ignore_already_exists").(bool)
 		if alreadyExists {
-			sa = &iam.ServiceAccount{
-				Name: fmt.Sprintf("projects/%s/serviceAccounts/%s@%s.iam.gserviceaccount.com", project, aid, project),
-			}
+			err = transport_tpg.Retry(transport_tpg.RetryOptions{
+				RetryFunc: func() (operr error) {
+					sa, saerr := config.NewIamClient(userAgent).Projects.ServiceAccounts.Get(d.Id()).Do()
```
trodge: Can this client be created outside the retry func?

mmv1/third_party/terraform/services/resourcemanager/resource_google_service_account.go
```diff
@@ -122,54 +122,71 @@ func resourceGoogleServiceAccountCreate(d *schema.ResourceData, meta interface{}
...
+		return fmt.Errorf("Error setting member: %s", err)
 	}
 
 	// We can't guarantee complete consistency even after polling,
 	// so sleep for some additional time to reduce the likelihood of
 	// eventual consistency failures.
 	time.Sleep(10 * time.Second)
 
-	return resourceGoogleServiceAccountRead(d, meta)
+	return nil
```
trodge, line 173: Was this change intentional? I think we still want to read the resource after creation to confirm there are no errors.

trodge: Oh I see, we're now calling it from `PollingWaitTime`.

Edit- or not, since we're not actually calling `populateResourceData` from `resourceServiceAccountPollRead`. What does this mean for server owned fields?

chasevedder: Yes this is intentional, we've come to this decision after discussion with @c2thorn . We do read the resource and wait for 3 consecutive successful responses on line 157. The final read was intended as a fail-safe to ensure that the service account was consistent, but if it fails it leaves the terraform in a broken state since the service account was created, but the resource wasn't actually present in the terraform resource.

We can't make any guarantees about eventual consistency, so the best we can do is wait for the 3 consecutive successes and then return success. Even if this were to result in downstream errors due to the eventual consistency of the service account, those failures would at least be retriable and leave terraform in a normal state.

> since we're not actually calling populateResourceData from resourceServiceAccountPollRead. What does this mean for server owned fields?

I believe all of the server owned fields are being set here, but I'll update this to use `populateResourceData` to make it clearer and less error prone. All of the data we need is in the response from CreateServiceAccount, so we don't actually need to call GetServiceAccount to populate any of the data, only to try waiting until the account is consistent in the backend.

mmv1/third_party/terraform/services/resourcemanager/resource_google_service_account.go
chasevedder: Updated

mmv1/third_party/terraform/services/resourcemanager/resource_google_service_account.go
```diff
@@ -122,54 +122,59 @@ func resourceGoogleServiceAccountCreate(d *schema.ResourceData, meta interface{}
 		ServiceAccount: sa,
 	}
 
-	sa, err = config.NewIamClient(userAgent).Projects.ServiceAccounts.Create("projects/"+project, r).Do()
+	d.SetId(fmt.Sprintf("projects/%s/serviceAccounts/%s@%s.iam.gserviceaccount.com", project, aid, project))
```
trodge, line 125: I missed this in review but setting the ID here means that it will remain set even if the resource fails to create, which causes our nightly tests to fail.

PR #14452, opened 2025-07-03
Feature gap: Implement `aggregateReservation` and `enableEmergentMaintenance`

Cezarus27:
This patch provides `aggregateReservation` fields to the `FutureReservation`. They are marked as `immutable` because of the error returned by API:
```
Error: Error updating FutureReservation "projects/<project>/zones/us-central1-a/futureReservations/tf-test-future-reservation": googleapi: Error 400: Invalid value for field 'updateMask': '"aggregateReservation"'. Invalid updateMask Path. Updating aggregate_reservation is not supported., invalid
```

```release-note:enhancement
compute: added `aggregate_reservation.0.vm_family`, `aggregate_reservation.0.reserved_resources.0.accelerator.0.accelerator_count`, `aggregate_reservation.0.reserved_resources.0.accelerator.0.accelerator_type` and `aggregate_reservation.0.workload_type` fields to`google_future_reservation` resource (beta)
```


mmv1/products/compute/FutureReservation.yaml
```diff
@@ -545,3 +561,62 @@ properties:
         type: Integer
         description: |
           Span of time that's a fraction of a second at nanosecond resolution. Durations less than one second are represented with a 0 seconds field and a positive nanos field. Must be from 0 to 999,999,999 inclusive.
+  - name: 'enableEmergentMaintenance'
+    type: Boolean
+    description: |
+      Indicates if this group of VMs have emergent maintenance enabled.
+    immutable: true
+    ignore_read: true
```
SirGitsalot: It doesn't look like this feature is fully rolled out yet, which is why `ignore_read` is necessary here - the attempt to set the field silently fails, so the field isn't returned in the response. This is also true for updates, but there the failure isn't silent, hence the need for `immutable`. Support for this field will need to be added in a future PR once the feature is complete (`aggregateReservation` is working just fine!)

SirGitsalot (changes requested):

mmv1/products/compute/FutureReservation.yaml
Cezarus27: Sure, I removed `enableEmergentMaintenance` from the PR.

PR #14454, opened 2025-07-03
Add deletion policy to Secure Source Manager instance

ericpang777:

```release-note:enhancement
securesourcemanager: added `deletion_policy` field to `google_secure_source_manager_instance` resource
```

### DIFF
--- a/mmv1/products/securesourcemanager/BranchRule.yaml
+++ b/mmv1/products/securesourcemanager/BranchRule.yaml
@@ -51,10 +51,10 @@ examples:
       deletion_policy: 'PREVENT'
     test_vars_overrides:
       'prevent_destroy': 'false'
-      'deletion_policy': 'DELETE'
+      'deletion_policy': '"DELETE"'
     oics_vars_overrides:
       'prevent_destroy': 'false'
-      'deletion_policy': 'DELETE'
+      'deletion_policy': '"DELETE"'
   - name: 'secure_source_manager_branch_rule_with_fields'
     primary_resource_id: 'default'
     vars:
@@ -62,13 +62,13 @@ examples:
       repository_id: 'my-initial-repository'
       instance_id: 'my-initial-instance'
       prevent_destroy: 'true'
-      deletion_policy: 'PREVENT'
+      deletion_policy: '"PREVENT"'
     test_vars_overrides:
       'prevent_destroy': 'false'
-      'deletion_policy': 'DELETE'
+      'deletion_policy': '"DELETE"'
     oics_vars_overrides:
       'prevent_destroy': 'false'
-      'deletion_policy': 'DELETE'
+      'deletion_policy': '"DELETE"'
 parameters:
   - name: 'branch_rule_id'
     type: String
--- a/mmv1/products/securesourcemanager/BranchRule.yaml
+++ b/mmv1/products/securesourcemanager/BranchRule.yaml
@@ -48,10 +48,13 @@ examples:
       repository_id: 'my-basic-repository'
       instance_id: 'my-basic-instance'
       prevent_destroy: 'true'
+      deletion_policy: 'PREVENT'
     test_vars_overrides:
       'prevent_destroy': 'false'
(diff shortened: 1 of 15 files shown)


ericpang777:
I was reading https://googlecloudplatform.github.io/magic-modules/develop/client-side-fields/#update-data-source and didn't understand what this means. Is there any action to be taken?

trodge:
> I was reading https://googlecloudplatform.github.io/magic-modules/develop/client-side-fields/#update-data-source and didn't understand what this means. Is there any action to be taken?

A data source is a read-only version of the resource to allow its fields to be referenced by other resources, but it doesn't look like there's one implemented for this resource yet, so no action to be taken there.

Currently we want to avoid adding `deletion_protection` fields with default value `false` because we plan on making `deletion_policy` (an enum which supports `PREVENT` `ABANDON` or `DELETE`) a universal field in the future.

In the mean time, given that we are in a [major release window](https://googlecloudplatform.github.io/magic-modules/breaking-changes/make-a-breaking-change/#in-the-700-major-release), there are two viable options, both involving targeting the 7.0.0 feature branch:

1. Add `deletion_policy` with default value `PREVENT`
2. Add `deletion_protection` with default value `true`

trodge (changes requested):
See above comment suggesting making this change in a major release.

Edit- it would also be helpful to use this PR to make `deletion_policy` available in a 6.X release with default value `DELETE` for an easier upgrade path for users.

ericpang777:
We would prefer `deletion_protection` with default `true`, so that sounds good to me. Looking into `deletion_policy`..

ericpang777:
Ok so do you mean that `deletion_protection` is a field that we should avoid adding now. And we should go with `deletion_policy`. For `deletion_policy`, we set the default to `DELETE` for 6.X releases, and make the breaking change to a default value of `PREVENT` (basically equivalent to `deletion_prevention: true`) for 7.X releases?

That sounds good to me then 👍 , I'll make the change here

trodge:
Yes, ideally add `deletion_policy` with default `DELETE` for 6.x and change default to `PREVENT` in 7.0

mmv1/products/securesourcemanager/BranchRule.yaml
```diff
@@ -48,21 +48,27 @@ examples:
...
       repository_id: 'my-initial-repository'
       instance_id: 'my-initial-instance'
       prevent_destroy: 'true'
+      deletion_policy: 'PREVENT'
     test_vars_overrides:
       'prevent_destroy': 'false'
+      'deletion_policy': 'DELETE'
     oics_vars_overrides:
       'prevent_destroy': 'false'
+      'deletion_policy': 'DELETE'
```
trodge:
```suggestion
      deletion_policy: '"PREVENT"'
    test_vars_overrides:
      'prevent_destroy': 'false'
      'deletion_policy': '"DELETE"'
    oics_vars_overrides:
      'prevent_destroy': 'false'
      'deletion_policy': '"DELETE"'
  - name: 'secure_source_manager_branch_rule_with_fields'
    primary_resource_id: 'default'
    vars:
      branch_rule_id: 'my-initial-branchrule'
      repository_id: 'my-initial-repository'
      instance_id: 'my-initial-instance'
      prevent_destroy: 'true'
      deletion_policy: '"PREVENT"'
    test_vars_overrides:
      'prevent_destroy': 'false'
      'deletion_policy': '"DELETE"'
    oics_vars_overrides:
      'prevent_destroy': 'false'
      'deletion_policy': '"DELETE"'
```

trodge: This should fix the build issue when applied to all 3 product yamls.

trodge (changes requested):

mmv1/products/securesourcemanager/BranchRule.yaml
ericpang777: Ahhh thank you so much!

PR #14459, opened 2025-07-04
Adding flexible webhook support;

zhihaos:
```release-note: enhancement
dialogflowcx: added `flexible` support to `google_dialogflow_cx_webhook` resource.
```

### DIFF
--- a/mmv1/products/dialogflowcx/Webhook.yaml
+++ b/mmv1/products/dialogflowcx/Webhook.yaml
@@ -258,6 +258,7 @@ properties:
                   The client secret provided by the 3rd party platform.  If the
                   `secret_version_for_client_secret` field is set, this field will be
                   ignored.
+                ignore_read: true
               - name: scopes
                 type: Array
                 description: The OAuth scopes to grant.
--- a/mmv1/products/dialogflowcx/Webhook.yaml
+++ b/mmv1/products/dialogflowcx/Webhook.yaml
@@ -327,3 +327,20 @@ properties:
             enum_values:
               - 'STANDARD'
               - 'FLEXIBLE'
+  - name: 'startFlow'
+    type: String
+    description: |
+      Deprecated. Name of the start flow in this agent. A start flow will be automatically created when the agent is created, and can only be deleted by deleting the agent. Format: projects/<Project ID>/locations/<Location ID>/agents/<Agent ID>/flows/<Flow ID>.
+    output: true
+  - name: 'securitySettings'
+    type: String
+    description: |
+      Deprecated. Name of the SecuritySettings reference for the agent. Format: projects/<Project ID>/locations/<Location ID>/securitySettings/<Security Settings ID>.
+  - name: 'enableStackdriverLogging'
+    type: Boolean
+    description: |
+      Deprecated. Determines whether this agent should log conversation queries.
+  - name: 'enableSpellCorrection'
+    type: Boolean
+    description: |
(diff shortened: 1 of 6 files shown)


zhihaos:
> Looks like this PR introduced couple breaking changes.
> 
> Are `uri` and `requestHeaders` fields removed from the API?

Yes, there are couple breaking changes. `uri` and `requestHeaders` are still in the API. The removed fields are: `enable_spell_correction` `enable_stackdriver_logging` `security_settings` and `start_flow`.

hao-nan-li:
> > Looks like this PR introduced couple breaking changes.
> > Are `uri` and `requestHeaders` fields removed from the API?
> 
> Yes, there are couple breaking changes. `uri` and `requestHeaders` are still in the API. The removed fields are: `enable_spell_correction` `enable_stackdriver_logging` `security_settings` and `start_flow`.

Thanks for confirming. The contribution window is open for major release, see https://googlecloudplatform.github.io/magic-modules/breaking-changes/make-a-breaking-change/#in-the-700-major-release

zhihaos:
@hao-nan-li I am adding back those fields and marking them as deprecated in the description, so that I can push the change sooner. wdyt?

hao-nan-li:
> @hao-nan-li I am adding back those fields and marking them as deprecated in the description, so that I can push the change sooner. wdyt?

Sounds good, as long as the change does not include any breaking-change.

zhihaos:
> Could you add a test for the added fields? Or add these fields to an existing test?

@hao-nan-li I added the test, it seems the test fails because the client secrets are not matched due to redaction. 
```
client_secret                    = "REDACTED" -> "projects/example-proj/secrets/example=secret/versions/example-version"
```
How can I ignore this other than skipping VCR?

hao-nan-li (commented):
The missing test report shows that some fields are not added to `google_dialogflow_cx_webhook`. I noticed that some of the fields are deprecated and shouldn't be tested, but could you double check if that's the case for the missing fields and add what's missing?

zhihaos:
> The missing test report shows that some fields are not added to `google_dialogflow_cx_webhook`. I noticed that some of the fields are deprecated and shouldn't be tested, but could you double check if that's the case for the missing fields and add what's missing?

Thanks for the review. As I checked, all fields that are deprecated are not in any tests before and now. I covered most of the fields, is it expected to reach 100% coverage?

hao-nan-li:
Yeah, we target 100% field coverage in the tests. In other word, every field should be tested at least once in their corresponding resource. We don't need to test all possible values for an existing field.

zhihaos:
@hao-nan-li Please review, I think it is all covered. Thanks!

mmv1/products/dialogflowcx/Webhook.yaml
```diff
@@ -71,27 +83,134 @@ properties:
...
+        type: KeyValuePairs
+        description: |-
+          Maps the values extracted from specific fields of the flexible webhook
+          response into session parameters.
+          - Key: session parameter name
+          - Value: field path in the webhook response
+      - name: requestBody
+        type: String
+        description: Defines a custom JSON object as request body to send to flexible webhook.
+      - name: requestHeaders
```
hao-nan-li, line 164: Just want to make sure, this field looks to be immutable previously, is removing `immutable` the right move?

zhihaos: Thanks for the catch. Yes, it is mutable now.

hao-nan-li (approved):

PR #14462, opened 2025-07-07
feat: support PSC outbound's network attachment field

kgala2:


mmv1/third_party/terraform/services/sql/resource_sql_database_instance.go.tmpl
```diff
@@ -1596,6 +1601,7 @@ func expandPscConfig(configured []interface{}) *sqladmin.PscConfig {
 		return &sqladmin.PscConfig{
 			PscEnabled:              _entry["psc_enabled"].(bool),
 			AllowedConsumerProjects: tpgresource.ConvertStringArr(_entry["allowed_consumer_projects"].(*schema.Set).List()),
+			NetworkAttachmentUri:	 _entry["network_attachment_uri"].(string),
```
rileykarson:
```suggestion
			NetworkAttachmentUri:    _entry["network_attachment_uri"].(string),
```

mmv1/third_party/terraform/services/sql/resource_sql_database_instance.go.tmpl
```diff
@@ -532,6 +532,11 @@ is set to true. Defaults to ZONAL.`,
 													Set:         schema.HashString,
 													Description: `List of consumer projects that are allow-listed for PSC connections to this instance. This instance can be connected to with PSC from any network in these projects. Each consumer project in this list may be represented by a project number (numeric) or by a project id (alphanumeric).`,
 												},
+												"network_attachment_uri": {
+													Type:			schema.TypeString,
+													Optional:		true,
+													Description:	`Name of network attachment resource used to authorize a producer service to connect a PSC interface to the consumer's VPC. For example: "projects/myProject/regions/myRegion/networkAttachments/myNetworkAttachment". This is required to enable outbound connection on a PSC instance.`,
```
rileykarson:
```suggestion
													Type:        schema.TypeString,
													Optional:    true,
													Description: `Name of network attachment resource used to authorize a producer service to connect a PSC interface to the consumer's VPC. For example: "projects/myProject/regions/myRegion/networkAttachments/myNetworkAttachment". This is required to enable outbound connection on a PSC instance.`,
```

mmv1/third_party/terraform/services/sql/resource_sql_database_instance.go.tmpl
```diff
@@ -2641,6 +2647,7 @@ func flattenPscConfigs(pscConfig *sqladmin.PscConfig) interface{} {
 	data := map[string]interface{}{
 		"psc_enabled":               pscConfig.PscEnabled,
 		"allowed_consumer_projects": schema.NewSet(schema.HashString, tpgresource.ConvertStringArrToInterface(pscConfig.AllowedConsumerProjects)),
+		"network_attachment_uri":	 pscConfig.NetworkAttachmentUri,
```
rileykarson:
```suggestion
		"network_attachment_uri":    pscConfig.NetworkAttachmentUri,
```

mmv1/third_party/terraform/services/sql/resource_sql_database_instance_test.go.tmpl
```diff
@@ -4843,6 +4890,49 @@ func verifyPscAutoConnectionsOperation(resourceName string, isPscConfigExpected
 	}
 }
 
+func verifyPscNetorkAttachmentOperation(resourceName string, isPscConfigExpected bool, expectedPscEnabled bool, expectedNetworkAttachmentUri string ) func(*terraform.State) error {
```
rileykarson, line 4926: note: This is mostly redundant with import's builtin ImportStateVerify logic- which should do nearly the same thing- but no harm in keeping it around!

mmv1/third_party/terraform/services/sql/resource_sql_database_instance_test.go.tmpl
```diff
@@ -1183,6 +1183,53 @@ func TestAccSqlDatabaseInstance_withPSCEnabled_withIpV4Enabled(t *testing.T) {
...
+	networkAttachmentUri := fmt.Sprintf("projects/%s/regions/%s/networkAttachments/%s", projectId, region, networkAttachmentName)
+
+
+	acctest.VcrTest(t, resource.TestCase{
+		PreCheck:                 func() { acctest.AccTestPreCheck(t) },
+		ProtoV5ProviderFactories: acctest.ProtoV5ProviderFactories(t),
+		CheckDestroy:             testAccSqlDatabaseInstanceDestroyProducer(t),
+		Steps: []resource.TestStep{
+			{
+				Config: testAccSqlDatabaseInstance_withPSCEnabled_withoutPscOutbound(instanceName),
```
rileykarson, line 1208: Are we able to create the resource w/ the setting, i.e. could start with `testAccSqlDatabaseInstance_withPSCEnabled_withNetworkAttachmentUri` directly? No need to actually do so if it works- the update case is more interesting and we're not getting enough new information to justify the carrying cost of the test, since it's an API responsibility- but if it _doesn't_ we may want special handling & a dedicated test to confirm that it works as intended.

kgala2: Adding a Network Attachment URI is an update operation on a PSC only instance, and we do not support NetworkAttachmentUri flag during instance creation.

rileykarson: We could optionally bridge over this with a `post_create` in Terraform and call the update immediately after creation when the value is specified. That would make recreation cases easier as you could do an all-in-one apply, provided it makes sense for the domain. The API supporting it and doing that serverside would be even better.

That's optional and non-blocking though; either way would be backwards compatible to add.

rileykarson (approved):

mmv1/third_party/terraform/services/sql/resource_sql_database_instance_test.go.tmpl
nasmart: The inability to specify the Network Attachment URI at the point of Cloud SQL instance creation appears to break what I believe is a general expectation of Terraform: The ability to specify a desired state, with dependencies (`depends_on`) where required, and run `terraform apply` **once** to achieve that state.

Unless I'm missing something, I currently need to follow this process to create a Cloud SQL instance with outbound PSC:

1. Declare a `google_sql_database_instance` resource with `psc_config.psc_enabled = true`
2. Declare a `google_compute_network_attachment` resource.
3. Run `terraform apply`.
4. Update the `google_sql_database_instance` resource with `psc_config.network_attachment_uri = google_compute_network_attachment.<network_attachment>.id`.
5. Run `terraform apply`.

Please let me know if I'm missing a way to achieve the desired state with a single `terraform apply`.

mmv1/third_party/terraform/services/sql/resource_sql_database_instance_test.go.tmpl
```diff
@@ -1183,6 +1183,86 @@ func TestAccSqlDatabaseInstance_withPSCEnabled_withIpV4Enabled(t *testing.T) {
...
+
+
+	acctest.VcrTest(t, resource.TestCase{
+		PreCheck:                 func() { acctest.AccTestPreCheck(t) },
+		ProtoV5ProviderFactories: acctest.ProtoV5ProviderFactories(t),
+		CheckDestroy:             testAccSqlDatabaseInstanceDestroyProducer(t),
+		Steps: []resource.TestStep{
+			{
+				Config: testAccSqlDatabaseInstance_withPSCEnabled_withNetworkAttachmentUri(instanceName, networkAttachmentUri),
+				ExpectError: regexp.MustCompile(`.*Network attachment used for Private Service Connect interfaces can not be assigned with instance creation.*`),
```
nasmart, line 1260: Nit: "can not" should be one work, i.e. "cannot".

mmv1/third_party/terraform/services/sql/resource_sql_database_instance_test.go.tmpl
rileykarson: This is correct! A post-create is how we'd support this in the provider unless there's a circular dependency in which case we'd need a different approach.

PR #14470, opened 2025-07-08
alloydb: Revert incorrect removal of machine_type from machine_config (it is GA)

jahnavi2k:
Restored support for the machine_type field in the machine_config block. The field was previously removed due to an incorrect assumption that it was not GA.  While the field is GA, the documentation had not yet reflected this; it is currently being updated.

```release-note:bug
Restored support for the machine_type field in the machine_config block. It is GA.
```

### DIFF
--- a/mmv1/products/alloydb/Instance.yaml
+++ b/mmv1/products/alloydb/Instance.yaml
@@ -257,7 +257,6 @@ properties:
   - name: 'observabilityConfig'
     type: NestedObject
     description: 'Configuration for enhanced query insights.'
-    min_version: 'beta'
     default_from_api: true
     properties:
       - name: 'enabled'
--- a/mmv1/third_party/terraform/services/alloydb/resource_alloydb_instance_test.go.tmpl
+++ b/mmv1/third_party/terraform/services/alloydb/resource_alloydb_instance_test.go.tmpl
@@ -81,9 +81,7 @@ resource "google_alloydb_instance" "default" {
 
   machine_config {
     cpu_count = 4
-	{{- if ne $.TargetVersionName "ga" }}
     machine_type = "n2-highmem-4"
-    {{ end }}
   }
 
   labels = {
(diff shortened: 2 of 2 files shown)


trodge:
This looks correct to me, but I'm not sure how to confirm that https://github.com/hashicorp/terraform-provider-google/issues/23383 won't re-surface.

sands-123:
Hey @trodge 

Thanks for taking a look at this one.
Let us confirm with the author of https://github.com/hashicorp/terraform-provider-google/issues/23383 to understand how they were able to reproduce that issue ?

Looking at the description, they tried to update an existing "read-pool instance of 8 vCPU" via terraform.
We need to understand how the read-pool instance was initially created to be able to troubleshoot it further.

q2w (approved):

jahnavi2k:
Hi @trodge 

The machine_type field was introduced in v6.31.0, whereas in the issue, the author used v6.3.0.
I have verified that the issue doesn't come up with versions >= 6.31. Updating cpu count/machine_type works as expected for both primary and read instances. 
@sands-123 is also trying to reproduce the issue to cross verify!

sands-123:
Confirming that the issue observed in https://github.com/hashicorp/terraform-provider-google/issues/23383 occurred because they were using version v 6.3

Support for machine-type was added in v 6.31.0 and I can verify this works as intended

trodge (approved):

sammy-hughes:
> This looks correct to me, but I'm not sure how to confirm that [hashicorp/terraform-provider-google#23383](https://github.com/hashicorp/terraform-provider-google/issues/23383) won't re-surface.

I believe this is perfectly safe, and do not expect it to cause future issues. As @jahnavi2k and @sands-123 noted, our team was using an outdated version of the google-provider plugin. That is more likely the cause of our issue, rather than feature status. I can confirm that the changelog shows that our issue was corrected as of April 2025, in version 6.31.0.

PR #14476, opened 2025-07-08
adding all the release diff testing logic

luckyswaminathan:
As a part of the addition of a weekly diff test run, I have opened a PR that has all the diff-testing logic. This is in conjunction with the proposal: https://docs.google.com/document/d/1c7gbM1A-ugtZXcrtQ0CwEBqRjwc4xJdvWJ9jk6KB-XI/edit?resourcekey=0-FsWNtDZ9uI3hT05LwfFFCw&tab=t.0#heading=h.xzptrog8pyxf 

```release-note:none
added augmented RELEASE_DIFF logic and extra testing, which affects no functionality since diff tests are not used
```

### DIFF
--- a/mmv1/third_party/terraform/acctest/vcr_utils.go
+++ b/mmv1/third_party/terraform/acctest/vcr_utils.go
@@ -401,7 +401,7 @@ func writeOutputFileDeferFunction(tempFile *os.File, failed bool) {
 	if failed {
 		// Check if the output line starts with "[Diff]"
 		if isDiff {
-			fmt.Fprintf(os.Stdout, "Breaking Change Detected \n")
+			fmt.Fprintf(os.Stdout, "[Diff] Breaking Change Detected] \n")
 			fmt.Fprintf(diffFailureFile, "[Diff] %s\n", testOutput)
 		} else {
 			fmt.Fprintf(regularFailureFile, testOutput)
--- a/mmv1/third_party/terraform/acctest/vcr_utils.go
+++ b/mmv1/third_party/terraform/acctest/vcr_utils.go
@@ -335,32 +335,36 @@ func ReformConfigWithProvider(config, provider string) string {
 // parseReleaseDiffOutput reads the temporary file created during the release diff test and returns the last line of output
 // This is useful for extracting the diff output from the file after the test has run
 
-func ParseReleaseDiffOutput(temp *os.File) (string, string, error) {
+func ParseReleaseDiffOutput(temp *os.File) (bool, string, error) {
+	var isDiff = false
 	if temp == nil {
-		return "", "", errors.New("temporary file is nil")
+		return isDiff, "", errors.New("temporary file is nil")
 	}
 	_, err := temp.Seek(0, io.SeekStart)
 	if err != nil {
-		return "", "", fmt.Errorf("failed to seek to beginning of temporary file: %w", err)
+		return isDiff, "", fmt.Errorf("failed to seek to beginning of temporary file: %w", err)
 	}
 	var lastLine string
-	var allContentBuffer bytes.Buffer
+	var testOutputBuffer bytes.Buffer
 	scanner := bufio.NewScanner(temp)
 
 	for scanner.Scan() {
 		line := scanner.Text()
 		lastLine = line
-		allContentBuffer.WriteString(line)
-		allContentBuffer.WriteByte('\n')
+		testOutputBuffer.WriteString(line)
+		testOutputBuffer.WriteByte('\n')
+	}
+	if strings.HasPrefix(lastLine, "[Diff]") {
+		isDiff = true
 	}
 	if err := scanner.Err(); err != nil {
-		return "", "", fmt.Errorf("error reading temporary file: %w", err)
+		return isDiff, "", fmt.Errorf("error reading temporary file: %w", err)
 	}
-	allContent := allContentBuffer.String()
-	if len(allContent) > 0 && allContent[len(allContent)-1] == '\n' {
-		allContent = allContent[:len(allContent)-1]
+	testOutput := testOutputBuffer.String()
+	if len(testOutput) > 0 && testOutput[len(testOutput)-1] == '\n' {
+		testOutput = testOutput[:len(testOutput)-1]
 	}
-	return lastLine, allContent, nil
+	return isDiff, testOutput, nil
 }
 
 func writeOutputFileDeferFunction(tempFile *os.File, failed bool) {
@@ -379,17 +383,17 @@ func writeOutputFileDeferFunction(tempFile *os.File, failed bool) {
 		fmt.Fprintf(os.Stderr, "Error creating file: %v\n", err)
 		return
 	}
-	var lastLine, output, error = ParseReleaseDiffOutput(tempFile)
+	var isDiff, testOutput, error = ParseReleaseDiffOutput(tempFile)
 	if error != nil {
 		fmt.Fprintf(os.Stderr, "Error parsing release diff output: %v\n", err)
 	} else if failed {
 		// Check if the output line starts with "[Diff]"
-		if strings.HasPrefix(lastLine, "[Diff]") {
-			fmt.Fprintf(os.Stdout, "Breaking Change at \n%s\n", lastLine)
-			fmt.Fprintf(diffFailureFile, "[Diff] %s\n", output)
+		if isDiff {
+			fmt.Fprintf(os.Stdout, "Breaking Change Detected \n")
+			fmt.Fprintf(diffFailureFile, "[Diff] %s\n", testOutput)
 		} else {
-			fmt.Fprintf(regularFailureFile, output)
-			fmt.Fprintf(regularFailureFile, "FAILED --- %s\n", lastLine)
+			fmt.Fprintf(regularFailureFile, testOutput)
+			fmt.Fprintf(regularFailureFile, "FAILED --- %s\n", testOutput)
 		}
 	}
 	tempFile.Close() // Close the file handle
--- a/mmv1/third_party/terraform/acctest/vcr_utils.go
+++ b/mmv1/third_party/terraform/acctest/vcr_utils.go
@@ -149,6 +149,7 @@ func VcrTest(t *testing.T, c resource.TestCase) {
 		defer closeRecorder(t)
 	} else if isReleaseDiffEnabled() {
 		// creates temporary file for the individual test, will be a temporary to store the output
+		// this will only be used by the diff weekly teamcity project
 		temp_file, error := os.CreateTemp("", "release_diff_test_output_*.log")
 
 		if error != nil {
@@ -257,7 +258,7 @@ func initializeReleaseDiffTest(c resource.TestCase, testName string, temp_file *
 			// TODO: make a github action to get most recent release + current head, this is not a fix just for testing
 			// this should not be hardcoded
 			releaseProvider: {
-				VersionConstraint: "= 6.33.0", // if left empty fetches most recent release provider, which is actually optimal
+				// if left empty fetches most recent release provider, which is actually optimal
 			},
 		}
 	}
--- a/mmv1/third_party/terraform/acctest/vcr_utils.go
+++ b/mmv1/third_party/terraform/acctest/vcr_utils.go
@@ -281,7 +281,6 @@ func InsertDiffSteps(c resource.TestCase, tempOutputFile *os.File, releaseProvid
 	var replacementSteps []resource.TestStep
 	for _, testStep := range c.Steps {
 		countSteps++
-		// todo: add preconfig - categorize test failures (add flag to steps that if they fail is a diff failure)
 		if testStep.Config != "" {
 			ogConfig := testStep.Config
 			fmt.Fprintf(tempOutputFile, "[DEBUG] Original config: %s\n", ogConfig)
--- a/mmv1/third_party/terraform/acctest/vcr_utils.go
+++ b/mmv1/third_party/terraform/acctest/vcr_utils.go
@@ -1,11 +1,13 @@
 package acctest
 
 import (
+	"bufio"
 	"bytes"
 	"context"
 	"encoding/json"
 	"errors"
 	"fmt"
+	"io"
 	"io/ioutil"
 	"log"
 	"math/rand"
@@ -141,14 +143,41 @@ func vcrFileName(name string) string {
 // Can be called when VCR is not enabled, and it will behave as normal
 func VcrTest(t *testing.T, c resource.TestCase) {
 	t.Helper()
-
 	if IsVcrEnabled() {
 		defer closeRecorder(t)
 	} else if isReleaseDiffEnabled() {
-		c = initializeReleaseDiffTest(c, t.Name())
-	}
+		// creates temporary file for the individual test, will be a temporary to store the output
+		temp_file, error := os.CreateTemp("", "release_diff_test_output_*.log")
 
-	c = extendWithTGCData(t, c)
+		if error != nil {
+			t.Fatalf("Error creating temporary file: %v", error)
+			return
+		}
+		var dir, _ = os.Getwd()
+		fmt.Printf("current wd: %s\n", dir)
+
+		var regularFailureFile, err = os.Create(filepath.Join("../../../", "regular_failure_file.log"))
+		if err != nil {
+			fmt.Fprintf(os.Stderr, "Error creating file: %v\n", err)
+			return
+		}
+		var diffFailureFile *os.File
+		// todo = get rid of this hardcoding once again is just meant for visibility
+		diffFailureFile, err = os.Create(filepath.Join("../../../", "diff_failure_file.log"))
+		if err != nil {
+			fmt.Fprintf(os.Stderr, "Error creating file: %v\n", err)
+			return
+		}
+		fmt.Printf("Temporary file created at: %s\n", temp_file.Name())
+		fmt.Printf("Regular failure file created at: %s\n", regularFailureFile.Name())
+		fmt.Printf("Diff failure file created at: %s\n", diffFailureFile.Name())
+
+		defer func() {
+			writeOutputFileDeferFunction(t, temp_file, diffFailureFile, regularFailureFile, t.Failed())
+		}()
+		c = initializeReleaseDiffTest(c, t.Name(), temp_file)
+
+	}
 
 	// terraform_labels is a computed field to which "goog-terraform-provisioned": "true" is always
 	// added by the provider. ImportStateVerify "checks for strict equality and does not respect
@@ -164,6 +193,7 @@ func VcrTest(t *testing.T, c resource.TestCase) {
 	c.Steps = steps
 
 	resource.Test(t, c)
+
 }
 
 // We need to explicitly close the VCR recorder to save the cassette
@@ -207,7 +237,7 @@ func isReleaseDiffEnabled() bool {
 	return releaseDiff != ""
 }
 
-func initializeReleaseDiffTest(c resource.TestCase, testName string) resource.TestCase {
+func initializeReleaseDiffTest(c resource.TestCase, testName string, temp_file *os.File) resource.TestCase {
 	var releaseProvider string
 	packagePath := fmt.Sprint(reflect.TypeOf(transport_tpg.Config{}).PkgPath())
 	if strings.Contains(packagePath, "google-beta") {
@@ -220,10 +250,15 @@ func initializeReleaseDiffTest(c resource.TestCase, testName string) resource.Te
 		c.ExternalProviders[releaseProvider] = resource.ExternalProvider{}
 	} else {
 		c.ExternalProviders = map[string]resource.ExternalProvider{
-			releaseProvider: {},
+			// TODO: make a github action to get most recent release + current head, this is not a fix just for testing
+			// this should not be hardcoded
+			releaseProvider: {
+				VersionConstraint: "= 6.33.0", // if left empty fetches most recent release provider, which is actually optimal
+			},
 		}
 	}
 
+	// create files to config this step(flag), do this in temp file created within this method? preconfig/postconfig
 	localProviderName := "google-local"
 	if c.Providers != nil {
 		c.Providers = map[string]*schema.Provider{
@@ -243,13 +278,35 @@ func initializeReleaseDiffTest(c resource.TestCase, testName string) resource.Te
 		}
 	}
 
+	c.Steps = InsertDiffSteps(c, temp_file, releaseProvider, localProviderName)
+
+	return c
+}
+
+// InsertDiffSteps inserts a new step into the test case that reformats the config to use the release provider - this allows us to see the diff
+// between the local provider and the release provider. for a certain test, this will be used to see the diff between the local provider and the release provider.
+func InsertDiffSteps(c resource.TestCase, temp_file *os.File, releaseProvider string, localProviderName string) []resource.TestStep {
+	var countSteps = 0
+
 	var replacementSteps []resource.TestStep
 	for _, testStep := range c.Steps {
+		// todo: add preconfig - categorize test failures (add flag to steps that if they fail is a diff failure)
 		if testStep.Config != "" {
 			ogConfig := testStep.Config
+			fmt.Fprintf(os.Stdout, "Original config: %s\n", ogConfig)
+			fmt.Fprintf(os.Stdout, "Reformatting config to use provider: %s\n", localProviderName)
 			testStep.Config = ReformConfigWithProvider(ogConfig, localProviderName)
-			if testStep.ExpectError == nil && testStep.PlanOnly == false {
+			fmt.Fprintf(os.Stdout, "Reformatted config: %s\n", testStep.Config)
+			testStep.PreConfig = func() {
+				fmt.Fprintf(temp_file, "[Diff] Step %d\n", countSteps)
+				countSteps++
+			}
+			if testStep.ExpectError == nil && !testStep.PlanOnly {
 				newStep := resource.TestStep{
+					PreConfig: func() {
+						fmt.Fprintf(temp_file, "Step %d\n", countSteps)
+						countSteps++
+					},
 					Config: ReformConfigWithProvider(ogConfig, releaseProvider),
 				}
 				testStep.PlanOnly = true
@@ -261,11 +318,38 @@ func initializeReleaseDiffTest(c resource.TestCase, testName string) resource.Te
 			replacementSteps = append(replacementSteps, testStep)
 		}
 	}
-
-	c.Steps = replacementSteps
-
-	return c
+	return replacementSteps
 }
+
+func writeOutputFileDeferFunction(t *testing.T, temp_file *os.File, regularFailureFile *os.File, diffFailureFile *os.File, failed bool) {
+	if temp_file != nil {
+		// parses the temporary file created during the release diff test and returns the last line of output
+		// This is useful for extracting the diff output from the file after the test has run
+
+		var output, err = ParseReleaseDiffOutput(temp_file)
+		if err != nil {
+			fmt.Fprintf(os.Stderr, "Error parsing release diff output: %v\n", err)
+		} else if failed {
+			// Check if the output line starts with "[Diff]"
+			if diffRegexpMatch.MatchString(output) {
+				fmt.Printf(output)
+				fmt.Fprintf(diffFailureFile, "%s\n", output)
+			} else {
+				fmt.Fprintf(regularFailureFile, "%s\n", output)
+			}
+		}
+		temp_file.Close() // Close the file handle
+		err = os.Remove(temp_file.Name())
+		if err != nil {
+			fmt.Fprintf(os.Stdout, "Test passed, but failed to delete log file: %v\n", err)
+		}
+	}
+}
+
+// reformConfigWithProvider reformats the config to use the given provider
+// The method matches a regex for the provider block and replaces it with the given provider.
+// For example: ' data "google_compute_network" "default" { provider = "google-local" } '
+// will be reformatted to ' data "google_compute_network" "default" { provider = "google-beta" } '
 
 func ReformConfigWithProvider(config, provider string) string {
 	configBytes := []byte(config)
@@ -285,6 +369,31 @@ func ReformConfigWithProvider(config, provider string) string {
 
 	resourceHeader := regexp.MustCompile(`((resource|data) .*google_.* .*\w+.*\{ *)`)
 	return string(resourceHeader.ReplaceAll(configBytes, providerReplacementBytes))
+}
+
+func ParseReleaseDiffOutput(temp *os.File) (string, error) {
+	if temp == nil {
+		return "", errors.New("temporary file is nil")
+	}
+
+	_, err := temp.Seek(0, io.SeekStart)
+	if err != nil {
+		return "", fmt.Errorf("failed to seek to beginning of temporary file: %w", err)
+	}
+
+	var lastLine string
+	scanner := bufio.NewScanner(temp)
+
+	for scanner.Scan() {
+		line := scanner.Text()
+		lastLine = line
+	}
+
+	if err := scanner.Err(); err != nil {
+		return "", fmt.Errorf("error reading temporary file: %w", err)
+	}
+
+	return lastLine, nil
 }
 
 // HandleVCRConfiguration configures the recorder (github.com/dnaeon/go-vcr/recorder) used in the VCR test
--- a/mmv1/third_party/terraform/acctest/vcr_utils.go
+++ b/mmv1/third_party/terraform/acctest/vcr_utils.go
@@ -47,6 +47,8 @@ func IsVcrEnabled() bool {
 	return envPath != "" && vcrMode != ""
 }
 
+var diffRegexpMatch = regexp.MustCompile(`^\[Diff\]`)
+
 var configsLock = sync.RWMutex{}
 var sourcesLock = sync.RWMutex{}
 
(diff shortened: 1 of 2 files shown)


hao-nan-li:
Adding @roaks3 and @trodge as reviewers as you were contributors to the doc.

mmv1/third_party/terraform/acctest/vcr_utils.go
```diff
@@ -141,11 +145,41 @@ func vcrFileName(name string) string {
 // Can be called when VCR is not enabled, and it will behave as normal
 func VcrTest(t *testing.T, c resource.TestCase) {
 	t.Helper()
-
 	if IsVcrEnabled() {
 		defer closeRecorder(t)
 	} else if isReleaseDiffEnabled() {
-		c = initializeReleaseDiffTest(c, t.Name())
+		// creates temporary file for the individual test, will be a temporary to store the output
```
trodge, line 151: nit: this could go in a helper method

mmv1/third_party/terraform/acctest/vcr_utils_test.go
```diff
@@ -449,3 +451,111 @@ func TestReformConfigWithProvider(t *testing.T) {
...
+	defer os.Remove(temp_file.Name())
+	// Write some dummy data to the temp file
+	_, err = temp_file.WriteString("This is a test release diff output.\n")
+	if err != nil {
+		t.Fatalf("Failed to write to temp file: %v", err)
+	}
+
+	var expectedOutput = "This is a test release diff output."
+	var output string
+	output, err = acctest.ParseReleaseDiffOutput(temp_file)
```
trodge:
```suggestion
	lastLine, output, err = acctest.ParseReleaseDiffOutput(temp_file)
```

trodge (approved):

mmv1/third_party/terraform/acctest/vcr_utils.go
```diff
@@ -164,6 +173,31 @@ func VcrTest(t *testing.T, c resource.TestCase) {
 	c.Steps = steps
 
 	resource.Test(t, c)
+
+}
+
+func createFilesForReleaseDiffTest() (*os.File, *os.File, *os.File) {
```
roaks3: I would consider tweaking this function, since returning 3 file handles is a bit unusual and non-obvious. Do you actually need to create the failure files upfront, or could that be done in `writeOutputFileDeferFunction`?

mmv1/third_party/terraform/acctest/vcr_utils.go
```diff
@@ -164,6 +173,31 @@ func VcrTest(t *testing.T, c resource.TestCase) {
...
+	// creates temporary file for the individual test, will be a temporary to store the output
+	var tempFile *os.File
+	var err error
+	tempFile, err = os.CreateTemp("", "release_diff_test_output_*.log")
+	if err != nil {
+		log.Fatalf("Error creating temporary file: %v", err)
```
roaks3: It might make more sense to throw a regular error here, as part of the return value. Same for the other files as well I would assume, since it seems like the diff test results can't effectively be consumed without them.

mmv1/third_party/terraform/acctest/vcr_utils.go
```diff
@@ -207,7 +241,7 @@ func isReleaseDiffEnabled() bool {
 	return releaseDiff != ""
 }
 
-func initializeReleaseDiffTest(c resource.TestCase, testName string) resource.TestCase {
+func initializeReleaseDiffTest(c resource.TestCase, testName string, tempFile *os.File) resource.TestCase {
```
roaks3: Maybe `tempOutputFile`? Only because `tempFile` is pretty generic.

mmv1/third_party/terraform/acctest/vcr_utils.go
```diff
@@ -243,13 +282,35 @@ func initializeReleaseDiffTest(c resource.TestCase, testName string) resource.Te
...
 	}
 
+	c.Steps = InsertDiffSteps(c, tempFile, releaseProvider, localProviderName)
+
+	return c
+}
+
+// InsertDiffSteps inserts a new step into the test case that reformats the config to use the release provider - this allows us to see the diff
+// between the local provider and the release provider. for a certain test, this will be used to see the diff between the local provider and the release provider.
+func InsertDiffSteps(c resource.TestCase, tempFile *os.File, releaseProvider string, localProviderName string) []resource.TestStep {
```
roaks3: Nit: could pass in `c.Steps` to make it slightly more clear (ie. original list of steps is passed in, new set of steps is returned)

mmv1/third_party/terraform/acctest/vcr_utils.go
```diff
@@ -243,13 +282,35 @@ func initializeReleaseDiffTest(c resource.TestCase, testName string) resource.Te
...
 		if testStep.Config != "" {
 			ogConfig := testStep.Config
+			fmt.Fprintf(tempFile, "Original config: %s\n", ogConfig)
+			fmt.Fprintf(tempFile, "Reformatting config to use provider: %s\n", localProviderName)
 			testStep.Config = ReformConfigWithProvider(ogConfig, localProviderName)
-			if testStep.ExpectError == nil && testStep.PlanOnly == false {
+			fmt.Fprintf(tempFile, "Reformatted config: %s\n", testStep.Config)
+			testStep.PreConfig = func() {
+				fmt.Fprintf(tempFile, "[Diff] Step %d\n", countSteps)
+				countSteps++
```
roaks3: I think you may want this incremented as the first line under the for loop, but could you confirm the behavior? I think this will basically count each of the new steps, which seems like it would be of limited use. If if was the count of the step in the original test, then I think it would be more useful as a lookup of sorts for someone to go back to the test source code and look for where the problem occurred.

mmv1/third_party/terraform/acctest/vcr_utils.go
```diff
@@ -287,6 +345,73 @@ func ReformConfigWithProvider(config, provider string) string {
...
+
+	return lastLine, allContent, nil
+}
+
+func writeOutputFileDeferFunction(tempFile *os.File, regularFailureFile *os.File, diffFailureFile *os.File, failed bool) {
+	if tempFile != nil {
+		// parses the temporary file created during the release diff test and returns the last line of output
+		// This is useful for extracting the diff output from the file after the test has run
+
+		var lastLine, output, err = ParseReleaseDiffOutput(tempFile)
```
roaks3: I would consider rearranging the logic a bit for clarity and so you don't need to pass around as many variables. One option would be to first read the temp file to a string, then have an `isDiffStep` function that returns a boolean, then write to the failure file based on that result.

mmv1/third_party/terraform/acctest/vcr_utils.go
```diff
@@ -287,6 +345,73 @@ func ReformConfigWithProvider(config, provider string) string {
...
+
+	if len(allContent) > 0 && allContent[len(allContent)-1] == '\n' {
+		allContent = allContent[:len(allContent)-1]
+	}
+
+	return lastLine, allContent, nil
+}
+
+func writeOutputFileDeferFunction(tempFile *os.File, regularFailureFile *os.File, diffFailureFile *os.File, failed bool) {
+	if tempFile != nil {
```
roaks3: Go favors go/go-style/decisions#indent-error-flow, so you'd write this in the reverse like `if tempFile == nil return`

mmv1/third_party/terraform/acctest/vcr_utils.go
```diff
@@ -287,6 +345,73 @@ func ReformConfigWithProvider(config, provider string) string {
...
+		// parses the temporary file created during the release diff test and returns the last line of output
+		// This is useful for extracting the diff output from the file after the test has run
+
+		var lastLine, output, err = ParseReleaseDiffOutput(tempFile)
+		if err != nil {
+			fmt.Fprintf(os.Stderr, "Error parsing release diff output: %v\n", err)
+		} else if failed {
+			// Check if the output line starts with "[Diff]"
+			if diffRegexpMatch.MatchString(lastLine) {
```
roaks3: You might be able to use https://pkg.go.dev/strings#HasPrefix instead, and `[Diff]` could be a constant that is defined at the package level (also used in `InsertDiffSteps`)

mmv1/third_party/terraform/acctest/vcr_utils_test.go
```diff
@@ -449,3 +451,112 @@ func TestReformConfigWithProvider(t *testing.T) {
 		})
 	}
 }
+
+func TestDiffTestStepInjection(t *testing.T) {
```
roaks3: I couldn't find a reference, but I believe it is a Go convention to name tests `Test[FuncName]`, so I think this should be `TestInsertDiffSteps`

roaks3 (changes requested):
I would suggest one more pass on this to clean up the code and make it a bit easier to understand (however, some of this is of course subjective).

1. Some of the newline changes could be cleaned up. Unless you find the changes particularly more helpful for reading, I would generally avoid arbitrary newline changes, and I also saw a few places where there was a trailing newline within a block, which is a bit unusual stylistically.
2. There were also some comments that appeared to be out of place (reformConfigWithProvider comment) or might not make sense to check in (like TODOs). It would probably be good to look those over and make sure they're how you want for pushing to main.
3. I found some function / parameter names a bit hard to follow in places, in that they seem to describe what they literally are as opposed to their purpose. For example, `writeOutputFileDeferFunction` is somewhat opaque when I read it from the caller function. It might alternatively be named `writeTestRunnerOutput` or `saveTestRunnerOutput` (which also isolates the details of the function a bit)

mmv1/third_party/terraform/acctest/vcr_utils.go
luckyswaminathan: yeah this makes sense

mmv1/third_party/terraform/acctest/vcr_utils.go
luckyswaminathan: I don't think that passing c.Steps would really do anything here - probably easier to just add a comment that describes it

mmv1/third_party/terraform/acctest/vcr_utils.go
```diff
@@ -141,15 +143,19 @@ func vcrFileName(name string) string {
 // Can be called when VCR is not enabled, and it will behave as normal
 func VcrTest(t *testing.T, c resource.TestCase) {
 	t.Helper()
-
```
roaks3: What I meant regarding whitespace was not that it should be eliminated, but just that I wouldn't change areas of the code that are unrelated to your work. It basically creates a level of noise in both the PR and the git blame/history, and we'll inevitably have people changing things back and forth as preferences vary.

I would remove the newline changes from your changeset that aren't part of the new work (like this line, line 152, etc).

mmv1/third_party/terraform/acctest/vcr_utils.go
```diff
@@ -164,6 +170,20 @@ func VcrTest(t *testing.T, c resource.TestCase) {
 	c.Steps = steps
 
 	resource.Test(t, c)
+
```
roaks3: Is there a reason for adding this extra newline? It seems a bit odd/unconventional.

mmv1/third_party/terraform/acctest/vcr_utils.go
```diff
@@ -164,6 +170,20 @@ func VcrTest(t *testing.T, c resource.TestCase) {
...
+}
+
+func createTemporaryFile() *os.File {
+	// creates temporary file for the individual test, will be a temporary to store the output
+	var tempOutputFile *os.File
+	var err error
+	tempOutputFile, err = os.CreateTemp("", "release_diff_test_output_*.log")
```
roaks3: Nit: can be `:=` and then you don't need the var statements above

mmv1/third_party/terraform/acctest/vcr_utils.go
```diff
@@ -164,6 +170,20 @@ func VcrTest(t *testing.T, c resource.TestCase) {
...
+	if err != nil {
+		fmt.Fprintf(os.Stdout, "Error creating temporary file: %v", err)
+		return nil
```
roaks3: Is there a reason we aren't returning (*os.File, error), and returning an error here? I think the code will fail elsewhere if we allow it to continue without a temp file, like when writing to it.

mmv1/third_party/terraform/acctest/vcr_utils.go
roaks3: Nit: could still be changed if you agree here

mmv1/third_party/terraform/acctest/vcr_utils.go
```diff
@@ -220,10 +240,13 @@ func initializeReleaseDiffTest(c resource.TestCase, testName string) resource.Te
 		c.ExternalProviders[releaseProvider] = resource.ExternalProvider{}
 	} else {
 		c.ExternalProviders = map[string]resource.ExternalProvider{
-			releaseProvider: {},
+			releaseProvider: {
+				// if left empty fetches most recent release provider, which is actually optimal
```
roaks3: `// if left empty fetches most recent release provider` might be enough here, stating that it's optimal might be confusing to an arbitrary reader

mmv1/third_party/terraform/acctest/vcr_utils.go
```diff
@@ -220,10 +240,13 @@ func initializeReleaseDiffTest(c resource.TestCase, testName string) resource.Te
...
+			},
 		}
 	}
 
+	// create files to config this step(flag), do this in temp file created within this method? preconfig/postconfig
```
roaks3: This seems like maybe it shouldn't be pushed to main, or otherwise could be made more clear

mmv1/third_party/terraform/acctest/vcr_utils.go
```diff
@@ -242,14 +265,37 @@ func initializeReleaseDiffTest(c resource.TestCase, testName string) resource.Te
...
 	}
+	// InsertDiffSteps adds modified steps to the test that run with an external provider
+	// these steps do the actual infrastructure provisioning, and c.Steps is updated to have these modified steps
+	c.Steps = InsertDiffSteps(c, tempFile, releaseProvider, localProviderName)
+
+	return c
+}
+
+// InsertDiffSteps inserts a new step into the test case that reformats the config to use the release provider - this allows us to see the diff
+// between the local provider and the release provider. for a certain test, this will be used to see the diff between the local provider and the release provider.
```
roaks3: Nit: I'm not sure the last sentence here adds any information, could maybe be dropped

mmv1/third_party/terraform/acctest/vcr_utils.go
```diff
@@ -242,14 +265,37 @@ func initializeReleaseDiffTest(c resource.TestCase, testName string) resource.Te
...
+func InsertDiffSteps(c resource.TestCase, tempFile *os.File, releaseProvider string, localProviderName string) []resource.TestStep {
+	var countSteps = 0
 
 	var replacementSteps []resource.TestStep
 	for _, testStep := range c.Steps {
+		countSteps++
+		// todo: add preconfig - categorize test failures (add flag to steps that if they fail is a diff failure)
```
roaks3: Is this still a todo? it seems like that's what is implemented here

mmv1/third_party/terraform/acctest/vcr_utils.go
roaks3: I believe we want to remove this line now that we are counting at the top of the loop (my assumption is that the original step count is more meaningful, which is also shown for regular test failures too, vs using the count of the "new" set of steps)

mmv1/third_party/terraform/acctest/vcr_utils.go
roaks3: That's fine, but the reason the current interface is a little confusing is because the function actually alters the steps (by rewriting config and adding the preconfig) while also returning a new set of steps that then need to be assigned on line 270. So the mutation paradigm is a bit mixed. If you are going to pass in the full test case like this, I would recommend keeping the `c.Steps = replacementSteps` and `return c` lines at the end, and just say that the function modifies the test case.

mmv1/third_party/terraform/acctest/vcr_utils.go
```diff
@@ -282,11 +324,85 @@ func ReformConfigWithProvider(config, provider string) string {
 	providerReplacementBytes = []byte(providerReplacement)
 	// Match resource and data blocks that use google_ provider
 	// regex matches for labels resource and data blocks that use google_ provider
-
 	resourceHeader := regexp.MustCompile(`((resource|data) .*google_.* .*\w+.*\{ *)`)
 	return string(resourceHeader.ReplaceAll(configBytes, providerReplacementBytes))
 }
 
+// reformConfigWithProvider reformats the config to use the given provider
```
roaks3: This reformConfigWithProvider comment seems out of place?

mmv1/third_party/terraform/acctest/vcr_utils.go
```diff
@@ -282,11 +324,85 @@ func ReformConfigWithProvider(config, provider string) string {
...
+// The method matches a regex for the provider block and replaces it with the given provider.
+// For example: ' data "google_compute_network" "default" { provider = "google-local" } '
+// will be reformatted to ' data "google_compute_network" "default" { provider = "google-beta" } '
+// parseReleaseDiffOutput reads the temporary file created during the release diff test and returns the last line of output
```
roaks3: I think this comment needs to be updated now

mmv1/third_party/terraform/acctest/vcr_utils.go
```diff
@@ -282,11 +324,85 @@ func ReformConfigWithProvider(config, provider string) string {
...
+// This is useful for extracting the diff output from the file after the test has run
+
+func ParseReleaseDiffOutput(temp *os.File) (bool, string, error) {
```
roaks3: This is still an odd function signature, and I would recommend breaking into 2 functions you call one after another: `ReadDiffOutput(temp *os.File) (string, error)` would basically call os.ReadFile (you don't need to scan the file line-by-line), and `ParseReleaseDiffOutput(output string) (bool, error)` would check for [Diff] on the last line, maybe just using `strings.Split(s, "\n").

mmv1/third_party/terraform/acctest/vcr_utils.go
```diff
@@ -282,11 +324,85 @@ func ReformConfigWithProvider(config, provider string) string {
...
+func writeOutputFileDeferFunction(tempFile *os.File, failed bool) {
+	if tempFile == nil {
+		return
+	}
+	// parses the temporary file created during the release diff test and returns the last line of output
+	// This is useful for extracting the diff output from the file after the test has run
+	regularFailureFile, err := os.Create(filepath.Join("", "regular_failure_file.log"))
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "Error creating file: %v\n", err)
+		return
```
roaks3, line 390: I don't think we want to return here because tempFile would remain unclosed. Might want to move the parse + close parts up to the top.

mmv1/third_party/terraform/acctest/vcr_utils.go
```diff
@@ -282,11 +324,85 @@ func ReformConfigWithProvider(config, provider string) string {
...
+	}
+	diffFailureFile, err := os.Create(filepath.Join("", "diff_failure_file.log"))
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "Error creating file: %v\n", err)
+		return
+	}
+	var isDiff, testOutput, error = ParseReleaseDiffOutput(tempFile)
```
roaks3: Nit: s/error/err/

Here and on the few lines below. error is the name of the type.

mmv1/third_party/terraform/acctest/vcr_utils_test.go
```diff
@@ -449,3 +451,87 @@ func TestReformConfigWithProvider(t *testing.T) {
...
+
+	if len(releaseDiffSteps) != len(expectedSteps) {
+		t.Fatalf("Expected %d steps, but got %d", len(expectedSteps), len(releaseDiffSteps))
+	}
+	for i, step := range releaseDiffSteps {
+		if step.Config != expectedSteps[i].Config {
+			t.Fatalf("Expected step %d config to be:\n%q\nbut got:\n%q", i, expectedSteps[i].Config, step.Config)
+		}
+		if (i % 2) == 1 {
+			if step.ExpectNonEmptyPlan != false {
```
roaks3: Could `ExpectNonEmptyPlan` and `PlanOnly` just go in the expected steps?

mmv1/third_party/terraform/acctest/vcr_utils.go
luckyswaminathan: forgot to change within this method

mmv1/third_party/terraform/acctest/vcr_utils_test.go
luckyswaminathan: the intention moreso is to see if the steps were correctly altered within the insertdiff steps function

mmv1/third_party/terraform/acctest/vcr_utils.go
```diff
@@ -145,11 +146,20 @@ func VcrTest(t *testing.T, c resource.TestCase) {
 	if IsVcrEnabled() {
 		defer closeRecorder(t)
 	} else if isReleaseDiffEnabled() {
-		c = initializeReleaseDiffTest(c, t.Name())
+		// creates temporary file for the individual test, will be a temporary to store the output
+		tempOutputFile, err := createTemporaryFile()
+		if err != nil {
+			fmt.Printf("Error creating temporary file %s", err)
+			t.Fail()
```
roaks3: I think this would be more conventionally `t.Errorf("creating temporary file %v", err)`

mmv1/third_party/terraform/acctest/vcr_utils.go
```diff
@@ -145,11 +146,20 @@ func VcrTest(t *testing.T, c resource.TestCase) {
...
+		}
+		defer func() {
+			writeOutputFileDeferFunction(tempOutputFile, t.Failed())
+		}()
+		initializeReleaseDiffTest(c, t.Name(), tempOutputFile)
+
```
roaks3: Stray newline

mmv1/third_party/terraform/acctest/vcr_utils.go
```diff
@@ -145,11 +146,20 @@ func VcrTest(t *testing.T, c resource.TestCase) {
...
 	}
 
 	c = extendWithTGCData(t, c)
-
```
roaks3: Unrelated to your changes, so could be removed

mmv1/third_party/terraform/acctest/vcr_utils.go
```diff
@@ -166,6 +176,17 @@ func VcrTest(t *testing.T, c resource.TestCase) {
 	resource.Test(t, c)
 }
 
+func createTemporaryFile() (*os.File, error) {
+	// creates temporary file for the individual test, will be a temporary to store the output
+	tempOutputFile, err := os.CreateTemp("", "release_diff_test_output_*.log")
+	if err != nil {
+		fmt.Fprintf(os.Stdout, "Error creating temporary file: %v", err)
+		return nil, err
```
roaks3, line 184: The caller of this is already adding "creating temporary file:", so here you can just do `return nil, err`, and basically remove the print above it.

mmv1/third_party/terraform/acctest/vcr_utils.go
roaks3: Looks good but you can remove the Fail() line. It is done for you in Errorf() https://pkg.go.dev/testing#T.Errorf

mmv1/third_party/terraform/acctest/vcr_utils.go
```diff
@@ -242,14 +264,34 @@ func initializeReleaseDiffTest(c resource.TestCase, testName string) resource.Te
 			},
 		}
 	}
+	// InsertDiffSteps adds modified steps to the test that run with an external provider
+	// these steps do the actual infrastructure provisioning, and c.Steps is updated to have these modified steps
+	c.Steps = InsertDiffSteps(c, tempOutputFile, releaseProvider, localProviderName)
```
roaks3: I'd recommend just `InsertDiffSteps(c, tempOutputFile, releaseProvider, localProviderName)`, since you're mutating the test case anyway (comment would need to be updated too)

mmv1/third_party/terraform/acctest/vcr_utils.go
```diff
@@ -242,14 +264,34 @@ func initializeReleaseDiffTest(c resource.TestCase, testName string) resource.Te
...
+
+	return
+}
+
+// InsertDiffSteps inserts a new step into the test case that reformats the config to use the release provider - this allows us to see the diff
+// between the local provider and the release provider
+func InsertDiffSteps(c resource.TestCase, tempOutputFile *os.File, releaseProvider string, localProviderName string) []resource.TestStep {
```
roaks3: I'd recommend removing the return type

mmv1/third_party/terraform/acctest/vcr_utils.go
```diff
@@ -242,14 +264,34 @@ func initializeReleaseDiffTest(c resource.TestCase, testName string) resource.Te
...
 	for _, testStep := range c.Steps {
+		countSteps++
 		if testStep.Config != "" {
 			ogConfig := testStep.Config
+			fmt.Fprintf(tempOutputFile, "[DEBUG] Original config: %s\n", ogConfig)
 			testStep.Config = ReformConfigWithProvider(ogConfig, localProviderName)
-			if testStep.ExpectError == nil && testStep.PlanOnly == false {
+			fmt.Fprintf(tempOutputFile, "[DEBUG] Reformatted config: %s\n", testStep.Config)
+			testStep.PreConfig = func() {
+				fmt.Fprintf(tempOutputFile, "[Diff] Step %d\n", countSteps)
```
roaks3: Would it make sense to keep `[Diff]` in a constant at the top, since it is used again later during parse? Mainly to prevent accidental changes to the logic that is inherent in these print lines.

mmv1/third_party/terraform/acctest/vcr_utils.go
```diff
@@ -261,18 +303,18 @@ func initializeReleaseDiffTest(c resource.TestCase, testName string) resource.Te
 			replacementSteps = append(replacementSteps, testStep)
 		}
 	}
-
-	c.Steps = replacementSteps
-
-	return c
+	return replacementSteps
```
roaks3: I'd recommend just keeping `c.Steps = replacementSteps` with no return value, since the steps are being mutated anyway

mmv1/third_party/terraform/acctest/vcr_utils.go
```diff
@@ -261,18 +303,18 @@ func initializeReleaseDiffTest(c resource.TestCase, testName string) resource.Te
...
+// reformConfigWithProvider reformats the config to use the given provider
+// The method matches a regex for the provider block and replaces it with the given provider.
+// For example: ' data "google_compute_network" "default" { provider = "google-local" } '
+// will be reformatted to ' data "google_compute_network" "default" { provider = "google-beta" } '
 func ReformConfigWithProvider(config, provider string) string {
 	configBytes := []byte(config)
 	providerReplacement := fmt.Sprintf("provider = %s", provider)
 	providerReplacementBytes := []byte(providerReplacement)
 	providerBlock := regexp.MustCompile(`provider *=.*google-beta.*`)
-
```
roaks3: Nit: stray whitespace change

mmv1/third_party/terraform/acctest/vcr_utils.go
```diff
@@ -282,11 +324,87 @@ func ReformConfigWithProvider(config, provider string) string {
 	providerReplacementBytes = []byte(providerReplacement)
 	// Match resource and data blocks that use google_ provider
 	// regex matches for labels resource and data blocks that use google_ provider
-
```
roaks3: Nit: stray whitespace change

mmv1/third_party/terraform/acctest/vcr_utils.go
```diff
@@ -282,11 +324,87 @@ func ReformConfigWithProvider(config, provider string) string {
...
+func ParseReleaseDiffOutput(output string) (isDiff bool) {
+	trimmedOutput := strings.TrimSpace(output)
+	if trimmedOutput == "" {
+		return false
+	}
+
+	lines := strings.Split(trimmedOutput, "\n")
+	lastLine := lines[len(lines)-1]
+
+	isDiff = strings.HasPrefix(lastLine, "[Diff]")
```
roaks3: This would be the other place where `[Diff]` is being used

mmv1/third_party/terraform/acctest/vcr_utils.go
```diff
@@ -282,11 +324,87 @@ func ReformConfigWithProvider(config, provider string) string {
...
+func writeOutputFileDeferFunction(tempOutputFile *os.File, failed bool) {
+	if tempOutputFile == nil {
+		return
+	}
+	// parses the temporary file created during the release diff test and returns the last line of output
+	// This is useful for extracting the diff output from the file after the test has run
+
+	testOutput, err := ReadDiffOutput(tempOutputFile)
+	if err != nil {
+		fmt.Fprintf(os.Stdout, "Error reading temporary file: %v\n", err)
```
roaks3: Nit: unless there's a specific reason, these statements could be `fmt.Printf("Error reading ...`, which is the more common form.

mmv1/third_party/terraform/acctest/vcr_utils_test.go
```diff
@@ -449,3 +451,93 @@ func TestReformConfigWithProvider(t *testing.T) {
...
+				}`,
+			ExpectNonEmptyPlan: false,
+			PlanOnly:           true,
+		},
+	}
+
+	if len(releaseDiffSteps) != len(expectedSteps) {
+		t.Fatalf("Expected %d steps, but got %d", len(expectedSteps), len(releaseDiffSteps))
+	}
+	for i, step := range releaseDiffSteps {
```
roaks3: This loop would need to change to check for the `ExpectNonEmptyPlan` and `PlanOnly` fields. So probably add some more `if`s like `step.ExpectNonEmptyPlan != expectedSteps[i].ExpectNonEmptyPlan`

roaks3 (commented):
Some remaining nits, and the test should be fixed, but otherwise LGTM

mmv1/third_party/terraform/acctest/vcr_utils_test.go
```diff
@@ -449,3 +451,93 @@ func TestReformConfigWithProvider(t *testing.T) {
...
+					provider = google-beta
+				}`,
+				ExpectNonEmptyPlan: true,
+			},
+			{
+				Config: `resource "google_new_resource" {
+					provider = google-beta
+				}`,
+				ExpectNonEmptyPlan: true,
+			},
```
roaks3, line 493: Could we add a few more test cases? Some examples that I found throughout the code for some ideas:
```
 {
                ResourceName:            "google_pubsub_subscription.example",
                ImportStateVerifyIgnore: []string{"topic"},
},

{
        Config: testAccExampleResource(rName),
          testAccCheckExampleResourceExists("example_widget.foo", &widgetBefore),
        ),
},

{
				// Invalid input check - cannot pass in both sources
				Config:      testAccAlloydbClusterAndInstanceAndBackup_OnlyOneSourceAllowed(context),
				ExpectError: regexp.MustCompile("\"restore_continuous_backup_source\": conflicts with restore_backup_source"),
},
```

roaks3 (commented):
LGTM pending unit tests being added/fixed

mmv1/third_party/terraform/acctest/vcr_utils.go
```diff
@@ -145,10 +148,18 @@ func VcrTest(t *testing.T, c resource.TestCase) {
...
+			t.Errorf("creating temporary file %v", err)
+		}
+		defer func() {
+			writeOutputFileDeferFunction(tempOutputFile, t.Failed())
+		}()
+		c = initializeReleaseDiffTest(c, t.Name(), tempOutputFile)
 	}
 
-	c = extendWithTGCData(t, c)
+	//c = extendWithTGCData(t, c)
```
roaks3: I think this may need to be uncommented (also would presumably affect things outside of Diff Test)

mmv1/third_party/terraform/acctest/vcr_utils_test.go
roaks3: Looks like unit tests are passing again, but I still think some of these should be added.

Also, is the `ExpectNonEmptyPlan: true,` from steps 2 and 3 getting removed? I'm surprised the test is passing for them.

luckyswaminathan: all original steps are the ones modified to expectnonemptyplan = false, so they are overwritten

mmv1/third_party/terraform/acctest/vcr_utils.go
```diff
@@ -287,6 +331,83 @@ func ReformConfigWithProvider(config, provider string) string {
...
+		if isDiff {
+			fmt.Fprintf(os.Stdout, "%s Breaking Change Detected] \n", diffFlag)
+			fmt.Fprintf(diffFailureFile, "%s %s\n", diffFlag, testOutput)
+		} else {
+			fmt.Fprintf(regularFailureFile, testOutput)
+			fmt.Fprintf(regularFailureFile, "FAILED --- %s\n", testOutput)
+		}
+	}
+}
+
```
trodge, line 410: If all this is only used for diff test, we might want to put it in its own file, as well as the unit tests. Maybe something like `diff_utils.go` and `diff_utils_test.go`.

luckyswaminathan: can do so in followup pr - really just want this one merged so I can have a functional release_diff to use within teamcity

roaks3: Yea agreed, I think it's a good point, but would also include other functions not modified as part of this PR. A follow-up PR makes sense to me.

roaks3 (approved):

PR #14481, opened 2025-07-09
[DO NOT MERGE] Adding Multiregion support to Listings and Listing Subscriptions.

arnavdham:
Adding Multiregion support to Listings and Listing Subscriptions. This adds support to create multi-region listings and multi-region listing subscriptions. @shashambhavi

```release-note:enhancement
bigqueryanalyticshub: added `replica_locations` and 'effective_replicas' field to `google_bigquery_analytics_hub_listing` resource
```


mmv1/third_party/terraform/acctest/bootstrap_test_utils.go.tmpl
```diff
@@ -34,6 +34,7 @@ import (
 	"google.golang.org/api/servicenetworking/v1"
 	"google.golang.org/api/serviceusage/v1"
 	sqladmin "google.golang.org/api/sqladmin/v1beta4"
+	"cloud.google.com/go/bigquery"
```
c2thorn, line 37: we have a guide on how to update dependencies to include this. You'll need to update the correct go.mod and go.sum at least. https://googlecloudplatform.github.io/magic-modules/develop/update-dependencies/

mmv1/third_party/terraform/acctest/bootstrap_test_utils.go.tmpl
```diff
@@ -2311,3 +2312,99 @@ func BootstrapIntegrationsClient(t *testing.T, locationID string) BootstrapClien
...
+	}
+	defer client.Close() 
+
+	log.Printf("INFO: Attempting to create dataset '%s' in project '%s' at location '%s'", datasetID, projectID, primaryLocation)
+	datasetRef := client.Dataset(datasetID)
+	datasetMetadata := &bigquery.DatasetMetadata{
+		Location: primaryLocation,
+	}
+
+	err = datasetRef.Create(ctx, datasetMetadata)
```
c2thorn, line 2331: is the bigquery client necessary to do this? any equivalent rest API we can just call?

mmv1/products/bigqueryanalyticshub/Listing.yaml
```diff
@@ -86,6 +86,14 @@ examples:
       listing_id: 'tf_test_pubsub_listing'
       pubsub_topic_name: 'test_pubsub'
       description: 'Example for pubsub topic source'
+# - name: 'bigquery_analyticshub_listing_multiregion'
```
c2thorn: if you end up needing to have the create test be handwritten-only, you can make these examples documentation-only by adding skip_test: true. Just also provide an explanatory comment

c2thorn: if you end up needing to have the create test be handwritten-only, you can make these examples documentation-only by adding skip_test: true. Just also provide an explanatory comment

mmv1/products/bigqueryanalyticshub/Listing.yaml
```diff
@@ -86,6 +86,14 @@ examples:
...
+#   primary_resource_id: 'listing'
+#   primary_resource_name: 'fmt.Sprintf("tf_test_my_data_exchange%s", context["random_suffix"]), fmt.Sprintf("tf_test_my_listing%s", context["random_suffix"])'
+#   region_override: 'US'
+#   vars:
```
c2thorn: You can try passing the bootstrap function into test_vars_overrides like in https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/products/alloydb/Backup.yaml#L63

mmv1/third_party/terraform/acctest/bootstrap_test_utils.go.tmpl
arnavdham: Nope there is no API support for creating a replica. This is the only way.

mmv1/products/bigqueryanalyticshub/Listing.yaml
arnavdham: Used exclude_test instead of skip_test. Skip_test was still creating a test.

mmv1/products/bigqueryanalyticshub/Listing.yaml
arnavdham: Used the handwritten way instead.

mmv1/third_party/terraform/acctest/bootstrap_test_utils.go.tmpl
wj-chen: IIUC this is trying to 1) create a dataset then 2) run a DDL to add a replica, both of which should have REST API support. 1) would be a `POST` to the Dataset API, and 2) would be a `POST` to the Job API to create a query Job with the DDL as the `query` text and `useLegacySql = false` (https://github.com/GoogleCloudPlatform/magic-modules/blob/490a414a5b05cadaef67a1c8c792019e8d6d8bcf/mmv1/third_party/terraform/services/bigquery/resource_bigquery_table.go.tmpl#L2170 has an example of which input parameters to set).

The clean-up steps should be similar. Give it a try and let me know if you run into any issues.

arnavdham: This uses the "google.golang.org/api/bigquery/v2" package which is deprecated. The new advised package to use is "cloud.google.com/go/bigquery" which I have done. Also the PR for adding this Dependency has been merged so I don't think this should be an issue in any case now. Thanks!

mmv1/third_party/terraform/services/bigqueryanalyticshub/resource_bigquery_analytics_hub_listing_subscription_test.go
```diff
@@ -119,3 +119,81 @@ resource "google_bigquery_analytics_hub_listing_subscription" "subscription" {
 }
 `, context)
 }
+
+func TestAccBigqueryAnalyticsHubListingSubscription_bigqueryAnalyticshubListingSubscriptionMultiregionExample(t *testing.T) {
+	t.Parallel()
```
c2thorn, line 132:
```suggestion
	// Skip if not running in a acc test environment,
	// as acc test environment variables needed to get use 
	// the bigquery client for bootstrapping
	if v := os.Getenv("TF_ACC"); v == "" {
		t.Skip("Acceptance tests skipped unless env 'TF_ACC' set")
	}
	t.Parallel()
```

c2thorn (changes requested):

arnavdham:
Closed as this was taken up and merged

PR #14482, opened 2025-07-09
fix: (storage) added validation for cors parameters method, origin and response_header cannot be empty

gurusai-voleti:

```release-note:bug
storage: added validation for cors parameters method, origin and response_header cannot be empty
```

### DIFF
--- a/mmv1/third_party/terraform/services/storage/resource_storage_bucket.go.tmpl
+++ b/mmv1/third_party/terraform/services/storage/resource_storage_bucket.go.tmpl
@@ -27,6 +27,23 @@ import (
 	"google.golang.org/api/storage/v1"
 )
 
+func corsValidation(_ context.Context, d *schema.ResourceDiff, meta interface{}) error {
+	method, methodOk := d.GetOkExists("cors.0.method")
+	responseHeader, responseHeaderOk := d.GetOkExists("cors.0.response_header")
+	origin, originOk := d.GetOkExists("cors.0.origin")
+
+	if methodOk && len(method.([]interface{})) == 0 {
+		return fmt.Errorf("method cannot be empty")
+	}
+	if responseHeaderOk && len(responseHeader.([]interface{})) == 0 {
+		return fmt.Errorf("response_header cannot be empty")
+	}
+	if originOk && len(origin.([]interface{})) == 0 {
+		return fmt.Errorf("origin cannot be empty")
+	}
+	return nil
+}
+
 func ResourceStorageBucket() *schema.Resource {
 	return &schema.Resource{
 		Create: resourceStorageBucketCreate,
@@ -39,6 +56,7 @@ func ResourceStorageBucket() *schema.Resource {
 		CustomizeDiff: customdiff.All(
 			customdiff.ForceNewIfChange("retention_policy.0.is_locked", isPolicyLocked),
 			tpgresource.SetLabelsDiff,
+			corsValidation,
 		),
 
 		Timeouts: &schema.ResourceTimeout{


mmv1/third_party/terraform/services/storage/resource_storage_bucket.go.tmpl
```diff
@@ -2208,8 +2256,10 @@ func setStorageBucket(d *schema.ResourceData, config *transport_tpg.Config, res
 	if err := d.Set("location", res.Location); err != nil {
 		return fmt.Errorf("Error setting location: %s", err)
 	}
-	if err := d.Set("cors", flattenCors(res.Cors)); err != nil {
-		return fmt.Errorf("Error setting cors: %s", err)
+	if res.Cors != nil {
```
kautikdk: In case of nil value of CORS, can we set empty empty list similar to lifecycle rules?

kautikdk (commented):
While I agree that validation logic will work in this case, can we not fix the issue with diffSupressFunc? If that works, we don't have to go through separate customize diff and validation logic from client side.

If CustomizeDiff is the only solution, I would recommend adding unit and acceptance tests for the same. Current tests are essentially not validating the issue scenario rather they just ensure working of the new field.

mmv1/third_party/terraform/services/storage/resource_storage_bucket.go.tmpl
gurusai-voleti: added empty list

mmv1/third_party/terraform/services/storage/resource_storage_bucket.go.tmpl
```diff
@@ -27,6 +27,35 @@ import (
...
+		responseHeader := rule["response_header"].([]interface{})
+		origin := rule["origin"].([]interface{})
+
+		// check if all cors attributes provided have values
+		if !(len(method) > 0 || len(origin) > 0 || len(responseHeader) > 0) {
+			return fmt.Errorf("cors attributes cannot be empty or use send_max_age_seconds_if_zero flag to send max_age_seconds as zero")
+		}
+	}
+	return nil
+}
```
melinath: Just want to verify: are the error cases here things that would otherwise be API errors? Otherwise this would be a [breaking change](https://googlecloudplatform.github.io/magic-modules/breaking-changes/breaking-changes/).

gurusai-voleti: these are not replication API errors even API accepts empty values for cors field but in API response we wont get anything back for cors field, so validating here to ensure values are provided and not empty

gurusai-voleti:
> While I agree that validation logic will work in this case, can we not fix the issue with diffSupressFunc? If that works, we don't have to go through separate customize diff and validation logic from client side.
> 
> If CustomizeDiff is the only solution, I would recommend adding unit and acceptance tests for the same. Current tests are essentially not validating the issue scenario rather they just ensure working of the new field.

List types cannot be handled by diffsuppress func as per terraform SDK its not recommended approach

mmv1/third_party/terraform/services/storage/resource_storage_bucket.go.tmpl
melinath: If it's valid from the API's perspective, we can't add stricter client-side validation once the field in question is introduced. Since cors is already released, adding validation to it would be a breaking change.

melinath (changes requested):
Marking as changes requested due to the feedback above.

> List types cannot be handled by diffsuppress func as per terraform SDK its not recommended approach

Just to clarify, I don't think this is correct - diffsuppress definitely works for lists - but diffsuppress _can't_ handle validation across multiple fields.

gurusai-voleti:
> Marking as changes requested due to the feedback above.
> 
> > List types cannot be handled by diffsuppress func as per terraform SDK its not recommended approach
> 
> Just to clarify, I don't think this is correct - diffsuppress definitely works for lists - but diffsuppress _can't_ handle validation across multiple fields.

sorry it supports, I tried diffsuppress first but seems not feasible for this solution to handle multiple fields, customize diff seems a better option here as multiple fields needed for validation

melinath (changes requested):

gurusai-voleti:
Diffsuppress only provides way to show diff or not in tf  plan but it will not to throw any error, here we need to validate the data and let customer/user know about the data validation error, so customDiff seems a feasible solution than diffsuppress,  please correct me if I am wrong

melinath:
> Diffsuppress only provides way to show diff or not in tf plan but it will not to throw any error, here we need to validate the data and let customer/user know about the data validation error, so customDiff seems a feasible solution than diffsuppress, please correct me if I am wrong

yes, that's also correct. Single-field validation would use validation functions, not diff suppress. 🤦

mmv1/third_party/terraform/services/storage/resource_storage_bucket.go.tmpl
kautikdk: Agree!

gurusai-voleti: removed the customDiff validation, fixed expandCors which is root cause of the issue which user has reported

kautikdk:
Since the proposed solution is changed, can you please add tests to verify behavior? I.e tests with empty CORS fields.

mmv1/third_party/terraform/services/storage/resource_storage_bucket.go.tmpl
```diff
@@ -420,6 +421,44 @@ func ResourceStorageBucket() *schema.Resource {
 			"cors": {
 				Type:     schema.TypeList,
 				Optional: true,
+				DiffSuppressFunc: func(k, oldValue, newValue string, d *schema.ResourceData) bool {
+					if d.HasChange("cors") && (k == "cors.#") {
```
melinath: hm, yeah, calling d.GetChange seems to be something we do on other resources. I guess a DiffSuppress actually can take other fields into account! Interesting.

mmv1/third_party/terraform/services/storage/resource_storage_bucket.go.tmpl
```diff
@@ -420,6 +421,44 @@ func ResourceStorageBucket() *schema.Resource {
 			"cors": {
 				Type:     schema.TypeList,
 				Optional: true,
+				DiffSuppressFunc: func(k, oldValue, newValue string, d *schema.ResourceData) bool {
```
melinath: Rather than handling this with a diff suppress func, it would be better to handle it with a flattener, if possible. For example, if the cors block is returned empty from the API, you could check whether the value set in config had all-empty children (using `d.Get("cors")`) and then return the value of `d.Get("cors")` to preserve the value from config in that case (or whatever cases you'd need to preserve the value, if I didn't capture that quite right). This would suppress the diff, and if they make any additional changes, the diff will better match what the user might expect.

What would you think about doing it that way? Would it be possible?

gurusai-voleti: the above suggestion, if API response doesn't contain cors it will fill an empty array (cors: []) for field cors in state file, currently it is filling the cors as array with all fields set to null values,  both cases will act as zero length  for old values in diffsuppress, 

the below terraform config will always show diff for cors fields unless we suppress it in diffsuppress, if it shows the diff for un changed cors this is a problem for user, 

resource "google_storage_bucket" "bucket1" {
  name     = "or-hello"
  location = "US"
  uniform_bucket_level_access = true
  force_destroy=true
  cors{}
  cors{}
}

diffsuppress helps to suppress it as both the cases will consider the old value as an empty array, and when testing the suggested fix it doesn't fully fix the issue with diff in other words if no change in config still tf plan/apply shows cors as new fields to be added every time we run tf plan/apply which is not good for users

melinath: Yes, I understand why you're using diffsuppress, but I'm not sure I understand why any of that means that using a flattener doesn't work - if the flattener stores a value in state that matches what's in config, users shouldn't see a diff. Generally speaking, preventing diffs with flatteners is better than preventing them with diffsuppress, because then when there _is_ a diff, it only shows what's actually changing (whereas with a diffsuppress, everything that was suppressed will also show up when a diff is reported - even if the reported diff is on some other field).

gurusai-voleti: @melinath  in the input terraform config provided by users as below

resource "google_storage_bucket" "bucket1" {
name = "or-hello"
location = "US"
uniform_bucket_level_access = true
force_destroy=true
cors{}
}

the cors {} from input is considered an empty map as nil [<nil>] and the array size is not zero, where cors value from state file the array length is zero, this causes the diff everytime so in diffsuppress verifying the input cors to suppress the diff but this is restricting the customer to use empty cors if really needed initially

melinath: yes, so if the user set `cors{}` in their config (or if the API returns the corresponding empty value), the flattener could make sure the value set in state matches what's in the config.

melinath (changes requested):
marking as changes requested per above conversation

melinath:
There's enough noise on the PR that it's hard to tell offhand.

gurusai-voleti:
> @gurusai-voleti have you addressed the previously-requested changes?

yes all the comments are addressed and handled only with flattener

mmv1/third_party/terraform/services/storage/resource_storage_bucket_test.go
```diff
@@ -2184,11 +2225,38 @@ resource "google_storage_bucket" "bucket" {
   }
 
   cors {
-    origin          = ["ghi", "jkl"]
-    method          = ["z9z"]
-    response_header = ["000"]
-    max_age_seconds = 5
+    origin                       = ["ghi", "jkl"]
+    method                       = ["z9z"]
+    response_header              = ["000"]
```
kautikdk: Spacing is incorrect for this block, can we just revert these unnecessary changes?

mmv1/third_party/terraform/website/docs/r/storage_bucket.html.markdown
```diff
@@ -38,6 +38,13 @@ resource "google_storage_bucket" "static-site" {
     response_header = ["*"]
     max_age_seconds = 3600
   }
+  cors {
+    origin                       = ["http://image-store.com"]
+    method                       = ["GET", "HEAD", "PUT", "POST", "DELETE"]
+    response_header              = ["*"]
+    max_age_seconds              = 0
+    send_max_age_seconds_if_zero = true
```
kautikdk: I don't think we are introducing this field in the PR. Please remove this and other references!

mmv1/third_party/terraform/services/storage/resource_storage_bucket.go.tmpl
```diff
@@ -1330,20 +1330,50 @@ func expandCors(configured []interface{}) []*storage.BucketCors {
...
+			continue
+		}
 		data := raw.(map[string]interface{})
-		corsRule := storage.BucketCors{
-			Origin:         tpgresource.ConvertStringArr(data["origin"].([]interface{})),
-			Method:         tpgresource.ConvertStringArr(data["method"].([]interface{})),
-			ResponseHeader: tpgresource.ConvertStringArr(data["response_header"].([]interface{})),
-			MaxAgeSeconds:  int64(data["max_age_seconds"].(int)),
+		corsRule := storage.BucketCors{}
+		if v, ok := data["method"].([]interface{}); ok {
```
kautikdk, line 1340: While I understand flattener based solution from the PR discussion, why to modify the expander which essentially modifies the API request?
Also what is the difference that we are making here that was not present in the`tpgresource.ConvertStringArr`?

mmv1/third_party/terraform/services/storage/resource_storage_bucket_test.go
```diff
@@ -1149,6 +1149,47 @@ func TestAccStorageBucket_cors(t *testing.T) {
 	})
 }
 
+func TestAccStorageBucket_emptyCors(t *testing.T) {
```
kautikdk, line 1153: Can you please add some more test steps including multiple `cors` blocks? Current ones only covers simple scenarios with one `cors` config, we should test this fix with more complicated scenarios.

mmv1/third_party/terraform/services/storage/resource_storage_bucket.go.tmpl
gurusai-voleti: the user's reported issue is due to nil error is being throwed when the user entered empty cors{} block
when an nil is trying to convert to array "data["method"].([]interface{})"

while fixing this issue we encountered an issue with diff and flattener  which is also fixed

mmv1/third_party/terraform/services/storage/resource_storage_bucket_test.go
gurusai-voleti: main issue is with empty cors{} and completely removing cors when previously those are set so covered those cases with more than one cors blocks and ensuring users can enter empty arrays fro cors attributes and empty cors{} blocks allowed and no diff occurs also and no error in tf apply

mmv1/third_party/terraform/services/storage/resource_storage_bucket_test.go
gurusai-voleti: formatted the terraform config

mmv1/third_party/terraform/website/docs/r/storage_bucket.html.markdown
gurusai-voleti: removed the unncessary attribute from the documentation

mmv1/third_party/terraform/services/storage/resource_storage_bucket.go.tmpl
kautikdk: Okay,
What about using `tpgresource.ConvertStringArr` inside the condition block? Current implementation has some repetitive code which can be replaced by common utility functions.

mmv1/third_party/terraform/services/storage/resource_storage_bucket_test.go
kautikdk: I am not able to completely understand your reasoning, can you please explain it in a bit more detail and with an example if possible?

gurusai-voleti: The user reported issue is they cannot use empty cors{} and empty arrays with cors {method=[], origin=[]} 

covered the above cases with multiple entries of cors, like below
completely filled cors -> empty cors blocks-> partially filled cors ->  and finally completely removed cors from terraform config all cases covered in tests

mmv1/third_party/terraform/services/storage/resource_storage_bucket.go.tmpl
gurusai-voleti: replaced repeating logic with tpgresource.ConvertStringArr, please verify

mmv1/third_party/terraform/services/storage/resource_storage_bucket_test.go
kautikdk: Got your point. My ask was to add some tests related to the modified flattener logic in which we are adding multiple blank CORS block if some are absent in the API response.

gurusai-voleti: ~~this already handles it, in testGoogleStorageBucketsEmptyCors case we have empty cors and empty arrays in cors in request and in API response wont return any for both, for this I added cors to importverifyignore as it wont present in API response and in next step we update bucket with partiallyfilled cors and we added cors to importverify so this both cases implicitly tests the flattener logic~~

gurusai-voleti: added a test verification to ensure the cors length is equal in config and state file, this ensures the flattener is working as expected which ensure there is an empty cors in state file for any empty cors in user terraform config, as API won't return any empty cors in response and this doesn't create any diff as well

kautikdk:
@melinath, can you please continue your review and check if the proposed solution in the PR is the one that you suggested?  I will just do the final pass after your review to reduce review noice in the PR.

melinath:
yeah, I can do that tomorrow!

melinath:
Rerunning just in case there was a race condition with the last three commits.

mmv1/third_party/terraform/services/storage/resource_storage_bucket.go.tmpl
```diff
@@ -1352,9 +1370,21 @@ func flattenCors(corsRules []*storage.BucketCors) []map[string]interface{} {
...
+		for ; differnce > 0; differnce-- {
+			cors := map[string]interface{}{
+				"origin":          []string{},
+				"method":          []string{},
+				"response_header": []string{},
+				"max_age_seconds": 0,
+			}
+			corsRulesSchema = append(corsRulesSchema, cors)
+		}
+	}
```
melinath: This logic isn't quite correct - it just puts empty cors blocks at the end. But if there's an empty cors block in the middle, this creates a permadiff. For example:

```tf
  cors {
    origin          = ["*"]
    method          = ["GET"]
    response_header = ["*"]
    max_age_seconds = 3600
  }
  cors {
    origin          = []
    method          = []
    response_header = []
    # max_age_seconds = 0
  }

  cors {
    origin          = ["*"]
    method          = ["HEAD", "PUT", "POST", "DELETE"]
    response_header = ["*"]
    max_age_seconds = 10
  }
```

Instead, you need to add an empty cors block at every index where one is present in the value in state.

gurusai-voleti: Updated the logic to insert empty cors as per the config

PR #14486, opened 2025-07-09
Fix TRACE log level override in VCR tests by setting TF_LOG_CORE to match TeamCity

ScottSuarez:
## Problem
VCR tests in Google Cloud Build were showing excessive TRACE-level logs despite setting `TF_LOG=DEBUG`, making test output difficult to read and parse.

## Root Cause
The terraform-plugin-testing framework has a behavior where if `TF_LOG_PATH_MASK` is set but `TF_ACC_LOG` or `TF_LOG_CORE` is not explicitly set, it defaults the log level to TRACE. Our VCR tests were setting `TF_LOG_PATH_MASK` for per-test log files but not setting any core log level control, causing the unintended TRACE level override.

https://github.com/hashicorp/terraform-plugin-testing/blob/cf1f0ecc8af7b06b7e72adbb122a5c592b78e469/internal/plugintest/environment_variables.go#L22-L31

## Solution
Set `TF_LOG_CORE=WARN` in the VCR test environment to match the TeamCity configuration and prevent the automatic TRACE default behavior.

## Why This Wasn't Noticed Before
TeamCity tests work correctly because they already set `TF_LOG_CORE=WARN`. GCB tests only set `TF_LOG=DEBUG` and `TF_LOG_PATH_MASK`, triggering the framework's TRACE default.

```release-note:none

```


melinath (approved):
heck yeah

PR #14473, opened 2025-07-08
Add tags field to Metastore Federation for TagsR2401

rajeshgupthar:
Add tags field to federation resource to allow setting tags on federation resources at creation time.
Part of b/367307391

IMPORTANT: Please do not merge this PR unless we confirm the respective APIs are rolled out in prod. We will ping here on this PR once the APIs are rolled out. 

**Release Note Template for Downstream PRs (will be copied)**

See [Write release notes](https://googlecloudplatform.github.io/magic-modules/contribute/release-notes/) for guidance.

```release-note:enhancement
metastore: added `tags' field to metastore to allow setting tags for federations at creation time
```

Requesting mandatory review from @aniket-gupta

PR #14489, opened 2025-07-09
Firestore Tags R2401 terraform support

jingqizz:
This commit adds support for the tags field to the Firestore Database resource. The tags field is a map<string, string>. This allows users to perform database actions and setting tag bindings at the same time.

`firestore: add `tags` field to `database` resource`

### DIFF
--- a/mmv1/templates/terraform/examples/firestore_database_with_tags.tf.tmpl
+++ b/mmv1/templates/terraform/examples/firestore_database_with_tags.tf.tmpl
@@ -1,0 +1,11 @@
+resource "google_firestore_database" "{{$.PrimaryResourceId}}" {
+  project                           = "{{index $.TestEnvVars "project_id"}}"
+  name                              = "{{index $.Vars "database_id"}}"
+  location_id                       = "nam5"
+  type                              = "FIRESTORE_NATIVE"
+  delete_protection_state           = "{{index $.Vars "delete_protection_state"}}"
+  deletion_policy                   = "DELETE"
+  tags = {
+    "{{index $.Vars "tag_key_id"}}": "{{index $.Vars "tag_value_id"}}"
+  }
+}
--- a/mmv1/products/firestore/Database.yaml
+++ b/mmv1/products/firestore/Database.yaml
@@ -72,6 +72,22 @@ examples:
       - 'project'
       - 'etag'
       - 'deletion_policy'
+  - name: 'firestore_database_with_tags'
+    primary_resource_id: 'database'
+    vars:
+      database_id: 'database-with-tags-id'
+      delete_protection_state: 'DELETE_PROTECTION_ENABLED'
+      tag_key_id: 'keyname'
+      tag_value_id: 'valuename'
+    test_env_vars:
+      project_id: 'PROJECT_NAME'
+    test_vars_overrides:
+      'delete_protection_state': '"DELETE_PROTECTION_DISABLED"'
+    ignore_read_extra:
+      - 'project'
+      - 'etag'
+      - 'deletion_policy'
+    exclude_test: true
   - name: 'firestore_cmek_database'
     primary_resource_id: 'database'
     vars:
@@ -315,3 +331,14 @@ properties:
         output: true
         item_type:
           type: String
+  - name: 'tags'
+    type: KeyValuePairs
+    description: |
+      Input only. A map of resource manager tags. Resource manager tag keys
+      and values have the same definition as resource manager tags.
+      Keys must be in the format tagKeys/{tag_key_id}, and values are in the format tagValues/456.
+      The field is ignored when empty. The field is immutable and causes
+      resource replacement when mutated. To apply tags to an existing resource, see
+      the `google_tags_tag_value` resource.
+    immutable: true
+    ignore_read: true
--- a/mmv1/third_party/terraform/services/firestore/resource_firestore_database_test.go.tmpl
+++ b/mmv1/third_party/terraform/services/firestore/resource_firestore_database_test.go.tmpl
@@ -1,0 +1,54 @@
+package firestore_test
+
+import (
+  "testing"
+
+  "github.com/hashicorp/terraform-plugin-testing/helper/resource"
+	"github.com/hashicorp/terraform-provider-google/google/acctest"
+	"github.com/hashicorp/terraform-provider-google/google/envvar"
+)
+
+func TestAccFirestoreDatabase_tags(t *testing.T) {
+  t.Parallel()
+
+  // Bootstrap shared tag key and value
+  tagKey := acctest.BootstrapSharedTestProjectTagKey(t, "firestore-databases-tagkey", map[string]interface{}{})
+  context := map[string]interface{}{
+    "pid":           envvar.GetTestProjectFromEnv(),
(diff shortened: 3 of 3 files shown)


jingqizz:
cc @rwhogg 
Sorry I messed up the previous PR by running a super wrong rebase. I copied and pasted the changes here. The local run `make testacc TEST=./google/services/firestore TESTARGS='-run=^TestAccFirestoreDatabase_tags -count=1'` passed.

jingqizz:
> quick question: What's the reason for handwriting the test instead of generating it from the example? If you only need to use the bootstrapped value, `test_vars_overrides` should work for that.

Hello Shuya (@shuyama1 ), thank you for the question and taking a look at the PR!

(1) Based on my understanding, I cannot leverage the `acctest` helper methods for dynamically pre-creating tags within `Database.yaml` file since go codes cannot directly be invoked within a YAML configuration file. 

(2) Then, if we rely on the test generated from the example, we will need to perform the following steps:
- create tagKey, then tagValue
- create db with tags
- then the clean up process will start trying to delete the db resource, tagValue, and tagKey
- that's where we observed issues - tagValue cannot be deleted because there exists some bindings using this tagValue being associated to the **deleted** db.
If I understand correctly, the backend mechanism of tags will realize that "oh the resource has been deleted, so the binding is no longer valid", but that seems to take some time. Hence, we decided to add a go test manually instead of relying on the example since the clean-up process will complain.

(3) Another reason is that the other teams who adopt the horizontal Tags R2401 also manually added Go tests. (e.g. PR #14473, PR #13994, PR #13568) 

Please let me know if you have any questions!

mmv1/templates/terraform/examples/firestore_database_with_tags.tf.tmpl
```diff
@@ -0,0 +1,11 @@
+resource "google_firestore_database" "{{$.PrimaryResourceId}}" {
+  project                           = "{{index $.TestEnvVars "project_id"}}"
+  name                              = "{{index $.Vars "database_id"}}"
+  location_id                       = "nam5"
+  type                              = "FIRESTORE_NATIVE"
+  delete_protection_state           = "{{index $.Vars "delete_protection_state"}}"
+  deletion_policy                   = "DELETE"
+  tags = {
+    "{{index $.Vars "tag_key_id"}}": "{{index $.Vars "tag_value_id"}}"
```
shuyama1:
```suggestion
    "{{index $.Vars "tag_key_id"}}" = "{{index $.Vars "tag_value_id"}}"
```
Please correct me if I'm wrong

mmv1/third_party/terraform/services/firestore/resource_firestore_database_test.go
```diff
@@ -0,0 +1,54 @@
...
+func testAccFirestoreDatabaseTags(context map[string]interface{}) string {
+	return acctest.Nprintf(`
+    resource "google_firestore_database" "database" {
+      name                              = "tf-test-database-%{random_suffix}"
+      location_id                       = "nam5"
+      type                              = "FIRESTORE_NATIVE"
+      delete_protection_state           = "DELETE_PROTECTION_DISABLED"
+      deletion_policy                   = "DELETE"
+      tags = {
+	      "%{pid}/%{tagKey}" = "%{tagValue}"
```
shuyama1: only nit-pick: spaces for indentation instead of tabs

shuyama1 (commented):
> > quick question: What's the reason for handwriting the test instead of generating it from the example? If you only need to use the bootstrapped value, `test_vars_overrides` should work for that.
> 
> Hello Shuya (@shuyama1 ), thank you for the question and taking a look at the PR!
> 
> (1) Based on my understanding, I cannot leverage the `acctest` helper methods for dynamically pre-creating tags within `Database.yaml` file since go codes cannot directly be invoked within a YAML configuration file.
> 
> (2) Then, if we rely on the test generated from the example, we will need to perform the following steps:
> 
> * create tagKey, then tagValue
> * create db with tags
> * then the clean up process will start trying to delete the db resource, tagValue, and tagKey
> * that's where we observed issues - tagValue cannot be deleted because there exists some bindings using this tagValue being associated to the **deleted** db.
>   If I understand correctly, the backend mechanism of tags will realize that "oh the resource has been deleted, so the binding is no longer valid", but that seems to take some time. Hence, we decided to add a go test manually instead of relying on the example since the clean-up process will complain.
> 
> (3) Another reason is that the other teams who adopt the horizontal Tags R2401 also manually added Go tests. (e.g. PR #14473, PR #13994, PR #13568)
> 
> Please let me know if you have any questions!


Thanks for the detailed explanation! You can still use bootstrapped resources in example-based tests with `test_vars_overrides`. I've created a demo [PR](https://github.com/GoogleCloudPlatform/magic-modules/pull/14497/files) to show how it can be configured. However, this is not a blocker for this PR. I just wanted to share it in case it's helpful for your future work.

I've also left some small comments. Overall looks good.

jingqizz:
> > > quick question: What's the reason for handwriting the test instead of generating it from the example? If you only need to use the bootstrapped value, `test_vars_overrides` should work for that.
> > 
> > 
> > Hello Shuya (@shuyama1 ), thank you for the question and taking a look at the PR!
> > (1) Based on my understanding, I cannot leverage the `acctest` helper methods for dynamically pre-creating tags within `Database.yaml` file since go codes cannot directly be invoked within a YAML configuration file.
> > (2) Then, if we rely on the test generated from the example, we will need to perform the following steps:
> > 
> > * create tagKey, then tagValue
> > * create db with tags
> > * then the clean up process will start trying to delete the db resource, tagValue, and tagKey
> > * that's where we observed issues - tagValue cannot be deleted because there exists some bindings using this tagValue being associated to the **deleted** db.
> >   If I understand correctly, the backend mechanism of tags will realize that "oh the resource has been deleted, so the binding is no longer valid", but that seems to take some time. Hence, we decided to add a go test manually instead of relying on the example since the clean-up process will complain.
> > 
> > (3) Another reason is that the other teams who adopt the horizontal Tags R2401 also manually added Go tests. (e.g. PR #14473, PR #13994, PR #13568)
> > Please let me know if you have any questions!
> 
> Thanks for the detailed explanation! You can still use bootstrapped resources in example-based tests with `test_vars_overrides`. I've created a demo [PR](https://github.com/GoogleCloudPlatform/magic-modules/pull/14497/files) to show how it can be configured. However, this is not a blocker for this PR. I just wanted to share it in case it's helpful for your future work.
> 
> I've also left some small comments. Overall looks good.

Hello Shuya, thank you very much for the demo PR and the comments. We will definitely use that for future reference. The comments have been resolved now. Please kindly let me know if you have any other questions or concerns.

mmv1/third_party/terraform/services/firestore/resource_firestore_database_test.go
```diff
@@ -0,0 +1,54 @@
...
+      }
```
shuyama1:
```suggestion
      tags = {
        "%{pid}/%{tagKey}" = "%{tagValue}"
      }
```

Sorry, I meant spaces only in the resource config blocks. This should fix it. Thanks!

shuyama1 (approved):
LGTM! I've modified the changlog a bit.

PR #14490, opened 2025-07-09
tgc-revival: rework comparison of terraform configs

trodge:
Step 1 is now passing for cloudfunctions2 function, but tf plan is failing because the post-cai2hcl asset lacks `location`.

```release-note:none

```

### DIFF
--- a/mmv1/third_party/tgc_next/test/hcl.go
+++ b/mmv1/third_party/tgc_next/test/hcl.go
@@ -1,0 +1,120 @@
+package test
+
+import (
+	"fmt"
+	"log"
+
+	"github.com/hashicorp/hcl/v2"
+	"github.com/hashicorp/hcl/v2/hclparse"
+	"github.com/hashicorp/hcl/v2/hclsyntax"
+)
+
+func parseHCLBytes(src []byte, filePath string) (map[string]map[string]struct{}, error) {
+	parser := hclparse.NewParser()
+	hclFile, diags := parser.ParseHCL(src, filePath)
+	if diags.HasErrors() {
+		return nil, fmt.Errorf("parse HCL: %w", diags)
+	}
+
+	if hclFile == nil {
+		return nil, fmt.Errorf("parsed HCL file %s is nil cannot proceed", filePath)
+	}
+
+	parsed := make(map[string]map[string]struct{})
+
+	for _, block := range hclFile.Body.(*hclsyntax.Body).Blocks {
+		if block.Type == "resource" {
+			if len(block.Labels) != 2 {
+				log.Printf("Skipping address block with unexpected number of labels: %v", block.Labels)
+				continue
+			}
+
+			resType := block.Labels[0]
+			resName := block.Labels[1]
+			addr := fmt.Sprintf("%s.%s", resType, resName)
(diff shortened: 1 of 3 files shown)


zli82016 (approved):
Great job. Thanks.

zli82016:
cc @hao-nan-li

mmv1/third_party/tgc_next/test/hcl.go
```diff
@@ -0,0 +1,120 @@
...
+			}
+			flatten(value, newPrefix, result)
+		}
+	case []interface{}:
+		if len(v) == 0 && prefix != "" {
+			result[prefix] = struct{}{}
+		}
+		for i, value := range v {
+			newPrefix := fmt.Sprintf("%s.%d", prefix, i)
+			flatten(value, newPrefix, result)
```
trodge, line 113: Just realized that this won't actually solve the problem of lists of nested objects being in the wrong order after conversion. If we want to ignore order of lists entirely, we could do something like this:
```suggestion
			flatten(value, prefix, result)
```
But that would mean that we don't check if the arrays have the same elements after conversion, so for example something like this:

```
resource "google_compute_firewall" "default" {
  name    = "test-firewall"
  network = google_compute_network.default.name

  allow {
    protocol = "icmp"
  }

  allow {
    protocol = "tcp"
    ports    = ["80", "8080", "1000-2000"]
  }

  source_tags = ["web"]
}
```
would compare equal to
```
resource "google_compute_firewall" "default" {
  name    = "test-firewall"
  network = google_compute_network.default.name

  allow {
    protocol = "tcp"
    ports    = ["80", "8080", "1000-2000"]
  }

  allow {
    protocol = "icmp"
  }

  source_tags = ["web"]
}
```
but also to

```
resource "google_compute_firewall" "default" {
  name    = "test-firewall"
  network = google_compute_network.default.name

  allow {
    protocol = "tcp"
    ports    = ["80", "8080", "1000-2000"]
  }

  source_tags = ["web"]
}
```

trodge: I've decided to do this in a better way in the next PR.

zli82016: I have a thought. We can sort the array of the nested objects [in this line](https://github.com/GoogleCloudPlatform/magic-modules/pull/14490/files#diff-db94763256bb8c7fd7d94abbcd8272a0a5b0ae2f91f32228b7b34e49e18c8194R43)  before flattening the attributes. 

One option is to  hash the nested objects in the arrays and then sort the array by the items' hashed values. 

The function [encodeToBase64JSON](https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/third_party/terraform/acctest/tgc_utils.go#L63) can be used to hash the nested objects.

```
allow.0: eyJwb3J0cyI6e30sInByb3RvY29sIjp7fX0= 
allow.1: eyJwcm90b2NvbCI6e319 
```

This is just my thought. Feel free to explore it if you have different idea.

trodge: I have a very similar idea

PR #14491, opened 2025-07-09
tgc-revival: support google_compute_subnetwork

zli82016:

```release-note:none

```

### DIFF
--- a/mmv1/products/compute/Subnetwork.yaml
+++ b/mmv1/products/compute/Subnetwork.yaml
@@ -45,6 +45,7 @@ references:
 docs:
 base_url: 'projects/{{project}}/regions/{{region}}/subnetworks'
 has_self_link: true
+include_in_tgc_next_DO_NOT_USE: true
 immutable: true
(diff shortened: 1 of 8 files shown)


mmv1/products/compute/Subnetwork.yaml
```diff
@@ -472,6 +475,7 @@ properties:
   - name: 'ipCollection'
     type: String
     ignore_read: true
+    is_missing_in_cai: true
```
hao-nan-li, line 478: Does this mean this field is not existent in CAI?

zli82016: Right. This field is in API object, but not in CAI asset.

hao-nan-li (approved):

zli82016:
cc @trodge

zli82016:
Made the changes in this PR to automatically check if the referenced resource is in the same product or not. We don't need to manually handle the case that the referenced resource is not in the same product.

PR #14499, opened 2025-07-10
Fix TestAccDNSManagedZone_dnsManagedZoneCloudLoggingExample

nbodepudi:

```release-note:none
dns: Fix broken TestAccDNSManagedZone_dnsManagedZoneCloudLoggingExample 
```

### DIFF
--- a/mmv1/products/dns/ManagedZone.yaml
+++ b/mmv1/products/dns/ManagedZone.yaml
@@ -116,6 +116,7 @@ examples:
     primary_resource_id: 'cloud-logging-enabled-zone'
     vars:
       zone_name: 'cloud-logging-enabled-zone'
+      dns_name: 'fmt.Sprintf("services.example.com-%s.", context["random_suffix"])'
 virtual_fields:
   - name: 'force_destroy'
     description: 'Set this true to delete all records in the zone.'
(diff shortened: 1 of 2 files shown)


nbodepudi:
@rileykarson, can you take a look at it?

rileykarson:
Sorry for the delay! Approved the test run.

mmv1/products/dns/ManagedZone.yaml
```diff
@@ -116,6 +116,7 @@ examples:
     primary_resource_id: 'cloud-logging-enabled-zone'
     vars:
       zone_name: 'cloud-logging-enabled-zone'
+      dns_name: 'fmt.Sprintf("services.example.com-%s.", context["random_suffix"])'
```
rileykarson: I think we want to use the docs value (`services.example.com.`) here and can do the fmt in a `test_vars_overrides` entry! https://github.com/GoogleCloudPlatform/magic-modules/blob/95927b93ff81923ab376c5e2abbd283827202261/mmv1/products/workstations/WorkstationCluster.yaml#L70-L72 is a similar example. `random_suffix` isn't accessible because it doesn't exist yet at that point iirc so we just need a fresh suffix.

nbodepudi: Thanks for the suggestion, I am bit new to terraform. Looks like it worked now.

PR #14500, opened 2025-07-10
tgc-revival: sort all list elements before comparing hcl

trodge:

```release-note:none

```

### DIFF
--- a/mmv1/third_party/tgc_next/test/hcl.go
+++ b/mmv1/third_party/tgc_next/test/hcl.go
@@ -3,6 +3,8 @@ package test
 import (
 	"fmt"
 	"log"
+	"sort"
+	"strings"
 
 	"github.com/hashicorp/hcl/v2"
 	"github.com/hashicorp/hcl/v2/hclparse"
@@ -94,9 +96,9 @@ func insert(data any, key string, parent map[string]any) {
 	}
 }
 
-func flatten(data interface{}, prefix string, result map[string]struct{}) {
+func flatten(data any, prefix string, result map[string]struct{}) {
 	switch v := data.(type) {
-	case map[string]interface{}:
+	case map[string]any:
 		for key, value := range v {
 			newPrefix := key
 			if prefix != "" {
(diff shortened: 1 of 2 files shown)


mmv1/third_party/tgc_next/test/hcl.go
```diff
@@ -94,27 +96,69 @@ func insert(data any, key string, parent map[string]any) {
...
+		return
+	}
+
+	type sortableElement struct {
+		flatKeys  string
+		flattened map[string]struct{}
+	}
+
+	sortable := make([]sortableElement, len(v))
+	for i, value := range v {
```
hao-nan-li, line 130: If I understood correctly, this function sorts top level elements with the same field name by checking the sub-fields keys, and sort them in alphabetical order?

For example:
allow {
   test: "test"
}

allow {
  "port": "port"
}

into allow {"port": "port", "test", "test"}

mmv1/third_party/tgc_next/test/hcl_test.go
```diff
@@ -0,0 +1,167 @@
+package test
+
+import (
+	"testing"
+
+	"github.com/google/go-cmp/cmp"
+)
+
+var (
+	basicHCL = `resource "google_project_service" "project" {
```
hao-nan-li: Could you fix the indentation?

mmv1/third_party/tgc_next/test/hcl.go
trodge: Yes, it will sort and flatten them into `allow.0.port` and `allow.1.test`

hao-nan-li:
@zli82016 cc

mmv1/third_party/tgc_next/test/hcl_test.go
```diff
@@ -0,0 +1,172 @@
...
+
+  allow {
+    protocol = "tcp"
+    ports    = ["80", "8080", "1000-2000"]
+  }
+
+  source_tags = ["web"]
+}
+`
+	listOfMultiLevelNestedObjectsHCL = `
```
zli82016, line 57: This test case is the same with `listOfNestedObjectsHCL`.  Should it be a different test case?

trodge: Ah, yes it should. Adding that now.

PR #14519, opened 2025-07-14
Add support for GPU redundancy to Cloud Run v2 job

yanweiguo:
Adds support for GPU redundancy to Cloud Run v2 job.

```release-note:enhancement
cloudrunv2: added `gpu_zonal_redundancy_disabled` field to `google_cloud_run_v2_job` resource.
```


melinath:

Reassigning since @c2thorn is OOO

yanweiguo:
> Just to double-check, would it be possible to _not_ use default_from_api? That is, any chance this isn't a complex default?

There's a default value currently set from API side. We are going to remove that default value. In order to do that, we need to support TF. In that case, is it ok to not use default_from_api at this point? Or is it better to implement it with default_from_api now and remove it later when we remove the default value from API side?

yanweiguo:
> If "no default" means that the field will default to false 

At this moment, it defaults to `true` explicitly. It would default to `false` after TF support is added. I removed `default_from_api `.

> please make sure there's an update test that tries to set this field from true to false/unset

At this moment, API would throw an error is this field is set to `false` or unset. It is planned to be allowed in the future. A test added now would be broken when API changes the behavior in the future. Also to use `false` or unset value, the test projects should have zonal redundancy GPU quota, which requires quota increase request for each test project. So I didn't add a test for this.

melinath:
After discussion offline - we're going to move this forward with default_from_api: true + send_empty_value: true and no test for `false` for now.

yanweiguo:
> send_empty_value: true

Done

melinath (changes requested):
looks like you need to add the new field to https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/third_party/tgc/tests/data/example_cloud_run_v2_job.json as well (right after maxRetries should be fine) - false should be fine as a value even though the API won't allow it currently.

yanweiguo:
> looks like you need to add the new field to https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/third_party/tgc/tests/data/example_cloud_run_v2_job.json as well (right after maxRetries should be fine) - false should be fine as a value even though the API won't allow it currently.

What does https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/third_party/tgc/tests/data/example_cloud_run_v2_job.json do? Does it to be a valid example? The field is not allowed if GPU resource is not configured. Wondering whether I need to configure this field as well as the GPU resource.

melinath:
Merging since @paridhishah18 approved an earlier version and it hasn't changed substantially since then.

PR #14522, opened 2025-07-14
Add HA Policy to Regional Backend Services

uberferg:
Adds Terraform support for HA Policy as per http://b/353278740 and https://github.com/hashicorp/terraform-provider-google/issues/22615

```release-note:enhancement
compute: added `ha_policy` field to `google_compute_region_backend_service` resource
```


uberferg:
I noticed the bot added the "compute-l7-load-balancer" label, but this change is not for L7LB. I think the label should be "compute-regional-pass-through-lb".

Seems the bot did the same thing on the issue I created a while ago (https://github.com/hashicorp/terraform-provider-google/issues/22615) and I didn't notice. Not sure how to change the label in either location though...

uberferg:
Hi @rileykarson sorry for the delay. Finally got the VCR tests working, should be ready for review now.

mmv1/templates/terraform/examples/region_backend_service_ha_policy.tf.tmpl
```diff
@@ -0,0 +1,17 @@
+resource "google_compute_network" "default" {
```
rileykarson, line 1: This & the file below have mixed tabs and spaces that make them fairly messy- IDEs can be a bit annoying about template files and assign the wrong behaviour to tab. As HCL (.tf) files both can be entirely spaces, with the left side being two-spaced.

mmv1/products/compute/RegionBackendService.yaml
```diff
@@ -1472,3 +1485,62 @@ properties:
...
+                       network endpoint that should become the new leader sends either a
+                       Gratuitous ARP (GARP) packet (IPv4) or an ICMPv6 Router Advertisement(RA)
+                       packet (IPv6). Google Cloud immediately but temporarily associates the
+                       forwarding rule IP address with that VM, and both new and in-flight packets
+                       are quickly delivered to that VM.
+        immutable: true
+        enum_values:
+          - 'DISABLED'
+          - 'GARP_RA'
+      - name: 'leader'
```
rileykarson, line 1526: Could you add a test that changes the `leader` for a GARP_RA RBS?

uberferg: I think this should be covered by the the ManualLeaderExample .tf.tmpl. When adding that template, I had to spend a while debugging errors caused by updating an existing "leader" field, so it seems to follow the generated tests from that template cover changing/updating this field? Or perhaps I misunderstood the errors I was getting

rileykarson: That will set all the values on create, where we want to ensure users can successfully update an already-created resource as well!

uberferg: Added an update test for the leader. Sorry that took me a while to figure out how to do it.

mmv1/templates/terraform/examples/region_backend_service_ha_policy.tf.tmpl
uberferg: Done.

uberferg:
@rileykarson just a nudge to review this PR when you get the chance :)

rileykarson (approved):
LGTM, thanks!

PR #14527, opened 2025-07-15
Router tags at creation

Gorlami96:
Router tagging at creation Hz - b/323924842

```
release-note:compute: add `params.resourceManagerTags` field to the `google_compute_router`
```

### DIFF
--- a/mmv1/third_party/terraform/services/compute/resource_compute_router_test.go.tmpl
+++ b/mmv1/third_party/terraform/services/compute/resource_compute_router_test.go.tmpl
@@ -7,6 +7,7 @@ import (
 	"github.com/hashicorp/terraform-plugin-testing/helper/resource"
 	"github.com/hashicorp/terraform-provider-google/google/acctest"
 	"github.com/hashicorp/terraform-provider-google/google/envvar"
+	"github.com/hashicorp/terraform-plugin-testing/terraform"
 
 	{{- if ne $.TargetVersionName "ga" }}
 	compute "google.golang.org/api/compute/v0.beta"
@@ -227,7 +228,11 @@ func TestAccComputeRouter_resourceManagerTags(t *testing.T) {
 	sharedTagkey,_ := tagKeyResult["shared_tag_key"]
 	tagValueResult := acctest.BootstrapSharedTestTagValueDetails(t, "crm-routers-tagvalue", sharedTagkey, org)
 	routerName := fmt.Sprintf("tf-test-router-resource-manager-tags-%s", suffixName)
+	networkName := fmt.Sprintf("tf-test-network-resource-manager-tags-%s-net", suffixName)
+	subnetName := fmt.Sprintf("tf-test-subnet-resource-manager-tags-%s-subnet", suffixName)
 	context := map[string]interface{}{
+		"network_name": networkName,
+		"subnet_name": subnetName,
 		"router_name": routerName,
 		"tag_key_id": tagKeyResult["name"],
 		"tag_value_id": tagValueResult["name"],
@@ -458,7 +463,7 @@ func testAccCheckComputeRouterExists(t *testing.T, n string, router *compute.Rou
 }
 
 func testAccComputeRouter_resourceManagerTags(context map[string]interface{}) string {
-	return fmt.Sprintf(`
+	return acctest.Nprintf(`
 		resource "google_compute_network" "foobar" {
 		name                    = "%s-net"
 		auto_create_subnetworks = false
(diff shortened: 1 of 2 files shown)


mmv1/third_party/terraform/services/compute/resource_compute_router_test.go.tmpl
```diff
@@ -211,6 +219,46 @@ func TestAccComputeRouter_addAndUpdateIdentifierRangeBgp(t *testing.T) {
...
+		"router_name": routerName,
+		"tag_key_id": tagKeyResult["name"],
+		"tag_value_id": tagValueResult["name"],
+	}
+
+	acctest.VcrTest(t, resource.TestCase{
+		PreCheck:                 func() { acctest.AccTestPreCheck(t) },
+		ProtoV5ProviderFactories: acctest.ProtoV5ProviderBetaFactories(t),
+		CheckDestroy:             testAccCheckComputeRouterDestroyProducer(t),
+		Steps: []resource.TestStep{
```
rileykarson, line 239: Can you add an import step, similar to other cases, or explain why one doesn't work? We should also be able to remove the `Check` func as the import step makes it redundant.

mmv1/third_party/terraform/services/compute/resource_compute_router_test.go.tmpl
```diff
@@ -211,6 +219,46 @@ func TestAccComputeRouter_addAndUpdateIdentifierRangeBgp(t *testing.T) {
...
+		"network_name": networkName,
+		"subnet_name": subnetName,
+		"router_name": routerName,
+		"tag_key_id": tagKeyResult["name"],
+		"tag_value_id": tagValueResult["name"],
+	}
+
+	acctest.VcrTest(t, resource.TestCase{
+		PreCheck:                 func() { acctest.AccTestPreCheck(t) },
+		ProtoV5ProviderFactories: acctest.ProtoV5ProviderBetaFactories(t),
```
rileykarson, line 237:
```suggestion
		ProtoV5ProviderFactories: acctest.ProtoV5ProviderFactories(t),
```

Gorlami96: The new params field being tested is only available in beta. Will using ProtoV5ProviderFactories work?

mmv1/third_party/terraform/services/compute/resource_compute_router_test.go.tmpl
Gorlami96: I was following the examples of https://github.com/GoogleCloudPlatform/magic-modules/pull/14119 where the tagging support was added for VPC. The tags are not returned as part of GET call so I suppose they were just checking for existence of resource. But if import step does that I am fine with using Import.

mmv1/third_party/terraform/services/compute/resource_compute_router_test.go.tmpl
rileykarson: Yes! The standard factory aliases the beta provider as `google` in the test when run from the beta provider repo, while the BetaFactories factory uses its external `google-beta` alias. This only matters for examples where we want to use the appropriate external alias, and tests can use the `google` one for ease of promotion later.

mmv1/third_party/terraform/services/compute/resource_compute_router_test.go.tmpl
```diff
@@ -387,3 +426,39 @@ resource "google_compute_router" "foobar" {
...
+		name                    = "%{network_name}"
+		auto_create_subnetworks = false
+		}
+
+		resource "google_compute_subnetwork" "foobar" {
+		provider                = google-beta
+		name          = "%{subnet_name}"
+		network       = google_compute_network.foobar.self_link
+		ip_cidr_range = "10.0.0.0/16"
+		region        = "%s"
```
rileykarson: We need to fix the value, use `context`, or omit the field. I'd recommend omitting it, and using the test default.

mmv1/third_party/terraform/services/compute/resource_compute_router_test.go.tmpl
rileykarson: You can ImportStateVerifyIgnore `tags`!

mmv1/third_party/terraform/services/compute/resource_compute_router_test.go.tmpl
Gorlami96: Not sure but the test starts failing with Inconsistent Lock file error when this is changed to ProtoV5ProviderFactories. So leaving this as it is for now.

Gorlami96:
@rileykarson can you PTAL?

rileykarson (approved):
LGTM, thanks!

PR #14530, opened 2025-07-15
add Developer Connect Triggers to Cloudbuild

MattLaddy:

```release-note:enhancement
cloudbuild: added `developer_connect_event_config` field to `google_cloudbuild_trigger` resource
```


SirGitsalot (changes requested):
Looks good, but please add an example a la https://github.com/GoogleCloudPlatform/magic-modules/blob/ba3f222be34ecf70b7195219db164fbca4e81da8/mmv1/templates/terraform/examples/cloudbuild_trigger_github.tf.tmpl and add it to the list of examples up around line 55 (there's a bunch of Cloud Build trigger examples in the mmv1/templates/terraform/examples/ you can copy and paste from). Thanks!

mmv1/templates/terraform/examples/cloudbuild_trigger_developer_connect.tf.tmpl
```diff
@@ -0,0 +1,12 @@
+resource "google_cloudbuild_trigger" "{{$.PrimaryResourceId}}" {
+  location = "us-central1"
+
+  developer_connect_event_config {
+    git_repository_link = "projects/ci-test-project-188019/locations/us-central1/connections/my-connection/gitRepositoryLinks/YOUR_GIT_REPOSITORY_LINK_NAME"
```
SirGitsalot:
```suggestion
    git_repository_link = "projects/{{index $.TestEnvVars "project_id"}}/locations/us-central1/connections/my-connection/gitRepositoryLinks/YOUR_GIT_REPOSITORY_LINK_NAME"
```

That should work with whatever project the test happens to be running in.

SirGitsalot (commented):
Does the backend care if the connection actually exists? If so, [this example](https://github.com/GoogleCloudPlatform/magic-modules/blob/ab923c8c57704663a42c375ab0ccb4df2202b1a0/mmv1/templates/terraform/examples/cloudbuild_trigger_pubsub_with_repo.tf.tmpl) shows how to set up a connection before creating the trigger. That may explain the service account error (I'm  not very familiar with the Cloud Build API, but some backend return 403 instead of 404 - or vice versa -  when a resource is not found in order to avoid leaking information)

MattLaddy:
Hey @SirGitsalot, Thank you for all the help. I think im still confused on what a proper template looks like and what this file is actually doing. I am setting a fake project ID on line 140 in the trigger.yaml. Should I be setting a real project ID somewhere else/ on that same line instead?

mmv1/templates/terraform/examples/cloudbuild_trigger_developer_connect.tf.tmpl
```diff
@@ -0,0 +1,12 @@
+resource "google_cloudbuild_trigger" "{{$.PrimaryResourceId}}" {
+  location = "us-central1"
+  
+  developer_connect_event_config {
+    git_repository_link = "projects/{{index $.Vars "project_id"}}/locations/us-central1/connections/my-connection/gitRepositoryLinks/YOUR_GIT_REPOSITORY_LINK_NAME"
```
SirGitsalot:
```suggestion
    git_repository_link = "projects/{{index $.TestEnvVars "project_name"}}/locations/us-central1/connections/my-connection/gitRepositoryLinks/YOUR_GIT_REPOSITORY_LINK_NAME"
```

mmv1/products/cloudbuild/Trigger.yaml
```diff
@@ -134,6 +134,10 @@ examples:
       'installation_id': '31300675'
       'pat_secret': '"projects/gcb-terraform-creds/secrets/github-pat/versions/latest"'
       'repo_uri': '"https://github.com/gcb-repos-robot/tf-demo.git"'
+  - name: 'cloudbuild_trigger_developer_connect'
+    primary_resource_id: 'developer-connect-trigger'
+    vars:
```
SirGitsalot:
```suggestion
    test_env_vars:
      project_name: 'PROJECT_NAME'
```

mmv1/products/cloudbuild/Trigger.yaml
```diff
@@ -134,6 +134,10 @@ examples:
...
+      'project_id': "PROJECT_ID"
```
SirGitsalot:
```suggestion
```

SirGitsalot (commented):
> Should I be setting a real project ID somewhere else/ on that same line instead?

I updated my broken code suggestions :-(

It's not obvious, but the variables in example templates come from two sources: `vars`, which defines the variables in the resource YAML file (and which can include inlined Go code), and `test_env_vars` which draws them from environment variables. When you're testing locally the env vars are set [in your shell](https://googlecloudplatform.github.io/magic-modules/test/run-tests/#before-you-begin) and when the tests are running in the CI system they're preset for whichever CI system they're running on (we have several, one for nightly tests, one for running PR checks, one for beta, and more) and come from one of [several possible env var names](https://github.com/GoogleCloudPlatform/magic-modules/blob/4ab6c4db0aa1540b58a15a5d88028996507d7f7a/mmv1/third_party/terraform/envvar/envvar_utils.go#L38).

MattLaddy:
It appears that the syntax is correct but looks like im getting 403 errors on the service account. 

Im also unrelated having issues doing the local testing that you mentioned. Im setting local variables but running into Santa blocking my command from being run as its saying some of the trustworthiness of the applications being run are in quesiton. 

Thanks in advance for the help

SirGitsalot (commented):
You can get a Santa exception to run arbitrary binaries: go/terraform-team-noogler#initialize-work-environment

Being able to run the integration test locally will be muuuuuuuuch less painful than waiting for our CI runs.

MattLaddy:
Hey @SirGitsalot finally got these acceptance tests working 😆. let me know if theres anything else I need to do before submitting. Thanks!

SirGitsalot (approved):

PR #14532, opened 2025-07-15
Add deletion policy to Secure Source Manager repository

ericpang777:

```release-note:enhancement
securesourcemanager: added `deletion_policy` field to `google_secure_source_manager_repository` resource
```

### DIFF
--- a/mmv1/products/securesourcemanager/Repository.yaml
+++ b/mmv1/products/securesourcemanager/Repository.yaml
@@ -47,7 +47,7 @@ iam_policy:
     - 'projects/{{project}}/locations/{{location}}/repositories/{{repository_id}}'
     - '{{repository_id}}'
 custom_code:
-  pre_delete: 'templates/terraform/pre_delete/securesourcemanager_repository.go.tmpl'
+  pre_delete: 'templates/terraform/pre_delete/securesourcemanager_deletion_policy.go.tmpl'
 examples:
   - name: 'secure_source_manager_repository_basic'
     primary_resource_id: 'default'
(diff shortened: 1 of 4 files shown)


ericpang777:
@c2thorn  Hi could you take a look? I would like this change to be in the same release as https://github.com/GoogleCloudPlatform/magic-modules/pull/14454 so it doesn't make the docs ugly (one resource has deletion policy, the other has lifecycle rules)

mmv1/products/securesourcemanager/Repository.yaml
```diff
@@ -91,6 +86,18 @@ parameters:
       The ID for the Repository.
     url_param_only: true
     required: true
+virtual_fields:
+  - name: 'deletion_policy'
+    type: String
+    description: |
+      The deletion policy for the repository. Setting `ABANDON` allows the resource
+      to be abandoned, rather than deleted. Setting `DELETE` deletes the resource
+      and all its contents. Setting `PREVENT` prevents the resource from being deleted.
```
c2thorn:
```suggestion
      and all its contents. Setting `PREVENT` prevents the resource from accidental deletion by erroring out during plan.
```
Do you think this is more useful for users to make the behavior and incentive for PREVENT more clear?

mmv1/templates/terraform/pre_delete/securesourcemanager_repository.go.tmpl
```diff
@@ -0,0 +1,7 @@
+deletionPolicy := d.Get("deletion_policy");
```
c2thorn: We may want to actually generalize [pre_delete/securesourcemanager_instance.go.tmpl](https://github.com/ericpang777/magic-modules/blob/1228d87cc2b4c3b2d5663d7b42208929f999ef3e/mmv1/templates/terraform/pre_delete/securesourcemanager_instance.go.tmpl) to be used for both and prevent later possible drift 

I don't think the error message difference is worthwhile.

ericpang777: Done, made the error message generic.

mmv1/products/securesourcemanager/Repository.yaml
ericpang777: Thanks! Good point, updated.

c2thorn (approved):
LGTM thank you. Planning to merge tomorrow.

PR #14534, opened 2025-07-15
Update dataproc Batch and dataproc SessionTemplate resource to support authentication config.

animeshnandanwar:
Update dataproc Batch and dataproc SessionTemplate resource to support authentication config.
The API reference is still not updated, but WIP.

```release-note:enhancement
dataproc: added `authentication_config` field to `google_dataproc_batch` and `google_dataproc_session_template` resource
```

### DIFF
--- a/mmv1/products/dataproc/Batch.yaml
+++ b/mmv1/products/dataproc/Batch.yaml
@@ -377,6 +377,19 @@ properties:
               Subnetwork configuration for workload execution.
             conflicts:
               - environment_config.0.execution_config.0.network_uri
+          - name: 'authenticationConfig'
+            type: NestedObject
+            description: |
+              Authentication configuration for a workload is used to set the default identity for the workload execution.
+            properties:
+              - name: userWorkloadAuthenticationType
+                type: Enum
+                description: |
+                  Authentication type for the user workload running in containers.
+                enum_values:
+                  - AUTHENTICATION_TYPE_UNSPECIFIED
+                  - SERVICE_ACCOUNT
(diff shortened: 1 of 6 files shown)


mmv1/products/dataproc/Batch.yaml
```diff
@@ -377,6 +377,19 @@ properties:
...
+            type: NestedObject
+            description: |
+              Authentication configuration for a workload is used to set the default identity for the workload execution.
+            properties:
+              - name: userWorkloadAuthenticationType
+                type: Enum
+                description: |
+                  Authentication type for the user workload running in containers.
+                enum_values:
+                  - AUTHENTICATION_TYPE_UNSPECIFIED
```
shuyama1:
```suggestion
```
Is `AUTHENTICATION_TYPE_UNSPECIFIED` a value that we'd expect users to specify explicitly? If not, we can just remove it from the available value list

mmv1/products/dataproc/SessionTemplate.yaml
```diff
@@ -189,6 +189,19 @@ properties:
...
+            type: NestedObject
+            description: |
+              Authentication configuration for a workload is used to set the default identity for the workload execution.
+            properties:
+              - name: userWorkloadAuthenticationType
+                type: Enum
+                description: |
+                  Authentication type for the user workload running in containers.
+                enum_values:
+                  - AUTHENTICATION_TYPE_UNSPECIFIED
```
shuyama1:
```suggestion
```
same question here

PR #14539, opened 2025-07-16
tgc-revival: support cloudfunctions2 function in tgc next

trodge:

```release-note:none

```

### DIFF
--- a/mmv1/api/type.go
+++ b/mmv1/api/type.go
@@ -913,7 +913,7 @@ func (t Type) UserProperties() []*Type {
 
 		return google.Reject(t.Properties, func(p *Type) bool {
 			if t.ResourceMetadata.IsTgcCompiler() {
-				return p.Exclude || p.Output || p.IsMissingInCai
+				return p.Exclude || p.Output
 			}
 			return p.Exclude
 		})


zli82016:
Can you try to fix the test `TestAccComputeBackendService_backendServiceExternalIapExample`? Thanks.

```
=== CONT  TestAccComputeBackendService_backendServiceExternalIapExample
    assert_test_files.go:56: Test for the primary resource google_compute_backend_service.default begins.
    assert_test_files.go:59: Test fails: missing fields in resource google_compute_backend_service.default after cai2hcl conversion:
        [iap.oauth2_client_secret]
--- FAIL: TestAccComputeBackendService_backendServiceExternalIapExample (0.00s)
```

mmv1/api/type.go
```diff
@@ -912,6 +912,9 @@ func (t Type) UserProperties() []*Type {
 		}
 
 		return google.Reject(t.Properties, func(p *Type) bool {
+			if t.ResourceMetadata.IsTgcCompiler() {
```
zli82016, line 915: This change also affects `AllUserProperties`, as AllUserProperties = UserProperites + UserParameters
```
func (r Resource) AllUserProperties() []*Type {
	return google.Concat(r.UserProperites(), r.UserParameters())
}
```

`AllUserProperties` is also used to render the schemas for resources https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/templates/tgc_next/services/resource.go.tmpl#L43

I think this is the reason for the test failure of `TestAccComputeBackendService_backendServiceExternalIapExample`.

Instead of modifying this function, we add the logic to the [flatten_property_method.go.tmpl](https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/templates/terraform/flatten_property_method.go.tmpl)

zli82016 (changes requested):

mmv1/api/type.go
```diff
@@ -912,6 +912,9 @@ func (t Type) UserProperties() []*Type {
...
+				return p.Exclude || p.Output || p.IsMissingInCai
```
trodge:
```suggestion
				return p.Exclude || p.Output
```

mmv1/api/type.go
trodge: We might still want to exclude output fields when making schemas for resources, since they won't be useful in either tfplan2cai or cai2hcl.

zli82016: That makes sense.

zli82016 (approved):
Thanks.

PR #14543, opened 2025-07-16
fixing permadiff create_without_validation

luckyswaminathan:
Fixes https://github.com/hashicorp/terraform-provider-google/issues/21780 - a permadiff regarding the create_without_validation field within datastream

```release-note: bug
datastore: Fixed a permadiff related to datastream 'create_without_validation' field
```

### DIFF
--- a/mmv1/templates/terraform/constants/datastream_connection_profile.go.tmpl
+++ b/mmv1/templates/terraform/constants/datastream_connection_profile.go.tmpl
@@ -14,7 +14,7 @@
 func resourceDataStreamStreamCreateWithoutValidationDiffSuppress(k, old, new string, d *schema.ResourceData) bool {
 	// If the old value was "false" and the new value is now unset (empty string),
 	// return true to suppress the diff.
-	if old == "false" && new == "" {
+	if (old == "" && new == "false") || (old == "false" && new == "") {
 		d.Set("create_without_validation", "false")
 		return true
 	}
--- a/mmv1/templates/terraform/constants/datastream_connection_profile.go.tmpl
+++ b/mmv1/templates/terraform/constants/datastream_connection_profile.go.tmpl
@@ -1,0 +1,24 @@
+{{/*
+	The license inside this block applies to this file
+	Copyright 2024 Google Inc.
+	Licensed under the Apache License, Version 2.0 (the "License");
+	you may not use this file except in compliance with the License.
+	You may obtain a copy of the License at http://www.apache.org/licenses/LICENSE-2.0
+	Unless required by applicable law or agreed to in writing, software
+	distributed under the License is distributed on an "AS IS" BASIS,
+	WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+	See the License for the specific language governing permissions and
+	limitations under the License.
+*/ -}}
+
+func resourceDataStreamStreamCreateWithoutValidationDiffSuppress(k, old, new string, d *schema.ResourceData) bool {
+	// If the old value was "false" and the new value is now unset (empty string),
+	// return true to suppress the diff.
+	if old == "false" && new == "" {
+		d.Set("create_without_validation", "false")
+		return true
+	}
+
+	// Otherwise, do not suppress the diff.
+	return false
+}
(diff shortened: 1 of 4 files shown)


mmv1/templates/terraform/constants/datastream_connection_profile.go.tmpl
```diff
@@ -0,0 +1,24 @@
...
+	WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+	See the License for the specific language governing permissions and
+	limitations under the License.
+*/ -}}
+
+func resourceDataStreamStreamCreateWithoutValidationDiffSuppress(k, old, new string, d *schema.ResourceData) bool {
+	// If the old value was "false" and the new value is now unset (empty string),
+	// return true to suppress the diff.
+	if (old == "" && new == "false") || (old == "false" && new == "") {
+		d.Set("create_without_validation", "false")
```
trodge: This is what's causing the panic, because terraform expects a boolean. I would try putting
```
d.Set("create_without_validation", false)
```
in a `post_import`.

mmv1/templates/terraform/constants/datastream_connection_profile.go.tmpl
```diff
@@ -0,0 +1,23 @@
...
+	Unless required by applicable law or agreed to in writing, software
+	distributed under the License is distributed on an "AS IS" BASIS,
+	WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+	See the License for the specific language governing permissions and
+	limitations under the License.
+*/ -}}
+
+func resourceDataStreamStreamCreateWithoutValidationDiffSuppress(k, old, new string, d *schema.ResourceData) bool {
+	// If the old value was "false" and the new value is now unset (empty string),
+	// return true to suppress the diff.
```
trodge, line 16: This describes what the conditional does, but the important thing to explain here is why we need to do it. You can either link the issue in the comment or summarize it.

mmv1/templates/terraform/pre_create/datastream_connection_profile.go.tmpl
```diff
@@ -0,0 +1,18 @@
...
+	WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+	See the License for the specific language governing permissions and
+	limitations under the License.
+*/ -}}
+
+_, exists := d.GetOk("create_without_validation")
+if !exists {
+	url += "false"
+	d.Set("create_without_validation", false)
+}
```
trodge: Sorry for the back and forth here, but after reviewing [a discussion about `GetOkExists`](https://discuss.hashicorp.com/t/terraform-sdk-usage-which-out-of-get-getok-getokexists-with-boolean/41815) I think our best approach is actually to get the url parameters out of the base URL entirely.

So that would mean deleting `&force={{create_without_validation}}` from the `create_url` and calling `transport.AddQueryParams` here with something like `map[string]string{"force": "true"}` only if `d.Get("create_without_validation")` returns `true`.

We might also want to make sure there's at least one test that creates the resource with the field set to false or unspecified.

luckyswaminathan: would this not be a breaking change?

trodge: Is the `force` url parameter required in the API endpoint?

mmv1/templates/terraform/pre_create/datastream_connection_profile.go.tmpl
```diff
@@ -11,8 +11,8 @@
...
-if !exists {
-	url += "false"
-	d.Set("create_without_validation", false)
-}
\ No newline at end of file
+if d.Get("create_without_validation").(bool) {
+		url, err = transport_tpg.AddQueryParams(url, map[string]string{"force": "true"})
+	} else {
+		url, err = transport_tpg.AddQueryParams(url, map[string]string{"force": "false"})
+	}
```
trodge: nit: the indentation is looking weird to me, although it should be fine as long as gofmt still works on it.

mmv1/templates/terraform/pre_create/datastream_connection_profile.go.tmpl
trodge: It would also be a breaking change if `force=false` has different behavior than omitting `force` altogether. Keeping it as is should be safest.

trodge (approved):

PR #14545, opened 2025-07-16
tgc-revival: add google_compute_network

zli82016:

```release-note:none

```

### DIFF
--- a/mmv1/api/type.go
+++ b/mmv1/api/type.go
@@ -434,7 +434,7 @@ func (t Type) Lineage() string {
 // This format is intended for resource metadata, to be used for connecting a Terraform
 // type with a corresponding API type.
 func (t Type) MetadataLineage() string {
-	if t.ParentMetadata == nil {
+	if t.ParentMetadata == nil || t.ParentMetadata.FlattenObject {
 		return google.Underscore(t.Name)
 	}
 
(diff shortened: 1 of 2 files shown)


mmv1/api/type.go
```diff
@@ -434,7 +434,7 @@ func (t Type) Lineage() string {
 // This format is intended for resource metadata, to be used for connecting a Terraform
 // type with a corresponding API type.
 func (t Type) MetadataLineage() string {
-	if t.ParentMetadata == nil {
+	if t.ParentMetadata == nil || t.ParentMetadata.FlattenObject {
```
zli82016, line 437: @roaks3, can you help to confirm if this change will break something? Thanks.

For example: `- field: 'dataset_reference.dataset_id'` is changed to `  - field: 'dataset_id'` in [google-beta/services/bigquery/resource_bigquery_dataset_generated_meta.yaml](https://github.com/modular-magician/terraform-provider-google-beta/compare/d426a62f8183a1f1dcd63096d69f64e73369a15f..c9c3741958e530cd7abd7bf23e5e0e0ec93ac50e)

roaks3: Yep, this looks correct, and appears to be logic that was missed previously. It shouldn't have any impact to coverage metrics from what I can tell, but it makes our data more accurate wrt the TF field.

Thanks!

roaks3 (approved):

mmv1/api/type.go
zli82016: np. Thanks for the confirmation.

PR #14547, opened 2025-07-16
Update service account creation to be more resilient to eventual cons…

chasevedder:
This changes service account creation to no longer require a final GetServiceAccount request to succeed. Instead, we will wait until we get 3 consecutive successful responses from GetServiceAccount before considering the create successful.

closes https://github.com/hashicorp/terraform-provider-google/issues/22546
```release-note:bug
iam: Update service account creation to prevent failures due to eventual consistency
```

### DIFF
--- a/mmv1/third_party/terraform/services/resourcemanager/resource_google_service_account.go
+++ b/mmv1/third_party/terraform/services/resourcemanager/resource_google_service_account.go
@@ -122,54 +122,60 @@ func resourceGoogleServiceAccountCreate(d *schema.ResourceData, meta interface{}
 		ServiceAccount: sa,
 	}
 
-	sa, err = config.NewIamClient(userAgent).Projects.ServiceAccounts.Create("projects/"+project, r).Do()
+
+	iamClient := config.NewIamClient(userAgent)
+	sa, err = iamClient.Projects.ServiceAccounts.Create("projects/"+project, r).Do()
 	if err != nil {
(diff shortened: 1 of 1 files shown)


mmv1/third_party/terraform/services/resourcemanager/resource_google_service_account.go
```diff
@@ -122,54 +122,59 @@ func resourceGoogleServiceAccountCreate(d *schema.ResourceData, meta interface{}
...
+						return saerr
+					}
+					return populateResourceData(d, sa)
+				},
+				Timeout: d.Timeout(schema.TimeoutCreate),
+				ErrorRetryPredicates: []transport_tpg.RetryErrorPredicateFunc{
+					transport_tpg.IsNotFoundRetryableError("service account creation"),
+				},
+			})
+
```
trodge, line 148: I think we still want to set id in this case. We only want to avoid setting id if creation fails (i.e. we got a non-409 error or `create_ignore_already_exists` is false.

trodge (changes requested):

mmv1/third_party/terraform/services/resourcemanager/resource_google_service_account.go
chasevedder: Updated

zli82016 (approved):

PR #14552, opened 2025-07-17
google_cloudfunctions_function base image policy fields 

dixuswe:
Adding 2 new fields to google_cloudfunctions_function (Cloud Function 1st gen) automatic_update_policy and on_deploy_update_policy (the former is the API default) 

```release-note: enhancement
cloudfunction: added `automatic_update_policy` and `on_deploy_update_policy` to google_cloudfunctions_function resource
```


mmv1/products/cloudfunctions/CloudFunction.yaml
maemayve, line 1: Do these new fields need to have a min version associated with them?

mmv1/third_party/terraform/services/cloudfunctions/resource_cloudfunctions_function.go
```diff
@@ -808,6 +848,24 @@ func resourceCloudFunctionsRead(d *schema.ResourceData, meta interface{}) error
 	if err := d.Set("version_id", strconv.FormatInt(function.VersionId, 10)); err != nil {
 		return fmt.Errorf("Error setting version_id: %s", err)
 	}
+	if function.OnDeployUpdatePolicy != nil {
```
maemayve, line 851: can you add a comment that says this order is mandatory? want to make sure that down the line someone doesn't reorder these two branches and put automatic update policy first which would break since it is alway set by the API as the default value

mmv1/third_party/terraform/services/cloudfunctions/resource_cloudfunctions_function_test.go.tmpl
maemayve, line 1: can you add a test case that goes automatic -> on deploy -> no policy specified?

Do you think it would be worth adding a test case that does CRUD the other direction? on deploy -> automatic -> on deploy?

mmv1/third_party/terraform/services/cloudfunctions/resource_cloudfunctions_function.go
dixuswe: I think the order doesn't really matter, but I am not quite sure. 
Added the comment

mmv1/third_party/terraform/services/cloudfunctions/resource_cloudfunctions_function_test.go.tmpl
dixuswe: the test case exactly do "automatic -> on deploy -> no policy specified"

I did more manually test with all possible combinations

mmv1/products/cloudfunctions/CloudFunction.yaml
dixuswe: this .yaml probably don't have any effect and I think it can go directly to GA

mmv1/products/cloudfunctions/CloudFunction.yaml
```diff
@@ -200,3 +200,30 @@ properties:
...
+    allow_empty_object: true
+    exactly_one_of:
+      - 'automatic_update_policy'
+      - 'on_deploy_update_policy'
+    properties:
+      - name: 'runtimeVersion'
+        type: String
+        description: |
+          The runtime version which was used during latest function deployment.
+        output: true
```
melinath: This configuration is unused & doesn't match what's in the handwritten code; we should just not include it.

```suggestion
```

mmv1/third_party/terraform/services/cloudfunctions/resource_cloudfunctions_function_meta.yaml
```diff
@@ -51,3 +51,6 @@ fields:
   - field: 'version_id'
   - field: 'vpc_connector'
   - field: 'vpc_connector_egress_settings'
+  - field: 'automatic_update_policy'
+  - field: 'on_deploy_update_policy'
```
melinath: Thanks for updating this file! However, it only needs to contain leaf fields.

```suggestion
```

mmv1/third_party/terraform/services/cloudfunctions/resource_cloudfunctions_function.go
```diff
@@ -1232,3 +1307,45 @@ func flattenSecretVersion(secretVersions []*cloudfunctions.SecretVersion) []map[
...
+		return nil
+	}
+	// Have to append an empty element for empty message type
+	result = append(result, map[string]interface{}{})
+	log.Printf("flatten automatic_update_policy to: %s", result)
+	return result
+}
+
+func expandOnDeployUpdatePolicy(configured []interface{}) *cloudfunctions.OnDeployUpdatePolicy {
+	log.Printf("on_deploy_update_policy configued: %s", configured)
```
melinath: nit: sp

```suggestion
	log.Printf("on_deploy_update_policy configured: %s", configured)
```

melinath (changes requested):
> Missing test report (`on_deploy_update_policy.runtime_version`)
> Missing doc report (`on_deploy_update_policy.runtime_version`)

Please fix these, thanks!

mmv1/products/cloudfunctions/CloudFunction.yaml
dixuswe: reverted change to this file

dixuswe:
> > Missing test report (`on_deploy_update_policy.runtime_version`)
> > Missing doc report (`on_deploy_update_policy.runtime_version`)
> 
> Please fix these, thanks!

this should now have been fixed

PR #14562, opened 2025-07-17
tgc-revival: support google_compute_global_address

zli82016:

```release-note:none

```

### DIFF
--- a/mmv1/api/resource.go
+++ b/mmv1/api/resource.go
@@ -226,16 +226,6 @@ type Resource struct {
 	// If true, resource is not importable
 	ExcludeImport bool `yaml:"exclude_import,omitempty"`
 
-	// If true, exclude resource from Terraform Validator
-	// (i.e. terraform-provider-conversion)
-	ExcludeTgc bool `yaml:"exclude_tgc,omitempty"`
-
-	// If true, include resource in the new package of TGC (terraform-provider-conversion)
-	IncludeInTGCNext bool `yaml:"include_in_tgc_next_DO_NOT_USE,omitempty"`
-
-	// Name of the hcl resource block used in TGC
-	TgcHclBlockName string `yaml:"tgc_hcl_block_name,omitempty"`
-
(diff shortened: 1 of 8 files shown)


mmv1/api/resource.go
```diff
@@ -362,6 +352,26 @@ type Resource struct {
 
 	ImportPath     string `yaml:"-"`
 	SourceYamlFile string `yaml:"-"`
+
```
trodge, line 355: I think we should move all this and the `Resource` methods that use it into a new type called `TGCResource` which embeds a `Resource`. That way we can at least be sure that nothing relies on this outside of TGC.

mmv1/api/resource.go
```diff
@@ -2060,9 +2070,9 @@ func (r Resource) ReadPropertiesForTgc() []*Type {
 // Rarely, it is the API "resource type kind".
 // For example, the API resource type of "google_compute_autoscaler" is "ComputeAutoscalerAssetType".
 // The API resource type of "google_compute_region_autoscaler" is also "ComputeAutoscalerAssetType".
-func (r Resource) ApiResourceType() string {
-	if r.ApiResourceTypeKind != "" {
-		return fmt.Sprintf("%s%s", r.ProductMetadata.Name, r.ApiResourceTypeKind)
+func (r Resource) CaiResourceType() string {
```
trodge, line 2077: If we're renaming this function we should consider also renaming [ResourcesGroupedByApiResourceType](https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/provider/terraform_tgc_next.go) to `ResourcesByCaiResourceType`

trodge (approved):

mmv1/api/resource.go
zli82016: Changed.

mmv1/api/resource.go
zli82016: Added the new type TGCResource

PR #14387, opened 2025-06-26
Make the subgroup include @ explicitly and trim it in the command

trodge:
Fixes [regression](https://github.com/GoogleCloudPlatform/magic-modules/pull/14385#issuecomment-3009730200)

**Release Note Template for Downstream PRs (will be copied)**

See [Write release notes](https://googlecloudplatform.github.io/magic-modules/contribute/release-notes/) for guidance.

```release-note:none

```

PR #14565, opened 2025-07-18
Tweaked reassign-reviewer regex

melinath:
Ran into a problem where this didn't work properly without an @ symbol even though the @ symbol is optional: https://github.com/GoogleCloudPlatform/magic-modules/pull/14519#issuecomment-3081283134 We should just make the `-er` optional instead of allowing arbitrary content there. We can also make the `re-` prefix optional.

```release-note:none

```

### DIFF
--- a/.github/workflows/reassign-reviewer.yml
+++ b/.github/workflows/reassign-reviewer.yml
@@ -22,7 +22,7 @@ jobs:
         uses: actions-ecosystem/action-regex-match@d50fd2e7a37d0e617aea3d7ada663bd56862b9cc # v2.0.2
         with:
           text: ${{ github.event.comment.body }}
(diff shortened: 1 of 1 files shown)


trodge (approved):

trodge:
The intent was actually to make the @ required if and only if reassigning to a specific persion so that they get pinged, but it's hard to do that in regex in a way that's readable. https://github.com/GoogleCloudPlatform/magic-modules/pull/14387 would do it more explicitly, but it still allows arbitrary content between "review" and "@"

edit- @melinath does this ping you?

melinath:
I think they should still get pinged by the review request and by the @-mention in the assignment comment.

edit: @trodge - here's an edit @ mention for you too!

melinath:
> edit- @melinath does this ping you?

@trodge yes

PR #14566, opened 2025-07-18
Fix teamcity runs for modelarmor

shuyama1:
This PR should fix the service name not matching issue for `modelarmor` and `modelarmorglobal`, which was introduced in https://github.com/GoogleCloudPlatform/magic-modules/pull/14457

```release-note:none

```


zli82016 (approved):
Thanks for fixing it.

PR #14483, opened 2025-07-09
Fixing hive partitioning diff

sachinpro:
Added a flag to prevent Terraform from showing diff for server generated schema columns(like hive partitioned ones)

b/420688695, b/300616880

Fixes https://github.com/hashicorp/terraform-provider-google/issues/12465, https://github.com/hashicorp/terraform-provider-google/issues/16907

**Release Note Template for Downstream PRs (will be copied)**

See [Write release notes](https://googlecloudplatform.github.io/magic-modules/contribute/release-notes/) for guidance.

```release-note:enhancement
bigquery: added `ignore_auto_generated_schema` virtual field to `google_bigquery_table` resource to ignore server-added columns in the `schema` field
```

PR #14569, opened 2025-07-18
fix(bigquery_dataset): fixed handling of non-legacy roles for access …

sachinpro:
Added a flag to prevent Terraform from showing diff for server generated schema columns(like hive partitioned ones)

Fixes https://github.com/hashicorp/terraform-provider-google/issues/8370

```release-note:enhancement
bigquery: fixed handling of non-legacy roles for access block inside `google_bigquery_dataset`
```

### DIFF
--- a/mmv1/templates/terraform/constants/bigquery_dataset.go.tmpl
+++ b/mmv1/templates/terraform/constants/bigquery_dataset.go.tmpl
@@ -1,5 +1,11 @@
 const datasetIdRegexp = `^[0-9A-Za-z_]+$`
 
+var bigqueryDatasetAccessPrimitiveToRoleMap = map[string]string{
+	"OWNER":  "roles/bigquery.dataOwner",
+	"WRITER": "roles/bigquery.dataEditor",
+	"READER": "roles/bigquery.dataViewer",
+}
+
 func validateDatasetId(v interface{}, k string) (ws []string, errors []error) {
     value := v.(string)
     if !regexp.MustCompile(datasetIdRegexp).MatchString(value) {
@@ -23,3 +29,30 @@ func validateDefaultTableExpirationMs(v interface{}, k string) (ws []string, err
 
     return
 }
+
+{{- if ne $.Compiler "terraformgoogleconversion-codegen" }}
+// bigqueryDatasetAccessHash is a custom hash function for the access block.
+// It normalizes the 'role' field before hashing, treating legacy roles
+// and their modern IAM equivalents as the same.
+func resourceBigqueryDatasetAccessHash(v interface{}) int {
+	m, ok := v.(map[string]interface{})
+	if !ok {
+		return 0
+	}
+	// Make a copy of the map to avoid modifying the underlying data.
+	copy := make(map[string]interface{}, len(m))
+	for k, val := range m {
+		copy[k] = val
+	}
+
+	// Normalize the role if it exists and matches a legacy role.
+	if role, ok := copy["role"].(string); ok {
+		if newRole, ok := bigqueryDatasetAccessPrimitiveToRoleMap[role]; ok {
+			copy["role"] = newRole
+		}
+	}
+
+	// Use the default HashResource function on the (potentially modified) copy.
+	return schema.HashResource(bigqueryDatasetAccessSchema())(copy)
+}
+{{- end }}
--- a/mmv1/templates/terraform/examples/bigquery_dataset_basic.tf.tmpl
+++ b/mmv1/templates/terraform/examples/bigquery_dataset_basic.tf.tmpl
@@ -10,7 +10,7 @@ resource "google_bigquery_dataset" "{{$.PrimaryResourceId}}" {
   }
 
   access {
-    role          = "OWNER"
+    role          = "roles/bigquery.dataOwner"
     user_by_email = google_service_account.bqowner.email
   }
 
--- a/mmv1/products/bigquery/Dataset.yaml
+++ b/mmv1/products/bigquery/Dataset.yaml
@@ -129,6 +129,7 @@ properties:
     is_set: true
     default_from_api: true
     is_missing_in_cai: true
+    set_hash_func: 'resourceBigqueryDatasetAccessHash'
     item_type:
       type: NestedObject
       properties:


sachinpro:
> 🔴 Tests failed during RECORDING mode: `TestAccBigQueryExternalDataTable_CSV_WithSchemaAndConnectionIDAndHivePartitioning` [[Error message](https://storage.cloud.google.com/ci-vcr-logs/beta/refs/heads/auto-pr-14569/artifacts/e790d2fb-bb83-4a62-9f9a-dc9bd7bfa566/build-log/recording_build/TestAccBigQueryExternalDataTable_CSV_WithSchemaAndConnectionIDAndHivePartitioning_recording_test.log)] [[Debug log](https://storage.cloud.google.com/ci-vcr-logs/beta/refs/heads/auto-pr-14569/artifacts/e790d2fb-bb83-4a62-9f9a-dc9bd7bfa566/recording/TestAccBigQueryExternalDataTable_CSV_WithSchemaAndConnectionIDAndHivePartitioning.log)]
> 
> 🔴 Errors occurred during RECORDING mode. Please fix them to complete your PR.
> 
> View the [build log](https://storage.cloud.google.com/ci-vcr-logs/beta/refs/heads/auto-pr-14569/artifacts/e790d2fb-bb83-4a62-9f9a-dc9bd7bfa566/build-log/recording_test.log) or the [debug log](https://console.cloud.google.com/storage/browser/ci-vcr-logs/beta/refs/heads/auto-pr-14569/artifacts/e790d2fb-bb83-4a62-9f9a-dc9bd7bfa566/recording) for each test

@BBBmau, I am not sure why this is failing. `TestAccBigQueryExternalDataTable_CSV_WithSchemaAndConnectionIDAndHivePartitioning` shouldn't be affected by this change.

mmv1/templates/terraform/examples/bigquery_dataset_basic.tf.tmpl
```diff
@@ -10,7 +10,7 @@ resource "google_bigquery_dataset" "{{$.PrimaryResourceId}}" {
   }
 
   access {
```
wj-chen, line 12: In the documentation we can advertise the user to use the new role syntax. In the test at least, we should keep the old case and add a new test for the new role so we have both to prevent regression. We also want to make sure the case in https://github.com/hashicorp/terraform-provider-google/issues/8370#issuecomment-1380206756 is tested and handled.

sachinpro: > In the documentation we can advertise the user to use the new role syntax. 

 It is not shown in the diff but this test validates both cases:
 ```
   access {
     role          = "roles/bigquery.dataOwner"
    user_by_email = google_service_account.bqowner.email
  }
   access {
    role   = "READER"
    domain = "hashicorp.com"
  }
 ```
 This test is reflected in the documentation showing both ways to specify role.
 Do you still think we need another test?


> We also want to make sure the case in https://github.com/hashicorp/terraform-provider-google/issues/8370#issuecomment-1380206756 is tested and handled.

Let me check this. I don't think it is an issue anymore.

sachinpro:
> General question: Is there any difference between implementing it this way vs through a diff suppression?

@wj-chen The DiffSuppressFunc on a field within a set element doesn't work because Terraform hashes the entire element to see if it has changed. Since "OWNER" and "roles/bigquery.dataOwner" are different strings, they produce different hashes, and Terraform sees one access block being removed and another being added, rather than an in-place change.
That's why I *have to* use a custom hash function.

I can make the change if there is a better way to handle this.

mmv1/templates/terraform/examples/bigquery_dataset_basic.tf.tmpl
wj-chen: If the test covers both legacy and new syntax then it should be sufficient.

wj-chen:
> > 🔴 Tests failed during RECORDING mode: `TestAccBigQueryExternalDataTable_CSV_WithSchemaAndConnectionIDAndHivePartitioning` [[Error message](https://storage.cloud.google.com/ci-vcr-logs/beta/refs/heads/auto-pr-14569/artifacts/e790d2fb-bb83-4a62-9f9a-dc9bd7bfa566/build-log/recording_build/TestAccBigQueryExternalDataTable_CSV_WithSchemaAndConnectionIDAndHivePartitioning_recording_test.log)] [[Debug log](https://storage.cloud.google.com/ci-vcr-logs/beta/refs/heads/auto-pr-14569/artifacts/e790d2fb-bb83-4a62-9f9a-dc9bd7bfa566/recording/TestAccBigQueryExternalDataTable_CSV_WithSchemaAndConnectionIDAndHivePartitioning.log)]
> > 🔴 Errors occurred during RECORDING mode. Please fix them to complete your PR.
> > View the [build log](https://storage.cloud.google.com/ci-vcr-logs/beta/refs/heads/auto-pr-14569/artifacts/e790d2fb-bb83-4a62-9f9a-dc9bd7bfa566/build-log/recording_test.log) or the [debug log](https://console.cloud.google.com/storage/browser/ci-vcr-logs/beta/refs/heads/auto-pr-14569/artifacts/e790d2fb-bb83-4a62-9f9a-dc9bd7bfa566/recording) for each test
> 
> @BBBmau, I am not sure why this is failing. `TestAccBigQueryExternalDataTable_CSV_WithSchemaAndConnectionIDAndHivePartitioning` shouldn't be affected by this change.

Could you check if it's related to the test added in https://github.com/GoogleCloudPlatform/magic-modules/pull/14483?

Error message:
```
POST /bigquery/v2/projects/ci-test-project-188019/queries?alt=json&prettyPrint=false HTTP/1.1
Host: bigquery.googleapis.com
User-Agent: google-api-go-client/0.5 Terraform/1.11.0 (+https://www.terraform.io) Terraform-Plugin-SDK/2.36.0 terraform-provider-google-beta/acc
Content-Length: 142
X-Goog-Api-Client: gl-go/1.23.11 gdcl/0.238.0
Accept-Encoding: gzip

{
 "query": "ALTER TABLE `ci-test-project-188019.tf_test_1byqthgjxe.tf_test_6w7ugd6t64` DROP COLUMN `country_partitioned`",
 "useLegacySql": false
}

2025/07/19 00:07:21 [DEBUG] Google API Response Details:
HTTP/2.0 404 Not Found
Date: Sat, 19 Jul 2025 00:07:21 GMT
Server: ESF
Vary: Origin
Vary: X-Origin
Vary: Referer
X-Frame-Options: SAMEORIGIN
X-Xss-Protection: 0

{
  "error": {
    "code": 404,
    "message": "Not found: Connection tf_test_nyfa7l3x1h was not found for table ci-test-project-188019:tf_test_1byqthgjxe.tf_test_6w7ugd6t64",
    "errors": [
      {
        "message": "Not found: Connection tf_test_nyfa7l3x1h was not found for table ci-test-project-188019:tf_test_1byqthgjxe.tf_test_6w7ugd6t64",
        "domain": "global",
        "reason": "notFound"
      }
    ],
    "status": "NOT_FOUND"
  }
}
```

So looks like our table column drop logic still tried to run on the partition column, and it happened after the connection resource was deleted to cause the failure. Let's prioritize fixing this.

mmv1/templates/terraform/examples/bigquery_dataset_basic.tf.tmpl
sachinpro: > We also want to make sure the case in https://github.com/hashicorp/terraform-provider-google/issues/8370#issuecomment-1380206756 is tested and handled.

I couldn't reproduce this by adding access blocks with `group_by_email` and `special_group`:
```
	access {
    role   = "READER"
    group_by_email = "magic-modules@google.com"
  }

	access {
    role   = "READER"
    special_group = "projectWriters"
  }
```

I did find one issue with case sensitivity though, which is there on main and this branch. Updating this [line](https://github.com/hashicorp/terraform-provider-google/blob/94c82becbf3fee673f0f62f708380b3a59f16f73/google/services/bigquery/resource_bigquery_dataset_generated_test.go#L79) to `domain = "hashicorp.coM"`(for example) will give this error:

```
  |   # google_bigquery_dataset.dataset will be updated in-place
  |   ~ resource "google_bigquery_dataset" "dataset" {
  |         id                              = "<redacted>"
  |         # (18 unchanged attributes hidden)
  | 
  |       - access {
  |           - role          = "OWNER" -> null
  |           - user_by_email = "<redacted>" -> null
  |         }
  |       - access {
  |           - domain = "hashicorp.com" -> null
  |           - role   = "READER" -> null
  |         }
  |       + access {
  |           + domain = "hashicorp.coM"
  |           + role   = "READER"
  |         }
  |       + access {
  |           + role          = "OWNER"
  |           + user_by_email = "<redacted>"
  |         }
  |     }
```

Case sensitivity doesn't seem to cause error in that comment though.

wj-chen: Thanks for testing. The case permadiff sounds related to https://github.com/hashicorp/terraform-provider-google/issues/21928. I added this observation in the tracking issue and we can track and fix that separately.

wj-chen:
> > General question: Is there any difference between implementing it this way vs through a diff suppression?
> 
> @wj-chen The DiffSuppressFunc on a field within a set element doesn't work because Terraform hashes the entire element to see if it has changed. Since "OWNER" and "roles/bigquery.dataOwner" are different strings, they produce different hashes, and Terraform sees one access block being removed and another being added, rather than an in-place change. That's why I _have to_ use a custom hash function.
> 
> I can make the change if there is a better way to handle this.

This approach LGTM.

sachinpro:
@BBBmau This PR is ready for review. The test failure `TestAccBigQueryExternalDataTable_CSV_WithSchemaAndConnectionIDAndHivePartitioning`(not caused by this change) is fixed in a separate PR(https://github.com/GoogleCloudPlatform/magic-modules/pull/14659).

melinath:

mmv1/templates/terraform/constants/bigquery_dataset.go.tmpl
```diff
@@ -23,3 +23,28 @@ func validateDefaultTableExpirationMs(v interface{}, k string) (ws []string, err
...
+	// Normalize the role if it exists and matches a legacy role.
+	if role, ok := copy["role"].(string); ok {
+		if newRole, ok := bigqueryAccessPrimitiveToRoleMap[role]; ok {
+			copy["role"] = newRole
+		}
+	}
+
+	// Use the default HashResource function on the (potentially modified) copy.
+	return schema.HashResource(bigqueryDatasetAccessSchema())(copy)
+}
```
melinath: In the provider downstreams, the schema function will be implicitly available because it's part of the resource file (which is in the same package). However, in terraform-google-conversion, the schema isn't included in the downstream. While you could try to set up that import - I'm pretty sure you only need this hash func in the providers. You can exclude it from TGC with:

```suggestion
{{- if ne $.Compiler "terraformgoogleconversion-codegen" }}
// bigqueryDatasetAccessHash is a custom hash function for the access block.
// It normalizes the 'role' field before hashing, treating legacy roles
// and their modern IAM equivalents as the same.
func resourceBigqueryDatasetAccessHash(v interface{}) int {
	m, ok := v.(map[string]interface{})
	if !ok {
		return 0
	}
	// Make a copy of the map to avoid modifying the underlying data.
	copy := make(map[string]interface{}, len(m))
	for k, val := range m {
		copy[k] = val
	}

	// Normalize the role if it exists and matches a legacy role.
	if role, ok := copy["role"].(string); ok {
		if newRole, ok := bigqueryAccessPrimitiveToRoleMap[role]; ok {
			copy["role"] = newRole
		}
	}

	// Use the default HashResource function on the (potentially modified) copy.
	return schema.HashResource(bigqueryDatasetAccessSchema())(copy)
}
{{- end }}
```

melinath (changes requested):

mmv1/templates/terraform/constants/bigquery_dataset.go.tmpl
sachinpro: Thanks! I've made this change.

melinath (approved):
LGTM for Terraform.

melinath:
Moving forward with merge since it looks like there aren't meaningful changes since @wj-chen's last review.

As a side note, I tested this manually locally & confirmed that there's no diff on upgrade. Like with any diff suppress, users will see a more complex diff if anything in the access block _does_ change, but that's expected.

PR #14572, opened 2025-07-18
Initial bazel setup for magic-modules binary build

melinath:
It seems like it might be necessary (or maybe just easier) to make magic-modules have a toplevel binary that does everything we need (instead of handling it in a makefile) but I'm going to do that separately (which will probably break this) - I want to make sure the non-bazel path still works there first. I could just wait to merge this until I can rebase it & fix any issues.

```release-note:none

```


melinath:
Closing this one for now - should be simple to recreate properly as needed.

PR #14574, opened 2025-07-18
Add a message for the change in default value for view.use_legacy_sql in google_bigquery_table

wj-chen:
The corresponding breaking change: 

**Release Note Template for Downstream PRs (will be copied)**

See [Write release notes](https://googlecloudplatform.github.io/magic-modules/contribute/release-notes/) for guidance.

```release-note:none

```

PR #14575, opened 2025-07-18
Remove the default value of view.use_legacy_sql in google_bigquery_table

wj-chen:
Context of the breaking change: https://github.com/GoogleCloudPlatform/magic-modules/pull/12390#discussion_r1941991099
Documentation update in `main`: https://github.com/GoogleCloudPlatform/magic-modules/pull/14574

```release-note:breaking-change
bigquery: removed the default value of `view.use_legacy_sql` in `google_bigquery_table`
```

### DIFF
--- a/mmv1/third_party/terraform/website/docs/guides/version_7_upgrade.html.markdown
+++ b/mmv1/third_party/terraform/website/docs/guides/version_7_upgrade.html.markdown
@@ -102,6 +102,12 @@ Description of the change and how users should adjust their configuration (if ne
 
 ## Resources
 
+## Resource: `google_bigquery_table`
+
+### `view.use_legacy_sql` no longer has a default value of `True`
+
+If `view.use_legacy_sql` is not specified in the configuration, no value is sent to the API.
+
 ## Resource: `google_bigtable_table_iam_binding`
 
 ### `instance` is now removed
--- a/mmv1/third_party/terraform/services/bigquery/resource_bigquery_table.go.tmpl
+++ b/mmv1/third_party/terraform/services/bigquery/resource_bigquery_table.go.tmpl
@@ -1051,13 +1051,12 @@ func ResourceBigQueryTable() *schema.Resource {
 						},
 
 						// UseLegacySQL: [Optional] Specifies whether to use BigQuery's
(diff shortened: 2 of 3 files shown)


mmv1/third_party/terraform/website/docs/guides/version_7_upgrade.html.markdown
```diff
@@ -102,6 +102,12 @@ Description of the change and how users should adjust their configuration (if ne
 
 ## Resources
 
+## Resource: `google_bigquery_table`
+
+### `view.use_legacy_sql` no longer has a default value of `True`
+
+If `view.use_legacy_sql` is not specified in the configuration, no value is sent to the API.
```
rileykarson: Just to make sure I understand this- the API will only accept null (no value) or `true`, and will reject `false`. By removing the default here we're sending null when unset (or false, I think, w/ how we serialize empty values) and users can enable the setting with an explicit true.

A user who configured a resource without setting the field today had the default set to `true`; they'll get a diff from `true`-> _what appears to be_ `false` in Terraform, but is actually null, prompting an update to set the API value to null.

(You haven't done anything to make the null and false thing a specific issue, to be clear! The values are collapsed into each other by several layers by default- in Terraform schema, in json serialization, etc. I'm being very explicit about where values are false/null because that's how they're displayed in user prompts/logs at those points)

wj-chen: The API accepts three values: `null`, `true`, `false`, and treats `null` and `false` as separate values. I tested in TF before to confirm they are different since the input of not setting vs setting to `false` comes back from the API response as `null` vs `false`.

For a user without it in their config, after upgrading to this version, they will see `true` -> `(none)`.

This change is required for an upcoming feature where this field can't be set, so TF users of that feature need to leave it out in their config, and also there can't be a TF default value when they leave it unset.

Hope this helps. Let me know if you have suggestions on how to make the release note more clear.

wj-chen: Accidentally closed this while syncing...new PR is https://github.com/GoogleCloudPlatform/magic-modules/pull/14751.

PR #14578, opened 2025-07-20
Fixed incorrect immutable flag on linked_router_appliance_instances instances

r-teller:
```release-note:bug
networkconnectivity: fixed `instances[].ip_address` & `instances[].virtual_machine` fields in `linked_router_appliance_instances` block being incorrectly treated as immutable for `google_network_connectivity_spoke` resource
```

### DIFF
--- a/mmv1/products/networkconnectivity/Spoke.yaml
+++ b/mmv1/products/networkconnectivity/Spoke.yaml
@@ -226,7 +226,6 @@ properties:
         type: Array
         description: The list of router appliance instances
         required: true
-        immutable: true
         item_type:
           description: The list of router appliance instances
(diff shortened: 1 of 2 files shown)


r-teller:
This PR will allow terraform to add or remove instances to linked_router_appliance_instances without requiring deletion of the existing spoke. This behavior aligns with how the gcloud and console behave

roaks3 (approved):
LGTM

Is there a GitHub issue we should tie this to before we merge?

r-teller:
> LGTM
> 
> Is there a GitHub issue we should tie this to before we merge?

There is no issue, I noticed the issue during a deployment and it was quicker to fix

PR #14581, opened 2025-07-21
Adding support for refresh subscription resource.

arnavdham:
This Pull Request introduces a new resource refresh resource for Analytics Hub.  This resource is made for calling the Refresh API which is an important API for the subscription resource.

```release-note:new-resource
`google_bigquery_analytics_hub_refresh_data_exchange_subscription` (beta)
```


melinath (commented):
This seems like an imperative action; I'm not sure it makes sense to add to the Terraform provider.

melinath:
@modular-magician reassign-reviewer @c2thorn

Handing off since Cameron has additional context here

c2thorn:
Noting that the plan was for this to be picked back up by the service team

arnavdham:
Would've been picked by now if it was important :)

PR #14582, opened 2025-07-21
fix: (storage) remove client side bucket name validation as same validations being performed on API

gurusai-voleti:

```release-note:bug
storage: remove bucket name validations for `google_storage_bucket` as the same validations being performed on API
```


gurusai-voleti:
Unrelated test case failures

pavandagagoogle:
> Are the same validations being performed on the API side, or different ones?
> 
> If they're the same, we should prefer to keep the validation here as well so users get feedback earlier in the process as validation runs at `terraform plan` time rather than API-side validation happening at `terraform apply`.
> 
> If the validation changes or is not easily captured here then it makes sense to remove

The validations are quickly getting out of sync and replicating them here is like duplicating efforts. We are keeping minimal to no validations on the client side for GCS except for type checks, etc.. and delegating validations like these where bucket name prefixes, naming conventions, etc. to the backend to avoid issues like this issue.

PR #14583, opened 2025-07-21
compute: fix documentation naming conflict, added `network_name` to data source `google_compute_subnetworks` and deprecated `network_self_link`

ramonvermeulen:
Closes https://github.com/hashicorp/terraform-provider-google/issues/23663

Fixing it this way seems most logical to me, just in-case `network_self_link` is used even though not documented, so that there will be no breaking change in a `6.x.x` release.

```release-note:enhancement
compute: added `network_name` field to `google_compute_subnetworks` data source
```

```release-note:deprecation
compute: deprecated `network_self_link` field in `google_compute_subnetworks` data source. Use ``network_name` instead.
```

### DIFF
--- a/mmv1/third_party/terraform/website/docs/d/compute_subnetworks.html.markdown
+++ b/mmv1/third_party/terraform/website/docs/d/compute_subnetworks.html.markdown
@@ -42,7 +42,7 @@ The following arguments are supported:
 * `ip_cidr_range` - The IP address range represented as a CIDR block.
 * `name` - The name of the subnetwork.
 * `network` - The self link of the parent network.
-* `network_self_link` - The name of the parent network computed from `network` attribute. (will be deprecated in the future, use `network_name` instead)
+* `network_self_link` - (Deprecated) The name of the parent network computed from `network` attribute. (deprecated and will be removed in a future major release. Use `network_name` instead.)
 * `network_name` - The name of the parent network computed from `network` attribute.
 * `private_ip_google_access` - Whether the VMs in the subnet can access Google services without assigned external IP addresses.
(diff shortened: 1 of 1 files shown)


mmv1/third_party/terraform/services/compute/data_source_google_compute_subnetworks.go
```diff
@@ -99,8 +99,9 @@ func dataSourceGoogleComputeSubnetworksRead(d *schema.ResourceData, meta interfa
 			"description":              subnet.Description,
 			"ip_cidr_range":            subnet.IpCidrRange,
 			"name":                     subnet.Name,
-			"network_self_link":        filepath.Base(subnet.Network),
+			"network_self_link":        filepath.Base(subnet.Network), // TODO: remove in next major release (7.0.0) also from docs
 			"network":                  subnet.Network,
+			"network_name":             filepath.Base(subnet.Network),
```
shuyama1, line 110: You'll also want to add this field to the resource schema and may also want to deprecate the `network_self_link` field there following the handwritten section in https://googlecloudplatform.github.io/magic-modules/breaking-changes/make-a-breaking-change/#field-deprecation-due-to-removal-or-rename

ramonvermeulen: Just added the field to the resource schema, and use `Deprecated` to deprecate the old field in https://github.com/GoogleCloudPlatform/magic-modules/pull/14583/commits/01f125e3938c9f06f5236769424f77af24eaa878.

shuyama1 (approved):

PR #14584, opened 2025-07-21
Remove extra step

slevenick:
Removes extra step that is causing tests to fail due to deleted connection. I'm not sure why the second step was added here, it seems like it was copied from a test designed around removing the connection. I don't think it needs to exist

Tested locally

```release-note:none

```


melinath (changes requested):
This seems like something that should be possible to do, though? It looks like a bug in the test - for some reason `google_bigquery_table` isn't updated before the `google_bigquery_connection` is deleted. Possibly routing through the `locals` is breaking the ordering somehow?

PR #14586, opened 2025-07-21
feat: add OdbNetwork support in OracleDatabase AutonomousDatabase

nimish-khurana:
ref https://github.com/hashicorp/terraform-provider-google/issues/23651

```release-note:enhancement
oracledatabase: added `odb_network` and `odb_subnet` fields, and make `network` and `cidr` fields optional in `google_oracle_database_autonomous_database` resource
```

### DIFF
--- a/mmv1/products/oracledatabase/AutonomousDatabase.yaml
+++ b/mmv1/products/oracledatabase/AutonomousDatabase.yaml
@@ -73,6 +73,23 @@ examples:
       deletion_protection: 'false'
       database_name: 'fmt.Sprintf("tftestdatabase%s", acctest.RandString(t, 10))'
       endpoint_name: 'fmt.Sprintf("tftestendpoint%s", acctest.RandString(t, 10))'
+  - name: 'oracledatabase_autonomous_database_odbnetwork'
+    primary_resource_id: 'myADB'
+    vars:
+      project: 'my-project'
+      autonomous_database_id: 'my-instance'
+      database_name: 'mydatabase'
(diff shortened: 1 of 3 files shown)


mmv1/products/oracledatabase/AutonomousDatabase.yaml
```diff
@@ -623,11 +640,25 @@ properties:
...
     description: "The name of the VPC network used by the Autonomous Database.\nFormat:
     projects/{project}/global/networks/{network} "
-    required: true
+    required: false
   - name: 'cidr'
     type: String
     description: 'The subnet CIDR range for the Autonmous Database. '
-    required: true
+    required: false
+  - name: odbNetwork
```
roaks3, line 661: Should these be using `exactly_one_of`, since required was removed? Otherwise it's possible to omit both options (I assume there is still a server-side error, but I think for cases like this we tend to prefer modeling the validation at the client level)

roaks3 (commented):
LGTM, but just wanted to check about `exactly_one_of`

mmv1/products/oracledatabase/AutonomousDatabase.yaml
nimish-khurana: It is actually possible to omit both, then no private connectivity to the database will be established.

nimish-khurana: Also added a test for this case.

nimish-khurana:
@roaks3 Gentle ping to review.

PR #14589, opened 2025-07-21
Remove token from github repo string if not necessary

iyabchen:
https://github.com/hashicorp/terraform-provider-google/issues/17307

- Clone does not require a token, but generate downstream diff(involves git commit) needs a token. 
- Tested in https://github.com/GoogleCloudPlatform/magic-modules/pull/14597 where a downstream can be generated. 

```release-note:none

```


shuyama1:
Trying to rotate a reviewer from the contribution team, but it doesn't seem terraform-contribution exists anymore. Use [terraform-validator-reviewers](https://github.com/orgs/GoogleCloudPlatform/teams/terraform-validator-reviewers) instead, as it seems to be the closest match

ScottSuarez:
I commented on this parent bug but i think we should always use a token when interacting with github. Even if we are using a public GET operation we might get hit with a 429.

ScottSuarez:
closing as of https://github.com/hashicorp/terraform-provider-google/issues/17307#issuecomment-3114977052

@melinath 

Thanks @iyabchen for taking a look, sorry for any throw away here!

PR #14593, opened 2025-07-22
Update docs for `google_compute_forwarding_rule`

FilipKubawskiOkta:
Hi this doesn't work for this resource. It only works with `google_compute_global_forwarding_rule`.

You don't have any tests regarding the `all-apis` keyword and i wasn't able to find a working configuration. The Google API translates it to a selflink by default and throws an error
image

```release-note:fix
update docs regarding API bundles for `google_compute_forwarding_rule`
```


BBBmau (approved):
thanks for catching this!

PR #14598, opened 2025-07-22
Lakshman single datasource generation

luckyswaminathan:
https://github.com/hashicorp/terraform-provider-google/issues/12595 this issue discusses the problems that can be faced due to handwritten singular datasources, this PR gives the ability to autogenerate singular datasources(read only)

```release-note:none
allow autogeneration of singular datasources
```


mmv1/api/resource.go
```diff
@@ -226,6 +226,9 @@ type Resource struct {
 	// If true, resource is not importable
 	ExcludeImport bool `yaml:"exclude_import,omitempty"`
 
+	// If true, resource should be autogenerated as a data source
+	DataSourceGenerationFlag bool `yaml:"datasource_generation_flag,omitempty"`
```
melinath: This would be a slightly better name - active voice ("generate" instead of "generation") to match other fields like "exclude_import" etc., and omits "flag" (which doesn't add anything semantically)

```suggestion
	GenerateDatasource bool `yaml:"generate_datasource,omitempty"`
```

mmv1/provider/terraform.go
```diff
@@ -188,6 +189,20 @@ func (t *Terraform) GenerateResourceSweeper(object api.Resource, templateData Te
...
+	if !object.ShouldGenerateSingularDataSource() {
+		return
+	}
+
+	productName := t.Product.ApiName
+	targetFolder := path.Join(outputFolder, t.FolderName(), "services", productName)
+	if err := os.MkdirAll(targetFolder, os.ModePerm); err != nil {
+		log.Println(fmt.Errorf("error creating parent directory %v: %v", targetFolder, err))
+	}
+	targetFilePath := path.Join(targetFolder, fmt.Sprintf("data_source_%s_generated.go", t.ResourceGoFilename(object)))
```
melinath: I'd recommend leaving off the "_generated" suffix - we don't use one for resources either. (We do use it for tests, but that's to distinguish between handwritten and generated tests. But there won't ever be multiple versions of the actual datasource definition.)

This will also make it easier to compare downstream diffs when a datasource is converted from handwritten to generated.

```suggestion
	targetFilePath := path.Join(targetFolder, fmt.Sprintf("data_source_%s.go", t.ResourceGoFilename(object)))
```

mmv1/provider/template_data.go
```diff
@@ -93,6 +93,14 @@ func (td *TemplateData) GenerateMetadataFile(filePath string, resource api.Resou
 	td.GenerateFile(filePath, templatePath, resource, false, templates...)
 }
 
+func (td *TemplateData) GenerateDataSourceFile(filePath string, resource api.Resource) {
+	templatePath := "templates/terraform/singular_datasource.go.tmpl"
```
melinath: nit: singular is the "default" / common case for datasources, so maybe just call it `datasource`. But I don't feel particularly strongly about this.

```suggestion
	templatePath := "templates/terraform/datasource.go.tmpl"
```

melinath (changes requested):
Same question for datasource tests!

mmv1/templates/terraform/singular_datasource.go.tmpl
```diff
@@ -0,0 +1,117 @@
...
+{{ if $.FlattenedProperties }}
+    "google.golang.org/api/googleapi"
+{{- end}}
+)
+
+func DataSource{{ .ResourceName -}}() *schema.Resource {
+	rs := Resource{{ .ResourceName -}}().Schema
+
+	dsSchema := tpgresource.DatasourceSchemaFromResourceSchema(rs)
+	uriParts := strings.Split("{{.SelfLinkUri}}", "/")
```
roaks3: Since you have the `SelfLinkUri` at the time this template is rendered, it might be slightly better to do the required/optional calculation at that time and just emit the `tpgresource.AddRequiredFieldsToSchema()`/`tpgresource.AddOptionalFieldsToSchema()`. Mainly I think it might just make migrations a little easier because the new and old version will look more similar.

That might look like functions you create in `resource.go` like `DatasourceRequiredFields()`/`DatasourceOptionalFields()` to return the list of strings, and then in this template you just use them directly.

luckyswaminathan:
> Are you planning to handle docs generation in this PR or a separate PR?
(ignore the location of the comment the gh ui was bugging a bit)
for both of these they will be done in followups! @melinath

mmv1/templates/terraform/singular_datasource.go.tmpl
luckyswaminathan: I have tried to do so, but I have come to the conclusion that this is genuinely not possible from a template file, as calling methods defined in resource.go on a resource doesn't compile(happens for all helpers defined in resource.go)

luckyswaminathan: unless I am missing something which could very well be the case

melinath: could you give an example of how you're calling them?

luckyswaminathan: oh, sorry didn't see this - from what I have seen, none of the methods in resource.go actually are used in resource.go.tmpl (and also go doesn't compile if I try to do something like $.DatasourceRequiredFields(). Am I calling this wrong? seems like that would be a logical way of doing it if it worked

melinath: Go templates call functions implicitly - you have to leave off the `()` or it will fail to compile. https://github.com/hashicorp/terraform-provider-google/issues/23891. https://pkg.go.dev/text/template#hdr-Arguments has some more details about how to notate function calls.

melinath: So for example, https://github.com/GoogleCloudPlatform/magic-modules/blob/884aa539cdbf2c55efcf1def253aea8f1f13c662/mmv1/templates/terraform/resource.go.tmpl#L104

That's calling https://github.com/GoogleCloudPlatform/magic-modules/blob/884aa539cdbf2c55efcf1def253aea8f1f13c662/mmv1/api/resource.go#L651

luckyswaminathan: sounds good! just pushed changes

melinath (commented):
It looks like this is basically all correct; however, the generated diff shows one unusual change: it's moving the `location` field from optional to required, and (relatedly) removing the setting of location with `tpgresource.GetLocation` / removing the setting of `location` in state. That's most likely a breaking change, unless there is no valid configuration with the current provider that omits `location`.

(The current logic also seems a little surprising - like it would create a permadiff if location is unset.)

I'm not sure whether handling `location` this way on datasources is universal, standard, or non-standard. We should probably make the default be whatever happens most frequently, but may need to allow overriding to avoid breaking changes. Could you investigate and see?

mmv1/templates/terraform/datasource.go.tmpl
```diff
@@ -0,0 +1,96 @@
...
+    "{{ $.ImportPath }}/tpgresource"
+    transport_tpg "{{ $.ImportPath }}/transport"
+    "{{ $.ImportPath }}/verify"
+
+{{ if $.FlattenedProperties }}
+    "google.golang.org/api/googleapi"
+{{- end}}
+)
+
+func DataSourceGoogle{{ .ResourceName -}}() *schema.Resource {
```
melinath: no need to change the naming for backwards-compatibility - DataSource{{.ResourceName}} is better / matches generated resource names better.

```suggestion
func DataSource{{ .ResourceName -}}() *schema.Resource {
```

melinath:
It looks like this is adding a `tpgresource.SetDataSourceLabels` section to the datasource - what will the impact of that be?

luckyswaminathan:
> It looks like this is adding a `tpgresource.SetDataSourceLabels` section to the datasource - what will the impact of that be?

id say roughly 40 datasources include it (all the method does is set metadata via effective_labels, terraform_labels, etc)  - is referenced: https://www.hashicorp.com/en/blog/terraform-google-provider-adds-updates-to-default-labels 

I'm a bit unsure on this method's failure rate, but from what I have discerned it seems relatively fine(and is kinda randomly inserted into a lot of the newer datasources, but also could be misinterpreting the gitlens activity)

melinath (changes requested):
`SetDataSourceLabels` is intented to be called if the datasource has `labels` / `effective_labels` /  `terraform_labels` fields. It looks like it fails silently if `effective_labels` isn't present, but I think we probably just shouldn't include it unless those fields are on the resource.

I assume the "random" distribution is because not all resources have labels fields in their schemas.

For now, it might be easiest to just remove this from the template, and then add the logic for including it properly when adding a generated datasource that needs it.

luckyswaminathan:
> `SetDataSourceLabels` is intented to be called if the datasource has `labels` / `effective_labels` / `terraform_labels` fields. It looks like it fails silently if `effective_labels` isn't present, but I think we probably just shouldn't include it unless those fields are on the resource.
> 
> I assume the "random" distribution is because not all resources have labels fields in their schemas.
> 
> For now, it might be easiest to just remove this from the template, and then add the logic for including it properly when adding a generated datasource that needs it.

sounds good, I think ill do a follow up set of prs to add these parts as I continue to migrate(can add different blocks within template files to handle some of these variants so datasource generation can be relatively expressive)

mmv1/templates/terraform/datasource.go.tmpl
```diff
@@ -0,0 +1,92 @@
...
+func DataSource{{ .ResourceName -}}() *schema.Resource {
+	rs := Resource{{ .ResourceName -}}().Schema
+
+	dsSchema := tpgresource.DatasourceSchemaFromResourceSchema(rs)
+
+  // Set 'Required' schema elements
+  tpgresource.AddRequiredFieldsToSchema(dsSchema, {{range $index, $field := .DatasourceRequiredFields}}{{if gt $index 0}}, {{end}}{{printf "%q" $field}}{{end}})
+
+  // Set 'Optional' schema elements
+  tpgresource.AddOptionalFieldsToSchema(dsSchema, {{range $index, $field := .DatasourceOptionalFields}}{{if gt $index 0}}, {{end}}{{printf "%q" $field}}{{end}})
```
melinath, line 65: technically this is only necessary if there actually are optional fields; I'm not blocking this PR on it (since it's ultimately a no-op) but it would be nice to fix in a follow-up.

PR #14600, opened 2025-07-22
Update boot disk config for Hyperdisk provisioned iops and throughput

nathanielford:

```release-note:REPLACEME

```

### DIFF
--- a/mmv1/third_party/terraform/services/container/node_config.go.tmpl
+++ b/mmv1/third_party/terraform/services/container/node_config.go.tmpl
@@ -146,6 +146,8 @@ func schemaNodeConfig() *schema.Schema {
 					Description:  `Type of the disk attached to each node. Such as pd-standard, pd-balanced or pd-ssd`,
 				},
 
+				"boot_disk": schemaBootDiskConfig(),
+
 				"guest_accelerator": {
 					Type:     schema.TypeList,
 					Optional: true,
@@ -899,6 +901,46 @@ func schemaNodeConfig() *schema.Schema {
 	}
 }
 
+
+func schemaBootDiskConfig() *schema.Schema {
+	return &schema.Schema{
+		Type: schema.TypeList,
+		Optional: true,
+		Computed: true,
+		MaxItems: 1,
+		Description: `Boot disk configuration for node pools nodes.`,
+		Elem: &schema.Resource{
+			Schema: map[string]*schema.Schema{
+				"disk_type": {
+					Type:         schema.TypeString,
+					Optional:     true,
+					Computed:     true,
+					Description:  `Type of the disk attached to each node. Such as pd-standard, pd-balanced or pd-ssd`,
+				},
+				"size_gb": {
+					Type:         schema.TypeInt,
+					Optional:     true,
+					Computed:     true,
+					ValidateFunc: validation.IntAtLeast(10),
+					Description:  `Size of the disk attached to each node, specified in GB. The smallest allowed disk size is 10GB.`,
+				},
+				"provisioned_iops": {
+					Type:         schema.TypeInt,
+					Optional:     true,
+					Computed:     true,
+					Description:  `Configured IOPs provisioning. Only valid with disk type hyperdisk-balanced.`,
+				},
+				"provisioned_throughput": {
+					Type:         schema.TypeInt,
+					Optional:     true,
+					Computed:     true,
+					Description:  `Configured throughput provisioning. Only valid with disk type hyperdisk-balanced.`,
+				},
+			},
+		},
+	}
+}
+
 // Separate since this currently only supports a single value -- a subset of
 // the overall NodeKubeletConfig
 func schemaNodePoolAutoConfigNodeKubeletConfig() *schema.Schema {
@@ -1032,6 +1074,10 @@ func expandNodeConfig(v interface{}) *container.NodeConfig {
 
 	if v, ok := nodeConfig["disk_type"]; ok {
 		nc.DiskType = v.(string)
+	}
+
+	if v, ok := nodeConfig["boot_disk"]; ok {
+		nc.BootDisk = expandBootDiskConfig(v)
 	}
 
 	if v, ok := nodeConfig["local_ssd_count"]; ok {
@@ -1305,6 +1351,36 @@ func expandNodeConfig(v interface{}) *container.NodeConfig {
 	}
 
 	return nc
+}
+
+func expandBootDiskConfig(v interface{}) *container.BootDisk {
+	bd := &container.BootDisk{}
+	if v == nil {
+		return nil
+	}
+	ls := v.([]interface{})
+	if len(ls) == 0 {
+		return nil
+	}
+	cfg := ls[0].(map[string]interface{})
+
+	if v, ok := cfg["disk_type"]; ok {
+		bd.DiskType = v.(string)
+	}
+
+	if v, ok := cfg["size_gb"]; ok {
+		bd.SizeGb = int64(v.(int))
+	}
+
+	if v, ok := cfg["provisioned_iops"]; ok {
+		bd.ProvisionedIops = int64(v.(int))
+	}
+
+	if v, ok := cfg["provisioned_throughput"]; ok {
+		bd.ProvisionedThroughput = int64(v.(int))
+	}
+
+	return bd
 }
 
 func expandResourceManagerTags(v interface{}) *container.ResourceManagerTags {
@@ -1688,11 +1764,12 @@ func flattenNodeConfig(c *container.NodeConfig, v interface{}) []map[string]inte
 		"containerd_config":        flattenContainerdConfig(c.ContainerdConfig),
 		"disk_size_gb":             c.DiskSizeGb,
 		"disk_type":                c.DiskType,
+		"boot_disk":                flattenBootDiskConfig(c.BootDisk),
 		"guest_accelerator":        flattenContainerGuestAccelerators(c.Accelerators),
 		"local_ssd_count":          c.LocalSsdCount,
 		"logging_variant":          flattenLoggingVariant(c.LoggingConfig),
 {{- if ne $.TargetVersionName "ga" }}
-		"ephemeral_storage_config":  flattenEphemeralStorageConfig(c.EphemeralStorageConfig),
+		"ephemeral_storage_config": flattenEphemeralStorageConfig(c.EphemeralStorageConfig),
 {{- end }}
 		"local_nvme_ssd_block_config": flattenLocalNvmeSsdBlockConfig(c.LocalNvmeSsdBlockConfig),
 		"ephemeral_storage_local_ssd_config": flattenEphemeralStorageLocalSsdConfig(c.EphemeralStorageLocalSsdConfig),
@@ -1738,6 +1815,22 @@ func flattenNodeConfig(c *container.NodeConfig, v interface{}) []map[string]inte
 		config[0]["oauth_scopes"] = schema.NewSet(tpgresource.StringScopeHashcode, tpgresource.ConvertStringArrToInterface(c.OauthScopes))
 	}
 
+	return config
+}
+
+func flattenBootDiskConfig(c *container.BootDisk) []map[string]interface{} {
+	config := []map[string]interface{}{}
+
+	if c == nil {
+		return config
+	}
+
+	config = append(config, map[string]interface{}{
+		"disk_type": c.DiskType,
+		"size_gb": c.SizeGb,
+		"provisioned_iops": c.ProvisionedIops,
+		"provisioned_throughput": c.ProvisionedThroughput,
+	})
 	return config
 }
 
--- a/mmv1/third_party/terraform/services/container/resource_container_node_pool_test.go.tmpl
+++ b/mmv1/third_party/terraform/services/container/resource_container_node_pool_test.go.tmpl
@@ -317,6 +317,119 @@ func TestAccContainerNodePool_withNodeConfig(t *testing.T) {
 	})
 }
 
+func TestAccContainerNodePool_withNodeConfigWithBootDiskConfig(t *testing.T) {
+	t.Parallel()
+
+	cluster := fmt.Sprintf("tf-test-cluster-%s", acctest.RandString(t, 10))
+	nodePool := fmt.Sprintf("tf-test-nodepool-%s", acctest.RandString(t, 10))
+	networkName := acctest.BootstrapSharedTestNetwork(t, "gke-cluster")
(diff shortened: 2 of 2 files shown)


mmv1/third_party/terraform/services/container/resource_container_node_pool_test.go.tmpl
```diff
@@ -317,6 +317,119 @@ func TestAccContainerNodePool_withNodeConfig(t *testing.T) {
...
+resource "google_container_node_pool" "np" {
+  provider           = google.user-project-override
+  name               = "%s"
+  location           = "us-central1-a"
+  cluster            = google_container_cluster.cluster.name
+  initial_node_count = 3
+
+  node_config {
+	machine_type    = "c3-standard-4"
+	boot_disk {
```
slevenick: Spacing is off here

mmv1/third_party/terraform/services/container/resource_container_node_pool_test.go.tmpl
```diff
@@ -317,6 +317,119 @@ func TestAccContainerNodePool_withNodeConfig(t *testing.T) {
...
+      size_gb                = 100
+      disk_type              = "hyperdisk-balanced"
+	  provisioned_iops       = 3456
+	  provisioned_throughput = 234
+	}
```
slevenick: Spacing

mmv1/third_party/terraform/services/container/resource_container_node_pool_test.go.tmpl
```diff
@@ -317,6 +317,119 @@ func TestAccContainerNodePool_withNodeConfig(t *testing.T) {
...
+resource "google_container_node_pool" "np" {
+  provider           = google.user-project-override
+  name               = "%s"
+  location           = "us-central1-a"
+  cluster            = google_container_cluster.cluster.name
+  initial_node_count = 3
+
+  node_config {
+	machine_type    = "c3-standard-4"
+	boot_disk {
```
slevenick: Spacing

mmv1/third_party/terraform/services/container/node_config.go.tmpl
```diff
@@ -899,6 +901,46 @@ func schemaNodeConfig() *schema.Schema {
...
+func schemaBootDiskConfig() *schema.Schema {
+	return &schema.Schema{
+		Type: schema.TypeList,
+		Optional: true,
+		Computed: true,
+		MaxItems: 1,
+		Description: `Boot disk configuration for node pools nodes.`,
+		Elem: &schema.Resource{
+			Schema: map[string]*schema.Schema{
+				"disk_type": {
```
slevenick, line 914: Should this be required? If someone specifies a boot_disk it seems like they should be required to set the type and size

mmv1/third_party/terraform/services/container/resource_container_node_pool_test.go.tmpl
nathanielford: Done!

mmv1/third_party/terraform/services/container/resource_container_node_pool_test.go.tmpl
nathanielford: Done!

mmv1/third_party/terraform/services/container/node_config.go.tmpl
nathanielford: The primary reason is that changing this would be backwards breaking. (People using the legacy field would be forced to update their configs.) That said, there are defaults that the API puts into place, making this non-required.

slevenick: I'm not sure how this would be breaking

Adding a required field as a sub-field of a new field isn't breaking because it only applies if the top-level field is specified. So disk_type would only be required if the user specifies boot_disk, which they can't currently.

nathanielford: The GKE API is migrating the top level `disk_type` and `disk_size_gb` to sub-fields of `boot_disk`, `boot_disk.disk_type` and `boot_disk.size_gb` respectively. The API cross-populates those fields. That is, if the legacy (top level) field is specified, then it fills in the sub-field of `boot_disk`. So the `boot_disk` field will become immediately populated. The goal is to keep these in sync, and to not allow the user to specify both unless the specification matches.

Now, I might be misunderstanding 'Required' - if the `boot_disk.disk_type` field gets populated by the API does it satisfy the 'Required' flag? Or is it required that the user-applied plan (the tf file the user is writing) has the Required field if `boot_disk` is sent up at all? If the former, then maybe it's fine to set it to required. 

The intended plan was to support any of the following configurations:

```
// Current status quo
node_config {
  disk_type = "hyperdisk-balanced"
  disk_size_gb = 100
  ...
}

// Support iops/throughput but change nothing else
node_config {
  disk_type = "hyperdisk-balanced"
  disk_size_gb = 100
  boot_disk = {
    provisioned_iops = 2345
    provisioned_throughput = 345 
  }
  ...
}

// Move to the new configuration field locations
node_config {
  boot_disk = {
    disk_type = "hyperdisk-balanced"
    disk_size_gb = 100
    provisioned_iops = 2345
    provisioned_throughput = 345 
  }
  ...
}

// Both, expected to be required during migration interim (note legacy and new fields must match)
node_config {
  disk_type = "hyperdisk-balanced"
  disk_size_gb = 100
  boot_disk = {
    disk_type = "hyperdisk-balanced"
    disk_size_gb = 100
    provisioned_iops = 2345
    provisioned_throughput = 345 
  }
  ...
}

```

If there is a better way to structure the Schema to handle this, am definitely open to changing it!

> disk_type would only be required if the user specifies boot_disk, which they can't currently.

I'm not sure I quite understand what you mean by "can't currently [specify boot_disk]"? It is live in the API, and just this week supported by gcloud, so they can specify it - though it won't be supported in the terraform provider until this PR is in. 

(By the way, sorry for not setting this as a draft - meant to just create this PR so a coworker see the changes and could help fix some testing problems and did not mean to submit it for a full review. If it's helpful, I can cancel it and re-PR it when it's ready. There are several doc things that need sorting out still.)

slevenick: Oh, so you plan to let people control the setting through either field? That's a little scary, but may work out since disk_type is already Optional+Computed. Makes sense to not require it inside boot_disk then.

> (By the way, sorry for not setting this as a draft - meant to just create this PR so a coworker see the changes and could help fix some testing problems and did not mean to submit it for a full review. If it's helpful, I can cancel it and re-PR it when it's ready. There are several doc things that need sorting out still.)

Yeah if there are outstanding docs/decisions that need to be made then marking as draft would be good

mmv1/third_party/terraform/services/container/resource_container_node_pool_test.go.tmpl
```diff
@@ -317,6 +317,264 @@ func TestAccContainerNodePool_withNodeConfig(t *testing.T) {
...
+		CheckDestroy:             testAccCheckContainerNodePoolDestroyProducer(t),
+		Steps: []resource.TestStep{
+			{
+				Config: testAccContainerNodePool_withBootDisk(cluster, nodePool, networkName, subnetworkName),
+			},
+			{
+				ResourceName:      "google_container_node_pool.np",
+				ImportState:       true,
+				ImportStateVerify: true,
+				// autoscaling.# = 0 is equivalent to no autoscaling at all,
```
tonyzhc: Remove the ImportStateVerifyIgnore and comments related to autoscaling.

mmv1/third_party/terraform/services/container/resource_container_node_pool_test.go.tmpl
```diff
@@ -317,6 +317,264 @@ func TestAccContainerNodePool_withNodeConfig(t *testing.T) {
...
+  cluster            = google_container_cluster.cluster.name
+  initial_node_count = 3
+
+  node_config {
+    disk_type    = "pd-balanced"
+    disk_size_gb = 202
+  }
+}
+`, cluster, networkName, subnetworkName, np)
+}
```
tonyzhc, line 684: What's the expected behavior if the terraform config is
```
boot_disk {
      size_gb                = 0
      disk_type              = ""
      provisioned_iops       = 0
      provisioned_throughput = 0
    }
```

nathanielford: The API would reject that - the disk_type isn't a valid string. (Unless the other disk_type is set, in which case it accepts the other value.) 

The others being 0 (assuming size matches what is specified elsewhere) is fine, and causes default values to be used on cluster.

mmv1/third_party/terraform/services/container/resource_container_node_pool_test.go.tmpl
nathanielford: Done!

mmv1/third_party/terraform/services/container/resource_container_node_pool_test.go.tmpl
```diff
@@ -317,6 +317,258 @@ func TestAccContainerNodePool_withNodeConfig(t *testing.T) {
...
+
+
+
```
tonyzhc, line 721: Can you add another test case that goes from 
```
node_config {
...
}
```
to 
```
node_config {
    boot_disk {
        ...
    }
}
```
just to test the permadiff behavior on the entire field?

slevenick (commented):
Is this ready for review?

nathanielford:
@slevenick Yep! Sorry, was trying to fix an (apparently) failing test on Friday and it looks like it passed. Docs got updated too I think. LMK if there is anything else you'd like me to work on!

mmv1/third_party/terraform/services/container/node_config.go.tmpl
```diff
@@ -899,6 +901,46 @@ func schemaNodeConfig() *schema.Schema {
 	}
 }
 
+
+func schemaBootDiskConfig() *schema.Schema {
```
slevenick, line 905: We still need handwritten docs to be added to the markdown page for this resource: https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/third_party/terraform/website/docs/r/container_cluster.html.markdown

mmv1/third_party/terraform/services/container/resource_container_node_pool_test.go.tmpl
```diff
@@ -317,6 +317,301 @@ func TestAccContainerNodePool_withNodeConfig(t *testing.T) {
...
+				ImportState:       true,
+				ImportStateVerify: true,
+				ImportStateVerifyIgnore: []string{"deletion_protection"},
+			},
+		},
+	})
+}
+
+
+func testAccContainerNodePool_withClusterBootDisk(cluster, networkName, subnetworkName string) string {
```
slevenick, line 355: Can you add some update tests for this field within the cluster resource also? This seems to only be a create test

mmv1/third_party/terraform/services/container/resource_container_node_pool_test.go.tmpl
slevenick: Yes, please add this type of tests as well

mmv1/third_party/terraform/services/container/resource_container_node_pool_test.go.tmpl
nathanielford: Done!

mmv1/third_party/terraform/services/container/resource_container_node_pool_test.go.tmpl
nathanielford: Done!

nathanielford:
I'm not sure why the VCR-test is reporting failure, given that the details seem to suggest everything went off fine. If you have guidance there lmk.

nathanielford:
Apologies - the debug logs that finally came through don't match what I saw locally. Will figure this out and update this once that is done. No need to re-review until the tests are passing.

nathanielford:
All the tests are passing... you can review whenever you're ready, @slevenick. Thanks again!

mmv1/third_party/terraform/services/container/node_config.go.tmpl
```diff
@@ -2350,6 +2446,34 @@ func nodePoolNodeConfigUpdate(d *schema.ResourceData, config *transport_tpg.Conf
...
+				bootDiskTypeOld, bootDiskTypeNew := d.GetChange("node_config.0.boot_disk.0.disk_type")
+				if bootDiskTypeOld == bootDiskTypeNew {
+					req.BootDisk.DiskType = ""
+				}
+				// If the legacy (top level) disk_size_gb field is not changing, nil it out to allow the API to fill it in.
+				legacyDiskSizeGbOld, legacyDiskSizeGbNew := d.GetChange("node_config.0.disk_size_gb")
+				if legacyDiskSizeGbOld == legacyDiskSizeGbNew {
+					req.DiskSizeGb = 0
+				}
+				// if the new boot disk configuration size_gb field is not changing, nit it out to allow the API to fill it in.
```
slevenick:
```suggestion
				// if the new boot disk configuration size_gb field is not changing, nil it out to allow the API to fill it in.
```

mmv1/third_party/terraform/services/container/node_config.go.tmpl
```diff
@@ -2350,6 +2446,34 @@ func nodePoolNodeConfigUpdate(d *schema.ResourceData, config *transport_tpg.Conf
 				req.StoragePools = storagePools
 			}
 
+			if v, ok := d.GetOk("node_config.0.boot_disk"); ok {
```
slevenick, line 2449: I'm getting into a weird permadiff loop here. If I have a config that specifies both of these fields with different values the apply "succeeds" but the request is rejected with a 400 error complaining that both are set with different values.

It seems like we aren't correctly checking the result of the update call in this case.

Would it make sense to prevent users from specifying both versions of the field, like have a conflict between `disk_size_gb` and `boot_disk.size_gb` to prevent this?

nathanielford: That is intentional; the API will reject it with an error message if the two values are set differently.

We could make a change in the terraform code to handle this, but the goal was to allow as much of the logic as possible to live in the backend, since the client code is difficult to change in backwards compatible ways. If you think it's a blocker (i.e. the error isn't clear enough to the user that those need to be set similarly), we can discuss options.

mmv1/third_party/terraform/services/container/node_config.go.tmpl
nathanielford: Fixed!

PR #14601, opened 2025-07-23
Add capability related fields to the folder schema

blacken57:
Adds two new fields to the Folders Schema: configured_capabilities and management_project. These fields are populated when a capability is enabled on a Folder. A capability is enabled/disabled as part of a UpdateCapability call. These fields are output only and cannot be manipulated by directly updating the Folder.  

Fixes b/422638425

```release-note:enhancement
resourcemanager: Added `management_project` and `configured_capabilities` fields to the `google_folder` resource.
```

### DIFF
--- a/mmv1/third_party/terraform/services/resourcemanager/resource_google_folder.go
+++ b/mmv1/third_party/terraform/services/resourcemanager/resource_google_folder.go
@@ -78,6 +78,17 @@ func ResourceGoogleFolder() *schema.Resource {
 				Elem:        &schema.Schema{Type: schema.TypeString},
 				Description: `A map of resource manager tags. Resource manager tag keys and values have the same definition as resource manager tags. Keys must be in the format tagKeys/{tag_key_id}, and values are in the format tagValues/456. The field is ignored when empty. This field is only set at create time and modifying this field after creation will trigger recreation. To apply tags to an existing resource, see the google_tags_tag_value resource.`,
 			},
+			"configured_capabilities": {
+				Type:        schema.TypeList,
+				Computed:    true,
+				Elem:        &schema.Schema{Type: schema.TypeString},
+				Description: `A list of capabilities that are configured for this folder.`,
+			},
+			"management_project": {
+				Type:        schema.TypeString,
+				Computed:    true,
+				Description: `The Management Project associated with the folder's configured capabilities.`,
+			},
 		},
 		UseJSONNumber: true,
 	}
@@ -178,6 +189,12 @@ func resourceGoogleFolderRead(d *schema.ResourceData, meta interface{}) error {
 	}
 	if err := d.Set("create_time", folder.CreateTime); err != nil {
 		return fmt.Errorf("Error setting create_time: %s", err)
+	}
+	if err := d.Set("configured_capabilities", folder.ConfiguredCapabilities); err != nil {
+		return fmt.Errorf("Error setting configured_capabilities: %s", err)
+	}
+	if err := d.Set("management_project", folder.ManagementProject); err != nil {
+		return fmt.Errorf("Error setting management_project: %s", err)
 	}
 
 	return nil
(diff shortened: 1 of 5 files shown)


mmv1/third_party/terraform/services/resourcemanager3/resource_resource_manager_capability_test.go.tmpl
```diff
@@ -27,6 +30,10 @@ func TestAccResourceManagerCapability_resourceManagerCapabilityExample_basic(t *
 		Steps: []resource.TestStep{
 			{
 				Config: testAccResourceManagerCapability_resourceManagerCapabilityExample_basic(context),
+				Check: resource.ComposeTestCheckFunc(
```
BBBmau, line 37: it seems like the fields are never added into the test configuration which is why we see a missing test report

## Missing test report
Your PR includes resource fields which are not covered by any test.

Resource: `google_folder` (177 total tests)
Please add an acceptance test which includes these fields. The test should include the following:

```hcl
resource "google_folder" "primary" {
  configured_capabilities = # value needed
  management_project      = # value needed
}


```

blacken57: Hi, could you clarify in which configuration these fields should be added? These new fields are output only, and are only populated after a capability is updated with the folder as a parent. Owing to this, the fields cannot be populated from the folder tests. 
Since these fields make sense from a Capability's perspective, we have added tests verifying that these fields are populated for the parent folder in the Capability unit tests. If this is not enough, could you provide some guidance on how we can improve things? Or which configurations to change to make sure the error subsides?

BBBmau: in that case these fields should be marked with just the `computed: true` in both fields, including the `optional: true` makes it seem like it's possible to input a value.

https://github.com/GoogleCloudPlatform/magic-modules/pull/14601/files#diff-8939c1291b39b0d1d28d168c049ea46c2e9aa9be58b3b1dd80bece057dc2cfcfR57-R68

blacken57: Understood. Removed the extra annotations. Thanks

blacken57:
Hi @BBBmau a friendly ping for the change. The new fields are being used in the tests, and have been only marked as compute=true.

melinath:

mmv1/third_party/terraform/services/resourcemanager/resource_google_folder.go
```diff
@@ -78,6 +78,21 @@ func ResourceGoogleFolder() *schema.Resource {
 				Elem:        &schema.Schema{Type: schema.TypeString},
 				Description: `A map of resource manager tags. Resource manager tag keys and values have the same definition as resource manager tags. Keys must be in the format tagKeys/{tag_key_id}, and values are in the format tagValues/456. The field is ignored when empty. This field is only set at create time and modifying this field after creation will trigger recreation. To apply tags to an existing resource, see the google_tags_tag_value resource.`,
 			},
+			"configured_capabilities": {
+				Type:        schema.TypeList,
+				Optional:    true,
```
SirGitsalot: `Optional` and `ForceNew` should be removed from both of these fields since they're purely output-only.

SirGitsalot (changes requested):

mmv1/third_party/terraform/services/resourcemanager/resource_google_folder.go
blacken57: Hi @SirGitsalot  made the changes.

SirGitsalot (changes requested):
LGTM on the code, but please also update:

https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/third_party/terraform/website/docs/d/folder.html.markdown

and

https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/third_party/terraform/website/docs/r/google_folder.html.markdown

Thanks!

blacken57:
Makes sense. Made the changes to add the documentation. It mimics the publicly available docs.

SirGitsalot (approved):

PR #14602, opened 2025-07-23
Add Feature Service attachment tunneling config

Dawid212:
This PR adds tunneling_config to google_compute_service_attachment resource.

```release-note:enhancement
compute: added `tunneling_config` field to `google_compute_service_attachment` resource (beta)
```


mmv1/products/compute/ServiceAttachment.yaml
```diff
@@ -209,6 +236,21 @@ properties:
     immutable: true
     item_type:
       type: String
+  - name: 'tunnelingConfig'
+    type: NestedObject
+    description: |
+      Tunneling configuration for this service attachment.
+    min_version: beta
+    ignore_read: true
```
slevenick, line 248: Why do we need ignore read here?

mmv1/products/compute/ServiceAttachment.yaml
```diff
@@ -209,6 +236,21 @@ properties:
     immutable: true
     item_type:
       type: String
+  - name: 'tunnelingConfig'
```
slevenick, line 243: Please add an update test for this field

slevenick (changes requested):
Please format the changelog entry to match https://googlecloudplatform.github.io/magic-modules/code-review/release-notes/#type-specific-guidelines-and-examples

It should use `snake_case` for field names, and the resource name should match the resource name in Terraform rather than the API

mmv1/products/compute/ServiceAttachment.yaml
Dawid212: Thanks for the feedback. Updated it!

slevenick (changes requested):
To add an update test you'll need to add it to the handwritten test file: https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/third_party/terraform/services/compute/resource_compute_service_attachment_test.go

We don't currently support update tests via the YAML, though it's on our roadmap!

mmv1/products/compute/ServiceAttachment.yaml
Dawid212: It seems the API doesn't return the TunnelingConfig field in read responses, although it works correctly during create and update operations.

PR #14610, opened 2025-07-23
Fix partial match regexes

NickElliot:
fixes https://github.com/hashicorp/terraform-provider-google/issues/21012

With https://github.com/hashicorp/terraform-provider-google/issues/14731 being resolved a couple years ago the majority of import format validation issues were resolved for generated resources. These missed changing the vast majority of handwritten imports though, and this PR covers all of the ones that could experience this issue that have been identified.

The specific scenario in this issue is for resources like the referenced `google_dns_record_set` that do have stricter validation for their abbreviated IDs (most easily identifiable by usage of capture groups such as `(?P<project>[^/]+)` where the `[^/]` bit enforces the pattern match to have a specific number of `/`s. Due to our regex going for a sub string match, this usually allowed looser validation than was intended. 

Within this PR I have adjusted remaining instances of import formats that do have this stricter validation to use proper `^$` tokens so they are as strict as intended, and should prevent false positives like in this token.

That said, there are another several resources that do use much looser validation and could still experience this issue, most often older handwritten code for child-resources that aligns with the following comment/structure:

```
    // current import_formats cannot import fields with forward slashes in their value
    if err := tpgresource.ParseImportId([]string{
    		"(?P<instance_id>.+)/attachments/(?P<name>.+)",
		"(?P<instance_id>.+)/(?P<name>.+)",
```

wherein `foo/bar/attachments/foo/bar/foo/bar` would be valid and accepted by the first pattern, and `foo/barfoo/bar/foobar/fsdfjsdlasdf` would be accepted by the second, functionally allowing any import to be attempted as long as it had at least one `/`.

This scoped ticket specifically is looking to prevent sub-matching, and this PR resolves all cases of erroneous submatching identified at this time but it would be worth considering a seperate ticket/PR to resolve those unintentionally loose handwritten import IDs.

Lastly, it is possible that some of these will end up causing repeats of issues like https://github.com/hashicorp/terraform-provider-google/issues/16160

```release-note:bug
provider: fixed many import functions throughout the provider that erroneously matched a subset of the provided input, leading to unclear error messages when using `terraform input` with invalid resource IDs
```

### DIFF
--- a/mmv1/templates/terraform/custom_import/extract_taxonomy.go.tmpl
+++ b/mmv1/templates/terraform/custom_import/extract_taxonomy.go.tmpl
@@ -1,7 +1,7 @@
 	config := meta.(*transport_tpg.Config)
 
 	if err := tpgresource.ParseImportId([]string{
-		"(?P<taxonomy>projects/[^/]+/locations/[^/]+/taxonomies/[^/]+)/policyTags/(?P<name>.+)"}, d, config); err != nil {
+		"^(?P<taxonomy>projects/[^/]+/locations/[^/]+/taxonomies/[^/]+)/policyTags/(?P<name>.+)$"}, d, config); err != nil {
 		return nil, err
 	}
 
--- a/mmv1/templates/terraform/custom_import/vertex_ai_tensorboard_import.go.tmpl
+++ b/mmv1/templates/terraform/custom_import/vertex_ai_tensorboard_import.go.tmpl
@@ -1,9 +1,9 @@
     config := meta.(*transport_tpg.Config)
 	if err := tpgresource.ParseImportId([]string{
-		"projects/(?P<project>[^/]+)/locations/(?P<region>[^/]+)/tensorboards/(?P<name>[^/]+)",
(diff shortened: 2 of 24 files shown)


hao-nan-li (commented):
Is this a breaking-change? I'm just wondering why is it included in the major release?

shuyama1:
This change breaks TestAccApigeeSharedflowDeployment_apigeeSharedflowDeploymentTestExample

Error:
`Import id "organizations/tf-testxd19pbjofv/environments/tf-testxd19pbjofv/sharedflows/tf-test-apigee-sharedflow/revisions/1/deployments" doesn't match any of the accepted formats: [^organizations/(?P<org_id>[^/]+)/environments/(?P<environment>[^/]+)/sharedflows/(?P<sharedflow_id>[^/]+)/revisions/(?P<revision>[^/]+)$ ^(?P<org_id>[^/]+)/(?P<environment>[^/]+)/(?P<sharedflow_id>[^/]+)/(?P<revision>[^/]+)$]`

[TeamCity Link](https://hashicorp.teamcity.com/test/8277136818848289339?currentProjectId=TerraformProviders_GoogleCloud_GOOGLE_BETA_NIGHTLYTESTS)

PR #14617, opened 2025-07-24
vertexai: marked `enable_secure_private_service_connect` in `google_vertex_ai_endpoint` as beta isntead of GA (beta)

ramonvermeulen:
Closes: https://github.com/hashicorp/terraform-provider-google/issues/23696

Would probably be flagged as a breaking change, however, I would assume this is an exception given the field was never usable in the GA API.

```release-note:bug
vertexai: marked `enable_secure_private_service_connect` in `google_vertex_ai_endpoint` as beta isntead of GA (beta)
```


c2thorn (commented):
@ramonvermeulen Is it okay if I change the base to point to the 7.0.0 feature branch? We are a few weeks away from the major release cutoff anyway, and might as well add this in there. I don't think there's much difference in waiting for that? Especially with the base API change.

ramonvermeulen:
> @ramonvermeulen Is it okay if I change the base to point to the 7.0.0 feature branch? We are a few weeks away from the major release cutoff anyway, and might as well add this in there. I don't think there's much difference in waiting for that? Especially with the base API change.

Sure, I opened another PR at https://github.com/GoogleCloudPlatform/magic-modules/pull/14665, this one can be closed.

PR #14627, opened 2025-07-24
feat: define smt examples for pubsub topics and subscriptions

abbrowne126:

Add [MessageTransform](https://github.com/googleapis/googleapis/blob/master/google/pubsub/v1/pubsub.proto#L716-L728) field to `google-pubsub-topic`.

```release-note:none
pubsub: added `message_transform` examples for `google_pubsub_topic` and `google_pubsub_subscription` resources
```

### DIFF
--- a/mmv1/products/pubsub/Subscription.yaml
+++ b/mmv1/products/pubsub/Subscription.yaml
@@ -116,12 +116,12 @@ examples:
       subscription_name: 'example-subscription'
       bucket_name: 'example-bucket'
       service_account_id: 'example-stw'
-  - name: 'pubsub_subscription_using_single_smt'
+  - name: 'pubsub_subscription_single_smt'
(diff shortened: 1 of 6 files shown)


abbrowne126:
> As written, these will solely exist in the repo, which is normally not what we want. Is the intention to have these visible on the docs site, or perhaps to have them run as automated tests?

We'd like these to show up on the terraform.io site. I followed an older PR of ours which indicated this is where they should be added, but didn't see it mentioned on the docs. Where should these go?

abbrowne126:
> Ah ok, you do have the files in the correct location, but you need to add example blocks for them in https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/products/pubsub/Subscription.yaml#L47

Thank you - this is done now

roaks3 (approved):

PR #14628, opened 2025-07-24
`teamcity`: add release_tests subproject

BBBmau:
I've setup a prototype of how the project will appear in TeamCity, the scheduling triggers are currently disabled for now.

`Weekly Diff Test` subproject in Google Beta within `Development` root project: https://hashicorp.teamcity.com/project/Development_TerraformProviderGoogleDiffTestProject?branch=refs%2Fheads%2Fauto-pr-12783&mode=builds

It adds `RELEASE_DIFF` envvar and sets it to true for the `Weekly Diff Test` project while also updating the cronjobs to ensure that we don't run nightly tests on saturday in order to run the weekly diff test that day. Sweepers remain the same as the sweepers are searching for resources in both GA + Beta upon build regardless of nightly or weekly

```release-note:none

```

### DIFF
--- a/mmv1/third_party/terraform/.teamcity/components/constants.kt
+++ b/mmv1/third_party/terraform/.teamcity/components/constants.kt
@@ -20,8 +20,8 @@ const val DefaultParallelism = 6
 // this is updated semi-regularly
 const val DefaultTerraformCoreVersion = "1.11.0"
 
-// This represents a cron view of days of the week - we skip saturday to allow release diff tests to run
-const val DefaultDaysOfWeek = "1-6"
+// This represents a cron view of days of the week
+const val DefaultDaysOfWeek = "*"
 
 // Cron value for any day of month
 const val DefaultDaysOfMonth = "*"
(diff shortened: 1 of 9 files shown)


mmv1/third_party/terraform/.teamcity/components/constants.kt
```diff
@@ -19,8 +20,8 @@ const val DefaultParallelism = 6
 // this is updated semi-regularly
 const val DefaultTerraformCoreVersion = "1.11.0"
 
-// This represents a cron view of days of the week
-const val DefaultDaysOfWeek = "*"
+// This represents a cron view of days of the week - we skip saturday to allow release diff tests to run
+const val DefaultDaysOfWeek = "1-6"
```
roaks3: I think we should leave this off for now, ie. revert these lines. To start out, we'll just be experimenting with a low number of tests, and we don't want to have to handle the missing day of regular tests quite yet (particularly regarding metrics).

roaks3 (commented):
LGTM, but I think along with not setting up the trigger for the new Diff Tests, we should also keep the regular tests running every night (for now)

mmv1/third_party/terraform/.teamcity/components/constants.kt
BBBmau: just reverted in latest commit

PR #14637, opened 2025-07-25
Adding support for missing AH fields

arnavdham:
This PR identifies the missing fields in the whole of Analytics Hub API and adds support for the missing fields including 

DataExchange [API](https://cloud.google.com/bigquery/docs/reference/analytics-hub/rest/v1/projects.locations.dataExchanges)
DiscoveryType

Listings [API](https://cloud.google.com/bigquery/docs/reference/analytics-hub/rest/v1/projects.locations.dataExchanges.listings)
state
DiscoveryType
allowOnlyMetadataSharing

@c2thorn @shashambhavi

```release-note:enhancement
bigqueryanalyticshub: added `discovery_type` field to `google_bigquery_analytics_hub_data_exchange` resource
```

```release-note:enhancement
bigqueryanalyticshub: added `state`, `discovery_type`, and `allow_only_metadata_sharing` fields to `google_bigquery_analytics_hub_listing` resource
```


mmv1/products/bigqueryanalyticshub/DataExchange.yaml
```diff
@@ -144,8 +152,20 @@ properties:
...
+        conflicts:
+          - 'discovery_type'
         properties:
           []
+  - name: 'discoveryType'
+    type: Enum
+    enum_values:
+      - 'DISCOVERY_TYPE_PRIVATE'
+      - 'DISCOVERY_TYPE_PUBLIC'
+    default_value: 'DISCOVERY_TYPE_PRIVATE'
```
wj-chen: Is this a service-side default? e.g. if user sends a REST API directly without a discovery type value, does the API response have `discoveryType = DISCOVERY_PRIVATE` (vs. not return it)? Changing `default_value` is a breaking change so we should be careful about whether to set it and what to set it to. Might also be able to simply/remove the diff suppression if we don't need to set it.

mmv1/third_party/terraform/services/bigqueryanalyticshub/resource_bigquery_analytics_hub_listing_test.go
```diff
@@ -130,3 +130,69 @@ resource "google_bigquery_analytics_hub_listing" "listing_pubsub" {
...
+}
+
+resource "google_bigquery_analytics_hub_listing" "listing" {
+  location         = "US"
+  data_exchange_id = google_bigquery_analytics_hub_data_exchange.listing.data_exchange_id
+  listing_id       = "tf_test_my_listing%{random_suffix}"
+  display_name     = "tf_test_my_listing%{random_suffix}"
+  description      = "example public listing%{random_suffix}"
+  discovery_type   = "DISCOVERY_TYPE_PRIVATE"
+	allow_only_metadata_sharing= false
```
wj-chen: Indentation

mmv1/products/bigqueryanalyticshub/DataExchange.yaml
arnavdham: Yes, default is DISCOVERY_PRIVATE. https://cloud.google.com/bigquery/docs/reference/analytics-hub/rest/v1/DiscoveryType.
The diff_suppress is needed because sending the value for a DCR or a restricted listing makes no sense as DiscoveryType is just a feature for a normal Data Exchange and Normal listing. The existing tests start to fail without the diff_suppress as the plan isn't empty after the creation of a DCR.

mmv1/third_party/terraform/services/bigqueryanalyticshub/resource_bigquery_analytics_hub_listing_test.go
arnavdham: fixed.

mmv1/products/bigqueryanalyticshub/DataExchange.yaml
zli82016: `default_value` should not be used as the API doesn't return the default value in all cases.

https://googlecloudplatform.github.io/magic-modules/develop/diffs/#default

mmv1/products/bigqueryanalyticshub/Listing.yaml
```diff
@@ -281,6 +292,25 @@ properties:
...
+    type: String
+    description: |-
+      Current state of the listing.
+    output: true
+  - name: 'discoveryType'
+    type: Enum
+    enum_values:
+      - 'DISCOVERY_TYPE_PRIVATE'
+      - 'DISCOVERY_TYPE_PUBLIC'
+    default_value: 'DISCOVERY_TYPE_PRIVATE'
```
zli82016: If the API side doesn't return the default value in all cases, `default_value` shouldn't be used.

zli82016 (changes requested):

mmv1/products/bigqueryanalyticshub/Listing.yaml
arnavdham: If the listing is not restricted it returns DISCOVERY_TYPE_PRIVATE. If restricted,it does not return the field as a whole itself and the case was handled using diff suppress. Same thing for a Data Exchange,it returns DISCOVERY_TYPE_PRIVATE, if the Data exchange is a DCR it does not return the field as a whole itself and the case was handled using diff suppress.  I think this would make the whole thing work. 
1)Is there any issue you see with the current handling of the code?
2)If yes, what workaround do you propose?

arnavdham: If we don't set a default value and a user doesn't set discovery_type in his terraform config and the API returns DISCOVERY_TYPE_PRIVATE for non-restricted listings. It will always cause a diff from private->null and the tf plan will never be empty.

mmv1/products/bigqueryanalyticshub/DataExchange.yaml
zli82016: >  sending the value for a DCR or a restricted listing makes no sense as DiscoveryType is just a feature for a normal Data Exchange and Normal listing. 

Thanks for the information. Can you clarify the following question? 
What will happen when sending the value for a DCR or a restricted listing on API side? Will API return an error?

Thanks.

mmv1/third_party/terraform/services/bigqueryanalyticshub/resource_bigquery_analytics_hub_listing_test.go
```diff
@@ -188,13 +188,77 @@ resource "google_bigquery_analytics_hub_listing" "listing" {
...
+			{
+				Config: testAccBigqueryAnalyticsHubListing_bigqueryAnalyticshubPublicListingUpdate(context),
+			},
+			{
+				ResourceName:            "google_bigquery_analytics_hub_listing.listing",
+				ImportState:             true,
+				ImportStateVerify:       true,
+				ImportStateVerifyIgnore: []string{"data_exchange_id", "listing_id", "location"},
+			},
+		},
```
zli82016, line 232: Can you add one more test step to change the `discovery_type` to `DISCOVERY_TYPE_PUBLIC` in resource `google_bigquery_analytics_hub_data_exchange`? Will the `discovery_type` in resource `google_bigquery_analytics_hub_listing` also be updated automatically?

mmv1/products/bigqueryanalyticshub/DataExchange.yaml
arnavdham: No,not an error. It just ignores the field which is being sent and doesn't return it back in the response for a DCR/Restricted Listing. https://paste.googleplex.com/6544530735890432 for reference.

mmv1/third_party/terraform/services/bigqueryanalyticshub/resource_bigquery_analytics_hub_listing_test.go
arnavdham: Yes, the changing the discovery_type on data_exchange level causes it to change(overwrite) on all the listings present inside it. [Relevant Proto](https://source.corp.google.com/piper///depot/google3/google/cloud/bigquery/exchange/v1main/analyticshub.proto;l=1339;rcl=788357801)

zli82016: Thanks for the information. 

As `discovery_type` of listings is overwritten, there will be a perma-diff for `discovery_type` in listings.
 
`default_from_api: true` should be used for `discoveryType` instead of the `default_value` to fix the perma-diff.

mmv1/products/bigqueryanalyticshub/Listing.yaml
zli82016: `default_from_api: true` should be used for discoveryType instead of `default_value`. And `diff_suppress_func` is not needed.

mmv1/products/bigqueryanalyticshub/DataExchange.yaml
zli82016: `default_from_api: true` can be used instead of `default_value`. 

And `diff_suppress_func` is not needed.

mmv1/third_party/terraform/services/bigqueryanalyticshub/resource_bigquery_analytics_hub_listing_test.go
arnavdham: Thank you! Have made the necessary changes.

mmv1/products/bigqueryanalyticshub/DataExchange.yaml
```diff
@@ -146,6 +153,13 @@ properties:
           - 'sharing_environment_config.0.dcr_exchange_config'
         properties:
           []
+  - name: 'discoveryType'
+    type: Enum
+    enum_values:
+      - 'DISCOVERY_TYPE_PRIVATE'
+      - 'DISCOVERY_TYPE_PUBLIC'
+    default_from_api: true
+    description: Type of discovery on the discovery page for all the listings under this exchange. Updating this field also updates (overwrites) the discoveryType field for all the listings under this exchange.
```
zli82016: Can you add a note that DiscoveryType is a feature for a normal Data Exchange, not for DCR? 

Thanks.

mmv1/products/bigqueryanalyticshub/Listing.yaml
```diff
@@ -281,6 +289,22 @@ properties:
...
+    description: |-
+      Current state of the listing.
+    output: true
+  - name: 'discoveryType'
+    type: Enum
+    enum_values:
+      - 'DISCOVERY_TYPE_PRIVATE'
+      - 'DISCOVERY_TYPE_PUBLIC'
+    default_from_api: true
+    description: Specifies the type of discovery on the discovery page. Note that this does not control the visibility of the exchange/listing which is defined by IAM permission.
```
zli82016: Can you add a note that DiscoveryType is a feature for a normal listing, not for a restricted listing? 

Thanks.

zli82016 (approved):
Thanks for working on this.

zli82016:
@wj-chen , do you mind reviewing the PR again? Thanks.

wj-chen (approved):

wj-chen:
> @wj-chen , do you mind reviewing the PR again? Thanks.

Approved. Thank you for your review as well.

zli82016:
BTW, I modified the release notes a little bit.

PR #14645, opened 2025-07-25
tgc-revival: make test data fall back to most recent run

trodge:
Currently the behavior is to always use the most recent run. This will make it so that we use the most recent of the last 30 runs that has  at least one `ResourceMetadata`. We could still fail to find data if a run results in resource metadata without having the resource we're interested in, but I think we can prevent this from happening by controlling how the data gets stored.

```release-note:none

```

### DIFF
--- a/mmv1/third_party/tgc_next/test/setup.go
+++ b/mmv1/third_party/tgc_next/test/setup.go
@@ -27,6 +27,11 @@ type CaiData struct {
 	CaiAsset caiasset.Asset `json:"cai_asset,omitempty"`
 }
 
+type NightlyRun struct {
+	MetadataByTest map[string]TgcMetadataPayload
+	Date           time.Time
+}
+
 type TgcMetadataPayload struct {
 	TestName         string                       `json:"test_name"`
 	RawConfig        string                       `json:"raw_config"`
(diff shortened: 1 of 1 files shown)


mmv1/third_party/tgc_next/test/setup.go
```diff
@@ -45,49 +50,50 @@ type Resource struct {
...
-			if err != nil {
-				return nil, fmt.Errorf("storage.NewClient: %v", err)
-			}
-			defer client.Close()
-
-			currentDate = currentDate.AddDate(0, 0, -1)
+		bucket := client.Bucket(bucketName)
 
-			rc, err := client.Bucket(bucketName).Object(objectName).NewReader(ctx)
+		var allErrs error
```
trodge, line 76:
```suggestion
		var allErrs error
		retriesRemaining := 30
```

mmv1/third_party/tgc_next/test/setup.go
```diff
@@ -45,49 +50,50 @@ type Resource struct {
...
 				}
 			}
-			defer rc.Close()
-
-			data, err := io.ReadAll(rc)
-			if err != nil {
-				return nil, fmt.Errorf("io.ReadAll: %v", err)
+			if metadata == nil {
+				// Keep looking until we find a date with metadata.
+				i--
```
trodge, line 89:
```suggestion
				i--
				retriesRemaining--
				if retriesRemaining < 0 {
					return nil, fmt.Errorf("too many retries, %v", allErrs)
				}
```

mmv1/third_party/tgc_next/test/setup.go
```diff
@@ -45,49 +50,55 @@ type Resource struct {
...
-			defer client.Close()
-
-			currentDate = currentDate.AddDate(0, 0, -1)
+		bucket := client.Bucket(bucketName)
 
-			rc, err := client.Bucket(bucketName).Object(objectName).NewReader(ctx)
+		var allErrs error
+		retriesRemaining := 30
+		for i := 0; i < len(TestsMetadata); i++ {
+			metadata, err := readTestsDataFromGCSForRun(ctx, currentDate, bucketName, bucket)
```
trodge, line 79: We could make this a goroutine if fetching 30 separate objects takes too long.

mmv1/third_party/tgc_next/test/setup.go
```diff
@@ -45,49 +50,55 @@ type Resource struct {
...
 			}
-			defer rc.Close()
-
-			data, err := io.ReadAll(rc)
-			if err != nil {
-				return nil, fmt.Errorf("io.ReadAll: %v", err)
+			if metadata == nil {
+				// Keep looking until we find a date with metadata.
+				i--
+				retriesRemaining--
```
zli82016: In what use cases  `retriesRemaining` is used? Is `i` enough to download the data for 30 days?

trodge: If all our requests fail to get metadata `i` will just keep getting decremented and never get over 30. `retriesRemaining` will eventually reach 0 in that case.

mmv1/third_party/tgc_next/test/setup.go
zli82016: got it. Thanks for the information. Can you add a comment for this case? 

Also, we can use a constant for `30`.

zli82016 (changes requested):

mmv1/third_party/tgc_next/test/setup.go
trodge: Done.

Also in this implementation we could potentially go back as far as 60 days fetching data if a test has a >50% failure rate, but I think this is acceptable since we'll still be using the most recent available data when running tests.

zli82016: Sure. We can go with `30` for now.

zli82016 (approved):
Thanks.

zli82016:
Can you use the goroutines to download the files from the GCS bucket in parallel? 

The time to run one test becomes much longer.

Thanks.

PR #14649, opened 2025-07-25
Revert "Fixing hive partitioning diff (#14483)"

slevenick:
This reverts commit 43ca35a159e375c2668ddf5b45122fdeab748f89.

Due to https://github.com/hashicorp/terraform-provider-google/issues/23718

```release-note:bug
bigquery: fixed a crash in `google_bigquery_table` due to `parquet_options` being specified with an external table that was introduced in 6.45.0
```


wj-chen (approved):

mmv1/third_party/terraform/services/bigquery/resource_bigquery_table.go.tmpl
```diff
@@ -1345,13 +1345,6 @@ func ResourceBigQueryTable() *schema.Resource {
 				Description: `Whether Terraform will be prevented from destroying the instance. When the field is set to true or unset in Terraform state, a terraform apply or terraform destroy that would delete the table will fail. When the field is set to false, deleting the table is allowed.`,
 			},
 
-			"ignore_auto_generated_schema": {
```
rileykarson, line 1348: This is going to remove the field, which can be a bit annoying- should we keep it but make it nonfunctional in the meantime, and add behaviour back with a fix, or try to fix first?

wj-chen: I think removing vs making nonfunctional have about the same effect on users that rely on this field in the next version? But yeah either way we should call it out in the release note clearly that the field will be temporarily removed/nonfunctional.

\+ @sachinpro to assess the likelihood of a fix-forward before the next regular release cut. If we can make it, shouldn't need the rollback.

sachinpro: https://github.com/GoogleCloudPlatform/magic-modules/pull/14659 should fix this. Then we won't need this PR.

PR #14650, opened 2025-07-25
Add deletion support for Endpoint With Model Garden Deployment resource

francis-ohara:
This PR adds the necessary custom code templates and resource configuration fields to enable the `vertexai_endpoint_with_model_garden_deployment`  resource via `terraform destroy`.

Specifically, this PR:
- Adds a new field to the EndpointWithModelGardenDeployment.yaml file for storing the deployed model IDs of the models deployed to the Endpoint.  
 - Updates the post_create template to read the deployed model IDs from an API call and populate the field's value.  
 - Adds a custom_delete template that undeploys all models deployed to the endpoint using the deployed model IDs and deletes the endpoint.  
 - Updates acceptance tests to verify that resource is deleted.

Vertex AI Model Garden service
Based on:
 - [ModelGardenService.deploy](https://cloud.google.com/vertex-ai/docs/reference/rest/v1beta1/projects.locations/deploy)   
 -  [EndpointService.undeployModel](https://cloud.google.com/vertex-ai/docs/reference/rest/v1beta1/projects.locations.endpoints/undeployModel)
 - [EndpointService.deleteEndpoint](https://cloud.google.com/vertex-ai/docs/reference/rest/v1beta1/projects.locations.endpoints/delete)

Addresses:
 - https://github.com/hashicorp/terraform-provider-google/issues/23217 
 - https://github.com/hashicorp/terraform-provider-google/issues/22478

`google_vertex_ai_endpoint_with_model_garden_deployment`

### DIFF
--- a/mmv1/products/vertexai/EndpointWithModelGardenDeployment.yaml
+++ b/mmv1/products/vertexai/EndpointWithModelGardenDeployment.yaml
@@ -33,8 +33,8 @@ self_link: "projects/{{project}}/locations/{{location}}/endpoints/{{endpoint}}"
 create_url: "projects/{{project}}/locations/{{location}}:deploy"
 immutable: true
 exclude_read: true
-exclude_delete: true  # the resource does not yet support deletion
-exclude_import: true  # the resource does not support import
+exclude_delete: true # the resource does not yet support deletion
+exclude_import: true # the resource does not support import
 timeouts:
   insert_minutes: 180
 autogen_status: RW5kcG9pbnRXaXRoTW9kZWxHYXJkZW5EZXBsb3ltZW50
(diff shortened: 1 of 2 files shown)


francis-ohara:
> Non-blocking, but only question is around the number of tests. Are these resources particularly expensive or slow, and how does that weigh against the benefit of including the tests (like mulitple models in parallel)? Just curious, if they're sufficiently cheap/fast then the set of tests looks great.

Thanks @roaks3. Deploying these models can be slow but the multiple models in parallel test specifically is only included as an example in the YAML file to generate documentation and isn't actually included as one of the tests in the [handwritten test file](https://github.com/francis-ohara/magic-modules/blob/6e2862e315695f6891ae25f4219a7117ec9118e1/mmv1/third_party/terraform/services/vertexai/resource_vertex_ai_deploy_test.go).

PR #14659, opened 2025-07-28
fix(hashicorp/terraform-provider-google#23718): removing unnecessary …

sachinpro:
fixes hashicorp/terraform-provider-google#23718.

```release-note:bug
bigquery: fixing crash during call to `expandParquetOptions` from redundant `resourceTable` call.
```


wj-chen:
From my understanding of the `TestAccBigQueryExternalDataTable_CSV_WithSchemaAndConnectionIDAndHivePartitioning` failure - Since we remove the generated columns on the TF side, when there is an Update, we only send the user-provided columns in the request, and that gets rejected by the server because it seems that the schema in the request is missing the generated columns.

One option for the fix may be when we filter out the generated columns in `Read()`, collect them and save separately into an output-only virtual field `generated_schema_columns` or so to be merged with `schema` for future Updates. Let me know what you think and if there are other ideas.

wj-chen:
Also add https://github.com/hashicorp/terraform-provider-google/issues/23655 to "fixes" in Description and update the release note (e.g. "fixing" -> "fixed").

sachinpro:
> Also add [hashicorp/terraform-provider-google#23655](https://github.com/hashicorp/terraform-provider-google/issues/23655) to "fixes" in Description and update the release note (e.g. "fixing" -> "fixed").

Done.

sachinpro:
> From my understanding of the `TestAccBigQueryExternalDataTable_CSV_WithSchemaAndConnectionIDAndHivePartitioning` failure - Since we remove the generated columns on the TF side, when there is an Update, we only send the user-provided columns in the request, and that gets rejected by the server because it seems that the schema in the request is missing the generated columns.
> 
> One option for the fix may be when we filter out the generated columns in `Read()`, collect them and save separately into an output-only virtual field `generated_schema_columns` or so to be merged with `schema` for future Updates. Let me know what you think and if there are other ideas.

@wj-chen I have added this logic in the [commit](https://github.com/GoogleCloudPlatform/magic-modules/pull/14659/commits/ab3d860c3ba682d8051693f89c79e384d8163e38).

It is unfortunate that we need to ignore the virtual fields in all our tests.

sachinpro:
Note that the first [commit](https://github.com/GoogleCloudPlatform/magic-modules/pull/14659/commits/1806bd356f2aa9f6b3b279583475f4e7a367104b) independently fixes both the issues I mentioned in the description. The other commit is only for fixing a test which was already failing.

wj-chen:
> > From my understanding of the `TestAccBigQueryExternalDataTable_CSV_WithSchemaAndConnectionIDAndHivePartitioning` failure - Since we remove the generated columns on the TF side, when there is an Update, we only send the user-provided columns in the request, and that gets rejected by the server because it seems that the schema in the request is missing the generated columns.
> > One option for the fix may be when we filter out the generated columns in `Read()`, collect them and save separately into an output-only virtual field `generated_schema_columns` or so to be merged with `schema` for future Updates. Let me know what you think and if there are other ideas.
> 
> @wj-chen I have added this logic in the [commit](https://github.com/GoogleCloudPlatform/magic-modules/pull/14659/commits/ab3d860c3ba682d8051693f89c79e384d8163e38).
> 
> It is unfortunate that we need to ignore the virtual fields in all our tests.

Thank you for fixing this!

To keep documentations consistent, let's add `ignore_auto_generated_schema` and `generated_schema_columns` to https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/products/bigquery/Table.yaml, and `generated_schema_columns` to https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/third_party/terraform/website/docs/r/bigquery_table.html.markdown?plain=1. Otherwise the change LGTM.

sachinpro:
> To keep documentations consistent, let's add `ignore_auto_generated_schema` and `generated_schema_columns` to https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/products/bigquery/Table.yaml, and `generated_schema_columns` to https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/third_party/terraform/website/docs/r/bigquery_table.html.markdown?plain=1. Otherwise the change LGTM.

I have added them to `virtual_fields` in [Table.yaml](https://github.com/GoogleCloudPlatform/magic-modules/pull/14659/files#diff-3ac5e2e884d581df6e035f060f7dad7c7c66923cfc5e1ddb2de1fb183a66fdcbR641).

As we discussed, to avoid confusion, I am not updating the documentation to include `generated_schema_columns`.

ScottSuarez:
@modular-magician reassign-reviewer @slevenick 

sam could you take this over since you have context on the original change?

sachinpro:
@slevenick gentle reminder to review this. Thanks!

slevenick (approved):

PR #14676, opened 2025-07-29
Add allow_fewer_zones_deployment to Redis Cluster

NA2047:

```release-note:enhancement
memorystore: added `allow_fewer_zones_deployment` field to `google_redis_cluster` resource
```

### DIFF
--- a/mmv1/templates/terraform/examples/redis_cluster_aof.tf.tmpl
+++ b/mmv1/templates/terraform/examples/redis_cluster_aof.tf.tmpl
@@ -13,6 +13,7 @@ resource "google_redis_cluster" "{{$.PrimaryResourceId}}" {
     maxmemory-policy	= "volatile-ttl"
   }
   deletion_protection_enabled = {{index $.Vars "deletion_protection_enabled"}}
+  allow_fewer_zones_deployment = true
   zone_distribution_config {
     mode = "MULTI_ZONE"
   }
--- a/mmv1/third_party/terraform/services/redis/resource_redis_cluster_test.go
+++ b/mmv1/third_party/terraform/services/redis/resource_redis_cluster_test.go
@@ -322,47 +322,6 @@ resource "google_redis_cluster" "cluster_abc" {
 `, context)
 }
 
-// Validate that allow_fewer_zones_deployment is set correctly for the cluster
-func TestAccRedisCluster_allowFewerZonesDeployment(t *testing.T) {
-	t.Parallel()
-
-	context := map[string]interface{}{
-		"random_suffix":                acctest.RandString(t, 10),
-		"allow_fewer_zones_deployment": "true",
-	}
-
-	acctest.VcrTest(t, resource.TestCase{
-		PreCheck:                 func() { acctest.AccTestPreCheck(t) },
-		ProtoV5ProviderFactories: acctest.ProtoV5ProviderFactories(t),
-		CheckDestroy:             testAccCheckRedisClusterDestroyProducer(t),
-		Steps: []resource.TestStep{
-			{
-				// create cluster with allow_fewer_zones_deployment set to false (default)
-				Config: testAccRedisCluster_allowFewerZonesDeployment(context),
-			},
-			{
-				ResourceName:            "google_redis_cluster.cluster_afz_main",
-				ImportState:             true,
-				ImportStateVerify:       true,
-				ImportStateVerifyIgnore: []string{"psc_configs"},
-			},
-		},
-	})
-}
-
-func testAccRedisCluster_allowFewerZonesDeployment(context map[string]interface{}) string {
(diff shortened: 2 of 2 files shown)


mmv1/products/redis/Cluster.yaml
```diff
@@ -189,10 +189,13 @@ examples:
       subnet_name: "my-subnet"
       network_name: "my-network"
       deletion_protection_enabled: 'true'
+      allow_fewer_zones_deployment: 'false'
     test_vars_overrides:
       'deletion_protection_enabled': 'false'
+      'allow_fewer_zones_deployment': 'false'
```
ScottSuarez:
```suggestion
```

mmv1/products/redis/Cluster.yaml
```diff
@@ -189,10 +189,13 @@ examples:
...
     oics_vars_overrides:
       'deletion_protection_enabled': 'false'
+      'allow_fewer_zones_deployment': 'false'
```
ScottSuarez:
```suggestion
```
override identical to value above

mmv1/templates/terraform/examples/redis_cluster_aof.tf.tmpl
```diff
@@ -13,6 +13,7 @@ resource "google_redis_cluster" "{{$.PrimaryResourceId}}" {
     maxmemory-policy	= "volatile-ttl"
   }
   deletion_protection_enabled = {{index $.Vars "deletion_protection_enabled"}}
+  allow_fewer_zones_deployment = {{index $.Vars "allow_fewer_zones_deployment"}}
```
ScottSuarez:
```suggestion
  allow_fewer_zones_deployment = false
```

mmv1/products/redis/Cluster.yaml
```diff
@@ -189,10 +189,13 @@ examples:
       subnet_name: "my-subnet"
       network_name: "my-network"
       deletion_protection_enabled: 'true'
+      allow_fewer_zones_deployment: 'false'
```
ScottSuarez:
```suggestion
```
recommend not using variable if doesn't need to be different depending on environment

mmv1/products/redis/Cluster.yaml
```diff
@@ -378,6 +381,14 @@ properties:
...
         description: |
           Immutable. The zone for single zone Memorystore Redis cluster.
+  - name: 'allowFewerZonesDeployment'
+    type: Boolean
+    description: |
+      Allows customers to specify if they are okay with deploying a multi-zone
+      cluster in less than 3 zones. Once set, if there is a zonal outage during
+      the cluster creation, the cluster will only be deployed in 2 zones, and
+      stay within the 2 zones for its lifecycle.
+    default_value: false
```
ScottSuarez: Is there reason we don't set to true in the test. We are setting to a value different from API value. Does api return false by default?

NA2047: Good point @ScottSuarez Ill update and set to true

mmv1/products/redis/Cluster.yaml
NA2047: i'll set this to true

mmv1/templates/terraform/examples/redis_cluster_aof.tf.tmpl
NA2047: same as below

mmv1/products/redis/Cluster.yaml
NA2047: same as below

mmv1/products/redis/Cluster.yaml
NA2047: same as below

mmv1/products/redis/Cluster.yaml
```diff
@@ -189,10 +189,13 @@ examples:
       subnet_name: "my-subnet"
       network_name: "my-network"
       deletion_protection_enabled: 'true'
+      allow_fewer_zones_deployment: 'true'
```
ScottSuarez:
```suggestion
```

we should remove this variable and simply encode true the test.

mmv1/templates/terraform/examples/redis_cluster_aof.tf.tmpl
ScottSuarez:
```suggestion
  allow_fewer_zones_deployment = true
```

ScottSuarez (changes requested):

ScottSuarez:
Hi, I am wondering if there was some confusion with my suggestion. I seems like you are migrating to a handwritten test now. That is okay but I am wondering why.

ScottSuarez:
Looks like it's being returned by the API now. I am wondering what has changed?

PR #14677, opened 2025-07-29
Add allow_fewer_zones_deployment to Memorystore

NA2047:

```release-note:enhancement
memorystore: added `allow_fewer_zones_Deployment` field to `google_memorystore_instance` resource
```

### DIFF
--- a/mmv1/products/memorystore/Instance.yaml
+++ b/mmv1/products/memorystore/Instance.yaml
@@ -65,10 +65,12 @@ examples:
       subnet_name: 'my-subnet'
       network_name: 'my-network'
       prevent_destroy: 'true'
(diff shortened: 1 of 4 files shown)


NickElliot (commented):
LGTM but wanted to verify, are existing instances considered as "false" here? since its only settable on creation I want to verify nobody would become broken on upgrade

NA2047:
> LGTM but wanted to verify, are existing instances considered as "false" here? since its only settable on creation I want to verify nobody would become broken on upgrade

Hey @NickElliot The default is false on the API, I confirmed with the Service Team

NickElliot (approved):
LGTM

PR #14678, opened 2025-07-29
Update Instance.yaml to add new trial editions to Looker

efeelaiho:
Looker is adding new trial platform editions for Looker

```release-note:enhancement
looker: Added `LOOKER_CORE_TRIAL_STANDARD`, `LOOKER_CORE_TRIAL_ENTERPRISE`,  and `LOOKER_CORE_TRIAL_EMBED` editions to  `google_looker_instance` resource.
```

### DIFF
--- a/mmv1/products/looker/Instance.yaml
+++ b/mmv1/products/looker/Instance.yaml
@@ -401,6 +401,7 @@ properties:
     type: Enum
     description: |
       Platform editions for a Looker instance. Each edition maps to a set of instance features, like its size. Must be one of these values:
+      - LOOKER_CORE_TRIAL: trial instance (Currently Unavailable)
       - LOOKER_CORE_STANDARD: pay as you go standard instance (Currently Unavailable)
       - LOOKER_CORE_STANDARD_ANNUAL: subscription standard instance
       - LOOKER_CORE_ENTERPRISE_ANNUAL: subscription enterprise instance
@@ -414,6 +415,7 @@ properties:
     immutable: true
     default_value: "LOOKER_CORE_TRIAL"
     enum_values:
+      - 'LOOKER_CORE_TRIAL'
       - 'LOOKER_CORE_STANDARD'
       - 'LOOKER_CORE_STANDARD_ANNUAL'
       - 'LOOKER_CORE_ENTERPRISE_ANNUAL'


mmv1/products/looker/Instance.yaml
```diff
@@ -401,25 +401,29 @@ properties:
...
       - LOOKER_CORE_NONPROD_STANDARD_ANNUAL: nonprod subscription standard instance
       - LOOKER_CORE_NONPROD_ENTERPRISE_ANNUAL: nonprod subscription enterprise instance
       - LOOKER_CORE_NONPROD_EMBED_ANNUAL: nonprod subscription embed instance
+      - LOOKER_CORE_TRIAL_STANDARD: A standard trial edition of Looker (Google Cloud core) product.
+      - LOOKER_CORE_TRIAL_ENTERPRISE: An enterprise trial edition of Looker (Google Cloud core) product.
+      - LOOKER_CORE_TRIAL_EMBED: An embed trial edition of Looker (Google Cloud core) product.
     immutable: true
     default_value: "LOOKER_CORE_TRIAL"
     enum_values:
-      - 'LOOKER_CORE_TRIAL'
```
rileykarson: This is technically backwards incompatible, even though the API probably rejects this by now- let's keep the value for now, and clean it up as part of the major release . Removing from docs is fine atm.

Our contribution window is open now and will be for a couple more weeks; you'd just do this removal against the major release branch (https://googlecloudplatform.github.io/magic-modules/breaking-changes/make-a-breaking-change/#make-the-change-on-feature-branch-major-release-700) and add a brief note to the upgrade guide (https://googlecloudplatform.github.io/magic-modules/breaking-changes/make-a-breaking-change/#in-the-700-major-release). Most other notes on that page are for larger changes/removals which need more prep than an enum value.

rileykarson (commented):
You could consider dropping this field to a string instead of an enum if new editions will be regularly (~more than every 2y) added. That will remove the validation against unknown values at Terraform's "plan" time, and we'd fail against the server at apply time.

mmv1/products/looker/Instance.yaml
efeelaiho: Sounds good! Updated

rileykarson (approved):

PR #14682, opened 2025-07-29
Add unique index support in Firestore

HansiMou:

firestore: added `unique` field to `google_firestore_index` resource

### DIFF
--- a/mmv1/products/firestore/Index.yaml
+++ b/mmv1/products/firestore/Index.yaml
@@ -92,6 +92,12 @@ examples:
       database_id: 'database-id-sparse-any'
     test_env_vars:
       project_id: 'PROJECT_NAME'
+  - name: 'firestore_index_unique'
+    primary_resource_id: 'my-index'
+    vars:
+      database_id: 'database-id-unique'
+    test_env_vars:
+      project_id: 'PROJECT_NAME'
 parameters:
 properties:
   - name: 'name'
@@ -152,6 +158,12 @@ properties:
       definition reach or traverse an array, except via an explicit array
       index. Violations will result in errors. Note this field only applies to
       indexes with MONGODB_COMPATIBLE_API ApiScope.
+  - name: 'unique'
+    type: Boolean
+    default_value: false
+    description:
+      Optional. Whether it is an unique index. Unique index ensures all values for the
+      indexed field(s) are unique across documents.
   - name: 'fields'
     type: Array
     description: |
@@ -206,7 +218,7 @@ properties:
               send_empty_value: true
               allow_empty_object: true
               properties:
- # Meant to be an empty object with no properties.
+                # Meant to be an empty object with no properties.
                 []
     # Most composite indexes require at least two fields, but it is possible
     # for a user to require a single field index such as `__name__ DESC`.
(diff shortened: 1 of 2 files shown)


mmv1/products/firestore/Index.yaml
```diff
@@ -152,6 +158,12 @@ properties:
       definition reach or traverse an array, except via an explicit array
       index. Violations will result in errors. Note this field only applies to
       indexes with MONGODB_COMPATIBLE_API ApiScope.
+  - name: 'unique'
+    type: Boolean
+    default_value: false
+    description:
+      Optional. Whether it is an unique index. Unique index ensures all values for the
+      indexed field(s) are unique across documents.
```
bw-yuhogg, line 166: "unique across documents" means "unique across all documents indexed by this index as defined by its collection and query scope", right?

mmv1/templates/terraform/examples/firestore_index_unique.tf.tmpl
```diff
@@ -0,0 +1,32 @@
...
+	location_id              = "nam5"
+	type                     = "FIRESTORE_NATIVE"
+	database_edition         = "ENTERPRISE"
+
+	delete_protection_state = "DELETE_PROTECTION_DISABLED"
+	deletion_policy         = "DELETE"
+}
+
+resource "google_firestore_index" "{{$.PrimaryResourceId}}" {
+	project     = "{{index $.TestEnvVars "project_id"}}"
```
bw-yuhogg: [Nitpick] Can you remove a single blank space before the "=" here?

rwhogg (approved):

mmv1/products/firestore/Index.yaml
```diff
@@ -152,6 +158,12 @@ properties:
       definition reach or traverse an array, except via an explicit array
       index. Violations will result in errors. Note this field only applies to
       indexes with MONGODB_COMPATIBLE_API ApiScope.
+  - name: 'unique'
+    type: Boolean
+    default_value: false
```
BBBmau: can you provide information on where this new field is coming from? the [API reference](https://cloud.google.com/firestore/docs/reference/rest/v1/projects.databases.collectionGroups.indexes) docs do not include `unique`

it would probably be best to use the default that comes from the API. we also don't need to include `Optional.` in the description

mmv1/products/firestore/Index.yaml
HansiMou: Correct. Don't want to add too many modifiers here as the list can be large(shard_count, multikey, density)

mmv1/products/firestore/Index.yaml
HansiMou: > can you provide information on where this new field is coming from? the [API reference](https://cloud.google.com/firestore/docs/reference/rest/v1/projects.databases.collectionGroups.indexes) docs do not include `unique`

Hey,

This is a new feature we are to launch today or tomorrow. The docs are still in the draft state as there are many access point we need to update first (terraform, gcloud). Nothing in public and everything is in corp, not sure whether I can add more details here.


> it would probably be best to use the default that comes from the API. we also don't need to include `Optional.` in the description

Removed it from comments.

BBBmau: is it expected for the value to be set to false by default if not being used?

HansiMou: Yes, by default it's false.

HansiMou: Hey Mauricio,

Let me know if there is anything outstanding. I need to merge it in today so that it can be released on Monday. Otherwise it will miss a whole week.

BBBmau: we'd want to have it be `default_from_api: true` then

HansiMou: Added

PR #14687, opened 2025-07-30
Pagination function plural data

BBBmau:
Fixes https://github.com/hashicorp/terraform-provider-google/issues/19157

This simplifies the implementation of plural data-sources by creating a generic function that can be used regardless of the data-source.

The PR represents two uses of the function, with one simplifying implementation within an already existing List function and the second showcasing the function implementation within a Read function 

- [X] google_kms_crypto_keys - [pagination loop](https://github.com/GoogleCloudPlatform/magic-modules/blob/53302465ca4fafd09b0e44f2259b5290297258a5/mmv1/third_party/terraform/services/kms/data_source_google_kms_crypto_keys.go#L129-L171)
- [X] google_kms_key_rings  - [pagination loop](https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/third_party/terraform/services/kms/data_source_google_kms_key_rings.go#L108-L149)
- [X] google_kms_crypto_key_versions - [pagination loop](https://github.com/GoogleCloudPlatform/magic-modules/blob/b1c656e86f2efce6b3043082e7e15147ce32723e/mmv1/third_party/terraform/services/kms/data_source_google_kms_crypto_key_versions.go#L170-L212)
- [X] google_storage_buckets - [pagination loop](https://github.com/hashicorp/terraform-provider-google-beta/blob/c45269d8a2805240fe84577251dc73a66c157924/google-beta/services/storage/data_source_google_storage_buckets.go#L70-L110)
- [x] google_storage_bucket_objects - [pagination loop](https://github.com/hashicorp/terraform-provider-google-beta/blob/c45269d8a2805240fe84577251dc73a66c157924/google-beta/services/storage/data_source_google_storage_bucket_objects.go#L73-L112)

```release-note:none

```

### DIFF
--- a/mmv1/third_party/terraform/transport/transport.go
+++ b/mmv1/third_party/terraform/transport/transport.go
@@ -191,15 +191,15 @@ func IsApiNotEnabledError(err error) bool {
 }
 
 type GetPaginatedItemsOptions struct {
-	ResourceData         *schema.ResourceData
-	Config               *Config
-	BillingProject       *string
-	UserAgent            string
-	URL                  string
-	ListFlattener        func(config *Config, res []map[string]interface{}) ([]map[string]interface{}, error)
-	Params               map[string]string
-	ResourceToList       string
-	ErrorRetryPredicates []RetryErrorPredicateFunc
+	ResourceData         *schema.ResourceData                                                                 // Data for the resource that is being paginated
+	Config               *Config                                                                              // Config for the provider
+	BillingProject       *string                                                                              // Billing project for the request
+	UserAgent            string                                                                               // User agent for the request
+	URL                  string                                                                               // URL for the request, e.g. "https://www.googleapis.com/compute/v1/projects/my-project/zones/us-central1-a/instances"
+	ListFlattener        func(config *Config, res []map[string]interface{}) ([]map[string]interface{}, error) // Function to flatten the list of items
+	Params               map[string]string                                                                    // Parameters for the request
+	ResourceToList       string                                                                               // Resource to list, e.g. "items"
+	ErrorRetryPredicates []RetryErrorPredicateFunc                                                            // Predicates for retrying the request
 }
 
 func GetPaginatedItems(paginationOptions GetPaginatedItemsOptions) ([]map[string]interface{}, error) {


kautikdk:
Hi @BBBmau, I believe I may not be the best person to review this PR. As a Storage Service POC, I can review code changes which are specific to storage** services but this change appears to be wider than that so you might want re-roll reviewer assignment. Thanks!

BBBmau:
for some reason the original PR was closed, had to make a new one. I addressed some of the comments you left from the previous PR let me know your thoughts. I'm specifically wondering about the handling of certain behaviors and leaving the user to handle them @rileykarson

mmv1/third_party/terraform/transport/transport.go
```diff
@@ -189,3 +189,140 @@ func IsApiNotEnabledError(err error) bool {
 	}
 	return false
 }
+
+type GetPaginatedItemsSliceOptions struct {
```
rileykarson: nit: Can you document the fields on this, given it's a public function's interface?

mmv1/third_party/terraform/transport/transport.go
```diff
@@ -189,3 +189,140 @@ func IsApiNotEnabledError(err error) bool {
...
+		}
+
+		headers := make(http.Header)
+		opts := SendRequestOptions{
+			Config:               paginationOptions.Config,
+			Method:               "GET",
+			RawURL:               url,
+			UserAgent:            paginationOptions.UserAgent,
+			Headers:              headers,
+			ErrorRetryPredicates: []RetryErrorPredicateFunc{Is429RetryableQuotaError},
```
rileykarson: Is it intentional that we're newly retrying this in some cases?

mmv1/third_party/terraform/transport/transport.go
```diff
@@ -189,3 +189,140 @@ func IsApiNotEnabledError(err error) bool {
...
+	Config         *Config
+	BillingProject *string
+	UserAgent      string
+	URL            string
+	ListFlattener  func(config *Config, res interface{}) ([]interface{}, error)
+	Params         map[string]string
+	ResourceToList string
+}
+
+func GetPaginatedItemsSlice(paginationOptions GetPaginatedItemsSliceOptions) ([]interface{}, error) {
```
rileykarson: As far as I can tell this just returns a weaker type than GetPaginatedItemsMapOptions in all cases. Could we just strengthen the type in the callers? (We also appear to be weakening the type in AlloyDB locations right now)

mmv1/third_party/terraform/transport/transport.go
BBBmau: we actually need to treat this as an input where the default value is `Is429RetryableQuotaError`

found that `sql_databases` requires a different `ErrorRetryPredicate`:
https://github.com/GoogleCloudPlatform/magic-modules/blob/79e70b1c649fdb914d2968a6c28d7bb5a2a8a8e0/mmv1/third_party/terraform/services/sql/data_source_sql_databases.go#L59

mmv1/third_party/terraform/transport/transport.go
BBBmau: just so i understand, it's a weaker type due to the new structs not containing all the fields that we see in `SendOptions` struct correct? Seeing as it's occurring in alloydb locations like you said i imagine that's what you're referring to?

https://github.com/GoogleCloudPlatform/magic-modules/pull/14687/files#diff-0702f7783f542336226fad855351f5878a27a47c3ab0a02d72ee0b91f5b54162L89-L96

rileykarson: No, I just mean that []interface{} is a weaker type than []map[string]interface{}. I think even when we type as []interface{} we're always returning a group of `map[string]interface{}`'s? Where we could probably just convert all the callers to use the stronger type if they care about the type at all.

(Remember that type conversion can be a bit silly here- []interface{} and []map[string]interface{} are incompatible, but an array of map[string]interface{} could be represented with either one- and if we feed them to a (de)serialization function that accepts interface{} we'll produce the same result regardless of which type our slice nominally holds), w/ d.Get and d.Set behaving like (de)serialization

mmv1/third_party/terraform/transport/transport.go
rileykarson: This appears to add a new retry condition for data_source_alloydb_locations.go which previously had none; plumbing through settings from the callsite makes sense. If we're changing the default I'd want to see a breakdown of why it's appropriate (i.e. proving every user already had it or a case for why it's safe/a good idea).

(My list is inexhaustive- I saw one case where we added a new setting and jumped up to ask about it)

mmv1/third_party/terraform/transport/transport.go
BBBmau: went with the stronger type in the latest commits, we now just rely on just a singular `GetPaginatedItems` that returns `[]map[string]interface{}`

mmv1/third_party/terraform/transport/transport.go
BBBmau: Went with a default value of `nil` since from looking at implementations of list operations that are existing in the repo there are only a handful of data-sources that utilize a `ErrorRetryPredicate`

BBBmau:
pinging for another review @rileykarson

PR #14688, opened 2025-07-30
tgc-revival: add google_pubsub_subscription

zli82016:

```release-note:none

```


mmv1/products/pubsub/Subscription.yaml
```diff
@@ -55,6 +57,8 @@ examples:
     vars:
       topic_name: 'example-topic'
       subscription_name: 'example-subscription'
+    tgc_test_ignore_extra:
+      - enable_message_ordering # ignore its false value in configuration
```
trodge:
```suggestion
      - enable_message_ordering  # ignore its false value in configuration
```

mmv1/templates/tgc_next/custom_expand/pubsub_subscription_attributes.go.tmpl
```diff
@@ -0,0 +1,17 @@
...
+	}
+	m := make(map[string]string)
+
+	// The default value is present in CAI asset
+	if len(m) == 0 {
+		m["x-goog-version"] = "v1"
+		return m, nil
+	}
+
+	for k, val := range v.(map[string]interface{}) {
```
trodge:
```suggestion
	vMap, ok := v.(map[string]interface{})
	if !ok {
		return m, fmt.Errorf("non-map v: %v (%T)", v, v)
	}
	for k, val := range vMap {
```

mmv1/templates/tgc_next/custom_expand/pubsub_subscription_attributes.go.tmpl
```diff
@@ -0,0 +1,17 @@
+func expand{{$.GetPrefix}}{{$.TitlelizeProperty}}(v interface{}, d tpgresource.TerraformResourceData, config *transport_tpg.Config) (interface{}, error) {
+	if v == nil {
+		return map[string]string{}, nil
+	}
+	m := make(map[string]string)
```
trodge:
```suggestion
	m := make(map[string]string)
	if v == nil {
		return m, nil
	}
```

mmv1/templates/tgc_next/custom_expand/pubsub_subscription_attributes.go.tmpl
```diff
@@ -0,0 +1,17 @@
...
+
+	// The default value is present in CAI asset
+	if len(m) == 0 {
+		m["x-goog-version"] = "v1"
+		return m, nil
+	}
```
trodge: `m` will always be an empty map here, did you mean to check if `v` is empty?

trodge (changes requested):

mmv1/templates/tgc_next/custom_expand/pubsub_subscription_attributes.go.tmpl
zli82016: Good catch. I would like to check if `v` is empty or not.

PR #14701, opened 2025-07-31
Removes unnecessary wrapping in the documentation for google_firebaserules_release

rainshen49:
Fixes https://github.com/hashicorp/terraform-provider-google/issues/23661

```release-note: none
```


rainshen49:
Uhh this change is not as simple as I imagined. The underlying test uses the legacy tpgtools system, which is very hard to change. Since this is not a usage-impacting issue, I'll close this PR and keep the issue open until such time when we migrate it off of tpgtools.

PR #14709, opened 2025-07-31
feat: (vertexai) add rag_engine_config resource support

darshanmehta17:
Fixes https://github.com/hashicorp/terraform-provider-google/issues/23824

```release-note:new-resource
`google_vertexai_rag_engine_config`
```

### DIFF
--- a/mmv1/third_party/terraform/services/vertexai/resource_vertex_ai_rag_engine_config_test.go
+++ b/mmv1/third_party/terraform/services/vertexai/resource_vertex_ai_rag_engine_config_test.go
@@ -1,0 +1,37 @@
+package vertexai_test
+
+import (
+	"testing"
+
+	"github.com/hashicorp/terraform-plugin-testing/helper/resource"
+	"github.com/hashicorp/terraform-provider-google/google/acctest"
+	"github.com/hashicorp/terraform-provider-google/google/envvar"
+)
+
+func TestAccVertexAIRagEngineConfig_basic(t *testing.T) {
+	t.Parallel()
+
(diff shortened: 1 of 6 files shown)


mmv1/third_party/terraform/services/vertexai/resource_vertex_ai_rag_engine_config_test.go
```diff
@@ -0,0 +1,37 @@
...
+func TestAccVertexAIRagEngineConfig_basic(t *testing.T) {
+	t.Parallel()
+
+	context := map[string]interface{}{
+		"project": envvar.GetTestProjectFromEnv(),
+	}
+	acctest.VcrTest(t, resource.TestCase{
+		PreCheck:                 func() { acctest.AccTestPreCheck(t) },
+		ProtoV5ProviderFactories: acctest.ProtoV5ProviderFactories(t),
+		Steps: []resource.TestStep{
```
SirGitsalot, line 20: Everything is otherwise LGTM, but could you add some update steps? In particular, it'd be good to see an update step that goes to `unprovisioned` so that the custom delete is getting exercised regularly.

SirGitsalot (changes requested):

mmv1/third_party/terraform/services/vertexai/resource_vertex_ai_rag_engine_config_test.go
darshanmehta17: Added the test.

SirGitsalot (approved):

PR #14711, opened 2025-07-31
Switched to API field name as primary metadata

melinath:
This allows us to preserve the exact API field name in metadata while still being able to programmatically derive the terraform field name. Since conversion to terraform name is lossy, we can't reliably go the other direction.

This PR only handles generated meta.yaml files. If this approach works, I'll handle the handwritten files separately later.

```release-note:none

```


melinath:
> I think we may want to update docs/content/reference/metadata.md, at least slightly. api_field will now be camelcased, and I assume field will default to the snakecased version of api_field?

Done. Yes, those are the changes.

> > Since conversion to terraform name is lossy, we can't reliably go the other direction
> 
> Could you clarify, or possibly have an example to share?

Conversion to terraform name uses this function: https://github.com/GoogleCloudPlatform/magic-modules/blob/4e1621cca1333af17148fcba6bf6c0d7fba147f5/mmv1/google/string_utils.go#L27-L34

It looks like we don't have any fields that contain `.` or `-` in their API names, so the fact those get collapsed doesn't matter (unless fields like that are added in the future).

However, we can't tell the difference between fields like `IPAddress` vs `IpAddress` - both would be converted to `ip_address`. Examples: InterconnectAttachment has `cloudRouterIpAddress`, but RegionBackendService has `fastIPMove`.

roaks3 (approved):
> However, we can't tell the difference between fields like IPAddress vs IpAddress - both would be converted to ip_address. Examples: InterconnectAttachment has cloudRouterIpAddress, but RegionBackendService has fastIPMove.

Ah right, the acronym handling makes it lossy. Thanks for explaining!

PR #14722, opened 2025-08-02
tgc-revival: support google_certificate_manager_certificate

trodge:

```release-note:none

```

### DIFF
--- a/mmv1/templates/tgc_next/cai2hcl/flatten_property_method_tgc.go.tmpl
+++ b/mmv1/templates/tgc_next/cai2hcl/flatten_property_method_tgc.go.tmpl
@@ -26,10 +26,6 @@ func flatten{{$.GetPrefix}}{{$.TitlelizeProperty}}(v interface{}, d *schema.Reso
 	}
 	return tpgresource.GetResourceNameFromSelfLink(v.(string))
 }
-	{{- else if $.IgnoreRead }}
-func flatten{{$.GetPrefix}}{{$.TitlelizeProperty}}(v interface{}, _ *schema.ResourceData, _ *transport_tpg.Config) interface{} {
-	return v
-}
 	{{- else }}
 {{ template "flattenPropertyMethod" $ -}}
 	{{- end }}
--- a/mmv1/templates/tgc_next/cai2hcl/flatten_property_method_tgc.go.tmpl
+++ b/mmv1/templates/tgc_next/cai2hcl/flatten_property_method_tgc.go.tmpl
@@ -26,6 +26,10 @@ func flatten{{$.GetPrefix}}{{$.TitlelizeProperty}}(v interface{}, d *schema.Reso
 	}
 	return tpgresource.GetResourceNameFromSelfLink(v.(string))
 }
+	{{- else if $.IgnoreRead }}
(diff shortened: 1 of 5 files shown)


mmv1/templates/tgc_next/cai2hcl/flatten_property_method_tgc.go.tmpl
```diff
@@ -25,8 +25,12 @@ func flatten{{$.GetPrefix}}{{$.TitlelizeProperty}}(v interface{}, d *schema.Reso
 		return v
 	}
 	return tpgresource.GetResourceNameFromSelfLink(v.(string))
+}
+	{{- else if $.IgnoreRead }}
+func flatten{{$.GetPrefix}}{{$.TitlelizeProperty}}(v interface{}, _ *schema.ResourceData, _ *transport_tpg.Config) interface{} {
+	return v
```
zli82016: I found the reason for the error `error unmarshaling JSON as cty.Value: missing expected ["`.

The handling for fields with `IgnoreRead` should be added to mmv1/templates/terraform/flatten_property_method.go.tmpl, instead of this file. 

In  the[ line 21 ](https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/templates/terraform/flatten_property_method.go.tmpl#L21)of that file, we can add the check that the compiler is not tgc. If the compiler is tgc, the field is treated as the normal field.

For `self_managed` field, its flattener should return an array of objects with flattened fields, instead of an object, similar to [flattenCertificateManagerCertificateManaged](https://github.com/modular-magician/terraform-google-conversion/compare/1a9091231e1fe9aed0fbdf5c9be15fa8f8affcbe..72c141b990566cd9cae29c52594e1d32dfc48f85#diff-bb3f8f1563642002dc23619c0b0e0c98a26b1d8b27e4f509785a2a6336bad66eR111)

Feel free to let me know if you have different ideas.

trodge: I think this still will not work in this case because the field is in `res` and not in `res["selfManaged"]`.

I'm going to try having the decoder call `d.Set` so that the normal flattener picks up the correct value from the resource data.

zli82016: I tested it locally. It is working with the decoder.

trodge: It looks like I was mistaken about `pemCertificate` at the root level of the cai asset data indicating that the certificate is self managed. It also shows up there for google-managed certificates.

zli82016: I see. In this case,  `is_missing_in_cai: true` could be used for the field `self_managed` as it is missing in CAI asset. And the decoder is not needed.

trodge: `selfManaged` is an empty object in the CAI asset, so I think we can use that to determine when `pemCertificate` should go under it. All the tests are passing locally now.

PR #14725, opened 2025-08-02
container: add support for transparent hugepage config

DrFaust92:

```release-note:enhancement
container: add support for `linux_node_config.transparent_hugepage_defrag` and `linux_node_config.transparent_hugepage_enabled`
```

### DIFF
--- a/mmv1/third_party/terraform/services/container/node_config.go.tmpl
+++ b/mmv1/third_party/terraform/services/container/node_config.go.tmpl
@@ -719,12 +719,12 @@ func schemaNodeConfig() *schema.Schema {
 							"transparent_hugepage_defrag": {
                             	Type:        schema.TypeString,
                             	Optional:    true,
-                            	Description: `Defines the maximum age an image can be unused before it is garbage collected.`,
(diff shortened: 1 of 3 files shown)


hao-nan-li (commented):
I'm not sure why VCR marked this test as passed and failed due to randomness. Could you confirm this test always pass locally?

mmv1/third_party/terraform/services/container/node_config.go.tmpl
```diff
@@ -716,6 +716,16 @@ func schemaNodeConfig() *schema.Schema {
 									},
 								},
 							},
+							"transparent_hugepage_defrag": {
+                            	Type:        schema.TypeString,
```
hao-nan-li, line 720: Looking at the doc, looks like these fields have default values. Could you add them in? Also is `enum` a possible better representation of these fields?

DrFaust92:
> I'm not sure why VCR marked this test as passed and failed due to randomness. Could you confirm this test always pass locally? Actually that test seems unrelated to your change..

for me locally it works fine

mmv1/third_party/terraform/services/container/node_config.go.tmpl
DrFaust92: added

DrFaust92:
@hao-nan-li , added and retested.

PR #14728, opened 2025-08-04
Secure Source Manager (SSM): make CA pool optional for private instances

LucaPrete:
SSM: makes CA pool optional for private instances that use Google-managed trusted certificates.

```release-note:enhancement
secure-source-manager: made `caPool` argument optional.
```

### DIFF
--- a/mmv1/products/securesourcemanager/Instance.yaml
+++ b/mmv1/products/securesourcemanager/Instance.yaml
@@ -80,6 +80,13 @@ examples:
       'deletion_policy': '"DELETE"'
     ignore_read_extra:
       - 'update_time'
+  - name: 'secure_source_manager_instance_private_trusted_cert'
+    primary_resource_id: 'default'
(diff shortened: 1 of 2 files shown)


LucaPrete:
@melinath this should now be ready from my side whenever you're ready for the review. Thank you!

LucaPrete:
[apologies, please WAIT to merge this because there was a rollback at API level]

mmv1/products/securesourcemanager/Instance.yaml
```diff
@@ -80,6 +80,13 @@ examples:
       'deletion_policy': '"DELETE"'
     ignore_read_extra:
       - 'update_time'
+  - name: 'secure_source_manager_instance_private_trusted_cert'
```
melinath, line 83: There are already a lot of examples here, and this one doesn't seem to add much value. Would it be possible to update a different example to not set this field? Otherwise, you could set `exclude_docs` to only generate a test.

melinath (changes requested):

melinath:
superseded

PR #14734, opened 2025-08-04
Set up action to override breaking changes failures

melinath:

```release-note:none

```

### DIFF
--- a/.github/workflows/override-labels.yml
+++ b/.github/workflows/override-labels.yml
@@ -1,0 +1,28 @@
+name: override-labels
+
+permissions:
+  issues: write
+
+on:
+  pull_request:
+    types: [labeled, unlabeled]
+
+
+jobs:
+  override-breaking-change:
+    if: github.event.label.name == 'override-breaking-change'
+    env:
+      GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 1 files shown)


melinath:
~I can't actually seem test this workflow at the moment because of needing to merge it first, but I think it should work!~ (I can totally test this without merging, my bad.) It looks like we're already running generation + breaking change tests even though this PR doesn't modify any go code.

.github/workflows/override-labels.yml
```diff
@@ -0,0 +1,30 @@
+name: override-labels
+
+permissions: read-all
+
+on:
+  pull_request:
```
ScottSuarez: This will need to be pull_request_target before you can check it in

melinath: yep

melinath: it's just this way for testing.

melinath: done

melinath:
This doesn't seem to want to work due to some kind of permission issue - I'd like to try merging it and see if that helps.

ScottSuarez:
> This doesn't seem to want to work due to some kind of permission issue - I'd like to try merging it and see if that helps.

Stephen, make this change on a fork of GoogleCloudPlatform/magic-modules instead of your private repo. That should fix it

melinath:
sure, can do.

PR #14739, opened 2025-08-04
Fixed GH token env var name

melinath:

```release-note:none

```


melinath:
context: https://docs.github.com/en/actions/tutorials/authenticate-with-github_token#example-1-passing-the-github_token-as-an-input

PR #14750, opened 2025-08-05
Add enable_advanced_cluster example to vmware admin cluster

chethangowda89:

```release-note:REPLACEME

```

### DIFF
--- a/mmv1/templates/terraform/examples/gkeonprem_vmware_admin_cluster_metallb.tf.tmpl
+++ b/mmv1/templates/terraform/examples/gkeonprem_vmware_admin_cluster_metallb.tf.tmpl
@@ -3,8 +3,9 @@ resource "google_gkeonprem_vmware_admin_cluster" "{{$.PrimaryResourceId}}" {
   location = "us-west1"
   description = "test admin cluster"
   bootstrap_cluster_membership = "projects/870316890899/locations/global/memberships/gkeonprem-terraform-test"
-  on_prem_version = "1.31.0-gke.35"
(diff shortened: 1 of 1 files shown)


mmv1/products/gkeonprem/VmwareAdminCluster.yaml
```diff
@@ -688,7 +688,6 @@ properties:
   - type: Boolean
     name: enableAdvancedCluster
     description: If set, the advanced cluster feature is enabled.
-    output: true
```
c2thorn, line 691: is this able to be set by the API at all? Since it was set to output before I'm inclined to believe that we should set 
`default_from_api: true`

c2thorn (changes requested):

mmv1/products/gkeonprem/VmwareAdminCluster.yaml
chethangowda89: Yes we can set this from the api. This is was added in the previous release https://github.com/GoogleCloudPlatform/magic-modules/commit/9e4bd2c7de6bfa52368727cc90e9c8c0a61edeb4. At that time only user cluster was supporting this configuration, now we support for admin cluster too. Hence removing `output: true`

c2thorn: Got it, then yes `default_from_api: true` is appropriate here

chethangowda89: Done. Setting `default_from_api: true`

PR #14751, opened 2025-08-05
Remove the default value of view.use_legacy_sql in google_bigquery_table

wj-chen:
Context of the breaking change: https://github.com/GoogleCloudPlatform/magic-modules/pull/12390#discussion_r1941991099
Documentation update in `main`: https://github.com/GoogleCloudPlatform/magic-modules/pull/14574

The API accepts three values: `null`, `true`, `false`, and treats `null` and `false` as separate values. I tested in TF before to confirm they are different since the input of not setting vs setting to false comes back from the API response as `null` vs `false`.

For a user without it in their config, after upgrading to this version, they will see `true -> (none)`.

This change is required for an upcoming feature where this field can't be set, so TF users of that feature need to leave it out in their config, and also there can't be a TF default value when they leave it unset.

```release-note:breaking-change
bigquery: removed the default value of `view.use_legacy_sql` in `google_bigquery_table`
```

### DIFF
--- a/mmv1/third_party/terraform/services/bigquery/resource_bigquery_table.go.tmpl
+++ b/mmv1/third_party/terraform/services/bigquery/resource_bigquery_table.go.tmpl
@@ -1051,13 +1051,12 @@ func ResourceBigQueryTable() *schema.Resource {
 						},
 
 						// UseLegacySQL: [Optional] Specifies whether to use BigQuery's
-						// legacy SQL for this view. The default value is true. If set to
-						// false, the view will use BigQuery's standard SQL:
+						// legacy SQL for this view. If set to false, the view will use
+						// BigQuery's standard SQL:
 						"use_legacy_sql": {
 							Type:        schema.TypeBool,
 							Optional:    true,
-							Default:     true,
-							Description: `Specifies whether to use BigQuery's legacy SQL for this view. The default value is true. If set to false, the view will use BigQuery's standard SQL`,
+							Description: `Specifies whether to use BigQuery's legacy SQL for this view. If set to false, the view will use BigQuery's standard SQL`,
 						},
 					},
 				},
--- a/mmv1/third_party/terraform/website/docs/guides/version_7_upgrade.html.markdown
+++ b/mmv1/third_party/terraform/website/docs/guides/version_7_upgrade.html.markdown
@@ -102,6 +102,12 @@ Description of the change and how users should adjust their configuration (if ne
 
 ## Resources
 
+## Resource: `google_bigquery_table`
+
+### `view.use_legacy_sql` no longer has a default value of `True`
+
+If `view.use_legacy_sql` is not specified in the configuration, no value is sent to the API.
+
 ## Resource: `google_bigtable_table_iam_binding`
 
 ### `instance` is now removed
--- a/mmv1/third_party/terraform/website/docs/r/bigquery_table.html.markdown
+++ b/mmv1/third_party/terraform/website/docs/r/bigquery_table.html.markdown
@@ -425,7 +425,7 @@ The following arguments are supported:
 * `query` - (Required) A query that BigQuery executes when the view is referenced.
 
 * `use_legacy_sql` - (Optional) Specifies whether to use BigQuery's legacy SQL for this view.
-    The default value is true. If set to false, the view will use BigQuery's standard SQL.
+    If set to false, the view will use BigQuery's standard SQL.
     -> **Note**: Starting in provider version `7.0.0`, no default value is
     provided for this field unless explicitly set in the configuration.
 


wj-chen:
Hi @BBBmau , could you take a look at this change and let me know if there are any questions or concerns? Thanks.

wj-chen:
> Hi @BBBmau , could you take a look at this change and let me know if there are any questions or concerns? Thanks.

@BBBmau  I'm hoping to get this merged by the next major release cutoff which is middle of this week. Could you prioritize this review or reassign to someone who could? Thank you.

ScottSuarez:

mmv1/third_party/terraform/website/docs/guides/version_7_upgrade.html.markdown
```diff
@@ -131,6 +131,12 @@ Use `google_beyondcorp_security_gateway_application` instead.
 
 `public_repository` fields have had their default values removed. If your state has been reliant on them, they will need to be manually included into your configuration now.
 
+## Resource: `google_bigquery_table`
+
+### `view.use_legacy_sql` no longer has a default value of `True`
+
+If `view.use_legacy_sql` is not specified in the configuration, no value is sent to the API.
```
ScottSuarez: Can this be message be more customer centric. What will this mean for me if I have resource with this field. Presumably a sequential apply will unset this field on API side. Will this have any material affect on me as a developer.

If so, what will that be and what do I need to do to mitigate?

If not, what is the impact? There will still be a plan time diff on my resource

wj-chen: Thank you for the feedback. I've added more details to the version upgrade notes to describe this change.

mmv1/third_party/terraform/website/docs/guides/version_7_upgrade.html.markdown
```diff
@@ -131,6 +131,12 @@ Use `google_beyondcorp_security_gateway_application` instead.
 
 `public_repository` fields have had their default values removed. If your state has been reliant on them, they will need to be manually included into your configuration now.
 
+## Resource: `google_bigquery_table`
+
+### `view.use_legacy_sql` no longer has a default value of `True`
+
+If `view.use_legacy_sql` is not specified in the configuration, no value is sent to the API. For configurations that were created without explicitly setting `view.use_legacy_sql` in previous versions, the planned diff report will contain `use_legacy_sql = true -> null`. To achieve the desired behavior, either set `view.use_legacy_sql = true` explicitly, or update `view.query` to a standard SQL query.
```
ScottSuarez: Thanks for this! The note clearly explains the "what" (the plan will show `true -> null`).

To make this even safer for our users, could we also explicitly state the "so what"? Like exactly what happens if they run `terraform apply` on that plan.


you mention in the issue the new tri-state value of this field. I am wondering the implication of null vs true vs false and how we are communicating that here and any implications.

wj-chen: Apologies for the additional review load, but I made a code change to make this field Optional+Computed now, which means there will be no plan diff for users with existing configs that rely on the old value. I updated the note here accordingly, let me know if it's clear.

mmv1/third_party/terraform/website/docs/guides/version_7_upgrade.html.markdown
```diff
@@ -131,6 +131,12 @@ Use `google_beyondcorp_security_gateway_application` instead.
 
 `public_repository` fields have had their default values removed. If your state has been reliant on them, they will need to be manually included into your configuration now.
 
+## Resource: `google_bigquery_table`
+
+### `view.use_legacy_sql` no longer has a default value of `True`
+
+The `view.use_legacy_sql` field no longer has a default value. Configurations that relied on the old default will show no diff in the plan, and there will be no change to existing views. For a new view, leaving this field unspecified in the configuration will result in the view being created with no `use_legacy_sql` value, which assumes the legacy SQL dialect for its query according to the [API documentation](https://cloud.google.com/bigquery/docs/reference/rest/v2/tables#ViewDefinition).
```
ScottSuarez: I am I understanding this correctly. If unset it is interpreted by the api as sort of a sort 'true'.

mmv1/third_party/terraform/website/docs/r/bigquery_table.html.markdown
```diff
@@ -425,7 +425,7 @@ The following arguments are supported:
 * `query` - (Required) A query that BigQuery executes when the view is referenced.
 
 * `use_legacy_sql` - (Optional) Specifies whether to use BigQuery's legacy SQL for this view.
-    The default value is true. If set to false, the view will use BigQuery's standard SQL.
+    If set to false, the view will use BigQuery's standard SQL.
```
ScottSuarez: Could we add if unset behavior or a link to the api docs as above

ScottSuarez (changes requested):
Thanks for your help here. I think we are pretty much there!

mmv1/third_party/terraform/website/docs/guides/version_7_upgrade.html.markdown
wj-chen: Yes, I updated some wording to hopefully make that clearer

mmv1/third_party/terraform/website/docs/r/bigquery_table.html.markdown
wj-chen: Yes, updated the description here.

mmv1/third_party/terraform/services/bigquery/resource_bigquery_table.go.tmpl
```diff
@@ -1051,13 +1052,13 @@ func ResourceBigQueryTable() *schema.Resource {
...
-						// legacy SQL for this view. The default value is true. If set to
-						// false, the view will use BigQuery's standard SQL:
+						// legacy SQL for this view. If set to false, the view will use
+						// BigQuery's standard SQL:
 						"use_legacy_sql": {
 							Type:        schema.TypeBool,
 							Optional:    true,
-							Default:     true,
-							Description: `Specifies whether to use BigQuery's legacy SQL for this view. The default value is true. If set to false, the view will use BigQuery's standard SQL`,
+							Computed:    true,
```
ScottSuarez, line 1060: I see we are using Computed: true
Is there a reason for this? This basically mean that the field is managed upstream by the API when unset.
Specifically it is tricky because there will be no way to unset this field if we go forward in this manner.
If this is expected/anticipated we are good to go.
I just want to double confirm because if you need tri state behavior (unset/false/true) this will be difficult to maintain when the user cannot unset the value

ScottSuarez: discussed offline, api behaves similarly

wj-chen: For future reference:

In my testing, if this field already has a value e.g. `true`, and a new Update request is sent without this field in the payload, the value afterwards remains `true`. This means the API treats absent-value-in-request as a no-op and not "unsetting". At the same time, there isn't really a way to "explicitly send a `null` in a Update request" to "unset" because the field itself is a boolean and only accepts `true`/`false`: https://cloud.google.com/bigquery/docs/reference/rest/v2/tables#ViewDefinition

So the summary of the behavior is, if a view is created without this field, this field remains absent in the resource, until it's set to have `true`/`false` for the first time, then the value stays `true`/`false` and can be explicitly changed to `false`/`true` but can't be "unset" to be absent again.

NickElliot:
hi @wj-chen was there a github or buganizer associated with this?

wj-chen:
> hi @wj-chen was there a github or buganizer associated with this?

It's part of b/416315686.

PR #14737, opened 2025-08-04
difftest logic moving to a different file

luckyswaminathan:
https://github.com/GoogleCloudPlatform/magic-modules/pull/14476/files as a follow up to this PR's comments, am separating diff testing logic into separate files, no new code written

**Release Note Template for Downstream PRs (will be copied)**

See [Write release notes](https://googlecloudplatform.github.io/magic-modules/contribute/release-notes/) for guidance.

```release-note:none
moving difftesting logic to a different file
```

PR #14763, opened 2025-08-06
diff flag nil fix 

luckyswaminathan:
https://github.com/GoogleCloudPlatform/magic-modules/pull/14737 had a small bug where one of the vars wasn't ported over so this would cause all tests to not work at head

```release-note:none
fixed diffFlag nil issue
```

### DIFF
--- a/mmv1/third_party/terraform/acctest/diff_utils.go
+++ b/mmv1/third_party/terraform/acctest/diff_utils.go
@@ -16,6 +16,8 @@ import (
 	"github.com/hashicorp/terraform-plugin-testing/helper/resource"
 )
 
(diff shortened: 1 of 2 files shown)


mmv1/third_party/terraform/acctest/diff_utils.go
```diff
@@ -16,6 +16,8 @@ import (
 	"github.com/hashicorp/terraform-plugin-testing/helper/resource"
 )
 
+const diffFlag = "[Diff]"
```
trodge: I would also rename this to "diffTag" instead of "diffFlag"

PR #14768, opened 2025-08-06
tgc-revival: add google_alloydb_cluster

zli82016:
If the property is type of bool and has `defaul_from_api: true`,
	 include empty value in CAI asset by default during tfplan2cai conversion.
	Use `exclude_false_in_cai: true` to override the default behavior
	when the default value on API side is true.

```release-note:none

```

### DIFF
--- a/mmv1/products/alloydb/Cluster.yaml
+++ b/mmv1/products/alloydb/Cluster.yaml
@@ -492,7 +492,6 @@ properties:
         description: |
           Whether automated backups are enabled.
         default_from_api: true
-        include_empty_value_in_cai: true
   - name: 'backupSource'
     type: NestedObject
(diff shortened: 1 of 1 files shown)


hao-nan-li (commented):
Might be a bit demanding, but can we split this PR into 2? 
One to add `google_alloydb_cluster` and another one to add `ExcludeFalseInCai`?

Mostly because the added property field caused a lot of changes in `bigquery` and in `compute`

zli82016:
Moved the logic for automatically checking if false value should be set in CAI assets to https://github.com/GoogleCloudPlatform/magic-modules/pull/14783

hao-nan-li (approved):

PR #14769, opened 2025-08-06
fix plural formatting

shuyama1:

```release-note:none

```


melinath (changes requested):
could you add unit tests for Plural?

PR #14770, opened 2025-08-06
DNM: `google_monitoring_dashboard` conversion to fwprovider

NickElliot:
[TestAccMonitoringDashboard_update](https://paste.googleplex.com/4795374421934080) covers every update/import scenario

`mmv1/third_party/terraform/services/monitoring/fw_monitoring_utils.go` is actually fairly small, a majority of the lines are due to reimplementing the hashicorp official jsontypes library to add in the recursive computed attribute removal. the actual "new" code is lines 1-69, and 153-183. This reimplementation might actually be wholly unecessary after the `export` attribute, will try a conversion back to the official jsontypes library in the morning

monitoring dashboard itself is a fairly straight forward resource due to the fact it is just sending the single json field as the http request body, I highlighted the weird solution to the data inconsistency rules. otherwise it should look fairly standard.

I also implemented the default project plan modifier at `mmv1/third_party/terraform/fwresource/framework_common_plan_modifiers.go`, could quickly be modified for region/zone to maintain parity with sdkv2 resources. neither of the resources you converted would use it though

```release-note:none

```

### DIFF
--- a/mmv1/third_party/terraform/fwresource/framework_common_plan_modifiers.go
+++ b/mmv1/third_party/terraform/fwresource/framework_common_plan_modifiers.go
@@ -8,7 +8,7 @@ import (
 	"github.com/hashicorp/terraform-plugin-framework/types"
 )
 
-func DefaultProjectModify(ctx context.Context, req resource.ModifyPlanRequest, resp *resource.ModifyPlanResponse, providerConfigProject string){
+func DefaultProjectModify(ctx context.Context, req resource.ModifyPlanRequest, resp *resource.ModifyPlanResponse, providerConfigProject string) {
 	var old types.String
 	diags := req.State.GetAttribute(ctx, path.Root("project"), &old)
 	resp.Diagnostics.Append(diags...)


mmv1/third_party/terraform/services/monitoring/fw_resource_monitoring_dashboard.go
```diff
@@ -350,16 +398,14 @@ func (r *MonitoringDashboardResource) Refresh(ctx context.Context, data *Monitor
...
-	)
-	data.Id = types.StringValue(id)
-
 	str, err := structure.FlattenJsonToString(res)
 	if err != nil {
 		diags.AddError("Error converting Dashboard:", fmt.Sprintf("%s", err))
 		return
 	}
-	data.DashboardJson = jsontypes.NewNormalizedValue(str)
+	if data.DashboardJson.IsNull() || data.DashboardJson.IsUnknown() {
```
NickElliot, line 406: this is the solution to the import problem:

the refresh function will only set DashboardJson if Read is called before Create or Update, 

this then allows the plan modifier/semantic equality to work as expected for our "state =/= config" pov, because import uses `Read` as the source of truth for data consistency rather than the actual config.

c2thorn:
closing as we declined to move forward with this.

PR #14773, opened 2025-08-07
tgc-revival: add google_alloydb_instance

zli82016:
template: {{cluster}}/instances/{{instance_id}}
assetName: //alloydb.googleapis.com/projects/ci-test-project/locations/us-central1/clusters/tf-test-cluster/instances/tf-test-instance
Handled the case that the `cluster` field in converted HCL should be set to `projects/ci-test-project/locations/us-central1/clusters/tf-test-cluster`

```release-note:none

```

### DIFF
--- a/mmv1/third_party/tgc_next/pkg/cai2hcl/converters/utils/utils.go
+++ b/mmv1/third_party/tgc_next/pkg/cai2hcl/converters/utils/utils.go
@@ -30,10 +30,10 @@ template: //bigquery.googleapis.com/projects/{{project}}/datasets/{{dataset_id}}
 assetName: //bigquery.googleapis.com/projects/my-project/datasets/my-dataset
 hclData: [project:my-project dataset_id:my-dataset]
 
-It also handles mutliple-path fragements.
+It also handles multi-fragment fields.
 template: {{cluster}}/instances/{{instance_id}}
 assetName: //alloydb.googleapis.com/projects/ci-test-project/locations/us-central1/clusters/tf-test-cluster/instances/tf-test-instance
-{{cluster}} field covers projects/ci-test-project/locations/us-central1/clusters/tf-test-cluster
+hclData: [cluster:projects/ci-test-project/locations/us-central1/clusters/tf-test-cluster instance_id:tf-test-instance]
 */
 func ParseUrlParamValuesFromAssetName(assetName, template string, outputFields map[string]struct{}, hclData map[string]any) {
 	templateFragments := strings.Split(template, "/")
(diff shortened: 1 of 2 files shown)


mmv1/api/resource.go
```diff
@@ -552,9 +556,6 @@ func (r Resource) UserVirtualFields() []*Type {
 }
 
 func (r Resource) ServiceVersion() string {
-	if r.CaiBaseUrl != "" {
```
zli82016: I don't find the cases that CaiBaseUrl starting with a specific version and it shouldn't, based on https://cloud.google.com/asset-inventory/docs/asset-names

mmv1/third_party/tgc_next/pkg/cai2hcl/converters/utils/utils.go
```diff
@@ -28,26 +29,60 @@ func ParseFieldValue(url string, name string) string {
...
+				for {
+					if endAssetIx == len(assetFragments) || assetFragments[endAssetIx] == nextNonFieldFragment {
+						break
+					} else {
+						endAssetIx++
+					}
+				}
+			} else {
+				endAssetIx = len(assetFragments)
+			}
```
trodge: I'm finding this block of code a little hard to follow given that it has multiple loops nested inside a conditional inside a loop. Would it be possible to move some of the logic into a helper function?

trodge (changes requested):
Overall LGTM, just one question about readability.

mmv1/third_party/tgc_next/pkg/cai2hcl/converters/utils/utils.go
zli82016: Sure. Added a helper function `getEndAssetIx`.

Feel free to let me know if more helper functions are needed.

mmv1/third_party/tgc_next/pkg/cai2hcl/converters/utils/utils.go
```diff
@@ -28,31 +29,73 @@ func ParseFieldValue(url string, name string) string {
...
+	assetFragments := strings.Split(assetName, "/")
+
+	// Iterate through the fragments and match fields.
+	assetIx := 0
+	for templateIx := 0; templateIx < len(templateFragments); templateIx++ {
+		templateFragment := templateFragments[templateIx]
+
+		// Check if the template fragment is a field (e.g., {{project}})
+		if fieldName, isField := strings.CutPrefix(templateFragment, "{{"); isField {
+			fieldName, _ = strings.CutSuffix(fieldName, "}}")
```
trodge: I think this change would make the `}}` suffix optional. We should keep it required to avoid confusing situations where it might be omitted.

mmv1/third_party/tgc_next/pkg/cai2hcl/converters/utils/utils.go
```diff
@@ -28,31 +29,73 @@ func ParseFieldValue(url string, name string) string {
 template: //bigquery.googleapis.com/projects/{{project}}/datasets/{{dataset_id}}
 assetName: //bigquery.googleapis.com/projects/my-project/datasets/my-dataset
 hclData: [project:my-project dataset_id:my-dataset]
+
+It also handles mutliple-path fragements.
```
trodge:
```suggestion
It also handles multi-fragment fields.
```

mmv1/third_party/tgc_next/pkg/cai2hcl/converters/utils/utils.go
```diff
@@ -28,31 +29,73 @@ func ParseFieldValue(url string, name string) string {
...
+template: {{cluster}}/instances/{{instance_id}}
+assetName: //alloydb.googleapis.com/projects/ci-test-project/locations/us-central1/clusters/tf-test-cluster/instances/tf-test-instance
+{{cluster}} field covers projects/ci-test-project/locations/us-central1/clusters/tf-test-cluster
```
trodge:
```suggestion
hclData: [cluster:projects/ci-test-project/locations/us-central1/clusters/tf-test-cluster instance_id:tf-test-instance]
```

We might also want to replace this comment with a unit test, but not necessarily a blocker for this PR.

zli82016: Added the unit test

mmv1/third_party/tgc_next/pkg/cai2hcl/converters/utils/utils.go
zli82016: Added a check that if `}}` exists or not.

If we want to return an error for missing `}}`, the [resource converter template](https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/templates/tgc_next/cai2hcl/resource_converter.go.tmpl#L115)  also needs the modification. 

Maybe we can return the error and modify the template in a separate PR.

trodge (approved):

PR #14777, opened 2025-08-07
Fix for desired_auto_created_endpoints incorrect update

NA2047:
Fix for issue https://github.com/terraform-google-modules/terraform-google-memorystore/issues/314

```release-note:bug
memorystore:  `desired_auto_created_endpoints` field incorrectly updated when desired_psc_auto_connections should have been 
```

### DIFF
--- a/mmv1/templates/terraform/decoders/memorystore_instance.go.tmpl
+++ b/mmv1/templates/terraform/decoders/memorystore_instance.go.tmpl
@@ -85,10 +85,25 @@
 					}
 				}
 			}
+			// NOTE: Due to how the api updates the backend. When mode is set to CLUSTER_DISABLED
+			// When desired_auto_created_endpoints is set this flow works as expected
+			// FLOW: Terraform  -> API Input -> API Output -> Terraform output
+			// CLUSTER_DISABLED and desired_auto_created_endpoints defined: desired_auto_created_endpoints -> endpoints -> Endpoints -> desired_auto_created_endpoints + Endpoints
+			// CLUSTER_DISABLED and desired_psc_auto_connections defined: desired_psc_auto_connections -> psc_auto_connections -> Endpoints -> desired_auto_created_endpoints + Endpoints
+			// CLUSTER and desired_psc_auto_connections defined: desired_psc_auto_connections -> psc_auto_connections -> psc_auto_connections -> desired_psc_auto_connections + psc_auto_connection
+			// CLUSTER and desired_auto_created_endpoints defined: desired_auto_created_endpoints -> endpoints -> Endpoints -> desired_auto_created_endpoints + Endpoints
+			// The issue is that when CLUSTE_DISABLED is set for that flow  API is updated Endpoints instead of psc_auto_connections.
+			// The if else if below checks what is define by the user in order to set the correct fielt in the state
 			if len(transformed) > 0 {
-				d.Set("desired_auto_created_endpoints", transformed)
-				log.Printf("[DEBUG] Setting desired_auto_created_endpoints in decoder for %#v", transformed)
-
+				_, okEndpoint := d.GetOk("desired_auto_created_endpoints")
+				_, okPsc := d.GetOk("desired_psc_auto_connections")
+				if okEndpoint {
+					d.Set("desired_auto_created_endpoints", transformed)
+					log.Printf("[DEBUG] Setting desired_auto_created_endpoints in decoder within endpoints for %#v", transformed)
+				} else if okPsc {
+					d.Set("desired_psc_auto_connections", transformed)
+					log.Printf("[DEBUG] Setting desired_psc_auto_connections in decoder within endpoints for %#v", transformed)
+				}
 			}
 		}
 
--- a/mmv1/third_party/terraform/services/memorystore/resource_memorystore_instance_test.go
+++ b/mmv1/third_party/terraform/services/memorystore/resource_memorystore_instance_test.go
@@ -1537,3 +1537,73 @@ data "google_project" "project" {
 }
 `, context)
 }
+
+func TestAccMemorystoreInstance_memorystorePscAutoInstanceClusterDisabled(t *testing.T) {
+	t.Parallel()
+
+	context := map[string]interface{}{
+		"random_suffix": acctest.RandString(t, 10),
+		"location":      "us-central1",
+	}
+
+	acctest.VcrTest(t, resource.TestCase{
+		PreCheck:                 func() { acctest.AccTestPreCheck(t) },
+		ProtoV5ProviderFactories: acctest.ProtoV5ProviderFactories(t),
+		CheckDestroy:             testAccCheckMemorystoreInstanceDestroyProducer(t),
+		Steps: []resource.TestStep{
+			{
+				Config: testAccMemorystoreInstance_memorystorePscAutoInstanceClusterDisabled(context),
+			},
+			{
+				ResourceName:      "google_memorystore_instance.instance-cluster-disabled",
+				ImportState:       true,
+				ImportStateVerify: true,
+			},
+		},
+	})
+}
+
+func testAccMemorystoreInstance_memorystorePscAutoInstanceClusterDisabled(context map[string]interface{}) string {
+	return acctest.Nprintf(`
+resource "google_memorystore_instance" "instance-cluster-disabled" {
+  instance_id                  = "tf-test-instance-psc%{random_suffix}"
+  shard_count                  = 1
+  desired_psc_auto_connections {
+    network                    = google_compute_network.producer_net.id
+    project_id                 = data.google_project.project.project_id
+  }
+  location                     = "%{location}"
+  deletion_protection_enabled  = false
+  mode                         = "CLUSTER_DISABLED"
+  depends_on = [
+    google_network_connectivity_service_connection_policy.default
+  ]
+}
+
+resource "google_network_connectivity_service_connection_policy" "default" {
+  name                           = "tf-test-my-policy%{random_suffix}"
+  location                       = "%{location}"
+  service_class                  = "gcp-memorystore"
+  description                    = "my basic service connection policy"
+  network                        = google_compute_network.producer_net.id
+  psc_config {
+    subnetworks                  = [google_compute_subnetwork.producer_subnet.id]
+  }
+}
+
+resource "google_compute_subnetwork" "producer_subnet" {
+  name                           = "tf-test-my-subnet%{random_suffix}"
+  ip_cidr_range                  = "10.0.0.248/29"
+  region                         = "%{location}"
+  network                        = google_compute_network.producer_net.id
+}
+
+resource "google_compute_network" "producer_net" {
+  name                           = "tf-test-my-network%{random_suffix}"
+  auto_create_subnetworks        = false
+}
+
+data "google_project" "project" {
+}
+`, context)
+}


mmv1/products/memorystore/Instance.yaml
```diff
@@ -109,7 +109,7 @@ examples:
 virtual_fields:
   - name: 'desired_psc_auto_connections'
     description: "`desired_psc_auto_connections` is deprecated  Use `desired_auto_created_endpoints` instead."
-    deprecation_message: '`desired_psc_auto_connections` is deprecated  Use `desired_auto_created_endpoints` instead.'
+    deprecation_message: '`desired_psc_auto_connections` is deprecated  Use `desired_auto_created_endpoints` instead. Terrafrom import will work with desired_auto_created_endpoints only'
```
rileykarson:
```suggestion
    deprecation_message: '`desired_psc_auto_connections` is deprecated  Use `desired_auto_created_endpoints` instead. `terraform import` will only work with desired_auto_created_endpoints`.'
```

mmv1/products/memorystore/Instance.yaml
```diff
@@ -109,7 +109,7 @@ examples:
 virtual_fields:
   - name: 'desired_psc_auto_connections'
     description: "`desired_psc_auto_connections` is deprecated  Use `desired_auto_created_endpoints` instead."
-    deprecation_message: '`desired_psc_auto_connections` is deprecated  Use `desired_auto_created_endpoints` instead.'
+    deprecation_message: '`desired_psc_auto_connections` is deprecated  Use `desired_auto_created_endpoints` instead. `terraform import` will only work with desired_auto_created_endpoints`.'
```
melinath:
```suggestion
    deprecation_message: '`desired_psc_auto_connections` is deprecated. Use `desired_auto_created_endpoints` instead. `terraform import` will only work with desired_auto_created_endpoints`.'
```

mmv1/templates/terraform/decoders/memorystore_instance.go.tmpl
```diff
@@ -85,10 +85,31 @@
...
+				if okEndpoint {
+					d.Set("desired_auto_created_endpoints", transformed)
+					log.Printf("[DEBUG] Setting desired_auto_created_endpoints in decoder within endpoints for %#v", transformed)
+				} else if okPsc {
+					d.Set("desired_psc_auto_connections", transformed)
+					log.Printf("[DEBUG] Setting desired_psc_auto_connections in decoder within endpoints for %#v", transformed)
+				} else {
+					d.Set("desired_auto_created_endpoints", transformed)
+					log.Printf("[DEBUG] Setting desired_auto_created_endpoints in decoder within endpoints for %#v", transformed)
+				}
```
melinath: It looks like the generated code for this resource [doesn't actually properly set `conflicts`](https://github.com/hashicorp/terraform-provider-google-beta/blob/8a8630000e4627242d1e12786eeb7ef7230d5cb6/google-beta/services/memorystore/resource_memorystore_instance.go#L929) for these fields, so they can both be set at the same time. What would you think about treating them independently here?

```suggestion
				if okEndpoint {
					d.Set("desired_auto_created_endpoints", transformed)
					log.Printf("[DEBUG] Setting desired_auto_created_endpoints in decoder within endpoints for %#v", transformed)
				}
				if okPsc {
					d.Set("desired_psc_auto_connections", transformed)
					log.Printf("[DEBUG] Setting desired_psc_auto_connections in decoder within endpoints for %#v", transformed)
				}
				// Set preferred field on import
				if !okPsc && !okEndpoint {
					d.Set("desired_psc_auto_connections", transformed)
					log.Printf("[DEBUG] Setting desired_psc_auto_connections in decoder within endpoints for %#v", transformed)
				}
```

mmv1/templates/terraform/decoders/memorystore_instance.go.tmpl
```diff
@@ -85,10 +85,31 @@
...
+			// 2. CLUSTER_DISABLED and desired_psc_auto_connections defined: desired_psc_auto_connections -> psc_auto_connections -> Endpoints -> desired_auto_created_endpoints + Endpoints
+			// 3. CLUSTER and desired_psc_auto_connections defined: desired_psc_auto_connections -> psc_auto_connections -> psc_auto_connections -> desired_psc_auto_connections + psc_auto_connection
+			// 4. CLUSTER and desired_auto_created_endpoints defined: desired_auto_created_endpoints -> endpoints -> Endpoints -> desired_auto_created_endpoints + Endpoints
+			// The issue is that when CLUSTE_DISABLED is set for that flow  API is updated Endpoints instead of psc_auto_connections
+			// The if else if below checks what is define by the user in order to set the correct fielt in the state
+			// The code below fixes use case 2
+			// The else conditon is to handle importing as there is no clear way to tell what the user has set desired_auto_created_endpoints or desired_psc_auto_connections
+			// Default is to set endpoints and user will need to adjust thier Terraform config as to use desired_auto_created_endpoints
```
melinath: This comment is currently describing what's being fixed in this PR. Could you instead make it explain why the code is written the way it is? Something like: "we want to make these fields detect API-side drift, so if the API returns a value for them we set them in state."

mmv1/templates/terraform/decoders/memorystore_instance.go.tmpl
```diff
@@ -85,10 +85,31 @@
...
 			if len(transformed) > 0 {
```
melinath, line 90: If the API doesn't return a value, we don't detect drift? That seems wrong? If these are removed in the console we should recreate the resource.

melinath (changes requested):

mmv1/templates/terraform/decoders/memorystore_instance.go.tmpl
NA2047: The issue here is that the API when `mode = CLUSTER`  will populate both `endpoints` as well as `pscConnections` . This causes an issue for import depending on if the user has set `desired_psc_auto_connections` or  `desired_auto_created_endpoints` the deprecation message lets the user know that if you are using import you will need to set  `desired_auto_created_endpoints`.

I will add something to the description of `desired_auto_created_endpoints` as well

mmv1/templates/terraform/decoders/memorystore_instance.go.tmpl
NA2047: Yes Conflicts with for virtual fields does not work currently. The API will not let you set both at the same time. So this shouldnt be an issue.

NA2047: But I feel we should leave them as separate checks

mmv1/templates/terraform/decoders/memorystore_instance.go.tmpl
melinath: right - that all makes sense, I think, but I'm not talking about Import here. The problem I'm trying to flag here is if `len(transformed) == 0` because of API drift - that is, the field was emptied using a tool other than Terraform. Currently, we will never update the state in that case, so we won't detect that drift on normal Reads.

We don't necessarily need to fix this in this PR since it's a separate issue, though.

NA2047: Ya thats a good Idea, I can come back to this in another PR. Want to get this out to customers

mmv1/templates/terraform/decoders/memorystore_instance.go.tmpl
melinath: It looks like the encoder for this resource will ensure we always only send one field to the API even if we set both on the resource, so I think we should make this change in this PR just to be safe. https://github.com/hashicorp/terraform-provider-google-beta/blob/aa1290e90e3a134192da51af01626eeba9ba1c69/google-beta/services/memorystore/resource_memorystore_instance.go#L3210

mmv1/templates/terraform/decoders/memorystore_instance.go.tmpl
```diff
@@ -85,10 +85,33 @@
...
+					d.Set("desired_auto_created_endpoints", transformed)
+					log.Printf("[DEBUG] Setting desired_auto_created_endpoints in decoder within endpoints for %#v", transformed)
+				} else if okPsc {
+					d.Set("desired_psc_auto_connections", transformed)
+					log.Printf("[DEBUG] Setting desired_psc_auto_connections in decoder within endpoints for %#v", transformed)
+			    // Set preferred field on import
+				} else {
+					d.Set("desired_auto_created_endpoints", transformed)
+					log.Printf("[DEBUG] Setting desired_auto_created_endpoints in decoder within endpoints for %#v", transformed)
+				}
```
melinath:
```suggestion
				if okEndpoint {
					d.Set("desired_auto_created_endpoints", transformed)
					log.Printf("[DEBUG] Setting desired_auto_created_endpoints in decoder within endpoints for %#v", transformed)
				}
				if okPsc {
					d.Set("desired_psc_auto_connections", transformed)
					log.Printf("[DEBUG] Setting desired_psc_auto_connections in decoder within endpoints for %#v", transformed)
				}
				// Set preferred field on import
				if !okPsc && !okEndpoint {
					d.Set("desired_auto_created_endpoints", transformed)
					log.Printf("[DEBUG] Setting desired_auto_created_endpoints in decoder within endpoints for %#v", transformed)
				}
```

mmv1/templates/terraform/decoders/memorystore_instance.go.tmpl
```diff
@@ -85,10 +85,33 @@
...
+			// 2. CLUSTER_DISABLED and desired_psc_auto_connections defined: desired_psc_auto_connections -> psc_auto_connections -> Endpoints -> desired_auto_created_endpoints + Endpoints
+			// 3. CLUSTER and desired_psc_auto_connections defined: desired_psc_auto_connections -> psc_auto_connections -> psc_auto_connections -> desired_psc_auto_connections + psc_auto_connection
+			// 4. CLUSTER and desired_auto_created_endpoints defined: desired_auto_created_endpoints -> endpoints -> Endpoints -> desired_auto_created_endpoints + Endpoints
+			// The issue is that when CLUSTE_DISABLED is set for that flow  API is updated Endpoints instead of psc_auto_connections.
+			// The if else if below checks what is define by the user in order to set the correct field in the state
+			// The code below fixes use case 2
+			// The else conditon is to handle importing as there is no clear way to tell what the user has set desired_auto_created_endpoints or desired_psc_auto_connections
+			// Default is to set endpoints and user will need to adjust thier Terraform config as to use desired_auto_created_endpoints
```
melinath:
```suggestion
			// We want to make these fields detect API-side drift, so if the API returns a value for them and they're set in config, we set them in state.
			// On import, we only set `desired_auto_created_endpoints` because that's the non-deprecated field.
```

mmv1/templates/terraform/decoders/memorystore_instance.go.tmpl
```diff
@@ -85,10 +85,24 @@
...
+				_, okEndpoint := d.GetOk("desired_auto_created_endpoints")
+				_, okPsc := d.GetOk("desired_psc_auto_connections")
+				if okEndpoint {
+					d.Set("desired_auto_created_endpoints", transformed)
+					log.Printf("[DEBUG] Setting desired_auto_created_endpoints in decoder within endpoints for %#v", transformed)
+				}
+				if okPsc {
+					d.Set("desired_psc_auto_connections", transformed)
+					log.Printf("[DEBUG] Setting desired_psc_auto_connections in decoder within endpoints for %#v", transformed)
+				}
```
melinath: Found a slight issue here. If both fields were previously set, and then you remove one of them, it ends up causing a permadiff because we never unset the field in state. I think we can solve this by doing the following:

```suggestion
				if okEndpoint {
					d.Set("desired_auto_created_endpoints", transformed)
					log.Printf("[DEBUG] Setting desired_auto_created_endpoints in decoder within endpoints for %#v", transformed)
				} else if okPsc {
					d.Set("desired_auto_created_endpoints", []interface{}{})
				}
				if okPsc {
					d.Set("desired_psc_auto_connections", transformed)
					log.Printf("[DEBUG] Setting desired_psc_auto_connections in decoder within endpoints for %#v", transformed)
				} else if okEndpoint {
					d.Set("desired_psc_auto_connections", []interface{}{})
				}
```

mmv1/third_party/terraform/services/memorystore/resource_memorystore_instance_test.go
```diff
@@ -1537,3 +1537,74 @@ data "google_project" "project" {
 }
 `, context)
 }
+
+func TestAccMemorystoreInstance_memorystorePscAutoInstanceClusterDisabled(t *testing.T) {
```
melinath, line 1541: could you add a couple more test steps here so that the test flow is:

1. setting both
2. setting just one
3. setting just the other
4. setting neither

melinath (changes requested):

mmv1/third_party/terraform/services/memorystore/resource_memorystore_instance_test.go
```diff
@@ -1537,3 +1537,236 @@ data "google_project" "project" {
...
+			},
+			{
+				Config: testAccMemorystoreInstance_memorystorePscAutoInstanceClusterDisabled_bothConnections(context),
+			},
+			{
+				ResourceName:            "google_memorystore_instance.instance-cluster-disabled",
+				ImportState:             true,
+				ImportStateVerify:       true,
+				ImportStateVerifyIgnore: []string{"desired_auto_created_endpoints.#", "desired_auto_created_endpoints.0.%", "desired_auto_created_endpoints.0.project_id", "desired_auto_created_endpoints.0.network", "desired_psc_auto_connections.#", "desired_psc_auto_connections.0.%", "desired_psc_auto_connections.0.network", "desired_psc_auto_connections.0.project_id"},
+			},
```
melinath:
```suggestion
			{
				Config: testAccMemorystoreInstance_memorystorePscAutoInstanceClusterDisabled_bothConnections(context),
			},
			{
				ResourceName:            "google_memorystore_instance.instance-cluster-disabled",
				ImportState:             true,
				ImportStateVerify:       true,
				ImportStateVerifyIgnore: []string{"desired_auto_created_endpoints.#", "desired_auto_created_endpoints.0.%", "desired_auto_created_endpoints.0.project_id", "desired_auto_created_endpoints.0.network", "desired_psc_auto_connections.#", "desired_psc_auto_connections.0.%", "desired_psc_auto_connections.0.network", "desired_psc_auto_connections.0.project_id"},
			},
			{
				Config: testAccMemorystoreInstance_memorystorePscAutoInstanceClusterDisabledPscAutoConnections(context),
			},
			{
				ResourceName:            "google_memorystore_instance.instance-cluster-disabled",
				ImportState:             true,
				ImportStateVerify:       true,
				ImportStateVerifyIgnore: []string{"desired_auto_created_endpoints.#", "desired_auto_created_endpoints.0.%", "desired_auto_created_endpoints.0.project_id", "desired_auto_created_endpoints.0.network", "desired_psc_auto_connections.#", "desired_psc_auto_connections.0.%", "desired_psc_auto_connections.0.network", "desired_psc_auto_connections.0.project_id"},
			},
```

melinath (approved):
LGTM.

PR #14778, opened 2025-08-07
Enable custom endpoints for all DCL-based resources

melinath:
Fixes https://github.com/hashicorp/terraform-provider-google/issues/23913.

We accidentally broke custom endpoints for a number of DCL-based resources as part of trying to add universe domain support for them: https://github.com/GoogleCloudPlatform/magic-modules/pull/10400.

Additionally, we didn't correctly add universe domain support - it functions based off replacing the values in DefaultBasePaths, but those aren't populated for the impacted DCL-based resources because the DCL doesn't provide a way to easily get product base paths at generation time.

Given how few DCL-based resources still manage their custom endpoints via the DCL, we should just hard-code the remainder as part of MMv1 (like we do with some already) to ensure consistent behavior.

This PR makes that change, as well as removing special handling for Cloud Build Worker Pool and GKEHub Feature Membership (so that they instead use the same custom endpoint as other resources in the same service.) Separate endpoints have been a source of confusion (yaqs/8708287123125436416).

Per https://registry.terraform.io/providers/hashicorp/google/latest/docs/guides/provider_reference#{{service}}_custom_endpoint-1:

> Support for custom endpoints is on a best-effort basis. The underlying endpoint and default values for a resource can be changed at any time without being considered a breaking change.

yaqs/1855794208267304960

I was able to test this manually for google_apikeys_key by setting up a config with universe_domain set, disabling the "auth matches universe" check in source, and trying to create a google_apikeys_key.

On main, this accesses apikeys.googleapis.com regardless of custom endpoint or universe domain.

On this branch, custom domain takes precedence over universe domain, which takes precedence over default.

```release-note:none

```


melinath:
The AssuredWorkload tests are failing because the DCL resource incorrectly [uses the provided BasePath](https://github.com/GoogleCloudPlatform/declarative-resource-client-library/blob/6de5db3e6a3825b60daa157dd367093971749d3f/services/google/assuredworkloads/assuredworkloads_utils.go#L35-L41) to determine the URL for the folder (which worked fine when we weren't actually setting a proper BasePath but doesn't work now that we are). (This is probably why we left it with d.Get() previously, since that would have only broken if someone set a custom endpoint.)

I'll need to fix that & release separately before we can merge this.

I think the other failures should be unrelated

hao-nan-li (commented):
Just want to make sure, when the universe_domain is set to an appropriate value, did the host change to the correct universe?

melinath:
> Just want to make sure, when the universe_domain is set to an appropriate value, did the host change to the correct universe?

yes, for google_apikeys_key. I didn't test each product individually yet but I'd like to before merging. Obviously with the current bug in assured workloads that one wouldn't work properly.

melinath:
Please feel free to double-check me though!

c2thorn:
does this need to wait on cl/793805595 being pulled into the DCL and upgraded in MM?

melinath:
> does this need to wait on cl/793805595 being pulled into the DCL and upgraded in MM?

yes

hao-nan-li (approved):

c2thorn:
just released https://github.com/GoogleCloudPlatform/declarative-resource-client-library/releases/tag/v1.81.0 will make a separate PR to upgrade the go.mod

c2thorn:
https://github.com/GoogleCloudPlatform/magic-modules/pull/14852

melinath:
Verified with manual local testing that universedomain / custom endpoints works with:

- apikeys
- assuredworkloads
- firebaserules
- recaptcha_enterprise
- google_cloudbuild_worker_pool
- google_gke_hub_feature_membership

Cloud resource manager should work too, but it doesn't seem to be used anywhere. Leaving its removal for a separate cleanup.

melinath:
- TestAccAssuredWorkloadsWorkload_SovereignControlsWorkload is a flakey test that fails like this in nightlies.
- other failures are unrelated

PR #14781, opened 2025-08-07
Set SecureSourceManager Instance and Repository deletion_policy default to PREVENT

ericpang777:

```release-note:breaking-change
securesourcemanager: changed `deletion_policy` default value from `DELETE` to `PREVENT`
```

### DIFF
--- a/mmv1/products/securesourcemanager/Instance.yaml
+++ b/mmv1/products/securesourcemanager/Instance.yaml
@@ -177,7 +177,7 @@ virtual_fields:
         * DELETE
         * PREVENT
         * ABANDON
-    default_value: 'DELETE'
+    default_value: 'PREVENT'
 properties:
   - name: 'name'
     type: String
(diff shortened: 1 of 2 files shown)


ScottSuarez (changes requested):
Hello, can you also update our tests so that they will delete. Looks like all the tests above are leaving persistent infrastructure around after running causing the failure.

ericpang777:
@ScottSuarez Ok I think I fixed it for the Update tests (haven't committed yet because I don't want to trigger another build run). But for these import state verify diffs, should this be ignored, or is it a legitimate error?

```
resource_secure_source_manager_repository_generated_test.go:41: Step 2/2 error running import: ImportStateVerify attributes not equivalent. Difference is shown below. The - symbol indicates attributes missing after import.
        
          map[string]string{
        - 	"deletion_policy": "DELETE",
        + 	"deletion_policy": "PREVENT",
          }
```

ScottSuarez:
You can add this field to the import ignore parameter on the test suite.

ScottSuarez (commented):
comment above

ericpang777:
> You can add this field to the import ignore parameter on the test suite.

@ScottSuarez  Done

ScottSuarez:
Thanks for the work eric!

PR #14788, opened 2025-08-08
test

zli82016:

```release-note:none

```


mmv1/api/type.go
```diff
@@ -1354,3 +1357,9 @@ func (t Type) TGCSendEmptyValue() bool {
 
 	return false
 }
+
+// When the property has ignore_read: true and have custom_flatten method,
+// the custom_flatten method should not be applied when compiling the tgc-next provider.
```
trodge:
```suggestion
```
I think the name of this function is a good enough description of what it does and the comment is only adding confusion.

mmv1/templates/terraform/flatten_property_method.go.tmpl
```diff
@@ -14,7 +14,7 @@
   limitations under the License. */ -}}
 {{- define "flattenPropertyMethod" }}
 {{- if $.WriteOnly }}
-{{- else if $.CustomFlatten }}
+{{- else if and $.CustomFlatten (not $.ShouldIgnoreCustomFlatten) }}
```
trodge, line 17: Not necessarily a blocker for this PR but we should probably plan on making a TGC-only version of this template if we keep needing to add TGC-specific logic to it.

trodge (changes requested):

mmv1/templates/terraform/flatten_property_method.go.tmpl
zli82016: Thanks for the suggestion.

zli82016: I will look at it and make a TGC-only version of the templates in a separate PR if needed.

PR #14789, opened 2025-08-08
tgc-revival: set an object to nil if its all values are nil during cai2hcl

zli82016:
In an CAI asset, it is possible that an object field only has output fields. But the output fields are not converted during cai2hcl. The field will be an empty object in converted HCL.

This PR set an object to nil if its all values are nil during cai2hcl. With this change, the field will not show up  instead of empty object in converted HCL.

```release-note:none

```


zli82016:
cc @trodge

hao-nan-li (approved):

PR #14793, opened 2025-08-08
tgc-revival: retry 3 times for missing fields error

zli82016:
`TestAccCertificateManagerCertificate_` failed in other PRs because the `data` in cai asset doesn't have all of the fields in some nightly test runs, only have `name` and `creationTime`. In this case, the test failed with the error of missing fields. 

But in previous nights' runs, `data` does have all of the fields. So retry 3 times to use the previous nights' test data.

```release-note:none

```

### DIFF
--- a/mmv1/third_party/tgc_next/test/assert_test_files.go
+++ b/mmv1/third_party/tgc_next/test/assert_test_files.go
@@ -56,7 +56,7 @@ func BidirectionalConversion(t *testing.T, ignoredFields []string, ignoredAssetF
 
 		logger := zaptest.NewLogger(t)
 
-		// If the primary resource is available, only test the primary resource.
+		// If the primary resource is specified, only test the primary resource.
 		// Otherwise, test all of the resources in the test.
 		if primaryResource != "" {
 			t.Logf("Test for the primary resource %s begins.", primaryResource)
--- a/mmv1/third_party/tgc_next/test/assert_test_files.go
+++ b/mmv1/third_party/tgc_next/test/assert_test_files.go
@@ -11,6 +11,7 @@ import (
 	"strings"
 	"sync"
 	"testing"
+	"time"
 
 	"github.com/GoogleCloudPlatform/terraform-google-conversion/v6/pkg/cai2hcl"
 	cai2hclconverters "github.com/GoogleCloudPlatform/terraform-google-conversion/v6/pkg/cai2hcl/converters"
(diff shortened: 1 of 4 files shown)


mmv1/third_party/tgc_next/test/assert_test_files.go
```diff
@@ -32,38 +34,57 @@ var (
...
-	}
-
-	// Create a temporary directory for running terraform.
-	tfDir, err := os.MkdirTemp(tmpDir, "terraform")
-	if err != nil {
-		log.Fatal(err)
-	}
-	defer os.RemoveAll(tfDir)
+	retries := 0
+	flakyAction := func(ctx context.Context) error {
```
zli82016, line 38: Added a function wrapper to the original code block https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/third_party/tgc_next/test/assert_test_files.go#L35-L68

mmv1/third_party/tgc_next/test/assert_test_files.go
```diff
@@ -32,38 +34,57 @@ var (
...
 		}
-	} else {
-		for _, testData := range resourceTestData {
-			err = testSingleResource(t, t.Name(), testData, tfDir, ignoredFields, ignoredAssetFields, logger, false)
+		defer os.RemoveAll(tfDir)
+
+		logger := zaptest.NewLogger(t)
+
+		// If the primary resource is available, only test the primary resource.
+		// Otherwise, test all of the resources in the test.
```
trodge:
```suggestion
		// If the primary resource is specified, only test the primary resource.
		// Otherwise, test all of the resources in the test.
```
To avoid confusion with the primary resource being unavailable as in missing from the test data.

trodge (approved):

mmv1/third_party/tgc_next/test/assert_test_files.go
```diff
@@ -32,38 +34,57 @@ var (
...
+		// Create a temporary directory for running terraform.
+		tfDir, err := os.MkdirTemp(tmpDir, "terraform")
 		if err != nil {
-			t.Fatal("Test fails:", err)
+			return err
 		}
-	} else {
-		for _, testData := range resourceTestData {
-			err = testSingleResource(t, t.Name(), testData, tfDir, ignoredFields, ignoredAssetFields, logger, false)
+		defer os.RemoveAll(tfDir)
```
zli82016, line 55: Removing the directory works locally.

PR #14799, opened 2025-08-08
google_sql_user doc change to indicate password policy is supported for PG

s-vakati:
google_sql_user documentation says password policy can only be set for MySQL users. Tested that this works for Postgres too. Making a documentation change reflecting the same.

```release-note:none

```


NickElliot (approved):
LGTM

PR #14800, opened 2025-08-08
updated datasource generation to handle certain other code blocks

luckyswaminathan:

```release-note:REPLACEME

```


mmv1/templates/terraform/datasource.go.tmpl
```diff
@@ -84,6 +84,12 @@ func dataSource{{ $.ResourceName -}}Read(d *schema.ResourceData, meta interface{
 		return err
 	}
 
+  {{if $.Datasource.UseSetDataSourceLabel}}
```
roaks3: I think we may not need a config for this, and instead could look at whether the schema includes the expected `labels` field. See https://googlecloudplatform.github.io/magic-modules/develop/add-handwritten-datasource/ which has the guidance for writing datasources, which we should be able to effectively formalize. (Note I would still keep the `datasource: ...` nested config so that it can be added to as needed)

mmv1/templates/terraform/datasource.go.tmpl
```diff
@@ -84,10 +84,12 @@ func dataSource{{ $.ResourceName -}}Read(d *schema.ResourceData, meta interface{
 		return err
 	}
 
-  {{if $.Datasource.UseSetDataSourceLabel}}
-    if err := tpgresource.SetDataSourceLabels(d); err != nil {
-    return err
-    }
+  {{range .Properties}}
+  {{if eq .Name "labels"}}
```
roaks3: You may actually need a custom function here. First, I believe we need to make sure the type is `KeyValueLabels` (there are some other types being used and I don't think they are meant to have this functionality), but also we don't want this to be added to the generated datasource multiple times (there are datasources with multiple `labels` fields).

luckyswaminathan: so you mean a method that identifies whether to use it in resource.go?

roaks3 (approved):
LGTM, could you confirm you ran a test of some sort (locally or otherwise) to confirm the behavior before we merge?

luckyswaminathan:
> LGTM, could you confirm you ran a test of some sort (locally or otherwise) to confirm the behavior before we merge?

yep! correctly does this

PR #14802, opened 2025-08-08
Treat skip_initial_version_creation as a create-only parameter for Cloud KMS keys

dannyqiu:
There is a lot of custom logic that sets the value of `skip_initial_version_creation` to `false`, which breaks imports for external keys. External keys are required to be set `skip_initial_version_creation = true`. Since this field is marked as immutable, Terraform currently attempts to replace the resource by deleting and recreating it. Cloud KMS keys cannot be deleted, which leaves these keys in permadiff. This change fixes this behavior to only update state for changes to `skip_initial_version_creation`.

`skip_initial_version_creation` is a create-only parameter on the crypto_key resource. It is neither returned from the Cloud KMS API, nor does it affect the resource after initial creation. Changing its value should not result in any remote changes.

```release-note:bug
kms: `skip_initial_version_creation` field is no longer immutable in `google_kms_crypto_key`, but is still only settable at-creation
```

### DIFF
--- a/mmv1/products/kms/CryptoKey.yaml
+++ b/mmv1/products/kms/CryptoKey.yaml
@@ -76,7 +76,7 @@ parameters:
       You must use the `google_kms_crypto_key_version` resource to create a new CryptoKeyVersion
       or `google_kms_key_ring_import_job` resource to import the CryptoKeyVersion.
     url_param_only: true
-    immutable: true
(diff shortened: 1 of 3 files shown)


dannyqiu:
Updated the field description.

NickElliot (approved):
LGTM!

PR #14685, opened 2025-07-30
Add support for new flags in Container node kubelet and linux config.

VeraQin:
The new flags in node_kubelet_config include:

- eviction_soft
- eviction_soft_grace_period
- eviction_minimum_reclaim
- eviction_max_pod_grace_period_seconds
- max_parallel_image_pulls

The new flags in linux_node_config include:

- transparent_hugepage_enabled
- transparent_hugepage_defrag

The new flags in sole_tenant_config include:

- sole_tenant_config

Change-Id: I45c6a35137f532b88380dfd9359f893d25fef859

**Release Note Template for Downstream PRs (will be copied)**

See [Write release notes](https://googlecloudplatform.github.io/magic-modules/contribute/release-notes/) for guidance.

```release-note:enhancement
container: added `eviction_soft`, `eviction_soft_grace_period`, `eviction_minimum_reclaim`, `eviction_max_pod_grace_period_seconds`, `max_parallel_image_pulls`, `transparent_hugepage_enabled`, `transparent_hugepage_defrag` and `min_node_cpus` fields to `node_config` block of `google_container_node_pool` and `google_container_cluster` resources
```

PR #14834, opened 2025-08-13
container: Removed instance type from resourceManagerTags test

wyardley:
Remove the explicit instance type from `TestAccContainerNodePool_resourceManagerTags `

`n1` instance types may have stockout issues. Also, it doesn't appear that SSDs are needed anyway, so this was probably cargo-culted. The tests appear to pass with the default instance type.

Part of https://github.com/hashicorp/terraform-provider-google/issues/23597

```release-note:none

```

### DIFF
--- a/mmv1/third_party/terraform/services/container/resource_container_node_pool_test.go.tmpl
+++ b/mmv1/third_party/terraform/services/container/resource_container_node_pool_test.go.tmpl
@@ -5559,7 +5559,6 @@ resource "google_container_node_pool" "primary_nodes" {
   node_count = 1
 
   node_config {
-    machine_type = "n1-standard-1" // can't be e2 because of local-ssd
     disk_size_gb = 15
 
     resource_manager_tags = {
@@ -5606,12 +5605,11 @@ resource "google_container_node_pool" "primary_nodes" {
   node_count = 1
 
   node_config {
-    machine_type = "n1-standard-1" // can't be e2 because of local-ssd
     disk_size_gb = 15
 
     resource_manager_tags = {
       "%{pid}/%{tagKey1}" = "%{tagValue1}"
-	  "%{pid}/%{tagKey2}" = "%{tagValue2}"
+      "%{pid}/%{tagKey2}" = "%{tagValue2}"
     }
   }
 }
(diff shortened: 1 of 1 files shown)


BBBmau (commented):
can you also resolve the following container_node_pool test here:

```hcl
    resource_container_node_pool_test.go:1197: Step 4/4 error: Check failed: Check 1/2 error: Check 2/2 error: google_container_node_pool.with_tier1_net: Attribute 'network_config.0.network_performance_config.0.total_egress_bandwidth_tier' expected "TIER_UNSPECIFIED", got "TIER_1"
    testing_new.go:90: Error running post-test destroy, there may be dangling resources: Cluster still exists
--- FAIL: TestAccContainerNodePool_withNetworkConfig (12.29s)
```

wyardley:
> can you also resolve the following container_node_pool test here:

I am not sure if this one is related; also not sure if it's a consistent failure. I'll test and see if I can reproduce it locally, though.

The "dangling resources" error makes me wonder if it's some sort of timing thing or just slow destroy? Don't see an existing ticket for `TestAccContainerNodePool_withNetworkConfig` as a test failure.

wyardley:
Side note: when trying to test `TestAccContainerNodePool_withNetworkConfig` locally, I get
`Error: error creating NodePool: googleapi: Error 400: Disabling pod cidr overprovision is not allowed for this project.`

IMO, it might make sense to use a zonal cluster for testing and / or set a more limited set of zone(s) for`node_locations`, to reduce the node count of all the different node pools. I can make a PR for that if someone thinks it makes sense.

I'm assuming `TestAccContainerCluster_withNodeConfigKubeletConfigSettingsInNodePool` is also unrelated, I'll see if I can repro that one too, but would probably PR it separately if I could?

wyardley:
Seems like `TestAccContainerCluster_withNodeConfigKubeletConfigSettingsInNodePool` was very recently added in #14685, so maybe the author of that PR could take a look? (*edit* looks like it's a sole tenant instance too, so will be difficult / expensive to test).

wyardley:
@BBBmau can you try rerunning to pull in recent changes? I also have a pass at making some improvements to `TestAccContainerNodePool_withNetworkConfig`, but for reasons mentioned there, I can't easily test it locally.

PR #14835, opened 2025-08-13
feat: (storage) added field aws_s3_compatible_data_source for google_storage_transfer_job resource

gurusai-voleti:

```release-note:enhancement
storage:  added `aws_s3_compatible_data_source` to `google_storage_transfer_job` resource
```


mmv1/third_party/terraform/services/storagetransfer/resource_storage_transfer_job_test.go
```diff
@@ -2399,3 +2399,169 @@ resource "google_storage_transfer_job" "transfer_job" {
 }
 `, project, dataSourceBucketName, project, dataSinkBucketName, project, transferJobDescription, project)
 }
+
+func TestAccStorageTransferJob_awsS3CompatibleDataSource(t *testing.T) {
+	t.Parallel()
+
+	testDataSourceBucketName := acctest.RandString(t, 10)
+	testDataSinkName := acctest.RandString(t, 10)
+	testTransferJobDescription := acctest.RandString(t, 10)
```
trodge: Please make sure these resource names are prefixed with `tf-test-` so that our sweepers can clean them up in the case of a test failure.

This would not currently apply to the `description` field of the storage transfer job itself since we don't seem to have a sweeper for this resource. If I'm reading the API documentation correctly it looks like the description is meant to be a human readable string and the `name` field is the server-generated ID for the resource. If that's correct I'd reccomend putting the description in the test config as something like 
```
description = "A test job using an aws s3 compatible data source"
```

trodge (changes requested):

mmv1/third_party/terraform/services/storagetransfer/resource_storage_transfer_job_test.go
gurusai-voleti: Updated the description

mmv1/third_party/terraform/services/storagetransfer/resource_storage_transfer_job.go
```diff
@@ -1461,6 +1533,74 @@ func flattenHdfsData(hdfsData *storagetransfer.HdfsData) []map[string]interface{
...
+
+	awsS3CompatibleData := awsS3CompatibleDataSchema[0].(map[string]interface{})
+	result := &storagetransfer.AwsS3CompatibleData{
+		BucketName: awsS3CompatibleData["bucket_name"].(string),
+		Path:       awsS3CompatibleData["path"].(string),
+		Endpoint:   awsS3CompatibleData["endpoint"].(string),
+		Region:     awsS3CompatibleData["region"].(string),
+	}
+
+	if awsS3CompatibleData["s3_metadata"] != nil {
```
kautikdk: I believe this condition is redundant since we are checking for nil and unset value in the expander itself.

mmv1/third_party/terraform/services/storagetransfer/resource_storage_transfer_job.go
```diff
@@ -1461,6 +1533,74 @@ func flattenHdfsData(hdfsData *storagetransfer.HdfsData) []map[string]interface{
...
+
+	return []map[string]interface{}{data}
+}
+
+func flattenS3MetaData(s3MetaData *storagetransfer.S3CompatibleMetadata, d *schema.ResourceData) []map[string]interface{} {
+	s3Metadata := map[string]interface{}{
+		"protocol":      s3MetaData.Protocol,
+		"request_model": s3MetaData.RequestModel,
+	}
+	if d.Get("transfer_spec.0.aws_s3_compatible_data_source.0.s3_metadata.0.auth_method") == "AUTH_METHOD_UNSPECIFIED" {
```
kautikdk, line 1628: Why this customized logic is required? If `AUTH_METHOD_UNSPECIFIED` is specified in the `auth_method` then we don't get it back it from the API response?

mmv1/third_party/terraform/services/storagetransfer/resource_storage_transfer_job_test.go
```diff
@@ -2399,3 +2399,168 @@ resource "google_storage_transfer_job" "transfer_job" {
...
+  role   = "roles/storage.admin"
+  member = "serviceAccount:${data.google_storage_transfer_project_service_account.default.email}"
+}
+
+resource "google_storage_transfer_job" "transfer_job" {
+  description = "A test job using an aws s3 compatible data source"
+  project     = "%s"
+
+  transfer_spec {
+    aws_s3_compatible_data_source {
```
kautikdk, line 2578: The only updates are happening with three nested `s3_metadata` fields and one top level `region` field. Other fields are as it is without any modification. Can we update other top level `aws_s3_compatible_data_source` fields as well if possible? and `s3_metadata.request_model` field. Preferable update would be to set them to zero values or unsetting them.

mmv1/third_party/terraform/services/storagetransfer/resource_storage_transfer_job.go
gurusai-voleti: yes API is not returning the value

mmv1/third_party/terraform/services/storagetransfer/resource_storage_transfer_job.go
gurusai-voleti: removed condition

trodge: To be safe against panics you may want to use two-value type assertions here (e.g. `if vSlice, ok := awsS3CompatibleData["s3_metadata"].([]interface{}); ok`), but it's not strictly necessary.

trodge (changes requested):
More thorough update test coverage as recommended by @kautikdk would be helpful.

gurusai-voleti:
> More thorough update test coverage as recommended by @kautikdk would be helpful.

yes  implemented the update test case as per Kautik comments

mmv1/third_party/terraform/services/storagetransfer/resource_storage_transfer_job_test.go
gurusai-voleti: added other fields in update test case

mmv1/third_party/terraform/services/storagetransfer/resource_storage_transfer_job.go
gurusai-voleti: updated with assertion

trodge (approved):
LGTM, pending approval from @kautikdk

gurusai-voleti:
@trodge  @kautikdk  please review/approve the PR

kautikdk (approved):

trodge (approved):

PR #14836, opened 2025-08-13
All squash support

raazanand:
```release-note:enhancement
volume: added `squashMode` & `anonUid ` field inside volume `exportPolicy`.
```

### DIFF
--- a/mmv1/products/chronicle/ReferenceList.yaml
+++ b/mmv1/products/chronicle/ReferenceList.yaml
@@ -34,6 +34,7 @@ examples:
     primary_resource_id: 'example'
     vars:
       reference_list_id: reference_list_id
+      data_access_scope_id: scope-id
     test_env_vars:
       chronicle_id: 'CHRONICLE_ID'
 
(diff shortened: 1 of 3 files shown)


mmv1/products/netapp/Volume.yaml
```diff
@@ -191,6 +191,20 @@ properties:
               type: Boolean
               description: |
                 If enabled (true) the rule defines read and write access for clients matching the 'allowedClients' specification. It enables nfs clients to mount using 'privacy' kerberos security mode. The 'kerberos5pReadOnly' value is ignored if this is enabled.
+            - name: 'squashMode'
+              type: Enum
+              description: |-
+                SquashMode defines how remote user privileges are restricted when accessing an NFS export. It controls how the user identities (like root) are mapped to anonymous users to limit access and enforce security.
+              enum_values:
+                - 'SQUASH_MODE_UNSPECIFIED'
```
SirGitsalot: Try removing `SQUASH_MODE_UNSPECIFIED` here and the `default_value` line below. We [don't normally include an enum's zero value](https://googlecloudplatform.github.io/magic-modules/reference/field/#enum-properties) and treat "field not present" as "XXX_UNSPECIFIED" instead.

PR #14837, opened 2025-08-13
Add resource_manager_tags support to Region Backend Service api

axelkossek:
Added resource manager tags support to Ragion Backend Service.
Part of https://github.com/hashicorp/terraform-provider-google/issues/23979

#### Release Note Template for Downstream PRs (will be copied)

```release-note:enhancement
compute: add `params.resourceManagerTags` field to the `google_compute_region_backend_service`
```

### DIFF
--- a/mmv1/products/compute/RegionBackendService.yaml
+++ b/mmv1/products/compute/RegionBackendService.yaml
@@ -1550,12 +1550,12 @@ properties:
     immutable: true
     description: |
       Additional params passed with the request, but not persisted as part of resource payload
-      properties:
-        - name: 'resourceManagerTags'
-          type: KeyValuePairs
-          description: |
-            Resource manager tags to be bound to the region backend service. Tag keys and values have the
-            same definition as resource manager tags. Keys must be in the format tagKeys/{tag_key_id},
-            and values are in the format tagValues/456.
-          api_name: resourceManagerTags
-          ignore_read: true
+    properties:
+      - name: 'resourceManagerTags'
+        type: KeyValuePairs
+        description: |
+          Resource manager tags to be bound to the region backend service. Tag keys and values have the
+          same definition as resource manager tags. Keys must be in the format tagKeys/{tag_key_id},
+          and values are in the format tagValues/456.
+        api_name: resourceManagerTags
+        ignore_read: true
--- a/mmv1/products/compute/RegionBackendService.yaml
+++ b/mmv1/products/compute/RegionBackendService.yaml
@@ -1544,3 +1544,18 @@ properties:
                 description: |
                   The name of the VM instance of the leader network endpoint. The instance must
                   already be attached to the NEG specified in the haPolicy.leader.backendGroup.
+  - name: 'params'
+    type: NestedObject
+    ignore_read: true
+    immutable: true
+    description: |
+      Additional params passed with the request, but not persisted as part of resource payload
+      properties:
+        - name: 'resourceManagerTags'
+          type: KeyValuePairs
+          description: |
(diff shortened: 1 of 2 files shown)


axelkossek:
Added resource manager tags support to Ragion Backend Service.
Part of https://github.com/hashicorp/terraform-provider-google/issues/23979

```
compute: add `params.resourceManagerTags` field to the `google_compute_region_backend_service`
```

mmv1/products/compute/RegionBackendService.yaml
```diff
@@ -1544,3 +1544,18 @@ properties:
...
+      Additional params passed with the request, but not persisted as part of resource payload
+    properties:
+      - name: 'resourceManagerTags'
+        type: KeyValuePairs
+        description: |
+          Resource manager tags to be bound to the region backend service. Tag keys and values have the
+          same definition as resource manager tags. Keys must be in the format tagKeys/{tag_key_id},
+          and values are in the format tagValues/456.
+        api_name: resourceManagerTags
+        ignore_read: true
```
trodge, line 1561: Should this field also be immutable, or is it only the parent field?

axelkossek: Must have missed it, added it.

mmv1/third_party/terraform/services/compute/resource_compute_region_backend_service_test.go.tmpl
```diff
@@ -476,6 +477,35 @@ func TestAccComputeRegionBackendService_withDynamicBackendCount(t *testing.T) {
...
+		CheckDestroy:             testAccCheckComputeRegionBackendServiceDestroyProducer(t),
+		Steps: []resource.TestStep{
+			{
+				Config: testAccComputeRegionBackendService_withTags(serviceName, checkName, tagKeyResult["name"], tagValueResult["name"]),
+			},
+			{
+				ResourceName:      "google_compute_region_backend_service.foobar",
+				ImportState:       true,
+				ImportStateVerify: true,
+        ImportStateVerifyIgnore: []string{"params"},
```
trodge:
```suggestion
				ImportStateVerify: true,
				ImportStateVerifyIgnore: []string{"params"},
```
nit: indentation fix

mmv1/third_party/terraform/services/compute/resource_compute_region_backend_service_test.go.tmpl
trodge, line 1: Formatting this file can be tricky because `go fmt` only works for actual `.go` files and this is a `.go.tmpl` file. We run it through the formatter after generation, so one trick is to just copy the equivalent code from the downstream diff comment. https://github.com/GoogleCloudPlatform/magic-modules/pull/14837#issuecomment-3208494735

As long as the section you're formatting doesn't contain any template syntax (e.g. `{{` and '}}`) the equivalent `.go` file in `terraform-provider-google` will contain the correctly formatted code.

Currently I see mixed tabs and spaces, which could display inconsistently in different editors.

axelkossek: Some other tests have inconsistent tabbing, some have spaces only, command strings have two space tabbing. Should I format all tests or just mine? I am not sure what is the etiquette in this repository.

trodge: Probably best to just format the test you're adding. The string literals for terraform configs should continue to use two space tabbing while the go code should use tab characters.

PR #14847, opened 2025-08-13
container: Used zonal cluster for network config test

wyardley:
- Use zonal cluster for `TestAccContainerNodePool_withNetworkConfig` test. Pick zone c (though a is used for most other zonal tests). This should result in 1/3 or 1/4 of the nodes being created for each of the multiple node pools in this test, which should improve both the speed and the reliability of this test
- Fix formatting

Noted while working on #14834

Note: I haven't been able to test this locally, both because of quota issues, and because of this error:
`Disabling pod cidr overprovision is not allowed for this project`
on my personal test project

```release-note:none

```

### DIFF
--- a/mmv1/third_party/terraform/services/container/resource_container_node_pool_test.go.tmpl
+++ b/mmv1/third_party/terraform/services/container/resource_container_node_pool_test.go.tmpl
@@ -4172,7 +4172,7 @@ resource "google_container_node_pool" "with_tier1_net" {
   cluster    = google_container_cluster.cluster.name
   node_count = 1
   node_locations = [
-    "us-central1-a",
+    "us-central1-c",
   ]
   network_config {
     network_performance_config {


slevenick:
I'm not sure why this error happened:
=== NAME  TestAccContainerNodePool_withConfidentialBootDisk
    bootstrap_iam_test_utils.go:119: [DEBUG] All bindings already present for project ci-project
=== NAME  TestAccContainerCluster_withConfidentialBootDisk
    bootstrap_iam_test_utils.go:119: [DEBUG] All bindings already present for project ci-project
panic: interface conversion: interface {} is *url.Error, not []interface {}

It shouldn't be related? I can retry

wyardley:
Huh, yeah, that's weird

There could be some kind of mistake in here, though it's really only terraform code changed. As mentioned in the description, I unfortunately can't easily test this one locally at the moment.

mmv1/third_party/terraform/services/container/resource_container_node_pool_test.go.tmpl
```diff
@@ -4110,85 +4111,84 @@ resource "google_container_cluster" "cluster" {
...
   }
   deletion_protection = false
 }
 
 resource "google_container_node_pool" "with_manual_pod_cidr" {
-  name               = "%s-manual"
-  location           = "us-central1"
-  cluster            = google_container_cluster.cluster.name
+  name       = "%s-manual"
+  location   = "us-central1-c"
```
wyardley: @slevenick not sure if it would be best / cleaner to use `google_container_cluster.cluster.location` in all of these if it works? obviously there's already an implicit dependency already with the cluster name, but it would keep them consistent without having to declare a variable. 🤷 
Happy to update it to that if you think it would be better... I think it will work either way.

slevenick: Yeah, that sounds reasonable. Then updating the location will be super simple in the future

slevenick:
The error is still happening, but I'm sure it's entirely unrelated to your change. I'll try to investigate that

slevenick:
Ok my PR passed, so it actually does seem to be something in your change. I'm really not sure why. One possible fix is changing the test name so it thinks it's entirely new. Can you give that a try?

wyardley:
Maybe it’s that extra empty line I removed before the backtick? I’m not sure why it would be an issue, though. I’ll try adding the location change and rebasing before I mess around with changing the test name.

mmv1/third_party/terraform/services/container/resource_container_node_pool_test.go.tmpl
wyardley: Ok, updated. Trying a local test with the two node pools that were causing me issues commented out.

wyardley:
I rebased, updated the locations, and also put that extra newline back, though I don't think it's the issue. Also fixed an issue where the large nodepool already had `node_locations` set explicitly... left it there, but updated to zone c to match the cluster.
Trying to repro the replaying issue locally by commenting out the two nodepools that I was having issues with temporarily.

wyardley:
Seems like we’re back to a “normal” replaying error for the right test, at least. Can you let me know what the error is now?

slevenick:
resource_container_node_pool_test.go:1197: Step 4/4 error: Check failed: Check 1/2 error: Check 2/2 error: google_container_node_pool.with_tier1_net: Attribute 'network_config.0.network_performance_config.0.total_egress_bandwidth_tier' expected "TIER_UNSPECIFIED", got "TIER_1"

wyardley:
So basically seems like the tier didn't get updated on the "update" step... but then it seems very odd that the test passed on the `RECORDING` run.

slevenick:
/gcbrun

Yeah... that's super unusual

PR #14810, opened 2025-08-11
Add support for default compute class on GKE

olagacek:
Enable Terraform support for cluster autoscaling default compute class config.

**Release Note Template for Downstream PRs (will be copied)**

```release-note:enhancement
container: added `cluster_autoscaling.default_compute_class_enabled` field to `google_container_cluster` resource
```

PR #14848, opened 2025-08-13
Update google.golang.org/api package to the 0.245.0 version.

olagacek:

Update google.golang.org/api package to the 0.245.0 version.

```release-note:none

```


olagacek:
> What's the purpose of this update? Does it block any incoming change?

Yes, it's needed for https://github.com/GoogleCloudPlatform/magic-modules/pull/14810, as the 0.245.0 is the earliest version in which default compute class config is available

hao-nan-li (approved):
Thanks for creating a separate PR for the version update

PR #14849, opened 2025-08-13
google_dialogflow_conversation_profile

c2thorn:
pulled from https://github.com/GoogleCloudPlatform/magic-modules/pull/14324, which I was previously reviewing but now taking over. This has a few minor changes related to adding `ignore_read` on a field and updating import state verify ignore on tests.

fixes https://github.com/hashicorp/terraform-provider-google/issues/12466

```release-note:new-resource
`google_dialogflow_convesation_profile`
```


c2thorn:
> For the missing fields, do we have a way to test them? Looks like it's mostly model_versions.

with the time remaining, I don't think it's worthwhile. It's a very large resource with a lot of configuration options, it would take a bit to find valid configurations. I think it's more beneficial to push it through before the major release than hold it or drop the fields.

PR #14857, opened 2025-08-14
write-only: modified expander behavior for write-only arguments

ramonvermeulen:
Closes https://github.com/hashicorp/terraform-provider-google/issues/23211

@melinath 

- encountered some potential duplicates with the `exactly_one_of` fields etc. Therefore added deduplication logic.
- added new method `CamelizeProperty` to return a camelized variant of the property name. Was indending to use `TitelizeProperty` first but noticed this breaks a lot of tests because it changes a lot of property names, makes more sense to try to keep the name closer to what it was. With `CamelizeProperty` I only had to change some `labelsProp` references to `effectiveLabelsProp`.
- added `GetRawConfigAttributeAsString` as util method, to be able to retrieve the value of the write-only argument so it can be used when creating the request data object.

```release-note:none
write-only: modified expander behavior for write-only arguments
```

Came across these shortcomings in, but probably better to put in a separate (this) PR:

https://github.com/GoogleCloudPlatform/magic-modules/pull/14138
https://github.com/GoogleCloudPlatform/magic-modules/pull/13847

### DIFF
--- a/mmv1/api/type.go
+++ b/mmv1/api/type.go
@@ -508,6 +508,10 @@ func (t Type) TitlelizeProperty() string {
 	return google.Camelize(t.Name, "upper")
 }
 
+func (t Type) CamelizeProperty() string {
+	return google.Camelize(t.Name, "lower")
+}
+
 // If the Prefix field is already set, returns the value.
 // Otherwise, set the Prefix field and returns the value.
 func (t *Type) GetPrefix() string {
--- a/mmv1/api/resource.go
+++ b/mmv1/api/resource.go
@@ -741,6 +741,20 @@ func (r Resource) GetIdentity() []*Type {
 	})
 }
 
+func deduplicateSliceOfStrings(slice []string) []string {
+	seen := make(map[string]bool, len(slice))
+	result := make([]string, 0, len(slice))
+
+	for _, str := range slice {
+		if !seen[str] {
+			seen[str] = true
+			result = append(result, str)
+		}
+	}
+
+	return result
+}
+
 func buildWriteOnlyField(name string, versionFieldName string, originalField *Type, originalFieldLineage string) *Type {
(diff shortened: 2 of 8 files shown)


melinath:

mmv1/api/type.go
```diff
@@ -508,6 +508,10 @@ func (t Type) TitlelizeProperty() string {
 	return google.Camelize(t.Name, "upper")
 }
 
+func (t Type) CamelizeProperty() string {
```
BBBmau, line 511: just so i understand, we run into duplicate fields in `exactly_one_of` due to the fields not being in the same style correct? i.e `accessContextName` vs `accesscontextname`

Reviewed `GetRawConfigAttributeAsString`, looks good!

ramonvermeulen: In the past the `{{ $prop.ApiName -}}Prop` was used as variable name for the expand method. The problem with this is that for a write-only field, the api_name is the same as the original field. Resulting in two variables with exactly the same name getting generated. That is why I changed this to the camelized version of the property name instead. This way most generated property names will stay exactly the same, except the ones that had `api_name` set, such as write-only properties.

The deduplicate fields is another thing, to ensure after programatically adding entries to `conflicts`, `exactly_one_of` or `at_least_one_of` it ensures the output can never have a duplicate entry. E.g. if the original field was for example already configured in the `exactly_one_of`, it shouldn't be added twice.

melinath:
(CamelizeProperty seems like a fine new addition since it uses lowercase for the first letter, which is what we want in this case.)

ramonvermeulen:
> Could you split the `CamelizeProperty` changes into a separate PR? It causes so many changes in the downstreams that it makes it much harder to evaluate the impact of the rest of this PR.

Check, moved to a separate PR at: https://github.com/GoogleCloudPlatform/magic-modules/pull/14906, will rebase this one after it is merged.

melinath (commented):
Removing from my queue for now pending rebase.

ramonvermeulen:
Moved to: https://github.com/GoogleCloudPlatform/magic-modules/pull/14933

PR #14863, opened 2025-08-14
tgc-revival: modify the testing logic by testing each test step

zli82016:

```release-note:none

```

### DIFF
--- a/mmv1/third_party/tgc_next/test/setup.go
+++ b/mmv1/third_party/tgc_next/test/setup.go
@@ -28,8 +28,8 @@ type CaiData struct {
 }
 
 type NightlyRun struct {
-	MetadataByTest map[string][]TgcMetadataPayload
-	Date           time.Time
+	MetadataByTestAndStep map[string]map[int]TgcMetadataPayload
+	Date                  time.Time
 }
 
 // The metadata for each step in one test
@@ -103,8 +103,8 @@ func ReadTestsDataFromGcs() ([]NightlyRun, error) {
 				}
 			} else {
 				TestsMetadata[i] = NightlyRun{
-					MetadataByTest: metadata,
-					Date:           currentDate,
+					MetadataByTestAndStep: metadata,
+					Date:                  currentDate,
 				}
 			}
 			currentDate = currentDate.AddDate(0, 0, -1)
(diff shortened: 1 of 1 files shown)


mmv1/third_party/tgc_next/test/setup.go
```diff
@@ -28,12 +28,14 @@ type CaiData struct {
 }
 
 type NightlyRun struct {
-	MetadataByTest map[string]TgcMetadataPayload
+	MetadataByTest map[string][]TgcMetadataPayload
```
trodge: Since the step numbers are not a sequence (some tests have non-config steps), we may want to store them in a `map[int]RawConfig` type. That way we'd be able to find the correct config for each step rather than iterate through the steps looking for it.

I'd also prefer to make the value type a struct or type alias rather than a slice, because the meaning of `map[string][]TgcMetadataPayload` is not obvious.

mmv1/third_party/tgc_next/test/setup.go
```diff
@@ -28,12 +28,14 @@ type CaiData struct {
...
 	Date           time.Time
 }
 
+// The metadata for each step in one test
 type TgcMetadataPayload struct {
 	TestName         string                       `json:"test_name"`
+	StepNumber       int                          `json:"step_number"`
```
trodge, line 38: Would it be possible to have a single metadata payload which contains multiple steps? The way this is currently structured implies that each step could have a different set of resource metadata objects. I was under the impression that resource metadata should be the same across a single test.

mmv1/products/certificatemanager/Certificate.yaml
```diff
@@ -66,28 +66,33 @@ examples:
       dns_auth_name2: 'dns-auth2'
       dns_auth_subdomain2: 'subdomain2'
       cert_name: 'dns-cert'
+    tgc_skip_test: 'Need to figure out to get the desired cai asset state.'
```
trodge, line 69: Why do these tests in particular need to be skipped? Have they been failing on main or are they only failing after these changes to the test logic?

trodge: Discussed offline and plan to fix in a follow-up PR.

mmv1/third_party/tgc_next/test/setup.go
trodge: Based on offline discussion, the preferred approach is to have a single payload for each step which we store in a two layer map by test name and step number.

mmv1/third_party/tgc_next/test/setup.go
```diff
@@ -28,12 +28,14 @@ type CaiData struct {
 }
 
 type NightlyRun struct {
-	MetadataByTest map[string]TgcMetadataPayload
-	Date           time.Time
+	MetadataByTestAndStep map[string]map[int]TgcMetadataPayload
```
zli82016, line 31: Modified the map name and to have two layers of map for now.

mmv1/third_party/tgc_next/test/assert_test_files.go
```diff
@@ -34,59 +34,77 @@ var (
...
-			err = testSingleResource(t, t.Name(), resourceTestData[primaryResource], tfDir, ignoredFields, ignoredAssetFields, logger, true)
-			if err != nil {
-				return err
-			}
-		} else {
-			for _, testData := range resourceTestData {
-				err = testSingleResource(t, t.Name(), testData, tfDir, ignoredFields, ignoredAssetFields, logger, false)
+	for _, stepN := range stepNumbers {
+		subtestName := fmt.Sprintf("step%d", stepN)
+		t.Run(subtestName, func(t *testing.T) {
```
trodge, line 58: Out of curiousity, why does this function need to be defined anonymously?

trodge (approved):

mmv1/third_party/tgc_next/test/assert_test_files.go
zli82016: This function refers some variables from the outer function. 

I will double check if it works to make it a separate function.

PR #14872, opened 2025-08-15
autogeneration of tests for singular datasources

luckyswaminathan:
https://github.com/hashicorp/terraform-provider-google/issues/12595 part of this initiative

```release-note:none
autogeneration of tests enabled
```


mmv1/products/redis/Cluster.yaml
```diff
@@ -108,6 +108,9 @@ timeouts:
   update_minutes: 120
   delete_minutes: 30
 autogen_async: true
+datasource:
+  generate: true
+  generate_tests: true
```
melinath: Given there are only four datasources generated at the moment, could we actually just convert all four in this PR (with simplified diffs) and make that the default behavior, instead of having a `generate_tests` flag that needs to be set?

luckyswaminathan: so the problem with this is that there are generatable datasources (ie storagecontrol) that don't have generatable tests, as they only have handrolled tests, so no real way of generating from that template

luckyswaminathan:
> Rather than converting a new datasource to have generated tests, could you take a resource that already has a generated datasource and convert its tests?
> 
> Ideally, it would be nice to rename the handwritten test file to have the same name as the generated file in a separate PR, then remove the handwritten test file in this PR. That would make the diff easier to review.


I can for certain datasources yeah, I think it might best to keep the flag(so datasources with handrolled resource tests are compatible), but still replacing tests for compatible datasources such as the redis/iapclient ones done in this PR, let me know if you think this isn't the right way

mmv1/products/redis/Cluster.yaml
melinath: yeah, we probably want to have an `exclude_test` attribute that you can set to _not_ generate, similar to the `exclude_docs` & `exclude_test` for examples, which defaults to `false`.

The alternative would be to make `generate_test` default to true (advantage being that we wouldn't end up with a double-negative) but we generally try to avoid that at the moment (I forget why - maybe hard to tell the difference between "unset" and "set to false"?)

melinath: Basically, the default behavior should be "generate tests" since that's what we expect people to do in most cases.

mmv1/provider/terraform.go
```diff
@@ -203,6 +204,21 @@ func (t *Terraform) GenerateSingularDataSource(object api.Resource, templateData
...
+	if !object.ShouldGenerateSingularDataSourceTests() {
+		return
+	}
+
+	productName := t.Product.ApiName
+	targetFolder := path.Join(outputFolder, t.FolderName(), "services", productName)
+	if err := os.MkdirAll(targetFolder, os.ModePerm); err != nil {
+		log.Println(fmt.Errorf("error creating parent directory %v: %v", targetFolder, err))
+	}
+	targetFilePath := path.Join(targetFolder, fmt.Sprintf("data_source_%s_test.go", t.ResourceGoFilename(object)))
```
melinath, line 222: 👍 doing this for now so the diff is clearer seems reasonable - we should switch to `_generated_test.go` in the near future, but that can be in a separate PR.

mmv1/templates/terraform/examples/base_configs/datasource_test_file.go.tmpl
```diff
@@ -0,0 +1,90 @@
...
+			"{{$provider}}": {},
+		{{- end }}
+		},
+	{{- end }}
+	{{- if not $.Res.ExcludeDelete }}
+		CheckDestroy: testAccCheck{{ $.Res.ResourceName }}DestroyProducer(t),
+{{- end }}
+		Steps: []resource.TestStep{
+			{
+				Config: testAcc{{ $e.TestSlug $.Res.ProductMetadata.Name $.Res.Name }}DataSource(context),
```
melinath, line 63: This should call `CheckDataSourceStateMatchesResourceState` or `CheckDataSourceStateMatchesResourceStateWithIgnores` as it did in the IAP/Redis tests.

mmv1/templates/terraform/examples/base_configs/datasource_test_file.go.tmpl
```diff
@@ -0,0 +1,90 @@
...
+
+func testAcc{{ $e.TestSlug $.Res.ProductMetadata.Name $.Res.Name }}DataSource(context map[string]interface{}) string {
+	return fmt.Sprintf(`%s
+
+		data "{{ $.Res.TerraformName }}" "default" {
+		{{- range $fieldName := $.Res.DatasourceRequiredFields }}
+		{{ $fieldName }} = {{ $e.ResourceType $.Res.TerraformName }}.{{ $e.PrimaryResourceId }}.{{ $fieldName }}
+		{{- end }}
+		}`,
+		testAcc{{ $e.TestSlug $.Res.ProductMetadata.Name $.Res.Name }}(context),
```
melinath: We should probably inline the configuration (similar to [what IAM tests do](https://github.com/GoogleCloudPlatform/magic-modules/blob/d6a1df847d320ce3dfc03e2e86535c30150fe331/mmv1/templates/terraform/examples/base_configs/iam_test_file.go.tmpl#L377)) rather than using string formatting - string formatting like this makes it harder to read the final result (both for humans and for the missing test detector).

mmv1/api/resource/datasource.go
```diff
@@ -16,4 +16,6 @@ package resource
 type Datasource struct {
 	// boolean to determine whether the datasource file should be generated
 	Generate bool `yaml:"generate"`
+	// boolean to determine whether tests should be generated for a datasource
+	ExcludeTestGeneration bool `yaml:"exclude_test_generation"`
```
melinath: nit: for consistency with https://github.com/GoogleCloudPlatform/magic-modules/blob/80031d2e77a91d48a3175ddac15f05c2a9f4b4ae/mmv1/api/resource/examples.go#L139

```suggestion
	ExcludeTest bool `yaml:"exclude_test"`
```

luckyswaminathan:
> removing this from my queue for now since it sounds like you're still tweaking it - let me know if that's not the case.

I think its good now - redis cluster was just a bad example of testing generation due to how their examples are structured, but this works with IAP client and I think will with a lot of others

mmv1/templates/terraform/examples/base_configs/datasource_test_file.go.tmpl
```diff
@@ -0,0 +1,95 @@
...
+func testAcc{{ $e.TestSlug $.Res.ProductMetadata.Name $.Res.Name }}DataSource(context map[string]interface{}) string {
+	return acctest.Nprintf(`
+
+		{{ $e.TestHCLText }}
+
+		data "{{ $.Res.TerraformName }}" "default" {
+		{{- range $fieldName := $.Res.DatasourceRequiredFields }}
+		{{ $fieldName }} = {{ $e.ResourceType $.Res.TerraformName }}.{{ $e.PrimaryResourceId }}.{{ $fieldName }}
+		{{- end }}
+		}`,
```
melinath:
```suggestion
{{ $e.TestHCLText }}

data "{{ $.Res.TerraformName }}" "default" {
{{- range $fieldName := $.Res.DatasourceRequiredFields }}
  {{ $fieldName }} = {{ $e.ResourceType $.Res.TerraformName }}.{{ $e.PrimaryResourceId }}.{{ $fieldName }}
{{- end }}
}`,
```

melinath (changes requested):
small whitespace change to better match existing tests

mmv1/products/redis/Cluster.yaml
```diff
@@ -108,6 +108,9 @@ timeouts:
   update_minutes: 120
   delete_minutes: 30
 autogen_async: true
+datasource:
+  generate: true
+  exclude_test: true
```
melinath: It looks like redis still isn't quite equivalent on generation - we're losing the call to `SetDataSourceLabels`. Let's just undo the generation (and put back the handwritten file) for now so this PR can focus on the test generation changes.

```suggestion
```

mmv1/templates/terraform/examples/base_configs/datasource_test_file.go.tmpl
```diff
@@ -0,0 +1,94 @@
...
+		},
+	{{- end }}
+	{{- if not $.Res.ExcludeDelete }}
+		CheckDestroy: testAccCheck{{ $.Res.ResourceName }}DestroyProducer(t),
+{{- end }}
+		Steps: []resource.TestStep{
+			{
+				Config: testAcc{{ $e.TestSlug $.Res.ProductMetadata.Name $.Res.Name }}DataSource(context),
+				Check: resource.ComposeTestCheckFunc(
+					acctest.CheckDataSourceStateMatchesResourceState("data.{{ $e.ResourceType $.Res.TerraformName }}.default", "{{ $e.ResourceType $.Res.TerraformName }}.{{ $e.PrimaryResourceId }}"),
```
melinath: We should be doing something like this here to match the [current code](https://github.com/modular-magician/terraform-provider-google-beta/compare/b017f813fe60c9435ab1f7960af86c092a3ab851..629c68228b261422eb44c63e7bc22eef442f931b#diff-870a7355f48cf4cb19916f78fdba32a314fd7bbf7bd459c3a998e1f35fd99251L43-L49). While IAP tests seem to pass regardless, that may not be the case for other data sources.

Note that this would require implementing `IgnoreReadProperties` - which should probably replace `IgnoreReadPropertiesToString`. For this PR we could just keep its usage to this PR and then replace the single other usage & remove the string version in a follow-on PR.

```suggestion
		{{- if gt (len ($.Res.IgnoreReadProperties $e)) 0 }}
					acctest.CheckDataSourceStateMatchesResourceStateWithIgnores(
						"data.{{ $e.ResourceType $.Res.TerraformName }}.default",
						"{{ $e.ResourceType $.Res.TerraformName }}.{{ $e.PrimaryResourceId }}",
						map[string]struct{}{
			{{- range _, $p := $.Res.IgnoreReadProperties }}
							"{{ $p.TerraformLineage }}": struct{}{},
			{{- end }}
						}
					),
		{{- else }}
					acctest.CheckDataSourceStateMatchesResourceState("data.{{ $e.ResourceType $.Res.TerraformName }}.default", "{{ $e.ResourceType $.Res.TerraformName }}.{{ $e.PrimaryResourceId }}"),
		{{- end }}
```

melinath (changes requested):

mmv1/templates/terraform/examples/base_configs/datasource_test_file.go.tmpl
```diff
@@ -0,0 +1,108 @@
...
+					{{- if gt (len ($.Res.IgnoreReadPropertiesToString $e)) 0 }}
+						acctest.CheckDataSourceStateMatchesResourceStateWithIgnores(
+							"data.{{ $e.ResourceType $.Res.TerraformName }}.default",
+							"{{ $e.ResourceType $.Res.TerraformName }}.{{ $e.PrimaryResourceId }}",
+							map[string]struct{}{
+								{{- range $name, $val := $.Res.Parameters }}
+									{{- if $val.IgnoreRead }}
+								"{{ $name }}": {},
+									{{- end }}
+								{{- end }}
```
melinath: This isn't equivalent to what IgnoreReadPropertiesToString is doing, and as a result it's not generating properly in the downstream. We should make a new function called `IgnoreReadProperties` that returns the properties as a slice that we can just iterate over. `IgnoreReadPropertiesToString` should at a minimum become a wrapper around that, but should really just be replaced by it. I could go either way on whether the replacement happens in this PR or not.

mmv1/templates/terraform/examples/base_configs/datasource_test_file.go.tmpl
```diff
@@ -0,0 +1,108 @@
...
+	{{- if not $e.ExcludeImportTest }}
+			{
+				ResourceName:      "{{ $e.ResourceType $.Res.TerraformName }}.{{ $e.PrimaryResourceId }}",
+				ImportState:       true,
+				ImportStateVerify: true,
+		{{- if $.Res.IgnoreReadPropertiesToString $e }}
+				ImportStateVerifyIgnore: {{ $.Res.IgnoreReadPropertiesToString $e }},
+		{{- end }}
+			},
+	{{- end }}
```
melinath: I don't think we should have an import step here - import tests resource behavior, but this is a datasource test.

melinath:
I think this is probably working but it's hard to tell because the test runs were interleaved. Rerunning.

/gcbrun

mmv1/products/alloydb/Cluster.yaml
```diff
@@ -299,6 +299,7 @@ properties:
     description: |
       Annotations to allow client tools to store small amount of arbitrary data. This is distinct from labels. https://google.aip.dev/128
       An object containing a list of "key": value pairs. Example: { "name": "wrench", "mass": "1.3kg", "count": "3" }.
+    required: true
```
melinath:
```suggestion
```

melinath (changes requested):

PR #14878, opened 2025-08-18
Fix Lustre timeout failures

stephanecharite:

```release-note:enhancement
lustre: increased `google_lustre_instance` resource create timeout to 120m from 20m
```


mmv1/third_party/terraform/services/lustre/data_source_lustre_instance_test.go
```diff
@@ -44,7 +44,10 @@ resource "google_lustre_instance" "instance" {
   capacity_gib                = 18000
   network                     = data.google_compute_network.lustre-network.id
   gke_support_enabled         = false
-  per_unit_storage_throughput = 1000   
+  per_unit_storage_throughput = 1000
+  timeouts {
+	create = "120m"
```
rileykarson: The tests are complaining because we have a setting here but not in the DS- this way the test is checking this is wrong. We could either update the test to work around the issue, or consider baking the revised setting into the resource itself. Will users hit a 20m timeout or is the test case exceptional in some way? If the 99.9% latency is over 20m we probably want to kick the resource timeout up.

stephanecharite: ACK, our P99 seems to dance around 40 to 80 minutes. We expect that to increase slightly due to seem changes we have coming in the pipeline. So I updated the create timeout to be 120 minutes. Let's see if that addresses the issue

rileykarson (approved):

PR #14879, opened 2025-08-18
fix noop for optional and required fields

luckyswaminathan:
addressing https://github.com/GoogleCloudPlatform/magic-modules/pull/14598#discussion_r2263909913

```release-note:none
removing noop from optional and required datasource fields if there are none
```


kautikdk (commented):
This PR mistakenly got tagged with `service/storage` tag but it appears to be outside of the Storage service scope. Changes look good to me but I would request @roaks3 for full review of this PR.

PR #14880, opened 2025-08-18
altered datasource url usage

luckyswaminathan:
part of: https://github.com/hashicorp/terraform-provider-google/issues/12595

changed template to prioritize import format

```release-note:none
altered datasource generation to prioritize import_format
```


mmv1/templates/terraform/datasource.go.tmpl
```diff
@@ -73,11 +73,19 @@ func DataSource{{ .ResourceName -}}() *schema.Resource {
 func dataSource{{ $.ResourceName -}}Read(d *schema.ResourceData, meta interface{}) error {
   config := meta.(*transport_tpg.Config)
 
-  id, err := tpgresource.ReplaceVars{{if $.LegacyLongFormProject -}}ForId{{ end -}}(d, config, "{{$.SelfLinkUri}}{{$.ReadQueryParams}}")
-    if err != nil {
-        return err
-    }
-  d.SetId(id)
+  {{if $.ImportFormat}}
+    id, err := tpgresource.ReplaceVars{{if $.LegacyLongFormProject -}}ForId{{ end -}}(d, config, "{{index $.ImportFormat 0}}{{$.ReadQueryParams}}")
```
roaks3: The line that I see being used in the resource is:
```
id, err = tpgresource.ReplaceVars{{if $.LegacyLongFormProject -}}ForId{{ end -}}(d, config, "{{ $.IdFormat -}}")
```

Could that be used here instead? If I walk through the code, it looks like `IdFormat` will already default to the SelfLink as needed. I think it would allow you to remove the conditional (could also be used in your functions).

roaks3 (commented):
Resulting code LGTM, but just one comment on the implementation. Ideally we stay aligned to the existing paradigms as much as possible, which also limits the potential for creating new dependencies on yaml fields.

mmv1/api/resource.go
```diff
@@ -2109,7 +2109,13 @@ func (r Resource) ShouldDatasourceSetAnnotations() bool {
...
 func (r Resource) DatasourceRequiredFields() []string {
 	requiredFields := []string{}
-	uriParts := strings.Split(r.SelfLink, "/")
+	var uri string
+	if len(r.ImportFormat) > 0 {
+		uri = r.ImportFormat[0]
+	} else {
+		uri = r.SelfLink
+	}
+	uriParts := strings.Split(uri, "/")
```
roaks3: Per the other comment, could this be changed to `uriParts := strings.Split(r.IdFormat, "/")` to simplify?

mmv1/templates/terraform/datasource.go.tmpl
```diff
@@ -73,10 +73,8 @@ func DataSource{{ .ResourceName -}}() *schema.Resource {
 func dataSource{{ $.ResourceName -}}Read(d *schema.ResourceData, meta interface{}) error {
   config := meta.(*transport_tpg.Config)
 
-  id, err := tpgresource.ReplaceVars{{if $.LegacyLongFormProject -}}ForId{{ end -}}(d, config, "{{$.SelfLinkUri}}{{$.ReadQueryParams}}")
-    if err != nil {
-        return err
-    }
+  id, err := tpgresource.ReplaceVars{{if $.LegacyLongFormProject -}}ForId{{ end -}}(d, config, "{{ $.IdFormat -}}")
```
roaks3, line 76: I think we want to keep the error check, right? Otherwise it looks like we are ignoring errors.

luckyswaminathan: yeah, seems to have been randomly deleted by me

PR #14885, opened 2025-08-18
VPCFlowLogs - Add Organization Support

maayanbeltzer:

```release-note: enhancement
networkmanagement: added org to `google_network_management_vpc_flow_logs_config` resource (beta)
```

### DIFF
--- a/mmv1/products/networkmanagement/OrganizationVpcFlowLogsConfig.yaml
+++ b/mmv1/products/networkmanagement/OrganizationVpcFlowLogsConfig.yaml
@@ -16,6 +16,7 @@ name: 'OrganizationVpcFlowLogsConfig'
 description:
   VPC Flow Logs Config is a resource that lets you configure
   Flow Logs for Organization.
+min_version: beta
 id_format: 'organizations/{{organization}}/locations/{{location}}/vpcFlowLogsConfigs/{{vpc_flow_logs_config_id}}'
 base_url: 'organizations/{{organization}}/locations/{{location}}/vpcFlowLogsConfigs'
 self_link: 'organizations/{{organization}}/locations/{{location}}/vpcFlowLogsConfigs/{{vpc_flow_logs_config_id}}'


mmv1/products/networkmanagement/OrganizationVpcFlowLogsConfig.yaml
```diff
@@ -0,0 +1,149 @@
+# Copyright 2024 Google Inc.
```
rileykarson:
```suggestion
# Copyright 2025 Google Inc.
```

mmv1/products/networkmanagement/OrganizationVpcFlowLogsConfig.yaml
```diff
@@ -0,0 +1,149 @@
...
+  - name: 'flowSampling'
+    type: Double
+    default_from_api: true
+    description: |
+      Optional. The value of the field must be in (0, 1]. The sampling rate
+      of VPC Flow Logs where 1.0 means all collected logs are reported. Setting the
+      sampling rate to 0.0 is not allowed. If you want to disable VPC Flow Logs, use
+      the state field instead. Default value is 1.0.
+  - name: 'metadata'
+    type: String
```
rileykarson, line 110: If this will be fixed in value range indefinitely we could use the Enum type, which will strictly validate values. If any changes are expected/possible String is more appropriate, though.

mmv1/products/networkmanagement/OrganizationVpcFlowLogsConfig.yaml
```diff
@@ -0,0 +1,149 @@
...
+    description: |
+      Output only. The time the config was created.
+    output: true
+  - name: 'updateTime'
+    type: String
+    description: |
+      Output only. The time the config was updated.
+    output: true
+  - name: 'crossProjectMetadata'
+    type: String
```
rileykarson: Same as the possible enum above!

mmv1/products/networkmanagement/OrganizationVpcFlowLogsConfig.yaml
```diff
@@ -0,0 +1,149 @@
...
+      Identifier. Unique name of the configuration using the form:     `organizations/{org_id}/locations/global/vpcFlowLogsConfigs/{vpc_flow_logs_config_id}`
+    output: true
+  - name: 'description'
+    type: String
+    description: |
+      Optional. The user-supplied description of the VPC Flow Logs configuration. Maximum
+      of 512 characters.
+  - name: 'state'
+    type: String
+    default_from_api: true
```
rileykarson, line 90: If these defaults are fixed & will stay fixed, we could consider `Default` instead here & on other `default_from_api` fields. That'll create a more accurate `terraform plan` when the user doesn't specify a value (it just says "known after apply" for `default_from_api` fields) & avoid a deficiency where it's difficult/impossible to explicitly update to zero values for the type ("", 0, 0.0, etc)

mmv1/products/networkmanagement/OrganizationVpcFlowLogsConfig.yaml
maayanbeltzer: Thanks! we have the same API at the project level that uses string so I think it's better to keep them similar

mmv1/products/networkmanagement/OrganizationVpcFlowLogsConfig.yaml
maayanbeltzer: This field is relevant only at the org level so we can change it as you suggested, how can I change it to value range so we can use the enum?

mmv1/products/networkmanagement/OrganizationVpcFlowLogsConfig.yaml
maayanbeltzer: "If these defaults are fixed & will stay fixed" > what do you mean by will stay fixed? are you referring to the defaults value in the API?

rileykarson: Yep!

mmv1/products/networkmanagement/OrganizationVpcFlowLogsConfig.yaml
rileykarson: There should be an enum section under https://googlecloudplatform.github.io/magic-modules/develop/add-fields/!

mmv1/products/networkmanagement/OrganizationVpcFlowLogsConfig.yaml
maayanbeltzer: I don't think there's a plan to change the default values right now, but it's possible. I'm concerned that someday we'll be asked to, and there will be nothing to alert us that we need to update Terraform as well, so we better keep it without the default value itself.

mmv1/products/networkmanagement/OrganizationVpcFlowLogsConfig.yaml
maayanbeltzer: Thanks, done!

PR #14889, opened 2025-08-18
Breaking Change: Remove allow_fewer_zones_deployment from Memorystore and Redis Cluster

NA2047:
Fixes  https://github.com/hashicorp/terraform-provider-google/issues/24027

Deprecation for fields in main branch https://github.com/GoogleCloudPlatform/magic-modules/pull/14887

```release-note:breaking-change
memorystore: remove `allow_fewer_zones_deployment` field from `google_memorystore_instance` resource because it isn't user-configurable
```

```release-note:breaking-change
redis: remove `allow_fewer_zones_deployment` field from `google_redis_cluster` resource because it isn't user-configurable
```

### DIFF
--- a/mmv1/third_party/terraform/website/docs/guides/version_7_upgrade.html.markdown
+++ b/mmv1/third_party/terraform/website/docs/guides/version_7_upgrade.html.markdown
@@ -238,3 +238,11 @@ Remove `template.containers.depends_on` from your configuration after upgrade.
 The default value for `disable_on_destroy` has been changed to `false`. The previous default (`true`) created a risk of unintended service disruptions, as destroying a single `google_project_service` resource would disable the API for the entire project.
 
 Now, destroying the resource will only remove it from Terraform's state and leave the service enabled. To disable a service when the resource is destroyed, you must now make an explicit decision by setting `disable_on_destroy = true`.
+
+## Resource: `google_memorystore_instance`
+
+ `allow_fewer_zones_Deployment` has been removed as user configurable but will be set to true at creation
+
+## Resource: `google_redis_cluster`
+
+ `allow_fewer_zones_Deployment` has been removed as user configurable and will be set to true at creation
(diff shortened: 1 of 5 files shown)


mmv1/third_party/terraform/website/docs/guides/version_7_upgrade.html.markdown
```diff
@@ -238,3 +238,11 @@ Remove `template.containers.depends_on` from your configuration after upgrade.
 The default value for `disable_on_destroy` has been changed to `false`. The previous default (`true`) created a risk of unintended service disruptions, as destroying a single `google_project_service` resource would disable the API for the entire project.
 
 Now, destroying the resource will only remove it from Terraform's state and leave the service enabled. To disable a service when the resource is destroyed, you must now make an explicit decision by setting `disable_on_destroy = true`.
+
+## Resource: `google_memorystore_instance`
+
+ `allow_fewer_zones_Deployment` has been removed as user configurable but will be set to true at creation
```
melinath: "removed as user-configurable" could be interpreted to mean "it is now output-only" but actually we're just fully removing it. I'm not sure about the reason I've proposed; that's not generally a reason to remove a field. But I'm not sure if there's a better way to put it?

```suggestion
 `allow_fewer_zones_deployment` has been removed because it isn't user-configurable.
```

melinath (changes requested):

mmv1/third_party/terraform/website/docs/guides/version_7_upgrade.html.markdown
```diff
@@ -238,3 +238,11 @@ Remove `template.containers.depends_on` from your configuration after upgrade.
...
 
 Now, destroying the resource will only remove it from Terraform's state and leave the service enabled. To disable a service when the resource is destroyed, you must now make an explicit decision by setting `disable_on_destroy = true`.
+
+## Resource: `google_memorystore_instance`
+
+ `allow_fewer_zones_deployment` has been removed because it isn't user-configurable.
+
+## Resource: `google_redis_cluster`
+
+ `allow_fewer_zones_Deployment` has been removed because it isn't user-configurable.
```
melinath:
```suggestion
 `allow_fewer_zones_deployment` has been removed because it isn't user-configurable.
```

PR #14890, opened 2025-08-19
Add a new resource google_discovery_engine_data_connector

jialei-chen:

```release-note:new-resource
`google_discovery_engine_data_connector`
```


mmv1/products/discoveryengine/DataConnector.yaml
```diff
@@ -0,0 +1,372 @@
...
+parameters:
+  - name: 'location'
+    type: String
+    description: |
+      The geographic location where the data store should reside. The value can
+      only be one of "global", "us" and "eu".
+    url_param_only: true
+    required: true
+    immutable: true
+  - name: 'collectionIdUrlParam'
```
jialei-chen: I have `collectionIdUrlParam` and `collectionId` below. I think these 2 variables should be merged to one, but failed. I asked https://yaqs.corp.google.com/eng/q/2640687180824969216 but so far there is no reply.

mmv1/templates/terraform/examples/discoveryengine_dataconnector_jira_basic.tf.tmpl
```diff
@@ -0,0 +1,36 @@
...
+  location                  = "global"
+  collection_id             = "{{index $.Vars "collection_id"}}"
+  collection_id_url_param   = "{{index $.Vars "collection_id"}}"
+  collection_display_name   = "tf-test-dataconnector-jira"
+  data_connector {
+    data_source             = "jira"
+    params {
+        instance_id         = "33db20a3-dc45-4305-a505-d70b68599840"
+        instance_uri        = "https://vaissptbots1.atlassian.net/"
+        client_secret       = "ATOAGAf7J5OIiM_q-fmgSZjbh0lWxOkj3-P-KgGK5Q2yRmAJkEWqiFWuVK4fjfwtTSqFEBDE636D"
```
jialei-chen: `client_secret` is a confidential key that, along with the `client_id`, obtains access tokens and makes authenticated requests to Atlassian products like Jira, Confluence, etc in the context of [Atlassian's OAuth 2.0](https://developer.atlassian.com/cloud/oauth/) implementation.

`client_secret` is a user-provided field in plaintext. When adding a test case and explicitly specifying `client_id` and `client_secret`, I have to leak the OAuth credentials on Github. Is there any way that can avoid this on Terraform side?

roaks3: So for this, it actually might be a new check (I haven't seen it before), but it looks like it might just be checking the website. If that's the case, you might try using a `var` inside the `example` block instead of hardcoding the value, so that the website uses something benign like ABC123. This is something we'd probably want regardless, because many people will be copying the values in the example into their configs.

The broader question here is if we need to leak a valid secret/id. What privileges does this have, and what could a user or adversary do with it? Also, is it appropriate for arbitrary devs running tests to have this value used? If this is a secret value that is needed solely for our automated tests internally, we may need to consider an alternative approach that makes use of a value in secret manager.

mmv1/products/discoveryengine/DataConnector.yaml
roaks3: I'll take a closer look at this, but a good callout. Perhaps there is a way to get the interface we want.

mmv1/templates/terraform/examples/discoveryengine_dataconnector_jira_basic.tf.tmpl
jialei-chen: Thanks! As suggested, 
1. I changed the hardcoded `client_id` and `client_secret` values to `vars` block.
2. I found that the real/valid `client_id` and `client_secret` values are not the must-have for a successful CREATE API call. I changed them to arbitrary test values and verified locally. Thus, we can avoid leaking real/valid id&secret on Github.
3. Our team is considering about changing the user-provided id/secret in plaintext to secret manager. But it needs extra code changes, tests, production releases, etc, which will take some time. Since 2. above has already avoided secret leaking, we will not consider/include secret manager alternative approach in this PR. We can create a follow-up PR after our secret manager approach is done.

mmv1/products/discoveryengine/DataConnector.yaml
roaks3: Just responded directly on the yaqs. There are a few possible options here, and I expect we'll want to collapse this down to the one field.

mmv1/products/discoveryengine/DataConnector.yaml
```diff
@@ -0,0 +1,331 @@
...
+      Should be human readable, used to display collections in the Console
+      Dashboard. UTF-8 encoded string with limit of 1024 characters.
+    required: true
+    ignore_read: true
+  - name: 'dataConnector'
+    type: NestedObject
+    description: |
+      The DataConnector to initialize in the newly created Collection.
+    required: true
+    ignore_read: true
```
roaks3, line 114: I'm a little confused to see this field marked as `ignore_read: true`, but then many of the fields within in are `output: true`. Could you clarify the intent?

mmv1/products/discoveryengine/DataConnector.yaml
```diff
@@ -0,0 +1,331 @@
...
+      - name: 'kms_key_name'
+        type: String
+        description: |
+          The KMS key to be used to protect the DataStores managed by this connector.
+          Must be set for requests that need to comply with CMEK Org Policy
+          protections.
+          If this field is set and processed successfully, the DataStores created by
+          this connector will be protected by the KMS key.
+      - name: 'actionState'
+        type: Enum
```
roaks3: Nit: We normally model output-only enums as Strings, so that it is easier to support additional values without a client-side update. Would apply here and to a few other enums below.

mmv1/products/discoveryengine/product.yaml
```diff
@@ -17,5 +17,7 @@ display_name: 'Discovery Engine'
 versions:
   - name: 'ga'
     base_url: 'https://{{location}}-discoveryengine.googleapis.com/v1/'
+  - name: 'beta'
```
roaks3: I'll check with my team to see if there is any issue with doing this, but in the meantime could you confirm that it follows go/terraform-launch-stages (particularly making sure that the expectations around the API like breaking changes and SLOs matches the requirements of the beta provider)?

roaks3 (commented):
Overall LGTM

The two main issues I see are 1) resolving the collection_id issue, and 2) verifying that the alpha API is usable here.

mmv1/products/discoveryengine/DataConnector.yaml
jialei-chen: Thanks Ryan! 
Replied on the YAQS https://yaqs.corp.google.com/eng/q/2640687180824969216#n2
1. I retried locally and couldn't reproduce the same error. It seems this issue has been fixed automatically.
2. I removed `collection_id_url_param` to use `collection_id` only, and verified locally: [9363982](https://github.com/GoogleCloudPlatform/magic-modules/pull/14890/commits/9363982378582d7e54c6c00881d393699614391a).

mmv1/products/discoveryengine/DataConnector.yaml
jialei-chen: If DataConnector block isn't marked with "ignore_read: true", "update in-place" error will occur: https://paste.googleplex.com/5775323941699584#l=301

I think this "update in-place" error is generated because the SetUpDataConnector API itself has multiple Spanner writes: [an insert](http://google3/cloud/ml/discoveryengine/external_service/v1main/data_connector_service/set_up_data_connector_handler.cc;l=663;rcl=797490101) and some follow-up [updates](http://google3/cloud/ml/discoveryengine/external_service/v1main/data_connector_service/set_up_data_connector_handler.cc;l=735,978,1026;rcl=797490101).

Please see details: https://docs.google.com/document/d/1dSYo_8rYI6RxCKLF5xhjASo1slY7yocyaeN2oSJj9kA/edit?resourcekey=0-RGlx_UAOmwffKjLat42pqA&tab=t.fwppndro5dt2#heading=h.m6bjacmuj5zl

jialei-chen: This field `dataConnector` has a lot of output-only subfields. I listed them out and marked them as `output: true`. It's because IIRC given the contribution guide, we'd better list all output-only fields.

mmv1/products/discoveryengine/DataConnector.yaml
jialei-chen: Thanks, got it. Updated as suggested [9593735](https://github.com/GoogleCloudPlatform/magic-modules/pull/14890/commits/95937354ca507590e579594bb9a429fa079ece1f).

mmv1/products/discoveryengine/product.yaml
jialei-chen: Thanks! Sure, I will go check with my team and get back to you once I have answers.

mmv1/products/discoveryengine/DataConnector.yaml
roaks3: Well I guess I'm just a bit confused about what the behavior is then. Perhaps the cascading is doing something different then I expect.

My assumption is that since this field is marked as `ignore_read: true` that the other fields are not stored in state (ie. they are also `ignore_read: true`). It seems like the intent is for them to NOT be specified by the user, which I guess would result in them doing nothing?

mmv1/products/discoveryengine/product.yaml
jialei-chen: Hi Ryan, talked with slevenick@ about this issue today. He clarified that Terraform doesn't support and accept v1alpha APIs in google-beta provider. Then we have 2 options:
Option 1. Use v1 in google provider.
Option 2. Use v1alpha in google-private provider. This option is not publicly available via the hashicorp registry, so it's much harder for customers to use and very different for users. This can be a risk: I cannot tell if such difference is acceptable for customers.

Talked with my team, we will choose Option 1. I'm moving current public preview v1alpha APIs to GA v1, and cherrypicking the changes to speed up the production rollout. Once it's fully released, I will roll back my changes in [mmv1/products/discoveryengine/product.yaml](https://github.com/GoogleCloudPlatform/magic-modules/pull/14890/files/faf37a650637f952ff9768f3b3854f0b32307637#diff-2d3dd3454c79f4e3eb20a3bc8694d22b51aa529ab121247b5a453066a13cb360).

mmv1/products/discoveryengine/DataConnector.yaml
jialei-chen: > Well I guess I'm just a bit confused about what the behavior is then. Perhaps the cascading is doing something different then I expect.
 
* I asked the owner team, "In api handler, we first insert the connector entry in db and then could mutate the object depends on different scenarios/connector types.".
* I read the API handler implementation, in one successful `SetUpDataConnector` API call, no matter which scenario/connector type this API call handles, the following steps must be done sequentially:
  * a DataConnector item **create** [code point](http://google3/cloud/ml/discoveryengine/external_service/v1main/data_connector_service/set_up_data_connector_handler.cc;l=706-707;rcl=806713146)
  * some non-DataConnector (DataStore / Schema) items creation/update [code pointer](http://google3/cloud/ml/discoveryengine/external_service/v1main/data_connector_service/set_up_data_connector_handler.cc;l=723-769;rcl=806713146)
  * this DataConnector item **update** [code pointer](http://google3/cloud/ml/discoveryengine/external_service/v1main/data_connector_service/set_up_data_connector_handler.cc;l=778-779;rcl=806713146)
  * other follow-up DB writes
In this API call, there is an update after create. The resource is modified without recreating the database. I think this is why Terraform show a `~ update in-place` action in its plan.

> My assumption is that since this field is marked as ignore_read: true that the other fields are not stored in state (ie. they are also ignore_read: true). It seems like the intent is for them to NOT be specified by the user, which I guess would result in them doing nothing?

Given https://googlecloudplatform.github.io/magic-modules/reference/field/#ignore_read, "If true, the provider sets the field’s value in the resource state based only on the user’s configuration.". 
The fields with `ignore_read: true` are (1) stored in state, and (2) needed to be specified by the users.

mmv1/third_party/terraform/services/discoveryengine/resource_discovery_engine_data_connector_test.go
```diff
@@ -0,0 +1,123 @@
...
+`, context)
+}
+
+func testAccDiscoveryEngineDataConnector_discoveryengineDataconnectorJiraBasicExample_update(context map[string]interface{}) string {
+	return acctest.Nprintf(`
+resource "google_discovery_engine_data_connector" "jira-basic" {
+  provider                  = google-beta
+  location                  = "global"
+  collection_id             = "tf-test-collection-id%{random_suffix}"
+  data_connector {
```
jialei-chen: Hi @roaks3, I added an update test here.

The same context as https://yaqs.corp.google.com/eng/q/2640687180824969216:
* In create method `SetUpDataConnector`, body clause in the google.api.http annotation is "*". The `SetUpDataConnectorRequest` object itself is the HTTP body. It means that an entire `data_connector` block needs to be passed when calling the API.
* In update method `UpdateDataConnector`, the `data_connector` object is the HTTP body. It means that API doesn't accept an entire `data_connector` block but the fields of a `data_connector`. 

This HTTP body difference is blocking the update test case:
* If not defining `data_connector` block, update test fails for "Insufficient data_connector blocks" error: https://paste.googleplex.com/6220739380379648#l=644
* If defining `data_connector` block, update test fails for INVALID_ARGUMENT error: https://paste.googleplex.com/4974961567596544#l=821

Is there any way to solve this issue, such as custom code?

mmv1/products/discoveryengine/product.yaml
jialei-chen: Solved at [e6fd2fb](https://github.com/GoogleCloudPlatform/magic-modules/pull/14890/commits/e6fd2fbbbadea71d1bd7a3ea082d761ce4d7396f).

mmv1/third_party/terraform/services/discoveryengine/resource_discovery_engine_data_connector_test.go
jialei-chen: As discussed in chat group, we changed API. And we updated this PR correspondingly to solve the issue at [5e087f9](https://github.com/GoogleCloudPlatform/magic-modules/pull/14890/commits/5e087f9617558264c10fdb014e9e8d9c3249dc77).

jialei-chen:
Hi Ryan @roaks3, I solved all known issues. Please see my replies in the open comments above for details. 

There is a new error generating from the update test case: Update needs to wait until create operation changes resource state from `INITIALIZING` to its next state `ACTIVE`: [[Error message](https://storage.cloud.google.com/ci-vcr-logs/beta/refs/heads/auto-pr-14890/artifacts/34af63c8-82ee-4428-81e9-43e36251a97a/build-log/recording_build/TestAccDiscoveryEngineDataConnector_discoveryengineDataconnectorJiraBasicExample_update_recording_test.log)] [[Debug log](https://storage.cloud.google.com/ci-vcr-logs/beta/refs/heads/auto-pr-14890/artifacts/34af63c8-82ee-4428-81e9-43e36251a97a/recording/TestAccDiscoveryEngineDataConnector_discoveryengineDataconnectorJiraBasicExample_update.log)]. It may take ~1 hour. 
An update test is an acceptance test that creates the target resource and then makes updates to fields that are updatable. We return "done" immediately at the create step, the LRO doesn't really track the real progress of the data connector creation workflow (there is also an async ucaip pipeline for creating data connector). Even if we are adding arbitrary sleep after creation, the update test case would likely be flaky.

As for the possible solutions:
Option1. Potential workaround in Terraform using [time_sleep](https://screenshot.googleplex.com/BQRXNJtYcrCKjJ9): I searched in the code base, the existing examples set the timer as multiple minutes. Is it okay for us to set the timer up to 1 hour? 
[Preferred] Option2: Remove the update test to unblock PR submission to unblock our customer ASAP. The resource is still mutable, but just has no update test case in this PR.
Do you have any concerns about Option2? If not, I will move forward correspondingly and then the PR will be ready.

mmv1/products/discoveryengine/DataConnector.yaml
roaks3: Ok, I think that makes sense, but I then still wonder what `output: true` is being used for on the child fields. Are they able to be populated by the API even though the parent is `ignore_read: true`?

roaks3: Oh actually, looks like this isn't actually present in the current version. So nevermind :)

roaks3:

roaks3:
hey @NickElliot , I'm going to be OOO so had to re-assign this. My last comment above should basically cover everything, so I expect this will be relatively easy to get merged once the items there are covered.

jialei-chen:
> Just to confirm because I didn't see these modified in the update test, the following fields should be mutable?
> 
> * `collectionDisplayName`
> * `dataSource`
> * `params`
> * `jsonParams`
> * `refreshInterval`
> * `entities`
> * `kmsKeyName`
> * `staticIpEnabled`
> 
> Everything else LGTM (as mentioned over chat, I would use a time_sleep for the test and `t.Skip()` it for now with a comment that it's skipped due to duration and flakiness)

Hi @NickElliot, 
1. Mutable fields or not: Referring to our [proto definition](http://google3/google/cloud/discoveryengine/v1main/data_connector_service.proto;l=1143-1164;rcl=810506992)
* `collectionDisplayName`: marked as immutable
* `dataSource`: marked as immutable
* either `params` or `jsonParams`: mutable & modified in update test
* `refreshInterval`: mutable & modified in update test
* `entities`: marked as immutable
* `kmsKeyName`: marked as immutable
* `staticIpEnabled`: marked as immutable

2. Update test: Added (1) 1-hour time_sleep and (2) `t.Skip()`as suggested.

3. Create test `TestAccDiscoveryEngineDataConnector_discoveryengineDataconnectorJiraBasicExample`: tested and passed locally. All other test cases (even if they fail) are unrelated to this PR and can be exluced.

NickElliot (approved):
all comments from previous review resolved and one last issue with `json_params` from me, LGTM!

PR #14892, opened 2025-08-19
Memcache fb deletion protection

rajeshgupthar:
Part of b/368232860

```release-note:enhancement
memcache: added `deletion_protection` field to `memcache_instance` to make deleting them require an explicit intent.`memcache_instance` resources now cannot be destroyed unless `deletion_protection = false` is set for the resource.
```

Requesting mandatory review from @aniket-gupta


trodge:
@aniket-gupta @rajeshgupthar I'm unclear about the status of this PR, as it was never taken out of draft mode. It looks like it will need to target the 8.0 major release if the intention is to change the default behavior in a non-backwards compatible way.

PR #14893, opened 2025-08-19
Added support for Deletion protection flag for Metastore service

rajeshgupthar:

```release-note:enhancement

```

Mandtory review from @aniket-gupta

### DIFF
--- a/mmv1/products/metastore/Service.yaml
+++ b/mmv1/products/metastore/Service.yaml
@@ -50,13 +50,16 @@ iam_policy:
   import_format:
     - 'projects/{{project}}/locations/{{location}}/services/{{service_id}}'
     - '{{service_id}}'
-custom_code:
+custom_code: !ruby/object:Provider::Terraform::CustomCode
+  pre_delete: 'templates/terraform/pre_delete/metastore_service.go.tmpl'
 examples:
   - name: 'dataproc_metastore_service_basic'
     primary_resource_id: 'default'
     primary_resource_name: 'fmt.Sprintf("tf-test-metastore-srv%s", context["random_suffix"])'
     vars:
       metastore_service_name: 'metastore-srv'
+    ignore_read_extra:
+      - 'deletion_protection'
   - name: 'dataproc_metastore_service_deletion_protection'
     primary_resource_id: 'default'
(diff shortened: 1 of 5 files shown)


aniket-gupta (changes requested):
This should done with CCFE phase 2 and FT changes

mmv1/products/metastore/Service.yaml
```diff
@@ -314,10 +317,13 @@ properties:
...
         required: true
-  - name: 'deletionProtection'
+  - name: 'deletion_protection'
     type: Boolean
     description: |
-      Indicates if the dataproc metastore should be protected against accidental deletions.
+      Indicates if the dataproc metastore should be protected against accidental deletions.Defaults to false.
+      When the field is set to true in Terraform state, a `terraform apply`
+      or `terraform destroy` that would delete the service will fail.
+    default_value: false
```
trodge: Adding a default value of false would not be a breaking change. Only making it default to true would be breaking.

As far as I know, the default value is already false.

trodge (changes requested):
This test failure:
```
2025-08-20T18:20:17.648Z [ERROR] sdk.helper_resource: Unexpected error: test_terraform_path=/bin/terraform test_working_directory=/tmp/plugintest385635592 test_step_number=1 test_name=TestAccDataprocMetastoreService_deletionProtection
  error=
  | After applying this test step, the plan was not empty.
  | stdout:
  | 
  | 
  |   ~ update in-place
  | 
  | 
  |   # google_dataproc_metastore_service.my_metastore will be updated in-place
  |   ~ resource "google_dataproc_metastore_service" "my_metastore" {
  |       ~ deletion_protection = false -> true
  |         id                  = "projects/ci-test-project-188019/locations/us-central1/services/tf-test-metastore-011lmmkr0u"
  |         name                = "projects/ci-test-project-188019/locations/us-central1/services/tf-test-metastore-011lmmkr0u"
  |         # (17 unchanged attributes hidden)
  | 
  |         # (2 unchanged blocks hidden)
  |     }
  | 
  | Plan: 0 to add, 1 to change, 0 to destroy.
  ```
  
  Suggests that the field is being read from the API as `false` and terraform expects it to be `true`. This would make sense if the API doesn't return the field at all. Should it have been added as a virtual field instead?

PR #14899, opened 2025-08-19
Add auto_ipam_config to container_cluster

translucens:

```release-note:enhancement
container: added `auto_ipam_config` field to `google_container_cluster` resource
```

### DIFF
--- a/mmv1/third_party/terraform/services/container/resource_container_cluster.go.tmpl
+++ b/mmv1/third_party/terraform/services/container/resource_container_cluster.go.tmpl
@@ -1862,7 +1862,7 @@ func ResourceContainerCluster() *schema.Resource {
 							Type:        schema.TypeList,
 							MaxItems:    1,
 							Optional:    true,
-							Computed:    true,
+							DiffSuppressFunc: suppressDiffAutoIpamConfig,
 							Description: `Configuration for auto IP address management.`,
 							Elem: &schema.Resource{
 								Schema: map[string]*schema.Schema{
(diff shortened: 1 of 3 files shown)


translucens:
I'll add `DiffSuppressFunc` to suppress the diff

```
                   - auto_ipam_config {
                      - enabled = false -> null
                    }
```

mmv1/third_party/terraform/services/container/resource_container_cluster.go.tmpl
```diff
@@ -1856,9 +1856,23 @@ func ResourceContainerCluster() *schema.Resource {
...
 								},
-
 							},
-
+						},
+						"auto_ipam_config": {
+							Type:        schema.TypeList,
+							MaxItems:    1,
+							Optional:    true,
+							Computed:    true,
```
zli82016: What is the reason to add `Computed:    true`?

Will the API return a default value for this field in the API response if the field is not provided in the API CREATE request?

zli82016: I checked the logs that the API returns a default value when no value is provided for this field `auto_ipam_config`.

mmv1/third_party/terraform/services/container/resource_container_cluster.go.tmpl
```diff
@@ -1856,9 +1856,23 @@ func ResourceContainerCluster() *schema.Resource {
...
 								},
-
 							},
-
+						},
+						"auto_ipam_config": {
+							Type:        schema.TypeList,
+							MaxItems:    1,
+							Optional:    true,
+							DiffSuppressFunc: suppressDiffAutoIpamConfig,
```
zli82016, line 1865: I checked the logs that the API returns a default value when no value is provided for this field auto_ipam_config.

Can you try to add back `Computed: true` and remove the DiffSuppressFunc? 
The failed tests such as `TestAccContainerCluster_additional_ip_ranges_config_on_create` should pass with `Computed: true`

Thanks.

zli82016 (changes requested):

PR #14906, opened 2025-08-19
write-only: use `CamelizeProperty` for the expander property variable name instead of `ApiName`

ramonvermeulen:
@melinath 

Moved to separate PR for easier review after the request in: https://github.com/GoogleCloudPlatform/magic-modules/pull/14857#pullrequestreview-3133293027

```release-note:none
write-only: use `CamelizeProperty` for the expander property variable name instead of `ApiName`
```


melinath:

PR #14910, opened 2025-08-20
Adds UnitKind provider for SaasRuntime

g-dreva:
Adds UnitKind resource provider for SaasRuntime API. (beta)

```release-note:new-resource
`google_saas_runtime_unit_kind`
```

### DIFF
--- a/mmv1/products/saasservicemgmt/UnitKind.yaml
+++ b/mmv1/products/saasservicemgmt/UnitKind.yaml
@@ -1,0 +1,213 @@
+# Copyright 2024 Google Inc.
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+---
+name: UnitKind
+description: A UnitKind serves as a template or type definition for a group of Units. Units that belong to the same UnitKind are managed together, follow the same release model, and are typically updated together through rollouts.
+base_url: projects/{{project}}/locations/{{location}}/unitKinds
+update_mask: true
+self_link: projects/{{project}}/locations/{{location}}/unitKinds/{{unit_kind_id}}
+create_url: projects/{{project}}/locations/{{location}}/unitKinds?unitKindId={{unit_kind_id}}
+update_verb: PATCH
+id_format: projects/{{project}}/locations/{{location}}/unitKinds/{{unit_kind_id}}
+import_format:
+  - projects/{{project}}/locations/{{location}}/unitKinds/{{unit_kind_id}}
+min_version: beta
+examples:
+  - name: saas_runtime_unit_kind_basic
+    primary_resource_id: "example"
+    min_version: 'beta'
+    vars:
+      saas_name: example-saas
+      cluster_unitkind_name: cluster-unitkind
+      app_unitkind_name: app-unitkind
+    bootstrap_iam:
+      - member: "serviceAccount:service-{project_number}@gcp-sa-saasservicemgmt.iam.gserviceaccount.com"
+        role: "roles/saasservicemgmt.serviceAgent"
+autogen_async: false
+autogen_status: VW5pdEtpbmQ=
+parameters:
+  - name: location
+    type: String
+    description: Resource ID segment making up resource `name`. It identifies the resource within its parent collection as described in https://google.aip.dev/122.
+    immutable: true
+    url_param_only: true
+    required: true
+  - name: unitKindId
+    type: String
+    description: The ID value for the new unit kind.
+    immutable: true
+    url_param_only: true
+    required: true
+properties:
+  - name: annotations
+    type: KeyValueAnnotations
+    description: |-
+      Annotations is an unstructured key-value map stored with a resource that
+      may be set by external tools to store and retrieve arbitrary metadata.
+      They are not queryable and should be preserved when modifying objects.
+
+      More info: https://kubernetes.io/docs/user-guide/annotations
+  - name: createTime
+    type: String
+    description: The timestamp when the resource was created.
+    output: true
+  - name: defaultRelease
+    type: String
+    description: |-
+      A reference to the Release object to use as default for creating new units
+      of this UnitKind (optional).
+
+      If not specified, a new unit must explicitly reference which release to use
+      for its creation.
+  - name: dependencies
+    type: Array
+    description: |-
+      List of other unit kinds that this release will depend on. Dependencies
+      will be automatically provisioned if not found. Maximum 10.
+    immutable: true
+    item_type:
+      type: NestedObject
+      properties:
+        - name: alias
+          type: String
+          description: An alias for the dependency. Used for input variable mapping.
+          required: true
+        - name: unitKind
+          type: String
+          description: The unit kind of the dependency.
+          immutable: true
+          required: true
+  - name: etag
+    type: String
+    description: |-
+      An opaque value that uniquely identifies a version or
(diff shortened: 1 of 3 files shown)


g-dreva:
Regarding the Missing test report:

Default release is a field that refers to a resource that we haven't added a provider for, and I'm following the guideline to add one resource at a time. I can add both resources in this PR if preferred so I can have full test coverage. Both resources depend on each other but the expected flow is 1) Create UnitKind, 2) Create Release with required UnitKind association, 3) Optionally set the default release on UnitKind to the release.

For OutputVariableMappings, although both input and output variable mappings have the same type, in practice output only uses "from" mapping as I've added in the test. The "to" mapping in output variables is not used.

hao-nan-li:
> Regarding the Missing test report:
> 
> Default release is a field that refers to a resource that we haven't added a provider for, and I'm following the guideline to add one resource at a time. I can add both resources in this PR if preferred so I can have full test coverage. Both resources depend on each other but the expected flow is 1) Create UnitKind, 2) Create Release with required UnitKind association, 3) Optionally set the default release on UnitKind to the release.
> 
> For OutputVariableMappings, although both input and output variable mappings have the same type, in practice output only uses "from" mapping as I've added in the test. The "to" mapping in output variables is not used.

Thanks for the explanation, if the `Default release` field refers to something not added, could we temporarily remove that field from this PR?

g-dreva:
> > Regarding the Missing test report:
> > Default release is a field that refers to a resource that we haven't added a provider for, and I'm following the guideline to add one resource at a time. I can add both resources in this PR if preferred so I can have full test coverage. Both resources depend on each other but the expected flow is 1) Create UnitKind, 2) Create Release with required UnitKind association, 3) Optionally set the default release on UnitKind to the release.
> > For OutputVariableMappings, although both input and output variable mappings have the same type, in practice output only uses "from" mapping as I've added in the test. The "to" mapping in output variables is not used.
> 
> Thanks for the explanation, if the `Default release` field refers to something not added, could we temporarily remove that field from this PR?


The issue I see is that the UnitKind provider without the "default release" field is much less useful.  If it's critical to test this field, and if it's ok with you, I would prefer to add both providers in this PR than remove a critical field from this resource and add it later. 

To explain why: Releases are typically created as part of a CI/CD pipeline outside of Terraform. Then you might update the UnitKind terraform configuration to point to a particular release ID, without having the Release itself managed through Terraform. So UnitKind without default release will not satisfy customer needs.

hao-nan-li:
> > > Regarding the Missing test report:
> > > Default release is a field that refers to a resource that we haven't added a provider for, and I'm following the guideline to add one resource at a time. I can add both resources in this PR if preferred so I can have full test coverage. Both resources depend on each other but the expected flow is 1) Create UnitKind, 2) Create Release with required UnitKind association, 3) Optionally set the default release on UnitKind to the release.
> > > For OutputVariableMappings, although both input and output variable mappings have the same type, in practice output only uses "from" mapping as I've added in the test. The "to" mapping in output variables is not used.
> > 
> > 
> > Thanks for the explanation, if the `Default release` field refers to something not added, could we temporarily remove that field from this PR?
> 
> The issue I see is that the UnitKind provider without the "default release" field is much less useful. If it's critical to test this field, and if it's ok with you, I would prefer to add both providers in this PR than remove a critical field from this resource and add it later.
> 
> To explain why: Releases are typically created as part of a CI/CD pipeline outside of Terraform. Then you might update the UnitKind terraform configuration to point to a particular release ID, without having the Release itself managed through Terraform. So UnitKind without default release will not satisfy customer needs.

I'm not sure I can fully understand what you said, but go ahead if you believe adding both providers is beneficial.

g-dreva:
While adding the Release provider, I ran into an issue because it requires an Artifact Registry Image that is built in a very specific way. I found https://yaqs.corp.google.com/eng/q/6819086453071413248 which mentions that you can assist with this.

I tried creating the resource in our own testing project and granting the Terraform test project service agent the necessary permissions, but I ran into the issue documented in go/terraform-team-testing#other-dependencies:
> google.com org-level constraints on providing access to resources outside the google.com organization

It's also not possible to create this resource through Terraform, so I can't have it set up as part of the test.

If I'm understanding correctly, there are some shared testing projects (like tf-static-<>) that are used to host static resources. Would it be possible for me to get access to one of those to create the images myself? 

Or if you're able to do it, can you create the images directly in the testing projects? The instructions are here:go/ez-experience:gcloud-cli#create-blueprints.
To make the commands work, you just need to make sure go/ez-experience:gcloud-cli#set-up-variables has the correct values, for example: https://screenshot.googleplex.com/8HBvCJqyWB6HoyF.

hao-nan-li:
> While adding the Release provider, I ran into an issue because it requires an Artifact Registry Image that is built in a very specific way. I found https://yaqs.corp.google.com/eng/q/6819086453071413248 which mentions that you can assist with this.
> 
> I tried creating the resource in our own testing project and granting the Terraform test project service agent the necessary permissions, but I ran into the issue documented in go/terraform-team-testing#other-dependencies:
> 
> > google.com org-level constraints on providing access to resources outside the google.com organization
> 
> It's also not possible to create this resource through Terraform, so I can't have it set up as part of the test.
> 
> If I'm understanding correctly, there are some shared testing projects (like tf-static-<>) that are used to host static resources. Would it be possible for me to get access to one of those to create the images myself?
> 
> Or if you're able to do it, can you create the images directly in the testing projects? The instructions are here:go/ez-experience:gcloud-cli#create-blueprints. To make the commands work, you just need to make sure go/ez-experience:gcloud-cli#set-up-variables has the correct values, for example: https://screenshot.googleplex.com/8HBvCJqyWB6HoyF.

Do you mind setting up a quick 15 minutes with me? Just to make sure we are on the same page.

mmv1/products/saasservicemgmt/UnitKind.yaml
```diff
@@ -0,0 +1,207 @@
...
+  - name: createTime
+    type: String
+    description: The timestamp when the resource was created.
+    output: true
+  - name: dependencies
+    type: Array
+    description: |-
+      List of other unit kinds that this release will depend on. Dependencies
+      will be automatically provisioned if not found. Maximum 10.
+    immutable: true
```
hao-nan-li, line 72: If the dependencies will be provisioned if not found, I wonder if `default_from_api` would be the most appropriate attribute here. https://googlecloudplatform.github.io/magic-modules/develop/diffs/#default. Please take a look if that's the case.

g-dreva: No, that would not be appropriate here.

That is describing the purpose of the field, which is that if you define a dependency on another UnitKind, and an instance of that UnitKind is not found when you try to provision an instance of this one, then provisioning will cascade and also provision an instance of your dependency.

hao-nan-li: Thanks for the explanation

hao-nan-li:
One thing I want to make sure before merging this: the added tests will run on a different project, do I need to re-do what we've done to create the blueprints on the other project?

g-dreva:
> One thing I want to make sure before merging this: the added tests will run on a different project, do I need to re-do what we've done to create the blueprints on the other project?

Yes, and actually I need to enable the API and create the P4SA in each testing project as well: https://yaqs.corp.google.com/eng/q/5003599241566748672

The commands needed before creating the blueprint are:
gcloud services enable saasservicemgmt.googleapis.com --project TEST-PROJECT
gcloud beta saas-runtime saas create tftest-saas --location=us-central1 --locations=name=us-central1 --project TEST-PROJECT

hao-nan-li:
> > One thing I want to make sure before merging this: the added tests will run on a different project, do I need to re-do what we've done to create the blueprints on the other project?
> 
> Yes, and actually I need to enable the API and create the P4SA in each testing project as well: https://yaqs.corp.google.com/eng/q/5003599241566748672
> 
> The commands needed before creating the blueprint are: gcloud services enable saasservicemgmt.googleapis.com --project TEST-PROJECT gcloud beta saas-runtime saas create tftest-saas --location=us-central1 --locations=name=us-central1 --project TEST-PROJECT

commands applied on nightly beta test project

hao-nan-li (approved):

PR #14913, opened 2025-08-20
secretmanager: set default for `secret_data_wo_version` in `google_secret_manager_secret_version` back to `0`

ramonvermeulen:
@melinath 

Closes https://github.com/hashicorp/terraform-provider-google/issues/24044

Still needs to be rebased onto `v6.49.0`, however, I am not sure what the normal flow is for hotfixes within magic-modules. Couldn't find a branch nor tag for this to start the branch from.

Ideally requires a new patch release.

```release-note:bug
secretmanager: set default for `secret_data_wo_version` in `google_secret_manager_secret_version` back to `0`
```

### DIFF
--- a/mmv1/api/resource.go
+++ b/mmv1/api/resource.go
@@ -786,9 +786,9 @@ func buildWriteOnlyVersionField(name string, originalField *Type, writeOnlyField
 		propertyWithClientSide(true),
 	}
 
-	// TODO(anymore) remove with next major release + add migration guide
+	// TODO: remove this branch with the next major release + add migration guide
 	if name == "secretDataWoVersion" {
-		options = append(options, propertyWithDefaultValue("0"))
+		options = append(options, propertyWithDefaultValue(0))
 	}
 
 	return NewProperty(name, name, options)
--- a/mmv1/api/resource.go
+++ b/mmv1/api/resource.go
@@ -737,7 +737,7 @@ func (r Resource) GetIdentity() []*Type {
 	})
 }
 
-func buildWriteOnlyField(name string, versionFieldName string, originalField *Type, originalFieldLineage string) *Type {
+func buildWriteOnlyField(resourceName string, name string, versionFieldName string, originalField *Type, originalFieldLineage string) *Type {
 	description := fmt.Sprintf("%s Note: This property is write-only and will not be read from the API. For more info see [updating write-only attributes](/docs/providers/google/guides/using_write_only_attributes.html#updating-write-only-attributes)", originalField.Description)
 	fieldPathOriginalField := originalFieldLineage
 	fieldPathCurrentField := strings.ReplaceAll(originalFieldLineage, google.Underscore(originalField.Name), google.Underscore(name))
@@ -759,7 +759,7 @@ func buildWriteOnlyField(name string, versionFieldName string, originalField *Ty
 
 	// TODO: remove this branch with the next major release + add migration guide
 	// In the future (next major) the secret_data_wo_version is required and should always be set explicitly (so no default, required with the _wo field)
-	if name != "secretDataWo" {
+	if name != "secretDataWo" && resourceName != "SecretVersion" {
 		options = append(options, propertyWithRequiredWith([]string{requiredWith}))
 	}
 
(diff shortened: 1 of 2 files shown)


ramonvermeulen:
Ideally we use something more robust than:

```go
	if name == "secretDataWoVersion" {
```

and

```go
	if name != "secretDataWo" {
```

**EDIT:** Added extra check on resourceName in https://github.com/GoogleCloudPlatform/magic-modules/pull/14913/commits/4694422d5d1e288f7527367e2210de45a9acab1b (for the scenario where 2 resources share the same write only argument names) and reran below tests with same result.

<hr>

Because I want to be sure this actually fixes the problem, I tried the following:

<details>

**1. Applied the following config, via the provider version `6.48.0`:**

Versions used:
```
➜ terraform version
Terraform v1.12.2
on darwin_arm64
+ provider registry.terraform.io/hashicorp/google v6.48.0
```

Config:
```terraform
locals {
  project_id = "my-gcp-project"
}

provider "google" {
  project = local.project_id
  version = "6.48.0"
}

resource "google_secret_manager_secret" "secret" {
  secret_id = "test-bug"

  replication {
    auto {

    }
  }
}

resource "google_secret_manager_secret_version" "version" {
  secret      = google_secret_manager_secret.secret.id
  secret_data_wo = "foo"
  # no version set so uses default ( 0 )
}
```

TF state after apply:
image

<hr>

**2. Build the provider binary based on this branch, and then re-applied using `TF_CLI_CONFIG_FILE` with the following config:**

Config:
```terraform
locals {
  project_id = "my-gcp-project"
}

provider "google" {
  project = local.project_id
}

resource "google_secret_manager_secret" "secret" {
  secret_id = "test-bug"

  replication {
    auto {

    }
  }
}

resource "google_secret_manager_secret_version" "version" {
  secret      = google_secret_manager_secret.secret.id
  secret_data_wo = "foo"
  # no version set so uses default ( 0 )
}
```

```
~/projects/personal/terraform_test_suite                                                                                                                                                                                 
❯ TF_CLI_CONFIG_FILE="$(pwd)/tf-dev-override.tfrc" terraform apply

╷
│ 
│ 
╵
google_secret_manager_secret.secret: Refreshing state... [id=projects/magic-modules-dev/secrets/test-bug]
google_secret_manager_secret_version.version: Refreshing state... [id=projects/1079685878740/secrets/test-bug/versions/1]

```
</details>

ScottSuarez:

melinath:
Thanks for flagging this issue yesterday! We ended up doing a full revert of the previous PR just to be safe, so the next thing to do would be to "revert the revert" and create a new PR that addresses the breaking change issue as well. I think the best option is likely to be a state migration - the best documentation we have about that is here: https://github.com/hashicorp/terraform-provider-google/issues/15584

Closing this PR for now.

ramonvermeulen:
> Thanks for flagging this issue yesterday! We ended up doing a full revert of the previous PR just to be safe, so the next thing to do would be to "revert the revert" and create a new PR that addresses the breaking change issue as well. I think the best option is likely to be a state migration - the best documentation we have about that is here: [hashicorp/terraform-provider-google#15584](https://github.com/hashicorp/terraform-provider-google/issues/15584)
> 
> Closing this PR for now.

Check, and that makes sense. With a full revert you can at least assure with full certainty that it will resolve the issue, and it is quicker. 

Sorry for causing this in the first place, felt bad since quite some people were impacted by this (given `google_secret_manager_secret_version` is a widely used resource). I think I focussed too much on testing the int to string change for the `wo_version` that somehow I missed the breaking change of removing the default.

Anyways, I will dive into state migrations and re-open a PR for the write-only generation.

ScottSuarez:
Hey @ramonvermeulen, thats okay. Things happen. I think we could of also done some things better on our end to detect these types of things so we can make changes like this more safely. Thanks for your work.

melinath:
+1, this isn't on you. We had multiple folks review your PR. I even noticed the problem but then failed to make sure we addressed it in a backwards-compatible way. This has been a very complex project - thanks again for taking it on & sticking with it!

PR #14917, opened 2025-08-20
Revert " provider: eliminated the need to manually add `*_wo` and `*_wo_version` for write-only properties"

ScottSuarez:
Reverts GoogleCloudPlatform/magic-modules#14230

```release-note:bug
secretmanager: reverted breaking type and default value changes for secret manager
```


melinath:
Why make an additional change to bigquerydatatransfer Config? I'd be a little concerned about expanding this beyond a strict revert.

melinath (approved):

PR #14923, opened 2025-08-20
add `TF_ACC_REFRESH_AFTER_APPLY` envvar for `plugin-testing` - `1.13.3` support

BBBmau:
this resolves a version block of being able to bump to the latest version of `plugin-testing` due to https://github.com/hashicorp/terraform-plugin-testing/issues/327

it'll allow us to test `resource_identity` without diffs from occurring as stated in the original issue opened in `plugin-testing`

```release-note:none

```


BBBmau:
> Is there any chance that we could get the bug fixed rather than working around it like this?

prior discussions with @bbasata led to there not being a clear solution for this bug to be resolved without a refactoring of how we generate tests

rileykarson:
Related PR in https://github.com/GoogleCloudPlatform/magic-modules/pull/14973

BBBmau:
anything else we need to get this approved? @melinath

melinath:
@BBBmau I think it's just pending an answer about whether it's possible to just fix the bug in terraform-plugin-testing.

BBBmau:
> @BBBmau I think it's just pending an answer about whether it's possible to just fix the bug in terraform-plugin-testing.

yes it should be possible just requires time on my part to investigate further

PR #14928, opened 2025-08-21
fix permadiff when specify IPv6 CIDR in a firewall

translucens:
Fixes https://github.com/hashicorp/terraform-provider-google/issues/23169

```release-note:bug
compute: fixed IPv6 CIDR perma-diff in `google_compute_firewall`
```

### DIFF
--- a/mmv1/templates/terraform/constants/firewall.tmpl
+++ b/mmv1/templates/terraform/constants/firewall.tmpl
@@ -107,6 +107,30 @@ func diffSuppressSourceRanges(k, old, new string, d *schema.ResourceData) bool {
 			return true
 		}
 	}
-	// For any other source_ranges value diff, don't suppress
-	return false
+
+	return diffSuppressCIDR("source_ranges", d)
+}
+
+func diffSuppressDestinationRanges(k, old, new string, d *schema.ResourceData) bool {
+	return diffSuppressCIDR("destination_ranges", d)
+}
+
+func diffSuppressCIDR (key string, d *schema.ResourceData) bool {
+	o, n := d.GetChange(key)
+	oldRanges := o.(*schema.Set).List()
+	newRanges := n.(*schema.Set).List()
+
+	if len(oldRanges) != len(newRanges) {
+		return false
+	}
+
+	for i := 0; i < len(oldRanges); i++ {
+		oldCIDR := netip.MustParsePrefix(oldRanges[i].(string))
+		newCIDR := netip.MustParsePrefix(newRanges[i].(string))
+
+		if oldCIDR != newCIDR {
+			return false
+		}
+	}
+	return true
 }
--- a/mmv1/products/compute/Firewall.yaml
+++ b/mmv1/products/compute/Firewall.yaml
@@ -172,6 +172,7 @@ properties:
       must be expressed in CIDR format. IPv4 or IPv6 ranges are supported.
     is_set: true
     default_from_api: true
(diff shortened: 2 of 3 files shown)


mmv1/templates/terraform/constants/firewall.tmpl
```diff
@@ -107,6 +107,30 @@ func diffSuppressSourceRanges(k, old, new string, d *schema.ResourceData) bool {
...
+	}
+
+	for i := 0; i < len(oldRanges); i++ {
+		oldCIDR := netip.MustParsePrefix(oldRanges[i].(string))
+		newCIDR := netip.MustParsePrefix(newRanges[i].(string))
+
+		if oldCIDR != newCIDR {
+			return false
+		}
+	}
```
cjgibson, line 132: [Comment]

Not a contributor, just happened to hit this issue and excited to find someone already working on a fix 😅 

I noticed they use common helpers for this elsewhere: https://github.com/GoogleCloudPlatform/magic-modules/blob/fd62c38302087df4545b6a3496df8aef38a9e02a/tpgtools/ignored_handwritten/common_diff_suppress.go#L58-L64

Might want to consider expanding those common helpers?

Thank you for working on this!

melinath: We generally try not to add diff suppress functions to the common library unless there's strong evidence that the behavior will be the same across at least a few different API surfaces. I'm not sure offhand whether that's the case here, so I'd be inclined to leave it as specific to `compute` for now.

translucens:
I found injecting diff to `schema.ResourceData` is difficult so added acceptance test cases which covers logic of `DiffSuppressSourceRanges`.

note: The workaround in https://github.com/hashicorp/terraform-plugin-sdk/issues/221 was not worked as intended

melinath:
Ah, yeah - the diff suppresses are actually using the resource data object. If you'd like to add more unit tests, we do have [ResourceDiffMock](https://github.com/hashicorp/terraform-provider-google/blob/55b615e10bb7ec185a4a8d0d47b67c82a22d17d7/google/tpgresource/resource_test_utils.go#L94-L100) which lets you provide mock data.

(You'd also need to change the function signature to take [TerraformResourceDiff](https://github.com/hashicorp/terraform-provider-google/blob/55b615e10bb7ec185a4a8d0d47b67c82a22d17d7/google/tpgresource/utils.go#L66).)

That said, I haven't taken a thorough look at the changes yet; if you're feeling good about the level of testing it may already be fine as-is.

melinath:
Test run failed due to a GH blip
/gcbrun

mmv1/templates/terraform/constants/firewall.tmpl
```diff
@@ -107,6 +107,28 @@ func diffSuppressSourceRanges(k, old, new string, d *schema.ResourceData) bool {
...
+}
+
+func CompareCIDRlist (oldRanges, newRanges []any) bool {
+	if len(oldRanges) != len(newRanges) {
+		return false
+	}
+
+	for i := 0; i < len(oldRanges); i++ {
+		oldCIDR := netip.MustParsePrefix(oldRanges[i].(string))
+		newCIDR := netip.MustParsePrefix(newRanges[i].(string))
```
melinath, line 127: Since these are user-supplied strings, we should use [`ParsePrefix`](https://pkg.go.dev/net/netip#ParsePrefix) instead to avoid panics. Please also add tests cases for handling errors in parsing either the old or new ranges.

PR #14933, opened 2025-08-21
 provider: automatically generate the `*_wo` and `*_wo_version` for write-only arguments + compute: added `write-only` support for `shared_secret ` argument for `google_compute_vpn_tunnel` resource

ramonvermeulen:
Closes https://github.com/hashicorp/terraform-provider-google/issues/23058

```release-note:enhancement
compute: added `shared_secret_wo` and `shared_secret_wo_version` fields for `google_compute_vpn_tunnel` resource, enabling write-only management of the shared secret.
```


ramonvermeulen:
@melinath 

I was just experimenting with state upgraders in https://github.com/GoogleCloudPlatform/magic-modules/pull/14933/commits/b860c8cf5c5f0ac0ffeb78016a9382b21e267d79 after your comment https://github.com/GoogleCloudPlatform/magic-modules/pull/14913#issuecomment-3211184736, I was just wondering if it makes sense at all in this case to move forward with this.

With the bugged release (`v6.49.0`) there were **two serious problems**:

<hr>

**Problem 1:** the default value of `secret_data_wo_version` changed from `0` to `nil`.

This caused force replacements, even for users that didn't use the write-only variant because under-the-hood the defaults are always still set. Example of this is:

```terraform
resource "google_secret_manager_secret_version" "db_host" {
  secret      = google_secret_manager_secret.google_secret_manager_secret["DB_HOST"].id
  secret_data = google_sql_database_instance.google_sql_database_instance.ip_address[0].ip_address
}
```

causing

```
Step #2 - "tf plan":   # module.dev.google_secret_manager_secret_version.db_host must be replaced
Step #2 - "tf plan": -/+ resource "google_secret_manager_secret_version" "db_host" {
Step #2 - "tf plan":       ~ create_time            = "2025-05-30T15:52:04.734092Z" -> (known after apply)
Step #2 - "tf plan":       + destroy_time           = (known after apply)
Step #2 - "tf plan":       ~ id                     = "projects/764710828782/secrets/dev_DB_HOST/versions/1" -> (known after apply)
Step #2 - "tf plan":       ~ name                   = "projects/764710828782/secrets/dev_DB_HOST/versions/1" -> (known after apply)
Step #2 - "tf plan":       - secret_data_wo_version = "0" -> null # forces replacement
Step #2 - "tf plan":       ~ version                = "1" -> (known after apply)
Step #2 - "tf plan":         # (5 unchanged attributes hidden)
Step #2 - "tf plan":     }
```

For this a state upgrader could make sense, however, only if I am able to determine if the user defined `secret_data_wo_version` is set to "0" because the user explicitly set it OR because it was due to the default value setting it under-the-hood. In the case of the default value, you could write a state upgrader that changes it from `0` to `nil`. However, if the user explicitly set it, you don't want to touch it. And as far I can see in the state upgrader method we can only work with raw state and not make the distinction.

<hr>

**Problem 2:** `secret_data_wo` because "required_with" `secret_data_wo_version` (and visa versa)

This caused problems for users that were already using the write-only variant, however, relied upon the default value (`0`) of the `secret_data_wo_version`. For example for users who had something configured like:

```terraform
resource "google_secret_manager_secret_version" "db_host" {
  secret         = google_secret_manager_secret.google_secret_manager_secret["DB_HOST"].id
  secret_data_wo = google_sql_database_instance.google_sql_database_instance.ip_address[0].ip_address
  # no `secret_data_wo_version` configured, because it had default 0 anyways
}
```

However, because it had the "required_with" configured both ways, this caused problems.

```
"secret_data_wo": all of `secret_data_wo,secret_data_wo_version` must be specified
```

<hr>

I was wondering if you have an idea how we could prevent both these scenarios using a state upgrader. I think (even though it is less elegant) a solution like https://github.com/GoogleCloudPlatform/magic-modules/pull/14913 is the only way to tackle both these cases. I think the required_with case will always be a problem until a next major release, where you can tell in the upgrade guide that every user should explicitly set the `secret_data_wo_version` moving forward.

melinath:
For Problem 1 - I wouldn't _expect_ that people have that field set to 0 explicitly, since that wouldn't do anything, but... stranger things have happened. Perhaps we should just add a diff suppress function to the field instead of messing with state migrations?

For Problem 2 - this is essentially a bug. The current code: https://github.com/GoogleCloudPlatform/magic-modules/blob/bc461dd7ef33d2e5095fcc3b92c692b5087b318d/mmv1/products/secretmanager/SecretVersion.yaml#L172

This line is incorrect - it should be `secret_data_wo_version` - probably actually `payload.0.secret_data_wo_version`. But it is a breaking change. @c2thorn @NickElliot could we potentially fix the required_with constraint as part of the major release, so that we don't have to work around it in write-only field generation? Alternatively, @BBBmau - could we remove the RequiredWith constraint to allow empty version values?

c2thorn:
> This line is incorrect - it should be `secret_data_wo_version` - probably actually `payload.0.secret_data_wo_version`. But it is a breaking change. @c2thorn @NickElliot could we potentially fix the required_with constraint as part of the major release, so that we don't have to work around it in write-only field generation?

If it can be done, tested, and merged by EOD then sure. Otherwise I don't recommend it given the time remaining.

ramonvermeulen:
For Problem 1 - I agree, then the code in `ResourceSecretManagerSecretVersionUpgradeV0` makes sense. However, I think if you want to be entirely sure, a diff suppress function seems better. Because then you just ignore (suppress) the diff if it's old and new value are both either `0` or `nil`.

For Problem 2 - I'll wait on other responses

BBBmau:
> Alternatively, @BBBmau - could we remove the RequiredWith constraint to allow empty version values?

It's expected for a `*_wo_version` to be inputted when a write-only argument is being used, so we wouldn't be able to move forward with removing the `RequiredWith` constraint unless we're fine with the case of the  version being `nil` despite the `*_wo` field being used. [hashicorp docs discussing write-only argument usage](https://developer.hashicorp.com/terraform/language/resources/ephemeral/write-only#declare-a-write-only-argument)

melinath:
I've made the changes to RequiredWith / ExactlyOneOf in https://github.com/GoogleCloudPlatform/magic-modules/pull/14941 - although they would've been lower impact, they are still technically breaking, so it's better to get them done as part of the major release. Since we currently believe it's possible to do the Int -> String conversion without breaking users, we can leave that for later.

ramonvermeulen:
> I've made the changes to RequiredWith / ExactlyOneOf in #14941 - although they would've been lower impact, they are still technically breaking, so it's better to get them done as part of the major release. Since we currently believe it's possible to do the Int -> String conversion without breaking users, we can leave that for later.

Great work! Thanks a lot for putting in the effort and making this ready before the `v7.0.0` release. This will make the move towards generated write-only arguments easier because the existing write-only supported resources now align with the generation logic in this PR. This tackles both problem 1 and 2 at the same time.

I guess it makes sense to wait with this PR until `v7.0.0` is released and merged back into `main`?

melinath:
I think we'll still need to tackle problem 1, because that comes from the conversion of int -> string + removal of default, but yeah, problem 2 should now be solved. I believe the plan is to merge the 7.0.0 branch into main today, so you should be able to merge or rebase shortly.

ramonvermeulen:
> I think we'll still need to tackle problem 1, because that comes from the conversion of int -> string + removal of default, but yeah, problem 2 should now be solved. I believe the plan is to merge the 7.0.0 branch into main today, so you should be able to merge or rebase shortly.

Check, I will rebase as soon as the `v7.0.0` branch is merged in! 

I thought because the required_with is now set both ways from `v7.0.0` onwards, a user always explicitly has to set the `wo_version`, not causing a problem.

But I think you are right, because existing state still has the `0` as default, and changing that to `nil` force replacement. So for that we should still implement the suppress diff. I'll tag you as soon as the PR is ready, probably somewhere next week.

ramonvermeulen:
@melinath 

Just rebased, but I think the `v7.0.0` changes are not merged yet into `main`? Given the left side of the diff shows the old values for `required_with` and `conflicts` (before your fix in https://github.com/GoogleCloudPlatform/magic-modules/pull/14941).

```yaml
        required_with:
          - 'SecretDataWoVersion'
        conflicts:
          - 'payload.0.secretData'
```

I was about to look into setting a `suppress_diff_func` specifically for the `secret_data_wo_version` it's `0` -> `nil`. However, with this PR I don't have direct access to this field anymore via the yaml (because with this PR the field is now generated), making it slightly more complicated.

What I could do is make a specific condition on field name / resource type and implement it directly into the generator in the `buildWriteOnlyVersionField` function, but my gut feeling says there is a better way to do this. Even though you have this rare scenario (https://github.com/GoogleCloudPlatform/magic-modules/pull/14933#issuecomment-3211584254), maybe moving back to the initial approach using a state_upgrader makes more sense after all (just like I did in https://github.com/GoogleCloudPlatform/magic-modules/commit/b860c8cf5c5f0ac0ffeb78016a9382b21e267d79, basically re-reverting that commit)?

melinath:
7.0.0 was merged into main but it looks like (some of) my changes were lost during a [merge conflict resolution](https://github.com/GoogleCloudPlatform/magic-modules/commit/570a8c3825e6892470168dc978134b21299dccb2#diff-324857dec85f1d8866b9eb21ae5ec1b7d34c6f663e637ed6db720f4e3c0cd201).

@c2thorn @NickElliot FYI

melinath:
because of the generation, if we set up a diff suppress func on one resource, we'd need to set it up on all of them (which would probably be fine). That said, it's possible that the state upgrader might work better. We'd just need to be sure to thoroughly test.

ramonvermeulen:
> 7.0.0 was merged into main but it looks like (some of) my changes were lost during a [merge conflict resolution](https://github.com/GoogleCloudPlatform/magic-modules/commit/570a8c3825e6892470168dc978134b21299dccb2#diff-324857dec85f1d8866b9eb21ae5ec1b7d34c6f663e637ed6db720f4e3c0cd201).
> 
> @c2thorn @NickElliot FYI

Check, I see the changes at least [made it to the `v7.x.x` release ](https://github.com/hashicorp/terraform-provider-google/blob/146c9f8b233fa903ef98dad66874b13eb9c2bcc0/google/services/bigquerydatatransfer/resource_bigquery_data_transfer_config.go#L295), so I will wait a bit till this is resolved.

> because of the generation, if we set up a diff suppress func on one resource, we'd need to set it up on all of them (which would probably be fine). That said, it's possible that the state upgrader might work better. We'd just need to be sure to thoroughly test.

Yes, do you have any preference? Pro for the state upgrader is that it is only on this specific `google_secret_manager_secret_version` resource instead of all current and future write_only_version fields. For testing I will do write-out of a couple of manual test scenarios once the PR is ready.

mmv1/api/resource.go
```diff
@@ -749,14 +749,108 @@ func (r Resource) GetIdentity() []*Type {
...
+
+	return NewProperty(name, originalField.ApiName, options)
+}
+
+func buildWriteOnlyVersionField(name string, originalField *Type, writeOnlyField *Type, originalFieldLineage string) *Type {
+	description := fmt.Sprintf("Triggers update of %s write-only. For more info see [updating write-only attributes](/docs/providers/google/guides/using_write_only_attributes.html#updating-write-only-attributes)", google.Underscore(writeOnlyField.Name))
+	requiredWith := strings.ReplaceAll(originalFieldLineage, google.Underscore(originalField.Name), google.Underscore(writeOnlyField.Name))
+
+	options := []func(*Type){
+		propertyWithType("String"),
```
melinath: let's just go with int since we don't need to be backwards-compatible at this point.

```suggestion
		propertyWithType("Int"),
```

melinath (changes requested):
> Check, I see the changes at least [made it to the v7.x.x release ](https://github.com/hashicorp/terraform-provider-google/blob/146c9f8b233fa903ef98dad66874b13eb9c2bcc0/google/services/bigquerydatatransfer/resource_bigquery_data_transfer_config.go#L295), so I will wait a bit till this is resolved.

That code was actually unchanged by my updates. The big change to that file (which is still not present) was adding RequiredWith to `secret_access_key_wo` (which is still missing).

---

We had our postmortem for the patch release today and have come to the conclusion that this change needs a more cautious approach: we want to separate out the new generation logic from conversions to the new generation logic and avoid any potentially-breaking changes while doing so.

That means that there are three phases for moving forward:

1. Make the existing write only fields use a separate flag that maintains the previous behavior (that is, rename write_only to write_only_legacy on main). This would happen in a separate PR.
2. Introduce the new generation logic - basically this PR. We should choose one write-only field to generate as part of this - probably a new one, since none of the existing fields perfectly match what we want.
3. Migrate the legacy write_only fields to the new generator logic - I'll create tickets to track this as part of the next major release & we or the service teams will handle cleanup then.

Given this change in strategy, we don't need to figure out how to handle the default value change here, or do anything else that's only for backwards-compatibility.

melinath:
Let me know if you have any questions / concerns about this approach!

ramonvermeulen:
> 1. Make the existing write only fields use a separate flag that maintains the previous behavior (that is, rename write_only to write_only_legacy on main). This would happen in a separate PR.
> 2. Introduce the new generation logic - basically this PR. We should choose one write-only field to generate as part of this - probably a new one, since none of the existing fields perfectly match what we want.
> 3. Migrate the legacy write_only fields to the new generator logic - I'll create tickets to track this as part of the next major release & we or the service teams will handle cleanup then.
> 
> Given this change in strategy, we don't need to figure out how to handle the default value change here, or do anything else that's only for backwards-compatibility.

Thanks! This sounds to me like a good plan.

1. Check, so you really mean the field definition in the [mmv1 field spec](https://googlecloudplatform.github.io/magic-modules/reference/field/#write_only) to be renamed from `write_only` to `write_only_legacy`, so that this will not trigger generation for **this pr** making the migration easier and not directly impacting the existing write-only resources. (Opened a PR at https://github.com/GoogleCloudPlatform/magic-modules/pull/15022)
2. Sounds good, I would then suggest to use either https://github.com/hashicorp/terraform-provider-google/issues/23058 or https://github.com/hashicorp/terraform-provider-google/issues/21855 for this.
3. Check, this can now be done one at a time, making review/impact analysis easier. Once all existing write-only resources are migrated the `write_only_legacy` flag can be removed.

melinath:
> 1. Check, so you really mean the field definition in the [mmv1 field spec](https://googlecloudplatform.github.io/magic-modules/reference/field/#write_only) to be renamed from `write_only` to `write_only_legacy`, so that this will not trigger generation for **this pr** making the migration easier and not directly impacting the existing write-only resources. (Opened a PR at [MMv1 field: changed `write_only` to `write_only_legacy` and marked field as deprecated #15022](https://github.com/GoogleCloudPlatform/magic-modules/pull/15022))

Yes, exactly.

> 2. Sounds good, I would then suggest to use either [Add support to `google_compute_vpn_tunnel` for write-only `shared_secret` argument. hashicorp/terraform-provider-google#23058](https://github.com/hashicorp/terraform-provider-google/issues/23058) or [Add support for write only values in `monitoring_notification_channel` hashicorp/terraform-provider-google#21855](https://github.com/hashicorp/terraform-provider-google/issues/21855) for this.

Either one is probably fine. I'd lean towards monitoring_notification_channel because it has a bunch of thumbs.

ramonvermeulen:
Downstream diff should be fairly minimal, and only change things related to `monitoring_notification_channel` and the new util method `GetRawConfigAttributeAsString` in `tpgresource`.

**Acceptance tests:**

```
❮ make testacc TEST=./google/services/monitoring TESTARGS='-run=TestAccMonitoringNotificationChannel_'
TF_ACC=1 TF_SCHEMA_PANIC_ON_ERROR=1 go test ./google/services/monitoring -v -run=TestAccMonitoringNotificationChannel_ -timeout 240m -ldflags="-X=github.com/hashicorp/terraform-provider-google/version.ProviderVersion=acc"
=== RUN   TestAccMonitoringNotificationChannel_notificationChannelBasicExample
=== PAUSE TestAccMonitoringNotificationChannel_notificationChannelBasicExample
=== RUN   TestAccMonitoringNotificationChannel_update
=== PAUSE TestAccMonitoringNotificationChannel_update
=== RUN   TestAccMonitoringNotificationChannel_updateLabels_slack
    resource_monitoring_notification_channel_test.go:57:
--- SKIP: TestAccMonitoringNotificationChannel_updateLabels_slack (0.00s)
=== RUN   TestAccMonitoringNotificationChannel_updateLabels
=== PAUSE TestAccMonitoringNotificationChannel_updateLabels
=== RUN   TestAccMonitoringNotificationChannel_updateSensitiveLabels_slack
    resource_monitoring_notification_channel_test.go:107:
--- SKIP: TestAccMonitoringNotificationChannel_updateSensitiveLabels_slack (0.00s)
=== RUN   TestAccMonitoringNotificationChannel_updateSensitiveLabels
=== PAUSE TestAccMonitoringNotificationChannel_updateSensitiveLabels
=== RUN   TestAccMonitoringNotificationChannel_updateSensitiveLabelsWo
=== PAUSE TestAccMonitoringNotificationChannel_updateSensitiveLabelsWo
=== RUN   TestAccMonitoringNotificationChannel_updateSensitiveLabelsWo_slack
    resource_monitoring_notification_channel_test.go:217:
--- SKIP: TestAccMonitoringNotificationChannel_updateSensitiveLabelsWo_slack (0.00s)
=== CONT  TestAccMonitoringNotificationChannel_notificationChannelBasicExample
=== CONT  TestAccMonitoringNotificationChannel_updateSensitiveLabels
=== CONT  TestAccMonitoringNotificationChannel_updateLabels
=== CONT  TestAccMonitoringNotificationChannel_updateSensitiveLabelsWo
=== CONT  TestAccMonitoringNotificationChannel_update
--- PASS: TestAccMonitoringNotificationChannel_updateSensitiveLabels (8.20s)
--- PASS: TestAccMonitoringNotificationChannel_notificationChannelBasicExample (8.50s)
--- PASS: TestAccMonitoringNotificationChannel_updateLabels (10.59s)
--- PASS: TestAccMonitoringNotificationChannel_update (12.66s)
--- PASS: TestAccMonitoringNotificationChannel_updateSensitiveLabelsWo (16.98s)
PASS
ok      github.com/hashicorp/terraform-provider-google/google/services/monitoring       17.741s
```

**Manual tests:**

Applied after building provider binary on this branch via:
```bash
TF_LOG=DEBUG TF_CLI_CONFIG_FILE="$(pwd)/tf-dev-override.tfrc" terraform apply --auto-approve
```

Config:

```terraform
resource "google_monitoring_notification_channel" "basicauth" {
  display_name = "TFTest Basicauth Channel"
  type         = "webhook_basicauth"
  labels = {
    "username" = "username"
    "url"      = "http://fakeurl.com"
  }

  sensitive_labels {
    password_wo         = "somepassword-very-secret"
    password_wo_version = 1
  }
}

resource "google_monitoring_notification_channel" "pagerduty" {
  display_name = "TFTest Pagerduty Channel"
  type         = "pagerduty"

  sensitive_labels {
    service_key_wo         = "some_service_key"
    service_key_wo_version = 1
  }
}
```

terraform.tfstate (no plain-text values for write_only arguments):

<details>

```
{
  "version": 4,
  "terraform_version": "1.13.1",
  "serial": 38,
  "lineage": "b158aaaf-9cae-aae1-464d-2cd00eeaec35",
  "outputs": {},
  "resources": [
    {
      "mode": "managed",
      "type": "google_monitoring_notification_channel",
      "name": "basicauth",
      "provider": "provider[\"registry.terraform.io/hashicorp/google\"]",
      "instances": [
        {
          "schema_version": 0,
          "attributes": {
            "description": "",
            "display_name": "TFTest Basicauth Channel",
            "enabled": true,
            "force_delete": false,
            "id": "projects/my-gcp-project/notificationChannels/3287287572745080595",
            "labels": {
              "url": "http://fakeurl.com",
              "username": "username"
            },
            "name": "projects/my-gcp-project/notificationChannels/3287287572745080595",
            "project": "my-gcp-project",
            "sensitive_labels": [
              {
                "auth_token": "",
                "auth_token_wo": null,
                "auth_token_wo_version": "",
                "password": "",
                "password_wo": null,
                "password_wo_version": "1",
                "service_key": "",
                "service_key_wo": null,
                "service_key_wo_version": ""
              }
            ],
            "timeouts": null,
            "type": "webhook_basicauth",
            "user_labels": null,
            "verification_status": ""
          },
          "sensitive_attributes": [
            [
              {
                "type": "get_attr",
                "value": "sensitive_labels"
              },
              {
                "type": "index",
                "value": {
                  "value": 0,
                  "type": "number"
                }
              },
              {
                "type": "get_attr",
                "value": "auth_token"
              }
            ],
            [
              {
                "type": "get_attr",
                "value": "sensitive_labels"
              },
              {
                "type": "index",
                "value": {
                  "value": 0,
                  "type": "number"
                }
              },
              {
                "type": "get_attr",
                "value": "password"
              }
            ],
            [
              {
                "type": "get_attr",
                "value": "sensitive_labels"
              },
              {
                "type": "index",
                "value": {
                  "value": 0,
                  "type": "number"
                }
              },
              {
                "type": "get_attr",
                "value": "service_key"
              }
            ]
          ],
          "identity_schema_version": 0,
          "private": "XXXX"
        }
      ]
    },
    {
      "mode": "managed",
      "type": "google_monitoring_notification_channel",
      "name": "pagerduty",
      "provider": "provider[\"registry.terraform.io/hashicorp/google\"]",
      "instances": [
        {
          "schema_version": 0,
          "attributes": {
            "description": "",
            "display_name": "TFTest Pagerduty Channel",
            "enabled": true,
            "force_delete": false,
            "id": "projects/my-gcp-project/notificationChannels/3287287572745077767",
            "labels": null,
            "name": "projects/my-gcp-project/notificationChannels/3287287572745077767",
            "project": "my-gcp-project",
            "sensitive_labels": [
              {
                "auth_token": "",
                "auth_token_wo": null,
                "auth_token_wo_version": "",
                "password": "",
                "password_wo": null,
                "password_wo_version": "",
                "service_key": "",
                "service_key_wo": null,
                "service_key_wo_version": "1"
              }
            ],
            "timeouts": null,
            "type": "pagerduty",
            "user_labels": null,
            "verification_status": ""
          },
          "sensitive_attributes": [
            [
              {
                "type": "get_attr",
                "value": "sensitive_labels"
              },
              {
                "type": "index",
                "value": {
                  "value": 0,
                  "type": "number"
                }
              },
              {
                "type": "get_attr",
                "value": "auth_token"
              }
            ],
            [
              {
                "type": "get_attr",
                "value": "sensitive_labels"
              },
              {
                "type": "index",
                "value": {
                  "value": 0,
                  "type": "number"
                }
              },
              {
                "type": "get_attr",
                "value": "password"
              }
            ],
            [
              {
                "type": "get_attr",
                "value": "sensitive_labels"
              },
              {
                "type": "index",
                "value": {
                  "value": 0,
                  "type": "number"
                }
              },
              {
                "type": "get_attr",
                "value": "service_key"
              }
            ]
          ],
          "identity_schema_version": 0,
          "private": "XXXX"
        }
      ]
    }
  ],
  "check_results": null
}
```

</details>

mmv1/products/monitoring/NotificationChannel.yaml
```diff
@@ -105,31 +105,31 @@ properties:
     properties:
       - name: 'authToken'
         type: String
+        write_only: true
         description: |
           An authorization token for a notification channel. Channel types that support this field include: slack
         sensitive: true
-        exactly_one_of:
-          - 'sensitive_labels.0.auth_token'
+        conflicts:
```
melinath: Optional fields that are already in an ExactlyOneOf should extend that instead of converting it to a Conflicts. But that probably doesn't Just Work because if you only specify the _wo field, you'll get an error because the exactly_one_of constraint for the other fields isn't fulfilled.

We would need to fix that - but it would require significant work on top of what's already in this PR so we should handle it separately. (Basically we'll need to have a way to modify the exactlyoneof constraints for all the fields in the set.)

Sorry for not realizing this would be a problem sooner! It looks like google_compute_vpn_tunnel is simpler - no complex conflicts / exactlyoneof / etc. constraints. Could we do that first instead?

ramonvermeulen: Yes, that is indeed the pain point here.. also quite a hard one to get a good solution for I think. Because at generation time the `_wo` field basically adds itself to the already existing list of `exactly_one_of`. However, it doesn't / can't know of the existence of the other `_wo` fields (I don't know if this explanation makes sense at all, but I hope it does :')).

To do this properly the generator somehow needs to map out all the fields with an exactly_one_of in combination with a write_only and the relationships on eachother upfront. Probably need to brainstorm and play around a bit to come up with a solution that can tackle this.

<hr>

No problem, this whole thing has been quite a learning experience so far and I don't mind spending more time on it. My plan is as follows:

1. Revert all changes related to https://github.com/hashicorp/terraform-provider-google/issues/21855
2. Choose https://github.com/hashicorp/terraform-provider-google/issues/23058 instead as first issue for the write-only generation logic and do that implementation in this PR
3. Add a warning in the documentation spec `write_only` that it cannot be used in combination with an `exactly_one_of` on multiple write-only arguments. This can be removed later once we have a solid solution for this problem. However, we will handle that seperately after this PR.

Let me know if you agree.

melinath: yeah that sounds good.

ramonvermeulen:
@melinath 

Updated the PR with what we discussed in https://github.com/GoogleCloudPlatform/magic-modules/pull/14933#discussion_r2323076470

**Acceptance tests locally** (ran with -parallel 3 because of network quota on my personal org):

```
❯ make testacc TEST=./google/services/compute TESTARGS='-run=TestAccComputeVpnTunnel_ -parallel 3'
TF_ACC=1 TF_SCHEMA_PANIC_ON_ERROR=1 go test ./google/services/compute -v -run=TestAccComputeVpnTunnel_ -parallel 3 -timeout 240m -ldflags="-X=github.com/hashicorp/terraform-provider-google/version.ProviderVersion=acc"
=== RUN   TestAccComputeVpnTunnel_vpnTunnelBasicExample
=== PAUSE TestAccComputeVpnTunnel_vpnTunnelBasicExample
=== RUN   TestAccComputeVpnTunnel_regionFromGateway
=== PAUSE TestAccComputeVpnTunnel_regionFromGateway
=== RUN   TestAccComputeVpnTunnel_router
=== PAUSE TestAccComputeVpnTunnel_router
=== RUN   TestAccComputeVpnTunnel_routerWithSharedSecretWo_update
=== PAUSE TestAccComputeVpnTunnel_routerWithSharedSecretWo_update
=== RUN   TestAccComputeVpnTunnel_defaultTrafficSelectors
=== PAUSE TestAccComputeVpnTunnel_defaultTrafficSelectors
=== CONT  TestAccComputeVpnTunnel_vpnTunnelBasicExample
=== CONT  TestAccComputeVpnTunnel_routerWithSharedSecretWo_update
=== CONT  TestAccComputeVpnTunnel_defaultTrafficSelectors
--- PASS: TestAccComputeVpnTunnel_vpnTunnelBasicExample (231.69s)
=== CONT  TestAccComputeVpnTunnel_router
--- PASS: TestAccComputeVpnTunnel_defaultTrafficSelectors (252.40s)
=== CONT  TestAccComputeVpnTunnel_regionFromGateway
--- PASS: TestAccComputeVpnTunnel_routerWithSharedSecretWo_update (354.89s)
--- PASS: TestAccComputeVpnTunnel_router (174.31s)
--- PASS: TestAccComputeVpnTunnel_regionFromGateway (208.58s)
PASS
ok      github.com/hashicorp/terraform-provider-google/google/services/compute  461.781s
```

**Manual test:**

Used similar config as the acceptance test.

```bash
TF_CLI_CONFIG_FILE="$(pwd)/tf-dev-override.tfrc" terraform apply --auto-approve
```

image

Please also take a look at: https://github.com/hashicorp/terraform-provider-google/issues/23058#issuecomment-2917196327 (comment on the returned hash)

melinath:
> Please also take a look at: [hashicorp/terraform-provider-google#23058 (comment)](https://github.com/hashicorp/terraform-provider-google/issues/23058#issuecomment-2917196327) (comment on the returned hash)

The question of whether the hash is sensitive should be considered separately from this PR. Avoiding storing the hash in state is not required to make this change, and it would be a significant uplift.

ramonvermeulen:
>   * Expected Document Path: `/website/docs/r/compute_vpn_tunnel.html.markdown`

I think this check doesn't support [`## Ephemeral Attributes Reference`](https://github.com/modular-magician/terraform-provider-google/compare/0f3bec09cdbd53c19422f6048967dc51feab9818..7a219ae2c1c8b5945516d98b8e16fb7c2e867f06#diff-57f6d9f4f8ca190a93b7ff796bb1e5bd49598b323585fbe60e1215c220ef3b20R361) section yet, might be a nice fun issue to resolve.

melinath:
> > * Expected Document Path: `/website/docs/r/compute_vpn_tunnel.html.markdown`
> 
> I think this check doesn't support [`## Ephemeral Attributes Reference`](https://github.com/modular-magician/terraform-provider-google/compare/0f3bec09cdbd53c19422f6048967dc51feab9818..7a219ae2c1c8b5945516d98b8e16fb7c2e867f06#diff-57f6d9f4f8ca190a93b7ff796bb1e5bd49598b323585fbe60e1215c220ef3b20R361) section yet, might be a nice fun issue to resolve.

I think that's exactly correct :-) please feel free to take a stab if you'd like! I've opened https://github.com/hashicorp/terraform-provider-google/issues/24287 to track.

mmv1/api/resource.go
```diff
@@ -749,14 +749,108 @@ func (r Resource) GetIdentity() []*Type {
...
+		propertyWithDescription(description),
+		propertyWithWriteOnly(true),
+		propertyWithApiName(apiName),
+		propertyWithIgnoreRead(true),
+		propertyWithRequiredWith([]string{requiredWith}),
+	}
+
+	if originalField.Required {
+		exactlyOneOf := append(originalField.ExactlyOneOf, fieldPathOriginalField, fieldPathCurrentField)
+		options = append(options, propertyWithExactlyOneOf(deduplicateSliceOfStrings(exactlyOneOf)))
```
melinath, line 789: Can we set ExactlyOneOf for the OriginalField too? I think it's not strictly "necessary" in that the constraint will be enforced anyway, but it would be better for consistency. It looks like this works since originalField is a pointer.

```suggestion
		options = append(options, propertyWithExactlyOneOf(deduplicateSliceOfStrings(exactlyOneOf)))
		originalField.ExactlyOneOf = deduplicateSliceOfStrings(exactlyOneOf)
```

docs/content/reference/field.md
```diff
@@ -107,14 +107,34 @@ Example:
...
 ```
 
+### `write_only`
+Set to true to enable write-only functionality for this field.
+If true, the [**Write-only Arguments**](https://developer.hashicorp.com/terraform/plugin/sdkv2/resources/write-only-arguments) will be automatically generated by the code generator (`[field_name]_wo` and `[field_name]_wo_version`). 
+When the write-only argument for a field is used, it means that its value will be obscured in Terraform output as well as not be stored in state. 
+This field is meant to replace `sensitive` as it doesn't store the value in state. 
+See [Ephemerality in Resources - Use Write-only arguments](https://developer.hashicorp.com/terraform/language/resources/ephemeral/write-only) for more information.
+
+Write-only arguments are only supported in Terraform v1.11+. Because the provider supports earlier Terraform versions, write only fields must be paired with (mutually exclusive) `sensitive` fields covering the same functionality for compatibility with those older versions.
```
melinath: Rewrote this a bit - the previous language assumed that `write_only` was being used on a separate field from `sensitive`.

```suggestion
The "base" version of the field will be marked as `sensitive` so that the value is obscured in Terraform output but it will still be stored in state. (This is necessary for compatibility with Terraform <v1.11.) The write-only version of the field will be obscured in Terraform output and _won't_  be stored in state. See [Ephemerality in Resources - Use Write-only arguments](https://developer.hashicorp.com/terraform/language/resources/ephemeral/write-only) for more information.
```

docs/content/reference/field.md
```diff
@@ -107,14 +107,34 @@ Example:
...
+This field cannot be used in conjunction with `immutable` or `sensitive`.
+
+**Note**: Due to write-only not being read from the API, it is not possible to update the field directly unless a sidecar field is used. (e.g. `password` as a write-only field and `password_wo_version` as an immutable field meant for updating).
```
melinath: These parts aren't relevant because we're generating the "sidecar" fields.

```suggestion
```

docs/content/reference/field.md
```diff
@@ -107,14 +107,34 @@ Example:
...
+
+Example:
+
+```yaml
+write_only: true
+```
+
+**Warning:** This field cannot be used in combination with `exactly_one_of` on multiple write-only fields. This is planned to be fixed in the future.
```
melinath: Created a ticket for this.

```suggestion
**Warning:** This field cannot be used in combination with `exactly_one_of` on multiple write-only fields. This is planned to be [fixed in the future](https://github.com/hashicorp/terraform-provider-google/issues/24327).
```

docs/content/reference/field.md
```diff
@@ -107,14 +107,34 @@ Example:
...
+
 ### `write_only_legacy` (deprecated)
 If true, the field is considered "write-only", which means that its value will
 be obscured in Terraform output as well as not be stored in state. This field is meant to replace `sensitive` as it doesn't store the value in state.
 See [Ephemerality in Resources - Use Write-only arguments](https://developer.hashicorp.com/terraform/language/resources/ephemeral/write-only)
 for more information.
 
-Write-only fields are only supported in Terraform v1.11+. Because the provider supports earlier Terraform versions, write only fields must be paired with (mutually exclusive) `sensitive` fields covering the same functionality for compatibility with those older versions.
-This field cannot be used in conjuction with `immutable` or `sensitive`.
+Write-only arguments are only supported in Terraform v1.11+. Because the provider supports earlier Terraform versions, write only fields must be paired with (mutually exclusive) `sensitive` fields covering the same functionality for compatibility with those older versions.
```
melinath:
```suggestion
Write-only fields are only supported in Terraform v1.11+. Because the provider supports earlier Terraform versions, write only fields must be paired with (mutually exclusive) `sensitive` fields covering the same functionality for compatibility with those older versions.
```

melinath (changes requested):
Sorry, I started this review but forgot to finish it! Here you go! I do also still need to double-check the downstream functionality but I'm not anticipating any issues there.

docs/content/reference/field.md
ramonvermeulen: I actually changed this from fields to arguments, because I think that is technically the "official" term that hashicorp uses ([source](https://developer.hashicorp.com/terraform/language/manage-sensitive-data/write-only)). On the other hand I also understand fields is maybe more logical within the mmv1 context? E.g. within mmv1 context it is a field, which in the downstream provider is being generated as an argument or attribute.

Either way, I am fine with either one of those ;)

ramonvermeulen:
> Sorry, I started this review but forgot to finish it! Here you go! I do also still need to double-check the downstream functionality but I'm not anticipating any issues there.

Thanks for the review and suggestions, just applied them 👍🏻

Let's have one final look at the generated code when the CI is done.

Also you can assign https://github.com/hashicorp/terraform-provider-google/issues/24327 to me, I will follow up on that in the upcoming weeks when this PR is through.

ramonvermeulen:
> `google` provider: [Diff](https://github.com/modular-magician/terraform-provider-google/compare/4429aada009bb9fd086f58339a435248c7f3e23c..2ea98050fb3db3e4507f967aaccfb49d31606bc3) ( 6 files changed, 267 insertions(+), 15 deletions(-)) `google-beta` provider: [Diff](https://github.com/modular-magician/terraform-provider-google-beta/compare/ff90840f1d3dd9c3e8f9aeae96ab0257129662fb..ba2ac39989e27c042804ed6b7024c99543620308) ( 6 files changed, 268 insertions(+), 16 deletions(-)) `terraform-google-conversion`: [Diff](https://github.com/modular-magician/terraform-google-conversion/compare/e7b0ab34a6e328500c27164ca49a073a77c73237..ddef3d847749744bdb61413a3bcfefa5c6ad1f11) ( 2 files changed, 50 insertions(+))

Downstream diff looks good to me, I am only not sure about the failing VCR tests. They seem unrelated, but maybe good to double check.

mmv1/api/resource.go
```diff
@@ -749,14 +749,109 @@ func (r Resource) GetIdentity() []*Type {
...
+	return NewProperty(name, originalField.ApiName, options)
+}
+
+func buildWriteOnlyVersionField(name string, originalField *Type, writeOnlyField *Type, originalFieldLineage string) *Type {
+	description := fmt.Sprintf("Triggers update of %s write-only. For more info see [updating write-only attributes](/docs/providers/google/guides/using_write_only_attributes.html#updating-write-only-attributes)", google.Underscore(writeOnlyField.Name))
+	requiredWith := strings.ReplaceAll(originalFieldLineage, google.Underscore(originalField.Name), google.Underscore(writeOnlyField.Name))
+
+	options := []func(*Type){
+		propertyWithType("Int"),
+		propertyWithImmutable(originalField.Immutable),
```
melinath: It turns out checking `Immutable` isn't enough. For `google_compute_vpn_tunnel`, the field isn't marked immutable, but the _resource_ is immutable - so in the downstream code, all fields are considered immutable.

We need to instead check [IsForceNew](https://github.com/GoogleCloudPlatform/magic-modules/blob/4c6257388aa8f5aff7b619043d809a7e47c63ebc/mmv1/api/type.go#L1217), which is what's [used in the template for the schema](https://github.com/GoogleCloudPlatform/magic-modules/blob/4c6257388aa8f5aff7b619043d809a7e47c63ebc/mmv1/templates/terraform/schema_property.go.tmpl#L40-L42)

```suggestion
		propertyWithImmutable(originalField.IsForceNew()),
```

(As a side note, this means that shared_secret_wo is an immutable client-side field! Very unusual!)

mmv1/third_party/terraform/services/compute/resource_compute_vpn_tunnel_test.go
```diff
@@ -59,6 +59,45 @@ func TestAccComputeVpnTunnel_router(t *testing.T) {
...
+				),
+			},
+			{
+				ResourceName:            "google_compute_vpn_tunnel.foobar",
+				ImportState:             true,
+				ImportStateVerify:       true,
+				ImportStateVerifyIgnore: []string{"shared_secret", "detailed_status"},
+			},
+			{
+				Config: testAccComputeVpnTunnelRouterWithSharedSecretWo_update(acctest.RandString(t, 10), router),
```
melinath: We should verify that this deletes the resource & recreates it.

This will require also importing `"github.com/hashicorp/terraform-plugin-testing/plancheck"`

```suggestion
				Config: testAccComputeVpnTunnelRouterWithSharedSecretWo_update(acctest.RandString(t, 10), router),
				ConfigPlanChecks: resource.ConfigPlanChecks{
					PreApply: []plancheck.PlanCheck{
						plancheck.ExpectResourceAction("google_compute_vpn_tunnel.foobar", plancheck.ResourceActionDestroyBeforeCreate),
					},
				},
```

mmv1/api/resource.go
ramonvermeulen: Thanks, I will change this to `IsForceNew()`. 
I see that method does quite a bunch of checks, including on the parent.

ramonvermeulen:
> It looks like there's still one blocking issue - the version field should be immutable to match the original field. (Left notes below about what's happening & how to fix.) I was going to suggest that we should check whether the value of shared_secret_hash changes after the apply but apparently it changes on every create even if I'm sending the same value with shared_secret (or shared_secret_wo). 🤔 but I can see that the sent value is correct in the API requests.

Thanks for the comprehensive review! I just applied your feedback in the last few commits, and reran the tests locally:

```
make testacc TEST=./google/services/compute TESTARGS='-run=TestAccComputeVpnTunnel_ -parallel 3'
sh -c "'/Users/ramon/go/src/github.com/hashicorp/terraform-provider-google/scripts/gofmtcheck.sh'"
go vet
TF_ACC=1 TF_SCHEMA_PANIC_ON_ERROR=1 go test ./google/services/compute -v -run=TestAccComputeVpnTunnel_ -parallel 3 -timeout 240m -ldflags="-X=github.com/hashicorp/terraform-provider-google/version.ProviderVersion=acc"
=== RUN   TestAccComputeVpnTunnel_vpnTunnelBasicExample
=== PAUSE TestAccComputeVpnTunnel_vpnTunnelBasicExample
=== RUN   TestAccComputeVpnTunnel_regionFromGateway
=== PAUSE TestAccComputeVpnTunnel_regionFromGateway
=== RUN   TestAccComputeVpnTunnel_router
=== PAUSE TestAccComputeVpnTunnel_router
=== RUN   TestAccComputeVpnTunnel_routerWithSharedSecretWo_update
=== PAUSE TestAccComputeVpnTunnel_routerWithSharedSecretWo_update
=== RUN   TestAccComputeVpnTunnel_defaultTrafficSelectors
=== PAUSE TestAccComputeVpnTunnel_defaultTrafficSelectors
=== CONT  TestAccComputeVpnTunnel_vpnTunnelBasicExample
=== CONT  TestAccComputeVpnTunnel_defaultTrafficSelectors
=== CONT  TestAccComputeVpnTunnel_routerWithSharedSecretWo_update
--- PASS: TestAccComputeVpnTunnel_defaultTrafficSelectors (233.54s)
=== CONT  TestAccComputeVpnTunnel_router
--- PASS: TestAccComputeVpnTunnel_vpnTunnelBasicExample (235.90s)
=== CONT  TestAccComputeVpnTunnel_regionFromGateway
--- PASS: TestAccComputeVpnTunnel_routerWithSharedSecretWo_update (323.89s)
--- PASS: TestAccComputeVpnTunnel_router (141.30s)
--- PASS: TestAccComputeVpnTunnel_regionFromGateway (259.90s)
PASS
ok      github.com/hashicorp/terraform-provider-google/google/services/compute  496.617s
```

mmv1/third_party/terraform/services/compute/resource_compute_vpn_tunnel_test.go
```diff
@@ -59,6 +60,50 @@ func TestAccComputeVpnTunnel_router(t *testing.T) {
...
+	t.Parallel()
+
+	router := fmt.Sprintf("tf-test-tunnel-%s", acctest.RandString(t, 10))
+	acctest.VcrTest(t, resource.TestCase{
+		PreCheck:                 func() { acctest.AccTestPreCheck(t) },
+		ProtoV5ProviderFactories: acctest.ProtoV5ProviderFactories(t),
+		CheckDestroy:             testAccCheckComputeVpnTunnelDestroyProducer(t),
+		Steps: []resource.TestStep{
+			{
+				Config: testAccComputeVpnTunnelRouterWithSharedSecretWo(acctest.RandString(t, 10), router),
```
melinath: here & below, this should take the suffix as a variable rather than generating a new suffix each time (because that will unnecessarily recreate the resources, which is why this test is passing.)

mmv1/api/resource.go
```diff
@@ -749,14 +749,109 @@ func (r Resource) GetIdentity() []*Type {
...
+
+// AddExtraFields processes properties and adds supplementary fields based on property types.
+// It handles write-only properties, labels, and annotations.
+func (r *Resource) AddExtraFields(props []*Type, parent *Type, lineage string) []*Type {
 	for _, p := range props {
+		currentPropLineage := r.buildCurrentPropLineage(p, lineage)
+		if p.WriteOnly && !strings.HasSuffix(p.Name, "Wo") {
+			props = r.addWriteOnlyFields(props, p, currentPropLineage)
+			p.WriteOnly = false
+			p.Required = false
```
melinath: let's handle this as part of addWriteOnlyFields

```suggestion
```

mmv1/api/resource.go
```diff
@@ -749,14 +749,109 @@ func (r Resource) GetIdentity() []*Type {
...
+}
+
+// AddExtraFields processes properties and adds supplementary fields based on property types.
+// It handles write-only properties, labels, and annotations.
+func (r *Resource) AddExtraFields(props []*Type, parent *Type, lineage string) []*Type {
 	for _, p := range props {
+		currentPropLineage := r.buildCurrentPropLineage(p, lineage)
+		if p.WriteOnly && !strings.HasSuffix(p.Name, "Wo") {
+			props = r.addWriteOnlyFields(props, p, currentPropLineage)
+			p.WriteOnly = false
```
melinath: This is causing issues because the original field gets treated as WriteOnly when checking IsForceNew (even though it's not actually.) We should set WriteOnly to false before calling addWriteOnlyFields to avoid that.

```suggestion
			p.WriteOnly = false
			props = r.addWriteOnlyFields(props, p, currentPropLineage)
```

mmv1/api/resource.go
```diff
@@ -749,14 +749,109 @@ func (r Resource) GetIdentity() []*Type {
...
+		propertyWithType("String"),
+		propertyWithRequired(false),
+		propertyWithDescription(description),
+		propertyWithWriteOnly(true),
+		propertyWithApiName(apiName),
+		propertyWithIgnoreRead(true),
+		propertyWithRequiredWith([]string{requiredWith}),
+	}
+
+	if originalField.Required {
```
melinath, line 786:
```suggestion
	if originalField.Required {
		originalField.Required = false
```

mmv1/main.go
```diff
@@ -235,7 +235,7 @@ func GenerateProduct(version, providerName, productName, outputPath string, prod
 		resource.SourceYamlFile = resourceYamlPath
 
 		resource.TargetVersionName = version
-		resource.Properties = resource.AddLabelsRelatedFields(resource.PropertiesWithExcluded(), nil)
+		resource.Properties = resource.AddExtraFields(resource.PropertiesWithExcluded(), nil, "")
```
melinath:
```suggestion
		// SetDefault before AddExtraFields to ensure relevant metadata is available on existing fields
		resource.SetDefault(productApi)
		resource.Properties = resource.AddExtraFields(resource.PropertiesWithExcluded(), nil, "")
		// SetDefault after AddExtraFields to ensure relevant metadata is available on generated fields
```

mmv1/main.go
```diff
@@ -268,7 +268,7 @@ func GenerateProduct(version, providerName, productName, outputPath string, prod
 			}
 
 			resource.TargetVersionName = version
-			resource.Properties = resource.AddLabelsRelatedFields(resource.PropertiesWithExcluded(), nil)
+			resource.Properties = resource.AddExtraFields(resource.PropertiesWithExcluded(), nil, "")
```
melinath:
```suggestion
			// SetDefault before AddExtraFields to ensure relevant metadata is available on existing fields
			resource.Properties = resource.AddExtraFields(resource.PropertiesWithExcluded(), nil, "")
			// SetDefault after AddExtraFields to ensure relevant metadata is available on generated fields
```

melinath (changes requested):
It looks like the downstream generated code is still missing `ForceNew`. I looked into it and there's two main causes:

- SetDefault (our old friend). Let's just call it before & after adding the extra fields.
- WriteOnly being set to true on the originalField when the new fields are added

ramonvermeulen:
> * SetDefault (our old friend). Let's just call it before & after adding the extra fields.

Oeff, not again haha ;)

I think if we call it both before and after we might get rid of `buildCurrentPropLineage` as well, and just call `.TerraformLineage()` directly (need to double check this). 

I don't know how big the performance impact will be of calling SetDefault twice, but maybe we want to apply some kind of caching strategy if necessary (e.g. flagging fields for which it already applied the SetDefault logic). Also I think we want to pay some extra attention to see if the method is fully idempotent.

I will update the PR in the upcoming days when I have some free time.

melinath:
Just using TerraformLineage seems reasonable. I don't think we need to do a rework of SetDefault logic / cache anything as part of this PR. In local runs I don't see a noticeable duration difference between with / without. It also seems likely idempotent; in my local testing, the generated downstream doesn't have extraneous changes.

If you'd want to investigate it as a separate follow-up PR that would be fine though!

ramonvermeulen:
```
➜ make testacc TEST=./google/services/compute TESTARGS='-run=TestAccComputeVpnTunnel_ -parallel 3'
sh -c "'/Users/ramon/go/src/github.com/hashicorp/terraform-provider-google/scripts/gofmtcheck.sh'"
go vet
TF_ACC=1 TF_SCHEMA_PANIC_ON_ERROR=1 go test ./google/services/compute -v -run=TestAccComputeVpnTunnel_ -parallel 3 -timeout 240m -ldflags="-X=github.com/hashicorp/terraform-provider-google/version.ProviderVersion=acc"
=== RUN   TestAccComputeVpnTunnel_vpnTunnelBasicExample
=== PAUSE TestAccComputeVpnTunnel_vpnTunnelBasicExample
=== RUN   TestAccComputeVpnTunnel_regionFromGateway
=== PAUSE TestAccComputeVpnTunnel_regionFromGateway
=== RUN   TestAccComputeVpnTunnel_router
=== PAUSE TestAccComputeVpnTunnel_router
=== RUN   TestAccComputeVpnTunnel_routerWithSharedSecretWo_update
=== PAUSE TestAccComputeVpnTunnel_routerWithSharedSecretWo_update
=== RUN   TestAccComputeVpnTunnel_defaultTrafficSelectors
=== PAUSE TestAccComputeVpnTunnel_defaultTrafficSelectors
=== CONT  TestAccComputeVpnTunnel_vpnTunnelBasicExample
=== CONT  TestAccComputeVpnTunnel_routerWithSharedSecretWo_update
=== CONT  TestAccComputeVpnTunnel_defaultTrafficSelectors
--- PASS: TestAccComputeVpnTunnel_routerWithSharedSecretWo_update (216.48s)
=== CONT  TestAccComputeVpnTunnel_router
--- PASS: TestAccComputeVpnTunnel_defaultTrafficSelectors (222.89s)
=== CONT  TestAccComputeVpnTunnel_regionFromGateway
--- PASS: TestAccComputeVpnTunnel_vpnTunnelBasicExample (234.57s)
--- PASS: TestAccComputeVpnTunnel_router (149.47s)
--- PASS: TestAccComputeVpnTunnel_regionFromGateway (226.42s)
PASS
ok      github.com/hashicorp/terraform-provider-google/google/services/compute  450.094s
```

Just tried to apply your feedback, `ForceNew` seems to be set correctly now on my local build. Let's take a look at the what the CI diff produces.

<details>

```go
			"shared_secret": {
				Type:     schema.TypeString,
				Optional: true,
				ForceNew: true,
				Description: `Shared secret used to set the secure session between the Cloud VPN
gateway and the peer VPN gateway.`,
				Sensitive:    true,
				ExactlyOneOf: []string{"shared_secret", "shared_secret_wo"},
			},
			"shared_secret_wo": {
				Type:     schema.TypeString,
				Optional: true,
				Description: `Shared secret used to set the secure session between the Cloud VPN
gateway and the peer VPN gateway.
 Note: This property is write-only and will not be read from the API. For more info see [updating write-only attributes](/docs/providers/google/guides/using_write_only_attributes.html#updating-write-only-attributes)`,
				WriteOnly:    true,
				ExactlyOneOf: []string{"shared_secret", "shared_secret_wo"},
				RequiredWith: []string{"shared_secret_wo_version"},
			},
			"shared_secret_wo_version": {
				Optional:     true,
				ForceNew:     true,
				Description:  `Triggers update of shared_secret_wo write-only. For more info see [updating write-only attributes](/docs/providers/google/guides/using_write_only_attributes.html#updating-write-only-attributes)`,
				RequiredWith: []string{"shared_secret_wo"},
			},
```

</details>

melinath:
I see ForceNew on the downstream 👍 Tests are still running, and I'll be OOO for a few days, so I won't be able to merge this this week.

ramonvermeulen:
Some thoughts about `SetDefault`, maybe it could also work to just call:

https://github.com/ramonvermeulen/magic-modules/blob/3e114d4d6cbee1d3502d52cc892d8bf616c01778/mmv1/api/type.go#L336

For every field directly after the field is generated. Then you only have to do SetDefault before AddExtraFields. 
But this requires some investigation, the [resource variant](https://github.com/ramonvermeulen/magic-modules/blob/3e114d4d6cbee1d3502d52cc892d8bf616c01778/mmv1/api/resource.go#L407) loops over the properties to call SetDefault on property level, but also does a lot of other things.

docs/content/reference/field.md
```diff
@@ -107,14 +107,27 @@ Example:
 sensitive: true
 ```
 
+### `write_only`
+Set to true to enable write-only functionality for this field.
+If true, the [**Write-only Arguments**](https://developer.hashicorp.com/terraform/plugin/sdkv2/resources/write-only-arguments) will be automatically generated by the code generator (`[field_name]_wo` and `[field_name]_wo_version`). 
+The "base" version of the field will be marked as `sensitive` so that the value is obscured in Terraform output but it will still be stored in state. (This is necessary for compatibility with Terraform <v1.11.) The write-only version of the field will be obscured in Terraform output and _won't_  be stored in state. See [Ephemerality in Resources - Use Write-only arguments](https://developer.hashicorp.com/terraform/language/resources/ephemeral/write-only) for more information.
```
melinath: We don't seem to actually do this automatically right now. I'm just updating the docs here; we can make that change in a small follow-on PR if we want. I've opened https://github.com/hashicorp/terraform-provider-google/issues/24484 to track it.

```suggestion
The base field should also be marked as `sensitive` so that the value is obscured in Terraform output but it will still be stored in state. (This is necessary for compatibility with Terraform <v1.11.) The write-only version of the field will be obscured in Terraform output and _won't_  be stored in state. See [Ephemerality in Resources - Use Write-only arguments](https://developer.hashicorp.com/terraform/language/resources/ephemeral/write-only) for more information.
```

ramonvermeulen: Check, sounds good! I will make a follow-up PR after this PR has been merged 👍🏼

BBBmau (approved):
I went and generated the provider and all looks good, walked through the generation code and it all looks good as well.

Thank you again for working on this! I'll be merging this since the latest commit was just a documentation change.

PR #14934, opened 2025-08-21
tgc-revival: add cloudasset feed resources

zli82016:

```release-note:none

```

### DIFF
--- a/mmv1/templates/tgc_next/decoders/cloud_asset_feed.go.tmpl
+++ b/mmv1/templates/tgc_next/decoders/cloud_asset_feed.go.tmpl
@@ -1,0 +1,4 @@
+// billing_project is the required url_param_only property, but is not in CAI asset name or data
+// TODO: handle it in a generic way
+hclData["billing_project"] = "null"
+return res, hclData, nil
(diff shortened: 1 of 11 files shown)


mmv1/templates/tgc_next/decoders/cloud_asset_feed.go.tmpl
```diff
@@ -0,0 +1,4 @@
+// billing_project is the required url_param_only property, but is not in CAI asset name or data
+// TODO: handle it in a generic way
+hclData["billing_project"] = "null"
```
zli82016, line 3: `billing_project` is a special property. It is just to set the billing project when sending the requests to APIs.  

https://github.com/hashicorp/terraform-provider-google-beta/blob/main/google-beta/services/cloudasset/resource_cloud_asset_folder_feed.go#L56

trodge: Would it be confusing to set it to "hidden" or "sensitive" instead of "null"? I suspect that might be the reason it's not in the cai asset.

trodge (approved):

trodge:
Added @hao-nan-li in case we want to merge this before the 2 reviewer requirement is lifted.

mmv1/templates/tgc_next/decoders/cloud_asset_feed.go.tmpl
zli82016: `billingProject` is a wired property. It is not a field in [the API object](https://cloud.google.com/asset-inventory/docs/reference/rest/v1/feeds) by design. I think it is just for billing purposes when sending the API requests, but is not included in the API request body.

zli82016:
We can wait until the main branch is unlocked.

hao-nan-li (approved):

PR #14941, opened 2025-08-21
Standardized required_with behavior for write-only fields

melinath:
Clean-up / standardization for write-only fields that requires some small breaking changes. https://github.com/GoogleCloudPlatform/magic-modules/pull/14933#issuecomment-3211584254 for context.

I've gone through and tested the plan errors locally to compare the latest release with these changes.

Note: I believe the bigquerydatatransfer Config change is not breaking, because we already enforce it via a [customize diff func](https://github.com/hashicorp/terraform-provider-google-beta/blob/886372902bda169fb44f8224a3a38561f6d36b59/google-beta/services/bigquerydatatransfer/resource_bigquery_data_transfer_config.go#L57-L59). I'm not removing that in this PR to keep things simple. 

```release-note:breaking-change
sql: on `google_sql_user`, made `password_wo_version` required when `password_wo` is set
```

```release-note:breaking-change
secretmanager: on `google_secret_manager_secret_version`, made `secret_data_wo` and `secret_data_wo_version` both required when one is set
```

```release-note:breaking-change
monitoring: on `google_monitoring_uptime_check_config`, made it required to set exactly one of `http_check.auth_info.password` and `http_check.auth_info.password_wo`
```

### DIFF
--- a/mmv1/products/bigquerydatatransfer/Config.yaml
+++ b/mmv1/products/bigquerydatatransfer/Config.yaml
@@ -222,30 +222,26 @@ properties:
       to a different credential configuration in the config will require an apply to update state.
     url_param_only: true
     properties:
-      - name: 'secretAccessKeyWoVersion'
-        type: Integer
-        url_param_only: true
-        required_with:
-          - 'sensitive_params.0.secretAccessKeyWo'
-        description: |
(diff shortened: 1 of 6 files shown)


melinath:
TGC failures are unrelated - this PR doesn't modify TGC. Breaking change failure is expected.

BBBmau (approved):
LGTM!

PR #14943, opened 2025-08-21
fix other scenario for node_pool not exist

ScottSuarez:
resolves b/438538527

```release-note:none

```


melinath (approved):
Seems straightforward / similar to the other change. LGTM.

Do we need to worry at all about the fact that we're now removing the node from state for all errors, instead of checking whether it's a "not found" error?

ScottSuarez:
> Seems straightforward / similar to the other change. LGTM.
> 
> Do we need to worry at all about the fact that we're now removing the node from state for all errors, instead of checking whether it's a "not found" error?

This is appropriate for the cache interface. There is no variance in the errors the cache returns.

see https://github.com/hashicorp/terraform-provider-google-beta/blob/c01e684d3065bae3fc26d3acbd23553bba9069d1/google-beta/services/container/resource_container_node_pool.go#L52-L61

melinath:
ack, thanks!

ScottSuarez:
failure unrelated

PR #14952, opened 2025-08-22
Increase robustness of disk_autoresize in sql_database_instance

same-id:
Solves: https://github.com/hashicorp/terraform-provider-google/issues/18296

Now instances that have `disk_autoresize` set to true (default) will automatically ignore the `disk_size` field (unless a higher size was specified).

Furethermore, both `disk_autoresize` and `ignore_changes = [settings[0].disk_size]` will now work even if the state was not refreshed before running.

Added tests.

This change also includes:
* Additional `resourceSqlDatabaseInstanceRead` at the begining of the update flow. This helps with updating the `settingsVersion` when the instance is updated outside terraform.
* Bump in version of `github.com/hashicorp/terraform-plugin-testing v1.13.2` to allow testing with the new `AdditionalCLIOptions`

```release-note:enhancement
sql: Increase robustness of disk_autoresize in sql_database_instance
```

### DIFF
--- a/mmv1/third_party/terraform/services/sql/resource_sql_database_instance.go.tmpl
+++ b/mmv1/third_party/terraform/services/sql/resource_sql_database_instance.go.tmpl
@@ -17,7 +17,6 @@ import (
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/validation"
 
-	"github.com/hashicorp/terraform-provider-google/google/services/compute"
 	"github.com/hashicorp/terraform-provider-google/google/services/servicenetworking"
 	"github.com/hashicorp/terraform-provider-google/google/tpgresource"
 	transport_tpg "github.com/hashicorp/terraform-provider-google/google/transport"
@@ -127,6 +126,41 @@ var (
 	}
 )
 
+func diskSizeCutomizeDiff(ctx context.Context, d *schema.ResourceDiff, meta interface{}) error {
+	key := "settings.0.disk_size"
+
+	old, new := d.GetChange(key)
+	// It's okay to remove size entirely.
+	if old == nil || new == nil {
+		return nil
+	}
+	autoResizeI, exists := d.GetOkExists("settings.0.disk_autoresize")
+	autoResize := !exists || autoResizeI.(bool)
+
+	if old.(int) <= new.(int) {
+		// If a resize up, always allow it - keep the diff.
+		return nil
+	}
+
+	if autoResize {
+		// Allow having disk size larger in the state than in config if auto resize is enabled.
+		// Delete the diff.
+		err := d.Clear(key)
+		if err != nil {
+			return err
+		}
+		return nil
+	}
+
+	// If we are here, we are trying to shrink the disk with auto resize disabled and no ignore changes on disk size.
+	// This will force a new resource.
+	if err := d.ForceNew(key); err != nil {
+		return err
+	}
+
+	return nil
+}
+
 func ResourceSqlDatabaseInstance() *schema.Resource {
 	return &schema.Resource{
 		Create: resourceSqlDatabaseInstanceCreate,
@@ -145,7 +179,7 @@ func ResourceSqlDatabaseInstance() *schema.Resource {
 
 		CustomizeDiff: customdiff.All(
 			tpgresource.DefaultProviderProject,
-			customdiff.ForceNewIfChange("settings.0.disk_size", compute.IsDiskShrinkage),
+			diskSizeCutomizeDiff,
 			customdiff.ForceNewIf("master_instance_name", func(_ context.Context, d *schema.ResourceDiff, meta interface{}) bool {
 				// If we set master but this is not the new master of a switchover, require replacement and warn user.
 				return !isSwitchoverFromOldPrimarySide(d)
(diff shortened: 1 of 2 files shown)


mmv1/third_party/terraform/go.mod
trodge, line 1: Please move these dependency version bumps into a separate PR so that we can analyze their effects independently from the new features.

same-id: See: https://github.com/GoogleCloudPlatform/magic-modules/pull/14973

trodge: If the version bump in the `terraform-plugin-testing` package isn't required to get this test to pass we can omit it from this PR.

same-id: Ok, I'll remove it

But I am thinking out loud here - I'm perhaps not sure about the addition of:

```go
	// Always read the instance before updating it.
	// In cases where the instance was edited outside of Terraform, this will ensure that
	// the we grab the latest settingsVersion in case we need to update the settings.
	err = resourceSqlDatabaseInstanceRead(d, meta)
	if err != nil {
		return err
	}
```

At the beginning of the flow here.

Since by default `terraform apply` refreshes the state of an object that it tries to update.

The reason I added it was to also make it work in `pulumi` which uses the terraform provider underneath and suffers from the fact that by default it does not refresh resources before applying change (I currently suspect that it's a problem that pulumi has in general, but haven't gotten a confirmation from them yet).

Currently they fix it by having a misplaced patch (IMO) here:
https://github.com/pulumi/pulumi-gcp/blob/master/patches/0008-Fix-794-with-an-unconditional-read.patch

Which is supposed to fix: https://github.com/pulumi/pulumi-gcp/issues/794

Aaron Friel (@AaronFriel) from Pulumi, suggested fixing it before:

https://github.com/hashicorp/terraform-provider-google/issues/15762

But I see that they haven't taken it to the magic-modules repo.

WDYT?

Do we want to support `pulumi` and `terraform -refresh=false apply` out of the box here?

same-id: Also @trodge, If I remove:

```go
		AdditionalCLIOptions: &resource.AdditionalCLIOptions{
			Plan: resource.PlanOptions{
				NoRefresh: true,
			},
		},
```

The entire test becomes pointless since every `terraform apply` refreshes the state.

So I also need to add this change without a test which is "okay" but I don't like it much.

Should I at least add the test commented out with `// XXX - Uncomment this test  when we upgrade terraform-plugin-testing to support NoRefresh: true`
?

same-id: There's been an interesting development.

See here:
https://github.com/GoogleCloudPlatform/magic-modules/pull/14973#issuecomment-3240303148

guineveresaenger:
Hi @same-id - Guinevere from Pulumi here. @AaronFriel is no longer at the company, which is probably why we lost track of making this change upstream. I've been looking into our patches, and came across this pull request.

We'd love to see an unconditional Read on this so we're following this development with interest. 💟

mmv1/third_party/terraform/go.mod
same-id: I did a small refactor (without changing the tests) which makes more sense - disk size will only be provided in the API (non-0) if there has been a change, instead of always - this removes much of the complexity and edge cases.

Would appreciate a review

trodge:
Sorry for the delayed review @same-id, the test should be running now in our CI.

Does the upstream change you linked obviate the need for this change or do we want to do both in parallel?

trodge:
Could this test failure (`TestAccSqlDatabaseInstance_BackupDRUpdate`) be related to your change?
```
PUT /sql/v1beta4/projects/ci-test-project-188019/instances/tf-test-ywizbedawo?alt=json&prettyPrint=false HTTP/1.1
Host: sqladmin.googleapis.com
User-Agent: google-api-go-client/0.5 Terraform/1.11.0 (+https://www.terraform.io) Terraform-Plugin-SDK/2.37.0 terraform-provider-google-beta/acc
Content-Length: 865
X-Goog-Api-Client: gl-go/1.23.12 gdcl/0.248.0
Accept-Encoding: gzip

{
 "instanceType": "CLOUD_SQL_INSTANCE",
 "settings": {
  "activationPolicy": "ALWAYS",
  "availabilityType": "ZONAL",
  "backupConfiguration": {
   "backupRetentionSettings": {
    "retainedBackups": 7,
    "retentionUnit": "COUNT"
   },
   "binaryLogEnabled": false,
   "enabled": false,
   "pointInTimeRecoveryEnabled": false,
   "startTime": "09:00",
   "transactionLogRetentionDays": 7
  },
  "connectorEnforcement": "NOT_REQUIRED",
  "dataCacheConfig": {},
  "dataDiskType": "PD_SSD",
  "edition": "ENTERPRISE",
  "enableDataplexIntegration": false,
  "enableGoogleMlIntegration": false,
  "ipConfiguration": {
   "enablePrivatePathForGoogleCloudServices": false,
   "ipv4Enabled": true,
   "requireSsl": null,
   "serverCaMode": "GOOGLE_MANAGED_INTERNAL_CA",
   "sslMode": "ALLOW_UNENCRYPTED_AND_ENCRYPTED"
  },
  "locationPreference": {
   "zone": "us-central1-f"
  },
  "pricingPlan": "PER_USE",
  "retainBackupsOnDelete": false,
  "settingsVersion": "1",
  "storageAutoResize": true,
  "tier": "db-g1-small"
 }
}

2025/09/15 18:24:48 [DEBUG] Google API Response Details:
HTTP/2.0 412 Precondition Failed
Date: Mon, 15 Sep 2025 18:24:48 GMT
Server: ESF
Vary: Origin
Vary: X-Origin
Vary: Referer
X-Frame-Options: SAMEORIGIN
X-Xss-Protection: 0

{
  "error": {
    "code": 412,
    "message": "Condition does not match.",
    "errors": [
      {
        "message": "Condition does not match.",
        "domain": "global",
        "reason": "staleData",
        "location": "If-Match",
        "locationType": "header"
      }
    ]
  }
}
```

same-id:
I ran `TestAccSqlDatabaseInstance_BackupDRUpdate` on `fd5b04dcc - (HEAD, origin/nightly-test, origin/main, origin/HEAD) Merge feature branch major release 7.0.0` which is the parent commit and got the same failure (before my changes) - So I am guessing this was not broken in this PR.

trodge (approved):

same-id:
Thanks for helping push this through!

PR #14971, opened 2025-08-23
cloudrunv2: Fixed permadiff in `template.scaling.min_instance_count`

wyardley:
Added `default_from_api` to `template.scaling.max_instance_count`, to resolve a permadiff if `min_instance_count` is set to the default (0) without a value set for `max_instance_count`

Part of https://github.com/hashicorp/terraform-provider-google/issues/20399

See linked ticket: currently, this does _not_ resolve the diff when one of CPU or memory is defined in limits but the other is unset.

```release-note:bug
cloudrunv2: fixed permadiff in `template.scaling.min_instance_count` when `max_instance_count` is unset.
```

### DIFF
--- a/mmv1/products/cloudrunv2/Service.yaml
+++ b/mmv1/products/cloudrunv2/Service.yaml
@@ -402,6 +402,7 @@ properties:
             description: |-
               Maximum number of serving instances that this resource should have. Must not be less than minimum instance count. If absent, Cloud Run will calculate
               a default value based on the project's available container instances quota in the region and specified instance size.
+            default_from_api: true
       - name: 'vpcAccess'
         type: NestedObject
         description: |-


wyardley:
> What conditions will a value get returned by Cloud Run? I think to solve this we probably need to work through the rules for that relatively exhaustively.

My brain is a little fuzzy on all this ~ 5 weeks later, but did you read this comment in the linked issue? This hopefully explains what I think is happening.
https://github.com/hashicorp/terraform-provider-google/issues/20399#issuecomment-3216108957

For example, see comments inline here:
```hcl
resource "google_cloud_run_v2_service" "default" {
  # [...]
  template {
    scaling {
      min_instance_count = 0   // This is the default, but max_instance_count isn't also set
    }

    containers {
      image = "us-docker.pkg.dev/cloudrun/container/hello"
      resources {
        limits = {
          memory = "1Gi"  // CPU not set; presumably you'd run into the same issue if CPU were set but memory wasn't
        }
      }
    }
  }
}
```

From my limited testing, if _all_ defaults are taken (i.e., `template.scaling` is empty), things are fine, but once you set one of them explicitly, the values returned back from the API start causing some confusion. 

I was going off the docs on permadiffs, and the settings of some existing parameters, and in my _limited_ testing, the fix here seemed to work, but agree that it may be worth doing some additional testing to confirm. Baking in the default values would probably also work, so will defer to you in terms of what's the better behavior (or if there's another way to "fix" this).

I wasn't able to come up with something easy for the `containers.resources.limits` settings that seemed to work.

PR #14973, opened 2025-08-23
Update `github.com/hashicorp/terraform-plugin-testing` to `v1.13.3`

same-id:
So we can now use `AdditionalCLIOptions` in tests.

See: https://github.com/GoogleCloudPlatform/magic-modules/pull/14952

```release-note: enhancement
go: Update github.com/hashicorp/terraform-plugin-testing for AdditionalCLIOptions
```

### DIFF
--- a/mmv1/third_party/terraform/go.mod
+++ b/mmv1/third_party/terraform/go.mod
@@ -25,7 +25,7 @@ require (
 	github.com/hashicorp/terraform-plugin-log v0.9.0
 	github.com/hashicorp/terraform-plugin-mux v0.20.0
 	github.com/hashicorp/terraform-plugin-sdk/v2 v2.37.0
-	github.com/hashicorp/terraform-plugin-testing v1.5.1
+	github.com/hashicorp/terraform-plugin-testing v1.13.3
 	github.com/mitchellh/go-homedir v1.1.0
 	github.com/mitchellh/hashstructure v1.1.0
 	github.com/sirupsen/logrus v1.8.1
@@ -54,7 +54,7 @@ require (
 	github.com/apparentlymart/go-textseg/v15 v15.0.0 // indirect
 	github.com/cenkalti/backoff v2.2.1+incompatible // indirect
 	github.com/cespare/xxhash/v2 v2.3.0 // indirect
-	github.com/cloudflare/circl v1.6.0 // indirect
+	github.com/cloudflare/circl v1.6.1 // indirect
 	github.com/cncf/xds/go v0.0.0-20250501225837-2ac532fd4443 // indirect
 	github.com/envoyproxy/go-control-plane/envoy v1.32.4 // indirect
 	github.com/envoyproxy/protoc-gen-validate v1.2.1 // indirect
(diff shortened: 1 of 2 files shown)


roaks3 (commented):
Running tests to verify, but this could have some implications to our various testing methods. It looks like this upgrades us from a version that is ~2 years old.

Depending how tests look and probably some spot checking (like verifying that diff test behavior still works), we should be able to move this forward, but recommendation in the near-term is to avoid requiring any new features in your other changes.

same-id:
> Running tests to verify, but this could have some implications to our various testing methods. It looks like this upgrades us from a version that is ~2 years old.
> 
> Depending how tests look and probably some spot checking (like verifying that diff test behavior still works), we should be able to move this forward, but recommendation in the near-term is to avoid requiring any new features in your other changes.

Hi @roaks3, TBH I don't require this in my #14952 PR - I only needed it to create a test that fails before my changes, that passes after my changes.

If we can merge it without requiring this test to fail before - It's fine by me - WDYT?

I can comment out the AdditionalCLIOptions option and keep it in-code commented out with "Waiting for terraform-plugin-testing upgrade to uncomment this"

trodge:
Our main reason for freezing the `terraform-plugin-testing` package was an issue with `Set` (as in sets of nested objects with no fixed order) fields behaving unexpectedly in the versions immediately after the one we pinned. If that issue has been resolved we could do this version bump.

https://github.com/hashicorp/terraform-plugin-testing/issues/327 for reference

roaks3:
Ah, thanks for that context. Yep, at a minimum we're seeing that same issue with the tests in the PR.

@same-id you can try to resolve referring to the issue that @trodge mentioned, but I think it will be challenging. This seems to be a stronger case for trying to make your other change without the version bump of this library.

same-id:
I'll try to avoid the library update

same-id:
By the way, I don't have access to the test results

same-id:
Seems like @bbasata recently fixed this in `v1.13.3`

(See: https://github.com/hashicorp/terraform-plugin-testing/releases/tag/v1.13.3)

I tested this locally:

- Went to main branch, ran tests locally with RECORDING
- Used this branch in v1.13.2 in REPLAYING mode - failed
- Changed to v1.13.3 and added in GNUMakefile: `TF_ACC_REFRESH_AFTER_APPLY=1` - *original replay works*

I have bumped the version again, however not sure where I should add the flag in this repo - seems like we need to still set this flag in both `terraform-provider-google` and `terraform-provider-google-beta` as additional PRs since the GNUMakefile is not generated from magic-modules.

Would appreciate if you will help me carry this PR if this indeed solves the issue :-)

Thanks all

rileykarson:
I approved this build so we can get a test run through. The changes mentioned in https://github.com/GoogleCloudPlatform/magic-modules/pull/14973#issuecomment-3240303148 should unblock updating the library in theory.

same-id:
> I approved this build so we can get a test run through. The changes mentioned in [#14973 (comment)](https://github.com/GoogleCloudPlatform/magic-modules/pull/14973#issuecomment-3240303148) should unblock updating the library in theory.

I am not sure that this will work without setting `TF_ACC_REFRESH_AFTER_APPLY=1` as an environment variable for the tests.

rileykarson:
Thanks! https://github.com/GoogleCloudPlatform/magic-modules/pull/14923 covers that and there's some discussion there. I'm gonna reassign this review to @BBBmau 

@modular-magician reassign-reviewer @BBBmau

BBBmau:
closing this since we are now on plugin-testing 1.14.1 after merging the following PR last week:
- https://github.com/GoogleCloudPlatform/magic-modules/pull/16694

PR #14974, opened 2025-08-23
Improve root_password and password_validation_policy updates in sql instance

same-id:
In cases where both the password_validation_policy and the root_password are being changed - it is important to first update the password_validation_policy and only then update the password.

Consider the case where mistakenly we set a too strict policy and now the password updates are failing - we now want to loosen the policy, but we always fail on updates since the password change always happens first. This requires now doing a "two-step" terraform apply instead of single one.

Also always update the settings first - this is needed in cases where update failed mid-run but we still have updated the settingsVersion which causes subsequent failures. (Similar change is being done here: https://github.com/GoogleCloudPlatform/magic-modules/pull/14952)

```release-note:enhancement
sql: Improve root_password and password_validation_policy updates in sql instance
```

### DIFF
--- a/mmv1/third_party/terraform/services/sql/resource_sql_database_instance.go.tmpl
+++ b/mmv1/third_party/terraform/services/sql/resource_sql_database_instance.go.tmpl
@@ -2060,70 +2060,6 @@ func resourceSqlDatabaseInstanceUpdate(d *schema.ResourceData, meta interface{})
 		}
 	}
 
-	// Check if the root_password is being updated, because updating root_password is an atomic operation and can not be
-	// performed with other fields, we first update root password before updating the rest of the fields.
-	if d.HasChange("root_password") {
-		oldPwd, newPwd := d.GetChange("root_password")
-		password := newPwd.(string)
-		dv := d.Get("database_version").(string)
-		name := ""
-		host := ""
-		if strings.Contains(dv, "MYSQL") {
-			name = "root"
-			host = "%"
-		} else if strings.Contains(dv, "POSTGRES") {
-			name = "postgres"
-		} else if strings.Contains(dv, "SQLSERVER") {
-			name = "sqlserver"
-			if len(password) == 0 {
-				if err := d.Set("root_password", oldPwd.(string)); err != nil {
-					return fmt.Errorf("Error re-setting root_password: %s", err)
-				}
-				return fmt.Errorf("Error, root password cannot be empty for SQL Server instance.")
-			}
-		} else {
-			if err := d.Set("root_password", oldPwd.(string)); err != nil {
-				return fmt.Errorf("Error re-setting root_password: %s", err)
-			}
-			return fmt.Errorf("Error, invalid database version")
-		}
-		instance := d.Get("name").(string)
-
-		user := &sqladmin.User{
-			Name:     name,
-			Instance: instance,
-			Password: password,
-		}
-
-		transport_tpg.MutexStore.Lock(instanceMutexKey(project, instance))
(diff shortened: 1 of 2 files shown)


mmv1/third_party/terraform/services/sql/resource_sql_database_instance.go.tmpl
```diff
@@ -2011,6 +2011,14 @@ func resourceSqlDatabaseInstanceUpdate(d *schema.ResourceData, meta interface{})
 
 	databaseVersion := d.Get("database_version").(string)
 
+	// Always read the instance before updating it.
+	// In cases where the instance was edited outside of Terraform, this will ensure that
+	// the we grab the latest settingsVersion in case we need to update the settings.
+	err = resourceSqlDatabaseInstanceRead(d, meta)
```
NickElliot: could you explain the scenario this is helping? as part of the terraform plan process, the resource is already refreshed. if a value was changed outside of terraform, for it to do anything here it would have to happen between plan generation and plan apply

NickElliot (commented):
for clarification, this scenario is fixing the "two step apply" right?

sorry for delays, due to major release preparation my reviews were deprioritized.

mmv1/third_party/terraform/services/sql/resource_sql_database_instance.go.tmpl
same-id: You are absolutely right, I am myself pondering about it, I commented about it here:
https://github.com/GoogleCloudPlatform/magic-modules/pull/14952#discussion_r2312435872

Maybe the best argument for it, that it will help `pulumi` and `terraform --refresh=false apply` scenarios and this resource already refreshes the state so many times on updates that one extra time should not hurt.

WDYT

same-id:
> for clarification, this scenario is fixing the "two step apply" right?
> 
> sorry for delays, due to major release preparation my reviews were deprioritized.

And not only "two step apply"

First you need to "revert the code change that changed the password", apply the configuration with only the "password_validation_policy" change in all environments (dev, staging, prod), then reintroduce the password change, and apply again in all environments - instead of just supporting all changes in the correct order, like it is done previously in the provider.

To be honest, I don't even like the fact that this resource support changing the root password in Update flows. The correct way to have done this is to treat the password as google does - an initial instance password, and then any change should have been made via the sql.User resource - but the horses have long bolted...

mmv1/third_party/terraform/services/sql/resource_sql_database_instance.go.tmpl
NickElliot: I am not familiar with pulumi so I do not understand the use case but doing this is already going against intended Terraform life cycle, and we especially shouldn't be superceding a `--refresh=false` flag like that. Please remove this section thanks!

same-id: I removed it, but I might still try to reintroduce it via https://github.com/GoogleCloudPlatform/magic-modules/pull/14952#discussion_r2312435872 where it is more needed (will expand my reasoning there)

For the context of this change we can just improve the order in the resource and add the tests.

same-id: Gentle ping

NickElliot:
sorry for the delays, going to spend time soon to research this situation more to make sure it works as expected

same-id:
Hi @NickElliot any news regarding this?

same-id:
Hi @roaks3 (saw you did some reviews in the vicinity)

Would appreciate a review here

PR #14979, opened 2025-08-25
Add Big Query Data Policy V2 resource

ankitiit84:
```release-note:new-resource
`google_bigquery_datapolicyv2_data_policy`
```


ankitiit84:
I ran the yaml formater but the check is still failing. Is there a different formater that I need to run?

wj-chen:
> I ran the yaml formater but the check is still failing. Is there a different formater that I need to run?

The file was created through the Terraform Autogen tool? @trodge for insights.

ankitiit84:
> > I ran the yaml formater but the check is still failing. Is there a different formater that I need to run?
> 
> The file was created through the Terraform Autogen tool? @trodge for insights.

It was created with Autogen but I had to make a few changes on top (add doc links, decoder, examples, add `default_from_api` tag etc. )

trodge:
> > > I ran the yaml formater but the check is still failing. Is there a different formater that I need to run?
> > 
> > 
> > The file was created through the Terraform Autogen tool? @trodge for insights.
> 
> It was created with Autogen but I had to make a few changes on top (add doc links, decoder, examples, add `default_from_api` tag etc. )

Looks like just a difference in how the yaml formatter you're using is configured compared to our yaml lint checker. As long as it's valid yaml we can either ignore the lint-yaml check or figure out a way to disable it for this resource.

As another option, you could configure your formatter to indent array elements like this:
```
some_array:
  - element_1
  - element_2
```
instead of like this:
```
some_array:
- element_1
- element_2
```
I believe both options are valid yaml, but the first is more common in our repo.

ankitiit84:
> > > > I ran the yaml formater but the check is still failing. Is there a different formater that I need to run?
> > > 
> > > 
> > > The file was created through the Terraform Autogen tool? @trodge for insights.
> > 
> > 
> > It was created with Autogen but I had to make a few changes on top (add doc links, decoder, examples, add `default_from_api` tag etc. )
> 
> Looks like just a difference in how the yaml formatter you're using is configured compared to our yaml lint checker. As long as it's valid yaml we can either ignore the lint-yaml check or figure out a way to disable it for this resource.
> 
> As another option, you could configure your formatter to indent array elements like this:
> 
> ```
> some_array:
>   - element_1
>   - element_2
> ```
> 
> instead of like this:
> 
> ```
> some_array:
> - element_1
> - element_2
> ```
> 
> I believe both options are valid yaml, but the first is more common in our repo.

Thanks for the details. I indented yaml files manually and seems like linter is happy now.

mmv1/templates/terraform/examples/bigquery_datapolicyv2_datapolicy_withgrantees.tf.tmpl
```diff
@@ -0,0 +1,10 @@
+resource "google_bigquery_datapolicyv2_data_policy" "{{$.PrimaryResourceId}}" {
+  location         = "us-central1"
+  data_policy {
+    data_policy_type = "RAW_DATA_ACCESS_POLICY"
+    grantees = [
+      "principal://goog/subject/mrankit@google.com"
```
wj-chen: Can we use a more generic name here (e.g. "jane@example.com" used by https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/bigquery_datapolicy_data_policy_iam today), this will show up verbatim in the public documentation, see this page generated from this PR: https://github.com/modular-magician/terraform-provider-google/blob/26842f56067c17880dbd358cf7713f9c1419c1b5/website/docs/r/bigquery_datapolicyv2_data_policy.html.markdown

mmv1/products/bigquerydatapolicyv2/DataPolicy.yaml
```diff
@@ -0,0 +1,172 @@
...
+  example_config_body: 'templates/terraform/iam/iam_attributes.go.tmpl'
+  import_format:
+    - 'projects/{{project}}/locations/{{location}}/dataPolicies/{{data_policy_id}}'
+    - '{{data_policy_id}}'
+custom_code:
+  decoder: templates/terraform/decoders/bigquery_datapolicyv2_datapolicy.go.tmpl
+examples:
+  - name: 'bigquery_datapolicyv2_datapolicy_basic'
+    primary_resource_id: 'basic_data_policy'
+    primary_resource_name: 'fmt.Sprintf("tf_test_data_policy%s", context["random_suffix"])'
```
wj-chen: This needs to match `primary_resource_id` to be `..."tf_test_basic_data_policy%s"...`, otherwise the IAM tests can't find the right resource to check. Same applies to the other `primary_resource_name`s here.

mmv1/templates/terraform/examples/bigquery_datapolicyv2_datapolicy_withgrantees.tf.tmpl
ankitiit84: Yes that makes sense. But the auto-generated test will not work after that since IAM will not be able to resolve the principal. AFAIK `examples` is used for both public documentation and auto generated tests.

wj-chen: So the API validates the IAM entities here? Is it possible to use `domain:google.com` or something else generic? If not, we can separate the test and the doc, add `exclude_docs: true` for a test-only example, and `exclude_test: true` for a doc-only example.

ankitiit84: Done.

mmv1/products/bigquerydatapolicyv2/DataPolicy.yaml
ankitiit84: I made the changes, though I couldn't find this guidance in official docs https://googlecloudplatform.github.io/magic-modules/reference/resource/#examples-1


I have run all the auto generated tests including IAM tests and they passed, so I believe change is right. Let me know if I am not doing it right.

mmv1/products/bigquerydatapolicyv2/DataPolicy.yaml
```diff
@@ -0,0 +1,179 @@
+# Copyright 2024 Google Inc.
```
wj-chen: nit: 2025, here and in the other YAML. Looks like the Autogen tool needs to update this...

wj-chen (approved):

mmv1/products/bigquerydatapolicyv2/DataPolicy.yaml
wj-chen: \+ @trodge about the question on documentation, and we also need another /gcbrun to check the tests.

mmv1/products/bigquerydatapolicyv2/DataPolicy.yaml
ankitiit84: Done

mmv1/products/bigquerydatapolicyv2/DataPolicy.yaml
trodge: The requirement for `primary_resource_name` to generate IAM tests is not documented and we're currently working on removing that field in favor of parsing the value from state. See https://github.com/hashicorp/terraform-provider-google/issues/8712

trodge (changes requested):
Everything looks reasonable so far, but I would like to see an update test for the fields that support update in place. See https://googlecloudplatform.github.io/magic-modules/test/test/#add-an-update-test for instructions.

ankitiit84:
> Everything looks reasonable so far, but I would like to see an update test for the fields that support update in place. See https://googlecloudplatform.github.io/magic-modules/test/test/#add-an-update-test for instructions.

Done. Thanks for asking this, this surfaced a bug as well. I had introduced a decoder earlier but update couldn't work with that (since backend expects update_mask values for flattened fields whereas decoder was always nesting the resource under data_policy, so update_mask was always set as "data_policy"). To resolve that I removed decoder and instead added encoder and update_encoder. This also better align TF resource with API resource.


All tests passed locally.

trodge (changes requested):
It looks like to fix the remaining failure you will need to add this new service to both `mmv1/third_party/terraform/.teamcity/components/inputs/services_ga.kt` and `mmv1/third_party/terraform/.teamcity/components/inputs/services_beta.kt`.

wj-chen:
> It looks like to fix the remaining failure you will need to add this new service to both `mmv1/third_party/terraform/.teamcity/components/inputs/services_ga.kt` and `mmv1/third_party/terraform/.teamcity/components/inputs/services_beta.kt`.

Could you share a bit more on what these configs are used for and if they are always required when registering a new service in the provider? Are there documentations we can use for future reference?

trodge:
> > It looks like to fix the remaining failure you will need to add this new service to both `mmv1/third_party/terraform/.teamcity/components/inputs/services_ga.kt` and `mmv1/third_party/terraform/.teamcity/components/inputs/services_beta.kt`.
> 
> Could you share a bit more on what these configs are used for and if they are always required when registering a new service in the provider? Are there documentations we can use for future reference?

They are used for setting up our nightly tests in Teamcity, and are unfortunately not well documented at the moment, but they are only required when adding a new service.

trodge:
On an unrelated note, is the `etag` field meant to be set by the user? If it is, can we add a test for it?

ankitiit84:
> On an unrelated note, is the `etag` field meant to be set by the user? If it is, can we add a test for it?

etag generated by backend server, customers need to pass it during update operations to avoid race conditions. My understanding was since this field is marked as `default_from_api: true` so it will automatically be passed by TF. And I can see that happening from the logs in custom update test `TestAccBigqueryDatapolicyv2DataPolicy_bigqueryDatapolicyv2DatapolicyBasicExample_update`.

ankitiit84:
> > > It looks like to fix the remaining failure you will need to add this new service to both `mmv1/third_party/terraform/.teamcity/components/inputs/services_ga.kt` and `mmv1/third_party/terraform/.teamcity/components/inputs/services_beta.kt`.
> > 
> > 
> > Could you share a bit more on what these configs are used for and if they are always required when registering a new service in the provider? Are there documentations we can use for future reference?
> 
> They are used for setting up our nightly tests in Teamcity, and are unfortunately not well documented at the moment, but they are only required when adding a new service.

Done

trodge:
> > On an unrelated note, is the `etag` field meant to be set by the user? If it is, can we add a test for it?
> 
> etag generated by backend server, customers need to pass it during update operations to avoid race conditions. My understanding was since this field is marked as `default_from_api: true` so it will automatically be passed by TF. And I can see that happening from the logs in custom update test `TestAccBigqueryDatapolicyv2DataPolicy_bigqueryDatapolicyv2DatapolicyBasicExample_update`.

Could we use the `Fingerprint` type instead of `String`?
Example here: https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/products/firestore/Database.yaml#L259-L265

Not sure if this field behaves differently than the one in the firestore API.

ankitiit84:
> > > On an unrelated note, is the `etag` field meant to be set by the user? If it is, can we add a test for it?
> > 
> > 
> > etag generated by backend server, customers need to pass it during update operations to avoid race conditions. My understanding was since this field is marked as `default_from_api: true` so it will automatically be passed by TF. And I can see that happening from the logs in custom update test `TestAccBigqueryDatapolicyv2DataPolicy_bigqueryDatapolicyv2DatapolicyBasicExample_update`.
> 
> Could we use the `Fingerprint` type instead of `String`? Example here: https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/products/firestore/Database.yaml#L259-L265
> 
> Not sure if this field behaves differently than the one in the firestore API.

Done

PR #14993, opened 2025-08-26
`resourceIdentity`: testing generation template

BBBmau:

```release-note:REPLACEME

```


c2thorn (commented):
This is a ton of tests that we are now duplicating the first example of. I have some concerns like:

- What is the overall test time added to our test suite with 500+ new tests?
- Should we be introducing resource identity + tests to a subset first?
- Can we just append the resource identity step as an extra test step to the first generated test? Instead of a completely new test that spins up new real resources?

BBBmau:
> * What is the overall test time added to our test suite with 500+ new tests?
it's considerably long, we can look into adding this to an existing example such as the first one like you mentioned

> * Should we be introducing resource identity + tests to a subset first?

are you talking about only having certain resources containing resource identity to start out? I'd wonder how'd we determine which ones to move forward with.

> * Can we just append the resource identity step as an extra test step to the first generated test? Instead of a completely new test that spins up new real resources?

We can do that, something like adding a third step followiing the regular import can work
```hcl
func TestAccPubsubTopic_pubsubTopicBasicExample(t *testing.T) {
	t.Parallel()

		"random_suffix": acctest.RandString(t, 10),
	}

		CheckDestroy:             testAccCheckPubsubTopicDestroyProducer(t),
		Steps: []resource.TestStep{
			{
				Config: testAccPubsubTopic_pubsubTopicBasicExample(context),
			},
			{
				ResourceName:            "google_pubsub_topic.example",
				ImportStateVerifyIgnore: []string{"labels", "terraform_labels"},
			},
			{
				ResourceName:    "google_sql_user.user1",
				ImportState:     true,
				ImportStateKind: resource.ImportBlockWithResourceIdentity,  <--- this third step is what we can add
			},
		},
	})
}
```

c2thorn:
> > * Should we be introducing resource identity + tests to a subset first?
> 
> are you talking about only having certain resources containing resource identity to start out? I'd wonder how'd we determine which ones to move forward with.

In generation we could filter out products alphabetically. Mostly a suggestion just to avoid the sheer amount of tests needed to be added/run per PR.

> > * Can we just append the resource identity step as an extra test step to the first generated test? Instead of a completely new test that spins up new real resources?
> 
> We can do that, something like adding a third step followiing the regular import can work

great, that seems like a much better way forward and should be a much smaller time addition

c2thorn (commented):
yeah looks like VCR timed out, we'll need to batch the tests by product letter

PR #15008, opened 2025-08-28
Update provider generation to use generic naming for imports

roaks3:
This slightly changes the logic for how we replace imports throughout the generated provider to match the provider version. Basically, besides the GA version, imports will use a simple pattern based on the version, instead of hardcoded values.

* GA - no impact or diff
* beta - no impact or diff
* private - `github.com/hashicorp/terraform-provider-google-private` will be used instead of `internal/terraform-next`, and we will use a simple find/replace function to adjust those imports during terraform-next generation (there is already a matching internal CL)

```release-note:none

```


roaks3:
Internal CL has been merged to support this

PR #15010, opened 2025-08-28
publicca: use `RawURLEncoding` instead of `URLEncoding` for unpadded base64 encoding

ramonvermeulen:
closes: https://github.com/hashicorp/terraform-provider-google/issues/24156

```release-note:bug
publicca: use `RawURLEncoding` instead of `URLEncoding` for unpadded base64 encoding
```


mmv1/third_party/terraform/verify/validation.go
```diff
@@ -290,7 +290,7 @@ func ValidateBase64String(i interface{}, val string) ([]string, []error) {
 }
 
 func ValidateBase64URLString(i interface{}, val string) ([]string, []error) {
```
hao-nan-li, line 292: This change seems to affect all resources. Is this something intended?

ramonvermeulen: Good point, when I am at the desk I will take a look at the usage. Might be a safer approach to add an extra validation method for the raw encoding (without padding).

**Small update:**
Let me know what you like, it seems that `ExternalAccountKey.yaml` is the only resource reference using this `ValidateBase64URLString` method.

Image

hao-nan-li: ok sounds good.

champtar:
@ramonvermeulen The API returns padded base64, using `RawStdEncoding` is incorrect, tried to report it here https://github.com/hashicorp/terraform-provider-google/issues/25799 but without success

ramonvermeulen:
> @ramonvermeulen The API returns padded base64, using `RawStdEncoding` is incorrect, tried to report it here https://github.com/hashicorp/terraform-provider-google/issues/25799 but without success

Interesting one, because that means it would be in contradiction with this issue https://github.com/hashicorp/terraform-provider-google/issues/24156 that was solved with this PR. 

Did the API potentially change in the meanwhile? If that is really the case I guess your the best way moving forward would be to revert the change in this PR?

champtar:
`b64_mac_key` is the raw response from the API, and it's basically `URLEncoding(RawUrlEncoding(binary key))`

What people want is just `mac_key` == `RawUrlEncoding(binary key)`, but following some misunderstanding we instead got `b64url_mac_key` which is `RawURLEncoding(RawUrlEncoding(binary key))` and not always working

I will send a fix + new `mac_key` output

champtar:
https://github.com/GoogleCloudPlatform/magic-modules/pull/16198

PR #15018, opened 2025-08-29
`resourceIdentity`: testing generation; products A-C

BBBmau:

```release-note:REPLACEME

```


BBBmau:
there shouldn't be any skipped tests since we moved away from a single `TestAcc` for resourceIdentity and instead just add it as a step in the first example. looking into this.

PR #15024, opened 2025-08-29
Add deletion_protection field to Redis Instance

rajeshgupthar:
Part of b/367733375

```release-note:enhancement
redis: added `deletion_protection` field to `redis_instance` to make deleting them require an explicit intent. `redis_instance` resources now cannot be destroyed unless `deletion_protection = false` is set for the resource.
```

### DIFF
--- a/mmv1/third_party/terraform/services/redis/resource_redis_instance_test.go
+++ b/mmv1/third_party/terraform/services/redis/resource_redis_instance_test.go
@@ -433,57 +433,21 @@ func TestAccRedisInstance_deletionprotection(t *testing.T) {
 				ExpectError: regexp.MustCompile("deletion_protection"),
 			},
 			{
-				Config: testAccRedisInstance_deletionprotectionFalse(name, "us-central1", true),
+				Config: testAccRedisInstance_deletionprotection(name, "us-central1", false),
 			},
 		},
 	})
 }
 
-func testAccRedisInstance_deletionprotection(name string, region string, preventDestroy bool) string {
-	lifecycleBlock := ""
-	if preventDestroy {
-		lifecycleBlock = `
-		lifecycle {
-			prevent_destroy = false
-		}`
-	}
+func testAccRedisInstance_deletionprotection(name string, region string, deletionProtection bool) string {
 	return fmt.Sprintf(`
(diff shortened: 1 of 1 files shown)


mmv1/products/redis/Instance.yaml
```diff
@@ -126,6 +129,17 @@ examples:
...
+  - name: 'deletion_protection'
+    description: |
+      Whether Terraform will be prevented from destroying the instance. Defaults to true.
+      When a`terraform destroy` or `terraform apply` would delete the instance,
+      the command will fail if this field is not set to false in Terraform state.
+      When the field is set to true or unset in Terraform state, a `terraform apply`
+      or `terraform destroy` that would delete the instance will fail.
+      When the field is set to false, deleting the instance is allowed.
+    type: Boolean
+    default_value: false
```
aniket-gupta: is this correct? Description says Defaults to true but default_value is set to false.

c2thorn: agreed plus see https://github.com/GoogleCloudPlatform/magic-modules/pull/15025#discussion_r2334925852

mmv1/third_party/terraform/services/redis/resource_redis_instance_test.go
```diff
@@ -407,3 +408,91 @@ resource "google_redis_instance" "test" {
...
+    maxmemory-policy       = "allkeys-lru"
+    notify-keyspace-events = "KEA"
+  }
+  redis_version = "REDIS_4_0"
+}
+`, name, region, lifecycleBlock)
+}
+
+func testAccRedisInstance_deletionprotectionFalse(name string, region string, preventDestroy bool) string {
+	lifecycleBlock := ""
```
c2thorn: Why do we have the lifecycle blocks here at all? They are just sent to `prevent_destroy = false` which does nothing. We should remove it entirely from both, and have a boolean input field that controls the `deletion_protection` field so we only need one `testAccRedisInstance_deletionprotection` function

SamanthaMathews: This is addressed In the new commit

PR #15025, opened 2025-08-29
Add deletion_protection field to Memcache Instance

rajeshgupthar:
Part of b/368232860

```release-note:enhancement
memcache: added `deletion_protection` field to `memcache_instance` to make deleting them require an explicit intent. `memcache_instance` resources now cannot be destroyed unless `deletion_protection = false` is set for the resource.
```


mmv1/products/memcache/Instance.yaml
```diff
@@ -330,3 +333,14 @@ properties:
...
+  - name: 'deletion_protection'
+    description: |
+      Whether Terraform will be prevented from destroying the instance. Defaults to true.
+      When a `terraform destroy` or `terraform apply` would delete the instance,
+      the command will fail if this field is not set to false in Terraform state.
+      When the field is set to true or unset in Terraform state, a `terraform apply`
+      or `terraform destroy` that would delete the instance will fail.
+      When the field is set to false, deleting the instance is allowed.
+    type: Boolean
+    default_value: false
```
aniket-gupta: Is this correct?

c2thorn: The only way to introduce this field without it being a breaking change is to make it default to false.

I would prefer to remove the default attribute entirely, as that is the standard way of assigning a false default.

mmv1/third_party/terraform/services/memcache/resource_memcache_instance_test.go
```diff
@@ -97,3 +98,93 @@ data "google_compute_network" "memcache_network" {
...
+	})
+}
+
+func testAccMemcacheInstance_deletionprotection(prefix, name, network, region string) string {
+	return fmt.Sprintf(`
+resource "google_memcache_instance" "test" {
+  name = "%s"
+  region = "%s"
+  authorized_network = data.google_compute_network.memcache_network.id
+  deletion_protection = true
```
c2thorn: if we change

```suggestion
  deletion_protection = %s
```
we can add a variable to move deletion protection to true / false. and remove the duplicate function

SamanthaMathews: This is addressed in the latest commit

c2thorn (approved):

PR #15028, opened 2025-08-29
Update google_bigquery_table schema change detection to take into account presence of row access policy

wj-chen:
Fixes b/385157254.

In-place table schema column drop was added in https://github.com/hashicorp/terraform-provider-google-beta/releases/tag/v5.25.0. However for a table that has row access policy, in-place schema update would result in `Error: googleapi: Error 400: Cannot DROP column from a table ... with Row Access Policy.` This PR adds a check for row access policy on the table if any top-level column drop is detected and determines whether to proceed with in-place update or revert to the old default behavior of forcing-new depending on the presence of row access policy.

```release-note:bug
bigquery: updated the schema change detection for `google_bigquery_table` to take into account presence of row access policy
```


c2thorn (approved):
thank you for adding to the tests for this.
I think generally we don't like calling API requests outside of the typical CRUD methods, but I think it should be warranted in the customize diff here. 

Waiting to for @sachinpro approval, but let me know if you want me to merge sooner @wj-chen

wj-chen:
> thank you for adding to the tests for this. I think generally we don't like calling API requests outside of the typical CRUD methods, but I think it should be warranted in the customize diff here.
> 
> Waiting to for @sachinpro approval, but let me know if you want me to merge sooner @wj-chen

Thanks for your review. The plan is to restrict this call to an edge case (table schema changed in a way that some columns may need to be dropped) when diffing.

@sachinpro Could you review it too?

mmv1/third_party/terraform/services/bigquery/resource_bigquery_table.go.tmpl
```diff
@@ -410,8 +444,10 @@ func resourceBigQueryTableSchemaCustomizeDiffFunc(d tpgresource.TerraformResourc
 			// same as above
 			log.Printf("[DEBUG] unable to unmarshal json customized diff - %v", err)
 		}
+		datasetId := d.Get("dataset_id").(string)
+		tableId := d.Get("table_id").(string)
 		_, isExternalTable := d.GetOk("external_data_configuration")
-		isChangeable, err := resourceBigQueryTableSchemaIsChangeable(old, new, isExternalTable, true)
+		isChangeable, err := resourceBigQueryTableSchemaIsChangeable(old, new, isExternalTable, true, config, project, datasetId, tableId)
```
sachinpro: To simplify this change, can we call `bigQueryTableHasRowAccessPolicy` outside `resourceBigQueryTableSchemaIsChangeable`? 

This way we won't need to pass `config, project, datasetId, tableId`(which are only passed for `bigQueryTableHasRowAccessPolicy`) to `resourceBigQueryTableSchemaIsChangeable`.

mmv1/third_party/terraform/services/bigquery/resource_bigquery_table.go.tmpl
```diff
@@ -271,9 +271,32 @@ func bigQueryTableNormalizePolicyTags(val interface{}) interface{} {
...
+		UserAgent: config.UserAgent,
+	})
+
+	if err != nil {
+		return false, err
+	}
+
+	if policies, ok := res["rowAccessPolicies"]; ok {
+		if policiesList, ok := policies.([]interface{}); ok && len(policiesList) > 0 {
+			log.Printf("[INFO] Table has row access policies, schema change will be forced to recreate.")
```
sachinpro: nit: `Table has row access policies, schema change will force us to re-create the table` sounds clearer to me.

mmv1/third_party/terraform/services/bigquery/resource_bigquery_table.go.tmpl
```diff
@@ -271,9 +271,32 @@ func bigQueryTableNormalizePolicyTags(val interface{}) interface{} {
 	return val
 }
 
+func bigQueryTableHasRowAccessPolicy(config *transport_tpg.Config, project, datasetId, tableId string) (bool, error) {
+	url := fmt.Sprintf("https://bigquery.googleapis.com/bigquery/v2/projects/%s/datasets/%s/tables/%s/rowAccessPolicies", project, datasetId, tableId)
+	res, err := transport_tpg.SendRequest(transport_tpg.SendRequestOptions{
+		Config:    config,
+		Method:    "GET",
```
sachinpro, line 278: Do we really need to get the table from the server or can we assume that the "old" terraform state will have the same row access policies? I understand your approach is probably more robust.

sachinpro: I get why you need a separate call to read the policies. I earlier assumed that it is a part of schema, so you can do a comparison within `resourceBigQueryTableSchemaIsChangeable` using the old schema.

This comment is resolved.

mmv1/third_party/terraform/services/bigquery/resource_bigquery_table.go.tmpl
wj-chen: Done.

mmv1/third_party/terraform/services/bigquery/resource_bigquery_table.go.tmpl
wj-chen: I want to restrict row access policy API calls to only happen when a specific condition is met (new schema has dropped columns) and not every time a diff is generated, and that condition is checked inside `isChangeable`, so the invocation needs to happen inside `isChangeable`, but I since updated to pass the row access policy function itself inside `isChangeable` instead of individual params, thanks for the suggestion and hope it's cleaner now.

sachinpro: Thanks! It looks a bit cleaner now.

sachinpro (approved):

PR #15036, opened 2025-08-30
Fixes issue #24052 related to google_chronicle_reference_list resource

ankitgoyal0301:

```release-note:note
chronicle: made the `scope_info` field in `google_chronicle_reference_list` configurable
```


SirGitsalot (approved):
Adding a `override-breaking-change` label because it appears to be a false positive. The detection is correct: the `scope_info` field is now `MaxItems: 1` where it was unset before, but that change seems to be caused by the removal of `output: true` (that is, AFAICT `MaxItems: 1` *should* have been present before).  In any event, `MaxItems: 1` on a `NestedObject` is correct.

PR #15037, opened 2025-08-30
dialogflow: support non-global endpoints for `google_dialogflow_conversation_profile`

ramonvermeulen:
Closes: https://github.com/hashicorp/terraform-provider-google/issues/24132

```release-note:bug
dialogflow: added support for non-global endpoints for `google_dialogflow_conversation_profile`
```

Took over example from https://github.com/GoogleCloudPlatform/magic-modules/blob/b6cb99b6361868d2885c75566314192ebff32fd6/mmv1/products/dialogflow/EncryptionSpec.yaml#L36
Where it uses this pre_create method to overwrite the `base_url`.

### DIFF
--- a/mmv1/templates/terraform/pre_create/dialogflow_set_endpoint.go.tmpl
+++ b/mmv1/templates/terraform/pre_create/dialogflow_set_endpoint.go.tmpl
@@ -1,7 +1,11 @@
 location := d.Get("location").(string)
+universeDomain := config.UniverseDomain
 
-// insert location into url for a different endpoint.
-if strings.HasPrefix(url, "https://dialogflow.googleapis.com/v2/") {
+if universeDomain != "" && universeDomain != "googleapis.com" {
+  url = strings.Replace(url, "googleapis.com", universeDomain, 1)
+}
+
+if strings.HasPrefix(url, "https://dialogflow") {
     if location != "" && location != "global" {
         url = strings.Replace(url, "https://dialogflow", fmt.Sprintf("https://%s-dialogflow", location), 1)
     }


ramonvermeulen:
I was not able to run all tests locally, because I hit the project quota on my testing org. Therefore locally I changed the test slightly to use a predefined project which seemed to work:

```
❮ make testacc TEST=./google/services/dialogflow TESTARGS='-run=TestAccDialogflowConversationProfile_regional'

TF_ACC=1 TF_SCHEMA_PANIC_ON_ERROR=1 go test ./google/services/dialogflow -v -run=TestAccDialogflowConversationProfile_regional -timeout 240m -ldflags="-X=github.com/hashicorp/terraform-provider-google/version.ProviderVersion=acc"
=== RUN   TestAccDialogflowConversationProfile_regional
=== PAUSE TestAccDialogflowConversationProfile_regional
=== CONT  TestAccDialogflowConversationProfile_regional
--- PASS: TestAccDialogflowConversationProfile_regional (148.85s)
PASS
ok      github.com/hashicorp/terraform-provider-google/google/services/dialogflow       149.633s
```

Waiting for VCR to review all the tests.

mmv1/templates/terraform/pre_create/dialogflow_set_endpoint.go.tmpl
```diff
@@ -2,5 +2,7 @@ location := d.Get("location").(string)
 
 // insert location into url for a different endpoint.
 if strings.HasPrefix(url, "https://dialogflow.googleapis.com/v2/") {
```
c2thorn: This is already a bug, but this endpoint needs to be made universe domain-aware.

Can you do something like the following:

https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/third_party/terraform/services/cloudasset/data_source_google_cloud_asset_resources_search_all.go.tmpl#L100

c2thorn (changes requested):

mmv1/templates/terraform/pre_create/dialogflow_set_endpoint.go.tmpl
melinath: will this handle custom endpoints properly as well? (Probably worth doing either way.)

ramonvermeulen: Check, thanks! I already had the feeling this might not be the desired approach. I will look into this (probably) later this week.

melinath: > will this handle custom endpoints properly as well? (Probably worth doing either way.)

Responding to myself having slept on it: I think we should be fine with the solution above. The thing we lose is being able to specify separate custom endpoints for regional vs non-regional resources, but I'm not sure folks actually need to be able to do that.

ramonvermeulen: Would this be the appropriate approach here?:
https://github.com/GoogleCloudPlatform/magic-modules/pull/15037/commits/382f5e953090a563f43e855389df71ddf0466286

c2thorn: yes, that looks correct

c2thorn:
pulling this and trying it out locally just to be sure

PR #15038, opened 2025-08-31
Add auto_upgrade_enabled setting to google_sql_database_instance.

ankushkhanna1998:
This PR adds support for the new field `auto_upgrade_enabled` to `google_sql_database_instance` settings.

```release-note:enhancement
sql: added `auto_upgrade_enabled` field to `google_sql_database_instance` resource.
```

### DIFF
--- a/mmv1/third_party/terraform/services/sql/resource_sql_database_instance.go.tmpl
+++ b/mmv1/third_party/terraform/services/sql/resource_sql_database_instance.go.tmpl
@@ -425,6 +425,12 @@ API (for read pools, effective_availability_type may differ from availability_ty
 							Optional:    true,
 							Description: `Enables Vertex AI Integration.`,
 						},
+						"auto_upgrade_enabled": {
+							Type:        schema.TypeBool,
+							Optional:    true,
+							Default:     true,
+							Description: `Enables Automatic Version Upgrade feature. Can be used with MySQL only.`,
+						},
 						"enable_dataplex_integration": {
 							Type:             schema.TypeBool,
 							Optional:         true,
@@ -1487,7 +1493,7 @@ func expandSqlDatabaseInstanceSettings(configured []interface{}, databaseVersion
 		Tier:                          _settings["tier"].(string),
 		Edition:                       _settings["edition"].(string),
 		AdvancedMachineFeatures:       expandSqlServerAdvancedMachineFeatures(_settings["advanced_machine_features"].([]interface{})),
-		ForceSendFields:               []string{"StorageAutoResize", "EnableGoogleMlIntegration", "EnableDataplexIntegration", "RetainBackupsOnDelete"},
+		ForceSendFields:               []string{"AutoUpgradeEnabled", "StorageAutoResize", "EnableGoogleMlIntegration", "EnableDataplexIntegration", "RetainBackupsOnDelete"},
 		ActivationPolicy:              _settings["activation_policy"].(string),
 		ActiveDirectoryConfig:         expandActiveDirectoryConfig(_settings["active_directory_config"].([]interface{})),
 		DenyMaintenancePeriods:        expandDenyMaintenancePeriod(_settings["deny_maintenance_period"].([]interface{})),
@@ -1505,6 +1511,7 @@ func expandSqlDatabaseInstanceSettings(configured []interface{}, databaseVersion
 		PricingPlan:                   _settings["pricing_plan"].(string),
 		DeletionProtectionEnabled:     _settings["deletion_protection_enabled"].(bool),
 		EnableGoogleMlIntegration:     _settings["enable_google_ml_integration"].(bool),
+		AutoUpgradeEnabled:            _settings["auto_upgrade_enabled"].(bool),
 		EnableDataplexIntegration:     _settings["enable_dataplex_integration"].(bool),
 		RetainBackupsOnDelete:         _settings["retain_backups_on_delete"].(bool),
 		UserLabels:                    tpgresource.ConvertStringMap(_settings["user_labels"].(map[string]interface{})),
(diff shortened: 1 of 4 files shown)


mmv1/third_party/terraform/services/sql/resource_sql_database_instance.go.tmpl
```diff
@@ -1526,6 +1531,10 @@ func resourceSqlDatabaseInstanceCreate(d *schema.ResourceData, meta interface{})
 	cloneContext, cloneSource := expandCloneContext(d.Get("clone").([]interface{}))
 	pointInTimeRestoreContext := expandPointInTimeRestoreContext(d.Get("point_in_time_restore_context").([]interface{}))
 
+	if valueI, ok := d.GetOk("settings.0.auto_upgrade_enabled"); ok && !(valueI.(bool)) {
+		return fmt.Errorf("The setting auto_upgrade_enabled cannot be set to false.")
```
SirGitsalot: Why can't this be set to false? I would assume that setting it to false would behave the same way as not setting it at all.

mmv1/third_party/terraform/services/sql/resource_sql_database_instance.go.tmpl
```diff
@@ -516,6 +516,11 @@ API (for read pools, effective_availability_type may differ from availability_ty
 							Optional:    true,
 							Description: `Enables Vertex AI Integration.`,
 						},
+						"auto_upgrade_enabled": {
+							Type:        schema.TypeBool,
+							Optional:    true,
```
SirGitsalot, line 522: If this field is immutable you can add `ForceNew: true` and Terraform will take care of disallowing changes to the field.

Based on the code in the create and update methods, it might not be immutable though? It looks like maybe an instance can start out with `auto_upgrade_enabled` off and later have it turned on, but once it's turned on it can't be turned off. Is that right?

mmv1/third_party/terraform/services/sql/resource_sql_database_instance.go.tmpl
```diff
@@ -2591,9 +2604,25 @@ func resourceSqlDatabaseInstanceUpdate(d *schema.ResourceData, meta interface{})
 		instance.PscServiceAttachmentLink = d.Get("psc_service_attachment_link").(string)
 	}
 
+	if d.HasChange("settings.0.auto_upgrade_enabled") {
```
SirGitsalot: If this field can be enabled if disabled, but not disabled if enabled, then a custom diff that conditionally applies `ForceNew` might be a better approach. See `privateNetworkCustomizeDiff()` above for an example.

mmv1/third_party/terraform/services/sql/resource_sql_database_instance.go.tmpl
ankushkhanna1998: We want the field to behave this way:
- Can be `unset` at first.
(This is the same behavior as having `false` as the field's value, but this is not an explicitly put `false` value by the user)
- Once `set`, then cannot be `unset`.
- Can never be `set` to `false` explicitly.

We want to do this to make the field's behavior consistent with `gcloud`. The `gcloud` APIs are written this way.

mmv1/third_party/terraform/services/sql/resource_sql_database_instance.go.tmpl
ankushkhanna1998: Yes, that's correct. So, it's not exactly immutable in that sense. An instance can start out without `auto_upgrade_enabled` (the field being `unset`), and can later `set` it to `true`. But that's it, that's the only kind of change allowed.

mmv1/third_party/terraform/services/sql/resource_sql_database_instance.go.tmpl
ankushkhanna1998: Nice suggestion, thanks. Changed.
Please check and let me know if this would work the same way as before, and as intended, now.

mmv1/third_party/terraform/services/sql/resource_sql_database_instance.go.tmpl
SirGitsalot: Would it be acceptable to treat `false` and `unset` both as `unset` at create time rather than returning an error for `false`? Having `false` be an invalid value for a boolean field will be surprising to users, and won't play well with [Terraform Modules](https://developer.hashicorp.com/terraform/language/modules).

Then at update code could handle:
- Update `unset` to `false` = leave `unset`
- Update `false` to `unset` = update to `unset` (although this should never trigger, if both create and update will never set the field to false in the first place)
- Update `unset` to `true` = `true`
- Update `false` to `true` = `true` (ditto on the "should never trigger")
- Update `true` to `unset` = error
- Update `true` to `false` = error

This *should* work, since the default value for an unset boolean field is `false`. This meets Terraform users' expectations that a boolean field can take a `false` value while respecting the API's "explicit false can never be update to true" behavior.

ankushkhanna1998: Thanks, this looks nice. I've implemented this, please check.

mmv1/third_party/terraform/website/docs/r/sql_database_instance.html.markdown
```diff
@@ -407,6 +407,13 @@ The `settings` block supports:
 
 * `user_labels` - (Optional) A set of key/value user label pairs to assign to the instance.
 
+* `auto_upgrade_enabled` - Enables
```
SirGitsalot:
```suggestion
* `auto_upgrade_enabled` - (Optional) Enables
```

ankushkhanna1998: Done

PR #15050, opened 2025-09-02
Add google secure source manager hook resource

ericzzzzzzz:
Add  secure_source_manager hook resource and test

```release-note:new-resource
`google_secure_source_manager_hook`
```

### DIFF
--- a/mmv1/products/securesourcemanager/Hook.yaml
+++ b/mmv1/products/securesourcemanager/Hook.yaml
@@ -1,0 +1,155 @@
+# Copyright 2024 Google Inc.
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+---
+name: Hook
+description: Hook is a user-defined HTTP callback triggered by an event.
+references:
+  guides:
+    Official Documentation: https://cloud.google.com/secure-source-manager/docs/overview
+docs: null
+id_format: projects/{{project}}/locations/{{location}}/repositories/{{repository_id}}/hooks/{{hook_id}}
+base_url: projects/{{project}}/locations/{{location}}/repositories/{{repository_id}}/hooks?hook_id={{hook_id}}
+self_link: projects/{{project}}/locations/{{location}}/repositories/{{repository_id}}/hooks/{{hook_id}}
+import_format:
+  - projects/{{project}}/locations/{{location}}/repositories/{{repository_id}}/hooks/{{hook_id}}
+  - "{{hook_id}}"
+timeouts:
+  insert_minutes: 20
+  update_minutes: 20
+  delete_minutes: 20
+autogen_async: true
+update_verb: PATCH
+update_mask: true
+async:
+  actions:
+    - create
+    - delete
+  type: OpAsync
+  operation:
+    base_url: "{{op_id}}"
+  result:
+    resource_inside_response: false
+custom_code: null
+examples:
+  - name: secure_source_manager_hook_basic
+    primary_resource_id: basic
+    vars:
+      hook_id: my-basic-hook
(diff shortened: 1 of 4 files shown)


slevenick (commented):
A couple questions:
Did you check out the autogen tooling to generate this resource's schema automatically?
I'm seeing timeouts on running the tests, do you know why that would be? Does this resource take a long time to create?

ericzzzzzzz:
@slevenick I didn't now check the autogenerated schema, but the test was passing https://github.com/GoogleCloudPlatform/magic-modules/pull/12525 that one has been opened for a while and I closed it and opened this with the same content. The test projects are now running into quota issues.  securesourcemanager api seems also having issues now. And yes, test should take more than an hour.

ericzzzzzzz:
hi @slevenick  the tests are failing due to `Quota limit 'InstancePerProjectPerLocation' has been exceeded. Limit: 20 in region us-central1`, could you help check if there are any orphan securesourcemanagers from tests are not cleaned up? Could you help clean those manually?  
delete instance command: 

```bash
curl -X DELETE  -H "Authorization: Bearer $(gcloud auth print-access-token)" \
  https://securesourcemanager.googleapis.com/v1/projects/<project>/locations/<location>/instances/<instance_Id>
```

slevenick:
Ok they're being deleted. Can you kick the tests by pushing a new commit?

ericzzzzzzz:
Hi @slevenick tests are good now, please take a look when you have time. thanks!

mmv1/products/securesourcemanager/Hook.yaml
```diff
@@ -0,0 +1,155 @@
...
+      Determines if the hook disabled or not.
+      Set to true to stop sending traffic.
+  - name: events
+    type: Array
+    description: |
+      The events that trigger hook on.
+    item_type:
+      type: Enum
+      enum_values:
+        - UNSPECIFIED
```
slevenick: Remove `UNSPECIFIED` as it's not a valid value to specify

mmv1/products/securesourcemanager/Hook.yaml
```diff
@@ -0,0 +1,155 @@
...
+    sensitive: true
+    ignore_read: true
+    description: |
+      The sensitive query string to be appended to the target URI.
+  - name: pushOption
+    type: NestedObject
+    description: |
+      The trigger option for push events.
+    properties:
+      - name: branchFilter
```
slevenick, line 152: This should probably be required. Would a user specify pushOption without branchFilter?

ericzzzzzzz: this field is optional, `If empty or *, events for all branches are
          reported` , user can also update this field from '<some value>' to empty to make hook report events for all branches.

mmv1/products/securesourcemanager/Hook.yaml
ericzzzzzzz: removed.

mmv1/products/securesourcemanager/Hook.yaml
ericzzzzzzz: b/352536114#comment10

slevenick: Is it valid for a user to specify this in their TF config?
```
push_option {
}
```

Usually it's not, and empty objects in Terraform can end up being confusing, so we mark at least one of the nested fields as being required to prevent this. It's not 100% necessary, and there are some times when having the empty object is useful.

If you want to keep this field as Optional in Terraform for now, can you add a step to a test that sets `push_option {}` to make sure the behavior is what we want?

mmv1/products/securesourcemanager/Hook.yaml
```diff
@@ -0,0 +1,154 @@
+# Copyright 2024 Google Inc.
```
slevenick:
```suggestion
# Copyright 2025 Google Inc.
```

mmv1/products/securesourcemanager/Hook.yaml
ericzzzzzzz: ahhh thanks! let me try that.

ericzzzzzzz: test with 
```bash
push_option {
}
``` 
is good now after adding 

```
    allow_empty_object: true
    send_empty_value: true
    default_from_api: true
```

ericzzzzzzz: PTAL, let me know if you have any concerns. thanks

PR #15055, opened 2025-09-03
workbench: Fix issue with apply failing with labels was null

bcreddy-gcp:

```release-note:REPLACEME

```


bcreddy-gcp:
@zli82016: Fixes b/434918880

zli82016:
@bcreddy-gcp , can you add the test for the use case in b/434918880?

You can use the random value generated during terraform apply for the test. Example: https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/third_party/terraform/services/bigquery/resource_bigquery_dataset_test.go#L680

Thanks.

mmv1/products/workbench/Instance.yaml
```diff
@@ -612,10 +612,10 @@ properties:
     immutable: true
   - name: 'labels'
     type: KeyValueLabels
+    default_from_api: true
```
zli82016: Can you please remove this line? It is enough to remove the diff_suppress_func. Thanks.

zli82016:
Though it is detected as the breaking change to remove the diff suppress function, it is not a breaking change in this PR, as the labels model has been applied already.

For most use cases there are not diffs after upgrading the provider to the one without the diff suppress function.

The one time diffs are noticed only for two edge cases. Then the diffs are fixed by the Terraform apply.
1. The resource is created with a service label in the labels field and then the service label is deleted. Then upgrade the provider to the one without the diff suppress function
2. The resource is created with a service label in the labels field and then the labels block is deleted. Then upgrade the provider to the one without the diff suppress function.

zli82016:
[TestAccWorkbenchInstance_updatelabels](https://hashicorp.teamcity.com/buildConfiguration/TerraformProviders_GoogleCloud_GOOGLE_BETA_MMUPSTREAMTESTS_GOOGLEBETA_PACKAGE_WORKBENCH/463014?buildTab=tests&expandedTest=build%3A%28id%3A463014%29%2Cid%3A2000000000) passed.

PR #15057, opened 2025-09-03
Added validationConfig to healthcare FHIR store.

canuck3141:

```release-note:enhancement
healthcare: added `validation_config` to `google_healthcare_fhir_store` resource
```

AFAICT this `validationConfig` is GA based on [Google Cloud Healthcare Release Notes](https://cloud.google.com/healthcare-api/docs/release-notes#:~:text=It%20is%20now%20possible%20to%20validate%20a%20FHIR%20resource%20for%20a%20specific%20profile%20or%20for%20all%20profiles%20defined%20for%20a%20FHIR%20store%20to%20ensure%20that%20your%20FHIR%20resource%20conforms%20to%20one%20more%20profiles.) where it discusses profile validation happening in v1 without mention of beta or alpha.

### DIFF
--- a/mmv1/third_party/terraform/services/healthcare/resource_healthcare_fhir_store_test.go.tmpl
+++ b/mmv1/third_party/terraform/services/healthcare/resource_healthcare_fhir_store_test.go.tmpl
@@ -3,7 +3,6 @@ package healthcare_test
 import (
 	"fmt"
 	"path"
-	"slices"
 	"testing"
 
 	"github.com/hashicorp/terraform-provider-google/google/acctest"
@@ -310,7 +309,7 @@ func testAccCheckGoogleHealthcareFhirStoreUpdate(t *testing.T, pubsubTopic strin
 	}
 }
 
-// Returns true if the size and contents (order independent) of enabledImplementationGuides1 and enabledImplementationGuides2 are identical,
+// Returns true if the size and contents (order dependent) of enabledImplementationGuides1 and enabledImplementationGuides2 are identical,
 // else false.
 func checkEnabledImplementationGuidesArraysAreEqual(enabledImplementationGuides1 []string, enabledImplementationGuides2 []string) bool {
 	// Same length?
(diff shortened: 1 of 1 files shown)


canuck3141:
`TestAccHealthcarePipelineJob_healthcarePipelineJobBackfillExample` has been failing since May and is not related to this PR.

mmv1/third_party/terraform/services/healthcare/resource_healthcare_fhir_store_test.go.tmpl
```diff
@@ -277,3 +309,29 @@ func testAccCheckGoogleHealthcareFhirStoreUpdate(t *testing.T, pubsubTopic strin
 		return nil
 	}
 }
+
+// Returns true if the size and contents (order independent) of enabledImplementationGuides1 and enabledImplementationGuides2 are identical,
+// else false.
+func checkEnabledImplementationGuidesArraysAreEqual(enabledImplementationGuides1 []string, enabledImplementationGuides2 []string) bool {
```
shuyama1, line 314: Since this is implemented as list, I think the order matters, otherwise we risk a permadiff. https://googlecloudplatform.github.io/magic-modules/develop/diffs/#list-order

Do you happen to know if the API might return this list in a different order than it was sent?

canuck3141: Hi @shuyama1 !  I don't know if it returns the array in order because the docs don't say AFAICT.

So I assumed it could be returned in any order.  However, in terms of behavior of the FHIR store's validation, the arrays `["ig1", "ig2"]`and `["ig2", "ig1"]` should be identical, which is why I added the sorts.

Good point about the potential permadiff.

Go doesn't compare arrays easily, which is why I added the helper function.  Should I drop the `sort` and just compare the arrays for length and item equality and return false if array contents are the same but in different orders?

WDYT?

shuyama1: Yeah, I'd suggest that we drop the `sort` and let it fail if the order changes

canuck3141: @shuyama1 I updated the function `checkEnabledImplementationGuidesArraysAreEqual` to NOT sort to avoid the permadiff risk as suggested.  Please see the latest commit pushed to this PR for the changes.  Thanks!

canuck3141: Hi @shuyama1 - would you mind taking a look again and if everything looks good approving this PR?  I wanted to get it merged to `main` before Friday so that it could be included in the release on Tue Sept 16.

Thanks!

shuyama1:
I've removed the VCR cassettes so that the newly added tests will be rerun to confirm if theupdated check works. Re-triggering tests now

/gcbrun

PR #15079, opened 2025-09-08
Make allow_global_access conditionally immutable for INTERNAL_MANAGED forwarding rules

JamesDuncanNz:
- Force recreation when allow_global_access changes for INTERNAL_MANAGED load balancing scheme
- Add tests for allow_global_access updates with both INTERNAL and INTERNAL_MANAGED schemes
- INTERNAL scheme remains updatable while INTERNAL_MANAGED requires recreation

Docs reference: ["Note that for INTERNAL_MANAGED forwarding rules, this field cannot be changed after the forwarding rule is created."](https://cloud.google.com/compute/docs/reference/rest/v1/forwardingRules)

Fixes https://github.com/hashicorp/terraform-provider-google/issues/23095

```release-note:bug
compute: fixed `allow_global_access` to correctly be immutable for `google_compute_forwarding_rule` resources with load balancing scheme of INTERNAL_MANAGED
```

### DIFF
--- a/mmv1/templates/terraform/constants/compute_forwarding_rule.go.tmpl
+++ b/mmv1/templates/terraform/constants/compute_forwarding_rule.go.tmpl
@@ -12,6 +12,14 @@ func forwardingRuleCustomizeDiff(_ context.Context, diff *schema.ResourceDiff, v
 			}
 		}
 	}
+
+	// Force recreation if allow_global_access changes for INTERNAL_MANAGED load balancing scheme
+	if diff.Id() != "" && diff.HasChange("allow_global_access") {
+		if loadBalancingScheme, ok := diff.Get("load_balancing_scheme").(string); ok && loadBalancingScheme == "INTERNAL_MANAGED" {
+			diff.ForceNew("allow_global_access")
+		}
+	}
+
 	return nil
 }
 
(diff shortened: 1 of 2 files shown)


JamesDuncanNz:
- Correct, the current that users see is: 
```Terraform will perform the following actions:

  # google_compute_forwarding_rule.http will be updated in-place
  ~ resource "google_compute_forwarding_rule" "http" {
      ~ allow_global_access     = false -> true
        id                      = "projects/xxxxxxxx/regions/us-central1/forwardingRules/fr-test-rule"
        name                    = "fr-test-rule"
        # (32 unchanged attributes hidden)
    }

google_compute_forwarding_rule.http: Modifying... [id=projects/xxxxxxxx//regions/us-central1/forwardingRules/fr-test-rule]
╷
│ Error: Error updating ForwardingRule "projects/xxxxxxxx//regions/us-central1/forwardingRules/fr-test-rule": googleapi: Error 400: Invalid value for field 'resource.allowGlobalAccess': 'true'. Updating the allow-global-access flag is only supported for regional INTERNAL forwarding rules with backend service/target instance., invalid
│
│   with google_compute_forwarding_rule.http,
│   on main.tf line 55, in resource "google_compute_forwarding_rule" "http":
│   55: resource "google_compute_forwarding_rule" "http" {
```
- Correct (I think this aligns with the [Immutable fields](https://googlecloudplatform.github.io/magic-modules/best-practices/immutable-fields/#immutable-fields) guidance) 
- Changing the `allowGlobalAccess` field is a destructive operation. Due to the immutability of the field in the API, the existing forwarding rule must be destroyed and a new one created. This will cause a brief service interruption where the IP address is unassociated with a forwarding rule, and traffic will be dropped. After the new rule is created, the Load Balancer will be accessible to internal clients in other regions, which is the intended behavior when `allowGlobalAccess` is `True`.

PR #15081, opened 2025-09-08
make label updatable for google_compute_instance under initialize_params

chenir0219:

```release-note: enhancement
compute: labels under initialize_params are now updatable on `google_compute_instance`
```

### DIFF
--- a/mmv1/third_party/terraform/services/compute/resource_compute_instance_test.go.tmpl
+++ b/mmv1/third_party/terraform/services/compute/resource_compute_instance_test.go.tmpl
@@ -1260,6 +1260,12 @@ func TestAccComputeInstance_bootDiskUpdate(t *testing.T) {
 			},
 			{
 				Config: testAccComputeInstance_bootDiskUpdate(context2),
+				ConfigPlanChecks: resource.ConfigPlanChecks{
+          PreApply: []plancheck.PlanCheck{
+            // Check that the update is done in-place
+            plancheck.ExpectResourceAction("google_compute_instance.foobar", plancheck.ResourceActionUpdate),
+          },
+        },
 				Check: resource.ComposeTestCheckFunc(
 					testAccCheckComputeInstanceExists(
 						t, "google_compute_instance.foobar", &instance),


mmv1/third_party/terraform/services/compute/resource_compute_instance.go.tmpl
```diff
@@ -2732,6 +2732,31 @@ func resourceComputeInstanceUpdate(d *schema.ResourceData, meta interface{}) err
...
 	}
 
+	if d.HasChange("boot_disk") {
+  		//default behavior for the disk is to have the same name as the instance
+  		diskName := instance.Name
+  		if v := tpgresource.GetResourceNameFromSelfLink(d.Get("boot_disk.0.source").(string)); v != diskName {
+  			diskName = v
+  		}
+
+  		disk, err := config.NewComputeClient(userAgent).Disks.Get(project, zone, diskName).Do()
```
roaks3, line 2742: Can this fail for legitimate configs? It looks like `source` is not required, and I can't tell if that implies the disk is guaranteed to exist with the same name as the instance. If the disk can't be found, I think the update will fail, which is a bigger deal if that could occur when updating some other boot_disk field (like perhaps auto_delete).

mmv1/third_party/terraform/services/compute/resource_compute_instance_test.go.tmpl
```diff
@@ -1230,6 +1230,45 @@ func TestAccComputeInstance_attachDisk_forceAttach(t *testing.T) {
...
+				Config: testAccComputeInstance_bootDiskUpdate(context1),
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckComputeInstanceExists(
+						t, "google_compute_instance.foobar", &instance),
+				),
+			},
+			{
+				Config: testAccComputeInstance_bootDiskUpdate(context2),
+				Check: resource.ComposeTestCheckFunc(
+					testAccCheckComputeInstanceExists(
```
roaks3, line 1270: Do you happen to know if this check ensures the instance was not recreated? I suppose the server name would be changed?

mmv1/third_party/terraform/services/compute/resource_compute_instance.go.tmpl
chenir0219: Thanks for the question! According to [the official documentation](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/compute_instance#nested_boot_disk), it is not supported to specify the disk_name directly for the instance’s boot disk (the device_name is not the same as the disk_name). So, the disk name should always either come from the source field if provided, or be automatically set to match the instance name.

Given this, I don't think there are other valid cases where the disk name would differ. Please let me know if this makes sense!

mmv1/third_party/terraform/services/compute/resource_compute_instance_test.go.tmpl
chenir0219: Good point. Added a plan check.

mmv1/third_party/terraform/services/compute/resource_compute_instance.go.tmpl
chenir0219: https://screenshot.googleplex.com/9gtkp8Zbi3xghq6 "If not specified, the default is to use the name of the instance."

chenir0219:
My new commit only changed my test instead of the logic in code. All checks passed before.

However, this time VCR test failed on the latest commit due to a unchanged test 'TestAccComputeRegionBackendService_regionBackendServiceHaPolicyManualLeader_update'. And the error msg doesn't look like related to my change. ('Root object was present, but now absent.')

Can I know how to retry this failed test? @roaks3

chenir0219:
Maybe /retest or /re-run tests

mmv1/third_party/terraform/services/compute/resource_compute_instance.go.tmpl
roaks3: Thank you, that rationale sounds reasonable to me!

PR #15084, opened 2025-09-08
Update google.golang.org/api package to the latest version

printchard:

```release-note:note
provider: updated google.golang.org/api to v0.249.0
```

### DIFF
--- a/mmv1/third_party/terraform/go.mod
+++ b/mmv1/third_party/terraform/go.mod
@@ -37,8 +37,8 @@ require (
 	golang.org/x/oauth2 v0.30.0
 	google.golang.org/api v0.249.0
 	google.golang.org/genproto/googleapis/rpc v0.0.0-20250818200422-3122310a409c
-	google.golang.org/grpc v1.74.2
-	google.golang.org/protobuf v1.36.7
+	google.golang.org/grpc v1.75.0
+	google.golang.org/protobuf v1.36.8
 	gopkg.in/yaml.v2 v2.4.0
 )
 


printchard:
I am trying to update the dependencies in order to implement a feature requiring the new release of the Google APIs Go Client. I followed the steps described in the repo to update the dependencies and generated the new `go.mod` and `go.sum` files. 

All the changes to these files come from running `go get` and `go mod tidy` as specified in the docs.

I don't understand where the big diff is coming form. Generating the provider locally does not generate those diffs. I apologize for opening the PR with failing CI but I'd appreciate any insight from the maintainers as to why this might be happening.

NickElliot (commented):
I'm guessing your local repo is out of date in either TPG or in MMv1, that is the main reason for diffs being different from local and in CI

printchard:
Thanks for the quick reply. I made sure to be up to date in both repos. I even tried it right now on this [commit](https://github.com/GoogleCloudPlatform/magic-modules/commit/aa289e7096997fce4b51ec8f64370a64608edb2e), on both MMv1 and TPG and saw no diff (additional to the go.mod and go.sum themselves).

Additionally, the CI diff seems to be inserting the same imports on different files that don't need them.

mmv1/third_party/terraform/go.mod
```diff
@@ -1,6 +1,8 @@
 module github.com/hashicorp/terraform-provider-google
 
-go 1.23.0
+go 1.24.0
+
+toolchain go1.24.4
```
NickElliot: could you remove this toolchain addition?

printchard: I did it, and the generated diff still inserted import statements in unrelated files. I do not know what might be causing this issue.

mmv1/third_party/terraform/go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/hashicorp/terraform-provider-google
 
-go 1.23.0
+go 1.24.0
```
NickElliot, line 3: could you remove the go version bump?

NickElliot (changes requested):

mmv1/third_party/terraform/go.mod
printchard: It now fails during generation due to not executing `go mod tidy`. However, if i run the command, the version gets bumped to `1.24.0`.

printchard: Actually, the [dependency](https://github.com/googleapis/google-api-go-client/blob/v0.249.0/go.mod) I am trying to update now declares `1.24` as their min version. I am guessing that's the reason for the version upgrade.

printchard:
@NickElliot Hello Nick. I have updated the PR. The test failures seem unrelated to the changes made. These are mainly validation errors from different resources.

PR #15086, opened 2025-09-08
List and cache InstanceGroupManagers instead of getting them separately to improve node pool reconciliation time

martin-guillen:
Fixes https://github.com/hashicorp/terraform-provider-google/issues/23128

The previous changes to list node pools did not have nearly the expected speed up when handling a large amount of node pools due to InstanceGroupManagers being fetched separately for each node pool. Now they are all listed and cached and this has been verified to speed up reconciliation time when there is a large amount of node pools

```release-note:none
```

### DIFF
--- a/mmv1/products/firebaseapphosting/Domain.yaml
+++ b/mmv1/products/firebaseapphosting/Domain.yaml
@@ -55,6 +55,9 @@ examples:
       service_act_id: '"tf-test-domain-m"'
   - name: firebase_app_hosting_domain_full
     primary_resource_id: example
+    ignore_read_extra:
+      # This isn't guaranteed to be set right away.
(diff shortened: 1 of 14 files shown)


martin-guillen:
I need to change the IGM cache refresh to use Pages() instead of Do() so it processes all pages of the List call, only handling the first page is likely the cause of the test failures.

SirGitsalot (commented):
Re-running the tests now, a bunch failed due to IGM quota so it may have been bad luck running when a bunch of other tests were also creating IGMs (for comparison, I just checked and the test project is at 2% of quota).

martin-guillen:
The remaining test failures don't look to be related to my changes and most likely require test changes

PR #15099, opened 2025-09-09
Add ephemeral `google_client_config` resource

busser:
Fixes https://github.com/hashicorp/terraform-provider-google/issues/24318.

```release-note:enhancement
Added ephemeral `google_client_config` resource
```

### DIFF
--- a/mmv1/third_party/terraform/services/resourcemanager/ephemeral_google_client_config.go
+++ b/mmv1/third_party/terraform/services/resourcemanager/ephemeral_google_client_config.go
@@ -1,0 +1,142 @@
+package resourcemanager
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/hashicorp/terraform-plugin-framework/ephemeral"
+	"github.com/hashicorp/terraform-plugin-framework/ephemeral/schema"
+	"github.com/hashicorp/terraform-plugin-framework/types"
+	transport_tpg "github.com/hashicorp/terraform-provider-google/google/transport"
+)
+
+// Ensure the ephemeral resource satisfies the expected interfaces.
+var (
+	_ ephemeral.EphemeralResource              = &GoogleClientConfigEphemeralResource{}
+	_ ephemeral.EphemeralResourceWithConfigure = &GoogleClientConfigEphemeralResource{}
+)
+
+func GoogleEphemeralClientConfig() ephemeral.EphemeralResource {
+	return &GoogleClientConfigEphemeralResource{}
+}
+
+type GoogleClientConfigEphemeralResource struct {
+	providerConfig *transport_tpg.Config
+}
+
+type GoogleClientConfigEphemeralModel struct {
+	// Id could/should be removed in future as it's not necessary in the plugin framework
+	// https://github.com/hashicorp/terraform-plugin-testing/issues/84
+	Id            types.String `tfsdk:"id"`
+	Project       types.String `tfsdk:"project"`
+	Region        types.String `tfsdk:"region"`
+	Zone          types.String `tfsdk:"zone"`
+	AccessToken   types.String `tfsdk:"access_token"`
+	DefaultLabels types.Map    `tfsdk:"default_labels"`
+}
+
+func (e *GoogleClientConfigEphemeralResource) Metadata(ctx context.Context, req ephemeral.MetadataRequest, resp *ephemeral.MetadataResponse) {
+	resp.TypeName = req.ProviderTypeName + "_client_config"
+}
+
+func (e *GoogleClientConfigEphemeralResource) Schema(ctx context.Context, req ephemeral.SchemaRequest, resp *ephemeral.SchemaResponse) {
+
+	resp.Schema = schema.Schema{
+
+		Description:         "Use this ephemeral resource to access the configuration of the Google Cloud provider.",
+		MarkdownDescription: "Use this ephemeral resource to access the configuration of the Google Cloud provider.",
+		Attributes: map[string]schema.Attribute{
+			"id": schema.StringAttribute{
+				Computed:            true,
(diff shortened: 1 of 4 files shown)


ramonvermeulen:
@busser nice work, PR Looks good to me!

Since ephemeral resources are slightly harder to test, it might be worth to build the binary locally and also do a [manual test](https://googlecloudplatform.github.io/magic-modules/test/run-tests/#optional-test-manually) with a local config. To ensure the access_token and other attributes are set properly.

busser:
> it might be worth to build the binary locally and also do a [manual test](https://googlecloudplatform.github.io/magic-modules/test/run-tests/#optional-test-manually) with a local config

Yep, I ran it on one of our projects and it behaves as expected.

I just created a more contrived example that shows that the resource works correctly:

```terraform
provider "google" {}

ephemeral "google_client_config" "default" {}

resource "null_resource" "test" {
  # Create a file with all the fields in google_client_config so we can inspect them.
  provisioner "local-exec" {
    command = "echo '${jsonencode(ephemeral.google_client_config.default)}' > ${path.module}/google_client_config.json"
  }
}
```

I applied this code with this command:

```sh
GOOGLE_PROJECT=fake-project GOOGLE_REGION=fake-region GOOGLE_ZONE=fake-zone TF_CLI_CONFIG_FILE=./tf-dev-override.tfrc terraform apply
```

After which, `google_client_config.json` contains:

```json
{
  "access_token": "ya29.CENSORED",
  "default_labels": {},
  "id": "projects/fake-project/regions/fake-region/zones/fake-zone",
  "project": "fake-project",
  "region": "fake-region",
  "zone": "fake-zone"
}
```

Seems to work as expected :)

Side note: we could possibly use this trick to write automatic tests for ephemeral resources.

ramonvermeulen:
> Side note: we could possibly use this trick to write automatic tests for ephemeral resources.

Nice, looks good! Quite a neat approach for manually testing ephemeral resources. Just in case you didn't know, if your PR is ready and you upgrade it to `ready for review` someone from the core team will be automatically assigned.

---

After diving a bit deeper, it seems that the echo provider is the way hashicorp advices you to test ephemeral resources.

They have some documentation about it here:
https://developer.hashicorp.com/terraform/plugin/testing/acceptance-tests/ephemeral-resources#testing-ephemeral-data-with-echo-provider

Might see if I can incorporate that in my ephemeral secret manager secret version PR as well. I see that this does require to upgrade the `github.com/hashicorp/terraform-plugin-testing` dependency to at least [`v1.11.0`](https://github.com/hashicorp/terraform-plugin-testing/releases/tag/v1.11.0) or newer.

Any idea if there is something blocking us from upgrading this? @BBBmau @rileykarson @melinath 
Would be quite nice to use this functionality, however, I suspect there is a reason why this dependency is still pinned to `v1.5.1`.

melinath:
> Any idea if there is something blocking us from upgrading this? @BBBmau @rileykarson @melinath Would be quite nice to use this functionality, however, I suspect there is a reason why this dependency is still pinned to `v1.5.1`.

terraform-plugin-testing has a bug in it that blocked us from upgrading past 1.5.1: https://github.com/hashicorp/terraform-plugin-testing/issues/327

https://github.com/GoogleCloudPlatform/magic-modules/pull/14923 contains a workaround that would let us keep the current behavior, but as discussed on the PR I'd prefer to see the root bug fixed if possible. (If not we'll move forward with the workaround.)

busser:
Hey @BBBmau 👋 You seem to be the author of several existing ephemeral resources. Could you please have a look at this PR? All feedback is welcome!

busser:
Hey @melinath 👋 I can see that the new team in charge of the provider is now active. Any chance someone can take a look at this PR? :)

melinath:
@rileykarson FYI this can be reviewed now - the blocker has been resolved.

busser:
I can't view the error messages or debug logs; I don't have access to the `ci-vcr-logs` GCS bucket.

The breadth of impacted resources makes me suspect the failures may be unrelated to my changes.

Rebasing my PR in case the issue is drift from the main branch.

mmv1/third_party/terraform/services/resourcemanager/ephemeral_google_client_config.go
```diff
@@ -0,0 +1,142 @@
...
+		val := v
+		m[k] = &val
+	}
+	dls, diags := types.MapValueFrom(ctx, types.StringType, m)
+	resp.Diagnostics.Append(diags...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	data.DefaultLabels = dls
```
BBBmau: im confused about this new map being created, it uses the `DefaultLabels` which is already a map and loops through in order to store the key and value inside the new map `m` yet there isn't a change that's occurred? I imagine this is from a first iteration of this PR since i believe you can just utilize `e.providerConfig.DefaultLabels` as the argument in `types.MapValueFrom()`

busser: I copied this from the `google_client_config` datasource's implementation without giving it much thought. It seems like we can indeed pass `e.providerConfig.DefaultLabels` to `types.MapValueFrom()` directly, as you say. I ran a few local sanity checks and it seems to work the same.

I've updated the file I added as well as a couple other places in the codebase where we did the same thing.

BBBmau: makes sense to do, i noticed that `google_client_config` data source written using the `plugin-framework`.

feel free to open a PR that also updates the logic in the data-source so that it aligns here.

shuyama1:
Fixed VCR GKE issues and cleaned up VCR cassettes. Triggering a new run now. Sorry for the noise!

/gcbrun

BBBmau (changes requested):
one last request is to include this new ephemeral in the following documentation page
- https://registry.terraform.io/providers/hashicorp/google/latest/docs/guides/using_ephemeral_resources

busser:
> one last request is to include this new ephemeral in the following documentation page

Done 👍

BBBmau (approved):

PR #15101, opened 2025-09-09
add cai_base_url to bigtable product

zli82016:

```release-note:none

```

### DIFF
--- a/mmv1/products/bigtable/product.yaml
+++ b/mmv1/products/bigtable/product.yaml
@@ -17,5 +17,6 @@ display_name: 'Cloud Bigtable'
 versions:
   - name: 'ga'
     base_url: 'https://bigtableadmin.googleapis.com/v2/'
+    cai_base_url: 'https://bigtable.googleapis.com/v2/'
 scopes:
   - 'https://www.googleapis.com/auth/bigtable'


zli82016:
@roaks3 , can you help to confirm if [these changes](https://github.com/modular-magician/terraform-provider-google-beta/compare/53d553112fac4b3fc0ed221669edf71b0723a594..8041291f3984a8a389358eca5309b4885a41affe)  will break anything?  I just realized that adding `cai_base_url` will change the service name based on current logic. And `cai_base_url` is also applied to other products.

Thanks.

zli82016:
`api_service_name` for sql database instance is also changed in this PR.

roaks3:
Since they are both relatively old services, I'm suspecting that we will need to give them special treatment for CAI vs ServiceConfig matching.

mmv1/products/bigtable/product.yaml
```diff
@@ -17,5 +17,6 @@ display_name: 'Cloud Bigtable'
 versions:
   - name: 'ga'
     base_url: 'https://bigtableadmin.googleapis.com/v2/'
+    cai_base_url: 'https://bigtable.googleapis.com/v2/'
```
melinath, line 20: According to https://cloud.google.com/asset-inventory/docs/asset-types, Bigtable resources use bigtableadmin for their CAI asset types:

Screenshot 2025-09-11 at 4 03 38 PM

Do we have evidence that page is incorrect?

melinath: I also don't see any resources in the "bigtable" service, and it's not documented at https://cloud.google.com/bigtable/docs/apis

I'm fairly confident bigtableadmin is correct.

mmv1/third_party/terraform/services/sql/resource_sql_database_instance_meta.yaml
```diff
@@ -1,6 +1,6 @@
 resource: 'google_sql_database_instance'
 generation_type: 'handwritten'
-api_service_name: 'sqladmin.googleapis.com'
```
melinath, line 3: sqladmin is correct. [CAI](https://cloud.google.com/asset-inventory/docs/asset-types): 
Screenshot 2025-09-11 at 4 08 49 PM

https://cloud.google.com/sql/docs/mysql/admin-api/rest

melinath: I don't see any resources in the "cloudsql" API either.

melinath:
(In general I would expect CAI asset name and API resource type to be the same, but it's possible there are exceptions.)

mmv1/products/bigtable/product.yaml
zli82016: In the page CAI asset names page (https://cloud.google.com/asset-inventory/docs/asset-names), the asset name format for bigtable resources are 
```

bigtableadmin.googleapis.com/Backup | //bigtable.googleapis.com/projects/PROJECT_ID/instances/INSTANCE/clusters/CLUSTER/backups/BACKUP
-- | --
bigtableadmin.googleapis.com/Cluster | //bigtable.googleapis.com/projects/PROJECT_ID/instances/INSTANCE/clusters/CLUSTER


```

Also, in the CAI asset data fetched from the CAI API for the test resources, the asset names are 
```
"name": "//bigtable.googleapis.com/projects/ci-test-project-nightly-beta/instances/tf-test-yookbgvyus"
"name": "//bigtable.googleapis.com/projects/ci-test-project-nightly-beta/instances/tf-test-8182sa4e9k/clusters/tf-test-8182sa4e9k"
```

These seems to be annoying exceptions.

mmv1/third_party/terraform/services/sql/resource_sql_database_instance_meta.yaml
zli82016: In the page CAI asset names page (https://cloud.google.com/asset-inventory/docs/asset-names), the asset name format for sql resources are
```

sqladmin.googleapis.com/Backup | //cloudsql.googleapis.com/projects/PROJECT_ID/backups/BACKUP
-- | --
sqladmin.googleapis.com/BackupRun | //cloudsql.googleapis.com/projects/PROJECT_ID/instances/INSTANCE/backupRuns/BACKUPRUN
sqladmin.googleapis.com/Instance | //cloudsql.googleapis.com/projects/PROJECT_ID/instances/INSTANCE


```

Look like these are annoying exceptions.

zli82016:
The reason to modify the `cai_base_url` or `api_service_name` is that the `api_service_name` is needed to build the cai asset names for the resources provisioned during nightly tests runs. Then these asset names are used to call the API [batchGetAssetsHistory](https://cloud.google.com/asset-inventory/docs/reference/rest/v1/TopLevel/batchGetAssetsHistory)

If we have concerns or these changes could break launch blocker or coverage metrics, I can add a new field for cai service name.

melinath:
Thanks for the references, Zhenhua! It looks like I was using slightly wrong terminology before, sorry about that!

So it looks like the CAI _Resource Types_ use bigtableadmin and sqladmin - it's just the _asset names_ (i.e. resource paths) that use bigtable and cloudsql. +1 that these are annoying exceptions.

They definitely need to be bigtableadmin and sqladmin for metrics purposes, to match the API resource types.

I wonder if we've reached the point where we should take a step back and reconsider the way we're storing the data for CAI vs metrics. `base_url` was originally chosen because it was similar to the product & resource base_url fields, but it doesn't really map to CAI concepts well. For example, we could do:

```yaml
# product.yaml
(api_)service_name: sqladmin.googleapis.com

# ResourceName.yaml
cai_resource_type: whatever.googleapis.com/ResourceType  # Default: product service name + api_resource_type_kind
cai_asset_name: //cloudsql.googleapis.com/projects/{project}/instances/{instance}  # Default: //{product service name}/{{ base_url }}
```

I think something like this might be more clear than trying to provide cai_* overrides for every individual part of what makes up the resource type & asset name. What do you think?

zli82016:
No worries, Stephen. I should put the details and reason for the changes earlier in the description of this PR. 

Your suggestion makes sense to me. Thank you.

The following changes will be made:

ResourceName.yaml 
1. Add `cai_asset_name` if needed, for example for bigtable resources
2. We can consider deprecating `cai_base_url`

Resource_meta.yaml files
1. Add `cai_asset_name` to the template file [metadata.yaml.tmpl](https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/templates/terraform/metadata.yaml.tmpl) for generated resources.
2. Add `cai_asset_name` to the resource_meta.yaml files for handwritten resources

Let me know if you have any concerns, @roaks3 and @melinath 

Thanks.

melinath:
Why do the meta.yaml files need the cai_asset_name?

zli82016:
In providers, a check function is added for each test step to get the CAI asset names for the provisioned resources.

`cai_asset_name` from meta.yaml file will be read from meta.yaml file and then generate the asset names with Terraform state.

 For example, `cai_asset_name` in meta.yaml is `//bigtable.googleapis.com/projects/{{project}}/instances/{{name}}`. With the value of fields in Terraform state, the cai asset name will be `//bigtable.googleapis.com/projects/tf-test-project/instances/tf-test-instance`

zli82016:
Currently, in the check function, api service name + resource id from the Terraform state are used to generate the cai asset name. https://github.com/hashicorp/terraform-provider-google-beta/blob/main/google-beta/acctest/tgc_utils.go#L131


We found the issues that  `api service name + terraform resource id != cai asset names` sometimes.

melinath:
Thanks! Yeah, adding cai_asset_name to the meta.yaml files seems like a great solution.

zli82016:
The changes are made in https://github.com/GoogleCloudPlatform/magic-modules/pull/15414

PR #13422, opened 2025-03-21
Wbi metadata

bcreddy-gcp:
**Release Note Template for Downstream PRs (will be copied)**

See [Write release notes](https://googlecloudplatform.github.io/magic-modules/contribute/release-notes/) for guidance.

```release-note:REPLACEME

```

PR #15103, opened 2025-09-09
workbench: Prevent setting server-managed metadata fields in config

### LINKED ISSUE #24330: Workbench: Error out when a sever managed metadata key is present in the resource config
### Community Note

Terraform v1.5.7
on linux_amd64

+ provider registry.terraform.io/hashicorp/google vX.X.X
+ provider registry.terraform.io/hashicorp/google-beta vX.X.X

### Affected Resource(s)

google_workbench_instance

### Terraform Configuration

```tf
resource "google_workbench_instance" "instance" {
  name     = "tf-test-workbench-instance"
  location = "us-central1-a"
  gce_setup {
	metadata = {
        "proxy-url"  = "https://custom.proxy"
        "image-url"  = "https://custom.image"
        "custom-key" = "custom-value"
		"gce-software-declaration" = "custom-value"
	}
  }
  enable_managed_euc = true
}
```

### Debug Output

_No response_

### Expected Behavior

Error out when a sever managed metadata key is present in the resource config

### Actual Behavior

Sever managed metadata keys are ignored

### Steps to reproduce

1. `terraform apply`

### Important Factoids

_No response_

### References

_No response_

b/515829702

bcreddy-gcp:
Fixes https://github.com/hashicorp/terraform-provider-google/issues/24330

This is a breaking change targeted at the `8.0.0` major release. `google_workbench_instance` now returns a plan-time error when `gce_setup.metadata` contains a server-managed key, instead of silently accepting and overwriting it (which caused a permadiff). A `version_8_upgrade` guide entry documents the change.

```release-note:breaking-change
workbench: setting a server-managed key in `gce_setup.metadata` on `google_workbench_instance` now returns an error instead of being silently ignored
```


bcreddy-gcp:
Original PR got closed by mistake: https://github.com/GoogleCloudPlatform/magic-modules/pull/13422

Corresponding issue: https://github.com/hashicorp/terraform-provider-google/issues/24330

cc: @melinath

melinath:

melinath (commented):
Removing from my queue since there's not a major release for a while.

melinath (changes requested):
It looks like the current logic is breaking a bunch of tests. I'm not sure offhand whether the behavior is wrong or the tests need to change.

bcreddy-gcp:
I think the validation is currently checking if the get response has any of the protectedKeys. Instead we should be checking if the config has the protectedKeys. cc: @melinath

melinath:
> I think the validation is currently checking if the get response has any of the protectedKeys. Instead we should be checking if the config has the protectedKeys. cc: @melinath

yeah, that makes sense, since the field is only partially server managed.

Normally for fields like this, we'd do something like:

- Remove server-managed values in the flattener for this field. (You might be able to remove the diff suppress at that point.) Normally the client-managed field would be called `desired_metadata` but that's not required.
- If having access to the server-managed values is important, add a separate field that is output-only and contains all metadata values.

Would that work in this case?

melinath (commented):

c2thorn:
@bcreddy-gcp this should be moved to https://github.com/GoogleCloudPlatform/magic-modules/tree/FEATURE-BRANCH-major-release-8.0.0 or recreated there. Do you have time to do so before end of next week?

bcreddy-gcp:
I haven't worked on this PR is a while. I can take a look this week.

bcreddy-gcp:
Rebased onto current `main` and reconciled with the metadata handling that landed since this PR was opened (`metadataDefaults`, `WorkbenchInstanceEUCSettableUnmodifiableDefaultMetadata`, the `ForceNew` path, and #15998 / #15968 / #16650 making `post-startup-script[-behavior]`, `enable-jupyterlab4`, and `install-nvidia-driver` settable).

Key changes since the last review:

- **Root cause of the earlier test breakage:** the validation read the state-merged metadata (`diff.Get`), which includes the server-managed keys the API returns and stores in state, so it failed on every post-create plan. It now reads `diff.GetRawConfig()` and validates **only keys the user set in config**.
- **No overlap with the now-settable keys:** the reject-set (`WorkbenchInstanceProvidedMetadata` + EUC additional) does not include `install-nvidia-driver`, `enable-jupyterlab4`, or `post-startup-script[-behavior]`, so upstream's settable/ForceNew behavior is preserved. The new validation is layered above the existing `ForceNew` logic, not replacing it.
- Removed `resource-url` / `image-url` from two existing update-metadata test configs (they're server-managed and would now be rejected; nothing was asserted on them), and added `TestAccWorkbenchInstance_metadataValidation`.

Verified locally: GA + tgc generation, gofmt clean (source + generated), and `go build` + `go vet` of `google/services/workbench/...` in a `terraform-provider-google` checkout. Acceptance tests still need a maintainer test run.

@melinath — since `main` now also addresses the #24330 permadiff via diff-suppression, the remaining intent of this PR is to **hard-error when a user sets a server-managed key** instead of silently dropping it. Could you confirm that's still the direction you want before I chase a full test run?

melinath (changes requested):
yeah, that seems to be the remaining work here.

Like Cameron mentioned, this should be moved to be against the 8.0.0 release branch instead of main. https://googlecloudplatform.github.io/magic-modules/breaking-changes/make-a-breaking-change/ has guidance on making breaking changes. In particular, the release note should call out that it's a breaking change and the upgrade guide should explain the breaking change.

bcreddy-gcp:
Thanks both. Addressed the review:

- **Retargeted this PR to `FEATURE-BRANCH-major-release-8.0.0`** (was `main`) and rebased the branch onto it. The workbench `customizeDiff` and metadata lists are identical on `main` and the release branch, so the change applies cleanly (diff is 3 files, +105/-2).
- **Release note is now `breaking-change`.**
- **Added a `version_8_upgrade.html.markdown` entry** ("Setting server-managed `gce_setup.metadata` keys is now an error") explaining what users must remove when upgrading.

Verified locally: GA + tgc generation, gofmt (source + generated), `go build` of `google/services/workbench/...`, and an end-to-end `apply -> re-plan -> destroy` on a real project — the post-create re-plan is empty (no permadiff) and no longer errors, since the validation reads only `GetRawConfig` (user config), not the server-managed keys stored in state.

@melinath one thing worth a look: I compared the provider's protected lists against the backend source of truth (`cloud/hosted/notebooks/clh/common/BUILD` constgen) and they've drifted a bit. Most notably `enable-guest-attributes` is in the provider's protected list but the backend treats it as a user-configurable feature key, and a few EUC keys the backend protects (`container-custom-params`, `status-config-url`, `status-variable-path`) aren't in the provider list. Happy to align those in this PR or a follow-up — let me know your preference.

bcreddy-gcp:
One design concern I'd like opinions on before this lands.

This PR reuses the same lists (`WorkbenchInstanceProvidedMetadata` / `WorkbenchInstanceEUCProvidedAdditionalMetadata`) for **both** the existing `WorkbenchInstanceMetadataDiffSuppress` and the new hard error. That couples "suppress a permadiff" to "reject the key", which creates an awkward lifecycle for any *future* server-managed key:

1. The backend starts returning a new managed key `K`. It isn't in the provider list yet, so users get a **permadiff** (the API returns `K`, their config doesn't have it).
2. The natural workaround is to add `K` to `gce_setup.metadata` so config matches state.
3. We later add `K` to `WorkbenchInstanceProvidedMetadata` to suppress the permadiff.
4. With this PR, step 3 *also* makes `K` a hard error — so every user who applied the step-2 workaround now has a previously-valid config that **fails to plan**.

Net effect: **adding any new key to the protected list becomes a breaking change**, and it breaks precisely the users who followed the permadiff workaround. Today (silent-suppress behavior) growing the list is non-breaking.

Options I can see:

- **Decouple the lists**: keep a broad *suppress* list (new managed keys stop causing permadiffs, non-breaking) and a separate, narrower *reject* list that only grows during major releases with upgrade-guide entries. New keys are suppress-only by default and only promoted to reject deliberately.
- **Warn instead of error**, so growing the list is never breaking. Caveat: SDKv2 `CustomizeDiff` can only return an error, not a warning; a field-level `ValidateDiagFunc` on `metadata` could emit warnings but can't see `enable_managed_euc`, so the EUC-conditional keys couldn't be handled there.
- **Keep them coupled** and treat every list addition as a major-release breaking change (deprecation warning on `main` first, then error in the next major). Correct per policy, but heavy, and the treadmill remains.

I lean toward decoupling (suppress != reject), but wanted your take before I change anything. @melinath @c2thorn wdyt?

bcreddy-gcp:
Following up on my own decoupling suggestion above — I don't think it actually solves the problem, so I want to correct course before touching any code.

**Decoupling only changes the cadence, not the fact of the breakage.** The harm happens the moment a key *joins the reject list* — that breaks every config that already has the key. A separate "reject" list grown only in majors just moves those additions into major releases; the treadmill (and the "we break exactly the users who followed the permadiff workaround") is still there.

**The tension looks intrinsic to hard-erroring:** rejecting key `K` is breaking for precisely the users who have `K` in their config, and the keys worth rejecting are the server-managed ones that *cause* permadiffs — i.e. exactly the keys users add to config to silence those permadiffs. So the reject set and the "already in real configs" set overlap by construction. Any mechanism that grows the reject set will break real configs, no matter which list it lives in.

So the real question is whether a hard error is the right primitive at all. The two directions that are non-breaking *by nature* both stop erroring:

1. **Warn instead of error** — growing the list never breaks anyone. Caveat: SDKv2 `CustomizeDiff` can't emit a warning, and a field `ValidateDiagFunc` on `metadata` can't see `enable_managed_euc`, so EUC-conditional keys couldn't be covered there — it'd be "warn on the base keys, best-effort."
2. **Just suppress/ignore robustly and drop the error** — fix the permadiff so users never need the workaround; accept that setting a managed key is a silent no-op (documented), optionally with a warning.

If instead we want to keep the hard error, I think it has to be an explicit one-time 8.0.0 cleanup with the reject list **frozen** afterward (future managed keys are suppress-only forever) — otherwise every future addition is a breaking change.

@melinath @c2thorn — before I rework anything: do you want this to stay a hard error (one-time, frozen list), or move to warn/suppress so the list is safe to grow? Happy to implement whichever you prefer.

c2thorn:
> Normally for fields like this, we'd do something like:
> 
> * Remove server-managed values in the flattener for this field. (You might be able to remove the diff suppress at that point.) Normally the client-managed field would be called `desired_metadata` but that's not required.
> * If having access to the server-managed values is important, add a separate field that is output-only and contains all metadata values.
> 
> Would that work in this case?

@bcreddy-gcp I back @melinath here, and recommend this solution over any of the options I see in your subsequent  comments (they are hard to parse really)

melinath (commented):
It sounds like a key can move from "protected" to "settable" status, and this is a fairly common occurrence. In that case, we probably shouldn't add validation to Terraform, since we want to be forwards-compatible. Any validation we add could restrict users from taking valid actions in the future without upgrading (potentially forcing upgrade across major release boundaries.)

Does the API actually prevent sending these fields? According to the issue, it sounds like they're ignored by the API. Do we need to flag them on the client side at all? I suppose we could want to make sure people know the API will ignore the value; in that case, a warning could be sufficient (but the fact it'll be constantly outdated makes it less useful.)

Separating the server-managed fields into a separate output-only field could make sense, and it might be non-breaking enough to divert new values there later, but that's more a fix for the permadiff than preventing people setting these fields (and it'll also be constantly outdated, apparently?)

The shifting allowed values just makes this all hard to manage, basically. Really, if people shouldn't be sending certain values to the API, it might be best to make the API return errors when they're sent. Alternatively / additionally, it would be great if the API put server-managed values in a separate field, or at least use a specific prefix for server-managed values so that diff suppression / flattening / whatever logic could key off that to be future-compatible without needing constant management on the client side. For comparison, see: https://github.com/hashicorp/terraform-provider-google-beta/blob/59f3945ae09c68eb61383036eff5bb12d804d1cb/google-beta/services/dataproc/resource_dataproc_cluster.go#L147

c2thorn:
There is not enough time remaining for the 8.0.0 release, as a solution was not even decided on yet. 

This will have to be re-opened for next year's major release, or worked around in a non-breaking way.

bcreddy-gcp:
We are working on implementing a output only system provided metadata field. See https://docs.google.com/document/d/1FFJ3HXUOF5KaL2deS9AE1JZl4L7ZWkT4uTHgjz6CkzM/edit?disco=AAACFxC1h8I.    We won't be able to remove the system provided metadata from the original metadata field since that would be breaking a lot of other things. That said, the custom diff can be updated to show diff only if (metadata - system_metadata) is having a diff. 

Even otherwise, silently dropping system metadata seems like a much better experience than constantly breaking the users. Closing this PR seems to be the right thing to do.

PR #15106, opened 2025-09-10
Support the new `Broker` type of mirroring endpoint group.

duvni:
Mirroring endpoint groups support DIRECT and BROKER types.
The default type (and existing type) is DIRECT, and this PR allows users of the beta provider to create resources with the new BROKER type.

```release-note:enhancement
networksecurity: added `type` and `mirroringDeploymentGroups` fields to `google_network_security_mirroring_endpoint_group` resource
```


duvni:
@roaks3 the tests found a potential breaking change, the backend does guarantee that the default values specified is the default value returned from the API. However I do want to validate this better so I'll close this PR and create a new one afterwards.

PR #15107, opened 2025-09-10
fix(google_sql_database_instance): Connection Pool requires Enterprise Plus

AdamSkarboJonsson:
I got burnt when trying to only add a connection pool to a non-Enterprise Plus instance: 
```
googleapi: Error 400: Invalid request: Connection pool is only supported for Enterprise Plus edition.., invalid
```

```release-note:note
Fixed the `google_sql_database_instance` examples to include `edition = "ENTERPRISE_PLUS"` when using `connection_pool_config`, since connection pooling requires Enterprise Plus.
```


shuyama1 (approved):
Interesting - Our nightly tests pass consistently without specifying `edition = "ENTERPRISE_PLUS"`, I checked the API responses for those runs and they do return the `ENTERPRISE_PLUS` value though. My assumption is this might be due to a default setting in our test project (?). However, explicitly setting the value in the example configuration seems more correct. Thanks for adding it.

PR #15113, opened 2025-09-11
Skip the acceptance test for Dialogflow CMEK

zhihaos:
```release-note:none
```

### DIFF
--- a/mmv1/products/dialogflow/EncryptionSpec.yaml
+++ b/mmv1/products/dialogflow/EncryptionSpec.yaml
@@ -61,3 +61,21 @@ properties:
           If empty, the resource is secured by the default Google encryption key.
           Only the key in the same location as this resource is allowed to be used for encryption.
           Format: projects/{project}/locations/{location}/keyRings/{keyRing}/cryptoKeys/{key}
+
+examples:
+  - name: "dialogflow_encryption_spec_basic"
+    primary_resource_id: "my-encryption-spec"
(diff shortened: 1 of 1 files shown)


zhihaos:
> we have seen an increase in the number of CMEK enabled DF projects to be created, This has resulted in an increasing number of CMEK DBs in us-central1. Although, the project is deleted right away. currently there is no automation/manual process which is deleting this CMEK DBs with deleted project, increasing in our group count quota for Spanner dbs. we need to turndown cmek acceptance test until we have way to handle db deletion for deleted cmek projects

@hao-nan-li do you have any suggestions? or is there other way to disable the tests?

mmv1/products/dialogflow/EncryptionSpec.yaml
```diff
@@ -61,18 +61,3 @@ properties:
           If empty, the resource is secured by the default Google encryption key.
           Only the key in the same location as this resource is allowed to be used for encryption.
           Format: projects/{project}/locations/{location}/keyRings/{keyRing}/cryptoKeys/{key}
-
-examples:
-  - name: "dialogflow_encryption_spec_basic"
```
hao-nan-li: I think you could add `exclude_test: true` instead of removing it.
Please add a comment on why do we skip this test

zhihaos: Thanks! Done.

PR #15124, opened 2025-09-11
Adds RolloutKind provider for SaasRuntime

g-dreva:
```release-note:new-resource
`google_saas_runtime_rollout_kind`
```

### DIFF
--- a/mmv1/products/saasservicemgmt/RolloutKind.yaml
+++ b/mmv1/products/saasservicemgmt/RolloutKind.yaml
@@ -1,4 +1,4 @@
-# Copyright 2024 Google Inc.
+# Copyright 2025 Google Inc.
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
 # You may obtain a copy of the License at


mmv1/products/saasservicemgmt/RolloutKind.yaml
```diff
@@ -0,0 +1,153 @@
+# Copyright 2024 Google Inc.
```
rileykarson:
```suggestion
# Copyright 2025 Google Inc.
```

mmv1/products/saasservicemgmt/RolloutKind.yaml
```diff
@@ -0,0 +1,153 @@
...
+        type: Integer
+        description: |-
+          The maximum number of failed units allowed in a location without pausing
+          the rollout.
+      - name: allowedPercentage
+        type: Integer
+        description: |-
+          The maximum percentage of units allowed to fail (0, 100] within a location
+          without pausing the rollout.
+  - name: etag
```
rileykarson: You can remove this, probably! It should probably be `type: Fingerprint` if not.

mmv1/products/saasservicemgmt/RolloutKind.yaml
```diff
@@ -0,0 +1,153 @@
...
+      scheme:
+
+      "projects/{project}/locations/{location}/rolloutKinds/{rollout_kind_id}"
+    output: true
+  - name: rolloutOrchestrationStrategy
+    type: String
+    description: |-
+      The strategy used for executing a Rollout. This is a required field.
+
+      There are two supported values strategies which are used to control
```
rileykarson: Seems to have some missing text.

mmv1/products/saasservicemgmt/RolloutKind.yaml
```diff
@@ -0,0 +1,153 @@
...
+  - name: updateUnitKindStrategy
+    type: Enum
+    description: |-
+      The config for updating the unit kind. By default, the unit kind will be
+      updated on the rollout start.
+      Possible values:
+      UPDATE_UNIT_KIND_STRATEGY_ON_START
+      UPDATE_UNIT_KIND_STRATEGY_NEVER
+    enum_values:
+      - "UPDATE_UNIT_KIND_STRATEGY_UNSPECIFIED"
```
rileykarson:
```suggestion
```

APIs turn this into null (not returned) on response, no need to allow sending it.

mmv1/products/saasservicemgmt/RolloutKind.yaml
```diff
@@ -0,0 +1,153 @@
...
+    type: String
+    description: |-
+      The timestamp when the resource was last updated. Any
+      change to the resource made by users must refresh this value.
+      Changes to a resource made by the service should refresh this value.
+    output: true
+  - name: updateUnitKindStrategy
+    type: Enum
+    description: |-
+      The config for updating the unit kind. By default, the unit kind will be
```
rileykarson, line 138: "By default..."

What does the API return when nothing (null) is sent?

mmv1/templates/terraform/examples/saas_runtime_rollout_kind_basic.tf.tmpl
```diff
@@ -0,0 +1,30 @@
...
+  provider          = google-beta
+  location          = "global"
+  rollout_kind_id   = "{{index $.Vars "rolloutkind_name"}}"
+  unit_kind         = google_saas_runtime_unit_kind.example_unitkind.id
+
+  rollout_orchestration_strategy = "Google.Cloud.Simple.OneLocationAtATime"
+  error_budget {
+    allowed_count = 1
+  }
+  update_unit_kind_strategy = "UPDATE_UNIT_KIND_STRATEGY_ON_START"
```
rileykarson:
```suggestion
```

We cover both values below, so let's use unset (null) here

mmv1/products/saasservicemgmt/RolloutKind.yaml
g-dreva: The API will not return any value either, but when parsed it will be the default value, which is "UPDATE_UNIT_KIND_STRATEGY_UNSPECIFIED". 

What the description is saying is that the behavior of unspecified/unset is the same as "UPDATE_UNIT_KIND_STRATEGY_ON_START".

rileykarson: Thanks- so the effective return is null.

PR #7202, opened 2023-02-02
Add support for Apigee Sharedflow

5fff:
If this PR is for Terraform, I acknowledge that I have:

- [ ] Searched through the [issue tracker](https://github.com/hashicorp/terraform-provider-google/issues) for an open issue that this either resolves or contributes to, commented on it to claim it, and written "fixes {url}" or "part of {url}" in this PR description. If there were no relevant open issues, I opened one and commented that I would like to work on it (not necessary for very small changes).
- [X] [Generated Terraform](https://github.com/GoogleCloudPlatform/magic-modules#generating-the-terraform-providers), and ran `make test` and `make lint` to ensure it passes unit and linter tests.
- [X] Ensured that all new fields I added that can be set by a user appear in at least one [example](https://github.com/GoogleCloudPlatform/magic-modules/tree/main/mmv1/templates/terraform/examples) (for generated resources) or [third_party test](https://github.com/GoogleCloudPlatform/magic-modules/tree/main/mmv1/third_party/terraform/tests) (for handwritten resources or update tests).
- [X] [Ran](https://github.com/hashicorp/terraform-provider-google/blob/main/.github/CONTRIBUTING.md#tests) relevant acceptance tests (If the acceptance tests do not yet pass or you are unable to run them, please let your reviewer know).
- [x] Read the [Release Notes Guide](https://github.com/GoogleCloudPlatform/magic-modules/blob/main/.ci/RELEASE_NOTES_GUIDE.md) before writing my release note below.

**Release Note Template for Downstream PRs (will be copied)**

```release-note:enhancement
apigee: Added support for resource apigee sharedflow
https://cloud.google.com/apigee/docs/reference/apis/apigee/rest/v1/organizations.sharedflows

Using handwritten approach as create API requires a file to be uploaded as `octet-stream` and mmv1 does not support this pattern

Using direct api call instead of golang library because golang library for this api is broken see: b/265084250

Other relevant background/bugs:
Apigee Sharedflow get method response missing `latestRevisionId` field b/266146203
(recently fixed) Apigee SharedFlow Create Documentation issue b/256739387
```

PR #15126, opened 2025-09-11
add resources to support apigee proxy deployments

jcaldwell-lb:
Fixes https://github.com/hashicorp/terraform-provider-google/issues/24218

Provides a resource called `google_apigee_api_deployment` that manages Apigee proxy deployments, similar to the existing resource `google_apigee_sharedflow_deployment`.

```release-note:new-resource
`google_apigee_api_deployment`
```

### DIFF
--- a/mmv1/products/apigee/ApiDeployment.yaml
+++ b/mmv1/products/apigee/ApiDeployment.yaml
@@ -86,4 +86,3 @@ examples:
     external_providers: ['time']
     skip_vcr: true
     exclude_docs: true
-


mmv1/third_party/terraform/services/apigee/resource_apigee_api_deployment_test.go
```diff
@@ -0,0 +1,182 @@
...
+  org_id   = google_apigee_organization.apigee_org.id
+  name         = "tf-test%{random_suffix}"
+  description  = "Apigee Environment"
+  display_name = "environment-1"
+}
+
+resource "google_apigee_api" "test_apigee_api" {
+  name            = "tf-test-apigee-api"
+  org_id          = google_project.project.project_id
+  config_bundle   = "%{config_bundle}"
```
c2thorn: By changing the config_bundle here, are we expecting the api_deployment resource to force recreate?

jcaldwell-lb: We are - when I ran this acceptance test I saw that both a revision 1 and 2 were deployed. This is akin to the shared flow test.

jcaldwell-lb: This resource is handwritten because the related `google_apigee_sharedflow_deployment` is also handwritten, and the use case is very similar. I can look into the MMv1 format if you'd prefer that.

c2thorn: Thanks for sharing the similar resource. Looking at its creation PR
https://github.com/GoogleCloudPlatform/magic-modules/pull/7202, I see
```
Using handwritten approach for sharedflow as create API requires a file to be uploaded as `octet-stream` and mmv1 does not support this pattern
```
It looks like Sharedflow needed be handwritten, but I'm very doubtful SharedflowDeployment needed to be. It does have a [non-standard create URL](https://github.com/GoogleCloudPlatform/magic-modules/pull/7202/files#diff-501161be421e75a6a475e3665110d31ca784d85e8accf11ab1e8057960d48879R70) but it seems google_apigee_api_deployment does not.

Handwritten resources are considered only to be created when absolutely necessary and make it harder for maintainability over time (divergence from large sweeping changes to the engine, harder for our metric tracking). I appreciate that this resource is already complete, but I have to ask you to please recreate it following the standard "MMv1" method here: https://googlecloudplatform.github.io/magic-modules/develop/add-resource/. This resource is fairly small, so hopefully it should be quick.

This will reduce the need for the handwritten documentation file as well. You could keep the test file you already have, but there would need to be at least one [create test example](https://googlecloudplatform.github.io/magic-modules/test/test/#add-a-create-test) for the generated documentation.

jcaldwell-lb: It took me some time to figure out but I believe I have constructed a new resource using the MMv1 templating format. It looks like the generated test file passes as well. Please take a look and let me know if there's anything else you'll need!

jcaldwell-lb:
> kicking off the build

Does it look successful from your end?

jcaldwell-lb:
I see there is a failing check, that seems to be a test. I cannot see the reason behind the issue, do I need to make a change?

zli82016:
> I see there is a failing check, that seems to be a test. I cannot see the reason behind the issue, do I need to make a change?

Feel free to ignore the failed google conversion integration tests. Sorry for the confusion.

c2thorn (approved):

PR #15123, opened 2025-09-11
Fix network field in google_compute_region_backend_service to force new on change

andrecardinalli-cit:
Fixes hashicorp/terraform-provider-google#23039

## Description
The network field of google_compute_region_backend_service cannot be modified after creation according to the GCP API. This change marks the field as immutable so Terraform will know to recreate the resource instead of trying to update it in place. This resolves the "Network field cannot be modified" error that occurs when users try to add a network to an existing resource.

## Affected resources
- google_compute_region_backend_service

## Self Review Checklist
- [x] The PR title follows the [contributing guide](https://googlecloudplatform.github.io/magic-modules/contribute/)
- [x] Issue being fixed is linked: https://github.com/hashicorp/terraform-provider-google/issues/23039
- [ ] Tests are added/updated to cover the changes (not needed for this schema-only change)
- [x] Documentation is updated to cover the changes
- [x] Changelog entry added if necessary

```release-note:bug
google_compute_region_backend_service: The network field in 'google_compute_region_backend_service' is now correctly handled, forcing a new resource on change. This resolves the "Network field cannot be modified" error when trying to add a network to an existing resource
```

PR #15136, opened 2025-09-13
Fix network field in google_compute_region_backend_service to force new on change

andrecardinalli-cit:
Fixes hashicorp/terraform-provider-google#23039

## Description
The network field of google_compute_region_backend_service cannot be modified after creation according to the GCP API. This change marks the field as immutable so Terraform will know to recreate the resource instead of trying to update it in place. This resolves the "Network field cannot be modified" error that occurs when users try to add a network to an existing resource.

## Affected resources
- google_compute_region_backend_service

This replaces PR #15123 with a clean branch containing only the necessary change.

```release-note:bug
compute: The network field in 'google_compute_region_backend_service' is now correctly handled, forcing a new resource on change. This resolves the "Network field cannot be modified" error when trying to add, remove, or modify a network on an existing resource
```

### DIFF
--- a/mmv1/products/compute/RegionBackendService.yaml
+++ b/mmv1/products/compute/RegionBackendService.yaml
@@ -1439,10 +1439,13 @@ properties:
     type: ResourceRef
     description: |
       The URL of the network to which this backend service belongs.
-      This field can only be specified when the load balancing scheme is set to INTERNAL.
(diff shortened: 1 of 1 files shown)


andrecardinalli-cit:
Hey @rileykarson, could you take a look on this PR?!

Tks in advance!

rileykarson (approved):
LGTM

Doesn't seem there are any exceptions to the rule

rileykarson (approved):
Failing test not realted

PR #15140, opened 2025-09-15
Added List DsRefs API and improved some documentation for some API's

kushallunkad-google:

```release-note:new-datasource
`google_backup_dr_list_data_source_references`
```

### DIFF
--- a/mmv1/third_party/terraform/services/backupdr/data_source_backup_dr_data_source_references_list.go
+++ b/mmv1/third_party/terraform/services/backupdr/data_source_backup_dr_data_source_references_list.go
@@ -24,16 +24,6 @@ func DataSourceGoogleCloudBackupDRListDataSourceReferences() *schema.Resource {
 				Computed:    true,
 				Description: "The ID of the project in which the resource belongs.",
 			},
-			"filter": {
-				Type:        schema.TypeString,
-				Optional:    true,
-				Description: "The filter to apply to the list results.",
-			},
-			"order_by": {
-				Type:        schema.TypeString,
-				Optional:    true,
-				Description: "The order to sort results by.",
-			},
 			"data_source_references": {
 				Type:        schema.TypeList,
 				Computed:    true,
@@ -98,24 +88,6 @@ func dataSourceGoogleCloudBackupDRListDataSourceReferencesRead(d *schema.Resourc
 
 	location := d.Get("location").(string)
 	url := fmt.Sprintf("%sprojects/%s/locations/%s/dataSourceReferences", config.BackupDRBasePath, project, location)
-
-	params := make(map[string]string)
-	if v, ok := d.GetOk("filter"); ok {
-		params["filter"] = v.(string)
-	}
-	if v, ok := d.GetOk("order_by"); ok {
-		// API expects `orderBy` as the query parameter name
-		params["orderBy"] = v.(string)
-	}
-
-	// Attach query params to the URL so the API receives filter/orderBy
-	if len(params) > 0 {
-		var err error
-		url, err = transport_tpg.AddQueryParams(url, params)
-		if err != nil {
-			return fmt.Errorf("Error adding query params to URL: %s", err)
-		}
-	}
 
 	res, err := transport_tpg.SendRequest(transport_tpg.SendRequestOptions{
 		Config:    config,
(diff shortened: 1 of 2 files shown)


mmv1/third_party/terraform/services/backupdr/data_source_backup_dr_data_source_reference.go
```diff
@@ -10,6 +10,133 @@ import (
 	"google.golang.org/api/googleapi"
 )
 
+func DataSourceGoogleCloudBackupDRListDataSourceReferences() *schema.Resource {
```
shuyama1: Just to confirm my understanding - this new datasource `google_backup_dr_list_data_source_references` lists the same dataSourceReferences as `google_backup_dr_data_source_references`, except that one is targeted for a specific resource type by using the `:fetchForResourceType` endpoint?

I'm wondering if we need a separate datasource for this or we should combine it with `google_backup_dr_data_source_references`. We could make `resource_type` optional and only append `:fetchForResourceType` to the url if the field is set.

kushallunkad-google: We currently have two separate APIs, each requiring different IAM permissions. If we use the same data source name (like combining listDSRefs and FetchDSRefs), users might inadvertently call an API for which they lack authorization. Therefore, we must keep the data source names distinct to enforce access control based on IAM roles.

mmv1/third_party/terraform/services/backupdr/data_source_backup_dr_data_source_reference.go
```diff
@@ -10,6 +10,133 @@ import (
...
+			},
+			"filter": {
+				Type:        schema.TypeString,
+				Optional:    true,
+				Description: "The filter to apply to the list results.",
+			},
+			"order_by": {
+				Type:        schema.TypeString,
+				Optional:    true,
+				Description: "The order to sort results by.",
```
shuyama1: Can you add a test coverage on these two parameters? Thanks

mmv1/third_party/terraform/services/backupdr/data_source_backup_dr_data_source_reference.go
shuyama1: Okay, understand, then let's keep them as separate datasources. Can you move this to a separate file please? We'd want one resource/datasource per file.

Plus, I'm checking with my team on the naming of this datasource.

mmv1/third_party/terraform/services/backupdr/data_source_backup_dr_data_source_reference_test.go
```diff
@@ -9,6 +9,121 @@ import (
 	"github.com/hashicorp/terraform-provider-google/google/acctest"
 )
 
+func TestAccDataSourceGoogleBackupDRListDataSourceReferences_basic(t *testing.T) {
```
shuyama1: Also, here. Can you move them to a separate test file please? Thanks!

mmv1/third_party/terraform/services/backupdr/data_source_backup_dr_data_source_reference.go
```diff
@@ -10,6 +10,133 @@ import (
...
+	flattenedDataSourceReferences, err := flattenDataSourceReferences(items)
+	if err != nil {
+		return err
+	}
+
+	if err := d.Set("data_source_references", flattenedDataSourceReferences); err != nil {
+		return fmt.Errorf("Error setting data_source_references: %s", err)
+	}
+
+	d.SetId(url)
```
shuyama1: I'd suggest the id to be `projects/%s/locations/%s/dataSourceReferences` without the basepath part. Maybe also include the filter like `projects/%s/locations/%s/dataSourceReferences/filter/%`

mmv1/third_party/terraform/services/backupdr/data_source_backup_dr_data_source_reference.go
kushallunkad-google: Done

mmv1/third_party/terraform/services/backupdr/data_source_backup_dr_data_source_reference.go
kushallunkad-google: Done

mmv1/third_party/terraform/services/backupdr/data_source_backup_dr_data_source_reference_test.go
kushallunkad-google: Done

mmv1/third_party/terraform/services/backupdr/data_source_backup_dr_data_source_reference.go
kushallunkad-google: Done

PR #15141, opened 2025-09-15
fix(apihub): custom decoder to resolve permadiff

ramonvermeulen:
Closes https://github.com/hashicorp/terraform-provider-google/issues/24355

```release-note:bug
apihub: fix permadiff on `supported_auth_types`
```

### DIFF
--- a/mmv1/products/apihub/Plugin.yaml
+++ b/mmv1/products/apihub/Plugin.yaml
@@ -31,6 +31,8 @@ examples:
     external_providers: ["time"]
 autogen_async: true
 autogen_status: UGx1Z2lu
+custom_code:
+  decoder: 'templates/terraform/decoders/apihub_plugin_auth_config_template.go.tmpl'
 parameters:
   - name: location
     type: String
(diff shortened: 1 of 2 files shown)


mmv1/products/apihub/Plugin.yaml
```diff
@@ -130,6 +130,7 @@ properties:
   - name: configTemplate
     type: NestedObject
     description: ConfigTemplate represents the configuration template for a plugin.
+    custom_flatten: 'templates/terraform/custom_flatten/apihub_plugin_config_template.go.tmpl'
```
shuyama1: Will adding `default_from_api: true` to `configTemplate` resolve the issue? https://googlecloudplatform.github.io/magic-modules/develop/diffs/#fix-diffs

ramonvermeulen: Check, I initially tried that on the `supported_auth_types` field itself but that didn't work. But just tested it and if you put it on the whole configTemplate that actually works fine.

shuyama1 (approved):

PR #15148, opened 2025-09-16
Add the Resourcemanager Capability Service to the google provider.

blacken57:
Adds the Capability API in ga version of terraform.

Fixes b/445323472

```release-note:enhancement
resourcemanager: promoted `google_resource_manager_capability` to ga
```

### DIFF
--- a/mmv1/templates/terraform/examples/resource_manager_capability.tf.tmpl
+++ b/mmv1/templates/terraform/examples/resource_manager_capability.tf.tmpl
@@ -1,5 +1,4 @@
 resource "google_folder" "folder" {
-  provider         = google-beta
   display_name     = "{{index $.Vars "display_name"}}"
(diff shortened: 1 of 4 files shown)


mmv1/templates/terraform/examples/resource_manager_capability.tf.tmpl
```diff
@@ -1,5 +1,5 @@
 resource "google_folder" "folder" {
-  provider         = google-beta
+  provider         = google
```
hao-nan-li: Perhaps this line should be removed from the test so that the tests could run on both beta and GA.

Same for all the following lines with `provider = google`

blacken57: Makes sense! Changed.

blacken57:
@hao-nan-li  would appreciate a review. Implemented your suggestion already.

PR #15160, opened 2025-09-17
added default_from_api: true for squashMode, hasRootAccess & accessType

raazanand:
```release-note: bug
netapp: fixed incorrect default value handling in `google_netapp_volume` for `export_policy.rules` attributes `has_root_access` and `squash_mode`. When not specified, these fields will now take on the API default value with no diff.
```


rileykarson (approved):
LGTM

PR #15162, opened 2025-09-17
feat: (storageinsights) added `force_destroy` to `google_storage_insights_report_config` resource

gurusai-voleti:

```release-note:enhancement
storageinsights: added `force_destroy` to `google_storage_insights_report_config` resource
```

### DIFF
--- a/mmv1/products/storageinsights/ReportConfig.yaml
+++ b/mmv1/products/storageinsights/ReportConfig.yaml
@@ -23,6 +23,7 @@ docs:
 base_url: 'projects/{{project}}/locations/{{location}}/reportConfigs'
 self_link: 'projects/{{project}}/locations/{{location}}/reportConfigs/{{name}}'
 create_url: 'projects/{{project}}/locations/{{location}}/reportConfigs'
+delete_url: projects/{{project}}/locations/{{location}}/reportConfigs/{{name}}?force={{force_destroy}}
 update_verb: 'PATCH'
 update_mask: true
 timeouts:
@@ -35,6 +36,8 @@ examples:
     primary_resource_id: 'config'
     vars:
       bucket_name: 'my-bucket'
+    ignore_read_extra:
+      - 'force_destroy'
 parameters:
   - name: 'location'
     type: String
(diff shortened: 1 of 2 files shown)


mmv1/products/storageinsights/ReportConfig.yaml
```diff
@@ -23,6 +23,7 @@ docs:
 base_url: 'projects/{{project}}/locations/{{location}}/reportConfigs'
 self_link: 'projects/{{project}}/locations/{{location}}/reportConfigs/{{name}}'
 create_url: 'projects/{{project}}/locations/{{location}}/reportConfigs'
+delete_url: projects/{{project}}/locations/{{location}}/reportConfigs/{{name}}?force={{force_destroy}}
```
hao-nan-li, line 26: Is this only applicable if force_destroy is set to true?

mmv1/third_party/terraform/services/storageinsights/resource_storage_insights_report_config_test.go
```diff
@@ -167,6 +167,7 @@ resource "google_storage_insights_report_config" "config" {
   depends_on = [
 	google_storage_bucket_iam_member.admin,
   ]
+  force_destroy = true
```
sameer-google, line 170: Do we have tests for the default case, force_destroy = False and destroy the resource

sameer-google (commented):
Overall the change LGTM, +1 for https://github.com/GoogleCloudPlatform/magic-modules/pull/15162#discussion_r2369389169, was delete verified for default value -> False.

mmv1/products/storageinsights/ReportConfig.yaml
gurusai-voleti: default is false, if not set by user in tf config force_destroy will be set to false in URL when calling DELETE  API

mmv1/third_party/terraform/services/storageinsights/resource_storage_insights_report_config_test.go
gurusai-voleti: default is false so if not set to true in tf config by user, force_destroy goes as false in API call

PR #15163, opened 2025-09-17
datastream: add support for `mongodb_profile` in `google_datastream_connection_profile`

ramonvermeulen:
Partly closes https://github.com/hashicorp/terraform-provider-google/issues/24391

**Release Note Template for Downstream PRs (will be copied)**

See [Write release notes](https://googlecloudplatform.github.io/magic-modules/contribute/release-notes/) for guidance.

```release-note:enhancement
datastream: added support for `mongodb_profile` in `google_datastream_connection_profile`
```

PR #15164, opened 2025-09-17
datastream: add support for `mongodb_source_config` in `google_datastream_stream`

ramonvermeulen:
Partly closes https://github.com/hashicorp/terraform-provider-google/issues/24391

```release-note:enhancement
datastream: add support for `mongodb_source_config` in `google_datastream_stream`
```

### DIFF
--- a/mmv1/products/datastream/Stream.yaml
+++ b/mmv1/products/datastream/Stream.yaml
@@ -240,6 +240,13 @@ examples:
       source_connection_profile_id: 'source-profile'
       destination_connection_profile_id: 'destination-profile'
     exclude_test: true
+  - name: 'datastream_stream_mongodb'
+    primary_resource_id: 'default'
+    vars:
+      stream_id: 'mdb-stream'
+      source_connection_profile_id: 'source-profile'
+      destination_connection_profile_id: 'destination-profile'
+    exclude_test: true
 virtual_fields:
   - name: 'desired_state'
     description: |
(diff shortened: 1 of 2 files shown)


ramonvermeulen:
Upgrading to ready for review, looking for suggestions on how to write acceptance tests for this. Because essentially you would like to test this by provisioning a MongoDB instance and using that credentials in the test (via the resource from the other PR `google_datastream_connection_profile`). However AFAIK there is no native MongoDB on GCP (only via marketplace in collab with Atlas, but probably not that suited for an acceptance test?), maybe [firestore with mongodb compatibility](https://cloud.google.com/products/firestore/mongodb-compatibility?hl=en) is an option?

BBBmau (commented):
we have fields that are missing from the test report, can you add these in? The PR itself looks just fine

ramonvermeulen:
> we have fields that are missing from the test report, can you add these in? The PR itself looks just fine

@BBBmau See comment at https://github.com/GoogleCloudPlatform/magic-modules/pull/15164#issuecomment-3312720376, looking for ways to tests these fields properly. I think this is quite hard since it requires a MongoDB instance.

ramonvermeulen:
> to answer your question regarding testing this, it seems that in order to fully support this we first must work on adding `mongoDB` support in `datastream_connection_profile`
> 
> Once that is supported we can move to provisioning a `connection_profile` of type `MongoDB` as part of the acceptance test.
> 
> creating the connection profile for MongoDB is supported in the API [here](https://cloud.google.com/datastream/docs/create-connection-profiles#cp4mongodb)
> 
> I see that you've opened a PR to include this also:
> 
> * #15163
> 
> let's wait until that gets merged to add the `mongodb_source_config` tests

Thanks for the update! I made quite some steps in testing `google_datastream_connection_profile` locally via a free Atlas MongoDB instance. Also I figured out it has a `create_without_validation` argument which can be used to acceptance test some parts of it (you can create the connection_profile without actually "testing" the real connectivity to MongoDB).

For now I reverted both PRs back to draft, I will first finalizehttps://github.com/GoogleCloudPlatform/magic-modules/pull/15163 and after that come back to this PR.

PR #15166, opened 2025-09-17
Terraform Integration for TagBindingCollection

amukul-25:
Add support for TagBindingCollection in terraform.
TagBindingCollection is a singleton resource which is part of Resource Manager Tags product.

```release-note:new-resource
google_tags_tag_binding_collection
```

### DIFF
--- a/mmv1/third_party/terraform/services/tags/resource_tags_tag_binding_collection.go.tmpl
+++ b/mmv1/third_party/terraform/services/tags/resource_tags_tag_binding_collection.go.tmpl
@@ -1,6 +1,6 @@
-{{- if ne $.TargetVersionName "ga" }}
-
 package tags
+
+{{- if ne $.TargetVersionName "ga" }}
 
 import (
 	"context"
@@ -11,12 +11,11 @@ import (
 	"strings"
 	"time"
 
-	dcl "github.com/GoogleCloudPlatform/declarative-resource-client-library/dcl"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 
-	"github.com/hashicorp/terraform-provider-google-beta/google-beta/tpgresource"
-	transport_tpg "github.com/hashicorp/terraform-provider-google-beta/google-beta/transport"
+	"github.com/hashicorp/terraform-provider-google/google/tpgresource"
+	transport_tpg "github.com/hashicorp/terraform-provider-google/google/transport"
 )
 
 func ResourceTagsTagBindingCollection() *schema.Resource {
@@ -155,6 +154,7 @@ func resourceTagsTagBindingCollectionUpsert(ctx context.Context, d *schema.Resou
 		mergedTags[k] = v
 	}
 	for k, v := range configuredTags {
+		// This will also overwrite the current bindings with the bindings in terraform config, nullifying any updates done outside of terraform config
 		mergedTags[k] = v
 	}
 
@@ -203,12 +203,7 @@ func resourceTagsTagBindingCollectionDelete(ctx context.Context, d *schema.Resou
 	// Step 1: Read current state
 	currentTags, currentEtag, err := getCurrentTagBindingCollectionState(config, userAgent, location, encodedResourceName)
 	if err != nil {
-		if dcl.IsNotFound(err) {
-			log.Printf("[DEBUG] TagBindingCollection not found for %s during delete, nothing to do.", resourceName)
-			d.SetId("")
-			return nil
-		}
-		return diag.Errorf("Error reading current state for delete: %s", err)
+		return diag.FromErr(transport_tpg.HandleNotFoundError(err, d, fmt.Sprintf("TagBindingCollection %q", resourceName)))
 	}
 
 	// Step 2. Get Terraform-managed old tags from state
@@ -229,6 +224,7 @@ func resourceTagsTagBindingCollectionDelete(ctx context.Context, d *schema.Resou
 	for k := range oldTags {
 		// Check if the tag managed by Terraform is actually present on the resource.
 		if _, existsInCurrent := currentTags[k]; existsInCurrent {
+			// This will delete all the bindings managed by terraform, even if such bindings were updated outside of terraform as well
 			delete(finalTags, k)
 			hasTagsToDelete = true
 		}
--- a/mmv1/third_party/terraform/services/tags/resource_tags_tag_binding_collection_test.go.tmpl
+++ b/mmv1/third_party/terraform/services/tags/resource_tags_tag_binding_collection_test.go.tmpl
@@ -1,6 +1,6 @@
-{{- if ne $.TargetVersionName "ga" }}
-
 package tags_test
+
+{{- if ne $.TargetVersionName "ga" }}
 
 import (
 	"context"
@@ -10,15 +10,14 @@ import (
 	"regexp"
 	"testing"
 
-	dcl "github.com/GoogleCloudPlatform/declarative-resource-client-library/dcl"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 	"github.com/hashicorp/terraform-plugin-testing/helper/resource"
 	"github.com/hashicorp/terraform-plugin-testing/terraform"
 
-	"github.com/hashicorp/terraform-provider-google-beta/google-beta/acctest"
-	"github.com/hashicorp/terraform-provider-google-beta/google-beta/envvar"
-	transport_tpg "github.com/hashicorp/terraform-provider-google-beta/google-beta/transport"
-	"github.com/hashicorp/terraform-provider-google-beta/google-beta/services/tags"
+	"github.com/hashicorp/terraform-provider-google/google/acctest"
+	"github.com/hashicorp/terraform-provider-google/google/envvar"
+	transport_tpg "github.com/hashicorp/terraform-provider-google/google/transport"
+	"github.com/hashicorp/terraform-provider-google/google/services/tags"
 )
 
 // Acceptance Tests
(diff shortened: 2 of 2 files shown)


mmv1/third_party/terraform/services/tags/resource_tags_tag_binding_collection.go.tmpl
```diff
@@ -0,0 +1,421 @@
...
+		obj["etag"] = currentEtag
+	}
+
+	configuredTags, diags := getStringMapFromSchema(d, "tags")
+	if diags != nil {
+		return diags
+	}
+
+	mergedTags := make(map[string]interface{})
+	for k, v := range currentTags {
```
abd-goog, line 153: Lets add a test case where a tag k1,v1 was applied via 'tags' field but then later a tag k1,v2 was applied outside of terraform. In this case TF config should override the k1,v2 tag

mmv1/third_party/terraform/services/tags/resource_tags_tag_binding_collection.go.tmpl
```diff
@@ -0,0 +1,421 @@
...
+	if err != nil {
+		return diag.FromErr(err)
+	}
+
+	location, resourceName, encodedResourceName := getTagBindingCollectionInfo(d)
+
+	// Step 1: Read current state
+	currentTags, currentEtag, err := getCurrentTagBindingCollectionState(config, userAgent, location, encodedResourceName)
+	if err != nil {
+		if dcl.IsNotFound(err) {
```
rileykarson: Drive-by comment: I think we want `transport_tpg.HandleNotFoundError` for these cases!

rileykarson:
@modular-magician reassign-reviewer slevenick

mmv1/third_party/terraform/services/tags/resource_tags_tag_binding_collection.go.tmpl
```diff
@@ -0,0 +1,421 @@
...
+import (
+	"context"
+	"fmt"
+	"log"
+	"net/http"
+	"net/url"
+	"strings"
+	"time"
+
+	dcl "github.com/GoogleCloudPlatform/declarative-resource-client-library/dcl"
```
slevenick: We'll want to remove this dependency and any uses of it throughout this PR. There should be an alternative 404 check method available

mmv1/third_party/terraform/services/tags/resource_tags_tag_binding_collection.go.tmpl
```diff
@@ -0,0 +1,421 @@
...
+		Schema: map[string]*schema.Schema{
+			"location": {
+				Type:        schema.TypeString,
+				Required:    true,
+				Description: `The location of the TagBindingCollection.`,
+			},
+			"full_resource_name": {
+				Type:        schema.TypeString,
+				Required:    true,
+				Description: `The full resource name of the resource to which the tags are bound. E.g. //cloudresourcemanager.googleapis.com/projects/123`,
```
slevenick, line 47: Can we validate the format of this at all? Because it's such a unique format that we don't use in other resources I'd expect users to make mistakes here. If we can validate that the value is `//[a-z]\.googleapis\.com.*` if that makes sense I think this could be helpful.

It may be impossible to do though if this is going to support non-googleapis.com resources

mmv1/third_party/terraform/services/tags/resource_tags_tag_binding_collection_test.go.tmpl
```diff
@@ -0,0 +1,582 @@
...
+
+import (
+	"context"
+	"fmt"
+	"net/http"
+	"net/url"
+	"regexp"
+	"testing"
+
+	dcl "github.com/GoogleCloudPlatform/declarative-resource-client-library/dcl"
```
slevenick: Remove this and all uses of it

mmv1/third_party/terraform/services/tags/resource_tags_tag_binding_collection.go.tmpl
```diff
@@ -0,0 +1,421 @@
...
+		},
+
+		Timeouts: &schema.ResourceTimeout{
+			Create: schema.DefaultTimeout(20 * time.Minute),
+			Update: schema.DefaultTimeout(20 * time.Minute),
+			Delete: schema.DefaultTimeout(20 * time.Minute),
+		},
+
+		Schema: map[string]*schema.Schema{
+			"location": {
```
slevenick, line 39: Why do we need location? Is it not included in the full_resource_name?

slevenick: And is it possible to add tag binding collections to resources without locations, such as projects?

amukul-25: No not for all resource types, location has to be specified in the API as well, see https://cloud.google.com/resource-manager/reference/rest/v3/locations.tagBindingCollections. 

Additionally, for global resources like projects, the location is to be specified as `global`.

mmv1/third_party/terraform/services/tags/resource_tags_tag_binding_collection.go.tmpl
amukul-25: It is actually same as the other tag binding resources in Terraform - https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/tags_tag_binding#argument-reference

Also, full resource name is validated in the API, my understanding is the tf plan will fail when the read request goes to the Get API.

mmv1/third_party/terraform/services/tags/resource_tags_tag_binding_collection.go.tmpl
amukul-25: Yes, I've made the change to use `transport_tpg.HandleNotFoundError` here.

mmv1/third_party/terraform/services/tags/resource_tags_tag_binding_collection.go.tmpl
amukul-25: Removed from both main and test files.

mmv1/third_party/terraform/services/tags/resource_tags_tag_binding_collection_test.go.tmpl
amukul-25: Done

mmv1/third_party/terraform/services/tags/resource_tags_tag_binding_collection.go.tmpl
slevenick: Interesting! Ok that works to keep this required

mmv1/third_party/terraform/services/tags/resource_tags_tag_binding_collection.go.tmpl
slevenick: Was this resolved?

mmv1/third_party/terraform/services/tags/resource_tags_tag_binding_collection.go.tmpl
```diff
@@ -0,0 +1,417 @@
...
+	if err != nil {
+		return diag.FromErr(err)
+	}
+
+	log.Printf("[INFO] Successfully completed PATCH request for %s", operationDesc)
+	return nil
+}
+
+// Helper to get a map[string]string from schema.ResourceData
+func getStringMapFromSchema(d *schema.ResourceData, key string) (map[string]string, diag.Diagnostics) {
```
slevenick, line 382: This is only used in one place, can we just inline it?

mmv1/third_party/terraform/website/docs/r/google_tags_tag_binding_collection.html.markdown
```diff
@@ -0,0 +1,127 @@
...
+- `create` - Default is 20 minutes.
+- `update` - Default is 20 minutes.
+- `delete` - Default is 20 minutes.
+
+## Import
+
+
+TagBindingCollection can be imported using any of these accepted formats:
+
+* `locations/{{location}}/tagBindingCollections/{{encoded_full_resource_name}}`
```
slevenick, line 112: How possible will it be for a user to derive this import format?

Because we're URL encoding it, they're going to have to URL encode the string to pass it to the import statement. It should be possible to accept the full_resource_name in the import and then URL encode it there so a user doesn't have to try to handle that themselves.

mmv1/third_party/terraform/services/tags/resource_tags_tag_binding_collection_test.go.tmpl
```diff
@@ -0,0 +1,581 @@
...
+				t.Errorf("Expected location %q, got %q", tc.ExpectedLoc, got)
+			}
+			if got := data.Get("full_resource_name").(string); got != tc.ExpectedResourceName {
+				t.Errorf("Expected full_resource_name %q, got %q", tc.ExpectedResourceName, got)
+			}
+		})
+	}
+}
+
+// TODO: Add more unit tests for Upsert and Delete logic, mocking API calls.
```
slevenick, line 579: Todo?

mmv1/third_party/terraform/services/tags/resource_tags_tag_binding_collection_test.go.tmpl
```diff
@@ -0,0 +1,581 @@
...
+}
+resource "google_tags_tag_value" "new_value" {
+  provider = google-beta 
+  parent      = google_tags_tag_key.new_key.id
+  short_name  = "%{new_value_short_name}"
+}
+
+resource "google_tags_tag_binding_collection" "basic" {
+  provider = google-beta 
+  full_resource_name    = "//cloudresourcemanager.googleapis.com/projects/${google_project.project.number}"
```
slevenick, line 267: Can we add a test that uses a resource other than projects? Just to make sure other resource types work as well

mmv1/third_party/terraform/services/tags/resource_tags_tag_binding_collection.go.tmpl
amukul-25: Not yet, I had made some more changes to address this, will send in next commit.

PR #15180, opened 2025-09-18
fix(cloudbuild): Fix persistent diff for git_file_source in google_cloudbuild_trigger

MattLaddy:
```release-note:bug
cloudbuild: Fixed a persistent diff that occurred when `git_file_source` was configured, as the field was not being saved to state correctly.
```


rileykarson (commented):
This code is a `post_import` and may not be what you intend (I think it's appropriate for an encoder?). Can you share more context and any bugs or issues?

MattLaddy:
@rileykarson Hi this was an attempt to fix b/357704840

PR #15185, opened 2025-09-18
Swap the order of applying default settings and

rabiamohiuddin:
Apply the default settings before checking if the edition field has been changed.

Fixes https://github.com/hashicorp/terraform-provider-google/issues/22030

```release-note:bug
sql: fixed order of applying edition updates in `google_sql_database_instance`
```


mmv1/third_party/terraform/services/sql/resource_sql_database_instance.go.tmpl
```diff
@@ -2504,6 +2508,8 @@ func resourceSqlDatabaseInstanceUpdate(d *schema.ResourceData, meta interface{})
 		}
 	}
 
+	// If there is no change detected in disk size, no need to try and update the disk size, send 0 always
+	instance.Settings.DataDiskSizeGb = 0
```
c2thorn, line 2512: this seems like a non-op change, isn't instance immediately getting overwritten with a new object in the next line?

c2thorn: did you mean to add this as an `else` to the 
```
if d.HasChange("settings.0.disk_size") {
```
block?
That would also remove the need for the addition above.

c2thorn: Also, I am not confident about eroding the disk size. Why not set it to what the current value is instead of always zero? Are you sure this is safe?

c2thorn (changes requested):

PR #15193, opened 2025-09-19
Looker gemini_enabled field added to Instance

efeelaiho:

```release-note:enhancement
looker: added `gemini_enabled` field to `google_looker_instance` resource
```


mmv1/templates/terraform/examples/looker_instance_enterprise_full.tf.tmpl
```diff
@@ -4,6 +4,7 @@ resource "google_looker_instance" "{{$.PrimaryResourceId}}" {
   region             = "us-central1"
   private_ip_enabled = true
   public_ip_enabled  = false
+  gemini_enabled     = true
```
c2thorn, line 7: This example is skipped in tests (I'm not sure/don't remember why). Can you add the field to another one that is not skipped (doesn't have `exclude_test: true` in the examples YAML block)

efeelaiho: All done!

PR #15199, opened 2025-09-22
apigee: added `deletion_protection` to `google_apigee_instance`

ramonvermeulen:
Closes https://github.com/hashicorp/terraform-provider-google/issues/24422

```release-note:enhancement
apigee: added `deletion_protection` to `google_apigee_instance`
```

There was no `deletionProtection` field available in the API, so I figured a virtual field  in combination with a `pre_delete` hook seemed like a logical solution to me. Default value is `false` for now to prevent a breaking change.

I did some manual testing:

<details>

**Config used:**
```terraform
data "google_client_config" "current" {}

resource "google_compute_network" "apigee_network" {
  name = "apigee-network"
}

resource "google_compute_global_address" "apigee_range" {
  name          = "apigee-range"
  purpose       = "VPC_PEERING"
  address_type  = "INTERNAL"
  prefix_length = 16
  network       = google_compute_network.apigee_network.id
}

resource "google_service_networking_connection" "apigee_vpc_connection" {
  network                 = google_compute_network.apigee_network.id
  service                 = "servicenetworking.googleapis.com"
  reserved_peering_ranges = [google_compute_global_address.apigee_range.name]
}

resource "google_apigee_organization" "apigee_org" {
  analytics_region   = "us-central1"
  project_id         = data.google_client_config.current.project
  authorized_network = google_compute_network.apigee_network.id
  depends_on         = [google_service_networking_connection.apigee_vpc_connection]
}

resource "google_apigee_instance" "apigee_instance" {
  name                = "my-instance-name"
  location            = "us-central1"
  org_id              = google_apigee_organization.apigee_org.id
  deletion_protection = true
}
```

(also did full apply/destroy cycle without `deletion_protection` set and that still works as intended)

Ran:

```

```
google_apigee_instance.apigee_instance: Destroying... [id=organizations/apigee-test-ramon/instances/my-instance-name]
╷
│ Error: cannot destroy instance without setting deletion_protection=false and running `terraform apply`
│
│
╵
```


</details>

### DIFF
--- a/mmv1/products/apigee/Instance.yaml
+++ b/mmv1/products/apigee/Instance.yaml
@@ -44,6 +44,7 @@ async:
 custom_code:
   constants: 'templates/terraform/constants/apigee_instance.go.tmpl'
   custom_import: 'templates/terraform/custom_import/apigee_instance.go.tmpl'
+  pre_delete: 'templates/terraform/pre_delete/apigee_instance.go.tmpl'
 error_retry_predicates:
   - 'transport_tpg.IsApigeeRetryableError'
 exclude_sweeper: true
@@ -226,3 +227,9 @@ properties:
           The statusCode is the only expected/supported filter field. (Ex: statusCode)
           The filter will parse it to the Common Expression Language semantics for expression
           evaluation to build the filter condition. (Ex: "filter": statusCode >= 200 && statusCode < 300 )
+virtual_fields:
+  - name: 'deletionProtection'
+    type: Boolean
+    description: "Optional. If set to true deletion of the instance will fail."
+    # TODO: change to `true` in the next major release (8.0.0)
+    # default_value: true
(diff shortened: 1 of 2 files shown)


slevenick (commented):
The implementation is fine here, but we are looking at a more systematic way to solve this issue in the same way across resources in the provider. At this point I'd hesitate to introduce more one-off implementations of deletion protection unless there is an important need for it. I don't think this is such a case, so I want to hold off on this PR until we have a standard way of doing this or we see more friction here

ramonvermeulen:
> The implementation is fine here, but we are looking at a more systematic way to solve this issue in the same way across resources in the provider. At this point I'd hesitate to introduce more one-off implementations of deletion protection unless there is an important need for it. I don't think this is such a case, so I want to hold off on this PR until we have a standard way of doing this or we see more friction here

Check, and also makes sense. Is essentially Introducing quite a bit of tech debt to have so much custom implementations for deletion protection. Shall I close the PR for now, or is this something on the near term roadmap? Then I will keep it open and change the implementation accordingly as soon as there is a standardized approach.

ramonvermeulen:
> I'd close for now, it'll remain linked in the associated bug so we can always come back to it as needed

Check, thanks! Is there an issue in which the more standardized deletion protection implementation can be tracked?

PR #15209, opened 2025-09-22
Add util function for waiting 1 min

chengliwm:
Fixes https://github.com/hashicorp/terraform-provider-google/issues/24421

Add a util function for waiting 1 min. This PR is a preparation for adding new resources (KAJ Policy Config), which will require waiting 1 min after creation/update.

```release-note:none
```


chengliwm:
> Looks like this template is not used by any resource yet, do you plan to merge this first then use it in different PRs? I feel like this change could be included in other PRs so that this could be tested before it's merged.

Yes, my plan was to merge this first and used in a different PR. I can put this util together with the resource that uses it.

PR #15212, opened 2025-09-23
Support the new `Broker` type of mirroring endpoint group.

duvni:
Mirroring endpoint group resource now supports DIRECT and BROKER types.
The default type (and existing type) is DIRECT, and this PR allows users of the beta provider to create resources with the new BROKER type.

```release-note:enhancement
networksecurity: added `type` and `mirroring_deployment_groups` fields to `google_network_security_mirroring_endpoint_group` resource (beta)
```

### DIFF
--- a/mmv1/products/networksecurity/MirroringEndpointGroup.yaml
+++ b/mmv1/products/networksecurity/MirroringEndpointGroup.yaml
@@ -108,8 +108,9 @@ properties:
       `projects/123456789/locations/global/mirroringDeploymentGroups/my-dg`.
       See https://google.aip.dev/124.
     immutable: true
-    conflicts:
-      - 'mirroring_deployment_groups'
+    exactly_one_of:
+      - mirroring_deployment_group
+      - mirroring_deployment_groups
   - name: 'mirroringDeploymentGroups'
     type: Array
     description: |-
@@ -121,8 +122,9 @@ properties:
     immutable: true
     item_type:
       type: String
-    conflicts:
-      - 'mirroring_deployment_group'
+    exactly_one_of:
+      - mirroring_deployment_group
+      - mirroring_deployment_groups
   - name: 'state'
     type: String
     description: |-


duvni:
@zli82016, following on the [terraform-provider-breaking-change-test](https://console.cloud.google.com/cloud-build/builds;region=global/0f53d2ee-222e-4b85-8d5c-d60758643805;step=17?project=graphite-docker-images) check that failed.
It is failing on the new `type` field that was added with a default value. I can confirm that at the API-level, the default matches the schema default which is DIRECT. This is true for all existing resources and is documented appropriately, so this change is safe and `override-breaking-change` label can be added.

See: https://googlecloudplatform.github.io/magic-modules/breaking-changes/breaking-changes#no-new-optional-default

mmv1/products/networksecurity/MirroringEndpointGroup.yaml
```diff
@@ -99,8 +107,22 @@ properties:
...
+      A list of the deployment groups that this BROKER endpoint group is
+      connected to, for example:
+      `projects/123456789/locations/global/mirroringDeploymentGroups/my-dg`.
+      See https://google.aip.dev/124.
+    min_version: 'beta'
+    immutable: true
+    item_type:
+      type: String
+    conflicts:
+      - 'mirroringDeploymentGroup'
```
zli82016:
```suggestion
      - 'mirroring_deployment_group'
```

mmv1/products/networksecurity/MirroringEndpointGroup.yaml
```diff
@@ -99,8 +107,22 @@ properties:
       The deployment group that this DIRECT endpoint group is connected to, for example:
       `projects/123456789/locations/global/mirroringDeploymentGroups/my-dg`.
       See https://google.aip.dev/124.
-    required: true
     immutable: true
+    conflicts:
+      - 'mirroringDeploymentGroups'
```
zli82016:
```suggestion
      - 'mirroring_deployment_groups'
```

mmv1/products/networksecurity/MirroringEndpointGroup.yaml
```diff
@@ -99,8 +107,22 @@ properties:
...
+  - name: 'mirroringDeploymentGroups'
```
zli82016, line 114: I wonder if one of `mirroringDeploymentGroups`, `mirroringDeploymentGroup` must be provided by the users? Or it is OK if neither is provided?

duvni: I used `conflicts` instead of `exactly_one_of` as I was concerned that it'll be treated as a breaking change.
However, following your suggestion I re-read the docs:
https://googlecloudplatform.github.io/magic-modules/breaking-changes/breaking-changes/

And since the `mirroring_deployment_group` field was required and not optional, this is actually a safe change, so changed to `exactly_one_of` for both fields, thanks!

mmv1/products/networksecurity/MirroringEndpointGroup.yaml
duvni: Done.

mmv1/products/networksecurity/MirroringEndpointGroup.yaml
duvni: Done.

zli82016 (approved):

PR #15230, opened 2025-09-24
tgc-revival: fix the service package in resource metadata

zli82016:
The TGC tests for service certificatemanager and networksecurity services  passed.

```release-note:none

```


zli82016:
> What happened with this particular product? Given that there're so many failed tests.



In the resource metadata, the value of `service` field was wrong. Instead of api_service_name, `service` should be the service package name.

When get the matched cai asset state from their [history](https://cloud.google.com/asset-inventory/docs/reference/rest/v1/TopLevel/batchGetAssetsHistory), these two services `certificatemanager` and `networksecurity` need special handling by checking their `service` field in resource metadata, as their cai assets history is not standard.

hao-nan-li (approved):

zli82016:
Failed VCR tests are unrelated.

PR #15232, opened 2025-09-25
Added `reservation` field to BigQuery Job configuration

vasilyl:
This allows a user to specify slots reservation name for a BigQuery job.

```release-note:enhancement
bigquery: added `reservation` field to `google_bigquery_job` resource (beta)
```

### DIFF
--- a/mmv1/products/bigquery/Job.yaml
+++ b/mmv1/products/bigquery/Job.yaml
@@ -876,6 +876,7 @@ properties:
         description: |
           The reservation that job would use. User can specify a reservation to execute the job. If this field is not set, reservation is determined based on the rules defined by the reservation assignments.
           The expected format is `projects/{project}/locations/{location}/reservations/{reservation}`.
+        min_version: beta
   - name: 'jobReference'
     type: NestedObject
     description: |


mmv1/products/bigquery/Job.yaml
```diff
@@ -859,6 +871,12 @@ properties:
                 type: String
                 description: 'The ID of the model.'
                 required: true
+      - name: 'reservation'
+        type: String
+        description: |
+          The reservation that job would use. User can specify a reservation to execute the job. If this field is not set, reservation is determined based on the rules defined by the reservation assignments.
+          The expected format is `projects/{project}/locations/{location}/reservations/{reservation}`.
+        required: false
```
rileykarson:
```suggestion
```

Being optional is implicit if no other modifiers are set!

mmv1/products/bigquery/Job.yaml
```diff
@@ -859,6 +871,12 @@ properties:
                 type: String
                 description: 'The ID of the model.'
                 required: true
+      - name: 'reservation'
+        type: String
+        description: |
+          The reservation that job would use. User can specify a reservation to execute the job. If this field is not set, reservation is determined based on the rules defined by the reservation assignments.
```
rileykarson, line 877: > If this field is not set, reservation is determined based on the rules defined by the reservation assignments.

Will a value be returned if the user didn't set one, if there's a match?

mmv1/products/bigquery/Job.yaml
```diff
@@ -859,6 +871,12 @@ properties:
                 type: String
                 description: 'The ID of the model.'
                 required: true
+      - name: 'reservation'
+        type: String
+        description: |
```
rileykarson, line 876: The central docs refer to setting this field as "Override a reservation on a query", at https://cloud.google.com/bigquery/docs/reservations-assignments#reservation

@wj-chen the description is currently from the API reference, but doesn't make the fact that this is an override clear. Should we leave it as-is or consider amending it to tie it to the product CUJ more?

mmv1/templates/terraform/examples/bigquery_job_query_reservation.tf.tmpl
```diff
@@ -0,0 +1,12 @@
...
+  provider = google-beta
+
+  job_id     = "{{index $.Vars "job_id"}}"
+
+  query {
+    query = "SELECT state FROM [lookerdata:cdc.project_tycho_reports]"
+
+  }
+
+  reservation = "projects/{{index $.TestEnvVars "project"}}/locations/US/reservations/{{index $.Vars "reservation_name"}}"
```
rileykarson, line 11: We'll ideally make a reservation and use it in our test- but currently have 0 slots on most/all test reservations. Will get back to you on this one.

mmv1/products/bigquery/Job.yaml
vasilyl: No, the effective reservation can be determined after the job created. It can be found in the job execution statistics response.
This field is a part of job insert request only.

mmv1/products/bigquery/Job.yaml
vasilyl: The widely used term _reservation override_ is not always correct. If a job doesn't have any default reservation assigned, this field would set the reservation, as overriding emptiness doesn't make sense.
The name _user specified reservation_ is better but too long.

wj-chen: I think this name is ok. It's simply "use this reservation for the job if specified" and consistent with other clients e.g. the BQ CLI.

mmv1/products/bigquery/Job.yaml
```diff
@@ -859,6 +871,11 @@ properties:
                 type: String
                 description: 'The ID of the model.'
                 required: true
+      - name: 'reservation'
```
wj-chen, line 874: Add `min_version: beta` for this property too if it's meant for a public preview release to the `google-beta` provider.

vasilyl: Done

vasilyl:
> @vasilyl @wj-chen is it expected that a reservation that doesn't exist gets accepted by the API? I think that's what the passing test indicates.

No, the server should fail to create a job if the specified reservation doesn't exist or located in another location or organization. Is it possible to confirm the test reservation with the generated name suffix exists or does not exist at the test execution time?

rileykarson:
Not authoritatively, easily, but we pick a random suffix such that it's unlikely it can succeed by change- let me run a second live test run & if that passes something is almost certainly up.

rileykarson:
Restricted access, but for convenience when I check back: https://hashicorp.teamcity.com/buildConfiguration/TerraformProviders_GoogleCloud_GOOGLE_BETA_MMUPSTREAMTESTS_GOOGLEBETA_PACKAGE_BIGQUERY/482365

The original live run's logs are at https://github.com/GoogleCloudPlatform/magic-modules/pull/15232#issuecomment-3335412749. You can see there that we didn't create a reservation (the config didn't ask to!) and we successfully create the job resource. Note that the job itself failed to execute due to being an invalid query, if error ordering matters, not sure which would happen first.

vasilyl:
> The original live run's logs are at [#15232 (comment)](https://github.com/GoogleCloudPlatform/magic-modules/pull/15232#issuecomment-3335412749). You can see there that we didn't create a reservation (the config didn't ask to!) and we successfully create the job resource. Note that the job itself failed to execute due to being an invalid query, if error ordering matters, not sure which would happen first.

This looks as expected: the created job resource contains failed job. Do you think we need another test for a successful job using existing reservation?

rileykarson:
I would have expected the service to validate the resource reference before starting parsing/execution of the job, but that may be appropriate due to the domain. That solves the problem that we'd need a valid reservation for testing, if we don't! I'll LGTM and merge.

> Do you think we need another test for a successful job using existing reservation?

Generally we would prefer a positive test, but reservations are a pain due to the billing implications so I think it's fine to forgo for this case. If you've confirmed that the success path works locally manually that'd be awesome- or if you could, if you haven't. If there are new issues that uncovers we can revert or fix forward as needed.

rileykarson (approved):

PR #15239, opened 2025-09-26
adding source_instance_deletion_time

tambalo-lgtm:

```release-note:enhancement
sql: added `source_instance_deletion_time` field to `google_sql_database_instance` resource
```
```release-note:enhancement
sql: added `source_instance_deletion_time` field to `google_sql_database_instance_latest_recovery_time` data source
```


mmv1/third_party/terraform/services/sql/resource_sql_database_instance.go.tmpl
```diff
@@ -1251,6 +1251,12 @@ API (for read pools, effective_availability_type may differ from availability_ty
 							Optional:    true,
 							Description: `The name of the allocated ip range for the private ip CloudSQL instance. For example: "google-managed-services-default". If set, the cloned instance ip will be created in the allocated range. The range name must comply with [RFC 1035](https://tools.ietf.org/html/rfc1035). Specifically, the name must be 1-63 characters long and match the regular expression [a-z]([-a-z0-9]*[a-z0-9])?.`,
 						},
+						"source_instance_deletion_time": {
```
melinath, line 1254: This sounds like it's an output-only field but currently it's marked as writable. Just wanted to double-check that's correct? If it's output only it should be marked `Computed` instead of `Optional`

mmv1/third_party/terraform/services/sql/data_source_sql_database_instance_latest_recovery_time.go
```diff
@@ -29,6 +28,11 @@ func DataSourceSqlDatabaseInstanceLatestRecoveryTime() *schema.Resource {
 				Computed:    true,
 				Description: `Timestamp, identifies the latest recovery time of the source instance.`,
 			},
+			"source_instance_deletion_time": {
```
melinath, line 31: same question for the data source.

mmv1/third_party/terraform/services/sql/resource_sql_database_instance.go.tmpl
tambalo-lgtm: We use the source_instance_deletion_time and instance name to identify the deleted instance for clone

mmv1/third_party/terraform/services/sql/data_source_sql_database_instance_latest_recovery_time.go
tambalo-lgtm: We use the source_instance_deletion_time and instance name to identify the deleted instance for latest recovery time

mmv1/third_party/terraform/services/sql/data_source_sql_database_instance_latest_recovery_time_test.go
```diff
@@ -52,8 +61,49 @@ resource "google_sql_database_instance" "main" {
...
+resource "time_sleep" "wait_for_instance" {
+  // Wait 30 seconds after the instance is created
+  depends_on = [google_sql_database_instance.main]
+  create_duration = "330s"
+}
+
 data "google_sql_database_instance_latest_recovery_time" "default" {
-  instance = resource.google_sql_database_instance.main.name
+  instance = google_sql_database_instance.main.name
+  source_instance_deletion_time = "2025-06-20T17:23:59.648821586Z"
```
melinath: I'm concerned about hardcoding this value in a test. How long will this string be good for? Given the main instance's name is random, can we actually use the same value for all test runs? Is there an API we can call to get this based on main's name instead of hardcoding it? For example, the other test sets this value using `data "google_sql_backup_run" "backup"`.

melinath (changes requested):
TestAccDataSourceSqlDatabaseInstanceLatestRecoveryTime_basic is failing on this PR with:

```
 Error: googleapi: Error 400: Request contains an invalid argument., badRequest
```

It's passing in nightlies, and you're modifying it, so this seems like a real failure caused by this PR (though the error from the API is unhelpful.)

mmv1/third_party/terraform/services/sql/data_source_sql_database_instance_latest_recovery_time_test.go
tambalo-lgtm: Functionally the timestamp string + random instance name should be non-breaking for all test runs since tf will pass them to sqladmin and will still not find a backup to complete the request.

melinath: ah, I missed that this was also expecting an error 👍 thanks!

mmv1/third_party/terraform/services/sql/data_source_sql_database_instance_latest_recovery_time_test.go
```diff
@@ -15,10 +16,19 @@ func TestAccDataSourceSqlDatabaseInstanceLatestRecoveryTime_basic(t *testing.T)
 	}
 	resourceName := "data.google_sql_database_instance_latest_recovery_time.default"
 
+	expectedError := regexp.MustCompile(`.*No backups found for instance.* and deletion time seconds: .*`)
+
+	if acctest.IsVcrEnabled() {
+		expectedError = regexp.MustCompile(`.*Error 400.*`)
+	}
```
melinath: Why would VCR being enabled change the error?

melinath:
Manual non-VCR test run (you won't be able to access this link): https://hashicorp.teamcity.com/buildConfiguration/TerraformProviders_GoogleCloud_GOOGLE_BETA_MMUPSTREAMTESTS_GOOGLEBETA_PACKAGE_SQL/487986

mmv1/third_party/terraform/services/sql/data_source_sql_database_instance_latest_recovery_time_test.go
tambalo-lgtm: https://yaqs.corp.google.com/eng/q/5932846691254272

Found that debugInfo field which contains the text we assert on is not visible to customers, so in the external test env we can't have this assertion

mmv1/third_party/terraform/services/sql/data_source_sql_database_instance_latest_recovery_time_test.go
tambalo-lgtm, line 1: Leaving the changes here since the test was originally not getting the latest recovery time, added a delay

tambalo-lgtm:
Recapping that the plan is to write a follow-up pr with the happy case integration tested, and to increase the helpfulness of the customer visible error logs which are currently opaque

Manual testing done to ensure feature works:
https://paste.googleplex.com/6008867393765376

melinath:
The failure for TestAccSqlDatabaseInstance_MysqlReadPoolEnableDisableSuccess happens in nightlies as well - unrelated to these changes.

PR #15240, opened 2025-09-26
feat(write-only): mark base field as `sensitive` during generation

ramonvermeulen:
Closes https://github.com/hashicorp/terraform-provider-google/issues/24484

```release-note:enhancement
provider: mark base field as sensitive for generated write-only arguments
```

@melinath 
Will not show a diff, because the only implementation has it already "manually" marked on the base field:
https://github.com/GoogleCloudPlatform/magic-modules/blob/e9c7b3bf64eca845688218688261064fbff27f2b/mmv1/products/compute/VpnTunnel.yaml#L181


melinath:

melinath:
attn @BBBmau FYI

ramonvermeulen:
> This PR should remove `sensitive` from the base field's yaml and still show no diff. It would also be good to update the contribution docs to reflect that this happens automatically.

Indeed makes sense to directly test it via removing the `senstive` key. Just updated the PR with https://github.com/GoogleCloudPlatform/magic-modules/pull/15240/commits/7eab39bf66605adde792d5376855a6d24e263f5b and https://github.com/GoogleCloudPlatform/magic-modules/pull/15240/commits/919b02896f3e06a6166da1900b70e59f1b9b508a, local it still doesn't produce a diff.

melinath (approved):

PR #15254, opened 2025-09-29
cloudquotas: resolved permadiff for `preferred_value` in `google_cloud_quotas_quota_preference`

ramonvermeulen:
Closes https://github.com/hashicorp/terraform-provider-google/issues/24480

```release-note:bug
cloudquotas: resolved permadiff for `preferred_value` in `google_cloud_quotas_quota_preference`
```

### DIFF
--- a/mmv1/products/cloudquotas/QuotaPreference.yaml
+++ b/mmv1/products/cloudquotas/QuotaPreference.yaml
@@ -91,6 +91,7 @@ properties:
         description: |
           The preferred value. Must be greater than or equal to -1. If set to -1, it means the value is "unlimited".
         required: true
+        custom_flatten: 'templates/terraform/custom_flatten/cloudquotas_quota_preference_preferred_value.go.tmpl'
       - name: 'stateDetail'
         type: String
         description: Optional details about the state of this quota preference.
--- a/mmv1/templates/terraform/custom_flatten/cloudquotas_quota_preference_preferred_value.go.tmpl
+++ b/mmv1/templates/terraform/custom_flatten/cloudquotas_quota_preference_preferred_value.go.tmpl
@@ -1,0 +1,4 @@
+func flatten{{$.GetPrefix}}{{$.TitlelizeProperty}}(v interface{}, d *schema.ResourceData, config *transport_tpg.Config) interface{} {
+    // ignore read on this field
+    return d.Get("quota_config.0.preferred_value")
+}


NickElliot (changes requested):
this would not be the ideal approach here:
would be better off making a diff suppress func that will suppress the diff on update if the plan is null->0

ramonvermeulen:
> this would not be the ideal approach here: would be better off making a diff suppress func that will suppress the diff on update if the plan is null->0

Check, I will change this to a diff_suppress_func. I went with custom_flatten because of the documentation on diff_suppress_func states:

_"Specifies the name of a [diff suppress function](https://developer.hashicorp.com/terraform/plugin/sdkv2/schemas/schema-behaviors#diffsuppressfunc) to use for this field. **In many cases, a [custom flattener](https://googlecloudplatform.github.io/magic-modules/develop/custom-code/#custom_flatten) is preferred because it will allow the user to see a clearer diff** when the field actually is being changed."_

ramonvermeulen:
Just changed to a diff_suppress_func in https://github.com/GoogleCloudPlatform/magic-modules/commit/dbad345a5d4443e960a6796889d1087801eced5a + I added it as a common diff suppress func, because I think this is quite a common case (string value not returned by API) and couldn't find an existing function for it yet.

I tested this by first reproducing the permadiff by apply-ing via the latest provider version. Then I applied again with my own build, which did not cause the diff.

mmv1/third_party/terraform/tpgresource/common_diff_suppress.go
```diff
@@ -22,6 +22,11 @@ func EmptyOrFalseSuppressBoolean(k, old, new string, d *schema.ResourceData) boo
 	return (o == nil && !n.(bool))
 }
 
+func EmptyStringOrNilDiffSuppress(k, old, new string, d *schema.ResourceData) bool {
+	o, n := d.GetChange(k)
+	return (o == nil || o == "") && (n != nil && n != "")
```
ScottSuarez: I don't believe this solves for the usecase from your yaqs/3061527455379488768. In your specific case you should be checking `(o == nil || o == "" || o == "0")`, no?

ScottSuarez: Which then we should make this a function specific to this resource instead of generalized.

ramonvermeulen: What is yaqs/3061527455379488768 about? I manually tested this and for me it resolves the permadiff. Why would you like to extend the validation to check if `old == "0"`? In that scenario there is no diff suppress required right? Because if old == "0" the `preferred_value` was already set correct in state in the first place right? and no permadiff would occur?

I'm not sure if I fully understand the scenario, but in my mind the permadiff would occur because the resource is created with e.g. `preferred_value = "0"`, however the GET endpoint doesn't return `preferred_value` resulting in that the key/value pair not written to state during the fetch directly after the create, the next plan/apply this gives the permadiff. Therefore with a diff suppress func you suppress on exactly that scenario where the old value is not set and the new value is.

ScottSuarez: Based on the name `EmptyStringOrNilDiffSuppress`, it seems like it should suppress diffs only when going from empty/nil to empty/nil (treating them as equivalent). But instead, it's suppressing any change from empty/nil to ANY non-empty value.

This is very broad diff suppress and essentially ignoring API value value when empty. That's not what diff suppress is typically used for as we want to target the specific scenario in question.

| would prefer we do something specific to the resource. Like

```
func QuotaPreferredValueDiffSuppress(k, old, new string, d *schema.ResourceData) bool {
	o, n := d.GetChange(k)
	
	// API doesn't return preferred_value when it's "0", so suppress the diff when:
	// - Old (from API) is empty/nil and New (from config) is "0"
	// - OR Old (from API) is "0" and New (from config) is empty/nil (for safety)
	oldEmpty := o == nil || o == ""
	newEmpty := n == nil || n == ""
	
	return (oldEmpty && n == "0") || (o == "0" && newEmpty)
}
```

This function can be specific in the resource file directly by adding it to the constants import like so

https://github.com/ScottSuarez/magic-modules/blob/c5c6f66b096d3e24c16f84e3ffd15acfd3f37fa6/mmv1/products/accesscontextmanager/ServicePerimeter.yaml#L66

ramonvermeulen:
> Based on the name `EmptyStringOrNilDiffSuppress`, it seems like it should suppress diffs only when going from empty/nil to empty/nil (treating them as equivalent). But instead, it's suppressing any change from empty/nil to ANY non-empty value.
> 
> This is very broad diff suppress and essentially ignoring API value value when empty. That's not what diff suppress is typically used for as we want to target the specific scenario in question.
> 
> | would prefer we do something specific to the resource. Like
> 
> ```
> func QuotaPreferredValueDiffSuppress(k, old, new string, d *schema.ResourceData) bool {
> 	o, n := d.GetChange(k)
> 	
> 	// API doesn't return preferred_value when it's "0", so suppress the diff when:
> 	// - Old (from API) is empty/nil and New (from config) is "0"
> 	// - OR Old (from API) is "0" and New (from config) is empty/nil (for safety)
> 	oldEmpty := o == nil || o == ""
> 	newEmpty := n == nil || n == ""
> 	
> 	return (oldEmpty && n == "0") || (o == "0" && newEmpty)
> }
> ```
> 
> This function can be specific in the resource file directly by adding it to the constants import like so
> 
> https://github.com/ScottSuarez/magic-modules/blob/c5c6f66b096d3e24c16f84e3ffd15acfd3f37fa6/mmv1/products/accesscontextmanager/ServicePerimeter.yaml#L66

Check, I moved towards a more scenario specific `diff_suppress_func` in https://github.com/GoogleCloudPlatform/magic-modules/pull/15254/commits/031999da127c6cc6b2d5c5b68f33ad69a2334059.

Manual testing:

<details>

1) Reproduce permadiff:

```terraform
resource "google_cloud_quotas_quota_preference" "project_gpus_quotas_preference" {
  parent = "projects/${local.project_id}"
  dimensions = {
    region = "europe-west6"
  }
  service  = "aiplatform.googleapis.com"
  quota_id = "RestrictedImageTrainingV100GPUsPerProjectPerRegion"
  quota_config {
    preferred_value = "0"
  }
}
```

```console
➜ terraform apply -auto-approve

  + create

  # google_cloud_quotas_quota_preference.project_gpus_quotas_preference will be created
  + resource "google_cloud_quotas_quota_preference" "project_gpus_quotas_preference" {
      + create_time          = (known after apply)
      + dimensions           = {
          + "region" = "europe-west6"
        }
      + etag                 = (known after apply)
      + id                   = (known after apply)
      + ignore_safety_checks = "QUOTA_SAFETY_CHECK_UNSPECIFIED"
      + name                 = (known after apply)
      + parent               = "projects/<masked>"
      + quota_id             = "RestrictedImageTrainingV100GPUsPerProjectPerRegion"
      + reconciling          = (known after apply)
      + service              = "aiplatform.googleapis.com"
      + update_time          = (known after apply)

      + quota_config {
          + granted_value   = (known after apply)
          + preferred_value = "0"
          + request_origin  = (known after apply)
          + state_detail    = (known after apply)
          + trace_id        = (known after apply)
        }
    }

google_cloud_quotas_quota_preference.project_gpus_quotas_preference: Creating...
google_cloud_quotas_quota_preference.project_gpus_quotas_preference: Creation complete after 5s [id=projects/<masked>/locations/global/quotaPreferences/134fe860-31c6-4489-920c-b489be4247e9]

Apply complete! Resources: 1 added, 0 changed, 0 destroyed.
```

Next apply:

```console
➜ terraform apply -auto-approve
google_cloud_quotas_quota_preference.project_gpus_quotas_preference: Refreshing state... [id=projects/<masked>/locations/global/quotaPreferences/134fe860-31c6-4489-920c-b489be4247e9]

  ~ update in-place

  # google_cloud_quotas_quota_preference.project_gpus_quotas_preference will be updated in-place
  ~ resource "google_cloud_quotas_quota_preference" "project_gpus_quotas_preference" {
        id                   = "projects/<masked>/locations/global/quotaPreferences/134fe860-31c6-4489-920c-b489be4247e9"
        name                 = "134fe860-31c6-4489-920c-b489be4247e9"
        # (9 unchanged attributes hidden)

      ~ quota_config {
          + preferred_value = "0"
            # (5 unchanged attributes hidden)
        }
    }

google_cloud_quotas_quota_preference.project_gpus_quotas_preference: Modifying... [id=projects/<masked>/locations/global/quotaPreferences/134fe860-31c6-4489-920c-b489be4247e9]
google_cloud_quotas_quota_preference.project_gpus_quotas_preference: Modifications complete after 5s [id=projects/<masked>/locations/global/quotaPreferences/134fe860-31c6-4489-920c-b489be4247e9]

Apply complete! Resources: 0 added, 1 changed, 0 destroyed.
```

2) With the cusom build of commit https://github.com/GoogleCloudPlatform/magic-modules/pull/15254/commits/031999da127c6cc6b2d5c5b68f33ad69a2334059.

```console
➜ TF_CLI_CONFIG_FILE="$(pwd)/tf-dev-override.tfrc" terraform apply -auto-approve

╷
│
│
╵
google_cloud_quotas_quota_preference.project_gpus_quotas_preference: Refreshing state... [id=projects/<masked>/locations/global/quotaPreferences/134fe860-31c6-4489-920c-b489be4247e9]

```
</details>

ScottSuarez (approved):

ScottSuarez:
Thanks for the work

NickElliot (approved):

PR #15257, opened 2025-09-29
Add new resource ProjectKajPolicyConfig

chengliwm:
Fixes https://github.com/hashicorp/terraform-provider-google/issues/24421

```release-note:new-resource
`google_kms_project_kaj_policy_config`
```

### DIFF
--- a/mmv1/products/kms/ProjectKajPolicyConfig.yaml
+++ b/mmv1/products/kms/ProjectKajPolicyConfig.yaml
@@ -54,8 +54,12 @@ examples:
   - name: 'kms_project_kaj_policy_config_basic'
     primary_resource_id: 'example'
     min_version: 'beta'
+    vars:
+      project_id: 'my-project'
     test_env_vars:
-      project_id: 'PROJECT_NAME'
+      org_id: 'ORG_ID'
+      billing_account: 'BILLING_ACCT'
+    external_providers: ["time"]
 parameters:
 properties:
   - name: 'defaultKeyAccessJustificationPolicy'
--- a/mmv1/third_party/terraform/services/kms/resource_kms_project_kaj_policy_config_test.go.tmpl
+++ b/mmv1/third_party/terraform/services/kms/resource_kms_project_kaj_policy_config_test.go.tmpl
@@ -15,13 +15,17 @@ func TestAccKMSProjectKajPolicyConfig_update(t *testing.T) {
 	t.Parallel()
 
 	context := map[string]interface{}{
-		"project_id":    envvar.GetTestProjectFromEnv(),
-		"random_suffix": acctest.RandString(t, 10),
-	}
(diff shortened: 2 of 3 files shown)


shuyama1:
/gcbrun

rerunning the tests to confirm if they can be executed in parallel

chengliwm:
> /gcbrun
> 
> rerunning the tests to confirm if they can be executed in parallel

The acc test cannot run in parallel because this resource is a singleton resource. Tests will run into data racing issue if run in parallel. In this case, shall I disable VCR tests?

mmv1/products/kms/ProjectKajPolicyConfig.yaml
```diff
@@ -0,0 +1,91 @@
...
+  - 'projects/{{project}}/kajPolicyConfig'
+timeouts:
+  insert_minutes: 20
+  update_minutes: 20
+  delete_minutes: 20
+custom_code:
+  post_create: 'templates/terraform/post_create/sleep_1_min.go.tmpl'
+  post_update: 'templates/terraform/post_create/sleep_1_min.go.tmpl'
+examples:
+  - name: 'kms_project_kaj_policy_config_basic'
```
shuyama1, line 54: Our tests are running in parallel in presubmit and nightly. To prevent test failures from parallel runs on this singleton resource, please skip this basic test([`exclude_test: true`](https://googlecloudplatform.github.io/magic-modules/reference/resource/#examples) and this example will be used to generate the example usage for documentation). The update test will provide sufficient test coverage.

mmv1/third_party/terraform/services/kms/resource_kms_project_kaj_policy_config_test.go.tmpl
```diff
@@ -0,0 +1,79 @@
...
+			{
+				ResourceName:      "google_kms_project_kaj_policy_config.example",
+				ImportState:       true,
+				ImportStateVerify: true,
+			},
+		},
+	})
+}
+
+func testAccKMSProjectKajPolicyConfig_basic(context map[string]interface{}) string {
```
shuyama1, line 55: I'd recommend creating a new project within the test for project-level policy testing instead of using the test project, which is used for all Terraform tests and may affect other tests due to policy change.

chengliwm: Done. Created a new project in the example and the update test.

mmv1/products/kms/ProjectKajPolicyConfig.yaml
chengliwm: Since new projects will be created in the test, is the "exclude_test:true" still needed?

shuyama1: Good point! Probably not needed, given they're in different projects now.

PR #8050, opened 2023-05-31
add vmwareengine network resource and datasource

codding-buddha:
Added basic support for VMwareEngine Network resource and datasource.
Version for the resource/datasource is currently set to beta
Fixes part of https://github.com/hashicorp/terraform-provider-google/issues/8516

If this PR is for Terraform, I acknowledge that I have:

- [X] Searched through the [issue tracker](https://github.com/hashicorp/terraform-provider-google/issues) for an open issue that this either resolves or contributes to, commented on it to claim it, and written "fixes {url}" or "part of {url}" in this PR description. If there were no relevant open issues, I opened one and commented that I would like to work on it (not necessary for very small changes).
- [X] Ensured that all new fields I added that can be set by a user appear in at least one [example](https://github.com/GoogleCloudPlatform/magic-modules/tree/main/mmv1/templates/terraform/examples) (for generated resources) or [third_party test](https://github.com/GoogleCloudPlatform/magic-modules/tree/main/mmv1/third_party/terraform/tests) (for handwritten resources or update tests).
- [X ] [Generated Terraform providers](https://github.com/GoogleCloudPlatform/magic-modules/blob/main/docs/content/docs/getting-started/generate-providers.md), and ran [`make test` and `make lint`](https://googlecloudplatform.github.io/magic-modules/docs/getting-started/run-provider-tests/#run-unit-tests) in the generated providers to ensure it passes unit and linter tests.
- [X] [Ran](https://github.com/GoogleCloudPlatform/magic-modules/blob/main/docs/content/docs/getting-started/run-provider-tests.md) relevant acceptance tests using my own Google Cloud project and credentials (If the acceptance tests do not yet pass or you are unable to run them, please let your reviewer know).
- [X] Read the [Release Notes Guide](https://github.com/GoogleCloudPlatform/magic-modules/blob/main/.ci/RELEASE_NOTES_GUIDE.md) before writing my release note below.

**Release Note Template for Downstream PRs (will be copied)**

```release-note:new-resource
`google_vmwareengine_network`
```

```release-note:new-datasource
`google_vmwareengine_network`
```

PR #15259, opened 2025-09-29
Promote some remaining vmwareengine fields

roaks3:
These were originally included in EAP, but then removed during promotion (see comment https://github.com/GoogleCloudPlatform/magic-modules/pull/8050#discussion_r1214466910). We have since changed our stance on these types of fields since it is more straightforward to simply cover all fields in most cases (and this makes coverage calculation and autogen behavior more predictable).

@codding-buddha could you confirm that it is reasonable to include these fields?

```release-note:none

```

### DIFF
--- a/mmv1/products/vmwareengine/Network.yaml
+++ b/mmv1/products/vmwareengine/Network.yaml
@@ -80,6 +80,20 @@ parameters:
     required: true
     immutable: true
 properties:
+  - name: 'createTime'
+    type: Time
+    description: |
+      Creation time of this resource.
+      A timestamp in RFC3339 UTC "Zulu" format, with nanosecond resolution and up to nine fractional digits.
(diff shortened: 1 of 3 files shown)


roaks3:

swamitagupta:
LGTM!

mmv1/products/vmwareengine/Network.yaml
```diff
@@ -132,3 +146,9 @@ properties:
     description: |
       System-generated unique identifier for the resource.
     output: true
+  - name: 'etag'
```
slevenick, line 149: @swamitagupta mentioned not adding etag here, but I don't see the comment anymore?

Do we want to remove this or keep it?

swamitagupta: We can keep etag here.

PR #15267, opened 2025-09-29
POC: Stop using goimports

SirGitsalot:
This PR is intended for discussion for now. It:

- Adds blank imports to templates to avoid unused import errors
- Adds the most common imports that weren't already included to avoid undefined package errors
- Adds uncommon imports to the custom code in `constants` for each resource that requires them
- Removes `if` conditions around imports

Placing an import at the top of the constant custom code template is a convenient location, as it gets merged into the top of the resource template just below the imports. That won't work for tests and data sources, but they're not currently needed for any of those.

With this PR the beta provider builds successfully and the unit tests pass (the integration tests build, but I haven't run them). I also have a set of changes for the TGC templates, but they're not building for me (unrelated to imports, it's getting a lot of big changes right now and I probably just got unlucky when I synced).

I think the best way to implement this would be to:

1. Make the template changes w/o disabling goimports 
2. Add a flag to the MM compiler to skip goimports and enable the flag for TPG
3. Repeat for TGC (and other targets) when it's convenient to do so

```release-note:none
```

### DIFF
--- a/mmv1/third_party/terraform/fwprovider/data_source_provider_config_plugin_framework.go
+++ b/mmv1/third_party/terraform/fwprovider/data_source_provider_config_plugin_framework.go
@@ -230,11 +230,7 @@ func (d *GoogleProviderConfigPluginFrameworkDataSource) Read(ctx context.Context
 	data.RequestReason = types.StringValue(d.providerConfig.RequestReason)
 	data.RequestTimeout = types.StringValue(d.providerConfig.RequestTimeout.String())
 
-	lbs := make(map[string]attr.Value, len(d.providerConfig.DefaultLabels))
-	for k, v := range d.providerConfig.DefaultLabels {
-		lbs[k] = types.StringValue(v)
-	}
-	labels, di := types.MapValueFrom(ctx, types.StringType, lbs)
+	labels, di := types.MapValueFrom(ctx, types.StringType, d.providerConfig.DefaultLabels)
 	if di.HasError() {
 		resp.Diagnostics.Append(di...)
 	}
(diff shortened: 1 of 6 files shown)


SirGitsalot:
OK, I figured out the problem I was having with building TGC and it's now working sans `goimports`

melinath:
@ScottSuarez what do you think?

melinath:
> This seems like a reasonable approach & LGTM. I'd be inclined to make this change and disable goimports globally ASAP afterwards. The thing to solve / make sure it still works is that we'd want to keep running goimports for non-PR downstream builds (but stop running it for PR downstream builds.)

To clarify, I want to make PRs start breaking ASAP so that we don't have to redo this work in a few months, but I think having latitude for things that have already run tests & might just be approved & merged would make sense, so that we don't accidentally break main.

ScottSuarez:
This approach definitely works. I wouldn't want this to be the long term solution but if we leave a comment explaining the code or maybe put it towards the bottom of the file. It would make things more readable. Either/or. This opinion isn't blocking though

ScottSuarez:
Are we looking to merge this?

melinath:
yes

ScottSuarez:
> yes

okay, just making sure. I'll dismiss my review.

SirGitsalot:
I rebased and fixed a merge conflict but didn't make any other changes, if you want to take another loook.

mmv1/provider/template_data.go
melinath, line 298: Should this entire FixImports function be removed now?

melinath (approved):
LGTM - there's a bit of cleanup to do but it could happen in a separate PR.

mmv1/provider/template_data.go
SirGitsalot: No need for it now, although I suppose we could have an optional compiler flag that calls it as a debugging tool for things like yesterday's mystery missing import issue.

melinath: I'd be inclined to just get rid of it personally, on the theory that we can run `goimports` to figure out what's missing... but I haven't actually worked with this as much as you have, so I'm happy to leave it in since you think it's useful.

PR #15268, opened 2025-09-29
GKE: Conditionally send the cpu_cfs_quota field based on presence in config

rileykarson:
Fixes https://github.com/hashicorp/terraform-provider-google/issues/15767

This conditionally sets the value of the field to true in the API request when null in config. This is technically a breaking change, as recreating a resource with the value set to null would result in different behaviour, but well-contained enough that I feel comfortable making it in a minor release. It may also be sufficient to not need to introduce a `Default` in the future, which would retroactively impact other users as well, causing a lengthy (10+ minutes even on test pools) upgrade.

There are a lot of cases to consider; I've gone through most of them manually by creating failing request messages and verifying them by hand below. In acceptance tests, to capture the new behaviour I've added a case that checks that explicit `false` is set correctly yielding no diff, and null yields a `true` in the request/server-side as expected. This exercises the more difficult to handle path- `google_container_cluster.node_pool.N.node_config.0.kubelet_config.0.cpu_cfs_quota`- and we should see no-ops on other GKE tests (after a VCR rerecord for some, since I did change the request message!).

I did not add acceptance tests for the simpler variants- `google_container_cluster.node_config.0.kubelet_config.0.cpu_cfs_quota` and `google_container_node_pool.node_config.0.kubelet_config.0.cpu_cfs_quota`- as other tests should cover the deep RawConfig access, the logic is trivial to review (we don't enter the prefix block), and GKE tests are pretty weighty. If you feel they'd add enough value, please request 'em.

I used a bit to break up the configs: http://go/mm-pr-15268-bit Please verify that for each of the 3 paths the field appears at:

* If kubelet config is undefined, no `cpuCfsQuota` value is sent
* If cpu_cfs_quota is null, `true` is sent for `cpuCfsQuota`
* If cpu_cfs_quota is true, `true` is sent for `cpuCfsQuota`
* If cpu_cfs_quota is false, `false` is sent for `cpuCfsQuota`

```release-note:breaking-change
container: `node_config` blocks that had set `kubelet_config` without explicitly setting `cpu_cfs_quota` implicitly set `cfu_cfs_quota` to `false` when unset. From this version onwards, an unset `cpu_cfs_quota` will instead match the API default of true `true`. Resources that are recreated will receive the new value; old resources are unaffected, and may change values by explicitly setting the intended one.
```

```release-note:bug
container: Fixed the default for `node_config.kubelet_config.cpu_cfs_quota` on `google_container_cluster`, `google_container_node_pool`, `google_container_cluster.node_pool` to align with the API. Terraform will now send a `true` value when the field is unset on creation, and preserve any previously set value when unset. Explicitly set values will work as defined in configuration.
```

### DIFF
--- a/mmv1/third_party/terraform/services/container/node_config.go.tmpl
+++ b/mmv1/third_party/terraform/services/container/node_config.go.tmpl
@@ -1,6 +1,7 @@
 package container
 
 import (
+	"fmt"
 	"log"
 	"strconv"
 	"strings"
--- a/mmv1/third_party/terraform/services/container/node_config.go.tmpl
+++ b/mmv1/third_party/terraform/services/container/node_config.go.tmpl
@@ -2,9 +2,11 @@ package container
 
 import (
 	"log"
+	"strconv"
 	"strings"
 	"time"
 
+	"github.com/hashicorp/go-cty/cty"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/validation"
 
@@ -659,6 +661,7 @@ func schemaNodeConfig() *schema.Schema {
 							},
 							"cpu_cfs_quota": {
 								Type:     schema.TypeBool,
+								Computed: true,
 								Optional: true,
 								Description: `Enable CPU CFS quota enforcement for containers that specify CPU limits.`,
 							},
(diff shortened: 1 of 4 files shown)


rileykarson:
Remaining failing tests are unrelated; GKE has a number of standing failures atm.

NickElliot:
Would we consider returning to the standard breaking-solution in a major release or leave it like this?

rileykarson:
I'd recommend leaving it as-is, since the breaking approach will retroactively impact old pools and kick off a long upgrade time, as well as restore drift detection when the field is unset. It'd be fine to do technically, just possibly not worth the end user impact.

PR #15170, opened 2025-09-17
`resourceIdentityGeneration`: use nil check instead of error when calling `identity()`

BBBmau:
after running tests, we ran into missing identity() errors being shown due to

- handwritten resources not having the identity schema but utilizing the identity in `importStateUpgrade`
- using `identity()` in data sources

this PR makes it so that we do a nil check of the identityData type and prevent the use of identtiy setting and getting in the event that identityData is nil

first discussed here:
- https://github.com/GoogleCloudPlatform/magic-modules/pull/15074

**Release Note Template for Downstream PRs (will be copied)**

See [Write release notes](https://googlecloudplatform.github.io/magic-modules/contribute/release-notes/) for guidance.

```release-note:none

```

PR #15273, opened 2025-09-30
Revert "`resourceIdentityGeneration`: use nil check instead of error when calling `identity()`"

BBBmau:
Reverts GoogleCloudPlatform/magic-modules#15170


BBBmau:
@shuyama1 we get a timeout issue with the VCR testing on the revert

PR #15202, opened 2025-09-22
Add activity journal feature for google_storage_insights_dataset_config

sameer-google:
```release-note:enhancement
storageinsights: added `activityDataRetentionPeriodDays` fields to `google_storage_insights_dataset_config` resource
```

PR #15279, opened 2025-10-01
Revert "Add activity journal feature for google_storage_insights_dataset_config"

sameer-google:
Reverts GoogleCloudPlatform/magic-modules#15202

```release-note:breaking-change
storageinsights: removed `activity_data_retention_period_days` field from `google_storage_insights_dataset_config` resource due to a delayed launch. It will be readded when the feature launches.
```

### DIFF
--- a/mmv1/products/storageinsights/DatasetConfig.yaml
+++ b/mmv1/products/storageinsights/DatasetConfig.yaml
@@ -119,10 +119,6 @@ properties:
     description: |
       Number of days of history that must be retained.
     required: true
-  - name: 'activityDataRetentionPeriodDays'
-    type: Integer
-    description: |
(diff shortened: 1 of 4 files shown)


sameer-google:
> Can you provide more context on this (Internal is fine)? The feature went out in a release, so a removal is a little complicated.

The feature was expected to be GA Sept 30, and terraform CLI support was planned accordingly, due to few reasons the GA timelines have been postponed hence the test failures have been raised and changes are released in prod. We are reverting and these will be added back once the API is completely GA.

zli82016:
Some information is in b/448219812, if it is helpful.

rileykarson:
@hao-nan-li can you review and resolve as oncall? Since the API returned a default value, we need to evaluate whether the revert works for customers who applied a resource on the affected versions, or if we need to mark the field O+C and keep it around until proper launch.

@modular-magician reassign-reviewer hao-nan-li

PR #15281, opened 2025-10-01
Keep base yaml source file location when overrides are used

roaks3:
See EAP Diff at https://cloud-internal-review.git.corp.google.com/c/cloud-graphite-eng/terraform-next/+/53490

TLDR: We were losing the path to the source yaml file when overrides were used, and it creates a nosier diff when we try to compare across providers.

```release-note:none

```


roaks3:

PR #15290, opened 2025-10-02
feat(oracledatabase): `google_oracle_database_autonomous_database` `private_endpoint_ip` should be output

adamstrawson:
When creating an Oracle autonomous database via terraform, it's often needed to retrieve the endpoint IP so that it can later be used in additional resources. 

This change sets the `output` flag to `true` to enable this. 

```release-note:enhancement
oracledatabase: output `private_endpoint_ip ` field in `google_oracle_database_autonomous_database` resource
```

### DIFF
--- a/mmv1/products/oracledatabase/AutonomousDatabase.yaml
+++ b/mmv1/products/oracledatabase/AutonomousDatabase.yaml
@@ -195,6 +195,7 @@ properties:
       - name: 'privateEndpointIp'
         type: String
         description: 'The private endpoint IP address for the Autonomous Database. '
+        output: true
         default_from_api: true
       - name: 'privateEndpointLabel'
         type: String


mmv1/products/oracledatabase/AutonomousDatabase.yaml
```diff
@@ -195,6 +195,7 @@ properties:
       - name: 'privateEndpointIp'
         type: String
         description: 'The private endpoint IP address for the Autonomous Database. '
+        output: true
```
zli82016, line 198: I don't think `Output: true` is needed here. 
1. `default_from_api: true` means the field is Computed + Optional. `Output: true` means the field is computed. Withe the usage of `default_from_api: true`, `Output: true` is redundant.


2. The field `private_endpoint_ip` can be retrieved  and used in additional resources now. Do you see any issues? 
 
Thanks.

adamstrawson: Thanks for clarification, useful to understand. 

Also found that `private_endpoint_ip` does indeed exist, it wasn't apparent as `properties` is a list of objects, rather than just a map, so needed to be `properties[0].private_endpoint_ip`

PR #15292, opened 2025-10-02
`.ci`: update Dockerfile to reference go v1.24

BBBmau:
We need to have this merged first in order for CI to have the latest go version initialized prior to merging:
- https://github.com/GoogleCloudPlatform/magic-modules/pull/15293

**Release Note Template for Downstream PRs (will be copied)**

See [Write release notes](https://googlecloudplatform.github.io/magic-modules/contribute/release-notes/) for guidance.

```release-note:none

```

PR #15293, opened 2025-10-02
`GO`: Upgrade Go version from `1.23` -> `1.24`

BBBmau:
This version bump is required in order to bump the version for sdk/v2 from `0.27.0`- > `0.28.1`. The latest version of the SDK/v2 contains a bug fix for resource identity that was merged into sdk/v2 repo here:

- https://github.com/hashicorp/terraform-plugin-sdk/pull/1527

The following PRs are also required in order to complete the Go version upgrade:
- https://github.com/GoogleCloudPlatform/magic-modules/pull/15292
- https://github.com/GoogleCloudPlatform/terraform-google-conversion/pull/4566

```release-note:none

```


docs/go.mod
```diff
@@ -1,7 +1,3 @@
 module github.com/GoogleCloudPlatform/magic-modules/docs
 
-go 1.23
-
-require (
-	github.com/alex-shpak/hugo-book/v12 v12.0.1-0.20250927134827-c58c73a59c10 // indirect
-)
+go 1.24
```
melinath, line 3: This doesn't seem like it should delete the rest of the go.mod & go.sum files?

tools/template-check/go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/GoogleCloudPlatform/magic-modules/tools/template-check
 
-go 1.23.0
+go 1.24.0
 
 toolchain go1.23.1
```
melinath: nit: we should be able to delete this line, I think?

```suggestion
```

melinath (changes requested):
This also needs to update `mmv1/third_party/tgc_next/go.mod`.

Just want to double-check that you ran `go mod tidy` for all of the changed go.mod files?

docs/go.mod
BBBmau: ah yeah, let me add this back in

melinath:
@BBBmau just to double-check, you ran `go mod tidy` for all the modified go.mod files, right?

BBBmau:
> @BBBmau just to double-check, you ran `go mod tidy` for all the modified go.mod files, right?
@melinath 

yes went through each directory and ran `go mod tidy` while also adding in any required dependencies that wre removed from running `tidy`

I did just go through one pass through and we get errors when attempting to do `go mod tidy` in `tools/diff-processor` though the same errors appears even when `go 1.23` is set:

```
go: github.com/GoogleCloudPlatform/magic-modules/tools/diff-processor/cmd imports
        google/provider/new/google/provider imports
        github.com/hashicorp/terraform-provider-google/google/services/accessapproval: module github.com/hashicorp/terraform-provider-google@latest found (v1.20.0), but does not contain package github.com/hashicorp/terraform-provider-google/google/services/accessapproval
```

melinath:
It looks like the build is failing due to the workspace not running go 1.24 yet (because we haven't upgraded the images yet.)

melinath (approved):

PR #15296, opened 2025-10-02
add new fields and update examples

kefang2:

```release-note:enhancement
discoveryengine: added `connector_modes`, `sync_mode`, `incremental_refresh_interval`, `auto_run_disabled`, and `incremental_sync_disabled` fields to `google_discovery_engine_data_connector` resource
```

```release-note:enhancement
discoveryengine: added `kms_key_name` field to `google_discovery_engine_search_engine` resource
```

```release-note:bug
discoveryengine: fixed bug where it wasn't possible to specify values for `knowledgeBaseSysId` or `catalogSysId` in `google_discovery_engine_data_connector.entities.params`.
```

```release-note:breaking-change
discoveryengine: changed type of `google_discovery_engine_data_connector.entities.params`. Previously, it was a map of string keys to string values; now, it must be a [JSON-encoded](https://developer.hashicorp.com/terraform/language/functions/jsonencode) string containing an object. This change is being made in a minor release because the field wasn't usable as intended – specifically, all current valid uses require mapping strings to _lists_ of strings.
```

### DIFF
--- a/mmv1/third_party/terraform/services/discoveryengine/resource_discovery_engine_search_engine_test.go
+++ b/mmv1/third_party/terraform/services/discoveryengine/resource_discovery_engine_search_engine_test.go
@@ -29,7 +29,7 @@ func TestAccDiscoveryEngineSearchEngine_discoveryengineSearchengineBasicExample_
 				ResourceName:            "google_discovery_engine_search_engine.basic",
 				ImportState:             true,
 				ImportStateVerify:       true,
-				ImportStateVerifyIgnore: []string{"engine_id", "collection_id", "location"},
+				ImportStateVerifyIgnore: []string{"engine_id", "collection_id", "location", "kms_key_name"},
 			},
 			{
 				Config: testAccDiscoveryEngineSearchEngine_discoveryengineSearchengineBasicExample_update(context),
@@ -38,7 +38,7 @@ func TestAccDiscoveryEngineSearchEngine_discoveryengineSearchengineBasicExample_
 				ResourceName:            "google_discovery_engine_search_engine.basic",
 				ImportState:             true,
 				ImportStateVerify:       true,
-				ImportStateVerifyIgnore: []string{"engine_id", "collection_id", "location"},
+				ImportStateVerifyIgnore: []string{"engine_id", "collection_id", "location", "kms_key_name"},
 			},
 		},
 	})


melinath:

mmv1/products/discoveryengine/DataConnector.yaml
```diff
@@ -275,3 +273,36 @@ properties:
       'STATE_UNSPECIFIED', 'CREATING', 'ACTIVE', 'FAILED', 'RUNNING', 'WARNING',
       'INITIALIZATION_FAILED', 'UPDATING'.
     output: true
+  - name: 'connectorModes'
+    type: Array
+    description: |
+      The modes enabled for this connector. The possible value can be:
+      'CONNECTOR_MODE_UNSPECIFIED', 'DATA_INGESTION', 'ACTIONS', 'FEDERATED'
```
melinath: Generally we prefer not to document `UNSPECIFIED` values because they're not meaningful for users.

```suggestion
      'DATA_INGESTION', 'ACTIONS', 'FEDERATED'
```

mmv1/products/discoveryengine/DataConnector.yaml
```diff
@@ -275,3 +273,36 @@ properties:
...
+      'EUA', 'FEDERATED_AND_EUA'.
+    item_type:
+      type: String
+  - name: 'syncMode'
+    type: String
+    description: |
+      The data synchronization mode supported by the data connector. The possible value can be:
+      'UNSPECIFIED', 'PERIODIC', 'STREAMING'.
```
melinath:
```suggestion
      'PERIODIC', 'STREAMING'.
```

mmv1/products/discoveryengine/DataConnector.yaml
```diff
@@ -168,9 +163,12 @@ properties:
             method, a DataStore is automatically created for each source entity.
           output: true
         - name: 'params'
-          type: KeyValuePairs
-          description: |
-            The parameters for the entity to facilitate data ingestion.
+          description: 'The parameters for the entity to facilitate data ingestion.'
```
melinath, line 171: Changing from KeyValuePairs to JSON is going to be a [breaking change](https://github.com/GoogleCloudPlatform/magic-modules/pull/15296#issuecomment-3387572394), so we probably can't do this as written until next year's major release. There may be a workaround but I'd recommend just removing this change for now if possible and discussing it separately in a follow-up PR.

kefang2: The reason for the change is because initially we thought that the params field was a flatten struct. The field can actually contain a nested struct value, and that's why we have to change from KeyValuePairs -> JSON

mmv1/products/discoveryengine/DataConnector.yaml
kefang2: Done

mmv1/products/discoveryengine/DataConnector.yaml
kefang2: Done

mmv1/third_party/terraform/services/discoveryengine/resource_discovery_engine_data_connector_test.go
```diff
@@ -1,20 +1,19 @@
...
 import (
+	"testing"
+
 	"github.com/hashicorp/terraform-plugin-testing/helper/resource"
 	"github.com/hashicorp/terraform-provider-google/google/acctest"
-	"testing"
 )
 
-func TestAccDiscoveryEngineDataConnector_discoveryengineDataconnectorJiraBasicExample_update(t *testing.T) {
+func TestAccDiscoveryEngineDataConnector_discoveryengineDataconnectorServicenowBasicExample_update(t *testing.T) {
```
melinath, line 11: could you make this test update `auto_run_disabled` and `incremental_sync_disabled` from true to false just to make sure that works. That should also resolve the missing tests for those fields.

mmv1/products/discoveryengine/SearchEngine.yaml
```diff
@@ -162,3 +162,13 @@ properties:
     type: KeyValuePairs
     description: |
       A map of the feature config for the engine to opt in or opt out of features.
+  - name: 'kmsKeyName'
```
melinath, line 165: This field is currently untested - have you been able to confirm that it works with manual testing / would it be possible to add tests for it? (CmekConfig tests are currently failing, so automated testing of it might not be easy if CmekConfig is involved.) (Generally we bootstrap KMS CryptoKeys for tests: https://googlecloudplatform.github.io/magic-modules/test/test/#cryptokeys)

mmv1/products/discoveryengine/DataConnector.yaml
kefang2: Discussed offline. We will use custom templates to handle data transformation

mmv1/products/discoveryengine/SearchEngine.yaml
kefang2: Added kms key name to the test

mmv1/third_party/terraform/services/discoveryengine/resource_discovery_engine_data_connector_test.go
kefang2: Done

mmv1/third_party/terraform/services/discoveryengine/resource_discovery_engine_search_engine_test.go
```diff
@@ -77,6 +78,7 @@ resource "google_discovery_engine_search_engine" "basic" {
     search_tier = "SEARCH_TIER_ENTERPRISE"
     search_add_ons = ["SEARCH_ADD_ON_LLM"]
   }
+  kms_key_name: "%{kms_key_name}"
```
melinath:
```suggestion
  kms_key_name = "%{kms_key_name}"
```

mmv1/third_party/terraform/services/discoveryengine/resource_discovery_engine_search_engine_test.go
```diff
@@ -118,6 +120,7 @@ resource "google_discovery_engine_search_engine" "basic" {
   features = {
     feedback = "FEATURE_STATE_OFF"
   }
+  kms_key_name: "%{kms_key_name}"
```
melinath:
```suggestion
  kms_key_name = "%{kms_key_name}"
```

melinath:
Manual test run for the test skipped in VCR mode: https://hashicorp.teamcity.com/buildConfiguration/TerraformProviders_GoogleCloud_GOOGLE_BETA_MMUPSTREAMTESTS_GOOGLEBETA_PACKAGE_DISCOVERYENGINE/495381

mmv1/third_party/terraform/services/discoveryengine/resource_discovery_engine_data_connector_test.go
```diff
@@ -1,20 +1,19 @@
...
 	// Skips this update test due to duration and flakiness.
 	t.Skip()
```
melinath, line 13: oh, this isn't just skipped in VCR - it's always skipped. Does this still need to be the case?

mmv1/templates/terraform/examples/discoveryengine_dataconnector_servicenow_basic.tf.tmpl
```diff
@@ -0,0 +1,50 @@
...
+    client_secret              = "SECRET_MANAGER_RESOURCE_NAME"
+    static_ip_enabled          = "false"
+    user_account               = "connectorsuserqa@google.com"
+    password                   = "SECRET_MANAGER_RESOURCE_NAME"
+  }
+  refresh_interval             = "86400s"
+  incremental_refresh_interval = "21600s"
+  entities {
+    entity_name                = "catalog"
+    params                     = {
```
melinath: is there a simple string -> string param that you could add here to make sure that works as expected too?

mmv1/third_party/terraform/services/discoveryengine/resource_discovery_engine_data_connector_test.go
kefang2: Yes, the update test are flaky because updates are not allow until the connector is initialized. See:
https://github.com/GoogleCloudPlatform/magic-modules/pull/14890#issuecomment-3321663930
https://github.com/GoogleCloudPlatform/magic-modules/pull/14890#pullrequestreview-3259648999

mmv1/products/discoveryengine/SearchEngine.yaml
melinath, line 165: looks like this isn't returned in the API response.

```suggestion
  - name: 'kmsKeyName'
    ignore_read: true
```

mmv1/third_party/terraform/services/discoveryengine/resource_discovery_engine_data_connector_test.go
melinath: It could make sense to unskip this at some point and add a time_sleep but I'm fine to keep it as-is for now.

mmv1/templates/terraform/examples/discoveryengine_dataconnector_servicenow_basic.tf.tmpl
melinath: never mind

mmv1/products/discoveryengine/DataConnector.yaml
```diff
@@ -169,8 +164,9 @@ properties:
           output: true
         - name: 'params'
           type: KeyValuePairs
-          description: |
-            The parameters for the entity to facilitate data ingestion.
+          description: 'The parameters for the entity to facilitate data ingestion.'
+          custom_flatten: 'templates/terraform/custom_flatten/discoveryengine_data_connector_params.go.tmpl'
+          custom_expand: 'templates/terraform/custom_expand/discoveryengine_data_connector_params.go.tmpl'
```
melinath: Sorry for the back and forth on this but we'll likely want to go forward with conversion to a json field after all since usage of this field requires being a JSON field - I just need to double-check the upgrade paths.

```suggestion
          custom_flatten: 'templates/terraform/custom_flatten/json_schema.tmpl'
          custom_expand: 'templates/terraform/custom_expand/json_schema.tmpl'
          state_func: 'func(v interface{}) string { s, _ := structure.NormalizeJsonString(v); return s }'
          validation:
            function: 'validation.StringIsJSON'
```

mmv1/products/discoveryengine/DataConnector.yaml
```diff
@@ -169,8 +164,9 @@ properties:
           output: true
         - name: 'params'
```
melinath, line 170: I ran into a problem where `entities` is immutable but the children aren't - due to a [bug in the generator](https://github.com/hashicorp/terraform-provider-google/issues/23893), this results in the children not being possible to change.

Is there any chance the children of `entities` should be marked immutable as well?

mmv1/products/discoveryengine/DataConnector.yaml
kefang2: Done

mmv1/products/discoveryengine/DataConnector.yaml
melinath: (Technically this could wait for a separate PR but I'd be happy to address it here if it's that simple.)

kefang2: The param is mutable, what is our option here? I am not sure why it was marked as immutable initially

mmv1/products/discoveryengine/DataConnector.yaml
```diff
@@ -136,19 +132,20 @@ properties:
     type: Array
     description: |
       List of entities from the connected data source to ingest.
-    immutable: true
     item_type:
```
melinath, line 140: entities should also stay immutable for now.

```suggestion
    immutable: true
    item_type:
```

mmv1/products/discoveryengine/DataConnector.yaml
melinath: We ended up deciding to leave this alone for this PR & may revisit separately later.

melinath:
we can ignore the unused-template-check failure - it's not relevant for state migrations.

melinath:
The TestAccDiscoveryEngineCmekConfig* failures are unrelated to this PR.

PR #15297, opened 2025-10-02
Modify override merging to ignore blocks with an incompatible version

roaks3:
This changes how the override merging works for mmv1 generation.

Previously, mmv1 yaml objects were (mostly) overridden naively, with the set of "override" fields essentially replacing the original set of fields. The problem with this is that overrides can then only change behavior through mmv1 mechanisms.

This PR changes the merge behavior to ignore (or prune) entire "override" objects if they target a version that is not compatible with the provider version being built. This allows overrides to exclude behavior with a higher degree of control by using `min_version` tags.

See https://cloud-internal-review.git.corp.google.com/c/cloud-graphite-eng/magic-modules-private-overrides/+/53590 for how this change impacts EAP (no diff).

```release-note:none

```

### DIFF
--- a/mmv1/api/product.go
+++ b/mmv1/api/product.go
@@ -282,9 +282,20 @@ func (p Product) Lineage() string {
 	return p.Name
 }
 
-func Merge(self, otherObj reflect.Value) {
-
+func Merge(self, otherObj reflect.Value, version string) {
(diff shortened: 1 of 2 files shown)


roaks3:

slevenick (approved):

PR #15300, opened 2025-10-03
Support for Tags Update and Delete in google_memcache_instance resource 

SamanthaMathews:
Modifying tags field to enable update and Delete in Memcache Instance resource.
Part of b/445884840

```release-note:enhancement
memcache: modified Modifying field to enable update and Delete in Memcache Instance resource
```

### DIFF
--- a/mmv1/third_party/terraform/services/memcache/resource_memcache_instance_test.go
+++ b/mmv1/third_party/terraform/services/memcache/resource_memcache_instance_test.go
@@ -5,7 +5,12 @@ import (
 	"testing"
 
 	"github.com/hashicorp/terraform-plugin-testing/helper/resource"
+	"github.com/hashicorp/terraform-plugin-testing/terraform"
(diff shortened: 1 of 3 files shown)


mmv1/third_party/terraform/services/memcache/resource_memcache_instance_test.go
```diff
@@ -97,3 +102,412 @@ data "google_compute_network" "memcache_network" {
...
+  labels = {
+    env = "test"
+  }
+  tags = {
+    "%{org}/%{tagKey}" = "%{tagValue}"
+  }
+}`, context)
+}
+
+func TestAccMemcacheInstance_tagsUpdateDelete(t *testing.T) {
```
abd-goog, line 268: In its present form this function is just checking if we can create/delete bindings for a resource via location tag bindings resource. I don't think that's needed. We are supposed to test tagging operations via the newly added 'tags' field instead.

PR #15306, opened 2025-10-03
WIP Auto generate import

trodge:
fixes https://github.com/hashicorp/terraform-provider-google/issues/8712

```release-note:none

```


trodge:
*IamGenerated test failures are definitely related:
```
  error=
  | exit status 1
  | 
  | Error: Cannot import non-existent remote object
  | 
  | While attempting to import an existing object to
  | "google_iap_web_type_app_engine_iam_policy.foo", the provider detected that
  | no object exists with the given id. Only pre-existing objects can be
  | imported; check that the id is correct and that it is associated with the
  | provider's configured region or endpoint, or use "terraform apply" to create
  | a new remote object for this resource.
  ```
  Seems like the resource name parameter is not correct, but I'm not sure where to get the correct one from based on @melinath 's [comment](https://github.com/hashicorp/terraform-provider-google/issues/8712#issuecomment-3353523402)

File: mmv1/templates/terraform/examples/base_configs/iam_test_file.go.tmpl [hunk 1]
```diff
@@ -362,7 +362,7 @@ func TestAcc{{ $.ResourceName }}IamPolicyGenerated_withCondition(t *testing.T) {
 {{- if not $.IamPolicy.ExcludeImportTest }}
 			{
 				ResourceName:      "{{ $.IamTerraformName }}_policy.foo",
-				ImportStateId:     fmt.Sprintf("{{ $.IamImportFormat }}", {{ if ne $.IamImportQualifiersForTest "" }}{{ $.IamImportQualifiersForTest }}, {{ end }}{{ $example.PrimaryResourceName }}),
+				ImportStateIdFunc: generate{{ $.ResourceName }}IAMPolicyStateID("foo"),
```
melinath: For the ImportStateIdFunc lines, I'd recommend passing the full resource name so that it's easier to see this is referencing the same thing as the line above.

```suggestion
				ImportStateIdFunc: generate{{ $.ResourceName }}IAMPolicyStateID("{{ $.IamTerraformName }}_policy.foo"),
```

shuyama1:
/gcbrun 

VCR issue should be addressed now. Sorry for the disruption!

melinath:
I've manually canceled the builds for the older new commits for clarity.

mmv1/api/resource/iam_policy.go
```diff
@@ -82,7 +81,10 @@ type IamPolicy struct {
 
 	// Resource name may need a custom diff suppress function. Default is to use
 	// CompareSelfLinkOrResourceName
-	CustomDiffSuppress *string `yaml:"custom_diff_suppress,omitempty"`
+	CustomDiffSuppress *string `yaml:"custom_diff_suppress"`
+
+	// StateIDFuncs may use a custom template if default funcs don't work.
+	CustomStateIDFuncs string `yaml:"custom_state_id_funcs"`
```
melinath: I'm a little concerned about this toggle, for two reasons:

1. The name is pretty close to [StateFunc](https://github.com/hashicorp/terraform-plugin-sdk/blob/8bd650e03a438b9e1a08fb97fea5bce0cd7ae529/helper/schema/schema.go#L215-L219) ([mmv1](https://github.com/GoogleCloudPlatform/magic-modules/blob/f3186132d8e380eb71ba3b447f5ef4624262b071/mmv1/api/type.go#L178)). Since both this and StateFunc are rarely used, I think it'll be confusing that this behavior is doing something very different. The name needs to very clearly identify how the function will be used.
2. This feels like a problem that we should be solving somewhere else. I'm not exactly sure where, though. In both cases the state id func seems to be paired with a diff suppress func - is that related, or just coincidence? Could we replace the diff suppress with a flattener to solve this? (This point isn't a blocker if it's too complicated to resolve now; we'd need to create a ticket to look at removing this toggle later, but I think it would still be a win overall since the need for overrides is rare and the pain is common. However, if we leave this toggle in place, we'd want to explicitly validate that we don't want to add new usage of this.)

(I'm actually surprised that (given it's possible for this kind of case to happen) there are only two exceptions!)

mmv1/templates/terraform/iam/iap_web_appengine_state_id_funcs.go.tmpl
```diff
@@ -0,0 +1,58 @@
...
+		for _, m := range state.Modules {
+			if len(m.Resources) > 0 {
+				if v, ok := m.Resources[iamResourceAddr]; ok {
+					rawState = v.Primary.Attributes
+				}
+			}
+		}
+		fmt.Printf("raw state %s\n", rawState)
+		project := tpgresource.GetResourceNameFromSelfLink(rawState["project"])
+		appId := appIdShortName(rawState["app_id"])
```
melinath, line 47: what would you think about (if we stick with custom code override) overriding this logic specifically instead of the whole function, since the rest of the function is standard?

mmv1/third_party/terraform/acctest/resource_test_utils.go
```diff
@@ -111,6 +111,14 @@ func TestCheckAttributeValuesEqual(i *string, j *string) resource.TestCheckFunc
 	}
 }
 
+// ConditionTitleIfPresent returns empty string if condition is not preset and " {condition.0.title}" if it is.
+func ConditionTitleIfPresent(state map[string]string) string {
+	if conditionTitle, ok := state["condition.0.title"]; ok {
+		return " " + conditionTitle
+	}
+	return ""
+}
```
melinath: This function isn't very clear in terms of its usage. That is, I can't tell from `acctest.ConditionTitleIfPresent` what the purpose of the function is unless I already know. This is also operating on a rawState map but that's not clear from context either. I'd recommend restructuring how the function works to be more specific. For example, something like:


```suggestion
func BuildIAMImportId(name, role, member, condition string) string {
    ret := name
    if role != "" {
        ret += " " + role
    }
    if member != "" {
        ret += " " + member
    }
    if condition != "" {
        ret += " " + condition
    }
    return ret
}
```

mmv1/templates/terraform/iam/iap_web_appengine_state_id_funcs.go.tmpl
trodge: It's not easy to override only this logic because this code is generated by a range statement. The yaml would need to set the field name as well as the template.

mmv1/api/resource/iam_policy.go
trodge: I think the names won't be confused because this field is nested under iam policy. I can't think of a more accurate name for this override.

I'm not sure if the diff suppress funcs are a coincidence or not. I think replacing them with flatteners has the potential to break resources.

mmv1/templates/terraform/iam/iap_web_appengine_state_id_funcs.go.tmpl
melinath: I think we could do this by overriding a larger section (not just this line) but I don't care that much. Happy to leave as-is.

PR #15311, opened 2025-10-03
Remove output_only restriction on OnPremVersion field

chethangowda89:

```release-note:enhancement
gkeonprem: made it possible to set the `on_prem_version` field on `google_gkeonprem_vmware_node_pool` (previously output-only)
```

### DIFF
--- a/mmv1/products/gkeonprem/VmwareNodePool.yaml
+++ b/mmv1/products/gkeonprem/VmwareNodePool.yaml
@@ -298,4 +298,3 @@ properties:
     type: String
     description: |
       Anthos version for the node pool. Defaults to the user cluster version.
-    output: true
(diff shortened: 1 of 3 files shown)


chethangowda89:
> This LGTM - old tests without on_prem_version set continue to work, and this test also passes. Just to double-check, if you set on_prem_version and then _remove_ on_prem_version in a subsequent apply, Terraform will omit it from the API request to update the resource. Will the API properly unset the field?

If we do not set then the field operation defaults to cluster's version, that is the intended behavior.

melinath:
> If we do not set then the field operation defaults to cluster's version, that is the intended behavior.

yes, that's good! But if the field isn't sent in the API request, will that unset the value in the API? Or does it need to be sent as some empty value?

chethangowda89:
> > If we do not set then the field operation defaults to cluster's version, that is the intended behavior.
> 
> yes, that's good! But if the field isn't sent in the API request, will that unset the value in the API? Or does it need to be sent as some empty value?

The platform controller is the one which defaults to cluster version. During update the clh will merge pointers from the request object with that of platform CR for all the fields that are specified in update mask. nil or empty will be defaulted to clusters version by controller during reconciliation

melinath:
while you're there could you also add a ConfigPlanChecks as described in https://googlecloudplatform.github.io/magic-modules/test/test/#add-an-update-test?

melinath (approved):
LGTM, thanks!

PR #15317, opened 2025-10-06
Add a new 'data_classification' field on 'dataplex_aspect_type' resource.

mll:
Add a new 'data_classification' field on 'dataplex_aspect_type' resource.

```release-note:enhancement
compute: added `data_classification` field to `google_dataplex_aspect_type` resource
```


roaks3 (commented):
To clarify the intent: you want `DATA_CLASSIFICATION_UNSPECIFIED` to be an explicit value that the user can specify, as opposed to just not including this field?

And per the failing test, it looks like the `dataClassification` is sent to the server, but not returned in the response. Is this how the API is supposed to work?

mll:
Hi @roaks3 , thanks for the review. Relating to your questions:

I want the user to be able to either not specify the field at all (then defaults to DATA_CLASSIFICATION_UNSPECIFIED in our API) or specify it as either DATA_CLASSIFICATION_UNSPECIFIED or METADATA_AND_DATA.

The test failed because our API was not visible from outside Google yet. Now we have changed the visibility. Is there any way to re-trigger tests without submitting an additional commit?

mll:
@roaks3 Do you know why terraform-google-conversion-test-integration failed?

mmv1/third_party/terraform/services/dataplex/resource_dataplex_aspect_type_test.go
```diff
@@ -51,6 +51,7 @@ resource "google_dataplex_aspect_type" "test_aspect_type" {
   project = "%{project_name}"
   location = "us-central1"
 
+  data_classification = "DATA_CLASSIFICATION_UNSPECIFIED"
```
roaks3: Nit: I would just omit this line and the one below, to verify the "unset" behavior in an aspect type test, and because there is no behavior being tested beyond what is already covered in the example tests.

PR #15319, opened 2025-10-06
fw provider plural datasource template

NickElliot:
Was not sure based on the current template state if recreating multiple nested resource models was still necessary, but this template can effectively copy the remaining resource.go.tmpl implementation + 1 layer of nested, with the most important plural specific pieces already done

```release-note:none

```

### DIFF
--- a/mmv1/templates/terraform/datasource_plural_fw.go.tmpl
+++ b/mmv1/templates/terraform/datasource_plural_fw.go.tmpl
@@ -1,0 +1,284 @@
+{{/* The license inside this block applies to this file
+  Copyright 2025 Google LLC. All Rights Reserved.
+
+  Licensed under the Apache License, Version 2.0 (the "License");
+  you may not use this file except in compliance with the License.
+  You may obtain a copy of the License at
+
+      http://www.apache.org/licenses/LICENSE-2.0
+
+  Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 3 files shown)


mmv1/templates/terraform/datasource_plural_fw.go.tmpl
```diff
@@ -0,0 +1,284 @@
...
+
+    log.Printf("[DEBUG] Refreshing {{ $.Name -}} data: %s", data.Id.ValueString())
+
+    headers := make(http.Header)
+{{- if $.CustomCode.PreRead }}
+    {{ $.CustomTemplate $.CustomCode.PreRead false -}}
+{{- end}}
+    res, err := fwtransport.SendRequest(fwtransport.SendRequestOptions{
+        Config: r.providerConfig,
+        Method: "LIST",
```
BBBmau, line 252: i'm not sure if this would actually work in cases where the method isn't `LIST`

[`compute_disks` uses `GET`](https://cloud.google.com/compute/docs/reference/rest/v1/disks/list) instead just like all other data-sources, unless there's something else that i'm missing?

fwiw i was working on standardizing how we handle pagination within plural data-sources in this PR:
- https://github.com/GoogleCloudPlatform/magic-modules/pull/14687

though it's been a while since i've touched it due to other things coming up

PR #15320, opened 2025-10-06
Add terraform support for EntryLinks

Sachin-Rungta:

```release-note:new-resource
`google_dataplex_entry_links`
```

### DIFF
--- a/mmv1/third_party/terraform/services/dataplex/resource_dataplex_entry_link_test.go
+++ b/mmv1/third_party/terraform/services/dataplex/resource_dataplex_entry_link_test.go
@@ -77,7 +77,7 @@ resource "google_dataplex_entry_link" "basic_entry_link" {
 	type = "SOURCE"
   }
   entry_references {
-    name = "projects/${google_dataplex_entry_group.entry-group-basic.project}/locations/us-central1/entryGroups/@dataplex/entries/${google_dataplex_glossary_term.term_test_id_full.name}"
+    name = "projects/${google_dataplex_entry_group.entry-group-basic.project}/locations/us-central1/entryGroups/@dataplex/entries/projects/${google_dataplex_entry_group.entry-group-basic.project}/locations/us-central1/glossaries/${google_dataplex_glossary.term_test_id_full.glossary_id}/terms/${google_dataplex_glossary_term.term_test_id_full.term_id}"
 	type = "TARGET"
   }
 }


Sachin-Rungta:
@melinath When I am passing "UNSPECIFIED" as a entry type in test, it is getting failed with some diff issue. Is there a way to ignore diff ?

melinath:
Generally we recommend not supporting `UNSPECIFIED` as a value in Terraform, since it doesn't usually have a special meaning (that is, different from omitting the field.) Can you omit `UNSPECIFIED` as an option and set it to some other value?

melinath:
In this case that would also mean making the field optional so that it can be omitted.

Sachin-Rungta:
> In this case that would also mean making the field optional so that it can be omitted.

I have some doubts
1. In Java code if I don't pass anything in type it will automatically consider it as unspecified as that is a required field and Default set to 0 for enum.
2. In terraform, will this be a issue if I make the type optional in YAML although the proto definition is expecting to be present ?
3. I couldn't understand why this job is stuck and terraform-google-conversion-test-integration is failing.
Thanks

melinath:
So, the reason for the diff is that when we send `UNSPECIFIED` to the API, it's being normalized and returned as the field being omitted. You can see this in the [debug log](https://ff82c06a9dab761f6a8b14d4a258b91d28f3d79dad77505e831d20b-apidata.googleusercontent.com/download/storage/v1/b/ci-vcr-logs/o/beta%2Frefs%2Fheads%2Fauto-pr-15320%2Fartifacts%2F1dc7ece0-c227-4ddc-9f74-ea18f8a150fa%2Frecording%2FTestAccDataplexEntryLink_dataplexEntryLinkBasicExample.log?jk=AUDkFIm_azM297GilEZm9nKaZFSCUvXsz9nnkYWPhC5_1lrTYxwFpGnwXwZZUDGBJCZ1PtdFwSqofMx7Iqayg8MHnXBTPycxg2aZyK_BGHU920d6vH-9iuOP6vJb_Fi1EczGxRW-Z0sPhgbaOJqwfWN3WtHPuRcTfuepCJDybJTQZ60dK9UpAP9TG7w83nyFrELJ4GV3bHB4f2MZvvJ4zjlpO9gGbtHRtIGJruFqbVvlZmQnDrJiOCCvXa_0WZHHisgiuZPMcoKQUGi57wvTNqguhgB-souNYglPp8BQOKENrwwIxQmHLZ8attp1G2MvrC3UZjbLhFA8hKEtYQGCYmkKSYWrz7vbI4uY5kOcEV-v5cg9JQZ1XNWGXWs7FbbZLXv1w-Tek2vNKOO-b9b88Ceg6NvUECiYlzQw_Nu9LOrE0dYJutlGXjENmRMk8u-0b0VwZ3xIOcfeu0TnvRwyz64AITaltohekAQfTiE54kvbJ5ZMYpR26z8Xhf6M1f0_tFgWrJpUKkEOOxIBiBNvPo6L_iUCW8__Me4l7iEb8h3hNFZmtj5qE-pPkAwJBDrve9q7rvAxNy_B6x5ioDhfC-Untcw8-DgUbQTUepw6wufuW5vWQusu8C9CeEpFNT4G8Y2kxFavlU1PI3cJCKH-ROW8G3u-odoHcb7bTssVg18_rxarCTSJ1RoMIcECg3BAx8NwtvVq7ep7WrCVTbh7cNAbb_2b8pzxWduv_PAa770gt8xRg3e4s_rvXO6tZXCBTTNQSa65hcsE3J5zqBK9-1KmswhMopkJb-jcY_GxHt2_5_-obD-IVjjKAT1DnOJVmqEfER3A3RZTxXmwtra7c5pAu9ie1pgZIXrYlCw9gdaW8jaRbm04F6OEMVFobttHfojxk22UtWkIOAQLmQEx7_C9BvFLx0oygVA-yxqBXeuXj7Odtq04Rc2mbdkLquXR5nw0YdybwXe7cuHgX6wU_1wew-H6jxPffRjAMxe1uP7m-Vx_9FRrMCW4sHb1CaWXwylPlRr833pMg8qeetQEUT1U0EtndgRgH9eQpytmslHANjYU0AMx2SsDs9-NVXXYcn6N9-547zusM6pZvNvEqthx22v4d-0D3fKC5yycvIBp5evU-A5S5r9BH5W8JZfVdC_8eOM-iQySNs3wr9SaekCaE8dNUnckxsepNvdTXOavHBiwKKu58Ct94FTj5L_Kj76oxYTp6rlVAw-xne3evPNa6xF3-FGJDCFPVglwO-P3k7xugP8jq0G96sNZL9CzI0aCz1AScR_LOEEyVwf2WGTJdEbV3ajGZ6lp6nyCNCfo9aYSZ1u4Ru4LZSJwa9hYydVSR9p_DeDPnfaO5m0dkBi4&isca=1). I'd expect that if we omit the field in the API request, it should be treated as "UNSPECIFIED" by the API. That's generally how GCP APIs behave. It may technically be "required" in the proto, but that just means that the API needs to set the default value if it's not provided by the user. Please try making the field optional to see if it works!

Regarding terraform-google-conversion-test-integration, it looks like there was a failure due to TestAccNetworkSecuritySecurityProfile_networkSecuritySecurityProfileUrlFilteringExample, which isn't related to your change. (Attn @zli82016 FYI)

Sachin-Rungta:
@melinath I think omit logic worked but there is some issue with the import format, although I passed standard format. Could you please help in this ?

zli82016:
The tgc integration test has been fixed in https://github.com/GoogleCloudPlatform/magic-modules/pull/15329 and merged into the main branch. Thanks Stephen for the reminder.

mmv1/products/dataplex/EntryLink.yaml
```diff
@@ -0,0 +1,124 @@
...
+    required: true
+    immutable: true
+properties:
+  - name: 'name'
+    type: String
+    description: |
+      The relative resource name of the Entry Link, of the form:
+      projects/{project_id_or_number}/locations/{location_id}/entryGroups/{entry_group_id}/entryLinks/{entry_link_id}
+    output: true
+    immutable: true
```
melinath: output-only fields can't be immutable.

```suggestion
```

mmv1/products/dataplex/EntryLink.yaml
```diff
@@ -0,0 +1,124 @@
...
+            The path in the Entry that is referenced in the Entry Link.
+            Empty path denotes that the Entry itself is referenced in the Entry Link.
+          immutable: true
+        - name: 'type'
+          type: Enum
+          description: |
+            The reference type of the Entry.
+          immutable: true
+          enum_values:
+            - 'UNSPECIFIED'
```
melinath: We generally don't list/allow "unspecified" unless it has a special meaning separate from "unset"

```suggestion
```

mmv1/products/dataplex/EntryLink.yaml
```diff
@@ -0,0 +1,124 @@
...
+    output: true
+  - name: 'updateTime'
+    type: Time
+    description: |
+      The time when the Entry Link was last updated.
+    output: true
+  - name: 'entryReferences'
+    type: Array
+    description: |
+      Specifies the Entries referenced in the Entry Link. There should be exactly two entry references.
```
melinath, line 91: Is the "exactly two" limitation likely to always be the case going forward? If so, you could add client-side validation here: 
```suggestion
    description: |
      Specifies the Entries referenced in the Entry Link. There should be exactly two entry references.
    min_size: 2
    max_size: 2
```

(The drawback with client-side validation is just that it will mean that the resource isn't future-compatible if the constraints loosen over time.)

mmv1/products/dataplex/EntryLink.yaml
```diff
@@ -0,0 +1,124 @@
...
+references:
+  guides:
+    'Official Documentation': 'https://cloud.google.com/dataplex/docs' # Update with specific EntryLink docs when available
+  api: 'https://cloud.google.com/dataplex/docs/reference/rest/v1/projects.locations.entryGroups.entryLinks'
+docs:
+base_url: 'projects/{{project}}/locations/{{location}}/entryGroups/{{entry_group_id}}/entryLinks/{{entry_link_id}}'
+self_link: 'projects/{{project}}/locations/{{location}}/entryGroups/{{entry_group_id}}/entryLinks/{{entry_link_id}}'
+create_url: 'projects/{{project}}/locations/{{location}}/entryGroups/{{entry_group_id}}/entryLinks?entryLinkId={{entry_link_id}}'
+update_verb: 'PATCH'
+update_mask: true
```
melinath: It looks like this resource doesn't actually support update at all - rather than marking each field as immutable, the whole resource should be immutable.

```suggestion
immutable: true
```

mmv1/products/dataplex/EntryLink.yaml
```diff
@@ -0,0 +1,124 @@
...
+import_format:
+  - 'projects/{{project}}/locations/{{location}}/entryGroups/{{entry_group_id}}/entryLinks/{{entry_link_id}}'
+timeouts:
+  insert_minutes: 20
+  update_minutes: 20
```
melinath:
```suggestion
```

mmv1/products/dataplex/EntryLink.yaml
```diff
@@ -0,0 +1,124 @@
...
+    test_env_vars:
+      project_number: 'PROJECT_NUMBER'
+parameters:
+  - name: 'entryGroupId'
+    type: String
+    description: |
+      The id of the entry group this entry link is in.
+    url_param_only: true
+    required: true
+    immutable: true
```
melinath: immutable can be removed on all fields

mmv1/products/dataplex/EntryLink.yaml
Sachin-Rungta: I am thinking to skip validation as scope can be increase in future.

melinath: sounds good!

mmv1/products/dataplex/EntryLink.yaml
```diff
@@ -0,0 +1,117 @@
...
+    item_type:
+      type: NestedObject
+      properties:
+        - name: 'name'
+          type: String
+          description: |
+            The relative resource name of the referenced Entry, of the form:
+            projects/{project_id_or_number}/locations/{location_id}/entryGroups/{entry_group_id}/entries/{entry_id}
+          required: true
+          immutable: true
```
melinath:
```suggestion
```

mmv1/products/dataplex/EntryLink.yaml
```diff
@@ -0,0 +1,117 @@
...
+        - name: 'path'
+          type: String
+          description: |
+            The path in the Entry that is referenced in the Entry Link.
+            Empty path denotes that the Entry itself is referenced in the Entry Link.
+          immutable: true
```
melinath:
```suggestion
```

mmv1/products/dataplex/EntryLink.yaml
```diff
@@ -0,0 +1,117 @@
...
+        - name: 'type'
+          type: Enum
+          description: |
+            The reference type of the Entry.
+          immutable: true
```
melinath:
```suggestion
```

mmv1/third_party/terraform/services/dataplex/resource_dataplex_entry_link_test.go
```diff
@@ -0,0 +1,77 @@
+package dataplex_test
```
melinath, line 1: This file can be deleted - no need for an update test since the resource can't be updated. (Unless this is testing something different than the generated tests?)

mmv1/products/dataplex/EntryLink.yaml
```diff
@@ -0,0 +1,117 @@
...
+      The relative resource name of the Entry Link, of the form:
+      projects/{project_id_or_number}/locations/{location_id}/entryGroups/{entry_group_id}/entryLinks/{entry_link_id}
+    output: true
+  - name: 'entryLinkType'
+    type: String
+    description: |
+      Relative resource name of the Entry Link Type used to create this Entry Link. For example:
+      projects/dataplex-types/locations/global/entryLinkTypes/definition
+    required: true
+    immutable: true
```
melinath:
```suggestion
```

Sachin-Rungta:
> could you add usage of `entry_references.type` to a test? That will resolve the missing test: [#15320 (comment)](https://github.com/GoogleCloudPlatform/magic-modules/pull/15320#issuecomment-3395231068)

That will be unspecified for most of the case where entryGroups is not `@dataplex` . Due to which we are skipping it.

melinath:
> That will be unspecified for most of the case where entryGroups is not `@dataplex` .

It probably works, since you'd tested it with `UNSPECIFIED` before... I'm just concerned we won't have automated coverage of it. Would it be possible to contrive a test where it has some other value, or is that difficult for some reason?

mmv1/templates/terraform/examples/dataplex_entry_link_basic.tf.tmpl
```diff
@@ -0,0 +1,44 @@
...
+  project = "{{index $.TestEnvVars "project_number"}}"
+}
+
+resource "google_dataplex_entry_link" "{{$.PrimaryResourceId}}" {
+  project = "{{index $.TestEnvVars "project_number"}}"
+  location = "us-central1"
+  entry_group_id = google_dataplex_entry_group.entry-group-basic.entry_group_id
+  entry_link_id = "tf-test-entry-link%{random_suffix}"
+
+  entry_link_type = "projects/655216118709/locations/global/entryLinkTypes/related"
```
melinath, line 35: This is currently hardcoding a project number - could it use the one from TestEnvVars instead?

mmv1/templates/terraform/examples/dataplex_entry_link_full.tf.tmpl
```diff
@@ -0,0 +1,45 @@
...
+  project = "{{index $.TestEnvVars "project_number"}}"
+}
+
+resource "google_dataplex_entry_link" "{{$.PrimaryResourceId}}" {
+  project = "{{index $.TestEnvVars "project_number"}}"
+  location = "us-central1"
+  entry_group_id = google_dataplex_entry_group.entry-group-full.entry_group_id
+  entry_link_id = "tf-test-entry-link-full%{random_suffix}"
+
+  entry_link_type = "projects/655216118709/locations/global/entryLinkTypes/related"
```
melinath, line 35: same here

mmv1/third_party/terraform/services/dataplex/resource_dataplex_entry_link_test.go
```diff
@@ -0,0 +1,85 @@
...
+  labels = { "tag": "test-tf" }
+  display_name = "terraform term"
+  description = "term created by Terraform"
+}
+resource "google_dataplex_entry_link" "basic_entry_link" {
+  project = "%{project_number}"
+  location = "us-central1"
+  entry_group_id = google_dataplex_entry_group.entry-group-basic.entry_group_id
+  entry_link_id = "tf-test-entry-link%{random_suffix}"
+  entry_link_type = "projects/655216118709/locations/global/entryLinkTypes/definition"
```
melinath, line 74: same here

melinath (changes requested):

mmv1/templates/terraform/examples/dataplex_entry_link_basic.tf.tmpl
Sachin-Rungta: We need this project number to be hardcoded as these are 1P types project number and same has to be used by user also.

mmv1/templates/terraform/examples/dataplex_entry_link_full.tf.tmpl
Sachin-Rungta: Same as per above comments.

PR #15331, opened 2025-10-07
Fix google_bigquery_analytics_hub_listing_subscription import

wj-chen:
Fixes https://github.com/hashicorp/terraform-provider-google/issues/24507.

There are two changes:

1. Remove `ignore_read: true` from `destination_dataset`; the API is returning the field in response
2. Add a `post_read` to set `data_exchange_id` and `listing_id`; these are `url_params_only` properties and the API response instead has `listing` which is an URI containing these IDs, during the `Read()` after `Import()`, these properties need to be set explicitly by parsing `listing`, otherwise their values in the state file would be `null` which would cause a force-replacement for a `plan` after `import`.

```release-note:bug
bigqueryanalyticshub: fixed `google_bigquery_analytics_hub_listing_subscription` import
```


c2thorn:
waiting for @sachinpro

mmv1/templates/terraform/post_read/bigqueryanalyticshub_listing_subscription.go.tmpl
```diff
@@ -0,0 +1,12 @@
+// Set data_exchange_id and listing_id from res["listing"]
+listing := res["listing"].(string)
+parts := strings.Split(listing, "/")
+if len(parts) != 8 {
+	return fmt.Errorf("Listing name %q is not in the expected format projects/*/locations/*/dataExchanges/*/listings/*", listing)
+}
+if err := d.Set("data_exchange_id", parts[5]); err != nil {
+  return fmt.Errorf("Error reading ListingSubscription: %s", err)
```
sachinpro, line 8: nit(optional): it would be more verbose if we have something like: "Error setting data_exchange_id".

mmv1/templates/terraform/post_read/bigqueryanalyticshub_listing_subscription.go.tmpl
```diff
@@ -0,0 +1,12 @@
...
+}
+if err := d.Set("listing_id", parts[7]); err != nil {
+  return fmt.Errorf("Error reading ListingSubscription: %s", err)
```
sachinpro, line 11: nit(optional): it would be more verbose if we have something like: "Error setting listing_id"

mmv1/templates/terraform/post_read/bigqueryanalyticshub_listing_subscription.go.tmpl
wj-chen: Thanks, doing it this way to align with the generated code when setting other properties: https://github.com/hashicorp/terraform-provider-google/blob/332992365cbe621adc44f10821319fad79d85a33/google/services/bigqueryanalyticshub/resource_bigquery_analytics_hub_listing_subscription.go#L399

PR #15338, opened 2025-10-08
added GKE maintenance exclusion option EndTimeBehavior.

yliaog:
added GKE maintenance exclusion option EndTimeBehavior.

```release-note:feature
added GKE maintenance exclusion option EndTimeBehavior.
```

### DIFF
--- a/mmv1/third_party/terraform/services/container/resource_container_cluster.go.tmpl
+++ b/mmv1/third_party/terraform/services/container/resource_container_cluster.go.tmpl
@@ -5668,8 +5668,9 @@ func expandMaintenancePolicy(d *schema.ResourceData, meta interface{}) *containe
 				meo := exclusionOptions.([]interface{})[0].(map[string]interface{})
 				mex := exclusions[exclusion["exclusion_name"].(string)]
 				mex.MaintenanceExclusionOptions = &container.MaintenanceExclusionOptions{
-					Scope:           meo["scope"].(string),
-					ForceSendFields: []string{"Scope"},
+					Scope:             meo["scope"].(string),
+					EndTimeBehavior:   meo["end_time_behavior"].(string),
+					ForceSendFields:   []string{"Scope", "EndTimeBehavior"},
 				}
 				exclusions[exclusion["exclusion_name"].(string)] = mex
 			}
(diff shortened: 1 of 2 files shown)


yliaog:
the api is in the process of being rolled out, let's wait for the rollout to complete in a couple of weeks.

yliaog:
@slevenick the API has been rolled out, could you please help review this PR?

slevenick (changes requested):
You also need to add this to the schema for the resource. You've only added it to the flatten/expander, but it won't be specifiable without being in the schema

mmv1/third_party/terraform/services/container/resource_container_cluster.go.tmpl
```diff
@@ -5731,8 +5674,9 @@ func expandMaintenancePolicy(d *schema.ResourceData, meta interface{}) *containe
 				meo := exclusionOptions.([]interface{})[0].(map[string]interface{})
 				mex := exclusions[exclusion["exclusion_name"].(string)]
 				mex.MaintenanceExclusionOptions = &container.MaintenanceExclusionOptions{
-					Scope:           meo["scope"].(string),
-					ForceSendFields: []string{"Scope"},
+					Scope:             meo["scope"].(string),
+					EndTimeBehavior:   meo["end_time_behavior"].(string),
+					ForceSendFields:   []string{"Scope", "EndTimeBehavior"},
```
yliaog: EndTimeBehavior is Optional, does it need to be in ForceSendFields: ?

yliaog: @slevenick

slevenick: It depends on the API's behavior for this field.

There seems to only be one valid value for the field, is it possible to unset that? Would that make sense?

yliaog: yes, the field has only one valid value, and it is possible to unset it. the behavior is: it is either unset, or set with that one single valid value.

for this case, does it need to be in ForceSendFields: ? i don't really know what does ForceSendFields: do, i guess that's also my question.

slevenick (commented):
Test failing with:     resource_container_cluster_test.go:3472: Step 1/4 error: Check failed: Check 1/2 error: Check 1/2 error: google_container_cluster.with_maintenance_exclusion_options: Attribute 'maintenance_policy.0.maintenance_exclusion.1.exclusion_options.0.end_time_behavior' expected "UNTIL_END_OF_SUPPORT", got ""

yliaog:
@slevenick the VCR-test failures do not seem to be related to this PR, could you help take a look?

yliaog:
@slevenick could you please take a look at the PR? and also the test failures that are unrelated to this PR

PR #15525, opened 2025-10-28
Added new resource for "RegionNetworkPolicy"

matheusaleixo-cit:
Adds new resource `google_compute_region_network_policy`.

```release-note:new-resource
`google_compute_region_network_policy`
```


mmv1/products/compute/RegionNetworkPolicy.yaml
```diff
@@ -0,0 +1,97 @@
...
+  delete_minutes: 20
+async:
+  actions: ['create', 'delete', 'update']
+  type: 'OpAsync'
+  operation:
+    base_url: '{{op_id}}'
+  result:
+    resource_inside_response: false
+custom_code:
+legacy_long_form_project: true
```
roaks3: Do you need this?

mmv1/products/compute/RegionNetworkPolicy.yaml
```diff
@@ -0,0 +1,97 @@
...
+samples:
+  - name: "region_network_policy_full"
+    primary_resource_id: "policy"
+    min_version: beta
+    steps:
+      - name: "region_network_policy_full"
+        resource_id_vars:
+          policy_name: "tf-test-policy"
```
roaks3: I think this doesn't require the tf-test prefix, but does require a hyphen. Maybe `rnp-full`?

mmv1/products/compute/RegionNetworkPolicy.yaml
```diff
@@ -0,0 +1,97 @@
...
+async:
+  actions: ['create', 'delete', 'update']
+  type: 'OpAsync'
+  operation:
+    base_url: '{{op_id}}'
+  result:
+    resource_inside_response: false
+custom_code:
+legacy_long_form_project: true
+samples:
```
roaks3, line 36: You should be able to generate the update test now: https://googlecloudplatform.github.io/magic-modules/reference/sample/

roaks3 (commented):
Could you please split out the traffic_classification_rule into a separate PR, for simpler review? We rarely allow multiple resources in a single PR.

mmv1/products/compute/RegionNetworkPolicy.yaml
```diff
@@ -0,0 +1,112 @@
...
+async:
+  actions: ['create', 'delete', 'update']
+  type: 'OpAsync'
+  operation:
+    base_url: '{{op_id}}'
+  result:
+    resource_inside_response: false
+custom_code:
+samples:
+  - name: "region_network_policy_full"
```
roaks3: Could you remove this sample (and keep the following sample)? Because of the way it works, this is now actually a duplicate test/docs example

roaks3 (approved):

PR #15818, opened 2025-11-26
WIP: compute: Add `protection_tier` and `enable_emergent_maintenance` fields to `compute_future_reservation`

Asiderr:
This patch adds `protection_tier` and `enable_emergent_maintenance` to the `google_compute_future_reservation` resource.

```release-note:enhancement
compute: added `protection_tier` field to `google_compute_future_reservation` resource
```
```release-note:enhancement
compute: added `enable_emergent_maintenance` field to `google_compute_future_reservation` resource
```

### DIFF
--- a/mmv1/products/compute/FutureReservation.yaml
+++ b/mmv1/products/compute/FutureReservation.yaml
@@ -192,6 +192,15 @@ properties:
     type: String
     description: |
       Name prefix for the reservations to be created at the time of delivery. The name prefix must comply with RFC1035. Maximum allowed length for name prefix is 20. Automatically created reservations name format will be -date-####.
+  - name: 'protectionTier'
+    type: Enum
+    description: |
+      Protection tier for the workload.
+    enum_values:
+      - 'STANDARD'
+      - 'CAPACITY_OPTIMIZED'
+    default_from_api: true
+    immutable: true
   - name: 'status'
     type: NestedObject
     description: |
@@ -428,6 +437,11 @@ properties:
     type: Boolean
     description: |
       Indicates whether the auto-created reservation can be consumed by VMs with affinity for "any" reservation. If the field is set, then only VMs that target the reservation by name can consume from the delivered reservation.
+  - name: 'enableEmergentMaintenance'
+    type: Boolean
+    description: |
+      Indicates if this group of VMs have emergent maintenance enabled.
+    immutable: true
   - name: 'reservationName'
     type: String
     description: |
--- a/mmv1/third_party/terraform/services/compute/resource_compute_future_reservation_test.go.tmpl
+++ b/mmv1/third_party/terraform/services/compute/resource_compute_future_reservation_test.go.tmpl
@@ -17,8 +17,8 @@ func TestAccComputeFutureReservation_update(t *testing.T) {
 	context := map[string]interface{}{
 		"random_suffix": acctest.RandString(t, 10),
 		"project":       envvar.GetTestProjectFromEnv(),
-		"end_time":        time.Date(time.Now().Year(), 12, 31, 0, 0, 0, 0, time.Now().Location()).AddDate(0, 0, 10).Format(time.RFC3339),
-		"start_time":      time.Date(time.Now().Year(), 12, 31, 0, 0, 0, 0, time.Now().Location()).AddDate(0, 0, 1).Format(time.RFC3339),
+		"end_time":      time.Date(time.Now().Year(), 12, 31, 0, 0, 0, 0, time.UTC).AddDate(0, 0, 10).Format(time.RFC3339),
+		"start_time":    time.Date(time.Now().Year(), 12, 31, 0, 0, 0, 0, time.UTC).AddDate(0, 0, 1).Format(time.RFC3339),
 	}
 
 	acctest.VcrTest(t, resource.TestCase{
@@ -47,6 +47,33 @@ func TestAccComputeFutureReservation_update(t *testing.T) {
 				ImportState:             true,
 				ImportStateVerify:       true,
 				ImportStateVerifyIgnore: []string{"auto_delete_auto_created_reservations"},
+			},
+		},
+	})
+}
+
+func TestAccComputeFutureReservation_maintenance_protection(t *testing.T) {
+	t.Parallel()
+
+	context := map[string]interface{}{
+		"random_suffix": acctest.RandString(t, 10),
+		"project":       envvar.GetTestProjectFromEnv(),
+		"end_time":      time.Date(time.Now().Year(), 12, 31, 0, 0, 0, 0, time.UTC).AddDate(0, 0, 10).Format(time.RFC3339),
+		"start_time":    time.Date(time.Now().Year(), 12, 31, 0, 0, 0, 0, time.UTC).AddDate(0, 0, 1).Format(time.RFC3339),
+	}
+
+	acctest.VcrTest(t, resource.TestCase{
+		PreCheck:                 func() { acctest.AccTestPreCheck(t) },
+		ProtoV5ProviderFactories: acctest.ProtoV5ProviderBetaFactories(t),
+		Steps: []resource.TestStep{
+			{
+				Config: testAccComputeFutureReservation_maintenance_protection(context),
+			},
+			{
+				ResourceName:            "google_compute_future_reservation.gce_future_reservation",
+				ImportState:             true,
+				ImportStateVerify:       true,
+				ImportStateVerifyIgnore: []string{"auto_delete_auto_created_reservations", "enable_emergent_maintenance"},
 			},
 		},
 	})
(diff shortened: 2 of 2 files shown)


mmv1/third_party/terraform/services/compute/resource_compute_future_reservation_test.go.tmpl
```diff
@@ -17,8 +17,8 @@ func TestAccComputeFutureReservation_update(t *testing.T) {
 	context := map[string]interface{}{
 		"random_suffix": acctest.RandString(t, 10),
 		"project":       envvar.GetTestProjectFromEnv(),
-		"end_time":        time.Date(time.Now().Year(), 12, 31, 0, 0, 0, 0, time.Now().Location()).AddDate(0, 0, 10).Format(time.RFC3339),
-		"start_time":      time.Date(time.Now().Year(), 12, 31, 0, 0, 0, 0, time.Now().Location()).AddDate(0, 0, 1).Format(time.RFC3339),
+		"end_time":        time.Date(time.Now().Year(), 12, 31, 0, 0, 0, 0, time.UTC).AddDate(0, 0, 10).Format(time.RFC3339),
```
c2thorn: does this change in the test matter?

mmv1/third_party/terraform/services/compute/resource_compute_future_reservation_test.go.tmpl
```diff
@@ -61,6 +61,8 @@ resource "google_compute_future_reservation" "gce_future_reservation" {
   project  = "%{project}"
   planning_status = "DRAFT"
   auto_delete_auto_created_reservations = true
+  enable_emergent_maintenance = false
```
c2thorn: we should change the values of these two fields in either this block or the updated one. So that way we are actually testing the update functionality

c2thorn (changes requested):

mmv1/third_party/terraform/services/compute/resource_compute_future_reservation_test.go.tmpl
Asiderr: It's needed when you're running tests locally - Now().Location() causes diff when you're running test from different location than UTC.

mmv1/third_party/terraform/services/compute/resource_compute_future_reservation_test.go.tmpl
Asiderr: In place update is not supported by those fields. I marked them as immutable in mmv1 config and I created new test for force new update.

mmv1/products/compute/FutureReservation.yaml
```diff
@@ -428,6 +438,12 @@ properties:
     type: Boolean
     description: |
       Indicates whether the auto-created reservation can be consumed by VMs with affinity for "any" reservation. If the field is set, then only VMs that target the reservation by name can consume from the delivered reservation.
+  - name: 'enableEmergentMaintenance'
+    type: Boolean
+    description: |
+      Indicates if this group of VMs have emergent maintenance enabled.
+    immutable: true
+    ignore_read: true
```
c2thorn: Why is this ignore_read? A field with this but no other related custom logic is a red flag for `ignore_read` not being the correct way to go. What is the behavior around this field?

mmv1/third_party/terraform/services/compute/resource_compute_future_reservation_test.go.tmpl
```diff
@@ -52,6 +52,47 @@ func TestAccComputeFutureReservation_update(t *testing.T) {
 	})
 }
 
+func TestAccComputeFutureReservation_maintenance_protection_update(t *testing.T) {
```
c2thorn: If update is not supported, we do not need to have an update test. We generally assume ForceNew will work as intended.

This can be modified to be a simpler non-update test.

c2thorn (changes requested):

c2thorn:
I'm going to be out of office for a while, I'm reassigning a new reviewer to take over after addressing my latest review comments.

c2thorn:

mmv1/products/compute/FutureReservation.yaml
```diff
@@ -189,6 +189,16 @@ properties:
     type: String
     description: |
       Name prefix for the reservations to be created at the time of delivery. The name prefix must comply with RFC1035. Maximum allowed length for name prefix is 20. Automatically created reservations name format will be -date-####.
+  - name: 'protectionTier'
+    type: Enum
+    description: |
+      Protection tier for the workload.
+    enum_values:
+      - 'STANDARD'
+      - 'PROTECTION_TIER_UNSPECIFIED'
```
melinath: Generally we don't want to include UNSPECIFIED values unless they're semantically distinct from an empty string (which is very unusual).

mmv1/products/compute/FutureReservation.yaml
```diff
@@ -189,6 +189,16 @@ properties:
...
+      - 'CAPACITY_OPTIMIZED'
+    default_from_api: true
```
melinath, line 202: Would it make sense to use a default_value instead of default_from_api? https://googlecloudplatform.github.io/magic-modules/develop/diffs/#default

melinath (changes requested):
a couple things in addition to Cameron's notes.

mmv1/products/compute/FutureReservation.yaml
Asiderr: It's not needed once I marked this field immutable. I deleted this flag.

mmv1/third_party/terraform/services/compute/resource_compute_future_reservation_test.go.tmpl
Asiderr: I simplified this test to non-update.

mmv1/products/compute/FutureReservation.yaml
Asiderr: I deleted UNSPECIFIED value

mmv1/products/compute/FutureReservation.yaml
Asiderr: I replaced default_from_api with default_value.

mmv1/products/compute/FutureReservation.yaml
```diff
@@ -195,6 +195,15 @@ properties:
...
     description: |
       Name prefix for the reservations to be created at the time of delivery. The name prefix must comply with RFC1035. Maximum allowed length for name prefix is 20. Automatically created reservations name format will be -date-####.
+  - name: 'protectionTier'
+    type: Enum
+    description: |
+      Protection tier for the workload.
+    enum_values:
+      - 'STANDARD'
+      - 'CAPACITY_OPTIMIZED'
+    default_value: 'STANDARD'
```
melinath: The breaking change detector is rightly pointing out that this could force recreation for existing configurations unless this matches the API default.

Looking at previous nightly test logs, it seems like the current API default is to not return this field at all, rather than having a default value. So, this probably isn't safe. We should go back to default_from_api. Sorry for leading you down a wrong path!

melinath (changes requested):

mmv1/products/compute/FutureReservation.yaml
Asiderr: No worries, I brought back `default_from_api`

mmv1/third_party/terraform/services/compute/resource_compute_future_reservation_test.go.tmpl
```diff
@@ -101,4 +128,31 @@ resource "google_compute_future_reservation" "gce_future_reservation" {
...
+func testAccComputeFutureReservation_maintenance_protection(context map[string]interface{}) string {
+	return acctest.Nprintf(`
+resource "google_compute_future_reservation" "gce_future_reservation" {
+  provider = google-beta
+  name     = "tf-test-fr%{random_suffix}"
+  name_prefix = "fr-%{random_suffix}"
+  project  = "%{project}"
+  planning_status = "DRAFT"
+  auto_delete_auto_created_reservations = true
+  enable_emergent_maintenance = false
```
melinath:
```suggestion
  enable_emergent_maintenance = true
```

melinath: I think this is necessary to exercise the field

melinath (commented):
It looks like there's an API bug, maybe? It's not returning a value for enableEmergentMaintenance.

Or maybe this is only writable and that's expected? If so, you can add `ignore_read: true` to the field to indicate it's not getting returned by the API.

melinath:
It seems like it will be returned from the API in the future. We'll either need ignore_read or to wait for that.

Asiderr:
> It seems like it will be returned from the API in the future. We'll either need ignore_read or to wait for that.

I added `ignore_read` flag for this field. Right now test is passing for me.

melinath:
It seems like this might be an API bug. Working on determining the best path forward.

melinath (commented):
Removing this from my review queue until the question around it is resolved. attn @mskrynnik FYI

melinath (changes requested):
it looks like ignore_read isn't the correct workaround; possibly this field needs to be updated after create. (Which is weird because the field is currently marked immutable.)

b/480938910 has more context.

melinath (changes requested):
for clarity, if this API behavior isn't changed, the fix here would be:

- remove ignore_read and immutable from enableEmergentMaintenance
- add [post_create custom code](https://googlecloudplatform.github.io/magic-modules/develop/custom-code/#pre_post_injection) that does a [PATCH Request](https://github.com/hashicorp/terraform-provider-google-beta/blob/9c07b58566d2cca22c7cc42393562e66f1043031/google-beta/services/compute/resource_compute_future_reservation.go#L1260) to update just this field.

But an API fix would be preferable.

PR #15996, opened 2025-12-18
adding a flag to serialize generated tests

NickElliot:
was reached out to by a service team regarding https://github.com/hashicorp/terraform-provider-google/issues/25577 https://github.com/hashicorp/terraform-provider-google/issues/25576 https://github.com/hashicorp/terraform-provider-google/issues/25523 where the flakiness seemed to be caused by these tests being ran in parallel. rather than recreating all the tests as handwritten, this feels like a straight forward solution for long term maintainability.

```release-note:none

```

### DIFF
--- a/mmv1/api/resource/examples.go
+++ b/mmv1/api/resource/examples.go
@@ -154,6 +154,11 @@ type Examples struct {
 	// See test_vars_overrides for more details
 	OicsVarsOverrides map[string]string `yaml:"oics_vars_overrides,omitempty"`
 
+	//If the generated tests should be ran in serial.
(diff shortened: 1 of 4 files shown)


NickElliot:

NickElliot:
tests dont actually need changes per yaqs/7977712224003162112, but I think this still would be a useful thing to have

roaks3 (approved):
Seems fine to me (might want to raise with the team as a change to the mm interface). The one thing I would mention just in case it's worth calling out: I believe in most test environments we run a separate `go test` per product, so the serial behavior would only apply within a particular product (ie. tests from other products could be running at the same time).

PR #16203, opened 2026-01-23
feat(firestore): Support for skip wait on Firestore index resource.

### LINKED ISSUE #13422: google_firestore_index non-blocking option
### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment. If the issue is assigned to the "modular-magician" user, it is either in the process of being autogenerated, or is planned to be autogenerated soon. If the issue is assigned to a user, that user is claiming responsibility for the issue. If the issue is assigned to "hashibot", a community member has claimed the issue already.

### Description

A Firestore index can take a while to create. For a [create index](https://firebase.google.com/docs/firestore/reference/rest/v1/projects.databases.collectionGroups.indexes/create) call, the API returns an operation, which the caller poll to wait. Currently, the [Firebase CLI](https://github.com/firebase/firebase-tools/blob/9c2eeba83eff5d8276d828bb36264f18f0c12dbb/src/firestore/indexes.ts#L103) doesn't wait on this operation (just succeeds), and the UI surfaces (e.g. Cloud Console, Firebase Console) shows a non-blocking in-progress state in the UI. However, Terraform blocks on the index creation due to [polling the returned operation](https://github.com/hashicorp/terraform-provider-google-beta/blob/3ea6e9b176e7ef0be49e2370947b2e63f276b175/google-beta/resource_firestore_index.go#L229), which makes `terraform apply` much slower (4 minutes per index for me on an empty database).

### New or Affected Resource(s)

* google_firestore_index

### Potential Terraform Configuration

It's ok to CRUD Firestore documents as normal when the index is being built, so I can't see why we need to wait for the index to be fully built. This is probably why none of the other surfaces (CLI, Console) blocks anything on index creation. For backward compatibility, I suggest a `skip_wait` field to the Terraform resource that doesn't wait on the returned operation, and maybe make it the default behavior in a major release.

```tf
# Propose what you think the configuration to take advantage of this feature should look like.
# We may not use it verbatim, but it's helpful in understanding your intent.
resource "google_firestore_index" "my-index" {
  project = "my-project-name"
  ...
  
  skip_wait = true
}
```

### References

`skip_wait` may also circumvent the problem experienced in #11316 because the permission "datastore.operations.get" is no longer necessary.

b/275100809

pcostell:
Add `skip_wait` field to `google_firestore_index` resource

This field allows users to skip waiting for the long-running operation to complete during index creation. This is useful for creating multiple indexes in parallel or when usage patterns do not require immediate index readiness.

Also adds support for updating virtual fields on otherwise immutable resources.

Fixes https://github.com/hashicorp/terraform-provider-google/issues/13422

```release-note:enhancement
firestore: added `skip_wait` field to `google_firestore_index` resource, skipping the wait for index creation
```

### DIFF
--- a/mmv1/products/firestore/Index.yaml
+++ b/mmv1/products/firestore/Index.yaml
@@ -51,6 +51,7 @@ custom_code:
   constants: 'templates/terraform/constants/firestore_index.go.tmpl'
   encoder: 'templates/terraform/encoders/index.go.tmpl'
   custom_import: 'templates/terraform/custom_import/index_self_link_as_name_set_project.go.tmpl'
+  custom_create: 'templates/terraform/custom_create/firestore_index.go.tmpl'
 error_retry_predicates:
 
   - 'transport_tpg.FirestoreIndex409Retry'
@@ -98,7 +99,22 @@ examples:
       database_id: 'database-id-unique'
     test_env_vars:
       project_id: 'PROJECT_NAME'
+  - name: 'firestore_index_skip_wait'
+    primary_resource_id: 'my-index'
+    vars:
+      database_id: 'database-id-skip-wait'
+    test_env_vars:
+      project_id: 'PROJECT_NAME'
+    ignore_read_extra:
+      - 'skip_wait'
+virtual_fields:
+  - name: "skip_wait"
+    type: Boolean
+    default_value: false
+    description: "Whether to skip waiting for the index to be created."
+    ignore_read: true
 parameters:
+# Firestore uses a custom create function. Any new fields must be explicitly handled there.
 properties:
   - name: 'name'
     type: String
--- a/mmv1/templates/terraform/custom_create/firestore_index.go.tmpl
+++ b/mmv1/templates/terraform/custom_create/firestore_index.go.tmpl
@@ -1,0 +1,145 @@
+userAgent, err := tpgresource.GenerateUserAgentString(d, config.UserAgent)
+if err != nil {
+	return err
+}
+
+obj := make(map[string]interface{})
+databaseProp, err := expandFirestoreIndexDatabase(d.Get("database"), d, config)
+if err != nil {
+	return err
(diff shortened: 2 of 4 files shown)


mmv1/products/firestore/Index.yaml
```diff
@@ -51,6 +51,7 @@ custom_code:
   constants: 'templates/terraform/constants/firestore_index.go.tmpl'
   encoder: 'templates/terraform/encoders/index.go.tmpl'
   custom_import: 'templates/terraform/custom_import/index_self_link_as_name_set_project.go.tmpl'
+  custom_create: 'templates/terraform/custom_create/firestore_index.go.tmpl'
```
rileykarson, line 54: drive-by comment: @shuyama1 can you evaluate making a template change to make this possible w/o a full custom create? This needs minimal structural change, so a custom method is pretty heavyhanded but there's no appropriate injection point atm.

pcostell: I did first attempt a template version. The primary issue I ran into was that normally the LRO API will only return the resource when complete. For Firestore, we put the index as a field in metadata, but this looks to be fairly specific to Firestore.

Slightly proposed alternative, lmk what you think:
1) introduce a skip_wait to async.go
2) add a new "extractor" for resource ids from pending LRO (custom decoder?)
3) add the same "if skip_wait" logic to the resource.go.tmpl

pcostell: Generic alternative in https://github.com/GoogleCloudPlatform/magic-modules/pull/16225.

shuyama1 (commented):
Sorry for the delay. The change itself looks reasonable overall. I am now taking a look at the template change PR https://github.com/GoogleCloudPlatform/magic-modules/pull/16225

shuyama1 (approved):

PR #16225, opened 2026-01-26
feat: Add skip_wait option to async operations.

### LINKED ISSUE #13422: google_firestore_index non-blocking option

pcostell:
This allows users to skip waiting for the long-running operation to complete during index creation. This is useful for creating resources in parallel, when usage patterns do not require immediate resource readiness, and especially when resource creation can take hours.

Resource providers may provide a custom decoder to extract the resource from the in-progress operation.

See Firestore Index.yaml for an example of usage.

Fixes https://github.com/hashicorp/terraform-provider-google/issues/13422

```release-note:enhancement
firestore: added `skip_wait` field to `google_firestore_index` resource, skipping the wait for index creation
```

### DIFF
--- a/mmv1/api/async.go
+++ b/mmv1/api/async.go
@@ -34,6 +34,9 @@ type Async struct {
 	// The list of methods where operations are used.
 	Actions []string `yaml:"actions,omitempty"`
 
+	// If true, the system will allow the user to skip waiting for the operation result.
+	AllowSkipWait bool `yaml:"allow_skip_wait,omitempty"`
+
 	OpAsync   `yaml:",inline"`
 	PollAsync `yaml:",inline"`
 }
--- a/mmv1/api/resource.go
+++ b/mmv1/api/resource.go
@@ -935,6 +935,24 @@ func (r *Resource) AddExtraFields(props []*Type, parent *Type) []*Type {
 	return props
 }
 
+func (r *Resource) AddExtraVirtualFields(virtualFields []*Type, parent *Type) []*Type {
+	if parent == nil && r.Async != nil && r.Async.AllowSkipWait {
+		virtualFields = append(virtualFields, buildSkipWaitField())
+	}
+	return virtualFields
+}
+
+func buildSkipWaitField() *Type {
+	options := []func(*Type){
+		propertyWithType("Boolean"),
+		propertyWithIgnoreWrite(true),
+		propertyWithImmutable(false),
+		propertyWithClientSide(true),
+		propertyWithDescription("If true, the asynchronous operation will not be awaited."),
+	}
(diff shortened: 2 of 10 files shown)


pcostell:
This is a more generic alternative to a Firestore-specific approach in https://github.com/GoogleCloudPlatform/magic-modules/pull/16203.

cc @rileykarson and @shuyama1

shuyama1:
@modular-magician reassign-reviewer shuyama1

PR #16247, opened 2026-01-28
Added flexible catalog support for BigLake IcebergCatalog resource.

nika-qubit:

```release-note:enhancement
biglakeiceberg: added `default_location` and `additional_locations` fields to `google_biglake_iceberg_catalog` resource
```


nika-qubit:
Please do not review or submit this draft PR until it's ready since the related product change hasn't gone public yet.

mmv1/templates/terraform/custom_create/biglake_iceberg_catalog_create.go.tmpl
```diff
@@ -0,0 +1,91 @@
...
+  obj["catalog-type"] = catalogTypeProp
+}
+defaultLocationProp, err := expandBiglakeIcebergIcebergCatalogDefaultLocation(d.Get("default_location"), d, config)
+if err != nil {
+  return err
+} else if v, ok := d.GetOkExists("default_location"); !tpgresource.IsEmptyValue(reflect.ValueOf(defaultLocationProp)) && (ok || !reflect.DeepEqual(v, defaultLocationProp)) {
+  obj["default-location"] = defaultLocationProp
+}
+
+// This custom logic determines whether the default-location is required or output only.
```
Noremac201, line 26: If we can forgo all this create.go.tmpl code and just defer the actual validation to the control plane, that would keep future updates easy to this TF code rather than needing to maintain custom logic

nika-qubit: Thanks for the review! I'll check again when we have the feature public and remove the custom logic if the server side can handle it.

mmv1/products/biglakeiceberg/IcebergCatalog.yaml
```diff
@@ -106,19 +107,21 @@ properties:
...
   - name: 'default_location'
     api_name: 'default-location'
     type: String
-    description: Output only. The default storage location for the catalog, e.g.,
+    description: The default storage location for the catalog, e.g.,
       `gs://my-bucket`.
-    output: true
+      Output only when the catalog type is CATALOG_TYPE_GCS_BUCKET.
+      Required when the catalog type is CATALOG_TYPE_BIGLAKE.
+    required: false
```
rileykarson, line 124:
```suggestion
```

mmv1/products/biglakeiceberg/IcebergCatalog.yaml
```diff
@@ -106,19 +107,21 @@ properties:
...
     immutable: true
     output: false
     enum_values:
       - 'CATALOG_TYPE_GCS_BUCKET'
+      - 'CATALOG_TYPE_BIGLAKE'
   - name: 'default_location'
     api_name: 'default-location'
     type: String
-    description: Output only. The default storage location for the catalog, e.g.,
+    description: The default storage location for the catalog, e.g.,
```
rileykarson, line 120:
```suggestion
    default_from_api: true
    description: The default storage location for the catalog, e.g.,
```

This will make it so that if the field is in config we detect drift, and if it's not we won't. That means when CATALOG_TYPE_GCS_BUCKET is set we'll record the outputted value with no diff, and if the values don't match for CATALOG_TYPE_BIGLAKE we'll show one.

There is a corner case where if a user creates the resource with CATALOG_TYPE_BIGLAKE and then removes the value from config we'll preserve the old one as a result- but that typically isn't a big deal.

mmv1/products/biglakeiceberg/IcebergCatalog.yaml
```diff
@@ -164,3 +167,14 @@ properties:
...
+    api_name: 'additional-locations'
+    type: Array
+    item_type:
+      type: String
+    description: Additional GCS locations (e.g., `gs://my-other-bucket/...`)
+      that are permitted for use by resources within this catalog. This field
+      can be used to specify either a GCS bucket or a path within it.
+    required: false
+    immutable: false
+    output: false
```
rileykarson, line 180:
```suggestion
```

Same as above-  we can omit `false` defaults.

mmv1/templates/terraform/custom_create/biglake_iceberg_catalog_create.go.tmpl
rileykarson: What was the result here? There's another spot that we can put the code that will be less intrusive + validate earlier, but it'll take some work to move. Falling back to the control plane is fine.

nika-qubit:
It's ready for review for the main gist. I'll remove the create custom template and restore create_url if approved for merging.

mmv1/products/biglakeiceberg/IcebergCatalog.yaml
```diff
@@ -106,19 +107,21 @@ properties:
...
     type: Enum
-    description: The catalog type of the IcebergCatalog. Currently only supports
-      the type for Google Cloud Storage Buckets.
+    description: The catalog type of the IcebergCatalog.
     required: true
     immutable: true
     output: false
     enum_values:
       - 'CATALOG_TYPE_GCS_BUCKET'
+      - 'CATALOG_TYPE_BIGLAKE'
```
rileykarson, line 116: Would it be possible to add an example or test for this type? They're pretty different from each other and that'll flush out any potential issues.

rambleraptor:
I've replaced this with #16778 and made all of the changes. @nika-qubit handed the work off to me.

rambleraptor:
@nika-qubit @rileykarson can one of you close this PR? We've merged this feature in with a different PR.

PR #16277, opened 2026-02-02
Converted write_only_legacy fields to write_only

melinath:

```release-note:none

```

### DIFF
--- a/mmv1/products/bigquerydatatransfer/Config.yaml
+++ b/mmv1/products/bigquerydatatransfer/Config.yaml
@@ -222,30 +222,10 @@ properties:
       to a different credential configuration in the config will require an apply to update state.
     url_param_only: true
     properties:
-      - name: 'secretAccessKeyWoVersion'
-        type: Integer
-        url_param_only: true
-        required_with:
-          - 'sensitive_params.0.secretAccessKeyWo'
-        description: |
-          The version of the sensitive params - used to trigger updates of the write-only params. For more info see [updating write-only arguments](/docs/providers/google/guides/using_write_only_arguments.html#updating-write-only-arguments)
       - name: 'secretAccessKey'
         type: String
         description: |
           The Secret Access Key of the AWS account transferring data from.
         sensitive: true
-        at_least_one_of:
-          - 'sensitive_params.0.secretAccessKey'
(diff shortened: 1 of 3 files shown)


mmv1/products/bigquerydatatransfer/Config.yaml
```diff
@@ -222,30 +222,14 @@ properties:
...
         type: String
         description: |
           The Secret Access Key of the AWS account transferring data from.
         sensitive: true
+        write_only: true
         at_least_one_of:
           - 'sensitive_params.0.secretAccessKey'
           - 'sensitive_params.0.secretAccessKeyWo'
         conflicts:
           - 'sensitive_params.0.secretAccessKeyWo'
```
wj-chen: Can remove now

wj-chen (commented):
The Breaking Change report shows generated provider changes that don't quite match this config change, is that expected?

mmv1/products/bigquerydatatransfer/Config.yaml
melinath: good point, thanks!

melinath:
@wj-chen I'm not quite sure what doesn't match, but for context, this PR isn't actually intended to make the change - just to demonstrate what the diff _would_ be. I'm creating a ticket to track this work for the next major release.

wj-chen:
> @wj-chen I'm not quite sure what doesn't match, but for context, this PR isn't actually intended to make the change - just to demonstrate what the diff _would_ be. I'm creating a ticket to track this work for the next major release.

Ack. For bigquerydatatransfer looks like the change removes `secretAccessKeyWo` and `secretAccessKeyWoVersion` but the generated diff still shows these params just with different types and other property values, hence the confusion: https://github.com/modular-magician/terraform-provider-google/compare/c874c8109350082d1e4c75e7281013b690da3fed..a129b9af534cdee7d60b56940cc230e583c9df4b

melinath:
Ah, yeah - if `write_only: true` is set, we automatically add the `_wo` and `_wo_version` fields in the downstream without needing to specify them explicitly in the config. That ensures that they're set up correctly in terms of, for example, conflicting with each other. (It's easy to do things wrong / inconsistently otherwise, as demonstrated by the fields modified in this PR.)

PR #16319, opened 2026-02-06
`MMv1`: move all inline calls of `flatteners + more` in `Read()` to new `resource{{resourceName}}Flatten` 

BBBmau:
## Purpose

With the introduction of List resources, we are in need of being able to flatten entire resources when iterating through a response containing items each representing individual resources.

This PR introduces a refactor of all per-field flattening logic from the inline `Read()` function into a dedicated `resource<ResourceName>Flatten()` method for every MMv1-generated resource. 

The resource flattener now contains all  inline calls to each property's flatten function, the decoder, the nested query handler, and post-read custom code that was previously only seen in `Read()`

## Edge cases covered

| Edge case | What it exercises | Example resource(s) |
|---|---|---|
| **Decoder only** | Decoder runs inside the flatten method; nil-check removes resource from state if decoder returns nil | `apigee/DeveloperApp`, `appengine/DomainMapping`, `kms/SecretCiphertext` |
| **Decoder + HasSelfLink** | `self_link` is set from `res["selfLink"]` inside the flatten method after the decoder transforms `res` | `compute/Route`, `compute/Snapshot`, `compute/RegionTargetHttpsProxy` |
| **Decoder + FlattenObject** | Properties with `flatten_object: true` iterate over the flattened map and call `d.Set(k, v)` for each key | `cloudrun/Service` |
| **Decoder + HasSelfLink + FlattenObject** | Combines all three: decoder, self_link, and flatten_object in one method | `compute/BackendService`, `compute/Network`, `compute/RegionBackendService` |
| **Decoder + NestedQuery** | NestedQuery extraction runs inside the flatten method (not duplicated in Read); decoder runs after | `activedirectory/DomainTrust`, `compute/DiskResourcePolicyAttachment`, `compute/NetworkEndpoint` |
| **Decoder + PostRead** | PostRead custom code runs at the top of the flatten method, before the decoder | `bigqueryanalyticshub/DataExchangeSubscription`, `bigqueryanalyticshub/ListingSubscription` |
| **Decoder + PostRead + HasSelfLink** | Full combination: PostRead, then decoder with nil-check, then property flattening, then self_link | `compute/Disk`, `compute/RegionDisk` |
| **PostRead + NestedQuery** | PostRead runs first, then NestedQuery extraction (no decoder involved) | `accesscontextmanager/ServicePerimeterEgressPolicy` and 4 sibling perimeter policy resources |
| **NestedQuery + Decoder (naming clash prevention)** | The method is named `resource<Name>Flatten` (not `flatten<Name>Resource`) to avoid collisions with property flatten functions like `flattenBackupDRManagementServerResource` | `backupdr/ManagementServer` (has a property named `resource`) |
| **No gettable properties** | `var err error` is conditionally emitted only when the method body actually uses it, avoiding unused-variable compile errors | `bigqueryreservation/ReservationGroup` |
| **ExcludeRead** | The flatten method is not generated at all for resources that cannot be read from the API | Any resource with `exclude_read: true` |
| **Property name = resource name** | `resource<Name>Flatten` naming avoids the clash that `flatten<Name>` would cause when a resource has a property with the same name (e.g. `SpannerInstance` has a `config` property → `flattenSpannerInstanceConfig` already exists) | `spanner/Instance`, `spanner/InstanceConfig`, `kms/CryptoKey`, `kms/KeyRing` |

## Template changes

- **`resource.go.tmpl` Read()**: Removed all inline PostRead, NestedQuery, Decoder, and per-property `d.Set()` calls. The function now always delegates to `resource<ResourceName>Flatten()`

```release-note:none
refactor of resource.go.tmpl to support `resource{{resourceName}}Flattener` to process entire resource `responses` from `SendRequest - res`
```

### DIFF
--- a/mmv1/templates/terraform/resource.go.tmpl
+++ b/mmv1/templates/terraform/resource.go.tmpl
@@ -1551,9 +1551,6 @@ func resource{{ $.ResourceName -}}PostCreateSetComputedFields(d *schema.Resource
 {{ if $.FlattenAllInMethod }}
 func flatten{{ $.ResourceName -}}(d *schema.ResourceData, meta interface{}, res map[string]interface{}) error {
     config := meta.(*transport_tpg.Config)
-    {{- if $.CustomCode.PostRead }}
-    {{ customTemplate $ $.CustomCode.PostRead false -}}
-{{- end }}
 
 {{if $.NestedQuery -}}
     res, err := flattenNested{{ $.ResourceName -}}(d, meta, res)
--- a/mmv1/templates/terraform/resource.go.tmpl
+++ b/mmv1/templates/terraform/resource.go.tmpl
@@ -816,7 +816,7 @@ func resource{{ $.ResourceName -}}Read(d *schema.ResourceData, meta interface{})
     }
 {{- end}}
 {{ if $.FlattenAllInMethod }}
-    err = flatten{{ $.ResourceName -}}(d, meta, res, config, {{- if $.HasProject }}project,{{ end }} userAgent, url, headers)
+    err = flatten{{ $.ResourceName -}}(d, meta, res, config, {{- if $.HasProject }}project,{{ end }} userAgent, billingProject, url, headers)
     if err != nil {
         return err
     }
@@ -1549,7 +1549,7 @@ func resource{{ $.ResourceName -}}PostCreateSetComputedFields(d *schema.Resource
 {{- end }}
 
 {{ if $.FlattenAllInMethod }}
-func flatten{{ $.ResourceName -}}(d *schema.ResourceData, meta interface{}, res map[string]interface{}, config *transport_tpg.Config,{{if $.HasProject -}}project string, {{end -}}userAgent string, url string, headers http.Header) error {
+func flatten{{ $.ResourceName -}}(d *schema.ResourceData, meta interface{}, res map[string]interface{}, config *transport_tpg.Config,{{if $.HasProject -}}project string, {{end -}}userAgent string, billingProject string, url string, headers http.Header) error {
 {{- if and $.CustomCode.PostRead }}
     {{ customTemplate $ $.CustomCode.PostRead false -}}
 {{- end }}
--- a/mmv1/templates/terraform/resource.go.tmpl
+++ b/mmv1/templates/terraform/resource.go.tmpl
@@ -745,6 +745,38 @@ func resource{{ $.ResourceName -}}Read(d *schema.ResourceData, meta interface{})
     }
 
     log.Printf("[DEBUG] Finished reading {{ $.ResourceName }} %q: %#v", d.Id(), res)
+{{- if $.CustomCode.PostRead }}
+    {{ customTemplate $ $.CustomCode.PostRead false -}}
+{{- end }}
+
+{{if $.NestedQuery -}}
+    res, err = flattenNested{{ $.ResourceName -}}(d, meta, res)
+    if err != nil {
+        return err
+    }
+
+    if res == nil {
+        // Object isn't there any more - remove it from the state.
+        log.Printf("[DEBUG] Removing {{ $.ResourceName }} because it couldn't be matched.")
+        d.SetId("")
+        return nil
+    }
+
+{{- end}}
+
+{{if $.CustomCode.Decoder -}}
+    res, err = resource{{ $.ResourceName -}}Decoder(d, meta, res)
+    if err != nil {
+        return err
+    }
+
+    if res == nil {
+        // Decoding the object has resulted in it being gone. It may be marked deleted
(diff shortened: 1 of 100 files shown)


mmv1/templates/terraform/resource.go.tmpl
```diff
@@ -745,38 +745,6 @@ func resource{{ $.ResourceName -}}Read(d *schema.ResourceData, meta interface{})
     }
 
     log.Printf("[DEBUG] Finished reading {{ $.ResourceName }} %q: %#v", d.Id(), res)
-{{- if $.CustomCode.PostRead }}
```
c2thorn, line 748: Does this chunk need to be moved as well? By moving it into the single function that's called later, we are making the 

{{- if $.VirtualFields }}
{{- if $.HasProject }}
{{if $.HasRegion -}}
{{if $.HasZone -}}

chunks happen before this logic. I'd be worried about some one-off custom code actually depending on the order and breaking.

c2thorn: If it does need to be moved for the sake of sharing code, then maybe two separate functions are appropriate to preserve logic order?

mmv1/templates/terraform/resource.go.tmpl
```diff
@@ -1541,3 +1486,71 @@ func resource{{ $.ResourceName -}}PostCreateSetComputedFields(d *schema.Resource
     return nil
 }
 {{- end }}
+
+{{ if not $.ExcludeRead }}
+func Resource{{ $.ResourceName -}}Flatten(d *schema.ResourceData, meta interface{}, res map[string]interface{}, config *transport_tpg.Config,{{if $.HasProject -}}project string, {{end -}}userAgent string, billingProject string, url string, headers http.Header) error {
+{{- if or $.NestedQuery $.CustomCode.Decoder $.CustomCode.PostRead $.GettableProperties $.HasSelfLink }}
+var err error
+{{- end }}
+{{- if and $.CustomCode.PostRead }}
```
c2thorn:
```suggestion
{{- if $.CustomCode.PostRead }}
```

if this works now, it's at least redundant

c2thorn (commented):
mainly a question about logic ordering

mmv1/templates/terraform/resource.go.tmpl
BBBmau: I reverted this since i did notice after your review that this isn't necessary to move. 

This may have made more sense in how the flattener would be used at the time of implementing `list-resources` in MMV1 but some changes were made to allow for customCode to be used within the flattener portion in order to include any necessary custom code that is also present in the resource's read prior to getting to the flattener logic.

Here's more context around how list would be using the `ResourceFlattener`. This is [list_cloud_run_service](https://github.com/hashicorp/terraform-provider-google/commit/8baecf170b03fe06f3b5e36502361e9b5d028231#diff-1d1f3ee1c4e560eb637e4c4cf155c1e6f470c7b8f876af8f6176cb924b0eef92R146-R177) where decoder is used with the [supporting MMv1 generation code](https://github.com/BBBmau/magic-modules/pull/29/changes#diff-06fbf06968ac8f95b7297ff67bfeb00dbbd09b9f4febc41bdeb1d3352fc4e367R42-R100) that produced it.

The VCR tests themselves have been failing in nightly, they aren't unrelated to the refactor that's being introduced in this PR

c2thorn:
Do you know why these tests are failing in VCR?

BBBmau:
> Do you know why these tests are failing in VCR?

they've all been failing consistently in `nightly-test` except for `TestAccDataformConfig_update`. Currently looking into what's causing this right now.

Edit: I ran `TestAccDataformConfig_update` in a [TeamCity Build](https://hashicorp.teamcity.com/buildConfiguration/TerraformProviders_GoogleCloud_GOOGLE_BETA_MMUPSTREAMTESTS_GOOGLEBETA_PACKAGE_DATAFORM/656591?buildTab=tests&expandedTest=build%3A%28id%3A656591%29%2Cid%3A2000000000) and we're getting a passing result

The diff on the dataform_config resource itself is very minimal, essentially just moving one flattener to inside the `resourceDataformConfigFlatten` method:

```go
func ResourceDataformConfigFlatten(d *schema.ResourceData, meta interface{}, res map[string]interface{}, config *transport_tpg.Config, project string, userAgent string, billingProject string, url string, headers http.Header) error {
	var err error

	if err = d.Set("default_kms_key_name", flattenDataformConfigDefaultKmsKeyName(res["defaultKmsKeyName"], d, config)); err != nil {
		return fmt.Errorf("Error reading Config: %s", err)
	}

	return nil
}
```

c2thorn (approved):
Think we are good now. This is really just moving code to a new function, no more reordering of logic.

PR #16523, opened 2026-02-25
Adding SnapshotRecycleBinPolicy resource for beta

jaiswalakshat:
### Issue

We are adding terraform support for Snapshot Recycle Bin on Compute Disk. This PR add a snapshot recycle bin policy resource on terraform. The resource will be only enabled for beta. 

For now the testing has been done against an allowlisted alpha staging project. As the feature is not available on beta we cannot at integration tests for it. Integration tests will come in follow up PR.

### Testing

Tested manually against alpha endpoint. 

[logs link](https://paste.googleplex.com/5615501314228224)

```release-note:REPLACEME
SnapshotRecycleBinPolicy resource for Disk Recycle Bin feature.

```

### DIFF
--- a/mmv1/products/compute/SnapshotRecycleBinPolicy.yaml
+++ b/mmv1/products/compute/SnapshotRecycleBinPolicy.yaml
@@ -33,13 +33,16 @@ examples:
 base_url: 'projects/{{project}}/global/snapshotRecycleBinPolicy'
 update_url: 'projects/{{project}}/global/snapshotRecycleBinPolicy'
 immutable: false
+import_format:
+  - 'projects/{{project}}/global/snapshotRecycleBinPolicy/'
 timeouts:
   insert_minutes: 20
   update_minutes: 20
   delete_minutes: 20
 create_verb: 'PATCH'
 update_verb: 'PATCH'
-update_mask: true
+custom_code:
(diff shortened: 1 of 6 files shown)


mmv1/products/compute/SnapshotRecycleBinPolicy.yaml
```diff
@@ -0,0 +1,79 @@
...
+      applied if snapshots do not have any of these tags. The value is the rule for the key.
+    key_name: 'name'
+    value_type:
+      type: NestedObject
+      properties:
+      - name: 'standardSnapshots'
+        type: NestedObject
+        description: The rule config for standard snapshots.
+        properties:
+        - name: 'retentionDurationDays'
```
BBBmau, line 79: the missing test report comes from not including an update test for this field. Refer to the Magic-Modules docs about [update tests](https://googlecloudplatform.github.io/magic-modules/test/test/#add-an-update-test)

BBBmau (commented):
aside from the linting errors that need to be addressed, we also have a missing test report to address as well.

BBBmau (changes requested):
this PR is introducing multiple resources, we should instead break this up so that one PR is focusing on a single resource

PR #16565, opened 2026-02-27
Add exclude_from_gen_ai_features field to Datastore

ranlz77:
google_discovery_engine_data_store.

This PR adds `exclude_from_gen_ai_features` as a valid field to the `google_discovery_engine_data_store` resource.
FIxes b/483792089

```release-note:enhancement
discoveryengine: added `exclude_from_gen_ai_features` field to `google_discovery_engine_data_store` resource
```

### DIFF
--- a/mmv1/products/discoveryengine/DataStore.yaml
+++ b/mmv1/products/discoveryengine/DataStore.yaml
@@ -419,6 +419,11 @@ properties:
                   required: false
                   item_type:
                     type: String
+  - name: 'excludeFromGenAiFeatures'
+    type: 'Boolean'
+    description: |
+      If set true, the DataStore will not be used for Generative AI features.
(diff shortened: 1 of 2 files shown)


SirGitsalot (commented):
Do you know if this is fully rolled out? It looks like there is still a visibility label on the field. I just kicked off the test run, that might tell us.

ranlz77:
@SirGitsalot Thanks for the heads up! I noticed the field was available in the Cloud Console's payload, but it is not mentioned in the [v1 API document](https://docs.cloud.google.com/gemini/enterprise/docs/reference/rest/v1/projects.locations.collections.dataStores) of `projects.locations.collections.dataStores`. So you're right, it might still be behind a visibility gate.

PR #16609, opened 2026-03-04
Add recurring maitenance window to maintenance policy.

Ewa-M:
Add recurring maitenance window to maintenance policy. 

```release-note:new-resource
`google_container_recurring_maintenance_window`
```


Ewa-M:
Tests have the same scope as recurring_window (both don't check individual fields), docs are updated

slevenick (changes requested):
Looking at the diff, something has gone wrong here. Please isolate the change to maintenance window/policy from the rest of the changes contained in this PR

slevenick (changes requested):
Still need to isolate changes. One way to do that is through creating a git patch of your local changes and applying it on a fresh copy of the main branch

PR #16685, opened 2026-03-11
Fix perma-diff in google_service_networking_connection for reserved_peering_ranges ordering

SCSAndre:
## Description
Fixes PR #20398 (https://github.com/hashicorp/terraform-provider-google/issues/20398)
Replaces PR #16191 with clean commit history
Reopening PR #16259 

This PR resolves a perma-diff issue in `google_service_networking_connection` where the order of `reserved_peering_ranges` caused Terraform to detect changes during every plan, even if the set of ranges remained identical.

## Changes
- **Schema Change**: Added a `DiffSuppressFunc` to `reserved_peering_ranges` in `resource_service_networking_connection.go` to ignore ordering differences between config and state.
- **Helper Function**: Added `stringListDiffSuppress` to handle the unordered comparison of the string list.
- **Test**: Added `TestAccServiceNetworkingConnection_reorder` to `resource_service_networking_connection_test.go` to verify that changing the order of ranges does not trigger an update.

## Breaking Change Note
This change utilizes `DiffSuppressFunc` on the existing `TypeList` schema rather than converting to `TypeSet`.

**Impact**: None. This is not a breaking change. Existing state files remain compatible as the schema type has not changed.

## Verification
Test added: `TestAccServiceNetworkingConnection_reorder` to verify order changes don't trigger updates.

```release-note:bug
servicenetworking: fixed a permadiff issue of `reserved_peering_ranges` in `google_service_networking_connection`
````

### DIFF
--- a/.changelog/16191.txt
+++ b/.changelog/16191.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+servicenetworking: fixed perma-diff issue in `google_service_networking_connection` when `reserved_peering_ranges` order changes
+```
--- a/.changelog/16191.txt
+++ b/.changelog/16191.txt
@@ -1,3 +1,0 @@
-```release-note:bug
-servicenetworking: fixed perma-diff issue in `google_service_networking_connection` when `reserved_peering_ranges` order changes
-```
--- a/mmv1/third_party/terraform/services/servicenetworking/resource_service_networking_connection.go
+++ b/mmv1/third_party/terraform/services/servicenetworking/resource_service_networking_connection.go
@@ -231,10 +231,13 @@ func resourceServiceNetworkingConnectionRead(d *schema.ResourceData, meta interf
 		return fmt.Errorf("Error setting peering: %s", err)
 	}
 
-	ranges := connection.ReservedPeeringRanges
-	if err := d.Set("reserved_peering_ranges", ranges); err != nil {
+	// removed the intermediate `ranges` variable — it was a
+	// leftover from when sort.Strings() lived here. The DiffSuppressFunc
+	// already handles ordering, so we write the API response directly to state.
+	if err := d.Set("reserved_peering_ranges", connection.ReservedPeeringRanges); err != nil {
 		return fmt.Errorf("Error setting reserved_peering_ranges: %s", err)
 	}
+
 	return nil
 }
 
@@ -438,7 +441,6 @@ func RetrieveServiceNetworkingNetworkName(d *schema.ResourceData, config *transp
 
 	// return the network name formatting unique to this API
 	return fmt.Sprintf("projects/%v/global/networks/%v", project.ProjectNumber, networkName), nil
-
 }
 
 const parentServicePattern = "^services/.+$"
--- a/mmv1/third_party/terraform/services/servicenetworking/resource_service_networking_connection.go
+++ b/mmv1/third_party/terraform/services/servicenetworking/resource_service_networking_connection.go
@@ -465,25 +465,23 @@ func init() {
 }
 
 func stringListDiffSuppress(k, old, new string, d *schema.ResourceData) bool {
-	// Extract the root field name from k (e.g., "reserved_peering_ranges.0" -> "reserved_peering_ranges")
-	// DiffSuppressFunc is called for each element and for the count, so we need to handle both
-	rootKey := k
-	if idx := strings.Index(k, "."); idx != -1 {
-		rootKey = k[:idx]
-	}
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

**Comment by SCSAndre** [comment]
Hey, @shuyama1, if possible, please share the reasons why the VCR tests are failing?

**Comment by shuyama1** [comment]
> Hey, @shuyama1, if possible, please share the reasons why the VCR tests are failing?

Sorry, it doesn't seem the VCR check is for your latest commit. Let me rerun the tests

/gcbrun

**Review by shuyama1 — COMMENTED** [review]
Could you double check if you've submitted the correct commits?

**Comment by SCSAndre** [comment]
@shuyama1, sorry for the confusion! I accidentally submitted the wrong branch for this PR, as you saw. Thanks for letting me know! Everything seems to be correct now.

File: mmv1/third_party/terraform/services/servicenetworking/resource_service_networking_connection.go [hunk 1]
```diff
@@ -226,7 +229,10 @@ func resourceServiceNetworkingConnectionRead(d *schema.ResourceData, meta interf
 	if err := d.Set("peering", connection.Peering); err != nil {
 		return fmt.Errorf("Error setting peering: %s", err)
 	}
-	if err := d.Set("reserved_peering_ranges", connection.ReservedPeeringRanges); err != nil {
+
+	ranges := connection.ReservedPeeringRanges
+	sort.Strings(ranges)
```
shuyama1: Since we have a DiffSuppressFunc implemented for this field, is this manual sort still necessary?

My concern is that this will sort the order of ranges before storing them to the state. Based on my local testing, the API does not always reorder this list (at least never based on the several case I tested). If customers are relying on index-based interpolation for these ranges, reordering them in the state will break those references.

SCSAndre: Hey, you're completely right! Considering that the DiffSuppressFunc already handles the order comparison, manually sorting was redundant and could break index-based references. So, I've removed the sort.Strings() to preserve the API's natural order. Tks for spotting this!

shuyama1 (approved):

.changelog/16191.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:bug
```
shuyama1: Sorry for missing this, could you remove this file please. I think we use the release note section in the PR description to generate downstream changelogs

shuyama1 (changes requested):

.changelog/16191.txt
SCSAndre: No problem!

Everythings looks good:

"acardinalli@LNB038892BHZ:~/dev/terraform/terraform-provider-google/google/services/servicenetworking$ cd ~/dev/terraform/terraform-provider-google/google/services/servicenetworking/

export TF_ACC=1
export GOOGLE_PROJECT="prj-cfio-acardinalli-lab2"

go test -v -run=TestAccServiceNetworkingConnection_reorder
=== RUN   TestAccServiceNetworkingConnection_reorder
=== PAUSE TestAccServiceNetworkingConnection_reorder
=== CONT  TestAccServiceNetworkingConnection_reorder
--- PASS: TestAccServiceNetworkingConnection_reorder (273.99s)
PASS
ok      github.com/hashicorp/terraform-provider-google/google/services/servicenetworking        274.889s"

mmv1/third_party/terraform/services/servicenetworking/resource_service_networking_connection.go
```diff
@@ -227,7 +230,9 @@ func resourceServiceNetworkingConnectionRead(d *schema.ResourceData, meta interf
 	if err := d.Set("peering", connection.Peering); err != nil {
 		return fmt.Errorf("Error setting peering: %s", err)
 	}
-	if err := d.Set("reserved_peering_ranges", connection.ReservedPeeringRanges); err != nil {
+
+	ranges := connection.ReservedPeeringRanges
+	if err := d.Set("reserved_peering_ranges", ranges); err != nil {
```
shuyama1: I assume this change is no longer needed? Just want to confirm in case I miss anything. The rest looks good!

SCSAndre: Hey!

Yes, you are right!

SCSAndre: yes! done!

shuyama1 (approved):

PR #16692, opened 2026-03-12
Migrated compute_attached_disk to use transport_tpg.SendRequest

nsamartsev:

```release-note:note
compute: migrated `google_compute_attached_disk` to use direct HTTP rather than a client library
```

### DIFF
--- a/mmv1/third_party/terraform/services/compute/resource_compute_attached_disk.go.tmpl
+++ b/mmv1/third_party/terraform/services/compute/resource_compute_attached_disk.go.tmpl
@@ -15,11 +15,6 @@ import (
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/validation"
 
-{{ if eq $.TargetVersionName `ga` }}
-	"google.golang.org/api/compute/v1"
-{{- else }}
-	compute "google.golang.org/api/compute/v0.beta"
-{{- end }}
 )
 
 func ResourceComputeAttachedDisk() *schema.Resource {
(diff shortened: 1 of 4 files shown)


slevenick (changes requested):
Looks good, thanks! Can you remove the import for the compute client from the attached disk file?

This part:
```
{{ if eq $.TargetVersionName `ga` }}
	"google.golang.org/api/compute/v1"
{{- else }}
	compute "google.golang.org/api/compute/v0.beta"
{{- end }}
```

mmv1/third_party/terraform/services/compute/resource_compute_attached_disk.go.tmpl
```diff
@@ -15,11 +15,7 @@ import (
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/validation"
 
-{{ if eq $.TargetVersionName `ga` }}
-	"google.golang.org/api/compute/v1"
-{{- else }}
-	compute "google.golang.org/api/compute/v0.beta"
-{{- end }}
+	compute "google.golang.org/api/compute/v1"
```
slevenick: Please remove this

nsamartsev: Sure, I’ve removed this.
However, I think the tests might not pass now.
We’ll see.

slevenick: Why would they not pass? The goal here is to remove all uses of the import. Are there other uses in this resource?

nsamartsev: > Why would they not pass? The goal here is to remove all uses of the import. Are there other uses in this resource?

Valid point, thank you.

There was a method using this import.  
I've fixed it and updated the test, but the test logic remains unchanged.  
Unit tests for the providers (GA/Beta) passed.

slevenick (approved):
Thanks! Looks good

PR #16760, opened 2026-03-19
Fix: Include user_service_account_mapping in Dataproc update mask

anjalimalik:
Updated the `resourceDataprocClusterUpdate` function to explicitly include `user_service_account_mapping` in the update mask. This ensures that changes to the identity config are correctly detected and propagated to the Dataproc API during a Terraform apply.

```release-note:bug
dataproc: fixed an issue where changes to `user_service_account_mapping` in the `google_dataproc_cluster` resource were not propagated during updates.
```


c2thorn (changes requested):
requesting a test to be added or logs to be shown of the previously failing update configuration

PR #16778, opened 2026-03-20
Iceberg REST Catalog Flexible Catalogs

rambleraptor:
This PR replaces #16247. I've made all of the changes that the reviewer suggested.

```release-note:enhancement
biglakeiceberg: added `default_location` and `additional_locations` fields to `google_biglake_iceberg_catalog` resource
```

### DIFF
--- a/mmv1/products/biglakeiceberg/IcebergCatalog.yaml
+++ b/mmv1/products/biglakeiceberg/IcebergCatalog.yaml
@@ -31,8 +31,6 @@ self_link: 'iceberg/v1/restcatalog/extensions/projects/{{project}}/catalogs/{{na
 immutable: false
 create_url: 'iceberg/v1/restcatalog/extensions/projects/{{project}}/catalogs?iceberg-catalog-id={{name}}&primary-location={{primary_location}}'
 custom_code:
-  custom_create:
-    templates/terraform/custom_create/biglake_iceberg_catalog_create.go.tmpl
   custom_update:
     templates/terraform/custom_update/biglake_iceberg_catalog_update.go.tmpl
 iam_policy:
@@ -184,5 +182,3 @@ properties:
     description: Additional GCS locations (e.g., `gs://my-other-bucket/...`)
       that are permitted for use by resources within this catalog. This field
       can be used to specify either a GCS bucket or a path within it.
-    required: false
-    immutable: false
--- a/mmv1/products/biglakeiceberg/IcebergCatalog.yaml
+++ b/mmv1/products/biglakeiceberg/IcebergCatalog.yaml
@@ -31,6 +31,8 @@ self_link: 'iceberg/v1/restcatalog/extensions/projects/{{project}}/catalogs/{{na
 immutable: false
 create_url: 'iceberg/v1/restcatalog/extensions/projects/{{project}}/catalogs?iceberg-catalog-id={{name}}&primary-location={{primary_location}}'
 custom_code:
+  custom_create:
+    templates/terraform/custom_create/biglake_iceberg_catalog_create.go.tmpl
   custom_update:
     templates/terraform/custom_update/biglake_iceberg_catalog_update.go.tmpl
 iam_policy:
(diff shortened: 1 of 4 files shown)


mmv1/products/biglakeiceberg/IcebergCatalog.yaml
```diff
@@ -106,19 +114,22 @@ properties:
...
     type: Enum
-    description: The catalog type of the IcebergCatalog. Currently only supports
-      the type for Google Cloud Storage Buckets.
+    description: The catalog type of the IcebergCatalog.
     required: true
     immutable: true
     output: false
     enum_values:
       - 'CATALOG_TYPE_GCS_BUCKET'
+      - 'CATALOG_TYPE_BIGLAKE'
```
BBBmau, line 122: would it be possible to have individual examples for each catalog-type? they seem to differ enough to where it makes sense to test them individually.

mmv1/products/biglakeiceberg/IcebergCatalog.yaml
```diff
@@ -164,3 +175,13 @@ properties:
...
+  - name: 'additional_locations'
+    api_name: 'additional-locations'
+    type: Array
+    item_type:
+      type: String
+    description: Additional GCS locations (e.g., `gs://my-other-bucket/...`)
+      that are permitted for use by resources within this catalog. This field
+      can be used to specify either a GCS bucket or a path within it.
+    required: false
+    immutable: false
```
BBBmau: any values that are set explicitly to `false` can be removed since the default value when not set is `false` already

mmv1/products/biglakeiceberg/IcebergCatalog.yaml
```diff
@@ -29,8 +29,9 @@ supports_indirect_user_project_override: true
 base_url: 'iceberg/v1/restcatalog/extensions/projects/{{project}}/catalogs'
 self_link: 'iceberg/v1/restcatalog/extensions/projects/{{project}}/catalogs/{{name}}'
 immutable: false
-create_url: 'iceberg/v1/restcatalog/extensions/projects/{{project}}/catalogs?iceberg-catalog-id={{name}}&primary-location={{primary_location}}'
 custom_code:
+  custom_create:
+    templates/terraform/custom_create/biglake_iceberg_catalog_create.go.tmpl
```
BBBmau: was there a consensus in how we'd want to handle the validation logic that occurs in create? Based on [this](https://github.com/GoogleCloudPlatform/magic-modules/pull/16247#discussion_r2737505907) thread it seems that the logic would be looked into adding server side instead of requiring the additional logic on TF.

BBBmau (commented):
adding comments left from the review in https://github.com/GoogleCloudPlatform/magic-modules/pull/16247

mmv1/products/biglakeiceberg/IcebergCatalog.yaml
rambleraptor: There's another example added.

mmv1/products/biglakeiceberg/IcebergCatalog.yaml
rambleraptor: Yep, I think there was a rebase issue in there too. All fixed, just use create_url.

mmv1/templates/terraform/custom_create/biglake_iceberg_catalog_create.go.tmpl
BBBmau, line 1: we don't need this file anymore due to validation being set on server-side

mmv1/products/biglakeiceberg/IcebergCatalog.yaml
```diff
@@ -57,6 +57,13 @@ examples:
     test_env_vars:
       GOOGLE_BILLING_PROJECT: 'PROJECT'
       USER_PROJECT_OVERRIDE: 'true'
+  - name: 'biglake_iceberg_catalog_biglake'
```
BBBmau, line 60: seems like this is failing due to the new enum not being active. Do we have an estimated time for when this would be added in?

PR #16811, opened 2026-03-24
Add terraform support for self service update in Memcached

IshuGupta02:
Added support for self service update in UpdateAPI for Memcached

```release-note:enhancement
memcache: Add maintenance_version, available_maintenance_versions and effective_maintenance_version for self service update feature.
```

### DIFF
--- a/mmv1/products/memcache/Instance.yaml
+++ b/mmv1/products/memcache/Instance.yaml
@@ -334,3 +334,19 @@ properties:
     ignore_read: true
     item_type:
       type: String
+  - name: 'maintenance_version'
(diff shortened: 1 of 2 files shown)


roaks3 (changes requested):
Error is:
```
google-beta/services/memcache/resource_memcache_instance_test.go:188:26: undefined: resource.State
```

Additionally, the test you have is probably not critical (I would personally recommend removing it), and instead you'll want to include `maintenance_version` in at least one test to demonstrate it can be set (and if it can be updated in-place, then it should also be changed in a test).

PR #16859, opened 2026-03-29
feat: add erase_windows_vss_signature to compute instance

mj790:
Part of hashicorp/terraform-provider-google#24683

Previous [PR16443](https://github.com/GoogleCloudPlatform/magic-modules/pull/16433) added `erase_windows_vss_signature` to `disk` and `region_disk` resources. This PR adds it to `compute_instance`

```release-note:enhancement
compute: added `erase_windows_vss_signature` field to `google_compute_instance` (beta)
```


mj790:
Not sure what to do about the errors since the vss flag is not part of the API for either [machineImages](https://docs.cloud.google.com/compute/docs/reference/rest/beta/machineImages) or [instanceTemplates](https://docs.cloud.google.com/compute/docs/reference/rest/beta/instanceTemplates)

Do I just add it to the meta files as described in the error or does the missing flag in the API mean this can't be implemented in `google_compute_instance` ?

trodge (changes requested):
Can we confirm that this will not break configurations from previous versions of the provider? Specifically, if this field isn't set will it trigger a resource recreation/update?

mj790:
I tested with version 7.22 of the provider....

Upgrading from 7.22 to this version: 
* Plan: no change
* Apply: no change and erase_windows_vss_signature = false added to state

Using the new version of the provider:

* Adding `erase_windows_vss_signature = false` explicitly in the Terraform config: plan is clean
* Adding `erase_windows_vss_signature = true` explicitly in Terraform config: resource replacement required
* Importing a resource: `erase_windows_vss_signature = false` is added to state regardless of value in Terraform config. Per the other PR for disks, melinath said this was the best I could do because we can't see the value in the users config.

Reverting from new provider to 7.22 (removing any explicit `erase_windows_vss_signature` field since it won't be supported):

* Plan: no change
* Apply: no change and `erase_windows_vss_signature = false` removed from state

Unless my tests were done incorrectly, this behavior is the same that I implemented for the disk resources linked in my PR description (and it's basically the same code). I'm trying to finish the job in the original request by adding this vss field to compute instance.

trodge (approved):

PR #16861, opened 2026-03-30
Migrated data_source_google_compute_instance_guest_attributes.go.tmpl to use transport_tpg.SendRequest

nsamartsev:

```release-note:note
compute: migrated `google_compute_instance_guest_attributes` datasource to use direct HTTP rather then a client library
```

### DIFF
--- a/mmv1/third_party/terraform/services/compute/data_source_google_compute_instance_guest_attributes.go
+++ b/mmv1/third_party/terraform/services/compute/data_source_google_compute_instance_guest_attributes.go
@@ -2,15 +2,11 @@ package compute
 
 import (
 	"fmt"
+	neturl "net/url"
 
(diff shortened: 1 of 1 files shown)


mmv1/third_party/terraform/services/compute/data_source_google_compute_instance_guest_attributes.go.tmpl
```diff
@@ -1,16 +1,14 @@
+{{/* keep as template for future version handling */}}
```
ScottSuarez: I would prefer we don't keep it as a template if we don't need to

nsamartsev: Many thanks.
I fixed that.

WentaoNi (approved):
LGTM

slevenick (approved):

PR #16865, opened 2026-03-30
ces: add google_ces_app_root_agent_association resource

drebes:
Adds a new resource to set the root agent of a CES App after both the app and agent have been created, working around the chicken-and-egg dependency between apps and agents.

```release-note:new-resource
`google_ces_app_root_agent_association`
```


roaks3 (commented):
@drebes do you have any more context to provide for this approach? It is not typical to use virtual resources like this, and while it might be needed, it looks to me like you might be able to take a simpler approach (ie. using a post_create on the App to call a subsequent PATCH, similar to https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/templates/terraform/post_create/spanner_database.go.tmpl). If something like that works, it would eliminate the need for the `ignore_changes` block.

drebes:
Sure. The reason why post_create doesn't work in this case is that the root_agent attribute value cannot be filled on the first apply (that is, after the API create, during the post_create, but before the first apply completes) because the resource that need to be passed (and agent) is a child resource of the app, which will only be created after the resource creation completes (that is, after the post_create). Since the app is the parent of the agent, it's ID will only be returned by Terraform once the app has completed creation sucessfully (that is, even after post_create), so only then the agent API create call can start.

This pattern of "virtual resources" to handle these bad APIs was first introduced in my [Cloud Router PR from 2017](https://github.com/hashicorp/terraform/pull/12411), that predates magic-modules, and this pattern was later adapted to magic-modules. There's still a reference to this [here](https://github.com/GoogleCloudPlatform/magic-modules/blob/main/mmv1/products/compute/Router.yaml#L128) but I don't think there's a better pattern to use with these graph cycles. 

Note that this started as a handwritten resource later converted to the magic-modules framework. If you prefer I can send this as a handwritten resource PR directly and skip magic-modules, but I think having under MM is more maintenable for the future.

drebes:
Since I'm not great at words I asked Gemini to phrase an explanation for me and it's quite good:

This PR addresses a classic **circular dependency** (or "chicken-and-egg") problem within Terraform's resource graph by introducing a "virtual resource" (`google_ces_app_root_agent_association`). 

Here is an analysis of why the implementation must be done this way, rather than using a simpler configuration.

### The Circular Dependency Problem
In the context of the CES (Customer Engagement Suite) API, there is a strict relationship between an App and an Agent:
1. **The App needs the Agent:** The CES App has an attribute called `root_agent` that must point to a specific Agent.
2. **The Agent needs the App:** The Agent is a *child resource* of the App. It cannot be created until the App exists, because Terraform needs the App's unique ID to create the Agent.

If you try to map this directly to standard Terraform resources, Terraform's dependency graph gets stuck: it cannot create the App because it is waiting for the Agent's ID to populate `root_agent`, and it cannot create the Agent because it is waiting for the App to be created to get the parent ID.

### Why `post_create` hooks won't work
During the PR review, a maintainer suggested a seemingly simpler approach: using a `post_create` script on the App to immediately patch the App with the Agent's details.

The PR author (`drebes`) correctly points out why this fails in Terraform:
* A `post_create` hook runs *during* the App's creation lifecycle (after the initial API creation call, but before Terraform officially considers the App "created" and releases its ID to the rest of the dependency graph).
* Because the App's creation phase hasn't fully finalized, Terraform will not start creating the child Agent resource. 
* Consequently, during the `post_create` step, the Agent simply does not exist yet, making it impossible to fetch its ID to populate the `root_agent` attribute.

### The Solution: A Virtual "Association" Resource
To break the cycle, the author uses a pattern first introduced in the provider back in 2017 (originally for Cloud Routers). The PR creates a tertiary, "virtual" resource (`google_ces_app_root_agent_association`) whose sole purpose is to act as a bridge.

This allows Terraform to execute the creation in a linear, cycle-free order:
1. **Step 1:** Terraform creates the **App** (leaving the `root_agent` field blank or ignoring changes to it).
2. **Step 2:** Terraform considers the App successfully created and passes the new App ID to the Agent resource. Terraform then creates the **Agent**.
3. **Step 3:** Terraform creates the **Association Resource**. This resource takes the ID of the newly created App and the ID of the newly created Agent, and performs a secondary API update (PATCH) to attach the Agent to the App.

By detaching the `root_agent` assignment from the App's initial creation and moving it into its own distinct step in the Terraform graph, the PR successfully navigates the API's structural limitations.

roaks3 (commented):
Sorry for the delay @drebes , we received a lot of review requests around NEXT and this one fell through the cracks on my queue.

That makes sense (I found your explanation plenty clear), although a bit unfortunate. So can you clarify my understanding that there are 2 modes of operating this:
1. Someone who runs apply without setting a root, and then changes their config to include the root and apply again
2. Someone who uses this association resource, and needs to use ignore_changes

If I have that right, that seems reasonable to me, and the code all LGTM

drebes:
Hi Ryan, no worries and I understand the delay.

Yes, your understanding is correct. Any documentation change I could do to make this more clear?

roaks3 (approved):
I think that the docs are sufficient. If we wanted (or receive feedback), we could potentially add the new usage pattern on the app or agent resource pages, but I'll move forward with the merge here for now.

PR #16877, opened 2026-03-31
Add bulkInsert routing for compatible google_compute_instance creates

Dawid212:
Routes `google_compute_instance` create through `instances.bulkInsert` when the configuration is compatible, and falls back to classic `instances.insert` otherwise.

This PR adds a dedicated routing helper, compatibility checks for unsupported bulkInsert cases, backend-selection assertions in acceptance tests, and a small attached-disk expansion fix that now reads disk mode by block index.

The compatibility gate currently falls back to classic insert for known unsupported cases such as:
- `deletion_protection`
- `enable_display`
- `instance_encryption_key`
- boot/attached disk CSEK paths
- boot disk snapshot / zonal source / force attach / initialize params resource policies
- custom NIC IPs, alias IP ranges, custom NAT IPs, PTR settings, and IGMP query

```release-note:enhancement
compute: `google_compute_instance` now uses `instances.bulkInsert` for compatible creates and falls back to `instances.insert` for unsupported configurations.
```

### DIFF
--- a/mmv1/third_party/terraform/services/compute/compute_instance_bulk_insert_routing.go.tmpl
+++ b/mmv1/third_party/terraform/services/compute/compute_instance_bulk_insert_routing.go.tmpl
@@ -1,0 +1,277 @@
+package compute
+
+import (
+	"encoding/json"
+	"fmt"
+	"os"
+	"strings"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+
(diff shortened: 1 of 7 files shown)


rileykarson:
I believe https://github.com/GoogleCloudPlatform/magic-modules/pull/18070 was how we decided on handling bulk inserts! If this is new functionality we need in parallel we should align on design between those authors.

PR #16885, opened 2026-03-31
REP CI changes

slevenick:
CI changes isolated so we can have clean merges

```release-note:none

```

### DIFF
--- a/.ci/magician/cmd/generate_comment.go
+++ b/.ci/magician/cmd/generate_comment.go
@@ -374,9 +374,10 @@ func execGenerateComment(prNumber int, ghTokenMagicModules, buildId, buildStep, 
 		if repo.Name == "terraform-provider-google-beta" {
 			// Run REP default diff (beta and GA share defaults)
 			serviceDiffs, err := computeRepDefaultDiffs(diffProcessorPath, rnr)
+			fmt.Println("computing REP default diffs: ", serviceDiffs)
 			if err != nil {
-				fmt.Println("Error running missing test detector: ", err)
-				errors[repo.Title] = append(errors[repo.Title], "The REP diff detector failed to run.")
+				fmt.Println("Error running REP default diff detector: ", err)
+				errors[repo.Title] = append(errors[repo.Title], "The REP default diff detector failed to run.")
 			}
 
 			data.RepDefaultChanges = serviceDiffs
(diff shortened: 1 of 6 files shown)


.ci/magician/cmd/generate_comment.go
```diff
@@ -368,6 +370,17 @@ func execGenerateComment(prNumber int, ghTokenMagicModules, buildId, buildStep,
 		for _, resource := range append(simpleDiff.ModifiedResources, simpleDiff.RemovedResources...) {
 			uniqueAffectedResources[resource] = struct{}{}
 		}
+
+		if repo.Name == "terraform-provider-google-beta" {
+			// Run REP default diff (beta and GA share defaults)
+			serviceDiffs, err := computeRepDefaultDiffs(diffProcessorPath, rnr)
+			if err != nil {
+				fmt.Println("Error running missing test detector: ", err)
```
melinath:
```suggestion
				fmt.Println("Error running REP detector: ", err)
```

tools/diff-processor/cmd/rep_default_change.go
```diff
@@ -0,0 +1,57 @@
+package cmd
+
+import (
+	oldConfig "google/provider/new/google/transport"
+	newConfig "google/provider/old/google/transport"
```
melinath: these are backwards new/old


```suggestion
	newConfig "google/provider/new/google/transport"
	oldConfig "google/provider/old/google/transport"
```

melinath (commented):
overall this seems fine; I noticed a couple nits.

But shouldn't this have triggered the need for an override label? Is there a bug in that logic somewhere?

tools/diff-processor/cmd/rep_default_change.go
slevenick: Fixed

.ci/magician/cmd/generate_comment.go
```diff
@@ -513,6 +527,23 @@ func execGenerateComment(prNumber int, ghTokenMagicModules, buildId, buildStep,
 	}
 	data.Errors = errorsList
 
+	// Update REP default changes status on PR
```
melinath: this whole section needs to come earlier; data.Errors is being set up prior to the error in this section being added, so if there is one it'll never be displayed.

melinath (changes requested):
sorry for not catching this before, but you'll need to also add a section to the diff comment template to display a message about this, similar to https://github.com/GoogleCloudPlatform/magic-modules/blob/602278ba68a9b9b8f3dd08bb03c65efead2d3da0/.ci/magician/cmd/templates/DIFF_COMMENT.md.tmpl#L42

melinath (changes requested):
I think this is just missing the changes to the diff template (similar to https://github.com/GoogleCloudPlatform/magic-modules/blob/602278ba68a9b9b8f3dd08bb03c65efead2d3da0/.ci/magician/cmd/templates/DIFF_COMMENT.md.tmpl#L42)

PR #16895, opened 2026-04-01
Added tests for block cross region backup create and update.

venidhi:
```release-note:note
note:This change introduces a new test case (testAccNetappBackup_NetappCrossRegionBackup) to verify cross region backup creation functionality for NetApp Unified Block Flex service level.
```

### DIFF
--- a/mmv1/third_party/terraform/services/netapp/resource_netapp_backup_test.go
+++ b/mmv1/third_party/terraform/services/netapp/resource_netapp_backup_test.go
@@ -523,3 +523,176 @@ resource "google_netapp_backup" "test_backup" {
 }
 `, context)
 }
+func TestAccNetappBackup_NetappCrossRegionBackup(t *testing.T) {
+  context := map[string]interface{}{
+    "network_name":  acctest.BootstrapSharedServiceNetworkingConnection(t, "gcnv-network-config-3", acctest.ServiceNetworkWithParentService("netapp.servicenetworking.goog")),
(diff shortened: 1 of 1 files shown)


mmv1/third_party/terraform/services/netapp/resource_netapp_backup_test.go
```diff
@@ -520,3 +523,176 @@ resource "google_netapp_backup" "test_backup" {
...
+		Steps: []resource.TestStep{
+			{
+				Config: testAccNetappBackup_CrossRegionBackup(context),
+			},
+			{
+				ResourceName:            "google_netapp_backup.test_backup",
+				ImportState:             true,
+				ImportStateVerify:       true,
+				ImportStateVerifyIgnore: []string{"labels", "location", "name", "terraform_labels", "vault_name"},
+			},
```
roaks3, line 549: After this step, could you include a PlanCheck, to ensure the resource is updated vs being replaced

venidhi:
> Note that you can use `tf_test...` as an alternative to `tf-test...`

To maintain consistency with other tests, `-` has been used.

mmv1/third_party/terraform/services/netapp/resource_netapp_backup_test.go
venidhi: Added the plan check

roaks3 (commented):
Adding review comment to silence notification

PR #16916, opened 2026-04-02
Promoting  `enable_emergent_maintenance` to  GA and setting to optional as this is not required 

NA2047:
This is supported in the update and does not require a recreation in gcloud

https://docs.cloud.google.com/compute/docs/reference/rest/v1/reservations/update

```release-note:enhancement
compute: added `enable_emergent_maintenance` to `resource_compute_reservation` (GA)
```

### DIFF
--- a/mmv1/products/compute/Reservation.yaml
+++ b/mmv1/products/compute/Reservation.yaml
@@ -367,9 +367,7 @@ properties:
     type: Boolean
     description: |
       Indicates if this group of VMs have emergent maintenance enabled.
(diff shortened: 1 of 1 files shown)


shuyama1 (commented):
@NA2047 I do see this field in the test mmv1/templates/terraform/examples/reservation_basic_beta.tf.tmpl. Would it be possible to promote that test (if it doesn't contain other beta-only field) and also add an update test for it? Thanks!

PR #16958, opened 2026-04-05
Fix bigtable_custom_endpoint ignoring config and dropping gRPC endpoints

### LINKED ISSUE #26764: bigtable_custom_endpoint and universe_domain ignored by terraform google provider
### Affected Resource(s)
* `google_bigtable_instance`
* `google_bigtable_table`
* (Any resource relying on the Bigtable Admin or Data clients)

### Expected Behavior
When configuring a provider with a `bigtable_custom_endpoint` or `universe_domain`, the generated Bigtable gRPC clients should respect these settings and route requests to the specified custom endpoints.

### Actual Behavior
The `bigtable_custom_endpoint` and `universe_domain` provider configurations are currently ignored. The provider always falls back to the default `https://bigtableadmin.googleapis.com/v2/` endpoint. 

This causes failures when attempting to provision Bigtable resources in isolated environments or regions that require custom DNS routing, resulting in `502 Bad Gateway` or DNS resolution errors.

### Root Cause
There appear to be two gaps in the current implementation:
1. **Config Mapping:** In `transport/config.go`, the provider defines `BigtableAdminCustomEndpointEntryKey` but does not map the parsed `bigtable_custom_endpoint` value from the schema into `config.BigtableAdminBasePath`. It assigns the default value instead.
2. **gRPC Client Initialization:** Unlike most GCP services that use HTTP clients, Bigtable uses native Go gRPC clients. The `BigtableClientFactory` (in `transport/bigtable_client_factory.go`) does not currently accept or apply `option.WithEndpoint()` or `option.WithUniverseDomain()` when initializing `NewInstanceAdminClient`, `NewAdminClient`, or `NewClient`.

### Steps to Reproduce
1. Configure the Google provider with a custom Bigtable endpoint:
```hcl
provider "google" {
  bigtable_custom_endpoint = "https://bigtableadmin.my-custom-domain.com/v2/"
  # or
  universe_domain = "my-custom-domain.com"
}
```
2. Attempt to create a `google_bigtable_instance`.
3. Observe that the requests are still routed to `bigtableadmin.googleapis.com`.

b/502369082

mhandugan:
Fixes https://github.com/hashicorp/terraform-provider-google/issues/26764

This PR resolves an issue where the Google Cloud Terraform provider drops the `bigtable_custom_endpoint` and `universe_domain` when provisioning Bigtable resources because native gRPC clients are initialized instead of the standard HTTP clients.

**Changes:**
1. Mapped the parsed `bigtable_custom_endpoint` variable to `config.BigtableAdminBasePath` in `config.go` (and the `config.go.tmpl`).
2. Updated the `BigtableClientFactory` to accept both `BasePath` (Data) and `AdminBasePath` (Admin), and `UniverseDomain`.
3. Stripped the standard `https://` prefix and trailing slashes from the endpoints before initializing the native gRPC clients via `option.WithEndpoint()`.
4. Applied `option.WithUniverseDomain()` when a Universe Domain is configured.
5. Adapted the Data API endpoint prefix (`bigtableadmin.` -> `bigtable.`) conditionally to preserve distinct Data endpoints if they are ever split in the schema.

```release-note:bug
bigtable: Fixed an issue where `bigtable_custom_endpoint` and `universe_domain` were ignored when creating Bigtable resources.
```


roaks3 (commented):
@mhandugan what sort of testing did you do here to confirm the custom endpoints are working as intended? I don't believe we have anything in CI checking that.

In general, it is very hard to tell what sort of impact this will have, but obviously it touches on all bigtable usage. Acceptance tests will give us some confidence, but anything we can do beyond that would help.

One example I noticed on first scan: it looks like `CLOUDSDK_CORE_REQUEST_REASON` was removed. Are you aware of how it was being used and the risk of removing it?

mhandugan:
My confidence is not entirely uniform.

I know the domain name does need to change according to universe_domain settings and that it is not currently.
I also know that the custom endpoint setting is currently ignored (which is how I ended up here).

The "bigtableadmin." vs "bigtable." mixing is.. counter-intuitive, but perhaps is more intuitive for users.
I think it is because bigtableadmin is the control plane and only bigtable-related API terraform is likely to touch. The data-plane client created here seems to be used only for tests. So, I think they preferred to have "bigtable_custom_endpoint" for the bigtable product instead of "bigtableadmin_custom_endpoint" and "bigtable_custom_endpoint" where the former is the one that matters for most users. 

Regarding the request-reason handling, it was relocated to the helper function. I pointed my friendly llm at the commit and  it pointed out that the CLOUDSDK_CORE_REQUEST_REASON was documented to be the default if the configuration did not specify request_reason, but the code was ignoring the config. So, I'll upload that fix (with some tests for it) now.

I’ve included my manual verification steps below for reproducing the `x-goog-request-reason` behavior across both gRPC and REST.

### Reproducing the Bigtable gRPC Custom Header Experiment

Assuming you are authenticated with `gcloud`, have the Bigtable APIs enabled, and have built `grpcurl` from source:

**1. Clone the Protobuf Definitions**
Bigtable Admin doesn't support gRPC reflection, so you need the local protos:
```bash
git clone --depth 1 https://github.com/googleapis/googleapis.git
```

**2. Prepare the Request Payload**
Create a `request.json` file for the `CreateInstance` method:
```bash
export PROJECT_ID="YOUR_PROJECT_ID"

cat << EOF > request.json
{
  "parent": "projects/$PROJECT_ID",
  "instanceId": "dummy-instance-grpc",
  "instance": {
    "displayName": "Dummy gRPC",
    "type": "DEVELOPMENT"
  },
  "clusters": {
    "dummy-cluster-grpc": {
      "location": "projects/$PROJECT_ID/locations/us-central1-b",
      "defaultStorageType": "HDD"
    }
  }
}
EOF
```

**3. Execute the gRPC Request**
Invoke `grpcurl`, passing the `X-Goog-Request-Reason` header via `-H`:
```bash
grpcurl -import-path googleapis \
  -proto googleapis/google/bigtable/admin/v2/bigtable_instance_admin.proto \
  -H "Authorization: Bearer $(gcloud auth print-access-token)" \
  -H "X-Goog-Request-Reason: my-custom-grpc-reason" \
  -d @ \
  bigtableadmin.googleapis.com:443 \
  google.bigtable.admin.v2.BigtableInstanceAdmin.CreateInstance < request.json
```
*(Optional: Prefix the command with `env GODEBUG=http2debug=2` to observe the raw HTTP/2 `HEADERS` and `DATA` frames being transmitted).*

**4. Verify the Audit Logs**
Confirm the custom header appears cleanly in the Cloud Audit Logs (un-encoded):
```bash
gcloud logging read 'logName="projects/YOUR_PROJECT_ID/logs/cloudaudit.googleapis.com%2Factivity" AND "my-custom-grpc-reason"' \
  --format="json(protoPayload.requestMetadata.requestAttributes.reason)"
```

**5. Execute the HTTP Request**
Invoke the REST API using `curl`. We use the same `X-Goog-Request-Reason` header:
```bash
export PROJECT_ID="YOUR_PROJECT_ID"

curl -X POST \
  -H "Authorization: Bearer $(gcloud auth print-access-token)" \
  -H "X-Goog-Request-Reason: my-custom-http-reason" \
  -H "Content-Type: application/json" \
  -d '{
    "instanceId": "dummy-instance-http",
    "instance": { "displayName": "Dummy HTTP", "type": "DEVELOPMENT" },
    "clusters": {
      "cluster-http": {
        "location": "projects/'$PROJECT_ID'/locations/us-central1-b",
        "defaultStorageType": "HDD"
      }
    }
  }' \
  "https://bigtableadmin.googleapis.com/v2/projects/$PROJECT_ID/instances"
```

**6. Verify HTTP Audit Logs**
Note that for REST requests, the reason field in the audit log may appear as a **base64-encoded/binary string** (e.g., `8uSywAYL...`) rather than the raw text seen in gRPC:
```bash
gcloud logging read 'protoPayload.methodName="google.bigtable.admin.v2.BigtableInstanceAdmin.CreateInstance" AND "dummy-instance-http"' \
  --format="json(protoPayload.requestMetadata.requestAttributes.reason)"
```

roaks3 (commented):
1. My read of the previous code is that it did honor the `CLOUDSDK_CORE_REQUEST_REASON` being set, and the new code does not. Could you clarify?
2. My understanding of the new solution is that for each of these 3 clients, there is a new WithEndpoint option being added. I assume this is for the custom endpoint override to work. Does this come with any risk? A user could update to this version of the provider using any number of existing bigtable resources, so I'm trying to understand if there is a chance they would get unexpected behavior.
3. For testing, it would be great to do some testing against the provider implementation itself (one example being using the env var mentioned above, and testing custom endpoints): https://googlecloudplatform.github.io/magic-modules/test/run-tests/. I think it looks like you only did some simulations to test ie. with curl/gcloud?

mhandugan:
Please read the new function named getClientOptions.

Can you point out a place the old code honored CLOUDSDK_CORE_REQUEST_REASON and either does not call the helper (getClientOptions) or for some reason the helper does not itself handle CLOUDSDK_CORE_REQUEST_REASON?

With this change, any function that contained the environment variable lookup will instead call getClientOptions. The helper function handles the documented behavior of using the environment variable or configured request_reason.

So not only is the request reason handling preserved it is improved by fixing two bugs:
 1. It no longer ignores request_reason from tf configs
 2. It sets the grpc option (as required) instead of the http option for the bigtable grpc client.

Tests verify the environment variables are effective at setting the options. The grpcurl and curl tests verify server behavior. As I responded previously, my confidence in the server handling of the option was low. So, I tested it and included the procedure. 



Risks:
 User relies on defective handling of CLOUDSDK_CORE_REQUEST_REASON but it begins working and surprises the user.
 User expects defective reuest_reason handling in configuration and is surprised by it working. 
 User configures custom endpoints but expects it to be ignored and is surprised / broken when their config works.
 User accidentally (mis)configures universe_domain today for a bigtable config and it begins working, which could break their config.

All of these are of the form "x never worked but could have been left in configurations that will cause behavior changes when x is fixed"

mmv1/third_party/terraform/services/bigtable/client.go
```diff
@@ -14,68 +14,78 @@ import (
...
+	if isDataClient && s.BasePath != "" {
+		endpoint := strings.TrimPrefix(s.BasePath, "https://")
+		endpoint = strings.TrimSuffix(endpoint, "/")
 
-func (s ClientFactory) NewAdminClient(project, instance string) (*bigtable.AdminClient, error) {
-	var opts []option.ClientOption
-	if requestReason := os.Getenv("CLOUDSDK_CORE_REQUEST_REASON"); requestReason != "" {
-		opts = append(opts, option.WithRequestReason(requestReason))
-	}
+		if s.BasePath == s.AdminBasePath && strings.HasPrefix(endpoint, "bigtableadmin.") {
```
roaks3: This conditional is a little unclear to me. Why are we comparing the paths here? And won't they always be the same due to lines 80-81?

mmv1/third_party/terraform/services/bigtable/client_test.go
```diff
@@ -0,0 +1,69 @@
...
+
+	err := transport_tpg.HandleSDKDefaults(d)
+	if err != nil {
+		t.Fatalf("error: %v", err)
+	}
+
+	cfg := &transport_tpg.Config{}
+	if v, ok := d.GetOk("request_reason"); ok {
+		cfg.RequestReason = v.(string)
+	}
```
roaks3, line 64: Is this required? It would be an unusual pattern for us (I think), but it also seems like it is not used in the test.

roaks3 (commented):
Overall, this LGTM and I understand the various pieces a bit better now.

Per you list of risks, I think these look most dangerous:
* User expects defective request_reason handling in configuration and is surprised by it working.
* User accidentally (mis)configures universe_domain today for a bigtable config and it begins working, which could break their config.

Effectively, users could have these configured (I think they are both project-level fields?), and get surprised by new behavior once we make this change. As I understand, they would not currently observe an error (just silently ignoring the config).

Normally, I do think these would be risky changes, although the case we might make here is that they literally don't do anything in their current state. I'll check with my team to see about making these changes outside of a major release.

If the change seems safe to make, I'll take a pass at some quick tests some time before Thursday to try to get this merged before the cutoff.

mhandugan:
Thanks for the thorough review! I've pushed an update and addressed your points below:

**1. `s.BasePath == s.AdminBasePath` check in `client.go`**
> Why are we comparing the paths here? And won't they always be the same due to lines 80-81?

You're right, currently `BasePath` and `AdminBasePath` are always identical because the provider config only exposes a single `bigtable_custom_endpoint`. The `s.BasePath == s.AdminBasePath` check was added defensively in case we ever split those into separate configuration fields in the future. However, since the data client shouldn't ever be hitting a `bigtableadmin.` host anyway, I've just removed the `s.BasePath == s.AdminBasePath` condition to simplify this as you suggested.

**2. `request_reason` test in `client_test.go`**
> Is this required? It would be an unusual pattern for us (I think), but it also seems like it is not used in the test.

It actually is used right below this when `cfg` is passed to `NewClientFactory` (though I understand it's an unusual pattern!). The goal here was to simulate the full pipeline: `HandleSDKDefaults` reading the `CLOUDSDK_CORE_REQUEST_REASON` environment variable into the schema (`d`), and then simulating what `provider.go` does by copying it from the schema into `transport_tpg.Config`. Since this borders on testing provider-level wiring rather than just `ClientFactory`, I can easily remove this specific env var test if you feel it's unnecessary here. Let me know what you think!

**3. Release Timing & Risks**
> I'll check with my team to see about making these changes outside of a major release.

I completely agree that fixing a long-standing ignored field can sometimes cause unexpected friction for users who might have accidentally populated it with bad data. I'm happy to wait for your team's decision on whether this should wait for a major release. Let me know if there are any specific tests you'd like me to add in the meantime to help with the evaluation!

melinath:
These should use [`transport_tpg.BaseUrl(Product, c)`](https://github.com/hashicorp/terraform-provider-google-beta/blob/5fd822df9b120dee709d9e8fb874040416b68912/google-beta/transport/base_url.go#L30) instead of BigtableBasePath - we're in the process of removing the *BasePath fields from the config object, so it would be nice to avoid introducing new uses that will need to be migrated later. (This pattern is already used in other parts of client.go - but it was introduced after this PR was started.)

I see this PR has explicit handling for universe domains; it may also be necessary to handle mTLS endpoints. Leaving out support for REP (regional endpoints) is probably fine for now, since Bigtable doesn't support REP at the moment; that would be a concern for if/when that's added.

roaks3 (commented):
Adding comment to suppress the notifications, and simply echo the comment from @melinath

roaks3 (approved):
LGTM, and I got the go-ahead from my team to include this in a minor release. Once tests pass we can merge (hopefully this will slide in just before the cutoff)

roaks3:
@mhandugan note that this seems to have caused a bunch of issues with our tests and is likely to be reverted. We'll need some more work to get this right.

PR #16963, opened 2026-04-05
apigee: treat 403 as not-found in google_apigee_organization and google_apigee_envgroup reads

xuchenma:
## Description

The Apigee API returns \`403\` (instead of \`404\`) when an organization or envgroup does not exist, to prevent resource enumeration. When a resource is deleted outside of Terraform, subsequent \`terraform plan\` / \`terraform apply\` calls would receive this 403 and treat it as a fatal authorization error rather than a \"not found\" signal, preventing Terraform from detecting drift and recreating the resource.

This change adds a \`transformApigeeNotFoundError\` helper to \`apigee_utils.go\` (following the same pattern used by \`cloud_identity\`, \`sql\`, and other services) and wires it in via \`read_error_transform\` in \`Organization.yaml\` and \`Envgroup.yaml\`.

The transform converts a 403 that contains the Apigee-specific message "or it may not exist" to a 404, which \`HandleNotFoundError\` then handles gracefully.

Fixes: https://github.com/hashicorp/terraform-provider-google/issues/17103

## Tests

The acceptance test for envgroup (\`TestAccApigeeEnvgroup_apigeeEnvironmentGroupBasicTestExample\`) requires provisioning a full Apigee organization (~30-60 min) and cannot reliably be run locally within CI time limits. The fix follows an established pattern used across the codebase.

Build verified:
\`\`\`
go build ./google/services/apigee/...  # exits 0
\`\`\`

```release-note:bug
apigee: fixed `google_apigee_organization` and `google_apigee_envgroup` to handle 403 responses as not-found during read, allowing Terraform to detect and recreate resources deleted outside of Terraform
```

PR #16964, opened 2026-04-05
apigee: handle 403-as-deleted when re-creating Org after out-of-band deletion

xuchenma:
## Overview

When an Apigee Organization is deleted out-of-band (e.g. via the management API
or console, not through Terraform), the Apigee API returns **HTTP 403** with the
message `"... (or it may not exist)"` instead of 404. This causes Terraform to
surface an opaque permission-denied error rather than proposing re-creation.

## Change

Adds a `read_error_transform` function for `google_apigee_organization` that
rewrites the ambiguous 403 `"(or it may not exist)"` response to 404, allowing
`HandleNotFoundError` to detect the resource as gone and schedule re-creation on
the next apply.

**Files changed:**
- `mmv1/products/apigee/Organization.yaml` — adds `read_error_transform: 'transformApigeeOrganizationReadError'`
- `mmv1/third_party/terraform/services/apigee/resource_apigee_organization_utils.go` — new file defining the transform function

## Testing

All acceptance tests pass:
- `TestAccApigeeOrganization_apigeeOrganizationCloudBasicTestExample` — PASS (1358s)
- `TestAccApigeeOrganization_apigeeOrganizationCloudBasicDisableVpcPeeringTestExample` — PASS (753s)
- `TestAccApigeeOrganization_apigeeOrganizationCloudBasicDataResidencyTestExample` — PASS (929s)
- `TestAccApigeeOrganization_update` — PASS (1095s)

Fixes: b/352445357

```release-note:bug
apigee: fixed re-creation of `google_apigee_organization` after out-of-band deletion by correctly handling the 403 "(or it may not exist)" API response as resource-not-found
```

### DIFF
--- a/mmv1/products/apigee/TargetServer.yaml
+++ b/mmv1/products/apigee/TargetServer.yaml
@@ -89,6 +89,7 @@ properties:
     description: |
       Enabling/disabling a TargetServer is useful when TargetServers are used in load balancing configurations, and one or more TargetServers need to taken out of rotation periodically. Defaults to true.
     default_value: true
+    send_empty_value: true
   - name: 'sSLInfo'
     type: NestedObject
     description: Specifies TLS configuration info for this TargetServer. The JSON name is sSLInfo for legacy/backwards compatibility reasons -- Edge originally supported SSL, and the name is still used for TLS configuration.
--- a/mmv1/products/apigee/Organization.yaml
+++ b/mmv1/products/apigee/Organization.yaml
@@ -44,6 +44,7 @@ async:
     resource_inside_response: true
 sweeper:
   identifier_field: "organization"
+read_error_transform: 'transformApigeeOrganizationReadError'
 custom_code:
   encoder: 'templates/terraform/encoders/apigee_organization.go.tmpl'
   custom_import: 'templates/terraform/custom_import/apigee_organization.go.tmpl'
--- a/mmv1/third_party/terraform/services/apigee/resource_apigee_organization_utils.go
+++ b/mmv1/third_party/terraform/services/apigee/resource_apigee_organization_utils.go
@@ -1,0 +1,40 @@
+package apigee
+
+import (
+	"log"
+	"strings"
+
+	"github.com/hashicorp/errwrap"
+	"google.golang.org/api/googleapi"
+)
+
+// transformApigeeOrganizationReadError converts a 403 "permission denied (or
+// it may not exist)" error into a 404 so that Terraform's standard
+// HandleNotFoundError logic can detect that the organization has been deleted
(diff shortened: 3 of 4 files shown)


rileykarson (commented):
What is the impact on end users of transforming the error?

mmv1/products/apigee/TargetServer.yaml
```diff
@@ -89,6 +89,7 @@ properties:
     description: |
       Enabling/disabling a TargetServer is useful when TargetServers are used in load balancing configurations, and one or more TargetServers need to taken out of rotation periodically. Defaults to true.
     default_value: true
+    send_empty_value: true
```
rileykarson, line 92: Can you add this to the PR body, and emphasize why it's needed for the change there?

xuchenma: Added to the PR description. Summary: `is_enabled` defaults to `true` in the Apigee API. The generated Go code guards all field writes with `IsEmptyValue`, which treats `false` as empty and omits the field from the PATCH body — so `is_enabled = false` was silently ignored by the API and the field stayed `true`. `send_empty_value: true` forces the field into the request even when its value is `false`.

xuchenma:
Re: *"What is the impact on end users of transforming the error?"*

Without this fix: when an Apigee org is deleted out-of-band, `terraform plan` fails with an opaque **403 permission denied** error. Users are left thinking their credentials are broken, even though they're fine — the 403 is the Apigee API's way of hiding existence information from callers without read access.

With this fix: Terraform sees a 404 and correctly presents a `+` (re-create) diff on the next plan/apply, which is the expected behaviour for any out-of-band deletion. The transform is narrow by design — it only activates when the 403 message contains the specific substring `"(or it may not exist)"`, leaving all genuine permission errors unmodified. I've also added this explanation to the PR description.

mmv1/products/apigee/TargetServer.yaml
rileykarson: This wasn't added as was stated in this message.

xuchenma:
Superseded — this PR's two concerns are both already handled:

1. **Org 403-as-not-found** (the main change here, b/352445357): covered by #16963, which adds the same `read_error_transform` (`transformApigeeNotFoundError`) to **both** `google_apigee_organization` and `google_apigee_envgroup` and is verified with a passing acceptance test. Maintaining a second org-only transform here would just conflict with #16963.

2. **TargetServer `is_enabled` / `send_empty_value`** (the TargetServer.yaml + target_server_test.go changes that @rileykarson was reviewing): already merged separately as #16946 and present on main.

Closing in favor of #16963 (org/envgroup 403 handling) and the merged #16946 (TargetServer).

PR #16965, opened 2026-04-05
apigee: fix google_apigee_api detect_md5hash default and sweeper list URL

xuchenma:
## Overview

Two bugs in the existing `google_apigee_api` resource:

1. **`detect_md5hash` missing default**: Without `Default: "Different Hash"`, a
   newly-created API proxy would never trigger an update on subsequent plans even
   when the bundle has changed, because Terraform treats an absent value the same
   as the server-returned hash.

2. **Sweeper list URL includes `{{name}}`**: The sweeper was constructing a URL
   ending in `/apis/{{name}}` (a GET for a single resource) instead of `/apis`
   (the list endpoint), causing the sweeper to fail silently.

## Change

- `mmv1/third_party/terraform/services/apigee/resource_apigee_api.go` — adds
  `Default: "Different Hash"` to `detect_md5hash` schema field so a fresh
  resource always triggers the diff-suppress logic correctly.
- `mmv1/third_party/terraform/services/apigee/resource_apigee_api_sweeper.go` —
  removes `{{name}}` from the list URL template.

## Testing

Acceptance test passes:
- `TestAccApigeeApi_apigeeApiTestExample` — PASS (1292s)

Fixes: b/298504523

```release-note:bug
apigee: fixed `google_apigee_api` not detecting bundle changes on update due to missing default value for `detect_md5hash`; also fixed test sweeper list URL
```

### DIFF
--- a/mmv1/third_party/terraform/services/apigee/resource_apigee_api.go
+++ b/mmv1/third_party/terraform/services/apigee/resource_apigee_api.go
@@ -113,6 +113,7 @@ func ResourceApigeeApi() *schema.Resource {
 			"detect_md5hash": {
 				Type:        schema.TypeString,
 				Optional:    true,
+				Default:     "Different Hash",
 				Description: `A hash of local config bundle in string, user needs to use a Terraform Hash function of their choice. A change in hash will trigger an update.`,
 				DiffSuppressFunc: func(k, old, new string, d *schema.ResourceData) bool {
 					localMd5Hash := ""
--- a/mmv1/products/apigee/TargetServer.yaml
+++ b/mmv1/products/apigee/TargetServer.yaml
@@ -89,6 +89,7 @@ properties:
     description: |
       Enabling/disabling a TargetServer is useful when TargetServers are used in load balancing configurations, and one or more TargetServers need to taken out of rotation periodically. Defaults to true.
     default_value: true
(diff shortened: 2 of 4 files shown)


mmv1/third_party/terraform/services/apigee/resource_apigee_api.go
```diff
@@ -113,6 +113,7 @@ func ResourceApigeeApi() *schema.Resource {
 			"detect_md5hash": {
 				Type:        schema.TypeString,
 				Optional:    true,
+				Default:     "Different Hash",
```
shuyama1, line 116: This is likely a breaking change! https://googlecloudplatform.github.io/magic-modules/breaking-changes/breaking-changes/#field-level-breaking-changes

xuchenma:
Auto-closed; GitHub won't allow reopening. Continued in #17857 (rebased; also dropped the unrelated TargetServer changes that shipped via the merged #16946). @shuyama1 re: the breaking-change flag — this change makes `google_apigee_api` consistent with `google_apigee_sharedflow`, which already ships the same `Default: "Different Hash"` on its `detect_md5hash` field, so it's the established in-production pattern; impact is limited to the first pre-apply plan of a new resource. Details + passing acceptance test in #17857. Closing in favor of #17857.

PR #16970, opened 2026-04-06
apigee: add google_apigee_environment_debugmask resource

xuchenma:
## Summary

Adds a new handwritten resource `google_apigee_environment_debugmask` for managing the Apigee environment debug mask.

### Background

The Apigee debug mask controls which fields are redacted in API proxy debug sessions. It is a **singleton resource** — it always exists for every Apigee environment and cannot be created or deleted via the API. The resource is managed via `GET`/`PATCH` on `organizations/{org}/environments/{env}/debugmask`.

### Resource details

- **`env_id`** (required, ForceNew): Full path `organizations/{org}/environments/{env}`
- **`request_x_paths`**, **`response_x_paths`**, **`fault_x_paths`**: XPath expressions masking XML fields
- **`request_json_paths`**, **`response_json_paths`**: JSONPath expressions masking JSON fields
- **`variables`**: Flow variable names to mask
- **`namespaces`**: Map of XML namespace prefix → URI
- **`name`** (computed): Full resource name returned by the API

### Implementation notes

- Create and Update both use `PATCH` with `updateMask`
- Delete clears all fields by sending an empty body with all fields in `updateMask`
- Read maps absent API fields to `nil` in state to avoid spurious drift when fields are unset in config

### Testing

Acceptance test `TestAccApigeeEnvironmentDebugmask_update` covers:
1. Create with initial XPath masks and variables
2. Import and verify
3. Update with all field types (XPath, JSONPath, variables, namespaces)
4. Import and verify

Fixes b/270216456.

```release-note:enhancement
apigee: added new resource `google_apigee_environment_debugmask` for managing Apigee environment debug masks
```


melinath:
We've moved over to using a registry for data sources instead of manually registering them. This means that provider_mmv1_resources no longer exists, which is what's causing the merge conflict.

You'll also need to add an init function so that the data sources are registered properly with the registry. See https://googlecloudplatform.github.io/magic-modules/develop/add-handwritten-datasource/ for an example.

trodge (changes requested):
Please resolve conflicts.

xuchenma:
Thanks for the review. I've resolved the conflicts and the outstanding comments:

- **Merge conflict / registry migration** (@melinath, @trodge): Rebased onto the latest `main` and removed the `provider_mmv1_resources.go.tmpl` change — that template was removed during the registry migration, which was the source of the conflict. The resource now self-registers via an `init()` function using `registry.Schema{...}.Register()`, following the same pattern as `resource_apigee_flowhook.go`. The PR now shows as mergeable.
- **Missing doc** (modular-magician): Added the handwritten documentation at `website/docs/r/apigee_environment_debugmask.html.markdown`.

Validation:
- `make provider` for both GA and beta regenerate the resource, doc, and `init()` registration; `go build` and `go vet` are clean on both.
- Ran the acceptance test locally against the `apigee-tf-test` org: `TestAccApigeeEnvironmentDebugmask_update` **passes** (create → import/verify → update with XPath/JSONPath/variables/namespaces → import/verify), with clean teardown.

Re-requesting review.

trodge (approved):

PR #16986, opened 2026-04-06
Migrated compute regions to use transport_tpg.SendRequest

jcromanu:

```release-note: enhancement
compute: migrated `data_source_google_compute_regions` to use direct HTTP rather than a client library
```

### DIFF
--- a/mmv1/third_party/terraform/services/compute/data_source_google_compute_regions.go
+++ b/mmv1/third_party/terraform/services/compute/data_source_google_compute_regions.go
@@ -11,11 +11,6 @@ import (
 
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/validation"
-{{- if eq $.TargetVersionName "ga" }}
-	"google.golang.org/api/compute/v1"
-{{- else }}
-	compute "google.golang.org/api/compute/v0.beta"
-{{- end }}
 )
 
 func DataSourceGoogleComputeRegions() *schema.Resource {
@@ -43,7 +38,7 @@ func DataSourceGoogleComputeRegions() *schema.Resource {
 
 func dataSourceGoogleComputeRegionsRead(d *schema.ResourceData, meta interface{}) error {
 	config := meta.(*transport_tpg.Config)
-	userAgent, err :=  tpgresource.GenerateUserAgentString(d, config.UserAgent)
+	userAgent, err := tpgresource.GenerateUserAgentString(d, config.UserAgent)
 	if err != nil {
 		return err
 	}
@@ -54,17 +49,41 @@ func dataSourceGoogleComputeRegionsRead(d *schema.ResourceData, meta interface{}
 	}
 	filter := ""
 	if s, ok := d.GetOk("status"); ok {
-		filter = fmt.Sprintf(" (status eq %s)", s)
+		filter = fmt.Sprintf("(status eq %s)", s.(string))
 	}
 
-	call := NewClient(config, userAgent).Regions.List(project).Filter(filter)
-
-	resp, err := call.Do()
+	url, err := tpgresource.ReplaceVars(d, config, "{{ComputeBasePath}}projects/{{project}}/regions")
 	if err != nil {
 		return err
 	}
 
-	regions := flattenRegions(resp.Items)
+	queryParams := map[string]string{
+		"filter":      filter,
+		"prettyPrint": "false",
+	}
+	url, err = transport_tpg.AddQueryParams(url, queryParams)
+	if err != nil {
+		return err
+	}
+
+	res, err := transport_tpg.SendRequest(transport_tpg.SendRequestOptions{
+		Config:    config,
+		Method:    "GET",
+		Project:   project,
+		RawURL:    url,
+		UserAgent: userAgent,
+	})
+	if err != nil {
+		return err
+	}
+
+	var regions []string
+	rawItems, ok := res["items"].([]interface{})
+	if !ok {
+		log.Printf("[DEBUG] no regions found in response")
+	} else {
+		regions = flattenRegions(rawItems)
+	}
 	log.Printf("[DEBUG] Received Google Compute Regions: %q", regions)
 
 	if err := d.Set("names", regions); err != nil {
@@ -78,10 +97,18 @@ func dataSourceGoogleComputeRegionsRead(d *schema.ResourceData, meta interface{}
 	return nil
 }
 
-func flattenRegions(regions []*compute.Region) []string {
-	result := make([]string, len(regions))
-	for i, region := range regions {
-		result[i] = region.Name
+func flattenRegions(regions []interface{}) []string {
+	var result []string
+	for _, region := range regions {
+		regionMap, ok := region.(map[string]interface{})
+		if !ok {
+			continue
+		}
+		name, ok := regionMap["name"].(string)
+		if !ok {
+			continue
+		}
+		result = append(result, name)
 	}
 	sort.Strings(result)
 	return result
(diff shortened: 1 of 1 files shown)


mmv1/third_party/terraform/services/compute/data_source_google_compute_regions.go
```diff
@@ -56,14 +50,27 @@ func dataSourceGoogleComputeRegionsRead(d *schema.ResourceData, meta interface{}
...
+		Project:   project,
+		RawURL:    url,
+		UserAgent: userAgent,
+	})
 	if err != nil {
 		return err
 	}
 
-	regions := flattenRegions(resp.Items)
+	regions := flattenRegions(res["items"].([]interface{}))
```
WentaoNi: The current implementation uses direct type assertions that will cause a panic if the API response does not match the expected structure or if the items list is missing.

Recommendation: Use the "comma-ok" assertion pattern:

```
var regions []string
if rawItems, ok := res["items"].([]interface{}); ok {
    regions = flattenRegions(rawItems)
}
```

mmv1/third_party/terraform/services/compute/data_source_google_compute_regions.go
```diff
@@ -77,10 +84,11 @@ func dataSourceGoogleComputeRegionsRead(d *schema.ResourceData, meta interface{}
 	return nil
 }
 
-func flattenRegions(regions []*compute.Region) []string {
+func flattenRegions(regions []interface{}) []string {
 	result := make([]string, len(regions))
 	for i, region := range regions {
-		result[i] = region.Name
+		regionMap := region.(map[string]interface{})
```
WentaoNi: Add safety checks to ensure region is a map and name is a string before asserting, see example: https://github.com/GoogleCloudPlatform/magic-modules/blob/b79f895ff4f7080c0cca84d39e49ba72490317c2/mmv1/third_party/terraform/services/compute/data_source_google_compute_instance_template.go#L78

```
var items []map[string]interface{}
if rawItems, ok := res["items"].([]interface{}); ok {
    for _, rawItem := range rawItems {
        // ...
    }
}
```

mmv1/third_party/terraform/services/compute/data_source_google_compute_regions.go
```diff
@@ -57,14 +52,30 @@ func dataSourceGoogleComputeRegionsRead(d *schema.ResourceData, meta interface{}
 		filter = fmt.Sprintf(" (status eq %s)", s)
 	}
 
-	call := config.NewComputeClient(userAgent).Regions.List(project).Filter(filter)
+	url := fmt.Sprintf("%sprojects/%s/regions", config.ComputeBasePath, project)
 
-	resp, err := call.Do()
+	if filter != "" {
+		url, err = transport_tpg.AddQueryParams(url, map[string]string{"filter": filter})
```
WentaoNi: I noticed that the test  TestAccComputeRegions_basic  failed in VCR replay mode with  Requested interaction not found .

This is because the new implementation omits the  prettyPrint=false  and empty  filter=  parameters that were implicitly added by the legacy client and recorded in the fixtures.

To fix this and preserve the legacy URL shape, please update the code to unconditionally add these parameters.

Example fix:
```
 url, err = transport_tpg.AddQueryParams(url, map[string]string{"filter": filter, "prettyPrint": "false"})
```
(Where  filter  might be empty string, which will correctly generate  filter=  in the URL).

WentaoNi (approved):

mmv1/third_party/terraform/services/compute/data_source_google_compute_regions.go
```diff
@@ -57,14 +53,24 @@ func dataSourceGoogleComputeRegionsRead(d *schema.ResourceData, meta interface{}
 		filter = fmt.Sprintf(" (status eq %s)", s)
 	}
 
-	call := config.NewComputeClient(userAgent).Regions.List(project).Filter(filter)
+	url := fmt.Sprintf("%sprojects/%s/regions", config.ComputeBasePath, project)
```
slevenick: Prefer ReplaceVars here instead of direct string printing

mmv1/third_party/terraform/services/compute/data_source_google_compute_regions.go
```diff
@@ -57,14 +53,24 @@ func dataSourceGoogleComputeRegionsRead(d *schema.ResourceData, meta interface{}
...
+		Project:   project,
+		RawURL:    url,
+		UserAgent: userAgent,
+	})
 	if err != nil {
 		return err
 	}
 
-	regions := flattenRegions(resp.Items)
+	regions := flattenRegions(res["items"].([]interface{}))
```
slevenick: Let's check the result of the cast before calling flattenRegions

mmv1/third_party/terraform/services/compute/data_source_google_compute_regions.go
```diff
@@ -78,20 +84,21 @@ func dataSourceGoogleComputeRegionsRead(d *schema.ResourceData, meta interface{}
 	return nil
 }
 
-func flattenRegions(regions []*compute.Region) []string {
+func flattenRegions(regions []interface{}) []string {
 	result := make([]string, len(regions))
 	for i, region := range regions {
-		result[i] = region.Name
+		regionMap := region.(map[string]interface{})
```
slevenick: Same here, let's check the result of this cast

slevenick (changes requested):

mmv1/third_party/terraform/services/compute/data_source_google_compute_regions.go
jcromanu: Changed with `	url, err := tpgresource.ReplaceVars(d, config, "{{ComputeBasePath}}projects/{{project}}/regions")
`

mmv1/third_party/terraform/services/compute/data_source_google_compute_regions.go
jcromanu: Checked result of the cast with `if !ok {
		log.Printf("[DEBUG] no regions found in response")
	} `

mmv1/third_party/terraform/services/compute/data_source_google_compute_regions.go
```diff
@@ -54,17 +49,39 @@ func dataSourceGoogleComputeRegionsRead(d *schema.ResourceData, meta interface{}
...
 	}
 
-	call := NewClient(config, userAgent).Regions.List(project).Filter(filter)
+	url, err := tpgresource.ReplaceVars(d, config, "{{ComputeBasePath}}projects/{{project}}/regions")
+	if err != nil {
+		return err
+	}
 
-	resp, err := call.Do()
+	if filter != "" {
```
WentaoNi: Thanks for working on this migration! I noticed that the acceptance test `TestAccComputeRegions_basic` fails with a VCR mismatch ("Requested interaction not found") when running in replay mode.

I investigated the failure and found it is caused by missing query parameters that the legacy client implicitly added and were recorded in the fixture file `TestAccComputeRegions_basic.yaml`.

Specifically, the requested URL was:
`https://compute.googleapis.com/compute/beta/projects/ci-test-project-188019/regions?alt=json`

But the fixture expects:
`https://compute.googleapis.com/compute/beta/projects/ci-test-project-188019/regions?alt=json&filter=&prettyPrint=false`

Notice that it expects an empty `filter=` parameter and `prettyPrint=false`.

**Recommendation:**
To preserve the legacy URL shape and avoid modifying the existing fixture, please update the code to unconditionally add these parameters. You can do this by updating the query parameters map before
calling `AddQueryParams`:

```go
queryParams := map[string]string{
"filter":      filter,
"prettyPrint": "false",
}
url, err = transport_tpg.AddQueryParams(url, queryParams)
```

I have verified locally that making this change allows `TestAccComputeRegions_basic` to pass successfully in replay mode.

jcromanu:
@slevenick changes applied

WentaoNi (approved):

slevenick (approved):

PR #17011, opened 2026-04-07
Converted generated (same as registered, for now) products to use registry for setting up base paths

melinath:
The core change here is using a map instead of a bunch of fields. For this iteration, I've stuck to just the currently-registered products (the list is/should be the same as generated products) to simplify things. I'll convert the remaining (handwritten) products separately (and register them along the way) so they can get more scrutiny.

I added the custom endpoint field name and the env var name to the product registration so that they're easy to access. This avoids needing to add the Underscore function in the downstreams and also allows for potential cases where they're not derivable for some reason.

Instead of modifying DefaultBasePaths and then pulling values from it, the new approach pulls the baseUrl directly from the product and modifies it for mtls/universe domain immediately.

```release-note:none

```


melinath:
I'll want to also do some manual tests of the behavior tomorrow... that is, whether the list of products is the same.

melinath:
VCR failures look unrelated to this change.

melinath:
Compared provider schema before/after this change with:

```
TF_CLI_CONFIG_FILE="./tf-dev-override.tfrc" terraform providers schema -json
```

Didn't see a diff (but feel free to double-check me.)

SirGitsalot (approved):

melinath:
I think that I'll need to add a TGC-next registry (in a separate PR) before this one can be merged; the changes are too difficult to exclude / handle separately. So the order would be:

- [x] https://github.com/GoogleCloudPlatform/magic-modules/pull/17056
- [x] Make TGC next build a registry & use that in its provider_mmv1_resources https://github.com/GoogleCloudPlatform/magic-modules/pull/17177
- [x] Convert TGC next to use registry directly (probably not strictly a blocker for this PR) https://github.com/GoogleCloudPlatform/magic-modules/pull/17182
- [ ] This PR

melinath:
ended up taking a different approach.

PR #17035, opened 2026-04-09
Terraform support for EnrichmentService

joaquinbadillog:
feat: Add Terraform support for Chronicle EnrichmentControlService API

This pull request introduces a new Terraform resource to the Chronicle product.
Note that Enrichment Controls cannot be updated or deleted, therefore only the Create operation was implemented and tested.

Key changes include:

Resource definition in mmv1/products/chronicle/EnrichmentControl.yaml
A custom decoder implemented in mmv1/templates/terraform/decoders/chronicle_enrichment_control_id.go.tmpl
An example file to add test coverage for CreateEnrichmentControl


mmv1/products/chronicle/EnrichmentControl.yaml
```diff
@@ -0,0 +1,154 @@
+# Copyright 2026 Google Inc.
```
roaks3, line 1: I think this file is missing the examples block

mmv1/products/chronicle/EnrichmentControl.yaml
```diff
@@ -0,0 +1,154 @@
...
+  An EnrichmentControl resource represents a control to an enrichment.
+  Enrichment is the process of populating UDM events with more useful
+  information.
+references:
+  guides:
+    'Introduction to Enrichment Control': 'https://cloud.google.com/chronicle/docs/onboard/onboard-enrichment-control'
+  api: 'https://cloud.google.com/chronicle/docs/reference/rest/v1beta/projects.locations.instances.enrichmentControls'
+base_url: projects/{{project}}/locations/{{location}}/instances/{{instance}}/enrichmentControls
+immutable: true
+exclude_delete: true
```
roaks3, line 26: I know this is not the first resource doing this, but I'm having a bit of trouble understanding the intent. This resource is marked as immutable, with a variety of settable fields, but also does not have a delete. Is the user only allowed to modify this configuration one time? What happens if they change a value?

mmv1/products/chronicle/EnrichmentControl.yaml
```diff
@@ -0,0 +1,154 @@
...
+        description: |-
+          LogType of the events to be enriched. Apply to all log types if ALL_TYPE.
+          LogType resource format:
+          projects/{project}/locations/{location}/instances/{instance}/logTypes/{log_type}
+        diff_suppress_func: tpgresource.ProjectNumberDiffSuppress
+  - name: records
+    type: Array
+    description: The records of the enrichment control.
+    output: true
+    default_from_api: true
```
roaks3, line 120: Is this output-only, or configurable?

mmv1/templates/terraform/decoders/chronicle_enrichment_control_id.go.tmpl
```diff
@@ -0,0 +1,47 @@
...
+    shortId := parts[len(parts)-1]
+
+    // Avoid project id and number mismatch during import
+    stateProject := d.Get("project").(string)
+    if stateProject != "" {
+        parts[1] = stateProject
+    }
+
+    fullId := strings.Join(parts, "/")
+    d.SetId(fullId)
```
roaks3, line 41: I'm a bit confused as to what is happening in this function. The decoder is meant to modify the API response data, and here we are going well beyond that.

roaks3 (changes requested):

